Issue #869, opened 2017-11-26
South Australia Battery


corradio:
Apparently there's data here:
https://ausrealtimefueltype.global-roam.com/expanded

- [x] Add UI
- [x] Add data
- [ ] Add LCA GHG parameters

corradio:
@systemcatch take a look at this article:
https://electrek.co/2017/11/30/tesla-100mw-battery-flexes-muscles-early-this-morning-delivers-70mw-of-stored-wind-energy-shows-off-fast-switching/?lipi=urn%3Ali%3Apage%3Ad_flagship3_feed%3B0Dkt70O6RUe9LRMLgFNNHw%3D%3D
I think the data source mentioned there might be better. For now I see only little charging..

systemcatch:
@corradio You think the current data source is wrong?  From what I can see the source in the tweet shows only very slow charging as well.

NEMO does not have it yet.

corradio:
Indeed. Let's keep an eye on this. At some point it was charging with 30MW - just want to make sure our data source is right.

systemcatch:
@corradio Reading the nemlog api I think this is a better data source and we should switch the parser to use it.  Doesn't look like the battery has discharged to the grid yet, only charged in bursts.

corradio:
I actually agree

systemcatch:
Ok I will make the necessary changes.

systemcatch:
There was a big charge at about 8am today, 70MW for 1hr.  Don't know if we missed it due to being offline though.

corradio:
yes unfortunately the server was down, and because the parsers don't fetch data back in time we missed it :(

PR #905, opened 2017-12-11
Change data source for South Australia battery

systemcatch:
Use nemlog api rather than global-roam.

ref #869


parsers/lib/AU_battery.py
```diff
@@ -1,31 +1,40 @@
...
-    json_content = req.json()
-    dicts = json_content["seriesCollection"]
-    discharge_id = (item for item in dicts if item["id"] == "41").next()
-    storage_id = (item for item in dicts if item["id"] == "57").next()
+    today = arrow.now('Australia/Adelaide')
+    current = today.format('YYYYMMDD')
+    old = today.shift(days=-2).format('YYYYMMDD')
+    nemlog_url = 'http://nemlog.com.au/api/unit/HPRL1/{}/{}/json'.format(old, current)
+
+    req = requests.get(nemlog_url)
```
corradio: is it possible to pass a session object in order to reuse existing one?

parsers/lib/AU_battery.py
```diff
@@ -1,31 +1,40 @@
...
 
-    #Verify set to false to prevent SSL error due to domain mismatch.
-    req = requests.get(base_url, verify = False)
-    json_content = req.json()
-    dicts = json_content["seriesCollection"]
-    discharge_id = (item for item in dicts if item["id"] == "41").next()
-    storage_id = (item for item in dicts if item["id"] == "57").next()
+    today = arrow.now('Australia/Adelaide')
+    current = today.format('YYYYMMDD')
+    old = today.shift(days=-2).format('YYYYMMDD')
```
corradio: why two days back?

systemcatch: The default behavior of the api with no dates is to return data up to 2 days back.  I've kept that for the request but tbh as we only use the last value it doesn't make much difference.

corradio:
btw did you test using the mockserver?

systemcatch:
Yes it's working fine though the 0.2 MW of charging isn't shown on the breakdown (assuming rounded down).

corradio:
ok great. Feel free to commit the update state.json so we have it as reference (and for others)

PR #906, opened 2017-12-11
Add missing strings to Swedish

magol:


web/locales/sv.json
```diff
@@ -17,7 +17,7 @@
     },
     "country-panel": {
         "source": "källa",
-        "carbonintensity": "Utsläpp av koldioxid",
+        "carbonintensity": "Utsläpp av CO2",
```
corradio: Why the discrepancy between country-panel.carbonintensity and tooltip.carbonintensity?
Let's keep koldioxid for all?

magol: It is a hack to try to remove the new line in 
image

But, it is probably beter to fix the layout issue.
I can restore it to "koldioxid"

corradio: ok got it. We shouldn't change the translation in order to fit on one line (we should at the UI instead).
Furthermore, you won't have a guarantee that the hack works because on small devices the space might small enough to cause a linebreak.

magol: You're right, I've restored the string

PR #918, opened 2017-12-14
Add statnett exchange parser for SE

brunolajoie:


parsers/statnett.py
```diff
@@ -43,35 +39,76 @@ def fetch_production(country_code='SE', session=None):
...
+    return {
+        'sortedBiddingZones': '->'.join(sorted([bidding_zone1, bidding_zone2])),
+        'netFlow': net_flow if bidding_zone1 == sorted([bidding_zone1, bidding_zone2])[0] else -1 * net_flow,
+        'datetime': arrow.get(obj[0]['MeasureDate'] / 1000).datetime,
+        'source': 'driftsdata.stattnet.no',
+    }
+
+def _fetch_exchanges_from_sorted_bidding_zones(sorted_bidding_zones, session=None):
+    zones = sorted_bidding_zones.split('->')
+    return {
```
corradio: why not simply
```python
return fetch_exchange_by_bidding_zone(zones[0], zones[1], session)
```
?
Here's you're launching 3 requests.

brunolajoie:
@corradio ready to merge on my side. Flow are much better than with ENTSOE ideed. I suggest that we also use this Statnett parser to update flows in the nordic. What do you think?

README.md
```diff
@@ -288,12 +288,20 @@ Once you're done, add your parser to the [zones.json](https://github.com/tmrowco
 For more info, check out the [example parser](https://github.com/tmrowco/electricitymap/tree/master/parsers/example.py) or browse existing [parsers](https://github.com/tmrowco/electricitymap/tree/master/parsers).
 
 ### Testing parsers locally
-We've added a testing server locally. In order to test your parser, you can run
+We've added a testing server locally. In order to test your parser, make sure first that you have installed the required modules as described in parsers/requirements.txt: for that you can run
+```
+pip install -r parsers/requirements.txt
+```
+
+Then, run the following command to test your parser locally:
```
corradio: 👍

README.md
```diff
@@ -288,12 +288,20 @@ Once you're done, add your parser to the [zones.json](https://github.com/tmrowco
...
 ```
 PYTHONPATH=. python mockserver/update_state.py <zone_name>
 ```
+
 from the root directory, replacing `<zone_name>` by the zone identifier of the parser you want to test. This will fetch production and exchanges and assign it a random carbon intensity value. It should appear on the map as you refresh your local browser.
```
corradio: This doesn't read well (as you now have a ":" before the command.
I suggest replacing the sentence "Then, run the following command to test your parser locally:" by "Then, you can run" in the sentence just before the command.

corradio:
Agreed (we should add the other nordic exchanges).
Did you test that in the frontend the arrows have same values (roughly) and same direction as the ones currently in production on electricitymap.org?

brunolajoie:
@corradio all good now. Tested locally. The parser does not crash, even if called by two zone at non alphabetical order, but always return the correct pair of ['XX->YY', netflow value]

corradio:
@brunolajoie did you push your commits? I don't see them.

brunolajoie:
@corradio sorry, done now

parsers/statnett.py
```diff
@@ -43,35 +82,78 @@ def fetch_production(country_code='SE', session=None):
...
+        'netFlow': sum(map(lambda e: e['netFlow'], exchanges)),
+        'datetime': exchanges[0]['datetime'],
+        'source': exchanges[0]['source']
+    }
+
+def fetch_exchange(country_code1='DK', country_code2='NO', session=None):
+    r = session or requests.session()
+
+    sorted_exchange = '->'.join(sorted([country_code1, country_code2]))
+    print 'sorted_exchange: '
```
corradio: Clean up required

README.md
```diff
@@ -288,12 +288,20 @@ Once you're done, add your parser to the [zones.json](https://github.com/tmrowco
 For more info, check out the [example parser](https://github.com/tmrowco/electricitymap/tree/master/parsers/example.py) or browse existing [parsers](https://github.com/tmrowco/electricitymap/tree/master/parsers).
 
 ### Testing parsers locally
-We've added a testing server locally. In order to test your parser, you can run
+We've added a testing server locally. In order to test your parser, make sure first that you have installed the required modules as described in parsers/requirements.txt: for that you can run
+```
+pip install -r parsers/requirements.txt
+```
+
+Then, you can run:
```
corradio: Remove ":"

PR #920, opened 2017-12-14
Repairing Bulgaria

### LINKED ISSUE #890: Bulgaria is missing

systemcatch:
Fixes #890 

Working ok on the mockserver.

bg_repaired

Side note; when trying to test using update_state.py the script failed due to calling several exchange methods that required ENTSOE api keys.  Easy to comment out and bypass the exchanges but still something to think about.


stgram:
Thank you so much! Now if only there was a quick fix like that for our coal emmisions...

Issue #787, opened 2017-10-04
California (CAISO)


corradio:
Unfortunately we only have renewable data live.
Can someone help us get access to fossil?

corradio:
I've been able to gather the following potential sources:
- https://github.com/WattTime/pyiso/blob/master/pyiso/caiso.py

And I got a reply from CAISO:
> Good morning and thank you for the inquiry. This data can be found on oasis.caiso.com where it can be downloaded in CSV or XML format under the System Demand tab. There is also an API specification that allows users to pull this data without going to the User Interface found here - 
http://www.caiso.com/Documents/OASIS-InterfaceSpecification_v4_3_3Clean_Independent2016Release.pdf 
The ISO also offers the below option in CSV as well. http://www.caiso.com/Pages/TodaysOutlookCSVFileInstructions.aspx

systemcatch:
They responded to your email which is a good sign, and the API spec looks quite hefty.  If we can get a renewable breakdown that includes hydro we may be in business.

corradio:
New response:
> Thank you for your e-mail. Yes, we have changed our Daily Renewable Watch. The information that we have provided is what is sent to that report.

!

http://content.caiso.com/green/renewrpt/20171009_DailyRenewablesWatch.txt now reports breakdown over last day. Can someone take the lead and make a parser?

systemcatch:
When I try to set the date for today it gives a 404 error.  Assume I'm missing something?
http://content.caiso.com/green/renewrpt/20171101_DailyRenewablesWatch.txt

corradio:
I'm reaching out to them.

systemcatch:
At a guess today's txt file is on the server and a script writes new lines to the file every hour.  Or it's possible that the data is stored elsewhere and the txt file is auto generated at the end of each day.  Their live renewable graph is just an image so it must be fed today's data from elsewhere then updated on the page.

7hibault:
Hello, I'm new here, I'd like to help. I assume this is generated at the end of each day, as the latest available is yesterday's: http://content.caiso.com/green/renewrpt/20171112_DailyRenewablesWatch.txt

I could work on the parser if there's no urgency, I should be able to start working on it in a week or so

corradio:
Yes but data is only available at midnight every day.
I'm trying to get in touch with them but I'm struggling to get a response.

Olivier

corradio:
CAISO is not being very opened to increasing the rate at which they publish data.
[...]

PR #922, opened 2017-12-15
Near real-time CAISO

7hibault:
See #787 

screenshot_20171215_150137


mockserver/public/v3/state
```diff
@@ -1 +0,0 @@
...22}, "countryCode": "SE", "exchange": {"DK": 411.0, "NO": 194.0, "DE": -71.0, "LT": -732.0, "FI": -1983.0, "PL": -595.0}, "storage": {}, "datetime": "2017-12-14T09:43:00+00:00", "source": "driftsdata.stattnet.no", "maxProduction": 10912.0, "production": {"nuclear": 8626.0, "unknown": 2046.0, "wind": 1420.0, "hydro": 10912.0}, "co2intensity": 107.23353790918422, "totalProduction": 23004.0}, "BR-NE": {"exchange": {"BR-CS": -235.9166}}, "BR-N": {"exchange": {"BR-CS": -1702.00964}}, "FI": {"exchangeCo2Intensities": {"SE": 107.23353790918422}, "exchange": {"SE": 1983.0}}, "IE": {"exchangeCo2Intensities": {"GB": 491.87272191893015, "GB-NIR": 491.87272191893015}, "countryCode": "IE", "exchange": {"GB": 173.0, "GB-NIR": 83.8}, "storage": {"hydro": 0}, "datetime": "2017-10-09T09:30:00+00:00", "source": "entsoe.eu", "production": {"oil": 401.0, "nuclear": null, "gas": 1652.0, "biomass": 238.0, "coal": 456.0, "geothermal": null, "solar": null, "unknown": 1.0, "wind": 385.0, "hydro": 74.0}, "co2intensity": 491.87272191893015}, "PL": {"exchangeCo2Intensities": {"SE": 107.23353790918422}, "exchange": {"SE": 595.0}}, "GB-NIR": {"exchangeCo2Intensities": {"IE": null}, "exchange": {"IE": -83.8}}}}}
```
corradio: it seems like this file was deleted?

parsers/US_CA.py
```diff
@@ -34,37 +34,51 @@ def fetch_production(country_code='US-CA', session=None):
...
-    yesterday = arrow.utcnow().to('US/Pacific').shift(days=-1).replace(hour = 0, minute = 0, second = 0, microsecond = 0)
-    url = 'http://content.caiso.com/green/renewrpt/' + yesterday.format('YYYYMMDD') +'_DailyRenewablesWatch.txt'
-
-    renewableResources = pandas.read_table(url, sep='\t\t', skiprows=2, header=None, names=['Hour', 'GEOTHERMAL','BIOMASS', 'BIOGAS', 'SMALL HYDRO','WIND TOTAL', 'SOLAR PV', 'SOLAR THERMAL'], skipfooter=27, skipinitialspace=True, engine='python')
-    otherResources = pandas.read_table(url, sep='\t\t', skiprows=30, header=None, names=['Hour', 'RENEWABLES', 'NUCLEAR', 'THERMAL', 'IMPORTS', 'HYDRO'], skipinitialspace=True, engine='python')
-
-    dailyData = []
+    url = 'http://www.caiso.com/outlook/SP/fuelsource.csv'
+    csv = pandas.read_csv(url)
+    latest_index = len(csv)-1
```
corradio: are you able to return an array of items instead of just the latest one? Then the parser will be more robust to downtimes.

parsers/US_CA.py
```diff
@@ -34,37 +34,51 @@ def fetch_production(country_code='US-CA', session=None):
...
+    h, m = map(int, csv['Time'][latest_index].split(':'))
+    date = arrow.utcnow().to('US/Pacific').replace(hour = h, minute = m, second = 0, microsecond = 0)
```
corradio: `utcnow` and the time of the csv might not be the same date.
Can't you directly parse `csv['Time']`?

parsers/US_CA.py
```diff
@@ -34,37 +34,51 @@ def fetch_production(country_code='US-CA', session=None):
...
-        data['production']['geothermal'] = renewableResources['GEOTHERMAL'][i]
-        data['production']['unknown'] = otherResources['THERMAL'][i] # this is not specified in the list
-        # set the date at the end of the hour
-        data['datetime'] = yesterday.shift(hours = i+1).datetime
-        dailyData.append(data)
+    for key, value in storage_map.items():
+        storage = csv[key][latest_index]
+        # if another mean of storage created a value, sum them up
+        if value in data['production']:
+            data['storage'][value] += storage
```
corradio: Are you sure storage is positive when energy is charged, and negative when it is discharged?

7hibault: Not 100% sure. This is not explicitly expressed. Should we ask them?
> An increasing amount of energy storage is being connected to the ISO grid, including non-generator resources, such as batteries, that can generate and store electricity until dispatched.

parsers/US_CA.py
7hibault: No. csv['Time'] only countains the time of the day (hour and minute). That's why I'm using the date "now" in US/Pacific time, which should be the same as the latest value from this file. Except if this is called somewhere between 00:00 and 00:05 in California, then the latest value might be from the day before and the date will indeed be wrong...

parsers/US_CA.py
7hibault: I could but that makes the issue stated below more problematic.

parsers/US_CA.py
corradio: it seems like a positive value means discharge:
image
image

This is the opposite convention of ours, so we need to multiply by -1

mockserver/public/v3/state
7hibault: My bad, I thought this was only generated for the mockserver

parsers/US_CA.py
7hibault: Actually the date is not in the fuelsource csv file but it is in other files like CAISO-renewables. I can use the date from that file instead.

mockserver/public/v3/state
corradio: yeah but let's keep it for the next contributors to be able to test. Can you restore and replace by the file where CA has been processed?

parsers/US_CA.py
corradio: If you can, please do (it will be more robust).
Else, we can do later.

mockserver/public/v3/state
```diff
@@ -0,0 +1 @@
...e": "entsoe.eu", "datetime": "2017-12-13T10:00:00+00:00"}, "GB->IE": {"sortedCountryCodes": "GB->IE", "netFlow": 173.0, "datetime": "2017-10-09T10:00:00+00:00", "source": "entsoe.eu", "co2intensity": null}, "AR->BR-S": {"sortedCountryCodes": "AR->BR-S", "netFlow": 0.0077993, "co2intensity": 337.7410457816576, "source": "ons.org.br", "datetime": "2017-10-05T08:31:00-03:00"}, "DE->SE": {"sortedCountryCodes": "DE->SE", "netFlow": -71.0, "co2intensity": 107.23353790918422, "source": "entsoe.eu", "datetime": "2017-12-13T10:00:00+00:00"}, "GB-NIR->IE": {"datetime": "2017-10-10T10:15:00+01:00", "source": "soni.ltd.uk", "netFlow": 83.8, "co2intensity": null, "sortedCountryCodes": "GB-NIR->IE"}, "BR-CS->BR-NE": {"sortedCountryCodes": "BR-CS->BR-NE", "netFlow": -235.9166, "datetime": "2017-10-02T07:58:00-03:00", "source": "ons.org.br", "co2intensity": null}, "LT->SE": {"sortedCountryCodes": "LT->SE", "netFlow": -732.0, "co2intensity": 107.23353790918422, "source": "entsoe.eu", "datetime": "2017-12-13T10:00:00+00:00"}, "AR->BR-CS": {"sortedCountryCodes": "AR->BR-CS", "netFlow": -0.008066349, "datetime": "2017-10-02T07:58:00-03:00", "source": "ons.org.br", "co2intensity": 97.54917499086058}}}}
```
7hibault: I've readded the file

PR #925, opened 2017-12-17
split CA-ON and update CA-NB exchanges with US

jarek:
* Split Ontario exchanges into separate items for MI, MN, and NY.
* Changed New Brunswick exchange from generic "US" to US-NEISO
* Did not changed Alberta and B.C. yet because we don't yet have western U.S. figured out

Screenshot from mockserver:

image

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -385,7 +385,7 @@
     },
     "rotation": -45
   },
-  "CA-NB->US": {
+  "CA-NB->US-NEISO": {
     "lonlat": [
       -67.7771,
       46.139
@@ -405,7 +405,17 @@
     },
     "rotation": 90
   },
-  "CA-ON->US": {
+  "CA-ON->US-MI": {
+    "lonlat": [
+      -82.1195,
+      43.5896
+    ],
+    "parsers": {
+      "exchange": "CA_ON.fetch_exchange"
+    },
+    "rotation": -90
+  },
+  "CA-ON->US-MN": {
     "lonlat": [
       -92.722024,
       48.566155
@@ -414,6 +424,16 @@
       "exchange": "CA_ON.fetch_exchange"
     },
     "rotation": 180
+  },
+  "CA-ON->US-NY": {
+    "lonlat": [
+      -78.9147,
+      43.5523
+    ],
+    "parsers": {
+      "exchange": "CA_ON.fetch_exchange"
+    },
+    "rotation": 135
   },
   "CH->DE": {
     "lonlat": [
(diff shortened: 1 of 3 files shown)


HansHyde:
@jarek 
ontario interconnections

jarek:
Yes, and those are all implemented with data from IESO. Am I missing something?

corradio:
Great!!

Looks good on my side. Waiting for confirmation from @HansHyde.

HansHyde:
5 exchanges with Ontario - looks correct

HansHyde:
@jarek can you direct me to where the coding is for **Bold Text**...

"CA-ON->US-NY": {
    "lonlat": [
      -78.9147,
      43.5523
    ],
    "parsers": {
      **"exchange": "CA_ON.fetch_exchange"**
    },
    "rotation": 135

I know you get the exchange data from IESO, and this is the code in "config/exchanges.json" but where is the code behind "CA_ON.fetch_exchange"?  Sorry, noob question. Thanks

corradio:
@HansHyde all parsers are here:
https://github.com/tmrowco/electricitymap/tree/master/parsers

I suggest you take a look at the README on the homepage of this github repository as it explains where parsers are, how they work etc..

HansHyde:
@jarek - a very nerdy endeavor (by myself), but when we get PJM & MISO parsers up, you will be able to cleanup IESO (hourly reporting) exchanges to 5 minute via PJM & it's "Lake Erie Loop Flows" with NYISO 5 minute exchanges. Here's the table you linked Ontario exchanges to (with some mods)...

http://reports.ieso.ca/public/IntertieScheduleFlow/PUB_IntertieScheduleFlow_20171222.xml

lake erie loop flow

Basically it is NYISO rtd (IESO imports) + PJM rtd (Lake Erie Loop Flow)

ie  
- IESO = 800  (from NYISO rtd stream)
- LELF = -200  (from PJM rtd stream)
- NYISO = 600  (actual NYISO imports from IESO)

http://www.pjm.com/markets-and-operations/interregional-map.aspx

On the flip side, this only represents about 200 MW of data  (so 1/100th the ~200GW of generation in NY, ON, PJM & MISO at any one time.

jarek:
To confirm I understand correctly: "flow" is the physical value while "export" is the accounting value? Does this mean Ontario booked 674 MW exports to Michigan, but on the wire there were actually only 410 MW and the balance went to NY despite not being on the books there?

If yes, we should probably use the flow values - and we can just switch to using http://reports.ieso.ca/public/IntertieScheduleFlow/PUB_IntertieScheduleFlow_20171222.xml as source rather than http://live.gridwatch.ca/WebServices/GridWatchWebApp.asmx/GetHomeViewData_v2 - it seems to have more granularity, too.

HansHyde:
@jarek , roughly yes.  It's a bit of a "ghost" in the electric system.  PJM is reporting it, as it goes through their transmission lines.

I think down the road, with both NYISO & PJM (hopefully MISO?) reporting in realtime (5 min) and parsers for each, we can link the exchanges related to Ontario to NYISO/PJM, instead of the once an hour reporting of IESO.

PR #931, opened 2017-12-19
Update zones.json

admetny:
Capacity info for Belarus has been added.


brunolajoie:
Thanks a lot for the PR guys,

I have taken the responsibility to revert back BY biogas capacity into biomass category, as our "biomass" is meant to encompass everything that comes from photosynthesis, from near-and-far (biomass, biogas, wood pellet, waste, etc...)

PR #935, opened 2017-12-19
New York exchange with New England

systemcatch:
Tested and working on the mockserver.  The direction and flow look correct to me.

nyne


corradio:
@HansHyde does this look good to you? Just looking for your confirmation before I merge.

HansHyde:
nice job Chris @systemcatch 

those are the 3 I have, and as of 5 minutes ago the flow was....

ne - ny exchanges

HansHyde:
Oli, on this does EM in the backend have ability to confirm to datastream sources are balanced with each other? 

Syscatch data is from NEISO, the data I posted is from NYISO (so they should be opposite each other). Not a big deal, more curiosity, as this would act as a check on the data reporting.

Of course, NYISO data would be linked within the NY Parser, and might not show as a 'second' exchange of the frontend.

corradio:
Good point. I've thought about it but currently we can't.

PR #940, opened 2017-12-22
Create tests for validation functions

systemcatch:
Adding some tests for the quality check module.  They're fairly simple at the moment, taking mock datapoints and checking to see if Exceptions are correctly raised and if the good datapoints pass as expected.


corradio:
Good idea :)

Issue #909, opened 2017-12-12
Ranking in alphabetical order


magol:
It's starting to be many countries, and finding a particular country in the ranking list can be hard. I suggest that you can choose to see the list in alphabetical order.

As an alternative (or complement) you may add a search box. If you write the first letters of a country, the country will be highlighted in the list and on the map.

This may be good if you do not know where on the map a particular country is located, or if it's so small that it's hard to select on the map.

corradio:
Makes sense!

systemcatch:
Search box is now in place!

PR #945, opened 2017-12-24
Add search bar #909

DeOldSax:
Added a simple search bar for looking up countries in the ranking list.


corradio:
Quite cool @DeOldSax ! I'll take a look at this as soon as I have a chance.

corradio:
Screenshot of mobile view:
image
Screenshot of desktop view:
image

web/locales/en.json
```diff
@@ -29,7 +29,8 @@
     },
     "rankings-panel": {
         "country-ranking": "Ranking",
-        "show-ranking": "Show ranking"
+        "show-ranking": "Show ranking",
+        "search": "Search..."
```
corradio: Can you rename to `Search a zone`, without the trailing `.`?

web/app/main.js
```diff
@@ -1111,6 +1111,24 @@ function dataLoaded(err, clientVersion, argCallerLocation, state, argSolar, argW
         countryTooltip.hide();
     });
 
+    // Add search bar handler
+    var selector = d3.select('.country-search-bar input')
```
corradio: Would be nice to write ES6 code instead (we're slowly moving everything to ES6).
It's ok if you can't do it right now.

corradio:
Looks very nice. @jarek @brunolajoie @systemcatch any comments?

DeOldSax:
@corradio thx for the review - let me know if you want me clean this up to have a single commit

corradio:
@DeOldSax all is good - just making sure we're all on the same page regarding the issue.
I'll squash commits into one when merging.
Can you however merge `master` into your branch?

brunolajoie:
Nie feature! But when I test it locally on chrome it does not looks similar to @corradio screenshot and the search does not work when pressing enter. 
image

corradio:
@brunolajoie there's probably something wrong with your setup.
Can you try building from master? Also, check the console, maybe an error is hidden there.

Olivier

jarek:
Works for me on the prod site now. Maybe it was a cache problem for Bruno?

PR #949, opened 2017-12-26
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3.  There might be other changes required to complete a port to Python 3 but this PR is a minimal, safe first step.

Run: [futurize --stage1 -w **/*.py](http://python-future.org/futurize_cheatsheet.html)

    See Stage 1: "safe" fixes http://python-future.org/automatic_conversion.html#stage-1-safe-fixes

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,12 @@ stages:
     steps:
       - type: checkout
       - type: shell
+        name: Set Python version
+        command: pyenv global 3.5.2
+      - type: shell
+        name: Upgrade pip
+        command: pip install --upgrade pip
+      - type: shell
         name: Install Docker Compose
         command: pip install docker-compose==1.17.1
       - type: shell
@@ -19,9 +25,15 @@ stages:
         name: Lint
         command: |
           set -exu
-          pip install pylint
+          python --version
+          pip --version
+          pip install flake8 pylint
           sudo apt-get install libxml2-dev
           pip install -r parsers/requirements.txt
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+          # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
           pylint -E parsers/*.py
           python -m unittest discover parsers/test
       - save_cache:
--- a/mockserver/update_state.py
+++ b/mockserver/update_state.py
@@ -1,7 +1,13 @@
-from __future__ import print_function
+#!/usr/bin/env python3
 # This script should be run from the root directory
-import arrow, importlib, json, pprint, sys
+import importlib
+import json
+import pprint
+import sys
 from random import random
+
+import arrow
+
 pp = pprint.PrettyPrinter(indent=2)
 
 # Read parser import list from config jsons
@@ -19,14 +25,17 @@ production_parser = zone_config['parsers']['production']
 exchange_parser_keys = []
 for k in exchanges_config.keys():
     zones = k.split('->')
-    if zone_name in zones: exchange_parser_keys.append(k)
+    if zone_name in zones:
+        exchange_parser_keys.append(k)
 
 # Import / run production parser
-print('Finding and executing %s production parser %s..' % (zone_name, production_parser))
+print('Finding and executing %s production parser %s..' % (zone_name,
+                                                           production_parser))
 mod_name, fun_name = production_parser.split('.')
 mod = importlib.import_module('parsers.%s' % mod_name)
 production = getattr(mod, fun_name)(zone_name)
-if type(production) == list: production = production[-1]
+if type(production) == list:
+    production = production[-1]
 pp.pprint(production)
 
 # Import / run exchange parser(s)
@@ -38,7 +47,8 @@ for k in exchange_parser_keys:
     mod = importlib.import_module('parsers.%s' % mod_name)
     sorted_zone_names = sorted(k.split('->'))
     exchange = getattr(mod, fun_name)(sorted_zone_names[0], sorted_zone_names[1])
-    if type(exchange) == list: exchange = exchange[-1]
+    if type(exchange) == list:
+        exchange = exchange[-1]
     exchanges.append(exchange)
     pp.pprint(exchange)
 
@@ -67,11 +77,11 @@ with open('mockserver/public/v3/state', 'r') as f:
 
         for z in exchange_zone_names:
             other_zone = exchange_zone_names[(exchange_zone_names.index(z) + 1) % 2]
-            if not z in obj['countries']:
+            if z not in obj['countries']:
                 obj['countries'][z] = {}
-            if not 'exchange' in obj['countries'][z]:
+            if 'exchange' not in obj['countries'][z]:
                 obj['countries'][z]['exchange'] = {}
-            if not 'exchangeCo2Intensities' in obj['countries'][z]:
+            if 'exchangeCo2Intensities' not in obj['countries'][z]:
                 obj['countries'][z]['exchangeCo2Intensities'] = {}
             obj['countries'][z]['exchange'][other_zone] = e['netFlow']
             if z == exchange_zone_names[0]:
@@ -83,7 +93,7 @@ with open('mockserver/public/v3/state', 'r') as f:
                 obj['countries'].get(other_zone, {}).get('co2intensity',
                     obj['countries'][z].get('co2intensity', None)) if is_import \
                 else obj['countries'][z].get('co2intensity', None)
-            
+
     # Set state datetime
     obj['datetime'] = production['datetime']
     # Save
--- a/mockserver/update_state.py
+++ b/mockserver/update_state.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # This script should be run from the root directory
 import arrow, importlib, json, pprint, sys
 from random import random
@@ -21,7 +22,7 @@ for k in exchanges_config.keys():
     if zone_name in zones: exchange_parser_keys.append(k)
 
 # Import / run production parser
-print 'Finding and executing %s production parser %s..' % (zone_name, production_parser)
+print('Finding and executing %s production parser %s..' % (zone_name, production_parser))
 mod_name, fun_name = production_parser.split('.')
 mod = importlib.import_module('parsers.%s' % mod_name)
 production = getattr(mod, fun_name)(zone_name)
@@ -32,7 +33,7 @@ pp.pprint(production)
 exchanges = []
 for k in exchange_parser_keys:
     exchange_parser = exchanges_config[k]['parsers']['exchange']
-    print 'Finding and executing %s exchange parser %s..' % (k, exchange_parser)
+    print('Finding and executing %s exchange parser %s..' % (k, exchange_parser))
     mod_name, fun_name = exchange_parser.split('.')
     mod = importlib.import_module('parsers.%s' % mod_name)
     sorted_zone_names = sorted(k.split('->'))
@@ -42,7 +43,7 @@ for k in exchange_parser_keys:
     pp.pprint(exchange)
 
 # Load and update state
-print 'Updating and writing state..'
+print('Updating and writing state..')
 with open('mockserver/public/v3/state', 'r') as f:
     obj = json.load(f)['data']
     obj['countries'][zone_name] = {}
@@ -88,4 +89,4 @@ with open('mockserver/public/v3/state', 'r') as f:
     # Save
 with open('mockserver/public/v3/state', 'w') as f:
     json.dump({'data': obj}, f)
-print '..done'
+print('..done')
--- a/parsers/AR.py
+++ b/parsers/AR.py
@@ -1,11 +1,17 @@
 #!/usr/bin/python
 
+from __future__ import print_function
 import arrow
 import string
 import requests
 from bs4 import BeautifulSoup
 import re
 import itertools
+
+try:
+    unicode         # Python 2
+except NameError:
+    unicode = str   # Python 3
 
 
 #This parser gets hourly electricity generation data from portalweb.cammesa.com/Memnet1/default.aspx
@@ -533,7 +539,7 @@ def fetch_price(country_code='AR', session = None):
         rprice = lprice.split('[')[0]
         price = float(rprice.replace(',','.'))
 
-    except AttributeError, ValueError:
+    except AttributeError as ValueError:
         #Price element not present or no price stated.
         price = None
 
@@ -623,7 +629,7 @@ def get_thermal(session=None):
         try:
             #avoids including titles and headings
             if all((item.isupper(), not item.isalpha(), not ' ' in item)):
-                print '{} is missing from the AR plant mapping!'.format(item)
+                print('{} is missing from the AR plant mapping!'.format(item))
         except AttributeError:
             #not a string....
             continue
--- a/parsers/AR.py
+++ b/parsers/AR.py
@@ -1,12 +1,12 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 
-from __future__ import print_function
-import arrow
+import itertools
+import re
 import string
+
+import arrow
 import requests
 from bs4 import BeautifulSoup
-import re
-import itertools
 
 try:
     unicode         # Python 2
@@ -14,19 +14,19 @@ except NameError:
     unicode = str   # Python 3
 
 
-#This parser gets hourly electricity generation data from portalweb.cammesa.com/Memnet1/default.aspx
-#for Argentina.  Currently wind and solar power are small contributors and not monitored but this is
-#likely to change in the future.
+# This parser gets hourly electricity generation data from portalweb.cammesa.com/Memnet1/default.aspx
+# for Argentina.  Currently wind and solar power are small contributors and not monitored but this is
+# likely to change in the future.
 
-#Useful links.
-#https://en.wikipedia.org/wiki/Electricity_sector_in_Argentina
-#https://en.wikipedia.org/wiki/List_of_power_stations_in_Argentina
-#http://globalenergyobservatory.org/countryid/10#
-#http://www.industcards.com/st-other-argentina.htm
+# Useful links.
+# https://en.wikipedia.org/wiki/Electricity_sector_in_Argentina
+# https://en.wikipedia.org/wiki/List_of_power_stations_in_Argentina
+# http://globalenergyobservatory.org/countryid/10#
+# http://www.industcards.com/st-other-argentina.htm
 
 
-#Map of power plants to generation type.
-#http://portalweb.cammesa.com/memnet1/revistas/estacional/base_gen.html
+# Map of power plants to generation type.
+# http://portalweb.cammesa.com/memnet1/revistas/estacional/base_gen.html
 
 power_plant_type = {
                      'ABRODI01': 'gas',
@@ -469,7 +469,7 @@ power_plant_type = {
                      }
 
 
-#URL's for thermal and hydro pages and data sources respectively.
+# URL's for thermal and hydro pages and data sources respectively.
 
 url = ('http://portalweb.cammesa.com/MEMNet1/Pages/Informes%20por'
        '%20Categor%C3%ADa/Operativos/VisorReporteSinComDesp_minimal.aspx'
@@ -500,6 +500,7 @@ thurl = ('http://portalweb.cammesa.com/Reserved.ReportViewerWebControl.'
 
 cammesa_url = 'http://portalweb.cammesa.com/default.aspx'
 
+
 def webparser(req):
     """Takes content from webpage and returns all text as a list of strings"""
 
@@ -509,7 +510,8 @@ def webparser(req):
 
     return data_table
 
-def fetch_price(country_code='AR', session = None):
+
+def fetch_price(country_code='AR', session=None):
     """
     Requests the last known power price of a given country
     Arguments:
@@ -527,20 +529,20 @@ def fetch_price(country_code='AR', session = None):
       """
     s = session or requests.Session()
     price_req = s.get(cammesa_url)
-    psoup = BeautifulSoup(price_req.content,'html.parser')
-    find_price = psoup.find('td', class_ = "cssFuncionesLeft", align = "left")
+    psoup = BeautifulSoup(price_req.content, 'html.parser')
+    find_price = psoup.find('td', class_="cssFuncionesLeft", align="left")
 
     try:
         price_text = find_price.getText()
 
-        #Strip all whitespace and isolate number.  Convert to float.
+        # Strip all whitespace and isolate number.  Convert to float.
         price_nws = "".join(price_text.split())
         lprice = price_nws.rpartition(':')[2]
         rprice = lprice.split('[')[0]
-        price = float(rprice.replace(',','.'))
+        price = float(rprice.replace(',', '.'))
 
-    except AttributeError as ValueError:
-        #Price element not present or no price stated.
+    except (AttributeError, ValueError):
+        # Price element not present or no price stated.
         price = None
 
     datetime = arrow.now('UTC-3').floor('hour').datetime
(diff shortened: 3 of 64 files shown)


parsers/AR.py
```diff
@@ -533,7 +539,7 @@ def fetch_price(country_code='AR', session = None):
         rprice = lprice.split('[')[0]
         price = float(rprice.replace(',','.'))
 
-    except AttributeError, ValueError:
+    except AttributeError as ValueError:
```
systemcatch: This still catches both errors I assume?

parsers/ES_CN.py
```diff
@@ -1,151 +1,152 @@
...
-    }
-
-    return data
-
-
-if __name__ == '__main__':
-    session = Session
-    print fetch_consumption('ES-CN', session)
-    print fetch_production('ES-CN', session)
+from __future__ import print_function
```
systemcatch: I'm not sure why the whole diff has changed?

parsers/ES_IB.py
```diff
@@ -1,98 +1,99 @@
...
-
-        return data
-
-
-if __name__ == '__main__':
-    session = Session
-    print fetch_consumption('ES-IB', session)
-    print fetch_production('ES-IB', session)
-    print fetch_exchange('ES', 'ES-IB', session)
+from __future__ import print_function
```
systemcatch: Same here.

parsers/TW.py
```diff
@@ -9,10 +10,7 @@ def fetch_production(country_code='TW', session=None):
     r = session or requests.session()
     url = 'http://data.taipower.com.tw/opendata01/apply/file/d006001/001.txt'
     response = requests.get(url)
-    content = response.content
```
systemcatch: This is safe to do?  Can you run this parser and check it still works.

parsers/test/test_quality.py
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/python
 
 """Tests for quality.py."""
+from __future__ import absolute_import
 
 import unittest
 from parsers.lib.quality import validate_consumption, validate_exchange, validate_production
-from mocks.quality_check import *
+from .mocks.quality_check import *
```
systemcatch: Nice.

systemcatch:
Looking further ahead I don't see a problem with dependencies not being available in Python 3.  The main issue we face is probably the change in unicode handling.  Since we get data from all over the world some of the parsers will need to be checked carefully to ensure they still handle text ok.

Of course there are python modules on the backend which might have extra depends or problems to be dealt with, @corradio any thoughts on this?

parsers/AR.py
cclauss: Good catch.  I did this wrong.  Fixing now.

This is exactly why Python 3 changed the syntax.  The old way was too ambiguous.

parsers/ES_CN.py
cclauss: This is usually when indentation is done with tabs instead of four spaces.  PEP8 recommends spaces.

parsers/ES_IB.py
cclauss: This is usually when indentation is done with tabs instead of four spaces.  PEP8 recommends spaces.

parsers/TW.py
cclauss: I do not know how to run the parser but perhaps we should have Travis CI run this test on every pull request.

systemcatch:
To run the Taiwan parser change into the parsers directory on the command line then do `python TW.py`, it will output results/errors.

cclauss:
$ __python3 TW.py__
```
TW.py:30: FutureWarning: convert_objects is deprecated.  To re-infer data dtypes for object columns, use DataFrame.infer_objects()
For all other conversions use the data-type specific converters pd.to_datetime, pd.to_timedelta and pd.to_numeric.
  objData = objData.convert_objects(convert_numeric=True)
{'countryCode': 'TW', 'datetime': datetime.datetime(2017, 12, 26, 22, 0, tzinfo=tzfile('/usr/share/zoneinfo/Asia/Taipei')), 'production': {'coal': 8783.8999999999996, 'gas': 10563.300000000001, 'oil': 1426.1000000000001, 'hydro': 337.10000000000002, 'nuclear': 2874.5999999999999, 'solar': 0.0, 'wind': 210.99999999999994, 'unknown': 685.20000000000005}, 'capacity': {'coal': 12297.200000000001, 'gas': 14481.199999999997, 'oil': 2573.1999999999998, 'hydro': 2089.2999999999997, 'nuclear': 5144.0, 'solar': 1345.4000000000001, 'wind': 690.29999999999995, 'unknown': 623.20000000000005}, 'storage': {'hydro': -453.29999999999995}, 'source': 'taipower.com.tw'}
```

corradio:
Thanks for submitting this.
This is a big step forward.
The main question at this stage is whether we want to keep Py2/3 compatibility, or if we want to switch completely to python3.
@cclauss what are the reasons to switch to python3 from your side?

cclauss:
Python 2 end of life is just two years away.  Python 3 is faster than Python 2 for most use cases.  Code that is compatible with both is commonplace.

corradio:
Got it.
Any reasons why we shouldn't move to Python 3 (exclusively)? I think it's
acceptable to ask for contributors to use Python 3.
Furthermore it would reduce code complexity (I'm not a fan of maintaining
two versions at the same time)

cclauss:
The out-of-box experience might be better if we support both because Mac OSX and many Linux distros have Python 2 installed by default but not Python 3.

We can add flake8 tests to our circle.yml to ensure that all future PRs are syntax compatible with both.

corradio:
That sounds like a good idea. Can you add it? Feel free to remove pylint if
flake8 removes the need for it.

systemcatch:
Most people who want to develop can install python 3 easily enough.  OSX has homebrew and distros like Ubuntu have 3 installed by default in 18.04.  Besides if we require supporting 2/3 then everyone will have to check their updates run on both which adds complexity to writing code.  I think it's better to move fully to 3.

corradio:
Sounds reasonable. @cclauss can you prepare a full switch to python 3 in this PR?

cclauss:
https://circleci.com/gh/tmrowco/electricitymap/1843 __Show Versions__ says that Circle CI is still using Python 2 instead of Python 3.  :-(  How do we change that?!?

corradio:
Have you tried this:
https://circleci.com/docs/1.0/configuration/#python-version

corradio:
Actually you can use `pyenv`.
I have successfully been adding `pyenv global 3.5.2` in my circle.yml

parsers/CL_SING.py
```diff
@@ -85,7 +86,7 @@ def data_processer(data):
 
         for key in datapoint.keys():
             if not key in plant_map.keys():
-                print '{} is missing from the CL_SING plant mapping.'.format(key)
+                print('{} is missing from the CL_SING plant mapping.'.format(key))
 
         mapped_plants = [(plant_map.get(plant, 'unknown'), val) for plant, val in datapoint.iteritems()]
```
corradio: Doesn't all `iteritems()` calls need to be changed to `items()`?

cclauss: Where is circle.yml did you add the pyenv stuff?

On iteritems(), Yes.  This is the kinda stuff that flake8 would catch if it was running on Python 3.

circle.yml
```diff
@@ -1,10 +1,15 @@
 version: 2
 executorType: machine
+docker:
+- image: circleci/python:3.6.4
 stages:
   build:
     workDir: ~/electricitymap
     steps:
       - type: checkout
+      - type: shell
```
corradio: Do you really need to upgrade pip? Upgrading to latest version can always be problematic if the latest version breaks for some reason.
You can add the `pyenv global 3.5.2` shell command to force python3 by the way

cclauss: WHERE do I add `pyenv global 3.5.2`?  That is not working for me.

Yes.  Upgrading pip is the right thing to do.  The old pip was designed for Python 2 and a lot of important Python 3 related changes have happened in the past 18 months.  https://pip.pypa.io/en/stable/news/

corradio: Gotcha.
For pyenv, just add it as a shell command as a first step (before upgrading pip)
```yaml
steps:
  - type: shell
    name: Set python version
    command: pyenv global 3.5.2
```

cclauss:
I could use help on these remaining __map/filter on lambda could be replaced by comprehension (deprecated-lambda)__ errors:
```
pylint -E parsers/AR.py parsers/AU.py parsers/AU_WA.py parsers/BG.py parsers/BO.py parsers/BR.py parsers/CA_AB.py parsers/CA_BC.py parsers/CA_NB.py parsers/CA_NS.py parsers/CA_ON.py parsers/CA_PE.py parsers/CA_YT.py parsers/CL_SING.py parsers/CR.py parsers/CY.py parsers/DO.py parsers/ENTE.py parsers/ENTSOE.py parsers/ES_CN.py parsers/ES_IB.py parsers/ESIOS.py parsers/example.py parsers/FO.py parsers/FR.py parsers/GB_NIR.py parsers/GT.py parsers/IN_AP.py parsers/IN_CT.py parsers/IN_DL.py parsers/__init__.py parsers/IN_KA.py parsers/IN_PB.py parsers/IS.py parsers/MD.py parsers/MY_WM.py parsers/NI.py parsers/NZ.py parsers/PA.py parsers/PE.py parsers/SG.py parsers/statnett.py parsers/SV.py parsers/TR.py parsers/TW.py parsers/UA.py parsers/US_CA.py parsers/US_NEISO.py parsers/US_NY.py parsers/UY.py
No config file found, using default configuration
************* Module electricitymap.parsers.AU
E:480,10: Value 'filter(lambda o: o['REGIONID'] == mapping['region_id'], response.json()['ELEC_NEM_SUMMARY'])' is unsubscriptable (unsubscriptable-object)
E:490,25: Value 'filter(lambda f: f['name'] == interconnector_name, flows)' is unsubscriptable (unsubscriptable-object)
E:537,10: Value 'filter(lambda o: o['REGIONID'] == PRICE_MAPPING_DICTIONARY[country_code], response.json()['ELEC_NEM_SUMMARY'])' is unsubscriptable (unsubscriptable-object)
************* Module electricitymap.parsers.statnett
E: 86,29: Value 'filter(lambda x: x['titleTranslationId'] == ('ProductionConsumption.%s%sDesc') % (('Nuclear', country_code)), obj['NuclearData'])' is unsubscriptable (unsubscriptable-object)
E: 89,27: Value 'filter(lambda x: x['titleTranslationId'] == ('ProductionConsumption.%s%sDesc') % (('Hydro', country_code)), obj['HydroData'])' is unsubscriptable (unsubscriptable-object)
E: 92,26: Value 'filter(lambda x: x['titleTranslationId'] == ('ProductionConsumption.%s%sDesc') % (('Wind', country_code)), obj['WindData'])' is unsubscriptable (unsubscriptable-object)
E: 96,22: Value 'filter(lambda x: x['titleTranslationId'] == ('ProductionConsumption.%s%sDesc') % (('Thermal', country_code)), obj['ThermalData'])' is unsubscriptable (unsubscriptable-object)
E: 99,22: Value 'filter(lambda x: x['titleTranslationId'] == ('ProductionConsumption.%s%sDesc') % (('NotSpecified', country_code)), obj['NotSpecifiedData'])' is unsubscriptable (unsubscriptable-object)
E:118,15: Value 'filter(lambda x: set([x['OutAreaElspotId'], x['InAreaElspotId']]) == set([bidding_zone_a, bidding_zone_b]), obj)' is unsubscriptable (unsubscriptable-object)
Exited with code 2
```

corradio:
# ENTSOE parser:
https://github.com/tmrowco/electricitymap/pull/949/files#diff-413aa4ec1affb048245f2be81a7c3baaR491
The `filter` methods return iterables, therefore, `not len(iterable)` does not check if the list is not empty. Can you go through the ENTSOE parser and double check those places?
Other parsers might be affected by that, and this won't be caught by the linter.
Simple solution is to wrap in a `list()` method, that will cast the iterable to list and force it to run.

The errors you are getting for the AU and statnett parsers are of the same order.
You can either just convert the resulting object to list in order to force the iterable to run, or you can convert to list comprehension.

I've converted the first one, but the pattern should be straighforward:
python2:
```python
filter(lambda o: o['REGIONID'] == mapping['region_id'], response.json()['ELEC_NEM_SUMMARY'])
```
```python3
[o for o in response.json()['ELEC_NEM_SUMMARY'] if o['REGIONID'] == mapping['region_id']]
```

cclauss:
I like the list comprehension approach.  I believe that you were missing an: __[0]__

obj = [o for o in response.json()['ELEC_NEM_SUMMARY']__[0]__
           if o['REGIONID'] == mapping['region_id']]

cclauss:
I am getting confused on these filter() changes.  Too many moving parts.  Perhaps what would be best would be for you to make these filter() conversions in the current codebase and then let me know when you are satisfied that they work correctly.  I can then merge those changes into the current PR.  Your thoughts?

corradio:
Hmm are you sure? For me those filter changes are quite straightforward.
What moving parts are you talking about? Ping me on Slack so we can fix this quickly.
I already have the backend ready to do a full end-to-end test.

cclauss:
Usually when I do these conversions, I try to do them in an incremental fashion so I can make sure I am not breaking the Python 2 as I go.  Modifying 53 files in one PR is a lot of moving parts and if I break something, it will be difficult to isolate the change that caused the troubles.  I figured that the filter() changes were easy for you.  I will head out now to enjoy the snow for several hours.  When I return, I will try to reach you on Slack.

corradio:
On further investigation, let's not dive into style issues right now.
The only thing you need to do is to wrap failing filters inside a `list()` call to force execution, and that should fix things.
We'll make sure to move to list comprehension as time goes.

Enjoy the snow ❄️

cclauss:
`grep iteritems parsers/*.py`

corradio:
```
❯ grep iteritems parsers/*.py
parsers/CL_SING.py:        mapped_plants = [(plant_map.get(plant, 'unknown'), val) for plant, val in datapoint.iteritems()]
parsers/CR.py:        for index, value in df[column].iteritems():
parsers/DO.py:        mapped_plants = [(thermal_plants.get(plant, 'unknown'), val) for plant, val in current_plants.iteritems()]
parsers/MY_WM.py:    mapped_generation = [(fuel_mapping.get(gen_type, 'unknown'), val) for gen_type, val in current_generation.iteritems()]
parsers/NI.py:    total = sum([v for k, v in d['production'].iteritems()])
parsers/PE.py:    if sum([v for k,v in datapoint['production'].iteritems() if v is not None]) > 0 \
parsers/US_NEISO.py:        for k, v in datapoint.iteritems():
parsers/US_NY.py:        mapped_types = [(mapping.get(k, k), v) for k, v in item.iteritems()]
```

jarek:
I can also help with Py2/3 differences, including sorting out Unicode on the parsers.

Re: `iteritems()`: I would say this can be just replaced with `items()`. Technically speaking `items()` in Python 2 is slower (returns a copy rather than an iterator) but at scales of data we're dealing with, any difference is likely to be dwarfed by time taken to connect to external servers around the world. And having just `items()` will work in both Pythons without needing compatibility libraries and whatnot. Only catch is we might need to wrap things in list-evaluations if code elsewhere cannot deal with generators which Py3 will return.

I know nothing about CircleCI though.

jarek:
Easy fixes first:

```
diff --git a/parsers/PA.py b/parsers/PA.py
index 406baa0..e77c629 100644
--- a/parsers/PA.py
+++ b/parsers/PA.py
@@ -39,6 +39,7 @@ def fetch_production(country_code='PA', session=None):
     url = 'http://sitr.cnd.com.pa/m/pub/gen.html'
     response = r.get(url)
+    response.encoding = 'utf-8'
     html_doc = response.text
     soup = BeautifulSoup(html_doc, 'html.parser')
     productions = soup.find('table',{'class':'sitr-pie-layout'}).find_all('span')
@@ -57,7 +58,7 @@ def fetch_production(country_code='PA', session=None):
     }
     for prod in productions:
         prod_data = prod.string.split(' ')
-        production_mean = map_generation[prod_data[0].encode("latin-1")]
+        production_mean = map_generation[prod_data[0]]
         production_value = float(prod_data[1])
         data['production'][production_mean] = production_value
 
diff --git a/parsers/PE.py b/parsers/PE.py
index 71c83fd..242aebf 100644
--- a/parsers/PE.py
+++ b/parsers/PE.py
@@ -20,7 +20,7 @@ MAP_GENERATION = {
 }
 
 def validate(datapoint):
-    if sum([v for k,v in datapoint['production'].iteritems() if v is not None]) > 0 \
+    if sum([v for k,v in datapoint['production'].items() if v is not None]) > 0 \
         and datapoint['production'].get('gas', None) is not None:
         return datapoint
     else:
@@ -46,7 +46,7 @@ def fetch_production(country_code='PE', session=None):
     datetimes = []
 
     for serie in obj:
-        k = serie['Name'].encode('utf-8')
+        k = serie['Name']
         if not k in MAP_GENERATION:
             raise Exception('Unknown production type %s' % k)
         for v in serie['Data']:
```

cclauss:
We are going whole hog on this PR... Dropping Python 2 support altogether and __only__ supporting Python 3.

jarek:
To make UY.py work in Python 3 (but not Python 2):

```
(venv3)jarek@x1:~/projects/electricitymap/parsers$ git diff UY.py
diff --git a/parsers/UY.py b/parsers/UY.py
index 779612e..42353ca 100644
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -42,7 +42,7 @@ def parse_page(session):
         key = tds[0].find_all('b')
         # Go back one level up if the b tag is not there
         if not len(key): key = tds[0].find_all('font')
-        k = key[0].contents[0].encode('utf-8')
+        k = key[0].contents[0]
 
         value = tds[1].find_all('b')
         # Go back one level up if the b tag is not there
```

Similarly GB_NIR, this seems to work in Python 3 as well as 2:

```(venv3)jarek@x1:~/projects/electricitymap/parsers$ git diff GB_NIR.py
diff --git a/parsers/GB_NIR.py b/parsers/GB_NIR.py
index bd65ebb..2e12e6d 100644
--- a/parsers/GB_NIR.py
+++ b/parsers/GB_NIR.py
@@ -88,7 +88,7 @@ def create_thermal_df(text_data):
     """
 
     cols_to_use = [0,1,2,3,4,5]
-    df_thermal = pd.read_csv(StringIO(text_data.decode('utf-8')), usecols = cols_to_use)
+    df_thermal = pd.read_csv(StringIO(text_data), usecols = cols_to_use)
     df_thermal.fillna(0.0, inplace = True)
 
     return df_thermal
@@ -100,7 +100,7 @@ def create_wind_df(text_data):
     """
 
     cols_to_use = [0,1]
-    df_wind = pd.read_csv(StringIO(text_data.decode('utf-8')), usecols = cols_to_use)
+    df_wind = pd.read_csv(StringIO(text_data), usecols = cols_to_use)
     df_wind.fillna(0.0, inplace = True)
 
     return df_wind
@@ -111,7 +111,7 @@ def create_exchange_df(text_data):
     Turns exchange csv data into a usable dataframe.
     """
 
-    df_exchange = pd.read_csv(StringIO(text_data.decode('utf-8')))
+    df_exchange = pd.read_csv(StringIO(text_data))
     df_exchange.fillna(0.0, inplace = True)
 
     return df_exchange
```

systemcatch:
Nice @jarek I've submitted a PR containing those fixes to @cclauss's branch.

systemcatch:
Second PR removing all iteritems() is done.  @corradio when I run 2to3 on ENTSOE some changes are suggested but since you have the API key I'll leave that for you to do.

```
--- ENTSOE.py	(original)
+++ ENTSOE.py	(refactored)
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 from bs4 import BeautifulSoup
 import arrow, os, re, requests
@@ -26,7 +26,7 @@
     'B19': 'Wind Onshore',
     'B20': 'Other',
 }
-ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in ENTSOE_PARAMETER_DESC.items()}
+ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in list(ENTSOE_PARAMETER_DESC.items())}
 # Define all ENTSOE country_code <-> domain mapping
 ENTSOE_DOMAIN_MAPPINGS = {
     'AL': '10YAL-KESH-----5',
@@ -413,7 +413,7 @@
     # Create a double hashmap with keys (datetime, parameter)
     production_hashmap = defaultdict(lambda: {})
     # Grab production
-    for k in ENTSOE_PARAMETER_DESC.keys():
+    for k in list(ENTSOE_PARAMETER_DESC.keys()):
         parsed = parse_production(query_production(k, domain, session, now))
         if parsed:
             productions, datetimes = parsed
@@ -423,18 +423,16 @@
 
     # Remove all dates in the future
     production_dates = sorted(set(production_hashmap.keys()), reverse=True)
-    production_dates = filter(lambda x: x <= arrow.now(), production_dates)
+    production_dates = [x for x in production_dates if x <= arrow.now()]
     if not len(production_dates): return None
     # Only take fully observed elements
-    max_counts = max(map(lambda date: len(production_hashmap[date].keys()),
-        production_dates))
-    production_dates = filter(lambda d: len(production_hashmap[d].keys()) == max_counts,
-        production_dates)
+    max_counts = max([len(list(production_hashmap[date].keys())) for date in production_dates])
+    production_dates = [d for d in production_dates if len(list(production_hashmap[d].keys())) == max_counts]
 
     data = []
     for production_date in production_dates:
 
-        production_values = {ENTSOE_PARAMETER_DESC[k]: v for k, v in production_hashmap[production_date].items()}
+        production_values = {ENTSOE_PARAMETER_DESC[k]: v for k, v in list(production_hashmap[production_date].items())}
 
         data.append({
             'countryCode': country_code,
@@ -457,7 +455,7 @@
             'source': 'entsoe.eu'
         })
 
-    return filter(validate_production, data)
+    return list(filter(validate_production, data))
 
 def fetch_exchange(country_code1, country_code2, session=None, now=None):
     if not session: session = requests.session()
@@ -487,7 +485,7 @@
 
     # Remove all dates in the future
     exchange_dates = sorted(set(exchange_hashmap.keys()), reverse=True)
-    exchange_dates = filter(lambda x: x <= arrow.now(), exchange_dates)
+    exchange_dates = [x for x in exchange_dates if x <= arrow.now()]
```

cclauss:
Looking for advise on solutions to the __test_ESIOS.py__ json issue at the end of https://circleci.com/gh/tmrowco/electricitymap/1887

systemcatch:
resource_string returns bytes, making this change fixes the test.
```
json_data = resource_string("parsers.test.mocks", "ESIOS_ES_MA.json")
json_data = json_data.decode("utf-8")
self.adapter.register_uri(ANY, ANY, json=loads(json_data))
```

systemcatch:
Now that all the tests are fixed I've ran all the parsers and these are the last changes needed.

DO - L173 list needed on range
ENTSOE - various changes still needed
statnett - L137 map object not subscriptable

Also update_state.py and server.py need some small fixes.

cclauss:
> I've ran all the parsers

How do I do that locally or in Travis?  I have made the DO.py and statnett.py changes but I am unclear on the others (ENTSOE, update_state.py, server.py).

corradio:
@cclauss you need to run them manually, there's no option to run them all.
I'll submit a PR to your fork with latest changes (hopefully).

mockserver/update_state.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
corradio: do we still need those?

cclauss: They prevent Python 2 contributors from putting syntax error into Python 3 code.  They also enable the __end__ parameter in Python 2 and prohibit print statements with trailing comma.  If you are 100% certain that no one will try any of these things or paste in code that tries any of these things, then we can take the future imports out.

corradio: If we switch completely to Python3 then this should not be necessary, so we can go ahead and remove them!

systemcatch:
@cclauss update_state is done, I've submitted a PR to your branch.  It looks like server.py has had some of it's modules move around for python 3 but I won't be able to check that out any deeper until tomorrow evening.

cclauss:
@systemcatch Sorry but I believe that I lost your PR on my branch.  Please check.

@corradio Let's leave the future imports in place.  It is clear that contributors are still writing Python 2 that is getting merged into master because there are no safeguards in place to prevent them from doing so.

corradio:
@cclauss I still believe we should remove support for Python2 completely.
In the PR I submitted you, I changed the README to explain that we are using py3, and I changed the python command to test parsers such that it runs the `python3` executable.
The last safeguard we have is actually the pull request that I will validating, so I think the change is safe.
I have tested the backend and all parsers seem to work. If you agree to drop the future imports, then I can do a final review/test and we can merge?

systemcatch:
@cclauss nice job managing all the fixes/merges/changes.  This PR escalated quickly!

cclauss:
Yes.  Shoulda been done in bite size increments.

PR #953, opened 2017-12-27
Add Moldova

### LINKED ISSUE #926: Add Moldova
Hey guys,

another part of the puzzle: I just found a site with accessible data for generation and cross-border exchanges for **Moldova**.

http://www.moldelectrica.md/ro/activity/system_state

image

The best thing about it: **it refreshes every minute** and provides the latest numbers.

**POWER PLANTS**:
Numbers next to the symbols represent the electricity generation of the following plants:
- The two NHE are hydro power plants,
- the three units displayed as "CET" run on natural gas and are CHPPs, 
- and the "CERS" power plant may run on multiple fuels, such as gas, coal or oil.

**EMISSION FACTORS**:
For the **unknown CERS fuel-mix**, I would suggest using the "other" category with **700 g/kWh**.

For the **CETs**, I found a [document](https://www.google.de/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwjhvbnshpLYAhWIEVAKHaIIACQQFgguMAA&url=http%3A%2F%2Fget-moldau.de%2Fdownload%2Fpolicypapers%2F2010%2FGET%2520Moldova_PP%252001%25202010_en.pdf&usg=AOvVaw37WEhiS1RHEn9lR52Ono--) saying the CET's efficiency may be as low as 22% or lower:
 
image

The emission factor should be higher than for modern efficient gas power plants. Based on the efficiency of 22% and released emissions of ca.[ 220 g/kWh(gas)](http://www.klimaneutral-handeln.de/php/kompens-berechnen.php#rechner) **a country-specific gas emission factor of 1,000 g/kWh(el)** for generation of the CET-units might be applied. 

Seems like Estonia and Poland get another carbon-intensive companion ;-)

**INSTALLED CAPACITIES** (presentation from December 2016):

- hydro: 64 MW
- natural gas CHPP (CET): 335 MW
- other (CERS-mix): 2,520 MW 

image

There seem to be some industrial power plants with about 100 MW, and some renewables as PV and wind, but they may be neglected for now, as their generation is unknown anyways.

**CROSS-BORDER FLOWS**
Interconnectors to Ukraine and Romania are also part of the scheme ("Fluxurile MD-UA" and "Fluxurile MD-RO"), where positive values are imports and negative values are exports.

**MISC**
Moldova was previously mentioned in #575 in July, where it says, entso-e were working on implementing data from Moldova in near future. So far, there are no data on entso-e yet. 
So this source should work fine as well, as it is provided by the national TSO.

Greetings, Alex

systemcatch:
Another piece of the jigsaw.  Exchanges are implemented and working as well.

Thanks to @alixunderplatz for helping find the data source and map the plants to type.

Fixes #926  

moldova


alixunderplatz:
AWESOME!!!
Thank you @systemcatch for putting the pieces together and writing the parsers!!! **Great work!!!**

parsers/MD.py
```diff
@@ -0,0 +1,139 @@
...
+          'unknown': 0.0
+      },
+      'storage': {
+          'hydro': -10.0,
+      },
+      'source': 'mysource.com'
+    }
+    """
+
+    grid_status = get_data(session = None)
```
corradio: are you sure you want to pass `None` as session here?

parsers/MD.py
```diff
@@ -0,0 +1,139 @@
...
+          'biomass': 0.0,
+          'coal': 0.0,
+          'gas': 0.0,
+          'hydro': 0.0,
+          'nuclear': null,
+          'oil': 0.0,
+          'solar': 0.0,
+          'wind': 0.0,
+          'geothermal': 0.0,
+          'unknown': 0.0
```
corradio: All entries where numbers are set to 0 represent a **known** zero. Are you sure e.g. there's no wind production? Else, better put `None` everywhere and let the parser fill in the blanks.

systemcatch: That's in the docstring not the actual parser dict.  I have set wind and solar to None further on.

parsers/MD.py
systemcatch: session = None is the default argument for get_data() so I don't think it matters.  I've remove it if you prefer though.

parsers/MD.py
corradio: wow sorry my bad!

parsers/MD.py
corradio: The parser, in production, will be called with a non `None` session argument, so it's important to propagate it.

parsers/MD.py
```diff
@@ -0,0 +1,139 @@
...
+      },
+      'storage': {
+          'hydro': -10.0,
+      },
+      'source': 'mysource.com'
+    }
+    """
+
+    grid_status = get_data(session = None)
+    production = {'solar': None, 'wind': None, 'biomass': None}
```
systemcatch: I've set these 3 keys to None as there is a very small amount of generation for each that is not reported.  Nuclear can be set to 0.0 as there is definitely no generation.

parsers/MD.py
systemcatch: Ok so I'll leave it in?

corradio: In order to propagate the value of session, we need `get_data(session=session)`.

parsers/MD.py
corradio: 👍

alixunderplatz:
@systemcatch @corradio 
One brief remark: 
In #926, I mentioned that electricity production from natural gas in Moldova happens with an efficiency of ~22%, which leads to CO2-emissions of ~1000 g/kWh. However, this efficiency was based on numbers from 2008, but I don't think that more efficient gas units were put into operation in the past decade in MD.
We are using the standard factor for gas for MD. Should we add this country specific factor or leave it as it is?

corradio:
@alixunderplatz can you submit a new issue where you aggregate all that info? Then @brunolajoie will be able to correct the emission factors if need be.

Issue #357, opened 2017-02-09
Redraw wind/solar on resize


corradio:
..possibly with a debounce

systemcatch:
Were you able to fix this in the Mapbox PR?

corradio:
nope. I think this is easy to fix though.

maxbellec:
should this still be opened? Wind and solar get updated when I resize.

corradio:
Indeed I can't reproduce. I love it when issues close themselves

PR #958, opened 2017-12-29
Mapbox

### LINKED ISSUE #499: Add arrow navigation

### LINKED ISSUE #504: Add zoom in / zoom out button
Hello,

With the new map, we start with an unzoomed view. By double clicking somewhere, it zooms in. Then, you cannot zoom out. 

We could add a zoom in / zoom out button somewhere, to help with the navigation.

### LINKED ISSUE #670: Rewrite map layer
This should speed up performances considerably.
A good strategy seems to use a big canvas (world), and then only draw parts of the canvas which correspond to the active screen. One could pre-draw the (invisible) surroundings to ensure smooth transition. Of course, we should remember what has been drawn in order not to redraw things.
We need a fast algorithm in order to detect mouseovers and clicks. https://github.com/d3/d3/blob/master/API.md#quadtrees-d3-quadtree could be used.

This would improve the drawing speed, but would not reduce the need to transfer, and load in memory, all geojson polygons.
We could generate tiles as GMaps does, but that would probably make mouse hit tests complicated, as we need the coordinates of zones. Moreover, it would complicate the interactivity use cases (change color of hovered zone etc..)

### LINKED ISSUE #965: Tapping on a region makes the hover tooltip appear over the detail view

### LINKED ISSUE #980: Exchange arrows for 0 MW import/export are confusing
This doesn't have any influence on the data, but is there a way to make the exchange arrows disappear when export/import between countries is 0 MW? 
Is it because zero is considered a positive number, so the arrow is pointing in the positive/default direction?

image

image

corradio:
List of changes:
- using mapbox as map layer (based on webgl), thus improving drag/zoom FPS (fix #670)
- added arrow navigation (fix #499)
- added zoom controls (fix #504)
- rewrote solar / wind / exchange map layer APIs
- exchange animations are now CSS, based on two layered images
- exchange animations pause (and are hidden) during drag/zoom, which improves performances
- d3 loaded selectively in newly written modules
- newly written modules use ES6 `class`
- fix some performance bugs in the wind drawing code
- fix pinch intensity (fix https://github.com/tmrowco/electricitymap/issues/513)
- hide arrows with low flow (fix #980)

Testable at https://staging.electricitymap.org

List of devices tested and associated performance improvements (please add yours):
(I use ✅ to indicate it works, and 🐇 to indicate it's fast)
- iPad (non-retina) ✅ 🐇 (major improvement)
- iPad 3rd gen (retina) ✅ 🐇 (major improvement)
- iPhone 5S ✅ 🐇 (noticeable improvement)
- Nexus 6P ✅ 🐇 (panning is much smoother)
- Nexus 5X ✅ 🐇 
- Nexus 4 ✅ 🐇 (looks better)
- Thinkpad T430 ✅ 🐇
- HTC One M9 ✅ 🐇
- Samsung S7 + chrome : ✅ 🐇 
- Samsung Galaxy Tab S + chrome: ✅ Drag 🐇, Wind 🐢 

List of desktop navigators tested on
- Chrome MacOS ✅ 🐇 
- Chrome PC ?
- Safari (major improvement) ✅ 🐇 
- Firefox MacOS (major improvement) ✅ 🐇 
- Firefox PC ? 
- Firefox Mobile ✅ 🐇 
- Firefox Linux: ✅ 🐇 (noticeable improvement)
- IE, Edge ✅ 

List of remaining tasks:
- [x] Left panel exchange arrow missing
- [x] Zone colouring during mouseover is (very) slow
- [x] Double check loading indicators
- [x] Fix map drawing bugs
- [x] Race condition when loading on mobile
- [x] Pinch zoom on iPhone 5S does not trigger events
- [x] Rotating is not disabled on mobile
- [x] On Safari mobile old ipad, after country click, header shouldn't resize
- [x] Hide arrows before map loads
- [x] Compare perfs (speed + size of assets) of wind + GIF vs wind + CSS arrows
- [x] When loading on mobile, half the screen is draggable/drawn
- [x] Reduce size of arrows
- [ ] ~Wind layer resize bug on mobile landscape~ (duplicate of #357)
- [x] Blurry exchange arrows on zoom on Safari mobile
- [x] Arrows sizes are not similar
- [x] Arrows should not change opacity when hovered
- [x] Firefox/Chrome mobile layout bug
- [x] Webpack bug:
image
- [x] Twitter error:
image
- [x] Tooltip error (fix #965):
image
- [x] on old devices without webgl, the map should be hidden instead of throwing `Failed to initialize WebGL`


jarek:
Kudos for setting this up, looks great.

Firefox 57 mobile, Nexus 6P: panning is much smoother. On initial load, arrows appear before the underlying map. I don't know if it's related to the changes, but tapping a region makes the hover tooltip appear over the detail view, it goes away after being tapped. It looks like the animation of the exchange arrow goes a bit further than the underlying arrow. I wonder if map weirdness is caused by parts of Russia that cross latitude 180?

systemcatch:
Performance seems noticeably better on my older T430 Thinkpad running Ubuntu 14.04.  I really like the country hover being a lighter grey.  On Firefox for Android there is an improvement in panning speed but when I tap a country then close it takes a while to deselect the country on map view.  That may just be because my phone (HTC M9) is older.

corradio:
Noted. Thanks for testing.
Russia is a bit annoying and I do think it's because it crosses lat 180.

jarek:
Hovering on carvon-intensive countries (currently Estonia, Poland) makes the highlight colour somewhat confusingly close to the "no data" grey colour

jarek:
Tested in Edge 41 on Windows 10. Performance seems same as in Chrome. Found one bug, see screenshot

edge bug

brunolajoie:
Arrows are still too big when zooming in (it will prevent nice screenshot for journalists or other) 
image

brunolajoie:
Bug found on my samsung S7 with chrome
image

brunolajoie:
Tablet Samsung Tab S: When clicking on a country, it stays hover and an orange border appears. it only disappear when clicking somewhere else. 
image

corradio:
Noted. Will fix.

brunolajoie:
Macbookpro 2016 OS X Chrome: Wind arrows are twice slower

corradio:
@brunolajoie when you say slower, do you mean speed or CPU? and do you mean that the new version is slower? Is is a bad thing or a good thing?

brunolajoie:
I meant that on the new version, on my computer, wind arrows are moving at least twice slower, and with non-negligible saccade that makes the wind looks ugly :/

corradio:
All set! I've updated the staging website to the latest fixes.
@brunolajoie @jarek @systemcatch can you quickly review?

brunolajoie:
Another issue : color blind mode is not changing map colors (except the legend)

corradio:
Excellent catches @brunolajoie !
I've fixed both - and deployed to staging.

brunolajoie:
Hey @scarlac ! We need your front end performance expertise here :) 

I see a deterioration of the graphical performance due to aliasing on my 22 inch, medium-res screen that no one else seems to see on higher ppi screens: have a look & try it yourself:

- staging.electricitymap.org
image

- electricitymap.org
image

Alising is visible on country and arrows, but is especially visible on arrows. Would you happen to have an idea why alising appears on arrows? We did not change anything specifically about arrows in this PR, it's still the same gif! Yet they now appear aliased

scarlac:
Hey @brunolajoie ! Yeah I see they are pixelated. Testing with Firefox 57 (aka Quantum) in "Low Res" mode on macOS, the fix seems to be to apply `translateZ(.01px)` to the transform.

Take a look here:
https://www.useloom.com/share/adfeae95093d487b97f6a7dc45ee7b0f

scarlac:
I should add that I see the same issue on the production site.

jarek:
I'm seeing a bug with top bar on Firefox 57 mobile/Android, see screenshots

screenshot_20180105-221530

screenshot_20180105-221545

corradio:
@brunolajoie can you try the fix Seph mentions and post two screenshots to see if it helps?
@jarek I'll take a look at the bug.

systemcatch:
One small point, on the production map double click lets you zoom in.  On staging this doesn't happen though it looks like the arrows refresh.

corradio:
@systemcatch I disabled double tap zoom because I was having a bug with it.
Do you think it's important?

systemcatch:
@corradio It lets people zoom in/out easier than using the buttons on the top right, but I don't think it's that critical tbh.

brunolajoie:
@scarlac interestingly, the aliasing issue happens both on staging and production in firefox. Yet in chrome, aliasing only appears on staging, and the fix you suggested does not improve it.

scarlac:
bruno, Seems very odd but I was using retina but running in Low resolution mode which should be identical from a functional perspective.

Did you try increasing translateZ? The other solutions would be to move around scaling and transforming in hopes it’s an issue related to order of transforms. Perhaps moving transform strait to arrow image or parent. I don’t have a specific solution in mind though.

corradio:
@jarek bug fixed (I hope)

jarek:
Yep that bug seems fixed

corradio:
..and the last update: we revert to a simpler ui if webgl is not available

image

PR #963, opened 2017-12-31
Update german capacity

Maverick78de:
Updated german power plants 
https://www.bundesnetzagentur.de/DE/Sachgebiete/ElektrizitaetundGas/Unternehmen_Institutionen/Versorgungssicherheit/Erzeugungskapazitaeten/Kraftwerksliste/kraftwerksliste-node.html

PR #972, opened 2018-01-02
Update capacity source for Germany

Maverick78de:
Update capacity source for Germany as mentioned in #963


Maverick78de:
changed data source to "Federal Network Agency" of Germany (Bundenetzagentur)

PR #974, opened 2018-01-03
Upgrade Circle CI from Python 3.5 to 3.6

cclauss:
To match with mockserver https://github.com/tmrowco/electricitymap/blob/master/mockserver/Dockerfile#L1


corradio:
I don't think circle ci supports 3.6 yet :/

Issue #960, opened 2017-12-30
Use operational emissions instead of lifecycle emissions


HansHyde:
Just noting #926 "Moldova" discussion & the Type_Mapping for generators...

Would it be possible to create a sheet/file similar (possibly not "functional" at this time) to CO2eq_parameters.js, where we define power plant types (particularly thermal) by characteristics?  This could then be used as EM calculation methodology support (with required supporting documentation, not just _"source - 'Hans Hyde'"_) to back EM's CO2eq x kWh data aggregations.

These "energy conversion" characteristics of the power plant ultimately determine how much CO2 is produced per kWh by fuel type.  This is referred to as both "Heat Rate" and its mathematical inverse "Efficiency", and the point at which HR is optimal is under full, static power production --> highest kWh per unit energy input.  

Starting or Stopping, referred to as "ramping" returns an amount of CO2 without usable kWh produced based on the amount of fuel consumed during "ramping Up" & "ramping Down".  

- For so called "peakers", ramping occurs twice a day (almost like clockwork - and it has been happening this way for decades.
- Existing Pumped Storage does not exist today as a "pre-thought" to future intermittent renewable energy sources, it exists because of the "Nighttime Trough" - too much non-rampable generators (coal/nuclear) out producing the demand during the NT.  Without PS, transmission grids would burn down when generation far-exceeded load.

![daily_load_curve-typical](https://user-images.githubusercontent.com/33690988/34454003-771f9730-ed61-11e7-8461-5566f0fec963.jpg)
_Typical Daily Load (Generation) Curve - since the dawn of bulk electric power systems_

For anything using Steam as the prime mover [of the electric generator], this CO2 value is very high per fuel [excluding nuclear] and takes many hours to "ramp".  I honestly think we could capture these events within the backend coding from the realtime data streams.

A general observation, not specific to EM, is that plant types (if listed as abbreviations, or in different languages) are frequently misinterpreted by almost all in the "energy & electric power industry" - except for those that actually work with thermal generators themselves.  The IPCC 2014 Synthesis Report, Chapter 7 - Energy Systems (where CO2eq_parameters.js derives its values) IMHO steamsrolls over this, assigns a range almost generically based on fuel type (which works for low carbon generators to capture a "lifecycle value", but falls apart for thermal).  

![heatrate_co2emissions](https://user-images.githubusercontent.com/33690988/34454018-c793bfde-ed61-11e7-8c63-1d759d47125e.png)
[...]

PR #975, opened 2018-01-03
Create a FAQ section

systemcatch:
There's some recurring questions that people tend to ask about the electricitymap so I though it would be useful to start an FAQ section.  I've added some of things I think might be useful but will add anything else that @corradio  @jarek @brunolajoie think is important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,20 @@ You can also see a list of missing data displayed as warnings in the developer c
 
 Check the [contributing](#contribute) section for more details.
 
+##Frequently asked questions
+
+**How do you define real time data?**
+Real time data is defined as a data source that updates every 2 hours or better.  It should provide a breakdown by generation type.  Often fossil fuel generation (coal/gas/oil) is combined under a single heading like 'thermal' or 'conventional', this is not a problem.
+
+**Is non real time data useful?**
+While the map relies on having real-time data to work it's still useful to collect data from days/months past. This older data can used to show past emissions and build up a better dataset. So if there's a good data source that lags several days behind you can still build a parser for it.
+
+**Can scheduled generation data be used?**
+The electricitymap doesn't use scheduled or planned generation data. This is to avoid introducing extra uncertainty into emissions calculations.
+
+**Can areas other than countries be shown?**
+Yes providing there is a valid GeoJSON geometry (or another format that can be converted) for the area. As an example we already split several countries into states and grid regions.
+
 ## Data sources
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,20 @@ You can also see a list of missing data displayed as warnings in the developer c
 
 Check the [contributing](#contribute) section for more details.
 
+## Frequently asked questions
+
+**How do you define real time data?**
```
corradio: Spelling should be "real-time" (with a dash) in all the text I think.

README.md
```diff
@@ -19,6 +19,20 @@ You can also see a list of missing data displayed as warnings in the developer c
 
 Check the [contributing](#contribute) section for more details.
 
+## Frequently asked questions
+
+**How do you define real time data?**  
+Real time data is defined as a data source that updates every 2 hours or better.  It should provide a breakdown by generation type.  Often fossil fuel generation (coal/gas/oil) is combined under a single heading like 'thermal' or 'conventional', this is not a problem.
+
+**Is non real time data useful?**  
+While the map relies on having real-time data to work it's still useful to collect data from days/months past. This older data can used to show past emissions and build up a better dataset. So if there's a good data source that lags several days behind you can still build a parser for it.
```
corradio: Maybe insist on the fact it needs to be hourly at least?

README.md
```diff
@@ -19,6 +19,20 @@ You can also see a list of missing data displayed as warnings in the developer c
 
 Check the [contributing](#contribute) section for more details.
 
+## Frequently asked questions
+
+**How do you define real time data?**  
+Real time data is defined as a data source that updates every 2 hours or better.  It should provide a breakdown by generation type.  Often fossil fuel generation (coal/gas/oil) is combined under a single heading like 'thermal' or 'conventional', this is not a problem.
```
corradio: Maybe say that it is defined as a data source with an hourly (or better) frequency, delayed by less than 2hrs.

README.md
```diff
@@ -19,6 +19,20 @@ You can also see a list of missing data displayed as warnings in the developer c
...
+
+**Is non real time data useful?**
```
corradio: Maybe focus the title on the question of whether or not **delayed** data is still useful (instead of real-time)

systemcatch:
Sure I'll make the changes you've suggested.  Any other questions you want added?

corradio:
Maybe we could add a section about why we use lifecycle emissions, ref #960 ?
But I can add that later.

PR #997, opened 2018-01-11
Update ENTSO-E installed hydro capacities

alixunderplatz:
Some more countries now have data available for pumped storage generation. Installed hydro capacity was split for the relevant countries and hydro capacities were updated to the latest numbers.


corradio:
@brunolajoie I'll let you review/merge

config/zones.json
```diff
@@ -642,7 +644,7 @@
       "coal": 9535,
       "gas": 30723,
       "geothermal": 0,
-      "hydro": 14614,
+      "hydro": 20281,
```
brunolajoie: Where did you get this new figure (+6GW compared to previous)? I think (although i'm not sure) that I had chosed to used IRENA data cause spain entose hydro capacity figure was incorrect (see readme)

config/zones.json
```diff
@@ -525,8 +527,8 @@
       "coal": 46336,
       "gas": 29731,
       "geothermal": 39,
-      "hydro": 3945,
-      "hydro storage": 6357,
+      "hydro": 4671,
```
brunolajoie: How are we sure that these capacity are closer to the truth than what was there initially? (readme quotes a federal agency figure, but i haven't looked in detail).

alixunderplatz: For Germany, readme quotes Bundesnetzagentur (BNetzA), which actually should have all the data.
I downloaded the list of power stations [(Kraftwerksliste)](https://www.bundesnetzagentur.de/SharedDocs/Downloads/DE/Sachgebiete/Energie/Unternehmen_Institutionen/Versorgungssicherheit/Erzeugungskapazitaeten/Kraftwerksliste/Kraftwerksliste_2017_2.xlsx) and checked.
In one cell it says "Hinweis: In das deutsche Netz einspeisende Kraftwerksleistungen in Luxemburg, Schweiz und Österreich sind ebenfalls aufgeführt."
"Please note: Generating capacity feeding into the German grid from LUX, CH and AT are listed here as well."

Some of the TSO's area (I think Amprion) expands a bit over to Austria and Luxemburg.
Pumped Storage "Vianden" is located in Luxembourg right next to the border of Germany, but its production goes entirely to Germany. It's a part of the German market, connected to the German grid only and is also reported there. So are some of the generators located in western Austria.

6357 MW is the capacity of the Pumped Storages located in Germany exclusively according to BNetzA, without LUX and AT. If they are a part of the market and deliver to Germany, the proper number from BNetzA is 9440 MW. Entso-e says 8733 MW. Don't know if unavailabilities due to maintenance are included in ENTSO-E (would be more precise) and BNetzA reports entire installed capacities instead.
That's also why general "hydro" is higher.
Seems like the data from BNetzA was not properly/not entirely used. Checking if data is excluded on entso-e (so it will not double and appear in AT or LUX or CH) could be a way to find out. I'm sure at least for LUX it is not included in their installed capacities on entso-e.

config/zones.json
alixunderplatz: @brunolajoie 
Yeah, I was about to check the readme for sources and any different sources than ENTSO-E. Then github was offline for a few minutes and I simply forgot about it. 
For Spain, I just saw the annotations below the chart on entso-e.
Pumped consumption capacity is included in "Pumped storage", but generating capacity is assigned to "water reservoir" ;) I guess that's why we have ~+5.6 GW??!?! 
It says on entso-e: 1156 MW run-of-river and 19125 MW hydro reservoir (which is including about 5600 MW pumped storage gen capacity which might not be included in IRENA (?) data or wherever. By the way: in the readme it says ENTSO-E for Spain?)

image

I am confused now. :D Do I keep the new value for pumped storage as it is so that consumption is shown as ratio of installed capacity? Or do I better leave it blank?

alixunderplatz: and: I will check and update the readme tomorrow!

config/zones.json
brunolajoie: 🔍  Awsome job 🔍  Please add a condensed version of these details in "_comments" so we don't loose track of it for later on!

config/zones.json
brunolajoie: Readme says Spain hydro capacity is currently IRENA.  Unless proven wrong, I'd rather keep using IRENA as long as these confusing comments from ENTSEO are still there xD

alixunderplatz: IRENA is mentioned in the "real time generation"-part for Spain and not in "installed capacity"-part. That's why I had overseen it there. I will fix this one, too, and copy it to capacities ;)

alixunderplatz: @brunolajoie Maybe you can take another look at the data on IRENA for Spain.
It also has Pumped storage (you have to select it manually)

image

I also couln't  manage to calculate the currently installed "14614 MW" using the numbers from IRENA, maybe there is another error somewhere?

Latest available data for 2016 from IRENA for Spain:
Pumped storage: 3 280 MW
Large hydro: 12 133 MW
Medium hydro: 1 673
Small hydro: 280 MW
Mixed hydro: 2 687 MW

Sum of all hydro: 20 053 MW which is very close to ENTSO-E numbers.

brunolajoie: IRENA figures looks good then! But on your PR, hydro+storage tops up 25GW so it was a bit too high

config/zones.json
```diff
@@ -644,8 +646,7 @@
       "coal": 9535,
       "gas": 30723,
       "geothermal": 0,
-      "hydro": 20281,
-      "hydro storage": 5645,
+      "hydro": 20053,
```
brunolajoie: I'm sorry to be picky, but why don't we breakdown spain storage capacity? It seems that Entsoe gives us live data for it.

alixunderplatz: @brunolajoie Be picky, it's absolutely okay! :) I wasn't sure anyway whether I should ask you again to make things totally clear for Spain, because I still had some doubts myself :)

I had a look at the generation and capacity data on entso-e, and here is the reason: 

I would not split Spanish hydro, because the generation from "pumped storage" is reported within "hydro reservoir". There would be absolutely no generation from PS shown in the map, it would be included in the usual "hydro" category. Generation from PS is always empty on entso-e for Spain.

However, the consumption is reported for pumped strorage. You can see it in the negative area of the PS category at night. So if we had the capacity for "pumped storage" on the map, there will only be the negative consumption values, the positive side would never be filled, just like it is now.

I want to point at entso-e once again, even if that is not the source of the data for Spain now:

image

So according to entso-e, the generation capacity from **run-of-river** plus **hydro reservoir** is 20281 MW. This is also including the **generation** capacity of pumped storage. 
That is probably a result of using **some of the hydro reservoirs as pumped storages** as well, so there is some natural water intake throughout the entire year and a partial pumped storage water intake for a few hours during "pumped storage consumption". In the IRENA numbers, they might be included in the "mixed hydro" category.

Listing the pumped storages capacity seperately would create the impression that Spain has about 20 GW of hydro capacity and another 5.6 GW of pumped storage (or 3.28 pumped storage according to IRENA), that are already included in the ~20 GW.
That is why I assigned 0 to pumped storage, because only the consumption does appear in the graph. Generation is included in "hydro", therefore the "more correct" capacity is ~20 GW.

We only need the negative part of the pumped storage capacity according to the entso-e generation data. I think, this is probably not possible to realize at the moment?

brunolajoie: Super clear, thanks! Indeed, we havent programmed yet the ability to distinguish storage capacity from discharge capacity. Lets go with this!

PR #998, opened 2018-01-11
Create module and method docstrings for ENTSOE

systemcatch:
Let me know if any need changing.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -127,7 +127,7 @@ def check_response(response, function_name):
 def query_ENTSOE(session, params, now=None, span=[-24, 24]):
     """
     Makes a standard query to the ENTSOE API with a modifiable set of parameters.
-    Allows an existing session to be passed. 7Default timespan of +/- 24 hours.
+    Allows an existing session to be passed. Default timespan of +/- 24 hours.
     Raises an exception if no API token is found.
     Returns a request object.
     """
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -1,5 +1,23 @@
...
+Production
+Exchanges
+Exchange Forecast
+Day-ahead Price
+Generation Forecast
+Consumption Forecast
+
+The zones below use methods from this parser.
+
+['AT', 'BA', 'BE', 'BG', 'CH', 'CZ', 'DE', 'DK', 'DK-DK1', 'DK-DK2', 'EE', 'ES',
```
corradio: Can we leave this out? The info is already in zones.json

systemcatch: The list or the whole module docstring?

corradio: just the list (I prefer having only one place to maintain instead of two).
Good initiative by the way!

systemcatch: Thanks, I'll change it.

parsers/ENTSOE.py
```diff
@@ -133,7 +127,7 @@ def check_response(response, function_name):
 def query_ENTSOE(session, params, now=None, span=[-24, 24]):
     """
     Makes a standard query to the ENTSOE API with a modifiable set of parameters.
-    Allows an existing session to be passed. Default timespan of +/- 24 hours.
+    Allows an existing session to be passed. 7Default timespan of +/- 24 hours.
```
corradio: typo?

systemcatch: indeed

PR #999, opened 2018-01-13
Set production by storage battery in Spain islands.

blackleg:
Setted that production by storage battery is zero in Canary Islands and Balearic Islands electricity systems.
Nowadays this islands haven't any system based in storage batteries.


blackleg:
@corradio Updated.

Issue #928, opened 2017-12-18
Get historical data for PJM


HansHyde:
**Parser** - ?
**Name** - PJM
**Geometry** - PA + OH + MD + WV + VA + NJ + DE (Delware)
- https://www.eia.gov/todayinenergy/detail.php?id=27152

**Exchanges**
- NYISO - 4 or 5 (both PJM & NYISO have accessible RTD)
- PJM shows price ($), NYISO show volume (MWH)
- PJM exchanges (all) - http://www.pjm.com/markets-and-operations/interregional-map.aspx 
- TBD

**Gen Mix** - http://www.pjm.com/markets-and-operations.aspx

corradio:
Unfortunately we need real-time generation data in order to add a new zone.
Feel free to re-open if you find real-time generation data!

HansHyde:
Thanks @corradio 

Can you or @systemcatch point me to the area of the code that represents the exchanges between the parsers pls?  Either Ontario or NEISO might be a good example???

corradio:
Actually a parser is responsible to fetch both generation mix and exchanges.
More info at https://github.com/tmrowco/electricitymap#adding-a-new-country (checkout the example file linked from there)

HansHyde:
@corradio ahhhh... I'm scared, I added gen capacity for PJM to the zones.json (on my local fork), but cautious about "pushing" it up.... as I don't want it to mess the master.

I think I have everything set up & linked together... python, Atom, Github, Docker

![added pjm to zones-json](https://user-images.githubusercontent.com/33690988/34315371-05c9248a-e77f-11e7-8e19-e906d6ca8059.png)

HansHyde:
PSS - @corradio if you were able to create a basic or template parser for US_PJM, I could mess around with it.  I'm looking at @systemcatch US_NYISO & @jarek CA_ON as examples, but they are two completely different code "builds" from what I can tell.

Is it ok for me to "commit to master"?  I assume, you have to approve it before it enters the Master?

![commit to master](https://user-images.githubusercontent.com/33690988/34315563-aa2c66ca-e781-11e7-86cf-45b1b77b9b46.png)

corradio:
I'll reply on Slack in order to keep this thread clean. Looking good however!

HansHyde:
Hey @corradio , Merry Christmas!  

[...]

PR #1010, opened 2018-01-16
Add New York exchange with PJM 

systemcatch:
These general functions will also allow us to easily add many other exchanges (e.g. MISO) whenever we have the data/zone ready.

ref #928 

pjm_ex


systemcatch:
fast merge!

Issue #984, opened 2018-01-07
Bolivia data handling needs to be improved


systemcatch:
Looks like there is sometimes an extra key present which is missing from the mapping dict.  This causes the parser to fail constantly.  We need to add this key to the mapping then setup error handling so that any new keys raise warnings in the logger rather causing total failure.

Good first issue for someone looking to get involved.
```
chris@ThinkPad:~/electricitymap/parsers$ python3 BO.py 
fetch_production() ->
Traceback (most recent call last):
  File "BO.py", line 155, in <module>
    print(fetch_production())
  File "BO.py", line 103, in fetch_production
    data_temp = fetch_hourly_production(country_code, obj, h, formatted_date)
  File "BO.py", line 68, in fetch_hourly_production
    data['production'][i_type] = obj[MAP_GENERATION[i_type]][obj.hour == hour].iloc[0]
  File "/usr/local/lib/python3.4/dist-packages/pandas/core/frame.py", line 2139, in __getitem__
    return self._getitem_column(key)
  File "/usr/local/lib/python3.4/dist-packages/pandas/core/frame.py", line 2146, in _getitem_column
    return self._get_item_cache(key)
  File "/usr/local/lib/python3.4/dist-packages/pandas/core/generic.py", line 1842, in _get_item_cache
    values = self._data.get(item)
  File "/usr/local/lib/python3.4/dist-packages/pandas/core/internals.py", line 3843, in get
    loc = self.items.get_loc(item)
  File "/usr/local/lib/python3.4/dist-packages/pandas/core/indexes/base.py", line 2527, in get_loc
    return self._engine.get_loc(self._maybe_cast_indexer(key))
  File "pandas/_libs/index.pyx", line 117, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 139, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1265, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1273, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'Intermitentes'
```

systemcatch:
So this isn't due to a Key missing.  When there is no generation for Intermitentes (wind) instead of zero a null tag is used which is odd but since this data is passed to flash later maybe there's a reason?  Anyway we will need to deal with this somehow to stop regular data loss.

[...]

PR #1011, opened 2018-01-18
Fix KeyError in Bolivia

systemcatch:
This should stop Bolivia losing all data when there is no wind data available. There is more work to be done in making the parser get only 1 dataframe for hourly data rather than creating up to 23 identical ones. But I'd like to make sure this fix is working ok before doing that.

ref #984

### DIFF
--- a/parsers/BO.py
+++ b/parsers/BO.py
@@ -65,7 +65,10 @@ def fetch_hourly_production(country_code, obj, hour, date):
         hour = 24
     # Fill production types
     for i_type in MAP_GENERATION.keys():
-        data['production'][i_type] = obj[MAP_GENERATION[i_type]][obj.hour == hour].iloc[0]
+        try:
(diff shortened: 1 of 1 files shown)


parsers/BO.py
```diff
@@ -65,7 +65,10 @@ def fetch_hourly_production(country_code, obj, hour, date):
         hour = 24
     # Fill production types
     for i_type in MAP_GENERATION.keys():
-        data['production'][i_type] = obj[MAP_GENERATION[i_type]][obj.hour == hour].iloc[0]
+        try:
+            data['production'][i_type] = obj[MAP_GENERATION[i_type]][obj.hour == hour].iloc[0]
+        except KeyError as e:
+            data['production'][i_type] = None
```
corradio: Ideally I think you want to drop the whole hour when wind data is not available (as this datapoint is not "consistent").
Maybe the best way is to abort the method and return `None` in that case, filtering out missing hours later on?

systemcatch: Ok I can do that instead.

PR #1013, opened 2018-01-19
add timezones to zones.json

maxbellec:


config/zones.json
```diff
@@ -1,725 +1,525 @@
 {
```
corradio: Can you sort keys alphabetically? The changes will be much smaller in that case :)

corradio:
Will merge after tests pass

Issue #1006, opened 2018-01-15
Assign dark blue to "pumped storage"


alixunderplatz:
It's hard to distinguish hydro generation from pumped storage generation, since they use the same "blue" and are placed next to each other. 
There is a thin line between them, but a slightly darker blue (like the blue of the "frontend" label) would be really nice to emphasize these peaks :)

![image](https://user-images.githubusercontent.com/25743609/34947198-b6f53798-fa09-11e7-9527-25185aeb8f43.png)

corradio:
Good idea!
Relevant code is here in case someone wants to play with this:
https://github.com/tmrowco/electricitymap/blob/master/web/app/main.js#L330

systemcatch:
@alixunderplatz this is using the colour for the frontend label, thoughts?
![hydro_storage](https://user-images.githubusercontent.com/30196510/35194935-c2af0330-feb3-11e7-81ad-1cd1720c1124.png)

alixunderplatz:
@systemcatch looks intense! I think it's good! hope it will also be okay with some biomass next to it in the diagram below.

alixunderplatz:
Looks good, so I'll close this :)

![image](https://user-images.githubusercontent.com/25743609/35282831-d42bca18-0056-11e8-8f68-beabd6e6274e.png)

![image](https://user-images.githubusercontent.com/25743609/35282903-145d4dbe-0057-11e8-85fb-43383018a2c2.png)

![image](https://user-images.githubusercontent.com/25743609/35283010-4ed7550c-0057-11e8-8e90-b96d9ac1dfa9.png)

PR #1018, opened 2018-01-21
Change hydro storage to dark blue

systemcatch:
Let me know if it looks right. /cc @alixunderplatz 

ref #1006


corradio:
Let's try it! Can you solve the conflicting file?

PR #1022, opened 2018-01-23
Split Canary Islands using new shapefiles

### LINKED ISSUE #626: Split generation by island for Canaries
Is it possible to split canary island into the islands.
they are really working for there own.
see  also:
http://ree.es/en/activities/canary-islands-electricity-system/canary-electricity-demand-in-real-time

El Hierro is very impressiv: rigth now 100% RE /  0 CO2
source:
https://demanda.ree.es/movil/canarias/el_hierro/total

systemcatch:
What I've done is open the natural earth file in QGIS, then cut out the islands into their own .shp file.  Then rename the NAME attribute to get a unique id, and finally do some hacking on the bash script to get it to work.

The bash script could definitely use some improvement, i.e. creating a country filter for countries that use other shapes. Overall this is a big improvement to the map.

I've left the Islands as one for now, once @blackleg has modified the Canary Islands parser to return data for each one it's easy to implement zones and topos.

ref #626 

Gif showing the split islands.

canary


web/app/countrytopos.js
```diff
@@ -351,6 +351,13 @@ exports.addCountryTopos = function(countries) {
   countries['SS'] = getCountry('SSD')
   countries['ES'] = getSubUnits(['ESX', 'SEC', 'SEM']); //Spain Peninsula
   countries['ES-CN'] = getSubUnits(['ESC']); //Spain Canary Islands
+  // countries['ESC-LP'] = getCountry('La Palma')
```
corradio: Can we keep the ISO names (https://en.wikipedia.org/wiki/ISO_3166-2:ES)?
If we subdivide at this level, I suggest having `ES-CN-...` in order to keep the prefix consistent.

web/topogen.sh
```diff
@@ -34,13 +36,25 @@ echo "Extract metadata"
...
 echo 'Parsing countries..'
 ("$NODE_MODULES_PATH/shp2json" -n build/${COUNTRIES_FILENAME}.shp \
   | "$NODE_MODULES_PATH/ndjson-map" '(d.id = (d.properties.ISO_A3 != "-99") ? d.properties.ISO_A3 : d.properties.ADM0_A3, d.oldProps = d.properties, d.properties = {}, d.properties.subid = d.oldProps.SU_A3, d.properties.code_hasc = d.oldProps.code_hasc, delete d.oldProps, d)' \
-  | "$NODE_MODULES_PATH/ndjson-filter" 'd.id' \
+  | "$NODE_MODULES_PATH/ndjson-filter" "d.id" \
 )> build/tmp.json
 
+# Parse 3rd party
+echo 'Parsing 3rd party..'
+("$NODE_MODULES_PATH/shp2json" -n build/Canary_Islands.shp \
```
corradio: Seems like you just need the shp, so maybe we can remove the others from git?

corradio:
Are you sure it fixes #626? I think we still need a parser.

web/app/countrytopos.js
systemcatch: Sure I will switch the names to all have a ES-CN- prefix, happy with the second parts (FV/LP etc)?

web/topogen.sh
systemcatch: I think the others are required for the shp file to work right. When we download natural earth shapes they include those extra files.

corradio: I think the shp file is a self-contained file - not sure though. Would be surprised otherwise.

web/app/countrytopos.js
corradio: yep! all good

web/topogen.sh
systemcatch: https://en.wikipedia.org/wiki/Shapefile, looks like the other parts are needed. Best to keep them for now imo.

corradio: 👍

PR #1024, opened 2018-01-25
Add US (MISO)

systemcatch:
Got this one up and running. Fixed the bug in update_state.py as well.  I've assigned their 'other' category to 'unknown' but maybe it needs special treatment.

To quote the MISO data source;
"The category listed as “Other” is the combination of Hydro, Pumped Storage Hydro, Diesel, Demand Response Resources, External Asynchronous Resources and a varied assortment of solid waste, garbage and wood pulp burners".

ref #983 

miso


sandeen:
Awesome!

FYI, the "Other" category does sometimes go negative, presumably when pumping hydro.  Not sure if you noticed that.  (It did so last night from 11pm to 4am, around -1MW.

systemcatch:
Hmmm no I hadn't noticed that, I find it weird they cram all those generation types into 'Other'.

README.md
```diff
@@ -130,6 +130,7 @@ Real-time electricity data is obtained using [parsers](https://github.com/tmrowc
 - Ukraine: [UKRENERGO](https://ua.energy/activity/dispatch-information/ues-operation/)
 - United States: [PYISO](https://github.com/WattTime/pyiso)
 - United States (California): [CAISO](http://www.caiso.com/Pages/default.aspx)
+- United States  (MISO):[MISO](https://api.misoenergy.org/MISORTWDDataBroker/DataBrokerServices.asmx?messageType=getfuelmix&returnType=json)
```
corradio: Maybe we should do one level of indent for the US (as done in capacities?

corradio:
LGTM but parser will fail if some data is negative. How do we handle it?

README.md
systemcatch: Sure will do.

systemcatch: I'll also get rid of the pyiso reference.

systemcatch:
A few options, only the 'other' key should ever go negative.

- Drop the negative key entirely.
- Set it to None.
- Move the key value to hydro storage.

corradio:
I merged. Let's see how often this happens

sandeen:
My best guess is the negative value indicates hydro pumping energy _consumption_, so I don't think that putting it in the hydro production category would be the right choice.

Issue #1023, opened 2018-01-24
Update Turkey TR parser and installed capacity


alixunderplatz:
According to the [latest news](https://www.pv-magazine.de/2018/01/22/tuerkei-baut-allein-im-dezember-1175-gigawatt-photovoltaik-zu/), Turkey's PV installations have quadrupled from 832 MW to 3420 MW in 2017.

So I checked the current data source https://ytbs.teias.gov.tr/ytbs/frm_login.jsf and it now has "Güneş" which means "solar". **In our parser, the category is still missing.**

The published numbers don't look correct at the moment (yesterday, the solar peak at noon was at ~6 MW), but I'm sure they will improve this.

So, [the parser ](https://github.com/tmrowco/electricitymap/blob/895561d6e2cb492439992fa327de3616fe6a9a74/parsers/TR.py)needs to be updated to include "Güneş" (I guess only the MAP_GENERATION part will do?).

I would also update the categories in general, since there seem to be changes.
Here is a mapping of the categories to the types. I am not sure how to change it in MAP_GENERATION, since there are no spaces and no capital and special letters. Somebody, please take a look :)

Fuel Oil - oil
Linyit - coal
Taş Kömür - coal
Asfaltit Kömür - coal
İthal Kömür - coal
Doğal Gaz - gas
LNG - gas
Kojenerasyon - unknown
Biyokütle -biomass
Jeotermal - geothermal
Barajlı - hydro
Akarsu - hydro
Rüzgar - wind
Güneş - solar

In the recent parser, there is "nafta" which is mapped to unknown (should be oil), but I'm not sure whether this category still exists. So somebody, please check the categories :)

Maybe the original author of the parser, @olivierSemet is up to do this?

There also is a good source for the latest **installed capacities** i found on https://www.teias.gov.tr/:

https://www.teias.gov.tr/sites/default/files/2018-01/2018.01.24%20TE%C4%B0A%C5%9E%20YTM%20G%C3%BCnl%C3%BCk%20Faaliyet%20Raporu-%C4%B0nternet%20Versiyonu.pdf

I can update the installed capacities if you approve these numbers. For renewables, there are numbers from IRENA, for other s it is enerji.gov.tr - we could remove these and use the numbers from the up-to-date source of the TSO "TEİAŞ".

OLD:
```
  "TR": {
    "capacity": {
      "biomass": 367,
      "coal": 17317,
      "gas": 22552,
      "geothermal": 821,
      "hydro": 27385,
[...]

PR #1026, opened 2018-01-26
Update Turkish installed capacity

alixunderplatz:
Based on latest figures from https://www.teias.gov.tr/ in daily report "Günlük İşletme Raporu"


corradio:
Ref #1023

PR #1053, opened 2018-01-31
Update en.json

strassburger:
change "Hierro" to "El Hierro"


strassburger:
May bee i am not patient enough.....
The changes still are not activated.
Did i mess  some up in not terminating the first wrong check?
found same here: -is ist normal, just have to wait or have i to clear up something somewhere???

grafik

corradio:
@strassburger once a change is done on github, I need to trigger a deployment to make update the version in production.
When that is done, you will see a "release" message on #electricitymap in Slack:
image

I've just started a new deployment so it should be up in the next minutes :)

Issue #1046, opened 2018-01-30
Dash lines to mark US states within RTOs


Hyric89:
A question for the US territories: could states within the RTOs be outlined with dash lines? This would make it easier to tell which state is located where within the RTO block. This is all aesthetics as placing the mouse of the state of Iowa versus the state of Minnesota will still display MISO as a whole regardless.

I will put forward two reasons to argue in favor of such a modification:

-When toggling the wind and/or solar potential function, it is difficult to distinguish which states are getting more wind and/or solar irradiance than others. Adding dash lines to mark state borders within RTOs would make it clear where the states' borders are within the RTO

-It could act as a (temporary) substitute for having specific, clear data of electricity generation and transport within specific state borders and at this moment, that doesn't seem to be available. The next best thing would be to simply add data for RTOs and the [unfettered boundaries](https://www.wapa.gov/newsroom/NewsFeatures/PublishingImages/ISO-RTOmap_slider.jpg) they come with. It would allow Electricitymap's representation of RTOs in the USA to be more accurate geographically while making it clear which states or which pieces of states the RTOs cover and which they do not. FOr example, it would make clear which parts of Texas are covered by SPP, MISO and ERCOT.

I would like to hear your thoughts.

systemcatch:
I like the idea of having state lines show when the RTO area is hovered over. Personally I don't think we will get generation by state anytime soon, see #899 .

systemcatch:
@Hyric89 what do you think of this?

![states](https://user-images.githubusercontent.com/30196510/35743839-a94b7dec-0836-11e8-9245-3f6cabe2c680.png)

Hyric89:
Yes, it's much more clear! Now I don't have to guess where the state borders are. It's also very pleasant on the eyes.

Issue #1058, opened 2018-02-01
Incomplete highlight on some zones


systemcatch:
Observed on Firefox 57 and Chromium 63;

When I hover the mouse on MISO (and sometimes Ontario) it doesn't always highlight the full area. It seems to occur more often when zoomed in, I haven't seen it happen on the first 3 zoom levels yet.

![miso_hover](https://user-images.githubusercontent.com/30196510/35693482-3e9f4e76-0776-11e8-8ce8-8ea7b8c92e7c.png)

alixunderplatz:
Have observed the same issue with other regions and countries, just like described. Happens for quite many regions starting from a certain zoomstage and depending on which side the cursor is coming from.

corradio:
Yeah it seems to be related to the mapbox-gl map implementation.. No clear clue on how to fix yet.

jarek:
The highlight boxes seem to be rectangles, and depend on where the hover area was entered.

It's particularly visible on Jan Mayen fairly reliably highlighting only a bit of Norway:

![screenshot - 050218 - 19 26 36](https://user-images.githubusercontent.com/47415/35821874-c5c1a73c-0aaa-11e8-933e-157fef1e713d.png)

Maybe there's just a default highlight area somewhere that's too small for our purposes?

corradio:
The way it works right now is that a new white geometry is draw with a
certain opacity. The geometry is the proper one, so it must be a rendering
issue.
Maybe we should open an issue on mapbox-gl to check if it's on their side.

PR #1061, opened 2018-02-03
Show US state lines within RTO areas

systemcatch:
As pointed out it's a little difficult to tell which states are within each RTO. Since state boundaries are very important to the way the country functions it's worth showing them in the merged areas. This will not effect the zone configurations.

ref #1046 

states


jarek:
This is cool. To quickly confirm - does the hover highlight still work as currently? I'm also curious how it'll work when we have boundaries that don't follow state borders.

systemcatch:
Hey jarek I made a gif showing the hover highlight on each area. I think the partial highlighting on MISO is related to #1058 and not to this change. For the future I don't think this method would work as there'd be overlap with some of the shape edges. Though maybe we could take a state boundary shapefile and merge it with our custom shapes so it's built in by default......

hover

corradio:
Wow I'm impressed it actually works.
How come? Is it because geometries are added in a specific order and the aggregated ones end up showing on top? Or maybe you forgot to commit some files?

systemcatch:
I think that's what's happening here, when the merged area is formed first then the underlying states won't show. Maybe this is due to how the internals of topojson.merge work (https://github.com/topojson/topojson-client/blob/master/README.md#merge). Looking at that page mergeArcs might help with performance?

The first time I saw this was when we were trying to split Chile into regions and I missed one state out when aggregating. Unintended but interesting feature.

corradio:
I'm spending some time rewriting the geoJSON generation script to improve performance and clean it up.
I'm my next PR we will be able to tell which merges we want as a union (keeping inner arcs) and which we want to prune (removing inner arcs).
I'm therefore closing this, but stay tuned!

systemcatch:
Sounds good let me know if you need a hand with anything.

Issue #1007, opened 2018-01-16
Add/Fix missing cross-border exchanges for Croatia (HR)


alixunderplatz:
I was looking at Croatia and was wondering why cross-border exchanges with Slovenia and Bosnia-Herzegovina are not displayed. Was there any reason in the past? Coudn't find anything yet in related issue #135
 
Exchange-data is completely available on entso-e and also here:

https://www.hops.hr/wps/portal/hr/web

![image](https://user-images.githubusercontent.com/25743609/34986098-970bf73e-fab6-11e7-9a43-891ef85bdadd.png)

Seems like the BA->HR arrow has no "lonlat" assigned and the  HR->SI arrow is yet missing in the code. Exchange with HR also does not appear in the bar chart for Slovenia. For the other connected countries it's there for the past 23 h.

Another issue related to this: the latest data for exchange with SI, RS and BA is only available for the previous hour. For the current hour, exchange is "0 MW". So, probably, that is the reason why the existing arrows are not displayed on the map? It's only there for the exchange with Hungary, where the latest data on entso-e is not zero. 

The entire required exchange data is probably retrieved in the entso-e parser already?

alixunderplatz:
Forgot to mention: data from https://www.hops.hr/wps/portal/hr/web seems to be "live" data. So this could replace the missing/zero values from entso-e.

alixunderplatz:
If the data from entso-e will be replacecd with first-hand live exchange-data (updating every 5-10 seconds) from HOPS (Hrvatski operator prijenosnog sustava d.o.o.), here is a small to-do list:
- write [parser](https://github.com/tmrowco/electricitymap/tree/master/parsers) for live-data from [HOPS](https://www.hops.hr/wps/portal/hr/web)
- add/fix missing arrows (HR->SI / BA->HR) in [exchanges.json](https://github.com/tmrowco/electricitymap/blob/master/config/exchanges.json)
- add / update exchange source in the graphs for RS, SI, BA and HU (where in the code is that?)
- **probably** exclude Croatia-related stuff from [ENTSO-E parser](https://github.com/tmrowco/electricitymap/blob/master/parsers/ENTSOE.py)
- add data source for Croatia (exchange only) in [readme](https://github.com/tmrowco/electricitymap/blob/master/README.md)

Please, tell me if something is missing here :)
Maybe someone in your upcoming hackathon can deal with this issue to warm-up? :)

systemcatch:
[...]

PR #1063, opened 2018-02-03
Implement missing Croatian exchanges

systemcatch:
BA->HR and HR->SI are now working. This parser can also do exchanges with Hungary and Serbia if we want to starting implementing multiple parsers per exchange.

/cc @alixunderplatz 

ref #1007 

hops


alixunderplatz:
nice, thanks @systemcatch !!! :)

PR #1064, opened 2018-02-05
Better topology serving

### LINKED ISSUE #992: Serve GeoJSON instead of TopoJSON
GeoJSON might be bigger in size, however, we might be able to save some loading time by running in-advance (precomputing) the `countries` object using `countrytopos.js` in a compile step.

### LINKED ISSUE #1046: Dash lines to mark US states within RTOs

corradio:
- ~Serves GeoJSON instead of TopoJSON~
- Precompiles the GeoJSON to avoid doing work on the client
- Supports showing state lines

Fixes #1046 
Ref #992 

The problem is that the bundle size is currently increased as world.json was ~500, and is now >1MB. This doesn't make any sense as the resulting topo should be less (we have removed unused geometries in the process?). This needs to be fixed.

@systemcatch let me know if you have any ideas.


systemcatch:
To be honest I was expecting a bigger bundle size than that.  Though download size for map might be bigger the increase in performance could make up for it.  Needs testing.

corradio:
Actually I just reverted to still serve topojson. The bundle size is still bigger than before, which shouldn't happen.

systemcatch:
Had a scan of the code but nothing leaps out. If you put both world.json files on slack we might see a difference when viewing them.

jarek:
@corradio check the JSON stringification - the new zonegeometries.json has a _lot_ of decimal digits ("33.52899174000001,35.055794169000066") which is not necessary at our scales (two should be more than enough) and could be increasing the file size.

jarek:
The world.json `arcs` list also has a lot of decimal places in its coordintes (but I don't know if the old one did as well)

maxbellec:
FYI, I only added a quantization step at the end in `generate-geometries`. I also reduced precision, from `1e5` to `1e4`, since I couldn't see the difference between both after simplification.

New size is `348KB` un-zipped, `102K` zipped

web/generate-geometries.js
```diff
@@ -0,0 +1,568 @@
...
+  }
+});
+zoneFeaturesNoStatesMerge = toListOfFeatures(zoneFeaturesNoStatesMergeInline, false);
+
+// Write unsimplified list of geojson, with and without state merges
+fs.writeFileSync('zonegeometries.json', zoneFeatures.map(JSON.stringify).join('\n'));
+fs.writeFileSync('zonegeometriesNoStateMerge.json', zoneFeaturesNoStatesMerge.map(JSON.stringify).join('\n'));
+
+// Simplify
+// fs.writeFileSync('src/zones.json', JSON.stringify(zones));
```
corradio: Remove

web/generate-geometries.js
```diff
@@ -0,0 +1,568 @@
...
+  }
+  else{
+    zoneFeaturesNoStatesMergeInline.push(data);
+  }
+});
+zoneFeaturesNoStatesMerge = toListOfFeatures(zoneFeaturesNoStatesMergeInline, false);
+
+// Write unsimplified list of geojson, with and without state merges
+fs.writeFileSync('zonegeometries.json', zoneFeatures.map(JSON.stringify).join('\n'));
+fs.writeFileSync('zonegeometriesNoStateMerge.json', zoneFeaturesNoStatesMerge.map(JSON.stringify).join('\n'));
```
corradio: Cleanup: we only need one of them

systemcatch:
Hi @maxbellec nice to have you on board. The size of world.json is looking better and documenting the process of how it's built is a good idea.

corradio:
Good on my side, let's merge once max has tested the frontend

PR #1065, opened 2018-02-06
New parser for Gujarat(India)

sahitya-pavurala:
I added a new parser for Gujarat(India), also updated zones.json


systemcatch:
Hey thanks for doing this. I will have a look through later on, one thing that needs to be done is to change the parser file name to IN_GJ.py to match the others.

sahitya-pavurala:
Changed the file name to IN_GJ.py

config/zones.json
```diff
@@ -1126,6 +1126,17 @@
...
+  "IN-GJ": {
+    "contributors": [
+      "https://github.com/sahitya-pavurala"
+    ],
+    "flag_file_name": "in.png",
+    "parsers": {
+      "consumption": "IN_GJ.fetch_consumption",
+      "production": "IN_GJ.fetch_production"
+    },
+    "timezone": null
```
corradio: Can you add the timezone?

parsers/IN_GJ.py
```diff
@@ -0,0 +1,130 @@
...
+
+    solar_html = web.get_response_soup(country_code, 'https://www.sldcguj.com/RealTimeData/GujSolar.php', session)
+    wind_html = web.get_response_soup(country_code, 'https://www.sldcguj.com/RealTimeData/wind.php', session)
+
+    india_date = get(solar_html.find_all('tr')[0].text.split('\t')[-1].strip(), 'D-MM-YYYY h:m:s')
+
+    solar_value = float(eval(solar_html.find_all('tr')[-1].find_all('td')[-1].text.strip()))
+    wind_value = float(eval(wind_html.find_all('tr')[-1].find_all('td')[-1].text.strip()))
+
+    hydro_value = thermal_value = gas_value = nuclear_value = coal_value = 0.0
```
corradio: If the values are missing, it's best to set to `None` in order to show the question mark on the UI.
Set it to 0 only if you know for sure that the value is 0.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,130 @@
...
+    :param session:
+    :return:
+    """
+    value_map = fetch_data(country_code)
+
+    data = {
+        'countryCode': country_code,
+        'datetime': value_map['date'],
+        'production': {
+            'biomass': 0.0,
```
corradio: Same here regarding the None vs 0

parsers/IN_GJ.py
```diff
@@ -0,0 +1,130 @@
...
+    Method to get production data of Gujarat
+    :param country_code:
+    :param session:
+    :return:
+    """
+    value_map = fetch_data(country_code)
+
+    data = {
+        'countryCode': country_code,
+        'datetime': value_map['date'],
```
corradio: This date needs to be a python datetime object (not an arrow object).
Don't you need to suffix `.datetime` to get it?
Would be good to double check.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,130 @@
...
+        "solar": solar_value,
+        "hydro": hydro_value,
+        "thermal": thermal_value,
+        "wind": wind_value,
+        "gas": gas_value,
+        "nuclear": nuclear_value,
+        "coal": coal_value
+    }
+
+    cookies_params = {
```
systemcatch: These are only used in the other_html variable which is commented out. Unless there is a need for them later both can be removed.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,130 @@
...
+            'hydro': value_map['hydro'],
+            'nuclear': value_map['nuclear'],
+            'oil': 0.0,
+            'solar': value_map['solar'],
+            'wind': value_map['wind'],
+            'geothermal': 0.0,
+            'unknown': 0.0
+        },
+        'storage': {
+            'hydro': 0.0
```
systemcatch: Lets set this to None rather than 0.0

sahitya-pavurala:
Addressed all the issues. We are on the same page with .datetime for dates.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,129 @@
...
+        'PHPSESSID': 'a301jk6p1p8d50dduflceeg6l1'
+    }
+
+    # other_html = requests.get('https://www.sldcguj.com/RealTimeData/RealTimeDemand.php', params=cookies_params)
+    rows = web.get_response_soup(country_code, 'https://www.sldcguj.com/RealTimeData/RealTimeDemand.php',
+                                 session).find_all('tr')
+
+    for row in rows:
+        if len(row.find_all('td')) > 3:
+            v1, v2 = (re.sub(r'[\n\t\r]', r'', x.text.strip()) for x in itemgetter(*[0, 3])(row.find_all('td')))
```
systemcatch: Can you explain this line? More interested in the why than the how, it works though so not looking for any changes.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,129 @@
...
+            energy_type = [k for k, v in station_map.items() if v1 in v]
+            if len(energy_type) > 0:
+                value_map[energy_type[0]] += float(eval(v2))
```
systemcatch: Maybe we can avoid eval() here and use https://docs.python.org/3/library/ast.html#ast.literal_eval or something else.

parsers/IN_GJ.py
```diff
@@ -0,0 +1,129 @@
...
+        },
+        'source': 'sldcguj.com',
+    }
+
+    return data
+
+
+def fetch_consumption(country_code='IN-GJ', session=None):
+    """
+    Method to get consumption data of Nepal
```
systemcatch: Gujarat instead of Nepal?

parsers/IN_GJ.py
sahitya-pavurala: Only because the html contents are formatted that way. The page is not well formatted and is not consistent for all the tags.

parsers/IN_GJ.py
sahitya-pavurala: using ast.literal_eval instead of eval everytime we need it is a good change.

parsers/IN_GJ.py
sahitya-pavurala: I prefer to keep it for now. I might add few changes later on.

corradio:
Thanks for the changes.
On my side, the only thing remaining would be to post a screenshot of the parser tested locally, using https://github.com/tmrowco/electricitymap#testing-parsers-locally (just to be sure we don't have bad surprises when pushing to prod)

parsers/IN_GJ.py
systemcatch: You weren't kidding about the html being filthy!

systemcatch:
Excellent, just needs the readme updating and we're done.

sahitya-pavurala:
screen shot 2018-02-06 at 12 38 29 pm

sahitya-pavurala:
Updated Readme.md

sahitya-pavurala:
Can we close this PR?

brunolajoie:
Awesome work @sahitya-pavurala !

Issue #757, opened 2017-09-17
Russia 🇷🇺 


denplis:
**Cumulative production per hour:**
http://so-ups.ru/index.php?id=ees_gen_consump_plan&no_cache=1

**Nuclear production, per few days update?**
http://www.rosenergoatom.ru/
url for parsing data:
http://www.rosenergoatom.ru/data.xml

**Capacities:** 
https://minenergo.gov.ru/node/532
From this [article](http://nnhpe.spbstu.ru/struktura-elektroenergetiki-v-rossii/), we 've got fossil stations structure:
gas - 71%
coal - 27.5%
oil - 1%
other - 0.5%

systemcatch:
@denplis Thanks for the info & links.  It looks like the first link has hourly production for each grid region which is useful.  I'm assuming that includes all types of generation.

The nuclear generation data is probably for the whole country using [this](https://en.wikipedia.org/wiki/Electricity_sector_in_Russia) as a capacity reference.

If we could separate out their hydroelectric generation then I think we could write a parser for the area of Russia covered by the grid.

denplis:
Production by type, updating per hour:
http://br.so-ups.ru/Public/MainPageData/BR/PowerGeneration.aspx

Export / Import data
http://br.so-ups.ru/Public/MainPage.aspx

denplis:
https://github.com/tmrowco/electricitymap/pull/766 parser draft

jarek:
Look like the data in http://br.so-ups.ru/Public/MainPageData/BR/PowerGeneration.aspx is scheduled and not necessarily real-time? It's 16:16 on 23.09 in Moscow right now and the page is returning data for up to 23.09.2017 23 hour.

denplis:
@jarek you right, that a schedule for the next day.
Real production displayed [here](http://br.so-ups.ru/Public/MainPageData/BR/GenConsum.aspx) but  
without the type of energy of production.
Anyway, difference between plan and real production isn't so big

corradio:
There's a decision to be made here. Up to know, we have refrained from using estimated numbers for everything as this is supposed to be a live view.
@denplis can you maybe try to write to them to know if they have a live breakdown?

systemcatch:
If we do use the planned generation there should probably be a note on the country breakdown saying it's estimated.  Also in the data returned by the parser we could have key like 'estimated'/'scheduled' set to True/False to allow easier filtering of those datapoints.

corradio:
[...]

PR #1066, opened 2018-02-06
Add Russia

systemcatch:
This is an updated version of #766 . I've made the requested changes and got it working with Python 3. Also rebased the branch to bring it up to date. Lets get Russia on the map! :ru: 

/cc @alixunderplatz , @denplis 

ref #757 

ru


parsers/requirements.txt
```diff
@@ -9,3 +9,4 @@ reescraper==1.6.0
 requests==2.18.4
 requests-mock==1.3.0
 xlrd==1.0.0
+tablib==0.12.1
```
corradio: Can you order alphabetically?

systemcatch:
I think it is measured once the the current hour is past (does that make sense?). @alixunderplatz you've done the work on this and can explain it better than me.

corradio:
ok waiting for confirmation from @alixunderplatz before merging!

alixunderplatz:
@corradio In #757 I assumed it was a "final schedule" for the upcoming hour.
You can track it [here](http://br.so-ups.ru/Public/MainPageData/BR/PowerGeneration.aspx) and see that the upcoming interval will change around minute :51 / :52 of each hour.

Just right now, values for "interval 15" (and following hours) were updated, while for "interval 14" they remain unchanged.

image

Assuming "interval 15" is 14-15 moscow time, it was "over" just in these minute. So either it is measured in an odd time frame (13:50 to 14:50) , or related to the previous (13-14?) hour and measured or something I don't understand. Maybe it is measured generation from 13:30 to 14:30 moscow time, which is published 20 min later. Or another timezone (Moscow time is central for rail grid, too!). or a final schedule. Assumptions.
Although, it clearly says "generation" or "Выработка".

I will look for a documentation that was somewhere on the site!

@systemcatch We could set 567 g/kWh to the "unknown" part as emission factor (my comment from Jan 5 #757 and Brunos replies) https://github.com/tmrowco/electricitymap/blob/master/config/co2eq_parameters.js


image

alixunderplatz:
@corradio @systemcatch ah it's tricky! I just saw there is an "interval 0" in the chart on http://br.so-ups.ru/Public/MainPageData/BR/PowerGeneration.aspx

So "interval 15" probably could mean "15 to 16", for which the data is updated 10 min ahead.
Will look for the documentation!

alixunderplatz:
Couldn't get anything out of the documentation of the publicly available data, because they left the third tab "БР" out ^^

On their main page they have the graph for "generation in the 15th hour" 

http://br.so-ups.ru/Public/MainPage.aspx
Unfortunately, clicking on "15 час" will just foward you to the generation table and not to the most recent hour.

In the top row it says "ч" which is the abbreviation for "час", so "hour".

image

I suggest interval "0" is 00:00 to 01:00. This is also supported by the time selector on the main page above the map in the upper right corner.

image

At 16:00 moscow time, data for the interval 14 to 15 shows actual generation on the map, too. So actual data is published with a delay of 1 hour. For interval 15 to 16 it is still empty (-), but planned generation is filled. But no seperation of types in that case.

I still think the data from the generation by type page is the "final 10-min-ahead schedule". 
Looking on the map on the front page, you can see that the differences between planned (План.) and actual (Факт.) are rather insignificant compared to the > 100 GW generation.

Relying on a final schedule for the upcoming hour seems legit.

systemcatch:
Currently the upcoming hour is dropped in the code, so at 3.59pm data for 4.00pm will still be ignored. This seems safer than including it for now. Otherwise I think we are ok with the parser as currently constructed.

alixunderplatz:
@systemcatch good!

By the way, it looks like Kaliningrad is also filled with colour. It will have no effect on exports to Lithuania, because that is defined as "from RU-KGD", right?

As a consecutive step, we should consider seperating the price zones by federal regions (called "oblast" or "krai"). Just like the states in the US. This will exclude the northern islands and thinly populated parts of Siberia. Will drop a list in a seperate issue when I have it.

corradio:
Awesome! Let's merge this and iterate.

brunolajoie:
💪  Be ready for some twitter activity!

PR #1067, opened 2018-02-06
Updating the 'updated x min ago' div every minute

### LINKED ISSUE #966: "Updated X min ago" text should be recalculated every minute

grady-lad:
Fixes #966

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -940,7 +940,6 @@ function dataLoaded(err, clientVersion, argCallerLocation, state, argSolar, argW
   d3.select('#new-version')
     .classed('active', (clientVersion !== getState().application.bundleHash && !getState().application.isLocalhost && !getState().application.isCordova));
 
-  // TODO: Code is duplicated
   dispatch({
     payload: state,
     type: 'GRID_DATA',
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -895,6 +895,18 @@ function renderMap() {
 }
 
 let countryListSelector;
+// inform the user the last time the map was updated.
+function setLastUpdated() {
+  currentMoment = (getState().application.customDate && moment(getState().application.customDate) || moment(getState().data.grid.datetime));
+  d3.selectAll('.current-datetime').text(currentMoment.format('LL LT'));
+  d3.selectAll('.current-datetime-from-now')
+    .text(currentMoment.fromNow())
+    .style('color', 'darkred')
+    .transition()
+    .duration(800)
+    .style('color', undefined);
+}
+setInterval(setLastUpdated, 60000);
 
 function dataLoaded(err, clientVersion, argCallerLocation, state, argSolar, argWind) {
   if (err) {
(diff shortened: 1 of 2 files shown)


web/src/main.js
```diff
@@ -927,16 +927,20 @@ function dataLoaded(err, clientVersion, argCallerLocation, state, argSolar, argW
...
   // TODO: Code is duplicated
-  currentMoment = (getState().application.customDate && moment(getState().application.customDate) || moment(state.datetime));
-  d3.selectAll('.current-datetime').text(currentMoment.format('LL LT'));
-  d3.selectAll('.current-datetime-from-now')
-    .text(currentMoment.fromNow())
-    .style('color', 'darkred')
-    .transition()
-    .duration(800)
-    .style('color', undefined);
+  const setLastUpdated = () => {
```
corradio: I think you're inside the method `dataLoaded`, which is called every time the frontend has new data.
This will create several `setInterval`.
You should probably move this outside of the method, and just call `setLastUpdated` from inside.

grady-lad: Ah apologies!

Ok if I move the function outside of `dataLoaded` there is a dependency against the state object due to the following line
```javascript
  currentMoment = (getState().application.customDate && moment(getState().application.customDate) || moment(state.datetime));
````
This fine when calling `setLastUpdated` from within `dataLoaded`, but if the function is called outside `dataLoaded` what can be used to pass the `state` parameter ?

Comment (corradio) [reply web/src/main.js]: I think the best is to use the new redux system we're putting into place (but didn't finish yet)

In a nutshell, you should
- add a new `case` in the reducer (in src/reducers/main.js) to handle a new `GRID_DATA` event, that updates the current application state, setting `data.grid` to the supplied state object (note: the application state and the grid state are two different objects)
- in `dataLoaded`, submit the grid state object to the application state by calling
```js
dispatch({
  payload: state,
  type: 'GRID_DATA',
});
```
- you will then be able to get the current date from `setLastUpdated` by calling `getState().grid.datetime`
- you should subscribe to datetime changes in order to call `setLastUpdated`. You can do that in the main.js using the `observe` method (look at existing implementations to see an example)
- set the setInterval method outside of the `dataLoaded`

web/src/main.js
```diff
@@ -926,17 +938,12 @@ function dataLoaded(err, clientVersion, argCallerLocation, state, argSolar, argW
 
   // Is there a new version?
   // d3.select('#new-version')
-  //   .classed('active', (clientVersion !== bundleHash && !store.getState().application.isLocalhost && !isCordova));
-
+  // .classed('active', (clientVersion !== bundleHash && !store.getState().application.isLocalhost && !isCordova));
   // TODO: Code is duplicated
```
corradio: can you remove this comment?

web/src/main.js
```diff
@@ -1280,6 +1287,13 @@ observe(state => state.application, (application) => {
   HistoryState.updateHistoryFromState(application);
 });
 
+// Observe for datetimechanes
+observe(state => state.data.grid, (grid) => {
```
corradio: Use `state => (state.data.grid || {}).datetime`, then you don't need the `if` after, and can directly put the `setLastUpdated` as handler

grady-lad: Using the if statement is necessary here as `setLastUpdated` is impure due to a direct reference to `getState().data.grid.datetime`.  The above suggestion doesn't solve accessing a property on a object that does not exist when within `setLastupdated`.

The only way to improve on this would be changing `setLastUpdated` to accept a param which would be `getState().data.grid.datetime`

So `observe` would look like
````javascript
observe(state => (state.data.grid || {}).datetime, (grid) => {
  setLastUpdated(grid.datetime);
});
````
And `setInterval` would be:
````javascript
setInterval(() => setLastUpdated(getState().data.grid.datetime), 60000);
````
IMO I would probably pass `state.application` as a parameter too, and add some default parameters to `setLastUpdated`.

The only problem is that the text that we set which is `moment(currentMoment).fromNow()`
will say _"a few seconds ago"_ if `currentmoment` is an empty string or undefined which is not true.

corradio: Indeed you are right!

corradio:
Can you solve conflicts before I merge?

grady-lad:
@corradio Thanks for the guidance and the review 😄

PR #1076, opened 2018-02-07
New parser for Sri Lanka

sahitya-pavurala:


sahitya-pavurala:
Tested the parser locally, everything seems fine. I just see that the hydro storage is not being displayed those there is a value for it.

parsers/LK.py
```diff
@@ -0,0 +1,107 @@
...
+    for row in rows:
+        values = row.text.lower().strip().split('\n')
+        if 'coal' in values[0]:
+            coal_value += float(values[1].split(' ')[0])
+        elif 'oil' in values[0]:
+            oil_value += float(values[1].split(' ')[0])
+        elif 'wind' in values[0]:
+            wind_value += float(values[1].split(' ')[0])
+        else:
+            hydro_value += float(values[1].split(' ')[0])
```
corradio: Can you add an else if hydro and an else that throws an exception? This will be more robust

parsers/LK.py
```diff
@@ -0,0 +1,107 @@
...
+            'hydro': value_map['hydro'],
+            'nuclear': value_map['nuclear'],
+            'oil': value_map['oil'],
+            'solar': value_map['solar'],
+            'wind': value_map['wind'],
+            'geothermal': None,
+            'unknown': None
+        },
+        'storage': {
+            'hydro': value_map['storage']
```
corradio: Note here that storage should be positive and discharge should be negative

parsers/LK.py
```diff
@@ -0,0 +1,107 @@
...
+
+from parsers import countrycode
+from parsers.lib import web
+
+
+def fetch_data(country_code, session=None):
+    countrycode.assert_country_code(country_code, 'LK')
+
+    html = web.get_response_soup(country_code, 'http://www.ceb.lk/yesterday-electricity/', session)
+    date = get(html.find_all('h4')[1].text.strip().split(':')[-1], 'MMMM D, YYYY')
```
corradio: Requires time zone

sahitya-pavurala:
screen shot 2018-02-07 at 12 21 03 pm
screen shot 2018-02-07 at 12 19 53 pm

parsers/LK.py
sahitya-pavurala: I did that because every other power plant is hydroelectric power plant.

corradio: Mmh but what would happen if they change that? You might not be around to maintain it so I prefer the parser being consistent as to what it expects. It will prevent future errors from being unnoticed.

sahitya-pavurala: Sure, Ill make that change

corradio:
Screenshots looking good. I see a bug on our local test server where the hydro storage scale is not correctly calculated. We'll fix that

parsers/LK.py
corradio: Do you confirm that is correct?

jarek:
As far as I can tell, this is total per day, not instantaneous. The units are watt-hours and presumably added up over the entire day. The parser returns it as a midnight data point, and the back-end will interpret it as watts as of midnight.

@sahitya-pavurala do you know if CEB publishes _current_ generation? Updated hourly or better is best for us.

sahitya-pavurala:
I see, so its the total for the day, even I felt the value is very high. I couldn't find a real time generation data in the website.

sahitya-pavurala:
Should we populate the storage as None for now?

parsers/LK.py
sahitya-pavurala: from jareks comment, it looks like its a total for the day,  i couldnt get any other info from their website.

corradio:
Hello @sahitya-pavurala ,
Unfortunately I didn't see this before, but as @jarek points out, this parser only gets data for a full day. We need data hour by hour in order to display it on the map :/
I will therefore close this pull request. If you come across hourly data, feel free to re-submit!
Sorry about not catching this before, and thanks for helping out.

Oli

sahitya-pavurala:
Right. That what i thought.

Thanks.

PR #1082, opened 2018-02-10
Map and translations update

Raffox97:
**_Updates for the translations:_**
* Fixed some indenting (Italian)
* Removed unused country strings (Italian)
* Added newly used country strings (Italian)
* Edited a string that was mismatching the country code (Italian)


web/locales/it.json
```diff
@@ -51,7 +55,7 @@
     },
     "tooltips": {
         "carbonintensity": "Intensità di carbonio",
-		"lowcarbon": "Basse emissioni",
+        "lowcarbon": "Basse emissioni",
```
corradio: Is there a better translation of low carbon? Right now it seems to say low émissions

web/locales/en.json
```diff
@@ -13,7 +13,10 @@
         "see": "see",
         "datasources": "data sources",
         "contribute": "Contribute by <a href=\"%s\" target=\"_blank\">adding your territory</a>",
-        "tip": "Tip: Click on a territory to start exploring"
+        "tip": "Tip: Click on a territory to start exploring",
+        "live": "Live",
+        "database": "Database",
```
corradio: Actually the toolbar items shouldn't be translated as they are the name of our products.

Raffox97: Ok, I'll rollback that edit

web/locales/it.json
Raffox97: Well, it's not the full phrase, because the full one is too long

Raffox97:
Rollbacked all the changes regarding the toolbar

PR #1093, opened 2018-02-13
trying to fix #893

### LINKED ISSUE #893: "Origin of carbon emissions last 24h" has incorrect title
When clickig on show "carbon emission", the graph title should change from "Origin of electricity in the last 24h" to "Origin of carbon emissions in the last 24h"

image

maxbellec:
this code **might** fix #893. Problem is when testing locally, there is no emission data (getting an error `dischargeCo2Intensities is undefined` when clicking on carbon emissions), so I'm not sure how to fix.

Not sure this fix is the right way to go, really experimenting here.

### DIFF
--- a/web/locales/de.json
+++ b/web/locales/de.json
@@ -34,7 +34,7 @@
     },
     "country-history": {
         "carbonintensity24h": "spezifische CO2-Emissionen in den letzten 24 Stunden",
-        "electricityorigin24h": "Ursprung der Elektrizität in den letzten 24 Stunden",
+        "electricityorgin24h": "Ursprung der Elektrizität in den letzten 24 Stunden",
         "electricityprices24h": "Strompreis in den letzten 24 Stunden",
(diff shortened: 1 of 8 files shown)


alixunderplatz:
@maxbellec Regarding the typo in de.json: I always wondered what was causing that particular text in the German version of the site to be in English. Thank you for finding it!!! ;) I had overseen this "i" is missing in "orgin" in the en.json while updating the translation^^

image

web/locales/de.json
```diff
@@ -34,7 +34,7 @@
     },
     "country-history": {
         "carbonintensity24h": "spezifische CO2-Emissionen in den letzten 24 Stunden",
-        "electricityorigin24h": "Ursprung der Elektrizität in den letzten 24 Stunden",
+        "electricityorgin24h": "Ursprung der Elektrizität in den letzten 24 Stunden",
```
corradio: Maybe we should update all translations to follow "origin" instead of "orgin"?

PR #1094, opened 2018-02-13
update PL translation

jarek:
Notes:
* "low-carbon" is translated as "low emission" following http://context.reverso.net/t%C5%82umaczenie/angielski-polski/low-carbon+energy?d=0 which actually seems like it uses some EU document :)
* country/region names are following [Polish Wikipedia](https://pl.wikipedia.org/wiki/Lista_pa%C5%84stw_%C5%9Bwiata_wed%C5%82ug_liczby_ludno%C5%9Bci), I've translated all the regions we currently have online, plus all countries with population over a million, plus a few smaller ones to make map hopefully complete in Europe and Africa
* country names that are the same as in English might not have been translated, since it seems the website falls back to English now
* For "United States of America" I used "USA" which is universally understood in Poland and much shorter than "Stany Zjednoczone Ameryki"
* US-MISO and US-PJM are translated as "MISO network" and "PJM network" since I expect these acronyms aren't well-known in Polish. "Network" is used for electrical grids in Polish
* Indian states follow [Polish Wikipedia](https://pl.wikipedia.org/wiki/Indie#Podzia%C5%82_administracyjny), some of them have Polish transliterations, others don't
* Brazil regions are named descriptively, with adjectives


systemcatch:
Nice! I'm thinking it would be a good idea to switch the English translation to USA (state) as well rather that United States of America (state) which is quite long winded and makes the hover boxes giant.

PR #1105, opened 2018-02-15
Fix bug 1081 solar and wind potential while hovering over water areas

jpierz:
Hi @corradio. Here's the PR about the ocean layer that we talked about. The other option is just an additional mousemove handler that does not target a specific layer.

I think if we go this route we have a couple nice benefits:
- Color for the ocean 🌊 
- Specific events targeting the ocean layer (if ever needed)
- Clean event handlers
- And of course hopefully fixing this bug

I think that if you agree with this approach, then there are some names in the files changed that should also be updated, like `countryMap` should probably be `worldMap` and where the word `world` describes countries, it should be `country/countries`. Just food for thought though, may not be 100% necessary.


web/src/components/map.js
```diff
@@ -27,9 +28,11 @@ export default class Map {
       type: 'FeatureCollection',
       features: this.nonClickableZoneGeometries,
     };
-    if (clickableSource && nonClickableSource) {
+    const oceanData = this.oceanData;
+    if (clickableSource && nonClickableSource && oceanSource) {
```
jpierz: Question for clarity:

When is this `if` check true?

corradio:
@jpierz did you have any chance to see my comments?

jpierz:
@corradio Yes, sorry I've been unable to update it this past week.

I agree with your comments and have made the changes as I understand them.

I tried first to have two separate `mousemove` events: one targeting the country layer and one global. Everything worked fine aside from moving the mouse over countries causing the `mapMouseOver()` function to be called twice. So I tried to condense them and this is what I landed on. You mentioned `this.map.queryRenderedFeatures(e.point);` potentially being expensive, but this way we can avoid double events firing off.

Let me know what you think :)

corradio:
Hi @jpierz,
Firing double events is not a problem. Here's my thinking:
1. Fire onMouseMove on all elements, without any filter (no need to call `queryRenderedFeatures`).
The handler should only call `mapMouseOver`.
2. Fire onZoneMouseMove (i.e. the old onCountryMouseMove) on zone elements (like previously).
This handler should only handle the tooltips.

That way we have one handler for each case. In some cases both will be fired at the same time (which is ok).

jpierz:
Hi @corradio,

Sounds good. I made the changes you requested.

web/src/components/map.js
```diff
@@ -184,28 +184,36 @@ export default class Map {
...
+          zoneId,
           rect.left + e.point.x,
           rect.top + e.point.y,
+        );
+      }
+    });
+    this.map.on('mousemove', (e) => {
+      // Disable for touch devices
+      if (this.userIsUsingTouch) { return; }
+      if (this.onMouseMove()) {
```
corradio: If this.onMouseMoveHandler

web/src/components/map.js
```diff
@@ -337,18 +345,24 @@ export default class Map {
     return this;
   }
 
-  onCountryMouseMove(arg) {
-    if (!arg) return this.countryMouseMoveHandler;
-    else this.countryMouseMoveHandler = arg;
-    return this;
-  }
-
   onCountryMouseOut(arg) {
```
corradio: Cam you also rename this one for consistency?

web/src/components/map.js
jpierz: Oops. This has been fixed.

web/src/components/map.js
jpierz: Renamed `onCountryMouseOut` to `onZoneMouseOut`. Should `onCountryClick` and `onCountryMouseOver` be renamed to `onZone...`?

maxbellec:
I have not had a look at the code, but it works great!

PR #1106, opened 2018-02-15
Swedish: Add missing regions

magol:


.editorconfig
```diff
@@ -8,5 +8,8 @@ charset = utf-8
 [*.{js,json}]
 indent_size = 2
 
+[web/locales/**.json]
```
corradio: Good call. At some point we should change that though (in one single PR)

magol: I agree, it is strange to have different coding styles for different files.

PR #1108, opened 2018-02-15
Add Chile (SIC)

systemcatch:
Sistema Interconectado Central de Chile (SIC) covers most of the generation in Chile.

Parser only gets retrospective data (1-2 days lag) but it's a full breakdown by type. I've not found a good source for capacities yet but there should be one as I found capacities for the SING region easily enough, will keep looking.

/cc @alixunderplatz 

ref #655 

cl-sic


alixunderplatz:
capacity from their app

screenshot_2018-02-15-18-55-56

alixunderplatz:
@systemcatch see my latest post in #655 for capacity 😉 The app seems to be for both SIC and SING (also because it is including geothermal which is only in SING as far as I know).

corradio:
Did you try to contact them for real-time data?

PR #1109, opened 2018-02-15
Adding parser for Uttar Pradesh, India

sahitya-pavurala:
Adding parser for UP, India. Not sure why the files .editorconfig and web/locales/sv.json changed.


systemcatch:
It shows changes in those 2 files because your merge of 'master' merged in #1106 which altered both of those files. I think it views that as an extra commit, shouldn't be an issue though. Don't know your setup with git/coding but using 'git rebase' can avoid many merge commits.

config/zones.json
```diff
@@ -1151,6 +1151,17 @@
     },
     "timezone": "Asia/Kolkata"
   },
+  "IN-UP": {
```
systemcatch: Is any information on capacities available?

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
+import re
```
systemcatch: Not used so the import can be removed.

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
...
+from ast import literal_eval
+from operator import itemgetter
```
systemcatch: Same here.

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
...
+import json
+from arrow import get
```
systemcatch: We can either do from arrow import get, now or just import arrow. No need for both.

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
...
+from requests import Session
+import arrow
+from parsers import countrycode
+from parsers.lib import web
+
+
+def fetch_data(country_code, session=None):
+    countrycode.assert_country_code(country_code, 'IN-UP')
```
systemcatch: Since this parser will only get data for Uttar Pradesh you can set countrycode='IN-UP' in the function and get rid of this line.

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
...
+    key_map = {
+        'total hydro generation': 'hydro',
+        'total thermal up generation': 'thermal',
+        'cogen-sent out': 'unknown',
+        'solar generation': 'solar',
+        'total up load/demand': 'demand'
+    }
+
+    value_map = {
+        "date": india_date,
```
systemcatch: response objects contains several "time_vals" keys, lets use them for datetime rather than arrow.now().

parsers/IN_UP.py
```diff
@@ -0,0 +1,117 @@
...
+    :return:
+    """
+
+    value_map = fetch_data(country_code,session)
+
+    data = {
+        'countryCode': country_code,
+        'datetime': value_map.get('date'),
+        'production': {
+            'biomass': value_map.get('biomass'),
```
systemcatch: Just a thought you could return value_map and date separately in fetch_data, so you could have 'production': value_map instead of getting every key.

config/zones.json
sahitya-pavurala: I couldn't find any capacity info on the page.

sahitya-pavurala:
Added all the suggested changes. Please check. Also for some reason when I run the build locally to test, I couldn't find any data for UP (also Gujarat ?), am I missing something.

sahitya-pavurala:
screen shot 2018-02-16 at 12 01 26 am

corradio:
@sahitya-pavurala did you run the `update_state` script to update with your new parser?
See https://github.com/tmrowco/electricitymap#testing-parsers-locally

config/zones.json
systemcatch: We'll leave it as an outstanding item in the issue.

sahitya-pavurala:
This file get modified everytime, i add a change. Do i need to checkin this file as well?
mockserver/public/v3/state

systemcatch:
Yes that can be committed as well.

sahitya-pavurala:
screen shot 2018-02-16 at 10 18 10 am

parsers/lib/web.py
```diff
@@ -13,6 +13,15 @@ def get_response(country_code, url, session=None):
...
 
 
+def get_response_with_params(country_code, url, session=None, params=None):
+    """Get response"""
+    ses = session or Session()
+    response = ses.get(url, params= params)
+    if response.status_code != 200:
+        raise ParserException(country_code, 'Response code: {0}'.format(response.status_code))
+    return response
+
```
sahitya-pavurala: I can as well add an extra parameter in get_response method. Just did this so that it doesn't break anything which is already working.

parsers/IN_UP.py
```diff
@@ -24,21 +26,21 @@ def fetch_data(country_code='IN-UP', session=None):
 
     key_map = {
         'total hydro generation': 'hydro',
-        'total thermal up generation': 'thermal',
+        'total thermal up generation': 'geothermal',
```
systemcatch: These are actual geothermal plants? Thermal normally refers to conventional fossil fuel generation, if we don't know the type breakdown (coal, gas etc) lets just set it to unknown.

systemcatch: They can't be geothermal plants. If there are 2 unknown keys then they will overwrite each in the dictionary so that needs to be avoided.

sahitya-pavurala: okay got it. setting it to unknown.

systemcatch:
Nice we should be all good to go now.

Issue #883, opened 2017-11-30
Åland / carbon intensity of zones fully powered by imports


Bjarkigd:
The Åland utility http://www.kraftnat.aland.fi/ However it says "Förändringar i bakomliggande system gör att statistiken inte visas rätt för tillfället." Which means that due to system work in the background the current stats aren't shown atm. 

Yesterday's (and further back) data http://www.kraftnat.ax/text2.con?iPage=24&iLan=1

However http://www.fingrid.fi/en/Pages/default.aspx has the live import data from Sweden to Åland, and the rest which is local is wind production http://www.allwinds.ax/vindkraftverken/?lang=en (no live data what I can find) 

Not much to go on, but it's a start.

corradio:
This looks terrific. Don't we all all we need in order to write a parser?

systemcatch:
Once kraftnat's stats start working again we should be good to go.  @Bjarkigd do you want to write this parser?

Bjarkigd:
I have to be honest and say that I have no idea at all about how to do that, I've no coding skills or experience at all, zilch, zero nada.

systemcatch:
That's fair enough, would you be prepared to contact them about live wind generation data?  They have it on http://www.kraftnat.ax/text2.con?iPage=28&iLan=1 but it's in an image format which is hard to scrape.  Contact page is [here](http://www.kraftnat.ax/text2.con?iPage=15&m=17).

Bjarkigd:
I could do that

7hibault:
@Bjarkigd If you get an answer from them, I could help with the coding part

tmslaine:
I think it is possible to parse the Åland power generation figures from that image,
which is updated every 30 seconds.
![stamnat](https://user-images.githubusercontent.com/25774514/36322324-8bab77ea-1355-11e8-9a7f-711cddfe3caf.jpg)

My idea was to use numpy and Image from PIL, recolor everything blue([0,0,255] in RGB) black and everything else white in the image, and scrape pixel masks for each digit and the decimal dot. Then one tries to fit these masks where the numbers appear on the image. I got it working but not very robust with respect to flow directions. My guess is that the arrow directions change if the flows reverse in which case the flow directions would also be machine readable using similar ideas.

http://194.110.178.135/grafik/stamnat.php is a source for the raw image
On deciphering the image:

- Senneby-Tellholm is a cable from Sweden-SE3 bidding zone to Åland
[...]

PR #1112, opened 2018-02-16
Åland graphical parser

tmslaine:
A parser creating Åland #883 generation and consumption stats from the image stamnat.php on kraftnat.aland.fi
Requires Python Imaging Library (PIL) to be added to extensions file


systemcatch:
I'm assuming you want #1113 to be the main PR and not this one?

tmslaine:
Yes, that is correct.

PR #1113, opened 2018-02-16
Åland parser

tmslaine:
Added PIL 5.0.0 to the requirements

ref #883


parsers/AX.py
```diff
@@ -0,0 +1,388 @@
...
+    obj = _fetch_data(session)
+
+    data = {
+        'countryCode': country_code,
+        'production': {},
+        'storage': {},
+        'source': 'kraftnat.aland.fi',
+        'datetime': arrow.get(obj['fetchtime']).datetime
+    }
+    data['production']['biomass'] = round(obj['biomass'],1)
```
systemcatch: We try to avoid assuming fuel types in general. Better to set something to 'unknown' if we can't be sure.

systemcatch:
The mask approach is nice indeed. While you're at it you can add Åland to the https://github.com/tmrowco/electricitymap/blob/master/config/zones.json and https://github.com/tmrowco/electricitymap/tree/master/web/locales

tmslaine:
Locales done, added AX to zones.json and the exchanges to Finland and Sweden to exchanges.json
Switched fetch_price function to be called from the ENTSOE parser (AX price = SE-SE3 price).
Added fetch_consumption function to the AX parser.
Managed to get the mask for minus sign from one of the wind parks last night, producing -0 MW, added this to the AX parser. 

Production not fossil-fuel or wind is now labeled 'unknown', assumes 100% biomass in the co2eq. Fossil-fuel production is assumed to be oil. Added capacities for oil, gas and wind. Not sure if gas production should be assumed if fossil generation exceeds installed oil capacity.

I got it running locally, the exchanges and production figures seem to work:
axtest

corradio:
If I understand correctly, "fossil" on the website is oil, and the mismatch between wind + fossil and consumption is "biomass"?
Why not put in biomass instead of unknown if you're assigning it that CO2 value?
Also, do you have any sources to support that fossil is oil, and the remaining is biomass?

tmslaine:
For assuming 'unknown' production is 'biomass', Kraftnät Åland statistics suggest that all non-fossil own production is renewable, for example: http://194.110.185.12/graphmonth5.php

I first did list unknown production as 'biomass'. But I changed that to unknown after a review.

There is no coal capacity in Åland. All gas plants were gas turbines 10 years ago http://www.kraftnat.ax/files/rapportdel_2.pdf, and I highly doubt they have built new ones after that. The data source does not really differentiate between production from oil generators and gas turbines. Since gas turbines typically have much lower efficiency than CCGT gas plants, I assume listing possible gas turbine use as oil is an OK estimate for actual CO2 emissions. 

An alternative source for the assumptions is given by statistics from the Finnish energy producers association Energiateollisuus, their statistics do not have small-scale production and solar:
They give 1.5-4 GWh of non-CHP thermal electricity production, and 12-16 GWh of CHP electricity production for year 2010, 2014, 2015 and 2016.
Practically 100% of total fuel use for district heating(CHP or not) for years 2010, 2014, 2015, 2016 has been either oil or biomass.

https://energia.fi/files/424/Sahkontuotanto_maakunnittain_2007-2016.xls has the electricity production per region and production type, Åland is referenced with its Finnish name Ahvenanmaa.
Hydro production in 2015 and 2016 might just be some of the local companies owning shares of hydropower plants elsewhere in Finland. Or there might be some small, possibly tidal hydropower plant in the islands.

For CHP fuel distribution one can look at the Finnish district heating fuel use statistics for 2016,2015,2014 and 2010:
https://energia.fi/files/2086/Vuositaulukot16_netti.xls
https://energia.fi/files/1185/Vuositaulukot15_netti.xls
https://energia.fi/files/777/vuositaulukot14_netti.xls
https://energia.fi/files/1189/Vuositaulukot10.xls
Sheet 3 has a list of district heating providers with their respective fuel uses
The companies in Åland are Mariehamns Energi, Mariehamns Bioenergi and Ålands Energi
Columns are split into fuel groups:
Coal | Oil | Peat | Natural gas | Biofuels | Waste | Process heat | Other sources |
In the far right there are columns:
"Kaukolämmön ja yhteistuotantosähkön tuotantoon käytetyt polttoaineet yhteensä"
- Total fuel use in district heating and CHP electricity production
"Edellisestä kaukolämmön erillistuotantoon käytetyt polttoaineet yhteensä"
- of which for heat-only production

Skimming through the statistics, the only fuel use from these three companies not under "Oil" or
"Biofuels" was Mariehamns Bioenergi using 0,1 GWh of peat out of a total of more than 100 GWh fuel use on two of the years.

brunolajoie:
Hello ! Could you indeed change-back unknown to biomass?
If we assume unknown is 100% biomass, let's put in under the biomass category.
Sorry for the late remark!

tmslaine:
All done!

corradio:
Seems like there still a conflict in exchanges.
Also, is it true that biomass capacity is so small? Just double checking

PR #1116, opened 2018-02-18
Add Russian exchanges with Eastern Europe & Asia

systemcatch:
This creates 7 new exchanges! I've designed the fetch_exchange function so that it will fail rather than provide bad data on flow direction. All of them look correct but please double check as there's lots of new parts being added.

/cc @alixunderplatz thanks for the help with translating everything.

ref #757 

Azerbaijan, Belarus, Georgia, Kazakhstan and Ukraine.

ru-w

China and Mongolia.

ru-asia


alixunderplatz:
Nice! 👍 
Some remarks: 
- the zone code for Kazakhstan is "KZ". In RU.py it says "KG" which is for Kirgistan
- the exchange arrow with China is located in the uncoloured/uncovered and electrically independent part of Russia (I think it's named RU-E now), which is the correct location for the arrow. 
The interconnector between RU's "far east" and China should therefore remain grey since it connects "CN->RU-E" and not ->RU.
So, if we seperated the uncovered Russian area in RU-E (for **E**urope) and RU-A (for **A**sia), this could point at "RU-A" and won't cause confusion.
- just noticed that when I hover over RU-E, RU-KGD will also be highlighted at small zoom stages

systemcatch:
Fixed the KZ zone name, also spotted where the problem is with RU-KGD so that should no longer highlight.

I originally used RU-E meaning the Eastern part of Russia but I like your idea of having separate unmapped zones. The only difference is I would call them RU-EU and RU-AS for extra clarity.

corradio:
This looks cool! I won't really have time to dig into the details until Wednesday.
@alixunderplatz and @systemcatch could you take a double look at the details to make sure everything works out? I can then merge with a very light review - but won't be able to double check all the details unfortunately.
@maxbellec feel free to take a peek here!

systemcatch:
Sure I will double check this again tomorrow, if you're busy then either @maxbellec or me can merge this to save time.

maxbellec:
I'll have a look this afternoon

corradio:
Greatness! @maxbellec feel free to merge when tests pass :)

PR #1120, opened 2018-02-19
Fixed width overflow issue in left panel

jbuverud:
Bug seen in firefox.


corradio:
Feel free to merge when tests pass!

Issue #715, opened 2017-08-25
Improve validation of NI


corradio:
![image](https://user-images.githubusercontent.com/1655848/29720186-066995ea-89b1-11e7-8391-f5019e12b4f7.png)

jarek:
Total production sent from ~500 MW to around 50 MW. Can probably reject any results that have less than 100 MW total production

jarek:
In the parser, I mean.

systemcatch:
Maybe invalidate the datapoint if too many generation types are missing and the production sum is below some floor value (higher than the current one).

corradio:
I think a very simple way is just to verify that hydro is not None.

jarek:
This could have been caught by https://github.com/tmrowco/electricitymap/commit/ef4c909d1c9fd2e470e6c787bd9c2d212a8d3bea if the error criterion was 8x or 5x rather than 10x. 

I am not a fan of looking for specific keys being None. If next time the upstream source reports 1 MW for hydro are we going to make another hardcoded fix?: :/

I still think this is optimally handled by the backend... otherwise we'll have each individual parser implementing their own validator, slightly differently. That won't make for fun maintenance. If you can't compare with past production in the backend, maybe have parsers define an "expected production" value and reference that in zones.json, and have backend discard data that is too different? Or at the very least, have one common validator function in `parsers/lib/`

corradio:
Hey @jarek you are totally right.
I'll work on that during the week. I'll change the error criterion. I however also think that it makes sense to add detailed validation inside each parser, because that's where we have an expectation of what the parser should output.

corradio:
Hi @jarek - just to say this hasn't been forgotten. I'm currently coding a major revamp of the backend to be able to do some big data analytics (there's so much data now!).
As soon as it's done I'll work on this.

systemcatch:
Hi @corradio, I quickly put together a data validator function yesterday.  It could be stored in parsers.lib and called for each individual parser with specific parameters.  I'm sure it could be extended further.

Let me know if it's something you're interested in using.

https://gist.github.com/systemcatch/256b06fca69921c605f0c04a16435255

corradio:
Awesome will take a look and incorporate it

systemcatch:
@corradio 
Cool take a look and let me know.  I'll create a proper branch & PR if it will work.

corradio:
Hi @systemcatch your code looks nice.
[...]

PR #1127, opened 2018-02-20
Create validation function for production datapoints

systemcatch:
This is a modified form of the gist I created in #715. It replaces the validation checks in FO/NI/PE, I've left ENTSOE alone for now. The idea behind it is to avoid creating the same function for each parser over and over again and to allow extensions as new situations come up. It will print failures so they are captured by the EM logger.


systemcatch:
Fixed the tests (relative import error), let me know if anything needs changing.

maxbellec:
just did a few changes so that code follows PEP 8 standards, and added examples in the docstring. @systemcatch, can you copy-paste this code in your `validation.py`?

```python
#!/usr/bin/env python3

"""Centralised validation function for all parsers."""


def validate(datapoint, **kwargs):
    """
    Validates a production datapoint based on given constraints.
    If the datapoint is found to be invalid then None is returned.

    Arguments
    ---------
    datapoint: a production datapoint. See examples
    optional keyword arguments
      required: list
        Generation types that must be present.
        For example ['gas', 'hydro']
        If any of these types are None the datapoint will be invalidated.
        Defaults to an empty list.
      floor: float
        Checks production sum is above floor value.
        If this is not the case the datapoint is invalidated.
        Defaults to None
      expected_range: tuple
        Checks production total against expected range.
        Tuple is in form (low threshold, high threshold).
        For example (1800, 12000)
        If the total is outside this range the datapoint will be invalidated.
        Defaults to None.

    Examples
    --------
    >>> test_datapoint = {
    >>>   'countryCode': 'FR',
    >>>   'datetime': '2017-01-01T00:00:00Z',
    >>>       'production': {
    >>>           'biomass': 50.0,
    >>>           'coal': 478.0,
    >>>           'gas': 902.7,
    >>>           'hydro': 190.1,
    >>>           'nuclear': None,
    >>>           'oil': 0.0,
    >>>           'solar': 20.0,
    >>>           'wind': 40.0,
    >>>           'geothermal': 0.0,
    >>>           'unknown': 6.0
    >>>       },
    >>>       'storage': {
    >>>           'hydro': -10.0,
    >>>       },
    >>>       'source': 'mysource.com'
    >>> }

    >>> validate(datapoint, required=['gas'], expected_range=(100, 2000))
    datapoint
    >>> validate(datapoint, required=['not_a_production_type'])
    None
    """

    required = kwargs.pop('required', [])
    floor = kwargs.pop('floor', False)
    expected_range = kwargs.pop('expected_range', None)
    if kwargs:
        raise TypeError('Unexpected **kwargs: %r' % kwargs)

    generation = datapoint['production']

    if required:
        for item in required:
            if generation.get(item, None) is None:
                print("Required generation type {} is missing from {}".format(
                    item, datapoint['countryCode']))
                return None

    if floor:
        total = sum(v for k, v in generation.items() if v is not None)
        if total < floor:
            print("{} reported total of {}MW does not meet {}MW floor value".format(
                datapoint['countryCode'], total, floor))
            return None

    if expected_range:
        low = min(expected_range)
        high = max(expected_range)
        total = sum(v for k, v in generation.items() if v is not None)
        if low <= total <= high:
            pass
        else:
            print("{} reported total of {}MW falls outside range of {}".format(
                datapoint['countryCode'], total, expected_range))
            return None

    return datapoint


test_datapoint = {
    'countryCode': 'FR',
    'datetime': '2017-01-01T00:00:00Z',
    'production': {
        'biomass': 50.0,
        'coal': 478.0,
        'gas': 902.7,
        'hydro': 190.1,
        'nuclear': None,
        'oil': 0.0,
        'solar': 20.0,
        'wind': 40.0,
        'geothermal': 0.0,
        'unknown': 6.0
    },
    'storage': {
        'hydro': -10.0,
    },
    'source': 'mysource.com'
}

if __name__ == '__main__':
    import doctest

    doctest.testmod()

    print(validate(test_datapoint, required=['gas'], expected_range=(100, 2000)))
```

systemcatch:
Good idea with the style changes, I've given you write access to my fork if you want to push changes there. Otherwise if you're busy I'll do it myself.

maxbellec:
Other than the minor style changes, it indeed looks much better!

systemcatch:
Yes I think trying to manage different functions everywhere will not scale well.

maxbellec:
@corradio I did some refactoring (well, my IDE did) on many parser files here (see last commit), but it's only PEP 8 cosmetics (spaces, newlines, encoding declaration). I checked the diffs visually, I'm very confident the code is functionnaly exactly the same. Are you OK with it?

systemcatch:
Not a problem PEP8 is a good guideline to aim for. 6/7 parsers have tests written so should be safe to refactor. As to the other ones we moved to python 3 wholesale in #949 with little test coverage, so living dangerously isn't a problem here! :laughing:

systemcatch:
I've had a look over all of the files altered and feel confident those PEP8 formatting changes will not break anything. Lets merge this in.

Issue #653, opened 2017-07-26
Singapore has solar data


jarek:
https://www.emcsg.com/ has generation mix for Singapore. It's overwhelmingly gas, but it is real time... Raw data comes from https://www.emcsg.com/ChartServer/blue/ticker (updated half-hourly).

The data has demand and "system loss" in MW, and mix in percent. Should generation be calculated based on demand plus system loss, then?

The data also contains prices for the following categories: USEP, LCP, Primary Reserve,
Secondary Reserve, Contingency Reserve, and Regulation. Is one of those of interest for the price parser?

jarek:
Found this for the explanation of the prices https://www.emcsg.com/marketdata/guidetoprices

systemcatch:
@jarek  nice find on the data.  I searched around and found this [report](https://www.ema.gov.sg/cmsmedia/Publications_and_Statistics/Publications/SES/2016/Singapore%20Energy%20Statistics%202016.pdf) from the government which may be useful in cross checking some figures.  System loss looks quite low but Singapore is very compact so it makes sense I guess.

corradio:
Indeed generation should be based on demand + loss.
For the prices I think that's a tricky issue. In general we have to rethink how we want to show prices, because currently they are very hard to compare.
Let's leave prices out for now, or simply take the spot price.

jarek:
OK, I'll write a parser for the generation and the USEP now then.

corradio:
Awesome!

jarek:
Now done :tada:

alixunderplatz:
@corradio @jarek 
I found a link to Singapore's PV production. 
https://www.ema.gov.sg/solarmap.aspx

Unfortunately, it's a png-image only.
https://www.ema.gov.sg/cmsmedia/irradiance/plot.png?ref=636512085214679676

Installed inverter capacity is ~109 MW, so only 1% of natural gas capacity / <2% of current daily peak demand.
Installed PV module capacity is ~141 MW according to:
https://www.ema.gov.sg/cmsmedia/Publications_and_Statistics/Statistics/31RSU.pdf

It's a quite insignificant share at the moment, but is there any way to use OCR or something similar for reading the data?

![image](https://user-images.githubusercontent.com/25743609/34770246-b7094592-f600-11e7-8e10-9a30cdb340d1.png)

systemcatch:
@alixunderplatz OCR would be fun to try.

Running the following commands on your image.
```
[...]

PR #1134, opened 2018-02-22
Use OCR to get Singapore solar data

systemcatch:
Implementation of OCR to get data from the solar map. For this to work the tesseract engine must be able to be invoked by the command 'tesseract' on the shell. I don't know the EM server setup or whether this can be handled by docker but will modify anyway needed. 

ref #653


parsers/SG.py
```diff
@@ -58,6 +61,41 @@
...
+    # At night format changes from 0.00 to 0
+    # tesseract cannot distinguish singular 0 and O in font provided by image.
+    # This try/except will make sure no invalid data is returned.
+    try:
+        solar = float(val)
+    except ValueError as err:
+        if len(val) == 1 and 'O' in val:
+            solar = 0.0
+        else:
+            print("Singapore solar data is unreadable - got {}.".format(val))
```
corradio: maybe raise an exception is in case?

jarek: If we raise exception, we lose data from main parser - there is no history backfill in SG

corradio: got it

systemcatch:
Right the image is still stuck on the same time. https://github.com/tmrowco/electricitymap/pull/1134/commits/0296bf172b6fa8b4b6423299031369d0f4de404c will make sure that any data older than 1 hour won't be used.

jarek:
We've jinxed it :(

Let's see if it gets fixed when it's Monday in Singapore

alixunderplatz:
The [image](https://www.ema.gov.sg/cmsmedia/irradiance/plot.png) was fixed and is working properly! :)

systemcatch:
@jarek happy with everything?

maxbellec:
loving this!

jarek:
This worked fine today morning for me as I recall. Changes look good, thanks!

PR #1139, opened 2018-02-23
Stop sudden production spikes in AU zones 

### LINKED ISSUE #704: Australia AEMO parser sometimes returns double the production
I'm guessing at some points, the source CSV contains information about production at two different timestamps (e.g. 03:30 as well as 03:25). The carbon intensity remains mostly correct, except the effect of solar production (which comes from a different source) is halved. Still, would be good to only return data for one timestamp per plant.

Compare the following screenshots, where generation from coal doubles from 5.32 GB to 10.6 GW - and so does the reported installed coal capacity.

au bug 1
au bug 2

systemcatch:
Make sure that only the most recent timestamp is used. @jarek do you see any problems with this approach?

fixes #704

### DIFF
--- a/parsers/AU.py
+++ b/parsers/AU.py
@@ -394,7 +394,6 @@ def fetch_production(country_code=None, session=None):
         # Parse the datetime and check it matches the most recent one.
         try:
             plant_timestamp = arrow.get(row['Most Recent Output Time (AEST)']).datetime
-            # TODO: We should check it's not too old..
         except (OSError, ValueError):
             # ignore invalid dates, they might be parsed as NaN
             continue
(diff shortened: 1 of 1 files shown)


parsers/AU.py
```diff
@@ -384,6 +391,19 @@ def fetch_production(country_code=None, session=None):
             print(row)
             continue
 
+        # Parse the datetime and check it matches the most recent one.
+        try:
+            plant_timestamp = arrow.get(row['Most Recent Output Time (AEST)']).datetime
+            # TODO: We should check it's not too old..
```
jarek: This TODO comment should now be up where you read in latest_ts. But maybe just remove it altogether, I think it's fine.

Issue #1122, opened 2018-02-19
Crimea is shown as part of Russia


systemcatch:
This is due to the natural earth shapefiles assigning administration of the area to Russia. We can alter this by using the state file instead and assigning Crimea to Ukraine. Lets discuss what we want to do here.

corradio:
Any statement by the UN about this? We should probably follow the position of the UN on this.

systemcatch:
They support Ukraine.
https://en.m.wikipedia.org/wiki/United_Nations_General_
Assembly_Resolution_68/262

jarek:
Do we know if UKRENERGO generation figures still include generation in Crimea? It looks like Crimea has about 500 MW of generation capacity and SO-UPS currently shows 757 MW of exports to Crimea (distinct from their report on interconnection to Ukraine). Perhaps we should have Crimea as a separate region altogether.

systemcatch:
https://sputniknews.com/russia/201601141033156976-crimea-switches-away-from-ukrainian-power/ suggests that Ukraine no longer is no longer supplying Crimea and that Russia is providing power.

systemcatch:
Maybe the simplest way to solve this is to create a new zone UKR-KR using data from http://br.so-ups.ru/Public/MainPageData/VSVGO/PowerVIEByRegions.aspx. Assuming that page updates scheduled to actual similar to the RU parser.

corradio:
If we can confirm that Crimea is not part of the RU parser, then let's create a new zone, and consider it as a "region" of Ukraine.

Maybe we should use the following codes?
https://en.wikipedia.org/wiki/ISO_3166-2:UA

systemcatch:
Yup those codes look good. I think that Crimea's generation is included in the RU parser at the moment which makes things difficult, @alixunderplatz do you think the generation schedule on http://br.so-ups.ru/Public/MainPageData/VSVGO/PowerVIEByRegions.aspx is updating once the hour has passed like in the main RU parser? I've been watching it but haven't seen anything yet.

alixunderplatz:
@systemcatch it should be included in the RU parser, that is correct, because it is part of price zone 1.

I just saw I gave the wrong link for thermal generation above - excuse me! The link in your comment is for the generation forecast for renewable energies (in this case, solar+wind), which probably remains constant through the day.
Here is the link for thermal generation (I fixed it now in my longer comment above):
[...]

PR #1154, opened 2018-02-26
Quick fix for Crimea

alixunderplatz:
Calling the zone "Russia & Crimea" is a possible temporary solution. 
Crimea is included in RU market data that is currently used and is exclusively connected to the RU electricity grid at the moment, so this is the "technical quick fix" for the common _bidding zone_ until parsers are adjusted/ready.
Ref #1122


systemcatch:
Good idea, we could even call it "Russia & Ukraine (Crimea) to remove any ambiguity.

PR #1155, opened 2018-02-27
Fix tw parser

### LINKED ISSUE #1148: Taiwan TW is offline
Currently, Taiwan is grey and 24-hours-history is empty as well.

- [The link ](http://data.taipower.com.tw/opendata01/apply/file/d006001/001.txt) in the [TW.py parser](https://github.com/tmrowco/electricitymap/blob/master/parsers/TW.py) seems to show data, though. 
Maybe there have been some crucial changes?

Some other TW related things that should be fixed:
- The source link on the electricitymap website for Taiwan is not shown / shown as "?".
image

- In general, the link in README.md will show 404 error and is leading nowhere. Seems like they updated/revised their website a bit?
Here is the website with the 10-min-generation mix that should be linked in README.md:
http://www.taipower.com.tw/tc/page.aspx?mid=206&cid=404&cchk=8ccc1918-8cae-4f40-a2d0-b43454f4f218

image

maxbellec:
Fixes #1148.
Thx @systemcatch and @alixunderplatz, it was indeed the `session` argument that is mandatory for `fetch_production`.


alixunderplatz:
@maxbellec @systemcatch 
Is it possible to add hydro pumped storage capacity (2602 MW) to the parser as static number just like this?:
 

```
'capacity': {
            'coal': coal_capacity,
            'gas': gas_capacity,
            'oil': oil_capacity,
            'hydro': production.ix['Hydro'].capacity,
            'hydro storage': 2602,
            'nuclear': production.ix['Nuclear'].capacity,
            'solar': production.ix['Solar'].capacity,
            'wind': production.ix['Wind'].capacity,
            'unknown': production.ix['Co-Gen'].capacity
        },
```

maxbellec:
I added the `hydro_storage` as suggested by @alixunderplatz. Maybe it would be better to add it in `zones.json`, but I'm not sure sure if capacities from `zones.json` and parsers are merged or if one overwrites the other.
So there might be a cleaner way, but I think this is fine.

maxbellec:
@corradio should be ready to merge

config/zones.json
```diff
@@ -1772,6 +1772,9 @@
     "timezone": "Europe/Istanbul"
   },
   "TW": {
+    "capacity": {
+      "hydro_storage": 2602
```
corradio: If I remember correctly the key is with a space instead of underscore. Can you check previous examples?

Issue #606, opened 2017-06-26
add Namibia


jarek:
http://www.nampower.com.na/Scada.aspx (linked from Nampower home page) has live generation and exchange data, but it's outputted as text within a PNG image. Optimally we would contact the utility and ask if they could publish the same data as text. Failing this, perhaps we could use some API to do 
optical character recognition (OCR), e.g. as in https://gist.github.com/dannguyen/a0b69c84ebc00c54c94d, but OCR is generally not trivial.

Mapping of the plant names to fuel at http://www.nampower.com.na/Page.aspx?p=182

It looks like most of Namibian electricity is imported from Eskom, the South African utility. ZESCO is the Zambian utility.

jarek:
There is also http://www.sappmarket.com/ which as far as I can tell provides a live or semi-live view onto the Southern African Power Pool. It seems like it should include the current generation as well, but I don't know the terms/lingo ("Day-Ahead Market", "Forward Physical Market", "APF-IDM", etc), so I don't know if I'm missing it or if it's not updated.

systemcatch:
A few notes on this.  I've run some preliminary tests using tesseract-ocr on generation images downloaded from Nampower's Scada dashboard.  Initial results are ok but not good enough to parse the data.  The images will have to be cleaned up before applying OCR on them.  It's likely a case of trial and error until the right method is found.

systemcatch:
Based on discussions in #817 with @alixunderplatz I can now get OCR to work reliably on the nampower scada images. The next step is to automate all of this into a parser.

@corradio 
There are some pros and cons with building this;

**Pro**
- First African country on the map (may lead to more sources/collaboration).
- OCR techniques can be reused in other parsers, allowing several issues to be resolved.
- The challenge is fun!

**Con**
- More python dependencies, PIL & pytesseract at a minimum.
- Have to install tesseract & image magick on the EM server.

Overall I think this is worth doing.

corradio:
how long does the parser take?

systemcatch:
It's not written yet but based on my tests the limiting factor will be the request to http://www.nampower.com.na/Scada.aspx . I expect this parser will be similar in speed to most of the current ones, maybe even a little faster than some.

jarek:
If getting tesseract on the EM parser server is a problem, I can try hosting a OCR proxy service for this and Singapore solar on my VPN

systemcatch:
[...]

PR #1170, opened 2018-03-01
Add Namibia

systemcatch:
After a lot of experimentation I've managed to get this working reliably. Namibia imports big amounts of power from South Africa and Zambia (no data for either yet) so it's CO2 intensity will be skewed somewhat.

/cc @alixunderplatz 

ref #606 

na


corradio:
Have you ever seen coal and oil values become > 0?

alixunderplatz:
@corradio I think this is something that might happen, but not very often. For example, when South Africa faces power shortages and load shedding and cannot supply Namibian demand. [On IEA](https://www.iea.org/statistics/statisticssearch/report/?country=NAMIBIA&product=electricityandheat&year=2015), it says 1500 GWh were generated by hydro, 27 GWh from oil and 7 GWh from coal in 2015. So fossil plants are operating barely. Wind and solar might be big enough to lower that generation even more.

alixunderplatz:
@systemcatch Just noticed, there are connections with Botswana and Angola, too!
It's in that 132 kV section:
- Omaere - Ghanzi is between Namibia -> Botswana
- Efundja - Ondjiva is between Namibia -> Angola

I think the header "cross border supplies" means, these places are connected to and supplied from the side of the Namibian system. So probably no export towards Namibia, but some more export could be displayed to other countries.

systemcatch:
It's weird that those aren't included under import/export though.

I've never seen the oil plants above 0. The coal plant often has -0.4 generation which is set to None by the parser.

corradio:
Maybe set coal to max(0, value) to make sure it is allowed to be >0 ?

systemcatch:
So you'd rather coal be 0.0 rather than None if it's negative?

corradio:
If it's only -0.45 we can assume it to be 0 I think? My main issue with
setting to None is to miss it the day it actually turns on.

systemcatch:
When it does turn on it will not be None. It is only set to None if
negative or unreadable.

corradio:
Perfect. That's even better.

parsers/NA.py
```diff
@@ -0,0 +1,183 @@
...
+          'unknown': 0.0
+      },
+      'storage': {
+          'hydro': -10.0,
+      },
+      'source': 'mysource.com'
+    }
+    """
+
+    raw_text = get_text_from_image(session=None, link=generation_link, \
```
jarek: Probably use `session` variable here and in fetch_exchange

systemcatch: so session=session to avoid resetting it again?

jarek: yeah, session=session will give get_text_from_image the same session object that fetch_production received

systemcatch:
@corradio happy with everything here? Let me know if the tesseract installation location is a problem and we'll find a workaround.

corradio:
All good on my side. Feel free to merge when you're good to go.

PR #1172, opened 2018-03-02
Fix AX flows

tmslaine:
Some changes motivated by information via email from Kraftnät Åland.
- Fixes incorrect assumption of export to Finland
- Unknown production cannot be estimated from the source. 
Details: see issue #883

### DIFF
--- a/parsers/AX.py
+++ b/parsers/AX.py
@@ -213,16 +213,16 @@ def _fetch_data(session=None):
     FIFlow = round(float(FIFlow),1)
 
 
-
+    # The shown total consumption is not reliable according to the TSO
     # Consumption
-    Cons = []
-    for x in range(650, 700-6):
-        for abr in shorts:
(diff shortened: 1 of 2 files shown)


parsers/AX.py
```diff
@@ -303,7 +306,7 @@ def fetch_production(country_code='AX', session=None):
         'source': 'kraftnat.aland.fi',
         'datetime': arrow.get(obj['fetchtime']).datetime
     }
-    data['production']['biomass'] = round(obj['unknown'],1)
+    data['production']['biomass'] = None
```
corradio: Maybe you should now assign `obj['unknown']` to `unknown` as it was removed from `biomass`?

tmslaine: We do not know the amount of any type of production except wind and oil based on the source. I can of course make obj['unknown'] equal None and make this change, but is there a point in doing so?

Issue #1069, opened 2018-02-06
Split NO and DK into bidding zones


corradio:
Requires #917, but could be great to start adding granularity!

tmslaine:
Different organizations have a bit different maps of NO bidding zones.
ENTSOE:
![entsoe-nordic](https://user-images.githubusercontent.com/25774514/36350124-d09bb75c-149a-11e8-8b28-f27cd3d09e38.png)
Norwegian TSO Statnett:
![statnett](https://user-images.githubusercontent.com/25774514/36350127-de7de412-149a-11e8-8735-4479650aee52.png)
Norwegian Water Resources and Energy Directorate NVE:
![nve-norway](https://user-images.githubusercontent.com/25774514/36350143-4beca786-149b-11e8-92ae-eef9352c8953.png)
I assume NVE has the most up-to-date map out of these ones, freehand-polygons from their map combined with ne10m_cultural gives:
![image](https://user-images.githubusercontent.com/25774514/36350180-61e58200-149c-11e8-9590-ab35abcc2a62.png)

For Denmark there is no such controversy. The ENTSOE map above agrees with what the TSO energinet shows, i.e. split on the Stor-Baelt bridge:
![energinet-dk](https://user-images.githubusercontent.com/25774514/36350287-cba1f1a4-149e-11e8-8fa2-520b8e5636a4.png)
This split also agrees with administrative zones, giving DK1 = Nordjylland, Midtjylland, Syddanmark and DK2 = Hovedstaden, Sjaelland
As Bornholm is already included as a separate zone, Bornholm can be excluded from DK2. The price should however be the same on "mainland" DK2 and Bornholm.
Again matching with ne_10m gives:
![image](https://user-images.githubusercontent.com/25774514/36350350-be908f56-149f-11e8-82c2-478948fd4ad6.png)

corradio:
I've started a branch at https://github.com/tmrowco/electricitymap/tree/olc/add-dk1-dk2 but I have some troubles excluding Bornholm (which is a subunit) from Sjælland (which is a state).

corradio:
I managed to split the geometries correctly. Can anyone help with installed capacities for DK1 / DK2?

alixunderplatz:
@corradio I'll take a look at capacities!

alixunderplatz:
This is **from ENTSO-E**, where numbers for 2016 and 2017 are exactly the same. So maybe not the most recent ones, but at least split by zones.

**For DK1:**
[...]

PR #1175, opened 2018-03-04
Split DK into DK1/DK2 with associated capacities and exchanges

corradio:
Ref #1069 

Split into DK1/DK2 with associated capacities and exchanges

image


parsers/lib/quality.py
```diff
@@ -41,8 +41,9 @@ def validate_production(obj, country_code):
     if (obj.get('production', {}).get('unknown', None) is None and
         obj.get('production', {}).get('coal', None) is None and
         obj.get('production', {}).get('oil', None) is None and
+        obj.get('production', {}).get('gas', None) is None and
```
systemcatch: Good change since the world is using gas more.

web/topogen.sh
```diff
@@ -46,7 +46,7 @@ echo "Extract metadata"
 echo 'Parsing countries..'
 ("$NODE_MODULES_PATH/shp2json" -n build/${COUNTRIES_FILENAME}.shp \
   | "$NODE_MODULES_PATH/ndjson-map" '(d.id = (d.properties.ISO_A3 != "-99") ? d.properties.ISO_A3 : d.properties.ADM0_A3, d.oldProps = d.properties, d.properties = {}, d.properties.subid = d.oldProps.SU_A3, d.properties.code_hasc = d.oldProps.code_hasc, delete d.oldProps, d)' \
-  | "$NODE_MODULES_PATH/ndjson-filter" "d.id && ${STATES_FILTER}.indexOf(d.id) === -1" \
```
systemcatch: No longer needed?

corradio: This will exclude countries that are in the states list. That's a harsh constraint. We might want to be able to access country geometries even though we split it by state.
Also we don't really need to filter anything at this stage anymore as we do not pass the whole list of possible geometries but instead keep only the ones we need.

corradio:
@brunolajoie look here:
https://github.com/tmrowco/electricitymap/pull/1175/files#diff-9196b58d1cdd3db6a09227448696ba09R640
Exchanges are given by Statnett and forecasts by ENTSOE

systemcatch:
I think we are good to go apart from the conflicts?

corradio:
Ifs blocked for now as we need to update a few things on the backend.
Should ready soon though!

Olivier

Issue #1003, opened 2018-01-15
Parser dashboard for real time data availability


brunolajoie:
I suggest creating a data quality dashboard accessible to everyone, in order to help contributors know which parsers are working and which one needs to be fixed.

Ideally, It could be accessible via a specific button on the electricity map welcoming left panel, with the following information (to start with)

- List of non-working parsers for power production and for exchanges (real time)
- Error message
- Link to the parser file on github directory

Step 2, think of how to display this dashboard for the past 24h or longer

systemcatch:
Further idea for this. The London Underground has a nice design showing service status for all lines. I'd recommend separating out the capacities from this avoid too much clutter.

![tube_status](https://user-images.githubusercontent.com/30196510/36641288-4bf22ce4-1a25-11e8-96f5-c3d6e130ed7b.png)

systemcatch:
oli's suggestion so it doesn't get lost.

>Maybe a simple first step would be to set up an ELK stack (ElasticSearch + Kibana) on which we can see logs of parsers on two levels (warnings + errors).

corradio:
FYI we have an internal PR ready about this, so it will be up soon.

corradio:
@maxbellec and I finished pushing it.
You can see a log history of warnings and errors here:
https://kibana.electricitymap.org/app/kibana#/discover/1710fdd0-2460-11e8-a779-9d01de8d7a71?_g=(refreshInterval:('$$hashKey':'object:6765',display:'10%20seconds',pause:!f,section:1,value:10000),time:(from:now-15m,mode:quick,to:now))&_a=(columns:!(level,extra.path,message),filters:!(('$state':(store:appState),exists:(field:level),meta:(alias:!n,disabled:!f,index:'93e631f0-245f-11e8-a779-9d01de8d7a71',key:level,negate:!f,type:exists,value:exists)),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'93e631f0-245f-11e8-a779-9d01de8d7a71',key:level,negate:!t,params:(query:INFO,type:phrase),type:phrase,value:INFO),query:(match:(level:(query:INFO,type:phrase))))),index:'93e631f0-245f-11e8-a779-9d01de8d7a71',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc))
and you can see a heatmap/dashboard here:
[...]

PR #1178, opened 2018-03-05
Add the possibility to parse historical data to parsers

maxbellec:
When we lost some data or we have a new source where we can parse historical data, we want parsers to be able to parse for a specific datetime.

The idea is to have all parsers have an additional `target_datetime` argument (by default `None`, which would mean last available date). When historical data is not possible / has not been implemented and `target_datetime` has been provided, we raise a `NotImplementedError`.

Also, we want to be all parser logs publicly available, so the parsers should also have a `logger` argument to write to. I'll also add it and, where existing, try to convert `print`s to use `logger`.


corradio:
Works perfectly in my opinion. Things left to do on the top of my head:
- update the example.py example parser
- go through the two READMEs (in / and in /parsers) to see if anything needs to be updated
- update the other parsers with the proper signature

(later)
- update all parsers to take into account new parameters

So excited about this!

maxbellec:
- [ ] update `example.py` example parser
- [ ]  through the two READMEs (in / and in /parsers) to see if anything needs to be updated
- [ ] update all parsers with the proper signature
- [ ] change parameter name `country_code` to `zone_key`

maxbellec:
I believe it's ready but it's late so I might have done bad things. Let's wait 'till tmrow morning

maxbellec:
screenshot at 2018-03-07 19-48-10

maxbellec:
for now, only ENTSOE parsers can parse a specific datetime. I believe we can already merge with this, and I'll continue with some other, with help from the community if possible

corradio:
Ref #1003. @jarek @systemcatch can you take a quick look to see if anything pops to your mind. Looks good on my side.

parsers/MD.py
```diff
@@ -8,24 +8,26 @@
...
+def get_data(session=None):
     """ Returns generation data as a list of floats."""
 
     s = session or requests.Session()
 
-    #In order for the data url to return data, cookies from the display url must be obtained then reused.
-    response = s.get(display_url)
+    # In order for the data url to return data, cookies from the display url must be
+    # obtained then reused.
+    s.get(display_url)
```
systemcatch: This can be removed.

parsers/example.py
```diff
@@ -1,16 +1,30 @@
...
 
     Arguments:
-    country_code (optional) -- used in case a parser is able to fetch multiple countries
-    session (optional)      -- request session passed in order to re-use an existing session
+    zone_key: used in case a parser is able to fetch multiple countries
+    session: request session passed in order to re-use an existing session
+    target_datetime: the datetime for which we want production data. If not provided, we should
+      default it to now. If past data is not available, raise a NotImplementedError. Beware that the
+      provided target_datetime is UTC.
+    logger: an instance of a `logging.Logger`. Information logged will be publicly available so that
```
systemcatch: Nice.

parsers/test/lib/test_countrycode.py
```diff
@@ -1,28 +1,28 @@
 import unittest
```
systemcatch: Shall we rename the test as well?

parsers/AX.py
```diff
@@ -347,25 +353,27 @@ def fetch_exchange(country_code1, country_code2, session=None):
...
     obj = _fetch_data(session)
 
     data = {
-        'sortedCountryCodes': '->'.join(sorted([country_code1, country_code2])),
+        'sortedCountryCodes': '->'.join(sorted([zone_key1, zone_key2])),
         'source': 'kraftnat.aland.fi',
         'datetime': arrow.get(obj['fetchtime']).datetime
     }
 
     # Country codes are sorted in order to enable easier indexing in the database
```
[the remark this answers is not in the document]
maxbellec: @systemcatch not sure what you mean here. Do you want to replace `sortedCountryCodes` by `sortedZoneKeys`? (I certainly do and will)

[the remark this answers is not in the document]
systemcatch: The comment says # Country codes are sorted... rather than zone codes are sorted, but tbh I don't think this matters much.

parsers/example.py
```diff
@@ -1,21 +1,35 @@
 #!/usr/bin/env python3
+import logging
+import datetime
+
 # The arrow library is used to handle datetimes
+from json import JSONDecodeError
```
jarek: This import doesn't appear to be used

maxbellec:
thanks @jarek, I didn't know about `inspect`. It probably would have been a better solution indeed.

PR #1179, opened 2018-03-05
Simplify exceptions for aruba

### LINKED ISSUE #1153: Some small islands like Aruba are not displayed on map
The simplification process contained into generate-geometries.js simplifies the geometries of the entire world by using a weight of 0.01.
This threshold is so high that some small islands like Aruba (and others from Caribbean) are simply removed from map.

With this threshold of 0.01, the generated file world.json weights 420kb.

With a threshold of 0.002, Aruba come back into map but the  world.json grows up to 896kb. E-Map still display well on my computer.

The good deal would be to :
- apply a threshold of 0.01 to all countries but Aruba.
- apply a threshold of 0.002 for Aruba
- merge the two previous in world.json

I tried to figure out but didn't succeed using topojson. My guess was that topojson.mergeArcs would be a could candidate for merging two topojson.

Please comment on this issue how to merge the two topojson or any other idea.

pascalheraud:
Fixes #1153 

I added a parameter "lessSimplify" to zones to avoid too much simplification on small zones like Aruba.

Diffing world.json is very hard due to huge single line. As far as I can see, it only adds Aruba. I did not notice any other changes.

I also added unzip to docker container that is required by topogen.sh. Some other various files needed for development using container were added too.

Another point is the third_party_maps folder that is missing inside container and causes download map not working.

This PR works well. Naming and comments can be changed according to your coding standards.


web/generate-geometries.js
```diff
@@ -551,13 +552,28 @@ const toListOfFeatures = (zones) => {
...
+  	zones[zone.zoneName] = getDataForZone(zone, true);
+  }
+});
+
+// create zonesLessSimplified by getting zone having lessSimplify===true
+let zonesLessSimplified = {};
+zoneDefinitions.forEach(zone => {
+  // Take only zones being small simplified
+  if (('lessSimplify' in zone) && zone.lessSimplify) {
+	if (zone.zoneName in zonesLessSimplified || zone.zoneName in zones)
```
corradio: Identation problems?

docker-compose.yml
```diff
@@ -27,3 +27,8 @@ services:
       - './web/webpack.config.js:/home/web/webpack.config.js'
       - './web/locales:/home/web/locales'
       - './config:/home/config'
+      - './web/topogen.sh:/home/web/topogen.sh'
```
corradio: can you sort alphabetically?

web/generate-geometries.js
```diff
@@ -580,11 +596,44 @@ zoneFeatures = toListOfFeatures(zoneFeaturesInline);
...
+		// Relocate arc into zone by adding srcArcsLength
+		topo1.objects[zoneName].arcs.forEach(arcList1=>{
+			arcList1.forEach(arcList2=>{
+				for(i=0; i<arcList2.length; i++) {
+					arcList2[i]+=srcArcsLength;
+				}
+			});
+		});
+	});
+
```
maxbellec: can you remove one or both blank lines?

web/generate-geometries.js
```diff
@@ -580,11 +596,44 @@ zoneFeatures = toListOfFeatures(zoneFeaturesInline);
...
+			});
+		});
+	});
+	
+	
+	// Add arcs from topo2
+	topo2.arcs.forEach(arc=> {
+		topo1.arcs.push(arc);
+	});
+
```
maxbellec: again, can you remove both blank lines?

web/generate-geometries.js
```diff
@@ -580,11 +596,44 @@ zoneFeatures = toListOfFeatures(zoneFeaturesInline);
 // Write unsimplified list of geojson, without state merges
 fs.writeFileSync('build/zonegeometries.json', zoneFeatures.map(JSON.stringify).join('\n'));
 
-// Simplify
+// Adds the zones from topo2 to topo1
+function mergeTopoJsonSingleZone(topo1, topo2) {
```
maxbellec: can you move the function definition above the script part (to line ~550)?

maxbellec:
this is great! I'd also like to see a screenshot of Aruba Island. Could you also post of Acores or Canaries so that we can see the differences? It might be nice to have a bit more precision on a few island (but we might not see a big difference due to the quantization step).

Not a necessary change, I'm not a fan of the `lessSimplify` name. Though it's clear for someone that knows we're using `topojson.simplify`, maybe `moreDetails` would be more explicit? Anyway, keep the name you prefer.

web/generate-geometries.js
pascalheraud: ok

docker-compose.yml
pascalheraud: sure

web/generate-geometries.js
pascalheraud: OK

web/generate-geometries.js
pascalheraud: OK

web/generate-geometries.js
pascalheraud: OK

pascalheraud: I changed to moreDetails, zonesMoreDetails and topoMoreDetails that seems better for me as well.

pascalheraud:
Here is a screen shot of Aruba using moreDetails:true : 
image

pascalheraud:
I tried to use moreDetails on Azores or Canaries but it produce the same geometries. As you said, another operation may changes geometries.

docker-compose.yml
```diff
@@ -19,11 +19,16 @@ services:
       - NODE_ENV=development
     ports: ['8000:8000']
     volumes:
+      - './config:/home/config'
+      - './web/generate-geometries.js:/home/web/generate-geometries.js'
       - './web/src:/home/web/src'
+      - './web/locales:/home/web/locales'
```
corradio: Can you order alphabetically?

Issue #657, opened 2017-08-01
Mobile interfaces should have floating wind/solar buttons


corradio:

jbuverud:
Fixed in #1187

PR #1187, opened 2018-03-09
Wind/solar buttons and legends panel UI update

jbuverud:
Solar/wind layer toggle buttons have been moved onto the map itself and styled like the existing zoom buttons. The buttons have grey icons when the corresponding layer is off, and black icons when the layer is on.

A new "legend panel" has been implemented in the bottom left, which now contains the CO2-bar as well as the wind and solar potential bars (these have been removed from the left-side panel). The solar/wind bars are shown/hidden automatically when the layers are toggled on or off. 

Desktop view:

image

On mobile devices, the old wind/solar buttons on the tab bar have removed, and the legend panel is expandable/collapsable. Its inital state is collapsed.

Mobile view (legend collapsed):

image

Mobile view (legend expanded):

image

Comments on the UX/design/whatever are very welcome :)


web/src/reducers/index.js
```diff
@@ -23,6 +23,7 @@ const initialApplicationState = {
   solarEnabled: Cookies.get('solarEnabled') === 'true' || false,
   useRemoteEndpoint: document.domain === '' || isLocalhost,
   windEnabled: Cookies.get('windEnabled') === 'true' || false,
+  legendVisible: false,
```
corradio: can you order alphabetically?

corradio:
Fix #657

corradio:
Thanks for doing this!

PR #1192, opened 2018-03-10
Add limit for remove_negative and fix zoneKey error

systemcatch:
I used countryCode instead of zoneKey in #1191. Also added a limit to remove_negative as bigger negative numbers likely indicate something is wrong and requires attention.

### DIFF
--- a/parsers/lib/validation.py
+++ b/parsers/lib/validation.py
@@ -2,8 +2,10 @@
 
 """Centralised validation function for all parsers."""
 
+from logging import getLogger
 
-def validate(datapoint, **kwargs):
+
+def validate(datapoint, logger=getLogger(__name__), **kwargs):
     """
     Validates a production datapoint based on given constraints.
     If the datapoint is found to be invalid then None is returned.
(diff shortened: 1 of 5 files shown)


parsers/lib/validation.py
```diff
@@ -71,8 +71,8 @@ def validate(datapoint, **kwargs):
 
     if remove_negative:
         for key, val in generation.items():
-            if val is not None and val < 0.0:
-                print("{} returned negative value for {}".format(key,datapoint['countryCode']))
+            if val is not None and -5.0 < val < 0.0:
+                print("{} returned {} for {}, setting to None".format(key, val, datapoint['zoneKey']))
```
corradio: can we use `logger.warn` to log the warning?

systemcatch: sure I will do the same for the other prints, do I put logger=None in the function args?

corradio:
if you us the logger and set logger=None we might get some errors if no
arguments is passed. Better force the logger argument?

systemcatch:
ok so have it like def validate(datapoint, logger, **kwargs): then change the validate uses in each parser?

corradio:
Yes indeed

systemcatch:
When altering the validate line in IN-GJ to:
```data = validate(data, logger, required=['hydro'], floor=10.0)```

The following happens because logger=None is the default when run as main.
```
(EM-env) chris@ThinkPad:~/electricitymap$ python -m parsers.IN_GJ
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/chris/electricitymap/parsers/IN_GJ.py", line 138, in <module>
    print(fetch_production('IN-GJ', session))
  File "/home/chris/electricitymap/parsers/IN_GJ.py", line 109, in fetch_production
    valid_data = validate(data, logger, remove_negative=True)
  File "/home/chris/electricitymap/parsers/lib/validation.py", line 77, in validate
    logger.warning("{} returned {}, setting to None".format(
AttributeError: 'NoneType' object has no attribute 'warning'
```
To get around this we could construct a temporary logger when running from the command line.
```
if __name__ == '__main__':
    logger = logging.getLogger(__name__)
    print(fetch_production())

corradio:
Or maybe the temporary logger could be set as default argument?

systemcatch:
Yes that is probably better as update_state.py does not run the functions as main so it'll fail.

systemcatch:
@corradio should be good to go here, let me know if you want any changes.

PR #1193, opened 2018-03-11
improve CA_NS validation and logging

jarek:
Hello,

Here are a few improvements to Nova Scotia:

* We are now following the validation logic used on their website (https://www.nspower.ca/site/renewables/assets/js/site.js), plus checking that reported generation doesn't go greatly over known capacity. The JSON data is sometimes just plain wrong, and doesn't correct afterward.
* Exchange parser now returns all available data rather than only the latest point.
* We now use logger. See question attached below about whether it should be warning or info.

Screenshot of problem:

ns

### DIFF
--- a/parsers/CA_NS.py
+++ b/parsers/CA_NS.py
@@ -6,8 +6,34 @@ import arrow
 import requests
 
 
-def _get_ns_info(requests_obj):
+def _get_ns_info(requests_obj, logger):
     zone_key = 'CA-NS'
+
+    # This is based on validation logic in https://www.nspower.ca/site/renewables/assets/js/site.js
+    # In practical terms, I've seen hydro production go way too high (>70%) which is way more
+    # than reported capacity.
+    valid_percent = {
+        # The validation JS reports error when Solid Fuel (coal) is over 85%,
+        # but as far as I can tell, that can actually be a valid result, I've seen it a few times.
+        # Use 98% instead.
+        'coal': (0.25, 0.98),
+        'gas': (0, 0.5),
+        'biomass': (0, 0.15),
+        'hydro': (0, 0.60),
+        'wind': (0, 0.55),
+        'imports': (0, 0.20)
+    }
+
+    # Sanity checks: verify that reported production doesn't exceed listed capacity by a lot.
+    # In particular, we've seen error cases where hydro production ends up calculated as 900 MW
+    # which greatly exceeds known capacity of 418 MW.
+    valid_absolute = {
+        'coal': 1300,
+        'gas': 700,
+        'biomass': 100,
+        'hydro': 500,
+        'wind': 700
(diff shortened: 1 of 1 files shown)


parsers/CA_NS.py
```diff
@@ -18,35 +44,73 @@ def _get_ns_info(requests_obj):
...
+        # validate again
+        valid = True
+        for gen_type, value in electricity_mix.items():
+            absolute_bound = valid_absolute.get(gen_type)  # imports are not in valid_absolute
+            if absolute_bound and value > absolute_bound:
+                valid = False
+                logger.warning(
+                    'CA-NS: discarding datapoint at {dt} '
+                    'due to {fuel} too high: {value}'.format(
+                        dt=data_date, fuel=gen_type, value=value))
```
jarek: This isn't a fatal problem, it happens every now and then due to data source problems, and we recover from it gracefully. Should we use logger.warning or logger.info?

parsers/CA_NS.py
```diff
@@ -18,35 +44,73 @@ def _get_ns_info(requests_obj):
...
             # if not found, assume 1244 MW, based on average yearly electricity available for use
             # in 2014 and 2015 (Statistics Canada table Table 127-0008 for Nova Scotia)
-            load = 1244 / 100.0
-
-        # datetime is in format '/Date(1493924400000)/'
-        # get the timestamp 1493924400 (cutting out last three zeros as well)
-        data_timestamp = int(mix['datetime'][6:-5])
-        data_date = arrow.get(data_timestamp).datetime
+            load = 1244
+            logger.warning('CA-NS: unable to find load for {}, assuming 1244 MW'.format(data_date))
```
corradio: can you add `extra={'key': 'CA-NS'}` as extra argument to make sure the zone code gets registered?

parsers/CA_NS.py
corradio: Warning seems like the proper level to me, as you're discarding data.

parsers/CA_NS.py
jarek: Sure - how should it look in total? Is there an existing example? Like this?

```
logger.warning(
    'CA-NS: unable to find load for {}, assuming 1244 MW'.format(data_date),
    extra={'key':'CA-NS'})
```

And do that in all warnings I log?

parsers/CA_NS.py
jarek: Yeah, on second thought, everything that explains why there's no data makes sense as a warning

parsers/CA_NS.py
corradio: exactly. You can even drop the `CA-NS: ` prefix from the message.

jarek:
Let's hope this works!

PR #1194, opened 2018-03-12
Added ad in header for pro version integration

jbuverud:


brunolajoie:
Can you post a screenshot plz? thanks!

jbuverud:
image

Link mouseover:

image

brunolajoie:
Plz just reduce text size a bit and make "contact-us" link similar to those in the left panel (blue by default)

jbuverud:
Sure. Made it mimic the links on the left: 

image

corradio:
Ok we need to think more about this as this won't work on mobile (and I think it's quite aggressive).
@jbuverud let's brainstorm tomorrow together.

jbuverud:
Sure thing. Note that I implemented it as hidden on mobile. I agree that it is quite aggressive though.

systemcatch:
When I look at the map it's not particularly obvious that any data is for sale or that an API could be available. I know that clicking database or forecast will take people to the buying sections but personally I wouldn't assume they will get that far.

This change is a good idea but putting it in the left panel is more natural than right at the top imo.

jbuverud:
@corradio and I agreed today that keeping the ad message in the top header is a good idea for now, despite it being slightly intrusive. I suggest we keep it there for a bit and see how well it works. In the future we should probably also have links in our left panel, but I suggest we wait with this until the left panel has been updated further (it is due to change a bit in the near future). 

Because we want to advertise both our database and forecasting services, I've added two separate ad messages, one of which is randomly chosen for being displayed in the header on each page load:

Message 1: 
"Looking for historical data? Check out our Database!"

Message 2:
"Want live data in your app? Try our Forecast API!"

The messages now link to our database/forecast sites instead of our email. 

Any comments on the wording are very welcome. I tried to word the two messages a bit differently so the users are more likely to notice if the messages has been changed on a reload.

corradio:
A like the wording. Some extra suggestions for message 2 (although I do realise it adds complexity):
"Want live data in your app *or on your device*? Try our ~Forecast~ API".
This also ties in with the internal discussion we're having about renaming the Forecast product into something else. But at least what you're proposing is a very good start.

brunolajoie:
Ah very nice idea for the random message choice.
Any news on that PR?

jbuverud:
I've updated the forecast ad text to reflect the changes suggested by @corradio. I think this is a good first step until the Forecast product is potentially renamed. Let me know if there's any other changes you'd like to see.

web/src/styles.css
```diff
@@ -146,7 +146,7 @@ body {
     -moz-box-shadow: 0 0 5px 1px rgba(0,0,0,0.3);
 }
 #header #header-content {
-    display: block;
+    display: flex;
```
corradio: Is that required? Changing the display type will require us to test on old devices (including my old iPad) because some bugs might arise..

web/views/pages/index.ejs
```diff
@@ -111,6 +111,11 @@ co2Sub = function(str) {
                         </span>
                     </div>
 
+                    <div class="pro-ad-header">
```
corradio: Can you remove the `pro-` prefix in general? It's not really related to "pro" (which is not even one of our products anymore).
Also, did you make sure it's not shown on small devices? There's a CSS class you can add that does exactly that.

web/src/styles.css
jbuverud: It makes it much easier to center the text properly when resizing. According to https://caniuse.com/#search=flex it would only be a potential issue in IE, and I've made sure it works correctly at least in IE11. If we're worried about support for older versions of IE then it might be a problem. What version is your ipad? I could also try to test on mine.

corradio:
I have the first iPad retina. I'll do a check when I have the time.

web/views/pages/index.ejs
jbuverud: Sure, I'll remove the pro prefix.

And yes (like I mentioned previously in this thread 😃), it is not shown on small devices. I'll update it to use the same visibility class though.

web/locales/en.json
```diff
@@ -65,7 +65,9 @@
         "lastupdate": "Updated",
         "oops": "Oops! We are having trouble reaching the server. We will try again in a few seconds.",
         "newversion": "A new version is available! Press <a onClick=\"location.reload(true);\">here</a> to reload.",
-        "retrynow": "Retry now"
+        "retrynow": "Retry now",
+        "database-ad": "Looking for historical data? <a href=\"https://data.electricitymap.org\">Check out our Database!</a>",
```
corradio: Can you add `&utm_source=electricitymap.org&utm_medium=referral` to each link? That's what we've been using so far in the ejs to properly mark the referrals.
Also, should we have `target=_blank`? What do you think?

jbuverud: Ah yes, I had completely forgotten our convo about the referral codes, my bad. :) 

Opening in new tab/window is a great idea - I'll add it.

corradio:
The "ad" is not centred here:

image
It seems to be fixed by removing `margin-right: auto` on `.logo`

corradio:
However it works fine on my iPad!

corradio:
@jbuverud we've now renamed Forecast to API

PR #1211, opened 2018-03-14
FR parser fetches whole day and is able to go back in time

corradio:


maxbellec:
could you just limit number of characters per line to 100?

PR #1217, opened 2018-03-15
Fetch ENTSOE per unit

corradio:
Added a fetch method that can fetch production of individual units.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -729,19 +729,21 @@ def fetch_production_per_units(zone_key, session=None, target_datetime=None, log
     # Iterate over all psr types
     for k in ENTSOE_PARAMETER_DESC.keys():
         try:
-            values = [ v for v in parse_production_per_units(
-                query_production_per_units(k, domain, session, target_datetime)) or [] if v is not None ]
+            values = parse_production_per_units(
+                query_production_per_units(k, domain, session, target_datetime)) or []
             for v in values:
-                if not v: continue
+                if not v:
+                    continue
                 v['datetime'] = v['datetime'].datetime
                 v['source'] = 'entsoe.eu'
                 if not v['unitName'] in ENTSOE_UNITS_TO_ZONE:
                     logger.warning('Unknown unit %s with id %s' % (v['unitName'], v['unitKey']))
                 else:
                     v['zoneKey'] = ENTSOE_UNITS_TO_ZONE[v['unitName']]
-            if values:
-                data.extend([ v for v in values if v.get('zoneKey', None) == zone_key ])
-        except QueryError as e: pass
+                    if v['zoneKey'] == zone_key:
+                        data.append(v)
+        except QueryError:
+            pass
 
     return data
 
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -115,7 +115,8 @@ ENTSOE_DOMAIN_MAPPINGS = {
 }
 # Generation per unit can only be obtained at EIC (Control Area) level
 ENTSOE_EIC_MAPPING = {
-    'DK': '10Y1001A1001A796'
+    'DK-DK1': '10Y1001A1001A796',
+    'DK-DK2': '10Y1001A1001A796',
 }
 
 # Some exchanges require specific domains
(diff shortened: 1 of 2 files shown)


parsers/ENTSOE.py
```diff
@@ -148,8 +190,6 @@ def query_ENTSOE(session, params, target_datetime=None, span=(-48, 24)):
     if target_datetime is None :
         target_datetime = arrow.utcnow()
     else:
-        # for a specific datetime, fetch 24h values starting from target datetime
-        span = (-1, 48)
```
maxbellec: is there a reason to remove this? Changing it in to span = (-48, 1) in a PR (soon)

maxbellec: Alright, it's because of `query_production_per_unit`.
Then I suggest changing to (-24, 0) whenever we call with `target_datetime` (and changing values in `feeder.lib.fetch_data.REFETCH_SUPPORT` to `1D` instead of `2D` for ENTSO).

parsers/ENTSOE.py
```diff
@@ -622,6 +717,35 @@ def fetch_production(zone_key, session=None, target_datetime=None, logger=loggin
...
+        session = requests.session()
+    domain = ENTSOE_EIC_MAPPING[zone_key]
+    data = []
+    # Iterate over all psr types
+    for k in ENTSOE_PARAMETER_DESC.keys():
+        try:
+            values = [ v for v in parse_production_per_units(
+                query_production_per_units(k, domain, session, target_datetime)) or [] if v is not None ]
+            for v in values:
+                if not v: continue
```
maxbellec: You're doing two loops in a row over values (counting the list comprehension), twice check if `v` is truthy. One could be enough - I'd remove the list comprehension?
Small detail, PEP 8 recommends avoiding one-liner `if`s.

parsers/ENTSOE.py
```diff
@@ -622,6 +717,35 @@ def fetch_production(zone_key, session=None, target_datetime=None, logger=loggin
...
+                v['datetime'] = v['datetime'].datetime
+                v['source'] = 'entsoe.eu'
+                if not v['unitName'] in ENTSOE_UNITS_TO_ZONE:
+                    logger.warning('Unknown unit %s with id %s' % (v['unitName'], v['unitKey']))
+                else:
+                    v['zoneKey'] = ENTSOE_UNITS_TO_ZONE[v['unitName']]
+            if values:
+                data.extend([ v for v in values if v.get('zoneKey', None) == zone_key ])
+        except QueryError as e: pass
```
maxbellec: Again small detail, PEP 8 recommends avoiding one-liner `except`s.

parsers/ENTSOE.py
maxbellec: does you linter not take PEP 8 recommandations into account ?

parsers/ENTSOE.py
```diff
@@ -622,6 +717,35 @@ def fetch_production(zone_key, session=None, target_datetime=None, logger=loggin
...
+            for v in values:
+                if not v: continue
+                v['datetime'] = v['datetime'].datetime
+                v['source'] = 'entsoe.eu'
+                if not v['unitName'] in ENTSOE_UNITS_TO_ZONE:
+                    logger.warning('Unknown unit %s with id %s' % (v['unitName'], v['unitKey']))
+                else:
+                    v['zoneKey'] = ENTSOE_UNITS_TO_ZONE[v['unitName']]
+            if values:
+                data.extend([ v for v in values if v.get('zoneKey', None) == zone_key ])
```
maxbellec: could you not remove the values not corresponding to the current `zone_key` while in the loop to avoid another list comprehension?
(and after / before brackets are not `PEP 8`-y)

parsers/ENTSOE.py
corradio: Why force the span inside the code? I find that counter-intuitive given that it's also given as a function argument (so it shouldn't be silently overridden).
Any reason to have it override?

parsers/ENTSOE.py
corradio: Very true. My linter is pretty crappy. I should take some time to upgrade it but basically I'm on my own..

corradio:
@maxbellec did the requested changes

parsers/ENTSOE.py
maxbellec: the reason was that `query_entsoe` is called in many places. So that when fetching for a `target_datetime`, we didn't have to specify the span since it's always the same. But I agree that it's not explicit, and Guido says `explicit is better than implicit`.

Issue #1197, opened 2018-03-12
Add missing exchange arrows in Europe


alixunderplatz:
Even though there is no data at the moment for some of the zones, I think it's nice to complete the cross-border exchanges with data from ENTSO-E, that are not yet displayed.
Maybe values were simply not available in the past, so consider this a list of possible updates for exchanges :)

1.) [BE-LU](https://transparency.entsoe.eu/transmission-domain/physicalFlow/show?name=&defaultValue=false&viewType=TABLE&areaType=BORDER_CTY&atch=false&dateTime.dateTime=12.03.2018+00:00|CET|DAY&border.values=CTY|10YLU-CEGEDEL-NQ!CTY_CTY|10YLU-CEGEDEL-NQ_CTY_CTY|10YBE----------2&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2))
2.) [DE-LU](https://transparency.entsoe.eu/transmission-domain/physicalFlow/show?name=&defaultValue=false&viewType=TABLE&areaType=BORDER_CTY&atch=false&dateTime.dateTime=12.03.2018+00:00|CET|DAY&border.values=CTY|10YLU-CEGEDEL-NQ!CTY_CTY|10YLU-CEGEDEL-NQ_CTY_CTY|10Y1001A1001A83F&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2))
3.) [UA-BY](https://transparency.entsoe.eu/transmission-domain/physicalFlow/show?name=&defaultValue=false&viewType=TABLE&areaType=BORDER_CTY&atch=false&dateTime.dateTime=12.03.2018+00:00|CET|DAY&border.values=CTY|10Y1001C--00003F!CTY_CTY|10Y1001C--00003F_CTY_CTY|BY&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2))
4.) IT-MT (exchange data is there if you select Malta on the electricitymap, just the arrow must be added)

corradio:
Good initiative! However I think that the LU arrows will never be relevant because LU is part of another bidding zone?

alixunderplatz:
@corradio LU is part of the bidding zone DE-AT-LU. This only means, the electricity price in these three countries is always the same. However, exchanges between the _countries_ are given seperately.
[...]

PR #1228, opened 2018-03-17
Add new exchanges and add entry to exchange domain override (#1197)

lorrieq:
* Add BE->LU, DE->LU, IT->MT and BY->UA to exchanges

* Add exchange domain override for BY->UA


corradio:
Thanks @lorrieq . Let's also not forget to reference the issue https://github.com/tmrowco/electricitymap/issues/1197

PR #1230, opened 2018-03-18
QC to NEISO exchange

jarek:
This usually provides around 1-1.5 GW to NEISO.

I also added ability to get historical data.

NEISO website and API is unreachable from many countries, but seems to work from Ireland at least, including via everyone's favourite cloud provider.

@systemcatch tagging you as author of the parser, can you verify?

@maxbellec @corradio can you confirm the assumption: The provided target_datetime is timezone-aware in UTC.


parsers/US_NEISO.py
```diff
@@ -138,8 +147,98 @@ def fetch_production(zone_key='US-NEISO', session=None, target_datetime=None, lo
...
+        multiplier = 1
+
+        postdata = {
+            '_nstmp_zone0': '4014',  # ".I.SHOREHAM138 99"
+            '_nstmp_zone1': '4017',  # ".I.NRTHPORT138 5"
+            '_nstmp_zone2': '4011'  # ".I.ROSETON 345 1"
+        }
+
+    else:
+        raise Exception('Exchange pair not supported: {}'.format(sorted_zone_keys))
```
systemcatch: I suggest we raise NotImplementedError here rather than a generic exception.

parsers/US_NEISO.py
```diff
@@ -138,8 +147,98 @@ def fetch_production(zone_key='US-NEISO', session=None, target_datetime=None, lo
...
+        for timestamp, value in summed_exchanges.items()
+    ]
+
+    return result
+
+
 if __name__ == '__main__':
     """Main method, never used by the Electricity Map backend, but handy for testing."""
 
+    from pprint import pprint
```
systemcatch: Like the use of pprint for readability, I suppose importing in line avoids the cost when the module isn't run as main?

parsers/US_NEISO.py
jarek: Wouldn't that potentially get confused with NotImplementedError from parsers that don't support target_datetime? What do you think, Oli and Max?

parsers/US_NEISO.py
jarek: I like importing it here to indicate that it's only used in the testing method. I doubt it makes any performance difference. Pylint doesn't complain about it, so I guess it's okay?

jarek:
BTW, when requesting target_datetime for which data is not available, an empty list is returned.

```
fetch_exchange("US-NEISO", "CA-QC", target_datetime=arrow.get("2007-03-13T12:00Z")) ->
[]
```

That's a falsey value in Python. Does that work, or should I explicitly return None?

parsers/US_NEISO.py
systemcatch: Yes it could though most of the parsers already raise NotImplementedError when an invalid exchange is provided. With a good message it's not an issue imo.

systemcatch:
Hmmm I'm not sure on that. https://github.com/tmrowco/electricitymap/blob/master/parsers/lib/quality.py#L13 will raise an error on an empty list. But the datapoint isn't really wrong (no data is not an error for the db). Maybe it should return the usual exchange dict with netflow = None?

parsers/US_NEISO.py
systemcatch: I think it's fine.

systemcatch:
Build looks good!
qbnei

parsers/US_NEISO.py
maxbellec: `NotImplementedError` is fine

parsers/US_NEISO.py
```diff
@@ -23,25 +23,33 @@
...
+
+def get_json_data(target_datetime, params, session=None):
+    """Fetches json data for requested params and target_datetime using a post request."""
 
     epoch_time = str(int(time.time()))
-    ne = arrow.now('America/New_York')
-    today = ne.format('MM/DD/YYYY')
-    yesterday = ne.shift(days=-1).format('MM/DD/YYYY')
+    
+    # when target_datetime is None, arrow.get(None) will return current time
```
maxbellec: cool, didn't know about that :) helpful in that context

maxbellec:
any falsey value is ignored by the backend, so returning `[]` or `None` is the same. I should add that to `example.py`

corradio:
@systemcatch can you merge if you approve?

systemcatch:
@jarek can you resolve the conflict?

parsers/US_NEISO.py
jarek: OK, I will do NotImplementedError in subsequent parsers/edits :+1:

Issue #959, opened 2017-12-30
CAISO imports


corradio:
Live data is available here:
http://www.caiso.com/TodaysOutlook/Pages/supply.aspx

However those are only net imports. We should determine who they import power from, and try to figure out if a live data source exists. Seems like OASIS has some of it?

HansHyde:
Information from California Emissions Commission shows Intermountain Power Project (coal) located 2 states away in Utah is owned by LADWP and reports end of the year data to California in MWHr produced and Primary Energy consumed.  I calculated g/kWh in #960 as it was a great dataset (1,460 power generation units over a year) to use.

I think the 3 of us (@brunolajoie) should try to talk this week.  The underlying CO2/kWh methodology is flawed, but I honestly think we have the data available & resources to fix it.

jarek:
As far as I can tell from a quick scroll through https://github.com/WattTime/pyiso/blob/master/pyiso/caiso.py they are also only getting absolute import/export, not who it's coming from.

Nevertheless it might be good to just add a generic import from "US" even if we don't get its properties. It seems to be about a third of the in-state reported supply and our supply chart looks particularly odd around "duck curve" time.

jarek:
Hmm checking Oasis I do see something that might be exchange data under Energy > EIM Transfer by Tie. Would have to write a script to see if the values there line up with summed up data on CAISO's Today's Outlook page.

systemcatch:
I see that one as well.  It's possible to eventually get a CSV containing 1000's of datapoints for the current day though I haven't checked to see if they match up with absolute totals.

However I'm not sure we gain much at the moment by having this level of detail.  All the regions/states surrounding CAISO have no data currently and I don't think we have any leads for sources yet.  Everything will be set to unknown anyway.  Let's start by using the absolute import/export number and improve from there.

To make that work we just need a fetch_exchange method in CAISO, also it would probably be a good idea to merge all states without data into a generic 'USA' to keep things simple.

corradio:
That could work in order to get the ball rolling.

alixunderplatz:
To get historical exchanges from the entire state of California to grid regions labeled as "Northwest", "Southwest" and Mexico, it is worth taking another look at the eia.gov "U.S. ELECTRIC SYSTEM OPERATING DATA" website.

Seperate data to these 3 zones seems to be about 1.5 days behind. 

[...]

PR #1234, opened 2018-03-18
imports to CAISO

jarek:
This is a quick-and-dirty implementation of #959. It's not split out by state, and absent carbon intensities of the import source it won't give any more accurate information, but should give a slightly more _informative_ view. California continuously imported between 5 and 10 GW yesterday and the data might be particularly useful to illustrate how imports compensate for the considerable solar production.

cc @alixunderplatz, you did some research on the grid neighbours, that will still be useful, but note this placeholder until we can crack all the other imports.

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -1322,6 +1322,16 @@
     },
     "rotation": 90
   },
+  "US->US-CA": {
+    "lonlat": [
+      -115.846,
+      35.962
+    ],
+    "parsers": {
+      "exchange": "US_CA.fetch_exchange"
+    },
+    "rotation": -135
+  },
   "US-MISO->US-PJM": {
     "lonlat": [
       -84.803,
--- a/parsers/US_CA.py
+++ b/parsers/US_CA.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python3
+
+from collections import defaultdict
+
 # The arrow library is used to handle datetimes
 import arrow
 # The pandas library is used to parse content through HTTP
 import pandas
+
+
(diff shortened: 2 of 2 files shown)


systemcatch:
:+1: I will take a look at this later on today. I've been playing around with the EIA API and have had some good results writing a simple wrapper for parts of it, could be an easy way to get the Mexico imports.

corradio:
Thanks for the work @jarek. Nice rewrite of the existing parser also!
I'm a bit wary of adding exchange to the US as this might pollute our historical database.
Would be nice to directly start looking at exchanges by state even if they are delayed.
@systemcatch let us know how the EIA API is coming, and based on that we can reach a decision.

systemcatch:
So this got me looking at the EIA's API again and unfortunately I've realized that all the exchanges lag behind by about a day (see https://www.eia.gov/opendata/qb.php?category=2122550) for a full list of CAISO exchanges. That method won't work here but I might still write a package for the api.

systemcatch:
3 of those exchanges can be added with no alterations to the map's representation. The rest will not work at the moment. I tend to think lumping imports/exports together is the way to go for now.

Could we create something new like a proxy zone which allows exchanges but won't show on the map or mess up the data? Kind of like a private method in a Class which is never used directly.

corradio:
In that case I think the best is to keep the using the "US" zone as that
proxy zone.

systemcatch:
Or maybe '_US' to provide an easy way of distinguishing it?

corradio:
It will show up on the UI so I think "US" is best

jarek:
If we're fine using "US" as import source, and won't have better data for a while, then as I understand it this can be merged?

PR #1237, opened 2018-03-19
PEP 8 and more comments on `example.py`

maxbellec:


maxbellec:
It seems I've broken the Github by accidentally pushing to master than reverting. This is the actual diff : https://github.com/tmrowco/electricitymap/compare/aaff26996fd2113d8c2952224873f504a2a18c67...68643c3

corradio:
Can you merge latest master onto this branch?

parsers/example.py
```diff
@@ -175,16 +219,17 @@ def fetch_exchange(zone_key1='DK', zone_key2='NO', session=None, target_datetime
         'source': 'someservice.com',
     }
 
-    # Country codes are sorted in order to enable easier indexing in the database
+    # Country codes are sorted in order to enable easier indexing in the DB
```
corradio: very minor comment: I think acronyms should be avoided here.

parsers/example.py
```diff
@@ -73,35 +91,48 @@ def fetch_production(zone_key='FR', session=None, target_datetime: datetime.date
...
 
     # Parse the datetime and return a python datetime object
     data['datetime'] = arrow.get(obj['datetime']).datetime
 
     return data
 
 
-def fetch_price(zone_key='FR', session=None, target_datetime=None, logger=None):
+def fetch_price(zone_key='FR', session=None, target_datetime=None,
+                logger=None):
```
corradio: Maybe in example.py we could set a default logger here?

corradio:
Few (very) minor comments. Feel free to merge/remove branch when you feel it!

Issue #319, opened 2017-01-26
Add cross-border exchange capacity


corradio:
Example: GB-NL is 1000MW. http://www.britned.com/

alixunderplatz:
Hey @corradio 
I can provide some help at least for data sources of the cross-border exchange capacities. As a student I've been working for a German grid operator, that's why I am a bit familiar with the topic and the difficulties.
I found some numbers for central europe. Actually cross border capacities are a really complex issue.
The values below show net transfer capacities (NTC) between the countries, which do not always represent the actual "technically possible" exchange capacities (TTC = total transfer capacity). The NTC values are the possible "failsafe" or "n-1" exchange values and mainly depend on seasonal load in the countries, weekday and daytime, available and unavailable grid infrastructure on both sides of the border, production capacities and cross-border power trading. Data on entso-e is not really consistent and reliable (it's rather an INtransparency plattform than a transparency plattform), therefore in my opinion the sites of the TSOs must be consulted.

**For France and its neighboring countries** i found this nice chart:
http://clients.rte-france.com/lang/an/visiteurs/vie/ntc_annuelles.jsp
![image](https://cloud.githubusercontent.com/assets/25743609/23019006/a8dcd56e-f440-11e6-9031-ab2bb48665ab.png)
Here you can already see that values differ from month to month because of said reasons. There are even weekly and intradaily changes of NTCs, depending on load and other stuff mentioned above. So maybe these weekly values are a better indicator (example here is France-Germany):
http://clients.rte-france.com/lang/an/visiteurs/vie/interconnexions/all/previ_capa_hebdo_ntc.jsp?codePays=ALL
It is important to consider that import and export NTCs deviate, too, although the technical transfer capacity of the power lines are the same for both directions. These restrictions in one certain direction also happen for grid safety reasons.

**German NTCs** must actually be gathered for each of the four TSO areas and then be combined.

Here are some intraday NTC values for **Tennet Germany** to [Czech Rep.](https://www.tennettso.de/site/Transparenz/veroeffentlichungen/engpassmanagement/vorschau-kapazitaetsvergabe/tagesvorschau?country_token=cz&lang=de_DE), and the [Netherlands](https://www.tennettso.de/site/Transparenz/veroeffentlichungen/engpassmanagement/vorschau-kapazitaetsvergabe/tagesvorschau?country_token=nl&lang=de_DE).
[...]

PR #1239, opened 2018-03-19
Add some European cross-border transmission capacities

alixunderplatz:
Ref #319

* added capacity values [min, max] for cross-border HVDC interconnectors in Europe that exclusively link two zones via DC
   * **FYI**: HVDC between Finland & Russia is working with 1400 MW only in the direction RU->FI, but not in the opposite direction. Additional AC capacity is 160 MW, only used for delivering power of two smaller Russian hydroPPs to Finland on 110 kV level (see ENTSO-E map and #319 for more details). 
(I also moved the arrow for FI->RU exchange a bit closer to the Baltic states.)

- adjusted DK-BHM->SE capacity (AC)
- added IT->MT capacity (AC)

- fixed rotation of BY->UA arrow to point in direction of flow - Ref #1197


corradio:
Hi @alixunderplatz !
This looks cool. We should probably document our sources somewhere. Maybe you can make a new section in the README?

alixunderplatz:
Sure! I will work on the list of sources as new section "Transmission capacity data sources" in the README!

Issue #1177, opened 2018-03-04
Add MISO->SPP exchange


tmslaine:
US Southwest Power Pool has live generation mix and load data as a csv file under https://marketplace.spp.org/pages/generation-mix
Timestamps seem to be UTC by default.
I do not know if they have flow data somewhere.

Like ERCOT, SPP is also geoblocked.

corradio:
Fyi I'm in the process of being cleared for access.

systemcatch:
Yup I started writing a parser for this one after I managed to get access. They allow access by ip but once you've got access to the csv you can [pickle](https://docs.python.org/3.6/library/pickle.html) the data for easy use anywhere.

systemcatch:
This is what I'm using as my data source when I'm using another ip address. You can load the pickle into pandas and see the full data.

[SPPgenmix.pkl.zip](https://github.com/tmrowco/electricitymap/files/1808879/SPPgenmix.pkl.zip)

corradio:
I don't get blocked for this content:
![image](https://user-images.githubusercontent.com/1655848/37388637-9953a88e-2761-11e8-8b1f-29c669b1fa2c.png)
@ ftp://pubftp.spp.org/Operational_Data/GEN_MIX/

That seems to be sufficient to build a parser/get historical data right?

systemcatch:
More than enough! Can you get https://marketplace.spp.org/pages/interchange-trend? That shows all of the exchanges with other zones.

corradio:
That one is blocked.. I'll reply to them.

systemcatch:
There are 2 keys that make up a very small part of generation, 'Waste Disposal Services' and 'Waste Heat'.  How do you want them mapped?  I've not been able to find any info on what type of waste they burn.

corradio:
I tend to put them those biomass when I don't know exactly what kind of
waste it is.

systemcatch:
@corradio One other interesting thing. The data source gives me timestamps in GMT (greenwich mean time) rather than US Central Time for some reason, 2 questions arise;

- Does the csv file return the same timestamp for you or is it taking into account the recipient's locale?
- How do you want to handle this, set to UTC, convert to US time, something else?

corradio:
The CSV of the FTP should not change depending on who calls it.
On our side, the parsers can return any timezone they want, as long as that information is in the datetime object.

systemcatch:
I assume not but it was surprising to see a GMT timestamp returned.

[...]

PR #1240, opened 2018-03-19
Add US (Southwest Power Pool)

systemcatch:
The production parser is working nicely. However the exchange data provided by SPP sometimes differs from the MISO scheduled data by a significant amount. I've therefore left the US-MISO->US-SPP exchange disabled until this discrepancy can be resolved (function is all built). Practically we might be best waiting until the MISO api provides actual exchange data. :crossed_fingers: 

/cc @tmslaine 

ref #1177 

spp


parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+    Takes a dataframe and logging instance as input.
+    Checks for new generation types and logs awarning if any are found.
+    Parses the dataframe row by row removing unneeded keys.
+    Returns a list of 2 element tuples, each containing a datetime object
+    and production dictionary.
+    """
+
+    # Remove leading whitespace in column headers.
+    df.columns = df.columns.str.strip()
+    print(df)
```
corradio: Maybe remove this?

systemcatch: :astonished: whoops that was left over from debugging.

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+import pandas as pd
+import requests
+
+generation_url = 'https://marketplace.spp.org/public-data-api/gen-mix/asFile'
+
+exchange_url = 'https://marketplace.spp.org/public-data-api/interchange-trend/asFile'
+
+mapping = {'coal': 'coal',
+           'unknown': 'unknown',
+           'Wind': 'wind',
```
jarek: Just checking - I can't load the source files, but do they really have "coal" lowercase and most everything else uppercase?

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+           'Nuclear': 'nuclear',
+           'Hydro': 'hydro',
+           'Solar': 'solar',
+           'Natural Gas': 'gas',
+           'Diesel Fuel Oil': 'oil',
+           'Waste Disposal Services': 'biomass'
+            }
+
+tie_mapping = {'US-MISO->US-SPP': ['AMRN', 'DPC', 'GRE', 'MDU', 'MEC', 'NSP', 'OTP']}
```
jarek: I think pylint's preferred style is to have the constants outside the function ALL_CAPS. This actually caught me out when I was reading data_processor() and was wondering where `mapping` came from.

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # Remove leading whitespace in column headers.
+    df.columns = df.columns.str.strip()
+
+    keys_to_remove = ['Coal Market', 'Coal Self', 'GMT MKT Interval', 'Average Actual Load',
+                      'Other', 'Waste Heat']
+
+    # Check for new generation columns.
+    known_keys = list(mapping.keys()) + keys_to_remove
+    column_headers = list(df)
```
jarek: Would `list(df.columns)` also work here? I think that's a bit clearer in terms of what it is

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+
+        extra_unknowns = sum([production[k] for k in unknown_keys])
+        production['unknown'] = production['Other'] + production['Waste Heat'] + extra_unknowns
+
+        dt_aware = parser.parse(production['GMT MKT Interval'])
+
+        for k in keys_to_remove:
+            production.pop(k, None)
+
+        mapped_production = {mapping.get(k,k):v for k,v in production.items()}
```
jarek: I think at this point we've assured that `production` only contains keys that are in `mapping`, so the default value in `mapping.get(k, k)` is not needed?

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+    if target_datetime:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+    raw_data = get_data(exchange_url, session=session)
+    sorted_codes = '->'.join(sorted([zone_key1, zone_key2]))
+
+    try:
+        exchange_ties = tie_mapping[sorted_codes]
+    except KeyError as e:
+        raise NotImplementedError('This exchange pair is not implemented')
```
jarek: Maybe include `sorted_codes` in the error message for possible later debugging

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
+#!usr/bin/env python3
+
+"""Parser for the Southwest Power Pool area of the United States."""
+
+from dateutil import parser, tz
```
jarek: It looks like `tz` is not actually used?

parsers/US_SPP.py
```diff
@@ -0,0 +1,198 @@
...
+    column_headers = list(df)
+
+    unknown_keys = []
+    for heading in column_headers:
+        if heading not in known_keys:
+            logger.warning('New column \'{}\' present in US-SPP data source.'.format(
+                heading), extra={'key': 'US-SPP'})
+            unknown_keys.append(heading)
+
+    keys_to_remove += unknown_keys
```
jarek: Completely optional, the above switched to using sets:

```
keys_to_remove = {'Coal Market', 'Coal Self', 'GMT MKT Interval', 'Average Actual Load',
    'Other', 'Waste Heat'}  # set literal
known_keys = mapping.keys() | keys_to_remove  # this is a set union
column_headers = set(df.columns)  # check exact syntax

unknown_keys = column_headers - known_keys  # set difference

for heading in unknown_keys:
    logger.warning('New column \'{}\' present in US-SPP data source.'.format(heading),
        extra={'key': 'US-SPP'})

keys_to_remove = keys_to_remove | unknown_keys  # can also do "|=" but it's a bit unclear I feel
```

Untested! But later on in the code, sets can be iterated over same as lists.

parsers/US_SPP.py
systemcatch: Nope coal and the unknown key are leftovers that can be removed.

parsers/US_SPP.py
systemcatch: It works and is clearer, swapping to that.

parsers/US_SPP.py
systemcatch: Would you say there is a benefit to using sets here rather than lists?

parsers/US_SPP.py
systemcatch: With the 'coal' and 'unknown' keys gone from MAPPING, get stops them raising KeyErrors in the dict comprehension. So I think it needs to stay.

parsers/US_SPP.py
maxbellec: it's certainly not necessary, but makes computing `unknown_keys` easier and clearer

systemcatch:
All good to go here guys.

alixunderplatz:
I have some doubts that only US-NE, US-KS and US-OK are within the borders. Seems like North and South Dakota could be withdrawn from MISO and added to SWPP as well.

image

image

image

image

alixunderplatz:
I hope one day all the US-RTOs will exchange their assets with each other to clean this mess up and create coherent areas.

This is from SPPs front page:

image

systemcatch:
It's very confusing. I've seen maps showing the Dakotas as part of MISO. Maybe we compromise and allocate South Dakota to SPP for now?

alixunderplatz:
@systemcatch I'd say MISO has rather a small influence on North Dakota. I also think the first two maps in my comment are the most recent ones and they don't show any MISO in ND anymore. So SPP covers most of the area of ND now.

Also, looking at population density, most of the consumption in MISO will come from areas in its "core" and not from ND.
image

So I'd also push North Dakota over to SPP.

systemcatch:
Good points I'll move them both over.

systemcatch:
All done I'll merge if tests pass.

dakota

Hyric89:
Capacity fleet by source as done in #1177 

Wind: 16.13 GW
Hydro: 3.42 GW
Nuclear: 2.11 GW
Gas: 34.26 GW
Coal: 26.06 GW
Oil: 1.69 GW
Other: 0.29 GW

PR #1244, opened 2018-03-20
add missing entsoe parsers

maxbellec:


config/zones.json
```diff
@@ -19,6 +19,10 @@
       "https://github.com/corradio"
     ],
     "parsers": {
+      "consumption": "ENTSOE.fetch_consumption",
```
corradio: AR is not an ENTSOE zone

maxbellec:
@corradio I've also added some consumption and price parsers. Should I remove or leave them?

config/zones.json
```diff
@@ -156,6 +160,8 @@
     "parsers": {
       "consumption": "AX.fetch_consumption",
       "price": "ENTSOE.fetch_price",
+      "consumptionForecast": "ENTSOE.fetch_consumption_forecast",
```
corradio: AX is actually in the Swedish zone (so we use the ENTSOE price parser), but is a small island.
Therefore, the consumption/generation forecasts will be inaccurate as they will use the SE ENTSOE domain.
I think we should just remove from here.

Issue #563, opened 2017-05-26
How to provide exchange CO2 intensities when unknown


jarek:
The Canadian province of Prince Edward Island generates only wind electricity (outside of emergencies), and imports electricity from New Brunswick otherwise.

Unfortunately we don't currently have real-time production breakdown for New Brunswick. Based on historical data from Statistics Canada, it is on the order of 300 gco2eq/kwh. (~32% nuclear, ~20% hydro, ~4% wind, remainder various fossils - I will provide more details on how I got the number.)

In absence of information on generation in New Brunswick, https://api.electricitymap.org/v1/state reports CA-PE's exchangeCo2Intensities from CA-NB as 12 gco2eq/kwh, so it assumes that the import is the same as the local generation. In this edge case this is pretty off, and most of the time draws PEI unreasonably green.

As far as I can tell the code behind api.electricitymap.org is not open source, so I can't find how exactly it determines the exchangeCo2Intensities.

Can you tell us if there is a way to specify CO2 intensity of an export in a way that the API will understand, short of hardcoding a value for New Brunswick current generation?

(I also have calculated the intensity for Quebec at around 28 gco2eq/kwh - they are vast majority hydro - that could help Ontario's imports be more accurate, especially if/when Ontario imports more in the summer. Currently that's assumed to be the same as Ontario generation, which will go up when air conditioning season starts and their gas peakers turn on.)

corradio:
Hi @jarek 
I can make the necessary changes, probably here:
https://github.com/tmrowco/electricitymap/blob/master/config/co2eq_parameters.js

Do you have any public trustable source/report that we can use to document the use of those specific intensities?

jarek:
I'm working on writeup/jupyter notebook of how to estimate the Canadian intensities - will update when done :)

jarek:
Here is the notebook: https://gist.github.com/jarek/bb06a7e1c5d9005b29c63562ac812ad7 and the write-up with links to the Statistics Canada data tables: http://piorkowski.ca/rev/2017/06/canadian-electricity-co2-intensities/

Around 300 to maybe 400 g for New Brunswick exports (depending on how conservative we want to be) should do it :). I'll try to agitate for proper real-time data in Canada...

corradio:
Will work on this as soon as I get the chance. Good job for pushing this forward @jarek !

jarek:
Hi @corradio do you think this will be possible soon?

corradio:
[...]

PR #1248, opened 2018-03-20
try providing default export data for Québec, ref #563

jarek:
re: #563 providing assumed exchange intensity when unknown

Providing Québec intensity would show up in US-NE and US-NY, and one can also observe how CA-NB will react (CA-NB has all of its production as unknown and significant imports from Québec). Also I already had the numbers crunched.

I wasn't really sure how or where the data should be specified, nor am I fluent in JS style - how is this? Suggestions/corrections welcome.


config/co2eq_parameters.js
```diff
@@ -95,6 +95,16 @@ var countryCo2eqFootprint = {
     }
 };
 
+var exportCo2eqFootprint = {
+    'CA-QC': {
+        value: 30,
```
corradio: maybe call it carbonIntensity

PR #1254, opened 2018-03-21
Add some capacities to exchanges.json and zones.json

alixunderplatz:
- added NZ North <-> NZ South interconnector capacity to exchanges.json
- added AUS-TAS <-> AUS-VIC interconnector capacity to exchanges.json
- added SG solar capacity (AC) to zones.json and updated "with help from" for SG
- added all sources to README and refined transmission capacity part


corradio:
The system will use the static one by default, which will be overwritten by the parser (at least that's the intended behaviour)

corradio:
The system will use the static one by default, which will be overwritten by the parser (at least that's the intended behaviour)

PR #1255, opened 2018-03-21
add a ValidationException

maxbellec:
useful to catch them in the backend


corradio:
Feel free to merge when tests pass.

PR #1258, opened 2018-03-21
hardcode export intensities for AZ, CA-NB, ZA, ZM

jarek:
re https://github.com/tmrowco/electricitymap/issues/563, cc @alixunderplatz @systemcatch 

This provides intensities for zones which are export-heavy and make up a significant amount of supply in at least one other zone for which we have real-time information.

Intensities are calculated by Electricity Map's fuel intensities and ratio of the fuels in the yearly generation. The values are rounded to nearest 10 since it's a rough estimate anyway.

CA-NB is a little conservative since it doesn't take NB's imports from QC into account, but I'd suggest better to overestimate than underestimate...

RIP PEI's days as nonstop climate leader on Electricity Map

### DIFF
--- a/config/co2eq_parameters.js
+++ b/config/co2eq_parameters.js
@@ -96,12 +96,41 @@ var countryCo2eqFootprint = {
 };
 
 var defaultExportCo2eqFootprint = {
+    'AZ': {
+        carbonIntensity: 470,
+        renewableRatio: 0.07,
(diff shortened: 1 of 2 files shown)


corradio:
@alixunderplatz feel free to review this.

alixunderplatz:
Looks good to me @corradio @jarek !

corradio:
ok I'll merge :)

Issue #1062, opened 2018-02-03
Add Albania AL (retrospective/scheduled data)


alixunderplatz:
Hey everyone! Very good news for another country! 😃 

Albanian electricity generation is based on ~100% hydro power. This allows us to take total generation data from the national TSO "OST" for the map, as long as it is still not reported to ENTSO-E.

There are 3 possibles sources for the data:

1.) Each day (with a delay of ~2 days) a PDF-file with a data chart is published, which contains hourly generation and load data. This will be something **for the database** (with a better precision than 2.) and 3.)).

https://www.ost.al/index.php/en/consume-generation-year-2018/
![image](https://user-images.githubusercontent.com/25743609/35770495-5fec98ca-091c-11e8-973c-a2e5cc022d94.png)

2.) **For the live-map**, we can actually take the day-ahead scheduled generation. It's not yet available for tomorrow (20:00 CET), but I'll keep watching it.
https://www.ost.al/Data%20Flow.html
https://www.ost.al/index.php/sq/scheduled-generation-day-ahead/

There is a data chart with hourly scheduled generation.
![image](https://user-images.githubusercontent.com/25743609/35770529-3b74861e-091d-11e8-8852-d291d7174198.png)

3.) Surprisingly, I found data on ENTSO-E for day ahead scheduled generation (already available for tomorrow as opposed to 2.) ) which is also a possible thing **for the live-map** and probably the best solution.
https://transparency.entsoe.eu/generation/r2/dayAheadAggregatedGeneration/show

I made a quick chart in Excel to compare scheduled generation of 2.) and 3.) vs. actual generation of 1.): the result is pretty good! 

![image](https://user-images.githubusercontent.com/25743609/35770655-82c8c000-091f-11e8-8d68-e2eb02f05bdf.png)

There are hydro plants only, so there won't be many unforeseen outages like for more complex fossil plants. So I'd say using scheduled data is absolutely not a problem here!
The installed hydro capacity is something between 1700 and 1800 MW, I'll try and look for more capacity data. On ENTSO-E it says 1492 MW, which is a bit too low and not containing smaller hydro.

brunolajoie:
Hey alex! great detective job again!! 🔍 

[...]

PR #1264, opened 2018-03-23
Albania hardcoded export emission factor

alixunderplatz:
Default source for generation in 2015:
https://www.iea.org/statistics/statisticssearch/report/?country=ALBANIA&product=electricityandheat&year=2015

New hydro power is supposed to come online this year. Currently up to 1800 MW generation based on the links I gave in #1062

Bit longer article on the current and future capacity development, if you're interested:
https://www.energyworldmag.com/electricity-generation-capacity-development-albania/
It says there is a CCGT gas power plant, but according to other [news](https://seenews.com/news/albania-plans-to-feed-gas-from-tap-pipeline-into-idled-vlora-tpp-532173), it was never put into operation since 2011 due to mechanical faults. 

Also there is some generation from waste but with max of 2.85 MW, so insignificant atm. 
PV capacity will slowly rise, tenders are expected. If any new capacity is added, it will still be represented by the "100% renewable" ratio.
Potential is between 1500 to 1800 kWh/kWp, so 100 MWp PV will translate into 150 to 180 GWh/a in the future.


jarek:
Does Albania often/ever import and export at the same time?

alixunderplatz:
@jarek very good point!
I guess your concern is also going in the direction that the intensity is wrong/worse if there is at least one import from the other zones.

Looks like Albania was heavily importing from Greece and Montenegro **in 2017**.
Here is the percentage of hours for the directions of flow to each neighbour (source ENTSO-E physical cross border flow):

image

Looking at the simultaneous occurrence of either imports or exports:
66 hours with export to all directions at the same time, 4175 hours of import from all directions at the same time.

This year, **so far** the situation looks totally different.

image

There already were 1196 hours (of 1968 analyzed) of simultaneous export in all directions, because probably there were some new hydro power plants that were put into operation and there is enough available water, too. No simultaneous imports from all sides. Exports to Serbia 100% of time so far.

jarek:
Hmm. Yeah we'd try to avoid situations where Albania transfers large amounts of electricity without generating much, but magically making it cleaner through the exchange hardcode.

Can you easily have an overview of the ratio of local production vs imports in 2018 in situations when they were both importing and exporting?

alixunderplatz:
@jarek 
Visualization of the imports and exports as time series for 2018:
Only some imports at all at the beginning of the year from ME, but still less than what was exported in the balance (red line, positive all the time):

image

What is a bit special is this: it seems like there is a constant amount of generation leaving to RS/Kosovo (yellow line), starting in February. First it is around 100 MW, then around 200 MW. It is possible that the grid is used to export the entire generation of one power plant or some units via a seperate power line.
Looking at the topology there is one 220 kV and one 380 kV line to RS/Kosovo, and the Fierze substation in AL has a HPP with 4x 125 MW connected to it.

From February, around half of the generation of AL (estimate by me, haven't yet calculated it) is exported and there barely were any imports.
So the intensity factor is valid at the moment.

jarek:
The chart is convincing for me and I have no more questions :+1: - let's wait for @brunolajoie

alixunderplatz:
@jarek @brunolajoie this situation may change as soon as the hydro generation decreases. (Generation diversity in the portfolio is a must for Albania.) 
I will track it and in case transit flows will be dominant, disabling the intensity for AL is the only option.

brunolajoie:
We couldn't validate hardcoding this export intnesity factor without a continuous check of AL import/export balance like the one you suggested above @alixunderplatz. Trusting your commitment on this, I'll approve the PR. Feel free to merge if you want to follow up on that 💪

PR #1265, opened 2018-03-23
Make NA->ZM exchange more reliable

systemcatch:
Trying to make the NA->ZM parse more reliably. See commit message for more details. @jarek do you think this regex will work? New group is supposed to match anything except digits and minus signs while avoiding grabbing too much.

### DIFF
--- a/parsers/NA.py
+++ b/parsers/NA.py
@@ -121,19 +121,21 @@ def fetch_production(zone_key = 'NA', session=None, target_datetime=None, logger
     return data
 
 
-def exchange_processor(text, exchange):
+def exchange_processor(text, exchange, logger):
     """
     Takes text produced from OCR and extracts exchange flow.
     Returns a float or None.
     """
 
     utility = exchange_mapping[exchange]
-
+    # ([\D]*?)([-+]?\d+\.\d\d)
     try:
-        pattern = re.escape(utility) + r"([\D^-]*?)-?(\d+\.\d\d)"
+        pattern = re.escape(utility) + r"([\D]*?)([-+]?\d+\.\d\d)"
         val = re.search(pattern, text).group(2)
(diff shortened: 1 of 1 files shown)


jarek:
Background: the specific value mentioned in Slack was that OCR was returning `ZESCOZ` instead of `ZESCO:`.

It does deal with the random Z and other characters better :+1: 

When I try the regex in regex101.com using Python setting on data "ZESCO: -11.22", it returns only "11.22" without the negative. It seems that the old regex would do this as well. This means we are ignoring when Namibia says it's exporting. To be fair I don't know if that ever happens currently.

regex101 also says `^-` "matches a single character in the list ^- (case sensitive)", which isn't what we want, I think. In effect, currently the `^-`  does nothing. I also don't think it's really necessary, anyway?

It might also be nice to check the result of `re.search(pattern, text)` and if it's None, logging a warning and including `text` in the log output. This would make it easier to debug future failures, which currently are logged only as "AttributeError: 'NoneType' object has no attribute 'group'" with no information as to what the malformed data was.

systemcatch:
Good idea on the logging I will implement it.

I've modified the regex to ([\D]*?)([-+]?\d+\.\d\d) which seems to be capturing negative numbers properly. Can you double check it works ok?

jarek:
One more note perhaps: `([\D]*?)([-+]?\d+[.,]\d\d)`. Currently `.\d\d` is literally "any character then two digits" because the `.` is a special char. If you do `[.,]\d\d` it'll match "either a comma or a period then two digits", in case the OCR misdetects the decimal separator. Otherwise looks like it will work :+1:

systemcatch:
I can do `\.` to escape the dot. If the decimal separator is misread the float operation will fail anyway.

Issue #1257, opened 2018-03-21
Add possibility to scrape historical data


maxbellec:
# Goal

The goal of this issue is to centralize what parsers already have the ability to scrap historical data, for which zones we can implement it and for which zones it is not currently possible.

We need historical as much historical data as possible so that we can build predictive models for carbon intensity or electricity mix for as many zones as possible.

## Already implemented
- [x] AR
- [x] CL-SIC
- [x] CL-SING
- [x] FR
- [x] ENTSOE
- [x] GB-NIR
- [x] Statnett
- [x] US-CA
- [x] US-NEISO
- [x] US-NY
- [x] US-PJM - see http://dataminer2.pjm.com/list. Not implemented on the parser but it's easy to get by hand

## Can be implemented
- [ ] CA-ON #987
- [ ] UY #1241
- [ ] DO - see comment

## Cannot (yet) be implemented
Note: I wrote almost every parser here without checking. It might be easy for a lot of them to fetch historical data.
- [ ] AU
- [ ] AW
- [ ] AX
- [ ] BO
- [ ] BR
- [ ] CA
- [ ] CR
- [ ] CY
- [ ] DK-BHM
- [ ] ENTE
- [ ] ES-CN
- [ ] ES-IB 
- [ ] ESIOS
- [ ] FO
- [ ] GB
- [ ] GT
- [ ] HOPS
- [ ] IN-AP
- [ ] IN-CT
- [ ] IN-DL
- [ ] IN-GJ
- [ ] IN-KA
- [ ] IN-PB
- [ ] MD
- [ ] MY-WM
- [ ] NA
- [ ] NI
- [ ] NZ
- [ ] PA
- [ ] PE
- [ ] RU
- [ ] SG
- [ ] SV
- [ ] TR
- [ ] TW
- [ ] UA
- [ ] US-MISO
- [x] US-SPP

maxbellec:
@jarek @systemcatch @alixunderplatz (and many others...) your help would be very much welcomed here, you probably have in mind for a few parsers if historical data can be fetched.

systemcatch:
Easier ones first.

**CL-SIC** 
It currently scrapes data from xls files on the site which use the date as the file name e.g. 180215.xls. All that needs to happen is to pass the desired url and everything should work ok.

**US-CA**
The old version of the parser could retrieve .txt files for any past date, e.g. http://content.caiso.com/green/renewrpt/20170601_DailyRenewablesWatch.txt

[...]

PR #1267, opened 2018-03-24
update CR parser, allow fetching past data

### LINKED ISSUE #1266: Costa Rica
Costa Rica is showing as N/A on your map.  They are 99% renewable.  https://www.independent.co.uk/news/world/americas/costa-rica-electricity-renewable-energy-300-days-2017-record-wind-hydro-solar-water-a8069111.html

jarek:
- HTML of the source had changed a bit, fix #1266
- past production data back to 2007 can be fetched, re #1257 
- update logging messages a bit

I hope that's enough to justify my shameless self-promotion in zones.json

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -560,6 +560,7 @@
   "CR": {
     "contributors": [
       "https://github.com/corradio",
+      "https://github.com/jarek",
       "https://github.com/mlucchini"
     ],
     "parsers": {
--- a/parsers/CR.py
+++ b/parsers/CR.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 # coding=utf-8
 
+import logging
+
 import arrow
 import pandas as pd
 import requests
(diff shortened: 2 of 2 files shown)


systemcatch:
Nothing wrong with a bit of self promotion :wink: ! Do you want to build this branch and update the mockserver state to check everything is ok?

jarek:
Ah, running front-end doesn't really work for me, my one machine has a super-broken Docker install and the other VM runs out of disk space... so I usually just test on commandline only

systemcatch:
Is your machine running Windows?

I got your branch to build and run ok, so we should be good to go here.

jarek_cr

jarek:
My one machine has a only-somewhat-updated Ubuntu 15.04 (that's the broken Docker), the other is a Windows I don't use much with too-small Linux VM... Sorting it all out is on the todo list :P 

I think I'll just merge it and see how it goes, it can't be any worse than not working at all.

systemcatch:
Heh I finally made the jump from Ubuntu 14.04 to 16.04 after delaying for ages so I know all about the todo list!

Issue #809, opened 2017-10-31
Malaysia 🇲🇾


jarek:
https://www.gso.org.my/LandingPage.aspx has data for Peninsular Malaysia (~80% of population) including total generation every 10 minutes, but fuel mix appears to only be on a per-day basis right now. There is already some fuel mix data for today (it is 18:00 in Malaysia), perhaps we can ask if they can publish at least hourly fuel mix?

As I understand it, https://www.gso.org.my/SystemData/TieLine.aspx  are interchanges - EGAT is the Thai operator so probably to Thailand (http://www2.egat.co.th/hvdc/INTRODUCTION.HTML), PLTG seems to be with Indonesia, and HVDC with Singapore (see https://github.com/tmrowco/electricitymap/blob/master/parsers/SG.py#L53). I don't know if positive means import or export.

jarek:
*Actually* it looks like https://www.gso.org.my/SystemData/FuelMix.aspx might be providing current values when no date range is specified. Currently it gives a sum of 15012 MW which *just happens* to be exactly equal to the latest current generation reported in https://www.gso.org.my/SystemData/CurrentGen.aspx ? Should keep an eye on this for a few days.

systemcatch:
Watching the Fuel mix link the sum total has matched the amount given in current generation for the past hour (6 datapoints).  How'd you find their website out of interest?

jarek:
Yeah, it also holds for me now. We can re-check later but it seems feasible.

So I guess a parser should read the FuelMix.aspx and CurrentGen.aspx pages. Sum up the values in fuel mix and verify that the sum is the same as the latest value in current gen. If yes, report the fuel mix, using the timestamp of the latest value in current gen table.

For map, we will have to include all of Malaysia's subdivisions except for MY.SK and MY.SA, which are the states not in Peninsular Malaysia.

Sounds like a good first issue for someone, too! Feel free to reach out if you need some more help.

--

[...]

PR #1273, opened 2018-03-27
Create Malaysian exchanges with Thailand & Singapore

systemcatch:
Scraping aspx pages is no fun. These functions can return 10min data for the current day. The Singapore arrow dwarfs the country!

/cc @maxbellec might be some inspirations for UY past dates here.

ref #809 

my_exchanges


jarek:
From first look code looks nice, I'll take a deeper look later.

That exchange arrow really does look a bit silly. How would it look moved a bit further north so that Singapore is more easily clickable? Though that might cause weirdness at different zoom levels.

alixunderplatz:
From what I can evaluate, everything is assigned to the proper zone.

I'd also move that exchange arrow to be completely inside the tip of Malaysia, so Singapore is still accessible. I think it will be quite helpful to enhance that tiny Singapore also has data.

systemcatch:
Ok I will move the arrow north into Malaysia and rotate counter clockwise slightly so Singapore is more visible.

jarek:
Code looks really good! You've managed to make the painful ASP.NET postbacks look logical!

It is currently 29.03. 01:52 in Malaysia, https://www.gso.org.my/SystemData/TieLine.aspx does not yet have any data for 29th (in fact it looks like exchanges have not updated since 28.03. 20:10?). If you care to catch the exception:

```
  File "MY_WM.py", line 345, in <module>
    print(fetch_exchange('MY-WM', 'TH'))
  File "MY_WM.py", line 312, in fetch_exchange
    processed_egat = xml_processor(egat_data)
  File "MY_WM.py", line 195, in xml_processor
    raw_data = ElementTree.fromstring(text)
  File "/usr/lib/python3.4/xml/etree/ElementTree.py", line 1325, in XML
    parser.feed(text)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 19, column 13
```

The offending line 19 seems to be: `<table width=100%><tr><td></td><td align=right><table><tr><td><span id="toolbar_User"></span></td><td><span id="toolbar_Station"></span></td><td><span id="toolbar_Date"></span></td></tr></table></td></tr></table></rl>`

I've no idea what the `</rl>` is supposed to be... Maybe xml lib of newer Pythons can deal with this better.

And at 01:58, I got a yet different error:

```
  File "MY_WM.py", line 346, in <module>
    print(fetch_exchange('MY-WM', 'TH'))
  File "MY_WM.py", line 305, in fetch_exchange
    egat_hidden_values = extract_hidden_values(req)
  File "MY_WM.py", line 176, in extract_hidden_values
    viewstategenerator = soup.find("input", attrs={'id': '__VIEWSTATEGENERATOR'})['value']
TypeError: 'NoneType' object is not subscriptable
```

These aren't really parser errors, of course, they're underlying data errors...

systemcatch:
Moved the arrow so that Singapore is visible again.

mysgarrow

systemcatch:
@jarek thanks I had to sacrifice (or post?) several small furry animals to the ASP.NET gods to get this to work but overall I understand these pages much better now.

It looks like https://www.gso.org.my/SystemData/TieLine.aspx has updated with data for 29th now. I think when there's no data available html is returned rather than xml which results in the failure.

The 2nd error is very weird as I thought the viewstategenerator tag always loads with the page. Only explanation is that the page failed to load properly.

systemcatch:
I'll merge this later on today unless there's any other points raised.

maxbellec:
that's awesome @systemcatch. Super clean code for a super ugly task :1st_place_medal:

PR #1279, opened 2018-03-30
add SE subzone exchanges

maxbellec:


maxbellec:
while at it, should I also add exchanges for `NO` subzones?

parsers/ENTSOE.py
```diff
@@ -69,6 +69,7 @@
     'CH': '10YCH-SWISSGRIDZ',
     'CZ': '10YCZ-CEPS-----N',
     'DE': '10Y1001A1001A83F',
+    'DE-AT-LU': '10Y1001A1001A63L',
```
corradio: Cool. Can you update `ENTSOE_PRICE_DOMAIN_OVERRIDE` now that you added this dict entry?

parsers/ENTSOE.py
```diff
@@ -127,6 +128,8 @@
 ENTSOE_EXCHANGE_DOMAIN_OVERRIDE = {
     'DE->DK-DK1': ['10Y1001A1001A63L', ENTSOE_DOMAIN_MAPPINGS['DK-DK1']],
     'DE->DK-DK2': ['10Y1001A1001A63L', ENTSOE_DOMAIN_MAPPINGS['DK-DK2']],
+    'DE->SE-SE4': [ENTSOE_DOMAIN_MAPPINGS['SE-SE4'],
```
corradio: Any need for adding `PL->SE-SE4`, `FI->SE-SE{1,3}`and `LT->SE-SE4`?
I'm not sure the exchanges between bidding zones and countries will work.

maxbellec: no need to add, it works

Issue #1275, opened 2018-03-28
Add Idaho (IPCO / US-ID)


alixunderplatz:
There is generation data for **Idaho Power Company**:
https://www.idahopower.com/energy/delivering-power/generation-and-demand/

It has a delay of about 2-3 hours. It's also available as a table for the timeframe displayed in the chart.

![image](https://user-images.githubusercontent.com/25743609/38018727-063001c2-3276-11e8-8940-145545641c3a.png)

If you select one of the bars, a new window will open with this link-structure:
https://www.idahopower.com/energy/delivering-power/generation-demand-detail/?date=2017undefined25&hr=19 

I tried to find out the format of how to replace the "undefined" with the month. Just replace it with 03 for March and there you go. This option could retrieve data from the past as well. 
See:
https://www.idahopower.com/energy/delivering-power/generation-demand-detail/?date=20180327&hr=19#tab-1-2

@systemcatch I am not sure if it is a sub-region of BPA #1040 , but BPA's transmission lines you see in the map below don't really cover much of Idaho. Do you have any insights? 
I assume it is not part of BPA, because exchanges with "BPAT" are also shown on the EIA website.

![image](https://user-images.githubusercontent.com/25743609/38019475-f63d086c-3277-11e8-9dea-8ae57ed396db.png)

Imports and exports seem to be relatively low, therer are mostly transit flows from "PACW" to "PACE" according to [the chart on EIA](https://www.eia.gov/realtime_grid/?src=home-undefined#/data/graphs?end=20180322T00&start=20180315T00&bas=0000000000g&regions=0)

![image](https://user-images.githubusercontent.com/25743609/38020274-13343b5a-327a-11e8-9ff4-9bfe05177fd1.png)

systemcatch:
Excellent find as usual Alex! I have some problems getting the chart to work in firefox but chromium handles it fine. Watching the network requests the chart is populated by data from https://api.idahopower.com/Energy/Api/V1/GenerationAndDemand/Subset

Looking at areas https://www.idahopower.com/about-us/service-area-map/ indicates they are separate from BPA.

alixunderplatz:
@systemcatch perfect! I also couldn't revisit the [hourly chart](https://www.idahopower.com/energy/delivering-power/generation-demand-detail/?date=20180327&hr=19#tab-1-2), only works when I have been on another page of their site - maybe some cookie issue?

[...]

PR #1280, opened 2018-03-30
Add Idaho Power Company

systemcatch:
I've set the 'PURPA-Other' key to biomass based on it's description [here](https://www.idahopower.com/energy/delivering-power/generation-and-demand/), let me know if that needs changing. Not yet found a good capacities source but I will keep searching. In reality the Idaho Power Company only supplies the southern half of the state but I think this is an ok approximation for now.

ref #1275 

us_ipc


parsers/US_IPC.py
```diff
@@ -0,0 +1,145 @@
...
+
+def fetch_production(zone_key = 'US-SPP', session=None, target_datetime=None, logger=getLogger(__name__)):
+    """
+    Requests the last known production mix (in MW) of a given zone
+    Arguments:
+    zone_key (optional) -- used in case a parser is able to fetch multiple zones
+    session (optional) -- request session passed in order to re-use an existing session
+    Return:
+    A dictionary in the form:
+    {
```
jarek: Maybe note that this parser actually returns a list of dictionaries.

parsers/US_IPC.py
```diff
@@ -0,0 +1,145 @@
...
+    for group in grouped:
+        dt = timestamp_converter(group[0])
+        generation = group[1]
+
+        production = {}
+        for gen_type in generation:
+            production[gen_type['name']] = float(gen_type['data'])
+
+        keys_to_remove = {'Load', 'Net Purchases', 'Inadvertent'}
+        known_keys = GENERATION_MAPPING.keys() | set() | keys_to_remove
```
jarek: The result of `known_keys` here won't change from iteration to iteration. You could move the whole thing up outside the for loop, and have it be something like:

```
keys_to_ignore = {'Load', 'Net Purchases', 'Inadvertent'}
known_keys = GENERATION_MAPPING.keys() | keys_to_ignore
```

then in the loop, `keys_to_remove = keys_to_ignore | unknown_keys`

parsers/US_IPC.py
```diff
@@ -0,0 +1,145 @@
...
+
+    return raw_data
+
+
+def timestamp_converter(timestamp):
+    """Converts a str timestamp into an aware datetime object."""
+
+    dt_naive = parser.parse(timestamp)
+    mountain_time = tz.gettz("America/Boise")
+    dt_aware = dt_naive.replace(tzinfo=dt_naive.tzinfo or mountain_time)
```
jarek: As far as I can tell, the source data never has timezone specified in its timestamps. So this check of `dt_naive.tzinfo` isn't really needed because it'll always be None. Also a new `tz.gettz` lookup is done for every timestamp, but that'll never change. We could move it to a constant up top:

```
TIMEZONE = tz.gettz("America/Boise")

....

def timestamp_converter(timestamp):
    """Converts a str timestamp into an aware datetime object."""

    dt_naive = parser.parse(timestamp)
    dt_aware = dt_naive.replace(tzinfo=TIMEZONE)

    return dt_aware
```

parsers/US_IPC.py
```diff
@@ -0,0 +1,145 @@
...
+
+        parsed_data.append((dt, production))
+
+    for key in unmapped:
+        logger.warning('Key \'{}\' in US-IPC is not mapped to type.'.format(key), extra={'key': 'US-IPC'})
+
+    return parsed_data
+
+
+def fetch_production(zone_key = 'US-SPP', session=None, target_datetime=None, logger=getLogger(__name__)):
```
jarek: zone_key probably should be `US-IPC` here, or even hardcode it to US-IPC in the function itself. Right now the testing function returns data with `'zoneKey': 'US-SPP'`. This won't be a problem in practice since the backend will call the parser with the right parameter, but it's a bit odd...

parsers/US_IPC.py
```diff
@@ -0,0 +1,145 @@
...
+                      }
+
+
+def get_data(session=None):
+    """Returns a list of dictionaries."""
+
+    s = session or requests.Session()
+
+    req = requests.get(PRODUCTION_URL)
+    json_data = req.json()
```
jarek: Hm... weird behaviour of the API. Sometimes I get parser failures:

```
  File "US_IPC.py", line 143, in <module>
    print(fetch_production())
  File "US_IPC.py", line 121, in fetch_production
    raw_data = get_data(session=session)
  File "US_IPC.py", line 34, in get_data
    json_data = req.json()  
  File "/home/jarek/projects/electricitymap/venv3/lib/python3.4/site-packages/requests/models.py", line 866, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.4/json/__init__.py", line 318, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.4/json/decoder.py", line 343, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.4/json/decoder.py", line 361, in raw_decode
    raise ValueError(errmsg("Expecting value", s, err.value)) from None
ValueError: Expecting value: line 1 column 11592 (char 11591)
```

where the JSON around that point looks like (linebreaks inserted):

```
...
{"name":"Coal","data":157.000000000000,"datetime":"2018-03-28T12:00:00"},
{"name":"Hydro","data":1318.000000000000,"datetime":"2018-03-28T12:00:00"},
{"name":"Net Purchases","data":0.000000000000,"datetime":"2018-03-28T12:00:00"},
{"name":"Inadvertent","data":***************,"datetime":"2018-03-28T13:00:00"},
{"name":"Load","data":1455.000000000000,"datetime":"2018-03-28T13:00:00"},
{"name":"Non-Utility Geothermal","data":32.000000000000,"datetime":"2018-03-28T13:00:00"},
{"name":"PURPA Other","data":39.000000000000,"datetime":"2018-03-28T13:00:00"},
...
```

That is quite clearly an API problem producing invalid JSON. I don't know how often it happens - right now it's about 30% of the time for me, but that datapoint is almost 3 days old and I don't know how often they insert such invalid things, or why it's only returned some of the time. Maybe let's merge and see in logger how often it causes problems?

parsers/US_IPC.py
systemcatch: No idea how I missed that.

parsers/US_IPC.py
systemcatch: Agreed. I saw this a few times but the weird thing was that when I ran the parser again immediately afterwards the json was fine.

parsers/US_IPC.py
jarek: No worries, that is why we have peer reviews :+1:

brunolajoie:
Thanks @systemcatch !

Do you know if the API results in MW are "commercial power delivery" or "physical power production"? Because the electricity map should only show physical power production & exchanges. I'm having the feeling that "PURPA" is actually commercial contracts via Renewable Energy Certiificates, which does not corresponds to physical power flows. Are they produced in the same region? If so, they could be included to the pysical power flows of this state indeed, but in the US you can buy REC from far away, in which case we might have already accounted for these electrons somewhere else in the electricitymap.

I haven't investigated it far enough, so just wanted to ask your pov here before deep diving into idahopower website (which charts only work with safari for me btw!)

systemcatch:
@brunolajoie , I think you might be right. Looking at their website in depth it appears they don't own any wind/solar/geothermal facilities but buy energy from the wholesale market using certificates. Some of these generators may be in Idaho but could also be in neighboring utilities. There's no breakdown by area.

I gathered some links that explain the situation. Can you take a look and give us your thoughts?
https://www.idahopower.com/energy/renewable-energy/
https://www.idahopower.com/energy/renewable-energy/renewable-energy-certificates-work/
https://www.idahopower.com/energy/delivering-power/energy-sources/
https://www.idahopower.com/energy/delivering-power/generation-and-demand/

brunolajoie:
I've send an email to idahopower to ask them whether the PURPA purchase comes from, geographically speaking

alixunderplatz:
@systemcatch On page 17 of the annual report for 2016 you can find some capacities. It looks like even some of the thermal power plants are located outside of Idaho - but close to it. Hydro is entirely within the borders of Idaho.
http://www.idacorpinc.com/~/media/Files/I/IDACorp/investor-docs/annual-reports/ar2016.pdf

Installed capcaity for Idaho Power Company in 2016 from that map in the report:
* hydro: 1709 MW,
* thermal: 1885 MW - the thermal breakdown is: 
  * coal 1118 MW, 
  * gas 762 MW,
  * oil 5 MW

(I hadn't read or searched any of the other chapters, maybe there are some figures on solar etc.)

Good catch with that PURPA-stuff @brunolajoie ! 👍

maxbellec:
code looks good, I'll wait to be sure we actually want this parser (it corresponds to electricity being physically generated) to look into more details

systemcatch:
@alixunderplatz nice I will use those capacities.

As a proposal if it's reported back that the PURPA generation is all outside of Idaho I suggest we remove those categories and collect only the hydro/thermal generation.

If that too is partly outside the boundaries of Idaho then I think we should run this parser as is for the dataset but not display it on the map.

brunolajoie:
Agree with you systemcatch! Safest way to start with is to stick to hydro/thermal.

systemcatch:
Ok @brunolajoie I will make those changes and set the capacities later on tonight.

systemcatch:
Alright we should be good to go here. If there's any other comments let me know, otherwise let's merge this.

parsers/US_IPC.py
```diff
@@ -0,0 +1,151 @@
...
+    raw_data = json_data['list']
+
+    return raw_data
+
+
+def timestamp_converter(timestamp):
+    """Converts a str timestamp into an aware datetime object."""
+
+    dt_naive = parser.parse(timestamp)
+    dt_aware = dt_naive.replace(tzinfo=TIMEZONE)
```
maxbellec: FYI, we usually use `arrow.get(timestamp, "America/Boise")` as it's a bit simpler, but it doesn't change the result

parsers/US_IPC.py
```diff
@@ -0,0 +1,151 @@
...
+
+    unmapped = set()
+    parsed_data = []
+    for group in grouped:
+        dt = timestamp_converter(group[0])
+        generation = group[1]
+
+        production = {}
+        for gen_type in generation:
+            production[gen_type['name']] = float(gen_type['data'])
```
maxbellec: just suggesting another code here:

```python
# None means we should ignore that key
GENERATION_MAPPING = {
  'Hydro': 'hydro',
  'Coal': 'coal',
  'Diesel': 'oil',
  'Natural Gas': 'gas'
  'load': None,
  'Net Purchases': None,
  'Inadvertent': None,
  'Non-Utility Geothermal': None,
  'PURPA/Non-Utility Wind': None,
  'PURPA/Non-Utility Solar': None,
  'PURPA Other': None,
}

[...]
production = {GENERATION_MAPPING[gen_type['name']]: float(gen_type['data'])
              for gen_type in generation
              if GENERATION_MAPPING.get(gen_type['name'])}

unknown_keys = {gen_type['name'] for gen_type in generation} - set(GENERATION_MAPPINGS)
unmapped = unmapped | unknown_keys
[...]
```

it eliminates the need for `current_keys`, `keys_to_remove`, `pop`s, `known_keys` and, I think, makes the code cleaner.

systemcatch: How about I wait until Bruno get a response regarding the renewables, then I can update the categories accordingly and use your suggestion.

brunolajoie: Ok for that: Let start with Idaho power company generation only, and then iterate from here, asking advice if we should mix the **geothermal, wind, solar and other** production to the state-mix. This should become clearer when we investigate neighboring regions. it's a puzzle!

PR #1281, opened 2018-03-31
SG: don't fail when unable to OCR

jarek:
Solar production is a very small part of Singapore electricity, so when it can't be retrieved, just log warning and move on.

Add the OCR'd text to log message to possibly aid debugging.

Currently the image is stuck on data from ~12 hours ago. The parser has code to discard too old data, but unfortunately the time 11:44 confuses terract, which reads it as `2018-03-31 1 1 :44`. Maybe the regexp could be improved to handle that, but in meanwhile, fall through and log better.

### DIFF
--- a/parsers/SG.py
+++ b/parsers/SG.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python3
 
 from collections import defaultdict
+import logging
+import re
 
 import arrow
 from PIL import Image
 from pytesseract import image_to_string
-import re
 import requests
 
 TIMEZONE = 'Asia/Singapore'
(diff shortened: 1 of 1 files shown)


parsers/SG.py
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env python3
 
 from collections import defaultdict
+import logging
```
systemcatch: Minor comment, since this only used once maybe it should be `from logging import getLogger`.

parsers/SG.py
```diff
@@ -61,52 +54,61 @@
...
     gray = solar_image.convert('L')
-    threshold_filter = lambda x: 0 if x<77 else 255
+    threshold_filter = lambda x: 0 if x < 77 else 255
     black_white = gray.point(threshold_filter, '1')
 
     text = image_to_string(black_white, lang='eng')
 
-    pattern = r'Est. PV Output: (.*)MWac'
-    val = re.search(pattern, text, re.MULTILINE).group(1)
+    try:
```
systemcatch: Much better handling, we shouldn't throw away data just because solar is not working.

parsers/SG.py
```diff
@@ -216,12 +219,16 @@ def fetch_production(zone_key='SG', session=None, target_datetime=None, logger=N
         'hydro': 0
     })
 
+    source = 'emcsg.com'
+    if generation_by_type['solar']:
```
systemcatch: Nice idea.

parsers/SG.py
jarek: Hm, this is a style thing ([there is no perf difference](https://softwareengineering.stackexchange.com/questions/187403/import-module-vs-from-module-import-function/187471)). My very personal reason for liking `logging.getLogger()` is because I don't have to scroll up to see if `getLogger` was imported and from which module, or maybe a function somewhere within this file. To give an extreme example, if we were to do `from re import search`, it's impossible for someone diving straight into the code to guess what `search` is without reading all imports.

I don't think we have particular consistency in the codebase on this right now. At the same time I myself did the `from collections import defaultdict` right above :D

jarek: Current codebase: `import logging` 9 times including SG (and in all cases it is only used for `getLogger`), `from logging import getLogger` 7 times. So if I change it, it'll be a tie. Do we have a style guide? :D

parsers/SG.py
jarek: I'm wondering if it's too clever. When there is no solar production, the data that solar is 0 also somes from EMA...

parsers/SG.py
systemcatch: Oh I thought there was a difference, never mind then it doesn't matter. If there was an EM style guide it would simply read "Chaos is good". :smiling_imp:

parsers/SG.py
systemcatch: Maybe both should be included by default?

parsers/SG.py
jarek: "Carbon is bad, chaos is good"

jarek:
I am on mobile next couple of days, please merge if you're happy with it :)

PR #1286, opened 2018-04-03
new parser for SE

maxbellec:


maxbellec:
I changed the default parser from statnett to SVK for `SE`

parsers/SE.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # creating a dict looking like
+    # {'2018-01-01 01:00': {'nuclear': xxx, 'hydro': ...}, ...}
+    productions = defaultdict(lambda: defaultdict(lambda: 0))
+    for sub_data in data:
+        # sub_data is a list of (x, y) values for that prod type
+        # x is a timestamp, y a value (MWH)
+        name = MAPPING[sub_data['name']]
+        if not name:
+            continue
```
corradio: I think we should issue a warning using our logger here.

parsers/SE.py
```diff
@@ -0,0 +1,77 @@
...
+            dt = datetime.datetime.fromtimestamp(value['x'] / 1000)
+            prod = value['y']
+            productions[dt][name] += prod
+
+    datetimes = sorted(list(productions))
+
+    # SVK returns almost a point per minute, but only nuclear is updated
+    # with that granularity. Other points are updated every hour, at minute 6
+    # So we'll keep only one datapoint per hour, and datetimes where minute > 6
+    # only keep one datapoint per hour, and only values where minute > 6
```
corradio: Seem like the two lines have the same comments?

parsers/SE.py
```diff
@@ -0,0 +1,77 @@
...
+            or dt.minute < 7):
+            continue
+        last_inserted_datetime = dt
+        prod = productions[dt]
+        to_return.append({
+            'production': dict(prod),
+            'datetime': dt.replace(minute=0),
+            'zoneKey': 'SE',
+            'storage': None,
+            'source': 'www.svk.se'
```
corradio: No need for the `www.` (we only keep the domain) -> `svk.se`.

parsers/SE.py
maxbellec: `MAPPING` intentionally maps two entries (corresponding to the total production and total consumption) to None. If it's another entry, it will raise a `KeyError`.

corradio: got it! I was still in Javascript world :)

Issue #1176, opened 2018-03-04
Extract components from main.js


corradio:
Components to extract/create:
- [ ] contributorList (on left panel)
- [x] countryList
- [x] searchBar

jbuverud:
The countrylist (now known as the zone list) and the search bar has been extracted into components in #1291.

PR #1291, opened 2018-04-05
Integrate rankings panel into left panel

jbuverud:
Screenshot: 

image

I have integrated the rankings list permanently into the left panel (and removed the button to open it). The app state 'highscore' still exists for mobile devices, however. On mobile devices the 'info panels' are also still separate from the rankings list. 

I realize the integration of the rankings list makes the left panel a bit cluttered. However, we have plans to extract the top box into something like a 'welcome modal' which is only shown for new users. This would give some more space for the rankings list in the left bar. 

Also, the rankings list has been updated with the following:

 - An overall ranking number has been added to each country in the rankings list. This is based on all countries in our data, and is therefore obviously not a perfect 'world rank', but I still think it is good enough for our purposes. 

- The rankings list component now automatically separates zones names from country names, assuming all zones are named with the 'country name (zone name)' format. If there are no parantheses, the full name is used. I believe this provides a much better overview in the list, especially when viewing a list of zones from the same country (for instance after searching for a country). 

- I have chosen to include country emissions in the rankings list. When discussing this with @corradio, he felt including emissions would be too much information and clutter the list. However, I believe it is necessary, because it makes obvious the criteria for the ranking. Furthermore, I believe it makes the map much more intuitive for new users, because the boxes with the emissions values is colored in the same way as the map itself, and in this way directly explains our coloring of CO2-intensities. 

- I have added custom, thinner scroll bars for Webkit browsers (Chrome and Safari). 

I would love to hear feedback! Especially regarding the inclusion of emissions in the list, because this was one area where @corradio and I couldn't convince eachother of what would be the best option 😉


jbuverud:
This PR also extracts 2 of the 3 components specified in #1176 .

brunolajoie:
Nice!! My main critics would be that we don't see enough countries at once, by a good margin. We risk always seeing the same top 10 and colors are not so insightful if we see no gradient of it

jbuverud:
If we were to show a significant change in gradient between the items on the list we would need to display approximately 25 items.. I don't think that's realistic without simplifying the list too much. 

However, my assumption is that it is pretty intuitive for users to try to scroll the list when they open the page - that way they will soon discover countries with a higher emission and different gradient. Furthermore I think it is pretty evident from looking at the map (which now always starts fully zoomed out) that there are many zones with higher emissions.

corradio:
Just a thought: if we explain (in the onboarding) that the main scale of
the map is gCO2eq/kWh, do we really need the value it in the list?
I think right now the purpose of showing the gCO2 value in the list is only
to teach the user what the list is sorted by. Maybe there's other ways to
do that, as I think the value is irrelevant for most users.
Just an idea: what about having a very transparent background of the cell
being coloured by the carbon intensity? Just trying to be creating about
how to add the color information to the cell (if we remove the value).
Finally, I agree with Bruno each country line takes a lot of space (and the
number of on-screen results shown is very low). I'd probably try to remove
the separators and have each row smaller (the flags should almost touch?)

Olivier

maxbellec:
Very happy to see the left panel being collapsable!
I'd also like to see a list with gradients. 

I believe there can be ways to show more countries and still have some information (though it will be more cluttered and very likely not very beautiful).

Maybe try that kind of chart with only country and flag on the left, and carbon intensity on the right (with bars of the right colors)?

top20-drone-company-ranking-q3-2016-1

maxbellec:
Very happy to see the left panel being collapsable!
I'd also like to see a longer list with gradients, though I'm not sure how... Maybe try with only the color (no number)?

jbuverud:
Thanks for the feedback guys!

> Just a thought: if we explain (in the onboarding) that the main scale of the map is gCO2eq/kWh, do we really need the value it in the list? I think right now the purpose of showing the gCO2 value in the list is only to teach the user what the list is sorted by. Maybe there's other ways to do that, as I think the value is irrelevant for most users.

I think the value is still very relevant - it allows the users to directly see how different countries compare (instead of only having to rely on color shades). From a UX perspective, relying solely on colors is pretty bad. I believe a 'rankings list' which doesn't show what the ranking is based on is very confusing to users. Furthermore, as mentioned before, showing the values directly and linking them to the background colors also acts as an explanation to the map colors.

However, I think you might have a point in that we should consider the knowledge users get via onboarding. I think one improvement could be to explain the complex CO2-equivalents-per-kilowat-hour unit during onboarding, and then just abbreviate the unit in the list to 'g', in order to reduce the visual clutter.

> Just an idea: what about having a very transparent background of the cell being coloured by the carbon intensity? 

I tried testing this, but it doesn't really look that good. Also, transparent colors are not supported by IE. 

I tried making the rows even smaller, removing the separators and simplifying the emissions unit. I think it looks pretty good! Let me know what you think: 

image


Making the rows smaller than this would require putting the country and zone names on the same line, which I feel severely clutters the list. The list is very information dense already - any more would lead to bad UX in my opinion. Of course it's hard to show interaction through screenshots, but I'm not sure why you think people are not going to scroll down the list to find the higher-emission entries! It's certainly the first thing I would do personally when seeing a design like this 😃 

Also, for reference, this is approx. how many countries will be shown on a HD monitor when we move away the top info box: 

image

If we want to show even more countries at once I think we should instead consider how to make the bottom part of the left panel smaller.

maxbellec:
I also like the last screenshots. It will look better indeed when we'll have removed the top part.

systemcatch:
I love it @jbuverud, significant improvement in design and presentation. The second image is very readable and I wouldn't want to squeeze it down anymore.

Quick idea, could there be a button that reverses the list and shows the dirtiest countries first? That might help with scroll issues.

jbuverud:
> Quick idea, could there be a button that reverses the list and shows the dirtiest countries first? That might help with scroll issues.

This might be a good idea, but I'm not sure how much value it will add. We should in any case keep it to a separate issue. 

During discussions today we agreed to remove the emissions from the list, as there is no good way to show the unit without it taking up too much space. We decided to instead put in a header with a subtitle explaining what the ranking is based on.  

We also decided to remove the "tip" from the panel text, assuming that users will figure out to either click the map zones or the zones in the rankings list.

Here is how it looks now:  

image


And without the top text:
image

brunolajoie:
Just a small text suggestion for the subtitle
>Ranked by carbon intensity of electricity consumed (gCO2eq/kWh)

PR #1294, opened 2018-04-05
EIA parser

maxbellec:
IEA website has quite a few data for the US. We can at least add all load (consumption) forecasts.

Still a work in progress, I'll add other zones and possibly other types of data.


maxbellec:
Here are the datasets available: https://www.eia.gov/opendata/qb.php?category=2123635

corradio:
@maxbellec feel free to merge

PR #1295, opened 2018-04-05
Make Namibian parser read scada timestamp

### LINKED ISSUE #1285: Namibia parser should read out data timestamp
Current time in Namibia is 2 April 14:41 but the [Nampower dashboard](http://www.nampower.com.na/Scada.aspx) is stuck on 1 April 11:45:40.

Timestamp of the data is available in http://www.nampower.com.na/gis/images/File_Info.png

We could OCR that file. A quick attempt using the current tesseract methods and upscaling from 600x20 to 1200x40 seems to have worked alright, I got `SCADA File Dene: 01/04/2018 11:45:40` (it should have been "File Date" but otherwise correct). There is always the danger of OCR misreading, so maybe only ignore the datapoint if the timestamp can be read and is too far in the past.

Screenshot from our page:
na

systemcatch:
fixes #1285 

@jarek do you think this is an ok solution?


parsers/NA.py
```diff
@@ -52,6 +50,31 @@ def get_text_from_image(link, expected_size, new_size, logger, session=None):
     return text
 
 
+def timestamp_checker(session=None, logger=None):
+
+    scada_info = get_text_from_image(session=session, link=timestamp_link, \
+                                    expected_size=(600,20), new_size=(1200,40), \
```
maxbellec: it's a detail, but `\` is not necessary when you are inside any type of brackets. Your indentation is off by 1 char.

parsers/NA.py
```diff
@@ -52,6 +50,31 @@ def get_text_from_image(link, expected_size, new_size, logger, session=None):
...
+                                    logger=logger)
+
+    timestamp = scada_info.split(':', 1)[1]
+
+    try:
+        scada_time = arrow.get(timestamp, ' DD/MM/YYYY HH:mm:ss')
+    except arrow.parser.ParserError as e:
+        logger.warning('Namibia scada timestamp cannot be read, got {}.'.format(timestamp))
```
maxbellec: you should probably fail here, no? you won't be able to use `scada_time` later anyway

parsers/NA.py
```diff
@@ -52,6 +50,31 @@ def get_text_from_image(link, expected_size, new_size, logger, session=None):
...
+
+    data_time = scada_time.replace(tzinfo='Africa/Windhoek')
+    current_time = arrow.now('Africa/Windhoek')
+    diff = current_time - data_time
+
+    # Need to be sure we don't get old data if image stops updating.
+    if diff.seconds > 3600:
+        raise ValueError('Namibia scada data is too old to use, data is {} hours old.'.format(diff.seconds/3600))
+    else:
+        return True
```
maxbellec: why does `check_timestamp` return a value? It never seems used.

parsers/NA.py
systemcatch: I will fix that.

parsers/NA.py
systemcatch: If it fails then we lose the datapoint entirely. The OCR of this timestamp is not super reliable (small, stretched, poor font) so we'd be dropping valid data quite often.

parsers/NA.py
systemcatch: Agree that the return in the else statement can go, the one in the try/except is used to exit if the timestamp can't be read.

parsers/NA.py
maxbellec: :+1:

parsers/NA.py
```diff
@@ -52,6 +50,31 @@ def get_text_from_image(link, expected_size, new_size, logger, session=None):
...
+                                    expected_size=(600,20), new_size=(1200,40), \
+                                    logger=logger)
+
+    timestamp = scada_info.split(':', 1)[1]
+
+    try:
+        scada_time = arrow.get(timestamp, ' DD/MM/YYYY HH:mm:ss')
+    except arrow.parser.ParserError as e:
+        logger.warning('Namibia scada timestamp cannot be read, got {}.'.format(timestamp))
+        return True
```
maxbellec: maybe change to (return value is not used so I'd return nothing)
```python
# the OCR of the is not super reliable, when it fails we assume the date was correct
return
```

parsers/NA.py
```diff
@@ -107,6 +130,7 @@ def fetch_production(zone_key = 'NA', session=None, target_datetime=None, logger
                                    logger=logger)
 
     production = data_processor(raw_text)
+    check_timestamp = timestamp_checker(session=session, logger=logger)
```
maxbellec: probably just
`timestamp_checker(session=session, logger=logger)`
since there is no return value

parsers/NA.py
```diff
@@ -177,6 +201,8 @@ def fetch_exchange(zone_key1, zone_key2, session=None, target_datetime=None, log
     if flow is not None:
         flow = -1 * flow
 
+    check_timestamp = timestamp_checker(session=session, logger=logger)
```
maxbellec: same

parsers/NA.py
```diff
@@ -52,6 +50,31 @@ def get_text_from_image(link, expected_size, new_size, logger, session=None):
     return text
 
 
+def timestamp_checker(session=None, logger=None):
```
maxbellec: very small detail, maybe rename to `check_timestamp`? I'd expect a "Checker" to be a class more than a function

Issue #304, opened 2017-01-20
Add India


corradio:
Possible data source:
- http://vidyutpravah.in/

blackleg:
I found the demand and production of the [State of Punjab](https://en.wikipedia.org/wiki/Punjab,_India), but to implement the parser we require split India in states.

I tried it but I don't found the India subunits.

- [Punjab Production](http://www.punjabsldc.org/pungenrealw.asp?pg=pbGenReal)
- [Punjab Demand](http://www.punjabsldc.org/nrrealw.asp?pg=nrGenReal)

corradio:
That's a pretty cool milestone if we can start adding parts of India.
If you want to look at states (and not subunits), you should try to add `IND` to https://github.com/tmrowco/electricitymap/blob/master/web/topogen.sh#L9, and then re-run that script. It will re-generate `world.json`, and another file, `world_states_props.json` (https://github.com/tmrowco/electricitymap/blob/master/web/topogen.sh#L35) that you can inspect in order to find the proper state.
Just ping me if you have troubles, but this might help get you started.

jarek:
Excellent catch @blackleg! It hasn't occured to me to check for data by state. 

Looks like Gujarat also has data: 
https://www.sldcguj.com/RealTimeData/RealTimeDemand.php

Maharashtra as well, but I only found the image format so far: http://mahasldc.in/wp-content/themes/mahasldc/report2.html

Google searches for "(state_name) electricity generation" have found these for me and might find subsequent ones

jarek:
I wonder how much power goes over interchanges... Karnataka http://kptclsldc.com/Snapshot.aspx looks to be about 10%?

We could really use someone well-versed in India electricity situation

systemcatch:
Wikipedia has some interesting information.  The states seem very well connected.
https://en.wikipedia.org/wiki/Electricity_sector_in_India
https://en.wikipedia.org/wiki/National_Grid_(India)

blackleg:
India Installed Capacity: http://www.cea.nic.in/reports/monthly/installedcapacity/2016/installed_capacity-03.pdf

prson:
I am aware of Indian power scenario and have been following this amazing initiative since sometime. There is a lack of data sources, lot of manual parsing has to be done for India power sector data. 
[...]

PR #1297, opened 2018-04-06
Add missing capacity for IN states with live-data

alixunderplatz:
Based on my comment in [#304](https://github.com/tmrowco/electricitymap/issues/304#issuecomment-379063439), I will add missing capacities to the Indian states which are covered at the moment


alixunderplatz:
I'll add the capacities state-by-state in seperate commits.
I'll give you a sign when it's ready for reviews ;)

alixunderplatz:
Allright, I think I did everything I could do here.

@corradio I updated the cap. sources in the README as well

**Please note:** in some states the installed capacity extremely exceeds the generation.
Strange but, that's what's on the National Power Portal. Maybe our data sources don't disclose the entire generation. Maybe it's just how it is.
For some states we may adapt what is currently mapped "unknown". But I'd like to add capacities first to have the visual comparison.

I added generation types that had "0 MW" installed instead of leaving them blank.
The capacity of types without any generation data were added as well to highlight the proportions

PR #1303, opened 2018-04-06
Add NL

### LINKED ISSUE #1277: Add static carbon intensity for NL exports

corradio:
Fixes #1277 

- Queries ENTSOE to get total load, total net imports, and from those estimate total generation
- Queries ENTSOE to get breakdown of production, setting fossil to None and calculating unknown as the difference between the measured production breakdown and total generation

I've also (as opposed to the screenshot), set biomass to None (appearing as "?")

image

FYI @brunolajoie


parsers/ENTSOE.py
```diff
@@ -569,48 +571,6 @@ def parse_price(xml_text):
     return prices, currencies, datetimes
 
 
-def parse_generation_forecast(xml_text):
```
maxbellec: 36 net deletions, cleaner code and shorter lines :+1:

parsers/NL.py
```diff
@@ -0,0 +1,73 @@
...
+        for e in exchange:
+            e['NL_import'] = e['netFlow'] if zone_2 == 'NL' else -1 * e['netFlow']
+            del e['source']
+        exchanges.extend(exchange or [])
+    df_exchanges = pd.DataFrame.from_dict(exchanges).set_index('datetime')
+    # Sum all exchanges to NL imports
+    df_exchanges = df_exchanges.groupby('datetime').sum()
+
+    # Fill missing values by propagating the value forward
+    df_consumptions_with_exchanges = df_consumptions.join(df_exchanges).fillna(
```
maxbellec: don't you want to put a limit to `fillna`?

parsers/NL.py
```diff
@@ -0,0 +1,73 @@
...
+        method='ffill')
+
+    # Load = Generation + netImports
+    # => Generation = Load - netImports
+    df_total_generations = df_consumptions_with_exchanges['consumption'] - \
```
maxbellec:
```python
df_total_generations = (df_consumptions_with_exchanges['consumption']
                        - df_consumptions_with_exchanges['NL_import'])
```
is a bit cleaner and recommended (but very small detail)

parsers/NL.py
corradio: didn't know you could! Love Pull Requests!

parsers/NL.py
corradio: roger that

parsers/NL.py
```diff
@@ -43,12 +43,12 @@ def fetch_production(zone_key='NL', session=None, target_datetime=None,
...
     # Fill missing values by propagating the value forward
     df_consumptions_with_exchanges = df_consumptions.join(df_exchanges).fillna(
-        method='ffill')
+        method='ffill', limit=3)  # Limit to 3 x 15min
 
     # Load = Generation + netImports
     # => Generation = Load - netImports
-    df_total_generations = df_consumptions_with_exchanges['consumption'] - \
-        df_consumptions_with_exchanges['NL_import']
+    df_total_generations = (df_consumptions_with_exchanges['consumption'] -
```
maxbellec: FYI (I'm totally fine leaving it like this, the comment is more when there are multiple signs)
https://legacy.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

corradio: ah ok got it. Will try to follow that for the next time.

brunolajoie:
And there are sun in the NL data from ENTSOE! 
good news! Many time in the past if was N/A.

PR #1305, opened 2018-04-06
Check year for every datapoint type

### LINKED ISSUE #1168: CL-SIC parser returns wrong datetime
CL-SIC parser returns datetime of 0018 instead of 2018, as illustrated below:

`datetime: 0018-02-18 (06:43:00.000) CET`
`production: {"unknown":0,"coal":2206.6,"hydro":2416.39660,"wind":309.314,....}`
`updatedAt: 2018-02-20 (23:40:06.979) CET`

@systemcatch any idea?

maxbellec:
fixes #1168

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -8,11 +8,18 @@ class ValidationError(ValueError):
     pass
 
 
-def validate_year(item, k):
+def validate_reasonable_time(item, k):
     data_time = arrow.get(item['datetime'])
     if data_time.year < 2000:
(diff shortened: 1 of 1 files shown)


parsers/lib/quality.py
```diff
@@ -28,9 +36,7 @@ def validate_exchange(item, k):
     if data_time > arrow.now():
```
corradio: Maybe validate_year should also validate points in the future?

maxbellec: good point

PR #1308, opened 2018-04-06
Multiple `expected_range` validation

maxbellec:
Add the possiblity to have an `expected_range` for each production type.


parsers/lib/validation.py
```diff
@@ -74,34 +103,36 @@ def validate(datapoint, logger=getLogger(__name__), **kwargs):
     if remove_negative:
         for key, val in generation.items():
             if val is not None and -5.0 < val < 0.0:
-                logger.warning("{} returned {}, setting to None".format(
+                logger.warning("{} returned {:.2f}, setting to None".format(
```
systemcatch: nice idea

parsers/lib/validation.py
```diff
@@ -5,13 +5,34 @@
...
+
+def check_expected_range(datapoint, value, expected_range, logger, key=None):
+    low, high = min(expected_range), max(expected_range)
+    if not (low <= value <= high):
+        key_str = 'for key `{}`'.format(key) if key else ''
+        logger.warning("{} reported total of {:.2f}MW falls outside range "
+                       "of {} {}".format(datapoint['zoneKey'], value,
+                                         expected_range, key_str),
+                       extra={'key': datapoint['zoneKey']})
+        return
```
systemcatch: You might just want to return None explicitly here, but that's really a style choice.

maxbellec:
good point @systemcatch, I tested with a few datapoints, but I should indeed test again with data coming from the parser

parsers/lib/validation.py
```diff
@@ -5,13 +5,34 @@
 from logging import getLogger
 
 
+def check_key_is_present(datapoint, key, logger):
```
corradio: It's more than the presence of a key: it's actually also that the value is not None.
I'd rename to `check_value_is_float`?

PR #1310, opened 2018-04-07
Improve test output formatting and add US-SPP test

systemcatch:
- New tests for US-SPP
- Cleaned up the test output to stop logs and useless warnings being generated.
- Note in readme on how to run all the tests.


systemcatch:
Good to hear, shall I merge or would you prefer to do it?

corradio:
too late I did it :)

PR #1318, opened 2018-04-09
remove sun from NL 'unknown' carbon intensity

brunolajoie:
As it is already part of ENTSOE data stream
image

### DIFF
--- a/config/co2eq_parameters.js
+++ b/config/co2eq_parameters.js
@@ -78,7 +78,7 @@ var countryCo2eqFootprint = {
         return (productionMode == 'unknown' || productionMode == 'other') ? {value: 700, source: null} : null;
     },
     'NL': function (productionMode) {
-        return (productionMode == 'unknown' || productionMode == 'other') ? {value: 556, source: 'assumes 54% gas, 33% coal, 5% biomass, 4% nuclear, 2% solar'} : null;
+        return (productionMode == 'unknown' || productionMode == 'other') ? {value: 563, source: 'assumes 57% gas, 33% coal, 5% biomass, 4% nuclear'} : null;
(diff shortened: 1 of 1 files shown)


corradio:
@alixunderplatz you seemed to think that solar data is suspicious right?

alixunderplatz:
@corradio exactly!
@brunolajoie, the solar data for NL on ENTSO-E is already displayed until 23:45 for today. So maybe it the forecast which is shown.

image

brunolajoie:
ah, good catch!
NL being a special test case for us (we acknowledge that data is going to be imperfect from the start); I'd say we leave these solar forecasts in our parsers. Order of magnitude are goods and who knows it might soon turn into real figures?

alixunderplatz:
@brunolajoie same for today:

image

Solar data was just included that one day, so I'd suggest putting solar back to unknown and using the previous intensity of 556 g/kWh.

brunolajoie:
You know what, thinking it twice, I think we can't adjust dynamically to data quality: PV data is supposed upposed_ to be here from ENTSOE because Tennet does report it in real time (contrary to themal power plants such as biomass, coal & gas which are clearly not available), and if this links btw entsoe and tennet breaks, let's not take the responsibility for it :) 💪

alixunderplatz:
@brunolajoie I think seeing the solar data for NL live on the map was an "accidental" one-time happening on monday. Or do you have some other news?

alixunderplatz:
@brunolajoie It's a mystery where that reliable wind value is coming from^^ 
As for the intensity, these 7 g/kWh won't make a big difference. When the solar values are added the next day, I think it will be ok to keep "unknown" w/o solar?

alixunderplatz:
@brunolajoie I was wrong, solar reappeared today! 👍 ;-)

image

PR #1323, opened 2018-04-10
test bounding boxes in zones.json

maxbellec:
Work in progress - not tested yet

### DIFF
--- a/parsers/test/test_zones_json.py
+++ b/parsers/test/test_zones_json.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python3
+
+"""Tests for BR.py"""
 import json
 import unittest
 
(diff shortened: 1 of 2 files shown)


maxbellec:
I had stupidely named my file `test_zones.json.py` so that it was not discovered by `unittest`. Renaming it to `test_zones_json.py` fixed it and preserves my mental health.

parsers/test/test_zones_json.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+"""Tests for BR.py"""
```
corradio: is this comment up to date?

maxbellec: no, thanks!

PR #1324, opened 2018-04-11
Many more bounding boxes

maxbellec:
FYI, we need this for weather forecasts.

The other minor changes are automatic alphabetical ordering.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -2,10 +2,90 @@
   "AD": {
     "timezone": null
   },
+  "AE": {
+    "bounding_box": [
+      [
+        51.5795186705,
+        22.4969475367
+      ],
+      [
+        56.3968473651,
+        26.055464179
+      ]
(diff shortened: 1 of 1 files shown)


config/zones.json
```diff
@@ -2115,28 +3902,28 @@
...
       "wind": 1740
     },
     "contributors": [
       "https://github.com/systemcatch"
     ],
     "flag_file_name": "us.png",
     "parsers": {
-      "production": "US_NY.fetch_production",
-      "consumptionForecast": "EIA.fetch_consumption_forecast"
+      "consumptionForecast": "EIA.fetch_consumption_forecast",
```
corradio: Seems like you need to merge the latest master onto your branch?
This chance shouldn't be there I think.

corradio: indeed! Tricky github

PR #1326, opened 2018-04-11
Fix localization strings

magol:
* Remove unused strings
* Add missing regions
* Use function co2sub
* Add missing strings to Swedish
* Fix key in pl.json


web/locales/en.json
```diff
@@ -177,6 +174,8 @@
         "DE": "Germany",
         "DJ": "Djibouti",
         "DK": "Denmark",
```
magol: @corradio Should DK remain now that the country is divided into zones? It is a little mixed among other countries.

corradio: We should probably have Denmark (West) and Denmark (East) as additional keys (let's keep DK just in case, even though it's not used).
I've merged the PR just to move fast. Feel free to submit a new one!

Issue #1330, opened 2018-04-13
missing oil data auto repair in  short time


strassburger:
not a mistake, just a report:
missing data and auto repair in  short time
observerved wrong ranking for short time.
was fixed after short time
looks like fine solution 
oil data was / is missing
Observerd at El Hierro and Lanzarote/Fuerteventura
![fuerte-lanza-2018-04-13 08 13 52](https://user-images.githubusercontent.com/28303688/38722717-b88b3652-3ef6-11e8-8a54-a452fe36c485.png)

![el-hierro-2018-04-13 08 11 19](https://user-images.githubusercontent.com/28303688/38722693-a1609db4-3ef6-11e8-980d-8462b87685c0.png)

jarek:
Here it would really help if the backend would reject datapoints that indicate total load in a zone drastically changing from one hour to next. For instance that screenshot from Lanzarote went from ~150 MW to ~10 MW.

It'd be a lot of effort to write parser checks to check for every zone and each possible combination individually.

A backend check against recent history would also help with cases I've seen recently where Germany reports incomplete data so that for example only half of coal generation is shown for a while, then half an hour later it is retroactively corrected.

jarek:
That could help for cases where we get more than one datapoint at a time, yeah

Something like: if the difference between any two datapoints is more than 50 MW and 10+%, raise

corradio:
However this parser doesn't return unknown values (it just returns 0 for all).
Is that supposed to happen? Can we fix it?

systemcatch:
@corradio @jarek for parsers that return data at close fixed intervals (e.g. 5min) we can figure out the max % change that's actually possible during that time and drop anything outside that. However we'd need to be careful handling the return to normal values.

corradio:
@blackleg are you able to make sure your library https://github.com/blackleg/reescraper returns `None` instead of 0 when a value is missing?

jarek:
Validation catches some data problems, e.g. 
![image](https://user-images.githubusercontent.com/47415/40046260-ce449822-582c-11e8-8b01-b8ebeb0edc3e.png)

![image](https://user-images.githubusercontent.com/47415/40046279-dd91e73a-582c-11e8-9f2b-669da396de1f.png)

I haven't seen this being an issue in live data recently. Can the ticket be closed?

strassburger:
Not sure...
@corradio ?
@blackleg ?

jarek:
I'll close this for now, please reopen if you see it happening again

PR #1331, opened 2018-04-13
add minimum generation for Canary Islands

jarek:
Hi,

This should help #1330. Reported total generation below the following values will be rejected:

* ES-CN-FVLZ: 50 MW,
* ES-CN-GC: 150 MW,
* ES-CN-IG: 3 MW,
* ES-CN-LP: 10 MW,
* ES-CN-TE: 150 MW,
* ES-CN-HI: 2 MW

@strassburger can you take a look at these values and see if they are reasonable? They are intended to be a little lower than the lowest value we would expect to normally see.


corradio:
Tests are failing:

```
======================================================================
ERROR: test_fetch_production (test_ES_CN.TestESIB)
----------------------------------------------------------------------
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/circleci/electricitymap/parsers/test/test_ES_CN.py", line 113, in test_fetch_production
    self.assertEqual(production['zoneKey'], 'ES-CN-FVLZ')
TypeError: 'NoneType' object is not subscriptable
```

systemcatch:
I think some of the mock datapoints (https://github.com/tmrowco/electricitymap/blob/master/parsers/test/test_ES_CN.py#L101) fail the new validate step and that results in None being appended to the list, which causes the assert to fail.

jarek:
Ah thanks! Let's see if that fixes it

jarek:
Actually this will still fail because test mocks 25 MW on FV/LZ and it expects to get it back from the parser. The parser won't return it now because that is way too low. The test is wrong IMO, I will update it later.

corradio:
@jarek feel free to merge when ready.

Issue #1304, opened 2018-04-06
Replace close button with back button in zone details view


jbuverud:
We want to replace the close button in the zone details view (shown when clicking on a zone) with a back button, to be shown in the top left part of the view.

phillipnunes:
PR #1333

phillipnunes:
Should we close this issue?

jbuverud:
Fixed in #1333

PR #1333, opened 2018-04-13
Changed icon to left arrow (Back button)

phillipnunes:
Issue #1304

### DIFF
--- a/web/src/styles.css
+++ b/web/src/styles.css
@@ -299,11 +299,12 @@ body {
 .left-panel-country-toolbar {
     display: flex;
     align-items: center;
+    justify-content: space-between;
     margin: 0 0 1rem;
     position: relative;
 }
 .left-panel-toolbar .left-panel-toolbar-back,
-#left-panel-country-back, #left-panel-highscore-back {
+#left-panel-highscore-back {
     line-height: 1;
     cursor: pointer;
     font-size: 1rem;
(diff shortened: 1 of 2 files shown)


jbuverud:
Do you have a screenshot of the new button?

phillipnunes:
@jbuverud 

image

jbuverud:
Can you center the header?

phillipnunes:
Done.

### **When the header is larger**
image

### **When the header is small**
image

jbuverud:
Hmm. It does indeed look a bit strange that it 'jumps around' when the length of the name changes. For now maybe just set it a fixed distance from the arrow (a slightly longer distance than in the USA screenshot)?

The header could probably use a visual overhaul anyway, to match the layout in the zone list (with title/subtitle separation), but that's for another issue/PR 😄

phillipnunes:
image

image

Ok?

jbuverud:
That looks better 😄 

Could you also maybe align it vertically with the top line of the area header? And maybe make it slightly bigger? 

Thanks for all the work you are putting into this! 👍

phillipnunes:
When you say "align vertically with top", you mean it?

image

And make bigger, you mean the width? I think increase the width it will not solve the space on the right because it occurs due to wrapping the text.

By the way...It is always a pleasure to help with projects like these, and actually I am not employed so I have some time available.  =)

jbuverud:
Yes, aligned like that :+1:

I was thinking make the font size bigger, so the icon is bigger and easier to click/tap.

phillipnunes:
Changed.

image

image

image

Issue #565, opened 2017-05-28
Add data for Argentina


jarek:
If my Google-Translate Spanish is correct, there is hourly generation data on http://portalweb.cammesa.com/Memnet1/default.aspx: open "operativos" on the left, then follow "Despacho Real de Generación Térmica" and "Despacho Real de Generación Hidráulica". Make sure to page through the results. Thermal plants will have to be mapped to type, but from other sources, they are mostly/vastly gas, with 6% diesel/oil and no coal. Note that thermal also includes a nuclear plan "ATUCNUCL". There is some delay, so data for 15:00 local time might not be uploaded until 15:30 or later...

Additional resources: 
* https://en.wikipedia.org/wiki/Electricity_sector_in_Argentina
* https://en.wikipedia.org/wiki/List_of_power_stations_in_Argentina

corradio:
@jarek want to take a crack at this one?

jarek:
Hey, sorry, realistically I won't have time to work on this within next 7 days. If anyone else is interested, feel free.

systemcatch:
I'm working on this at the moment.  Took me a while to figure out what was happening on the real time generation page but I now basically understand and should be able to finish the parser.

corradio:
Awesome @systemcatch!

corradio:
Here's an overview of what we still need to accomplish:
- [ ] Prices
- [ ] Interconnector

systemcatch:
@corradio I've got the price for electricity in $(Peso)/MWh all ready to go, one question when fetch_price returns a dict (like in the example) should the 'currency' key have a string value (e.g. 'ARS')?

jarek:
@systemcatch All the other parsers return currency code as string. The examples missing quotes around the EUR is probably a typo.

E.g. from AU.py

    data = {
        'countryCode': country_code,
        'currency': 'AUD',
        'price': obj['PRICE'],
        'source': 'aemo.com.au',
        'datetime': arrow.get(arrow.get(obj['SETTLEMENTDATE']).datetime, 'Australia/NSW').replace(minutes=-5).datetime
    }

corradio:
yes indeed!

corradio:
@systemcatch any progress here?

systemcatch:
[...]

PR #1334, opened 2018-04-15
Create Argentinian exchanges with Chile, Paraguay & Uruguay

systemcatch:
Based on Alex's research I've setup these exchanges. The UY one is kind of slow since it involves 2 network requests, ideally it could make just one but I've not found an easy way of doing that without significantly modifying other functions. Let me know if the target_datetime implementation needs changing.

ref #565 

The Chilean and Brasil exchanges are working but were 0.0 when I ran the parsers.

ar-exchanges


brunolajoie:
awesome! And you're totally right about negative consumption value for UY alex: the same issue happened to us with Switzerland, which we're about to remove from the EMap cause export exceed consumption+import

systemcatch:
I'll try to add the Salto Grande generation back into UY, despite what https://github.com/tmrowco/electricitymap/blob/master/parsers/UY.py#L66 says I think it's being dropped somewhere quite often.

corradio:
Do we have sufficient documentation to say that Salto Grande is connected to the UY grid (and not directly exported)? We should add that to the code if that's the case.

systemcatch:
Hey @corradio I added a reference to the issue where @alixunderplatz explained how the Salto Grande is connected to the UY grid. I will remove the old comment if you think it's no longer needed.

parsers/UY.py
```diff
@@ -69,13 +94,17 @@ def parse_page(session):
     # If this assumption is not correct, then the following line should be uncommented.
     # obj['Hidráulica'] -= max(0, obj['Interconexión Salto Grande'])
```
corradio: We should probably explain why we ignore `obj['Interconexión Salto Grande']` and use `http://www.cammesa.com/uflujpot.nsf/FlujoW?OpenAgent&Tensiones y Flujos de Potencia&` instead

corradio:
ok then let's link to the specific comment (https://github.com/tmrowco/electricitymap/issues/1325#issuecomment-380453296) and remove the old comment that is no longer applicable.

I'm just trying to make it very obvious for someone who might not be familiar with this issue. It's quite tricky.

systemcatch:
Happy for me to merge now Oli?

PR #1337, opened 2018-04-16
Onboarding modal

### LINKED ISSUE #1301: Extract left panel text into onboarding modal
The top part of the current left panel should be extracted into its own separate onboarding modal, shown to users on first page load. 

image

We should also find space for an icon somewhere allowing the users to re-open the modal if they want to see it again.

jbuverud:
Fixes #1301 

The modal is shown the first time the user loads the page. When dismissed, a cookie is set so the user will not see the modal for another 365 days.

image

(note: the modal looks bigger in the below images than in the first one because the screenshots are cropped - the size is the same!)
image

image

image

Bring on the feedback! 😅


maxbellec:
This looks really good!

Maybe you could be more explicit about imports
- In the "know where your electricity comes from" modal, maybe only talk about clicking on a country for details
- In another modal, talk about the fact that we show *consumption* mix taking imports and exports into account, and mention the arrows

jbuverud:
I agree that the 'know where your electricity comes from' modal is mixing two concepts, but I am afraid that splitting it into another separate modal would make the onboarding process too long and tedious. 

My idea was to group the group the concepts of import/exports and consumption mix in one modal, because they both are way of knowing 'where your electricity comes from'. But maybe mixing the two is too confusing? 

I also was reluctant to introduce the word 'consumption mix' directly, because it is kind of science-y and could be difficult to understand for newcomers. If we were to use it, then we should definitely also use it in the area details view on the map, to keep the terminology consistent.

corradio:
Looks good! We can go over the wording face-to-face tomorrow.
It would be nice to use the same font for "Electricity Map" as the logo in
upper left corner.
Also, how would it work on mobile? I was thinking that some users might
want to swipe.

Appart from that, it looks slick!!

brunolajoie:
Looking good! 

- We could potentially make the "next" and "back" arrows bigger cause users will have to click on it so the easier the better.
- How could we make the first screen a bit more interesting? It seems a bit useless to me. Could we merge it with the first? It would leave us with one more view to show a country details tooltip for instance?

Just suggestions here.
Lets discuss tmrow!

jbuverud:
The design (including the size of the buttons and the inclusion of a 'title' view) is taken pretty much directly from Google's own designs, so they are already very refined and user tested. I don't see a need to change what is already working :smile:

On mobile the modal is full-screen and the buttons moved to the bottom.

 I can post some screens on Wednesday when I come in (note: not tomorrow - I have a mandatory course at my other job :smile:)

maxbellec:
I also thought the first screen is not necessary

web/src/components/onboardingmodal.js
```diff
@@ -0,0 +1,167 @@
...
+
+const defaultViews = [{
+  headerImage: 'images/ms-icon-310x310.png',
+  headerCssClass: 'logo-header',
+  textCssClass: 'brand-text',
+  textHtml: `<div><h1>${translation.translate('onboarding-modal.view1.header')}</h1></div>
+      <div><h2>${translation.translate('onboarding-modal.view1.subtitle')}</h2></div>`,
+},
+{
+  headerImage: 'images/mapExtract.png',
```
corradio: maybe you want to put the images in an `onboarding` folder?

web/src/components/onboardingmodal.js
```diff
@@ -0,0 +1,167 @@
...
+);
+
+const defaultViews = [{
+  headerImage: 'images/ms-icon-310x310.png',
+  headerCssClass: 'logo-header',
+  textCssClass: 'brand-text',
+  textHtml: `<div><h1>${translation.translate('onboarding-modal.view1.header')}</h1></div>
+      <div><h2>${translation.translate('onboarding-modal.view1.subtitle')}</h2></div>`,
+},
+{
```
corradio: nicely structured 👍

web/src/main.js
```diff
@@ -762,7 +768,10 @@ d3.selectAll('.info-button').on('click', () => dispatchApplication('showPageStat
 d3.selectAll('.highscore-button')
   .on('click', () => dispatchApplication('showPageState', 'highscore'));
 
-
+// Onboarding modal
+if (onboardingModal) {
+  onboardingModal.onDismiss(() => Cookies.set('onboardingSeen', true, { expires: 365 }));
```
corradio: 365 days? I think we can put it longer than that.

web/src/styles.css
```diff
@@ -976,6 +970,155 @@ sub {
     text-decoration: underline !important;
 }
 
+.modal-background-overlay{
```
corradio: There's an `overlay` class that you might be able to re-use (if I remember correctly)

web/views/pages/index.ejs
```diff
@@ -152,18 +154,6 @@ co2Sub = function(str) {
 
                 <div class="panel left-panel">
 
-
-                    <div class="left-panel-info">
-                        <div class="left-panel-initial-text">
-                            <p>
-                                <%- __('panel-initial-text.thisshowsrealtime') %> <span style="color: lightgray"><%- __('panel-initial-text.whereelectricitycomesfrom') %></span> <%- __('panel-initial-text.and') %> <span style="color: lightgray"><%- co2Sub(__('panel-initial-text.howmuchco2')) %></span> <%- __('panel-initial-text.wasemitted') %>.
```
corradio: Would be nice to remove the associated keys in the translation

web/views/pages/index.ejs
```diff
@@ -55,10 +55,12 @@ co2Sub = function(str) {
         <link rel="stylesheet" type="text/css" href="dist/styles.<%= stylesHash %>.css">
         <link rel="stylesheet" type="text/css" href="dist/vendor.<%= vendorStylesHash %>.css">
         <link rel="stylesheet" href="css/font-awesome.min.css">
-
+        <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
```
corradio: Would be nice to embed it in `css/...` (so the app users have it locally).

web/views/pages/index.ejs
```diff
@@ -55,10 +55,12 @@ co2Sub = function(str) {
...
+        rel="stylesheet">
 
         <!-- Fonts -->
         <link href="https://fonts.googleapis.com/css?family=Catamaran" rel="stylesheet">
```
corradio: Same here in principle (but only if it's easy to do)

web/views/pages/index.ejs
```diff
@@ -331,8 +330,11 @@ co2Sub = function(str) {
                     </div>
                     <div class="floating-legend-container">
                         <div class="floating-legend-mobile-header">
-                            <span>Legend</span><img class="toggle-legend-button down" src="images/material-icons/ic_call_received_lightgrey_24px.svg">
-                            <img class="toggle-legend-button up visible" src="images/material-icons/ic_call_made_lightgrey_24px.svg">
+                            <span>Legend</span>
+                            <i class="material-icons toggle-legend-button up">call_made</i>
+                            <i class="material-icons toggle-legend-button down visible">call_received</i>
+                        <!-- <img class="toggle-legend-button down" src="images/material-icons/ic_call_received_lightgrey_24px.svg">
```
corradio: commented code

jbuverud:
Here's some new screenshots following the requested changes:

image

image

image

image
(in this last image the checkmark button is shown in its hover state)

web/views/pages/index.ejs
```diff
@@ -54,11 +54,13 @@ co2Sub = function(str) {
         <!-- CSS -->
         <link rel="stylesheet" type="text/css" href="dist/styles.<%= stylesHash %>.css">
         <link rel="stylesheet" type="text/css" href="dist/vendor.<%= vendorStylesHash %>.css">
-        <link rel="stylesheet" href="css/font-awesome.min.css">
-
+        <!-- <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
```
corradio: Commented code should be removed.

web/views/pages/index.ejs
```diff
@@ -54,11 +54,13 @@ co2Sub = function(str) {
...
+        rel="stylesheet"> -->
 
         <!-- Fonts -->
         <link href="https://fonts.googleapis.com/css?family=Catamaran" rel="stylesheet">
```
corradio: Can you add a local font?

jbuverud: Right, forgot about that one! Will do.

PR #1344, opened 2018-04-19
allow up to 2 hours for Namibian data

jarek:
As noticed by @systemcatch :

> There's lots of points that are just over 1 hour old. However Namibia stopped following DST in 2018, but maybe the scada dashboard hasn't been altered to reflect that.

### DIFF
--- a/parsers/NA.py
+++ b/parsers/NA.py
@@ -75,7 +75,9 @@ def check_timestamp(session=None, logger=None):
     diff = current_time - data_time
 
     # Need to be sure we don't get old data if image stops updating.
-    if diff.seconds > 3600:
+    # In April 2018, timestamp is regularly just over an hour old.
+    # We think it might be due to DST misconfiguration, so allow up to 2 hours.
+    if diff.seconds > 7200:
         raise ValueError('Namibia scada data is too old to use, data is {} hours old.'.format(diff.seconds/3600))
 
 


corradio:
@jarek silly question: why don't we put the dashboard datetime in the return object, and let the backend check for its validity?

jarek:
Initially this was not done because we weren't reading the timestamp at all. The timestamp was relatively recent, so the parser just returns current time.

Then we introduced this check when the data went out of date for a few days, but allowed fall-through in case the OCR fails, which it sometimes does.

We could change the logic to return either the time read from OCR, or current time in case the OCR fails. Will that be okay?

corradio:
Sorry to be adding my grain of salt even though I haven't followed thoroughly all of the discussions.
@systemcatch I'll let you decide and merge if applicable.

systemcatch:
@corradio It fails often enough that we would be dropping good data, I wouldn't trust using it for the parser datetime key, especially given that arrow can be lax parsing time strings.

corradio:
OK !

jarek:
No worries @corradio, these are good questions! It's just that dealing with OCR adds a lot of uncertainty - even in 2018 a sans-serif timestamp in black text on white background that is perfectly read by a human can confuse free OCR software.

corradio:
Makes sense indeed.

PR #1353, opened 2018-04-26
Fix ENTSOE forecast fetch

corradio:
ENTSOE forecasts sometime have two datasets in one api request. We need to select the right one.
This affects:
- consumption forecasts
- generation forecasts
- current consumption


maxbellec:
what were the effects of that bug?

corradio:
0 forecasted generation for large periods of time:

image

PR #1358, opened 2018-04-29
Update CA-BC capacities

daveachuk:
Taken from the https://en.wikipedia.org/wiki/List_of_generating_stations_in_British_Columbia wiki page which I have just updated to the latest figures. These capacities exclude non-grid-connected capacity (remote or industrial microgrids). Now to get BC Hydro to release real-time data......

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -591,12 +591,15 @@
       ]
     ],
     "capacity": {
-      "biomass": 356,
-      "gas": 514,
-      "hydro": 15004,
-      "oil": 46,
-      "wind": 673
+      "biomass": 838,
+      "gas": 507,
+      "hydro": 15898,
+      "solar": 1,
+      "wind": 704
     },
+    "contributors": [
+      "https://github.com/daveachuk"
+    ],
     "flag_file_name": "ca.png",
     "timezone": null
   },


corradio:
Thanks! No more `oil`? Also, can you update the README with your source for the capacities?

jarek:
According to the Wikipedia list, 58.8 MW of oil remains in off-grid installations. Of course that's basically rounding error.

I think we could also skip specifying 1 MW of solar - there is much more, of course, but decentralized and possibly not centrally tracked.

@corradio the README already links to https://en.wikipedia.org/wiki/List_of_generating_stations_in_Canada which links to the "in British Columbia" list... a bit more compact than linking to list for each province in turn?

@daveachuk Thanks for keeping up on this! I once asked the BC Sustainable Energy Association for who I could ask about real-time data - their suggested contacts were the Ministry of Energy, Mines and Petroleum Resources, in particular [Electricity Generation and Regulation](http://dir.gov.bc.ca/gtds.cgi?show=Branch&organizationCode=EMPR&organizationalUnitCode=EGR), or universities, "maybe someone at the Pacific Institute for Climate Solutions." [BCUC](http://www.bcuc.com/) is a regulator, they don't have any real-time data of their own, but after a months-long Proceeding they might require the generators to publish their data... I haven't contacted any of these groups yet, feel free to write them if you're interested. It seems to me that going through higher levels of provincial government and pointing out an opportunity to look green and climate-friendly (while demanding that complete real-time data is published) would be the most effective, but I don't have much experience on that level.

daveachuk:
@corradio As jarek said, the source is the [BC wiki page](https://en.wikipedia.org/wiki/List_of_generating_stations_in_British_Columbia) which is indirectly linked in the Readme. I spent the last a bunch of time over the last 2 days updating that page from various sources, mostly BC Hydro's list of contracts with Independent Power Producers as well as the more-comprehensive stats from their own facilities.

I excluded all off-grid installations in the stats in the attached edit, which is why `oil` went to 0, since that number was coming from diesel generators. I presumed that for this application you're only concerned with the energy people actually would be using from the grid, and any micro-grids should be excluded. There are also some off-grid hydro and biomass which are also excluded from the on-grid totals. `solar` at 1MW is more of a "look, we have solar too!" statement :)

@jarek I think you're definitely right in that the top-down approach to encouraging realtime data is the way to go. BC Hydro might be able to produce those stats but the BC grid is a bit more than BC Hydro-- [FortisBC](https://www.fortisbc.com/Pages/default.aspx) also runs their own grid which contains 4 hydro projects of their own, and another 5MW of IPPs with unspecified generation sources. Fortis also has their own cross-border entitlement agreements regarding Montana's Libby dam which impacts their hydro projects. They're connected to the BC Hydro grid and get ~25% of their energy from BC Hydro, but also have cross-US-border connection via shared access to a private Teck line, although I think their Libby entitlements actually flow through the main BC Hydro connection. So things get messy, but with enough data from all parties it should be possible to distill down to a simple answer. 

Thanks for the link to the Electricity Generation and Regulation branch, I'll ping someone there sometime to see what the situation is or what approach might be most effective.

jarek:
Let's get it in then :+1:

PR #1359, opened 2018-04-29
Update CA-BC default CO2 values

daveachuk:
Updated to 2017 numbers from here:
https://www.bchydro.com/content/dam/BCHydro/customer-portal/documents/corporate/environment-sustainability/environmental-reports/ghg-intensities-2007-2015.pdf


daveachuk:
Good thoughts. I'm not an expert on this stuff, just a hobby, so treat what I say as such...

One point is that BC is a net energy exporter on both of its exchanges (or it has been since 2015 when a number of additional hydro stations have come online, save for part of 2016 where low water levels affected hydro outputs). So one could make the argument that for any one point in time where BC's momentary emissions are higher due to importing US or AB energy, there will be always be another point in time (and then some) where the export of BC helps offset carbon produced in those other markets. This does diminish the 'realtime' aspect of the data to make assumptions like that however.

image

The realtime flow data and the above graph are based on total flows, which would include any Columbia River Treaty entitlements. The entitlements are basically there to prevent California or others from getting first choice of that 50% of the lower Columbia dams' energy, but do not represent a mandated flow. But overall in the last 3 years BC exported much more than they would have requested under the Treaty. As a bit of trivia, recently the controversial $10B+ Site C Dam under construction in BC had been called fiscally irresponsible because BC could have imported that much energy (1.1MW) from unused Columbia River Treaty entitlements at a lower price/MWh. Although doing so would have likely come at the cost of Washington and/or its connected grids having to use more thermal generation.

Yes, I suspect BC Hydro counts biomass as carbon-neutral. Like you say they don't provide any details of this on their annual report. Much of the larger-scale biomass is from pine-beetle killed timber, which is harvested en masse, and is turned into pulp & paper where possible, the rest left to rot or in this case burned as fuel. Although it'd decompose anyway there's something to be said about accelerating that process through combustion rather than spreading it over 30+ years of natural decay. I guess their logic is that by clearing the forest it allows regeneration of new vegetation which also absorbs CO2 faster than a mature forest, so the 'accelerated' nature of that CO2 release is offset?  The other source of biomass is logging waste, of which there is a lot in BC and falls into a similar category -- if not used for biofuel it's often burnt at the cutblocks, so using it as biofuel is obviously the better play, but whether or not it counts as CO2-neutral is a bit fuzzy.

The other part of BC biomass is where BC Hydro has helped paper and pulp mills, which once burned waste material biomass only for their own thermal/cogeneration use, has signed Power Purchase Agreements with the mills for them to run at their full capacity so that their boilers/generators are always operating at maximum efficiency. It would be interesting to see how they calculate the net CO2 impact for this. I have a friend who works at BC Hydro conservation and will maybe see if I can glean any answers from her.

Not disagreeing with anything, just sharing what I've learned in the last few days in case it's of use!

brunolajoie:
Super interesting findings. If we wanna stick to EMAP usual way of dealing with this, we shouls probably simply think first order (as long as reasoning it's traceable)
- We take the latest share of power generation by fuel, (like below for instance) 
- We compute a default carbon intensity based on EMAP default factor per fuel (IPCC's median value)
- We set the new default carbon intensity of the zone, and we write all these details in a transparent manner in the code as comments
If you're ok with that @daveachuk , feel free to improve your PR and i'll approve it! 
image

jarek:
OK, so as it is assigned to me, shall we give the export intensity `0.88*24 + 0.09*240 + 0.01*490 + 0.01*11 = 48` gCO2eq/kWh? It's based on data from a somewhat-neutral third party (National Energy Board), defensible based on Electricity Map methodology, and would still improve Alberta figures. I'm ignoring the oil "<1%" generation since that's offgrid and as far as I know we don't count offgrid in our other zones.

(When Alberta generation is at 550 gCO2eq/kWh and importing 10% from BC, using 48 gCO2eq/kWh gives 499.8 gCO2eq/kWh overall, while using 9 gCO2eq/kWh would give 495.9 gCO2eq/kWh.)

jarek:
I'm actually not sure if I have permission to do so, so you should be able to do it yourself. As I understand it, editing https://github.com/daveachuk/electricitymap/blob/patch-2/config/co2eq_parameters.js online should update the pull request as well. You should only need to change `carbonIntensity`, `source` and `URL`. I guess `renewableRatio` could be changed from 0.98 to 0.97, but that's rounding error anyway...

daveachuk:
Thanks! I've updated as requested -- one minor change the default for biomass in co2eq_parameters.js is `230` as opposed to your calculation which used `240`, which made it `47 gCO2eq/MWh`

brunolajoie:
Awesome! would you mind updating the comment with the % used, so that later we can refine this with better figures than those of IPCC:  For instance
`Average carbon intensity using 2016 mean production per fuel source (88% hydro, 9% biomass, 1% wind, 1% gas) and IPCC default emission factors)`

daveachuk:
Thanks, I've done so and also noted that this is using lifecycle emissions factors so if someday in the future production-only emissions factors are decided to be used for whatever reason it'll be obvious it should be changed.

systemcatch:
@brunolajoie all good to go here?

PR #1363, opened 2018-04-30
stattnet NO-NL exchange parser in NL production

maxbellec:


parsers/NL.py
```diff
@@ -38,10 +40,25 @@ def fetch_production(zone_key='NL', session=None, target_datetime=None,
...
-            del e['source']
         exchanges.extend(exchange or [])
+
+    # add NO data, fetch once for every hour
+    zone_1, zone_2 = sorted(['NO', zone_key])
+    exchange = [statnett.fetch_exchange(zone_key1=zone_1, zone_key2= zone_2,
+                                        session=r, target_datetime=dt.datetime,
+                                        logger=logger)
+                for dt in arrow.Arrow.range(
+            'hour',
```
corradio: The resolution must match the one we use in the exchanges parser.
Therefore, it should be 5min? (or whatever frequency we use for the statnett parser)

maxbellec: the NO resolution must match the ENTSOE resolution, which is hourly here, when I test the code just above

corradio: what I'm saying is: shouldn't we fetch at the highest resolution possible, aggregate, and then work with hourly data? Let's discuss tomorrow else.

maxbellec: the exchanges parser fetches, well, every time it's launched, so it's stochastic and will be hard to match. Is it really worth fetching every 5 minutes and align other exchanges? Will the client see the difference?

corradio:
If you only fetch hourly you might be unlucky and have an outlier compared
to the mean during an hour.
This will a discrepancy when computing our own consumption compared to
entsoe consumption.

maxbellec:
latest data is available every minute, historical data only every hour
screenshot at 2018-05-04 09-43-18

Issue #1040, opened 2018-01-29
Northwest USA - Bonneville Power Administration (BPA)


systemcatch:
There is 5 minute data broken down by type at https://transmission.bpa.gov/business/operations/Wind/baltwg.aspx

BPA covers several states in the Northwest US.

![bpa](https://user-images.githubusercontent.com/30196510/35506011-dba36fa2-04df-11e8-8dca-9e41c9d064cf.png)

alixunderplatz:
These borders of BPA are gratifiying 🙏 
Is that an old map ... what happened to MISO ??? 🤣 😆 

BPA have a geoportal:
http://bpagis.maps.arcgis.com/home/index.html

I assume, the grey lines are their balancing territory now?

![image](https://user-images.githubusercontent.com/25743609/35509997-8fe80bd8-04f6-11e8-8415-89a12b95ea4b.png)

alixunderplatz:
That BPA area seems to have huge generation excess (> 6GW), any infos on their site on power flows?

The DC line that is pointing to the upper left corner of Nevada in the image above is this one:
https://en.wikipedia.org/wiki/Pacific_DC_Intertie
It's going to the LA area and has 3100 MW transmission capacity

HansHyde:
BPA is a federal owned agency (like a Canadian "Crown" Company), that is only generation & transmission.

They, like TVA, generate & sell wholesale to any number of MOUs, Co-Ops, IOUs where their transmission lines cross-over.  At least 1/2 of the grey border territory above is PacifiCorp (IOU) - not a lot of people, but a large part of the land area.

HansHyde:
The ISO Today app (CAL-ISO website) shows the wholesale prices of every "Node", where a "Node" is a substation, switchyard or step-up sub at a generation station, for the entire Western EIM (Electricity Imbalance Market).  Nodes data is available in realtime, 5 minute frequency.

Prices & Congestion at Nodes does not tell flows, but it does offer clues as to how the system is operating.  How to convert that to flows (for use within EM) is beyond me, but there are 100s (thousands?) of Nodes that will ultimately determine 1) the generators (for the generation mix applicable) & 2) the end customers (regions of consumption applicable).

IMHO EM Parsers & their json territories are not designed for this, so not certain how the data can/will be scraped & parsed.

[...]

Issue #143, opened 2016-11-23
Split US


corradio:
- [x] Obtain generation mix data for each ISO (https://pyiso.readthedocs.io/en/latest/options.html)
- [ ] Obtain shapefile/GeoJSON for each ISO (DATA MISSING)
- [ ] Import/exports for each ISO (DATA MISSING)

martindaniel4:
👌

sYnfo:
Is this something that someone new to the project could help with? If so, could you please specify in more detail what would be needed to get US on the map?

corradio:
With pleasure! First thing to do is to use the library mentioned before to pull production mix informations from the US by writing a parser. Let's discuss in more details on http://slack.tmrow.co

Olivier

corradio:
@sYnfo are you still up for helping?

corradio:
I started working on this at https://github.com/corradio/electricitymap/tree/data/us

garciaev:
Hi - 

I am interested in getting the USA on the map. Where do I start?

corradio:
Hi @garciaev ,

Thanks! First, we need to get data for the boundaries of each ISO, for which we already have production data. Can you help? I didn't succeed in finding geoJSON or shapefiles..

garciaev:
OK just to be sure, we basically need geoJSON or shape files that have boundaries for the American ISOs here: http://www.isorto.org/about/default 

Correct? I am currently searching for the California ISO boundary map.

corradio:
That's right!

jacobbaratta:
http://www.eia.gov/beta/realtime_grid/ has two vector graphics that may be convertible with some work. I ripped the simpler vector graphic into a standalone svg, but I have zero experience with GIS and my short foray into investigating the feasibility of converting the SVG indicated that while it's probably possible, it's too complicated for a complete novice like me to attempt. I've attached the SVG to this post. Even if it can't be converted it can still probably help in some way with construction of a North American map.

[eia-mod.svg.zip](https://github.com/corradio/electricitymap/files/782157/eia-mod.svg.zip)

garciaev:
Hmm... not sure those EIA regions are the same as ISO member boundaries here? http://www.isorto.org/about/default

I think I am just gonna call the customer service and see if they can forward me to someone who can provide us with the boundary files.

jacobbaratta:
[...]

PR #1366, opened 2018-05-01
Add Bonneville Power Administration

systemcatch:
Currently this is just the parser until we figure out how to display the area on the map.

ref #1040


brunolajoie:
niiiice !

parsers/US_BPA.py
```diff
@@ -0,0 +1,131 @@
...
+
+    processed_data = []
+    for index, row in df.iterrows():
+        production = row.to_dict()
+
+        for key in keys_to_remove:
+            production.pop(key)
+
+        dt = production.pop('Date/Time')
+        mapped_production = {GENERATION_MAPPING[k]:v for k,v in production.items()}
```
maxbellec:
```python
mapped_production = {GENERATION_MAPPING[k]: v for k, v in production.items()
                     if k not in keys_to_remove}
```
might be slightly more elegant

systemcatch:
The parser is all good to go I think, does anyone have thoughts on how to represent this area on the map?

systemcatch:
Chatting to @alixunderplatz on slack we came to the conclusion that using the state of Washington as a rough representation is a good start.

Interestingly WattTime (developer of pyiso) have a map of the USA showing all of the balancing authority areas, see http://explorer.watttime.org/. They are using Leaflet to overlay these complex shapes onto a Mapbox base, @corradio could we do something similar?

corradio:
If we have the shapes defined in lonlat space we can integrate them on the map, but else it will be messy.. I've looked in the past at using the WattTime geometries to use in ElectricityMap, but that turned out not to be a good solution as the zoom features will become imprecise.
Issue #143 might have the geometries already defined? Else let's use Washington?

Oli

systemcatch:
Implemented the area.

us-bpa

corradio:
woop woop!

brunolajoie:
@systemcatch is on a killing spree! 💪

jarek:
I know the U.S. power systems situation is a royal pain, but 1.3 GW nameplate of semi-lignite at [Centralia](https://en.wikipedia.org/wiki/Centralia_Power_Plant) just got coloured hydro-green ¯\_(ツ)_/¯

systemcatch:
:scream: we should fix the US map areas asap.

brunolajoie:
Daaaaaaaaaamn 🙀 🙀 🙀 🙀

PR #1367, opened 2018-05-01
Split zone short names in translation files into zone names and count…

### LINKED ISSUE #1341: Descriptive zone names
I really like how subregions with definite names are rendered in the new list:

image

but I'm not so keen on those with descriptive names:

image

It's a bit hard to explain what seems wrong about those, but I think my main complaint is that these names don't stand on their own. For cases like Yukon and El Hierro the small-text is mostly only used to inform users who might not know where Yukon is, or which La Palma is meant. However "West" doesn't mean anything on its own, and one _needs_ either the smaller text or the flag (which people might not know).

Suggested guideline: if it wouldn't be the title of the region's Wikipedia article (as is the case with for example [North Island](https://en.wikipedia.org/wiki/North_Island)), don't use it as the main name of the zone.

Currently affected are:
- Denmark East/West
- Brazil zones
- Malaysia

Suggested solution: rename to "West Denmark", "Brazil North-East", "Peninsular Malaysia", or similar. Better yet if we could have their proper local names - though admittedly electrical subdivisions might not have names that ordinary people know.

While we're at it I would also flip the order in the "full name" displayed in the detail view, so that it's "California (United States)" and not "United States (California)".

I'm writing because this has been bugging me a bit, but this is only my opinion. Does anyone else have thoughts about this?

jbuverud:
…ry names. Added helper function in translation.js for getting full zone names (in format "Zone name (Country name)")

Fixes #1341

### DIFF
--- a/web/src/helpers/translation.js
+++ b/web/src/helpers/translation.js
@@ -37,8 +37,6 @@ exports.getFullZoneName = function getFullZoneName(zoneCode) {
   const countryName = exports.translate(`zoneShortName.${zoneCode}.countryName`);
   if (!countryName) {
     return zoneName;
-  }
-  var ret = `${zoneName} (${countryName})`;
- 
+  } 
   return `${zoneName} (${countryName})`;
 };
--- a/web/src/helpers/translation.js
+++ b/web/src/helpers/translation.js
@@ -2,7 +2,7 @@ var exports = module.exports = {};
 
 // Import all locales
 var locales = {};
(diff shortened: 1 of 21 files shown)


web/src/helpers/translation.js
```diff
@@ -28,3 +28,17 @@ exports.translate = function() {
...
+exports.getFullZoneName = function getFullZoneName(zoneCode) {
+  const zoneName = exports.translate(`zoneShortName.${zoneCode}.zoneName`);
+  if (!zoneName) {
+    return zoneCode;
+  }
+  const countryName = exports.translate(`zoneShortName.${zoneCode}.countryName`);
+  if (!countryName) {
+    return zoneName;
+  }
+  var ret = `${zoneName} (${countryName})`;
```
corradio: todo: remove

PR #1370, opened 2018-05-02
Moved tab bar to bottom. Changed font in zone list header. Mobile pad…

jbuverud:
…ding fixes.

I've used google's design guidelines for the general layout of the tabs (including text labels): https://material.io/guidelines/components/bottom-navigation.html#bottom-navigation-behavior

I used the "catamaran" font, which is also used elsewhere in the app, for the tab labels. If we want to adhere 100% to googles guidelines, we could change it to Roboto, but then that font would need to be embedded as well. I think Catamaran looks good. 

I tried changing the zone list title text to Catamaran, and I think it looks much better. What do you guys think?

There is an issue with the tab bar being the same color as the zone list/info tab, and I've therefore included a drop shadow on the top of the tab bar (similar to what we are doing with the header). I haven't seen many apps with a drop shadow from the tab bar, but I think it looks ok. Let me know what you think. 

image

image

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -1021,6 +1021,9 @@ function routeToPage(pageName, state) {
 
   d3.selectAll('#tab .list-item:not(.wind-toggle):not(.solar-toggle)').classed('active', false);
   d3.selectAll(`#tab .${pageName}-button`).classed('active', true);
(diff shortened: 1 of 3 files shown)


brunolajoie:
Top!

PR #1378, opened 2018-05-03
Fix #393 Peat combustion

### LINKED ISSUE #393: Peat combustion in Finland and Ireland / total gCO2eq/kWh 25-30% worse
Hello once again,
I had a look on the ENTSO-E data regarding Finnish and Irish electricity production.

They both have a significant number of **peat power plants** installed. And peat (together with Estonian oil shale ;) ) is even worse than coal when it comes to climate impact.
Currently on the map, the electricity generation from peat in Finland and Ireland seems to be contained in biomass with a CO2eq of 230 g/kWh.
But it is way worse than this, somewhere between 1,100 and 1,200 g/kWh or even higher. Since numbers for emissions from peat for electricity generation are kind of hard to find,  I tried to calculate that by using a specific emission value:
[380 g/kWh[thermal]](http://www.volker-quaschning.de/datserv/CO2-spez/index_e.php) / 33% efficiency = 1,140 g / kWh[electric]

Finland has 1,077 MW and Ireland has 344 MW of peat capacity installed. 
In the ENTSO-E data, it is declared as **FOSSIL peat**. 
Suprisingly, Finland [declared peat as "slowly renewable"](https://en.wikipedia.org/wiki/Peat_energy_in_Finland#Definition), which is actually taking some thousand years - let alone draining and destroying the swamp lands/bogs from which the peat is produced has a negative environmental and climate impact on its own.
That's why, in my opinion, peat should not be included in biomass on the map - it is not restored anually or in a few decades, such as usual biomass, but in hundreds or thousands of years. It should rather be included in an own category (e.g. "lignite/peat", if the coal split is implemented).

FYI, there is an analysis on whether peat is sustainable or not. Check out the first paragraph of the conclusion. 
https://www.researchgate.net/publication/222546437_How_sustainable_is_the_use_of_peat_for_commercial_energy_production
There it says:

> "On all undisturbed Finnish peatlands probably less peat grows than is being used, and on the 1200 km2 allotted to the industry growth is 85 times too slow. Also, peat that is growing beyond commercial reach cannot be said to re-plenish a renewable resource. The requirement that sustainable use of a renewable resource must be less or equal to its rate of regrowth is not fulﬁlled."

And since global warming is the main issue of this century, every additional release of unnecessary CO2, that can not be "recycled" within very short time, is "pure evil" for our climate.

Check out the Excel-file attached. I copied some Finnish and Irish data from ENTSO-E for 22-24 Feb. and applied the emission factors for most relevant types of generation to it. I hope I didn't do any mistakes.

[Peat in Finland and Ireland.xlsx](https://github.com/corradio/electricitymap/files/802161/Peat.in.Finland.and.Ireland.xlsx)

For Finland's domestic production the specific emissions in the period would rise from 233 g/kWh by about 30% to 300 g/kWh, considering Peat as non-renewable fuel (1,140 g/kWh CO2eq) instead of Biomass (230 g/kWh CO2eq).
For Ireland, they would increase from 315 g/kWh by about 25% to 388 g/kWh in the analyzed period. 
I did not consider any imports!

What is your opinion on peat - is it fossil or renewable? Should it be considered as "Biomass" or not? Is the difference high enough?

Greetings, Alex

maxbellec:
Thanks @alixunderplatz


parsers/ENTSOE.py
```diff
@@ -628,16 +628,17 @@ def validate_production(datapoint, logger):
...
+    if 'Biomass' in values or 'Waste' in values:
+        return (values.get('Biomass', 0)
+                + values.get('Waste', 0))
 
 
 def get_coal(values):
-    if 'Fossil Brown coal/Lignite' in values or 'Fossil Hard coal' in values:
-        return values.get('Fossil Brown coal/Lignite', 0) + \
-               values.get('Fossil Hard coal', 0)
+    if True and ('Fossil Brown coal/Lignite' in values or 'Fossil Peat' in values
```
corradio: True?

maxbellec: debatable, but I don't like this indentation
```
    if ('Fossil Brown coal/Lignite' in values or 'Fossil Peat' in values
        or 'Fossil Hard coal' in values):
        return (values.get('Fossil Brown coal/Lignite', 0)
```
but I agree that adding `True` is confusing

Issue #973, opened 2018-01-02
Moldova MD / emission factor for gas


alixunderplatz:
@brunolajoie

In #926, I mentioned that electricity production from natural gas in *Moldova* happens with an efficiency of ~22%, which leads to CO2-emissions of ~1000 g/kWh for the gas units. However, this efficiency was based on numbers from 2008, but I don't think that more efficient gas units were put into operation in the past decade in MD.

Here's the [source for the efficiency](https://www.google.de/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwjhvbnshpLYAhWIEVAKHaIIACQQFgguMAA&url=http%3A%2F%2Fget-moldau.de%2Fdownload%2Fpolicypapers%2F2010%2FGET%2520Moldova_PP%252001%25202010_en.pdf&usg=AOvVaw37WEhiS1RHEn9lR52Ono--) (page -3-) 
![image](https://user-images.githubusercontent.com/25743609/34484931-345a12fe-efc8-11e7-8187-772e643b2e1c.png)

I used [this as source for the CO2 emissions](http://www.klimaneutral-handeln.de/php/kompens-berechnen.php#rechner): 

burning 1 kWh of natural gas leads to 220 g CO2 emissions.
So 220 g/kWh(thermal) / 22% = 1000 g/kWh(electrical).

If you think the data is precise, let's use a country specific factor for MD ;)

HansHyde:
@alixunderplatz there is no such thing as "1 kWh of natural gas".

There are 52.16 kG of CO2 for every 1 million BTU of Methane combusted.  How the efficiency (heat rate) of the plant determines how many kWh of electricity produced from the combustion of a specific amount of natural gas releasing a given amount of CO2.  Everything else in the mass balance is lost as heat.

You can read a bit more here, but IPCC values are a very poor determiner of realtime CO2 emissions per kWH.  https://www.engineeringtoolbox.com/co2-emission-fuels-d_1085.html

alixunderplatz:
@HansHyde 
My intention for opening this MD-related issue: I just wanted to point out, that for this rather overseeable country and its few generation units with a poor efficiency (info that I found by accident while searching for data for MD), we might as well change the factor here, like we did for Norwegian hydro or Estonian oil shale.


1 million BTU of pure methane = 293 kWh of methane in derived SI units.
[...]

PR #1380, opened 2018-05-04
Switch unknown to gas in MD

alixunderplatz:
ref #973


systemcatch:
Looks good, do you want to add a link to the source as well?

PR #1384, opened 2018-05-05
switch CA-NB->CA-PE exchange to use PEI data

jarek:
Prince Edward Island has an area of 5660 km2 (roughly comparable to between Fyn and Sjælland), and has an installed wind capacity of around 200 MW. In practice this means that the production can be rather bumpy. Import from New Brunswick is used to provide extra energy when needed, but our data wasn't looking very intuitive:

PEI origin of electricity chart
PEI origin of electricity chart
PEI origin of electricity chart

We were using exchange data from the New Brunswick parser. PEI parser also can compute the exchange. Not to be outdone by Alex, I collected data computed by the two parsers, and compared them.

The PEI wind production and exchange data updates every 10 or 20 minutes. The New Brunswick exchange data updates once an hour. The PEI parser produces a much more intuitive-looking load-curve:

PEI load curve with PEI and NB data

Looking at the exchange data we can see that while the data is ultimately the same, the NB data doesn't update as frequently and thus misses smaller bumps and time-shifts some bigger ones:

PEI exchanges with PEI and NB data

Source CSV: https://gist.github.com/jarek/d8886475b750af2626da0f8ad787616c

(I am also collecting data on New Brunswick's exchanges with other neighbours. Pull requests or writeups, if any arise, will follow.)

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -505,7 +505,7 @@
       46.7349
     ],
     "parsers": {
-      "exchange": "CA_NB.fetch_exchange"
+      "exchange": "CA_PE.fetch_exchange"
     },
     "rotation": 90
   },


corradio:
Great!

PR #1389, opened 2018-05-05
Fix some localization issues

magol:


jbuverud:
Thanks @magol for catching these 🙂

PR #1390, opened 2018-05-06
New ELEXON parser

### LINKED ISSUE #1376: Change data source for UK-GB (5-min data)
I've seen this happen frequently in the last weeks: the wind energy production on ENTSO-E for the UK would drop and leave a very spiky and obviously wrong graph for the UK-GB zone.

image

This issue got much worse in the last three days. Here is what it looked like for 1st of May [on ENTSO-E](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=false&viewType=GRAPH&areaType=BZN&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=01.05.2018+00:00|CET|DAYTIMERANGE&dateTime.endDateTime=01.05.2018+00:00|CET|DAYTIMERANGE&area.values=CTY|GB!BZN|10YGB----------A&productionType.values=B17&productionType.values=B18&productionType.values=B19&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2)):

image

Offshore generation would almost completely disappear and onshore wind would drop by 30-40%.

Other sources, like these dashboards don't seem to be affected by that data quality issue.: 
http://electricinsights.co.uk/#/dashboard?&_k=dxtgvl 
http://gridwatch.co.uk/
http://grid.iamkate.com/

So maybe it's time to switch to data from ELEXON, which have a super short reporting interval of 5 minutes for both generation and all exchanges. Seems like there is a delay of 1 hour:

https://www.bmreports.com/bmrs/?q=generation/fueltype

image

Check out the XML and CSV links on the site that'll lead you to the data.
Historical data available, too!

lorrieq:
closes #1376 

Items that need review:
- datetimes in fetch_production. Currently half-hourly but can be 5 minute intervals once we resolve which datetime each row of data corresponds to (as mentioned in issue)
- allow multiple parsers per zone


parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+        return data_points
+
+
+def parse_production(csv_text, target_datetime=None,
+                     logger=logging.getLogger(__name__)):
+    if not csv_text:
+        return None
+
+    data_points = list()
+    lines = csv_text.split('\n')
```
corradio: Maybe simpler here to usage pandas with `pd.read_csv`. No need to change.

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+
+        data_points.append(data)
+
+    if not target_datetime:
+        return data_points[-1]
+    else:
+        return data_points
+
+
+def datetime_from_date_sp(date, sp):
```
corradio: why not use a parsing datetime format instead of doing string manipulation?

corradio:
This looks pretty neat!
One way to be sure of what we do would be to run both parsers (ENTSOE and ELEXON) in order to compare outputs.
Alternative: Run the elexon parser and compare the outputs with what's currently shown on the electricity map, to double check numbers and directions of flows.

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+        fields = line.split(',')
+
+        # create the datetime
+        datetime = datetime_from_date_sp(fields[1], fields[2])
+
+        data = {
+            'zoneKey': 'GB',
+            'datetime': datetime,
+            'production': {},
+            'storage': {},
```
corradio: can you confirm that positive storage values means using electricity from the grid, and that negative values means giving back (therefore acting as production)?
It's the opposite of the production logic.

lorrieq: I will need to check that. Actually I might have implemented hydro wrong, ELEXON provides values for hydro pumped storage and hydro non-pumped storage. At the moment I have using pumped storage for storage and non-pumped storage for production. I'm not fully sure what the terms mean.

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+            'Expected {} fields in FUELINST report, got {}'.format(
+                EXPECTED_FIELDS['FUELINST'], field_count
+            )
+        )
+
+    for line in lines[1:-1]:
+        fields = line.split(',')
+
+        # create the datetime
+        date_obj = dt.datetime.strptime(fields[3], '%Y%m%d%H%M%S')
```
corradio: Are we sure input time string is always formatted in UTC?

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+        datetime = arrow.get(date_obj).replace(minutes=-5).to('Europe/London')
```
corradio: I'm still a bit puzzled by this. Doesn't the Elexon datetime represent the (start) date of the datapoint?

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+            'Expected {} fields in INTERFUELHH report, got {}'.format(
+                EXPECTED_FIELDS['INTERFUELHH'], field_count
+            )
+        )
+
+    for line in lines[1:-1]:
+        fields = line.split(',')
+
+        # create the datetime
+        datetime = datetime_from_date_sp(fields[1], fields[2])
```
corradio: What does fields[2] represent? Is that an offset from fields[1] in minutes?

parsers/ELEXON.py
corradio: We should put non-pumped hydro in production and pumped hydro in storage, but multiplied by -1. We should double check that pumped hydro can be both positive and negative in the data

alixunderplatz: @lorrieq @corradio UK-GB previously hasn't been reporting to ENTSO-E what is stored into "pumped storage". 
ELEXON also looks like it does not contain this data. The PS-value never goes below 0. So there is nothing that shows the consumption for pumping water to the storage (= hydro storage bar does not go to negative side on the map).

So "pumped storage" reported by ELEXON is the *output* of electricity from discharging pumped storage. Positive values on ELEXON represent the output for the "hydro storage" category.

"Non-pumped storage" means, there either is a "reservoir type" hydro power with a natural inflow of water, which is released when it's needed, or a run-of-river hydro plant, which is using the recent flow of water of rivers.
Non-PS belongs to the usual "hydro" category.

parsers/ELEXON.py
lorrieq: image

Yes, see previous correspondence they sent me on this. All timestamps are in GMT, the only local times are settlement date - settlement period combinations.

parsers/ELEXON.py
lorrieq: fields[2] is the settlement period. So there are 48 settlement periods each day, each is 30 minutes long. The `datetime_from_date_sp` function contains the logic to convert a settlement date and settlement period to a local datetime.

parsers/ELEXON.py
lorrieq: Normally, yes. Note fields[3] is the publish time.

image
Settlement period 21 starts at 09:00GMT. For this settlement period, the first publication datetime is 09:05GMT and the last publication datetime is 09:30GMT. So I presume each publication datetime is referencing a datapoint with start date five minutes before that as the 5 minute observations for period 21 should be [09:00, 09:05, 09:10, 09:15, 09:20, 09:25] GMT. 
That's why I'm subtracting five minutes. Is that clear/correct?

parsers/ELEXON.py
lorrieq: (this conversion does not yet make considerations for the time change when one sees settlement periods 49 and 50)

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+    else:
+        return data_points
+
+
+def datetime_from_date_sp(date, sp):
+    date = dt.datetime.strptime(date, '%Y%m%d').date()
+    sp = int(sp)
+    hour = int((sp - 1) / 2)
+    minute = 0 if sp % 2 == 1 else 30
+    datetime = arrow.get(dt.datetime.combine(date, dt.time(hour, minute)))
```
corradio: once you have your date it might be easier just to use `arrow.get(date).shift(minutes=30*sp)` (assuming `sp` is zero-based).
Also it might be more logical to cast `sp` to int outside of the function.

parsers/ELEXON.py
corradio: Makes sense! Can you add that comment in the code to make it obvious for future maintainers?
Thanks!

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+    params['APIKey'] = os.environ['ELEXON_TOKEN']
+    return session.get(ELEXON_ENDPOINT.format(report), params=params)
+
+
+def query_exchange(session, target_datetime=None):
+    params = {
+        'ServiceType': 'csv'
+    }
+    response = query_ELEXON('INTERFUELHH', session, params, target_datetime)
+    response = response.content.decode()
```
systemcatch: You can do `response.text` to get the same result.

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+            'source': 'bmreports.com'
+        }
+
+        # positive value implies import to GB
+        multiplier = -1 if 'GB' in sorted_zone_keys[0] else 1
+        data['netFlow'] = float(fields[EXCHANGES[exchange]]) * multiplier
+        data_points.append(data)
+
+    if not target_datetime:
+        return data_points[-1]
```
systemcatch: It's fine to return a list of dictionaries, the parser is more robust that way.

parsers/ELEXON.py
```diff
@@ -0,0 +1,202 @@
...
+
+        data = {
+            'zoneKey': 'GB',
+            'datetime': datetime.datetime,
+            'production': {},
+            'storage': {},
+            'source': 'bmreports.com'
+        }
+
+        # read the fuel values into the datapoint
```
systemcatch: This is kinda confusing, is there a better way of mapping the fuels to type?

lorrieq:
Latest commit addresses many of your comments above. Appreciate them!

Sample output of parser (just last datapoint of each function call for readability):
screen shot 2018-05-07 at 22 25 25


**Remaining for this PR at the moment:**

1. Hydro production/storage (I know you've addressed above but haven't implemented yet)

2. Clean up mapping fuels to type

3.  GB-NIR -> IE should be GB -> GB-NIR. I just noticed this. I have mapped the Moyle interconnector to the exchange GB-NIR -> IE when it should be GB -> GB-NIR as the Moyle interconnector goes from Scotland to Northern Ireland. What exactly is the GB-NIR -> IE exchange?

systemcatch:
Looking good @lorrieq.

https://github.com/tmrowco/electricitymap/blob/master/parsers/GB_NIR.py implements the moyle and Irish exchanges already.  Does this Elexon data include generation for Northern Ireland as well?

lorrieq:
@systemcatch Ah, unfortunately the 5 minute data ignores negative values for the exchanges so we're still stuck with [30 minute data](https://www.bmreports.com/bmrs/?q=generation/avghalfhourIC) for the exchanges. So the GB_NIR.py parser is better for the Moyle exchange.

I'm pretty sure the ELEXON data is just for England/Scotland/Wales. Some of the reports can be viewed at a zonal level, zones B1-B17, and judging by [this map](https://test.bmreports.com/bmrs/sites/default/files//System_Zone_map.pdf) that does not include Northern Ireland.

systemcatch:
Well that's good news.  However we have to decide whether to use this parser or stick with ENTSOE.

ELEXON - 5min granularity, no solar, wind/biomass seem low.
ENTSOE - 30min granularity, wind spikes make the graph look bad.

brunolajoie:
We can't trade temporal granularity against solar/wind/biomass reliability. Let's just investigate a bit why Elexon solar/wind/biomass data is incorrect before chosing. There must be a way to get good elexon data. many other UK-centric webistes uses Elexon as data feed (http://grid.iamkate.com/, http://www.mygridgb.co.uk/dashboard/, http://carbonintensity.org.uk/) Can anyone have a look to see how they source solar data from?

systemcatch:
@brunolajoie most websites use https://www.solar.sheffield.ac.uk/pvlive/ for solar data. It's a fairly reliable estimate provided by the University of Sheffield at 30min intervals.

maxbellec:
is the generation mix displayed on Sheffield PVlive the Elexon data where PV is replaced by Sheffield PV live data?

systemcatch:
@maxbellec yes I think it is the 30min data though, not the 5min.

systemcatch:
Ways to move this forward;

1. Don't switch from ENTSOE (:sob: does not solve wind energy spikes, lots of work by @lorrieq not used.)
2. Use 30 min data instead (more reliable, solves wind spikes, suggest new PR for this option).
3. Contact ELEXON about discrepancy between 5min and 30min data.
4. Use this as a 'backup' to ENTSOE data (requires multiple parser per zone support).

brunolajoie:
Would that make sense to use the 5min data for fossil & hydro, and 30 min data for the rest?

systemcatch:
If you guys are happy with mixing granularity then I think that's an excellent solution.

brunolajoie:
If i'm not mistaken, we already have situation like that, where imports are 5min granular while prod is hourly (NL for instance) so I don't see any issue. @corradio you confirm?

maxbellec:
for different types of data (production, export, price), there's no problem mixing granularity. But currently it's not possible to have different productions fuels come from different parsers / at different granularities.

systemcatch:
@corradio @brunolajoie @maxbellec 
Lets make a decision on what to do here.

Personally I think we should use the the 30min data for now and get in touch with ELEXON about the disparity.

corradio:
@systemcatch if I understood correctly the 5min data is not usable because of interconnector issues, no solar and incorrect wind (correct me @lorrieq if I'm wrong).
The 30min ELEXON data however presents some discrepancy with the ENTSOE data.
I see conflicting statements about wind spikes (@systemcatch is it the 30min ELEXON data that presents this anomaly?)

If we want to be on the safe side, I suggest using the same methodology as:
- http://www.mygridgb.co.uk/dashboard/ (uses Elexon Portal + Sheffield University) - or -
- http://electricinsights.co.uk/ (read this: http://electricinsights.co.uk/#/reports/methodology?_k=icvu2n and https://www.sciencedirect.com/science/article/pii/S0301421516307017)

Electric Insights is produced and maintained by very knowledgeable people, so I'd probably use the same methodology as they used.
Unfortunately I don't have time to dig beyond the two links that I gave, but I think this should unblock us.

systemcatch:
@corradio 
Based on my reading of everything said so far.

- Only the ENTSOE data shows wind spikes.
- The 5min data does not include any solar, and wind and biomass seem low.
- The interconnectors in the 5min data do not show negative values.
- Solar data is only available at 30min intervals.
- 30min data at https://www.bmreports.com/bmrs/?q=actgenration/actualaggregated has no other issues.
- We cannot mix granularities.

Conclusion - we use the 30min data.

corradio:
Perfect! Just to make sure we're all aligned: what solar data do we use at 30min? Sheffield? National Grid? Elexon?

brunolajoie:
Very clear!

maxbellec:
thanks for the summary @systemcatch, I agree

lorrieq:
So to confirm, I must update the parser to:

- get production from https://www.bmreports.com/bmrs/?q=actgenration/actualaggregated
- get exchanges from https://www.bmreports.com/bmrs/?q=generation/avghalfhourIC

systemcatch:
@lorrieq exactly!

lorrieq:
Hi folks,

Sorry about the delay. I hope (and think!) that we are done now. Sample output file is [here](https://github.com/tmrowco/electricitymap/files/2063697/elexon_sample_output.txt).

If you can review some things I guessed from this [report](https://www.bmreports.com/bmrs/?q=actgenration/actualaggregated) (simple stuff for you guys I'm sure):
Fuel **Hydro Run-of-river and poundage** going into **production->hydro**
Fuel **Hydro Pumped Storage** going into **storage -> hydro**
Fuel **Wind Onshore** and **Wind Offshore** being summed into **production -> wind**
Fuel **Other** going into **production -> unknown**

I didn't update the config files yet to use this new parse because of fear for merge conflicts since this PR is quite old.

systemcatch:
Good effort!!!!

Those fuel type classifications are all fine. Sample output file is ok though the 'datetime' key is a pandas Timestamp object, normally we want a datetime object there but Timestamp is a subclass of datetime anyway so it might be ok, @corradio thoughts?  The conversion can be done with the `to_pydatetime()` method if needed.

I just tried rebasing your branch (on my machine) and it completed with no errors so I think you'll avoid any conflicts.

maxbellec:
this is great! I agree with @systemcatch, if you could only convert your datetime and update the config files we'll be done. (I merged master from the website, so just `git pull` before doing your changes)

parsers/ELEXON.py
```diff
@@ -0,0 +1,239 @@
...
+def query_production(session, target_datetime=None):
+    if target_datetime is None:
+        target_datetime = dt.date.today()
+
+    params = {
+        'SettlementDate': target_datetime.strftime('%Y-%m-%d'),
+        'Period': '*',
+        'ServiceType': 'csv'
+    }
+    response = query_ELEXON('B1620', session, params)
```
corradio: @maxbellec it seems here that ELEXON sends data back for D to D+1.
I think this is in contradiction with others APIs, where data is fetch *up to* `target_datetime`.
Will this be a problem?

corradio:
This is terrific @lorrieq ! I'm usually not very good in catching up with all that happened, but I'm quite good at being annoying with my questions. So I suggest the following: I'll put here a checklist of things I believe should be ok before we merge. Feel free to check each of them off if you're sure it's ok, and when all are checked, we can merge. I'll approve the PR anyway as the code looks good for:

- [ ] Check that storage is stored as a positive value when storing and negative value when discharging
- [ ] Check that ENTSOE *production* data and ELEXON data are approximately the same, for a given timestamp
- [ ] Check that ENTSOE *exchange* data and ELEXON data are approximately the same (at least they should have the same sign), for a given timestamp

@maxbellec this probably means we should a TODO on our side to refetch old data and purge our database of old ENTSOE data

maxbellec:
good points @corradio.
About the `target_datetime`, though this is not very well documented, a parser should return as much data as possible (in a single API call) up to `target_datetime`. Returning some data after `target_datetime` is not a problem, but if all data is after `target_datetime`, this is not the expected behaviour.
It seems the production data is available as a CSV for a full day. If the target datetime corresponds to midnight, the parser should fetch the day before. Else, it should fetch the requested day.

parsers/ELEXON.py
maxbellec: indeed, if the target datetime corresponds to midnight, the parser should fetch the day before. Else, it should fetch the requested day.

maxbellec:
that's great @lorrieq . The parsers are run ~every 10 minutes. So between 00:00 and 00:30 it'd fetch the day before, and the rest of the time it would continuously fetch current day, adding new entries as they arrive. It works great.

lorrieq:
Hey @corradio, here's a [spreadsheet](https://github.com/tmrowco/electricitymap/files/2087285/ENTSOE_vs_ELEXON.xlsx) showing ENTSOE and ELEXON production/exchange data for comparison. 

Production is a perfect match except ELEXON hydro storage sign is flipped. Also exchange data is hourly for ENTSOE and 30 minute for ELEXON but values are reasonably similar.

corradio:
@lorrieq is that the result of the output of our parsers (ENTSOE.py and ELEXON.py)? If the storage sign if flipped it might indicate something is wrong. Happy to see that the rest looks ok

lorrieq:
Yes, its the output of the parsers. So fixing the ELEXON hydro storage is all that's left to do I think

corradio:
Sounds great!

lorrieq:
Hydro storage fixed, hopefully thats it

corradio:
Sounds terrific. I suggest we put into production asap. As soon as I have a bit of time I'll merge and do that.

PR #1394, opened 2018-05-07
Enable fetching of historical data for CL-SING

systemcatch:
This allows us to get past days generation, which is useful because the real time data only provides a breakdown of (solar, wind, unknown).

ref #655

### DIFF
--- a/parsers/CL_SING.py
+++ b/parsers/CL_SING.py
@@ -5,11 +5,13 @@
 # https://www.engerati.com/article/chile%E2%80%99s-grid-be-unified-new-transmission-highway
 
 from __future__ import print_function
+import arrow
 from bs4 import BeautifulSoup
 from collections import defaultdict
 import datetime
 import json
 from pytz import timezone
+import pandas as pd
 import re
 import requests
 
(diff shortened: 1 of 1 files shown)


alixunderplatz:
@systemcatch You've mapped "cogeneracion" to biomass. 
Is this based on "PLANTA DE ÁCIDO SULFÚRICO MEJILLONES" from the table on the data website?

Really interesting: the power plant seems to be burning pure sulphur to create SO2 to create H2SO4. http://www.noracid.cl/
No idea where to put this, biomass seems a bit odd, too, but could be the right choice to account for LCA emissions^^ Maybe we should add a comment to the code?
_ _ _ _
I really liked the category "Fuel Oil Nro. 6"

"Fuel Oil Nro. 6" - new fragrance by C.L. Sing 👃😸

systemcatch:
@alixunderplatz Actually no I thought it meant they were burning crops or waste. The plant makes sulphuric acid and recovers heat/energy for electricity from the looks of it.  No idea how carbon intense the process is but we should add a comment as it's unusual.

:nose:  Price: you can't afford it.

systemcatch:
Shall we set the sulphuric acid plant to unknown since it's a really unusual generation type?

alixunderplatz:
I think biomass is totally okay ;) This will account for fuel (sulphur) mining, [delivery by ship](http://www.noracid.cl/productos/acido-sulfurico/), construction and operation.
Final output is sulphuric acid, seems like there are some minor emissions of acid mist from the values given [here](http://www.sulphuric-acid.com/sulphuric-acid-on-the-web/acid%20plants/NorAcid%20-%20Mejillones.htm), and thats it. SO2 êven has a cooling effect on the climate ^^

corradio:
@alixunderplatz do you approve then?

corradio:
Does this mean we can update #1257?

PR #1395, opened 2018-05-09
split NO

maxbellec:


maxbellec:
screenshot at 2018-05-09 16-08-30

corradio:
Test failing:
```
======================================================================
FAIL: test_bounding_boxes (test_zones_json.ZonesJsonTestcase)
----------------------------------------------------------------------
  File "/home/circleci/electricitymap/parsers/test/test_zones_json.py", line 14, in test_bounding_boxes
    self.assertLess(bbox[0][0], bbox[1][0])
AssertionError: 199.5 not less than -199.5
```

config/exchanges.json
```diff
@@ -1415,32 +1396,12 @@
...
   "NL->NO-NO2": {
     "parsers": {
       "exchangeForecast": "ENTSOE.fetch_exchange_forecast",
       "exchange": "statnett.fetch_exchange"
     }
   },
-  "NO->SE": {
-    "parsers": {
-      "exchange": "statnett.fetch_exchange",
-      "exchangeForecast": "ENTSOE.fetch_exchange_forecast"
```
corradio: @maxbellec we still want to keep the exchange forecast here. Maybe ENTSOE has forecasts on BZ level?

jarek:
I'll mention it here so that it doesn't get lost: the split ended up resetting Norway's hydro carbon intensity to the default 24 gCO2eq/kWh (because the production is now in e.g. NO-NO2 rather than NO)

PR #1401, opened 2018-05-11
Zone details no data update

jbuverud:
- Removed exchange information from zones in times when there is no production present

- Added "No data" overlay on electricity source graph and zone gauges when no data is present (disappears when hovering over a history time which has data associated)
- Added message to inform when live data is unavailable for a zone (does not disappear when hovering on history, as it only relates to the present data)

image

image

- Added a more descriptive message and removed data displays for zones with no parsers
- Made the social buttons (facebook/twitter not showing in my screenshots for some reason, but will show on production) stay at the bottom of the left panel, even when panel is mostly empty (for consistency)

image


jbuverud:
> Removed exchange information from zones in times when there is no production present
why is that? Exchange can still be interesting when there is no production, don't you think?

Because @brunolajoie said so when I asked in the slack channel. Maybe we misunderstood each other. What do you think @corradio  - should it be changed back?

> The red background might be a bit harsh. Maybe use a gray one?

The red background is the same color that we use for the drop-down warning when the map loses the connection to the server. I thought these two things were in the same category, which is why I chose that color. I'm afraid that putting it in grey will not make it visible enough. However, if we e.g. let the the bootstrap color conventions guide us (https://getbootstrap.com/docs/4.0/components/alerts/) we could say that it is now in the "danger" class, but that it should maybe be downgraded to the "warning" class (corresponding to some shade of yellow instead of red)? What do you think? Choosing a color is a bit more tricky in our case though, since we use light text/dark backgrounds.

> I'd remove the "no data" use case as I personally don't really like the "no data boxes" (I prefer having question marks). This is especially relevant in the case where we have partial data.

The "no data" overlays are what we agreed on in slack, no? The overlay makes it much easier to navigate the histories, in my opinion. When scrolling through the history it can be hard to see when a country has a very little amount of production and when it has none (two cases that are very different). Distinguishing between "a little data" and "no data" can be quite annoying actually - you have to visually search through the whole chart and see if there actually is something there. Having the overlay makes it much easier for the users to absorb the "complete set of information" in the graph in the case where there is no data. Let me know what you think.

> Text: I'd replace "You can contribute by adding area data" by "You can contribute by adding a live data source about this area".

Will do 🙂

corradio:
Agree with "alert" class. I think you and @brunolajoie misunderstood each other: exchanges should be visible even if there's no production data (some zones sometimes have exclusively exchanges and shut down all production).
For the rest I think we should discuss that IRL or on screensharing.
Let's do that asap?

jbuverud:
Here's a new design following our conversation today @corradio. 

image
image


I ended up placing the data message box in the middle of the overlay instead of on top. Putting it on top leads to the page growing in size when inserting it, meaning the history charts move, which results in the chart popping back and forth due to a loop of mouseleave/mouseenter-events when the mouse is held in one place. We could include permanent spacing between the zone title and the gauge area, so that the message could pop in and out without issues, but I think this doesn't look very good. I think putting it overlaid is a good compromise.

Also I ended up with a dirty-ish yellow for the "warning" background color.

Let me know how you feel about the changes 🙂

web/locales/en.json
```diff
@@ -38,7 +38,10 @@
         "bysource": "by source",
         "emissions": "Carbon emissions",
         "addeditsource": "<a href=\"%s\" target=\"_blank\">add or edit source</a>",
-        "helpfrom": "with help from"
+        "helpfrom": "with help from",
+        "nodata": "No data",
```
corradio: Might require some clean up

web/src/reducers/dataReducer.js
```diff
@@ -21,7 +21,7 @@ Object.entries(zonesConfig).forEach((d) => {
   zone.contributors = zoneConfig.contributors;
   zone.timezone = zoneConfig.timezone;
   zone.shortname = translation.getFullZoneName(key);
-  zone.hasParser = zoneConfig.parsers && zoneConfig.parsers.production !== undefined;
+  zone.hasParser = zoneConfig.parsers !== undefined && zoneConfig.parsers.production !== undefined;
```
corradio: Suggestion: `(zoneConfig.parsers || {}).production !== undefined`
No strong opinion.

web/src/reducers/dataReducer.js
```diff
@@ -123,7 +123,11 @@ module.exports = (state = initialDataState, action) => {
         });
         if (!zone.exchange || !Object.keys(zone.exchange).length) {
           console.warn(`${key} is missing exchanges`);
+        } 
+        else if (!Object.keys(zone.production).length) {
+          zone.exchange = {};
```
corradio: a comment would be welcome to understand why exchanges are changed due to empty production

web/src/reducers/dataReducer.js
```diff
@@ -150,13 +154,19 @@ module.exports = (state = initialDataState, action) => {
       // Create new histories
       const newHistories = Object.assign({}, state.histories);
 
-      // Add 'hasParser' flag to all zone history observations
       const zoneHistory = action.payload.map((observation) => {
         const ret = Object.assign({}, observation);
+        
         ret.hasParser = true;
+        if (observation.exchange && Object.keys(observation.exchange).length
```
corradio: Same here

web/src/styles.css
```diff
@@ -1246,6 +1258,66 @@ sub {
     padding-top: 0.5rem;
 }
 
+.no-data-overlay {
```
corradio: Is this still relevant now that the "no data" text is gone?

PR #1404, opened 2018-05-12
more validations for ENTSOE data

jarek:
Whole categories of generation have been dropping off ENTSOE quite often. Quite often several types of generation will drop from Germany, it's particularly noticeable with biomass and hydro which are otherwise constant. Recently I saw Poland and Czechia with no coal. This enhances the tests a bit to catch more of these problems, and makes it a bit cleaner to add further validations.


systemcatch:
Great work @jarek.

PR #1184, opened 2018-03-07
Localize Database and Forecast

magol:

PR #1406, opened 2018-05-12
PL translation for new strings

jarek:
Disclaimer! My Polish is a bit rusty for sentences, and I don't have much technical vocabulary. I spent a few hours checking translations on the internet, so I hope it's alright, but there might be errors. Is anyone else able to review? :poland: Czy ktoś mógłby spojrzeć i podzielić się swoim zdaniem?

A few questions:
- onboarding modal: should "The Electricity Map" to be translated? In Swedish it was.. but it's a product name? I just removed the "The" and kept "Electricity Map"
- In Database and API referral strings, should "Database" and "API" be translated? Swedish and Russian translated it, Italian did not (following #1184). I also went with "Electricity Map Database" interpreting the domain name, since the "our" doesn't translate cleanly to my ears...

Notes:
- "Map", "Areas", "About" at bottom of mobile view seems to not be localized at all
- `onboarding-modal view2 header` is the only onboarding header in English to have a full stop at the end. I didn't include it in Polish to keep consistent with the other onboarding headers
- I removed "cross-border" from "Cross-border export" since the word "transgraniczny" is a bit unelegant to me (mix of Latin prefix and Polish word); IMO it's not really confusing what kind of export it is; and "cross-border" is increasingly iffy with splitting more regions
- There was a left-over "Twoją" which I lowercased to "twoją". This is a semi-honorific rendering of the informal form of T-V distinctions, like capitalizing "Du" in German. Opinions on its necessity vary, but on the internet it seems a bit much to me.

If I may suggest - when you add a new string, if you have the time, also add it to another language you know, like French or Danish. That will make it much easier to triangulate the exact meaning :)

If any Polish people see this, kwestie, których nie jestem pewien:
- "region" czy może lepiej "obszar"?
- "eksport" czy może "wymiana"?
- "Twój" czy "twój"?
- "energia elektryczna", czy można czasem skrócić "prąd"?

### DIFF
--- a/web/locales/pl.json
+++ b/web/locales/pl.json
@@ -9,7 +9,7 @@
     "mobile-info-tab": {
         "map-info-text": "Mapa pokazuje, ile CO2 jest aktualnie emitowanie przy produkcji twojej energii elektrycznej. Regiony są kolorowane według intensywności emisji CO2, wliczając potencjał tworzenia efektu cieplarnianego, jednostkowo według zużywanej tam energii elektrycznej. Zieleńsze odcienie oznaczają bardziej przyjazną dla klimatu energię elektryczną.",
         "electricity-origins-text": "Strzałki wykazują, gdzie i ile energii płynie pomiedzy regionami. Naciśnij region aby dowiedzieć się więcej o źródłach energii elektrycznej tam zużywanej.",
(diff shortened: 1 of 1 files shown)


corradio:
@jbuverud agreed we shouldn't translate the brand / product names.

PR #1407, opened 2018-05-13
Consistent usage of "du" and "dein" in german locale

jld3103:


jarek:
I'll merge it now, sorry for the delay and all the master branch merges :) Thanks for the cleanup!

PR #1416, opened 2018-05-15
Re-added exchange information for last data point for zones without p…

### LINKED ISSUE #1403: Exchanges referencing "US" zone no longer appear on map
Change #1366 added a new zone which covers Washington State in the US under the Bonneville Power Administration, and I think since then the CA-BC->US exchange that used to show up on the map no longer does.

It looks like [config/exchanges.json](https://github.com/tmrowco/electricitymap/blob/master/config/exchanges.json) includes a few references to the "US" zone which no longer exists anywhere on the map, so that's probably why. Maybe.

Namely: 

- CA-BC->US  (should be changed to US-BPA?)
- CA-AB->US (should be changed to US-MT?)
- US->US-CA (Not sure about this one, the CSV imported in the parser contains only aggregate imports, not broken down by connection)

jbuverud:
…roduction (used in map exchanges). Instead removed this information directly in countrytable.js so it does not show in the consumption mix bar chart.

Fixes #1403

### DIFF
--- a/web/src/components/countrytable.js
+++ b/web/src/components/countrytable.js
@@ -466,6 +466,10 @@ CountryTable.prototype.data = function(arg) {
     this._data = arg;
 
     if (!this._data) { return this }
+
+    this.hasProductionData = this._data.production !== undefined && Object.keys(this._data.production).length > 0;
+    this.isMissingParser = this._data.hasParser === undefined || !this._data.hasParser;
+    
     if (this._exchangeKeys) {
         this._exchangeData = this._exchangeKeys
             .map(function(k) {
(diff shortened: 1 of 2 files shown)


maxbellec:
can you post a screenshot of the fix?

jbuverud:
A screenshot of what, exactly? Everything looks as before, except that exchange arrows now show on the map even when the originating country has no production data (e.g. between canada and US-BPA).

maxbellec:
Yes I meant just a screenshot with the arrows so that we can visually check that your PR works in US and NA, CA... Of course, you've checked this before submitting, but since there's not a lot of change in the code, and I'm lazy, I prefer to just check a screenshot than pull the branch and check locally

PR #1417, opened 2018-05-15
Split USA by balancing authorities

systemcatch:
There are many new zones being added with these shapes, I've tried to pick sensible names for them but some might need changing.

As you can see from the gif there's problems on highlighting/hover with zones overlapping everywhere.  Frontend expertise needed! :rofl: 

ref #143 

new_us


maxbellec:
great gif! Overlapping zones is a pain... We can probably start by removing all zones for which we have no data yet?

systemcatch:
Hey @maxbellec I'm not entirely sure this is what you want but in the gif below I've added back in some of the state shapes so that highlighting works better. The new zones are still there but covered by the states.

us_by states

I also tried laying the new shapes for areas we have data onto the continental US subunit. Looks quite good except for the Great Lakes being filled in (subunits ignore lakes?).

us_continental

systemcatch:
Hey @jarek I agree totally with your suggestions. The only problem with using the USB subunit is that it fills in the Great Lakes and leaves a weird looking grey strip in place.

BPA's area is pretty big, I'm not sure that includes only the generation from the parser.

corradio:
I'm not sure exactly how to move on from here.
We're definitely at a crossroad where we need to decide between accuracy and visual appearance.
I'd go mostly for visual appearance as long as the approximation mistakes are not huge.
Also going for a state approach might be good (when possible), as this might push people to ask their federal gov to change things. If we go to a grid level, then maybe it will feel a bit more impersonal. On the other hand we started splitting Denmark into East and West, so who knows.

If we go with a better subdivision, I however think we would need to remove most of the holes in the geometries you're showing @systemcatch. From a UI perspective it looks quirky (although I know it probably reflects reality better).

What do other people feel about this?

brunolajoie:
Mmm, tricky. I've build your PR locally and here's what i think:
I'd say its quite relevant to change US geometries by balancing authorities. 
It's gonna give us the credibility we currently miss in the US (compared to Europe).
It's been done [here](http://explorer.watttime.org/) and it does not look impossibly messy. 
image

However, the current PR is not yet ready to be merged, cause the geometry is much more messier than above. There's still lots of areas that are counted twice. 
image
 
So all in all, i'd say we should go for it, but only once we've a very clear and MECE (mutually exclusive and collectively exhaustive) division of the US

systemcatch:
Part of the problem is that the MISO and SWPP areas (also IPC and BPA to a lesser exent) overlap with these shapes.  I'm not really sure how that can be fixed at the moment.

systemcatch:
Closing this for now as the branch is really out of date and I can't make this work. The shapes are linked in #143 so anyone can have a go at getting this to work. Realistically we need someone who is far better with qgis than me to get the shapes to line up properly (overlaps might be impossible to resolve).

Watttime's shapes are very good, I think they are from another source though. That's the kind of result we need to be aiming for.

maxbellec:
I agree, need to clean up the files by hand to make non-overlapping ones (though that's not the reality), or find a better source.

PR #1432, opened 2018-05-30
No onboarding when embedded

### LINKED ISSUE #1425: Bypass onboarding when map is embedded

maxbellec:
fixes #1425 
Somehow the `getState().isEmbedded` was undefined when checking for it in main.js:213. Not sure why, but it's just as easy (though not as elegant) to directly check `window.top === window.self`

I've checked that
- I don't get onboarding when embedded and I have seen it
- I don't get onboarding when NOT embedded and I have seen it
- I don't get onboarding when embedded and I have NOT seen it
- I DO get onboarding when NOT embedded and I have NOT seen it

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -209,7 +209,8 @@ moment.locale(getState().application.locale.toLowerCase());
 // Analytics
 thirdPartyServices.trackWithCurrentApplicationState('Visit');
 
-if (!getState().application.onboardingSeen && !getState().isEmbedded) {
+// do not display onboarding when we've seen it or we're embedded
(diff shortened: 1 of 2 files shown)


web/src/main.js
```diff
@@ -209,7 +209,8 @@ moment.locale(getState().application.locale.toLowerCase());
 // Analytics
 thirdPartyServices.trackWithCurrentApplicationState('Visit');
 
-if (!getState().application.onboardingSeen && !getState().isEmbedded) {
```
corradio: needs to be `!getState().application.isEmbedded`!

Issue #1405, opened 2018-05-12
Data for Southwest USA (SVERI)


systemcatch:
There is generation by fuel type at https://sveri.energy.arizona.edu/#generation-by-fuel-type plus many other measurements. Looks like 1min data with a very good breakdown!

SVERI = **Southwest Variable Energy Resource Initiative**

From the site.
>The SVERI participants include Arizona Electric Power Cooperative, Arizona Public Service, El Paso >Electric, Imperial Irrigation District, Public Service Company of New Mexico, Salt River Project, Tucson >Electric Power and the Western Area Power Administration’s Desert Southwest Region.

Practically this covers parts of Arizona, Texas, California and New Mexico. But as far as I can tell there is no overlap in data with anything we get currently.

alixunderplatz:
Amazing find!

Just leaving the link to instructions for using the **API** here:
https://sveri.energy.arizona.edu/#howto

Looks like their large scale [solar and wind assets](https://sveri.energy.arizona.edu/#renewmap) are in **Arizona** and **New Mexico**, except for one smaller PV array (23 MW) which is located in California.

![image](https://user-images.githubusercontent.com/25743609/39965846-229c9c86-56a1-11e8-965d-e5cbfee0382e.png)

@systemcatch You're right, nothing overlaps. According to the [EIA map](https://www.eia.gov/realtime_grid/?src=home-undefined#/status?end=20180513T04) the area in Texas is rather small. Seems like the southern tip of Nevada (including Las Vegas) is within the service area, too, if these are the correct borders:

![image](https://user-images.githubusercontent.com/25743609/39966540-d107f35e-56ad-11e8-92cf-fea6206f8188.png)

Based on the EIA map, it looks like we could also implement exchanges of the Southwest with CAISO and SPP - at least for dates in the past.

jarek:
Warning, that EIA map is for Southwest balancing area, not SVERI area! Las Vegas is serviced by Nevada Power Co which is not part of that SVERI list above. Meanwhile Imperial Irrigation District is a rural part of southern California which is not actually serviced by CAISO.

Arizona Electric Power Cooperative = small area in southeast Arizona, I think?
Arizona Public Service = most of Arizona except for Phoenix and Tucson
El Paso Electric = western corner of Texas under New Mexico, and a bit of southern New Mexico
Imperial Irrigation District = rural southeastern California
Public Service Company of New Mexico = most of New Mexico, including Santa Fe and Albuquerque
[...]

PR #1435, opened 2018-06-03
Add Southwest Variable Energy Resource Initiative (SVERI)

systemcatch:
Geothermal is currently disabled because it's returning negative values.  I've contacted the site owner about the problem so hopefully he responds soon.

The area is set as just Arizona at the moment, @alixunderplatz @jarek what do you think about that?  No capacities found so far.

ref #1405 

sveri


parsers/US_SVERI.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if target_datetime is not None:
+        start_date = target_datetime.strftime('%Y-%m-%d')
+        shift_date = target_datetime + timedelta(days=1)
+        end_date = shift_date.strftime('%Y-%m-%d')
+        limits = (start_date, end_date)
+        raw_data = query_api(limits, session=session)
+    else:
+        today = datetime.now()
+        shift_date = today + timedelta(days=1)
```
maxbellec: some duplicated code here: if / else should only concert `start_date`. in both cases, `shift_date` should be start_date + 1 day, and the limts `(start_date, end_date)`

maxbellec:
:+1:

PR #1436, opened 2018-06-05
Test parsers

maxbellec:
The goal is to make testing parsers as easy as possible, so that it's easy to check when a country is down if the issue comes from the parser, the source or Tmrow's backend.
And I was sick of having to think / search where parsers are (ENTSOE / statnett / ...)

*from the electricitymap folder*, run a command looking like
- `python test_parser.py FR price` see output of the price parser for FR
- `python test_parser.py FR` defaults to production if no data type is given
- `python test_parser.py FR`
- `python test_parser.py DE --target_datetime 2018-04-17T04:00` to test past dates


maxbellec:
example output:

```python
python parsers/test_parser.py GB
[2018-06-05T18:47:25.511188+02:00] querying ENTSOE
parser result:
[{'zoneKey': 'GB', 'datetime': datetime.datetime(2018, 6, 3, 16, 0, tzinfo=tzutc()), 'production': {'biomass': 1996.0, 'coal': 0.0, 'gas': 12636.0, 'hydro': 223.0, 'nuclear': 7460.0, 'oil': 0.0, 'solar': 3990.0, 'wind': 900.0, 'geothermal': None, 'unknown': 127.0}, 'storage': {'hydro': -517.0}, 'source': 'entsoe.eu'}, ... , {'zoneKey': 'GB', 'datetime': datetime.datetime(2018, 6, 5, 15, 0, tzinfo=tzutc()), 'production': {'biomass': 2048.0, 'coal': 645.0, 'gas': 15869.0, 'hydro': 127.0, 'nuclear': 7697.0, 'oil': 0.0, 'solar': 4870.0, 'wind': 2718.0, 'geothermal': None, 'unknown': 127.0}, 'storage': {'hydro': -314.0}, 'source': 'entsoe.eu'}]
---------------------
took 12.35s
min returned datetime: 2018-06-03 16:00:00+00:00
max returned datetime: 2018-06-05T15:00:00+00:00 UTC  -- OK, <2h from now :) (now=2018-06-05T16:47:37.864948+00:00 UTC)
```

corradio:
Linter fails with

```
************* Module electricitymap.parsers.test_parser
E:  4, 0: Unable to import 'click' (import-error)
E: 46,10: No value for argument 'zone' in function call (no-value-for-parameter)
E: 46,10: No value for argument 'data_type' in function call (no-value-for-parameter)
E: 46,10: No value for argument 'target_datetime' in function call (no-value-for-parameter)
```

Feel free to update dockerfile to python:3.7 in the process.

maxbellec:
whoops, I had added `click` on master

parsers/test_parser.py
```diff
@@ -0,0 +1,46 @@
...
+                     '---------------------',
+                     'took {:.2f}s'.format(elapsed_time),
+                     'min returned datetime: {}'.format(min(dts)),
+                     'max returned datetime: {} UTC {}'.format(
+                         last_dt, max_dt_warning), ]))
+
+
+
+if __name__ == '__main__':
+    print(test_parser())
```
systemcatch: The build is failing because test_parser() is missing parameters here.

parsers/test_parser.py
```diff
@@ -0,0 +1,46 @@
+import time
+
+import arrow
+import click
+
+try:
+    from parsers.read_parser_config import parser_key_to_dict
+except Exception:
```
systemcatch: Maybe ImportError instead?

parsers/test_parser.py
```diff
@@ -0,0 +1,46 @@
...
+    raise ValueError("couldn't do a local import. Please run from the"
+                     "`electricitymap` folder: python parsers/test_parser.py "
+                     "[OPTIONS] ZONE [DATA_TYPE]")
+
+
+@click.command()
+@click.argument('zone')
+@click.argument('data-type', default='production')
+@click.option('--target_datetime', default=None, show_default=True)
```
systemcatch: Very nice I've not used click before but this is great.

parsers/read_parser_config.py
```diff
@@ -0,0 +1,75 @@
...
+    'productionPerUnit': PRODUCTION_PER_UNIT_PARSERS,
+    'exchange': EXCHANGE_PARSERS,
+    'price': PRICE_PARSERS,
+    'consumptionForecast': CONSUMPTION_FORECAST_PARSERS,
+    'generationForecast': GENERATION_FORECAST_PARSERS,
+    'exchangeForecast': EXCHANGE_FORECAST_PARSERS
+}
+# Read all zones
+for zone_id, zone_config in zones_config.items():
+    for parser_key, v in zone_config.get('parsers', {}).items():
```
systemcatch: I remember @jarek saying something about using the inspect module to get arguments from a function? Maybe that's something entirely different to here though.

parsers/test_parser.py
```diff
@@ -0,0 +1,46 @@
...
+def test_parser(zone, data_type, target_datetime):
+    if target_datetime:
+        target_datetime = arrow.get(target_datetime).datetime
```
systemcatch: Isn't target_datetime a python datetime already?

maxbellec:
it's actually a linter bug, it doesn't understand that the `click` decorators modify the function and that calling the decorated version without arguments is actually fine. I've moved test_parser to the root folder so that it's not linted (and more easily accessible)

test_parser.py
```diff
@@ -0,0 +1,46 @@
+import time
+
+import arrow
+import click
+
+try:
+    from parsers.read_parser_config import parser_key_to_dict
+except Exception:
```
systemcatch: ImportError is more specific.

parsers/read_parser_config.py
maxbellec: `inspect` is indeed great to know about function arguments but I'm not sure I see the link here. We're just reading a config json file

parsers/test_parser.py
maxbellec: :+1:

parsers/read_parser_config.py
systemcatch: Yeah I was thinking of something different, ignore what I said.

parsers/test_parser.py
maxbellec: we parse `target_datetime`, which is a string that we got from the command line (using click), using the `arrow` library, then convert the `arrow` object to a `datetime`

parsers/test_parser.py
maxbellec: see comment in discussion

test_parser.py
maxbellec: :+1:

maxbellec:
Regarding multiple zone, I've made this single purpose tool to make it easier to test a single zone that's gone dark.
I'm not sure I'd want to complicate it a bit to add more features such as multi-zone test (it's easy to run the command a few times with different zones).
We use Kibana to see global errors.

test_parser.py
systemcatch: I meant in the except clause, raising the ValueError is good. e.g.
```python
 try:
     from parsers.read_parser_config import parser_key_to_dict
 except ImportError:
# raise your ValueError here
```

PR #1440, opened 2018-06-07
update es.json

onmax:


corradio:
Hi @onmax, and thanks for the contribution. Maybe #1441 can be closed and we use this PR instead?

onmax:
@corradio Done

Issue #1430, opened 2018-05-28
NO4 remove Melkøya power plant (auto-consumption)


brunolajoie:
In NO4, our ENTSOE parser for real time generation per category seems to includes a gas fired power plant (Melkøya) that is not allowed to deliver to the grid. It is used for self-consumption in an LNG factory up there. 

We should try to find real-time power production from this plant in the Transparency platform, and net it from our current NO4 parser.

corradio:
Maybe this is something that needs to be told to entsoe so they update it
on their end?

brunolajoie:
Entsoe contacted

corradio:
@brunolajoie any news?

brunolajoie:
- ENTSOE answered me that Melkøya is supposed to be reported on the TP as it is right now for their computation of the total load/generation
- If our source saying that  Melkøya is disconnected is right, we should remove it from the grid mix ourself
- however the TP does not report individual outpur from Melkoya

I'll discuss it with my source

Gazer75:
Last I heard the plant at Melkøya is required to deliver power through a 132kV cable in emergency situations.
Statnett is currently extending the 400kV grid from Balsfjord to Alta(2021) and Skaidi(2023).
After that the chance of it delivering to the grid is pretty much 0.
A 400kV from Skaidi to Melkøya is also possible as Equinor indicates the need for an additional 270MW.

Research indicated 229MW capacity from five turbines.

The only gas power production for NO4 is from this plant afaik.

Power grid map with data from OpenStreetMap can be viewed at: https://openinframap.org

Gazer75:
The plant has been out of commission for months after a fire broke out during restart from maintenance.
My understanding is that it will not deliver power to the main grid under normal operation as it is there to power the refinery on the island.
Melkøya only has a single 132kV connection currently and that is nowhere near enough to run the entire operation.

It is kind of the same with the Mongstad plant that is hopefully shutting down its last turbine this year after several extensions.

Both of these can deliver to the grid if needed.

The offshore platforms have lots of gas turbines that have no production data available. The numbers for Norway would look a lot worse if we could include emissions from these.

[...]

PR #1446, opened 2018-06-09
Split Russia(#757)

tmslaine:
Splits Russia(RU) to European/Ural zone = zone 1 (RU-1) and Siberian zone = zone 2 (RU-2)


alixunderplatz:
Happy to see this split happening! :) 

We currently use a custom carbon intensity of 567 g/kWh for the unknown share based on generation shares of coal and gas in 2015 for Russia as a whole, which perhaps is not used for the seperated zones after splitting.

Based on the [map on resourcewatch](https://resourcewatch.org/data/explore?zoom=3&lat=0&lng=0&basemap=dark&labels=light&layers=%255B%257B%2522dataset%2522%253A%2522a86d906d-9862-4783-9e30-cdb68cd808b8%2522%252C%2522opacity%2522%253A1%252C%2522visible%2522%253Atrue%252C%2522layer%2522%253A%25222a694289-fec9-4bfe-a6d2-56c3864ec349%2522%257D%255D&page=1&sort=most-viewed&sortDirection=-1), price zone 2 seems to have thermal generation from coal exclusively, while in price zone 2 it's mainly gas power plants that make up P_TES and P_GRES.

We could try to find some figures for the zones, but these figures will be hard to find.
Another option is changing the type mapping for zone 1 and zone 2.

By the way: "BS" stands for "block stanzija", and refers to industrial auto-consumers with their own Co-generation facilities. Therefore, we could in general shift this category to gas, because that is very likely for such smaller and distributed units.

Based on these thoughts, the mapping for zone 2 could look like this:
```
MAP_GENERATION_2 = {
    'P_GES': 'hydro',
    'P_GRES': 'coal',
    'P_TES': 'coal',
    'P_BS': 'gas'
}
```

For zone 1, however, I have no real clue what to do. Gas is very, very dominant there, but thermal (P_GRES) and CHP (P_TES) types don't allow to put them into a certain fuel category. Maybe this is the way to go for the moment. 

```
MAP_GENERATION_1 = {
     'P_AES': 'nuclear',	     'P_AES': 'nuclear',
     'P_GES': 'hydro',	     'P_GES': 'hydro',
     'P_GRES': 'unknown',	     'P_GRES': 'unknown',
     'P_TES': 'unknown',	     'P_TES': 'unknown',
     'P_BS': 'unknown'	     'P_BS': 'gas'
 }
```

Any other ideas, @brunolajoie ?

tmslaine:
@alixunderplatz Is the map on resource watch reliable? For example Omsk has no generating facilities at all!
Do you think energybase.ru would be better source? It will be quite data-intensive to scrape the required information from there, but maybe it is possible to get results for fuels of estimated generation on a GRES/TES level

corradio:
@tmslaine that's great!
Are you able to provide screenshots of the frontend running, alongside the production mix for each zone so we can visually check?
Then we can move on to adjust carbon intensities for each zone.

loganbyers:
@tmslaine @alixunderplatz  Hi, I'm one of the contributors to the Global Power Plant Database that is on resourcewatch.org. Our Russia data that is geolocated covers about __70% of Russia's capacity__ (182.75 GW of ~263.5 GW), so it certainly yields an incomplete picture. Going forward we have improvements to our Russia data as a highest priority over the next ~4 months. 

I am happy to learn about the energybase.ru website that was posted, as it has some 277 GW of capacity. I haven't looked to see if that is all operational currently, but maybe our team has previously investigated how we could use this source.

tmslaine:
@corradio Screenshots:
ru1b
ru-2b

corradio:
Looking good! @alixunderplatz @brunolajoie can you guys go through the mapping?
Should we could autoconsumption at all? To me this map is about grid electricity, so maybe we should remove it?

brunolajoie:
We want to exclude industrial auto-consumption as much as possible from the electricityMap, cause electricity is not flowing to the transmission grid (ref #1430), so I think we should remove P_BS from our parser if we believe this P_BS is close to 100% auto-consumed

- For RU1, we'll have to map GES and GRES to unknown, but **we should probably find a rough estimate of the annual breakdown between gas and coal in terms of TWh produced (or if not in GW installed)** before merging the PR, in order to refine the GHG intensity factor of unknown for this zone proportionally to the split between coal and gas

- For RU2, we could use @tmslaine investigation to map it as follow
P_GRES' : coal
'P_TES': 96% goes to coal, 4% to gas
This would make the zone very close to 100% coal, and therefore I think it's acceptable to not put it all in "unknwown" as we usually do when no real-time breakdown of thermal production is available.
@tmslaine do you think you could prepare some short sentences to source your estimation so we can come back to it later if needed? A 1-2 sentence commentary directly the code could work. If its too long, a link to a static document published online could work too.

did I miss anything?

tmslaine:
If industrial complexes are connected to the grid and the transmission system operator estimates their production/consumption in their live reporting, isn't it cleaner to keep them in the parser and stay consistent with the source? I suspect industrial power stations do not always consume 100% of what they produce...

I think we can use the [ORDER OF THE MINISTRY OF ENERGY OF RUSSIA OF 28.02.18 NO. 121 "ON APPROVAL OF THE SCHEME AND PROGRAM FOR THE DEVELOPMENT OF THE UNIFIED ENERGY SYSTEM OF RUSSIA FOR 2018-2024"](https://minenergo.gov.ru/node/11323)
for estimating CO2 intensity and capacities.
Table 5.4, p.60: Capacities 2017 and capacity forecasts 2018-2024 for Nuclear, Hydro, Pump storage, Thermal, Renewable on a large grid area level (Northwest/Ural/Center etc.)
Table 7.3., p.80: Estimated consumption of Gas / Oil / Other / Coal in thermal generation
Appendix 11, p.227: Capacities 2017 and capacity forecasts 2018-2024 for Nuclear, Hydro, Thermal, Renewable on an oblast-level to account for RU-1 not having Kaliningrad, Komi, Arkhangelsk from the grid area Northwest.

Using estimated fuel consumption for 2018 from table 7.3. with RU-2=Siberia,
 RU-1=(NW+Centre+Volga+South+Ural), and assuming the same generating  efficiency for all fuels , thermal generation is on average:

- 88% gas, 0.5% oil, 7.7% coal, 3.8% other or 524g CO2eq/kWh for RU-1
- 8.3% gas, 0.4% oil, 86.7% coal, 4.6% other or 786g CO2eq/kWh for RU-2

It might be better to leave the GRES/TES/BS generation as unknown and use these estimates than to map generation to coal/gas.

brunolajoie:
Super nice! I think we can go with your suggested split for RU1 and RU2: Using fuel consumption is always much better than capacity installed indeed. I wasn't really able to double check, its all in Russian, but just to confirm: did you managed to get expected fuel use in 2018 for electricity production, and then, convert that into kWh using a similar fuel-to-power efficiency for all fuel?
We could use a slightly higher fuel efficiency for gas vs. oil or coal to back compute kWh, but at this stage, all we are doing is guesstimates so no pb to start from there.

Just to confirm: is your suggested split supposed to taking into account fuel use from industrial auto-consumption? If so, maybe it's better to integrate "BS generation" in our map indeed.

tmslaine:
@brunolajoie Yes, CO2 intensities are based on expected fuel use in 2018 for electricity production. I was wondering myself that gas probably might have higher efficiency. If we found an estimate like [this](https://www.ecofys.com/files/files/intern.-comparison-of-fossil-power-efficiency-and-co2-intensity-2016.pdf) for Russia, I am happy to tweak the numbers.
My impression is that Block Stations are probably included in the estimate, at least Russian System Operator lists of power plants over 25MW include power plants owned by industrial combinates.

Issue #1434, opened 2018-06-03
Data for Iraq


tmslaine:
http://109.224.53.139:8080/IPS.php has real-time data for Iraq, an example from today:
![iraq_gen](https://user-images.githubusercontent.com/25774514/40884963-480e7026-6726-11e8-8034-334a6ec622ad.png)
"TIE LINES" seem to be imports from other areas. The only one I was able link to Iran is "MERSAD" for Mersad-Diyala, but many of the other ones could be Iran->Iraq as well.
Edit: 
I think the dashboard shows data for Iraq excluding Kurdistan. As a quick check the "Loads of Governorates" dashboard is indeed missing the governorates Dohuk, Erbil and Sulaymaniyah and the load on that dashboard is roughly the same as the field "TOTAL" on "Iraqi Power System".
Perhaps Iraq should be split to Iraq and Iraq (Kurdistan) with Iraq (Kurdistan) shown on eMap similarly to Crimea?

Sources:

- MERSAD - Iran: http://ninanews.com/News_Details.aspx?3yN4%252bMn38qQxw8wdRO%252fpbw%253d%253d&AspxAutoDetectCookieSupport=1
- KRMS, KRKH, SBZB -Iran; if machine translation is to be trusted this recent article mentions the transmission lines Karkh (KRKH?), Serbil Zehab(SBZB?), Kermanshah(KRMS?): https://www.alsumaria.tv/news/231154/%D8%A7%D9%84%D8%B9%D8%B1%D8%A7%D9%82-%D9%8A%D8%AA%D9%81%D9%82-%D9%85%D8%B9-%D8%A7%D9%8A%D8%B1%D8%A7%D9%86-%D8%B9%D9%84%D9%89-%D8%AA%D8%AC%D8%AF%D9%8A%D8%AF-%D8%A7%D8%B3%D8%AA%D9%8A%D8%B1%D8%A7%D8%AF-%D8%A7%D9%84%D8%B7%D8%A7%D9%82%D8%A9-%D9%84%D8%B9%D8%A7%D9%85/ar

- KURDISTAN : This article mentions that Iraqi Kurdistan is separated from the national grid: http://www.iq.undp.org/content/iraq/ar/home/operations/projects/poverty_reduction/kr-electricity-reconstruction.html
I am inclined to think this refers to Iraq (Kurdistan) as the Iranian province Kurdistan is only connected to the the Iraqi national grid via Iraq (Kurdistan)

Edit2:
[...]

PR #1447, opened 2018-06-09
Add Iraq

alixunderplatz:
Feel free to comment and improve this attempt for a parser and to add/update the remaining files :)

ref #1434


systemcatch:
iraq

parsers/IQ.py
```diff
@@ -0,0 +1,122 @@
...
+
+    sortedZoneKeys = '->'.join(sorted([zone_key1, zone_key2]))
+
+    r = session or requests.session()
+    response = r.get(url)
+    iraq_json = response.json()
+    iraq_exch = iraq_json['d']
+    iraq_time = iraq_json['lastmodified']
+
+    #    tie_line_mapping = {
```
corradio: is this still needed?

corradio:
This is amazing guys! Looking good from here but need to merge master first.

parsers/IQ.py
systemcatch: Not sure, @alixunderplatz ?

systemcatch:
:+1:  I forgot to rebase before starting my additions, let me try and resolve the conflicts.

parsers/IQ.py
alixunderplatz: If there is a way to grab the data in one step in fetch_production and then reuse it for fetch_exchange (and/or fetch_consumption), then it is unnecessary. But I don't know yet how to do this, teach me, masters! ^^

systemcatch:
@corradio we are good to go!

PR #1449, opened 2018-06-13
Zone details new time navigation

jbuverud:
image

image

The slider adjusts the selected time index (including the markers on the graphs) automatically. If one hovers over the graphs, the slider gets adjusted, but after hover end it goes back to its previously (manually) assigned position. 

I also made the tooltips update their content when adjusting the slider, as they tooltips stay visible on mobile after pressing the graph until they are manually dismissed (meaning the slider can be adjusted at the same time as the tooltips are visible).


web/src/reducers/index.js
```diff
@@ -23,6 +23,7 @@ const initialApplicationState = {
   legendVisible: false,
   locale: window.locale,
   onboardingSeen: Cookies.get('onboardingSeen') === 'true' || false,
+  openTooltipMode: null,
```
corradio: Can we rename to something a bit more explicit? Open is a verb and therefore it's a bit hard to understand it as a state. Maybe just `tooltipDisplayMode`? Adding the word `display` makes it a bit more explicit that a `null` value should hide it.

web/src/styles.css
```diff
@@ -1390,6 +1402,128 @@ sub {
...
+    border-radius: 1.3px;
+    border: 0.2px solid #010101;
+  }
+  
+  .time-slider-input:focus::-webkit-slider-runnable-track {
+    background: white;
+  }
+
+
+  /* IE */
```
corradio: omg crazy IE style :(
Well done.
Can you maybe add a small comment on why IE and FF need separate styles?

brunolajoie:
I've tested it locally, it's working quite nicely.

Just thinking, that's not always obvious that the time span is 24h, especially with these annoying AM/PM notations. 

Any idea of visual clue to show that more clearly? One option would be to set a day/night indicator, for the local timezone of the country selected maybe? 
image

corradio:
yeah there's a couple of issues with ideas on how to do that already.
Good points but I believe this is for a later improvement.

brunolajoie:
yeppe!

PR #1452, opened 2018-06-14
Add consumption forecast to US_PJM parser

### LINKED ISSUE #1345: add PJM load forecast
Data available here : http://www.pjm.com/markets-and-operations/energy/real-time/7-day-load-forecast.aspx

lorrieq:
closes #1345 
* Fetches data through the PJM API
* Requires PJM_TOKEN to be added to secrets.env


parsers/US_PJM.py
```diff
@@ -115,6 +119,50 @@ def data_processer(data):
...
+        utc_datetime = elem['forecast_datetime_beginning_utc']
+        data_point = {
+            'zoneKey': zone_key,
+            'datetime': arrow.get(utc_datetime).replace(tzinfo='UTC').datetime,
+            'value': elem['forecast_load_mw'],
+            'source': 'pjm.com'
+        }
+        data_points.append(data_point)
+
+    return data_points
```
systemcatch: Can you post the output of the parser.

parsers/US_PJM.py
```diff
@@ -115,6 +119,50 @@ def data_processer(data):
...
+    url = API_ENDPOINT + 'load_frcstd_7_day'
+    resp = requests.get(url, params, headers=headers)
+    data = json.loads(resp.content)
+
+    data_points = list()
+    for elem in data:
+        utc_datetime = elem['forecast_datetime_beginning_utc']
+        data_point = {
+            'zoneKey': zone_key,
+            'datetime': arrow.get(utc_datetime).replace(tzinfo='UTC').datetime,
```
systemcatch: It's definitely returning UTC?

parsers/US_PJM.py
lorrieq: sample output [attached](https://github.com/tmrowco/electricitymap/files/2103816/pjm_output.txt)

maxbellec:
great to see this. I agree with @systemcatch's review, once we're set on the time zone we're good to go

parsers/US_PJM.py
lorrieq: Yes :) Below is a sample data point from the API. You can see a datetime is provided for local time (EPT) and UTC time. I take the UTC datetime and set the tzinfo accordingly.
pjm_snip

lorrieq:
To get an API key you must follow the instructions [here](http://www.pjm.com/-/media/etools/data-miner-2/data-miner-2-api-guide.ashx?la=en), it doesn't take long.

systemcatch:
Once you guys have the API key setup this can be merged. Thanks @lorrieq!

corradio:
Our system is currently not geared to have multiple load forecast.
This one seems to be 7 days ahead. We definitely don't want to overwrite the EIA one.
Also 7days ahead would need special care as we would need to store the time at which the forecast 
is made (as several forecasts will overlap during a day).
@maxbellec any ideas?

Worst case, and until we figure out how to do it properly, we can merge the code, rename the method to make it clear it's 7days ahead, and remove the change in `zones.json` to make sure it doesn't override the current forecast.

systemcatch:
You could also limit the data it returns to one day ahead.

lorrieq:
Hey @corradio, I'm happy to go with what you've suggested for now, to merge the code while maintaining EIA as the load forecast.

P.S. I think it would be ideal to store the time at which a forecast is made. It gives the forecast context and allows you to see the progression of the forecast. It would also be great if you could store data for multiple sources for comparison of forecasts. Not fully sure of the implications those would have on the backend or if this would even be useful for you guys, just my thoughts!

corradio:
Hi @lorrieq ,

Sorry for the late answer. I think we should proceed. Can you make the changes I mentioned in https://github.com/tmrowco/electricitymap/pull/1452#issuecomment-397550400 and then we can merge?

Thanks,

Olivier

parsers/US_PJM.py
```diff
@@ -372,6 +420,8 @@ def fetch_price(zone_key='US-PJM', session=None, target_datetime=None, logger=No
 
 
 if __name__ == '__main__':
+    print('fetch_consumption_forecast() ->')
```
corradio: those two probably need to be renamed

PR #1453, opened 2018-06-15
Add wind/solar ENTSOE forecast

corradio:
This adds the code but does not add it to the config and will thus not be inserted in the database.


parsers/ENTSOE.py
```diff
@@ -1065,3 +1081,43 @@ def fetch_consumption_forecast(zone_key, session=None, target_datetime=None,
...
+    domain = ENTSOE_DOMAIN_MAPPINGS[zone_key]
+    # Grab production
+    parsed = parse_production(
+        query_wind_solar_production_forecast(domain, session,
+                                             target_datetime=target_datetime))
+
+    if not parsed:
+        return None
+
+    productions, production_dates = parsed
```
maxbellec: not a big deal, but I find it weird that apart from the method name, everything is named production as if it was real values. Maybe rename `productions` to `prod_forecasts`?

parsers/ENTSOE.py
```diff
@@ -1065,3 +1081,43 @@ def fetch_consumption_forecast(zone_key, session=None, target_datetime=None,
...
+
+    data = []
+    for i in range(len(production_dates)):
```
maxbellec:
```
for prod_forecast, production_date in zip(productions, production_dates):
    production_values = {ENTSOE_PARAMETER_DESC[k]: v for k, v in prod_forecast.items()}
```
is more elegant

corradio:
Noted. merging anyway because it's mostly style issues and just want to move quickly here.

PR #1457, opened 2018-06-20
Improve hover explanation of emission percentage in en.json

jzlcdh:
Please see issue 1382


brunolajoie:
Your sentence will yield for instance:
>_Coal emitted 45% of greenhouse gases due to electricity being used now in Germany_

What the electricityMap system actually compute is:
>_45 % of greenhouse gas emissions from electricity produced or imported in Germany comes from local coal power plants_

It's a big long I agree. We could probably reduce to
>_45 % of GHG emissions from electricity produced or imported in Germany comes from local coal_

What do you think?

jzlcdh:
I will close this and and make a new pull request once I have looked at the code.

PR #1461, opened 2018-06-21
remove IN-GJ parser

jarek:
We lost thermal data and the 24 hour history chart now implies they use 100% wind and solar:

image

Production was 10 to 14 GW before:

image

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1759,7 +1759,6 @@
     "flag_file_name": "in.png",
     "parsers": {
       "consumption": "IN_GJ.fetch_consumption",
-      "production": "IN_GJ.fetch_production"
     },
     "timezone": "Asia/Kolkata"
   },


alixunderplatz:
Is this the first real loss of a source? really sad :(

What if we used the "Dutch method" and substituted the missing generation with data from the "Own generation" field displayed on this site.
 
http://meritindia.in/state-data/gujarat

Own generation - solar - wind = missing generation. Using the default "unknown" intensity of 700 gCO2/kWh should be fine for this category.

image

jarek:
@alixunderplatz not really the first to be lost - Delhi was there once, Ireland is coming and going, Switzerland as well (though perhaps Switzerland was never correct in the first place?), we also removed exchanges of El Salvador because they couldn't add up

corradio:
I've done a quickfix: see https://github.com/tmrowco/electricitymap/commit/2c8da8c735e44ab36971a04827504482ad71d2f1

Issue #1382, opened 2018-05-04
"available" not right word in "emissions available" 


jzlcdh:
After clicking the "carbon intensity" link and hovering over a bar there is a reference to "emissions available", which is not the best phrasing. Can't think of the best wording yet but feel free to assign this to me and I will get round to it eventually ("God willing" as my wife would make me say).

corradio:
I agree. This might be a duplicate of https://github.com/tmrowco/electricitymap/issues/1375.

jzlcdh:
Looks very similar but probably not duplicate as probably a different string when I find it.

corradio:
@jzlcdh feel free to move forward with this!

brunolajoie:
We could say, instead of 
>" 76% of emissions in Germany comes from coal"

the following fix:

>" 76% of carbon emissions related to electricity consumed in Germany comes from coal"

corradio:
It's actually not "consumed" because at this point the exports (and storage) haven't been subtracted yet. The proper technical phrasing would be

> X % of emissions from electricity available in Germany comes from coal

i.e. just adding "from electricity".

brunolajoie:
Oh, right, indeed!

jzlcdh:
Sorry I had forgotten about this - I hope to do it in next few days.

jzlcdh:
So just to be clear if electricity is generated in country Y but consumed in country Z are we assigning the emissions to Y or Z?

jzlcdh:
Duh sorry yes it is obvious on the readme. So as a native speaker but not a professional writer the best I can come up with is:

"Coal emitted X% of greenhouse gases due to electricity being used now in Z"

I have created pull request #1457. So would it be easy for me to test the change (I have a windows 10 desktop) or does someone else do that when batching up several minor changes?

brunolajoie:
I'm sorry to insist, but your suggestion is not correct strictly speaking. The exact definition is
"X % of emissions from electricity produced or imported in Germany comes from coal"

brunolajoie:
The resulting carbon intensity will be computed as the scalar product of each of these percentages (for all type of fuels, and all imports), multiplied by the carbon intensity of each type of fuel and each type of imports. It will represent the carbon intensity of electricity consumed in Germany.

jzlcdh:
@brunolajoie I understand and agree with your last comment and I think that is in accord with my suggestion.

[...]

Issue #1375, opened 2018-05-02
Change "...StoredUsing" and "hydro storage" strings in en.json


alixunderplatz:
While translating from en.json to ru.json, I found these lines a bit confusing/odd:

_ _ _ _

`"electricityStoredUsing": "...% of electricity available in [zone] is stored using [type]",
`

I'd change this line to:

`"electricityStoredUsing": "...% of electricity available in [zone] is used to charge [type]",
`
or
`"electricityStoredUsing": "...% of electricity available in [zone] is used for charging [type]",
`
_ _ _ _

Next one:

`"emissionsStoredUsing": " ...% of emissions available in [zone] are stored using [type of storage]"
`

Emissions happen at the moment of charging the storage (=consumption), therefore they are not stored and released later (only usual LCA emissions apply when discharging).

Some ideas for changing this line:
`"...% of emissions in [zone] are caused by charging [type of storage]", `
or
`"...% of emissions in [zone] are released due to charging [type of storage]",  `
or
`"...% of emissions in [zone] are released due to consumption for charging [type of storage]"`

I personally prefer the second idea. I think that's also how I had already translated it in the ru.json

_ _ _ _

In the same turn, I'd suggest changing `"hydro storage": "hydro storage"` 
to `"hydro storage": "pumped-storage hydro "` or to keep it shorter
to `"hydro storage": "pumped-storage"`

This will make it slightly more obvious that everything with a natural water inflow is summarized in the "hydro" category and only electrically charged hydro storages go into that pumped-storage category.

corradio:
Actually right now the co2 model doesn't consider emissions at time of storage but at time of discharge. Unfortunately we don't have the model completely implemented so it discharges with a hydro intensity, which is unfortunate because it transforms batteries charged using coal to hydro plants.

However I prefer your formulation. Feel free to change!
What do you think @brunolajoie ?

brunolajoie:
Right now, storing is like an export to a virtual neighbor that is the reservoir, which entails no emissions indeed. Hence the strange wording that we used. Therefore I wouldn't change the working for "emissionsStoredUsing". 

(As oli said, pb is that we haven't had time to implement proper discharge CO2 intensity computation yet. We really "store" dirty emissions into the batteries, to be released later as hydro.)

But feel free to change to wording for "electricityStoredUsing". My favorite wording for hydro storage is "pumped hydro storage" ("PHS" is a very common naming from what i've read)

jarek:
[...]

PR #1466, opened 2018-06-25
Correct tooltips grammar

jzlcdh:
As discussed in issue #1382 corrected grammar in info shown when hovering over generation and emission bars, and included some of the suggestions re storage in issue #1375

### DIFF
--- a/web/locales/en.json
+++ b/web/locales/en.json
@@ -93,7 +93,7 @@
     "wind": "wind",
     "solar": "solar",
     "hydro": "hydro",
-    "hydro storage": "hydro storage",
+    "hydro storage": "pumped hydro storage",
     "battery storage": "battery storage",
     "biomass": "biomass",
     "nuclear": "nuclear",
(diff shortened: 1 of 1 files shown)


web/locales/en.json
```diff
@@ -103,15 +103,15 @@
...
+    "emissionsStoredUsing": "Please see the Frequently Asked Questions for details",
 
-    "electricityExportedTo": "%1$s %% of electricity available in <img id=\"country-flag\"></img> <b>%2$s</b> is exported to <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
-    "emissionsExportedTo": "%1$s %% of emissions available in <img id=\"country-flag\"></img> <b>%2$s</b> are exported to <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
-    "electricityImportedFrom": "%1$s %% of electricity available in <img id=\"country-flag\"></img> <b>%2$s</b> is imported from <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
-    "emissionsImportedFrom": "%1$s %% of emissions available in <img id=\"country-flag\"></img> <b>%2$s</b> are imported from <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
+    "electricityExportedTo": "%1$s %% of electricity in <img id=\"country-flag\"></img> <b>%2$s</b> is being exported to <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
+    "emissionsExportedTo": "%1$s %% of emissions due to electricity in <img id=\"country-flag\"></img> <b>%2$s</b> are due to exports to <img id=\"country-exchange-flag\"></img> <b>%3$s</b>",
```
corradio: words are expensive (as we have little room in the tooltip).
Therefore, may I suggest changing "emissions due to electricity" to "emissions from electricity"?

jzlcdh:
@corradio Re your comments:

1) Yes I agree "emissions from electricity" is better than "emissions due to electricity".
2) I don't understand what you are saying here about changing the meaning as I think if you asked the average person the meaning of "electricity available in Germany" they would struggle to say what it means.
3) I think "an area is importing emissions from another country" is an implicit political statement about how the responsibility for emissions should be apportioned. Whilst I agree with you politically: e.g. that the emissions due to the manufacturing of the laptop I am now typing on should be apportioned to me I think that should be said explicitly.

So although this has been a stimulating discussion and made me think, because of 2) and 3) it is getting too complicated so I had better close this and put in my future pull requests for one simple and uncontroversial change at a time and leave 2) and 3) to yourself or other people.

PR #1467, opened 2018-06-25
Explain emissions related to storage

jzlcdh:
Hope this is a reasonable public explanation of what was discussed in issue  #1375. Also see issue #1382


web/src/components/faq.js
```diff
@@ -72,6 +72,10 @@ const faq = {
         question: 'Do you take into account imports and exports of electricity?',
         answer: 'Yes, we do. Imports and exports can be see on the map as small arrows between different areas [link to question about arrows]. Detailed information can be seen in the charts shown when you click on an area.',
       },
+      emissionsOfStored: {
+        question: 'What about emissions generating electricity which is then stored in batteries or used to fill reservoirs?',
```
corradio: can you change to "emissions from generating electricity"? I think that will read better.

corradio:
Great initiative @jzlcdh - thank you very much. I've added a minor comment, and I'll merge right after.

PR #1492, opened 2018-07-03
fixed country detail padding issue with longer text. changed Co2 <sub>

### LINKED ISSUE #1491: German translation hides top buttons
It seems like part of the emission toggle is not visible, probably because the element is positioned absolute. This doesn't play well when the header is higher than expected (due to an extra line break in German)

cc @ovbm 

image

ovbm:
fixes #1491 

Changes I made:

- Added the "Electricity | Carbon emissions" menu into the absolute header, so that it's always visible.
- Reformatted the co2-gage headers (line-height etc) so that 3 lines wouldn't take up much more space
- Increased padding on the content to account for the menu and 3 lines height.
- Changed the Co2 sub implementation <sub> increases the box-height of the text, which makes it hard to make consistent height predictions. replacing it with <span class="sub"> fixes that.
- I opted for this approach keeping the absolute header because i still think it's good to always see it. however if it causes more problems in the future we have two options: (1) remove absolute positioning, (2) javascript read header height and set padding (not recommended)

result:
screen shot 2018-07-03 at 12 43 03

### DIFF
--- a/web/src/helpers/formatting.js
+++ b/web/src/helpers/formatting.js
@@ -4,7 +4,7 @@ var d3 = require('d3-format');
 var translation = require('./translation');
 
 var co2Sub = module.exports.co2Sub = function (str) {
-  return str.replace('CO2', 'CO<sub>2</sub>');
+  return str.replace('CO2', 'CO<span class="sub">2</span>');
 };
 module.exports.formatPower = function (d, numDigits) {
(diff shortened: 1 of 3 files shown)


ovbm:
that's what I was looking for, something with the <span sub > is not working on german right now. can you revert the merge?

PR #1494, opened 2018-07-03
Ob/#1491

### LINKED ISSUE #1491: German translation hides top buttons

ovbm:
fixes #1491 

padding issue fixed


ovbm:
screen shot 2018-07-03 at 12 58 04

PR #1496, opened 2018-07-03
fix `test_parser.py` for exchanges

maxbellec:
and add some docs


maxbellec:
thanks @systemcatch for the fast review!

Issue #1497, opened 2018-07-03
California Elec. Imports


joshuasgoldstein:
Olivier asked me to post this issue to get responses. - Prof. Joshua Goldstein

On 4/9/2018 9:06 AM, Olivier Corradi wrote:
> Hi Joshua,
> For now we simply assume that the imports don't dilute the current mix - so no matter how much power is imported, the current mix will remain unchanged.
> That's the least wrong assumption we can do without any other information.
> Olivier

Dear Olivier,
    I'm following up on the question of California imports now that you have AZ and NM on the U.S. map at the moment.  I believe the "least wrong assumption" you can make is to take the monthly fuel mix for exporting regions, or the real-time fuel mix if you have it. Assuming that imports don't dilute the current mix is obviously wrong, and in the case of California that assumption makes CA seem "greener" than it really is.  Many people are using the electricity map app (I tell people about it all the time), and Calif. is the world's leading example of heavily integrating solar on the grid, so what happens at night is important to evaluating the results.
    I looked at the data yesterday morning at 6 am (July 2):

You showed CA at 183 g.  As the sun had not yet risen, there were (as every night) a LOT of imports, 32%, which you also showed at 183g. [23% gas @490g; 17% wind@11g]. A more accurate estimate, I believe, is about 325g for the imports, raising the state's overall intensity to 230g. Here is my reasoning:

  First, we know something about where the imports come from -- in summer, about equally from the Northwest and Southwest; in winter, about two-thirds from Northwest and one-third from Southwest.*
  The electricity generated in Oregon as of March (latest data) is three-quarters hydro and one-quarter gas. Counting hydro @24g and gas @490g, that mix is around 140g. But the whole of the Northwest grid has a lot of coal. That's the pool from which Calif. draws electricity, and seems more consistent with your methodology. Oregon shows the fuel mix of electricity consumed in Oregon as 32% coal (@820g), 17% gas (@490g), 41% hydro, which is overall something like 350g.**  That's a little out of date (2014-16). The main point is that you can't cherry-pick the clean energy from a grid region and say that's what's being exported, because if not exported it would displace the dirty sources in that region. Even with changes since 2014-16,the imports from the north seem to be at least something like 250g.
  Your real-time data for SVERI show 480g: from 38% coal @820g, 34% gas @490g, and 25% nuclear @12g. [Again, the nuclear is closer to Calif. but if not exported it would displace coal on the SVERI grid, so you can't say the imports are clean.]
[...]

PR #1500, opened 2018-07-03
hardcode imports into US-CA at 325 gco2eq/kwh

jarek:
ref #1497

To change the numbers or the assumptions, edit co2eq_parameters.js on this branch.

Otherwise this should work, I hope. I didn't test it though! Sorry, I don't have capacity right now, hopefully others can.

Mock-up of how I expect it'll work visually:

image

And in the exchange hover it'll say "US-intoCA -> United States of America (California)". That could be changed by updating files in web/locales to provide a string for "US-intoCA", but as it only shows on desktop in the hover it seems low priority.


jarek:
BTW, back-end question: if I were to change co2eq_parameters.js to return closer to 250 g in December, 400 g in June, and blend in between - would that work as expected? That is: is the export CO2 intensity override saved when the export value is saved, or is it every time the intensity is recomputed (and could e.g. return November's value when archive data for June is requested in November)?

Apart from the (very fair) argument that it would complicate our calculations and make them less transparent for relatively little gain...

corradio:
Changing the area name / import name will require us to do a data migration (cc @maxbellec).
@jarek the changes you make affect the past. So only a single carbon intensity number is used, and that's the one on the `master` branch (even for past data, as we recompute it on the fly).
I think the simplest right now is to keep the exchange named the same, but update the parameters?

systemcatch:
I built your branch @jarek but kept getting this error when running update_state.py, other zones work ok still.
```shell
(EM-env) chris@ThinkPad:~/electricitymap$ PYTHONPATH=. python mockserver/update_state.py US-CA
Finding and executing US-CA production parser US_CA.fetch_production..
{ 'datetime': datetime.datetime(2018, 7, 4, 2, 30, tzinfo=tzfile('/usr/share/zoneinfo/US/Pacific')),
  'production': defaultdict(<class 'float'>,
                            { 'biomass': 559.0,
                              'coal': 10.0,
                              'gas': 6668.0,
                              'geothermal': 1005.0,
                              'hydro': 2589.0,
                              'nuclear': 2261.0,
                              'solar': 0.0,
                              'unknown': 0.0,
                              'wind': 2390.0}),
  'source': 'caiso.com',
  'storage': defaultdict(<class 'float'>, {'battery': 20.0}),
  'zoneKey': 'US-CA'}
{ 'datetime': datetime.datetime(2018, 7, 4, 2, 30, tzinfo=tzfile('/usr/share/zoneinfo/US/Pacific')),
  'netFlow': -7165,
  'sortedZoneKeys': 'US-CA->US-intoCA',
  'source': 'caiso.com'}
{ 'datetime': datetime.datetime(2018, 7, 1, 8, 0, tzinfo=tzutc()),
  'netFlow': -48,
  'sortedZoneKeys': 'MX->US-CA',
  'source': 'mysource.com'}
Updating and writing state..
  File "mockserver/update_state.py", line 154, in <module>
    json.dump({'data': obj}, f)
  File "/usr/lib/python3.5/json/__init__.py", line 178, in dump
    for chunk in iterable:
  File "/usr/lib/python3.5/json/encoder.py", line 429, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.5/json/encoder.py", line 403, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.5/json/encoder.py", line 403, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.5/json/encoder.py", line 403, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.5/json/encoder.py", line 403, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.5/json/encoder.py", line 403, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.5/json/encoder.py", line 436, in _iterencode
    o = _default(o)
  File "/usr/lib/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: 7165 is not JSON serializable
```

systemcatch:
Further digging revealed that the netflow key was type `<class 'numpy.int64'>`, setting it to float makes the json dump work fine.  Not seeing the import having 325g intensity yet though.

ca_export

corradio:
I don't think our mock server supports that right now unfortunately

systemcatch:
Next mockserver milestone then? :grin:

jarek:
> I think the simplest right now is to keep the exchange named the same, but update the parameters?

That's fine by me, except then it might look like we've determined all of US to be 325 g, rather than just the bits that California is importing from.

jarek:
@systemcatch can you push a fix for the `float`/`numpy.int64`? I'm getting seemingly unrelated errors for EIA.py when running update_state

jarek:
@corradio How would you like to proceed? Should I change the import back to `US->US-CA`?

jarek:
CircleCI seems to be frozen, please merge as you see fit if/when it reports

PR #1501, opened 2018-07-03
remove Albania override

jarek:
per @alixunderplatz https://tmrow-team.slack.com/archives/C2DQPTHA6/p1530429840000010:

"The hardcoded carbon intensity for Albania (24 g/kWh) should be removed for now. 
There are regular power flows from Montenegro and Greece during the day with high emission intensities, and a high share of these imports is going to Serbia/Kosovo."


corradio:
@alixunderplatz @systemcatch let me know if I was too quickly on the merge.
Maybe I should wait for all reviewers to approve before merging?

PR #1505, opened 2018-07-05
rename pumped hydro storage into hydro storage

brunolajoie:
pumped hydro storage is a bit long and gets cut in 99% of the screenshot i've seen posted online
image


corradio:
I actually already did it. I think if you merge master it will show no changes.

corradio:
yep confirmed.

Issue #1242, opened 2018-03-19
Data for Mexico MX


alixunderplatz:
Promising news!

There is total generation (no breakdown yet), load data and power flow data for **Mexico**! 
Seems like [this site](https://www.gob.mx/cenace/articulos/graficas-de-demanda-y?idiom=es) has been online since 15th March, so that's an accidental find :)

Power flows and total generation and consumption per region (hit button on the right to switch):
http://www.cenace.gob.mx/Paginas/Publicas/Info/DemandaRegional.aspx

![image](https://user-images.githubusercontent.com/25743609/37621631-c3d10f04-2bbf-11e8-9c99-24449a74084f.png)

And generation and consumption as a time series (seems like it is updated every 15 min, see timestamp on the left):
http://www.cenace.gob.mx/GraficaDemanda.aspx

![image](https://user-images.githubusercontent.com/25743609/37623487-2fc0fae4-2bc5-11e8-9bcc-612749002b9d.png)

All live exchanges with the U.S. in the North are there, also with Belize (new) and Guatemala in the south.

Seems like the two regions on the left part of another system/synchronous area: "National Interconnected System, does not include the regions of Baja California and Baja California Sur."
So California is only linked with these and not with the Mexican "mainland", but Texas is.

_ _ _ _

Seems like this is the open data platform of the Mexican government:
https://datos.gob.mx/libreusomx

Maybe there will be or is more info on breakdown, so please take a look!

Terms of use: https://datos.gob.mx/libreusomx

systemcatch:
Going on http://www.cenace.gob.mx/Paginas/Publicas/Info/DemandaRegional.aspx and clicking 'intercambio entre regiones' (exchange between regions) seems to show live flows between Mexico and the USA. We could use this for #959!

alixunderplatz:
@systemcatch right, works for CAISO imports! 
The MX region "Baja California" in the upper left corner is not connected with the rest of the Mexikan mainland, so the influence on CAISO will be really low (24 MW at the moment, I've found up to ~100 MW on EIA's site in March) 

![image](https://user-images.githubusercontent.com/25743609/37722182-a408f098-2d2b-11e8-976f-5d82eed4d9d8.png)

brunolajoie:
What about adding it to the map as 100% "unknown", taking 2017 average mix to compute CO2 intensity of "unknown", and iterate from there when we'll have the proper breakdown per fuel?

alixunderplatz:
[...]

PR #1512, opened 2018-07-16
Create Mexican exchanges

systemcatch:
This is a prototype for a real time exchange between CAISO and Baja California (MX). We need to discuss how this will work with the US->US-CA exchange which has a fixed carbon value and if there needs to be a deduction somewhere. It would also be nice to keep using the EIA parser for historical data.

I had some problems running update_state due to json errors with the US->US-CA exchange which I'll look at later.

ref #959 , #1242, #1497 

ca_mx


alixunderplatz:
@brunolajoie The MX region "Baja California" is not connected with the rest of the Mexikan mainland! 
maybe we can use the global power plant database to find out what is installed in that  area or find a source with generation per Mexican state per type

jarek:
Also the US-CA->MX exchange values appear to be in hundreds-of-MW range, which won't impact the CO2 intensity nearly as much as 5-10 GW from elsewhere in the US

systemcatch:
http://datasets.wri.org/dataset/globalpowerplantdatabase provides the names of most of Mexico's power plants though sadly it doesn't state which regions they are in.

brunolajoie:
Don't hesitate to discuss that in the issue #1497 and improve directly on the spreadsheet I've done [there](
image
)

brunolajoie:
don't hesitate to comment on issue #1497 and edit the spreadhseet i've done here https://docs.google.com/spreadsheets/d/1grrWRVbmNbpfiUvXFbkh55QfY7CK5lizow7aNX-C9NU/edit?usp=sharing

jarek:
If we want to be really accurate, given the lack of electrical connections it might make sense to split out Baja California from Mexico at this stage already and make the exchange MX-BCN->US-CA

systemcatch:
For me the timestamp doesn't seem to update, it's stuck on hour 12 despite the exchange value updating regularly.

I agree we should switch to MX-BCN->US-CA, the natural earth shapes make this really easy.

jarek:
Hm, the timestamp seems updated right now, but I did see it report 0:00:00 a few hours ago when it definitely wasn't midnight or noon in Mexico. Good call on not using it then.

systemcatch:
Separated Baja California from Mexico, now we just need the carbon intensity.

baja_california

config/exchanges.json
```diff
@@ -1724,9 +1724,7 @@
       -115.846,
       35.962
     ],
-    "parsers": {
-      "exchange": "US_CA.fetch_exchange"
-    },
+    "parsers": {},
```
jarek: This change for US->US-CA should probably not be here?

systemcatch: Ok it should be fixed now, there's still a problem with US->US-CA not working though (json serialization error I think).

corradio:
Hi all,

Thank you for the terrific work. I'm discovering that I spend a lot of time getting up to speed with the background information in order to review a PR.
**Would it be possible to add a TLDR of the changes (and the justifications behind them) in the PR?** This would save a lot of time and simplify things.

The questions I have at this point are:

1. What is the reason to switch from EIA to a new source? (didn't the existing data source give us imports from MX / Baja California Norte?)
2. It seems that the new data source only gives net imports. Did we figure out how to extract the exchange with each zone?

systemcatch:
Hey @corradio I hope you had a good holiday! :beach_umbrella: 

To answer your questions;

1) We're not switching from the EIA to another source for the US->US-CA exchange, it will remain the same because it does not include any flows from Mexico per @brunolajoie.

However you are right that https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/EIA.py#L23 will need to be changed. That exchange is only able to get historical data with 1-2 days delay whereas the new source is 5min intervals that will show on the map.

2) Since the Baja California grid is isolated from the rest of Mexico I've split it into it's own separate region and created the MX-BC->US-CA exchange.

We're trying to find a good default carbon intensity for Baja California in #1497

corradio:
Yes I'm well energised again, thank you!
Thank you for your answers.
How come https://github.com/tmrowco/electricitymap-contrib/pull/1512/files#diff-9196b58d1cdd3db6a09227448696ba09R1435 was changed then? It seems like EIA was also used for the MX->US interconnector. Is the rationale related to the fact that EIA is delayed whereas the other source is real-time?

systemcatch:
@corradio you're right that was changed since the other source is real time and will show up on the map.

 @jarek it was only for MX->US-CA (CAISO).

corradio:
Thanks for pushing this forward @systemcatch .
I completely missed the toggle on right hand side of their website.
After toggling, we actually have access to flows, and not just net imports of each regions.

image

My suggestion: why don't we add all of the exchanges of MX in one go? You already did the split, so it should be easy?

I'll promise to be more reactive to review/help/merge :)

systemcatch:
Hey @corradio I can give that a shot though I'll have to further break Mexico into the control areas shown at http://epis.com/powermarketinsights/index.php/2016/02/17/the-new-electric-market-in-mexico/, hopefully they roughly follow the statelines.

Can you rerun CircleCI? It ran into the pylint bug that you fixed today.

Also with the conflict in world.json I think you regenerated it because of an error? I'll merge the good version then resplit if that's the case.

alixunderplatz:
hi @systemcatch , I hope this will help a bit:

image

Looks like norte and noreste share a border in the middle of the state of Coahuila and central and occidental in Morelos

systemcatch:
@alixunderplatz  That's great! We can split by states here.

How did you manage to make that image?

alixunderplatz:
@systemcatch That's just two images as an overlay (mode: multiplied) ;)

corradio:
@systemcatch you will need to merge master on your branch to fix the circle ci bug

systemcatch:
Hey everyone I've now managed to split Mexico and add in all the exchanges. The only things left to do are name the new regions in the english translation and decide on a carbon intensity for Baja California, cc @brunolajoie.  

split_mx_exchanges

brunolajoie:
Terrific! According to our discussion [here](https://github.com/tmrowco/electricitymap-contrib/issues/1497#issuecomment-412098334)   #1497, we can set the default Baja-->US-CA export intensity to 453gCO2eq/kWh!

magol:
I try to localize the regions of Mexico to Swedish, and I have a question about "occidental". 

If I understand correctly, "Occidental" means "Western" in English. My question is whether I'm should use "Occidental" or if I should translating it into "Västra" ("Western" is in Swedish)?

systemcatch:
@magol only just seen your question. You'd be better off using the Swedish words for Western and Eastern imo.

PR #1531, opened 2018-07-28
Fix mobile CSS

corradio:
Tested on:
- Desktop Chrome
- Cordova Android 5.1
- iOS Chrome

On mobile devices, when WebGL isn't available, the "map" tab shouldn't be available.
There was a CSS bug that prevented that.

Also fixed some layout errors when clicking on an area without a parser.

Also addressed (hopefully) a layout bug on Android 5.1.

image


web/views/pages/index.ejs
```diff
@@ -211,7 +211,7 @@ co2Sub = function(str) {
                                 </div>
                             </div>
                         <div class="info-text">
-                                <%- __('panel-initial-text.thisproject') %> <a href="https://github.com/tmrowco/electricitymap-contrib" target="_blank"><%- __('panel-initial-text.opensource') %></a> (<%- __('panel-initial-text.see') %> <a href="https://github.com/tmrowco/electricitymap-contrib#data-sources" target="_blank"><%- __('panel-initial-text.datasources') %></a>). <%- __('panel-initial-text.contribute', 'https://github.com/tmrowco/electricitymap#adding-a-new-country') %>.
+                            <%- __('panel-initial-text.thisproject') %> <a href="https://github.com/tmrowco/electricitymap-contrib" target="_blank"><%- __('panel-initial-text.opensource') %></a> (<%- __('panel-initial-text.see') %> <a href="https://github.com/tmrowco/electricitymap-contrib#data-sources" target="_blank"><%- __('panel-initial-text.datasources') %></a>). <%- __('panel-initial-text.contribute', 'https://github.com/tmrowco/electricitymap#adding-a-new-country') %>.
```
corradio: re-indented

web/views/pages/index.ejs
```diff
@@ -224,60 +224,57 @@ co2Sub = function(str) {
...
-                                <%- __('misc.faq') %>
-                        </div>
-                        <div class="mobile-faq">
+                            <%- __('misc.faq') %>
                         </div>
+                        <div class="mobile-faq"></div>
                     </div>
                     <div class="left-panel-zone-details">
 
-                        <div class="left-panel-zone-details-toolbar">
```
corradio: moved the toolbar (country name and flag) into the header (which is always visible).
Not sure why this wasn't done before.

web/src/styles.css
```diff
@@ -111,15 +111,15 @@ line, path {
...
+    overflow-y: scroll;
     padding-right: calc(2rem - 6px);
     padding-left: 2rem;
     margin: 0;
     -webkit-overflow-scrolling: touch;
 }
 .country-panel-wrap {
     position: relative;
-    padding-top: 160px;
+    padding-top: 200px;
```
corradio: now that the header is included, more padding must be applied to left panel content

web/src/styles.css
```diff
@@ -111,15 +111,15 @@ line, path {
 }
 
 .country-panel{
-    overflow-y: auto;
+    overflow-y: scroll;
```
corradio: scroll the country panel instead of the container

web/src/styles.css
```diff
@@ -452,11 +450,11 @@ body {
     margin: 0.6rem 0;
 }
 
-.left-panel-zone-details .detail-bottom-section{
-    padding: 0rem 1rem 1.25rem;
+.detail-bottom-section{
+    flex: 1 0 auto;
```
corradio: remember to flex grow

web/src/styles.css
```diff
@@ -1151,8 +1149,7 @@ div.highscore-button:hover {
     -moz-box-shadow: 0px 0px 12px 0px rgba(0,0,0,0.10);
     box-shadow: 0px 0px 12px 0px rgba(0,0,0,0.10);
     z-index: 2;
-    border-top-right-radius: 8px;
-    border-top-left-radius: 8px;
+    flex: 1 0 auto;
```
corradio: remember to flex grow

web/views/pages/index.ejs
ovbm: I think I remember this being a bit of a compromise to not have too much vertical space being taken up by fixed content, reducing the amount of info visible on smaller screens.

web/src/styles.css
ovbm: Did you test this on desktop? I think there was a specific reason for scrolling the container versus the panel. I think it would be nice to redo the html structure for the panel if we have time, now it's a bit of a patchwork and the lack of consistency is making it difficult to bugfix.

corradio:
Now additionally tested in Safari and Firefox desktop + responsive mode

web/src/styles.css
corradio: yes I tested it. I actually think it's better structured now. But let's review together

PR #1536, opened 2018-07-30
Add Italy bidding zones (GeoJSON)

antifa-ev:


alixunderplatz:
@antifa-ev Wow, this absolutely made my day/night! 😃😃😃
Awesome, thank you so much!!!!!

Zones all look good to me! 
@corradio is there any advantage/disadvantage of having these seperate GeoJSONs compared with "merging the regions the old way" like for the US, Russia etc. in generate-geometries.js?

corradio:
I think it's easier to maintain if we extract them from Natural Earth instead of having custom polygons.
Also the boundaries between countries might be more robust.
@antifa-ev can you tell us a bit about how you extracted the polygons? Is it the aggregation of Italian regions? In that case it might be simpler to extract them using generate-geometries.js.

Very excited about splitting Italy!

antifa-ev:
@corradio I've downloaded the boundaries from OpenStreetMap, reprocessed them (deleted all islands) and simplified them with simplify-geojson.

I could simplify them in a matter, that adjacent boundaries share the same simplification. But this would involve PostGIS, so I would like to avoid it!

And: I've indeed forgotten to fix Sicily.

corradio:
Hi @antifa-ev,
Thanks for the clarifications. However, do the individual areas correspond to regions? If that's the case, it might be better to use the generate-geometries.js script which will take care of everything for you, and will be more robust.

corradio:
@antifa-ev can you ping me on Slack? It will be easier!

jarek:
First thing I'm seeing is that these are LineStrings and our other maps use Polygons. That should be a really minor change though.

jarek:
From second glance, this should be doable from Natural Earth geometries by merging on region_cod, e.g. Toscana is region_cod=IT-52 (and there are several subdivisions of it - Grosseto, Livorno, Pisa, Siena... but they all have the same region_code).

To preview the Natural Earth, download the file http://naciscdn.org/naturalearth/10m/cultural/ne_10m_admin_1_states_provinces_lakes.zip (this is referenced in https://github.com/tmrowco/electricitymap-contrib/blob/master/web/topogen.sh), then upload it to http://mapshaper.org. Using the "i" icon in top-right you can preview properties of the regions.

corradio:
Closed in favor of https://github.com/tmrowco/electricitymap-contrib/pull/1604

PR #1538, opened 2018-07-30
add geothermal as baseload

jarek:
This mirrors de4064461615cfe as was done for nuclear and coal.

Geothermal is almost always used as baseload not varying its output greatly. The specific use case that can be seen is improving the intuitiveness of NZ-NZN origin of electricity chart.

Discussion item:

It looks like changing the order of the "origin of electricity" chart also changes the order of the bar-charts mode list. Thus this change might crowd the top of the sidebar for a value that's 0 for vast majority of regions, and pushing wind/solar/hydro down. Is it worth it?

screenshot - 300718 - 20 25 13

### DIFF
--- a/web/src/helpers/constants.js
+++ b/web/src/helpers/constants.js
@@ -17,12 +17,12 @@ const modeColor = {
 const modeOrder = [
   'nuclear',
   'coal',
+  'geothermal',
   'wind',
   'solar',
   'hydro',
   'hydro storage',
   'battery storage',
-  'geothermal',
   'biomass',
   'gas',
   'oil',


systemcatch:
Hey @jarek so if I understand you the yellow geothermal will drop below the wind and hydro in this image?
 
geo_move

corradio:
yes, and it will also show higher up in the table

jarek:
Sorry I didn't give a preview screenshot for NZ, but I don't think it's actually possible to get zone history from mockserver, so previewing is hard.

I'll merge and let's see how it goes then!

jarek:
New Zealand now has baseload :tada: 

image

Also El Salvador among others:

image

corradio:
Looks so much better!

alixunderplatz:
@corradio @jarek @systemcatch What about adding biomass to the bottom as well? Has a very constant output, too.

jarek:
Biomass seems to have varied between 405 and 215 MW in West Denmark today, and is also used rather intermittently on Bornholm or maybe Aland. It is true that it doesn't tend to change much elsewhere.

PR #1540, opened 2018-07-31
Read wind and solar values from print page

### LINKED ISSUE #1516: Restore access to India-Gujarat (IN-GJ) wind and solar data
Hey everyone,

looks like SLDCGUJ has also removed the access to real-time data websites for wind and solar, just like they did for the conventional real-time data.
The solution should be really simple: We can use the same link which still gives access to that ["print" page](http://www.sldcguj.com/RealTimeData/PrintPage.php?page=realtimedemand.php). It has wind and solar generation on the right side, too ;)

image

Greetings, Alex

lorrieq:
closes #1516

### DIFF
--- a/parsers/IN_GJ.py
+++ b/parsers/IN_GJ.py
@@ -10,6 +10,7 @@ from .lib import zonekey, web, IN
 from .lib.validation import validate
 from logging import getLogger
 
+SLDCGUJ_URL = 'http://www.sldcguj.com/RealTimeData/PrintPage.php?page=realtimedemand.php'
 
 station_map = {
      "coal": ["Ukai(1-5)+Ukai6",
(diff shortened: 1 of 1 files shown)


corradio:
Hi @lorrieq,
Thank you very much for fixing this.
To ease the review process, can you run `update_state.py` and post a screenshot? Or maybe just run the parser and post the output?
Thanks,

Olivier

lorrieq:
Hey Olivier, woops! Of course. See output below.

`{'zoneKey': 'IN-GJ', 'source': 'sldcguj.com', 'production': {'coal': 8635.0, 'geothermal': None, 'oil': None, 'solar': 639, 'unknown': 0, 'nuclear': 0, 'gas': 1357.0, 'biomass': None, 'wind': 3157, 'hydro': 95.0}, 'storage': {'hydro': None}, 'datetime': <Arrow [2018-07-31T14:23:41+05:30]>}`

corradio:
Thanks! The datetime object should be a datetime object (not arrow). Can you change that?

PR #1543, opened 2018-08-03
Add Kyushu (Japan)

tmslaine:
Adds Kyushu(solar/nuclear/unknown), flows between Japanese control areas and generation volumes for these areas calculated from consumption and imports.

Still requires some cleanup.
japan_parserb


corradio:
Hi @tmslaine,
This looks absolutely terrific. Could you tell us a bit more about the methodology and the assumptions? My japanese is non-existent so I have a hard time reading the websites.

Also, we should remember to update README.md with the data source.

Olivier

tmslaine:
Hi,

Each of the electric power companies balancing the control areas is I guess obliged to report consumption in their respective regions. Searching 電気予報(electricity forecast) will lead to their consumption report/forecast sites. The consumption for each area is reported on a 5-min frequency in a fairly standard csv format, which the JP parser reads. E.g. http://www.kyuden.co.jp/power_usages/csv/juyo-hourly-20180803.csv (encoding = 'Shift-JIS'), with Date / Time / Value columns starting from row 43

Organization for Cross-Regional Coordination of Transmission Operators (occto) reports flows, on a 5-minute basis, on their website http://occtonet.occto.or.jp/public/dfw/RP11/OCCTO/SD/LOGIN_login .
After clicking 地域間連系線情報(Cross-Regional transmission line info) on the left and 連系線潮流実績(Transmission lines actual flows) on the right, one can choose transmission line and date and press 検索(search). What you get is an image with region names on the left, actual flow in black. Clicking CSV保存saves the data as a csv. The Occtonet parser gets to this point and parses the columns 対象日付(Date), 対象時刻(Time),  潮流実績(Actual flow).
Notes:
- All of the transmission lines except 北陸フェンス go from one region to another.
- For simplicity 北陸フェンス is mapped to be between Chubu and Hokuriku, even though occto has it as Chubu/Kansai->Hokuriku
- It is assumed that the reported transmission lines cover essentially all transmission between regions

Generation is then calculated for each region like in the case of Netherlands, by summing consumption and net export from all of the transmission lines. Since both consumption data and exchanges have the same 5-minute granularity, it is easy to join them on timestamp. For areas other than Kyushu, the JP parser returns all generation as unknown production.
- Idea is to show regions other Kyushu greyed out on the map, like New Brunswick (Canada)

For Kyushu, the electricity forecast page also has an estimate on solar power, but with no historical values available. Thus for Kyushu we take the most recent consumption/solar values, add most recent nuclear information and match that with the first value of the flow Chugoku->Kyushu with a timestamp earlier than or equal to the timestamp of consumption/solar. 
Assumptions:
- Nuclear information is roughly as recent as reported solar estimate and consumption
- Transmission line information typically does not arrive much later than reported consumption

NB! Quite often values are reported as 万kW, 万=10.000, hence there are multiplications by 10 to convert them to MW.

parsers/JP-KY.py
```diff
@@ -0,0 +1,156 @@
...
+    data['production']['nuclear'] = nuclear
+    # add the exchange JP-CG->JP-KY
+    exch_list = occtonet.fetch_exchange('JP-KY', 'JP-CG')
+    # find the exchange timestamp that agrees with the consumption one
+    exch_newer = True
+    ind = 0
+    while exch_newer:
+        ind = ind - 1
+        exch_newer = (exch_list[ind]['datetime'] > dt)
+    exch = -1*exch_list[ind]['netFlow']
```
jarek: If you would like to be more Pythonic, you could do something like:

```
# find nearest exchanges in time
nearest_exchanges = sorted(exch_list, key=lambda exch: abs(exch['datetime'] - dt))

# take the nearest exchange
# TODO: maybe check that exch['datetime'] is not too far off (15 min?), return [] if so
exch = -1 * nearest_exchanges[0]['netFlow']
```

parsers/JP.py
```diff
@@ -0,0 +1,132 @@
...
+
+def fetch_consumption(zone_key='JP-TK', target_datetime=None,
+                      logger=logging.getLogger(__name__)):
+    """
+    Returns the consumption for an area as a pandas DataFrame
+    """
+    datestamp = arrow.get(target_datetime).to('Asia/Tokyo').strftime('%Y%m%d')
+    consumption_url = {
+        'JP-HKD': "".join(['http://denkiyoho.hepco.co.jp/area/data/juyo_01_',
+                           datestamp, '.csv']),
```
jarek: for these you could do `'http://denkiyoho.hepco.co.jp/area/data/juyo_01_{}.csv'.format(datestamp)`

parsers/JP.py
```diff
@@ -0,0 +1,132 @@
...
+        'JP-HR': "".join(['http://www.rikuden.co.jp/denki-yoho/csv/juyo_05_', datestamp, '.csv']),
+        'JP-CB': 'http://denki-yoho.chuden.jp/denki_yoho_content_data/juyo_cepco003.csv',
+        'JP-KN': 'http://www.kepco.co.jp/yamasou/juyo1_kansai.csv',
+        'JP-CG': "".join(['http://www.energia.co.jp/jukyuu/sys/juyo_07_', datestamp, '.csv']),
+        'JP-SK': 'http://www.yonden.co.jp/denkiyoho/juyo_shikoku.csv'
+        }
+    if zone_key == 'JP-KN':
+        startrow = 44
+    else:
+        startrow = 42
```
jarek: I would add a brief comment explaining why this number of rows is being skipped (even if it's just "this is the weird format of the data sources")

parsers/JP.py
```diff
@@ -0,0 +1,132 @@
...
+    df = pd.read_csv(consumption_url[zone_key], skiprows=[x for x in range(startrow)],
```
jarek: `skiprows=list(range(startrow))`

jarek:
And one more small tech tip. If you name your files as .geojson rather than .json, Github will show a graphical view of them overlaid on a map, allowing for quick preview and validation.

jarek:
Oh no! Looks like changing the extension to .geojson broke the build script.

Can someone more adept in Javascript look to see if that's resolvable? It's complaining about `Unexpected token : ` at `web/generate-geometries.js:30:5` which seems to be the `require('./third_party_maps/JP-CB.geojson'),` line. Can this be made to work?

The preview in GitHub https://github.com/tmslaine/electricitymap/blob/0cb4103db8788d4c1ceb31767e376ca36ff5dd7a/web/third_party_maps/JP-KY.geojson is great if it can work with the build...

corradio:
Hi @tmslaine,
Thank you for the explanation. Can you describe what feeds are real-time compared to forecasted? 
I'm a bit wary adding forecasted data to the electricityMap. It would be better to only have measured data (nuclear, consumption and solar).

corradio:
@jarek maybe we shouldn't read file with `require`. Instead we could do something like `JSON.parse(fs.readFileSync(filepath))`?

corradio:
Terrific! I'll add an issue to add the forecasted values.
As soon as the tests pass I think we can merge.

alixunderplatz:
@tmslaine What does the solar data for Kyushu look like? Is it a smooth or a bumpy graph?

tmslaine:
Solar is the red line, in February there was some noise:
image
Also today the curve is not a perfect bell:
image

It seems that prices for all of the regions except Okinawa are available:
http://www.jepx.org/market/ and clicking スポット市場取引結果(spot results) leads to http://www.jepx.org/market/excel/index_2018.csv with columns 7-15 having the area prices(yen/kWh).

corradio:
Great! Let's fix the geojson issue and let's merge!

tmslaine:
Thanks @jarek for fixing the geojson issue!

I wonder if it makes sense to adjust JP-KY unknown generation CO2 intensity before merging.
Japanese Ministry of Economy, Trade and Industry has statistics for the fiscal year 2017 (Apr 2017-Mar 2018) for generation in the 7 prefectures that make Kyushu (Table 2-(2), rows 44-50) and the national generation by fuel (Table 2-(1)) [link](http://www.enecho.meti.go.jp/statistics/electric_power/ep002/results_archive.html)

From table 2-(1) I get 56% gas, 39% coal, 5% oil for fossil generation in Japan,
from table 2-(2) I get 86% fossil, 3.3% bio, 8.5% hydro, 1.5% geothermal, 0.7% wind for the non-solar, non-nuclear generation in Kyushu. Thus the unknown CO2-intensity would be roughly 550gCO2eq/kWh.

corradio:
@tmslaine feel free to submit another PR (and put @brunolajoie as reviewer) for the specific CO2 intensities. In the interest of time I'll merge this one now.

PR #1564, opened 2018-08-17
bright/darkmode cookie fix

### LINKED ISSUE #1534: Darkmode setting is not remembered as cookie

ovbm:
Fixes #1534 

brightModeEnabled is enabled and application state now reads the cookie and sets state accordingly, I tried running this locally in chrome but it didn't open in darkmode after closing tab and reopening... any ideas?

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -725,24 +725,15 @@ window.retryFetch = () => {
 // cause events to be emitted
 
 // BrightMode
-const electricityMapHeader = d3.select('#header-content');
-const tmrowWatermark = d3.select('#watermark');
-
 function toggleBright() {
   dispatchApplication('brightModeEnabled', !getState().application.brightModeEnabled);
-  electricityMapHeader.classed('brightmode', !electricityMapHeader.classed('brightmode'));
-  tmrowWatermark.classed('brightmode', !tmrowWatermark.classed('brightmode'));
 }
-
 d3.select('.brightmode-button').on('click', toggleBright);
-
 const brightModeButtonTooltip = d3.select('#brightmode-layer-button-tooltip');
-
 if (!getState().application.isMobile) {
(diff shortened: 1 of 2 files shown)


web/src/reducers/index.js
```diff
@@ -12,7 +12,7 @@ const initialApplicationState = {
   callerLocation: null,
   clientType: window.isCordova ? 'mobileapp' : 'web',
   colorBlindModeEnabled: Cookies.get('colorBlindModeEnabled') === 'true' || false,
-  brightModeEnabled: Cookies.get('brightModeEnabled') === 'true' || true,
+  brightModeEnabled: Cookies.get('brightModeEnabled') === 'true' || false,
```
corradio: you're defaulting to false here. I don't think we want that?

corradio:
hmm so the state has brightModeEnabled = false?

ovbm:
Ah, i misinterpreted the code.

corradio:
@ovbm I've fixed the cookies. However, now, when starting in brightmode, some text is still black instead of white:

image

web/src/main.js
```diff
@@ -1383,11 +1370,17 @@ observe(state => state.application.legendVisible, (legendVisible) => {
 // Observe for left panel collapse
 observe(state => state.application.isLeftPanelCollapsed, (_, state) =>
   renderLeftPanelCollapseButton(state));
-
 // Observe for search query change
 observe(state => state.application.searchQuery, (searchQuery, state) => {
   zoneList.filterZonesByQuery(searchQuery);
 });
+// Observe for brightmode change
+observe(state => state.application.brightModeEnabled, (brightModeEnabled, _) => {
```
corradio: @ovbm This is the key change: I had to observe for a redux state change (before, we dispatched to redux but didn't listen to state changes)

PR #1568, opened 2018-08-19
Update German numbers of netto installation

Bohne13:
Updated the numbers of netto installed electricity in Germany.
My source came from "Frauenhofer ISE" the  data is from July 2018

Source link: https://www.energy-charts.de/power_inst.htm?year=2018&period=annual&type=power_inst

PR #1569, opened 2018-08-19
Updated production capacity data source of Germany

Bohne13:
Fixes #1568 
Changed data Source for production capacity from Germany like i did in the "zones.json


jarek:
Merged to complete #1568

PR #1571, opened 2018-08-19
Added missing german strings

jld3103:


alixunderplatz:
@jld3103 Thanks for filling the missing strings! I left some comments for improvements ;)

PR #1573, opened 2018-08-20
Updated dependencies and Docker

jld3103:


mockserver/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python
```
corradio: can you force a version? Else it will fall back on python:latest which might break suddenly.

web/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.5.0
+FROM node
```
corradio: Same here

corradio:
Hi @jld3103.
Sorry for not being clear enough. What I meant is that auto-updating to the latest version of node/python might cause some unforeseen bugs that would lie either in the code we wrote (not very probable) or in the dependencies we use (most probable).
That is why dependencies should always be frozen in order to have deterministic builds (another example is how npm includes package-lock.json which freezes dependencies -- same with yarn and yarn.lock).
That being said I'm happy to update the version of python/node, but we should specify it and not have it auto-update to the latest.

PR #1574, opened 2018-08-20
Auto-generated translations for missing strings

jld3103:
The translations are made using deepl. I checked most of the translations and they are all very very good.  
The script merges the English translation and the other translation and the translates all strings which were not translated before in the other language.

### DIFF
--- a/web/generate-translations.js
+++ b/web/generate-translations.js
@@ -1,0 +1,97 @@
+const {translate} = require('deepl-translator');
+const deepmerge = require('deepmerge');
+const fs = require('fs');
+
+const translatable = ['de', 'fr', 'es', 'it', 'nl', 'pl'];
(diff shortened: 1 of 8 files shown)


Raffox97:
Italian translations are pratically all out of context, please exclude them from the commits.

jld3103:
The German translations were good, so I thought the others would be good too.

alixunderplatz:
Hi @jld3103 
The zone/country translations in this commit are partially very, very weird and wrong for the German translation as well. 
I think having something translated by a machine is no option for files with a structure like ours, where many words are out of context. This needs to be done by humans, even if it is really annoying and time consuming, else it will lead to a wrong translation.

brunolajoie:
Many thanks for the try @jld3103!

PR #1577, opened 2018-08-21
Add JP prices

### LINKED ISSUE #1553: Add JP prices
See https://github.com/tmrowco/electricitymap-contrib/pull/1543#issuecomment-411281685 by @tmslaine

lorrieq:
closes #1553 

**Please review:**
- takes prices of current day and next day (relative to `target_datetime`).
- prices are half-hourly rather than hourly DAM that we're used to in Europe? Can we confirm this is how it is in Japan?
- timezone set to `Asia/Tokyo` which I have assumed is correct.
- column names / order is hardcoded, didn't want to deal with the Japanese symbols ;-)

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -3,6 +3,7 @@
 import logging
 # The arrow library is used to handle datetimes
 import arrow
+import datetime as dt
 import pandas as pd
 from . import occtonet
 
@@ -16,6 +17,8 @@ from . import occtonet
 # JP-SK  : Shikoku
 # JP-KY  : Kyushu
 # JP-ON  : Okinawa
+# JP-CG  : Chūgoku
+
 
 def fetch_production(zone_key='JP-TK', session=None, target_datetime=None,
                      logger=logging.getLogger(__name__)):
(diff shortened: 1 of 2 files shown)


parsers/JP.py
```diff
@@ -118,15 +121,55 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
     return df
 
+
+def fetch_price(zone_key='JP-TK', session=None, target_datetime=None,
+                logger=logging.getLogger(__name__)):
+    if target_datetime is None:
+        target_datetime = dt.datetime.now() + dt.timedelta(days=1)
+
+    url = 'http://www.jepx.org/market/excel/spot_{}.csv'.format(
+        target_datetime.year)
```
tmslaine: The year in the source url does not refer to calendar year, but to Japanese fiscal year, starting on Apr 1st. For example the the 2017 csv has prices from 2017/04/01 to 2018/03/31. Thus target_datetime needs to be shifted three months back before extracting year.

tmslaine: Based on the graph of system price at http://www.jepx.org/ and the fact that it matches the last available day, column 6, of the csv source I am quite sure that the prices are half-hourly.

lorrieq:
Well spotted @tmslaine, that's now accounted for!

PR #1448, opened 2018-06-09
Split ES-IB [work in progress]

alixunderplatz:
This is an attempt for a parser splitting the generation on the Balearic islands.
ref #1347 
 
As usual, feel free to comment and improve diectly, so we can move on to the next steps.

I copied and adapted most of the code from ES_CN. 

Some things are commented out (and in In fetch production, there is no "logger=logging.getLogger(__name__)" like in ES-CN) because I had local issues and I'm far, far away from being an expert in terms of Python ;)

PR #1580, opened 2018-08-22
Split ES-IB

blackleg:
Continue the work of @alixunderplatz at https://github.com/tmrowco/electricitymap-contrib/pull/1448, I fixed tests and add support to the old parser.

### DIFF
--- a/parsers/ES_IB.py
+++ b/parsers/ES_IB.py
@@ -5,7 +5,7 @@ from arrow import get
 from requests import Session
 from ree import (Formentera, Ibiza,
                  Mallorca, Menorca,
-                 BalearicIslands, IberianPeninsula)
+                 BalearicIslands)
 
 from .lib.exceptions import ParserException
 from .lib.validation import validate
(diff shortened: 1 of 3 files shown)


parsers/ES_IB.py
```diff
@@ -1,114 +1,209 @@
...
 
-    # TODO: Remove verify SSL config when working without it.
-    responses = BalearicIslands(ses, verify=False).get_all()
-
-    if not responses:
-        raise ParserException("ES-IB", "No response")
-    else:
+## biomass could probably be added as response.biomass in ree, including value['resid'] which I assume is for residuos=waste. Atm it seems unincluded.
+## I saw generation from "genAux" (generacion auxiliar) today on Formentera, which should probably be added to response.other in ree
+## Formentera mostly only has some solar generation during the day, importing from Ibiza all of the time, which probably has to be considered for response.production() > 0: ?
```
corradio: what about removing the condition then?

PR #1582, opened 2018-08-25
Added translation helper

jld3103:
Look [here](https://github.com/jld3103/electricitymap-contrib/blob/translation-helper/web/locales/README.md) for information.


corradio:
Awesome!

PR #1587, opened 2018-08-29
fix JP price parser

### LINKED ISSUE #1585: Decoding error in JP price parser
`JP.fetch_price` is failing due to a `UnicodeDecodeError` in `pandas.read_csv`, the CSV file contains Japanese characters. I don't get this issue locally and saw it for the first time on the Kibana logs. 

Maybe it's as simple as passing an `encoding` parameter to `pandas.read_csv`. But I'm using the same version of `pandas` that's specified in the `requirements.txt` file, so maybe that's not it. I can't replicate the issue locally. Any ideas?

maxbellec:
fixes #1585 
issue was identified by @lorrieq and fixed by @jarek, I'm only copying his code here

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -133,7 +133,7 @@ def fetch_price(zone_key='JP-TK', session=None, target_datetime=None,
     else:
         fiscal_year = target_datetime.year
     url = 'http://www.jepx.org/market/excel/spot_{}.csv'.format(fiscal_year)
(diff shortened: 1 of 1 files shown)


maxbellec:
works on mine also. There are no reason it wouldn't work on another one if it's the correct encoding. Anyway the parser is currently broken, we can't break it more

PR #1589, opened 2018-08-31
Time machine server.js

maxbellec:
for the premium subdomain, adds basic auth and replies with a `set-cookie` header so that back in time works

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -160,19 +160,21 @@ app.get('/', (req, res) => {
     const { locale } = res;
     const fullUrl = 'https://www.electricitymap.org' + req.originalUrl;
 
-    // basic auth and time machine cookie if we're in the back in time container
-    if (process.env['BASIC_AUTH_USERNAME;']){
-      let user = auth(req);
+    // basic auth for premium access
+    if (process.env.BASIC_AUTH_CREDENTIALS) {
+      const user = auth(req);
       let authorized = false;
-      if (user){
-        let ix = process.env['BASIC_AUTH_USERNAMES'].split(';').indexOf(user.name);
-        if (ix !== -1 && user.pass === process.env['BASIC_AUTH_PASSWORDS'].split(';')[ix]){
-          authorized = true;
-        }
+      if (user) {
+        process.env.BASIC_AUTH_CREDENTIALS.split(',').forEach((cred) => {
+          const [name, pass] = cred.split(':');
+          if (name === user.name && pass === user.pass) {
+            authorized = true;
+          }
+        });
       }
-      if (!authorized){
+      if (!authorized) {
         res.statusCode = 401;
-        res.setHeader('WWW-Authenticate', 'Basic realm="example"');
+        res.setHeader('WWW-Authenticate', 'Basic realm="Premium access to electricitymap.org"');
         res.end('Access denied');
         return;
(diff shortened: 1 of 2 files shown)


web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
     }
     const { locale } = res;
     const fullUrl = 'https://www.electricitymap.org' + req.originalUrl;
+
+    // basic auth and time machine cookie if we're in the back in time container
```
corradio: Can we change the comment to something even a contributor could understand
Something along the lines of:
// Basic auth for premium access

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
...
+    if (process.env['BASIC_AUTH_USERNAME;']){
+      let user = auth(req);
+      let authorized = false;
+      if (user){
```
corradio: Style: linter probably requires space between ) and {

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
     }
     const { locale } = res;
     const fullUrl = 'https://www.electricitymap.org' + req.originalUrl;
+
+    // basic auth and time machine cookie if we're in the back in time container
+    if (process.env['BASIC_AUTH_USERNAME;']){
+      let user = auth(req);
```
corradio: Linter will probably say it needs `const` instead.

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
     }
     const { locale } = res;
     const fullUrl = 'https://www.electricitymap.org' + req.originalUrl;
+
+    // basic auth and time machine cookie if we're in the back in time container
+    if (process.env['BASIC_AUTH_USERNAME;']){
```
corradio: why does the env variable name have a semicolon at the end?

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
...
+      let user = auth(req);
+      let authorized = false;
+      if (user){
+        let ix = process.env['BASIC_AUTH_USERNAMES'].split(';').indexOf(user.name);
```
corradio: More traditional to split using a comma

corradio: You can be more explicit by naming a variable `hasMatchingUserName` and directly do the `!== -1` comparaison here.

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
...
+        if (ix !== -1 && user.pass === process.env['BASIC_AUTH_PASSWORDS'].split(';')[ix]){
```
corradio: It's a tad dangerous because this forces us to have two lists (one for users, one for passwords), and is therefore prone to errors (if you offset one item for example).
A better way would be to serialise auth using `user1:pass1,user2:pass2`...

web/server.js
```diff
@@ -158,9 +159,28 @@ app.get('/', (req, res) => {
...
+          authorized = true;
+        }
+      }
+      if (!authorized){
+        res.statusCode = 401;
+        res.setHeader('WWW-Authenticate', 'Basic realm="example"');
```
corradio: example?

Issue #1596, opened 2018-09-11
Incomplete data being reported for Great Britain


bmckinlay:
As much as I would like the UK's grid to be 90% renewable, this doesn't seem correct. Data for gas & nuclear production missing?

<img width="478" alt="screen shot 2018-09-11 at 12 21 20" src="https://user-images.githubusercontent.com/13368547/45357241-a4046d00-b5bd-11e8-8c25-317b0f9ba7c4.png">

systemcatch:
Yes this is clearly wrong, unfortunately it looks like the quality filters haven't caught it so far.
![gb_drop](https://user-images.githubusercontent.com/30196510/45357599-9dc2c080-b5be-11e8-9270-bf2fd09f53b8.png)

corradio:
It seems like the validation code from the ENTSOE parser (https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTSOE.py#L293) hasn't been ported to the ELEXON parser:
https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ELEXON.py#L210
Someone fancy adding that check?

corradio:
This shouldn't happen anymore after the PR we merged.

PR #1597, opened 2018-09-11
Add validation to GB production

lorrieq:
The fuel values are set to zero when "missing" rather than not being returned by the source. Therefore the main validation is the `expected_range` check so please review the ranges I have included.

issue #1596

### DIFF
--- a/parsers/ELEXON.py
+++ b/parsers/ELEXON.py
@@ -20,6 +20,8 @@ import datetime as dt
 import pandas as pd
 from io import StringIO
 
+from .lib.validation import validate
+
 ELEXON_ENDPOINT = 'https://api.bmreports.com/BMRS/{}/v1'
 
 REPORT_META = {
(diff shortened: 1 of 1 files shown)


corradio:
@lorrieq any chance we can change the parser and set the fuel values to null when missing?

lorrieq:
Yes, I can make that change (this assumes the required fuels coal, gas and nuclear will never have an actual zero value because if so it will be set to `null`)

corradio:
Out of curiosity, why doesn't the data source differentiate between null
and 0?

lorrieq:
I don't know but will contact them about this

systemcatch:
I'm not sure coal should be a required type given there's often times when it doesn't run.

jarek:
If we can't distinguish between coal _missing_ and coal actually being 0 MW (which happens in UK sometimes but not all the time) then we cannot use it as a criterion.

corradio:
Indeed. Let's wait for reply from @lorrieq's message then. It should be doable to distinguish 0 and null in the datasource.

lorrieq:
Hey, I got a response from ELEXON. They said the issue occurred because of a "National Grid Balancing Mechanism" outage, which is where they get the data from. They linked me to a [schedule](https://bmreports.com/bmrs/?q=transmission/systemwarning) of these.

They mentioned their system does not distinguish between zero values and missing values and the 'missing' data is not even corrected retrospectively after an outage... Quite disappointing.

corradio:
Ok that's for having that info. What I suggest is to lower the coal lower bound to 0 (because they sometimes turn off the coal) but aside from that we should be good with this PR

lorrieq:
OK, at least when the zero value for coal means data is "missing" then the datapoint will be rejected because nuclear and gas will be zero/missing too. Otherwise it's accepted. Nice to tie this down but hopefully their system outages are rare from now on.

PR #1603, opened 2018-09-17
Raise when CL_SING production is constant

### LINKED ISSUE #1356: Chile (CL-SING): total generation stuck at fixed value
The total generation of the CL-SING zone is stuck at 1 939 MW. The renewable part varies, but the "unknown" category always fills the gap. The issue is on the data provider's side.

image

image

maxbellec:
This will raise an error when, in the last 48 hours, at least 3 of them have the same total production. 
Fixes #1356


parsers/CL_SING.py
```diff
@@ -203,6 +203,16 @@ def fetch_production(zone_key='CL-SING', session=None, target_datetime=None, log
         }
         production_mix_by_hour.append(production_mix)
 
+    # if production is the same for at least 3 in the last 48 hours, it's
+    # considered constant. Using round because of floating-point precision.
+    c = Counter([round(sum(v for _, v in e['production'].items()), 3)
+                 for e in production_mix_by_hour])
+    most_common = c.most_common(1)[0]
+    if most_common[1] > 3:
```
systemcatch: Maybe we should add the number of occurrences to the error message.

maxbellec: it is in the message, just two lines below

maxbellec:
yep the source is hourly

parsers/CL_SING.py
systemcatch: Sorry I missed that, looks good to go.

parsers/CL_SING.py
```diff
@@ -203,6 +203,16 @@ def fetch_production(zone_key='CL-SING', session=None, target_datetime=None, log
...
+        raise ValueError(
+            f'Detected constant production in CL-SING. Value {most_common[0]} '
+            f'occured {most_common[1]} times in the last 48 hours.')
```
jarek: I think you need to remove the initial `f` on the second line of the string continuation for the syntax to work

maxbellec: Strings are computed separately then concatenated, so it's right this way.

```python
a, b = 0, 1

(f'this is a test of {a}'
 f' and {b}')
>>> 'this is a test of 0 and 1'

(f'this is a test of {a}'
 ' and {b}')
>>> 'this is a test of 0 and {b}'

corradio:
@maxbellec status?

corradio:
@maxbellec I'll let you merge

maxbellec:
thanks for pinging me!

Issue #896, opened 2017-12-06
Orkney Islands


systemcatch:
See @jarek 's comment in #596.

There is live generation data at http://anm.ssepd.co.uk/.  They have a connector to the UK mainland and lots of wind generation.  There are Wave and Tidal energy test facility's but I think it's mainly for R&D rather than regular power generation at the moment.  However there is serious potential for that to change in the future.

The Islands are very committed towards renewable energy, innovation and sustainability.
http://www.orkney.com/invest/sectors/energy
http://www.oref.co.uk/

Bjarkigd:
If memory serves, then Orkney does from time to time produce even more than they can export, as the connection to the mainland can't handle it.

alixunderplatz:
Live generation data has moved to https://www.ssen.co.uk/ANM/

pascalheraud:
OK, I will start on this.

pascalheraud:
For Orkney Islands, the point is to display it on map. The code seems to be GB-ORK but I'm not sure coords are included into naturalearth. Any idea ?

jarek:
In the Natural Earth shapefiles, the Orkneys have their own shape, tagged with:
code_hasc = GB.MO
adm1_code = GBR-2744
gn_a1_code = GB.V9

Extracting them by one of those in topogen.sh should be doable.

While we are looking at the map, it might also be good to separate out Shetland from mainland GB, as there is no grid connection. The codes are GB.AR, GBR-2747, and GB.W3 respectively.

corradio:
Any news here @pascalheraud ?

systemcatch:
I'm going to take a look at getting this one up and running. Can you guys figure out some of the carbon intensities we should be using here.

systemcatch:
Quick question, can we create the GB shape from a mix of subunit and state shapes? There's 80+ adm1 regions which is going to be a pain to work with. It'd be nice to keep the England & Wales subunits and just build Scotland from regions.

corradio:
there's no reason why we couldn't code it that way!

systemcatch:
Reading the latest Energy audit from http://www.oref.co.uk/resources/orkney-energy-audit/ it mentions the following capacity figures.

Wind - 48MW
Solar - 1.2MW
Wave - research only at the moment but leases held for 550MW in future.
Tide - same as above for 530MW.

@alixunderplatz any idea for their carbon intensity?

Issue #1322, opened 2018-04-10
Improve developer documentation


systemcatch:
Few points to start with, feel free to add more.

- Getting nodejs installed and working correctly. :sob: 
- Installing python 3, Windows does not ship with it by default I think.
- Can topogen.sh be run on powershell or does it require bash for Windows?

corradio:
In principle that's why all of this should be done inside Docker. It simplifies things quite significantly! Although Docker can be a bit annoying to install..

jbuverud:
We could add something about docker's system requirements. It only runs on certain versions of windows 10 (does not run on Home editions, for instance). I had to resort to running it through a VM install on my laptop.

systemcatch:
@corradio That was my idea initially but afaik docker only works when building the map. To add new geometries you have to have a working node install.

Yes it would be nice to add something about using docker on windows. I know @alixunderplatz is currently trying to get the map built on his Windows machine but is having difficulties.

corradio:
Actually I think we made the changes required to be able to do all the
mapping inside the container.
It's not really documented though.

systemcatch:
@corradio if you can give me a rough guide to mapping in the container I will add a more polished version to the docs.

corradio:
I think it's as simple as `docker-compose run --rm web ./topogen.sh` but haven't tried it

corradio:
@systemcatch did it work?

systemcatch:
Running it on my machine worked. To be sure we'd need to run this on a fresh VM but it's a good sign. @jbuverud fancy having a go?
```
(EM-env) chris@ThinkPad:~/electricitymap$ docker-compose run --rm web ./topogen.sh
Starting electricitymap_mockserver_1 ... 
Starting electricitymap_mockserver_1 ... done
Downloading ne_10m_admin_0_map_subunits.zip
Warning: Illegal date format for -z, --timecond (and not a file name). 
Warning: Disabling time condition. See curl_getdate(3) for valid date syntax.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 5324k  100 5324k    0     0   157k      0  0:00:33  0:00:33 --:--:--    92
Archive:  build/ne_10m_admin_0_map_subunits.zip
  inflating: build/ne_10m_admin_0_map_subunits.README.html  
 extracting: build/ne_10m_admin_0_map_subunits.VERSION.txt  
[...]

PR #1615, opened 2018-09-27
Seperate Orkney Islands from Great Britain

systemcatch:
Things to do:

- [ ] Compute a default carbon intensity for the islands.
- [ ] Find capacity information.
- [ ] Generate shapes from a combination of subunits & states.
- [ ] Check it all works by running the mockserver.

ref #896


systemcatch:
@corradio so in terms of what needs to happen here, I can add the capacities & compute a carbon intensity. But I'll need someone else to alter to code so we can make shapes from a combination of subunits and states.

corradio:
@systemcatch I'm quite busy right now so not sure I'll be able to do it. However you should be able to edit the existing function that filters geometries. Once you have a list, you can simply call `geomerge` on them to merge them all.

systemcatch:
Hey @jarek that would be really cool if you'd help with that, I'd be happy to test by pulling from one of your branches.

Regarding your node environment could #1322 help?

corradio:
@maxbellec maybe you can help here if you have time?

maxbellec:
I've also separated Shetland islands, since @systemcatch tells me they're not connected to GB. We do not have a source yet, so it will stay grey for the moment

systemcatch:
Very nice @maxbellec! I'm trying to build the map but keep getting the following error. You could try running the parser and seeing if it displays ok.

```shell
(EM-env) chris@ThinkPad:~/electricitymap$ docker-compose run --rm web npm run watch
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

Docker version 18.06.1-ce, build e68fc7a
docker-compose version 1.16.1, build 6d1ac21

maxbellec:
I have `docker-compose==1.21.2` and `Docker version 1.13.1, build 092cba3`

web/generate-geometries.js
```diff
@@ -682,16 +681,25 @@ zoneDefinitions.forEach(zone => {
       'only once in zoneDefinitions');
   // Avoid zone with moreDetails
   if ( !('moreDetails' in zone) || !zone.moreDetails) {
-  	zones[zone.zoneName] = getDataForZone(zone, true);
+    zones[zone.zoneName] = getDataForZone(zone, true);
   }
 });
 
+// special case for GB, as we separate two islands from it
```
maxbellec: @corradio it's the only zone for which we do manual manipulation. Do you think that code is OK? At least it works fine.

systemcatch:
@maxbellec got it working (old docker-compose) and everything looks ok. Are you happy about the line between England and Scotland?

jarek:
Great job @maxbellec , thanks for sorting out the working code.

Though it is kind of strange that line is there between England and Scotland and not for example between England and Wales, I don't really understand why. We should be able to `geomerge` zones that have already been geomerged, I mean `geomerge(geomerge(...scotland_parts), geomerge(england, wales))`... is something not working there?

maxbellec:
hmm good point about the line, I'll try to understand why that happens

maxbellec:
I don't know how to remove the line. I also have it when doing a single `geomerge`. @corradio any idea here?

```javascript
zones['GB'] = geomerge(...allGeos.filter(d => (
  d.properties.subid === 'ENG'
  || d.properties.subid === 'WLS'
  || ((d.properties.gu_a3 === 'SCT') && (d.properties.adm1_code !== 'GBR-2747') && (d.properties.adm1_code !== 'GBR-2744'))
  )));
```

parsers/GB_ORK.py
```diff
@@ -0,0 +1,168 @@
...
+from bs4 import BeautifulSoup
+
+# There is a 2MW storage battery on the islands.
+# http://www.oref.co.uk/orkneys-energy/innovations-2/
+
+TZ = 'Europe/London'
+DATETIME_LINK = 'https://www.ssen.co.uk/ANM/'
+GENERATION_LINK = 'https://www.ssen.co.uk/Sse_Components/Views/Controls/FormControls/Handlers/ActiveNetworkManagementHandler.ashx?action=graph&contentId=14973&_=1537467858726'
+
+# TODO map correctly
```
corradio: Todo?

web/generate-geometries.js
corradio: let's move the code up to L236 then!

corradio:
`geomerge` simple does a union of shapes, so I think it will always create a line.
For example, SPP is the union of a certain number of states, and it shows lines between states:

image

Is there a previous situation where we merge items without any border? If yes, then we should look at how we did it for this case.

parsers/GB_ORK.py
systemcatch: Not anymore I'll remove it, are you happy with the carbon intensity I've assigned to the Orkney's?

corradio: @brunolajoie should answer that, he's more qualified than I am

maxbellec:
:tada: 
Now waiting for the Queen's phone call to thank me for saving peace in the UK.
screenshot at 2018-10-11 18-56-58

maxbellec:
FYI the script I used to remove all GeoJSON data is SCT above a certain latitude: 

```python
coords = sct['geometry']['coordinates']

new_coords = []
for coord in coords:
    assert len(coord) == 1
    coord = coord[0]
    
    new_coord = []
    append = True
    for x, y in coord:
        if y > 58.7:
            append = False
            new_coord.append([x, y])
    
    if append:
        new_coords.append([coord])

sct['geometry']['coordinates'] = new_coords

sct['properties']['subid'] = 'SCT-no-islands'

import json

with open('/home/m/temp/sct-no-islands.json', 'w') as f:
    f.write(json.dumps(sct))
```

where `sct` is a `json.loads` of the line where `properties.subid == 'SCT'` from `tmp_countries.json`

maxbellec:
I think we're just waiting for @brunolajoie to confirm carbon intensity

corradio:
Nice hack @maxbellec ! However I don't think this is the way to go, because having a 3rd party geom for SCT will be problematic as soon as we will split GB (we will have to regenerate the files, keep knowledge of your script etc..).
I don't think it's a problem that the line is there TBH. It will motivate people even more to open data up about GB (because we know there's data!)

maxbellec:
We'd still use `sct-no-islands` instead of SCT after the split, wouldn't we?

brunolajoie:
@systemcatch what does the parser returns as production mix? Only the total renewable generation? Is there any backup system ?

brunolajoie:
Yeah, looking at the live bar chart right now, and comparing with the load, it seems that we double count if we add ANM + Renewable generation, no? 
Or are they exporting about as much as their load, right now? Can you check comparing API load / export / generation results?

jarek:
Right now (site says 14 October 2018 20:24:49, my time 21:26...) I see the following:

Orkney Demand - Live Demand 16.88 MW
Orkney Generation - Orkney ANM 2.983 MW
Orkney Generation - Non-ANM Renewable Generation 7.391 MW
Orkney Demand - Winter Peak Demand 35.70 MW
Orkney Generation - Total Renewable Capacity 57.10 MW

So current generation would be 10.374 MW and demand 16.88 MW, with balance being imported.

Note that the chart includes peak demand and total production capacity, it has thrown me off more than once as well.

We don't have a separate source for import/export, it is calculated from demand minus generation.

brunolajoie:
Ok so clearly we need to add both ANM and non-ANM-renewable indeed.
But we need to find a way to understand what fuel is the ANM? Could it be a storage system?

PR #1617, opened 2018-09-28
included close button for modal, close modal on outside click, restyl…

### LINKED ISSUE #1616: button to discard onboarding
We should be able to skip the onboarding with a single, easy to see button.
Optionally, the onboarding should disappear if we click in the space around it.

ovbm:
…ed mobile navigation buttons

fixes #1616

### DIFF
--- a/web/src/components/onboardingmodal.js
+++ b/web/src/components/onboardingmodal.js
@@ -86,8 +86,8 @@ export default class OnboardingModal {
   }
 
   _setupLayout() {
-    this._setupBackgroundOverlay();
     this._setupRootContainer();
+    this._setupBackgroundOverlay();
     this._setupLeftButton();
     this._setupBody();
     this._setupFooter();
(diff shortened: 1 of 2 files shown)


corradio:
@ovbm can you add a small screenshot?

ovbm:
screenshot 2018-10-01 at 08 32 40

maxbellec:
@ovbm how does the button look on mobile? @brunolajoie is that big enough for you?

ovbm:
screenshot 2018-10-01 at 08 51 36

brunolajoie:
Looks great, letting you merge!
FYI, on a client phone, last time i sent him the link, he had to scroll down to see the right arrow, which was not visible at first, which makes him almost close the website. Lets hope this arrow will fix that. thanks!

corradio:
@ovbm what about we make a floating next button so it's always visible? Thanks for reacting so quickly on this issue.

ovbm:
Yes, a fixed button would be better, but require more extensive restyling. I'll merge this and if we find that it's still a hurdle in the future I'll propose a more thorough redesign. I for example don't think the full screen is a good solution on mobile, because it really dampens the "wow" and discovery effect of seeing the website for the first time.

corradio:
Roger that. Gogo merge!

ovbm:
screenshot 2018-10-01 at 10 00 25

this is how I would prefer it

corradio:
works for me. The real issue is what happens when the screen is smaller than the onboarding window. We'd like to avoid having people scroll to press the arrow. However I already think this solves it, so all good from my side!

PR #1619, opened 2018-10-02
Update AT, DE, LU price areas

### LINKED ISSUE #1605: DE-AT-LU Split
Seen on the ENTSOE TP:

> Dear Users,
> 
>
> Please be informed, that we will start preparation of the platform for DE-AT-LU bidding zone split on the 23rd of August. BZN|DE-AT-LU will be separated into 2 new bidding zones BZN|DE-LU and BZN|AT.
> 
> New bidding zones will be active from the 1st of October, however, first data submissions, like month ahead forecasts, are expected from the 1st of September. 
> 
> Validity end date for BZN|DE-AT-LU is the end of September 2018. 
> EIC codes for new bidding zones:
> BZN|DE-LU - 10Y1001A1001A82H 
> BZN|AT - 10YAT-APG------L (same as control area or country)
> Thanks for your patience and understanding.
> Best Regards
> Transparency Platform Team.

Does this affect us? cc @brunolajoie @alixunderplatz

lorrieq:
closes #1605


parsers/ENTSOE.py
```diff
@@ -77,6 +77,7 @@
     'CZ': '10YCZ-CEPS-----N',
     'DE': '10Y1001A1001A83F',
     'DE-AT-LU': '10Y1001A1001A63L',
```
corradio: you can probably remove that one!

parsers/ENTSOE.py
```diff
@@ -164,11 +165,10 @@
 }
 # Some zone_keys are part of bidding zone domains for price data
 ENTSOE_PRICE_DOMAIN_OVERRIDE = {
-    'AT': ENTSOE_DOMAIN_MAPPINGS['DE-AT-LU'],
```
corradio: AT by default gives the 15min prices as @alixunderplatz noted. We should probably set the right code for the AT bidding zone I think? Or did ENTSOE update such that AT now shows the correct prices?

alixunderplatz: @corradio we can use bidding zone Austria, it has the correct prices now as you assumed.

image

alixunderplatz: 'CTY|10YAT-APG------L' should be the correct ENTSO-E code for the bidding zone

alixunderplatz: @lorrieq I just saw your sample output, everything looks fine there pricewise! :)

lorrieq:
Oh, the key ``DE-AT-LU`` is used in ``ENTSOE_EXCHANGE_DOMAIN_OVERRIDE``. Has anyone noticed problems with the exchanges? Or they're still using the old code? I can investigate if we're unsure.

lorrieq:
The DE-LU code seems to now be used for old DE-AT-LU exchanges, see 
[sample output](https://github.com/tmrowco/electricitymap-contrib/files/2466089/entsoe_new_exchanges.pdf).

However I can't work out what's going on with the AT->IT-NO exchange. If you try and view it on the transparency platform, one gets all n/e results for AT->IT-North and DE-AT-LU->IT-North. I think the API is not yet correctly providing this information and therefore I removed it from the override mapping until the issue is resolved (so that the tests pass and this can be merged in the meantime).

corradio:
OK I'll merge.
@lorrieq @brunolajoie @alixunderplatz what about we open an issue about missing AT->IT_NO data so we don't forget?

corradio:
thx

PR #1622, opened 2018-10-03
Using NODE_ENV in all possible places

denplis:
- Remove client environment dependency on url in browser
- Same environment for client and server
- Using injected by webpack `ELECTRICITYMAP_PUBLIC_TOKEN` instead of hardcoded for `development` env
- const instead of let's for identifiers which won't be reassigned

### DIFF
--- a/web/src/services/dataservice.js
+++ b/web/src/services/dataservice.js
@@ -11,19 +11,14 @@ var moment = require('moment');
 
 // API
 function protectedJsonRequest(endpoint, path, callback) {
-  let t = new Date().getTime();
-  let md = forge.md.sha256.create();
-  let s
-  if (endpoint.indexOf('localhost') != -1) {
-    s = md.update('development' + path + t).digest().toHex();
-  }
-  else {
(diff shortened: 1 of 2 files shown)


web/src/services/dataservice.js
```diff
@@ -11,19 +11,14 @@ var moment = require('moment');
...
-  if (endpoint.indexOf('localhost') != -1) {
-    s = md.update('development' + path + t).digest().toHex();
-  }
-  else {
-    s = md.update(ELECTRICITYMAP_PUBLIC_TOKEN + path + t).digest().toHex();
-  }
-  let req = d3.json(endpoint + path)
+  const now = new Date().getTime();
+  const md = forge.md.sha256.create();
+  const signature = md.update(ELECTRICITYMAP_PUBLIC_TOKEN + path + now).digest().toHex();
```
corradio: Here we should use the "development" token in dev mode

denplis: I know, you've already pass "development" token here https://github.com/tmrowco/electricitymap-contrib/blob/master/web/webpack.config.js#L46

`

corradio:
good catch!

Issue #1621, opened 2018-10-03
US-NEISO is throwing a keyerror


systemcatch:
Been happening for at least a week.
```python
Traceback (most recent call last):
  File "/home/feeder/lib/fetch_data.py", line 173, in launch_parsers
    objs = res.get(timeout)
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/feeder/lib/fetch_data.py", line 171, in <lambda>
    **parser_kwargs))
  File "/home/contrib/parsers/US_NEISO.py", line 136, in fetch_production
    points = production_data_processer(production_json)
  File "/home/contrib/parsers/US_NEISO.py", line 82, in production_data_processer
    production[generation_mapping[k]] += v
KeyError: 'Other'
```

PR #1625, opened 2018-10-05
Logging and mapping fixes for US-NEISO

systemcatch:
- Map 'Other' key to battery storage.
- Log and warn about new keys rather than throwing an error.
- Make sure data without timestamp is caught.

ref #1621


maxbellec:
letting you merge

PR #1629, opened 2018-10-06
Setup react-devtools in common way

denplis:


corradio:
Hi @denplis ,

Thank you for the updated PR. At this stage I feel like we do not need to bring in another dependency, which might inflate the bundle size and increase complexity. I'm actually trying to get rid of other dependencies as well :) The existing solution works well, so I do not see the need for a change at this stage.

Let me know if you disagree. We all appreciate your help on this nonetheless.

Olivier

denplis:
@corradio 
Thanks for the response. I see your point and don't want to spent to much your time for this topic.

My point is: project is already using `__REDUX_DEVTOOLS_EXTENSION__` and [ here is official helper for handle it with redux store](https://github.com/zalmoxisus/redux-devtools-extension#13-use-redux-devtools-extension-package-from-npm) ( Actually that is the same git repo as for chrome extension and npm package )
So my propose to use what redux_devtools author suggest and not to use ternary operator such as `const store = process.env.NODE_ENV === 'production' ?` which doesn't look elegant and brings a bit more complexity for understanding how the `redux store` is created. 

Don't worry about bundle size ( [~572 bytes](https://bundlephobia.com/result?p=redux-devtools-extension@2.13.5) )

Issue #1630, opened 2018-10-09
FR: key coal has negative value -9.0


corradio:
```
invalid point: {'zoneKey': 'FR', 'datetime': datetime.datetime(2018, 10, 9, 11, 15, tzinfo=tzoffset(None, 7200)), 'production': {'nuclear': 41740.0, 'coal': -9.0, 'gas': 4057.0, 'oil': 188.0, 'wind': 1158.0, 'solar': 2762.0, 'biomass': 861.0, 'hydro': 3366.0}, 'storage': {'hydro': -1024.0}, 'source': 'opendata.reseaux-energies.fr', 'schemaVersion': 1}, reason:FR: key coal has negative value -9.0
```

Probably a good idea to set small negative values to 0

jarek:
ENTSOE parser does this https://github.com/tmrowco/electricitymap-contrib/blob/27d9f7e8f1641b3b536e9bdb1bf9c1a1c1520041/parsers/ENTSOE.py#L887

Probably makes sense to adapt https://github.com/tmrowco/electricitymap-contrib/blob/926ee622abb0c22cd15a26044c6b218183aa7a52/parsers/lib/quality.py#L79

Using 50 MW or less than 0.1% of total production or something

PR #1632, opened 2018-10-09
quality.py: don't raise on small negative values

jarek:
Further to https://github.com/tmrowco/electricitymap-contrib/issues/1630#issuecomment-428150514

Is it acceptable to change the values of `obj` within the quality validators?

If not, if we don't raise exception on a small negative value here, will the back-end complain instead?

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -73,10 +73,18 @@ def validate_production(obj, zone_key):
         if not_allowed_keys:
             raise ValidationError('unexpected keys in storage: {}'.format(
                 not_allowed_keys))
+    total_production = sum(obj['production'].values())
+    negative_threshold = total_production * 0.001
     for k, v in obj['production'].items():
         if v is None:
             continue
         if v < 0:
+            if abs(v) < negative_threshold:
+                # let through slightly negative values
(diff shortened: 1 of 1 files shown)


corradio:
@jarek the validation functions do not return anything, so they shouldn't be used to mutate data.
However, what about we change those functions https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/lib/validation.py to be able to remove small values automatically?
cc @brunolajoie

systemcatch:
https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/lib/validation.py#L106 can already be used to remove negative values (replaces them with None). The limits could altered from -5.0/0 to something else.

corradio:
ah nice! Didn't remember this one. I'll try to keep it mind :)

corradio:
@jarek I'm closing this as we can't allow negative values in our backend (it would screw up a lot of computations unfortunately). Setting them to 0 is definitely the way, but it needs to happen at parser level (I think). Thank you however for the proposal, and for your continued efforts :)
I'm still thinking we need to implement this derivative filter in quality.py by the way, but time is very limited on my side atm.

PR #1639, opened 2018-10-16
higher gas max production in GB

maxbellec:


lorrieq:
Maybe we shouldn't allow the max of the expected range to be greater than the installed capacity? (which according to `zones.json` is 27688 for GB)

corradio:
It's a good idea but in many cases the capacity information is outdated so
I wouldn't trust it.

maxbellec:
last commit passes the logger down to the validation function, so that errors be shown in Kibana

PR #1640, opened 2018-10-17
quality: derivative filter

maxbellec:
Introducing a simple quality filter that checks that the difference between consecutive production values for a specific energy is not too big. Warns about and removes datapoints that are too different from the previous one.

It solves the problem of some production being wrong for a single datapoint, but it will remove the wrong datapoint AND the next one. Shoud we compare to rolling median instead of only previous point?

### DIFF
--- a/parsers/lib/validation.py
+++ b/parsers/lib/validation.py
@@ -53,8 +53,7 @@ def validate_production_diffs(
     for energy, max_diff in max_diff.items():
         if 'energy' == 'total':
             series = pd.Series(
-                [sum([nan_to_zero(v)
-                     for en, v in datapoint['production'].items()])
+                [np.nansum([v for v in datapoint['production'].values()])
                  for datapoint in datapoints])
         else:
             series = pd.Series(
--- a/parsers/lib/validation.py
+++ b/parsers/lib/validation.py
@@ -70,8 +70,8 @@ def validate_production_diffs(
                 (datapoints[i]['datetime'], datapoints[i]['production'][energy])
                 for i in wrongs_ixs_and_previous if i > 0]
             logger.warning(
-                f'some datapoints have a too high production value difference '
-                f'for {energy}: {to_display}')
+                'some datapoints have a too high production value difference '
+                'for {}: {}'.format(energy, to_display))
         ok_diff &= new_diffs
     # first datapoint is always OK
     ok_diff.iloc[0] = True
--- a/parsers/lib/validation.py
+++ b/parsers/lib/validation.py
@@ -52,7 +52,8 @@ def validate_diffs(datapoints: list, max_diff: dict, logger: logging.Logger):
         series = pd.Series(
             [datapoint['production'].get(energy, np.nan)
              for datapoint in datapoints])
-        new_diffs = np.abs(series.diff()) < max_diff
+        # nan is always allowed (can be disallowed using `validate` function)
+        new_diffs = (np.abs(series.diff()) < max_diff) | series.isna()
         if not new_diffs[1:].all():
             wrongs_ixs = new_diffs[~new_diffs].index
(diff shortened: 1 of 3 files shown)


parsers/lib/validation.py
```diff
@@ -27,6 +31,46 @@ def check_expected_range(datapoint, value, expected_range, logger, key=None):
...
+                (datapoints[i]['datetime'], datapoints[i]['production'][energy])
+                for i in wrongs_ixs_and_previous if i > 0]
+            logger.warning(
+                f'some datapoints have a too high diff for {energy}: '
+                f'{to_display}')
+        ok_diff &= new_diffs
+    # first datapoint is always OK
+    ok_diff.iloc[0] = True
+
+    return [datapoints[i] for i in ok_diff[ok_diff].index]
```
corradio: It would be nice to issue a warning with the values being dropped, and the value of the derivative (so we can easily increase the threshold if needed

parsers/lib/validation.py
```diff
@@ -27,6 +31,46 @@ def check_expected_range(datapoint, value, expected_range, logger, key=None):
...
+        new_diffs = (np.abs(series.diff()) < max_diff) | series.isna()
+        if not new_diffs[1:].all():
+            wrongs_ixs = new_diffs[~new_diffs].index
+            wrongs_ixs_and_previous = sorted(
+                {ix - 1 for ix in wrongs_ixs} | set(wrongs_ixs))
+            to_display = [
+                (datapoints[i]['datetime'], datapoints[i]['production'][energy])
+                for i in wrongs_ixs_and_previous if i > 0]
+            logger.warning(
+                f'some datapoints have a too high diff for {energy}: '
```
corradio: can we rename `diff` to `time difference` (or something alike)? It makes the error more explicit.

maxbellec: yes, we can make it more explicit, but it's the absolute difference in the production value we're testing here, not the time difference

maxbellec:
any idea about comparing to rolling median instead of only previous datapoint?

parsers/lib/validation.py
corradio: I'm not sure I understand: aren't you comparing the difference between production at time t and production at time t-1 for a specific production mode?

maxbellec:
@corradio I didn't understand on the phone if you thought only comparing t and t-1 is better or rolling median with (t-2, t-1, t+1, t+2)?

corradio:
I think we should have a derivative filter like you did, and then later we can have another type of filter that compares a smotthed signal (rolling median for example) to detect outliers.
The derivative filter is enough for now I think - unless you want to code the other one as another method :)

parsers/lib/validation.py
```diff
@@ -27,6 +32,53 @@ def check_expected_range(datapoint, value, expected_range, logger, key=None):
...
+    """
+    # sort datapoins by datetime
+    datapoints = sorted(datapoints, key=lambda x: x['datetime'])
+
+    ok_diff = pd.Series(np.ones_like(datapoints, dtype=bool))
+    for energy, max_diff in max_diff.items():
+        if 'energy' == 'total':
+            series = pd.Series(
+                [sum([nan_to_zero(v)
+                     for en, v in datapoint['production'].items()])
```
jarek: if you're not using `en` here, you can loop over `datapoint['production'].values()` instead of `items()`.
```suggestion
                     for v in datapoint['production'].values()])
```

Or by numpy.nansum... `[np.nansum(datapoint['production'].items())]` (untested)

brunolajoie:
If I understand it well, the filter right now is only applied for France, and needs to be called manually for all regions we want to apply it for, right?

If I set a max diff only for some fuel, it will not prevent the remaning ones to fluctuates, right?

maxbellec:
yes and yes!

PR #1643, opened 2018-10-17
Add JP forecasts(#1550)

### LINKED ISSUE #1550: Add JP forecasts
@tmslaine discovered forecasted data for Japan, which we could add!

See https://github.com/tmrowco/electricitymap-contrib/pull/1543

tmslaine:
Resolves #1550

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -69,7 +69,8 @@ def fetch_production(zone_key='JP-TK', session=None, target_datetime=None,
     # add a row to production for each entry in the dictionary:
     
     datalist = []
-    for i in range(df.shape[0]):
+    #for i in range(df.shape[0]):
+    for i in df.index:
         data = {
             'zoneKey': zone_key,
             'datetime': df.loc[i, 'datetime'].to_pydatetime(),
@@ -116,7 +117,7 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
         startrow = 44
     else:
         startrow = 42
-    df = pd.read_csv(consumption_url[zone_key], skiprows=list(range(startrow)),
+    df = pd.read_csv(consumption_url[zone_key], skiprows=startrow,
                      encoding='shift-jis')
     df.columns = ['Date', 'Time', 'cons']
     # Convert 万kW to MW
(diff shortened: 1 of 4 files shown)


parsers/JP.py
```diff
@@ -121,6 +126,63 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+    """
+    # Currently past dates not implemented for areas with no date in their demand csv files
+    if target_datetime and (zone_key in ['JP-TK', 'JP-CB', 'JP-KN', 'JP-SK']):
+        raise NotImplementedError(
+            "Past dates not yet implemented for selected region")
+        
+    datestamp = arrow.get(target_datetime).to('Asia/Tokyo').strftime('%Y%m%d')
+    consumption_url = {
+                   'JP-HKD': 'http://denkiyoho.hepco.co.jp/area/data/juyo_01_{}.csv'.format(datestamp),
+                   'JP-TH': 'http://http://setsuden.tohoku-epco.co.jp/common/demand/juyo_02_{}.csv'.format(datestamp),
```
lorrieq: Typo in this url, remove the first http://

parsers/JP.py
```diff
@@ -121,6 +126,63 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+                   'JP-KY': 'http://www.kyuden.co.jp/power_usages/csv/juyo-hourly-{}.csv'.format(datestamp),
+                   'JP-ON': 'https://www.okiden.co.jp/denki/juyo_10_{}.csv'.format(datestamp)
+                   }
+    # Skip non-tabular data at the start of source files
+    if zone_key == 'JP-KN':
+        startrow = 9
+    else:
+        startrow = 7
+    # Read the 24 hourly values
+    df = pd.read_csv(consumption_url[zone_key],skiprows=list(range(startrow)), nrows = 24, encoding = 'shift-jis')
```
lorrieq: `skiprows=startrow` does the same and is more intuitive

parsers/JP.py
```diff
@@ -121,6 +126,63 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+    if zone_key == 'JP-KN':
+        df = df[['DATE', 'TIME', '予想値(万kW)']]
```
lorrieq: I get a `KeyError` when I run this code, something about the japanese symbols. But it works for you? Any idea why it's not working on my end?

parsers/JP.py
```diff
@@ -121,6 +126,63 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+    df['datetime'] = df.apply(parse_dt, axis=1)
+    df = df[['datetime', 'fcst']]
+
+    # convert from 万kW to MW
+    df['fcst'] = 10*df['fcst']
+    # validate
+    df = df.loc[df['fcst']>0]
+    # return format
+    data = []
+    for i in range(df.shape[0]):
```
lorrieq: How about `for i in df.index`? This will make sure the `df.loc` below always works too.

parsers/JP.py
```diff
@@ -121,6 +126,63 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+        df = df[['DATE', 'TIME', '予想値(万kW)']]
+    else:
+        try:
+            df = df[['DATE', 'TIME', '予測値(万kW)']]
+        except KeyError:
+            df = df[['DATE', 'TIME', '予測値（万kW)']]
+    df.columns = ['Date', 'Time', 'fcst']
+
+    df['datetime'] = df.apply(parse_dt, axis=1)
+    df = df[['datetime', 'fcst']]
```
lorrieq: This causes a `SettingWithCopyWarning` for me on line 173. We could probably remove this line as the columns are selected by name in the loop below.

parsers/JP.py
tmslaine: Yes, it is working for me.
Does https://en.wikipedia.org/wiki/Help:Multilingual_support#East_Asian render correctly?

tmslaine:
The forecasts only apply to past dates and current day in local time, i.e. at most 0 to 24 hrs ahead. As a possible future improvement, some areas might be publishing D+1 consumption forecast at 18:00.
Since the parser for Japanese exchanges is able to use Japanese characters, I expect that there will not be issues with them in code.

PR #1646, opened 2018-10-19
Map new plants and fix logging in DO

systemcatch:
Logger wasn't being created when running as main.


corradio:
feel free to merge if tests pass

PR #1651, opened 2018-10-22
Prepare production mix toggle

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/1554

The goal here is to add a toggle that enables to switch between production view and consumption view.
For now I
- added a new redux key `electricityMixMode`
- added a `toggleElectricityMixMode` global function that you can trigger from the console
- updated the code everywhere

untitled9

@ovbm are you able to help with designing the toggle? As action you just need to trigger a method similar to https://github.com/tmrowco/electricitymap-contrib/compare/olc/add-production-mix-toggle?expand=1#diff-42c3276bc26032ecad9d7061dd401cd4R1470, and remove that global function.

### DIFF
--- a/web/src/components/countrytable.js
+++ b/web/src/components/countrytable.js
@@ -92,7 +92,6 @@ function CountryTable(selector, modeColor, modeOrder) {
         .style('fill', 'darkgray')
         .attr('transform', 'translate(1, ' + this.TEXT_ADJUST_Y + ')')
         .style('display', 'none');
-    
 }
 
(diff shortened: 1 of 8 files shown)


brunolajoie:
Awesome! Let's highlight also on the left panel, if we are viewing the map as "production-view" of "consumption-view" !

corradio:
@ovbm I'm merging and then you can make a PR for the UI itself

corradio:
Merged because you opened a new PR

Issue #1647, opened 2018-10-19
Punjab (India) data format has changed


systemcatch:
http://www.punjabsldc.org/pungenrealw.asp?pg=pbGenReal

These changes are resulting in the datetime, solar & thermal regexs no longer matching. @blackleg do you want to correct them?

lorrieq:
At the moment it looks like everything is ok again except there is no solar value provided. We can either:
- remove lines 41-43 in [`IN_PB.py`](https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/IN_PB.py) and set solar production to 0. 
- keep the regex for solar but catch the exception that occurs if there's no match and set `solar_value` to 0. Not sure if solar is gone for good from that source but this would still record intermittent values.

systemcatch:
I believe the solar generation is now available at http://www.punjabsldc.org/solarppW.asp?pg=solarppW

lorrieq:
Nice!

systemcatch:
@alixunderplatz good spot on the biomass, they'll need to be specifically mapped to type.. We should be able to use the "SUSPECT" status quite easily.

lorrieq:
closed by #1652

PR #1652, opened 2018-10-23
Get IN-PB solar from new source

lorrieq:
Filtering out units with BIOMASS in name and units with `SUSPECT` . Is this strict enough or do we need a specific biomass mapping?

How do we deal with the `datetime` where we read different fuels from different sources? Do we want to return the latest `datetime` across the two webpages? I have not parsed out any timestamps in the solar source yet but we could select the latest updated timestamp from the table if needed.

issue #1647

### DIFF
--- a/parsers/IN_PB.py
+++ b/parsers/IN_PB.py
@@ -25,7 +25,7 @@ def read_text_by_regex(regex, text):
     return date_text
 
 
-def get_solar(zone_key, session):
+def get_biomass_solar(zone_key, session):
     response_text = web.get_response_text(
         zone_key, 'http://www.punjabsldc.org/solarppW.asp?pg=solarppW',
         session)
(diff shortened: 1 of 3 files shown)


systemcatch:
Good stuff, all that's left is to deal with the timestamps in the solar/biomass data. We could compare each one to india_date on https://github.com/tmrowco/electricitymap-contrib/pull/1652/files#diff-7038aab9af0a61c166116ea2b61de64dL37 to see if it's current enough to use?

lorrieq:
Currently implemented to only include values from biomass and solar units updated later than 30 minutes before the datetime of the main parser but this threshold can be changed.

systemcatch:
@lorrieq merge away!

corradio:
Thank guys!

Issue #1645, opened 2018-10-19
Corsica is missing from Entsoe domain mappings


systemcatch:
If this is expected maybe we could add an exception to avoid cluttering up the logger with errors.

```python
Traceback (most recent call last):
  File "/home/feeder/lib/fetch_data.py", line 174, in launch_parsers
    objs = res.get(timeout)
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/feeder/lib/fetch_data.py", line 172, in <lambda>
    **parser_kwargs))
  File "/home/contrib/parsers/ENTSOE.py", line 992, in fetch_exchange
    domain1 = ENTSOE_DOMAIN_MAPPINGS[zone_key1]
KeyError: 'FR-COR'
```

lorrieq:
Looking at the domain mapping for `FR-COR->IT-CNO` says `FR-COR`'s code is `10Y1001A1001A893` (which is labelled as `Italy_Saco_DC` in the ENTSOE [guide](https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html)). 
Maybe we just need to add a mapping to `ENTSOE_EXCHANGE_DOMAIN_OVERRIDE`: 
`'FR-COR->IT-SAR': ['10Y1001A1001A893', ENTSOE_DOMAIN_MAPPINGS['IT-SAR']]`

PR #1655, opened 2018-10-26
Add exchange mapping for FR-COR->IT-SAR

lorrieq:
Related to #1645 

This assumes the *FR-COR->IT-SAR* exchange is represented by *BZN|IT-SACODC > BZN|IT-Sardinia* on the ENTSOE transparency website. (following the code for the already implemented *FR-COR->IT-CNO* exchange).

@corradio do you know why in PR #1604 this exchange was commented out? Perhaps I have missed something.


corradio:
It's commented because we have some uncertainty about the validity of the data. cc @alixunderplatz 
I think there's two connections: AC and DC.
@alixunderplatz what do you think?

lorrieq:
Ok, thanks. If still uncertain I can simply remove the exchange parsers for *FR-COR->IT-SAR* in the config file.

corradio:
yes that you can do to avoid having it make errors

lorrieq:
@corradio As discussed, I've removed this parser from the config file until we have more clarity over that exchange's data. I will merge once you confirm the blank `parsers` dictionary for `FR-COR->IT-SAR` in `exchanges.json` won't cause issues with the backend. Thanks!

Issue #1613, opened 2018-09-25
ES-IB: coal should be required?


corradio:
![image](https://user-images.githubusercontent.com/1655848/45994560-7ecf2e80-c094-11e8-947f-aa2525af88af.png)

brunolajoie:
Very hard to say. It happened a few time in the past, often seemed buggy, but sometime very close to being truely 0
![image](https://user-images.githubusercontent.com/22095643/45999460-06716900-c0a6-11e8-8578-a8bba35edf78.png)

brunolajoie:
probably. But we'll need to be carefull is they really stop coal for any reason in the future
![image](https://user-images.githubusercontent.com/22095643/45999542-40426f80-c0a6-11e8-9721-1249643f9d08.png)

corradio:
how do you identify cases where it's truly 0?
We could try to double check that by finding cases where coal is close to 0 and total load is continuous (as opposed to the screenshot I sent where there's a sudden drop in load/production)

brunolajoie:
(@oli comment updated above)

jarek:
Hey guys, sorry to harp on about this. But this kind of problem comes up repeatedly for different parsers and it would really be easiest to solve on the backend by rejecting points with too big of a difference in local consumption from previous datapoint.

Here we could add a filter saying that more than 20 MW of coal is required. Then coal drops to 25 MW and we have a similar hole in the chart. Or we add a check in the parser for the time series returned not being too jumpy. Then we get a similar problem in Portugal. It's a bit cat and mouse...

corradio:
What bout coding à simple derivative filter in the data validation method
that we can apply to all inputs?

alixunderplatz:
The coal power plant on Mallorca has 2x125 MW and 2x130 MW hard coal units.
To have at least one of the blocks running seems like a basic requirement for save grid operation on the Balearic islands.

From brunos graph you see that coal usually is above 100 MW. 
Lowest I could find this year was [74 MW on 16th April](https://demanda.ree.es/visiona/baleares/mallorca/acumulada/2018-04-16) during a possible outage of one 120 MW block, while at least another unit is still running.

I guess coal > 50 MW seems like a good boundary.

The link with mainland has a capacity of 400 MW and is not yet able to power the islands on its own.
There could be a [second stage with 1000 MW](https://www.4coffshore.com/windfarms/interconnector-balearic-islands--spain-mallorca-second-link-icid125.html) in the future.

[...]

PR #1657, opened 2018-10-27
Set minimum requirement and validate diffs for coal in ES-IB

systemcatch:
Requires coal to be between 50-600MW and only allows a change of 100MW coal generation between datapoints (10mins).

ref #1613


systemcatch:
@corradio happy with my fix?

alixunderplatz:
@systemcatch the 100 MW in 10min condition might be a bit too strict. Would we skip outages of the coal blocks (>=125 MW) when they run at full power and fail spontaneously?

systemcatch:
@alixunderplatz I'll make it 150MW so we deal with that situation correctly.

PR #1659, opened 2018-10-30
Production Consumption Toggle / mobile modal fix

ovbm:
the toggle on mobile:

notice that I removed the top electricitymap logo to gain some vertical space and instead placed the logo in the tab of the app. I think this solution makes a lot of sense for mobile:
emapanim

What's missing is a modal or legend that explains what's happening, maybe with a link to a blogpost for further info. Can someone make a suggestion for this text?

todo:
- [x] explanation moda/legend


corradio:
Smart move. Small detail: I see an vertical alignment issue with the caption of the button compared to the two other buttons.
Do you think we should rename "Areas"?
For the modal, we don't have any blogpost, but for consumption we could say "takes into account imports and exports" and for production "ignores imports and exports"

web/public/images/electricitymap-icon.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
corradio: didn't we already have an icon?

web/src/main.js
```diff
@@ -794,6 +794,15 @@ if (!getState().application.isMobile) {
   });
 }
 
+// Production/Consumption
```
corradio: can you also remove the toggle that I coded in the bottom of main.js (that binds to `window`?)

web/src/main.js
```diff
@@ -794,6 +794,15 @@ if (!getState().application.isMobile) {
...
+function toggleElectricityMixMode() {
+  dispatchApplication('electricityMixMode', getState().application.electricityMixMode === 'consumption'
+    ? 'production'
+    : 'consumption');
+  const activeToggle = d3.select('.production-toggle-active-overlay');
```
corradio: This should be in the observer instead of here.

web/public/images/electricitymap-icon.svg
ovbm: no, this is just the emap symbol as svg, we only had emap smybol + text as svg

ovbm:
The vertical alignment was always there, it moves the "active" legend item a couple of pixels up compared to the inactive ones. not a fan but I kept it because it was already there.

Don't know about Areas, do you have an Idea?

Ok, so it's going to be a hover label with those two infos. I wonder if the toggle is too prominent and if it should be collapsed.

web/src/main.js
ovbm: do you mean the activeToggle part?

ovbm:
this is a solution for the mobile itunes thing:
screenshot 2018-10-30 at 10 52 49
screenshot 2018-10-30 at 10 52 27

web/public/images/electricitymap-icon.svg
corradio: ok!

web/src/main.js
corradio: yes

corradio:
yes indeed default should be consumption. OK for the other changes.

corradio:
you should take a look at the reducer where default values are set

ovbm:
I think my work is done for this now

web/src/main.js
ovbm: done

web/src/main.js
ovbm: done

corradio:
Merging now, we can always improve later.

PR #1669, opened 2018-11-08
Reverse Northern Ireland exchange directions

### LINKED ISSUE #1668: Northern Ireland data quality issues
Right now we show (timezone of screenshots UTC-5):

image

Link from GB (Scotland) is the "Moyle" link - SONI show it a large part of the _supply_:

image

Our data shows it as export, and at over 100% of the available supply at that:

image

Something is wrong. Is it just Moyle being flipped, or is there more problems? If someone has some time to investigate, it would be super.

brunolajoie:
This should fix #1668 (at least, the part about exchanges direction).


corradio:
How can we be sure about the exchange direction?

brunolajoie:
By comparing emap with the pie chart from http://www.soni.ltd.uk/ at times of imports on the soni website (right now, NIR is exporting on their website so hard to tell) but according to #1668 its really reversed and I checked the order of magnitude in MW were correct

corradio:
wow! If you can double check by running the parser manually and compare the exact numbers from their website and from the parser we'd have final closure on that matter.
Thanks for fixing. Already approved though.

systemcatch:
We need to check the numbers to be sure, this could effect other exchanges though as I'm not sure why just the moyle one would be different.

systemcatch:
At the same time the GB-NIR->IE exchange reports -177MW meaning Ireland is exporting to Northern Ireland. However that does not show up in the fuel mix pie with the Moyle connector. Makes me think that one is wrong as well.

systemcatch:
Given that Moyle is show in the fuel mix pie chart I think you're right to reverse these Bruno and keep a close eye for the next few days.

brunolajoie:
I'm deploying - on my check this afternoon, both echanges to Ireland and to Great Britain seemed reverse. Let's watch carefully!

PR #1673, opened 2018-11-14
Update Finland's capacities to those currently on ENTSO-E

jvarho:
The capacity numbers for Finland are both old and wrong. Seems #1378 / #393 moved peat from biomass to coal in the parser, but did not adjust Finland's capacities correspondingly.

These are the current 2018 numbers from ENTSO-E. I dropped the IRENA numbers, because they are almost the same, but would change the total unless you fudge "unknown" to compensate.


corradio:
cc @alixunderplatz in case you have some insights

Issue #1347, opened 2018-04-23
Split Balearic islands ES-IB generation


alixunderplatz:
**It is possible to split the Balearic islands** (just like the Canary Islands in #626)
This would solve the recent issue #1346, too.

Here is the equivalent website for the Balearic system:

http://www.ree.es/en/activities/balearic-islands-electricity-system/balearic-electricity-demand-in-real-time

![image](https://user-images.githubusercontent.com/25743609/39114040-df8fd26e-46dd-11e8-896f-f1e79e22c226.png)

It is using the same request structure like @systemcatch described in #626 , just a few changes in the URL brought me to this:

**Mallorca:**
https://demanda.ree.es/WSvisionaMovilesBalearesRest/resources/demandaGeneracionBaleares?callback=angular.callbacks._2&curva=MALLORCA&fecha=2018-04-23

**Menorca:**
https://demanda.ree.es/WSvisionaMovilesBalearesRest/resources/demandaGeneracionBaleares?callback=angular.callbacks._2&curva=MENORCA&fecha=2018-04-23

**Ibiza:**
https://demanda.ree.es/WSvisionaMovilesBalearesRest/resources/demandaGeneracionBaleares?callback=angular.callbacks._2&curva=IBIZA&fecha=2018-04-23

**Formentera:**
https://demanda.ree.es/WSvisionaMovilesBalearesRest/resources/demandaGeneracionBaleares?callback=angular.callbacks._2&curva=FORMENTERA&fecha=2018-04-23

Interconnectors are labeled "emm" for "Mallorca<->Menorca" (inoperative!), "eif" for "Ibiza<->Formentera" and "eim" for "Ibiza<->Mallorca".  "cb" is Mallorca<->Spanish peninsula.

Formentera is recently missing on the map due to its smaller size and should be added in the process.

alixunderplatz:
@blackleg is this possibly covered by your reescaper already?

alixunderplatz:
@blackleg that's lovely, thanks! :)

I guess these are the next steps:

* seperate the islands' geometry and create new zones:
  * ES-IB-MA for Mallorca
  * ES-IB-ME for Menorca
  * ES-IB-IZ for Ibiza
  * ES-IB-FO for Formentera
    * Formentera is not on the map and must be added to the geometries.
     @systemcatch Same approach as in #1022 required?

[...]

PR #1677, opened 2018-11-15
Split Balearic Islands

systemcatch:
Finally managed to get this done in qgis! Formentera is grey since it has no generation of it's own but really it should be the same colour as Ibiza.

Let me know what you think.

/cc @blackleg 
ref #1347 

split_balearic


corradio:
cool! Will take a look asap

alixunderplatz:
@corradio looks good to me, too! :)

systemcatch:
@corradio  :grin: lets merge it! Do you think Formentera will be coloured correctly on the main map?

corradio:
If it has no generation I need to add it to our specific exclusion list, else it will appear as grey. Do you confirm that's the thing we should do?

systemcatch:
Actually it has some solar generation running at the moment.

PR #1679, opened 2018-11-15
Add solar generation to Russia

alixunderplatz:
Adds the new renewable category "P_REN" as "solar".

Some of the sample output:
datetime.datetime(2018, 11, 15, 9, 0, tzinfo=tzfile('W-SU'))}, {'zoneKey': 'RU-1', 'production': {'nuclear': 26374.69140625, 'hydro': 7859.890625, 'unknown': 76245.1748046875, 'solar': 148.809997558594, 'biomass': None, 'geothermal': None}, 'storage': {}, 'source': 'so-ups.ru'

### DIFF
--- a/parsers/RU.py
+++ b/parsers/RU.py
@@ -18,14 +18,16 @@ MAP_GENERATION_1 = {
     'P_GES': 'hydro',
     'P_GRES': 'unknown',
     'P_TES': 'unknown',
-    'P_BS': 'unknown'
+    'P_BS': 'unknown',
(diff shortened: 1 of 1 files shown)


brunolajoie:
Dyou think we should update default CO2 intensity of unknown? Seems minuscule enough, but if you could just check thanks!

alixunderplatz:
@Bruno I think we can leave the intensity for unknown as is. Not much will change at the moment. By the beginning of next year we can see if we find some fresh publications on this

PR #1684, opened 2018-11-21
Added some US states and corresponding bounding box

brunolajoie:


brunolajoie:
All news zones are US states, already existing as geometries. I don't think we need to execute the geometry generation script! yay!

brunolajoie:
@corradio ready for review

PR #1685, opened 2018-11-23
rename ERCOT to TX

brunolajoie:
Just renamed ERCOT to TX, as we did for California (not named CAISO but CA)


parsers/EIA.py
```diff
@@ -22,7 +22,7 @@
 EXCHANGES = {
     'MX-BC->US-CA': 'EBA.CISO-CFE.ID.H',
     'US-BPA->US-IPC': 'EBA.BPAT-IPCO.ID.H',
-    'US-ERCOT->US-SPP': 'EBA.ERCO-SWPP.ID.H',
+    'US-SPP-->US-TX': 'EBA.SWPP-ERCO.ID.H',
```
corradio: I think the right hand side should be
```suggestion
    'US-SPP-->US-TX': 'SWPP.ID.H-EBA.ERCO',
```

systemcatch:
Would be nice to have ERCOT/TX on the map.

PR #1688, opened 2018-11-24
Added missing German strings

jld3103:


jarek:
Thanks @jld3103 and @alixunderplatz!

PR #1689, opened 2018-11-24
ignore CA-NS SSL problems

jarek:
In our Kibana and on my local Python I get:

`ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:847)`

In Kibana this started on Friday. I noticed that NSPower.ca now redirects web traffic to HTTPS. In a browser (Firefox 63.0.3 on Linux) https://www.nspower.ca/system_report/today/currentmix.json loads just fine, but Python struggles with it. I think it's something to do with certificate stores? We could probably upgrade Python to the latest, but having SSL verification on this domain isn't important, so just ignore the SSL errors for now.

### DIFF
--- a/parsers/CA_NS.py
+++ b/parsers/CA_NS.py
@@ -35,11 +35,14 @@ def _get_ns_info(requests_obj, logger):
         'wind': 700
     }
 
-    mix_url = 'http://www.nspower.ca/system_report/today/currentmix.json'
-    mix_data = requests_obj.get(mix_url).json()
+    # As of November 2018, NSPower redirects their web traffic to HTTPS. However we've been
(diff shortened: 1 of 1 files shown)


corradio:
Let me try to update python/ requests first

jarek:
On my local (admittedly Python 3.5.2) I still get the error even with `certifi==2018.10.15, requests==2.20.1, urllib3==1.24.1`

corradio:
We're running it from the Docker image `python:3.6`. Do you think that could be the issue?
Let's merge anyway. I've updated requests on the repo though.

PR #1691, opened 2018-11-27
Add co2eq json

### LINKED ISSUE #1576: Import intensity could fallback on yearly averages when missing/unknown
When a country, or area, is importing electricity from another country and the exporting country's production sources are unknown, it seems as if the intensity of the imported electricity is set to be equal to the intensity of the importing country. But this is hardly meaningful. Would it be possible to set the unknown intensity of imported electricity to an average or mean value from a historical period? E.g. the last month or the same month last year. Or to the last available dataset (depending on how old that is).

I can see that it happens quite often for Norway, that "Data [is] temporarily unavailable". The intensity of the electricity exported to Sweden is low, while it is medium high when exported to West Denmark.

### LINKED ISSUE #1690: Use yearly average carbon intensity for hydro discharge emission factor
Potentially, using a round-trip efficiency loss factor.

Fallback on "world-average carbon intensity" factor if country-specific yearly country average is unknown.

Step 1:  Input annual average emission factor for each eMap zones when possible, in `defaultExportCo2eqFootprint` (using eMap historical database)
Step 2:  code the link between hydro_discharge CO2 intensity factor and `defaultExportCo2eqFoorprint` values
Step 3: add a world average default fallback

corradio:
Fixes #1690 
Fixes #1576 

Default exporting CO2 intensities are added:

image

Hydro discharge factor uses annual averages:

image


corradio:
@brunolajoie ready for review

config/co2eq_parameters.json
```diff
@@ -0,0 +1,850 @@
+{
+  "emissionFactors": {
+    "defaults": {
+      "battery discharge": {
+        "source": "TODO",
+        "value": 0
```
brunolajoie: We should treat battery discharge like hydro discharge!

config/co2eq_parameters.json
```diff
@@ -0,0 +1,850 @@
...
+        "source": "IPCC 2014",
+        "value": 24
+      },
+      "hydro charge": {
+        "_comment": "Emissions are counted at discharge",
+        "source": "Tomorrow",
+        "value": 0
+      },
+      "hydro discharge": {
+        "_comment": "Here we use the world average",
```
brunolajoie:
```suggestion
        "_comment": "As a default fallback, we use here the annual mean carbon intensity of all electricityMap zones",
```

PR #1705, opened 2018-12-11
Create exchange between SIC and SING in Chile

systemcatch:
For the database only as it lags around 4 days behind live. Hopefully this might change soon with the revamp of the coordinador website.

ref #655

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -633,6 +633,16 @@
     },
     "rotation": 180
   },
+  "CL-SIC-CL_SING": {
+    "lonlat": [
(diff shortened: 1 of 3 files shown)


brunolajoie:
You rock!! Thanks a lot!!

brunolajoie:
Is the interconnection new? Like, is there historical data to refetch, or is that only starting from now on?

systemcatch:
There is lots of historical data to fetch (likely several years). It has to be run with a target_datetime argument though else it won't work.

PR #1718, opened 2018-12-26
Various improvements to AR.py

### LINKED ISSUE #1288: Revise and improve AR.py plant mapping
Data for categories ["thermal"](http://portalweb.cammesa.com/MEMNet1/Pages/Informes%20por%20Categor%C3%ADa/Operativos/DespachoGenTermica_Minimal.aspx) and ["hydro"](http://portalweb.cammesa.com/memnet1/Pages/Informes%20por%20Categor%C3%ADa/Operativos/DespachoGenHidra_Minimal.aspx) generation from Cammesa is used for the map.
Some recently comissioned power plants were causing a warning every once in a while, because they were not yet included in the type mapping. This could be fixed by improving the parser a bit. 

I just copy and edit the rest of my comment from #1284:
_ _ _ _

1. There is a clear pattern behind the power plant IDs: In general, everything that ends with "TG" before the numeric part is a gas turbine and should be running on natural gas as main fuel (e.g. ABCD**TG**01)
Actually, the plant mapping in the AR.py could be entirely dismissed, when only the the fourth and fifth letter of the plant IDs will be considered.

2. The mapping could be refined and updated by using the first column on
http://portalweb.cammesa.com/memnet1/revistas/estacional/base_gen.html for all of the existing power plants. 

image

Could that page also be used to map the reporting plants by their IDs automatically? (see mapping below)
- Here is the mapping of that first column to the fuel types (from what I could relate to the power plant IDs):
* TG is gas (turbina de gas)
* CC is gas (combined cycle)
* AG is gas (no idea what the A means, but the web leads me to CCGT plants)
* TV is mostly gas in the recent mapping - maybe this category needs a seperate review:
  * In the recent mapping, 5 of the steam turbines TV (turbina de vapor) are running on coal (~170 MW), which seems easy enough to consider, the rest of TV should be using gas, too. Maybe some biomass is included here as well.
* DI is oil (diesel)
* EO is wind (eolico)
* HI is hydro (hidro)
* FV is solar (fotovoltaica)
* BG is biomass (biogeneradora)
* BM is biomass (biomassa)
* NU is nuclear
* BRS and MH have a very tiny amount of MWs, I suggest mapping them to "unknown" together with any other abbreviation they might come up with in the future.

This would eliminate/limit this issue that pops up every once in a while when a new unit (or previously unused unit) suddenly enters operation.

I went through the mapping in the parser and found some errors as well, like "EO" wind power plants and "FV" solar power plants recently being mapped to "hydro". Probably because they are included in the page with hydro generation data. There are only 5 of these, that need to be changed, so that has no huge impact.

systemcatch:
resolves #1288 

@alixunderplatz can you check I've got the plant mapping right?

- Add new plants to mapping.
- Correctly map wind and solar plants to type.
- Improve handling of unknown plants.
- Small refactoring for readability.

### DIFF
--- a/parsers/AR.py
+++ b/parsers/AR.py
@@ -3,7 +3,7 @@
 import itertools
 import re
 import string
-
+import logging
 import arrow
 import requests
 from bs4 import BeautifulSoup
@@ -36,6 +36,7 @@ power_plant_type = {
     'ACAJTG05': 'gas',
     'ACAJTG06': 'gas',
     'ACAJTV07': 'gas',
+    'ACHIEO': 'wind',
     'ADTOHI': 'hydro',
     'AESPTG01': 'gas',
     'AESPTG02': 'gas',
@@ -56,8 +57,10 @@ power_plant_type = {
     'ANCHDI03': 'oil',
     'ANCHDI04': 'oil',
     'APARTV01': 'gas',
-    'ARA2EO': 'hydro',
-    'ARAUEO': 'hydro',
+    'ARA2EO': 'wind',
+    'ARA3EO': 'wind',
+    'ARAUEO': 'wind',
+    'ARC2EO': 'wind',
     'ARGETG01': 'gas',
     'ARISDI01': 'oil',
     'ARMATG01': 'gas',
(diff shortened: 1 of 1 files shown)


systemcatch:
Hmmm we did talk about adding all plants from http://portalweb.cammesa.com/memnet1/revistas/estacional/base_gen.html, let me look at that as well.

systemcatch:
I'll create an exclusion list for those 2 plants as they look like they're based in Uruguay. The other two I'll keep looking for any information, at the moment they aren't close to the exchange data we get  for AR->UY so it doesn't look like imports.

That link to renewable generation is a goldmine, we can simplify the parser significantly by incorporating it in the future.

PR #1721, opened 2018-12-30
El Salvador update

alixunderplatz:
- Updated & added renewable capacity
- Updated capacity sources in README.md
- Changed parser to fetch "termico" as generation from oil
  - sample output: 
`{'zoneKey': 'SV', 'datetime': datetime.datetime(2018, 12, 30, 5, 0, tzinfo=tzstr('UTC-6')), 'production': {'biomass': 149.89, 'coal': 0.0, 'gas': 0.0, 'hydro': 23.04, 'nuclear': 0.0, 'oil': 50.72, 'solar': 0.0, 'wind': None, 'geothermal': 168.02, 'unknown': 0.0}, 'storage': {'hydro': None}, 'source': 'ut.com.sv'}]`
  - based on generation figures [of IEA from the past](https://www.iea.org/statistics/?country=ELSALVADOR&year=2016&category=Key%20indicators&indicator=ElecGenByFuel&mode=chart&categoryBrowse=false&dataTable=ELECTRICITYANDHEAT&showDataTable=true), there is no other fossil source but oil/diesel
  - a new liquid gas power plant will start operating in 2020/2021 in El Salvador, then we'd have to look out for changes
  - see https://gastechinsights.com/article/what-energa-del-pacficos-lng-to-power-project-means-for-el-salvador

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,9 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
 - Denmark (Bornholm)
   - Wind: [stateofgreen.com](https://stateofgreen.com/en/profiles/regional-municipality-of-bornholm/solutions/kalby-wind-turbines)
 - Dominican Republic: [Climatescope](http://global-climatescope.org/en/country/dominican-republic/#/details)
+- El Salvador:
+  - Thermal: [CNE](http://estadisticas.cne.gob.sv/wp-content/uploads/2016/09/Plan_indicativo_2016_2026-1.pdf) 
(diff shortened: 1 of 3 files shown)


corradio:
@alixunderplatz what about adding the IEA link justifying that thermal = oil in the code itself? That way a future maintainer might already have an answer to his question :)

alixunderplatz:
@corradio will do so!
By the way: if desired, here is super-high-frequency, ultra-real-time data for El Salvador:
https://www.graficostr.ut.com.sv/

image

corradio:
wow! We should probably fetch data from there?

PR #1726, opened 2019-01-03
Slovak translation

PatriciaGolianova:
Labels, FAQ, areas


systemcatch:
@PatriciaGolianova I was able to build your branch, things look ok from a quick pass over.
sv_translation

corradio:
Thanks @PatriciaGolianova !
Were you able to run the system locally?
In order to make the bindings for a new language, you have to add it to the relevant parts of the following section:
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/server.js#L33-L91

systemcatch:
@alixunderplatz ahaha oh wow I should've seen that!

PatriciaGolianova:
Hello guys, 

I have added the required information about new translation to server.js and translation.js. 
But the things I have noticed: 

1. Consumption/Production is not translated - also in Swedish screenshot that @systemcatch madeas well.
snimka obrazovky 2019-01-07 o 11 42 34

2. They are lines that were probably outdated I am not sure about - server.js
snimka obrazovky 2019-01-07 o 11 38 14

Thank you for your patience :)

web/server.js
```diff
@@ -220,6 +222,6 @@ app.all('/dist/*.map', (req, res, next) => {
 app.use(express.static(STATIC_PATH, { etag: true, maxAge: isProduction ? '24h' : '0' }));
 
 // Start the application
-server.listen(process.env['PORT'], () => {
-  console.log(`Listening on *:${process.env['PORT']}`);
+server.listen(8000, () => {
```
corradio: Can you revert to the old code?

PatriciaGolianova: Fixed in the following commit.

corradio:
thanks!

Issue #1729, opened 2019-01-07
NO capacity outdated


Gazer75:
No idea where the data comes from, but ENSTOE is not to be trusted on this.

At any point you can get a csv file from here for hydro:
https://www.nve.no/energiforsyning-og-konsesjon/vannkraft/vannkraftdatabase
This can then be filtered to get the pure pumps out (Pumpe).
Can also filter for each of the five areas (Elspotområde).

File for wind can be found at:
https://www.nve.no/energiforsyning-og-konsesjon/vindkraft/utbygde-vindkraftverk/

As of today I summed ~32440MW total hydro power capacity in Norway.
NO1 = ~3592MW
NO2 = ~11317MW
NO3 = ~4571MW
NO4 = ~5088MW
NO5 = ~7873MW

Total wind capacity is ~1664MW, but the table do not show what area each windfarm is in.

jarek:
Note that the website currently does not display capacities in Norway because regional capacities were not added so far after the split into the five regions.

Gazer75:
Those regions have existed for at least 16 years. I found the history of the reservoir status back to 2002 divided by these areas.

Some strange values can be found at ENTSOE site here:
https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show

Only explanation is grid limitation or that a large portion of the generators are offline for maintenance. 

Either way the values are there.

Gazer75:
Closing this issue for now then. I'll try to update capacities later when more of the wind farms under construction start producing later this year. NVE expects almost 2000MW worth of wind to be finished in 2019 according to the 2018 3rd quarter report.

PR #1731, opened 2019-01-07
Production capacities for Norway

Gazer75:
Update production capacities for Norway #1729 
Values for wind and hydro calculated from NVE database. ENTSOE values were very low.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,8 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
 - Netherlands: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
 - Nicaragua: [Climatescope](http://global-climatescope.org/en/country/nicaragua/)
 - Norway
-  - Wind: [ieawind.org](http://www.ieawind.org/countries/norway.html)
+  - Hydro: [NVE](https://www.nve.no/energiforsyning-og-konsesjon/vannkraft/vannkraftdatabase/)
+  - Wind: [NVE](https://www.nve.no/energiforsyning-og-konsesjon/vindkraft/utbygde-vindkraftverk/)
   - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
 - Northern Ireland: [EIR Grid](http://www.eirgridgroup.com/site-files/library/EirGrid/Generation_Capacity_Statement_20162025_FINAL.pdf)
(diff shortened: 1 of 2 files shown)


systemcatch:
Hey @Gazer75, thanks for the contribution. Couple of things need to be tweaked before we can merge this. Can you update the readme source list with the new Norway regions. Also we tend to only set capacities to zero when we are absolutely sure there are no generators of that type.

Gazer75:
So new at all this. Is there some way I can edit that file and add it here?
As I edited about the wind and hydro is basically from the databases I listed in #1729

Gazer75:
Unless you consider district heating in the calculations here, then I can say with certainty that we have only hydro and wind with gas as backup.
Where the 5MW of gas listed in NO2 at ENTSOE is coming for I have no clue. 
Gas value for NO5 should probably be set to 0 unless they postponed the planned shutdown of Mongstad at end of 2018. 
In NO4 there is a mobile backup gas plant at Tjeldbergodden which is why there is 270MW listed.

brunolajoie:
Please find below what our database has registered as max instantaneous output in 2018 for each NO zone. 

We'll have conflicts for each zones, where output will be higher than reported capacity (especially for gas). It will report more than 100% capacity factor. Not such a big deal, but that should deserve a bit of investigation i'd say!

NO1
biomass          NaN
coal             NaN
gas             29.0
hydro         2959.0
nuclear          NaN
oil              NaN
solar            NaN
wind             0.0
geothermal       NaN
unknown         91.0

NO2
biomass          NaN
coal             NaN
gas              7.0
hydro         9440.0
nuclear          NaN
oil              NaN
solar            NaN
wind           566.0
geothermal       NaN
unknown        404.0

NO3
biomass          NaN
coal             NaN
gas            110.0
hydro         3765.0
nuclear          NaN
oil              NaN
solar            NaN
wind           350.0
geothermal       NaN
unknown        263.0

NO4
biomass          NaN
coal             NaN
gas            229.0
hydro         4295.0
nuclear          NaN
oil              NaN
solar            NaN
wind           250.0
geothermal       NaN
unknown         33.0

NO5
biomass          NaN
coal             NaN
gas            143.0
hydro         6602.0
nuclear          NaN
oil              NaN
solar            NaN
wind             0.0
geothermal       NaN
unknown         29.0

Gazer75:
I assume these production numbers are also coming from ENTSOE?
I could contact Statnett or something and see if someone is willing to explain the numbers reported

brunolajoie:
ENTSOE yes!

brunolajoie:
I think we can turst your capacity number more than our ENTSOE actual generation data.
I'm approving the PR as it is: what will have to be checked/changed in more our ENTSOE parser than your fixed capacity data!

PR #1734, opened 2019-01-08
update zone.json "de"

Bohne13:
updated netto installed capacity from Frauenhofer ISE (data from 27. DEC 2018)


jarek:
Just to double-check... where do figures for geothermal and hydro storage actually come from? https://www.energy-charts.de/power_inst.htm?year=2018&period=annual&type=power_inst is given as source in readme and has the values as in the PR, but I don't see geothermal and hydro there.

Bohne13:
@jarek Form geothermal data, I found a study or a website with data from 2018. I try to find it again!
For hydro storage, truly I haven't found current data from 2018.
With more time for searching the web, probably I could find some data.

corradio:
@brunolajoie I think you can merge

alixunderplatz:
@jarek I think we took hydro storage from the BNetzA Kraftwerksliste back then

PR #1738, opened 2019-01-11
switch oil-shale category from 'oil' to 'coal'

### LINKED ISSUE #1737: Emissions from oil shale
Estonia produces electricity from oil shale.

This is a solid fossil fuel that is in the same statistical group with coal.

It would be correct to dislpay the emissions from oil shale under "coal" (solid) not "oil" (liquid).

Source: https://ec.europa.eu/eurostat/statistics-explained/index.php?title=Glossary:Fossil_fuel&oldid=120658

brunolajoie:
fix #1737
We'll need to re-scrap historical Estonia dataset to maintain consistency

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -51,13 +51,13 @@ ENTSOE_PARAMETER_DESC = {
 ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in ENTSOE_PARAMETER_DESC.items()}
 ENTSOE_PARAMETER_GROUPS = {
     'biomass': ['B01', 'B17'],
-    'coal': ['B02', 'B05', 'B08'],
+    'coal': ['B02', 'B05', 'B07', 'B08'],
     'gas': ['B03', 'B04'],
     'geothermal': ['B09'],
     'hydro': ['B11', 'B12'],
     'hydro storage': ['B10'],
     'nuclear': ['B14'],
-    'oil': ['B06', 'B07'],
+    'oil': ['B06'],
     'solar': ['B16'],
     'wind': ['B18', 'B19'],
     'other': ['B20', 'B13', 'B15']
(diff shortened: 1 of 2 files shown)


parsers/ENTSOE.py
```diff
@@ -51,13 +51,13 @@
 ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in ENTSOE_PARAMETER_DESC.items()}
 ENTSOE_PARAMETER_GROUPS = {
     'biomass': ['B01', 'B17'],
-    'coal': ['B02', 'B05', 'B08'],
+    'coal': ['B02', 'B05', 'B07', 'B08'],
```
corradio: what is ENTSOE B07? This mapping is for all of ENTSOE data - not just EE, so we would have to be careful here.

brunolajoie: it's "oil shale"

brunolajoie: Yes, i should check for which country is B07 used ! 👀

brunolajoie: @alixunderplatz on top of your head dyou remember which other ENTSOE region could use oil shale by any chance?

alixunderplatz: I'm quite sure no other country of entsoe uses oil shale, just Estonia.

brunolajoie: That will make things easier. Good

alixunderplatz:
@jarek good point! That coal capacity of 5MW was assumed to be for peat in Estonia, so the oil shale will be added on top now. The comment wasn't added to the code though.
https://github.com/tmrowco/electricitymap-contrib/issues/393#issuecomment-386320742

@brunolajoie is this acutally problematic regarding the emission factor? The new Estonian "coal" category will be a mix of much oil shale and a very small part of peat... not sure how "exact" you want it to be or whether we need to calculate a new carbon intensity here

brunolajoie:
Peat goes into "coal", as well as shale oil, so the only thing to change if you wish is that of "coal carbon intensity estonia". It will change a tiny tiny bit, so no big deal!

Issue #1213, opened 2018-03-14
Add Bangladesh (d-1 data)


alixunderplatz:
⚠️ **WARNING** ⚠️ Using Win10, my Antivirus says their site is infected and blocks it - so I won't post any direct links here to keep everyone safe ;) The cause of the warning  is "HTML:iframe-BWC[Trj]". ⚠️  

I accessed it through a safe virtual browser environment instead. 😷
_ _ _ _

There is data by "**Power Grid Company of Bangladesh**" (PGCB) on the hourly generation mix and exchanges with India. I searched their site for useful info and found their "operational daily reports" with a full breakdown by type and even by unit.

⚠️Here is a screenshot of the link to the daily reports ⚠️:
![image](https://user-images.githubusercontent.com/25743609/37426459-cf752450-27c6-11e8-97f3-d5ae6536c901.png)

This is the link-structure to the XLSM-files (date in the end "dd-mm-yy"). They changed format from XLS to XLSM from 15th Jan 2018.
![image](https://user-images.githubusercontent.com/25743609/37426536-0d30e40a-27c7-11e8-9e78-a3a6fea8ba8c.png)

You can go further back to 01st Jan 2014 until 14th Jan 2018 to download the XLS-files. The link-structure is inconsistent and sometimes there is a space or "0" missing.

![image](https://user-images.githubusercontent.com/25743609/37418113-b8e80abe-27b1-11e8-80cd-ac1163b26c95.png)

The "daily report" download-link wasn't causing a virus warning in a usual browser and so didn't opening the file.
It is provided as XLS/XLSM file with several sheets. There is **half-hourly generation data** for yesterday and exchanges. The forecast is only for peak hours.

![image](https://user-images.githubusercontent.com/25743609/37413195-c5a15db0-27a6-11e8-8f04-3e52d72023ad.png)

Here is a copy of the data for 13th March. Mostly GAS and OIL. 
HVDC and TRIPURA are imports from India.

TIME | GAS | HVDC | P.GEN(GAS). | P.GEN(OIL). | COAL | HYDRO | OIL | SHORTAGE | TRIPURA
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
00:00 | 3027 | 467 | 1440 | 1343 | 364 | 69 | 652 | 0 | 257
00:30 | 2916 | 473 | 1450 | 1320 | 362 | 69 | 698 | 0 | 285
[...]

PR #1746, opened 2019-01-14
Add Bangladesh (historical)

systemcatch:
ref #1213 

This parser was annoying to write given all the url and formatting inconsistencies (see code comments), but in spite of that I got it to work. 

Couple of important points.
- Some days are simply missing an excel file so expect the odd 404 when scraping.
- While I've done my best to cover all the format options there's over 1500 excel files here, there may be some outliers.


jarek:
The CircleCI failure (at the bottom of a very long log) is:

```
./parsers/BD.py:74:9: F821 undefined name 'logger'
        logger.warning('New data columns may be present xlsm file.', extra={'key': 'BD'})
        ^
1     F821 undefined name 'logger'
1
Exited with code 1
```

systemcatch:
Thanks @jarek, I was pretty tired last night when finishing this off.

systemcatch:
@brunolajoie can you take a look over this?

PR #1747, opened 2019-01-19
DK real-time data

tmslaine:
Changes Danish production and exchange data sources to https://www.energidataservice.dk/en/group/production-and-consumption

Estimates generation by fuel for the time frame from -2hrs to -5minutes
by linear regression of 5-minute frequency small-scale and large-scale production against
recent reported hourly fuel data. When fuel data becomes available, corrects this estimate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Real-time electricity data is obtained using [parsers](https://github.com/tmrowc
 - Costa Rica: [ICE](https://appcenter.grupoice.com/CenceWeb/CencePosdespachoNacional.jsf)
 - Croatia (Exchanges): [HOPS](https://www.hops.hr/wps/portal/hr/web)
 - Cyprus : [TSO](http://www.dsm.org.cy/en/daily-system-generation-on-the-transmission-system-mw)
-- Denmark: [ENTSOE](https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html)
+- Denmark: [TSO](https://www.energidataservice.dk/en/group/production-and-consumption)
 - Denmark (Bornholm): [PowerlabDK](http://bornholm.powerlab.dk/)
 - Dominican Republic: [OC](http://www.oc.org.do/Reportes/postdespacho.aspx)
 - El Salvador: [Unidad de Transacciones](http://estadistico.ut.com.sv/OperacionDiaria.aspx)
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -712,7 +712,7 @@
       54.9
     ],
     "parsers": {
-      "exchange": "statnett.fetch_exchange",
+      "exchange": "DK.fetch_exchange",
       "exchangeForecast": "ENTSOE.fetch_exchange_forecast"
     },
     "rotation": 0
@@ -727,7 +727,7 @@
       54.486043
     ],
     "parsers": {
-      "exchange": "statnett.fetch_exchange",
+      "exchange": "DK.fetch_exchange",
       "exchangeForecast": "ENTSOE.fetch_exchange_forecast"
     },
     "rotation": 0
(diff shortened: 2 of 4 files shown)


brunolajoie:
Thanks @tmslaine I'll test this locally and compare with existing data.
Can you explain me a bit more your last sentence about linear regression?

tmslaine:
Trying to get real-time production by type is a bit greedy as the 5-minute source does not report this. Instead, you have wind, solar, >100MW production, <100MW production.

For hourly data you have wind, solar, hydro, biomass, waste, oil, gas and coal.
The parser tries to estimate each of hydro, bio(=biomass+(50% of waste)), oil, gas, coal and unknown(=50% of waste) as a weighted sum of >100MW and <100MW production, respectively.

To do this it
1) fetches production data **df** from the 5-minute frequency data source for the past _N_ hours, _N_=8
2) takes hourly means **df_1h** of the 5-minute data **df**
3) fetches production by type for the past N hours
4) joins the production by type data with hourly means **df_1h** by timestamp
5) for each of the fuels and hydro in the resulting hourly frequency dataframe,
optimizes in the least squares sense the weights _a_ and _b_
in  _aA_ +_bB_ = _F_,
 where 
_A_:= ">100MW prod.",
_B_:="<100MW prod.",
_F_:="Production from the fuel in question (or hydro)"
and applies weights _a_ and _b_ back to the ">100MW prod.",  "<100MW prod." of the original 5-minute data **df** of step 1. to estimate the real-time/5-minute frequency production from the fuel.


In addition, for each of the hours that have fuel data,
the parser scales the sum of ">100MW prod." and "<100MW prod."
in 5-minute data **df** to sum up to the realized production for each of the fuels and hydro.

If this seems too complicated, even implementing only the exchange part is probably an improvement.

jarek:
IMHO I would lean towards reporting only what we know for sure rather than trying to estimate precision. The estimate would be more useful in analysis code than in data collection code.

To give a dumb example: if we have a region that reports its usage in hourly increments, having a parser estimate wind production for 10:30 by calculating mean of production reported at 10:00 and at 11:00 is not useful IMO

corradio:
Hi @tmslaine,

That you for the very extensive work. Our philosophy at electricityMap is indeed to stay as close to the data by limiting the amount of interpolation/estimation.
However, the updated data source for the exchange and production data is certainly welcomed.
What I recommend:
- using the newest exchange data
- using the hourly mix data
- comment out / disable the 5min data, but maybe keep it coded in case we want to switch to it later on

How does that sound?
Furthermore, I don't understand the 50% waste assumption. Can you maybe explain that a bit more, or point to a document that does?

Thanks!

Olivier

tmslaine:
@corradio 
That sounds good, I will make the changes and update the PR.

Dividing waste into 50% renewable, 50% fossil is just a rule of thumb.
For example https://ens.dk/sites/ens.dk/files/Statistik/int.reporting_2016.xls
shows closer to 55%, 45% between MunWaste Renewable and MunWaste Non-renewable.
The historical division of waste into fossil and renewable is available for all EU countries in Eurostat table nrg_105a. If I recall correctly, eMap ENTSO-E parser maps all waste into biomass.

There are still at least two open questions before merging:
1) Should the production be changed from ENTSO-E to the new source? On a quick look
the local source has a decimal more but otherwise the values are the same as those reported
by ENTSO-E Transparency Platform.
2) Should the mapping of Waste be changed to 100% biomass?

corradio:
I'd say we follow ENS and add a clear comment/reference in the code.
I'd use your parser for the production data as it is closer to the data source.
What do you think @brunolajoie ?

brunolajoie:
- I'd say we follow ENS and add a clear comment/reference in the code.
Let's do this! We can assume safely that the ratio between REN-waste and non-REN waste and should remain approximately constant over within a year. Let's write it down clearly in the code so we we'll be able to update this ratio every year or so

- I'd use your parser for the production data as it is closer to the data source.
Agree

brunolajoie:
Tested the branch locally. 

The production parser
- goes back to 2017-01
- our historical dataset currently goes back 2015-01

The exchange parser
- goes back to 2015-01-01
- our historical dataset currently goes back 2015-01
- differs slightly that of stattnet in numbers. Tested now (16 vs. 8 MW) and tested in 2015-05-05 (316 vs 183MW). 

Can you guys have another independent check on data consistency before we merge that?

PR #1749, opened 2019-01-21
[Fix] Slider position should match 'stateDatetime'. Fix #1545

### LINKED ISSUE #1545: "Now" position of slider should match `stateDatetime`
Raised by @jarek :

Right now is 8:09 PM, when I load a page I get timestamp 7:57 (that's the stateDatetime): 

image

When I move the scrobbler a bit to left then back to as far right as it goes, I get shown timestamp 7:45. Because it moves in 15 minute intervals. Instead, it should show 7:57 at the "Now" position.

image

ewoken:


systemcatch:
:+1: lets test this locally (https://github.com/tmrowco/electricitymap-contrib#running-locally) to be sure it fixes the problem.

ewoken:
capture d ecran 2019-01-21 a 14 09 06capture d ecran 2019-01-21 a 14 08 38

First picture with the slider on the second to last value and the second with the slider on the last value.

web/src/main.js
```diff
@@ -1133,9 +1133,12 @@ function renderHistory(state) {
     .render();
 
 
-  zoneDetailsTimeSlider.onChange((selectedZoneTimeIndex) => {
-    if (getState().application.selectedZoneTimeIndex !== selectedZoneTimeIndex) {
-      dispatch({type: 'UPDATE_SLIDER_SELECTED_ZONE_TIME', payload: {selectedZoneTimeIndex: selectedZoneTimeIndex}})
+  zoneDetailsTimeSlider.onChange((selectedZoneTimeIndexInput) => {
+    if (getState().application.selectedZoneTimeIndex !== selectedZoneTimeIndexInput) {
+      const selectedZoneTimeIndex = selectedZoneTimeIndexInput === (history.length - 1)
```
corradio: can you add a small comment that states that if the slider is on the last value, then we set the time to null to effectively use the current state?

ewoken: Done !

PR #1750, opened 2019-01-21
[Fix] Arrow colors should change accordingly when colorblind mode is on. #410

ewoken:
Hi,

This is a first work to fix arrow colors when colorblind mode is on. I generate new .gif prefixed by 'colorblind-'.
I need some help or advice to connect exchange objects to colorblind value in state.

This is a preview:
screenshot_2019-01-21 live co2 emissions of electricity consumption

Thanks. :-)


corradio:
I'll help asap. Sorry for the delay

corradio:
Hi @ewoken ,

Thanks for the very helpful PR!
There's a few things to do indeed.
1. You need to react to changes in colorblindmode. Those changes happen here:
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/main.js#L1417-L1420
You probably want to do `ExchangeLayer.setColorblindmode(...)` there, to pass the update to the map arrow component, and subsequently call `.render()` on it to force it to change the images.

2. in exchange.js, you need to add a `setColorblindmode` (just look at how setData is coded) to make the component aware of the colorblindmode.

3. in render() of exchange.js, you can now select the image based on `this.colorblindmode`.

Ping me on slack if you have any questions.

web/src/main.js
```diff
@@ -337,6 +337,7 @@ try {
         .setData(getState().application.electricityMixMode === 'consumption'
           ? Object.values(getState().data.grid.exchanges)
           : [])
+        .setColorblindMode(getState().application.colorBlindModeEnabled)
         .render();
```
corradio: this will trigger a `render` twice. Either remove the render from here, or remove it from the `setColorblindMode` method.

corradio:
Hi @ewoken are you able to do the change? Else I'll do it.
Just ping me here (or on Slack) so we can move forward.

corradio:
Did the changes myself on master using this code. Thanks!

PR #1767, opened 2019-02-11
Add wind capacity for Australia

rbossler:
added capacity for Australia
source: https://en.wikipedia.org/wiki/Wind_power_in_Australia#Wind_power_by_state


brunolajoie:
Thanks. Can you add the source to the [readme](https://github.com/tmrowco/electricitymap-contrib#production-capacity-data-sources)?

PR #1776, opened 2019-02-16
Fix Italy South and Sicily

### LINKED ISSUE #1770: Italy South and Sicily need parser update
Production for Italian micro-bidding zones Priolo, Foggia and Brindisi isn't reported to ENTSO-E anymore since 1st January 2019. Probably it now is included in the respective surrounding zones? This is why we miss Italy South and Sicily on the map. Only micro-zone Rossano (included in our "Italy South") still has data which still must be added to ENTSO-E's "South".

alixunderplatz:
I hope this fixes #1770 
**Please test these changes locally, because I can't.**

Not sure what else needs to be deleted in case you want to re-grab data from the past for the database.


corradio:
@alixunderplatz have you tried to run the parser test command?
(https://github.com/tmrowco/electricitymap-contrib/blob/master/README.md#testing-parsers-locally)
Else @brunolajoie I'll let you test the parser and merge if it looks good

jarek:
I'm guessing this problem is because 'IT-SIC' is now a list of one item which somehow changes output of https://github.com/tmrowco/electricitymap-contrib/blob/204f4502c2ffbae79f24bc11a1319fddb5400450/parsers/ENTSOE.py#L930

parsers/ENTSOE.py
```diff
@@ -895,8 +895,8 @@ def fetch_production(zone_key, session=None, target_datetime=None,
 
 
 ZONE_KEY_AGGREGATES = {
-    'IT-SIC': ['IT-SIC', 'IT-PR'],
-    'IT-SO': ['IT-FO', 'IT-BR', 'IT-RO', 'IT-SO'],
+    'IT-SIC': ['IT-SIC'],
```
corradio: this should be removed, and in zones.json we should change IT-SIC parser from `ENTSOE.fetch_production_aggregate` to `ENTSOE.fetch_production`

corradio:
@alixunderplatz are you able to make the changes?

jarek:
I made the code change but am unable to test because I don't have a ENTSOE_TOKEN, can someone familiar with ENTSOE quickly check?

Issue #636, opened 2017-07-16
split Australia (Tasmania) to show King island off grid system


strassburger:
split Tasmania to show King island
Live data - King Island Advanced Hybrid Power Station

source
http://www.kingislandrenewableenergy.com.au/

systemcatch:
I'll take a crack at doing this one.

superlonglong:
the king island information page is notoriously unreliable.

hopefully their CSV is better, but dont assume that your code isnt working, it could just be garbage in, garbage out.

systemcatch:
@superlonglong I've had a look at this one and it seems difficult.  There are csv files but I don't think they've been updated for several years.  As regards getting getting the live data the dashboard uses a websocket to connect to ajenti who then send the data back to the dashboard.  It's probably possible to scrape this connection but it might take a while to figure out the best way.

If you want to have a go at this feel free as I'm working on the El Salvador parser right now.

superlonglong:
@systemcatch unfortunately im as useful as tits on a bull. i cant code :(

strassburger:
@superlonglong 
"the king island information page is notoriously unreliable"
no doubt, but right now 100%wind
PS:
i did copy the Page "http://www.kingislandrenewableenergy.com.au/" local and get same results.
in the folder there in the is a java - scripts which i  can not handle: "js_52fac29959611f022bfac1ab349c3dce.js"
And an html: "dashboard.html"
Both are to much for me

strassburger:
Have a little project now
[ki-2017-09-30.txt](https://github.com/tmrowco/electricitymap/files/1347552/ki-2017-09-30.txt)
 to read all data from King island:
May it help to put the data here too??
http://www.kingislandrenewableenergy.com.au/kiapp-dashboard/dashboard.html
Details:
[Readme.txt](https://github.com/tmrowco/electricitymap/files/1347548/Readme.txt)
[ki-script16.txt](https://github.com/tmrowco/electricitymap/files/1347549/ki-script16.txt)
[ki-2017-09-30.txt](https://github.com/tmrowco/electricitymap/files/1347555/ki-2017-09-30.txt)
Do not have the knowledge to integrate it in the co2 map.
Who can help?

systemcatch:
Hey @strassburger thanks for having a go at this one.  It's not a bad idea to scrape this using some browser automation tricks.  I can basically follow what your script is doing.  How about creating a repo so people can view it on github.

A few points;

[...]

PR #1780, opened 2019-02-18
Separate King Island from Tasmania

systemcatch:
ref #636 
/cc @strassburger 

king_island


alixunderplatz:
@systemcatch another step forward :) looks good!

strassburger:
Thanks,
sorry that i can not help really getting data

brandongalbraith:
Live data: http://www.kingislandrenewableenergy.com.au/

jarek:
@brandongalbraith please see #636 for discussion of that page and difficulties extracting data from it with a script. If you can help it would be great!

PR #1782, opened 2019-02-21
Updating translations

Raffox97:
**Front page**:
* Edited two buttons to make them translatable
* Edited a tooltip to make it translatable

**English**:
* Added the three new missing strings

**Italian**:
* Added King Island
* Added the three new missing strings


brunolajoie:
💯

PR #1786, opened 2019-02-25
Set UTC timezone for Danish parser

corradio:
Danish parser returned datetimes without timezone.
They are UTC at the source, so I simply added the UTC timezone to all dates.


corradio:
I'll let you merge and bump

PR #1787, opened 2019-02-26
ENTSOE refactor

### LINKED ISSUE #1742: ENTSOE parser cleanup
EE oil shale fix by @brunolajoie didn't work because the mapping
https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTSOE.py#L52-L64

is not used.

We should make sure those lines use the mapping directly, instead of using functions that do not depend on the mapping:

https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTSOE.py#L863-L875

Good catch @alixunderplatz !

lorrieq:
Makes use of `ENTSOE_PARAMETER_GROUPS` for code to fuel mapping as per issue. Fixes #1742.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -815,10 +815,11 @@ def fetch_production(zone_key, session=None, target_datetime=None,
             multiplier = -1 if key == 'storage' else 1
 
             for fuel, groups in parameter_groups.items():
-                value = sum([production_values.get(grp, 0) for grp in groups])
-                value *= multiplier
-
-                if fuel == 'oil' and value == -1.0:
+                has_value = any([grp in production_values.keys() for grp in groups])
+                if has_value:
+                    value = sum([production_values.get(grp, 0) for grp in groups])
+                    value *= multiplier
+                else:
                     value = None
 
                 production_types[key][fuel] = value
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -50,17 +50,21 @@ ENTSOE_PARAMETER_DESC = {
 }
 ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in ENTSOE_PARAMETER_DESC.items()}
 ENTSOE_PARAMETER_GROUPS = {
-    'biomass': ['B01', 'B17'],
-    'coal': ['B02', 'B05', 'B07', 'B08'],
-    'gas': ['B03', 'B04'],
-    'geothermal': ['B09'],
-    'hydro': ['B11', 'B12'],
-    'hydro storage': ['B10'],
-    'nuclear': ['B14'],
-    'oil': ['B06'],
-    'solar': ['B16'],
-    'wind': ['B18', 'B19'],
-    'other': ['B20', 'B13', 'B15']
+    'production': {
+        'biomass': ['B01', 'B17'],
+        'coal': ['B02', 'B05', 'B07', 'B08'],
+        'gas': ['B03', 'B04'],
+        'geothermal': ['B09'],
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -857,27 +806,22 @@ def fetch_production(zone_key, session=None, target_datetime=None,
...
-        production_values = {ENTSOE_PARAMETER_DESC[k]: v for k, v in
-                             productions[i].items()}
+        production_values = {k: v for k, v in productions[i].items()}
         production_date = production_dates[i]
 
+        production_types = {'production': {}, 'storage': {}}
+        for key in ['production', 'storage']:
+            parameter_groups = ENTSOE_PARAMETER_GROUPS[key]
+            for fuel, groups in parameter_groups.items():
+                value = sum([production_values.get(grp, 0) for grp in groups])
```
corradio: how will that behave if all `production_values.get(grp, 0)` are None?
In the previous behavior, if all group values are None then `value` should be None (indicates that all data is missing)

parsers/ENTSOE.py
```diff
@@ -857,27 +806,22 @@ def fetch_production(zone_key, session=None, target_datetime=None,
...
-                'oil': get_oil(production_values),
-                'solar': production_values.get('Solar', None),
-                'wind': get_wind(production_values),
-                'geothermal': get_geothermal(production_values),
-                'unknown': get_unknown(production_values)
-            },
+            'production': production_types['production'],
             'storage': {
-                'hydro': get_hydro_storage(production_values),
+                'hydro': production_types['storage'],
```
corradio: shouldn't the value be `production_types['storage']['hydro storage']`?

lorrieq: You're right, and I think I should multiply by -1

parsers/ENTSOE.py
lorrieq: Currently it will be 0. 

Can change to be None if all grp are 0. But gas is the only one I see that behaved like what you describe, the other fuels would get 0, is that desired?

lorrieq: Changed it to match what I think is the current behaviour, which is oil gets set to `None` if it is -1. Perhaps you want to improve this to something more along the lines of what you mentioned, let me know

corradio: The oil edge case should be removed (I don't know why it was there!)
However, what we want is to make sure that if all production_values are None, then value should be None.
I think the best is probably to add an `if` that checks if all values are None and sets value to None?

lorrieq: Thanks! That makes sense, I've added logic for that. Note its still possible for some very low values to get through if say one of the four coal codes has a value but I guess that's fine because it should be dealt with by `validate_production`?

parsers/ENTSOE.py
```diff
@@ -857,27 +806,30 @@ def fetch_production(zone_key, session=None, target_datetime=None,
...
+        production_values = {k: v for k, v in productions[i].items()}
         production_date = production_dates[i]
 
+        production_types = {'production': {}, 'storage': {}}
+        for key in ['production', 'storage']:
+            parameter_groups = ENTSOE_PARAMETER_GROUPS[key]
+            multiplier = -1 if key == 'storage' else 1
+
+            for fuel, groups in parameter_groups.items():
+                has_value = any([grp in production_values.keys() for grp in groups])
```
corradio: Here only the existence of the entry is checked for, but not the value itself.
I suggest
```suggestion
                has_value = any([production_values.get(grp) is not None for grp in groups])
```
As `production_values.get(grp)` would give None if a key is missing or if the value is None

Issue #166, opened 2016-12-13
Add Canada


corradio:
Ontario: http://live.gridwatch.ca/home-page.html

GuillaumeHo:
In Canada, the provinces have the control of their energy policy, hence the regulation of their electric power. 2 Provinces, Ontario and Alberta switched from regulated prices to an energy market. Here are the provinces and the pages or API I found to follow the production by type of energy and the exchanges with other provinces :

corradio:
Sounds terrific! I'll work on preparing the map itself so we have the right topologies.

jarek:
Hi, I saw that some Canadian data started appearing in Country Ranking and found the region=northamerica page. I would like to contribute a bit to this. Is anyone else working on any Canadian data right now? The Alberta data linked to by Guillaume looks like it'll be easy to integrate.

corradio:
Go ahead @jarek - would be awesome to have your contribution!
Let me know what you need, or ask us on Slack!

corradio:
Hey @jarek ? Are you on Slack? (http://slack.tmrow.co). We're preparing a big launch soon and it would be nice if AB would be there!

jarek:
Sorry @corradio, I ended up having no time so far :( I see mlucchini has done Alberta, congrats! I might try to integrate some other provinces.

I notice there are no capacities for Canada - should they be named with underscore (like `NZ-NZN` in zones.json) or with dash (like `CA_ON.py` in parsers)?

corradio:
Hi @jarek. No worries. Glad we did no do double work :)
Happy to get some capacities :)
They should be named with a dash (like in zones.json).

jarek:
There are no links to current generation numbers that I saw, but the "Provinces and Territories" sections of http://peoplepowerplanet.ca/ has a decent overview of electricity/energy sector per province, and might help people trying to figure out what numbers are what or what is roughly correct.

jarek:
I have also been working off Statistics Canada historical data, which is pretty good. Follow links from http://www5.statcan.gc.ca/cansim/a33?lang=eng&spMode=tables&themeID=1744&pattern=&stByVal=2&RT=TABLE, particularly "Table 127-0002: Electric power generation, by class of electricity producer; monthly (megawatt hour)" and "Table 127-0008: Supply and disposition of electric power, electric utilities and industry; annual"

[...]

PR #1791, opened 2019-03-10
update Prince Edward Island to new endpoint

jarek:
PEI data is currently stuck at source: https://www.princeedwardisland.ca/en/feature/pei-wind-energy/ says "Last updated March 10, 2019 08:45 AM" where current time there is 14:50.

The "old" API endpoint we were using returns the stuck data, but with updating timestamp, resulting in the following graph:

pei_not_updating

As we've known about a "new" endpoint from #166, and it (seemingly correctly) reports time of last update, switch to that one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Real-time electricity data is obtained using [parsers](https://github.com/tmrowc
 - Canada (New Brunswick): [NB Power](https://tso.nbpower.com/Public/en/op/market/data.aspx)
 - Canada (Nova Scotia): [Nova Scotia Power](http://www.nspower.ca/en/home/about-us/todayspower.aspx)
 - Canada (Ontario): [IESO](http://www.ieso.ca/power-data)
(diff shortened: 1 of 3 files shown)


jarek:
cc @reinvented, you might be aware, but just in case FYI: the request/response format seems to have changed a bit compared to the documentation in https://github.com/tmrowco/electricitymap-contrib/issues/166#issuecomment-441458506 and https://ruk.ca/content/new-api-endpoint-pei-wind - there is no more `prod` in the URL and `components` in response became `data`.

corradio:
@jarek should I merge?

jarek:
The old source meanwhile has updated data again. But this one is used by the official page https://www.princeedwardisland.ca/en/feature/pei-wind-energy/ and our old source http://www.gov.pe.ca/windenergy redirects there, so I guess this one's better.

jarek:
I'll merge, it seems fine, and currently gives same data but with seemingly more accurate timestamps.

Also this should fix #166 for all the available data sources in Canada.

Issue #1160, opened 2018-02-27
US: ERCOT access


jkfm:
Hi,

I was interested in adding data from ERCOT. I cannot access their website though and according to their helpdesk, one needs to fill out a detailed sheet to get your IP whitelisted. Did anyone here have communication with them?

Regards,
Johannes

jkfm:
Ok, so it seems American IP addresses can access the domain (I am in Germany). But the server running the parsers does not have a US IP address, right?

systemcatch:
Hi @jkfm I have tried going on to http://www.ercot.com/ before but as you say they implement geo blocking for all non American IP addresses. Unfortunately they also block their contact page and only provide a US phone number :roll_eyes:. I will check the site again later and see if anything else has changed.

corradio:
Same here. I can ask for access for the ElectricityMap project if you can
tell me the procedure!

jkfm:
Hi @corradio,

they sent me this via email: http://output.jsbin.com/zafexecasu
It'd be great if you could ask for access for the project, I guess I will try to write the parser with a proxy.

corradio:
Great I've sent them an email.

corradio:
I've been able to gain access for the Electricity Map project (from the EM server).
Is anyone able to work on a parser? I think that person would need to be cleared for access as well.

systemcatch:
What form is the data in Oli?

systemcatch:
I now have access to the site.  I've been looking around for a breakdown by fuel type but so far there's only a split into thermal, solar & wind.

systemcatch:
Quick update, though I have access to the main site I don't have access to mis.ercot.com which contains the solar and wind data.  I've sent an email about this but so far there's been no response (just like my access email).

corradio:
ok. They didn't answer me but gave me access anyway. Thanks for your
patience. Hope they fix it soon!

systemcatch:
Still no luck getting access to the part I need.  I suggest someone else build this parser so we aren't losing data, the EM server has full access right?

corradio:
There are DC flows here:
http://www.ercot.com/content/cdr/html/real_time_system_conditions.html
![image](https://user-images.githubusercontent.com/1655848/43391970-8975a0fe-93f2-11e8-8398-a25eef600fda.png)

THere's a lot of data but I didn't manage to find generation mix. I could only find wind/solar production and total production.

Here's the source code of the html in case someone wants to take a look at it:
[...]

PR #1796, opened 2019-03-15
Add US_TX production parser

typicalTYLER:
I've added a basic parser for ERCOT/US_TX as part of #1160 

Points to note:
- Didn't take into account the exchanges from this data (I can possibly add on later)
- There were multiple access points to ercot APIs used, I just put the most accessible one in the sources
- There's fluctuating total capacity in the real time data, I just sampled it at some point and put it in the capacities, not sure if this is best practice
- I'm new to this (and live in Texas :)) and guessed about a couple different things, let me know if I've implemented anything incorrectly and I can fix!


corradio:
Hi @typicalTYLER ,

Thank you for submitting this. From what I see the production mix only contains wind/solar.
In order to be show on the electricityMap, we need to be able to show a representative picture of the mix that sums up to the total production. Are you able to fetch the rest of the production? Maybe even putting it inside an "unknown" category?

Olivier

systemcatch:
Very nice @typicalTYLER! Can you post some output from your parsers?

typicalTYLER:
@systemcatch: here's the output from the parsers currently:
`fetch_production() ->
{'source': 'ercot.com', 'datetime': datetime.datetime(2019, 3, 15, 8, 25, tzinfo=tzfile('/usr/share/zoneinfo/US/Central')), 'production': {'wind': 9129.14, 'solar': 48.5}, 'zoneKey': 'US-TX', 'storage': {}}
fetch_consumption() ->
{'datetime': datetime.datetime(2019, 3, 15, 9, 29, 35, tzinfo=tzfile('/usr/share/zoneinfo/US/Central')), 'consumption': 39580.0, 'zoneKey': 'US-TX', 'source': 'ercot.eu'}`

(fortunately I live in the Texas so I'm not geo blocked by default)

@corradio: I was originally thinking unknown production was calculable from this API, but got a little confused by the terminology: image

Is the unknown production just calculated from: total demand - solar - wind - dc_ties? I also wasn't sure what the sign on the dc_ties implied (in vs out). Is the hourly data from the load report the same thing as the total demand from here as well?

typicalTYLER:
I've added the unknown production now, under the assumption that production + DC tie flows have to meet demand. Here's a sample output:
`fetch_production() ->
{'datetime': datetime.datetime(2019, 3, 15, 10, 40, tzinfo=tzfile('/usr/share/zoneinfo/US/Central')), 'production': {'solar': 1206.88, 'unknown': 25490.770000000004, 'wind': 11990.35}, 'source': 'ercot.com', 'storage': {}, 'zoneKey': 'US-TX'}
fetch_consumption() ->
{'consumption': 38876.0, 'datetime': datetime.datetime(2019, 3, 15, 11, 40, 35, tzinfo=tzfile('/usr/share/zoneinfo/US/Central')), 'source': 'ercot.eu', 'zoneKey': 'US-TX'}`

The math seems to cause a floating point error, not sure how y'all want me to handle that. Maybe round to two decimal places or just pass it as is.

corradio:
Looks great. Do anyone know what the average mix of "unknown" is? Do they have hydro or is it exclusively fossil-based?

typicalTYLER:
Looks like we have some Nuclear and some Hydro: image from here https://www.eia.gov/state/?sid=TX#tabs-4

Looks like this data is just from November though

typicalTYLER:
There's also this chart, but the data is older. Maybe it's more accurate for accounting for unknown, since I'm not sure what "net" electricity generation means in the previous graph. The large jump in fossil fuels is probably due to it accounting for the summer electricity production, which is a lot higher in magnitude due to how hot it gets here in Texas and how much air conditioning is used. image

typicalTYLER:
Found a better source and updated the unknown estimates and capacities: http://www.ercot.com/content/wcm/lists/172484/ERCOT_Quick_Facts_01.17.19.pdf

corradio:
Looks good to me. @systemcatch can you help doing another pass just to be sure?

typicalTYLER:
I was going to try to make the change on my phone, as I'm on vacation abroad currently and can't access a computer, but it looks like maybe my fork has a different file location for that file, so maybe it's out of date or something not trivial enough to do on my phone easily. Would it be possible for one of y'all to make the final change?

systemcatch:
@typicalTYLER I've made the change and will now merge this in, thanks again for the big contribution!

PR #1798, opened 2019-03-16
Improve DK logging when response is error or empty

lorrieq:
Addresses issue #1792

### DIFF
--- a/parsers/DK.py
+++ b/parsers/DK.py
@@ -1,16 +1,15 @@
 import logging
-import datetime
-
 import pandas as pd
-# The arrow library is used to handle datetimes
-import arrow
-# The request library is used to fetch content through HTTP
-import requests
+import arrow  # the arrow library is used to handle datetimes
+import requests  # the request library is used to fetch content through HTTP
 import pytz
 
 
-ids = {'real_time':'06380963-b7c6-46b7-aec5-173d15e4648b',
-       'energy_bal':'02356e88-7c4e-4ee9-b896-275d217cc1b9'}
+ids = {
+    'real_time': '06380963-b7c6-46b7-aec5-173d15e4648b',
+    'energy_bal': '02356e88-7c4e-4ee9-b896-275d217cc1b9'
+}
+
 
 def fetch_production(zone_key='DK-DK1', session=None,target_datetime=None,
                      logger: logging.Logger = logging.getLogger(__name__)):
@@ -19,8 +18,6 @@ def fetch_production(zone_key='DK-DK1', session=None,target_datetime=None,
     for Danish bidding zones
     """
     r = session or requests.session()
-    
-    
     
     if zone_key not in ['DK-DK1', 'DK-DK2']:
         raise NotImplementedError(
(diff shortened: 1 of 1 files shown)


parsers/DK.py
```diff
@@ -43,13 +40,18 @@ def fetch_production(zone_key='DK-DK1', session=None,target_datetime=None,
...
-    
-    assert response.status_code == 200 and response.json()['result']['records'] != [], \
-        'Exception when fetching production for ' \
-        '{}: error when calling url={}'.format(zone_key, url)
-        
+
+    if response.status_code != 200:
+        error = response.json()['error']['__type']
+        text = response.json()['error']['info']['orig']
+        logger.warning('"{}" fetching production data for {}: {}'.format(
```
corradio: here I would actually `raise Error(...)`

parsers/DK.py
```diff
@@ -43,13 +40,18 @@ def fetch_production(zone_key='DK-DK1', session=None,target_datetime=None,
...
+            error, zone_key, text))
+        return
+    if response.json()['result']['records']:
+        logger.info('API returned no data for {} production'.format(zone_key))
```
corradio: That's an error as well right? In that case, I would `raise`

parsers/DK.py
```diff
@@ -149,13 +146,18 @@ def fetch_exchange(zone_key1='DK-DK1', zone_key2='DK-DK2', session=None,
...
-    
-    assert response.status_code == 200 and response.json()['result']['records'] != [], \
-        'Exception when fetching flow for ' \
-        '{}: error when calling url={}'.format(sorted_keys, url)
-        
+
+    if response.status_code != 200:
+        error = response.json()['error']['__type']
+        text = response.json()['error']['info']['orig']
+        logger.warning('"{}" fetching exchange data for {}: {}'.format(
```
corradio: same here

parsers/DK.py
```diff
@@ -149,13 +146,18 @@ def fetch_exchange(zone_key1='DK-DK1', zone_key2='DK-DK2', session=None,
...
+            error, sorted_keys, text))
+        return
+    if response.json()['result']['records']:
+        logger.info('API returned no data for exchange {}'.format(sorted_keys))
```
corradio: same here

parsers/DK.py
lorrieq: Hmmm.. I don't think no data being returned is an error. I see this sort of thing on a lot of data sources where it temporarily doesn't return data (maybe when they're updating a file or whatever) so one may see it a number of times during the day.. which would be annoying if you do get all the data eventually just not on every call to the API and as a result you've got all these not very useful log messages.
Just my two cents, happy to go with whatever!

parsers/DK.py
lorrieq: Sure. `requests.exception.HTTPError` or `requests.exception.ConnectionError` would do?

corradio: right! You pick :)

parsers/DK.py
corradio: A parser should always return data. When a request for fetching data occurs, and that the parser is not able to deliver, then it should crash. This will ensure that the proper errors show up in kibana.
This is also helpful when collecting past data: it enables us to know when the data has been successfully retrieved.

corradio:
Thanks!

PR #1800, opened 2019-03-18
Change position of biomass and geothermal

alixunderplatz:
This will change the position of biomass and geothermal to a "baseload" position in the lower part of the area plots for the 24 hour production mix.
The almost constant production of these sources will be better comprehensible for the eye.

### DIFF
--- a/web/src/helpers/constants.js
+++ b/web/src/helpers/constants.js
@@ -16,14 +16,14 @@ const modeColor = {
 };
 const modeOrder = [
   'nuclear',
-  'coal',
   'geothermal',
+  'biomass',
+  'coal',
   'wind',
   'solar',
   'hydro',
   'hydro storage',
   'battery storage',
-  'biomass',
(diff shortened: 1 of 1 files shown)


corradio:
@alixunderplatz are you sure you want to put biomass before coal? I'd say coal is more baseload than biomass, but I might be wrong

brunolajoie:
It might be that biomass is more baseload, just because it burns less hot? A bit like lignite is more baseload than hard coal? And also because you'd prefer burn "renewable" (!) / "CO2 free" (!!) fuel than non-renewable one where you pay CO2 tax on it?

alixunderplatz:
I looked at all countries, biomass is very static and is rarely noticeably ramping up and down. Denmark might be the only exception because of biomass being used

alixunderplatz:
sorry for closing and reopening, I accidentally hit the wrong buttons ;)

PR #1802, opened 2019-03-21
Update default CO2 value to IPCC 2018

clementperon:
Data taken from the "Renewable Energy Sources and Climate Change Mitigation"
report from the IPCC publish in March 2018.

Annex II, page 982, Table A.II.4
Aggregated results of literature review of LCAs of GHG emissions from
electricity generation technologies as displayed in Figure 9.8.

Would like to point that there is a difference between Solar production:
Photovoltaics (PV) vs Concentrated Solar Power (CSP).

I propose to take the PV value as most of solar power (12TWh vs 1TWH in 2008)
is produced using this technology in the world.

Source: https://www.ipcc.ch/site/assets/uploads/2018/03/SRREN_Full_Report-1.pdf
Signed-off-by: Clément Péron <peron.clem@gmail.com>

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -12,24 +12,24 @@
         "value": 301.13684338512996
       },
       "biomass": {
-        "source": "IPCC 2014",
-        "value": 230
+        "source": "IPCC 2018",
+        "value": 18
       },
       "coal": {
-        "source": "IPCC 2014",
-        "value": 820
+        "source": "IPCC 2018",
+        "value": 1001
       },
       "gas": {
-        "source": "IPCC 2014",
-        "value": 490
+        "source": "IPCC 2018",
+        "value": 469
       },
       "geothermal": {
-        "source": "IPCC 2014",
-        "value": 38
+        "source": "IPCC 2018",
+        "value": 45
       },
       "hydro": {
-        "source": "IPCC 2014",
-        "value": 24
+        "source": "IPCC 2018",
+        "value": 4
       },
       "hydro charge": {
         "_comment": "Emissions are counted at discharge",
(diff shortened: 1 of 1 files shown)


clementperon:
image

jarek:
Technical note: if we update these values, we should also update all `zoneOverrides` except for hydro discharge and some of `fallbackZoneMixes` (at a first glance, `ZA`, `ZM`, `NL`, `RU-1`, `RU-2`, `SE`, `US-TX`, `GB-ORK`, `AZ`, `CA-BC`, `CA-QC`, `CA-NB`, `RU-KGD`, `US`). These are used for some zones where we don't have real-time information but wanted to estimate the carbon intensity of their exports, and in some cases where we have real-time information about "thermal" generation. They're calculated based on the IPCC 2014 values. I can do this if needed.

Also is the override for coal/heavy shale oil in Estonia as 1515 gCO2eq/kWh still valid?

brunolajoie:
Interesting! 

The IPCC report you mention is IPCC(2012) SRREN, which is not from march 2018. Let's compare these with our current median values, which comes from IPCC(2014) AR5 WGIII, which is more recent but less specialized in the topic of renewable. 

I see two big changes:

- Coal carbon intensity is significantly higher in the SRREN report (1001g) than on the AR5 (820g), which agrees with most feedback we've received so far complaining that coal emissions where too low at 820g.

- Biopower, on the other hand, has a much much lower carbon intensity in the SRREN (18) than in the AR5 (230)

I'm going to investigate why are these differences so large. Especially on the biopower side. If anyone has a clue, let us know

clementperon:
Oups sorry, you're right. I check the upload date on the URL not on the Doc :(.

PR #1812, opened 2019-03-29
Ukraine update

### LINKED ISSUE #1809: Discuss type for renewable category for Ukraine UA
Currently, there is no breakdown for the Ukrainian renewable category provided on https://ua.energy/diyalnist/dyspetcherska-informatsiya/dobovyj-grafik-vyrobnytstva-spozhyvannya-e-e/

The renewable category (ВДЕ) is mapped as "wind" in the parser, because wind used to be the dominant source in the past.
Going through the last few days on the website, you will notice a very clear solar peak at noon (~1.200. MW) each day. Wind at nighttimes reaches a maximum value of ~400 MW, mostly it is around 200 MW.

Here is an example for yesterday:
image

The installed capacity of solar grew very fast, and will continue because it's cheap and the potential in UA is huge:
image

Some suggestions to deal with this situation:
1. Any artificial boundaries (depending on x-axis-time or y-axis-megawatts or both) pushing production to wind or solar?
Like "from 06:00 to 18:00" -> solar if P > 200 MW, else wind". 
2. Put renewables to unknown category with a mixed carbon intensity (looking at the intalled capacity, 50% wind : 50% solar seems reasonable).
3. actively search for a breakdown of wind and solar
4. ask the data provider for a breakdown

alixunderplatz:
Closes #1809 
- add mixed "renewables" category from data provider to "unknown"
- add new carbon intensity based on estimated 50% solar/50% wind shares
- I also updated the renewable capacities to the latest available figures.

One more thing to do here: Set the "renewable ratio" for UA's "unknown" to 100%, if possible - where is the correct file to do this?

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -610,6 +610,11 @@
         "hydro discharge": {
           "source": "2018 average by Tomorrow",
           "value": 291.5736740216201
+        },
+        "unknown": {
+          "_comment": "estimated yearly share of solar & wind in this mixed renewable category based on installed capacity in 2018; Source: https://en.wikipedia.org/wiki/Renewable_energy_in_Ukraine",
+          "source": "renewable mix; assumes 50% solar, 50% wind",
+          "value": 28
         }
       },
       "US-CA": {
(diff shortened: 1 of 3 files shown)


corradio:
Thanks for the PR Alex. I've made the changes in the backend for the renewable ratio (that part isn't open source)

Issue #1563, opened 2018-08-16
Real-time production data for South Korea


scottlee120:
http://epsis.kpx.or.kr/epsisnew/selectMain.do

Data is provided in 5-minute intervals across the country as a whole. For now, there's only numbers on total production capacity (공급능력(MW)), actual production (현재부하(MW)), and a reserve capacity/ratio. Working on seeing if there's more detail on production mix. 

The EN and KR versions are not the same - for live data, it seems available only on the KR version although CSV exports are available only on the EN version seemingly.

scottlee120:
http://epsis.kpx.or.kr/epsisnew/selectMainEpsMep.ajax

Chart data raw format

scottlee120:
Historical data by year with a composition breakdown (thermal, nuclear, etc.)

http://epsis.kpx.or.kr/epsisnew/selectEkesKepGrid.do?menuId=010400

systemcatch:
Further info in #855  that might help you.

jarek:
Hi @scottlee120, were you able to find real-time production mix for South Korea?

alixunderplatz:
SUMMARY
-----
From the data sources presented here and in #855, we could write a parser with all nuclear + hydro generation and fill the rest as "unknown".
@brunolajoie @corradio do you think that is enough info to begin with?

__________________

- South Korea is an electrical island with no exchanges so far.
- Load is 70-90 GW at the moment.
- [Renewable electricity goal for 2030](https://renewablesnow.com/news/south-korea-plans-4-gw-solar-offshore-wind-complex-631803/) of 20% - oh, c'mon, thats all?!?
- General capacity data is available on [this website](http://cms.khnp.co.kr/content/163/main.do?mnCd=FN05040101) in MW:

nuclear | hydro (incl. pumped storage) | coal | oil | gas | RES | Sum
-- | -- | -- | -- | -- | -- | -- 
22,529 | 6,489 | 36,709 | 4,155 | 37,838 | 9,187 | 116,908

- RES breakdown [According to IRENA](http://resourceirena.irena.org/gateway/countrySearch/?countryCode=KOR) for 2017:
  - 5.06 GW solar
  - 1.2 GW wind energy
  - 1.07 GW biomass
  - That is 7.3 GW combined, so 1.4 GW are missing to fill the RES from the official Korean source.

- Monitored hydro capacity is 603 MW according to http://cms.khnp.co.kr/eng/content/566/main.do?mnCd=EN040201
[...]

PR #1816, opened 2019-04-05
Add South Korea

systemcatch:
This needs lots of cleanup and fixes.

- [x] CO2 value for unknown.
- [x] Config setup (zones, readme, en lang).
- [x] Cleanup all notes.
- [x] Make it more DRY.
- [x] Hydro capacity check.
- [x] Docstrings.
- [x] Add hydro and battery to storage.
- [x] Test on mockserver.

ref #1563


parsers/KR.py
```diff
@@ -0,0 +1,182 @@
...
+    production = {
+                  'production': {
+                    'nuclear': nuclear,
+                    'hydro': hydro,
+                    'unknown': unknown
+                  },
+                  'source': 'khnp.co.kr, kpx.or.kr',
+                  'zoneKey': zone_key,
+                  'datetime': dt_aware.datetime,
+                  'storage': {}
```
corradio: better just remove that line if you don't plan on putting storage

corradio:
Seems pretty cool if indeed wind/solar is such a small part

parsers/KR.py
systemcatch: actually according to Alex there's pumped storage and fuel cells (battery?) so I'll add them

systemcatch:
> Seems pretty cool if indeed wind/solar is such a small part

I think it's a good start, we can build on this further if/when more data becomes available. Also SK has no exchanges atm so that will stop it effecting the rest of the dataset.

systemcatch:
Seems to be basically working, looks a little funny with the coal/gas capacities included.

SK

config/co2eq_parameters.json
```diff
@@ -406,6 +406,14 @@
           "value": 413.76308999360253
         }
       },
+      "KR": {
+        "unknown": {
+          "_comment": "Source: Derived from 2016 IEA data",
```
brunolajoie:
```suggestion
          "_comment": "Source: Derived from 2016 annual average by IEA",
```

systemcatch:
@corradio good to go!

PR #1829, opened 2019-04-21
Update JP-CB parser, source format has changed

tmslaine:
JP-CB consumption CSV file has changed format
Makes the consumption readable again
Adds solar production parsing for JP-CB

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -59,8 +59,8 @@ def fetch_production(zone_key='JP-TK', session=None, target_datetime=None,
                 'geothermal': None,
                 'unknown': df.loc[i, 'unknown']
             },
-            'storage': {},
-            'source': ['occtonet.or.jp', sources[zone_key]]
+            'storage': None,
+            'source': 'occtonet.or.jp', sources[zone_key],
             }
         datalist.append(data)
     return datalist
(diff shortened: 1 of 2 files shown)


parsers/JP.py
```diff
@@ -38,12 +38,45 @@ def fetch_production(zone_key='JP-TK', session=None, target_datetime=None,
...
+                'hydro': None,
+                'nuclear': None,
+                'oil': None,
+                'solar': df.loc[i, 'solar'] if 'solar' in df.columns else None,
+                'wind': None,
+                'geothermal': None,
+                'unknown': df.loc[i, 'unknown']
+            },
+            'storage': {},
+            'source': ['occtonet.or.jp', sources[zone_key]]
```
corradio: can you write as a string instead of an array? Database schema will enforce this in the future

parsers/JP.py
```diff
@@ -38,12 +38,45 @@ def fetch_production(zone_key='JP-TK', session=None, target_datetime=None,
...
+                'gas': None,
+                'hydro': None,
+                'nuclear': None,
+                'oil': None,
+                'solar': df.loc[i, 'solar'] if 'solar' in df.columns else None,
+                'wind': None,
+                'geothermal': None,
+                'unknown': df.loc[i, 'unknown']
+            },
+            'storage': {},
```
corradio: Better put this key to `None` is no storage is expected

Issue #679, opened 2017-08-13
Data for the Philippines


jarek:
http://www.wesm.ph/ has hourly generation breakdown for Luzon and Visayas. These are two of Philippines' three major island groups, covering about 75% of the population.

According to news reports, the third island group, Mindanao, was to join or has joined WESM in June or July 2017, but there seems to be no data yet. Maybe they're still integrating it.

I would like to work on this next week :)

jarek:
I'm still digging through all the data and trying to get reliable real-time info for everything. Data so far:


## Major data sources

* http://wesm.ph/ has fuel breakdown and demand numbers for the current day
* http://wesm.ph/inner.php/downloads/ has past information and some future forecasts
* https://www.ngcp.ph/power-situation.asp seems to be updated a couple of times a day, including on weekends, but not really regularly, and doesn't have fuel mix breakdown of generation


## Regions and interchanges

Philippines electricity is split into three grids, roughly corresponding to Luzon, Visayas, and Mindanao (the three major island groups). These are not official "regions" or "provinces" and don't have their own ISO 3166-2 codes. I would probably suggest making codes up, e.g. PH-L, PH-V, and PH-M, to avoid conflicting with existing ISO 3166-2 codes. I looked into getting GeoJSON for them and it looks feasible from the Natural Earth dataset we use, with small changes to the shell script, I have it figured out.

There is currently no interchange between Visayas and Mindanao, see e.g. https://sg.news.yahoo.com/interconnecting-power-supply-mindanao-visayas-132845854.html

There is a connection between Luzon and Visayas, but its use is poorly reported (and not real-time). See schedules in https://www.ngcp.ph/weekly-outlook-vis.asp?location=LEYTE_SAMAR and reports in [previous-day summary (PDS)](http://wesm.ph/inner.php/the_market/wholesale/market_data_and_reports/daily_operations_reports/previous_day_summary/visayas_pds).

On 2017-09-01, [next-day summary (NDS)](http://wesm.ph/inner.php/the_market/wholesale/market_data_and_reports/daily_operations_reports/next_day_summary/visayas_nds) indicated that on 2017-09-02 (Saturday), Visayas would import a peak of 333 MW from Luzon on top of its own 1521 MW local generation, so a non-trivial amount.

[...]

PR #1830, opened 2019-04-22
Create PH.py

plutocrat:
Created stats file for Philippines.


jarek:
Hi @plutocrat, thanks for the pull request. It seems to be static monthly data, but Electricity Map focuses on real-time information.

There is some real-time data for the Philippines - see #679 for past exploration - when I looked, the numbers didn't add up, but it looks like there have been some changes and a new organization http://www.iemop.ph/ if you would like to take a look.

plutocrat:
OK, sorry, that wasn't clear. Just trying to help out. I'll withdraw the submission then .... erm ... how do I do that?

Issue #1833, opened 2019-04-25
Check/fix parsers of "lost" areas


alixunderplatz:
In the last couple of weeks, some areas on the map went grey.
Here's a list with regions that need a parser fix / investigation of the source:

- [x] **US-BPA** is grey (Washington (state))
 
- [x] **US-IPC** is grey (Idaho)
  - data url was changed, I think this data source could no more be available in real-time
  - see #1803 for latest commit

- [x] **IT-South** is grey + IT<->GR is missing (ENTSOE.py)
  - check if the ENTSOE-ID of the bidding zone has been changed
  - see #1770 and #1776 for latest commit
 
- [x] **CA-AB**
  - see @jarek's comment below

- [x] **GT**
  - pandas error seen repeatedly

- [x] **JP-HKD/JP-SK**
  - storage is wrong format

- [x] **DO**
  - source website does not respond
  - there's a note on the [website](http://www.sie.gob.do/) about server-issues for the market participants

- [x] **KR**
  - source for Korea not available at the moment
  - switching to another data source is discussed in #1563

jarek:
Alberta, too. I've been meaning to check on it, page http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet still loads...

systemcatch:
US-BPA data link seems to be working fine (https://transmission.bpa.gov/business/operations/Wind/baltwg.txt)

Logger is showing an error that's related to pandas.

```python
Traceback (most recent call last):
  File "/home/feeder/lib/fetch_data.py", line 171, in launch_parsers
    objs = res.get(timeout)
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/feeder/lib/fetch_data.py", line 169, in <lambda>
    **parser_kwargs))
  File "/home/contrib/parsers/US_BPA.py", line 114, in fetch_production
    processed_data = data_processor(raw_data, logger)
  File "/home/contrib/parsers/US_BPA.py", line 71, in data_processor
    dt = dt.to_datetime()
AttributeError: 'Timestamp' object has no attribute 'to_datetime'
```

systemcatch:
[...]

PR #1835, opened 2019-04-26
Update to fix US-BPA

systemcatch:
- pandas to_datetime() is deprecated, use to_pydatetime() instead.

ref #1833

### DIFF
--- a/parsers/US_BPA.py
+++ b/parsers/US_BPA.py
@@ -68,7 +68,7 @@ def data_processor(df, logger):
         production = row.to_dict()
 
         dt = production.pop('Date/Time')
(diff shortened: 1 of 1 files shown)


jarek:
As far as I can tell the circleci failure is just a timeout, try poking it again to see if it succeeds now?

systemcatch:
Tried rerunning the build but no luck. I see one error listed;

> ERROR: docker-compose 1.17.1 has requirement requests!=2.11.0,<2.12,>=2.6.1, but you'll have requests 2.20.1 which is incompatible.

corradio:
I'll try to fix the build error separately

PR #1837, opened 2019-04-30
Create language selector

### LINKED ISSUE #437: Let people switch between languages easily
I was just thinking that it can be a good idea to let people switch between languages.

jld3103:
I just had to create this feature because it bothered me a lot not to use the map in another language right away. This feature is my first real contribution to this project and I am sure that much of my implementation can be improved.

Fixes #437 

electricitMap_select_language


web/locales/de.json
```diff
@@ -1495,5 +1496,25 @@
...
+        "EN": "Englisch",
+        "ES": "Spanisch",
+        "FR": "Französisch",
+        "IT": "Italienisch",
+        "JA": "Janpanisch",
+        "NL": "Niederlöändisch",
+        "PL": "Polnisch",
+        "PT-BR": "Brazilianisches Portugiesisch",
+        "RU": "Russisch",
+        "SK": "Slowenisch",
```
jarek: Caution, SK is Slovakia / de: Slowakei, where the language is Slovakian / de: Slowakisch.

Slowenisch is for Slovenia / de: Slowenien / SI.

web/locales/de.json
```diff
@@ -1495,5 +1496,25 @@
...
+        "AR": "Arabisch",
+        "CS": "Tschechisch",
+        "DA": "Dänisch",
+        "DE": "Deutsch",
+        "EN": "Englisch",
+        "ES": "Spanisch",
+        "FR": "Französisch",
+        "IT": "Italienisch",
+        "JA": "Janpanisch",
+        "NL": "Niederlöändisch",
```
jarek: Typo here "Niederlöändisch"

web/locales/de.json
```diff
@@ -1495,5 +1496,25 @@
...
+    "localeNames": {
+        "AR": "Arabisch",
+        "CS": "Tschechisch",
+        "DA": "Dänisch",
+        "DE": "Deutsch",
+        "EN": "Englisch",
+        "ES": "Spanisch",
+        "FR": "Französisch",
+        "IT": "Italienisch",
+        "JA": "Janpanisch",
```
jarek: Typo here

web/locales/de.json
```diff
@@ -1495,5 +1496,25 @@
...
+        "NL": "Niederlöändisch",
+        "PL": "Polnisch",
+        "PT-BR": "Brazilianisches Portugiesisch",
+        "RU": "Russisch",
+        "SK": "Slowenisch",
+        "SV": "Schwedisch",
+        "ZH-CN": "Mainland China Chinesisch",
```
jarek: I think in German this could be called "Festlandchina", see https://de.wikipedia.org/wiki/Festlandchina

And how about phrasing these as "Chinesisch (Festlandchina)", "Chinesisch (Taiwan)", "Chinesisch (Hong Kong)?

jdauphant: We should have all languages in the language of the reader. If I speak French, it is more easy to select "Français" than "Französisch" ;)  
Something like :
Deutsch,
English,
Français
中文

web/locales/de.json
```diff
@@ -91,7 +91,8 @@
         "temporaryDataOutage": "Live-Daten sind momentan leider nicht verfügbar",
         "production": "Produktion",
         "consumption": "Verbrauch",
-        "cpinfo": "<b>Verbrauch</b> bezieht Importe und Exporte mit ein, <b>Produktion</b> ignoriert sie"
+        "cpinfo": "<b>Verbrauch</b> bezieht Importe und Exporte mit ein, <b>Produktion</b> ignoriert sie",
+        "selectLanguage": "Sprache auswählen"
```
jdauphant: "Sprache auswählen" is difficult to find in the page for people that don't speak German (there is more chance that if you want to switch language: you are not fluent in the current language). 
Maybe we can use a symbol (could be the country flag but it's not the best, language are not link to one country and it could be difficult if you don't know the flag) or simply use the English word

web/locales/de.json
jdauphant: And providing only the English translation if we don't have the language name in the langage. In that case, my hypothesis : there is more chance that a reader know the name of his language in English than German.

web/locales/de.json
jld3103: Didn't know that.

web/locales/de.json
jld3103: @jarek I Agree with you

web/locales/de.json
jld3103: I wanted the languages in their own languages, but I didn't know how to translate all those names.

jld3103:
@jdauphant There is not only a german translation!!! In the first commit is the default english translation. I added the german translation, because I could.

web/locales/de.json
jld3103: If you look in the screenshot I send you will see that there is a translate symbol. The text you mentioned is only the tooltip.

jld3103:
It looks like we have two choices:
1. Translate all language names into their own language
2. Have the language names only in english

systemcatch:
@jld3103 I think having the language names in their own language is a better idea (see https://ux.stackexchange.com/questions/37017/language-of-language-names-in-the-language-selector for some interesting ux discussion).

corradio:
Very nice to see this @jld3103 !
Are you able to put the language icon at the bottom and show a full-sized screenshot of the map when button is toggled/untoggled as well as in mobile mode?
Thanks!

systemcatch:
em_lang

jld3103:
@corradio If you put the button on the bottom the list of languages gets cut off by the screen.
I have no experience with the mobile app.

jld3103:
@systemcatch How do we translate the languages?

corradio:
Thanks @systemcatch this looks awesome.
@jld3103 I meant putting the button at the bottom of the existing list of buttons (wind, solar..) instead of at the top.

jld3103:
I know what you mean :) I had it previously, but the list was cut off.

corradio:
@jld3103 that is weird. The list has the same length regardless of the order, so it shouldn't be cut off.

jld3103:
I mean the list then appears lower.

systemcatch:
>  How do we translate the languages?

I'm not really sure how to implement it. There are a few text segments of the map that don't change with translation, the language list would need to be similar (static object). This would save doing the translations for every language as well.

corradio:
Why not hardcode the language list? In most systems, the list of languages is a list where each language is written in the language of the language. That list doesn't change for different languages.
See for example:
image

jld3103:
I'm going to copy a list of these languages and hardcode them.

PR #1840, opened 2019-05-02
Updated language selector

Raffox97:
* Added missing string in Italian for the new language selector
* Fixed wrong casing for language names in the language selector


web/src/helpers/translation.js
```diff
@@ -44,20 +44,20 @@ exports.getFullZoneName = function getFullZoneName(zoneCode) {
...
-  { shortName: 'fr', name: 'français' },
-  { shortName: 'it', name: 'italiano' },
+  { shortName: 'es', name: 'Español' },
+  { shortName: 'fr', name: 'Français' },
+  { shortName: 'it', name: 'Italiano' },
   { shortName: 'ja', name: '日本語' },
   { shortName: 'nl', name: 'Nederlands' },
-  { shortName: 'pl', name: 'polski' },
-  { shortName: 'pt-br', name: 'português (Brazilian)' },
+  { shortName: 'pl', name: 'Polski' },
```
jarek: I can't speak for other languages, but I can tell you for sure this is wrong - we don't capitalize language names in Polish as is done in English or German.

jld3103: And why did you remove pt-br?

jarek: pt-br is below, the display in Github is misleading because I attached my comment to the pl line

Raffox97:
I'm closing that pull request for now, since I thought it's better for every language, to decide for it's own.

PR #1845, opened 2019-05-05
Fix CA-AB

alixunderplatz:
Maybe the tables on the website were rearranged a bit. 
Setting  `total_net_generation = df_generations[1]..`. to     `total_net_generation = df_generations[2]...` has fixed CA-AB :)
ref #1833

Sample output:
```python
{'datetime': datetime.datetime(2019, 5, 5, 2, 33, tzinfo=<DstTzInfo 'Canada/Mountain' MDT-1 day, 18:00:00 DST>), 'zoneKey': 'CA-AB', 'production': {'coal': 3065.0, 'gas': 5269.0, 'hydro': 84.0, 'wind': 24.0, 'unknown': 192.0}, 'capacity': {'coal': 5723.0, 'gas': 7641.0, 'hydro': 894.0, 'wind': 1445.0, 'unknown': 438.0}, 'source': 'ets.aeso.ca'}
```

### DIFF
--- a/parsers/CA_AB.py
+++ b/parsers/CA_AB.py
@@ -64,8 +64,8 @@ def fetch_production(zone_key='CA-AB', session=None, target_datetime=None, logge
     dt = convert_time_str(time_string)
 
     df_generations = pd.read_html(response.text, match='GENERATION', skiprows=1, index_col=0, header=0)
-    total_net_generation = df_generations[1]['TNG']
-    maximum_capability = df_generations[1]['MC']
+    total_net_generation = df_generations[2]['TNG']
+    maximum_capability = df_generations[2]['MC']
 
     return {
         'datetime': dt,


alixunderplatz:
@corradio seems like the fix for CA-AB didn't work, the same error note appears on kibana :/ maybe we need to handle cookies of that website or so?

corradio:
@alixunderplatz what is the error reported by the parser?

alixunderplatz:
@corradio on kibana it says this:

Traceback (most recent call last):   File "/home/feeder/lib/fetch_data.py", line 171, in launch_parsers     objs = res.get(timeout)   File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get     raise self._value   File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker     result = (True, func(*args, **kwds))   File "/home/feeder/lib/fetch_data.py", line 169, in <lambda>     **parser_kwargs))   File "/home/contrib/parsers/CA_AB.py", line 67, in fetch_production     total_net_generation = df_generations[2]['TNG'] IndexError: list index out of range

"list index out of range" is what we used to see before as well, afaik

corradio:
It seems like `df_generations` has less than 3 rows, given the error.
Is there a way to reproduce the error? Is there anything we could log to help us get to a reproducible state?

PR #1854, opened 2019-05-15
Updated Austria

Bohne13:
Updated installed capacity of Austria from https://www.apg.at/de/markt/Markttransparenz/erzeugung/installierte-leistung

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -98,7 +98,7 @@
       "biomass": 500,
       "coal": 598,
       "geothermal": 1,
-      "storage": 2439,
+      "battery storage": 2439,
       "gas": 4468,
       "hydro": 5557,
       "hydro storage": 3120,
@@ -106,7 +106,7 @@
       "oil": 178,
       "solar": 1193,
       "wind": 3035,
-      "unknown": 115,
+      "unknown": 115
     },
     "contributors": [
       "https://github.com/corradio"
--- a/config/zones.json
+++ b/config/zones.json
@@ -95,15 +95,18 @@
       ]
     ],
     "capacity": {
-      "biomass": 474,
+      "biomass": 500,
       "coal": 598,
-      "gas": 4466,
-      "hydro": 8590,
-      "hydro storage": 3401,
+      "geothermal": 1,
+      "storage": 2439,
+      "gas": 4468,
+      "hydro": 5557,
+      "hydro storage": 3120,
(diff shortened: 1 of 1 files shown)


config/zones.json
```diff
@@ -95,15 +95,18 @@
...
     ],
     "capacity": {
-      "biomass": 474,
+      "biomass": 500,
       "coal": 598,
-      "gas": 4466,
-      "hydro": 8590,
-      "hydro storage": 3401,
+      "geothermal": 1,
+      "storage": 2439,
```
corradio: do you mean battery storage here?

config/zones.json
```diff
@@ -95,15 +95,18 @@
...
+      "gas": 4468,
+      "hydro": 5557,
+      "hydro storage": 3120,
       "nuclear": 0,
       "oil": 178,
-      "solar": 723,
-      "wind": 2696
+      "solar": 1193,
+      "wind": 3035,
+      "unknown": 115,
```
systemcatch: The comma here is causing the test failure I think.

config/zones.json
Bohne13: it's not completely clear if it's battery storage. I think it's all types of storage excluded hydro storage.

I don't know how to handle this?

corradio: Hmm that's tricky. The thing is that in order for the UI to be consistent, capacities should be tied to the data we're receiving from ENTSO-E.
Currently, ENTSO-E just has hydro storage. It would be interesting to understand what other type of storage is available. Because the storage you mention is non negligible, we need to know what it represents.

alixunderplatz: @Bohne13 "Speicher", in this case, refers to hydro storage, without the capability to pump back water to the storage (so non-pumped storage).
These hydro plants have natural sources of water (rivers, glaciers, ...). They are also referred to as "Saisonspeicher", because usually they fill during springtime and summer. 
In the past, we always combined this category with run-of-river/usual hydro. 
So you must add the 2439 MW to the hydro category, then it's fine.

config/zones.json
```diff
@@ -98,15 +98,15 @@
       "biomass": 500,
       "coal": 598,
       "geothermal": 1,
-      "storage": 2439,
+      "battery storage": 2439,
```
alixunderplatz: You must add 2439 to "hydro", see my other comment in the PR for full info.

PR #1856, opened 2019-05-18
Add partial translation to Croatian

fbarl:
* Added Croatian to the list of languages and added a JSON template
* Translated most of the country names and about 15 extra terms

#### Screenshots

image
image

### DIFF
--- a/mobileapp/generate-index.js
+++ b/mobileapp/generate-index.js
@@ -26,7 +26,7 @@ var VENDOR_STYLES_HASH = getHash('vendor', 'css');
 // TODO:
 // Currently, those variables are duplicated from server.js
 // We should instead have a central configuration file in the `config` folder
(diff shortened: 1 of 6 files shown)


fbarl:
@systemcatch thanks for such a quick response!

I've added some screenshots to the PR description - I'm aware a lot of terms remain untranslated, but I didn't want this PR to grow too big, so I'd rather try translating the rest of it in subsequent PR(s).

fbarl:
> Can you also update mobileapp/www/index.html and mobileapp/generate-index.js?

Actually I added `'hr'` to locales in both of those files in the first commit. Is there anything else I missed there?

> At some point it would make sense to unify all those.

Yes, I noticed that - I just opened #1862 to address it separately.

fbarl:
Thank you for a thorough review @systemcatch @corradio! :+1:

PR #1861, opened 2019-05-19
Add transfer capacties for Norway

Gazer75:
Values are based on max reported to ENTSOE and will obviously be lower in summer due to thermal limits.


corradio:
I'll let @alixunderplatz double check

Gazer75:
That's ok. This could be done for all countries within ENTSOE area that report this. I was a bit unsure about the NO<->SE due to the missing proper regions for SE.

alixunderplatz:
@corradio hehe thanks^^
In general, we have already added such AC-tie line capacities in Australia, but in that case, they were dynamic. Transmission capacities can be shifted with generation and demand in the respective areas and also with availability of transmission lines and power plants. 
I think it's okay to use the AC-transmission capacity for the interconnections between the zones here. But in this case, they are static, and in reality they will probably change over time, as @Gazer75 also mentioned. But as you can see on the different figures for the same interconnectors, imports/exports to some areas are limited in certain directions, which is what is important. I approve ;)
I don't have the time to look into NO<->SE, I guess we will see if it fits or not.^^

Parsing these transmission capacities from ENTSO-E in the future really could be an option, but let's discuss this in another issue.

Gazer75:
I was thinking the same. A parser might be able to read this as most are reported daily or monthly to ENTSO-E.

Issue #1860, opened 2019-05-19
Split Azerbaijan and Nagorno-Karabakh Republic networks


Varmin123:
Hi everyone,

First of all, I wanted to thank you for adding recently Armenia. However, it seems that there is a little detail lacking in the CO2 emissions.

Indeed, electricity imports from OHTL Artsakh and OHTL Shinuhayr are counted as imports from Azerbaijan. In practice, those imports come from Nagorno-Karabakh Republic (NKR), a de facto independant state. The networks of AZ and NKR have been separated (the transmission lines have been cut), and no electricity is exchanged between both.

I would suggest splitting AZ in half (following the Wikipedia map: https://commons.wikimedia.org/wiki/File:Republic_of_Artsakh_location_map_(dotted_de-facto)_(-NKR_provinces)_(%2BAzerbaijan).svg ), as there are actually two independent networks operating in the same country (hereafter the AZ network and the NKR network). 

The CO2 emissions of 490g/kWh given in co2eq_parameters.json are only for the AZ network, while the NKR network is made only of hydroelectric power plants (Sarsang HPP 50 MW, + other small HPP: Trghi-1 3MW, Trghi-2 6 MW, Mataghis-1 and -2 7,8 MW, Trghi-3 5 MW, Trghi-4 - 0.8 MW and Lev-1 - 7 MW + other HPP being built). The carbon content of the NKR network is assumed to be 24g/kWh (100% hydro).

After splitting the AZ network into two (AZ + NKR), the other networks would be modified as follows: 
- Armenia imports from NKR network
- Everybody else imports from AZ network
- No connection between AZ and NKR networks

I could submit the pull request from the last part, but I do not know how to do the changes on the map. If I could get some help, I would be really glad.

Thank you very much.

jarek:
Thanks for the great description of the system in place! That'll be really helpful.

Varmin123:
#1866

Varmin123:
Done.

PR #1866, opened 2019-05-23
Split Azerbaijan and Nagorno-Karabakh Republic networks

Varmin123:
Here is a first commit adding some of the backend features for splitting AZ into AZ + NKR (per #1860 ). 

I have no idea how to split the map though (do I need to write the polygons one by one ?).

Varmin123.


systemcatch:
Hi @Varmin123 

To split the map requires alterations to https://github.com/tmrowco/electricitymap-contrib/blob/master/web/generate-geometries.js#L132

First thing to do is download https://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-1-states-provinces/ then open it on https://mapshaper.org/

We normally split countries using adm1 codes or code hascs.

Varmin123:
Hi @systemcatch 

Well, it seems that the splitting is already done in the ADM0 ( https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_disputed_areas.zip ) : Admin 0 – Breakaway, Disputed Areas
NKR has the code NK from what I could understand. In my commit, I added a new line in generate_geometries (with the code NKR). Do you think that just changing it to NK would solve the issue ?

Varmin123:
I added NKR geometry as a third party map with the right ID. I load it in generate_geometries. Does it work as expected (we should see NKR as an overlay over AZ).

systemcatch:
> Well, it seems that the splitting is already done in the ADM0 ( https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_disputed_areas.zip ) : Admin 0 – Breakaway, Disputed Areas

Yes but we do not use that file. I didn't realize that Azerbaijan needed to be split rather than Armenia, but we don't need a new geometry to do that, so can you remove it?

Take a look at https://github.com/tmrowco/electricitymap-contrib/blob/master/web/generate-geometries.js#L468, this is how we split Russia. Each region/state has an adm1_code ('RUS-2280', 'RUS-2416', 'RUS-3200', 'RUS-2356' etc), Azerbaijan is the same (e.g. AZE-1715).

What we need to do is get every adm1_code for Azerbaijan then create 2 new areas from these codes, NKR and AZ.

Does this make sense?

web/generate-geometries.js
```diff
@@ -17,6 +17,7 @@ const countryGeos = readNDJSON('./build/tmp_countries.json');
 const stateGeos = readNDJSON('./build/tmp_states.json');
 const thirpartyGeos = readNDJSON('./build/tmp_thirdparty.json').concat([
     require('./third_party_maps/DK-DK2-without-BHM.json'),
+    require('./third_party_maps/NKR.json'),
```
systemcatch: No need to add a new geometry here.

web/generate-geometries.js
```diff
@@ -419,6 +420,7 @@ const zoneDefinitions = [
   { zoneName: 'NF', type: 'country', id: 'NFK'},
   { zoneName: 'NG', type: 'country', id: 'NGA'},
   { zoneName: 'NI', type: 'country', id: 'NIC'},
+  { zoneName: 'NKR', type: 'country', id: 'NKR'},
```
systemcatch: It needs to be in this form.

`{ zoneName: 'NKR', type: 'administrations', administrations: [insert adm1_codes here]},`

systemcatch: The same will need to be done for `AZ`.

Varmin123:
I see what you mean, but I found it hard to split, as the boundaries do to totally follow the ones in ADM (the following map shows the difference): https://upload.wikimedia.org/wikipedia/commons/d/d5/Artsakh_line_of_contact.png

I could try splitting following the lines, but some parts may have to be either split, or arbitrarily attached to one or the other part.

web/generate-geometries.js
Varmin123: In such case, the following rayons would go in NKR:
Kalbajar, Lachin, Qubadli, Zangilan, Jabrayil, Susha District, Susha City, Tartar, Khankendi, Khojali, Khojavend, Agdam. The only one strange is the Fizuli district that should be cut in half, but for simplicity we can leave it in main AZ. I'll try doing a commit for it.

systemcatch:
> I could try splitting following the lines, but some parts may have to be either split, or arbitrarily attached to one or the other part.

That's fine, just do the best you can to approximate the area.

Varmin123:
New trial. Note that I have no mockserver on my computer. If someone could double check that it renders correctly, that would be nice.

systemcatch:
This is what the map looks like once I got it to build.

NKR_test

Varmin123:
Oops. It seems I mixed some regions. I'll have to do it one by one. I'll fix it tomorrow.

Varmin123:
I changed the number of the regions, based this time on the region names (and not on their number as before). It is supposed to fix the problem. Could you please check ?

systemcatch:
You need to add `AZE` to https://github.com/tmrowco/electricitymap-contrib/blob/master/web/topogen.sh#L9

NKR_better

Varmin123:
Done.

systemcatch:
Ok good, here is how things look at the moment.

NKR_demo

To actually generate the new world.json you'll need to give me push access to your branch.

config/co2eq_parameters.json
```diff
@@ -1212,6 +1212,12 @@
       "fossilFuelRatio": 0.35349837676303403,
       "renewableRatio": 0.646501623236966
     },
+    "NKR": {
+      "_comment": "Average carbon intensity using 2016 mean production per fuel source (100% hydro) and IPCC 2014 default lifecycle emission factor",
+      "carbonIntensity": 24,
+      "fossilFuelRatio": 0.0,
+      "renewableRatio": 0.00
```
corradio: if it's 100% hydro, doesn't it mean renewableRatio should be 1.0?

Varmin123: Definitely, I patched it.

corradio:
@systemcatch I'll let you have the final word!

systemcatch:
Hey @Varmin123, thanks for being patient with this. All that needs to be done now is updating the readme with the data source info for NKR. Then we can merge this in!

Varmin123:
I added the source in the "Production capacity data sources" section. If I find other sources for transmission capacities between NKR and AM, I'll definitely add them in another PR. Thanks for the help on this PR.

systemcatch:
> I added the source in the "Production capacity data sources" section. If I find other sources for transmission capacities between NKR and AM, I'll definitely add them in another PR. Thanks for the help on this PR.

That's great, can you also add the production data to https://github.com/tmrowco/electricitymap-contrib#real-time-electricity-data-sources

Varmin123:
I fear that there is no production data for NKR yet (the section is for live data sources from what I understood). Armenia is already added.

systemcatch:
Great ok, thanks for your help and patience!

Issue #1867, opened 2019-05-24
Fixing Unknown Sources for Karnataka - India


zuron7:
The unknown sources (CGS - Central Generating Stations) consists of 20% Nuclear and 80% Thermal (Coal Based). The source for this can be found at this link.

http://srldc.org/2018-19srallocation.aspx

This sharing agreement changes quarterly. The data can be found in Table II. 

I've attached a picture of the data with Nuclear marked in green and Coal marked in brown.

![image](https://user-images.githubusercontent.com/7473369/58310634-bad5b180-7e24-11e9-9f03-5812fbe9f1ef.png)

Based on this, the percentage share is given as:

![image](https://user-images.githubusercontent.com/7473369/58310666-cb862780-7e24-11e9-8bac-54c798104dac.png)

Though the max draw is 3915 MW, the state has not drawn more than 3018 MW from the CGS. Proof: http://kptclsldc.com/Load%20curve.aspx

As for the Co Generation Station, all of it comes from the burning of Sugarcane Bagasse.
Information here: http://kredlinfo.in/General/KERC%20tariff%20orders/RE%20Tariff%20Order%202018.pdf
https://www.karnataka.gov.in/kerc/Court%20Orders/COURT-ORDERS-2005/CogenerationPPA.pdf

zuron7:
How do I change the proportion of Unknown sources? Like how it has been done for Russia and Netherlands.

PR #1868, opened 2019-05-25
Update Karnataka Unknown Types

zuron7:
From #1867 
Do tell me if I have to make any changes.

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -356,6 +356,14 @@
         "hydro discharge": {
           "source": "2017 average by Tomorrow",
           "value": 589.9391089569107
+        },
+        "unknown": {
+          "_comment": "Source: Souther Regional Load Dispatch Center Allocation Limits",
+          "_url": [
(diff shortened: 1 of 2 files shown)


systemcatch:
Hi @zuron7 thanks for the contribution, codewise this looks fine. I'd just like to have @alixunderplatz  check the CO2 figures.

corradio:
@alixunderplatz there's a `renewableRatio` key that you can use for this purpose. It's in the same configuration file if I'm not mistaken.

zuron7:
So, I just set
fossilFuelRatio to 0.80
and 
renewableRatio to 0.20
?

corradio:
Sorry I was a bit fast to respond. Those two keys represent a fallback when data is not present.
What we're looking for here is to override the fact that `unknown` maps to fossil by default. It shouldn't.
The code currently runs on our backend.
What I suggest is to create an issue to make sure this part of the code is open-sourced and can be configured from the JSON file.
How does that sound?

PR #1873, opened 2019-05-28
Show translation status through the API

### LINKED ISSUE #1838: Translation status
We should have some kind of translation report to know which languages need to be translated (a bot or something).
Maybe a small custom translation UI would be cool as well.

fbarl:
Resolves #1838, although in a bit different way.

Adds a couple of endpoints to the API that indicate translation status:

* `/translationstatus/:language` - returns a JSON with the translation status for a specific language, e.g. Croatian: `{"language":"hr","translated":"40.12%","legacyTerms":17,"color":"red"}`.
  * _translated_ percentage is taken against the valid terms present in English translation
  * _legacyTerms_ indicates legacy terms that have been removed from the English version
* `/translationstatus` - returns an array of translation statuses across all languages in the locales, sorted descendingly by the translation percentage
* `/translationstatus/badges.svg` - generates and returns the aggregate SVG image containing per-languages [badges](https://github.com/badges/shields) (see the screenshot) - hopefully, this is something that could be inserted directly into the README to encourage contributors in adding new translations and also staying on top of the existing ones as terms keep changing

image


web/locales/translation-status.js
```diff
@@ -0,0 +1,65 @@
+const { reverse, flatMap, difference, size, sortBy } = require('lodash');
```
corradio: I would recommend moving this file to web/translation-status.js as web/locales should only contain the translation files themselves.
We would also need to update the Dockerfile to make sure that this file is indeed added to the build.

web/server.js
```diff
@@ -123,6 +125,14 @@ function handleError(err) {
 app.get('/health', (req, res) => res.json({status: 'ok'}));
 app.get('/clientVersion', (req, res) => res.send(BUNDLE_HASH));
 
+// Translation status API
+app.get('/translationStatus/badges.svg', (req, res) => {
```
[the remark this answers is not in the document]
fbarl: Sure, I just did it like this because of the `/clientVersion` above - can I change it to `/client-version` while I'm at it?

fbarl:
@corradio PTAL

web/server.js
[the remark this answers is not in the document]
corradio: Good catch. `/clientVersion` is used by the frontend to figure out if a new version of the client is available. Changing it would mean update the frontend and would also mean causing errors for older versions of the frontend. I suggest keeping `clientVersion` as is but changing `/translationStatus` to `/transationstatus`.

What do you think?

web/translation-status.js
```diff
@@ -0,0 +1,65 @@
+const { reverse, flatMap, difference, size, sortBy } = require('lodash');
+const { LOCALES_LIST } = require('./src/helpers/translation');
```
corradio: requiring `./src/helpers/translation` means we need to add it to the Dockerfile as well.

web/server.js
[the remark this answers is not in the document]
fbarl: Sure, that makes sense! :+1:

web/translation-status.js
fbarl: :thinking: Hmm, I actually don't think either `web/src/helpers/translation.js` or `web/translation-status.js` need to be in the Dockerfile as they are already being copied across in https://github.com/tmrowco/electricitymap-contrib/blob/ddb84f400c34d6f6c85cf670e6673efcea3a9bce/web/Dockerfile#L23
not referenced before that, and only needed in the next line https://github.com/tmrowco/electricitymap-contrib/blob/ddb84f400c34d6f6c85cf670e6673efcea3a9bce/web/Dockerfile#L24

Indeed, running `docker-compose up --build` starts up the server normally and the `/translationstatus` endpoints seem to work on my machine ¯\\\_(ツ)\_/¯

Or is there also some other context that I'm not aware of?

corradio: indeed good catch! Let me try to merge and deploy this.

PR #1874, opened 2019-05-28
Almost complete translation to Croatian

fbarl:
#### Sections

* [x] `onboarding-modal`
* [x] `theMap`
* [x] `mapAreas`
* [ ] `methodology`
* [x] `data`
* [x] `aboutUs`
* [x] `zoneShortName`
* [x] the rest

---

* [x] Tested in the app

### DIFF
--- a/web/locales/hr.json
+++ b/web/locales/hr.json
@@ -6,12 +6,57 @@
         "datasources": "izvore podataka",
         "contribute": "Daj svoj doprinos <a href=\"%s\" target=\"_blank\">dodavanjem vlastitog područja</a>"
     },
+    "mobile-main-menu": {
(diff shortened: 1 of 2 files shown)


corradio:
@fbarl what is the status here? Should we already merge this and wait for the rest?
cc @martincollignon

fbarl:
> @fbarl what is the status here?

Sorry for slacking off on this one @corradio, it was a much bigger undertaking than I initially thought :)

> Should we already merge this and wait for the rest?

I think so. I just removed some legacy terms and the translation is still above 96% which is fairly good.

image

I also gave it a quick try in the app and it looked generally ok.

Issue #1862, opened 2019-05-20
List of supported languages is duplicated across the code


fbarl:
The locales array is currently being independently defined in 4 different places in the code:

* https://github.com/tmrowco/electricitymap-contrib/blob/40d3a1eaff42f13645fac86893a495cbdc7ef6c3/mobileapp/www/index.html#L5
* https://github.com/tmrowco/electricitymap-contrib/blob/40d3a1eaff42f13645fac86893a495cbdc7ef6c3/mobileapp/generate-index.js#L29
* https://github.com/tmrowco/electricitymap-contrib/blob/40d3a1eaff42f13645fac86893a495cbdc7ef6c3/web/server.js#L33
* https://github.com/tmrowco/electricitymap-contrib/blob/40d3a1eaff42f13645fac86893a495cbdc7ef6c3/web/src/helpers/translation.js#L5

This makes maintaining the list of translations a pain (e.g. adding a language requires changes in all the files above). Instead, the list should be defined at only _one_ place in the code and imported everywhere else where it's needed.

jld3103:
Maybe we should even generate the list from the json files.

corradio:
Still needs to be done for the mobile app

systemcatch:
Quoting from #1897 
> 
> 
> The problems comes from this line:
> https://github.com/jld3103/electricitymap-contrib/blob/8ab390672ead5e92a4108c29fd811f98c8ed4703/web/src/helpers/locales-list.js#L3
> `readdirSync` is not available on the browser.
> I didn't test it but somehow it should have failed when you tested it?
> The ideal solution is to create a webpack transform that injects (i.e. hard codes) the locale list into the source code.
> Else we can also make a script that simply creates a json with the list of locales. That script would have to be run as part of the build process.

PR #1888, opened 2019-06-15
Fix locales list duplicates across code

jld3103:
Fixes: #1862.
I don't know how to load the generated js file into the index.html of the cordova app. Maybe someone knows how to do it?
https://github.com/tmrowco/electricitymap-contrib/blob/40d3a1eaff42f13645fac86893a495cbdc7ef6c3/mobileapp/www/index.html#L5


jld3103:
Locally I'm not able to reproduce the error. In which way does the CI testing differ from local testing? @corradio

web/generate-locales-list.js
```diff
@@ -0,0 +1,19 @@
...
+const files = fs.readdirSync('./locales')
+  .filter(file => file.endsWith('.json'));
+
+const locales = files.map(file => file.replace('.json', ''));
+
+const localesString = JSON.stringify(locales)
+  .replace(/"/g, '\'')
+  .replace(/,/g, ', ');
+
+fs.writeFileSync('./locales/locales-list.js', `module.exports = ${localesString};\n`);
```
corradio: it would make more sense to store it in `src/` (it's not a locale file, so it shouldn't live in locales)

web/src/helpers/translation.js
```diff
@@ -1,6 +1,6 @@
 var exports = module.exports = {};
 
-const LOCALES_LIST = ['ar', 'cs', 'da', 'de', 'en', 'es', 'fr', 'hr', 'it', 'ja', 'nl', 'pl', 'pt-br', 'ru', 'sk', 'sv', 'zh-cn', 'zh-hk', 'zh-tw'];
+const LOCALES_LIST = require('../../locales/locales-list');
```
corradio: CI Error:
ModuleNotFoundError: Module not found: Error: Can't resolve '../../locales/locales-list' in '/home/web/src/helpers'

corradio:
Thanks for the PR. Super useful. I'll help on making it work.

web/src/helpers/translation.js
jld3103: Maybe it's, because it's outside of src?

corradio: I think that's the reason indeed

jld3103: I know what the issue is🤦

jld3103: The main issue was, that the transpiler can't use the file generated in the other container. I don't know if it also was, because it was outside of src.

corradio:
Great job!

jld3103:
Yes, but I think the issue should be updated, because I wasn't able to load the file in the mobile app.

jld3103:
Ping @corradio

corradio:
> Yes, but I think the issue should be updated, because I wasn't able to load the file in the mobile app.

what do you mean exactly? Did you encounter a compilation error or did you try to run the mobile app?
Note there's an issue with languages in the mobile app, as the mobile app doesn't respond to ?lang=<locale> querystring, but instead should be redirected to `/index_<locale>.html` as you can see here:
https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/www/index.html#L10-L13

Is that the issue you are referring to?

jld3103:
No. Didn't know how to load the file. So we need to do that. I've not worked with Cordova a lot.

corradio:
Indeed it's not very easy to load the file from here:
https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/www/index.html#L5
We'd need to change a bit the architecture of the mobile app.
I suggest merging this and updating our previous issue to mark that this is remaining to be done.
If I find a bit of time I might work on that.

Olivier

jld3103:
That is exactly what I think :)

PR #1893, opened 2019-06-24
Update AT capacities

scriptator:
I updated the production capacities for Austria because I noticed that the solar generation sometimes exceeded the stated capacity.

The aggregation of values has been done by using the mapping defined in `parsers.ENTSOE.ENTSOE_PARAMETER_GROUPS`. That is
    
    biomass -> "Biomass" + "Waste"
    coal -> "Fossil Brown coal/Lignite" + "Fossil Hard coal" + "Fossil Oil shale" + "Fossil Peat"
    gas -> "Fossil Coal-derived gas" + "Fossil Gas"
    geothermal -> "Geothermal"
    hydro -> "Hydro Run-of-river and poundage" + "Hydro Water Reservoir"
    nuclear -> "Nuclear"
    oil -> "Fossil Oil"
    solar -> "Solar"
    wind -> "Wind Offshore" + "Wind Onshore"
    unknown -> "Other" + "Marine" + "Other renewable"
    hydro storage -> "Hydro Pumped Storage"

To ease further updates I have written a script which is capable of performing this capacity update to the file `zones.json` automatically, given the corresponding CSV export file from ENTSOE. You can find the script [here](https://github.com/scriptator/electricitymap-contrib/blob/entsoe_capacity_parsing/utils/update_capacities_entsoe.py ). If you consider such a script a useful contribution and would like to have it in this repo, I'll open a separate pull request for that.

PR #1895, opened 2019-06-26
Automatic ENTSOE capacity updates

scriptator:
As already announced in #1893 I created a script which allows for updating production capacities in the file `zones.json` by using CSV files exported from the [ENTSOE page for installed capacities](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show).

The aggregation of values is done by using the mapping defined in parsers.ENTSOE.ENTSOE_PARAMETER_GROUPS. At the moment that is:

    biomass -> "Biomass" + "Waste"
    coal -> "Fossil Brown coal/Lignite" + "Fossil Hard coal" + "Fossil Oil shale" + "Fossil Peat"
    gas -> "Fossil Coal-derived gas" + "Fossil Gas"
    geothermal -> "Geothermal"
    hydro -> "Hydro Run-of-river and poundage" + "Hydro Water Reservoir"
    nuclear -> "Nuclear"
    oil -> "Fossil Oil"
    solar -> "Solar"
    wind -> "Wind Offshore" + "Wind Onshore"
    unknown -> "Other" + "Marine" + "Other renewable"
    hydro storage -> "Hydro Pumped Storage"

The mapping is not hard coded, though, but loaded directly from the parser file. That means if the names or the assignment change somehow the behaviour of my script will change as well.

For me, the following questions are still open:

- Is there a possibility to download data for multiple countries from ENTSOE simulatneously or is there even an API which I could query to fetch the capacity data? In this case it would be very easy to extend the script to handle multiple zones in the same run.
- Does the location of the file make sense where it is?
- The script depends on pandas for CSV parsing because it's so convenient. If this dependency is a problem for you it can be easily converted to use python's built-in csv parser.

I am happy to hear your feedback! :-)

### DIFF
--- a/utils/ENTSOE_capacity_update.py
+++ b/utils/ENTSOE_capacity_update.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+import os
+import sys
 import argparse
 import pandas as pd
 import json
@@ -14,6 +16,9 @@ def update_zone(zone, data, zonesfile):
 	with open(zonesfile) as zf:
 		zones = json.load(zf)
 
+	if zone not in zones:
+		raise ValueError(f"Zone {zone} does not exist in the zonesfile")
+
 	zones[zone]["capacity"].update(data)
 
 	with open(zonesfile, "w") as zf:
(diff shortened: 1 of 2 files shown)


corradio:
Terrific! There is an API to fetch capacities for ENTSOE (see https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html) but as a first try this is very good.
Location of the file looks good as well.
Thanks!

scriptator:
> There is an API to fetch capacities for ENTSOE (see https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html)

Good to know, thanks! Maybe I'll extend the script to directly query the data from ENTSOE sometime.

PR #1896, opened 2019-06-27
Fix CA-AB dataframe handling

systemcatch:
This is an another attempt to fix the problem with CA-AB (see #1845 before). The parser will try each df found rather than trying by index. However I'm not that happy with this solution as I still don't get why read_html is matching 2 dfs rather than just one.

ref #1833


jarek:
Sorry, I currently don't have a lot of time to check in depth on what is causing the problem. If it works decently reliably, go for it :+1: - can't be worse than not having data at all

PR #1898, opened 2019-07-12
Automatic ENTSOE capacity update: Parse from API

scriptator:
This is a follow-up of #1895 which adds possibility to parse the installed generation capacities directly from the ENTSOE API. Instead of downloading and passing CSV files, you only need to pass the API token. For example, you could update DE by executing:

    python3 -m utils.ENTSOE_capacity_update DE --api-token <TOKEN>

I am wondering, however, whether the data from ENTSOE is really correct. I tried several countries and got always quite different values than those which are currently in the `zones.json`. Are those values outdated or is the data which my script produces wrong?

Some examples for changes:

    Germany:
    "biomass": 7740 -> 9150
    "coal": 44910 -> 46498
    "gas": 29390 -> 31664
    "geothermal": 38 -> 42
    "hydro": 4800 -> 5282
    "hydro storage": 9600 -> 9422
    "nuclear": 9516 (unchanged)
    "oil": 4300 -> 4356
    "solar": 47200 -> 43269
    "unknown": 3137 -> 1860
    "wind": 59830 -> 58803


Gazer75:
I wouldn't trust the capacity from ETSOE. For Norway it is not accurate. I believe there is a cutoff on capacity that is required to be reported so many smaller plants are not included. I believe most of the plants with <10MW are missing.

scriptator:
>I believe most of the plants with <10MW are missing.

That may be, but it does not explain e.g. the increased amount of gas or coal capacity.

The changes to biomass look especially strange to me but I suspect the fault in this case rather in the current data. I believe that Germany's current numbers for biomass are not including the numbers for waste, as it is defined in the ENTSOE parser (which sounds questionable to me, but that's a different issue). That could explain the shift away from Unknown to Biomass, as ENTSOE data reports waste generation with 1686 MW.

systemcatch:
@corradio I'm away on holiday from tomorrow, happy to review this when I get back (28th). If you need it sooner I'd pass it to @jarek or @alixunderplatz

systemcatch:
hey @scriptator just had a quick look over this, I think something is wrong with the indentation (double tabs?).

systemcatch:
It might also be nice to have a way to update all ENTSOE zones at once?

scriptator:
> hey @scriptator just had a quick look over this, I think something is wrong with the indentation (double tabs?).

I don't really know what you mean. Can you point me to the exact position where you see the problem?


> It might also be nice to have a way to update all ENTSOE zones at once?

Good idea, that's easy to add to the script. However, I already heard that for some countries the data from ENTSOE is not accurate - in this case it wouldn't make sense to do an automatic update. How can we collect a list of zones which are suitable for performing automatic updates?

utils/ENTSOE_capacity_update.py
```diff
@@ -6,78 +6,131 @@
...
+        return aggregated
 
 
 def parse_args():
-	parser = argparse.ArgumentParser()
-	parser.add_argument("--zonesfile", default=ZONESFILE)
-	parser.add_argument("zone", help="The zone abbreviation (e.g. AT)")
-	parser.add_argument("data_file", help="The csv file from ENTSOE containing the installed capacities")
-	return parser.parse_args()
+        parser = argparse.ArgumentParser()
```
systemcatch: It's everywhere in this file, 8 space indentation rather than the original 4.

systemcatch:
> Good idea, that's easy to add to the script. However, I already heard that for some countries the data from ENTSOE is not accurate - in this case it wouldn't make sense to do an automatic update. How can we collect a list of zones which are suitable for performing automatic updates?

Hmmm well the domain mappings are already imported in this script so maybe use them to generate a list of zones when `all` is passed. Possibly we could use a blacklist approach with zones that shouldn't be automatically updated (I'm not up to date with which zones are problematic though).

utils/ENTSOE_capacity_update.py
scriptator: Indeed, I did this with a different computer - must have had a different setting there. I converted it to four spaces now.

scriptator:
> 
> 
> Hey @scriptator I kinda forgot about this, if you want to add the `all` option we talked about then go ahead otherwise I'll merge this in.

Hey, sorry I also forgot about this. Let's do the "--all" option if somebody misses it

PR #1902, opened 2019-07-18
Add US-SEC parser for Seminole Electric Cooperative in Florida

snarfed:
combines gas + coal data from EIA with solar data from apps.seminole.coop. still needs geojson. lots of background and details in #1713. example data:

```py
[{'datetime': datetime.datetime(2019, 7, 14, 23, 0, tzinfo=<UTC>),
  'production': {'solar': 0.6502, 'unknown': None},
  'source': 'eia.org, apps.seminole.coop',
  'storage': {},
  'zoneKey': 'US-SEC'},
  ...
]
```

both data sources usually lag by a day or two, sometimes much more. for example, right now EIA's data stops on 2019-07-09, almost 10d ago. :slightly_frowning_face:

minor quirk: this parser reuses code from both the EIA and ENTSOE parsers, which there isn't much other precedent for so far. i've added a TODO to move `ENTSOE.merge_production_outputs` to `lib.utils`, but the EIA reuse might be ok as is? see what you think.

thanks in advance for reviewing, and for all your support so far!


config/zones.json
```diff
@@ -3708,6 +3708,31 @@
     },
     "timezone": null
   },
+  "US-SEC": {
+    "_comment": "https://www.seminole-electric.com/facilities/generation/ lists two 650MW coal-fired plants, one 810MW gas plant, and a small 2.2MW solar farm. Hwoever, EIA combines coal and gas generation data, so we report those together as unknown",
```
corradio: can you put this comment in the parser itself? This is the best place to put this IMO

parsers/EIA.py
```diff
@@ -32,16 +36,28 @@
...
+
+def fetch_production(zone_key, session=None, target_datetime=None,
+                     logger=None):
+    return _fetch_production_or_consumption(
+        zone_key, PRODUCTION[zone_key], session=session,
+        target_datetime=target_datetime, logger=logger)
+
 
-    series_id = DAY_AHEAD[zone_key]
+def _fetch_production_or_consumption(zone_key, series_id, session=None,
```
corradio: how do we know if it's a production or consumption?

parsers/ENTSOE.py
```diff
@@ -857,6 +857,8 @@ def fetch_production(zone_key, session=None, target_datetime=None,
 }
 
 
+# TODO: generalize and move to lib.utils so other parsers can reuse it. (it's
```
corradio: 👍

config/zones.json
```diff
@@ -3708,6 +3708,31 @@
     },
     "timezone": null
   },
+  "US-SEC": {
```
corradio: do we already have a geometry for this?

snarfed: sadly, no. that's my next project!

config/zones.json
snarfed: good call! done.

parsers/EIA.py
snarfed: good question. the series id, specifically `DF` for consumption forecast vs `NG` for production. i've added a docstring.

config/zones.json
corradio: ok, then let's wait to merge this PR for when the geometry is ready

config/zones.json
corradio: great! Then I think you can remove it from here.

config/zones.json
snarfed: sure. i may need a bit of help, though. details in https://github.com/tmrowco/electricitymap-contrib/issues/1713#issuecomment-513506553 . if you know who i should talk to, please point me their way!

snarfed:
ok, this is close to ready for re-review! i've added geometry based on roughly half of Florida's counties. screenshot below; discussion in https://github.com/tmrowco/electricitymap-contrib/issues/1713#issuecomment-517704023 and onward.

two possible TODOs before merging this ([details](https://github.com/tmrowco/electricitymap-contrib/issues/1713#issuecomment-518012542)):
* merge county boundary lines so they're not visible
* fix bug in `topogen.sh` so that it correctly persists and caches `US-FL.geojson` instead of downloading again every time

image

systemcatch:
Hey @snarfed you don't need to do a check for the file. Just put it in https://github.com/tmrowco/electricitymap-contrib/tree/master/web/third_party_maps that way it will get copied with all the other shapes on https://github.com/tmrowco/electricitymap-contrib/blob/master/web/topogen.sh#L28

snarfed:
@systemcatch sure! agreed, that'd be easier. i didn't consider that originally because it's 3MB though. is that still ok to go into `third_party_maps`?

alternatively i could just check in the final 4-5KB of polygons, after compressing and quantizing to the `world.json` grid, but that would lose the provenance and steps to regenerate from the original Florida counties data.

snarfed:
oh, i see that the two NACIS zip files in `web/build/` are bigger, and they're checked in. ok then! will do.

web/generate-geometries.js
```diff
@@ -593,6 +606,7 @@ const zoneDefinitions = [
   { zoneName: 'US-PJM', type: 'states', countryId: 'USA', states: [
     'US.PA', 'US.NJ', 'US.MD', 'US.DE', 'US.VA', 'US.WV', 'US.OH', 'US.KY']},
   // { zoneName: 'US-RI', countryId: 'USA', stateId: 'US.RI', type: 'state' },
+  { zoneName: 'US-SEC', type: 'county', counties: ['ALACHUA', 'BAKER', 'BRADFORD', 'CITRUS', 'CLAY', 'COLUMBIA', 'DESOTO', 'DIXIE', 'GADSDEN', 'GILCHRIST', 'GLADES', 'HAMILTON', 'HARDEE', 'HENDRY', 'HERNANDO', 'HIGHLANDS', 'JEFFERSON', 'LAFAYETTE', 'LAKE', 'LEON', 'LEVY', 'LIBERTY', 'MADISON', 'MANATEE', 'OKEECHOBEE', 'OSCEOLA', 'PASCO', 'PUTNAM', 'SARASOTA', 'SUMTER', 'SUWANNEE', 'TAYLOR', 'UNION', 'WAKULLA']},
```
corradio: can you put this line after `US-SC` so it is ordered alphabetically?

web/generate-geometries.js
```diff
@@ -750,9 +767,14 @@ zoneFeatures = toListOfFeatures(zoneFeaturesInline);
 // Write unsimplified list of geojson, without state merges
 fs.writeFileSync('public/dist/zonegeometries.json', zoneFeatures.map(JSON.stringify).join('\n'));
 
-// Simplify all countries
+// Convert to TopoJSON
 const topojson = require('topojson');
 let topo = topojson.topology(zones);
+
+// merge Florida counties in US-SEC
```
corradio: maybe add a more explicit description that points to the fact that we do this to avoid having inner lines?

snarfed: agreed, good call, done.

web/generate-geometries.js
snarfed: oops, yes, done.

corradio:
Let's see if it works!

PR #1903, opened 2019-07-20
Fix electricity sources for IN_DL

anirudh-ramesh:
The data source for New Delhi in India has changed to reflect the shutting down of the Badarpur Coal Thermal Power Station (https://timesofindia.indiatimes.com/city/delhi/badarpur-thermal-plant-shut-for-good/articleshow/66228551.cms) in October 2018.

The parser had been designed to read the rows of an HTML table and assign data in sequential order. Following the change in the upstream data source, that sequence has changed as well. Therefore, the parser has been incorrectly assigning another plant's electricity generation data to Badarpur's. This means that since the time the dashboard had changed - quite a few months, at least - the parser has been malfunctioning.

This commit resolves this bug by using a library of names of power plants and the fuel source used in each. The electricity generation data is only assigned if the name of the plant from the HTML row matches the one in the library.

### DIFF
--- a/parsers/IN_DL.py
+++ b/parsers/IN_DL.py
@@ -5,6 +5,14 @@ from .lib import web
 from .lib import zonekey
 from .lib import IN
 
+plants = {
+    "CCGT-Bawana": "Gas",
+    "DMSWSL-Dsidc": "G2E",
+    "EDWPL-Gazipur": "G2E",
+    "GT": "Gas",
+    "Pragati": "Gas",
+    "TOWMP-Okhla": "G2E"
+}
 
 def fetch_consumption(zone_key='IN-DL', session=None, target_datetime=None, logger=None):
     """Fetch Delhi consumption"""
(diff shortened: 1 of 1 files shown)


corradio:
Hello and thanks for the PR! Can you send the output of this parser so we can double check everything is at it should? Thanks

anirudh-ramesh:
Hi, @corradio!

The parser's output (for production) is as follows.

`{'zoneKey': 'IN-DL', 'source': 'delhisldc.org', 'production': {'coal': 0, 'gas': 687.0, 'biomass': 35.0}, 'datetime': datetime.datetime(2019, 7, 22, 18, 6, 28, tzinfo=tzfile('/usr/share/zoneinfo/Asia/Kolkata'))}`

The following is a screenshot of the real-time dashboard at the same instant.

production

Over here, "CCGT-Bawana", "GT" and "Pragati" are powered by natural gas, while the others are powered by municipal solid waste.

PR #1910, opened 2019-07-26
RU.py - Reduced amount of requests

FelixDQ:
fixes tmrowco/electricitymap#1437

Now only fetches data from the last two hours when fetching live data.


corradio:
@FelixDQ I'll let you merge

PR #1923, opened 2019-08-05
Search for "CO2" globally in text strings

magol:


corradio:
Didn't it work previously?

magol:
In one FAQ text, it was two CO2, but only the first was replaced

Issue #1674, opened 2018-11-15
North and South Carolina States


corradio:
![image](https://user-images.githubusercontent.com/1655848/48536114-dd9f7000-e8ad-11e8-8a5b-12221706e622.png)
https://www.eia.gov/state/?sid=NC#tabs-4
![image](https://user-images.githubusercontent.com/1655848/48536132-ec862280-e8ad-11e8-8369-e4833fdb94c6.png)
https://www.eia.gov/state/?sid=SC#tabs-4

Given https://api.watttime.org/:
- NC would be Duke Energy
- SC would be South Carolina Electric & Gas Co and South Carolina Public Service Authority

Real-time generation + imports available here (but no mix):
https://www.eia.gov/realtime_grid/

corradio:
Closing due to inactivity. Thank you all for your inputs!

PR #1934, opened 2019-08-14
Add parser for US (South Carolina)

snarfed:
for #1674. historical production only right now, via EIA. combines SC (Santee Cooper) and SCEG (South Carolina Electric & Gas). good capacity data, but no mix, sadly, so everything is currently reported as unknown.

next to do, consumption.

i might also look into extending EIA.py to merge multiple keys itself, so that we can use it in place for multi-key zones like this.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -4050,6 +4050,14 @@
     "timezone": null
   },
   "US-SC": {
+    "_comment": "Production comes from two sources. 1) South Carolina Electric & Gas Co, a division of SCANA (https://www.scana.com/), acquired by Dominion 2) Santee Cooper (http://www.santeecooper.com/) aka South Carolina Public Service Authority. Capacity from http://www.energy.sc.gov/files/view/Santee%20Cooper_IRP_2018_FINAL.pdf averaged between summer and winter reported capacities.",
+    "capacity": {
+      "hydro": 962,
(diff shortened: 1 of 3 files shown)


systemcatch:
Hey @snarfed thanks for the contribution, I'll take a look at this over the weekend.

snarfed:
thanks! no hurry. I'm actually headed out on vacation today until 8/26, without a computer, so i'd need to wait until then to make any changes anyway.

systemcatch:
So the code itself is fine, my question to you @corradio is does historical production that is completely unknown have much value to the map?

snarfed:
friendly ping! do we have an answer to [systemcatch's question](https://github.com/tmrowco/electricitymap-contrib/pull/1934#issuecomment-522489920)? or is this PR now obsolete due to #1941?

i'm ok with any answer, just checking in.

corradio:
Sorry for the late reply, just came back from vacation.

> So the code itself is fine, my question to you @corradio is does historical production that is completely unknown have much value to the map?

Not really to be honest. However, with the new EIA parser capabilities, wouldn't this be obsolete?

systemcatch:
@corradio yes with #1941 this is now obsolete. So I think we can close this, thanks anyway @snarfed.

PR #1939, opened 2019-08-18
Create new historical production parser for Chile

systemcatch:
TODO
- [x] Docstrings
- [x] Configs (zone, locales)
- [x] Readme
- [x] New world.json
- [x] Check and bump any requirements needed
- [x] Mark "dead" parsers

ref #655


config/zones.json
```diff
@@ -800,70 +800,38 @@
     },
     "timezone": null
   },
-  "CL-SIC": {
+  "CL-SING-SIC": {
```
corradio: from a quick Google search it seems like this is now merged into a zone called "SING". I might be wrong but I think SIC doesn't exist anymore.

parsers/CL_SIC.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
-"""Parser for the SIC grid region of Chile."""
+"""Parser for the SIC grid region of Chile.
```
corradio: let's just delete that file

parsers/US_SVERI.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
-"""Parser for the SVERI area of the USA."""
+"""Parser for the SVERI area of the USA.
```
corradio: same here

web/locales/es.json
```diff
@@ -320,13 +320,9 @@
             "countryName": "Chile",
             "zoneName": "SEM"
         },
-        "CL-SIC": {
+        "CL-SING-SIC": {
```
corradio: good call 👍

systemcatch:
> Thank you for keeping this alive :)

Not a problem, it's good to have you back Oli.

PR #1940, opened 2019-08-21
Added parser for IN-KL

vigeesh:
Added parser for Kerala, India. The data is parsed from sldckerala.com


alixunderplatz:
@vigeesh Looks very promising, good job! One question: can you post a link to the data page from where you get the data? thanks!

vigeesh:
http://sldckerala.com/index.php?id=1

alixunderplatz:
@vigeesh thanks! is it hourly generation data or a daily sum? we need at least hourly data to display production and emissions on the map. unfortunately, processing cumulative daipy data is not possible at the moment.

vigeesh:
Yes, it is a daily sum. I couldn't find a source for the hourly data.

corradio:
Hi @vigeesh ,

Thank you for your contribution. Unfortunately we can only display hourly data at the moment. I will close this issue but please keep your code and let us know if hourly data pops up - it would be amazing to be able to add it ;-)

Olivier

PR #1944, opened 2019-08-29
remove the US-SEC zone from US-FL so they don't overlap

snarfed:
for #1713. thanks to @systemcatch for the drastically simplifying suggestion in https://github.com/tmrowco/electricitymap-contrib/issues/1713#issuecomment-526120874!

not a big deal, but if/when you all merge, maybe consider preserving the individual commits, ie merge or rebase instead of squashing. d239a74ac65f3fbbc904f89c6289f2d83a275740 in particular is a good example of how to subtract one arbitrary set of Topo/GeoJSON features from another, with `polygon-clipping.difference()`. basically:

```js
const clipping = require('polygon-clipping');
zones['US-FL'] = ... clipping.difference(
    zones['US-FL'].geometry.coordinates,
    zones['US-SEC'].geometry.coordinates)
```


corradio:
Hi @snarfed , can you post a screenshot just so we can double check?
Thanks for this!

snarfed:
oops, yes! sorry, should have done that originally.

image

Issue #1442, opened 2018-06-07
Missing generation in the Netherlands


matteodefelice:
I am aware that there are already issues with the information about generation in the Netherlands  ([https://github.com/tmrowco/electricitymap/issues/78](url)) but I was trying to understand where the data is currently obtained.
For example right now in the ENTSO-E TP we can see a very low generation (200 MW, a screenshot [here](https://www.dropbox.com/s/2rhyza1rihx8j3l/Pasted_Image_07_06_2018__12_26.png?dl=0)) , in the electricity map there is a value of about 18 GW while in the TenneT Dashboard (https://www.tennet.eu/electricity-market/data-dashboard/generation/) is lower, around 14 GW. 
What am I missing here?

alixunderplatz:
@matteodefelice 

Generation is calculated as mentioned in this pull request:
https://github.com/tmrowco/electricitymap/pull/1303

> * Queries ENTSOE to get total load, total net imports, and from those estimate total generation
> * Queries ENTSOE to get breakdown of production, setting fossil to None and calculating unknown as the difference between the measured production breakdown and total generation

matteodefelice:
Thank you for the information, I can see that this estimation tends to be higher than the actual values provided by TenneT especially during the evening, for example yesterday in the period 18:00-23:00 the generation on the electricity map was about 20 GW while TenneT reports about 5 GW less. In the same period the ENTSO-E TP reported a load of 15.5-16.5 GW and net import of about 2 GW...

PaulCornelissen:
Something seems to be going wrong:
<img width="854" alt="screenshot 2018-06-15 14 05 27" src="https://user-images.githubusercontent.com/5185181/41467404-b1c9b3ca-70a6-11e8-87a4-b879d2263494.png">

<img width="542" alt="screenshot 2018-06-15 14 05 56" src="https://user-images.githubusercontent.com/5185181/41467415-ba260744-70a6-11e8-95b6-855d8e7700a7.png">

I overlayed these matching the scales:
<img width="642" alt="screenshot 2018-06-15 14 05 56 overlay" src="https://user-images.githubusercontent.com/5185181/41467445-d8218d04-70a6-11e8-9f19-d206eb207489.png">

The underlaying graph doesn't show exports, but especially between 1am and 4am the increased error correlates with an increased import from Germany and decreased exports to UK and BE.

corradio:
[...]

Issue #1949, opened 2019-09-03
Add DK-DK1->NL new interconnector


corradio:
It is available from ENTSOE (although it shows no data obviously as it hasn't been put in operation).
It'd be great to have it ready when it opens!
cc @martincollignon

PaulCornelissen:
I think this one can be closed:
<img width="700" alt="image" src="https://user-images.githubusercontent.com/5185181/64697361-fadaa600-d4a0-11e9-87ef-1eb7a1fc6cea.png">

corradio:
Great!

PR #1955, opened 2019-09-09
Fix NL parser; imports were reversed and added new DK1 interconnector.

PaulCornelissen:
Line 82 fixes issue #1442. The problem was that imports and exports were reversed during calculation of production from the consumption figures.

Also I've added the missing NL-DK1 interconnector to the calculation of production. #1949

### DIFF
--- a/parsers/NL.py
+++ b/parsers/NL.py
@@ -4,6 +4,7 @@ import arrow
 
 from . import statnett
 from . import ENTSOE
+from . import DK
 import logging
 import pandas as pd
 import requests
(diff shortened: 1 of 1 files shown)


parsers/NL.py
```diff
@@ -43,8 +44,11 @@ def fetch_production(zone_key='NL', session=None, target_datetime=None,
         exchanges.extend(exchange or [])
 
     # add NO data, fetch once for every hour
+    # This introduces an error, because it doesn't use the average power flow
```
corradio: well spotted

Issue #1950, opened 2019-09-03
Croatia has no data, Slovenia lacks SI-HR data, but both is available.


seba1337:
I've found sources for both.

SI-HR data is available here https://www.eles.si/trzni-podatki, just seems you would need some work.

But even better, there is HR data, https://www.hops.hr/Home/PowerExchange in JSON, which gives also the missing SI-HR data and other interconnect data (SI = Slovenija, HU = Mađarska, RS+ME = Srbija i Crna Gora, BA = Bosna i Hercegovina) as well as total electricity production (Ukupna proizvodnja) and amount of electricity production by wind power (Proizvodnja VE).

I've found an old issue https://github.com/tmrowco/electricitymap-contrib/issues/1007, but seems that it wasn't realised that there is JSON data available, so no need for a complex parser. 

Citing wikipedia:
> Total installed capacity of generating objects built in Croatia amounts to 3745 MW, 2079 MW of which is hydropower's share, and 1666 MW comes from thermal power plants. There are 25 hydroelectric power plants, most of which are located on the coast (including the single pumped storage hydroelectric plant, RHE Velebit), and 7 thermal power stations, 3 of which also generate heat for industry and heating in cities. Also, 338 MW is available from the co-owned Krško Nuclear Power Plant, and 210 MW from the privately owned Plomin 2 thermal power plant

Seems that that's old data. Not having wind power, but we also see that a large share of installed power is hydro power of which there's no(?) public data. :'(

Croatian wikipedia has fresher data, there seems 570 MW wind power (2018), source: https://hr.wikipedia.org/wiki/Vjetroelektrane_u_Hrvatskoj. 

Seeing this site about alternative energy, http://www.poslovni.hr/hrvatska/u-hrvatskoj-je-u-pogonu-1207-elektrana-na-oie-koliko-nam-je-preostalo-do-ispunjenja-zadanih-ciljeva-do-2020-300342#, seems 40,4 MW solar, 15,9 MW biogas, 14,9 MW biomass, 5,5 MW landfill and waste water gas, 0 MW geothermal. The data on hydro is on microhydro, so not relevant. 

The data on hydroplants in the first citation (2079 MW) seems approx. correct.

Thermal power plants changed too,
* 303 MW oil (TE Rijeka, is in reserve, but used in peaks) 
* 199 MW coal (TE Plomin, block A is taken offline)
* 78 MW gas (KTE Jertovec)
* 300 MW gas (TE-TO Zagreb)
[...]

PR #1959, opened 2019-09-17
Croatia

PaulCornelissen:
- Parse exchanges from ENTSOE
- Configured zones.json to use ENTSOE parsers for consumption and consumptionForecast
- Wrote parser for HOPS json output according to issue #1950
- Add ENTSOE capacity figures to zones.json
- Took the liberty to add myself as contributor

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1698,6 +1698,25 @@
         46.5037509222
       ]
     ],
+    "capacity": {
+      "biomass": 70,
+      "coal": 332,
+      "gas": 743,
+      "hydro": 1857,
+      "hydro storage": 283,
+      "oil": 937,
+      "solar": 53,
+      "wind": 630,
+      "unknown": 67
+    },
+    "contributors": [
+      "https://github.com/PaulCornelissen"
+    ],
+    "parsers": {
+      "consumption": "ENTSOE.fetch_consumption",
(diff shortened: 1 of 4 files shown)


config/zones.json
```diff
@@ -2754,7 +2773,8 @@
       "wind": 4626
     },
     "contributors": [
-      "https://github.com/corradio"
+      "https://github.com/corradio",
```
corradio: are you sure you didn't mean to update the key "HR"?

parsers/NL.py
```diff
@@ -175,7 +175,7 @@ def get_production_data_energieopwek(date, session=None):
     # The API returns values per day from local time midnight until the last
     # round 10 minutes if the requested date is today or for the entire day if
     # it's in the past. 'sid' can be anything.
-    url = 'http://energieopwek.nl/jsonData.php?sid=2ecde3&Day=&Day=%s' % date.format('YYYY-MM-DD')
+    url = 'http://energieopwek.nl/jsonData.php?sid=2ecde3&Day=%s' % date.format('YYYY-MM-DD')
```
corradio: 👍

config/zones.json
PaulCornelissen: I meant both if that's ok? What's the criteria?

corradio: That's completely fine. But don't you also want to add yourself to CR as you update the CR production parser?

PR #1961, opened 2019-09-18
Update zones.json

seba1337:


config/zones.json
```diff
@@ -3462,9 +3462,12 @@
       "gas": 491,
       "hydro": 1233,
       "nuclear": 696,
+      "hydro storage": 180,
       "oil": 0,
       "solar": 263,
-      "wind": 3
+      "wind": 3,
+      "battery storage": 0,
+      "geothermal": 0,
```
corradio: there shouldn't be a comma at the end of this line

seba1337: OK I think I did it.

Issue #1871, opened 2019-05-28
NL renewable power estimates


dguzs:
Hello,

electricityMap lacks info on renewable power sources for the Netherlands. I suggest this information can be taken from http://energieopwek.nl/

BR
Dmitrijs Guzs

MartienVisser:
Thanks, as I'm the author of Energieopwek.nl :).

Actually, the Dutch CO2-emission per kWh is much lower than currently suggested. I think it is not too difficult to improve the data on electricitymap.org substantially using the information below:

1) currently, about 20% of electricity is generated by renewable sources (solar: 5%, biomass: 6%, onshore wind: 6% and offshore wind: 3%.
2) the Dutch coal fired power stations are new and have high efficiency (CO2 = 700 gram/kWh); moreover, they use some co-firing of biomass (see Energieopwek.nl); this is currently around 425 MW, thus 10% of the total coal-fired capacity in the NL.
3) gas-fired power in the Netherlands is generated by a mixture of cogeneration (300 gram/kWh) and high efficiency gas-fired power stations (<400 gram/kWh).

Real-time data of Dutch power stations is not publicly available due to regulation (coal-fired) and because Dutch cogen plants use a significant portion of their electricity production themselves; thus, the DSOs and TSOs don't see it. I do not expect a quick solution for this.

However, a good approximation could be using a fixed volume of 20% renewable, adding coal (90% of capacity at 700 gr/kWh) and the rest by gas (50.50 mixture STEG & Cogen) at average of 350 gr/kWh). 

If required, variable (solar and wind) in the Netherlands could be modelled using energieopwek.nl, or, maybe simpler, using the average utilisation factor of Belgium, UK and Germany.  

By the way, my estimate/calculated specific CO2 of Dutch electricity generation in April 2019 was about 375 gram/kWh. I publish this information monthly (in English) at the website of Entrance: https://www.en-tran-ce.org/onderzoek/duurzame-energie-in-nederland/

Best regards,

Martien Visser

jarek:
See also past and more recent comments about Netherlands data in #78

To confirm, @MartienVisser - is the data of renewable production on Energieopwek.nl official or estimated?

MartienVisser:
The data is calculated using models which are based on wind speed and solar radiation data from about 40 weather stations distributed over the Netherlands and offshore. The results have been calibrated against monthly (wind) and annual (solar) data from the national statistical office CBS.

dguzs:
[...]

PR #1964, opened 2019-09-20
Fixes to NL parser and adding pricing to Croatia

PaulCornelissen:
Fixes a bug in the NL parser. See also issue #1871 
Adds parsing of pricing for Croatia (issue #1959)


corradio:
let me know when it's ready to merge (want to be sure before merging)

PaulCornelissen:
This can be merged.

(It fixes a significant bug in the NL parser as explained ing #1871 and makes the output much better, but there's still another source of error that I haven't found yet.)

corradio:
you rock tx

PR #1965, opened 2019-09-21
Update co2eq_parameters.json for Croatia's unknown

seba1337:
https://www.hops.hr/page-file/oEvvKj779KAhmQg10Gezt2/temeljni-podaci/Temeljni%20podaci%202018.pdf

By suggestion from https://github.com/tmrowco/electricitymap-contrib/issues/1950#issuecomment-532624874
I've tried to make an estimate.
For 2018 it's 6691 GWh hydro, 3206 GWh thermo, 1240 GWh wind. So unknown is 6691 GWh + 3206 GWh = 9897 GWh. Hydro taking 68% of the unknown. Then I did it like this 24*0.68+820*(1-0.68), where 24 represents the CO2 from hydro and 820 the CO2 from coal and 68 the share of hydro from the PDF reference.

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -319,7 +319,7 @@
       "HR": {
         "unknown": {
           "source": "2018 average estimated from https://www.hops.hr/page-file/oEvvKj779KAhmQg10Gezt2/temeljni-podaci/Temeljni%20podaci%202018.pdf",
-          "value": 279
+          "value": 240
         }
       },
       "HU": {
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -316,6 +316,12 @@
           "value": 178.67099686684926
         }
       },
+      "HR": {
+        "unknown": {
+          "source": "2018 average estimated from https://www.hops.hr/page-file/oEvvKj779KAhmQg10Gezt2/temeljni-podaci/Temeljni%20podaci%202018.pdf",
+          "value": 279
+        }
+      },
       "HU": {
         "hydro discharge": {
           "source": "2017 average by Tomorrow",


PaulCornelissen:
That's a much better estimate I think. It's still a bit of a worst case though I think, because earlier you wrote:
> Thermal power plants changed too,
> 
> 303 MW oil (TE Rijeka, is in reserve, but used in peaks)
> 199 MW coal (TE Plomin, block A is taken offline)
> 78 MW gas (KTE Jertovec)
> 300 MW gas (TE-TO Zagreb)
> 50 MW gas (EL-TO Zagreb)
> 89 MW gas (TE-TO Osijek)
> 229 MW gas (TE-TO Sisak)

200MW coal, so even if it ran at 100% capacity factor that would only generate 1750GWh. So there must be a significant share of Gas in there. In the first post you also wrote about a co-owned nuclear plant with ~300MW capacity? Is that plant feeding directly into the Croatian grid?

If we can't find more accurate numbers I suggest to at least use the unknown CO2 intensity of 700gr/kWh for the remaining non hydro generation and not assume it's all coal. That would yield a CO2 intensity of 240gr/kWh.

seba1337:
> 
> 
> That's a much better estimate I think. It's still a bit of a worst case though I think, because earlier you wrote:
> 
> > Thermal power plants changed too,
> > 303 MW oil (TE Rijeka, is in reserve, but used in peaks)
> > 199 MW coal (TE Plomin, block A is taken offline)
> > 78 MW gas (KTE Jertovec)
> > 300 MW gas (TE-TO Zagreb)
> > 50 MW gas (EL-TO Zagreb)
> > 89 MW gas (TE-TO Osijek)
> > 229 MW gas (TE-TO Sisak)
> 
> 200MW coal, so even if it ran at 100% capacity factor that would only generate 1750GWh. So there must be a significant share of Gas in there. In the first post you also wrote about a co-owned nuclear plant with ~300MW capacity? Is that plant feeding directly into the Croatian grid?
> 
> If we can't find more accurate numbers I suggest to at least use the unknown CO2 intensity of 700gr/kWh for the remaining non hydro generation and not assume it's all coal. That would yield a CO2 intensity of 240gr/kWh.

The co-owned plant is in Slovenia and is counted already in Slovenia (whole capacity).

That's a good idea I guess, using 700 g/kWh for the remaining part. I've found also data for solar, which we could add and further refine the data https://github.com/tmrowco/electricitymap-contrib/issues/1966 :-)

Issue #1841, opened 2019-05-02
Language select not responsive


jld3103:
The language select should be display full screen on mobile devices.

corradio:
Also, in the mobile *app*, the redirect needs to be different, see
https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/www/index.html#L11

oscarmrom:
Hello! Initial contributor here. 

I started to take a look at this and give it a shot. At very minimum this is a rough draft of what it could look like. 

![image](https://user-images.githubusercontent.com/26697231/136024954-ad09c5fe-4aae-423d-8c09-0ba0740f4a3e.png)

Due to the way the toggle works based on the button I made the `Select Language` button have a larger z-index than the `language-select-container` so you can still toggle it off if you wanted to without clicking on a language and forcing a refresh. 

I added a media query to the `#language-select-container` of `max-width:700px;`.

What do people think?

jld3103:
To me the toggle button looks somewhat random placed (although it's not). I'd opt for an X at the top right corner to close it. Otherwise it looks good to me.

Kongkille:
In connection with historical view we are looking a bit further into this as well

PR #1971, opened 2019-09-26
#1841 add responsive language select

grayrattus:
It's my first PR on github ever so please don't get mad when I do something wrong ;)

image

image

Here is how it looks on Firefox 68.0 (64-bit) Linux Mint.


ovbm:
Hey there, thanks for the contrib, will review soon.

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/1841

web/src/styles.css
```diff
@@ -999,6 +1000,10 @@ body {
     cursor: pointer;
 }
 
+#language-header {
+    display: none;
+}
+
```
ovbm: this is overwritten below, should be removed.

grayrattus: It's overwritten because I don't want to show this header for desktops. The reason is that for desktops there is a tooltip on the button opening the language menu.

`display: none;` is in media query.

ovbm: my bad! :)

ovbm:
Hi, I pulled the branch and looked at it on my mobile screen and in my opinion, the selector is too large. Especially the fonts, they doesn't blend and mesh well with the other UI. 

Issue #1971 doesn't really give any reason for making the language select full screen, but I suspect that the reason it was opened was because **the language select is too small on mobile and the hit-zones are too close**.

That doesn't mean however, that we have to make it full-width. 

My suggestion is increasing the padding of the list items, increasing the fonts by maybe 10-20% and maybe adding horizontal lines, so it's clearer where to click to change the language.

In any case, sorry for not getting on the issue earlier - I think it would be great to solve the problem of the list item size, but we should have discussed this when the issue was raised so as to not cause unnecessary work to be done. 

Would you be up for giving my suggestions a go? Else I can take over. 

Thanks!

grayrattus:
@ovbm So you want to remove this PR and only increase the padding in current layout?

If so for me it's ok ;)

ovbm:
Yes, padding on the list-items, and fontSize, however not as extreme as in this PR.

corradio:
@ovbm @grayrattus status here?
Is it normal the PR is approved when changes seem to be requested?

grayrattus:
@corradio I though @ovbm wanted to add his changes regarding paddings, font-size and then merge this PR with his changes.
I'll try to fix it today. Sorry for making you wait.

corradio:
No worries, thank you for the help!

grayrattus:
I applied changes. Please check if it's ok.

corradio:
@grayrattus sorry for the delay, this must have slipped through the cracks..
@ovbm can you please review?

web/src/styles.css
```diff
@@ -2305,4 +2305,77 @@ div::-webkit-scrollbar-track-piece {
 div::-webkit-scrollbar-thumb {
     background: lightgray;
     border-radius: 3px;
+
```
ovbm: requires a "}". I made a mistake when merging the master branch.

corradio:
What is the status here? @ovbm @grayrattus

corradio:
Closing due to inactivity

web/src/styles.css
corradio: @ovbm if this was a mistake on your side, can you fix it?

Issue #1966, opened 2019-09-21
Croatia daily solar generation forecast available


seba1337:
Croatia has publicly available data (in excel :cry:) about the forecast of daily solar power generation.
https://www.hrote.hr/planning-electricity-production-for-the-eco-balance-group (at the bottom)

We could use this data and further refine the Croatia's unknown.

On ENTSOE for example Slovenia most likely reports similar data i.e. forecast. If I check the data generated for today I can see the solar MW for whole day even though it's only 9 AM, but not for other power sources.

systemcatch:
We might be able to bypass the excel by using the json url (https://files.hrote.hr/files/EKO_BG/FORECAST/SOLAR/FTP/TEST_DRIVE/22.9.2019.json) which supplies data for the graphs.

zilefaith:
Great and very informative post. I really appreciate your effort in this post. I got it what you have tried to explain about [Solar panels ](https://estimate.solar)devices.

OlgaCh:
Hello
I have large experience in web-scraping/data parsing.
Would love to work on this issue during `hacktoberfest`.
If you're ok with it could you please assign an issue to me? Thanks a lot!

martincollignon:
Hi Olga - sounds great! Would you start working on it now? Otherwise let me know when you start working on it and I'll assign it to you then :-)

OlgaCh:
@martincollignon I was planning to start working on it this weekend but should submit PR in October to stay compliant with the rules.

martincollignon:
I'll assign it to you then. I'll remove it on October 1st if not PR'ed/heard from you to let others help out :)

OlgaCh:
@martincollignon writing just to let you know that work is on track. Will submit PR tomorrow :)

PR #1987, opened 2019-09-30
Add solar production data for Croatia

OlgaCh:
Closes https://github.com/tmrowco/electricitymap-contrib/issues/1966

Updates HOPS parser with the additional data source with the estimated power production by solar panels: https://www.hrote.hr/planning-electricity-production-for-the-eco-balance-group


OlgaCh:
@martincollignon could you please review this PR?

parsers/HOPS.py
```diff
@@ -7,13 +7,46 @@
...
+    df = pd.DataFrame.from_dict(obj['FullPower']).set_index('Timestamp')
+    solar_production_dt = datetime.strftime(dt, '%Y-%m-%d %H:00:00')
+    try:
+        solar = df['Value'].loc[solar_production_dt]
+        # Value is in string format and needs to be converted to float
+        solar = float(solar.replace(",", "."))
+        # Converting to MW
+        solar *= 0.001
+    except KeyError:
+        logger.info("No value for Solar power production on {0}".format(solar_production_dt))
```
corradio: should probably be a warning instead?

parsers/HOPS.py
```diff
@@ -29,21 +62,26 @@ def fetch_production(zone_key='HR', session=None, target_datetime=None,
     df = pd.DataFrame.from_dict(obj['resources']).set_index('sourceName')
 
     # Get the wind power generation
-    wind = df['value'].loc['Proizvodnja VE']
+    # In some cases value may be negative. It looks like an issue at data source side.
+    wind = abs(df['value'].loc['Proizvodnja VE'])
```
corradio: maybe it makes more sense to consider negative values as 0? We could do max([0, ...])

parsers/HOPS.py
```diff
@@ -7,13 +7,46 @@
...
+    solar_production_dt = datetime.strftime(dt, '%Y-%m-%d %H:00:00')
+    try:
+        solar = df['Value'].loc[solar_production_dt]
+        # Value is in string format and needs to be converted to float
+        solar = float(solar.replace(",", "."))
+        # Converting to MW
+        solar *= 0.001
+    except KeyError:
+        logger.info("No value for Solar power production on {0}".format(solar_production_dt))
+        solar = 0
```
systemcatch: Probably needs to be `solar = None` here

parsers/HOPS.py
OlgaCh: Converted this one to the positive value since noticed that HOPS UI do the same.
They have negative value in API response but when display it at website it gets converted to the positive value. It makes me think that there may be some issues in the data itself.

parsers/HOPS.py
OlgaCh: @systemcatch solar set to 0 since in Python3.6+ it's not allowed to make numeric operations with `None`. So line `unknown = df['value'].loc['Ukupna proizvodnja'] - wind - solar` would fail.
Though if it's important for data post-processing I will set it to None at the returned json.

Sample output is:
```
[{
    'zoneKey': 'HR', 
    'datetime': datetime.datetime(2019, 10, 1, 11, 32, 16, tzinfo=tzutc()), 
    'production': {
        'wind': 12.0, 
        'solar': 33.006206, 
        'unknown': 1205.993794
    },
    'source': 'hops.hr'
}]
```

jarek: Internally in Electricity Map we do make a distinction: 0 means production reported at 0 (e.g. solar at night), while None means "unknown, not reported, no data available"

parsers/HOPS.py
OlgaCh: Updating to `warning`.

parsers/HOPS.py
OlgaCh: @jarek :+1: 
Indeed in this case makes sense to return `None` if solar data not available.

OlgaCh:
Updated PR to address requested changes.

Also noticed that in some cases data file with the solar data estimation may be not available for the current date since being updated with the delay.
Thus parser changed to get the list of all available data files and then download most recent one to get the solar power production.

OlgaCh:
@corradio @systemcatch kindly check updated pr.

parsers/HOPS.py
corradio: Let's do it the same way as they do it - thanks for clarifying

PR #1988, opened 2019-09-30
webpack: use mini-css-extract-plugin

simon04:
Ref: https://webpack.js.org/plugins/mini-css-extract-plugin/

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -45,12 +45,11 @@
     "eslint-plugin-import": "^2.9.0",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.7.0",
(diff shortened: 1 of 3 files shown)


corradio:
Hi @simon04 ,

Thanks for the PR. Could you detail a bit more the rationale here, and the consequences? Just need to make sure our deployment process will be compatible.

Thanks,

Olivier

simon04:
extract-text-webpack-plugin [is deprecated](https://github.com/webpack-contrib/extract-text-webpack-plugin) and mini-css-extract-plugin is its suggested drop-in replacement. For some advantages, please see https://github.com/webpack-contrib/mini-css-extract-plugin#mini-css-extract-plugin

PR #1994, opened 2019-10-01
Rebuild and refactor ENTE parser

### LINKED ISSUE #1930: ENTE exchanges no longer work
The enteoperador site has undergone a big refresh and the old url from which we scraped data no longer works (http://www.enteoperador.org/newsite/flash/data.csv)

New site + map
https://www.enteoperador.org/flujos-regionales-en-tiempo-real/

The map above is populated by data from the url below
https://mapa.enteoperador.org/WebServiceScadaEORRest/webresources/generic

Looks like they've moved from csv to json and also added a load more data to the graphic. This likely means rewriting some of the ENTE parser.

systemcatch:
Few more tweaks and this should be ready.

closes #1930


timrodz:
Awesome, @systemcatch! If this gets pulled, I'll be sure to add it to the CR one, which is breaking exactly because of this issue.

systemcatch:
Ready for review guys.

new_ente

timrodz:
> Ready for review guys.
> 
> new_ente

Hi @systemcatch - Looks like this does fix the regional parsers.

Based on your changes, I can guess we can remove the `fetch_exchange` entries for the individual parsers in Central America (Since they're already being handled by `ENTE` parser?

systemcatch:
> Based on your changes, I can guess we can remove the `fetch_exchange` entries for the individual parsers in Central America (Since they're already being handled by `ENTE` parser?

@timrodz yes though every exchange bar HN->NI (which is still working) already uses the ENTE parser. However if you want to do some old code cleaning then feel free to create another PR.

timrodz:
> > Based on your changes, I can guess we can remove the `fetch_exchange` entries for the individual parsers in Central America (Since they're already being handled by `ENTE` parser?
> 
> @timrodz yes though every exchange bar HN->NI (which is still working) already uses the ENTE parser. However if you want to do some old code cleaning then feel free to create another PR.

For sure, I'll be sure to check that parser for reference and update the existing codebase :)

Issue #968, opened 2018-01-02
Units missing from left panel charts


brunolajoie:
- Add units to the top right corner of each chart
- Convert second chart units in GW
![image](https://user-images.githubusercontent.com/22095643/34480219-5eb6622c-efab-11e7-8b00-eba20531365b.png)

corradio:
@jbuverud if I remember correctly you had a prototype of this working no?

jbuverud:
No prototype, just visual ideas. I was considering adding a vertical axis label with the unit, and then maybe add k/M/G-prefixes to the ticks so they don't overflow. However now that I think about it some more I think this will look strange, seeing that we want the ticks to be on the right side of the graph, and the unit label should be with the ticks. Maybe the best solution is simply to add the units (with prefixes) directly in the ticks.

max-arias:
This ticket should be closed, fyi

PR #1995, opened 2019-10-01
Added y axis label to Line- and Areagraph

DrFaust:
This is a PR for issue https://github.com/tmrowco/electricitymap-contrib/issues/968

I placed the unit labels parallel to the axis and not at the top because it provided more space this way.
The second chart is now in GW.

Happy for feedpack!


alixunderplatz:
@DrFaust cool! can you post a screenshot so we can see what it looks like?

DrFaust:
> @DrFaust cool! can you post a screenshot so we can see what it looks like?

screenshot

alixunderplatz:
@DrFaust Thank you, well done! 👍 

Two minor things that I noticed:
1. The specific emissions unit is "g/kWh".

2. A more general thing: In the related issue #968 it says to change the units of the middle chart to "GW" instead of "MW", which you have done. So for smaller zones with less production (islands like Aland, Bornholm, El Hierro,...), will we get decimal numbers now like "0.04 GW"? Sorry for asking, I can't run the code at the moment to check this. ;)
I'm just thinking if it is better to dynamically change to MW if the maximum value on the chart doesn't exceed 1 GW. Will that be easy to implement?
Any thoughts on this, @corradio & @martincollignon?

corradio:
Agreed with @alixunderplatz for the two points.
If memory serves well, I think the tooltip already has this functionality of writing "GW" or "MW" depending on the scale:
image

Maybe you can reproduce that behaviour? Thanks for looking into this though!

DrFaust:
Good point @alixunderplatz I updated the first label text and will have a look at the tooltip like @corradio proposed.

DrFaust:
Hi @alixunderplatz & @corradio. I changed it so that the unit of the second graph is dynamically set depending on the data. For a maximum with more than 1 GW its GW units otherwise it will be MW.

Would be glad if you would have time to take a look..

web/src/components/areagraph.js
```diff
@@ -70,6 +74,28 @@ AreaGraph.prototype.data = function (arg) {
...
+        } else {
+            // in tCO2eq/min
+            return (d.totalCo2Production + d.totalCo2Import + d.totalCo2Discharge) / 1e6 / 60.0;
+        }
+    })
+
+    var formattingFactor = 1;
+    if (!that._displayByEmissions) {
+        // Display in GW or MW
+        formattingFactor = maxTotalValue > 1e3 ? 1e3 : 1;
```
corradio: Instead I suggest using this:
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/helpers/formatting.js#L13

DrFaust: I tried to use d3.format before, but failed. Output of that function is a formatted string, which would not work here since we need to align the total dataset dynamically accordingly. Thats why this threshold is implemented.

corradio: What do you mean exactly by "align the total dataset dynamically accordingly"? I'm trying to figure out if we can find a more elegant solution because some areas have `kW` scale data.

DrFaust: I see. Sure 👍 . Added a new scalePower formatter.

DrFaust: I just meant that data and domain need to be scaled.

web/src/components/areagraph.js
```diff
@@ -126,14 +157,7 @@ AreaGraph.prototype.data = function (arg) {
...
         0,
-        1.1 * d3.max(arg, function(d) {
-            if (!that._displayByEmissions) {
-                return d.totalProduction + d.totalImport + d.totalDischarge;
-            } else {
-                // in tCO2eq/min
-                return (d.totalCo2Production + d.totalCo2Import + d.totalCo2Discharge) / 1e6 / 60.0;
-            }
-        })
+        maxTotalValue
```
corradio: can you multiply here by 1.1? It gives a bit of space between the graph and the axis (and then we keep the existing behavior)

Issue #1969, opened 2019-09-24
Toggle consumption / production looks weird at open


martincollignon:
Mac / Chrome. This is how the toggle looks like to me. Clicking on it makes it normal again.
<img width="304" alt="Screenshot 2019-09-24 at 10 21 22" src="https://user-images.githubusercontent.com/2604526/65494491-3e77db80-deb5-11e9-87cf-3fdfe1b15c43.png">

corradio:
@ovbm are you able to help? Maybe it's related to your last PR?

ovbm:
Can't reproduce on Firefox & Chrome. Could reproduce on Safari, on every other refresh O.o. No js errors.

Chopinsky:
I can reproduce this issue on my Chrome as well. I think the issue comes from the first-load, on any ensuing refreshes or reloads, the toggle button works just as expected. The problem seems to come from the code @jamescrowley has mentioned: the overlay's `left` and `width` properties are miscalculated on the first run.  

I'd like to work on this issue, it shouldn't be too complex imho, but if anyone could point me to where the first-load logic lives, that would be of great help!

PR #2005, opened 2019-10-02
Fix "Consumption/Production" toggle button's UI issue (#1969)

Chopinsky:
This PR is trying to fix the issue as raised by issue #1969, where the "Consumption / Production" toggle button has a weird UI issue on loading the web page for the first time in a (desktop) browser. 

The problem was caused by a typical out-of-the-order UI updates: when rendering a web page for the first time, browsers may choose to render components freely, meaning that certain components may be rendered before some others, causing the boundaries and sizes of the components changing as more elements are adding to the page. This abnormality (or a JavaScript feature?) usually cease to happen when the page is re-rendered or refreshed, because the browser have cached the sizes and positions of the elements, making them shifting less often.

So the fix will address the core problem in the heart of the issue: we'll now defer the render of the overlay of the "Consumption / Production" toggle button until most of the elements are drawn and in stable states, i.e. right before we lower the "map loading" shield. 

In addition, I've refactored code that will lower the "map loading" shield, such that if we encounter similar UI loading issues in the future, a remedy logic can be added in the same place in the future.


ovbm:
Nice, that's a very thorough way to fix the issue. I'll test it asap!

Issue #2009, opened 2019-10-03
[UI Issue] "color blind mode" checkbox text misalignment


Chopinsky:
It seems that the "color blind mode" checkbox's text is slightly misaligned with the box. This should be fixable with some CSS. 

- Chrome
![image](https://user-images.githubusercontent.com/5376280/66142465-0bc39680-e5cb-11e9-93a5-0ff03ecc899e.png)

- Firefox
![image](https://user-images.githubusercontent.com/5376280/66142813-c489d580-e5cb-11e9-9d34-967a5a31eb21.png)

- Edge
![image](https://user-images.githubusercontent.com/5376280/66142870-d9666900-e5cb-11e9-8899-0a20881d50e0.png)

In addition, it seems that the checkbox is quit buried in the lower right corner of the sidebar, making it less accessibility-friendly. 

Ideally, we probably should make it into a toggle button along with right side toggle buttons (e.g. right below the dark-mode button), but at least we can make the checkbox larger and right beneath the country selection list.

lanpai:
Could I handle this?

ystanwar:
Hi cab I take this issue

fbarl:
It seems that #2010 addressed this issue for the desktop mode:

![image](https://user-images.githubusercontent.com/5376280/66222550-0a5fa000-e697-11e9-8a7c-8c233a69472c.png)

So the only piece of work left is to update the mobile view with the same styles:

![image](https://user-images.githubusercontent.com/1216874/72525941-90c67900-3865-11ea-98ea-e0d9bfc9c831.png)

corradio:
OK. Closing this issue. feel free to re-open if the issue reappears

PR #2010, opened 2019-10-03
[UI Fix] Update checkbox styles and make it obvious for issue #2009

Chopinsky:
This PR is trying to fix a styling issue with the "color blind mode" checkbox, as stated in issue #2009. We're going to define a few container CSS classes that will help improve the user interaction with the checkbox, and moving the checkbox up, such that it will be more obvious to users. 

- Before:
image

- After:
image

The updates only concern the web version of the checkbox, the style of the mobile version is left untouched, yet we've moved the checkbox up, and now it's right below the country selection list as well, such that the layout is kept consistent with the web version.


systemcatch:
Hello @Chopinsky, can you post a full screenshot of the new map for context.

Chopinsky:
Sure, here's the screen shot of the map with the updated checkbox:

image

systemcatch:
@Chopinsky looks good! Let's see what @ovbm thinks.

ovbm:
hey, i like it, could we have a border-radius of 2-4px so that it’s more coherent with the rest of the UI?

Chopinsky:
> hey, i like it, could we have a border-radius of 2-4px so that it’s more coherent with the rest of the UI?

I've added a border radius to `6px` (which is consistent with map's right sidebar toggle buttons), and you can see the effect in the screen shot below. 

image

I've tried `2px` and `4px` options, the checkbox's angle curvature are less obvious, but it might be more consistent with the social media buttons ate the bottom of the left side panel.

web/src/styles.css
```diff
@@ -2184,3 +2280,76 @@ div::-webkit-scrollbar-thumb {
     border-radius: 3px;
 }
 
+/* 
+    custom "color blind mode" checkbox and others
+    adopted with tweeks from source: https://www.w3schools.com/howto/howto_css_custom_checkbox.asp 
+*/
+
+/* The container */
+.container {
```
ovbm: I think we should call this checkbox-container. "container" is too general

web/views/pages/index.ejs
```diff
@@ -158,22 +158,23 @@ co2Sub = function(str) {
                         </div>
 
                         <div class="info-text small-screen-hidden">
+                            <p>                                
+                                <label class="container"><%- __('legends.colorblindmode') %>
```
ovbm: class="checkbox-container"

ovbm:
6px looks great! thanks

PR #2013, opened 2019-10-04
Adding Indonesian translation in id.json

sahmura:


jarek:
Hi @sahmura, thank you very much for the translation!

Could you change the following:
- When country/region name is missing, the English name is automatically used. As far as I can tell, the country names haven't been translated to Indonesian - which is totally fine! - but in this case we should just leave out the entire `zoneShortName` section and it'll work the same.

Two changes that would be nice to have, but not required - we can do this for you if you're not sure what this means or how it works:
- If your editor supports it, it would be good to have the file indented with 4 spaces per indent
- To have the translation actually show up on the website, the `id` language code needs to be added in several places - there is a list of them in https://github.com/tmrowco/electricitymap-contrib/issues/1862

BTW, are you aware of a source of real-time electricity data for any of Indonesian regions? For example we know of http://wesm.ph/ for Philippines and https://www.gso.org.my/ for Malaysia - is there anything like this online for Indonesia?

sahmura:
Thank you for your feedback. I will correct it and add some translations that might still be wrong and follow your directions.

corradio:
Hi @sahmura ,
Did you have time to adresse the changes?
Also, could you add the locale id to:
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/server.js#L34-L99
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/helpers/translation.js#L3
and
https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/generate-index.js#L29-L78

Thank you ;-)

corradio:
@sahmura did you have time to look at this?

sahmura:
Hi!

I'm so sorry I can only give feedback to you guys now because of my busy life in lectures and many projects that I'm working on. I have corrected some of the wrong words and translated all the words into Indonesian.

sahmura:
Thank you

sahmura:
Let me know what should i do next ;-)

corradio:
Great! Last two places to add the locale definitions are:
- https://github.com/tmrowco/electricitymap-contrib/blob/master/web/locales-config.json
- https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/www/index.html#L5
then we should be good to go!

sahmura:
Done :)

PR #2015, opened 2019-10-04
Improve map navigation for keyboard-only users

ShahneRodgers:
Not all key presses should automatically focus on the input box; many others are used by people who can't use a mouse, and the arrow keys are used for navigating the map. I think just jumping there automatically for text input makes sense; it's unlikely the user will want to type anything else into that box.

I've also added 'tabindex' to the FAQ link and to hide the popover so that keyboard-only users can interact with them.

I'm happy to do more accessibility fixes if wanted. :)


systemcatch:
Hey @ShahneRodgers thanks for the PR, I'll build the map and see how your changes function. It'd be great to make the map more accessible for everyone.

Issue #1857, opened 2019-05-18
Make electricity consumption pop-ups more user friendly when no data


fbarl:
#### Suggestions

1. When there is no data available for a particular energy source (e.g. Iraq's biomass), the content would read more nicely if it was replaced by a single sentence:

    _Biomass consumption data is not available for Iraq._

2. When the consumption for a particular energy source equals 0 (e.g. Switzerland's geothermal), the last paragraph _with a carbon intensity..._ could probably be dropped as it's not relevant.
    * p.s In case of installed capacity being zero, the `NaN` should be replaced by `0%`.

3. When the installed capacity is unknown (e.g. Poland's geothermal), the last two paragraphs could probably be merged into a sentence like:

    _Utilizing 0.00W with a carbon intensity of..._

4. In all cases, making percentages **bold** and adding 1-2 decimal points to them is something I'd also consider

#### Screenshots

![image](https://user-images.githubusercontent.com/1216874/57972185-5388af00-7997-11e9-8d35-84a26a9c6f5c.png)

![image](https://user-images.githubusercontent.com/1216874/57972277-7cf60a80-7998-11e9-87bc-314edfc92c0a.png)

![image](https://user-images.githubusercontent.com/1216874/57972529-63ef5880-799c-11e9-9ce6-a2b607454c92.png)

WagnerMoreira:
@systemcatch can I work on this idea or are you planning to do this already?

lacexd:
@systemcatch I'd also love to work on this one.

systemcatch:
@WagnerMoreira @lacexd 

No I'm not planning on tackling this issue anytime soon, so feel free to have a crack at it. I suggest working together would make this easier.

lacexd:
@systemcatch @WagnerMoreira I've created a PR with the first 2 tasks done so you could take over with the last 2 if you wish.

IanvsPoplicola:
Hi everyone, I gave the last task a go, and made a PR; please let me know if you have any feedback!

Task 3 is pretty complex because it involves editing the actual text of each locale translation, so I don't think I'm qualified to take on that alone.

corradio:
Closing due to inactivity

fbarl:
cc @Kongkille in case you think parts of this could be useful for the next iteration of the UI :wink: 

[...]

PR #2016, opened 2019-10-05
Make electricity consumption pop-ups more user friendly when no data …

lacexd:
…#1857


corradio:
@lacexd friendly ping

PR #2035, opened 2019-10-14
Create Korean translation

mikkelrev22:
starting the korean translation for the locales


systemcatch:
Nice, let us know when it's ready for review.

corradio:
Hi @mikkelrev22 ,

Thanks for the PR! Can you add the locale to:

https://github.com/tmrowco/electricitymap-contrib/blob/master/web/server.js#L34-L99
https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/helpers/translation.js#L3
and
https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/generate-index.js#L29-L78
Thank you ;-)

PR #2039, opened 2019-10-17
Extend MX parser with historical production data

scriptator:
I created a parser for historical production data of Mexico. The data for a month gets available some weeks after it ended. See also the issue #1242 

Some questions:
- Should we add add an entry for Mexico to `zones.json` when it's only historical data?
- How many data rows before and after can I return from `fetch_production` when a certain `target_datetime` is requested.

### DIFF
--- a/parsers/MX.py
+++ b/parsers/MX.py
@@ -50,7 +50,7 @@ DATA_CACHE = {}
 
 
 def parse_date(date, hour):
-    tzoffset = tz.tzoffset("CDT", -3600 * 6)
+    tzoffset = tz.tzoffset("CST", -3600 * 6)
     dt = datetime.datetime.strptime(date, "%d/%m/%Y")
     dt = dt.replace(hour=int(hour) - 1, tzinfo=tzoffset)
     return dt
@@ -123,7 +123,7 @@ def fetch_production(zone_key, session=None, target_datetime=None):
         raise ValueError("Parser only supports fetching historical production data, please specify a terget_datetime in the past")
     
     # retrieve data for the month either from the cache or fetch it
-    cache_key = target_datetime.strftime("%Y-%d")
+    cache_key = target_datetime.strftime("%Y-%m")
     if cache_key in DATA_CACHE:
         df = DATA_CACHE[cache_key]
     else:
--- a/parsers/MX.py
+++ b/parsers/MX.py
@@ -1,10 +1,19 @@
 #!/usr/bin/env python3
 
 import arrow
+import datetime
 import pandas
 import requests
+import urllib
+import pandas as pd
+
 from bs4 import BeautifulSoup
 from collections import defaultdict
+from dateutil import tz, parser
+from io import StringIO
+
+
+MX_PRODUCTION_URL = "https://www.cenace.gob.mx/SIM/VISTA/REPORTES/EnergiaGenLiqAgregada.aspx"
 MX_EXCHANGE_URL = 'https://www.cenace.gob.mx/Paginas/Publicas/Info/DemandaRegional.aspx'
 
 EXCHANGES = {"MX-NO->MX-NW": "IntercambioNTE-NOR",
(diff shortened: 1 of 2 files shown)


parsers/MX.py
```diff
@@ -22,6 +31,115 @@
...
+    "Eolica": "wind",
+    "Fotovoltaica": "solar",
+    "Biomasa": "biomass",
+    "Carboelectrica": "coal",
+    "Ciclo Combinado": "gas",
+    "Geotermoelectrica": "geothermal",
+    "Hidroelectrica": "hydro",
+    "Nucleoelectrica": "nuclear",
+    "Combustion Interna": "unknown",
+    "Termica Convencional": "unknown",
```
scriptator: I don't know where to assign these two. Is unknown okay or do you have better ideas?

parsers/MX.py
```diff
@@ -22,6 +31,115 @@
...
+    "Turbo Gas": "gas"
+}
+
+# cache where the data for whole months is stored as soon as it has been fetched once
+DATA_CACHE = {}
+
+
+def parse_date(date, hour):
+    tzoffset = tz.tzoffset("CDT", -3600 * 6)
```
scriptator: I do not know whether that is correct also in case of daylight saving time

corradio:
Thanks for the PR!

> * Should we add add an entry for Mexico to `zones.json` when it's only historical data?

yes

> * How many data rows before and after can I return from `fetch_production` when a certain `target_datetime` is requested.

it's up to you. The backend will adapt

parsers/MX.py
corradio: I think unknown is the best we can do at this point

parsers/MX.py
scriptator: I could not find a source stating which time zone the data is reported in but I assume that is Central Standard Time (CST) where Mexico City belongs to, which is UTC-6. There are no holes in the data when daylight saving time changes happened so I guess that daylight saving time is never reported.

corradio:
I just realised we have MX split up as several zones in the current zones.json.
Unfortunately our system doesn't allow collecting data for both.
We'll have to put this on pause temporarily until we either unsplit MX or get the ability to gather data at two different levels.
Sorry for not catching this earlier.

scriptator:
> I just realised we have MX split up as several zones in the current zones.json.

@corradio are you sure? I can't find any zones in `zones.json` which have MX in them . However, I do see some exchanges with MX in `exchanges.json`.

corradio:
You're right sorry, I meant exchanges.json.

systemcatch:
@scriptator looks nice, can you post some output from the parser?

@corradio I think we should unsplit Mexico as historical production is better than unknown real-time exchange imo.

scriptator:
> @scriptator looks nice, can you post some output from the parser?

Sure, here is the data for 07, 08 and 09 of 2019. Contains both the list which the parser returns and the (time-zone corrected) raw API output.
[MX-parser-data.zip](https://github.com/tmrowco/electricitymap-contrib/files/3762607/MX-parser-data.zip)

systemcatch:
Checked the output and it looks ok, except we need a `datetime` object instead of `Timestamp('2019-07-01 00:00:00-0600', tz='tzoffset('CST', -21600)')` for the datetime key.

Do you want to rejoin Mexico into one or shall I do it?

scriptator:
> Checked the output and it looks ok, except we need a datetime object 

Oh yes, sure. I'll change that

> Do you want to rejoin Mexico into one or shall I do it?

It would be great if you could do that, thanks!

corradio:
Thanks guys!

systemcatch:
So there may be a few issues with rejoining Mexico into one zone.

1. The MX-BC (Baja California) exchange with California will be lost.

2. The exchange with Texas/ERCOT will stop working.

Both of these rely on fixed carbon intensities calaculated by us or third parties so they are not as good as actual data. However we need to decide about keeping these exchanges.

The long term solution is probably implementing the ability to gather data on multiple levels as per @corradio's suggestion.

systemcatch:
@corradio do you have any thoughts on my previous comment?

corradio:
Hi @systemcatch 
Sorry about the delay. We are staffing the team and from Q1 2020 there will be one more person able to help here.
In the meantime, I don't think there's much more we should do here.
Historical data is less important than real time, which means that we probably shouldn't touch anything as of now.
In January we will take a larger look at all initiatives (including this one) and validate what is the best direction. The code is definitely valuable and we'll see exactly how we can use it to not waste any data (data is precious!).

Thank you all and sorry for the delay

corradio:
Here's a summary of the issue (cc @robertahunt):
The data provider has historical data on national level, but has real-time data on interconnector level. This PR adds the national historical data.
Currently our system doesn't allow us to display both levels on the map. However, we can collect both levels.
The modifications requires updating the MX zone to add the `subZoneNames` key describing the hierarchy between the zones:

https://github.com/tmrowco/electricitymap-contrib/blob/3962a6a478fa26a7d090f7e573d47b100e6b700a/config/zones.json#L1000-L1003

@scriptator are you able to make those changes? Then we can merge and start collecting data, even though it won't show national data on the map.

scriptator:
Yes, I can do that. Probably next week. Thanks for taking the issue up again!

corradio:
ping @scriptator in case you forgot

scriptator:
> ping @scriptator in case you forgot (and thanks for helping us out with this!)

@corradio yes I completely forgot. Thx for reminding. I added now all subzones which I could find in `config/exchanges.json`. Hope, that's all

robertahunt:
One typo, otherwise it looks good to me!

PR #2043, opened 2019-10-18
Create Vietnamese translation

hoanglongtran:


corradio:
Hi @hoanglongtran ,

Thanks for the PR! Can you add your locale to:
- https://github.com/tmrowco/electricitymap-contrib/blob/master/web/server.js#L34-L99
- https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/helpers/translation.js#L3
and 
- https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/generate-index.js#L29-L78

Thank you ;-)

martincollignon:
@hoanglongtran could you do the requested changes so we can add your translations? :)

PR #2046, opened 2019-10-19
Update Poland solar capacity

wojciej:
Solar capacity update:
     old value  "solar": 882,
     new value "solar": 1007,


jarek:
Thanks for the update! What is the source of this figure?

Currently we indicate in our main README that source for Polish capacities is ENTSOE, however [ENTSOE's page](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show?name=&defaultValue=true&viewType=TABLE&areaType=BZN&atch=false&dateTime.dateTime=01.01.2019+00:00|UTC|YEAR&dateTime.endDateTime=01.01.2020+00:00|UTC|YEAR&area.values=CTY|10YPL-AREA-----S!BZN|10YPL-AREA-----S&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19) only has 430 MW for solar for Poland for 2019.

wojciej:
Hey Jarek,
PSE ( Polish electroenergetics network ) every month provide update about installed capacity of PV in Poland.
https://twitter.com/pse_pl/status/1182278197204934657
As I have noticed previous update was made based on that too I thought we could update it again.

I have contact Entsoe about how often the data in their system should be updated and in a reply I was sent to this documentation 
https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:32013R0543&from=EN
However I could really find any timeline of when TSO (Transmission System Operators in our case PSE )should send data to ENTSOE.
As far as I have noticed so far they have been doing this once a year... what kinda miss the point of having actual data...

With that information I have also contacted PSE with a question why dont they sent an update of this to ENTSOE but I have got no answer so far....


On a side note I have also contacted ENTSOE with a question why there i no information about electricity generation from PV and I have got the following answer:
_"The data for actual generation per production type solar for Polish area are not published on the platform due to not comply with the condition of the obligation to publish it according to article 16.2.c of the Regulation No 543/2013 “The information shall be provided for all bidding zones only in Member States with more than 1 % feed-in of wind or solar power generation per year or for bidding zones with more than 5 % feed-in of wind or solar power generation per year”._

So according to that once we hit 1% the data should be finally available.

Hope this helps with your question.
Thanks
Jarek

jarek:
@corradio looks good to me, open question how you want to cite this (extra line in data sources in readme linking to the tweet?)

wojciej:
@jarek  @corradio  in this case please consider this - we can either add url to the particular tweet or to PSE account on twitter https://twitter.com/pse_pl

systemcatch:
@wojciej let's add the direct link to the tweet in the production capacity sources in the readme.

PR #2060, opened 2019-10-22
Update translation.js by adding Korean/KR

mikkelrev22:
Added KR (Korean) to Locales_List array and languageNames array


corradio:
Closing as this should be part of https://github.com/tmrowco/electricitymap-contrib/pull/2035

Issue #2062, opened 2019-10-23
"Select language" menu has inconsistent capitalisation


Ogaday:
The names of some languages are capitalised and others are not. This looks inconsistent:

![image](https://user-images.githubusercontent.com/6689770/67429679-91d66a00-f5d8-11e9-95ce-70bda3ca1b6e.png)

I suggest that they should all be capitalised - this also matches how Wikipedia handles language names, for instance:

![image](https://user-images.githubusercontent.com/6689770/67429869-f42f6a80-f5d8-11e9-85f1-3e6c79d66c3f.png)

leternoz:
Hi, may I take this one ?

Ogaday:
Sorry @leternoz, I already made a PR but I messed up the commit message so the PR wasn't linked to this issue.

leternoz:
ok fine

jarek:
This has now been merged in #2063, closing issue

PR #2063, opened 2019-10-23
Capitalise language labels

Ogaday:
This commit updates the language names in the front end under the
"Select language" dropdown to be capitalised. This makes the menu appear
more congruous, and matches the behaviour of other internationalisation
menus, such as on Wikipedia.


Ogaday:
This closes #2062. I messed up the magic GitHub issue linking in the commit message.

jarek:
For what it's worth, I believe this to be incorrect for Polish. Polish does not capitalize language names. But I don't feel about this extremely strongly.

Wikipedia desktop version appears to capitalize all languages, but mobile version doesn't :man_shrugging: 

image

PR #2070, opened 2019-10-26
Update Nepal Capacity in zones.json

Subarna578:


Subarna578:
Please review the concerned file and merge it.

corradio:
Can you add the source of those capacity numbers in the README?

Subarna578:
Yes, I will add that. Thanks.

martincollignon:
As an exception I'll merge both. In the future it'd be great if you could work in one branch. Thanks for your help!

PR #2077, opened 2019-10-29
Improve formatting of percentage figures in tooltip text

IanvsPoplicola:
This implements _suggestion 4_ mentioned in #1857

The first commit increases the precision of the tooltip percentage figures to 2 decimal places.

The second commit changes the font of the tooltip percentage figures to bold.

image
_View on local build_

I'm perfectly happy to make further changes to this as required.


corradio:
thx!

PR #2084, opened 2019-11-06
Correct kr.json

ymin2570:
Countries that were not modified in the previous work were modified.


ymin2570:
Sir, I have a two questions.
First one is "SPP" and "MISO" and "PJM" of the United states(The same applies to Chile's SING / SIC). I think these are something different with actual USA's states name.
Second, Can't I translate anything that isn't in a 'en.json' file like US_MT or US_UT?

martincollignon:
I'll merge these edits, could you clarify your question? I'm not sure I understand. Certain zones are not states but grids (and thus should probably be not translated)

ymin2570:
@martincollignon 

> I'll merge these edits, could you clarify your question? I'm not sure I understand. Certain zones are not states but grids (and thus should probably be not translated)

Okay then I understand that it is right to leave grids in English. 

My other question is that when I access the web, I see regions that aren't in the kr.json(or other json files) file, so they can't be translated?

PR #2093, opened 2019-11-17
Updated capacity values for Denmark, 2019

nikBeier:
The full capacity of the offshore wind farm "Horns Rev 3 may not be included.
Bornholm not updated. Where is the source for that?
Two production types at ENTSO-E are not yet listed in this json file: "Waste" and "Other renewables"

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -976,15 +976,16 @@
       ]
     ],
     "capacity": {
-      "biomass": 197,
-      "coal": 4847,
-      "gas": 2941,
+      "biomass": 1722,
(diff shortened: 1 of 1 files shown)


corradio:
Hi @nikBeier . Thanks for the PR! Waste and other renewable are currently counted as biomass. Could you add them there if you didn't do so before?

nikBeier:
> Waste and other renewable are currently counted as biomass. Could you add them there if you didn't do so before?

I was not aware of that. Now they are added, @corradio

corradio:
well done!

PR #2116, opened 2019-12-09
added watermark with link to job page

ovbm:
Screenshot 2019-12-09 at 21 58 26

next up, banner.


ovbm:
it doesn’t show on mobile. the watermark is only on desktop. that’s why it’s just a partial solution byt something i could do quickly.

ovbm:
can make it smaller

corradio:
Got it, that's an important distinction. My size remark still applies but that being said you can go ahead with the solution you believe is best.

ovbm:
Screenshot 2019-12-10 at 08 54 31

logo is now 160x24 vs 180x28, flag is smaller. will merge

Issue #2134, opened 2020-01-03
Countries offline: AG, PE, RU-1+RU-2, AU-WA


alixunderplatz:
Several countries are / were offline for a while now or have reoccurring issues.
Creating a checklist below so everyone can see what the issue could be and which parser had been fixed already. [Kibana-logger](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=()&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',asc))) error output below.
Should you work on fixing one of the parsers, please comment below and then tag this issue in the PR. :)

- [x] Argentina (AR.py)
Could fail due to a html-format issue:
'NoneType' object has no attribute 'getText'
fixed by #2139 

- [ ] Peru (PE.py)
Could fail due to a faulty/outdated date-time format:
Failed to match '(?P<M>\d{1,2})/(?P<D>\d{1,2})/(?P<YYYY>\d{4}) (?P<h>\d{1,2}):(?P<mm>\d{2}):(?P<ss>\d{2}) (?P<A>(am|pm|AM|PM))' when parsing '2020/01/03 01:00:00'
partly fixed in #2135 

Quoting Alex
> Data for 3rd of January up until last hour. So we should rather find a way to check for
start "today-1" and end "today"
and also
start "today" and end "today + 1" and merge the two datasets when both are returned.

> This way we will get the most recent stuff. This could be why the parser failed in the past for a couple of hours each day, from what I can remember.

- [x] Russia West + Russia East (RU.py) + Exchanges of Russia with neighbouring states
Maybe there was a change in the source data format? But running it locally works fine for me. Maybe too many requests?
Expecting value: line 1 column 1 (char 0)

- [x] Western Australia (AU_WA.py or /lib/AU_solar.py)
Maybe there was a change in the source data format?
Expecting value: line 1 column 1 (char 0)
fixed by #2140

systemcatch:
[...]

PR #2135, opened 2020-01-03
Fix Peru-parser

alixunderplatz:
- fixed datetime-format of the raw data
- fixed date values of the data-request
   - **NOTE: seems like fetching real-time data is not possible anymore.** 
Only yesterdays data can be parsed. I haven't found a way to retrieve/view live-data from: http://www.coes.org.pe/Portal/portalinformacion/generacion

Sample output:
`[{'zoneKey': 'PE', 'datetime': datetime.datetime(2020, 1, 2, 0, 30, tzinfo=tzfile('America/Lima')), 'production': {'biomass': 0.0, 'unknown': 0.0, 'coal': 0.0, 'oil': 0.0, 'wind': 248.67762, 'gas': 1027.1763, 'hydro': 8934.01388, 'solar': 0.0}, 'source': 'coes.org.pe'}, {'zoneKey': 'PE', 'datetime': datetime.datetime(2020, 1, 2, 1, 0, tzinfo=tzfile('America/Lima')), 'production': {'biomass': 0.0, 'unknown': 0.0, 'coal': 0.0, 'oil': 0.0, 'wind': 242.39494, 'gas': 957.50884, 'hydro': 8836.51988, 'solar': 0.0}, 'source': 'coes.org.pe'},`

### DIFF
--- a/parsers/PE.py
+++ b/parsers/PE.py
@@ -22,7 +22,7 @@ MAP_GENERATION = {
 
 
 def parse_date(item):
-    return arrow.get(item['Nombre'], 'M/D/YYYY h:mm:ss A').replace(tzinfo=dateutil.tz.gettz(tz))
+    return arrow.get(item['Nombre'], 'YYYY/MM/DD hh:mm:ss').replace(tzinfo=dateutil.tz.gettz(tz))
 
 
 def fetch_production(zone_key='PE', session=None, target_datetime=None, logger=None):
@@ -32,8 +32,8 @@ def fetch_production(zone_key='PE', session=None, target_datetime=None, logger=N
     r = session or requests.session()
     url = 'http://www.coes.org.pe/Portal/portalinformacion/Generacion'
     response = r.post(url, data={
-        'fechaInicial': arrow.now(tz=tz).format('DD/MM/YYYY'),
-        'fechaFinal': arrow.now(tz=tz).replace(days=+1).format('DD/MM/YYYY'),
+        'fechaInicial': arrow.now(tz=tz).replace(days=-1).format('DD/MM/YYYY'),
+        'fechaFinal': arrow.now(tz=tz).format('DD/MM/YYYY'),
         'indicador': 0
     })
     obj = response.json()['GraficoTipoCombustible']['Series']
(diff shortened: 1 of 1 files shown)


alixunderplatz:
ref #2134

parsers/PE.py
```diff
@@ -32,8 +32,8 @@ def fetch_production(zone_key='PE', session=None, target_datetime=None, logger=N
     r = session or requests.session()
     url = 'http://www.coes.org.pe/Portal/portalinformacion/Generacion'
     response = r.post(url, data={
-        'fechaInicial': arrow.now(tz=tz).format('DD/MM/YYYY'),
-        'fechaFinal': arrow.now(tz=tz).replace(days=+1).format('DD/MM/YYYY'),
+        'fechaInicial': arrow.now(tz=tz).replace(days=-1).format('DD/MM/YYYY'),
```
systemcatch: We need to use `.shift()` here.

```shell
>>> import arrow
>>> arw=arrow.utcnow()
>>> arw
<Arrow [2020-01-03T12:41:19.470793+00:00]>
>>> arw.replace(days=-1)
  File "<stdin>", line 1, in <module>
  File "/home/chris/arrow/arrow/arrow.py", line 622, in replace
    raise AttributeError('unknown attribute: "{}"'.format(key))
AttributeError: unknown attribute: "days"
>>> arw.shift(days=-1)
<Arrow [2020-01-02T12:41:19.470793+00:00]>
```

systemcatch: Not sure what version of arrow is installed on your machine Alex but you can update to the latest dependencies with `pip install -r requirements.txt` from the parser directory.

alixunderplatz: Thanks @systemcatch, I upgraded my arrow and fixed it! ;)

systemcatch:
I'll merge this in and we'll keep an eye on data availability, hopefully the delay is temporary.

alixunderplatz:
@systemcatch update: 😉 
I just retried using "03/01/2020" and "04/01/2020" as start and enddate, this is what the website returns:
image
Data for 3rd of January up until last hour. So we should rather find a way to check for 
start "today-1" and end "today"
and also
start "today" and end "today + 1" and merge the two datasets when both are returned.

This way we will get the most recent stuff. This could be why the parser failed in the past for a couple of hours each day, from what I can remember.

PR #2142, opened 2020-01-07
Added historic data-grabbing capability to US_SPP parser

robertahunt:


parsers/US_SPP.py
```diff
@@ -66,14 +84,14 @@ def data_processor(df, logger):
...
     processed_data = []
     for index, row in df.iterrows():
         production = row.to_dict()
 
-        extra_unknowns = sum([production[k] for k in unknown_keys])
-        production['unknown'] = production['Other'] + production['Waste Heat'] + extra_unknowns
+        production['unknown'] = sum([production[k] for k in unknown_keys])
 
-        dt_aware = parser.parse(production['GMT MKT Interval'])
+        dt_aware = production['GMT MKT Interval']
```
corradio: are we sure that the datetime returned in the file object is of type `datetime`?

systemcatch: Yeah dateutil's `parser.parse()` will return a `datetime` object here.

parsers/US_SPP.py
```diff
@@ -52,7 +55,22 @@ def data_processor(df, logger):
...
+    #So first we need to combine those.
+    for col in df.columns:
+        if 'Market' in col:
+            combined_col = col.replace('Market','').strip()
+            self_col = col.replace('Market','Self')
+            if self_col in df.columns:
+                df[combined_col] = df[col] + df[self_col]
+                df.drop(self_col, inplace=True, axis=1)
+            else:
+                logger.warning(f'Corresponding column "{self_col}" to "{col}" not found in file', extra={'key':'US-SPP'})
```
systemcatch: I think the new log message is causing https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/test/test_US_SPP.py#L63 to break.

parsers/US_SPP.py
robertahunt: Ah yes, for the historical data I parsed that column earlier, so it is a timestamp, but I did not do that for the live data. I'll change that.

On a related note, for the downstream code, does the format have to be of type datetime.datetime? Or can it be a pandas timestamp? Because that may cause problems.

corradio: Unless pandas timestamp can be cast to datetime.datetime I think it needs to be datetime.datetime. The reason is that it gets inserted into the database.

robertahunt: This could be problematic in general since pandas wants all timestamps to be pandas timestamps, and it is more efficient to use parsed timestamps for filtering data. This is something we might need to think about finding a robust solution to in the future (unless there is one I am not seeing). In the meantime I will make it all datetimes without hopefully being too messy. :)

corradio: sounds good -  I suggest you add this to our internal issue tracker?

parsers/US_SPP.py
robertahunt: Hey systemcatch! That is strange indeed. Are you referring to when running "python test_parser.py US-SPP", because I do not get an error when running those tests. Could you give me some more details on how you would suggest fixing it? Thanks for checking on this!

robertahunt: I understand now! Nevermind, I am looking into it, you are probably right that the logger would cause an issue.

PR #2143, opened 2020-01-07
UA capacity update

amkozlov:
Update Ukraine capacity as of 11/2019 according to
https://ua.energy/vstanovlena-potuzhnist-energosystemy-ukrayiny/

- update renewables
- split hydro (ГЕС) and hydro_storage (ГАЕС)

Coal/gas/oil could be potentially extracted, but this requires more work:
- data is not reported aggregated by fuel type
- some power plants can run on multiple fuel types (coal/gas/oil), a pragmatic workaround would be to double-count respective capacities


corradio:
Thanks for the contribution @amkozlov !
Are you able to update the README as well so we keep track of the source?

amkozlov:
sure, here you go!

PR #2145, opened 2020-01-08
Migrate ZoneList component to React

fbarl:

#### Notes

* When testing locally and switching the toggle to `production`, the zone list will become empty as there is no production data is generated for the local runs. This should actually be a correct behavior as the zone list should presumably only display zones for which some data exists. In the _old_ `ZoneList` component, the list wouldn't really get updated when switching the toggle which is probably a bug.


web/src/helpers/scales.js
```diff
@@ -41,17 +42,24 @@ const solarRange = d3.range(10).map((i) => {
...
+export const solarColor = d3.scaleLinear()
   .domain(solarDomain)
   .range(solarRange)
   .clamp(true);
 
 // ** CO2
-const maxCo2 = 800;
+export const maxCo2 = 800;
+
+export const getCo2Scale = (colorBlindModeEnabled) => {
```
corradio: 👍

fbarl:
I just ran through all the interactions tests I could think of and it seems to work fine:

#### Web

* [x] Rendering of the zone list
* [x] Clicking on the zones
* [x] Keyboard interactions (up/down/enter/a-z)
* [x] Search bar filtering
* [x] Clicking on the map selects the zone

#### Mobile

* [x] Rendering of the zone list
* [x] Clicking on the zones
* [x] Search bar filtering
* [x] Clicking on the map selects the zone

The only bit of code I wanted to be careful about is the map centering event observer I added in this PR, but the map interactions seem to work fine as well!

```javascript
observe(state => state.application.selectedZoneName, (selectedZoneName, state) => {
  if (selectedZoneName) {
    centerOnZoneName(state, selectedZoneName, 4);
  }
});
```

**Note:** For mobile testing, I actually used Chrome in responsive mode as I couldn't get `ngrok http 8000` to work from my mobile due to too many connections error :/ I wonder how you normally test from your phone, couldn't find anything in the docs :thinking:

corradio:
That's perfect!

PR #2147, opened 2020-01-09
More localisation fixes + correct link for adding new regions doc

fbarl:
Inspired by #2146 and a similar bug I came across (see the screenshot), I went through the localisations that contained some HTML and wrapped them with `dangerouslySetInnerHTML` spans if they weren't made so previously (most of such localisations are presented in FAQ and they were already handled well).

Localisations fixed:

* `country-panel.noParserInfo`
* `tooltips.noParserInfo`
* `misc.newversion`

image

I also updated the links pointing to adding new regions doc to a correct header link https://github.com/tmrowco/electricitymap-contrib#adding-a-new-region.

### DIFF
--- a/parsers/README.md
+++ b/parsers/README.md
@@ -1,4 +1,4 @@
 # Parsers
 A parser is a file responsible for fetching data about the power system in a given region.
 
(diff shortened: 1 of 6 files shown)


corradio:
You rock!

PR #2148, opened 2020-01-10
Migrate AreaGraph component to React

fbarl:

#### Tested

* [x] AreaGraph renders well on small screens (`360px`) as well as desktop
  * [x] Consumption graphs render well
  * [x] Production graphs render well
  * [x] Axes render well
  * [x] Finland (empty graph)
  * [x] Southeast Norway (a lot of exchanges with the rest of Norway)
  * [x] Latvia (sharp graph)
  * [x] Germany

#### Notes

* I made the `AreaGraph` layers also have `cursor: pointer` like their background interaction rectangle and the whole of `LineGraph` already had - this was probably a tiny bug.
* I discovered one more subtle bug with componenents not always getting the freshest `selectedZoneTimeIndex` - turns out `UPDATE_SLIDER_SELECTED_ZONE_TIME` wasn't deep-copying the state so I prepended `{}` to every `Object.assign` (they were just missing in the reducers).
* I noticed that `AreaGraph` tooltip is not always ideally positioned when hovering on a small screen, which causes it to flash constantly. ~This was already true in prod https://www.electricitymap.org/ so I didn't try to fix it here. Presumably small screen means mobile mode so there should really be no hovering tooltips.~ I fixed that issue in https://github.com/tmrowco/electricitymap-contrib/pull/2148/commits/661ac357afc935cc8f4746fcfd929e5df0f69299 and I also centered tooltips to the top of the screen in mobile mode for readability.
* There seems to be no color-blind mode spectrum for `AreaGraph` - I opened #2158 to address it separately.


web/src/reducers/index.js
```diff
@@ -37,6 +37,8 @@ const initialApplicationState = {
   locale: window.locale,
   onboardingSeen: cookieGetBool('onboardingSeen', false),
   tooltipDisplayMode: null,
+  tooltipPosition: null,
+  tooltipCountryData: null,
```
corradio:
```suggestion
  tooltipZoneData: null,
```

corradio:
About the tooltips: Are you able to fix the issue on mobile? People actually use touch to simulate a hover and see the tooltip.

web/src/components/areagraph.js
```diff
@@ -1,466 +1,278 @@
...
+  id,
+  selectedIndex,
+  layerMouseMoveHandler,
+  layerMouseOutHandler,
+  mouseMoveHandler,
+  mouseOutHandler,
+}) => {
+  // Hack to create a reference via non-mutable state so that it's created
+  // only once at the initialization (like in a class constructor).
+  const [ref] = useState(React.createRef());
```
corradio: can't we use https://reactjs.org/docs/hooks-reference.html#useref?

web/src/components/areagraph.js
```diff
@@ -1,466 +1,278 @@
...
+  electricityMixMode: state.application.electricityMixMode,
+  selectedIndex: state.application.selectedZoneTimeIndex,
+});
+
+const AreaGraph = ({
+  colorBlindModeEnabled,
+  currentTime,
+  data,
+  displayByEmissions,
+  electricityMixMode,
```
corradio: `electricityMixMode` and `displayByEmissions` are two variables that in principle should be taken out of `AreaGraph` if we want the graph to be re-usable.
I was thinking, would it make sense to try to move those two out of the `AreaGraph`, and the replace the `data` prop with something less general that is agnostic of "exchanges", "countries" and "carbonintensity"?
Basically we would move `getGraphState` outside and pass directly `area` etc.. instead of `data`?

Let me know what you think. I think it could be worth it to clean up the code even more. Maybe we could do it in another PR?

web/src/components/graph/axis.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
corradio: well done

web/src/components/graph/exchangelineargradients.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
corradio: I was confused by the name but maybe we should call this exchangeseriesgradient? (to point out that it's the whole series and not just a list of exchanges)

web/src/components/areagraph.js
fbarl: Yes of course, good catch, not sure why it didn't cross my mind :D

web/src/components/graph/exchangelineargradients.js
fbarl: Sure, makes sense, the only thing is I'd like to keep it plural since it actually renders multiple gradients, so `ExchangeSeriesGradients`.

fbarl:
> About the tooltips: Are you able to fix the issue on mobile?

Yup, I managed to fix it relatively painlessly in https://github.com/tmrowco/electricitymap-contrib/pull/2148/commits/661ac357afc935cc8f4746fcfd929e5df0f69299, but please note that this code hangs on some loose ends (the specificity of `onClick` hooks for mobile vs `onMouseMove` etc..), which will be improved when we move `Tooltip` component to React and are able to embed them into the graph components.

web/src/components/areagraph.js
fbarl: Yeah, that's a good point!

I was looking into it and it looks like a couple of hours work to make the component fully generic, so I thought we might actually want to defer it to a later time - I opened #2163 to capture the problem and suggest a solution.

From a technical point it's definitely something I'd like to get around to, but maybe it shouldn't block the rest of the React transition as its motivation is somewhat orthogonal to this effort.

If you'd like to see it addressed soon, I'd be happy to play with #2163 in one of the next few days.

WDYT?

fbarl:
@corradio PTAL :)

web/src/components/areagraph.js
corradio: I think it'd be worth addressing soon while it's fresh in your mind.

fbarl: This comment has been addressed by https://github.com/tmrowco/electricitymap-contrib/pull/2171.

PR #2153, opened 2020-01-14
Update wind capacity for Norway

Gazer75:
More wind farms operational...


systemcatch:
Hi @Gazer75 is this data coming from ENTSOE or another source?

Gazer75:
I get this manually from NVE.no at their wind databse.

Issue #2166, opened 2020-01-16
US-SPP historical parser captures too much as 'Unknown'


robertahunt:
Because SPP named their columns 'Natural Gas Market' and 'Gas Self' instead of 'Natural Gas Self', the 'Gas Self' column is mistakenly added to Unknown.

fbarl:
@corradio @robertahunt can we close this now that #2168 has been merged?

robertahunt:
Yes!

PR #2168, opened 2020-01-16
Fixed US-SPP mapping of 'Gas Self' to Unknown

robertahunt:


corradio:
Just in case you didn't know, you can write "Fixes #2166" which will automatically close that issue once merged.

PR #2171, opened 2020-01-16
Generalize AreaGraph component

### LINKED ISSUE #2163: Generalize AreaGraph component
Motivated by https://github.com/tmrowco/electricitymap-contrib/pull/2148#discussion_r366791927.

`AreaGraph` component in its current version makes a lot of implicit assumptions about the structure of the data it shows and contains a lot of logic specific to countries carbon production - this is currently not necessarily a problem as `AreaGraph` is currently used in only one place, but if we'd want to reuse it in other contexts, we'd first need to pull out some non-generic logic and make it more robust.

I suggest creating a new component `CountryHistoryMixGraph` to wrap around `AreaGraph` and do all the specific calculations, transforming `AreaGraph` into a _pure_ component with no dependencies on the Redux state.

fbarl:
Resolves #2163.

#### Tested

* [x] AreaGraph renders well on small screens (`360px`) as well as desktop
  * [x] Consumption graphs render well
  * [x] Production graphs render well
  * [x] Axes render well
  * [x] Belarus (empty graph)
  * [x] Southeast Norway (a lot of exchanges with the rest of Norway)
  * [x] Latvia (sharp graph)
  * [x] Germany


web/src/helpers/redux.js
```diff
@@ -1,3 +1,14 @@
+import {
+  flatMap,
+  keys,
+  sortBy,
+  uniq,
+} from 'lodash';
+
+export function getExchangeKeys(zoneHistory) {
```
corradio: is this really a redux helper?

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  mouseOutHandler,
+  layerMouseMoveHandler,
+  layerMouseOutHandler,
+  selectedTimeIndex,
+  selectedLayerIndex,
+  gradientIdPrefix,
+  gradientLayers,
+  gradientStopColorSelector,
+  isMobile,
+}) => {
```
corradio: We plan to re-use this component quite a lot.
Therefore, I'd like to spend a bit of time here making sure all props are well defined.
I'll add some comments on each prop

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+const getTimeScale = (containerWidth, datetimes, currentTime) => scaleTime()
+  .domain([first(datetimes), currentTime ? moment(currentTime).toDate() : last(datetimes)])
+  .range([0, containerWidth]);
+
+const getValueScale = (containerHeight, maxTotalValue) => scaleLinear()
+  .domain([0, maxTotalValue * 1.1])
+  .range([containerHeight, Y_AXIS_PADDING]);
+
+const AreaGraph = React.memo(({
+  id,
```
corradio: why is this needed?

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  layers,
+  currentTime,
```
corradio: is this a `Date` object? If it is, maybe we should call it currentDate?

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  maxTotalValue,
```
corradio: can this be calculated / memoized?

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  .domain([first(datetimes), currentTime ? moment(currentTime).toDate() : last(datetimes)])
+  .range([0, containerWidth]);
+
+const getValueScale = (containerHeight, maxTotalValue) => scaleLinear()
+  .domain([0, maxTotalValue * 1.1])
+  .range([containerHeight, Y_AXIS_PADDING]);
+
+const AreaGraph = React.memo(({
+  id,
+  layers,
```
corradio: what is the schema here? Maybe a comment would help

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  currentTime,
+  maxTotalValue,
+  valueAxisLabel,
+  datetimes,
+  fillSelector,
```
corradio: what is this?

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,132 @@
...
+  mouseMoveHandler,
+  mouseOutHandler,
+  layerMouseMoveHandler,
+  layerMouseOutHandler,
+  selectedTimeIndex,
+  selectedLayerIndex,
+  gradientIdPrefix,
```
corradio: is there a way we can rid of id/selector for the gradient?

web/src/components/graph/areagraph.js
fbarl: There was this chunk in the styles:

```scss
#country-history-mix {
    height: 10rem;

    .layer path {
        opacity: 1.0

        &:hover {
            opacity: 0.8;
        }
    }
}
```

However it seems that the opacity effects weren't really working and the height is something that can probably be fixed for now (please let me know if you think there would be value in passing it as a prop instead) - updated in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/23c405160f733db0116c0583e52ff3c9a32247b8.

web/src/components/graph/areagraph.js
fbarl: I must confess I don't fully understand which values `currentTime` can assume and how it's being used, but to answer your questions, it looks like a `DateTime` object and it _can't_ be inferred from `selectedTimeIndex`.

[This](https://github.com/tmrowco/electricitymap-contrib/pull/2171/files#diff-d6c919ee9d0f0d9969d62718df324219R125) is its definition currently in use:

```
state.application.customDate || (state.data.grid || {}).datetime
```

So overall it seems there are currently 3 ways the end of the `AreaGraph` time scale interval can be set (in order):

1. By directly setting a `customDate` value
2. From `data.grid.datetime`
3. From the last timestamp from the data API response

I'm not really sure what the use cases for (1) and (2) are - if you think they are just legacy, then we can get rid of `currentTime` as a prop altogether and use a max between (3) and the actual current time for setting the time scale end.

web/src/components/graph/areagraph.js
fbarl: It was a function for selecting layer fill color based on layer index. I simplified this in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/edae3eccca2dc79a1d3ad013cea9dcb6c6fcb498 so that the fill colors are attached to the `layers` prop instead of being passed around as a separate prop.

web/src/components/graph/areagraph.js
fbarl: You're right, it was very easy to infer this info from the `layers` props :) See https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/cbb72f14063b532a73d0819d6ec973a89c26492a.

fbarl: By the way, similarly it was easy to infer `datetimes` from the `layers` prop so I did it in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/231759db2ba696ec053ac384f672e2aeb305bf23.

web/src/components/graph/areagraph.js
fbarl: Yes, that's a good point! I appended it to `layers` props instead of passing it through these 3 separate props so now gradients are conditionally being rendered together with their corresponding layers and their usage is a bit more robust - see https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/4a35f3d3093e400d3e7e9c895b4eb93903104ea7.

web/src/helpers/redux.js
fbarl: Good catch! I wasn't very sure where to put it though, created a new helpers file `zones.js` and moved it there in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/26885cd6d1f8aff209b3303aedcc919017942b81, hope that's better :)

web/src/components/graph/areagraph.js
fbarl: I agree! Added in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/58e38090e06d6c21940f48df3e0931495145a408.

fbarl:
Thanks for the review and the comments @corradio, they were spot on!

I tried to save some time working on this component and not force bringing everything in order before the need would occur, but as you pointed out this is a complex component which we already know is going to be reused, so the extra effort is probably worth it :)

Please take another look - all the redundant / inferable props have been stripped now with the exception of `currentTime` for which I wanted to see your answer first.

web/src/components/graph/areagraph.js
corradio: Actually it seems like `currentTime` is used here:
https://github.com/tmrowco/electricitymap-contrib/pull/2171/files#diff-e666df2f408c2db2102da3e7af69650aR28-R30
It's used to set the scale. The reasoning is this:
If the history data hasn't been updated in a while, the last datapoint will have a datetime that is much older than now(). Therefore, we pass the `currentTime` (which represents now()) in order to force the scale to go up to now().

There's two things we can do:
1. remove the prop and that logic in AreaGraph
2. rename the prop to make it less confusing (something like maxDatetime could be better)

I suggest going with 2. and make it optional.

web/src/components/graph/areagraph.js
```diff
@@ -0,0 +1,141 @@
...
+const getValueScale = (containerHeight, maxTotalValue) => scaleLinear()
+  .domain([0, maxTotalValue * 1.1])
+  .range([containerHeight, Y_AXIS_PADDING]);
+
+const AreaGraph = React.memo(({
+  /*
+    `layers` should be an array of objects containing the following fields:
+      * `key` - unique ID of the layer
+      * `fill` - fill color of the layer
+      * `datapoints` - represents the graph datapoints + metadata for the layer in D3 stack() format
```
corradio: Can we simplify even more and make sure the d3 stack step is done inside areaGraph?
This would make this component much more easy to use (and re-usable)

I'm thinking:
- data: array of datapoint objects. Each object represents a timestamp and must have a `datetime` key.
- layerAccessor: value or accessor function that returns a layer object from a data object.
- layerKeys: array of layer keys that should be present in layerAccessor. The order of this array defines the rendering order.
- layerFill: value or accessor function (e.g. key => some_color_mapping[d.key])
- layerHorizontalGradientFill: (e.g. datapoint => some_color_mapping[datapoint.cabonIntensity]) [do we need a gradient key here?]

Feel free to modify this based in order to optimize for performance (I don't remember what d3 stack requires)

web/src/components/graph/areagraph.js
fbarl: Indeed, so what I was suggesting above is we go with (1) but instead I went for a bit more robust version of (2) in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/2f4c419dbc9d66fe5f756cc9d5bef185f395056f.

Actually, the way I currently understand it, the same reasoning you provided could be applied to the other end of the graph to argue we need `startTime` to show gaps when data is missing at the beginning of graph, so I added support for it, but didn't fully transition as the other graphs are not using it yet and for the UI consistency, they should probably be switched at the same time.

I opened https://github.com/tmrowco/electricitymap-contrib/issues/2250 to capture the essence of the problem - please let me know if that makes sense.

corradio: That makes sense!

web/src/components/graph/areagraph.js
fbarl: That's a good point, I'm glad you brought it up as it slipped my radar somehow under all these iterations of `AreaGraph` :)

I addressed it in https://github.com/tmrowco/electricitymap-contrib/pull/2171/commits/1a9c53ec2ba09a37c9280b59e36a244424de291c by:

- using `data` and `layerKeys` as you suggested
- using `layerFill` which maps layer keys to either a fixed color or a (datapoint -> color) map which then automatically creates a horizontal gradient
- not using `layerAccessor` and `layerHorizontalGradientFill` as they are obsolete with regards to the above

This change shouldn't have any effect on the performance. Please let me know what you think about the current props and whether you see any other potential blockers for merging the PR.

fbarl:
\o/

> Would be awesome to have the same graph props for the line graph

Yes, that would be very easy to do now - however I was thinking how in the current two places where `LineGraph` is used, it's not really a _line_ graph but rather an _area_ graph with just one layer - so the only difference between the current `AreaGraph` and `LineGraph` is in the number of layers.

That made me think about merging the two together into one `Graph` component which we could achieve with some tiny modifications of the current version of `AreaGraph` (we could always extend the component by adding a boolean `stacked` / `area` if we ever wanted to support real line graphs) - wdyt?

corradio:
It makes *so* much sense. We could maybe have LineGraph as a thin wrapper over AreaGraph to simplify things the props?

fbarl:
> We could maybe have LineGraph as a thin wrapper over AreaGraph to simplify things the props?

Yup, I imagine we would probably want to do something like that, I'll see how it looks like when I get near that point :)

PR #2174, opened 2020-01-19
Improve Dutch localization

joppiesaus:
Lots of more improvements can be made, but this should improve the translation for now.
Great website by the way :+1:


corradio:
thx!

PR #2188, opened 2020-01-29
Fixed format of timestamp for fetching historic EIA data

robertahunt:
Previously entering a historic time would not cause an error, but would just return the last 24 hours of data instead. I think this is a problem on the eiapy side only recognizing timestamps of the format YYYYMMDDTHHZ, instead of all ISO 8601 formats.

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -11,7 +11,7 @@ import datetime
 import os
 
 import arrow
-from dateutil import parser
+from dateutil import parser, tz
 os.environ.setdefault('EIA_KEY', 'eia_key')
 from eiapy import Series
 import requests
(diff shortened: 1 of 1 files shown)


systemcatch:
@robertahunt that's interesting since I wrote the eiapy package.

I'd expect the EIA api to return an error if the format isn't `YYYYMMDDTHHZ` (the only one they accept), but the api was pretty janky last time I used it.

robertahunt:
Ah, okay. (also cool that you wrote it!!) Makes sense that it would be on the EIA side since the EIA API is still in beta. Would you prefer to handle this in eiapy instead then? or should we keep this conversion in the electricitymap contrib? 

There is some urgency here as we need to scrape some historic US data for a client within the next week.

systemcatch:
Probably better to handle this in electricitymap imo so things get moving faster.

robertahunt:
thanks!

Issue #2183, opened 2020-01-23
US-CA going offline overnight


typicalTYLER:
It seems like similar to the [US-TX issue](https://github.com/tmrowco/electricitymap-contrib/issues/2178) I posted recently, US-CA is going offline overnight as well, seemingly due to the same problem, negative solar overnight. Here is a [log](https://kibana.electricitymap.org/app/kibana#/doc/96f67170-0c49-11e9-85c1-1d63df8c862c/logstash-2020.01.23/doc?id=wyBc028BX-fmul22TGRZ&_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))) of one of the errors. Maybe this should be clamped and throw a warning as well?

robertahunt:
Hey! Yes, that makes sense. Is this something you want to/are able to take on fixing? Do you need anything from us?

typicalTYLER:
Just took a stab at it! Don't have an environment set up though so wasn't able to test, hopefully CI would catch any problems though

PR #2190, opened 2020-01-29
Added solar threshold so negative values are capped to 0

robertahunt:
Not sure if this is the right place to add this threshold, however, US-TN values from EIA apparently are negative quite often (basically every night) and thus the output is failing our quality checks and then no data can be imported into the database. It seems like this is a common issue for US data (see #2183 and #2178 ). This would cap any negative solar values from EIA to 0. 

I say this might not be the right place for this because perhaps we should make this a blanket measure for all parsers? Also does anyone know why these values are negative? perhaps it is to reflect solar-battery setups that charge in the night?

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -92,6 +92,9 @@ def fetch_production_mix(zone_key, session=None, target_datetime=None, logger=No
         if not mix:
             continue
         for point in mix:
+            if type == 'solar' and point['value'] < 0:
+                point['value'] = 0
+
             point.update({
                 'production': {type: point.pop('value')},
                 'storage': {},  # required by merge_production_outputs()
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -16,6 +16,7 @@ os.environ.setdefault('EIA_KEY', 'eia_key')
 from eiapy import Series
 import requests
 
+from .lib.validation import validate
 from .ENTSOE import merge_production_outputs
 
 EXCHANGES = {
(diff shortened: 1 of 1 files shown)


corradio:
Hi @robertahunt ,

Check the `validate` method in `validation.py`. We've made some utils function that allow to that but also enable us to give a minimum value (such that a very negative value causes an error)

robertahunt:
Sorry, but now I think I am confusing myself even more. It looks like 'validation.py' is no longer used, in favor of 'quality.py'. Is that correct? 

It also looks like you could pass validate in validation.py an argument (remove_negative) to tell it to set small negative values to 0. I can add this kind of thing to quality.py for all solar production, I just want to confirm that this is something we want/are okay with for small negative values for all parsers.

corradio:
We shoul probably refactor this to make it absolutely clear, but validation.py is used several places to validate each parser separately using custom rules , for example by the ENTSOE.py parser.
quality.py is used in our backend to validate all data data that comes in with a fixed set of rules.
I think quality.py should be moved outside of contrib (a quick search shows it isn't used here).

corradio:
To answer https://github.com/tmrowco/electricitymap-contrib/pull/2190#issuecomment-579931340 more precisely, I think we'd want to add the solar cap only to the EIA parser and not to all parsers by default as the threshold might vary per parser.

robertahunt:
Just a note on this, from the EIA website:
"Negative generation denotes that electric power consumed for plant use exceeds gross generation."
from: https://www.eia.gov/electricity/monthly/epm_table_grapher.php?t=epmt_1_07_a 

We may want to consider that negative generation is not always an error, especially as solar battery installations become more prevalent.

alixunderplatz:
Hi @robertahunt 
I had a look at the solar data for Tennessee. Mostly, it seems that negative values appear around sunrise and sunset. Sometimes they are pretty random during the middle of the night.
Here's what I think: It depends on the type of solar inverter, but from my experience, most of them tend to have a "deep sleep" mode during the night and then switch to "stand-by" mode as soon as it gets bright (DC-voltage rises), but there is not enough irradiation yet. Some inverters tend to start/stop feeding in at irradiation levels like ~20W/m², so below that threshold, they don't generate electricity on the AC-side, but their electronics consume a tiny bit of electricity during that period, which might be registered and reported by some of the solar-electricity meters. This would also underline the "negative gross generation" mentioned in your last comment.
Setting it to 0 was the way to go :) 👍

PR #2199, opened 2020-02-05
Rename "FYROM Rep. of Macedonia" to "North Macedonia"

tomi44g:
Country "MK" has changed name to "North Macedonia" effective Feb 2019. This has not been reflected in i18n files so I used Wiki as a resource to update all languages.
[Prespa Agreement](https://en.wikipedia.org/wiki/Prespa_agreement)


web/locales/zh-hk.json
```diff
@@ -254,7 +254,7 @@
             "zoneName": "黑山共和國"
           },
           "MK": {
-            "zoneName": "馬其頓共和國"
+            "zoneName": "北马其顿"
```
jarek: I'm not entirely sure, but it looks like the Traditional Chinese script name (as I guess would be used for zh-hk and zh-tw below) is 北馬其頓. Would you mind updating it in those two files?

```suggestion
            "zoneName": "北馬其頓"
```

web/locales/zh-tw.json
```diff
@@ -255,7 +255,7 @@
             "zoneName": "黑山共和國"
           },
           "MK": {
-            "zoneName": "馬其頓共和國"
+            "zoneName": "北马其顿"
```
jarek:
```suggestion
            "zoneName": "北馬其頓"
```

PR #2200, opened 2020-02-09
Fix AUS-WA for change in source data format

con-cat:
The AUS-WA parser is currently down with this error:
```
ValueError: Length mismatch: Expected axis has 4 elements, new values have 2 elements
```
It looks like there has been a change in the source data - line 69 was returning columns for latitude and longitude as well as the production and capacity data.

This PR fixes that issue, as well as replacing the [deprecated](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#ix-indexer-is-deprecated) `.ix` indexer with `.log`


robertahunt:
Thanks @con-cat

PR #2207, opened 2020-02-13
Update exchanges with Italy

### LINKED ISSUE #2131: Interconnectors IT-GR and IT-ME
The two interconnectors are available on the ENTSO-E transparency portal, but not on the map.

The IT-GR one, unlike the rest of the Greece data, seems to have a <2h delay, so it should be realtime enough, while the recently added IT-ME one started to return data just today [2019-12-28] (for now it sits at '0', but before it was N/A or n/e), so it may be operative soon.

The IT-ME interconnector, if added, should have a capacity of 600MW on both directions, as per https://www.terna.it/en/media/news-events/italy-montenegro-invisible-electrical-bridge

alixunderplatz:
Hi all,
1. The GR<->IT-SO exchange probably wasn't working due to a change in the domain names on ENTSOE (switched from "Italy-Brindisi" to "Italy South" after the Brindisi zone was integrated in "Italy South"). I changed it to the respective zone in ENTSOE.py.
2. The Montenegro<->Italy interconnector ("MONITA") was inaugurated in late December 2019.
I added the exchange override in ENTSOE.py, as well as the params in exchanges.json.

I tried to fix both, but couldn't test locally due to no ENTSOE API key, so please test before merging^^ :)

Fixes #2131 if everything is fine here


config/exchanges.json
```diff
@@ -1479,6 +1479,21 @@
     },
     "rotation": 90
   },
+  "IT-CSO->IT-ME": {
```
robertahunt:
```suggestion
  "IT-CSO->ME": {
```

PR #2214, opened 2020-02-16
Hungary (HU) Capacity Update

nessie2013:
also added Geothermal and Unknown data for Hungary


corradio:

PR #2215, opened 2020-02-16
Serbia (RS) Capacity Updates

nessie2013:
Updated and added Unknown and Solar Capacity


corradio:

PR #2221, opened 2020-02-18
Upd. Solar/Wind installed capacity for Portugal

Falconet8:
Per the 2018 IRENA data, Portugal installed solar capacity was 670 MW, up from the 460 MW currently listed.  Wind up from 5.03 GW to 5.191 GW, per the same source.


robertahunt:
Sorry about the late reply on this. Thanks for the changes @Falconet8. Could you also update the source in the readme.md? https://github.com/tmrowco/electricitymap-contrib/blob/master/README.md

Falconet8:
Hello Roberta,

I should have clarified. The source IRENA is already present on the "Production capacity data sources"

PR #2232, opened 2020-02-21
Updated SE Capacity

nessie2013:
(no source change needed as data is from original source)


robertahunt:
@nessie2013 how come this one was closed? I don't see were SE capacity was updated otherwise?
If you are going to reopen this though, I am curious why 'gas' capacity was removed.

PR #2234, opened 2020-02-21
Updated US-CA Capacity

nessie2013:
(no source update needed)


alixunderplatz:
@robertahunt @nessie2013 biomass and geothermal down?

robertahunt:
Good catch... Capacities match http://www.caiso.com/informed/Pages/CleanGrid/default.aspx and previous capacities were from https://ww2.energy.ca.gov/almanac/electricity_data/electric_generation_capacity.html which only has data up to 2018. 

It could be the case that CA-ISO does not cover all of the CAL balancing authority though. How would you handle these?

PR #2238, opened 2020-02-21
US-BPA Capacity Update

nessie2013:
No source update required


robertahunt:
Hey @nessie2013 , could you confirm that you got the data from https://transmission.bpa.gov/business/operations/Wind/baltwg.aspx? I don't see where they list solar capacity there.

alixunderplatz:
biomass and wind capacity lower, what's the reason?

robertahunt:
@alixunderplatz Seems like some wind farms changed balancing authorities in 2018: https://transmission.bpa.gov/business/operations/wind/WIND_InstalledCapacity_PLOT.pdf
I would assume the same for biomass (at least the numbers here match what BPA reports their capacity to be).

robertahunt:
For the records: https://transmission.bpa.gov/business/operations/wind/WIND_InstalledCapacity_LIST.pdf looks like they spun some of their wind farms off into the AVRN balancing authority

nessie2013:
Yup I don't know where that solar came from, was a mistake so I removed it

PR #2254, opened 2020-02-27
Migrate other graphs to React + deprecate LineGraph component

fbarl:

#### Tested

* [x] Graphs render well on small screens (`360px`) as well as desktop
  * [x] Axes render well
  * [x] Latvia (gradients working)
  * [x] Germany (half empty graphs)
  * [x] Israel (flat + empty graphs)

#### Changes

* Deprecated `LineGraph` component in favor of the extended `AreaGraph` component
* Made `CountryHistoryCarbonGraph` and `CountryHistoryPricesGraph` into no-prop wrappers around `AreaGraph`
* Made tooltips logic in `main.js` a bit more explicit
* Put a lot of shared logic in a new `history.js` helper file
* New props added to `AreaGraph`:
  * `layerStroke` - the prices graph had a light gray stroke, thought we might want to keep it
  * `focusFill` - the prices graph has a gray fill but the hover focus circle has a shade of orange as a fill, thought we might want to keep it
  * `height` - as the heights of all 3 graphs are currently different


web/src/components/countryhistorypricesgraph.js
```diff
@@ -0,0 +1,129 @@
...
+    [setSelectedLayerIndex]
+  );
+
+  return (
+    <AreaGraph
+      data={data}
+      layerKeys={layerKeys}
+      layerStroke={layerStroke}
+      layerFill={layerFill}
+      focusFill={focusFill}
```
corradio: what about calling this markerFill?

fbarl: I like that, it does sound more descriptive :) Addressed in https://github.com/tmrowco/electricitymap-contrib/pull/2254/commits/548ef41a2d9a2ec5447c38a057f5e1dc880d6f28.

PR #2256, opened 2020-03-01
Add parser for Maharashtra, India 

### LINKED ISSUE #1443: Add Maharashtra (India) IN-MH
Not sure if any of you OCR-experts already had a go at it in the past, but here is another challenge still waiting for you ;)
I tried to elaborate a bit more on what needs to be included and where to find this info in the images.
Feel free to comment below and reach out on Slack should you have any questions related to this. 😸 
_ _ _ _

This one may have gone lost a bit in the large #304 "Add India" (and #817 OCR) thread and could cover yet another large state on the west coast of India. So I decided to open a seperate issue on this and see what we can do here for **Maharashtra**.
_ _ _ _

On the [website of Maharashtra State Load Despatch Center](https://mahasldc.in/) there is a tab named "SCADA Data".

First off, there seems to be a **single image containing all of the relevant generation** for all types:
I played around with the URLs of the various images, and changing the end of the URL to "3.jpg" leads to this major dashboard:
https://mahasldc.in/wp-content/reports/sldc/mvrreport3.jpg

* **coal:** "THERMAL" in the lower left part of the image is for generation from coal. There is data for net generation ("behind bus") and gross generation, I suggest using the latter one to cover self-consumption related emissions (~+10%).
* **gas:** Below thermal, there is "GAS". Don't get distracted by the "Uran" label, it's not related to nuclear but just a city name of the unit's location south of Mumbai.
* **hydro:** The next column slightly left of the center has a field summarizing total "hydro" output.
  - Note: Within the hydro units, there is ["Ghatghar", which is a **pumped storage hydro** plant](https://en.wikipedia.org/wiki/Ghatghar_Dam)!!! If we decide to put this in its respective category, there is another field "Ghatgr pump" for the pumping consumption (I have not yet observed if it's showing any value, it's 0 at the moment.)
So the Ghatghar output could be subtracted from HYDRO and added to hydro storage, in case OCR works reliably enough to perform this.
* **solar, wind, biomass, unknown:** On the bottom of this column, there is a yellow box containing several types of generation.
Obviously, MS WIND is **wind** and MS SOLAR is **solar**. CO-GEN is from **biomass** ([bagasse](https://en.wikipedia.org/wiki/Bagasse)). 
I'd map the output of of "OTHER + SMHYD" entirely to "**unknown**", because small hydro has a tiny share (scroll down to the end for the direct link to that dashboard).
* **nuclear:** There are nuclear units of Tarapur power station TARPR PH-I (oldest operational in the world) and TARPR PH-II, displayed under the "C. S. GENERATION" section on the very right

There are three more crucial things to consider, which makes it really complicated.
1. The city of Mumbai, which is within Maharashtra on the west coast, has a seperate box with figures for generation and demand. It is split to "TPC THM" for thermal, "TPC HYD" for hydro and "REL GEN", probably for "relative share of generation in some power plants outside of Mumbai". 
1.1 TPC THM could be added to coal, because these are the types of plants in Mumbai operated by Tata Power Company (=TPC!). 
1.2 There is no hydro power plant directly within the city of Mumbai. But this figure is probably related to some hydro power plants of the private sector, which also belong to the private Mumbai based Tata Power Company with 310 MW of hydro installed very, very close to Mumbai in three locations. https://npp.gov.in/dashBoard/cp-map-dashboard So I think it's legit to include these hydro figures as well! 
1.3 "REL GEN" could be excluded or just added to unknown.
2. There is another huge chunk of "IPP / CPP" generation, which stands for Intependent and Central power producers. The list contains 9 of such power plants plus "other", that seem to produce for Maharashtra, but are not necessarily located within the state. I tried to locate them and their fuel types:

* JINDAL - [Jindal Tamnar **coal** power plant](https://en.wikipedia.org/wiki/Jindal_Tamnar_Thermal_Power_Plant) in **Chhattisgarh**
* ADANI - [Mundra thermal **coal** power station operated by Adani](https://en.wikipedia.org/wiki/Mundra_Thermal_Power_Station) in **Gujarat** (we've had this in the last Gujarat improvement issue)
* IDEAL ENR - [Bela **coal** power plant owned by ideal](https://www.industryabout.com/country-territories-3/757-india/fossil-fuels-energy/42622-bela-tata-coal-power-plant) in **Maharashtra**
* RATTAN_IND-AMST and _IND-NSK - [RattanIndia **coal** power plants](https://en.wikipedia.org/wiki/RattanIndia) in Amravati and Nasik in **Maharashtra**
* BUTIBORI-REL - [coal power plant](https://en.wikipedia.org/wiki/Butibori_Power_Project) in **Maharashtra**
* WARDHA - [Wardha-Warora coal power plant](https://en.wikipedia.org/wiki/Wardha_Warora_Power_Plant) in **Maharashtra**
* DHARIWAL - [Dhariwal **coal** power station](https://en.wikipedia.org/wiki/Dhariwal_Power_Station) in **Maharashtra**
* PIONEER - [Pioneer **gas** power plant](http://www.pgpl.in/) in **Maharashtra**

3. Central state generation is also displayed on the right side of the board. Some of these ultra super large power plants are in Maharashtra, some are outside of it. Not sure what to do with this data, maybe including all that is within IN-MH is the way to go since they feed in the grid of this state and make up the mix.
Here are the power plants of the central sector and their types + locations:
* KORBAN - [Korba **coal** power station](https://en.wikipedia.org/wiki/Korba_Super_Thermal_Power_Station) in **Chhattisgarh**
* VINCHAL - [Vindhyachal **coal** power station](https://en.wikipedia.org/wiki/Vindhyachal_Thermal_Power_Station) in **Madhya Pradesh**
* GANDHAR - [Jhanor-Ghandar **gas** power station](https://en.wikipedia.org/wiki/Jhanor-Gandhar_Thermal_Power_Station) in **Gujarat**
* KAWAS - [Kawas **coal** power station](https://en.wikipedia.org/wiki/Kawas_Thermal_Power_Station) in **Gujarat**
* KK' PARA - [Kakrapar **nuclear** power station](https://en.wikipedia.org/wiki/Kakrapar_Atomic_Power_Station) in **Gujarat**
* TARPR PH-I and TARPR PH-II - [Tarapur **nuclear**](https://en.wikipedia.org/wiki/Tarapur_Atomic_Power_Station)  in **Maharashtra (see comments above)**
* SIPAT - [Sipat **coal** power station](https://en.wikipedia.org/wiki/Sipat_Thermal_Power_Station) in **Chhattisgarh**
* SSD - [Sardar Sarovar **hydro** dam](https://www.water-technology.net/projects/sardar-sarovar-dam-gujarat/) in **Gujarat** (suffering from drouhgts)
* RGPPL - [Ratnagiri **gas** and power ](https://en.wikipedia.org/wiki/Ratnagiri_Gas_and_Power) in **Maharashtra**
* MAUDA STG-I and STG-II - [Mauda super thermal **coal** power station](https://en.wikipedia.org/wiki/Mauda_Super_Thermal_Power_Station) in **Maharashtra**
* CGPL MUNDRA - [Mundra ultra mega **coal** power plant](https://en.wikipedia.org/wiki/Mundra_Ultra_Mega_Power_Plant) in **Gujarat**
* SOLPRN - [Solapur super thermal **coal** power station](https://en.wikipedia.org/wiki/Solapur_Super_Thermal_Power_Station) in **Maharashtra**

Here is the image of the dashboard with all that info:

image

_ _ _ _
Further details for the types of generation, but not necessarily needed for the parser:

**Station wise generation, exchange and demand** (many, many figures):
https://mahasldc.in/report2/
https://mahasldc.in/wp-content/reports/sldc/report2.jpg

image

**Wind generation data:**
https://mahasldc.in/on-line-re-data/
https://mahasldc.in/wp-content/reports/sldc/mvrreport5.jpg

image

**Solar data:**
https://mahasldc.in/on-line-solar-data/
https://mahasldc.in/wp-content/reports/sldc/mvrreport8.jpg

image

Something called **"[Bagasse data" (= biomass)](https://en.wikipedia.org/wiki/Bagasse)**:
https://mahasldc.in/on-line-co-gen-others-data/
https://mahasldc.in/wp-content/reports/sldc/mvrreport7.jpg

image

And, finally, **"others and small hydro"**:
https://mahasldc.in/on-line-others-small-hydro/
https://mahasldc.in/wp-content/reports/sldc/mvrreport9.jpg

_ _ _ _ 

There are some more dashboards with **power line schemes and line loadings**, but this is wayyyy to complex:

https://mahasldc.in/wp-content/reports/sldc/mvrreport2.jpg
https://mahasldc.in/wp-content/reports/sldc/mvrreport4.jpg
https://mahasldc.in/wp-content/reports/sldc/mvrreport6.jpg

image

image

image

GitMatze:
The parser is based on OCR and requires [tesseract ](https://github.com/tesseract-ocr/tesseract) to be installed on the system (the pytesseract module is just a wrapper). 

Next to the preinstalled english language model a number specific model is used that  can be found here: https://github.com/Shreeshrii/tessdata_shreetest/blob/master/digits_comma.traineddata This model needs to be copied into the tessdata folder in the program directory. 

Resolves #1443


parsers/IN_MH.py
```diff
@@ -0,0 +1,386 @@
+from PIL import Image, ImageOps
+import pytesseract
+import cv2
+from imageio import imread
```
systemcatch: You'll need to add these packages to requirements.txt

parsers/IN_MH.py
```diff
@@ -0,0 +1,386 @@
...
+        ' ': ''
+    }
+    for key, value in chars.items():
+        str1 = str1.replace(key,value)
+        str2 = str2.replace(key,value)
+    return str1.lower() == str2.lower()
+
+
+def fetch_consumption(zone_key='GE', session=None, target_datetime: datetime.datetime = None,
+                     logger: logging.Logger = None):
```
systemcatch: This should be called `fetch_production`, the other keys need to be `target_datetime=None` and `logger=logging.getLogger(__name__)`

parsers/IN_MH.py
```diff
@@ -0,0 +1,386 @@
...
+        line = line + rec_time
+        file.write(line+'\n')
+        print(line)
+        #print(results)
+        plt.savefig('figures/' + filename_time + '.png')
+        #plt.show()
+        time.sleep(60*10)
+        plt.close()
+
+    file.close()
```
systemcatch: We can remove all of this and just have `print(fetch_production())`

parsers/IN_MH.py
GitMatze: I'm not so sure about fetch_production.. The output is of this function is the total state demand, ie consumption. Getting the production numbers is a different story since a lot of plants are outside of Maharashtra, see the list of plants in https://github.com/tmrowco/electricitymap-contrib/issues/1443#issue-330342377

GitMatze:
Sure, here's the test_parser.py output:

>python test_parser.py IN-MH consumption
parser result:
{'zoneKey': 'IN-MH', 'datetime': datetime.datetime(2020, 3, 3, 23, 7, tzinfo=tzfile('Asia/Calcutta')), 'production': {'biomass': 905.0, 'coal': 15914.158032053672, 'gas': 585.7323891166604, 'hydro': 129.0, 'nuclear': 601.1095788296684, 'solar': 67.0, 'wind': 1037.0, 'unknown': 279.0}, 'storage': {}, 'source': 'https://mahasldc.in'}
---------------------
took 9.03s
min returned datetime: 2020-03-03T17:37:00+00:00 UTC
max returned datetime: 2020-03-03T17:37:00+00:00 UTC  -- OK, <2h from now :) (now=2020-03-03T17:38:08.716862+00:00 UTC)

parsers/IN_MH.py
GitMatze: Ok! 
Should I post this code to the corresponding issue then? Without visualization anyone trying to debug this will have a really hard time.

parsers/IN_MH.py
systemcatch: Maybe I'm slightly confused since the returned json contains a production key.

parsers/IN_MH.py
```diff
@@ -0,0 +1,386 @@
...
+
+def fetch_consumption(zone_key='GE', session=None, target_datetime: datetime.datetime = None,
+                     logger: logging.Logger = None):
+
+
+    if target_datetime is not None:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+
+    time = arrow.now('Asia/Kolkata').floor('minute').datetime
```
systemcatch: We don't want to use `time` for the variable name here (python has a builtin time module), how about `maharashtra_time` or just `dt`?

parsers/IN_MH.py
systemcatch: Good idea, you've made me think of an interesting possibility (will open a new issue).

parsers/IN_MH.py
GitMatze: Oh, right, I'll change that!

GitMatze:
The circleci failure seems to be a common problem of pylint and cv2: https://github.com/PyCQA/pylint/issues/2426

systemcatch:
@GitMatze we should change the name of the function from `fetch_consumption` to `fetch_production` (andsame for the `consumption` key). As currently laid out this won't work on the map.

@alixunderplatz any thoughts here?

systemcatch:
We might be able to use https://stackoverflow.com/questions/28829236/is-it-possible-to-ignore-one-single-specific-line-with-pylint/29562085#29562085 to solve this.

Possibly using `# pylint: disable=no-member` on the offending import.

GitMatze:
To keep the code consistent with how I understand the dashboard I'd simply add a fetch_production function that calls fetch_consumption. That might be a bit confusing but it shows that we are actually mistaking production for consumption. What do you think?

systemcatch:
I'm afraid that won't work, the map is setup to use those function names in specific way.

`fetch_consumption` needs to look like https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTSOE.py#L765, `fetch_production` needs to look like https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/PE.py#L29 with those exact keys.

Let's make it a standard `fetch_production` function and add a note how this is likely an oversimplification. We can always improve it in the future.

GitMatze:
Yap, now I see! But it's ok to also add a consumption function to the parser, right? Since that is the data we actually have..

GitMatze:
Now the parser supports both consumption and production. The return data is the same, except for the production/consumption key. I hope this time it's ok!

```
(tesseract) >python test_parser.py IN-MH consumption
parser result:
{'zoneKey': 'IN-MH', 'datetime': datetime.datetime(2020, 3, 7, 12, 51, tzinfo=tzfile('Asia/Calcutta')), 'consumption': {'biomass': 860.0, 'coal': 19714.580069410014, 'gas': 476.504709965295, 'hydro': 547.0, 'nuclear': 732.9152206246902, 'solar': 703.0, 'wind': 122.0, 'unknown': 311.0}, 'storage': {}, 'source': 'https://mahasldc.in'}
---------------------
took 8.17s
min returned datetime: 2020-03-07T07:21:00+00:00 UTC
max returned datetime: 2020-03-07T07:21:00+00:00 UTC  -- OK, <2h from now :) (now=2020-03-07T07:21:14.276793+00:00 UTC)

(tesseract) >python test_parser.py IN-MH production
parser result:
{'zoneKey': 'IN-MH', 'datetime': datetime.datetime(2020, 3, 7, 12, 56, tzinfo=tzfile('Asia/Calcutta')), 'storage': {}, 'source': 'https://mahasldc.in', 'production': {'biomass': 862.0, 'coal': 19654.15141019297, 'gas': 476.75556655121227, 'hydro': 447.0, 'nuclear': 717.0930232558139, 'solar': 719.0, 'wind': 125.0, 'unknown': 316.0}}
---------------------
took 8.29s
min returned datetime: 2020-03-07T07:26:00+00:00 UTC
max returned datetime: 2020-03-07T07:26:00+00:00 UTC  -- OK, <2h from now :) (now=2020-03-07T07:26:20.438395+00:00 UTC)
```

systemcatch:
Maybe I'm not explaining this clearly enough (jarek made a good [reply](https://github.com/tmrowco/electricitymap-contrib/issues/2269#issuecomment-596096831) in #2269 ).

The consumption parser isn't needed.

parsers/IN_MH.py
```diff
@@ -0,0 +1,283 @@
...
+        'P': 'E',
+        ' ': ''
+    }
+    for key, value in chars.items():
+        str1 = str1.replace(key,value)
+        str2 = str2.replace(key,value)
+    return str1.lower() == str2.lower()
+
+
+def fetch_consumption(zone_key='IN-MH', session=None, target_datetime = None,
```
systemcatch: `fetch_production` here

parsers/IN_MH.py
```diff
@@ -0,0 +1,283 @@
...
+
+    if target_datetime is not None:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+    dt = arrow.now('Asia/Kolkata').floor('minute').datetime
+
+    data = {
+        'zoneKey': 'IN-MH',
+        'datetime': dt,
+        'consumption': {
```
systemcatch: Call this key `production`

parsers/IN_MH.py
```diff
@@ -0,0 +1,283 @@
...
+            data['consumption'][type] -= fac * values[plant]
+
+    #Sum over all production types is expected to equal the total demand
+    demand_diff = sum( data['consumption'].values() ) - values['DEMAND']
+    assert (abs( demand_diff) < 5), \
+        'Production types do not add up to total demand. Difference: {}'.format(demand_diff)
+
+    return data
+
+def fetch_production(zone_key='IN-MH', session=None, target_datetime = None,
```
systemcatch: This can be removed.

parsers/IN_MH.py
```diff
@@ -0,0 +1,283 @@
...
+    data['production'] = data.pop('consumption')
+
+    return data
+
+
+
+if __name__ == '__main__':
+    
+    consumption = fetch_consumption()
+    print(consumption)
```
systemcatch:
```suggestion
    print(fetch_production())
```

systemcatch:
That wasn't meant to sound angry btw, I'm just trying to avoid anymore confusion creeping in.

When I try to run the parser I get this error

```shell
(EM) chris@ThinkPad:~/electricitymap$ PYTHONPATH=. python mockserver/update_state.py IN-MH
Finding and executing IN-MH production parser IN_MH.fetch_production..
  File "mockserver/update_state.py", line 73, in <module>
    production = getattr(mod, fun_name)(k)
  File "/home/chris/electricitymap/parsers/IN_MH.py", line 272, in fetch_production
    data =  fetch_consumption(zone_key, session, target_datetime, logger)
  File "/home/chris/electricitymap/parsers/IN_MH.py", line 244, in fetch_consumption
    value, _ = recognize(locs['value'], image, 'digits_comma')
  File "/home/chris/electricitymap/parsers/IN_MH.py", line 189, in recognize
    text = pytesseract.image_to_string(img, lang=lang, config='--psm 7')
  File "/home/chris/envs/EM/lib/python3.7/site-packages/pytesseract/pytesseract.py", line 193, in image_to_string
    return run_and_get_output(image, 'txt', lang, config, nice)
  File "/home/chris/envs/EM/lib/python3.7/site-packages/pytesseract/pytesseract.py", line 140, in run_and_get_output
    run_tesseract(**kwargs)
  File "/home/chris/envs/EM/lib/python3.7/site-packages/pytesseract/pytesseract.py", line 116, in run_tesseract
    raise TesseractError(status_code, get_errors(error_string))
pytesseract.pytesseract.TesseractError: (1, "Tesseract Open Source OCR Engine v3.04.01 with Leptonica read_params_file: Can't open 7 Failed loading language 'digits_comma' Tesseract couldn't load any languages! Could not initialize tesseract.")
Updating and writing state..
..done
```

But I've installed the lang

```shell
(EM) chris@ThinkPad:~/electricitymap$ tesseract --list-langsList of available languages (4):
osd
equ
eng
digits_comma
```
Probably a path issue.

GitMatze:
Ok, sorry for the confusion. 
What happens if you don't use digits_comma? Just replace it with 'eng' in the code. 
If this makes it work, it might be a compability issue. I use tesseract v5, the author of the digits_comma model used v4.

parsers/IN_MH.py
```diff
@@ -0,0 +1,275 @@
...
+            'coal': 0.0,
+            'gas': 0.0,
+            'hydro': 0.0,
+            'nuclear': 0.0,
+            'solar': 0.0,
+            'wind': 0.0,
+            'unknown': 0.0
+        },
+        'storage': {},
+        'source': 'https://mahasldc.in',
```
systemcatch: You can make this `'mahasldc.in'`

systemcatch:
Yeah it fails later on in the code which probably points to a compatibility issue. I'm not entirely what tesseract version the backend is using (@robertahunt ) but hopefully there's no problem there.

You can install and run pylint to fix some of the spacing and indentation issues.

robertahunt:
Hey! Sorry about the wait on getting to review this. It took a little bit, but I managed to get it working on my system. It looks like the requirements should be okay, Unless I am mistaken though, it looks like the digits_comma file is not a standard tesseract file, so I think we should keep a local copy of that just in case.

robertahunt:
@GitMatze Thanks a bunch for all this work! 

A couple things, it looks like digits_comma is not a standard tesseract lang, could you perhaps add it to the parsers/lib folder? Just to make sure we have a persistent copy of it. 

Also, when running your code I get the following output, when the website shows the image below. To me it looks like the values are quite off. For example, i think solar should be 53, but it says 503, could you comment on these?

output:
{'zoneKey': 'IN-MH', 'datetime': datetime.datetime(2020, 3, 27, 16, 11, tzinfo=tzfile('/usr/share/zoneinfo/Asia/Kolkata')), 'production': {'biomass': 355.0, 'coal': 11653.81588816911, 'gas': 412.46062052505965, 'hydro': 863.3839072621889, 'nuclear': 1098.3395840436415, 'solar': 503.0, 'wind': 154.0, 'unknown': 88.0}, 'storage': {}, 'source': 'mahasldc.in'}

image:
image

Also, @systemcatch is completely correct, we need these values to be production, not consumption, but it looks from the image that they are reporting the production, so I think you can just remove your comment from the code, or if you could explain a bit more about why you think these are production values, that would be great.

GitMatze:
@robertahunt Thanks for getting this working!

I think it actually works just fine :)
For more details on how to derive all the output values check out https://github.com/tmrowco/electricitymap-contrib/issues/1443#issuecomment-590799851. In order to understand the dashbord I basically figured out how the sums, e.g. TTL(IPP/CPP+RE), are made up of other values in the dashboard.  You can click on the picture in the comment for further notes on that.  

Regarding solar for example,  I noticed that in order to derive  TTL(IPP/CPP+RE) you have to add RE (which includes MS SOLAR) but you have to _subtract_ SOLAR TTL, which leads to my conclusion that SOLAR TTL is already included in MS SOLAR, which means MS SOLAR is the overall solar production..

The ultimate proof that I don't miss anything or count anything twice is that all values add up to STATE DEMAND. Which, in turn, is why I'd argue that the values indeed refer to the consumption.
@alixunderplatz made the effort to find out all the plant's locations (https://github.com/tmrowco/electricitymap-contrib/issues/1443#issue-330342377) and they are spread over neighboring states as well. If I'm right declaring this as production would imply that all the plants are located in MH?

I'll add the model file and I'll also give this another test tonight to make sure that everything still works as expected.

robertahunt:
Hi @GitMatze Thanks for clarifying! After looking into the numbers in more detail, I agree your production numbers are correct, except where you include imports for other states (CS). 

In our model, we take the production values for all the states, and using the real-time imports we calculate the consumption carbon intensity ourselves. We do this in a way that allows us to account for the carbon intensity of imports from neighbouring zones, and imports from imports from neighbouring zones, and so on. What you are doing here is trying to precompute the consumption carbon intensity, (although without accounting for any imports the importing zone may have, which our flow model does take into account).

Please remove this import component. fetch_production should only return the production values per fuel type in the zone in question. Fetch_consumption should return the total demand for the zone (not broken down by fuel).

GitMatze:
Ok, I'll do that! 

Just to clarify: CS stands for Central Sector ( I wrongly wrote Central State! ), which means the plants are run by the federal government of India. Opposed to that there is the State Sector (i.e. State of Maharashtra) and the private sector. We therefore have more of an administrative than a geographical breakdown and as @alixunderplatz pointed out in his initial comment it is unclear how to assign the central sector production to the states. Since we are missing a geographical breakdown we won't get this perfectly right and ignoring CS production for now is a good option I think.

@alixunderplatz What's your take on that?

config/zones.json
```diff
@@ -2126,6 +2126,21 @@
...
     "timezone": null
   },
+  "IN-MH": {
+    "bounding_box": [
+      [72.799410, 15.613079],
+      [80.661984, 22.105644]
+    ],
+    "contributors": [
+      "https://github.com/GitMatze",
+      "https://github.com/alexunderplatz"
```
robertahunt:
```suggestion
      "https://github.com/alixunderplatz"
```

robertahunt:
Ah, I see. Thanks @GitMatze for explaining. In that case, I can see why you include the CS imports and that makes sense to me and I agree. 

Then if you could just add the digits_comma.traineddata under parsers/lib, and fix the one typo, we can get this merged :D :D

GitMatze:
Here we go!
I've already changed the code with regard to CS, but honestly I prefer CS being included. Since all three sectors have plants outside the state it would be kind of arbitrary to ignore CS production. And its acutally quite a big chunk. So let's leave it as it is ;)

Issue #2257, opened 2020-03-02
No data for most countries


StefanAO:
For the last couple of days, there is no data for most countries on my device and when I open the website.

![Screenshot_20200302-065151](https://user-images.githubusercontent.com/4057678/75649074-5157be80-5c52-11ea-92f9-3bf0e78aa793.jpg)

torbjornvatn:
I think it's related to this https://tmrow-team.slack.com/archives/C2DQPTHA6/p1582957262012400

robertahunt:
@StefanAO and @torbjornvatn Indeed, this is because we currently cannot get data for these countries through the entsoe api. I have contacted them, and we are working to get this fixed.

fbarl:
@robertahunt The [production](https://www.electricitymap.org/?page=map&solar=false&remote=true&wind=false) looks good to me now, I guess this issue can be considered resolved then?

robertahunt:
I am locally scraping the data and uploading it ... for our users it should be mostly resolved... until I have to bike to work :P #hacky resolved.

But I think we can indeed close this.

PR #2258, opened 2020-03-02
Added possibility to use second ENTSOE token

robertahunt:
Is there a better way to do this?

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -407,7 +407,13 @@ def query_ENTSOE(session, params, target_datetime=None, span=(-48, 24)):
     params['periodEnd'] = target_datetime.shift(hours=span[1]).format('YYYYMMDDHH00')
     if 'ENTSOE_TOKEN' not in os.environ:
         raise Exception('No ENTSOE_TOKEN found! Please add it into secrets.env!')
-    params['securityToken'] = os.environ['ENTSOE_TOKEN']
+    tokens = [os.environ['ENTSOE_TOKEN']]
+
+    # Due to rate limiting, we need to spread our requests across two different tokens
+    if 'ENTSOE_TOKEN_2' in os.environ:
+        tokens += [os.environ['ENTSOE_TOKEN_2']]
+    params['securityToken'] = np.random.choice(tokens)
+    
     return session.get(ENTSOE_ENDPOINT, params=params)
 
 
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -407,7 +407,13 @@ def query_ENTSOE(session, params, target_datetime=None, span=(-48, 24)):
     params['periodEnd'] = target_datetime.shift(hours=span[1]).format('YYYYMMDDHH00')
     if 'ENTSOE_TOKEN' not in os.environ:
         raise Exception('No ENTSOE_TOKEN found! Please add it into secrets.env!')
-    params['securityToken'] = os.environ['ENTSOE_TOKEN']
+    tokens = [os.environ['ENTSOE_TOKEN']]
```
corradio: Suggestion:
```suggestion
    tokens = os.environ['ENTSOE_TOKEN'].split(',')
```

That way we can pass one or multiple ENTSOE_TOKEN separated by a comma

corradio:
Do we also need to change the parser in entsoe-tp-app or does it use contrib?

robertahunt:
> Do we also need to change the parser in entsoe-tp-app or does it use contrib?

I don't know... I don't even know how I would find out.

fbarl:
Interesting - so the cause of #2257 was crossing the query limits to the ENTSOE API?

Their [API docs](https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html#_query_limits) say:

> Max 400 requests are accepted per user per minute.

I'm not very familiar with how the parsers are currently being employed, but if that's true, that should be enough to stay within limits if their responses were cached on the backend - is there any sort of caching mechanism in place right now? :thinking:

corradio:
Sent a private email

AyrtonB:
Have you tried just adding in something like `sleep(0.5)` in between calls?

robertahunt:
@AyrtonB I think we run all the parsers in parallel so I don't think this will solve it on our production server... that is what I am doing as part of my hacky fix though :P

AyrtonB:
How are the parsers parallelised? If they're always split into batches of the same domains you could have a dictionary mapping each of those batches to different keys, that could remove the chance of multiple parser operations randomly selecting the same key.

robertahunt:
Haha I like the way you think! But i'd like to avoid adding too much complexity to our system unless we have to. First we are following up with ENTSOE to see if we can arrange something.

PR #2263, opened 2020-03-03
Migrate CountryTable component to React

fbarl:

#### Tested

* [x] All 4 toggle combinations (consumption, emissions) x (production, consumption)
* [x] Table data renders well on small screens (`360px`) as well as desktop
  * [x] Tooltip shows up as expected
* [x] Renders well when no data
  * [x] No data overlay renders well when the data is not temporarily available
  * [x] Shows the no-parser message when nothing is in place for the zone
* [x] Axes render well
  * [x] Germany (most data is there)
  * [x] Poland (half of the data is there)
  * [x] Montenegro (most data not there)
  * [x] Belarus (data temporarily unavailable)
  * [x] Russia (no parser exists)

#### Changes

* Migrated `CountryTable` and `CountryTableOverlayIfNoData` (ex `NoDataOverlay`) components to React
* Fixed zone data reducer for immutability (see https://github.com/tmrowco/electricitymap-contrib/commit/c812583cfe66c3c4ec148eff8762673a1b5c70fe#diff-b6d55871d752b0241d1eaa476a4dc747R75)
* Moved all redux state _getters_ into `src/selectors` https://github.com/tmrowco/electricitymap-contrib/pull/2263/commits/3f804c6448e9a943b681295b02135ce4ab60f7c1
* Pulled out _width_ and _height_ observers into `src/effects` https://github.com/tmrowco/electricitymap-contrib/pull/2263/commits/832ebd3cefb600b9e4d3746d0157f77f01f663ac


web/src/components/countrytable.js
```diff
@@ -1,649 +1,502 @@
...
+const X_AXIS_HEIGHT = 15;
+const SCALE_TICKS = 4;
+
+function handleRowMouseMove(isMobile, mode, data, ev) {
+  // If in mobile mode, put the tooltip to the top of the screen for
+  // readability, otherwise float it depending on the cursor position.
+  const tooltipPosition = !isMobile
+    ? { x: ev.clientX - 7, y: ev.clientY - 7 }
+    : { x: 0, y: 0 };
+  dispatchApplication('tooltipPosition', tooltipPosition);
```
corradio: shouldn't we dispatch them at the same time as the two next (for performance?)

corradio:
One minor comment but aside from that, the rest looks very good!

web/src/components/countrytable.js
fbarl: Yeah that's a good point, addressed in https://github.com/tmrowco/electricitymap-contrib/pull/2263/commits/cdc1d8b2ed0d09c040bfa31de1c0208dbfcbc245 :+1:

corradio: great!

PR #2265, opened 2020-03-06
Update Singapore solar capacity

q--:
According to https://www.ema.gov.sg/solarmap.aspx, solar capacity is now
way higher. Because it's also higher than a check in the parser
(which considered everything over 200MW an error), solar would sometimes
wrongly considered invalid. I now updated the maximum value to 400.


brandongalbraith:
Consider adding yourself to the contributors list below your changes @q--

systemcatch:
@q-- looks good, add yourself to the contributors then I'll merge in.

q--:
Done

jarek:
For tracking: this cap was initially added in #1134 in early 2018 when the reported capacity was 108 MW, so indeed due for an update.

PR #2268, opened 2020-03-07
update SG PV production cap

jarek:
https://www.ema.gov.sg/solarmap.aspx currently specifies installed PV capacity as 231 MWac. The previous limit was added in early 2018 so there may well have been growth since. I've given it 400 MW to give some buffer for future growth.

The limit could arguably be 300 MW, comments welcome.

This was brought up in Slack by @q--, thank you!


jarek:
Didn't see that this was done in https://github.com/tmrowco/electricitymap-contrib/pull/2265, please ignore

PR #2275, opened 2020-03-10
Check that theme is defined when passing it to map setCo2color

fbarl:
`theme` param is not always passed to `setCo2color` (see below). Prior to this PR, it was unconditionally set in the method, which occasionally led to the state of `this.theme` being `undefined` in the map component.

https://github.com/tmrowco/electricitymap-contrib/blob/77b70262ec4b2a01a5d860d13672eb400947f59c/web/src/main.js#L309

That was a probable cause of [this Sentry issue](https://sentry.io/organizations/tomorrow/issues/1382469375/?project=1295430&query=is%3Aunresolved+device.brand%3ASamsung+device%3ASM-T720&statsPeriod=14d), so hopefully this PR resolves it.

### DIFF
--- a/web/src/components/map.js
+++ b/web/src/components/map.js
@@ -325,7 +325,9 @@ export default class Map {
 
   setCo2color(arg, theme) {
     this.co2color = arg;
-    this.theme = theme;
+    if (theme) {
+      this.theme = theme;
+    }
     this._setupMapColor();
     return this;
   }


web/src/components/map.js
```diff
@@ -325,7 +325,9 @@ export default class Map {
 
   setCo2color(arg, theme) {
     this.co2color = arg;
-    this.theme = theme;
+    if (theme) {
+      this.theme = theme;
+    }
```
corradio: does the map still work if this.theme is left undefined?

fbarl: The theme is actually _always_ set in the constructor (see below); this PR only makes the overrides in `setCo2color` conditional.

https://github.com/tmrowco/electricitymap-contrib/blob/77b70262ec4b2a01a5d860d13672eb400947f59c/web/src/main.js#L308

PR #2276, opened 2020-03-11
Migrate Tooltip component to React

fbarl:

#### Tested

* Desktop and mobile mode for _Germany_ and _Southeast Norway_ with combinations of consumption / emission and production / consumption toggles for all the tooltips:
  * Country panel production
  * Country panel exchange
  * Low carbon info
  * Price tooltip
  * Map zone info
  * Map exchange info

#### Changes

* Converted all tooltips into their own React components wrapped with new `Tooltip` component
* Added `co2ColorbarMarker` to Redux for global management of the colorbar marker state
* Made `getCo2Scale` helper more robust and `flagUri` safer with the unknown values
* Fixed the fallback interpolation state for  `CircularGauge` component https://github.com/tmrowco/electricitymap-contrib/pull/2276/commits/b8682a0c1b9f1474ed6e64fb81da8e8dbd430c93
* Some minor spacing consistency adjustments in the content of some toolbars

#### Out of scope

* Switch `CircularGauge` transitions to React (https://github.com/tmrowco/electricitymap-contrib/issues/2286)
* Render toolbars deeper in DOM (https://github.com/tmrowco/electricitymap-contrib/issues/2288)


web/src/components/circulargauge.js
```diff
@@ -70,7 +70,7 @@ export default class extends React.PureComponent {
               <path className="background" d={this.arc.endAngle(2 * Math.PI)()} />
               <path
                 className="foreground"
-                d={this.arc.endAngle((this.state.prevPercentage / 100) * 2 * Math.PI)()}
+                d={this.arc.endAngle((percentage / 100) * 2 * Math.PI)()}
```
corradio: why this change?

web/src/components/countrytable.js
```diff
@@ -25,12 +26,19 @@ const TEXT_ADJUST_Y = 11;
...
 const PADDING_X = 5;
-const FLAG_SIZE = 16;
 const RECT_OPACITY = 0.8;
 const X_AXIS_HEIGHT = 15;
 const SCALE_TICKS = 4;
 
-function handleRowMouseMove(isMobile, mode, zoneData, ev) {
+function handleRowMouseMove(isMobile, mode, zoneData, electricityMixMode, ev) {
+  dispatch({
+    type: 'SET_CO2_COLORBAR_MARKER',
```
corradio: 👍

web/src/components/countrytable.js
```diff
@@ -412,15 +409,14 @@ const CountryElectricityProductionTable = React.memo(({
...
             isMobile={isMobile}
             data={data}
           >
             <image
-              width={FLAG_SIZE}
-              height={FLAG_SIZE}
               style={{ pointerEvents: 'none' }}
               x={exchangeFlagX}
-              xlinkHref={flagUri(d.mode, FLAG_SIZE)}
+              xlinkHref={flagUri(d.mode)}
```
corradio: nice

web/src/components/faq.js
```diff
@@ -97,17 +98,12 @@ const FAQ = ({ className }) => {
...
-      // This seems to be the most browser-compatible way to iterate through a list of nodes.
-      // See: https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Example.
-      Array.prototype.forEach.call(links, (link) => {
-        const entryKey = link.href.split('#')[1];
-        if (entryKey) {
-          link.onclick = () => { setAnswerVisible(entryKey, true); };
-        }
-      });
-    }
+    modifyDOM(ref, '.entry-link', (link) => {
```
corradio: Maybe worth adding a TODO that states this should be done using react?

web/src/components/tooltips/mapexchangetooltip.js
```diff
@@ -0,0 +1,43 @@
...
+import { MAP_EXCHANGE_TOOLTIP_KEY } from '../../helpers/constants';
+import { __ } from '../../helpers/translation';
+import { co2Sub } from '../../helpers/formatting';
+import Tooltip from '../tooltip';
+
+import { CarbonIntensity, ZoneName } from './common';
+
+const mapStateToProps = state => ({
+  colorBlindModeEnabled: state.application.colorBlindModeEnabled,
+  // TODO: Should probably rename tooltipZoneData -> tooltipData
```
corradio: feel free to do it now

web/src/helpers/dom.js
```diff
@@ -0,0 +1,7 @@
+export function modifyDOM(ref, selector, callback) {
```
corradio: Maybe this function should only be in FAQ if it's only used there and it's supposed to be removed in the future?

web/src/layout/tooltips.js
```diff
@@ -1,183 +1,19 @@
...
-        <br />
-      </span>
-    </div>
-    <div id="lowcarb-info-tooltip" className="tooltip panel">
-      <span id="lowcarb-info-title" />
-      <br />
-      <small><span id="lowcarb-info-text" /></small>
-      <br />
-    </div>
+    <MapCountryTooltip />
```
corradio: That's what I'm talking about 💪

web/src/components/circulargauge.js
fbarl: I noticed that the D3 interpolation animations don't always get triggered properly with new tooltips when hovering over map (determining why is a separate question which motivated me to open #2286), so I thought it would be good to have a proper fallback value in such cases.

So if the animation doesn't get triggered, it should display the end result instead of the initial state - otherwise it will always be stuck at 0% in such cases when the gauge is first rendered.

Makes sense?

corradio: ok!

web/src/components/countrytable.js
fbarl: By the way, I noticed that a lot of the current Sentry errors are somehow caused by `flagUri` not checking for `undefined` - some of the changes I made in this PR will hopefully resolve some of these issues.

web/src/components/faq.js
fbarl: There's already a TODO comment a few lines above, I expect this will be easier to do via hash state once we add React Router.

web/src/helpers/dom.js
fbarl: Yeah good point, I used it temporarily in tooltips but now it's just used in FAQ again so I reverted it in https://github.com/tmrowco/electricitymap-contrib/pull/2276/commits/bb4e3f80792ec73aef1f56017089a72b49314547 :)

web/src/components/tooltips/mapexchangetooltip.js
fbarl: Done in https://github.com/tmrowco/electricitymap-contrib/pull/2276/commits/7e8a31a3b316a35dfa7c4745e596305185a79746.

This renaming much less painful than I expected :stuck_out_tongue:

Issue #2034, opened 2019-10-14
Data for Pakistan


UPstartDeveloper:
We haven't found the data for this region yet, but that doesn't mean you can't make it happen!

Ranked from best to useful information, here is what we could use:
- real-time data of energy generation by energy source
- historical data of energy generation by energy source (as recent and as granular as possible)
- contact details (contact form, email or social media) of someone who may have access to that data
- contact details (contact form, email or social media) of an organisation which may have access to that data

UPstartDeveloper:
Thanks to whoever wrote the text for the previous comment!

UPstartDeveloper:
Here is a list of data sources I think may be helpful. It looks like we may be able to build a parser for Pakistan!
https://energypedia.info/wiki/Pakistan_Energy_Situation 
https://www.ceicdata.com/en/pakistan/energy-consumption-and-supplies-annual
https://www.ceicdata.com/en/indicator/pakistan/electricity-production
https://www.indexmundi.com/g/g.aspx?c=pk&v=81
https://tradingeconomics.com/pakistan/electricity-production
https://www.worlddata.info/asia/pakistan/energy-consumption.php

UPstartDeveloper:
So far I have not been able to find any real-time data on how Pakistan generates its energy, or by which energy source its energy is coming from. 

For historical data, the CEIC website looks somewhat promising, but it's not very granular: https://www.ceicdata.com/en/indicator/pakistan/electricity-production. They have data on Pakistan's electricity production from for every calendar month between April 2018-March 2019. The website above DOES NOT break down the energy generation by source however. For example, data on coal and natural gas production are found on separate pages. 

Records go back as November 1987.

UPstartDeveloper:
As for an organization/person that could help, I think our best bet is the Government of Pakistan's (sometimes abbreviated GOP) Ministry of Energy (Power Division) http://www.mowp.gov.pk/. 

Here are the top 3 individuals I suggest contacting:

Officials:
1. Mr. Omar Ayub Khan, Federal Minister
PHONE: 9212442
FAX: 9224825
2. Mr. Zahoor Hussain Qureshi, Parliamentary Secretary
PHONE: 9222376
FAX: not found
3. Mr. Irfan Ali, Federal Secretary
PHONE: 9211852
FAX: 9206272
4. A full list of ministry officials can be found here: http://www.mowp.gov.pk/frmDetails.aspx 

[...]

PR #2283, opened 2020-03-15
Pakistan parser

UPstartDeveloper:
The parser function ```fetch_production``` is not implemented at the moment, nor are the other functions in this script. 

I want to just get the initial work out of the way, and make it easier for anyone else wanting to contribute to this parser.

In the meantime I will keep working on this and also look for data sources.

### DIFF
--- a/parsers/PK.py
+++ b/parsers/PK.py
@@ -87,7 +87,7 @@ def fetch_price(zone_key='PK', session=None, target_datetime=None,
     A dictionary in the form:
     {
       'zoneKey': 'PK',
-      'currency': EUR,
+      'currency': PKR,
       'datetime': '2017-01-01T00:00:00Z',
       'price': 0.0,
       'source': 'mysource.com'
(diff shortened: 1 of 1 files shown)


robertahunt:
Hi @UPstartDeveloper thanks for contributing! Usually we do not make any zone-specific parser files unless they have some working functions, just because having the file makes it look like we have a working parser in place, which would be misleading in this case, and I don't see much benefit in adding this over our example.py template that can be used for new zones. 

That said, have you had any luck finding real-time data sources for Pakistan?

UPstartDeveloper:
In terms of real-time data I have not. However on the Data for Pakistan issue (https://github.com/tmrowco/electricitymap-contrib/issues/2034) we've been able to find of handful of contacts that might help, and websites that show historical data to some extent.

I will keep looking though. One person who I recommend reaching out to is Adil Najam, he's the Inaugural Dean of the Pardee School of Global Studies, at Boston University. 

He's obviously preoccupied right now like everyone else, I reached out to him and he hasn't responded yet. All the same I know he really does care about climate change, he's also of Pakistani heritage and he obviously has better connections than I do - so it's probably still worth a shot.

The school has had to make some tough decisions about classes, residences, etc. due to coronavirus.

robertahunt:
Alright. Then I am closing this until we get some data source, but I hope we can add Pakistan to the map sometime soon!

PR #2290, opened 2020-03-18
Atomkraft -> Kernenergie

Spanni26:
- Use the more correct term "Kernenergie" instead of "Atomkraft"
- fixed an older type "Beinhaltet regenerative Energie**n**"


web/locales/de.json
```diff
@@ -134,7 +134,7 @@
...
-        "mapNuclearColors-question": "Warum werden Regionen mit Atomkraftwerken auf der Karte grün dargestellt?",
+        "mapNuclearColors-question": "Warum werden Regionen mit Kernkraftwerken auf der Karte grün dargestellt?",
         "mapNuclearColors-answer": "Die Erzeugung von Strom mittels Kernkraft hat eine niedrige Kohlenstoffintensität (<a href=\"https://en.wikipedia.org/wiki/Life-cycle_greenhouse-gas_emissions_of_energy_sources#2014_IPCC.2C_Global_warming_potential_of_selected_electricity_sources\" target=\"_blank\") Quelle</a>). Regionen mit hohem Atomstromanteil werden daher auf der Karte grüner dargestellt. Dies bedeutet aber nicht, dass es keine ökologischen Herausforderungen im Zusammenhang mit der Kernenergie gibt (wie dies auch bei anderen Formen der Energieerzeugung der Fall ist). Diese ökologischen Auswirkungen sind nicht Gegenstand dieser Karte.",
```
fbarl:
```suggestion
        "mapNuclearColors-answer": "Die Erzeugung von Strom mittels Kernkraft hat eine niedrige Kohlenstoffintensität (<a href=\"https://en.wikipedia.org/wiki/Life-cycle_greenhouse-gas_emissions_of_energy_sources#2014_IPCC.2C_Global_warming_potential_of_selected_electricity_sources\" target=\"_blank\") Quelle</a>). Regionen mit hohem Kernstromanteil werden daher auf der Karte grüner dargestellt. Dies bedeutet aber nicht, dass es keine ökologischen Herausforderungen im Zusammenhang mit der Kernenergie gibt (wie dies auch bei anderen Formen der Energieerzeugung der Fall ist). Diese ökologischen Auswirkungen sind nicht Gegenstand dieser Karte.",
```

In the same spirit as the rest of the changes in this PR, maybe `Atomstromanteil` -> `Kernstromanteil` would make sense?

web/locales/de.json
```diff
@@ -93,7 +93,7 @@
         "consumption": "Verbrauch",
         "cpinfo": "<b>Verbrauch</b> bezieht Importe und Exporte mit ein, <b>Produktion</b> ignoriert sie",
         "selectLanguage": "Sprache auswählen",
-        "lowCarbDescription": "Beinhaltet regenerative Energie and Atomkraft"
+        "lowCarbDescription": "Beinhaltet regenerative Energien and Kernenergie"
```
jarek:
```suggestion
        "lowCarbDescription": "Beinhaltet regenerative Energien und Kernenergie"
```

typo leftover

mberndt123:
The point here is that Kern is the german word for nucleus. While every kind of power plant involves atoms in one way or another, only nuclear energy involves the nuclei, and hence the term Kernenergie just makes more sense. The term Atomkraft is frequently used by the anti-nuclear crowd because it sounds more similar to Atombombe (nuclear bomb) than Kernenergie does. So I personally think this change makes sense as it's both more precise and less political.

alixunderplatz:
Seriously... we're debating changing that one single damn word "Atomkraft" in the code to a 100%-synonym "Kernenergie" which is equally used by a broad part of the Geman-speaking people...
Just for the record:
- Atomstrom, Atommüll, Atomausstieg, Atomkonsens, Atom-U-Boot, atombetrieben, Atommacht, Atomzeitalter, Atomreaktor, ... Words which are very common, and people are able to translate an relate "Atom-..." to "Kern-..." or "Nuklear-..." without any political connotation.

That being said, please leave it as "Atomstrom" as it was in that one part suggested by @fbarl:
Atomstromanteil ~~-> Kernstromanteil~~, because there is no such word or words derived from it. Searching the webs for it only leads to an electricity tariff promoted by the German nuclear-fanboy-lobby (as expected...lol).

Finally, sarcastic suggestion for the PR-title ;-) :

Spanni26:
Maybe it is a synonym. But it is technically wrong. Nuclear comes from "nucleus", the core of an atom. 
But if it is s synonym: why do you bother? 
"Kernstrom" is really a bit weird, "Strom aus Kernenergie" would be technically correct.

web/locales/de.json
Spanni26: hmm, when I think twice about it....Kernstrom ist not really a common word in germany. "Strom aus Kernenergie" would be better I think. But it will make the sentence a little bit longer.

web/locales/de.json
Spanni26: Yes, this was the old PullRequest. But it was closed because I didn't see the sesponse....

web/locales/de.json
fbarl: Seems like @alixunderplatz and you are on the same page here - sorry for pulling you in the wrong direction with my broken German :smile:

fbarl:
@alixunderplatz - assuming @Spanni26 reverts _Kernstrom_ -> _Atomstrom_ as suggested in https://github.com/tmrowco/electricitymap-contrib/pull/2290#discussion_r395010752, would you give this PR a pass or you strongly feel that the translation here is _worse_?

Spanni26:
@fbarl I think he wont. @alixunderplatz is against nuclear, so "Atomkraft" is the word he prefers. I don't know if he is objective here. 
Even wikipedia has his article named "Kernenergie"
https://de.wikipedia.org/wiki/Kernenergie 
Atomkraft is a synonym for "Kernenergie", but it is technically wrong.

freddiy:
The German Energiewende experience is one of the reasons why I became so pro-nuclear. How much of the German territory would one have to cover in solar parks to provide energy for a carbon neutral country. When is the next German coal plant opening?

Spanni26:
Any news here? I ust ask before the request is closed due to incativity

corradio:
I'll let @fbarl have the final word on this and merge when ready

fbarl:
@Spanni26 rest assured this PR won't be closed due to inactivity :slightly_smiling_face:

Because the issue is somewhat political, we've been taking our time getting everyone's input, but I'll try to make an informed decision in the next couple of days.

fbarl:
Sorry for keeping this on hold for this long @Spanni26, I got derailed with my own work on electricityMap :slightly_smiling_face:

Everyone is on board with merging this PR now, so I'm hitting the green button!

Thanks for opening the PR @Spanni26 and thank you @mberndt123 and @alixunderplatz for providing the extra arguments - we are all in this together :wink:

PR #2115, opened 2019-12-06
SASS

ovbm:
merges #2019 into master,
closes #1978

PR #2292, opened 2020-03-18
Fix mounting of OnboardingModal

fbarl:
While testing the `OnboardingModal` component, I figured I couldn't make it appear in the UI and then I realized that the mount CSS selector was wrong, which probably means none of the users were able to see the onboarding modal for a while now :astonished: 

So I changed the reference `#main` -> `#app` and removed some old SCSS code for `#main` that was never applied - [blame](https://github.com/tmrowco/electricitymap-contrib/blame/9ad12a36473fca289b2d69726b3780cc32c79730/web/src/scss/content/map/_map.scss) points to #2115, it seems like that code was introduced there.

### DIFF
--- a/web/src/scss/content/map/_map.scss
+++ b/web/src/scss/content/map/_map.scss
@@ -1,12 +1,3 @@
-#main {
-    position: fixed; /* This is done in order to ensure that dragging will not affect the body */
-    width: 100vw;
-    height: inherit;
-    display: flex;
-    flex-direction: column; /* children will be stacked vertically */
-    align-items: stretch; /* force children to take 100% width */
-}
-
 #map-container {
     position: relative;
     flex-grow: 1; /* This takes up available horizontal space */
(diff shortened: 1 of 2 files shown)


web/src/scss/content/map/_map.scss
```diff
@@ -1,12 +1,3 @@
-#main {
```
corradio: I'm super worried about removing this (as I remember testing on many devices etc..) - what is the reason? Do we have equivalent code copied somewhere? Maybe it's worth double checking in the scss PR to understand what happened.

fbarl: Ok so I think this is where it all started: https://github.com/tmrowco/electricitymap-contrib/pull/2111/files#diff-a344ab15f45451bd3fe93462dc8325d1L111-L112 (index.ejs)
image

So the container ID was renamed `main` -> `app` which I assume broke `OnboardingModal` at the time and also made this SCSS block unused.

I actually tried applying this SCSS to `#app`, but things rendered weirdly for me so I thought it'd be ok to remove it as it currently seems unused.

Of course, there is a chance I'm wrong and that `#main` is somehow set implicitly somewhere (even though I couldn't find it anywhere else in the code), so if you want to be on a very safe side, I can revert this bit and just keep the change to the `OnboardingModal` mounting point - WDYT?

corradio: Let's keep it removed, as right now, on the production, document.getElementById('main') returns null -- so this style is unused.

Issue #1157, opened 2018-02-27
Wind stream lines density unbalanced with map scale


nvbargen:
On the wind potential map, the scaling of the wind stream line density is not balanced with the overall map scale:  When taking a global view, hardly any stream lines are visible, and even big cyclone system hardly show up, whereas on a detailed scale (e.g. North Sea), even low-speed wind fields have a stream line density and length that visually would be more appropriate for a hurricane. The speed of the animation at large scale is also impressively fast (I don't get sea sick :-)).
This could probably be fine-tuned by reducing the scale sensitivity of the wind stream density (and stream line velocity).

![windstreamlinesok](https://user-images.githubusercontent.com/36658179/36722602-d714bf36-1ba5-11e8-94cf-0ed19c926b41.JPG)
![windstreamlinestoodense](https://user-images.githubusercontent.com/36658179/36722610-dbc61e26-1ba5-11e8-9fac-f22c92b0c6b7.JPG)
![windstreamlinestoofew](https://user-images.githubusercontent.com/36658179/36722614-de390eb6-1ba5-11e8-8e33-e2d2a241e431.JPG)

alixunderplatz:
@corradio @ovbm No similar issues for this stream-line related, visual topic were opened since then.
Do you want this to be fixed or leave it as it is?

corradio:
Let's keep the issue opened for now as indeed it could be nice to have this fixed

Kongkille:
For someone interested in looking into this, it would make sense to add wind/solar data to the mockserver as well (see #1876)

PR #2299, opened 2020-03-20
Migrate HorizontalColorbar component to React

fbarl:

#### Tested

Mobile + desktop mode, toggling wind and solar and hovering over map to see markers updated for all 3 colobars.

#### Notes

* There were some parts of the old `HorizontalColobar` component which didn't seem to have been used to I removed them (some ordinal scales code)
* I modified the colobars legend component a little by handing more DOM control from D3 to React
* I simplified the scales in `helpers/scales.js`, namely dropping the complex solar color scale code which wasn't used anywhere and also making max values for colorbars directly encoded in the color scale domains
* I slightly changed the number of ticks displayed with each colorbar, see screenshot below:

image


web/src/components/horizontalcolorbar.js
```diff
@@ -1,217 +1,83 @@
...
-    const inter = d3.interpolate(e[0], e[1]);
-    const value = inter(i / (10 - 1));
-    return that.originalColorScale(value);
-  });
-  this.scale = d3.scaleLinear()
-    .range(this.colors)
-    .domain(this._domain);
+const HorizontalColorbar = ({
+  colorScale,
+  currentMarker,
```
corradio: maybe rename to `currentValue` to make clear it's a float?

web/src/components/map.js
```diff
@@ -112,7 +112,7 @@ export default class Map {
       nonClickableFill: '#D4D9DE',
       oceanColor: '#FAFAFA',
       co2Scale: {
-        steps: [0, 150, 600, 750],
+        steps: [0, 150, 600, 800],
```
corradio: why this change?

web/src/helpers/scales.js
```diff
@@ -1,19 +1,16 @@
...
 import { themes } from './themes';
 
-const d3 = Object.assign(
-  {},
-  require('d3-array'),
-  require('d3-interpolate'),
-  require('d3-scale'),
-  require('d3-scale-chromatic'),
+const uniformSpread = ({ from, to, count }) => (
+  range(count).map(index => (from + (to - from) * index / (count - 1)))
```
corradio: you could consider using d3.range:
https://github.com/d3/d3-array#range

web/src/components/map.js
corradio: if I remember correctly we want the 750 to have the same color as all values >750, which is different from having the last color set at 800.

web/src/components/horizontalcolorbar.js
fbarl: Good point! I fixed it in https://github.com/tmrowco/electricitymap-contrib/pull/2299/commits/8d13810bfe0ca8808c7162cf20c3e441246dc5f2.

FYI I also uncovered a typo there which made me realize that solar layer markers are actually never being displayed - I opened https://github.com/tmrowco/electricitymap-contrib/issues/2305 to address that separately.

web/src/components/map.js
fbarl: Ah ok, I see. What I was trying to do there is get rid of the separate constraint constants (in this case `maxCo2`) and make the domain limits be encoded in scales themselves for simplicity.

> we want the 750 to have the same color as all values >750

I made this condition explicitly encoded across all `co2Scale` themes in https://github.com/tmrowco/electricitymap-contrib/pull/2299/commits/79e9082654657f73b50c27098c5cff2a4f995992 - does that make sense?

web/src/helpers/scales.js
fbarl: Thanks for the tip, but that didn't make it much nicer due to rounding-off errors and neither D3 nor lodash range including the end value by default.

What I decided to do instead is just be simple and list the domain steps explicitly in https://github.com/tmrowco/electricitymap-contrib/pull/2299/commits/1a2d2acd0c0bc63cad57ff46445808adbcda42b3. I also noticed that _15 m/s_ is very low for an upper cap on wind speed (the marker would often go off the scale), so I doubled it to _30 m/s_, makes sense?

image

fbarl:
@corradio PTAL, I made some changes to the scales.

web/src/components/map.js
corradio: perfect

web/src/helpers/scales.js
corradio: hmm are you sure it still looks nice visually?

fbarl: It looks different but to me it doesn't look worse - at least values don't go off scale anymore :)

_before_
image

_after_
image

I think the real problem now is with stream lines density, i.e. it would be nice to address https://github.com/tmrowco/electricitymap-contrib/issues/1157.

fbarl:
Thanks for approving @corradio!

I'll merge the PR now and we can keep an eye on the wind scale if any futher adjustments / improvements will be needed.

PR #2301, opened 2020-03-22
Split US into Balancing Authorities

robertahunt:
solves #143 (Hopefully!)

This PR uses two sets of geojsons:
1. The first is a simplified version of the US with the overlaps crudely removed (for each overlapping polygon/balancing authority, it will always show the smaller one).
2. The other will be used for the backend when querying by latitude/longitude, and for getting temperature and weather data. 

The original shape data was obtained from https://hifld-geoplatform.opendata.arcgis.com/datasets/control-areas 

Example 'look' (the carbon intensities values were chosen at random here, just to illustrate how it could look, they are not at all related to the true US carbon intensities):

image

Some still TBD items:
- [x] Test api lat/lon queries - Currently will only return one area when multiples discovered. Should fix this in private repo
- [x] Add zones to zones.json
- [x] Add exchanges to exchanges.json
- [x] replace EIA parsers with BA specific parsers where possible
- [x] add exchanges and zones to EIA.py
- [x] generate bounding boxes

TBD after contrib implementation:
- refetch historic data
- retrain forecasts


robertahunt:
@fbarl Ideally I would also like it to show the 'true' (larger) balancing authority area when hovering over the zone, Is this possible/easy? It looked to me like modifying web.js might do it, but I am not so familiar with the frontend.

parsers/EIA.py
```diff
@@ -20,18 +20,189 @@
...
     'US-BPA->US-IPC': 'EBA.BPAT-IPCO.ID.H',
     'US-SPP->US-TX': 'SWPP.ID.H-EBA.ERCO',
     'US-MISO->US-PJM': 'EBA.MISO-PJM.ID.H',
     'US-MISO->US-SPP': 'EBA.MISO-SWPP.ID.H',
     'US-NEISO->US-NY': 'EBA.ISNE-NYIS.ID.H',
-    'US-NY->US-PJM': 'EBA.NYIS-PJM.ID.H'
+    'US-NY->US-PJM': 'EBA.NYIS-PJM.ID.H',
+
+#Exchanges to non-US BAs
+    'US-CAL-CISO->MX-BC': 'EBA.CISO-CFE.ID.H', #Unable to verify if MX-BC is correct
```
corradio: given the comment, is there any action item here? Something we should check before merging?

martincollignon:
wait do we have the US zones ready??!?!

parsers/EIA.py
robertahunt: I tried to find more information on which Mexican BAs had exchanges, but was unable to find it on the Mexico side. Since we don't have any mexican production data at this point, I don't think this will affect anything yet, this is more a note for the future, if we find a source for mexican data we should check these exchanges too.

MaximeBaudette:
Seems like there is an error in mapping CAISO in CA, USA. Half of northern CA is included in WAPA, When I believe it is part of CAISO. 
Screen Shot 2020-08-25 at 16 08 47

PR #2308, opened 2020-03-24
Add CA-YT Hydro Capacity

nessie2013:


config/zones.json
```diff
@@ -960,8 +961,11 @@
       [
         -123.31910336003193,
         70.15420156500005
-      ]
-    ]
+     ]
+    ],
+    "capacity": {
+      "hydro": 92
+    },
```
jarek: This trailing comma breaks parsing JSON

jarek:
```suggestion
    }
```

Issue #2312, opened 2020-03-25
Simplify CountryHistoryMixGraph data processing based on displayByEmissions value


fbarl:
That graph actually shows very different content depending on whether the _Carbon emissions_ tab is selected in the country panel or not.

Splitting it into two separate graph wrappers and branching into one of them depending on `tableDisplayEmissions` would make sense semantically and would also make the code simpler.

corradio:
Would it be possible to pass a prop "displayByEmissions" to <CountryHistoryCarbonGraph> in order to selectively use a different `prepareGraphData` and re-use the underlying <AreaGraph> instance?
That would enable animations etc.. and avoid hiding/reconstructing a new graph everytime.

fbarl:
> Would it be possible to pass a prop "displayByEmissions" to CountryHistoryCarbonGraph in order to selectively use a different `prepareGraphData` ...

Yes, that's definitely what I had in mind as a part of the solution.

>  and re-use the underlying AreaGraph instance? That would enable animations etc.. and avoid hiding/reconstructing a new graph everytime.

I still have my doubts whether animations in this case are a good idea for the same reason as I argued in https://github.com/tmrowco/electricitymap-contrib/pull/2313#discussion_r398001302, but if you feel strongly that they would improve the user experience, we can add an issue for re-introducing them to the `AreaGraph` component and keep this case in mind.

The graph initialization and rendering is still very cheap, actually - e.g. with the current carbon / emissions graph, I get a very smooth transition when switching between the tabs even though a graph gets destroyed and rendered fresh since #2313.

corradio:
Upon thinking a bit more about it, I think it makes sense to allow for re-using the same underlying AreaGraph for some cases.
For example, when switching between countries, it makes sense that the bars & lines shrink/grow.
Furthermore, when switching to the emission tabs, it also makes sense that the bars & lines shrink/grow as it shows how much a % of the mix relates to a % of emissions.
I think we could do everything as part of this PR?

fbarl:
> Upon thinking a bit more about it, I think it makes sense to allow for re-using the same underlying AreaGraph for some cases. For example, when switching between countries, it makes sense that the bars & lines shrink/grow.

You are right, that is indeed a good use case! :+1:

> Furthermore, when switching to the emission tabs, it also makes sense that the bars & lines shrink/grow as it shows how much a % of the mix relates to a % of emissions.

[...]

Issue #2286, opened 2020-03-16
Replace CircularGauge D3 interpolation with React


fbarl:
D3 interpolation doesn't play will with React rendering cycle and there are many good alternatives for it in React like:

* CSS animations / transitions
* [React Motion](https://github.com/chenglou/react-motion)
* [React Transition](https://reactcommunity.org/react-transition-group/transition)

PR #2313, opened 2020-03-25
Add total emissions graph

### LINKED ISSUE #2300: Switching country panel to "emissions" doesn't change carbon intensity graph to emissions graph
When switching to "emissions" in the country panel, the carbon intensity graph should change title and show the absolute emission per interval instead of the carbon intensity (just like the mix graph).

cc @fbarl 

image

fbarl:
Resolves #2300.

This PR displays the total carbon emissions graph instead of carbon instensity graph in the country panel when _Carbon emissions_ tab is selected.

_before_
image

_after_
image

#### Notes

* A new entry in locales needed to be added for this (`emissions24h`) - hopefully the translations will follow up soon
* I created a new graph wrapper component for this - I realized this approach should probably be taken to split `CountryHistoryMixGraph` in a similar way so I opened #2312
* The newly added graph is in a way just a simpler version of the mix graph below (which displays emissions by source) which makes me a bit uncertain whether this PR is a better solution to #2300 than simply removing the carbon intensity graph when showing the emissions :thinking: WDYT @corradio?


corradio:
good points! I think it's fine to leave the graph for now

web/src/layout/leftpanel/countrypanel.js
```diff
@@ -211,24 +212,24 @@ class Component extends React.PureComponent {
...
+                  ))}
+                </span>
                 <br />
                 <small className="small-screen-hidden">
                   <i className="material-icons" aria-hidden="true">file_download</i> <a href="https://data.electricitymap.org/?utm_source=electricitymap.org&utm_medium=referral&utm_campaign=country_panel" target="_blank">{__('country-history.Getdata')}</a>
                   <span className="pro"><i className="material-icons" aria-hidden="true">lock</i> pro</span>
                 </small>
 
-                <CountryHistoryCarbonGraph />
+                {tableDisplayEmissions ? <CountryHistoryEmissionsGraph /> : <CountryHistoryCarbonGraph />}
```
corradio: I'm just thinking aloud here: the graphs won't animate using a transition when we toggle `tableDisplayEmissions`. That's a shame. Wouldn't it be simpler if we re-used the same elements, but just passed different data to it? That would achieve the animation effect.
Actually I didn't test it so I'm not sure about the statement I'm making.

fbarl: That's a good point, but graph animations are already not there as you can see by toggling `tableDisplayEmissions` and observing the mix graph (which passes two different datasets to one `AreaGraph`).

The animations were lost when we started rendering graphs with React instead of D3, but it wouldn't be too hard to reintroduce them with some combination of CSS and React animations (e.g. see https://github.com/tmrowco/electricitymap-contrib/issues/2286).

However, while having graph layers shrink and expand in a smooth motion make for a nice UX when the coordinate units are fixed, I think it might be misleading and undesirable when switching between completely different graphs. So I'd personally opt out of animations here even if there was no additional work involved :)

PR #2316, opened 2020-03-25
Fix Florida's takeover of world map

robertahunt:
No art today
before:
image

now:
image


corradio:
love the commit name!

PR #2318, opened 2020-03-26
Updated names for new us zones

robertahunt:
I updated the english names to reflect the names of the balancing authorities. Some of them spill off the list now, so I am wondering how we feel about that, and I am not sure how these would look on mobile devices. I could find a shorter description for them, but then I think it would be great if we could have the longer name somewhere so people can look up the name of the balancing authority in their area.

image

### DIFF
--- a/web/locales/en.json
+++ b/web/locales/en.json
@@ -1455,6 +1455,274 @@
             "countryName": "United States of America",
             "zoneName": "ERCOT"
           },
+          "US-CAL-BANC": {
+            "countryName": "United States of America",
+            "zoneName": "Balancing Authority Of Northern California"
+          },
+          "US-CAL-CISO": {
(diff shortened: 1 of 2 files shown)


fbarl:
> Some of them spill off the list now, so I am wondering how we feel about that...

I'd go with a simple solution here and just text ellipsis:

image

We can also add a simple tooltip (browser native one) with the full zone name, but otherwise I think not displaying the whole zone name should be fine as long it's shown when the zone is selected, which already works fine right now:

image

One more thing I noticed is that the tooltips can get very wide with these long names which sometimes makes them look weird, but maybe we can open a separate issue to address that.

image look

robertahunt:
would be great if you could commit the ellipsis :). I would not know how to do that!

Issue #2319, opened 2020-03-26
Imports into new US zones don't work


jarek:
Just noting - I guess it's because they've been renamed?

Looks great otherwise!

![Screenshot_2020-03-26_08-58-24](https://user-images.githubusercontent.com/47415/77649322-10488680-6f40-11ea-950d-bead5a3496f1.png)
![Screenshot_2020-03-26_08-58-40](https://user-images.githubusercontent.com/47415/77649326-12124a00-6f40-11ea-9f12-a890ea19a812.png)

fbarl:
It seems that the problem is on the backend - in case of New York, the `/history` API response doesn't contain the `exchange` field even when other data is present. This might be true for other US regions as well.

![image](https://user-images.githubusercontent.com/1216874/77650217-6d0d6680-6f6b-11ea-97c0-e6ab3ea34255.png)

It's possible that the exchanges are somehow not be available yet and the only reason they show up is because `/state` returns exchanges between _old_ US regions:

![image](https://user-images.githubusercontent.com/1216874/77650578-00469c00-6f6c-11ea-81b1-34eddc10cb8a.png)

cc @robertahunt - maybe you have an idea of what's going on here? :)

robertahunt:
Thanks @jarek. I thought it was just the exchanges to outside the US that weren't working (#2320 should fix that), but it appears none of the new exchanges are working. The plot thickens...

fbarl:
@jarek @robertahunt Seems like you guys already knew what needed to be done! :wink:

robertahunt:
Fixed by #2320

jarek:
Ontario's exports remain missing from NYISO and MISO balances:

![ontario](https://user-images.githubusercontent.com/47415/78955975-e69a6e00-7aae-11ea-85ab-daefee957faa.png)
---
![ny](https://user-images.githubusercontent.com/47415/78955974-e601d780-7aae-11ea-8fd0-14683d04c887.png)
---
![miso](https://user-images.githubusercontent.com/47415/78955977-e69a6e00-7aae-11ea-95ab-75d759db1dbc.png)

jarek:
Just wondering if this is because you area asking CA_ON parser for data for exchange `CA-ON->US-NY-NYIS`? Ontario, and likely the other non-American parsers, weren't updated to account for the new zone names.

PR #2320, opened 2020-03-26
Fixed key errors for exchanges

robertahunt:
Fixes a bug in the US exchanges that i introduced when moving over to the BA-level exchanges

Namely, when querying the EIA api for exchanges to zones outside the US, they only give them in one direction, which does not match our 'alphabetical' exchange naming convention. To fix this, I have a list of zones which need to be multiplied by -1 to get the right exchange data.


robertahunt:
Fixes #2319

robertahunt:
Okay, now it should fix #2319 . I think it is because EIA does not have realtime data so the exchanges do not show up on the map, so I updated the exchanges that we have more realtime data for to new sources.

PR #2321, opened 2020-03-27
Circle CI update

corradio:
Fixes #2294 

FYI @robertahunt @madsnedergaard

PR #2322, opened 2020-03-27
added clarification on data tooltip

tth4vb:
For some US zones, while we don't have live data we do have historical data. To make sure this is clear, I want to update the tooltip you get when hovering over zones with a temporary data outage to clarify that specifically "live" data is unavailable.


corradio:
Test tests are failing because https://github.com/tmrowco/electricitymap-contrib/pull/2321 needs to be merged first.

tth4vb:
> Test tests are failing because #2321 needs to be merged first.

Got it, thanks. was confused what wasn't working

PR #2327, opened 2020-03-29
Added US zones to exceptions

robertahunt:
Some of the new US zones do not have coal, oil, gas, or unknown production, but were not yet added to quality checks exceptions. This adds them.

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -63,7 +63,10 @@ def validate_production(obj, zone_key):
          obj.get('production', {}).get('coal', None) is None and
          obj.get('production', {}).get('oil', None) is None and
          obj.get('production', {}).get('gas', None) is None and zone_key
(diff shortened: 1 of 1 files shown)


corradio:
Just to make sure: is it because it's not present in the data, or is it because they don't have physical assets (i.e. capacity is 0)?

robertahunt:
It is slightly different in each case, but it is primarily because these BAs only have hydro capacity.

corradio:
Approving, but just to make sure I understood correctly: we agree that all those BAs have negligible fossil fuel (oil, gas, coal) capacity right? If not, then we should quickly discuss.

robertahunt:
EIA is not reporting any fossil fuel capacity for these zones. It is possible they have some fossil fuels that are not reported by EIA, but I find that unlikely.

corradio:
got it

PR #2328, opened 2020-03-29
Updated troubleshooting based on common issue

robertahunt:
The solution is based on responses from @corradio on #ask-anything on our slack channel, from a message on Jan 18. I have not been able to reproduce the problem though, so a quick check if the solution is correct would be nice.


robertahunt:
@corradio the build has failed twice on this. Any idea why?

corradio:
I think it's fixed with https://github.com/tmrowco/electricitymap-contrib/pull/2334

Issue #2331, opened 2020-03-30
CircleCI failing when running tests


fbarl:
![image](https://user-images.githubusercontent.com/1216874/77897902-95060e00-727a-11ea-8e02-fad03ad6e84c.png)

Perhaps a timeout longer than 20 seconds is needed to ensure everything boots up properly?

![image](https://user-images.githubusercontent.com/1216874/77897974-b36c0980-727a-11ea-8b4d-8778ef696667.png)

I hit this twice in a row on https://github.com/tmrowco/electricitymap-contrib/pull/2324 and I also noticed https://github.com/tmrowco/electricitymap-contrib/pull/2328 was failing because of it.

cc @corradio @robertahunt

corradio:
I'm on it

PR #2334, opened 2020-03-30
Fix circle ci

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2331

Turns out a `watch` container was also started, which would cause unnecessary reload of the API

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
           name: Start environment
           command: |
             set -euo pipefail
-            docker-compose up -d
+            docker-compose up -d web mockserver
             sleep 20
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           name: Start environment
           command: |
             set -euo pipefail
-            docker-compose up -d
+            docker-compose up -d web mockserver
```
fbarl: > Turns out a `watch` container was also started, which would cause unnecessary reload of the API

Cool, makes sense not to have it here, maybe I'd just add a comment above explaining that and linking to this PR but that's up to you :slightly_smiling_face:

PR #2335, opened 2020-03-30
[POC] CSS transitions for AreaGraph layers

### LINKED ISSUE #2336: Re-introduce animated transitions to left panel elements
Some D3 default transition were removed during migration to React components - it should be easy to re-introduce them either with CSS or React Motion (see https://github.com/tmrowco/electricitymap-contrib/issues/2286).

Only two elements come to my mind that we might want to animate again:

* [ ] `CountryTable` component horizontal bars
* [ ] `AreaGraph` component layers

@corradio can you think of anything else to add?

fbarl:
An attempt to resolve #2336.

### DIFF
--- a/web/src/scss/content/panel/_country-panel.scss
+++ b/web/src/scss/content/panel/_country-panel.scss
@@ -103,8 +103,9 @@
             margin-bottom: 1.2em;
         }
     }
+
     path {
-        /*shape-rendering: crispEdges;*/
+        transition: all 0.5s ease;
     }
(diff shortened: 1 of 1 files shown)


fbarl:
This approach didn't work ideally, but not great as the layers would often overlap or appear loosely connected during the transition, but the effects is still barely noticable and the animations are overall rather smooth.

I'm closing this PR for now but will probably get back to the issue in the next couple of weeks.

Issue #2309, opened 2020-03-24
Move Map component to React


fbarl:
Follow-up on https://github.com/tmrowco/electricitymap-contrib/issues/1681#issuecomment-572014604.

It would be good to have _Map_ as an isolated React component rendered in one place instead of having its state managed across multiple places in `main.js`.

fbarl:
As @corradio pointed out to me earlier, using https://github.com/visgl/react-map-gl might remove some of the complexity of this task.

PR #2338, opened 2020-03-30
Move some tooltips deeper in DOM

### LINKED ISSUE #2293: Time slider bug
Hi there,
it seems that moving the time slider doesn't change the displayed values in the tooltip anymore.
To reproduce, try the following:

App:
1.) select a country
2.) tap on one of the bars to show the tooltip box
3.) move the slider
--> the bars size will change, but the data in the tooltip stays the same

Browser:
1.) select a country
2.) select the time slider
3.) move your arrow to one of the bars/graphs
4.) use left and right arrows to move slider
--> the bars size will change, but the data in the tooltip stays the same

fbarl:
Moves the following tooltips deeper in React rendering tree:

* `CountryPanelEmissionsTooltip`
* `CountryPanelExchangeTooltip`
* `CountryPanelProductionTooltip`
* `LowCarbonInfoTooltip`
* `PriceTooltip`

That addresses most of #2288 keeping only the following tooltips global (and keeping the Redux tooltip state for them):

* `MapCountryTooltip`
* `MapExchangeTooltip`

Those two are blocked by `<Map />` not being wrapped in a React component yet (see #2309).

#### Tests

I tested all the tooltips above in my Chrome browser (both desktop and mobile mode), but didn't go beyond that.

#### Other changes

* Due to the way `z-index` works, the tooltips actually need to be rendered at the top of the DOM hierarchy to float above all other elements. However, this is not a problem if we can still get all the benefits of rendering the deeper in the _React render tree_. I wrapped `Tooltip` component with  [Portal](https://github.com/tajo/react-portal) which exists for this specific use case (see https://github.com/tmrowco/electricitymap-contrib/pull/2338/commits/3ae5bb2f3c1c940a19d0c4654822316236edded5)
* Made some **breaking changes** to the `<AreaGraph />` component props:
  * `backgroundMouseMoveHandler` now gets passed a single argument (_timestampIndex_)
  * `layerMouseMoveHandler` now gets passed two arguments (_timestampIndex_, _layerIndex_)
  * There are now two new callbacks than can be passed to the component: `markerUpdateHandler` and `markerHideHandler` - the former receiving 3 arguments (_position_, _datapoint_, _layerKey_), all three relating to the current graph marker. The motivation for this was supporting external manipulation of the graph marker state (e.g. by keyboard events) instead of assuming it's always triggered by mouse events. This should fix #2293.
* All the graph tooltips are now attached to the marker position instead of the cursor position, so that the marker itself always stays visible and the tooltips don't necessarily get updated so frequently so the experience is a little bit smoother
* I disabled pointer events on the `svg` of `CircularGauge` to prevent tooltips blinking and let the wrapping `div` have the full control of the hover events
* I also hopefully fixed the `CountryTable` row clicks issue on iOS mobile reported by one user - see https://github.com/tmrowco/electricitymap-contrib/pull/2338/commits/f2503e19639595fbed53b1d9aea72dce0ba52f01


web/src/components/countryhistorycarbongraph.js
```diff
@@ -63,40 +60,64 @@ const CountryHistoryCarbonGraph = ({
...
+        layerMouseMoveHandler={mouseMoveHandler}
+        layerMouseOutHandler={mouseOutHandler}
+        markerUpdateHandler={markerUpdateHandler}
+        markerHideHandler={markerHideHandler}
+        selectedTimeIndex={selectedTimeIndex}
+        selectedLayerIndex={selectedLayerIndex}
+        isMobile={isMobile}
+        height="8em"
+      />
+      {tooltip && (
```
corradio, line 114: niiiiice

PR #2341, opened 2020-03-31
Add test for consistency of co2eq

corradio:
Adds a list of required keys and checks that each country has the necessary fallbacks


corradio:
As expected, tests are failing for now on this commit

image

corradio:
Tests are now passing

Issue #2345, opened 2020-04-01
Migrate layerbuttons.js buttons into React


fbarl:
A small step in resolving #2311.

The three buttons involved are toggles for solar layer, wind layer and display mode:

![image](https://user-images.githubusercontent.com/1216874/78180216-be818e00-7462-11ea-83bb-82569d498497.png)

PR #2346, opened 2020-04-01
Generalized Toggle component

### LINKED ISSUE #2325: Turn prodconstoggle.js into a re-usable component
Create a re-usable component that can take a redux key alongside a list of (key, labels) and a tooltip description.

fbarl:
Resolves #2325.

Tested on desktop and mobile, also for cases of toggle containing more than 2 options and the info tip not being passed.

#### Major changes

* Generalized `ProdConsToggle` component into a more robust stateless `Toggle` component with the same styles
* Pulled out the fading info tooltip into a separate `InfoTooltip` component - this component can be reused when addressing #2345 and could potentially converge to the `Tooltip` component in a longer run
* I moved all the relevant CSS styles into the component files here and I chose to do it with the newly introduced `styled-components` - the main reason I didn't use JS to embed styles in props is because I didn't know how to manage `&:hover` state nicely on the CSS level (without introducing a React component variable that would track hovers). @corradio WDYT about that? I can always switch away in this PR if we find another suitable alternative (see #2261).

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -35,6 +35,7 @@
     "redux": "^3.7.2",
     "redux-logger": "^3.0.6",
     "sprintf-js": "^1.1.2",
+    "styled-components": "^5.0.1",
     "topojson": "^3.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


fbarl:
@corradio FYI most of the styles I moved directly `ProdConsToggle` -> `Toggle` and some of them might not be desirable generally (e.g. `box-shadow`) depending on how `Toggle` is going to be reused, but we can address that concern in separate issues if the need occurs.

PR #2347, opened 2020-04-02
Co2 eq parameters consistency

corradio:
This is a draft PR that brings back those reverted commits:

image

It's been prepared here so it can be merged (and possibly reverted) as one single atomic commit.


robertahunt:
Thanks for taking care of reopening a PR on this. I had it on my to-do list

PR #2349, opened 2020-04-02
Update German

jld3103:


robertahunt:
Thanks for this @jld3103 ! Looks good to me

PR #2350, opened 2020-04-02
Add wrapper to right-controls

corradio:
I've added a wrapper to start wrapper controls that appear on the right.
For now I've only wrapped the toggle but I'd be good if we could put them all there!

### DIFF
--- a/web/src/components/toggle.js
+++ b/web/src/components/toggle.js
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { findIndex, isEmpty } from 'lodash';
+import { isEmpty } from 'lodash';
 import styled, { css } from 'styled-components';
 
 import InfoTooltip from './infotooltip';
@@ -74,7 +74,6 @@ const InfoButton = styled.div`
 `;
 
 export default ({
-  className,
   infoHTML,
   onChange,
   options,
@@ -83,7 +82,7 @@ export default ({
   const [tooltipVisible, setTooltipVisible] = useState(false);
 
   return (
-    <Wrapper className={className}>
+    <Wrapper>
       <Options>
         {options.map(o => (
           <OptionItem key={o.value} active={o.value === value} onClick={() => onChange(o.value)}>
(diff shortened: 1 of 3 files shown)


web/src/layout/main.js
```diff
@@ -57,16 +57,17 @@ const Main = ({ brightModeEnabled, electricityMixMode, isLeftPanelCollapsed }) =
             </a>
           </div>
           <Legend />
-          <Toggle
-            className="production-consumption-toggle"
```
fbarl, line 61: Good idea with the wrappers! I think it's ok to remove the `className` prop from the`Toggle` component if we're going to be using wrappers instead.

corradio:
@fbarl I've added a commit that makes sure that a click on "consumption" switches to "consumption" instead of cycling through them.
This is necessary when we have more than two items
image

fbarl:
> @fbarl I've added a commit that makes sure that a click on "consumption" switches to "consumption" instead of cycling through them. This is necessary when we have more than two items

Good call, it makes more sense that way! :+1:

web/src/components/toggle.js
```diff
@@ -73,7 +73,7 @@ const InfoButton = styled.div`
   }
 `;
 
-const Toggle = ({
+export default ({
   className,
```
fbarl, line 77: Not sure if I made myself clear enough, but in https://github.com/tmrowco/electricitymap-contrib/pull/2350#discussion_r402182450, I was suggesting removing this prop if it's not being used anymore.

web/src/components/toggle.js
```diff
@@ -82,15 +82,11 @@ const Toggle = ({
 }) => {
   const [tooltipVisible, setTooltipVisible] = useState(false);
 
-  const activeIndex = findIndex(options, { value });
```
fbarl, line 85: I guess `findIndex` is not used here anymore, so it doesn't have to be imported (which reminds me, we should probably add a linting rule for that :wink:)

web/src/components/toggle.js
corradio: ah got it! Will do

PR #2352, opened 2020-04-03
Fixed exchange arrow directions

### LINKED ISSUE #2351: Double exchange arrows for some Canada - USA exchanges
Arrows for exchanges between US systems and Quebec, BC, and Alberta are showing double on the map. I'm guessing it's related to the US state split?

quebec double arrows

bc ab double arrows

robertahunt:
partially fixes #2351 - new exchange arrows are pointing in the wrong direction


config/exchanges.json
```diff
@@ -576,7 +576,7 @@
     "parsers": {
       "exchange": "EIA.fetch_exchange"
     },
-    "rotation": 0
+    "rotation": 180
```
jarek, line 579: I'm unsure if `CA-MB->US-MIDW-MISO` actually works? It's not showing anything right now. However, the rotation seems to be correct by analogy with `CA-BC->US-NW-BPAT` :+1:

robertahunt: I think the exchanges only show if they have 'live' data, which is kind of a shame. EIA has data for this exchange, it just comes in later (8-24 hours delay), so it won't show on the map, but it will be included in historic calculations.

Issue #2343, opened 2020-04-01
Maharashtra (India) parser: tesseract error


GitMatze:
(1, 'Error opening data file /usr/share/tesseract-ocr/4.00/tessdata/digits_comma.traineddata Please make sure the TESSDATA_PREFIX environment variable is set to your "tessdata" directory. Failed loading language \'digits_comma\' Tesseract couldn\'t load any languages! Could not initialize tesseract.')

GitMatze:
@robertahunt Does the file exist? 
I'd expect tesseract v4 to work.

robertahunt:
Hey @GitMatze. Nice catch!

Sorry, but I expected this to fail (I just expected to fix it before you noticed too ;) ). I wasn't sure where our production server would look for the file, so I just left it until it would break and tell me. I'll copy it over to the correct place and then it should work.

GitMatze:
Ok! 
I guess I was too excited to see this running ;)

robertahunt:
@GitMatze It looks like there must still be some problem because it isn't showing up on the map :(, but there are no kibana errors. I am digging into the issue.

robertahunt:
Just to give you an update, @GitMatze, our parser-feeder has a timeout of 5 minutes, and this takes 500 s to run on the server at the moment - we don't give it a lot of computing power for the parsers, so the parser is timing out and then the data is not being uploaded to the database. We are looking into if we can fix this (increase the timeout, log the error properly).

GitMatze:
@robertahunt That's almost as long as the update interval right? I might be able to reduce the computational demand a bit by further reducing the size of the image sections that go into tesseract for recognition. But if i look at it, there's not too much room for optimization. A last resort option would be not to recognize the names. I considered them to be a nice validation regarding possible longterm layout changes of the dashboard but it's probably not really necessary. And we could most likely cut computation time in half.  
![grafik](https://user-images.githubusercontent.com/45139485/78371875-d8d57c00-75c8-11ea-965d-0a18d189da55.png)

robertahunt:
Hey @GitMatze Unfortunately it seems like extending the time for individual parsers to run could be an issue, but it seems like if we don't recognize the labels, we can run the parser in 30 seconds on the production server, so I suggest we do that. I'll write a PR for that ASAP so hopefully this parser will be up and running.

PR #2359, opened 2020-04-07
Removed label recognition in IN-MH parser so it doesn't timeout

robertahunt:
FYI @GitMatze 

related: #2343


GitMatze:
I just realized that the 'labels' dictionary isn't used anyway. You can also delete those two occurences.

PR #2360, opened 2020-04-07
Add parser for Gulf Cooperation Council countries

### LINKED ISSUE #2357: Data for Gulf countries (United Arab Emirates, Bahrain, Saudi Arabia, Oman, Qatar, and Kuwait)
I did some digging and found https://www.gccia.com.sa/, which has a map with "Demand Now" data for Gulf Cooperation Council countries, which are United Arab Emirates, Bahrain, Saudi Arabia, Oman, Qatar, and Kuwait.

Comparing the value on the map for Kuwait with the [data source](https://www.mew.gov.kw/en) ElectricityMap currently uses for that country, the value on the GCCIA site is only slightly different, possibly due to latency or the fact that I didn't load both sites at the exact same time.

They also have [an app](https://www.gccia.com.sa/gcciamobile.html), with screenshots implying it has historical data graphs.

Edit: forgot to mention - a few countries show an `-` instead of electricity demand right now (Oman, Saudi  Arabia). This might just be temporary.

q--:
Which are the United Arab Emirates, Bahrain, Saudi Arabia, Oman, Qatar, and
Kuwait.

Oman and Saudi Arabia are currently not returning data. If/when data is available again, the parser will be able to retrieve it.

This pull request doesn't take care of adding the right data to config/co2eq_parameters.json and
config/zones.json, I'm not 100% sure if I know how you are doing the math for the former and don't yet understand how to add the bounding box for the latter. 

Do note that the UAE and Oman are in a different time zone than the other countries. The parser takes care of this, I don't know if it needs to be configured in zones.json as well.

Closes #2357 - but note that I haven't examined the app I mentioned there, which might have more data than I can scrape from the website. Might be worth opening a new issue (or keeping #2357 open).


README.md
```diff
@@ -142,7 +143,7 @@ Real-time electricity data is obtained using [parsers](https://github.com/tmrowc
 - Japan (Shikoku): [Yonden](http://www.yonden.co.jp/denkiyoho/)
 - Japan (Tōhoku-Niigata): [TH-EPCO](http://setsuden.tohoku-epco.co.jp/graph.html)
 - Japan (Tōkyō area): [TEPCO](http://www.tepco.co.jp/forecast/html/images/juyo-j.csv)
-- Kuwait: [Ministry of Electricity & Water](https://www.mew.gov.kw/en/)  
+- Kuwait: [Ministry of Electricity & Water](https://www.mew.gov.kw/en/) or [GCCIA](https://www.gccia.com.sa/)
```
robertahunt, line 146: Just to be more consistent with what we have done for other countries
```suggestion
- Kuwait (TSO): [Ministry of Electricity & Water](https://www.mew.gov.kw/en/)
- Kuwait (Power Market): [GCCIA](https://www.gccia.com.sa/)
```

robertahunt:
@q-- Thanks for this!

For the zones.json, It would be great if you could add everything except the bounding box - we have a javascript code that can automatically generate the bounding boxes - you can run this yourself if you like, or we can - it is under web/generate-zone-bounding-boxes.

What I think you should add for each of the zones is something along the lines of:
>   "AE": {
>     "contributors": [
>       "https://github.com/q--"
>     ],
>     "parsers": {
>       "production": "GCCIA.fetch_production"
>     },
>     "timezone": "Asia/Dubai"
>   },

The config/zones.json is arranged in alphabetical order.

robertahunt:
Hey! Sorry, looking at the website, they say these numbers are demand? so this should be consumption, correct? I'll make some suggestions in your code to change this.

config/zones.json
```diff
@@ -12,6 +12,15 @@
     ],
     "timezone": null
   },
+  "AE": {
+    "contributors": [
+      "https://github.com/q--"
+    ],
+    "parsers": {
+      "production": "GCCIA.fetch_production"
```
robertahunt, line 20:
```suggestion
      "consumption": "GCCIA.fetch_consumption"
```

config/zones.json
```diff
@@ -493,6 +502,15 @@
     },
     "timezone": null
   },
+  "BH": {
+    "contributors": [
+      "https://github.com/q--"
+    ],
+    "parsers": {
+      "production": "GCCIA.fetch_production"
```
robertahunt, line 510:
```suggestion
      "consumption": "GCCIA.fetch_consumption"
```

config/zones.json
```diff
@@ -3560,6 +3578,15 @@
     },
     "timezone": null
   },
+  "OM": {
+    "contributors": [
+      "https://github.com/q--"
+    ],
+    "parsers": {
+      "production": "GCCIA.fetch_production"
```
robertahunt, line 3586:
```suggestion
      "consumption": "GCCIA.fetch_consumption"
```

config/zones.json
```diff
@@ -3712,6 +3739,15 @@
     "flag_file_name": "pt.png",
     "timezone": null
   },
+  "QA": {
+    "contributors": [
+      "https://github.com/q--"
+    ],
+    "parsers": {
+      "production": "GCCIA.fetch_production"
```
robertahunt, line 3747:
```suggestion
      "consumption": "GCCIA.fetch_consumption"
```

config/zones.json
```diff
@@ -3923,6 +3959,15 @@
     "flag_file_name": "ru.png",
     "timezone": null
   },
+  "SA": {
+    "contributors": [
+      "https://github.com/q--"
+    ],
+    "parsers": {
+      "production": "GCCIA.fetch_production"
```
robertahunt, line 3967:
```suggestion
      "consumption": "GCCIA.fetch_consumption"
```

parsers/GCCIA.py
```diff
@@ -0,0 +1,67 @@
...
+TIME_ZONE_MAPPING = {
+  'AE': 'Asia/Dubai',
+  'BH': 'Asia/Bahrain',
+  'KW': 'Asia/Kuwait',
+  'OM': 'Asia/Muscat',
+  'QA': 'Asia/Qatar',
+  'SA': 'Asia/Riyadh',
+}
+
+def fetch_production(zone_key, session=None, logger=None):
```
robertahunt, line 35:
```suggestion
def fetch_consumption(zone_key, session=None, logger=None):
```

parsers/GCCIA.py
```diff
@@ -0,0 +1,67 @@
...
+
+    load = re.findall(pattern, response.text)
+    load = int(load[0])
+    production = {}
+    production['unknown'] = load
+
+    datapoint = {
+        'zoneKey': zone_key,
+        'datetime': arrow.now(TIME_ZONE_MAPPING[zone_key]).datetime,
+        'production': production,
```
robertahunt, line 50:
```suggestion
        'consumption': consumption,
```

parsers/GCCIA.py
```diff
@@ -0,0 +1,67 @@
...
+        'source': 'www.gccia.com.sa' # URL won't work without WWW
+    }
+
+    return datapoint
+
+if __name__ == '__main__':
+    """Main method, never used by the electricityMap backend, but handy for testing."""
+
+    for i in COUNTRY_CODE_MAPPING:
+      print('fetch_production(\'{0}\') ->'.format(i))
```
robertahunt, line 60:
```suggestion
      print('fetch_consumption(\'{0}\') ->'.format(i))
```

q--:
The KW.py parser, which I used as a starting point, might also need to be changed, then, as it also names it `fetch_production`

I was aware that I was actually dealing with consumption here, but because the Kuwait parser did that I thought ElectricityMap might not support it when there's only fetch_consumption and not fetch_production

parsers/GCCIA.py
q--: I adapted your suggestions, also changed a few more places that you'd missed

robertahunt:
Thanks for flagging that @q-- I will have a look at KW.py

q--:
As previously said, 

>This pull request doesn't take care of adding the right data to config/co2eq_parameters.json and

adding the bounding boxes to 

>config/zones.json

Should I open a new issue for that?

PR #2366, opened 2020-04-09
Updated bounding boxes for zones

robertahunt:
related - #2360
Should have been done there, my mistake.


robertahunt:
generated using generate-zone-bounding-boxes

PR #2367, opened 2020-04-09
Parser for ISEP historic data for Japan

robertahunt:
First crack at parsing Japan's historic data from https://isep-energychart.com/en/graphics/electricityproduction/?region=tokyo&period_year=2019&period_month=2&period_day=1&period_length=2days&display_format=plie_up 

I wouldn't replace our production parser with this one, but instead use this to repopulate historic data since the data is much higher quality than we can get live.

There is some trouble in that the live numbers should have different co2eq for 'unknown' from this - we should consider what to do there.

FYI @tth4vb


robertahunt:
It is also annoying that the only way I managed to get the data is by actually downloading the xls file, then reading it. If anyone has a better method, that would be great!

The trouble is the site runs javascript to generate the downloaded file instead of just linking to a url, and I couldn't figure out how to directly download the file's data.

corradio:
Agreed - if we could avoid using Selenium that'd be great, as the memory usage will be much much smaller if we do without it.

robertahunt:
Thanks @jarek! That was a great catch. I changed it to parse from the html now.

parsers/JP_ISEP.py
```diff
@@ -0,0 +1,239 @@
...
+   'interconnection':'exchanges',
+   'solar_suppression': 'solar curtailment',
+   'wind_suppression': 'wind curtailment'
+}
+
+def get_data(region, year, month, day):
+    r = requests.get(url.format(**{'region':region,
+                                    'year':year,
+                                    'month':month,
+                                    'day':day}))
```
jarek, line 43: I'm not sure if you know - this can be done a little shorter without having to go via the dict:

```suggestion
    r = requests.get(url.format(region=region,
                                year=year,
                                month=month,
                                day=day))
```

robertahunt: Didn't know that, thanks!

Issue #2203, opened 2020-02-11
Spanish mainland data from REE


pdunamuno:
We haven't found the data for this region yet, but that doesn't mean you can't make it happen!

Ranked from best to useful information, here is what we could use:
- real-time data of energy generation by energy source
- historical data of energy generation by energy source (as recent and as granular as possible)
- contact details (contact form, email or social media) of someone who may have access to that data
- contact details (contact form, email or social media) of an organisation which may have access to that data

pdunamuno:
Could you change the source of the data to https://demanda.ree.es/visiona/peninsula/demanda/total it is more accurate than the one that is now

jarek:
Can you briefly describe in which ways is it more accurate than the data we have which comes from ENTSOE?

pdunamuno:
The information in ENTSOE updates every hour and creates in the graph of electricity production like a stair boxines, instead if you use REE it updates more frecuently and there is not that stair looking graph.
In Mallorca electricity map uses REE and doesn´t has the stair looking graph.
thank you for answering.

liacds:
Hi, can I work on this issue?

jarek:
Yes that should be fine! You can use the current implementation of Balearic and/or Canary Islands as a template since they already use REE data.

corradio:
As discussed in https://github.com/tmrowco/electricitymap-contrib/pull/2370#issuecomment-640562408 we will need a PR with both real-time *and* historical data in order to be able to replace ENTSOE.
Work was already done here by @liacds https://github.com/tmrowco/electricitymap-contrib/pull/2370 so it would be great to re-use it.

CheneyZeng:
Can I work on adding historical data? @corradio @pierresegonne

Kongkille:
It seems the dashboard is still live and has quite a long historical range :)

VIKTORVAV99:
I'm going to close this as the `ES.py` parser now support parsing ES mainland data.

If anyone feels it should be re-opened let me know.

PR #2370, opened 2020-04-11
Spanish mainland data from REE #2203: In progress

liacds:
Hi, I think I did most of the job, but I have some questions:
1. is there a convention for creating zone_keys? 
2. to make this parser I looked at ES_IB and ES_CI and they found oil differently and I am a bit confused which one is right

### DIFF
--- a/parsers/ES_P.py
+++ b/parsers/ES_P.py
@@ -1,0 +1,37 @@
+import logging
+
+# The arrow library is used to handle datetimes
+from arrow import get
+# The request library is used to fetch content through HTTP
+from requests import Session
+from ree import (IberianPeninsula)
+from .lib.exceptions import ParserException
+from parsers.lib.validation import validate
+
+FLOORS = {
+    'ES-P': 0,
+}
+
+
+def fetch_consumption(zone_key='ES_P', session=None, target_datetime=None, logger=None):
+    ses = session or Session()
+    IberianPeninsula_data = IberianPeninsula(ses, verify=False).get()
+    if not IberianPeninsula_data:
+        raise ParserException(zone_key, "IberianPeninsula not response")
+    data = []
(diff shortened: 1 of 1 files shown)


robertahunt:
Hi @liacds , Sorry about the late reply getting back to you. There isn't one defined method of creating zone keys, so I think calling it ES-P has you have done, is alright.

This is data for mainland spain, is that correct? We currently get data from ENTSOE - I have been trying to find time to dig into the differences between these two datasources, but it is difficult. If you have time to quantify the difference between this data and what we get from ENTSOE, that would be great!

Otherwise, it is always nice to have extra parsers, but we wouldn't switch to a new one without a quantifiable reason.

martincollignon:
Hi @liacds would you have time to look at the differences between the two data sources?

liacds:
Hi, sorry for not responding earlier. I haven't looked at differences yet, but I can do it this week.

corradio:
Hi @liacds ,
Unfortunately we need parsers that are able to fetch historical data (like ENTSOE's)
Do you know if REE allows that?

liacds:
Hi @corradio, 
https://github.com/tmrowco/electricitymap-contrib/issues/2203 this is the issue that I worked on and it says it does.

corradio:
> Hi @corradio,
> #2203 this is the issue that I worked on and it says it does.

In that case, are you able to edit the parser to fetch historical data as well? Then we will be able to gather a sample historical dataset of REE and ENTSOE to compare them to double check the data quality of REE is superior!
Thanks,

Olivier

liacds:
No, sorry I don't want to do that

corradio:
In this case I will unfortunately have to close the PR. I hope someone else can add the historical part and build on your work!

Olivier

PR #2381, opened 2020-04-17
Add CR Capcity + Source

nessie2013:
Also calculated "unknown" carbon intensity


nessie2013:
hmm i cant see why its throwing up errors, its formatted the same as the rest

config/co2eq_parameters.json
```diff
@@ -213,6 +213,11 @@
         "hydro discharge": {
           "source": "2017 average by Tomorrow",
           "value": 52.16941300335089
+        },
+        "unknown": {
+          "_comment": "Source https://www.grupoice.com/wps/wcm/connect/579dfc1f-5156-41e0-807d-d6808f65d718/Fasciculo_Electricidad_2020_ingl%C3%A9s_compressed.pdf?MOD=AJPERES&CVID=m.pGzcp",
+          "source": "assumes 58% thermal (coal, gas, or oil), 42% solar and biomass",
+          "value": 463.75
```
robertahunt, line 220: @nessie2013 could you elaborate a bit more here? I couldn't find exactly where this value comes from in the document and it seems suspicious that you are assuming 58% thermal when there is no coal, oil or gas capacity.

robertahunt:
Otherwise it looks good! Thanks!

config/co2eq_parameters.json
nessie2013: on page 7 of the source it gives this information 
image
The "unknown" capacity is the "biomass and solar" and "thermal" combined. Using the 3.5 GW given as the total capacity of all sources in the country, I used the percentages to find there is 29MW of Biomass/Solar and 40MW of thermal capacity. This means Thermal accounts for ~58% of the total unknown capacity.

PR #2386, opened 2020-04-17
Dominican Republic Capacity and other updates

nessie2013:
Updated Capacities from new, up to date source.

Updated some generation from source (https://www.dropbox.com/sh/8dec0z1ftf2nqr0/AAAbDG-6lSttxBprxhPRNkjaa/2020?dl=0&preview=OC-2020-000190-GG-SIE-INFORME+ANUAL+2019+V1.pdf&subfolder_nav_tracking=1), some locations listed on this source look like they are not included in this file though. (changes also fixes gas being too high compared to the updated capacity)


README.md
```diff
@@ -236,7 +236,7 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
 - Denmark (DK1 and DK2): [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
 - Denmark (Bornholm)
   - Wind: [stateofgreen.com](https://stateofgreen.com/en/profiles/regional-municipality-of-bornholm/solutions/kalby-wind-turbines)
-- Dominican Republic: [Climatescope](http://global-climatescope.org/en/country/dominican-republic/#/details)
```
robertahunt, line 239: Can we keep this reference in place? You can add the new one below with a tagline, just because this one will be updated whereas the dropbox link is static, and has more potential to be invalid in the future

parsers/DO.py
```diff
@@ -65,13 +65,13 @@
...
                  u'QUISQUEYA 1': 'gas',
-                 u'QUISQUEYA 2': 'gas',
+                 u'QUISQUEYA 2': 'oil',
                  u'QUISQUEYA 1 SAN PEDRO': 'oil',
                  u'RIO SAN JUAN': 'oil',
                  u'SAN FELIPE': 'oil',
                  u'SAN FELIPE CC': 'gas',
                  u'SAN FELIPE VAP': 'oil',
-                 u'SAN LORENZO 1': 'gas',
+                 u'SAN LORENZO 1': 'oil',
```
robertahunt, line 74: Could you explain this change? I found this which suggests to me that it is a gas plant https://www.power-technology.com/projects/sanlorenzo/

config/zones.json
```diff
@@ -1357,15 +1357,16 @@
...
-      "hydro": 610,
-      "oil": 2200,
-      "solar": 3,
-      "wind": 85
+      "coal": 1094,
+      "gas": 678,
+      "hydro": 623,
+      "oil": 1938,
+      "solar": 188,
+      "wind": 370
```
robertahunt, line 1365: These look reasonable to me (just so you know I did a quick check against the UN data which might be helpful to you http://data.un.org/Data.aspx?d=EDATA&f=cmID%3AEC), but could you point to what page in the document these are referenced from? It is a large jump in coal capacity, so I want to double check.

nessie2013: Pages 26-28, I just recalculated it and got a slightly smaller capacity so will change the file with the new capacity.

parsers/DO.py
nessie2013: Looks like that was a mistake, sorry about that.

robertahunt:
Nice, thanks!!

README.md
brandongalbraith: @nessie2013 The following URL can be used as a permanent reference for the Dropbox file link your previously provided: https://archive.org/details/oc-2020-000190-gg-sie-informe-anual-2019-v-1

PR #2387, opened 2020-04-18
Add Italy Sardinia Generation Capacity

nessie2013:
found a new source for regional data, that gives the 2018 values for wind, solar, geothermal, wind and biomass, Entsoe is used to get other types. The data is a little out of date but as there is no data currently for the regional capacity i thought I could add this for now


README.md
```diff
@@ -261,8 +261,8 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
   - Biomass, Solar & Wind: [IRENA](http://resourceirena.irena.org/gateway/countrySearch/?countryCode=IRL)
 - Israel: [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)  
 - Italy
-  - Wind & Solar: [IRENA](http://resourceirena.irena.org/gateway/countrySearch/?countryCode=ITA)
```
robertahunt, line 264: I think this should stay, at least until all the individual regions have associated capacities

config/zones.json
```diff
@@ -2695,8 +2695,31 @@
...
+      "biomass": 1104,
+      "coal": 979, 
+      "geothermal": 0,
+      "hydro": 154,
+      "hydro-storage": 312,
+      "nuclear": 0,
+      "oil": 238,
+      "solar": 787,
+      "wind": 1054,
+      "unknown": 703
```
robertahunt, line 2718: On the website you link to, I can only find thermal, hydro, geothermal, wind and solar capacity. Could you explain where these numbers comes from?

nessie2013: I took the hydro, solar and wind and biomass numbers from Terna, then I used the Entsoe "installed capacity per production unit" source to find the capacity of the installed coal/oil/gas and any unknown capacity, I then also used the hydro capacity from Terna and took the hydro-storage from entsoe (taking the storage capacity out of the given hydro capacity) to get both hydro and hydro-storage.

Issue #2310, opened 2020-03-24
Move layer components to React


fbarl:
Follow-up on https://github.com/tmrowco/electricitymap-contrib/issues/1681#issuecomment-572014604.

Moving layers to React and having them managed in one place in the DOM tree would be simpler and might help us optimizing their render performance.

#### Layers

* [ ] ExchangeLayer
* [ ] SolarLayer
* [ ] WindLayer

PR #2388, opened 2020-04-18
Use Redux Saga for all API calls

### LINKED ISSUE #2384: Refactor UI data fetching logic
The code that fetches data from the API has become a little bit messy over time and it doesn't fit that well in the React + Redux paradigm that we've recently been transitioning to.

We should manage all calls to backend idiomatically and try to split it away from the rendering logic as much as possible - a good way to go might be to move it all to [Redux Sagas](https://redux-saga.js.org/).

fbarl:
Will resolve #2384.

**Note:** #2385 needs to be merged first as this PR branches off from that one.


web/src/components/layers/solar.js
```diff
@@ -102,20 +102,20 @@ class SolarLayer {
...
   }
 
   hide() {
+    if (this.hidden) return;
     if (this.canvas) {
-      d3.select(this.canvas).transition().style('opacity', 0)
-        .on('end', function() {
-          d3.select(this).style('display', 'none');
-        });
+      this.canvas.style.opacity = 0;
```
corradio, line 115: any reason to remove the transition?

web/src/helpers/api.js
```diff
@@ -0,0 +1,86 @@
...
+      } else if (!res) {
+        reject(new Error(`Empty response received for ${url}`));
+      } else {
+        resolve(res);
+      }
+    });
+  });
+}
+
+function trackError(err, appState) {
```
corradio, line 62: this should probably be in a different helper?

web/src/helpers/api.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export function handleConnectionReturnCode(err, appState = {}) {
+  if (err) {
+    if (err.target) {
+      // Avoid catching HTTPError 0
+      // The error will be empty, and we can't catch any more info
+      // for security purposes
+      // See http://stackoverflow.com/questions/4844643/is-it-possible-to-trap-cors-errors
+      if (err.target.status) {
```
corradio, line 76: can we also make sure not to catch errors 5xx? They appear a lot on Sentry as soon as the server experiences some downtime.

web/src/helpers/api.js
fbarl: Good catch, I moved it into `thirdPartyServices` in https://github.com/tmrowco/electricitymap-contrib/pull/2388/commits/dbd8406a638090ef773ee5f480d939da8e7b9b1d which also simplified the sagas a bit since the app state didn't need to be carried across :+1:

web/src/helpers/api.js
fbarl: Yes, it makes sense not to track them - I made the change in https://github.com/tmrowco/electricitymap-contrib/pull/2388/commits/507190a5fb9d7d1c228d91e96daaa855a01dee0b not to track them anymore and I also added tracking for zone history fetches in that commit which seems to have been missing from the code so far :thinking:

web/src/components/layers/solar.js
fbarl: Yes - due to the changes in polling logic from this PR, `hide()` and `show()` might now be called initially very soon after the app loads, sometimes before the layer is finished initializing, which would occasionally result in `d3.select(this.canvas).transition` being undefined, so I changed it here to CSS transitions ([see here](https://github.com/tmrowco/electricitymap-contrib/pull/2388/files/637bc3a7888cce7bdd9d9f070aceda52de314fed#diff-79e2defbaeeca7e8e55cd2a14fdb2f30R41)) which always work.

I agree the solution is kinda hacky, especially given the `setTimeout` below, but I'm going to start reworking this piece of code today as a part of https://github.com/tmrowco/electricitymap-contrib/issues/2310 so I think this should be an ok temporary compromise.

corradio: ok thx!

web/src/sagas/index.js
```diff
@@ -0,0 +1,85 @@
...
+  } catch (error) {
+    yield put({ type: 'ZONE_HISTORY_FETCH_FAILED', error });
+  }
+}
+
+function* fetchGridData(action) {
+  const { datetime } = action.payload;
+  try {
+    const payload = yield call(protectedJsonRequest, datetime ? `/v3/state?datetime=${datetime}` : '/v3/state');
+    thirdPartyServices.trackWithCurrentApplicationState('pageview');
```
fbarl, line 35: @corradio I kept this line here since it was there before, but I'm not sure it really belongs here - was the intention to track page views on every grid data fetch? We already do it on every route change - [see here](https://github.com/tmrowco/electricitymap-contrib/pull/2388/files/637bc3a7888cce7bdd9d9f070aceda52de314fed#diff-42c3276bc26032ecad9d7061dd401cd4R474).

corradio: yes I think the purpose was to track a pageview when the data refreshes (some people have TVs which are left running with emap)

fbarl: Ok thanks! Maybe it's the name `pageview` that confused me as it immediatelly made me assume the event is triggered by a user action. Maybe it's something worth discussing in general, but I won't touch it here then! :)

fbarl:
@corradio PTAL if you want - I just pushed a couple more commits that clean up the loading states and transitions:

* Created a component for the main loading overlay which now fades again smoothly with help of `react-transition` library
* Made the left panel content more stable when transitioning between loading <-> loaded state, mainly by using loaders as graph placeholders instead of overlays

corradio:
LGTM!

PR #2393, opened 2020-04-21
Historical view

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2114

This is for now just a backup of my internal branch. I've been able to work on this during internal hackdays :)


corradio:
@fbarl are you able to review? Still not ready to merge but it's getting very very close :)

web/src/sagas/index.js
```diff
@@ -25,9 +27,13 @@ function* fetchClientVersion() {
 }
 
 function* fetchZoneHistory(action) {
-  const { zoneId } = action.payload;
+  const { zoneId, timescale } = action.payload;
   try {
-    const payload = yield call(protectedJsonRequest, `/v3/history?countryCode=${zoneId}`);
+    let path = `/v3/history?countryCode=${zoneId}`;
+    if (timescale !== TIMESCALE.LIVE) {
+      path += `&timescale=${timescale}`;
```
fbarl, line 34: I see that the `TIMESCALE.MONTHLY` constant gets translated to `MONTHLY` instead of a duration format when making the API call: `/history?countryCode=IE&timescale=MONTHLY`

Wouldn't it be nicer if the backend could accept any integer duration in seconds (or at least duration string formats like `5y`, `10d`, `24h`) to make it work with more general durations?

If that seems over-engineered, I think something like `timescale=5y&step=1mo` / `timescale=24h&step=15m` would still be a nicer API (even if those were the only two supported options) :slightly_smiling_face:

web/src/sagas/index.js
```diff
@@ -36,9 +42,13 @@ function* fetchZoneHistory(action) {
 }
 
 function* fetchGridData(action) {
-  const { datetime } = action.payload;
+  const { datetime, timescale } = action.payload;
   try {
-    const payload = yield call(protectedJsonRequest, datetime ? `/v3/state?datetime=${datetime}` : '/v3/state');
+    let path = datetime ? `/v3/state?datetime=${datetime}` : '/v3/state';
+    if (timescale !== TIMESCALE.LIVE) {
+      path += `?timescale=${timescale}_${moment().startOf('month').subtract(1, 'month').format('YYYYMM')}`;
```
fbarl, line 49: Why is the `timescale` value necessary here (given that the `datetime` info is already passed)? :thinking: 

If this is to tell the backend to return the values averaged over a month, perhaps I'd prefer to go with something like `datetime=2020-07-01T00:00:00Z&avg_over=1mo`, but I'm not so sure myself :smile:

fbarl: By the way, this endpoint doesn't really work for me (while the history one does), I get an empty map and this message from `state?timescale=MONTHLY_202006`:

Screenshot from 2020-07-13 17-06-44

web/src/reducers/dataReducer.js
```diff
@@ -96,7 +115,10 @@ module.exports = (state = initialDataState, action) => {
         newGrid.zones[key] = zone;
       });
       Object.keys(newGrid.exchanges).forEach((key) => {
-        newGrid.exchanges[key].netFlow = undefined;
+        const exchange = Object.assign({}, newGrid.exchanges[key]);
```
fbarl, line 118: Good catch, these can be very dangerous! :grimacing: 

By the way, in ES6 this can be written idiomatically as `const exchange = { ...newGrid.exchanges[key] };`

web/src/reducers/dataReducer.js
```diff
@@ -53,6 +61,15 @@ const initialDataState = {
...
 module.exports = (state = initialDataState, action) => {
   switch (action.type) {
+    case 'APPLICATION_STATE_UPDATE': {
+      // Reset histories if timescale changes
+      const { key } = action;
+      if (key === 'timescale') {
+        return { ...state, histories: {} };
+      }
+      return state;
+    }
```
fbarl, line 71: This looks a little bit like an anti-pattern - I think it would make sense in this case to create a separate action `TOGGLE_TIMESCALE` that would both reset the histories and set the `timescale` key to prevent more logic piling up under the general `APPLICATION_STATE_UPDATE` action.

web/src/reducers/dataReducer.js
```diff
@@ -1,13 +1,21 @@
...
 const zonesConfig = require('../../../config/zones.json');
 
+function computeZoneAggregates(zone) {
+  zone.maxProduction = Math.max(...Object.values(zone.production || {}));
+  zone.maxDischarge = -Math.min(Math.min(...Object.values(zone.storage || {})), 0);
+  zone.maxStorage = Math.max(Math.max(...Object.values(zone.storage || {})), 0);
+  zone.maxExport = -Math.min(Math.min(...Object.values(zone.exchange || {})), 0);
+  zone.maxImport = Math.max(Math.max(...Object.values(zone.exchange || {})), 0);
+  return zone;
+}
```
fbarl, line 17: This logic looks quite different now (in particular the console warnings have been removed), what's the reasoning there? :slightly_smiling_face:

web/src/sagas/index.js
corradio: For your first question, indeed it is to tell the backend that the values should be averaged over 1 month.
For the second, it is because the backend is currently not updating itself by populating the monthly averaged values.
We'll need to do a bit more work to get there :)

web/src/reducers/dataReducer.js
corradio: Indeed I agree. However, I was hesitant to add a new action as this would mean there would be two ways to update the application state: 1/ APPLICATION_STATE_UPDATE 2/ TOGGLE_TIMESCALE.
However, only one of those two would have the right logic.

Any ideas?

web/src/components/countrytable.js
```diff
@@ -39,12 +41,15 @@ const RECT_OPACITY = 0.8;
...
 const SCALE_TICKS = 4;
 
-const getProductionData = data => modeOrder.map((mode) => {
+const getProductionData = (data, timescale) => modeOrder.map((mode) => {
   const isStorage = mode.indexOf('storage') !== -1;
   const resource = mode.replace(' storage', '');
 
   // Power in MW
-  const capacity = (data.capacity || {})[mode];
+  // Do not show capacity if timescale != live
```
fbarl, line 49: Interesting.. Maybe this is not such an obvious choice (the alternative being showing the monthly averages for capacities), but I agree it's better to remove it than risk wrong interpretation of the capacity value being lower than production :slightly_smiling_face:

web/src/components/graph/timeaxis.js
```diff
@@ -1,35 +1,40 @@
...
+    format = 'MMM YYYY';
+  } else {
+    interval = 'hour';
+    format = 'LT';
+  }
+  const range = moment.range(
+    startMoment.startOf(interval).add(1, interval), // add one at the end to make sure it includes NOW
+    endMoment,
+  );
+  const tickMoments = Array.from(range.by(interval, { step: interval === 'hour' ? 6 : 1 }));
```
fbarl, line 36: Given how prone to errors / imprecisions managing scale ticks generally tends to be, it'd be great if we could keep this logic for generating `tickMoments` somewhat extracted and _unit tested_. Probably out of scope of this PR, but maybe you could at least add a `TODO` comment here above :wink:

web/src/helpers/zonedata.js
```diff
@@ -34,18 +34,10 @@ export function getProductionCo2Intensity(mode, zoneData) {
...
 
-  return exchange > 0
-    ? (
-      exchangeCo2Intensity
-    ) : (
-      electricityMixMode === 'consumption'
-        ? zoneData.co2intensity
-        : zoneData.co2intensityProduction
-    );
+  return exchangeCo2Intensity;
```
fbarl, line 40: I'm happy for this simplification! :+1: 

I didn't quite understand this code when I was refactoring so I decided not to touch this logic before :smile:

web/src/layout/main.js
```diff
@@ -98,8 +104,28 @@ const Main = ({
...
+                onChange={value => dispatchApplication('timescale', value)}
+                options={[TIMESCALE.MONTHLY, TIMESCALE.LIVE].map((k) => {
+                  if (k === TIMESCALE.MONTHLY) {
+                    return { value: k, label: 'monthly historical' };
+                  }
+                  if (k === TIMESCALE.LIVE) {
+                    return { value: k, label: 'live' };
+                  }
+                  return { value: 'unknown', label: 'unknown' };
+                })}
```
fbarl, line 117: Why not simply `options={{ value: TIMESCALE.MONTHLY, label 'monthly historical' }, { value: TIMESCALE.LIVE, label: 'live }}`?

web/src/layout/main.js
```diff
@@ -98,8 +104,28 @@ const Main = ({
...
+                value={timescale}
+              />
             </div>
             <LayerButtons />
+            {
+              timescale !== TIMESCALE.LIVE ? (
+                <div className="text-title" style={{ color: brightModeEnabled ? '#000' : '#fff' }}>
```
fbarl, line 124: Could we add `textColor` to the theme and reuse it here instead?

web/src/layout/main.js
```diff
@@ -98,8 +104,28 @@ const Main = ({
...
+                  {moment(currentDate).format('MMMM YYYY')}
+                </div>
```
fbarl, line 126: Something to consider design-wise is a case of connection error in the monthly view, although I'm fine living it as it is:

Screenshot from 2020-07-13 18-23-41

web/src/reducers/dataReducer.js
fbarl: > Any ideas?

Yes, an alternative solution might be to add a new action `RESET_HISTORIES` here that would have to be called explicitly together with setting the `timescale` or alternatively create a `dispatchData` helper modelled by `dispatchApplication` for the same effect.

I know that these proposals make calling these actions a bit more complicated (two calls instead of one), but I think it pays better in a longer run to keep the reducers simple vs. branching into the cases in their logic.

Not a critical change though :stuck_out_tongue:

corradio:
Note to self: there's a bug where the interconnectors on the historical mix graph only appear when the slider is moved.

skovhus:
We are planning on re-iterating on showing historical data at a later point.

PR #2394, opened 2020-04-21
Use unicode subscripts for CO2 wherever possible

fbarl:
Motivated by @alixunderplatz reporting yet another case of `<span class="sub">2</span>` rendering in the UI because we missed a `co2Sub` helper, I thought it would be good to prevent this happening once and for all :slightly_smiling_face: 

Turns out there is actually a unicode subscript character "₂" which I used to replace `CO2` -> `CO₂` everywhere in the code and in all the translations, which enabled me to get rid of `co2Sub` helper altogether.

I took that one step further and I removed all the `dangerouslySetInnerHTML` uses which were there so far just because of the `co2Sub` - turns out it was most of them!


web/src/components/faq.js
```diff
@@ -76,7 +75,7 @@ const QuestionAnswer = ({
       <div
         className="answer"
         dangerouslySetInnerHTML={{
-          __html: co2Sub(__(`${groupKey}.${entryKey}-answer`)),
+          __html: __(`${groupKey}.${entryKey}-answer`),
```
corradio, line 78: no need for dangerouslySetInnerHTML?

fbarl: I wasn't so confident about removing it in the FAQ section as different translations might contain links (and afaik some of them do).

magol:
@fbarl I was the one who invented the co2Sub help function from the beginning, but I have never liked the solution. I first wanted to do the same solution as in this PR, but due to SEO I had to do the workaround with the help routine, see PR #635. I love that it is now solved as I intended to do it from the beginning and I hope that SEO not is a problem.

Issue #2311, opened 2020-03-24
Remove all DOM management from main.js


fbarl:
Once all the components are moved to React, there won't be need to manage DOM externally and `main.js` could ideally be made to contain only the React mounting point and some general initializers (e.g. Cordova app).

Depends on #2309 and #2310.

Issue #1590, opened 2018-08-31
Arrows are slowing down the map


corradio:
I've noticed that removing arrows speed up the map significantly, especially when the wind is present.
Maybe we should consider drawing them in canvas, or in webgl

systemcatch:
This includes the import/export arrows right? Maybe removing the animation could help?

corradio:
It would, but it would also make me sad :(

corradio:
Every arrow is a different GIF. Arrows are hidden/shown on zoom/drag. I
don't know of other formats that are universally supported.
We tried with SVG but that was slower due to CPU-bound animations.
I think an alternative could be to animate in canvas.

systemcatch:
Maybe we could reduce it to one object that just gets properties (rotation, colour, pulse), or possibly use CSS sprites in some way (sorry if I'm going over stuff you've already thought of, I'm just throwing out random ideas).

By canvas you mean the way countries are drawn?

corradio:
Countries are drawn using WebGL, but the wind animation is actually drawn
using canvas. We could do the same with arrows.
I don't know about CSS sprites or about how we would just have on object.
Are you able to investigate? That could be good ideas.

Olivier

systemcatch:
@corradio I've thought about this again.

Using [this article](https://moz.com/blog/how-to-improve-site-performance-using-gifs) as a reference I converted all of the arrows from gifs to html5 videos (mp4) with the following command.

```shell
for i in *.gif; do ffmpeg -i "$i" -movflags faststart -pix_fmt yuv420p -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" "${i%.*}.mp4"; done
```

Total size of all arrows dropped from 653KB to 185KB. There's likely some performance gains in the browser as well.

corradio:
That's pretty cool! Do you see any differences in perf in the browser? I'm
not sure the size changes a lot, but it shouldn't hurt!
The way I'm testing it is that I activate the wind layer, and see how
smooth it runs. Without arrows it runs very smooth on my laptop, and with
arrows it lags.

systemcatch:
I haven't had a chance to run it yet, it'll need some modification of the html tags I guess (any pointers for where to do that?).

I'll try testing with the wind layer once I get it working.

corradio:
I think there's a map layer called exchangelayer.js or something like that.
If you search for `img` or `gif` you should be able to see where it
generates the tag.

[...]

Issue #944, opened 2017-12-23
Investigate webgl-based wind layer


corradio:
Saw this demo from the mapbox team:
https://mapbox.github.io/webgl-wind/demo/

Quite impressive! Doesn't seem to work on my old iPad though :/

Another fork was done here: https://github.com/astrosat/windgl

We could implement as a custom mapbox layer:
https://github.com/mapbox/mapbox-gl-js/pull/7039

scriptator:
I think that's a great idea as the current wind visualization seems to be quite CPU intensive. I just tested the mapbox demo and it works on my iPad Air 2, which is 4,5 years old.

Kongkille:
I think this would be a great potential improvement to look further into. However since there's been no activity so far, I'm closing the issue :)

Issue #923, opened 2017-12-16
Sun layer: Use a GPU-based gaussian blur (CSS / webgl) instead of CPU-based


corradio:
Another idea is to use `createRadialGradient` to speed up the canvas drawing time.

corradio:
CSS blur filter not supported in IE11

Kongkille:
This seems like a good potential improvements, however since no ones has looked at it in a while, I am closing this issue. We can pick it up again, if we run into performance issues with the current sun layer :)

PR #2395, opened 2020-04-22
Migrate layers to React

### LINKED ISSUE #2310: Move layer components to React

fbarl:
Resolves #2310.

This PR migrates all 3 map layers to React (to various degrees), removing further parts of DOM management from `main.js`, bringing us one step closer in addressing https://github.com/tmrowco/electricitymap-contrib/issues/2311, leaving only https://github.com/tmrowco/electricitymap-contrib/issues/2309 as the final step to complete it.

As an intermediary step, I made `zoneMap` a global object that is consumed by all the layers for getting the projection maps and zoom levels. As we address https://github.com/tmrowco/electricitymap-contrib/issues/2309, we'll be able to get rid of the global object in favor of passing props if we embed layers as children of the map component.

##### Changes

* Fully migrated to the React component
* Moved `MapExchangeTooltip` into the layer, bringing #2288 to near completion
* Created `exchangeQuantizedIntensityScale` with explicit carbon steps that map into the corresponding arrow images

##### Didn't do

* The map zooming is not working now when hovering over exchange arrows as the `ExchangeLayer` is not mapped deep enough in the map DOM anymore so the scroll events don't bubble up. I don't know of an elegant way to fix this before addressing https://github.com/tmrowco/electricitymap-contrib/issues/2309, so I'd fix it there if that's ok.
* ~Decided not to hide or stop the arrows animation when _dragging_ the map as the performance seems good even without it (the arrows are still hidden when they are outside of the viewport)~
* Not sure if performance is still a problem (see https://github.com/tmrowco/electricitymap-contrib/issues/1590), I didn't experience any lags on my machine

##### Changes

* Fully migrated to React (keeping only some canvas image filter helpers outside)
* Different behavior when dragging on the map - instead of caching the canvas state, the layer smoothly fades out/in when the map dragging starts/ends
* Cut down on a few steps of canvas image transformation, making it refresh a bit faster
* Using `solarColor` scale directly when generating the layer canvas image

##### Didn't do

* Using WebGL here would probably make this _a lot_ faster, probably recalculation could be done with a high FPS so we wouldn't even need to _freeze_ it during map dragging + we'd get the blur effect out of the box instead of depending on that lengthy helper - https://github.com/tmrowco/electricitymap-contrib/issues/923

##### Changes

* Moved the layer to React, still keeping it as a wrapper around _Windy_ (which still makes it a bit slow and clunky)
* Different behavior when dragging on the map - instead of caching the canvas state, the layer smoothly fades out/in when the map dragging starts/ends

##### Didn't do

* Using WebGL here would probably make this _a lot_ faster, probably recalculation could be done with a high FPS so we wouldn't even need to _freeze_ it during map dragging - we should probably merge `Windy` into the wind layer component as we migrate it to WebGL - https://github.com/tmrowco/electricitymap-contrib/issues/944


corradio:
Hi @fbarl,

Thanks for this! Regarding performance, have you tested on very old devices?
It might be fast on your laptop but when I tested with old iPhones / iPads I had some lags especially during dragging. I think it would make sense to keep hiding the arrows while dragging if it is not too much work.

web/src/helpers/image.js
```diff
@@ -0,0 +1,175 @@
+/* eslint-disable */
```
corradio, line 1: I would put this in the solar component folder so we keep the solar component self-contained?

fbarl: I can do it if you want, but beware that `WindLayer` also has an external dependency `windy.js` and e.g. `AreaGraph` imports various methods from `helpers/*.js` :slightly_smiling_face: 

My idea here was that we'd be able to get rid of this file completely once we use e.g. WebGL blur filters as a part https://github.com/tmrowco/electricitymap-contrib/issues/923 and in the meantime I wanted to keep a clean separation between our internal `SolarLayer` React component and the temporary "external" blur helper which seems to be an adaptation of https://github.com/flozz/StackBlur and was never properly integrated in our code.

If you expect we'll be depending on that helper for a while, perhaps I can actually remove it in favor of using https://github.com/flozz/StackBlur directly, that would also somewhat simplify the code in `solarlayer.js`?

fbarl:
> Regarding performance, have you tested on very old devices?
It might be fast on your laptop but when I tested with old iPhones / iPads I had some lags especially during dragging. I think it would make sense to keep hiding the arrows while dragging if it is not too much work.

Good point! I didn't test on the old devices, I just didn't notice a _significant_ downgrade in performance compared to production on my machine, but you're right that's not a good way to test performance.

I added the code to hide arrows when dragging in https://github.com/tmrowco/electricitymap-contrib/pull/2395/commits/fb84c8e866e6f857b76d553859452c6bd9a715aa, the change was straightforward.

PTAL :slightly_smiling_face:

web/src/helpers/image.js
corradio: hmm let's keep it here for now. We're getting a bit pressed for time so I don't want to focus too much on these aspects which can be changed later as well :)

PR #2400, opened 2020-04-23
Use fallback mixes for zones we have them for

### LINKED ISSUE #2391: Wrong import co2eq values
See: https://tmrow-team.slack.com/archives/C2DQPTHA6/p1586621359085900

Imports from other countries where there is no current information had estimated (yearly) co2eq value before this issue. Now imports are assumed same co2eq as the countries average whose importing from another country.

robertahunt:
Fixes #2391

### DIFF
--- a/config/co2eq_parameters.js
+++ b/config/co2eq_parameters.js
@@ -14,7 +14,8 @@ exports.sourceOf = function(mode, zoneKey) {
 };
 
 exports.defaultExportIntensityOf = zoneKey =>
-  (co2eqParameters.fallbackZoneMixes.zoneOverrides[zoneKey] || {}).carbonIntensity;
+  (co2eqParameters.fallbackZoneMixes.zoneOverrides[zoneKey] || 
+    co2eqParameters.fallbackZoneMixes.defaults).carbonIntensity;
 
 exports.defaultRenewableRatioOf = (zoneKey) => {
     // if `zoneKey` has ratios in co2eqparameters, then use those ratios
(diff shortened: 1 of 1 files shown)


robertahunt:
The problem is our data pipeline was not finding the zone overrides since we made the update 20 days ago. We will need to rerun the pipeline for the last 20 days once this is merged.

If the pipeline doesn't find a zone override, it defaults to using the importing country's carbon intensity, so it doesn't dilute the model. We could discuss if this is more accurate than using the world average - at the moment I think it probably is. There is some correlation between the carbon intensity of interconnected zones.

config/co2eq_parameters.js
```diff
@@ -60,9 +60,10 @@ exports.renewableAccessor = (zoneKey, k, v) => {
 
 exports.defaultPowerOriginRatio = (zoneKey, fuelKey) => {
   //if no ratios found for that zoneKey, use defaults
-  const key = (co2eqParameters.fallbackZoneMixes[zoneKey] || {}).powerOriginRatios ? zoneKey : 'defaults' ;
-  
-  return co2eqParameters.fallbackZoneMixes[key].powerOriginRatios[fuelKey]
+  const ratio = (co2eqParameters.fallbackZoneMixes.zoneOverrides[zoneKey] ||
```
corradio, line 63: you could return directly at this line instead of declaring a variable and returning it after

config/co2eq_parameters.js
```diff
@@ -14,7 +14,7 @@ exports.sourceOf = function(mode, zoneKey) {
 };
 
 exports.defaultExportIntensityOf = zoneKey =>
-  (co2eqParameters.fallbackZoneMixes[zoneKey] || {}).carbonIntensity;
+  (co2eqParameters.fallbackZoneMixes.zoneOverrides[zoneKey] || {}).carbonIntensity;
```
corradio, line 17: for consistency, here you could also fallback on BP `defaults` (like you do for power origin).
Else we might have a power origin that uses defaults but a carbon intensity that is undefined (i.e. defaulting to country intensity instead of BP `defaults`)

PR #2401, opened 2020-04-23
France: Update with IRENA renewable data

nessie2013:
Also included ENTSO-E data for more up to date fossil fuel capacities


lorrieq:
Actually wind has been revised down from 16,494MW (previous RTE source) to 16,260MW (IRENA).. I would tend to favour a higher value (within reason) when there is a disagreement for wind and solar as these are growing unless there is some clear evidence the lower value is more accurate.

nessie2013:
Good point, I can put it back to the higher number.

PR #2404, opened 2020-04-23
Belgium: Update Capacities 

nessie2013:
Added IRENA as a source, Removed outdated source


README.md
```diff
@@ -212,8 +212,8 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
   - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
 - Belarus: [belstat.gov.by](http://www.belstat.gov.by/upload/iblock/7f7/7f70938f51eb9e49abc4a6e62f831a2c.rar), [RenEn](http://director.by/zhurnal/arkhiv-zhurnala/arkhiv-nomerov-2017/375-7-2017-iyul-2017/5456-zelenaya-energetika-nabiraet-oboroty)
 - Belgium
-  - Hydro, Oil: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
-  - Other: [ELIA](https://www.eliagroup.eu/-/media/project/elia/shared/documents/elia-group/publications-pdfs/20190411_gri18_elia_en.pdf)
+  - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
```
lorrieq, line 215: Hey @nessie2013, move Other to the bottom again and perhaps order Hydro, Wind, Solar, Biomass alphabetically

Fred-3000:
Hello! 
There is no active coal capacity in Belgium since 2016, I don't know how the 470 MW keeps popping up.  Source: http://www.caneurope.org/publications/press-releases/987-belgium-says-goodbye-to-coal-power-use

Also: the IRENA numbers look inaccurate, The 'outdated' source has been updated: https://www.elia.be/-/media/project/elia/shared/documents/elia-group/publications-pdfs/20200414_elia_annual-report-sustainability_en.pdf
Renewables:
PV: 4384 MW
Wind: 4037 MW
Hydro: 125 MW
Biomass: 811 MW (but this is without waste incineration)

Extra note: shouldn't the CO2 value for unknown account for the lack of coal?

lorrieq:
Hey @Fred-3000, thanks for your input!

Good spot on coal, I browsed over that focusing on the IRENA figures. Not sure why Entsoe are hanging on to that 470MW.

Hmmm.. that source looks good for wind and hydro, I'm not sure about solar. I would usually take the higher number for wind/solar considering they are getting more installed capacity all the time. Do you have any compelling reason for that 4384MW over 4531MW?

Do you have any info on the waste incineration figures and if that could account for the 811MW (without waste) vs 896MW (with waste?) disparity? If so we just need to add a _comment key with that information for the zone.

Fred-3000:
OK, the more I look into the PV situation, the more confusing it becomes. 

APERe (renewable energy observatory) : 4826 MW 
http://www.apere.org/fr/observatoire-photovoltaique

Elia, grid operator: 4384 MW but they report 4037 MW on their live monitoring page.  
https://www.elia.be/en/grid-data/power-generation/solar-pv-power-generation-data

So i think Irena information would be fine for PV. I am used to checking Elia's live solar production page  so I was a bit shocked to see 4531 MW :) 

As for waste incineration: per ENTSO-E data that should be 362 MW, so even a far bigger difference...

lorrieq:
OK thanks, so a little messy after all:
- biomass remains unclear due to waste element
- coal to be changed back to 0
- hydro to be updated to 125 (Elia)
- wind to be updated to 4037 (Elia)
- solar is OK at 4531

The readme will need to be updated to reflect that split in sources.

I don't know what the unknown split could reasonably be but yes, the simplest update would be average of gas + oil + biomass (given the default of "_assumes thermal (coal, gas, oil or biomass)_")

VIKTORVAV99:
What I did for Sweden (SE) was grouping renewable waste with biomass and non-renewable waste with oil and petroleum (I assume it's mostly plastic based).

Do you have know the ratio of renewable vs non-renewable waste? What I did for Sweden (SE) was grouping renewable waste with biomass and non-renewable waste with oil and petroleum (I assume it's mostly plastic based).

Do you have know the ratio of renewable vs non-renewable waste?

Otherwise you should be able to find 2018 production data here: https://ec.europa.eu/eurostat/web/energy/data/energy-balances

Fred-3000:
The Flemish Institute for Technological Research (VITO) estimated the renewable fraction of incinerated waste at 47.75% in 2016. (and rising)
https://www.vrt.be/vrtnws/nl/2016/09/21/5_vragen_over_afvalverbrandingengroeneenergie-1-2760219/

VIKTORVAV99:
That roughly lines up with the 41,6% renewable  waste power production from Eurostat 2018 data. (Might be some classification differences too) 

So my suggestion would be add the full waste capacity to unknown and then recalculate the co2eq based on Eurostat 2018 production levels.

lorrieq:
Hey @nessie2013, I can't add to your branch. Can you either allow me access or incorporate below changes?

image

Hey @VIKTORVAV99, thanks for your input! I'm not sure why we would put the waste capacity into unknown when currently waste is included in biomass on the electricitymap. However we could use your information to override the biomass co2eq for Belgium, what do you think of that?

VIKTORVAV99:
That should work perfectly actually, just need to know if any biomass or biofuel is included in unknown so I can exclude it from the biomass co2eq calculation.

VIKTORVAV99:
Actually we should recalculate unknown too as it today assumes some part is from biomass. (and we would get a more accurate coal/gas mix)

PR #2409, opened 2020-04-24
Add JP-ON

robertahunt:
It looks like it was just an oversight to not add JP-ON to the zones list? at least the only relevant issue I could find was the main issue for adding japan, and JP-ON is listed in the original parser, but was left out of zones.json in the original pr (https://github.com/tmrowco/electricitymap-contrib/pull/1543)

If I am mistaken, please let me know!!


alixunderplatz:
@robertahunt good catch, was wondering about Okinawa a few times
(by the way, Kyushu isn't online anymore and some of the other prefectures with solar data and some production curves are missing as well - maybe data format has changed somewhere in the sources?)

systemcatch:
@alixunderplatz  seems like the old link for Kyushu redirects to https://www.kyuden.co.jp/td_power_usages/pc.html, I managed to find the same div that is producing the error on the [logger](https://kibana.electricitymap.org/app/kibana#/doc/96f67170-0c49-11e9-85c1-1d63df8c862c/logstash-2020.04.25/doc?id=jyb5sHEBI9r-EpQawmKe&_g=()) so hopefully the layout hasn't changed.

robertahunt:
Good points @alixunderplatz and @systemcatch. I updated the urls for the JP parsers, and fixed some of the formatting. Unfortunately it seems like they change the format of their csvs often, and this parser is not very robust to those changes. Might be nice to overhaul it, but I didn't get that far.

parsers/JP.py
```diff
@@ -115,29 +117,31 @@ def fetch_consumption_df(zone_key='JP-TK', target_datetime=None,
...
+        startrow = 57
+    elif zone_key in ['JP-CB', 'JP-CG', 'JP-TK', 'JP-HKD', 'JP-TH', 'JP-HR', 'JP-SK', 'JP-KY']:
         startrow = 54
     else:
         startrow = 42
     df = pd.read_csv(consumption_url[zone_key], skiprows=startrow,
                      encoding='shift-jis')
-    
-    if zone_key in ['JP-CB', 'JP-TK']:
+    print(df.columns)
```
corradio, line 140: Should probably be removed

parsers/JP_ISEP.py
```diff
@@ -154,7 +154,7 @@ def fetch_production(zone_key='JP', session=None,
                 'unknown': row.get('unknown', None)
             },
             'storage': {
-                'hydro storage': row.get('hydro storage', None)
+                'hydro': row.get('hydro storage', None)
```
corradio, line 157: good catch

Issue #2410, opened 2020-04-24
Data on interconnectors between EU and non-EU countries


q--:
Page 15 and 16 of [this report](https://ec.europa.eu/energy/sites/ener/files/documents/2nd_report_ic_with_neighbouring_countries_b5.pdf) have a nice table.

>Situation in November 2018
so pretty recent

alixunderplatz:
nice discovery! seems like the values in the table represent the physical maximum transmission capacity, which is the right value to use.
a certain bit will always be used by reactive power. in addition, there usually are tramsmission limits from one side to the other (grid stability reasons).
Caution: I think Ukraine<>Romania with over 5GW is a bit high - if I remember correctly, it's due to an unused 750kV line from the Soviet times, which is partially not existing anymore. same probably applies for Ukraine<>Hungary and Ukraine<>Poland. capacity seems too high there opposed to what we regularly see on the map. background: A small western part of Ukraine is synchronous with the central EU countries. the rest is synchronous with the ex-soviet states. some Nuclear power plants in the west used to export power to the mentioned neighbouring states. but the high voltage level is not used anymore.
check blue lines on: https://www.entsoe.eu/data/map/

corradio:
Is there more to do here? Or have we already added all the capacities mentioned in the report?

q--:
Made a list. Don't have time to check them all at the moment, but I've seen that at least Estonia and Latvia appear to miss capacities 

- [ ] Russia-Estonia
- [ ] Russia-Latvia
- [x] Russia-Lithuania (is #2429 correct though... or does Lithuania also interconnect with Rusia on the eastern border?)
- [x] Belarus-Lithuania
- [ ] Ukraine-Poland but see https://github.com/tmrowco/electricitymap-contrib/issues/2410#issuecomment-619241587 w.r.t. Ukraine interconnections. Also, all interconnections already have import an import from Ukraine capacity. Whether or not that's correct
- [ ] Ukraine-Slovakia
- [ ] Ukraine-Hungary
- [ ] Ukraine-Romania
- [ ] Albania-Greece
- [ ] Bosnia and Herzegovina -Croatia
- [ ] North Macedonia-Bulgaria
- [ ] North Macedonia-Greece
- [ ] Serbia-Hungary
- [ ]  Serbia-Romania
- [ ] Serbia-Bulgaria
- [ ] Serbia-Croatia
- [ ] Turkey-Bulgaria
- [ ] Turkey-Greece
[...]

PR #2411, opened 2020-04-24
Add ES->MA interconnection capacity

q--:
Source from #2410

The report gives 8GWh of imports from Morocco to Spain, so even though almost al transmission is from Spain to Morocco the interconnection appears to work in the other direction as well


alixunderplatz:
@q-- I think it's an AC-interconnection, so it definitely would work in both directions.

PR #1764, opened 2019-02-09
Adjustment of capacity for Poland (PL)

wojciej:
Capacity volume has been adjusted according to https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show
Production Type	2019 [MW]
Fossil Peat	n/e
Nuclear	n/e
Fossil Hard coal	22444
Wind Onshore	5825
Fossil Brown coal/Lignite	7845
Geothermal	n/e
Hydro Run-of-river and poundage	436
Hydro Water Reservoir	157
Wind Offshore	n/e
Hydro Pumped Storage	1780
Other renewable	n/e
Solar	424
Fossil Oil shale	n/e
Waste	n/e
Fossil Gas	2129
Fossil Coal-derived gas	272
Fossil Oil	415
Marine	n/e
Other	n/e
Biomass	597
Total Grand capacity	42324

Those values has been added together as Hydro Storage
Hydro Run-of-river and poundage	436
Hydro Water Reservoir	157

Those values has been added together as coal:
Fossil Brown coal/Lignite	7845
Fossil Hard coal	22444

There is a question how to classify the following:
Fossil Coal-derived gas	272 
cause this is not a gas nor coal. Currently not included.

PR #2412, opened 2020-04-25
Poland Capacity Update

nessie2013:
Used the IRENA renewable data and some updated numbers from ENTSO-E

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,7 @@ Production capacities are centralized in the [zones.json](https://github.com/tmr
 - Peru: [IRENA](https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf)
 - Poland  
   - Solar: [PSE Twitter](https://twitter.com/pse_pl/status/1248502288152293378)
+  - Other Renewable: [IRENA](https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf)
   - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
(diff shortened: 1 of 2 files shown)


wojciej:
Hello, I need to disagree with this change and here are the reasons :

- gas capacity was changed from "gas": 2812, to "gas": 2540,  - and this is cause the lower value does not include Fossil Coal-derived gas | 272.
As this was agreed in here https://github.com/tmrowco/electricitymap-contrib/pull/1764 we should add those two values together

- hydro power and hydro storage power was changed from
      "hydro": 605,
      "hydro storage": 1780,
to 
      "hydro": 968,
      "hydro storage": 1423,
In total its all the same, its just that IRENA raport qualify those two differently than ENTSOE

- wind was changed from
      "wind": 5953
to
      "wind": 5917
And this is lower value is in aligned with what URE provides with its data
https://www.ure.gov.pl/download/9/10905/mocIVkw2019.pdf
however this does not include private generation sources that are small enough that do not need to be concessioned like private PV on a private home in Poland

The only data change I could agree with is the biomass - and this is even higher as per my current update and pull request and this confirms the URE data I have linked above. 

Please decide what to do in this situation, as this update puts confusion and inconsistency to the previous agreements.

jarek:
I'm not really familiar with the Polish energy market. Per #1764 yeah gas should include coal-derived gas. @nessie2013 any thoughts?

wojciej:
Hi Jarek, 
Thanks for your reply. 
I will include them in 2446 I just wanted to wait couple days to see if @nessie2013 would reply to this thread cause I dont want to play hide and seek with the data, changing them back and forth. I would rather understand the reasoning here.

wojciej:
Since we did not get any reply to this thread from @nessie2013 even tho the user has been active (pull request has been created on 5th of May #2474) I will revert all the changes done in this pull request.

PR #2423, opened 2020-04-27
Added singapore to 'unknown' fossil free exceptions

robertahunt:
After doing some digging into Singapore's makeup, found that the 'other' category from their website contains only biomass, waste, and solar. So I think this should be classified as fossil free.

Some references:
the government specifies the 'other' category as containing: "Others’ refer to other energy products that are not classified. These include Solar, Biomass, and Waste (such as Municipal Waste and Biogas). Municipal Waste is waste produced by residential, commercial and public services that is collected by local authorities for disposal in a central location for the production of electricity and/or heat."
https://www.ema.gov.sg/Singapore-Energy-Statistics-2019/Ch08/index8

https://www.ema.gov.sg/TemStatistic.aspx?pagesid=20140926wbNYp2Yh8iqy&pagemode=live&sta_sid=20140802NEeM2zyMguvz

### DIFF
--- a/config/co2eq_parameters.js
+++ b/config/co2eq_parameters.js
@@ -49,7 +49,7 @@ exports.fossilFuelAccessor = (zoneKey, k, v) => {
   return (k == 'coal' ||
           k == 'gas' ||
           k == 'oil' ||
-          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey != 'SG')) ||
+          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey !== 'SG')) ||
           k == 'other') ? 1 : 0;
 }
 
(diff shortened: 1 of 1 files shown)


config/co2eq_parameters.js
```diff
@@ -49,7 +49,7 @@ exports.fossilFuelAccessor = (zoneKey, k, v) => {
   return (k == 'coal' ||
           k == 'gas' ||
           k == 'oil' ||
-          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA')) ||
+          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey != 'SG')) ||
```
corradio, line 52:
```suggestion
          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey !== 'SG')) ||
```

q--:
Are you sure the "Other" on that website is the same as "unknown" in ElectricityMap's parser?

In the parser, 'unknown' is what's used for Steam Turbines, so it definitely doesn't include solar but might include e.g. oil

https://github.com/tmrowco/electricitymap-contrib/blob/224a3189bc7dd4f41daab74f4a2abe45feb60c47/parsers/SG.py#L14

robertahunt:
@q-- , Good point. That's an oversight on my part. I'll have to consider how we can account for the missing fossil-free production then. Any ideas?

PR #2427, opened 2020-04-27
Checked us zones, separated old US from new US

### LINKED ISSUE #2319: Imports into new US zones don't work

robertahunt:
Fixes #2319 .

This PR does a couple things:
- removes exchanges from old US zones to non-US zones - so the old US zones are now a virtual island (to be removed once customers have time to switch)
- adds a couple missing exchanges - US-NW-AVRN and US-SW-PNM were missing exchange configurations

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -518,16 +518,6 @@
     },
     "rotation": 90
   },
-  "CA-AB->US-MT": {
-    "lonlat": [
-      -111.920238,
-      49.016577
-    ],
-    "parsers": {
-      "exchange": "CA_AB.fetch_exchange"
-    },
-    "rotation": 180
-  },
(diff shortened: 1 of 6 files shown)


robertahunt:
@corradio any idea why pylint is screaming? When running locally I have no linting issues and the error message from the CI is incomprehensible to me.

corradio:
> @corradio any idea why pylint is screaming? When running locally I have no linting issues and the error message from the CI is incomprehensible to me.

From Googling around it could be an incompatibility between python3 and the older version of pylint used here. I would suggest trying to upgrade it in .circleci/config.yml

Issue #2304, opened 2020-03-24
Movemove performance


corradio:
I've noticed a performance degradation especially when I move the mouse around the map as there seems to be some lag that affects the tooltip update as its associated animations.
Has anyone experienced that? Maybe we're sending a lot of redux actions at each move move?

cc @fbarl

fbarl:
I haven't noticed this personally on my Chrome, but it's possible some performance leaks were introduced in https://github.com/tmrowco/electricitymap-contrib/pull/2276.

I'll assign myself to investigate.

fbarl:
By the way, if Redux is the bottleneck (which is quite possible), then addressing https://github.com/tmrowco/electricitymap-contrib/issues/2288 would also fix this.

corradio:
I've noticed that the SHOW_TOOLTIP event is dispatched for each movemove on the countrytable.
It would suffice to have it dispatched on mouveover

fbarl:
> I've noticed that the SHOW_TOOLTIP event is dispatched for each movemove on the countrytable. It would suffice to have it dispatched on mouveover.

Yes, we could do that indeed as a quickfix if we're ok with the tooltip not following the mouse cursor when hovering within a country - if that sounds like an ok compromise to you @corradio, I can put up a quick PR to see how it feels.

I've realized that it'd be very difficult to embed map tooltips deeper in DOM (and thus bypass Redux) as a part of #2288 before moving `Map` into a React component (see #2309), so maybe the change you're suggesting is indeed a way to go for now.

corradio:
It's not urgent so I would instead try to make sure the tooltips are embedded deeper into the DOM, potentially allowing for local state.
Another things is that maybe we should decouple the update of the tooltip content, with the positioning of it? Maybe that helps.
In any case, let's indeed try to bypass redux for the tooltips.

fbarl:
@corradio I wonder if you are still experiencing this now that #2434 has been merged?

corradio:
Unfortunately it's not. The test I'm making is that I'm moving my mouse quickly inside the same zone, and ideally the tooltip should follow closely the mouse position. However in this case there's some lag where the tooltip position lags behind the mouse position.

fbarl:
[...]

PR #2434, opened 2020-04-28
Migrate Map to React

### LINKED ISSUE #2288: Render tooltips deeper in the DOM
Extending on #2276, it would probably be a good idea to render tooltips next to the components which invoke them instead of having a global list of them.

That would also enable us to get rid of a global shared toolbar state, which would be necessary if we were to ever display multiple tooltips at once (or just have fade effects on them).

### LINKED ISSUE #2309: Move Map component to React

### LINKED ISSUE #2311: Remove all DOM management from main.js

fbarl:
This is another big one, which finally completes the migration to React! :tada: 

The performance looked uncompromised to me, but I'll probably run some metrics before merging to assure that's the case. The new `ZoneMap` component has a somewhat different interface, but I think I managed to make it sufficiently reusable.

#### Major Changes

* Created a _stateless_ React component `ZoneMap` that renders the map with [react-map-gl](https://github.com/visgl/react-map-gl) and a `Map` layer wrapper which plugs in all the handlers + data to it - resolves #2309 
* Moved `MapCountryTooltip` used when hovering the map inside the new `Map` layer, removing the need to track tooltip state in Redux - resolves #2288 (I also wonder how this will affect #2304)
* By removing the map code from `main.js`, not much was left to clean up that file completely so I did it in this PR - resolves #2311
* Embedded all the layers (exchange, wind, solar) into the map component with the help of a new proxy component `MapLayer`, treating them as Mapbox GL [custom overlays](https://visgl.github.io/react-map-gl/docs/advanced/custom-overlays) - in particular, that fixed the zooming over exchange arrows issue mentioned in https://github.com/tmrowco/electricitymap-contrib/pull/2395#issue-407131491
* Expanded 2 themes into 4 (dark, bright, colorblind dark, colorblind bright) + introduced `useTheme` hook for easier access to theme colors


robertahunt:
One thing I noticed when trying to run locally - exports seem to overlap the zone names now 
image

fbarl:
@robertahunt Indeed, the screenshot you posted is a result of the local mock data not containing some fields which in my observation are in fact always present on the production (in this case I think it's the CO2 _totals_ which are supposed cap the bar width):

image

We should probably update the mock data to fix this, I think it's also been around before this PR :slightly_smiling_face:

fbarl:
@corradio Please note two more commits I had to add to make `ZoneMap` more robust to be used externally:

* More robust styling via props - https://github.com/tmrowco/electricitymap-contrib/pull/2434/commits/83308d27428188dc3438544734f1c98b1f26a938
* Better internal tracking of map dragging state to enable smooth transitions when using zoom in/out navigation controls - https://github.com/tmrowco/electricitymap-contrib/pull/2434/commits/f276093f20c4e1afd2407b89dcb1ea51854cc357

PR #2438, opened 2020-04-29
Update test_parser to give error if parser returns pandas timestamps

### LINKED ISSUE #2437: Add check to test_parser that timestamps are datetime
Recently found an issue with Chile that the parser was returning pandas timestamps, not python datetimes, which currently passes our tests, but causes an error when trying to feed into the database. We should add a check to the test_parser to mitigate this in the future.

robertahunt:
Fixes #2437


test_parser.py
```diff
@@ -54,7 +55,7 @@ def test_parser(zone, data_type, target_datetime):
 
     elapsed_time = time.time() - start
     if isinstance(res, (list, tuple)):
-        res_list = iter(res)
+        res_list = list(res)
```
corradio, line 58: I think you can keep it an iterable?

test_parser.py
```diff
@@ -65,6 +66,9 @@ def test_parser(zone, data_type, target_datetime):
               'ouput. Full ouput: \n\n{}\n'.format(res))
         return
 
+    assert not any([isinstance(e['datetime'], pd.Timestamp) for e in res_list]), \
```
corradio, line 69: I would do the opposite: assert that all datetimes are instance of `datetime.datetime`.
That way, we also catch datetimes that are strings instead of datetime objects!

test_parser.py
robertahunt: Can't iterables only be used once? And we have two checks on this data now, so if that is the case, then no.

test_parser.py
robertahunt: Makes sense. I wasn't sure if we also accepted arrow instances, but I'll change it. Would be good to keep it consistent anyways.

test_parser.py
corradio: good point - agree

PR #2446, opened 2020-04-30
PL generation capacity update

wojciej:
I have noticed that live data on entsoe for biomas generation capacity are much higher than generation capacity in entsoe themself.
I dug deeper and I have end up on Polish Energy Regulator Office website where I have found data for Q4 2019 showing that available generation capacity for biomass in poland is actually: 1 492,875MW

https://www.ure.gov.pl/download/9/10905/mocIVkw2019.pdf

I will add this into the resources in readme

Also - I have updated PL solar capacity as per new reading from PSE.

I also have reverted changes introduced in #2412 as per discussion in that thread.

PR #2447, opened 2020-04-30
PL Biomass generation capacity source update

wojciej:
I have noticed that live data on entsoe for biomas generation capacity are much higher than generation capacity in entsoe themself.
I dug deeper and I have end up on Polish Energy Regulator Office website where I have found data for Q4 2019 showing that available generation capacity for biomass in poland is actually: 1 492,875MW

https://www.ure.gov.pl/download/9/10905/mocIVkw2019.pdf


wojciej:
Added to one pull request #2446

alixunderplatz:
@wojciej you see larger values for biomass electricity on the map, because since 10th April, solar data is reported within the biomass category to entso-e.
compare with this chart (select only biomass):
http://www.energodock.com/poland/electricity-generation

PR #2450, opened 2020-05-01
Readme improvement

steinwinde:
- added information regarding API keys
- corrected a statement regarding the state of the map after installation (dummy data instead of empty)
- removed invisible characters at line ends
- promoted a sub-sub-sub section to sub-sub section

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,12 +506,6 @@ Run all of the parser tests with the following command from the root directory:
 python -m unittest discover parsers/test/
 ```
 
-Many of the tests require API keys of the data or web service providers and will therefore fail with an error message like
-
-> Exception: No ENTSOE_TOKEN found! Please add it into secrets.env!
-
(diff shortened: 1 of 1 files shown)


steinwinde:
I have also removed the version of python from some command invocations, because it's not consistently used in the document (and would otherwise need to be added to the pip command too). Maybe state python version in a separate section?!

robertahunt:
This is great! Thanks @steinwinde

README.md
```diff
@@ -503,7 +503,7 @@ Once you're done, add your parser to the [zones.json](https://github.com/tmrowco
 
 Run all of the parser tests with the following command from the root directory:
 ```
-python3 -m unittest discover parsers/test/
```
q--, line 506: I found it helpful to know that these were python3 when getting things up and running, are you sure it's a good idea to remove this?

robertahunt: From my perspective this is system dependant. I, for example, don't have a 'python3' command - I use conda environments.

And in the other command line snippets in the readme, we use 'python', so I think this is better for consistency's sake...

But we should definitely specify somewhere that we use python 3.6.. I will add this to the readme now

PR #2451, opened 2020-05-01
Bugfix france parser price calc

steinwinde:
Replaced "hours" by "hour" to make the FR.py work.

### DIFF
--- a/parsers/FR.py
+++ b/parsers/FR.py
@@ -171,7 +171,7 @@ def fetch_price(zone_key, session=None, target_datetime=None,
                 if value.text == 'ND':
                     continue
                 period = int(value.attrib['periode'])
-                datetime = start_date.replace(hour=+period).datetime
+                datetime = start_date.shift(hours=+period).datetime
                 if not datetime in datas:
                     datas[datetime] = {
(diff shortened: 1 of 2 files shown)


ajoga:
@steinwinde you also pushed in this PR the commits you PR'd via #2450. Could you remove them from this PR please ?

parsers/FR.py
```diff
@@ -171,7 +171,7 @@ def fetch_price(zone_key, session=None, target_datetime=None,
                 if value.text == 'ND':
                     continue
                 period = int(value.attrib['periode'])
-                datetime = start_date.replace(hours=+period).datetime
+                datetime = start_date.replace(hour=+period).datetime
```
jarek, line 174:
```suggestion
                datetime = start_date.shift(hours=+period).datetime
```

see https://github.com/tmrowco/electricitymap-contrib/issues/2416#issuecomment-619422705

steinwinde:
@ajoga @jarek - thanks!
I've also removed the reference to an issue that I thought I fix, but fail to do so.

robertahunt:
Just want to get this included in the upcoming bump of contrib :)

Issue #2196, opened 2020-02-02
Add real-time data to Chile


robertahunt:
The parser for Chile appears to be failing for real time. (see [kibana](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=()&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:extra.key,negate:!f,params:(query:CL-SING-SIC,type:phrase),type:phrase,value:CL-SING-SIC),query:(match:(extra.key:(query:CL-SING-SIC,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))) and also for historic data (gives arrow timestamp formatting error)

I am also a bit confused about Chile in general. It looks like in zones.json the only parser we have set up is CL-SING-SIC, but this parser seems to then point to CL.py, which appears to (attempt to) fetch production for all of Chile. ([link to data source](https://www.coordinador.cl/operacion/graficos/operacion-real/generacion-real-del-sistema/)). 

I can fix the hiccup the parser is having, but I want to first understand if it is fetching data for all of chile or only CL-SING-SIC.

alixunderplatz:
@robertahunt the separated systems SING (Northern Chile) and SIC (central Chile) were merged through an interconnector a while ago, creating one common zone which was named SING-SIC here. The data from the link is for that region SING+SIC only.
Like in most Spanish-speaking countries, it seems that zone is called "SEN" now (sistema electrico nacional), so we could switch to that terminology while the parser is being fixed.
Fetching real-time data for Chile wasn't really possible in the past, since the data was partially delayed and incomplete, and complete generation per type data mostly was like 7-8 hours behind real-time.
At the moment, it seems like "Mix Generación real por tecnología" tab isn't showing any data on https://www.coordinador.cl/operacion/graficos/operacion-real/generacion-real-del-sistema .
[...]

Issue #1683, opened 2018-11-21
CL-SIC real-time data missing


brunolajoie:
Data for SL-SIC seems to be irregular, being present about 60% of the time only over the last 6 months (see picture below, since April 2018)
![image](https://user-images.githubusercontent.com/22095643/48847496-fa521100-eda1-11e8-8755-7829cfe6c8fe.png)

We need

1. To understand if its related to our parser not handling response correctly, or if its due to the source of data that breaks downs regularily. Calling parser returns
```
min returned datetime: 2018-11-06 01:00:00-03:00
max returned datetime: 2018-11-07T03:00:00+00:00 UTC  :( >2h from now !!!
```
2. To allow historical refetch to fix the missing datapoints. @systemcatch, You have kindly written a parser to fetch SL-SING historical data (#1394). Would you know if its possible to do the same with SL-SIC?

linked to #655 #1108 #1394

systemcatch:
Hey Bruno, I've taken a look at the parser response. All of the data comes from https://sic.coordinador.cl/informes-y-documentos/fichas/operacion-real/

1. The parser is working fine, data seems to be regularly delayed by a number of days (this is the same as when I wrote this parser). Never saw it 2 weeks behind though.

2. Looking at the parser it would be simple to modify to enable backfilling with historical data.

brunolajoie:
All right, so it's good news if I understand well, data is there, but just sometimes with more than 24h delay, therefore not visible on the electricityMap history graphs. 

Historical data seems to dates back several years. So, I guess we just need to add a fetch_history feature - cool! We'll just have to make it robust to potential change in power plant fleets

systemcatch:
Yes it has always been a delayed data source, we've never found a real time one.

I can add a target_datetime parameter to the fetch production function.

brunolajoie:
Very interesting. I'm sending them an email right now to decide what to do next. 

For now, lets keep it split in two regions (because we have the gas vs. coal breakdown on the sub zones, while here it's only solar/Wind vs. other) and add historical datasets for both.

We'll merge both regions when we'll get a better breakdown.
Meanwhile: have you seen a way to get power exchanges between SIC and SING from our current sources?

brunolajoie:
Also, how well interconnected are they, in terms of %of load being exchanged?

systemcatch:
@alixunderplatz I got creative with Wireshark and was able to glean where the app is getting it's data from.

[...]

PR #2454, opened 2020-05-01
Add real-time data to Chile (#2196)

peterhughgriffin:
I've added a new implementation to the CL parser to get live data for the whole of Chile, using the data behind the Coordinador app, as mentioned in [#1683 (comment)](https://github.com/tmrowco/electricitymap-contrib/issues/1683#issuecomment-441279533). My implementation was mostly based on the Brazilian parser (BR).

I'm not sure about the datetime implementation, so would appreciate some feedback on that.


parsers/CL.py
```diff
@@ -59,6 +70,38 @@ def data_processor(raw_data):
...
+def production_processor_live(json_data, zone_key):
+    """
+    Extracts data timestamp and sums regional data into totals by key.
+    Maps keys to type and returns a tuple.
+    """
+
+# unsure about this time object, but seems to be unix time, but in msecs
+        # 1) is this right?
+        # 2) does this need to have a timezone or anything??
+    dt = arrow.get(json_data['fecha']/1000)
```
robertahunt, line 90: I did a quick check that the times returned by your parser match the times given on the app, and they do. So although I am not used to APIs returning this datetime format, I think you've done it correctly! I found lots of google results for Unix milliseconds time, so I guess this is not so unheard of either.

parsers/CL.py
```diff
@@ -59,6 +70,38 @@ def data_processor(raw_data):
...
+
+def production_processor_live(json_data, zone_key):
+    """
+    Extracts data timestamp and sums regional data into totals by key.
+    Maps keys to type and returns a tuple.
+    """
+
+# unsure about this time object, but seems to be unix time, but in msecs
+        # 1) is this right?
+        # 2) does this need to have a timezone or anything??
```
robertahunt, line 89:
```suggestion
```
I would suggest removing these before we merge

peterhughgriffin: Great, will do that tomorrow. Thanks

alixunderplatz:
Hi @peterhughgriffin, I just had a look at the data returned by your parser.
For real-time data, you're using the URL to "http://panelapp.coordinadorelectrico.cl/api/chart/produccion", which from my understanding is "yesterday's daily sum of produced energy per category ".
This is returned for the live-data with that URL:
`{'zoneKey': 'CL', 'datetime': datetime.datetime(2020, 5, 3, 13, 37, 7, 497000, tzinfo=tzutc()), 'production': {'solar': 16413.6, 'wind': 18495.4, 'hydro': 27676.0, 'unknown': 192758.3}, 'storage': {'hydro': None}, 'source': 'panelapp.coordinadorelectrico.cl'}`
These values seemed too high when I checked, which made me aware of this issue.
Dividing "TOTAL GENERACION" by 24hours returns about 8.000 MW, which is the average production of Chile, so my assumption would make sense.

Actually, you'd need these two sources to get the correct 10-min live-data:
wind and solar should come from here:
http://panelapp.coordinadorelectrico.cl/api/chart/ernc
And unknown is "SIN SOLAR Y EÓLICO" from here:
http://panelapp.coordinadorelectrico.cl/api/chart/demanda
By combining these two, we can figure out the live-mix.

Actually this brings up a new problem which should be adressed in the code:
Using the recent version of the parser, we're able to get a part of the previous day's production:
grafik

This means, the parser for Chile should/could include three options to retrieve data:
1. Live data for the recent day, consisting of Solar+Wind+Unknown with 10-min-granularity 
   - high granularity / mediocre split to the sources

2. refined data for the last 24 hours, consisting of Solar+Wind**+Hydro**+Unknown with 60-min-granularity
   - decent granularity / better split of the sources

3. Historical data, consisting of Solar+Wind+Hydro**+Geothermal**+Unknown with 60-min-granularity
   - decent granularity / best split of the sources

@robertahunt Wouldn't it make sense if you to collected data from all three of these options seperately (avoiding overwriting data in the database)? 
I prefer the 10-min data from 1.) for the live-view on the map over 2.), as i feel hydro doesn't make much of a difference (at least at the moment). This would also bypass the issue of assigning different carbon intensities to "unknown", depending on the given production types from 1, 2 and 3.

parsers/CL.py
systemcatch, line 90:
```suggestion
    dt = arrow.get(json_data['fecha']/1000, tzinfo='Chile/Continental')
```
This will add the correct timezone to the datetime object.

parsers/CL.py
```diff
@@ -91,11 +134,25 @@ def fetch_production(zone_key='CL', session=None, target_datetime=None, logger=l
...
+        generation = production_processor_live(gd, zone_key)
+
+        datapoint = {
+                      'zoneKey': zone_key,
+                      'datetime': generation[0].datetime,
+                      'production': generation[1],
+                      'storage': {
+                          'hydro': None,
+                                  },
+                      'source': 'panelapp.coordinadorelectrico.cl'
```
systemcatch, line 148:
```suggestion
                      'source': 'coordinadorelectrico.cl'
```
We don't generally bother with the sub domain parts.

parsers/CL.py
```diff
@@ -132,7 +189,9 @@ def fetch_production(zone_key='CL', session=None, target_datetime=None, logger=l
 
 if __name__ == "__main__":
     """Main method, never used by the Electricity Map backend, but handy for testing."""
-    print('fetch_production() ->')
-    print(fetch_production())
+    #print('fetch_production() ->')
+    Prod=fetch_production()
+    print(Prod)
```
systemcatch, line 194: Not a fan of the old style? :wink:

parsers/CL.py
```diff
@@ -59,6 +70,38 @@ def data_processor(raw_data):
...
+    
+    totals = defaultdict(lambda: 0.0)
+    
+    breakdown = json_data['data']
+
+    for generation in breakdown:
+        totals[generation['key']] = generation['value']
+
+    mapped_totals = {TYPE_MAPPING_LIVE.get(name, 'unknown'): val for name, val
+                     in totals.items()}
```
systemcatch, line 100: We should probably raise a warning in the logger if a new energy type appears.

peterhughgriffin: In looking at doing this, I've got myself confused. If there is an unexpected source, then either the "TERMICAS" data or that of the unexpected one will be overwritten.

In fact I realised this was already happening, the TERMICAS data was being overwritten by the total production.

I've got round this by adding the total to TYPE_MAPPING_LIVE and using this to check for 'unexpected' values, which now don't overwrite 'unknown'. I then remove the total.

What do you think?

peterhughgriffin:
> Hi @peterhughgriffin, I just had a look at the data returned by your parser.
> For real-time data, you're using the URL to "http://panelapp.coordinadorelectrico.cl/api/chart/produccion", which from my understanding is "yesterday's daily sum of produced energy per category ".
> This is returned for the live-data with that URL:
> `{'zoneKey': 'CL', 'datetime': datetime.datetime(2020, 5, 3, 13, 37, 7, 497000, tzinfo=tzutc()), 'production': {'solar': 16413.6, 'wind': 18495.4, 'hydro': 27676.0, 'unknown': 192758.3}, 'storage': {'hydro': None}, 'source': 'panelapp.coordinadorelectrico.cl'}`
> These values seemed too high when I checked, which made me aware of this issue.
> Dividing "TOTAL GENERACION" by 24hours returns about 8.000 MW, which is the average production of Chile, so my assumption would make sense.
> 
> Actually, you'd need these two sources to get the correct 10-min live-data:
> wind and solar should come from here:
> http://panelapp.coordinadorelectrico.cl/api/chart/ernc
> And unknown is "SIN SOLAR Y EÓLICO" from here:
> http://panelapp.coordinadorelectrico.cl/api/chart/demanda
> By combining these two, we can figure out the live-mix.
> 
> Actually this brings up a new problem which should be adressed in the code:
> Using the recent version of the parser, we're able to get a part of the previous day's production:
> grafik
> 
> This means, the parser for Chile should/could include three options to retrieve data:
> 
> 1. Live data for the recent day, consisting of Solar+Wind+Unknown with 10-min-granularity
>    
>    * high granularity / mediocre split to the sources
> 2. refined data for the last 24 hours, consisting of Solar+Wind**+Hydro**+Unknown with 60-min-granularity
>    
>    * decent granularity / better split of the sources
> 3. Historical data, consisting of Solar+Wind+Hydro**+Geothermal**+Unknown with 60-min-granularity
>    
>    * decent granularity / best split of the sources
> 
> @robertahunt Wouldn't it make sense if you to collected data from all three of these options seperately (avoiding overwriting data in the database)?
> I prefer the 10-min data from 1.) for the live-view on the map over 2.), as i feel hydro doesn't make much of a difference (at least at the moment). This would also bypass the issue of assigning different carbon intensities to "unknown", depending on the given production types from 1, 2 and 3.

Thanks Alix, I'd blame my spanish, but really it just shows I didn't check!
I've addressed the easy bugs from the others first, but this needs a little more thought.

I've had a bash at correcting the live sources to get the solar and wind from "ernc" and total from "demanda" to give the live data, as suggested as option 1. Take a peek in the latest commit, it makes for a simpler program really, but I am making assumptions about the data structure given by the API that would be difficult to check (i.e. that ernc has wind data and then solar.)
Should I check the keys? 

I'm not sure about the consequences of the question you've directed at @robertahunt, as I don't yet understand how the data from parsers is accessed and stored.

parsers/CL.py
peterhughgriffin: This may be irrelevant now...

systemcatch:
@peterhughgriffin it's generally safer to check the keys. As for the 5min time difference you can try to match the timestamps then drop the ones that don't match up.

peterhughgriffin:
Keys: Right, will do. Not sure why it seemed like a big deal.

Time: As in take the latest coherent data, rather than the latest data from each source?

systemcatch:
@robertahunt do you think there's any benefit in having the parser return more than one datapoint here? i.e. up to an hour back.

@peterhughgriffin Yeah that was the idea but honestly if the difference is only 5 minutes I don't think it's that important.

robertahunt:
@systemcatch Yes, definitely the further back we can go the better, just in case the parsers fail for an hour, or if they update the data in some way. It's always nice to have a few buffer datapoints.

peterhughgriffin:
Great, so do I just return a list for each type of generation? What’s the format?

systemcatch:
Hey @peterhughgriffin it's exactly the same format as the current datapoint you're returning, only a list of them.

peterhughgriffin:
Am now returning a list of dictionaries, rather than just the most recent datapoint. I've also added key checking for the two renewable columns.

I discovered that the 5 minute difference between the total generation and renewable was actually due to a difference in data granularity, the renewable data updates every 5 minutes, while the total updates every 10. I have opted to return a dictionary for all the timestamps of the total and therefore ignore the extra renewable data.

peterhughgriffin:
Has anyone managed to review my updated version? @systemcatch @robertahunt

systemcatch:
Hey @peterhughgriffin looks good to me, the only thing that needs changing is to ensure we have python `datetime` objects rather than pandas `Timestamp` objects for the `"datetime"` key. You can use the [to_pydatetime()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timestamp.to_pydatetime.html#pandas.Timestamp.to_pydatetime) method for this.

peterhughgriffin:
Thanks Chris.
I guess you mean in the timestamp_creator function. That's in the historic data bit, which I've not really looked at it, but hopefully my addition sorts it out.
That whole bit could use arrow instead of pd I guess, but I had an "if it ain't broke mentality"

martincollignon:
Amazing! @systemcatch are we ready to merge? :)

systemcatch:
@peterhughgriffin ah sorry I was getting mixed up between the different processors. No harm fixing that though. In your live data part do you ensure that the datetime key is a datetime object?

When I run the parser I get the following output.

```shell
(EM) chris@ThinkPad:~/electricitymap$ python -m parsers.CL
fetch_production() ->
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
Required generation type hydro is missing from CL
[None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None]
```

Which is due to `hydro` being missing from the generation. I think we are good to merge, thanks for your patience.

Issue #2452, opened 2020-05-01
Canada Alberta lacks prices


steinwinde:
Search kibana and find:

> unknown attribute: "hours" | CA-AB

From what I can see, line 130 of CA_AB.py should be

> hour=hours

Mind the singular.

PR #2455, opened 2020-05-02
update CA_AB for arrow update

jarek:
In this case we can continue using replace(hour=...) because the object we're doing the replace on is certain to have 0 hours at the start (having just been parsed with format string 'MM/DD/YYYY')

closes https://github.com/tmrowco/electricitymap-contrib/issues/2452


robertahunt:
Looks good! Thanks!

PR #2456, opened 2020-05-02
Adding a parser for Australia's king island #636

ajoga:
Based on the discussion I tried to implement a parser for Australia's king island.

This very likely need help/rework as the endpoint I'm using to connect to the websocket has a token in it, and I'm guessing that this is going to expire at some point... Any help on implementing that better is welcome


parsers/AUS-TAS-KI.py
```diff
@@ -0,0 +1,67 @@
...
+    """ 
+    TODO
+        * What do I return for production mode that are not even on the island ? 
+            The main README says "The production values should never be negative. Use None, or omit the key if a specific production mode is not known." but doesn't answer this question
+    """
+    return {
+      'countryCode': 'AUS-TAS-KI',
+      'datetime': arrow.now(tz='Australia/Currie').datetime,
+      'production': {
+          'biomass': 0,
```
q--, line 48: The showcase has an icon that shows if they're using biodiesel (which they currenlty aren't), would there be a way to detect this so it can be switched from oil to biodiesel if they start using that?

ajoga: Mmh, I didn't consider biodiesel as counting as biomass... Is it undoubtedly ?

With no biodiesel currently being used, I'm not sure how to make sense of the payload : 
```
  "biodiesel": {
    "id": "biodiesel",
    "title": "Biodiesel",
    "value": 0,
    "unit": null,
    "percent": 0,
    "status": false
  }
```
My intuition is that when in use, `percent` is the percentage of bio-diesel blended with regular diesel in order to generate the total diesel-generated power, so I would have to subdivide the total diesel energy made into biomass and diesel. Or maybe the `value` and `unit` would be directly in kW. 

I suggest for now to add a TODO block for that. What do you think ?

q--: I've looked through the source code, and the only file where biodiesel is mentioned is https://data.ajenti.com.au/KIREIP/files/dashboard.js. The biodiesel code starts at line 170, and the only key from the data that's used is `percent`. If biodiesel and diesel values were reported separately and then added up to the total kW for the diesel gauge, I'd expect biodiesel to be mentioned elsewhere as well (in order to be able to do diesel.value + biodiesel.value).

I've also tried looking for 'diesel' and 'percent', but none of the results suggest anything like _diesel.value + biodiesel.value_ is done.  So I'm 99.99% sure that the total for diesel includes all biodiesel, so the correct thing to do would be
```
oil = diesel * (1- biodiesel.percent/100)
biomass = diesel * biodiesel.percent/100
```

Could you have a look if other things (e.g. wind and solar) have a unit reported when not generating? If so, that would further confirm my suspicion that  `value` and `unit` will always stay null for biodiesel, and are only included in order to have a consistent data structure.

ajoga: > I've also tried looking for 'diesel' and 'percent', but none of the results suggest anything like _diesel.value + biodiesel.value_ is done. So I'm 99.99% sure that the total for diesel includes all biodiesel, so the correct thing to do would be
> 
> ```
> oil = diesel * (1- biodiesel.percent/100)
> biomass = diesel * biodiesel.percent/100
> ```
> 

It makes sense with the presentation of the plant saying that the biodiesel is blended with regular diesel : https://www.hydro.com.au/docs/default-source/clean-energy/hybrid-energy-solutions/king_island.pdf

> Could you have a look if other things (e.g. wind and solar) have a unit reported when not generating? If so, that would further confirm my suspicion that `value` and `unit` will always stay null for biodiesel, and are only included in order to have a consistent data structure.

Yeah I found from one of my test payloads an instance of this : 
```
    {
      "id": "solar",
      "title": "Solar",
      "value": 0,
      "unit": "kW",
      "percent": 0,
      "status": false
    },
    {
      "burn": false,
      "id": "diesel",
      "title": "Diesel",
      "value": 0,
      "unit": "kW",
      "percent": 0,
      "status": false
    },
```

ajoga: See the next commits, especially 219590f

parsers/AU_TAS_KI.py
```diff
@@ -0,0 +1,81 @@
...
+          'nuclear': 0,
+          'oil': technologies_parsed['diesel']*(100-biodiesel_percent)/100,
+          'solar': technologies_parsed['solar'],
+          'wind': technologies_parsed['wind'],
+          'geothermal': 0,
+          'unknown': 0
+      },
+      'storage': {
+          'battery': technologies_parsed['battery']*-1,
+          'flywheel': technologies_parsed['flywheel']*-1,
```
q--, line 75: I've [searched](https://github.com/tmrowco/electricitymap-contrib/search?q=flywheel&type=Code) the project for any mentions of _flywheel_, and besides the original King island issue nothing comes up. I think that either this needs to be grouped with _battery_, or a new storage category might have to be added.

Probably worth asking someone from Tomorrow what the preferred course of action is.

ajoga: Do you know who would be the best to @ to get an opinion about that ?

alixunderplatz: 1. I think adding the flywheel to "battery" is the best option here.
It is used for "very short time storage and production" / for frequency balancing on the island in combo with the shunt resistor (which is an additional load to suppress overfrequency at times). The frequency of the changes in energy direction of the flywheel is of course very high. I'll add a more general comment below regarding that problem
2. "Battery" is "OFF" at the moment but consuming "3kW". In this case, -3kW don't go into the battery itself but probably are consumed for the auxiliary devices and controllers. @robertahunt is this something you'd set to 0 kW when "off" or keep the consumption to account for stand-by emissions?

alixunderplatz:
@ajoga @q-- Related to the flywheel-storage:
as i mentioned above, it's used for short term frequency balancing, and the direction might change quickly when the frequency drops or increases from one second to the other. That's why I have some concerns to include it in the storage part. Of course it is a storage, but only for seconds, unlike the battery.
@robertahunt How often will the parser collect the data? The source data page refreshes every few seconds, wouldn't it make sense to collect 10-15 datapoints to put them into an average for all the types and return the averaged values? This is the most-real-time source that we have, with the most variable production as well. As the system size is very small (up to 3 MW) and as it has a high share of variable sources in it, the frequency of the mini-grid will not be as stable, causing each fetch_production to be a snapshot of the last few seconds and not a timespan of a minute or more. Especially when it comes to storage.

parsers/AU_TAS_KI.py
ajoga: @alixunderplatz in relation to your (1), please note that I didn't do anything with the "resistor" part of [the monitoring](https://data.ajenti.com.au/KIREIP/index.html). I see it as a way to "burn" the excess electricity generated for grid regulation. Maybe the flywheel should be considered like the resistor and just not counted at all ?

robertahunt: Agreed with @alixunderplatz that flywheel could be included under 'battery' to simplify things. And in general for the battery, I think it's numbers should be split across storage and production - when it is charging, this should be included as a negative value under 'storage', and when it is discharging this should be a positive value under production - This is what we have done with hydro storage systems for example in EIA.py and JP_ISEP.py

robertahunt: > @robertahunt How often will the parser collect the data? The source data page refreshes every few seconds, wouldn't it make sense to collect 10-15 datapoints to put them into an average for all the types and return the averaged values? This is the most-real-time source that we have, with the most variable production as well. As the system size is very small (up to 3 MW) and as it has a high share of variable sources in it, the frequency of the mini-grid will not be as stable, causing each fetch_production to be a snapshot of the last few seconds and not a timespan of a minute or more. Especially when it comes to storage.

I am not sure I understand perfectly what is happening here, so correct me if my response doesn't seem to jive. The parsers run approximately every 15 minutes, but we are happy to get the highest granularity of data possible, so if data is available at a 1 minute frequency, we are happy to keep all those datapoints then our processor will take the average.

ajoga:
With all the previous discussions summarized, and please correct me if I'm wrong, the TODO here is : 

- [ ] When the battery system injects power onto the grid, state that as production (details here https://github.com/tmrowco/electricitymap-contrib/pull/2456#discussion_r431203840) 
- [ ] Aggregate the numbers I get from Flywheel & Battery together (pulled from https://github.com/tmrowco/electricitymap-contrib/pull/2456#issuecomment-627617394 )

Is everything there?

robertahunt:
@ajoga from my side I think that is everything :) Not sure if @alixunderplatz has any more to add.

FelixDQ:
@ajoga @alixunderplatz I'm sorry for the staleness on this PR. What do you see as the status of this PR? Is [this](https://github.com/tmrowco/electricitymap-contrib/pull/2456#issuecomment-636438196) the remaining TODOs?

I tried running this and I got an error: 
```
websockets.exceptions.InvalidStatusCode: server rejected WebSocket connection: HTTP 400
```
Do you think something changed with the data source?

I would like to do whatever I can to help this get merged. Is there something that I can do?

q--:
The site is still live, I think the error is because the `connectionToken` used [here](https://github.com/tmrowco/electricitymap-contrib/pull/2456/files#diff-c3b7bb94e70c63d689a577143fcab2a9R11) has expired.

Swapping it for a new token should probably fix it

ajoga:
> @ajoga @alixunderplatz I'm sorry for the staleness on this PR. What do you see as the status of this PR? Is [this](https://github.com/tmrowco/electricitymap-contrib/pull/2456#issuecomment-636438196) the remaining TODOs?

Yeah, I asked for a confirmation of the changes to write before doing them in order to minimize brainpower wastage :)

> 
> I would like to do whatever I can to help this get merged. Is there something that I can do?

Tell me if you agree with the changes in the TODO ; if so I'd like to implement them and have you review them. After that, I have no merge privileges here so we'll need maintainer support. 

Also, my code is written by a python newbie :) and the way I scrape the data is clearly suboptimal. It looks like the page uses SignalR but I failed to getting a proper signalR client to work on it. If you want, you can try that ; this would allow us to remove the hardcoded token in the parser that expired, as you and @q-- found out. 

Let me know.

FelixDQ:
Awesome! From looking at it, I agree with the TODO. 

Ping me when you have completed those, and then I'll try to implement the signalR client.

ajoga:
Folks, 
I've implemented everything from the TODO of https://github.com/tmrowco/electricitymap-contrib/pull/2456#issuecomment-636438196 . 
I've added some comments to explain how the parser works and what to do the next time the token expires. 
I'm sorry for the changes spam, I rebased from master instead of a merge ;( 

The parser now fetches data again : 

```
$ python test_parser.py AUS-TAS-KI
parser result:
{'zoneKey': 'AUS-TAS-KI', 'datetime': datetime.datetime(2020, 9, 28, 0, 12, 23, 614648, tzinfo=tzfile('/usr/share/zoneinfo/Australia/Currie')), 'production': {'battery': -0.005, 'biomass': 0.0, 'coal': 0, 'gas': 0, 'hydro': 0, 'nuclear': 0, 'oil': 1.131, 'solar': 0.0, 'wind': 0.013, 'geothermal': 0, 'unknown': 0}, 'storage': {'battery': 0}, 'source': 'https://data.ajenti.com.au/KIREIP/index.html'}
---------------------
took 2.46s
min returned datetime: 2020-09-27T14:12:23.614648+00:00 UTC
max returned datetime: 2020-09-27T14:12:23.614648+00:00 UTC  -- OK, <2h from now :) (now=2020-09-27T14:12:23.615488+00:00 UTC)
``` 

@FelixDQ if you wanna take a look at the SignalR part, feel free. To me this is good to go

FelixDQ:
Awesome! Thank you @ajoga :-)

I'll try to take a look at SignalR sometime this week, and if it is too complex, then I suggest that we merge this and add SignalR as a future improvement.

FelixDQ (approved):
Alright, I've implemented the SignalR client. 

We can merge this now if it looks good to you @ajoga

ajoga:
@FelixDQ super nice, thank you! I'm gonna take a look at this by next week

parsers/requirements.txt
```diff
@@ -18,3 +18,5 @@ testfixtures==6.0.0
 opencv-python==4.2.0.32
 xlrd==1.1.0
 freezegun==0.3.15
+websockets==8.1
```
q--, line 21: Is this (and the websockets import in the parser) still needed, or does the SignalR library take care of this?

Kongkille:
@ajoga just let me know when you are ready to have it merged!

parsers/requirements.txt
FelixDQ: No, nice catch! I removed it :-)

ajoga:
@FelixDQ @Kongkille this looks good to me, please go ahead! 

Thank you @FelixDQ for the SignalR implementation, and all other contributors for your patience.

Issue #2362, opened 2020-04-08
KW production is actually consumption


robertahunt:
As pointed out by @q-- the fetch_production in KW.py parser is actually fetching consumption. We should modify our script to treat this as consumption

Two things need to happen:
- [ ] parser KW.py should be changed to consumption
- [ ] historic data in database needs to be corrected (I will do this)

alixunderplatz:
@robertahunt Hi! I knew this, but I added it as production, because it must be equal to consumption due to no cross border links and therefore no exchanges with the neigbours.

alixunderplatz:
edit: at least that was what I knew back then. maybe things have changed, hadn't yet looked at the new data. but I'm sure it's still like this.

q--:
@alixunderplatz https://www.gccia.com.sa/Data/PressRelease/Press_43.pdf  page 37 (38th page in the PDF, page number in document is 37) says there's an interconnection between Kuwait and Saudi Arabia

alixunderplatz:
@q-- you're right, there is an interconnector. but is there any info if it is in use? can you look for any figures for import/export? The IEA data didnt show any exchanges back then.
https://openinframap.org/#6.49/29.136/47.699

q--:
Some googling gave me [this]() report

>The Gulf Cooperation Council established a regional power grid to support high voltage networks in member countries in 2001. The intended purpose of the grid was to provide backup  electricity  during  emergencies  caused  by  power  system  outages,  especially during the summer, and to share spinning reserves, optimize capital investments in electricity and reduce fuel costs.<sup>11</sup> The system to date, however, has been underutilized, but the GCC Interconnection Authority (GCCIA), the grid system’s operator, sees potential in expanding electricity trading among countries. The Northern System of the grid was completed in 2009, connecting the Az-Zour complex in Kuwait to grids in Saudi Arabia, Bahrain  and  Qatar.  The  GCCIA  manages  the  interconnection,  and  the  transmission capacity of the interconnector is 1.2 GW. According to the GCCIA, electricity trading by member states has increased over the past decade. Since it commenced trading, Kuwait has imported slightly more electricity than it has exported, but overall trade has generally been unchanged, at around 550 GWh for both imports and exports.<sup>12</sup> Kuwait imports power in an emergency and only for a few hours at a time. Since 2016, Kuwait has exported as much as 750 MW during the summer peak period to Bahrain and the UAE. In return, Kuwait has imported “in-kind” power during winter months.

[...]

PR #2457, opened 2020-05-02
KW production is actually consumption (#2362)

ajoga:
Hello @robertahunt. Is that PR all you expected for #2362, or am I missing something ?
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2362


parsers/KW.py
```diff
@@ -11,19 +11,19 @@
...
+def fetch_consumption(zone_key='KW', session=None, logger=None):
     r = session or requests.session()
     url = 'https://www.mew.gov.kw/en'
     response = r.get(url)
     load = re.findall(r"\((\d{4,5})\)", response.text)
     load = int(load[0])
-    production = {}
-    production['unknown'] = load
+    consumption = {}
+    consumption['unknown'] = load
```
robertahunt, line 21:
```suggestion
    consumption = load
```

robertahunt:
Nice! Thanks @ajoga .

Yes, except when we have a consumption parser we don't split it by fuel type, we just give the consumption straight out, I added a suggestion to your code. Otherwise looks good!

ajoga:
> Yes, except when we have a consumption parser we don't split it by fuel type, we just give the consumption straight out, I added a suggestion to your code. Otherwise looks good!

Thank you for the explanations, I overlooked that one. I accepted your suggestion, thanks

config/zones.json
```diff
@@ -3069,7 +3069,7 @@
       "https://github.com/alixunderplatz"
     ],
     "parsers": {
-      "production": "KW.fetch_production"
+      "production": "KW.fetch_consumption"
```
robertahunt, line 3072: This should also change to consumption! sorry I missed this the first time!
```suggestion
      "consumption": "KW.fetch_consumption"
```

robertahunt:
@alixunderplatz you are correct, that would make the zone appear grey. However, I am very reluctant to save data in the database which is labelled as production but is actually consumption - I think that sets a bad precedent. I am also reluctant to show 'live' carbon intensity values for KW that are completely static because we don't have per-fuel-type information. 

That said, I can understand wanting to display the consumption profile if we have it - perhaps this is something we could add to the frontend to display the consumption profiles for information. I am also curious what other contributors think about this - it would be nice to set a standard. @systemcatch  @jarek

jarek:
My initial thought was that 100% unknown is really pushing it.

But coming to think of it, 100% `some unknown combination of oil and gas` is not really fundamentally different from 100% `various hydro plants which have different lifetime emissions values`. And we definitely have 100% hydro areas.

With the inherent uncertainty we have on carbon intensity values, it's also not really fundamentally different from assuming that all gas plants in Singapore (which make up 95+% of generation) have the exact same intensity. Oil and gas aren't that far apart in CO2eq intensity and averaging them based on yearly statistics probably gives an error of no more than 2x. As a counterexample, it would probably be worse to have an unknown combination of gas and hydro.

I don't think it would also particularly greenwash the region (any more than the IPCC average intensity figures do), so that's also not an issue.

So in summary I think I'm fine with the region having 100% unknown-fossil if you're fine with it :+1:

systemcatch:
@robertahunt I'm not very keen on adding consumption data to the database as production. However don't we do that exact thing already for Israel?

In this instance it's probably ok since there are no significant interconnectors to other zones and I agree with Jarek's comments on the induced errors.

robertahunt:
Alright, sounds like there is a consensus then. Sorry @ajoga, I should have gotten more feedback before opening the issue, but then I suggest we close this, I'll make a note in the readme about this just for the future.

corradio:
Hi all!

Thank you for providing so many insights. I've read through the different threads, and here's what I suggest:
1. Revert this PR so we keep the previous behaviour
2. Add a comment quoting "Kuwait imports power in an emergency and only for a few hours at a time" from https://github.com/tmrowco/electricitymap-contrib/pull/2457#pullrequestreview-408781556 explaining why it's acceptable to use consumption as an approximation for production. Alternatively we can keep the code for `fetch_consumption` so we keep both methods. Although they will return the same data, it will be clearer to some future contributor how to iterate on the functions.

What do you think?

corradio (dismissed):
waiting for changes

FelixDQ (approved):

PR #2465, opened 2020-05-04
update Readme

robertahunt:
Fyi @q-- . Do you think this is acceptable?


README.md
```diff
@@ -434,6 +434,8 @@ We have a public [logger](https://kibana.electricitymap.org/app/kibana#/discover
 
 To get started, [clone](https://help.github.com/articles/cloning-a-repository/) or [fork](https://help.github.com/articles/fork-a-repo/) the repository, and install [Docker](https://docs.docker.com/engine/installation/).
 
+If you want to run python code outside the docker environment, for example when testing a new parser, please use python 3.6 .
```
ajoga, line 437: I think the instructions already present in a later section are enough. I +1 the idea of letting know the contributor that the docker things are not necessary to test the parser, hence this suggestion instead : 
```suggestion
Note : to test parsers, head directly to [Testing parsers locally](#testing-parsers-locally)
```

robertahunt: @ajoga I couldn't find any reference to the version of python (3.6) that we use though. Is that listed somewhere later on?

ajoga: I couldn't either. I'm unsure on what's best to offer to convey this information.

README.md
```diff
@@ -520,7 +522,7 @@ docker-compose run --rm web ./topogen.sh
 For a more detailed explanation of how the map is generated see [here](https://github.com/tmrowco/electricitymap-contrib/blob/master/web/README.md).
 
 ### Testing parsers locally
-In order to test your parser, make sure first that you have installed the required modules as described (consider using a [virtual environment](https://docs.python.org/3/library/venv.html)) in parsers/requirements.txt: for that you can run
+In order to test your parser, make sure first that you have installed the required modules as described, and are using python 3.6 (consider also using a [virtual environment](https://docs.python.org/3/library/venv.html)) in parsers/requirements.txt: for that you can run
```
robertahunt, line 525: @ajoga maybe here?

ajoga: I like this, yes

PR #2474, opened 2020-05-05
Added IT-SIC Capacity From Sources

nessie2013:
Added capacity from terna and entsoe the same as i added IT-SAR capacity


madsnedergaard:
Nice, thanks for doing this! :) 
Would you mind merging master into your branch on your fork? 
Also, what is the source of these capacities - is it from ENTSOE?

nessie2013:
Terna for renewables and ENTSOE for everything else, here is how I said I calculated it for #2387 

"I took the hydro, solar and wind and biomass numbers from Terna, then I used the Entsoe "installed capacity per production unit" source to find the capacity of the installed coal/oil/gas and any unknown capacity, I then also used the hydro capacity from Terna and took the hydro-storage from entsoe (taking the storage capacity out of the given hydro capacity) to get both hydro and hydro-storage."

madsnedergaard:
Ah nice, makes good sense :)
If you don't mind merging master once again, as @martincollignon was faster than me - then I'll watch it and make sure to merge this in next!

PR #2480, opened 2020-05-06
Some ZoneMap hover optimizations

fbarl:
An attempt at improving #2304.


fbarl:
Most of this PR is included in #2481 which seems to be the real performance bottleneck, would revisit further optimizations once that one makes it into production.

PR #2481, opened 2020-05-07
Use CSS for more efficient exchange arrow animations

fbarl:
While looking into how to optimize exchange arrows animations which turned out to be the main cause of #2304, I decided to give CSS animations a try and to my pleasant surprise ended up with what looks to me like a significant performance improvement!

_before_
before

_after_
after

#### Changes

* Fixed and simplified `generate-arrows.js` to only generate static carbon colored arrow images + an extra white arrow, which significantly decreases the number of assets (the script is now in ES6 and can be run via `yarn run generate-arrows`)
* Replaced animated GIFs with CSS animations, rendering a static PNG arrow with a white arrow overlay animated with the moving opacity gradient to achieve the same _sliding highlight_ effect - this potentially addresses #1590 and #2304


corradio:
@fbarl let us know what your conclusions on performance become. As I said I experienced worse perf on Chrome Mac OS.

martincollignon:
@fbarl any updates on this? :)

fbarl:
Hey @martincollignon, I got very excited when I opened this PR as it improved the performance significantly on my machine, but @corradio noticed it actually _makes things worse_ on some other platforms.

I think I'd like to have a play with it a bit more when I catch some free time but if I don't make it work better (or at least equally good) across all platforms within the next few weeks, I guess I'll just close the PR.

fbarl:
Closing for now, will probably pick it up again in a few weeks.

PR #2488, opened 2020-05-09
Additions to README Troubleshooting

rolanddosa:


README.md
```diff
@@ -554,3 +554,13 @@ It should appear on the map as you refresh your local browser.
...
 
 - No website found at `http://localhost:8000`: This can happen if you're running Docker in a virtual machine. Find out docker's IP using `docker-machine ip default`, and replace `localhost` by your Docker IP when connecting.
+
+- FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory: You can configure the memory allocation using `NODE_OPTIONS=--max-old-space-size=4096` in the "build-release" script in package.json, example: "export `NODE_OPTIONS=--max-old-space-size=4096` NODE_ENV=production && webpack --bail --mode production". Also update to the latest node version.
+
+- Expected linebreaks to be 'LF' but found 'CRLF': Fixed by adding `"linebreak-style": 0` in "rules" in .estlintrc
```
fbarl, line 560: I assume this error popped up after you edited a file? I think I'd prefer to keep the rule there but suggest configuring your editor to use LF for line breaks instead of CRLF (the problem is probably Windows specific as Linux and Mac don't use CRLF by default).

I'm not sure how common this problem is, it's probably something Windows users would generally be aware of as it looks like it'd pop up on every JS project that inherits the AirBnB eslint style :thinking:

README.md
```diff
@@ -554,3 +554,13 @@ It should appear on the map as you refresh your local browser.
...
 
 - No website found at `http://localhost:8000`: This can happen if you're running Docker in a virtual machine. Find out docker's IP using `docker-machine ip default`, and replace `localhost` by your Docker IP when connecting.
+
+- FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory: You can configure the memory allocation using `NODE_OPTIONS=--max-old-space-size=4096` in the "build-release" script in package.json, example: "export `NODE_OPTIONS=--max-old-space-size=4096` NODE_ENV=production && webpack --bail --mode production". Also update to the latest node version.
```
fbarl, line 558: > Also update to the latest node version.

How about we suggest trying the node version that actually worked for you - that's a safer advice than assuming the node backwards compatibility on this in the future :slightly_smiling_face: 

I'm also considering adding a `.nvmrc` file to the project to make sure everyone's using the same node version to make these sorts of errors more reproducible in the future..

README.md
```diff
@@ -554,3 +554,13 @@ It should appear on the map as you refresh your local browser.
...
+
+- Expected linebreaks to be 'LF' but found 'CRLF': Fixed by adding `"linebreak-style": 0` in "rules" in .estlintrc
+
+- ERROR: for electricitymap-contrib_mockserver_1  Cannot start service mockserver: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\".../server.py\\\" to rootfs \\\"/mnt/sda1/var/lib/docker/overlay2/.../merged\\\" at \\\".../server.py\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
+   - Check that the project is cloned under C:/Users/
+
+- nodemon not restarting on file changes: try adding the `-L` parameter to use the legacy watch: "server-dev": "nodemon server.js `-L`"
```
fbarl, line 565: :+1:

Maybe just add a link to https://www.npmjs.com/package/nodemon#application-isnt-restarting somewhere here for more context?

README.md
```diff
@@ -554,3 +554,13 @@ It should appear on the map as you refresh your local browser.
...
+
+- FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory: You can configure the memory allocation using `NODE_OPTIONS=--max-old-space-size=4096` in the "build-release" script in package.json, example: "export `NODE_OPTIONS=--max-old-space-size=4096` NODE_ENV=production && webpack --bail --mode production". Also update to the latest node version.
+
+- Expected linebreaks to be 'LF' but found 'CRLF': Fixed by adding `"linebreak-style": 0` in "rules" in .estlintrc
+
+- ERROR: for electricitymap-contrib_mockserver_1  Cannot start service mockserver: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\".../server.py\\\" to rootfs \\\"/mnt/sda1/var/lib/docker/overlay2/.../merged\\\" at \\\".../server.py\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
+   - Check that the project is cloned under C:/Users/
```
fbarl, line 563: Do you know why this solves the problem? More importantly, `C:/Users` implies user is on Windows - perhaps we can create a "**Windows specific**" subsection in "**Troubleshooting**" and move this bullet point there?

README.md
fbarl: Also, if we keep this around, maybe we could put it in the "**Windows specific**" subsection as well? (see the comment below)

README.md
rolanddosa: I read Unix using 'sudo' commands which is similar to increased admin privileges. My project was cloned outside my C:/Users directory, after I moved it there it worked. Might be a coincidence, but could help others

README.md
rolanddosa: Done

README.md
rolanddosa: Done

README.md
rolanddosa: Done

README.md
```diff
@@ -554,3 +554,15 @@ It should appear on the map as you refresh your local browser.
...
 
 - No website found at `http://localhost:8000`: This can happen if you're running Docker in a virtual machine. Find out docker's IP using `docker-machine ip default`, and replace `localhost` by your Docker IP when connecting.
+
+#### Windows Specific
+
+- FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory: You can configure the memory allocation using `NODE_OPTIONS=--max-old-space-size=4096` in the "build-release" script in package.json, example: "export `NODE_OPTIONS=--max-old-space-size=4096` NODE_ENV=production && webpack --bail --mode production". Also update the node version (last working version was v12.16.3).
```
fbarl, line 560:
```suggestion
- `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`. You can configure the memory allocation using `NODE_OPTIONS=--max-old-space-size=4096` in the "build-release" script in package.json, example: `export NODE_OPTIONS=--max-old-space-size=4096 NODE_ENV=production && webpack --bail --mode production`. Also update the node version (last working version was `v12.16.3`).
```

README.md
```diff
@@ -554,3 +554,15 @@ It should appear on the map as you refresh your local browser.
...
+
+- Expected linebreaks to be 'LF' but found 'CRLF': Configure your IDE to use LF instead of CRLF. If that doesn't fix it, add `"linebreak-style": 0` in "rules" in .estlintrc
+
+- ERROR: for electricitymap-contrib_mockserver_1  Cannot start service mockserver: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\".../server.py\\\" to rootfs \\\"/mnt/sda1/var/lib/docker/overlay2/.../merged\\\" at \\\".../server.py\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```
fbarl, line 564:
```suggestion
- `ERROR: for electricitymap-contrib_mockserver_1  Cannot start service mockserver: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\".../server.py\\\" to rootfs \\\"/mnt/sda1/var/lib/docker/overlay2/.../merged\\\" at \\\".../server.py\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type`.
```

README.md
```diff
@@ -554,3 +554,15 @@ It should appear on the map as you refresh your local browser.
...
+   - Check that the project is cloned under C:/Users/
+
+- nodemon not restarting on file changes: try adding the `-L` parameter to use the legacy watch: "server-dev": "nodemon server.js `-L`". See https://www.npmjs.com/package/nodemon#application-isnt-restartin
```
fbarl, line 567:
```suggestion
- nodemon not restarting on file changes: try adding the `-L` parameter to use the legacy watch: `"server-dev": "nodemon server.js -L"`. See https://www.npmjs.com/package/nodemon#application-isnt-restarting
```

fbarl:
Thanks for addressing my feedback @rolanddosa! There were still some typos / markdown formatting inconsistencies left so I suggested some changes, otherwise LGTM and will approve as soon as you address them :rocket:

rolanddosa:
> Thanks for addressing my feedback @rolanddosa! There were still some typos / markdown formatting inconsistencies left so I suggested some changes, otherwise LGTM and will approve as soon as you address them 🚀

Thanks for taking time to review it, I applied your suggestions.

PR #2496, opened 2020-05-11
Add shebang

ajoga:
stupid me keeps invoking `./test_parser.py`, and [that import](https://github.com/tmrowco/electricitymap-contrib/blob/c85368d052a58f40f48070713a65b1eb70464b5e/test_parser.py#L3) is a [valid command](https://linux.die.net/man/1/import)...

### DIFF
--- a/test_parser.py
+++ b/test_parser.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import time
 
 import arrow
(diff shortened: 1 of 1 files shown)


test_parser.py
```diff
@@ -43,7 +45,7 @@ def test_parser(zone, data_type, target_datetime):
     start = time.time()
 
     parser = PARSER_KEY_TO_DICT[data_type][zone]
-    if data_type in ['exchange', 'exchangeForecast']:
+    if data_type in ['exchange', 'exchangzoneeForecast']:
```
corradio, line 48: I think this shouldn't be changed right?

ajoga: Indeed, my apologies for letting that go through in the PR

PR #2501, opened 2020-05-12
Tidy up zone panel question marks

### LINKED ISSUE #2445: Question Mark Appears Next To Generation Type Despite Capacity Data Defined
It is my understanding that question marks should not appear next to generation types in the country panel when capacity information for the country has been defined in the zones.json config file. When reviewing generation mix data for Iceland (I have spot checked other country panels and the behavior appears consistent), the presentation deviates from expected behavior:

image

[Iceland zone.config](https://github.com/tmrowco/electricitymap-contrib/blob/master/config/zones.json#L2575-L2608)

Mentioned in [Slack](https://tmrow-team.slack.com/archives/CLEER5FQQ/p1587602754033900), @corradio suggested I create an issue for it.

fbarl:
image

#### Changes

* Assume electricity production is 0 when capacity is set to 0, reducing the number of question marks shown in the zones panel (see the screenshot above) - fixes #2445
* Did a similar change in the tooltip reusing the new `getElectricityProductionValue` helper
* Hide carbon intensity information from the tooltip when it doesn't apply, e.g. when we know the zone doesn't use that source of electricity
* Put the question mark inside of the `<Row />` component

### DIFF
--- a/web/src/components/countrytable.js
+++ b/web/src/components/countrytable.js
@@ -18,7 +18,11 @@ import {
 import { useCo2ColorScale } from '../hooks/theme';
 import { getTooltipPosition } from '../helpers/graph';
 import { modeOrder, modeColor, DEFAULT_FLAG_SIZE } from '../helpers/constants';
-import { getProductionCo2Intensity, getExchangeCo2Intensity } from '../helpers/zonedata';
+import {
+  getElectricityProductionValue,
+  getProductionCo2Intensity,
+  getExchangeCo2Intensity,
+} from '../helpers/zonedata';
 import { flagUri } from '../helpers/flags';
 import { __ } from '../helpers/translation';
 
@@ -124,6 +128,8 @@ const Row = ({
   index,
   isMobile,
   label,
+  scale,
+  value,
   onMouseOver,
   onMouseOut,
   width,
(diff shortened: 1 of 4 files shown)


corradio:
@fbarl I'm not sure I understand this one:
> Hide carbon intensity information from the tooltip when it doesn't apply, e.g. when we know the zone doesn't use that source of electricity
Could you send a screenshot?

robertahunt:
Looks good to me!.

fbarl:
Thanks for your review @robertahunt!

> > Hide carbon intensity information from the tooltip when it doesn't apply, e.g. when we know the zone doesn't use that source of electricity

> Could you send a screenshot? That would help a lot!

Sure, taking Iceland as an example:

1. We know that it doesn't generate any electricity from solar (the production is `0`)
2. We also don't have any information on its solar production carbon intensity explicitly provided

In cases like these, my idea was that we can confidently say the zone doesn't use this electricity source so we can hide its carbon impact:

image

Compare that to a _full_ tooltip e.g. for hydro:

image

My main idea was to avoid tooltips showing question marks when we actually know what the situation is:

image

@corradio Does that make sense or you think I should revert that change?

Issue #2483, opened 2020-05-08
Ibiza, Formentera & Montenegro deactivated without production


alixunderplatz:
When there is no production on Ibiza or Formentera, the carbon intensity of the islands will not be calculated. We need an exception here, because this is a regular and valid scenario.
Ibiza mostly imports from Mallorca. Formentera mostly imports from Ibiza.
That carbon-chain is often interrupted when one of the zones has no own generation.

For Formentera, when solar returns, the carbon intensity is calculated and displayed on the map.
At night times when importing the entire electricity from Ibiza, the island is grey. But it seems like the default databse value for the imports from Ibiza is used when Ibiza has no production.

![image](https://user-images.githubusercontent.com/25743609/81383723-09e12900-9100-11ea-8cc4-013bfa4e2422.png)

Ibiza seems to have no own production at the moment, because the load is so low (no tourism as usual = low demand). So instead of greying-it-out, it should get the carbon intensity of the imports from Mallorca. This should also be passed on to the imports of Formentera.

https://demanda.ree.es/visiona/baleares/ibiza/acumulada/2020-05-07
https://demanda.ree.es/visiona/baleares/formentera/acumulada/2020-05-07

![image](https://user-images.githubusercontent.com/25743609/81385352-bde3b380-9102-11ea-9e21-e6e87af31851.png)

By the way: The same exception should be included for Menorca<->Mallorac, just in case. 
The interconnector cable between these islands is undergoing reconstruction, after it was damaged back in 2018 or 19. When it'll be back online, it's possible that Menorca's demand will also be covered temporarily by imports from Mallorca.

**EDIT:** Same issue for Montenegro accordIng to the next comment:

![grafik](https://user-images.githubusercontent.com/25743609/81503166-71ef6500-92e2-11ea-85a8-f99abb6ddec3.png)

Fred-3000:
Montenegro has the same issue.

robertahunt:
@alixunderplatz Could this be because the current Ibiza parser returns no data instead of reporting the production as 0?

https://github.com/tmrowco/electricitymap-contrib/blob/db85902207871b01409054cd3eb2b7737b59316f/parsers/ES_IB.py#L99-L101

I am not 100% sure, but I suspect that if we let the parser return 0 production instead of giving an error, the carbon intensity will be calculated, and the imports would be properly included, then we don't need to maintain an exception.

alixunderplatz:
[...]

PR #2502, opened 2020-05-12
Fix parser for Ibiza & Formentera

alixunderplatz:
ref #2483
Let's see if that helps to keep Ibiza and Formentera on the map constantly when they are not producing electricity


parsers/ES_IB.py
```diff
@@ -87,17 +88,13 @@ def fetch_production(zone_key, session=None, target_datetime=None,
...
-## Formentera mostly only has some solar generation during the day, importing from Ibiza all of the time, which probably has to be considered for response.production() > 0: ?
-
     if zone_key == 'ES-IB':
         expected_range = {'coal': (50,600)}
     else:
         expected_range = None
 
     for response in island_data:
-        if response.production() > 0:
+        if response.production() >= 0:
```
robertahunt, line 97: I think we might have to actually remove this if statement - When I was looking at the ree site yesterday it was reporting small negative production values - my guess is onsite-consumption, but I am not sure - anyways, I think they are set to 0 if they are small enough in the validation steps later on (L120)

robertahunt: I just saw your comment on the issue! Let's try this and see! Approved.

Issue #2462, opened 2020-05-03
Offline zones and parser fixes: GB-ORK, GT, US-HI/Oahu, AM, CR and others


alixunderplatz:
- [x] GB Orkney Islands / GB-ORK.py
possible SSL-Error:
`HTTPSConnectionPool(host='www.ssen.co.uk', port=443): Max retries exceeded with url: 
...
(Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)'),))
`
fixed in #2504 
- [x] Guatemala / GT.py
maybe it's timing out? (too many requests?)
`Traceback (most recent call last):
  File "/home/src/electricitymap/feeder/lib/fetch_data.py", line 176, in launch_parsers
    objs = res.get(timeout) ...`
- [x] US Hawaii Oahu / US-HI.py
Arrow package issue? Or is it just the same issue as observed for other US-zones after implementing the new geometries?
`File "/usr/local/lib/python3.6/site-packages/arrow/parser.py", line 224, in parse
    "Failed to match '{}' when parsing '{}'".format(fmt, datetime_string)
arrow.parser.ParserMatchError: Failed to match 'YYYY-MM-DDTHH:mm:ss' when parsing '2020-05-03T13:30:00.000Z'`
Fixed in #2463 

- [x] Armenia / AM.py
See https://github.com/tmrowco/electricitymap-contrib/issues/2611

- [x] Croatia / HOPS.py
`solar = float(solar.replace(",", "."))
AttributeError: 'numpy.float64' object has no attribute 'replace'`
fixed in #2525  

- [x] Cyprus / CY.py
see #2376

- [x] Russia (both zones) / RU.py
see #2461

- [ ] Reunion (temporary) / FR-O.py
setting negative values for oil to "None" or "0" should fix this:
`invalid point: {'zoneKey': 'RE', 'datetime': datetime.datetime(2020, 5, 3, 14, 10, tzinfo=tzoffset(None, 14400)), 'source': 'opendata-reunion.edf.fr', 'production': {'nuclear': 0, 'biomass': 0.0, 'coal': 186.15, 'gas': 2.10047, 'hydro': 55.941, 'oil': -0.21995, 'solar': 56.498999999999995, 'wind': 0.91, 'geothermal': 0, 'unknown': 0}, 'schemaVersion': 3}, reason:RE: key oil has negative value -0.21995`
[...]

PR #2525, opened 2020-05-25
Fix Croatia parser: remove unneeded conversion from string to float

con-cat:
Addresses part of #2462 

This parser was returning the error `solar = float(solar.replace(",", ".")) AttributeError: 'numpy.float64' object has no attribute 'replace'`

It looks like the solar data source is now returning a float rather than a string, so this PR removes the conversion from string to float. This is the result when I test locally:
```
parser result:
[{'zoneKey': 'HR', 'datetime': datetime.datetime(2020, 5, 25, 6, 46, tzinfo=tzutc()), 'production': {'wind': 142.0, 'solar': 13.237, 'unknown': 1038.763}, 'source': 'hops.hr'}]
---------------------
took 4.67s
min returned datetime: 2020-05-25T06:46:00+00:00 UTC
max returned datetime: 2020-05-25T06:46:00+00:00 UTC  -- OK, <2h from now :) (now=2020-05-25T07:10:19.272656+00:00 UTC)
```


con-cat:
I see the test fixtures need updating to reflect the new data format too - I can do that

parsers/test/mocks/hrote.json
```diff
@@ -3,1202 +3,1202 @@
     {
       "ID": "portfolio",
       "Timestamp": "2019-11-03 16:00:00+01:00",
```
fbarl, line 5: Could you actually try to replace the content of this file with [this JSON](https://files.hrote.hr/files/EKO_BG/FORECAST/SOLAR/FTP/TEST_DRIVE/21.5.2020.json)?

My thinking is that it would be good to also update the timestamps since the old timestamps point at the time when the values used to be strings.

con-cat: Thanks for reviewing! I replaced the contents of that file. 

After that I found that for the tests to work I also needed to update dates in the tests and a couple of the other mock files to 21/5/2020 to match that file, and to update two of the assertions in the first test to match the content of the new JSON.

Issue #2530, opened 2020-05-30
Data for Puerto Rico (done: production parser; ToDo: installed capacity, forecast parser)


q--:
- [x] Add production parser - #2531 (need help with a few things)
- [ ] Add capacity data
- [ ] Add forecast parser<sup>Note: would marking 'unknown' renewable also affect the forecast? Because it shouldn't</sup>

---

* [Real-time generation data (+ forecast), updated every 30 minutes](https://aeepr.com/en-us/Pages/Generaci%C3%B3n.aspx)
  * Contains: 
     * Breakdown by fuel type: bunker C ([a type of fuel oil](https://en.wikipedia.org/wiki/Fuel_oil#United_States)), diesel, renewables, coal and natural gas
        * ***Click on the "renovables" part to view a [breakdown by renewable type](https://aeepr.com/es-pr/generacion/Documents/Unidades_renovables.aspx)***
     * Actual generation in MW
     * Generation forecast
* [Installed capacity](https://aeepr.com/es-pr/unidadesgeneratrices)  
  * Note that "Turbines gas" might include or even be 'Diesel GT', otherwise the large amount of generation from Diesel on the fuel type break down I'm seeing right now wouldn't make any sense (compared to the small amount of diesel generators listed).

The only thing that seems to be missing is hydro. It's listed on the _installed capacity_ page (separate from other renewables), but it doesn't seem present in the generation type  breakdown. The separate page with a breakdown of renewable generation doesn't list hydro either.  
At ​98.3MW installed capacity is pretty small though, so that shouldn't be a deal-breaker 

**Edit:** so it seems that wind/solar/landfill gas breakdown doesn't add up to the total generation from renewables. I assume that means the remainder is hydro

---

A few other possibly relevant things

https://aeepr.com/en-us/QuienesSomos/Pages/ElectricSystem.aspx contains the puzzling quote that "69 percent of production is from fuel" (renewable production isn't  even close to 31%). However, that seems about the amount generated right now from bunker + both types of diesel, so they probably mean generation from oil-based fuel and exclude nat. gas and coal.

https://aeepr.com/en-us/qui%C3%A9nes-somos/statisticalreports has links to various statistics. 1st link on this page takes you to https://aeepr.com/en-us/QuienesSomos/Pages/ley57/Estad%C3%ADsticos.aspx, which has annual reports

q--:
Seems like the forecast/demand graph is updated later than the breakdown by generation type;

[...]

PR #2531, opened 2020-05-30
Add Puerto Rico

q--:
Still working on this. Got most of the parser working.

Still To-Do:

- [x] Implement timestamp fetching
- [x] Add some logging in case things go wrong
- [x] Cleanup
- [ ] Add zone info to _zones.json_
  - [x] Add enough to be able to run ` python3 test_parser.py US-PREPA`
  - [ ] Add more stuff like installed capacity
- [ ] Bonus: Add fetching production forecast
   forecast only has total generation, no breakdown by type like real-time data has <sup>~~Note: would marking 'unknown' renewable also affect the forecast? Because it shouldn't~~ **Edit:** it doesn't [x]</sup>
- When downloading the renewables breakdown fails, the parser should fall back to a static carbon intensity:
  - [x] Mark all 'unknown' generation as renewable
  - [x] Calculate static carbon intensity for "unknown" - #2749

And I need someone to

- [ ] Generate the bounding boxes

for me

Related: #2530


parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+        'unknown': 0.0
+    },
+    'storage': {
+        'hydro': -10.0,
+    },
+    'source': 'mysource.com'
+  }
+  """
+
+  global renewable_output
```
systemcatch, line 87: Hey @q-- I don't think you need to set this as a global variable given what it's used for.

q--: To do https://github.com/tmrowco/electricitymap-contrib/pull/2531/files#diff-a83dfd9ee7c1fe0c19c4ca03828e9f5cR183 I had to make it global, otherwise it only updates the value inside the scope of the for loop

And I'm doing the above because in https://github.com/tmrowco/electricitymap-contrib/pull/2531/files#diff-a83dfd9ee7c1fe0c19c4ca03828e9f5cR195 I check if the page with the renewables breakdown was empty (seen it once, only the one empty data item is present) so I can add renewables to the 'unknown' category instead

There might be a more elegant solution, though

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+
+  #Step 1: fetch production by generation type
+  #Note: seems to be rounded down (to an integer)
+  #Total at the top of the page fetched in step 3 isn't rounded down, but seems to be lagging behind sometimes.
+  #Difference is only minor, so for now we will NOT try to parse the total and add the difference to "unknown"
+  res = r.get("https://aeepr.com/es-pr/generacion/Documents/combustibles.aspx")
+
+  assert res.status_code == 200, 'Exception when fetching production for ' \
+                               '{}: error when calling url={}'.format(
+                                   zone_key, url)
```
systemcatch, line 125: This will error because `url` is not defined.

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+      data['production']['hydro'] += renewable_output
+
+  #Step 3: fetch the timestamp, which is at the bottom of a different iframe
+  #Note: there's a race condition here when requesting data very close to <hour>:10 and <hour>:40, which is when the data gets updated
+  #Sometimes it's some seconds later, so we grab the timestamp from here to know the exact moment
+  res = r.get("https://aeepr.com/es-pr/generacion/Documents/CostosCombustible.aspx")#TODO do we know for sure the timestamp on this page gets updated *every time* the generation breakdown gets updated?
+
+  assert res.status_code == 200, 'Exception when fetching timestamp for ' \
+                               '{}: error when calling url={}'.format(
+                                   zone_key, url)
```
systemcatch, line 207: Same error here.

parsers/US_PREPA.py
systemcatch: I can comment out this line and the program runs exactly the same. All we need to do is clean up some of the variable names imo.

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+  raw_timestamp = re.search(r"Ultima Actualización:  ((?:0[1-9]|1[0-2])/(?:[0-2][0-9]|3[0-2])/2[01][0-9]{2}  [0-2][0-9]:[0-5][0-9]:[0-5][0-9] [AP]M)", res.text).group(1)#Extract timestamp
```
systemcatch, line 210: You can use beautiful soup and [arrow's tokens](https://arrow.readthedocs.io/en/latest/#supported-tokens) to make this much easier and more reliable.

systemcatch (dismissed):

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+    Arguments:
+    ----------
+    timestamp_string: timestamp in the format 06/01/2020 08:40:00 AM
+    """
+    timestamp_string = re.sub(r"\s+", " ", timestamp_string)#Replace double spaces with one
+
+    print("PARSED TIMESTAMP",arrow.get(timestamp_string, 'MM/DD/YYYY HH:mm:ss A', tzinfo=timezone_name))
+    return arrow.get(timestamp_string, 'MM/DD/YYYY HH:mm:ss A', tzinfo=timezone_name).datetime
+
+def fetch_production(zone_key='US-PR', session=None, target_datetime=None, logger: logging.Logger = logging.getLogger(__name__)):
```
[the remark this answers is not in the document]
q--: fixed the logging, thanks

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+    'source': 'aeepr.com'
+  }
+
+  renewable_output = 0.0 #Temporarily stored here. We'll subtract solar, wind and biomass (landfill gas) from it and assume the remainder, if any,  is hydro
+
+  #Step 1: fetch production by generation type
+  #Note: seems to be rounded down (to an integer)
+  #Total at the top of the page fetched in step 3 isn't rounded down, but seems to be lagging behind sometimes.
+  #Difference is only minor, so for now we will NOT try to parse the total and add the difference to "unknown"
+  res = r.get("https://aeepr.com/es-pr/generacion/Documents/combustibles.aspx")
```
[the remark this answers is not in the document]
q--: done

parsers/US_PREPA.py
q--: fixed this and the other two places where this occurs

parsers/US_PREPA.py
```diff
@@ -0,0 +1,225 @@
...
+    if(category == "BUNKER C" or category == "DIESEL CC" or category == "DIESEL GT"):
+      data['production']['oil'] += outputInMW
+    elif(category == "GAS NATURAL"):
+      data['production']['gas'] += outputInMW
+    elif(category == "CARBON"):
+      data['production']['coal'] += outputInMW
+    elif(category == "RENOVABLES"):
+      renewable_output += outputInMW #Temporarily store aggregate renewable output. We'll subtract solar, wind and biomass (landfill gas) from it and assume the remainder, if any,  is hydro
+    else:
+      print("HEY! UNKNOWN CATEGORY! " + category)#TODO log warning if unknown category detected
```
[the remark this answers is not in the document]
q--: Done, thanks

parsers/US_PREPA.py
q--: I already use arrow in `convert_timestamp`. Personally I think regular expressions are easier to read (and almost certainly use less memory than an HTML parser), so I kept this

systemcatch: Correct on the memory usage but the bottleneck for this parser is likely in the network requests. I guess we'll have to disagree on the readability part.

FelixDQ:
Hey @q--, what is the status on this? Do you need my help or input on some of this? :-)

q--:
Right, I should get back to finishing this... Need to figure out if that last code suggestion is accurate

It's related to a scenario where renewable generation is reported but the exact renewables breakdown, which is on another page, is unavailable. I now intend to report this renewable generation as "unknown" and mark it as renewable. Or would it be preferred to throw an error in that case (and not return anything)?

As for help, once I've added Puerto Rico to zones.json I need someone to generate the bounding boxes. And I've got a question,  because forecast only has total generation, no breakdown by type like real-time data has, would marking 'unknown' renewable also affect the forecast? (It should not.)

FelixDQ:
> It's related to a scenario where renewable generation is reported but the exact renewables breakdown, which is on another page, is unavailable. I now intend to report this renewable generation as "unknown" and mark it as renewable. Or would it be preferred to throw an error in that case (and not return anything)?

Yes, reporting it as unknown and marking it as renewable is probably the best solution.

> As for help, once I've added Puerto Rico to zones.json I need someone to generate the bounding boxes. 

Alright, just ping me when you have added the zone and then I can generate the bounding boxes.

> And I've got a question, because forecast only has total generation, no breakdown by type like real-time data has, would marking 'unknown' renewable also affect the forecast? (It should not.)

No, marking it as unknown is fine.

parsers/US_PREPA.py
q--: That's probably because that logic only kicks in when the graph data at https://aeepr.com/es-pr/generacion/Documents/Unidades_renovables.aspx is empty; it's a backup plan to avoid having the parser fail completely when the data on that page is unavailable.

It was required for that when I added it when writing the parser, so I assume it still is today. Leaving it it in now as said page is working correctly right now so I can't test this easily and I don't want to break the parser.

q--:
@FelixDQ Parser should be mergable. Someone else should generate the bounding boxes. I've left [this](https://github.com/tmrowco/electricitymap-contrib/pull/2531#discussion_r507809278) in in order to avoid breaking things.

This PR doesn't take care of capacity data or forecast, so I'm leaving  #2530 open to deal with those

config/co2eq_parameters.json
```diff
@@ -771,6 +771,13 @@
           "value": 399.77646355214347
         }
       },
+      "US-PREPA": {
+        "unknown": {
+          "_comment": "Source: https://indicadores.pr/dataset/generacion-consumo-costo-ingresos-y-clientes-del-sistema-electrico-de-puerto-rico/resource/fdad4f42-a4be-48bb-9478-a8fb75c000c6",
+          "source": "Based on average renewable generation September 2019-September 2020 (12% hydro, 57% solar, 31% wind)",
```
veqtrus, line 777: Off-by-one error: It's Oct-19 to Sep-20. A full year is Jan-Dec not Jan-Jan.

q--: Whoops

Fixed, thanks

FelixDQ:
Awesome job @q--! I have generated the new zone & updated the references from the old "PR" zone to "US-PREPA". We can merge this now if you agree with my commits :-)

q--:
Somehow I had missed that Puerto Rico has its own country code.

Since it's also used as e.g. their TLD I think it might be better to use PR for everything then. Everything else LGTM

FelixDQ (approved):
Cool! Let's merge then :-) 

Nice job!

PR #2535, opened 2020-06-02
Updated Italian

Raffox97:
* Added missing strings
* Edited some zone names
* Removed unused strings

### DIFF
--- a/web/locales/it.json
+++ b/web/locales/it.json
@@ -99,6 +99,7 @@
         "maintitle": "Emissioni di CO₂ in tempo reale del consumo elettrico",
         "oops": "Oops! Stiamo avendo dei problemi nel raggiungere il server. Proveremo di nuovo tra poco!.",
         "newversion": "È disponibile una nuova versione! Clicca <a onClick=\"location.reload(true);\">qua</a> per ricaricare.",
+        "webgl-not-supported": "La mappa non può essere mostrata perché questo browser non supporta WebGL.",
         "retrynow": "Riprova ora",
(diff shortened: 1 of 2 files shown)


web/locales/it.json
```diff
@@ -420,17 +421,17 @@
...
         "CL-SEM": {
             "countryName": "Cile",
-            "zoneName": "SEM"
+            "zoneName": "Sistema de Magallanes"
         },
-        "CL-SEA": {
+        "CL-SEN": {
             "countryName": "Cile",
-            "zoneName": "SEA"
+            "zoneName": "Sistema Eléctrico Nacional"
```
fbarl, line 434: Could you please provide a source for these 3 zone names? If that's how these zones are indeed called, we might want to re-use these zone names across all other translations. cc @robertahunt

Raffox97: Sure, there is a page on the Spanish wikipedia. Here's the link: https://es.wikipedia.org/wiki/Interconexi%C3%B3n_el%C3%A9ctrica_de_Chile :)

fbarl: Haha ok good point, the second sentence already explains it all:

> Hasta 2017 existían cuatro sistemas intercontectados que operaban aisladamente de los otros, pero ese año se unieron los dos principales –el Sistema Interconectado Central (SIC) y el Sistema Interconectado del Norte Grande (SING)– en el Sistema Eléctrico Nacional (SEN). Al SEN se le suman el Sistema de Aysén (SEA) y el Sistema de Magallanes (SEM), que operan en la zona austral del país.

Could you please update the english locale `en.json` with the same zone names while we're at it? Thanks!

Raffox97: Done :)

Issue #2499, opened 2020-05-11
Allow contribution via email


GitMatze:
Currently, clicking on a blank region prompts users to participate via GitHub. This is probably too big a hurdle for non-programmers (i.e. 98% of the people). Simply providing an email address or a contact form would make it a lot easier for users to leave valuable hints and sources.

Sorry if this has been discussed before.

fbarl:
I like your idea @GitMatze!

The closest thing in the app right now to my knowledge is the feedback form (which also nudges users to reach out via email):

![image](https://user-images.githubusercontent.com/1216874/81567539-b8a19580-939c-11ea-9d55-ca8bf180e257.png)

Perhaps we can extend on this and provide something similar also when clicking on a blank region?

![image](https://user-images.githubusercontent.com/1216874/81567685-ed155180-939c-11ea-8589-d4edbe70abfe.png)

GitMatze:
To avoid getting the same hints from different people over and over again we could show a little info message for each region briefly summarizing the current status on available and missing data plus the link to the relevant Github issues for those who want to know more. This will get people started right away and I feel like people might  also be intrigued to see if they can find out more.

robertahunt:
@GitMatze I really like the idea of giving people a summary of the current status of the region. I am more skeptical of e-mail feedback as we already receive a lot of e-mails on a daily basis and sorting through them is already quite a chore, so redirecting everyone to github where the eMap community can help direct and shape ideas and feedback is much preferred from my perspective. However, I think we could do a lot to make the github community more accessible for people - add more information to the readme, give a better introduction as to how we work, and what questions should be directed where, and like you said, point to relevant issues for each region.

GitMatze:
I totally get you @robertahunt. 
Seems like there is a GitHub API. So maybe a pretty neat alternative to email would be to provide a contact form that automatically posts a  message in a collective issue on GitHub? For further inquiries they could also leave their email address privately to the tomorrow team.

robertahunt:
That sounds like a good idea. If there is a way to auto-generate a github issue that would be really nice.

GitMatze:
There is a way, indeed. The GitHub API is, in fact, super extensive.
I created a little demo and it works like a charm. https://github.com/GitMatze/githubAPI

robertahunt:
[...]

PR #2542, opened 2020-06-08
Feedback Form with Github API [Demo]

GitMatze:
em_demo

Here's a demo of a feedback form that allows to post on GitHub directly from the site using the GitHub API.
See Issue: https://github.com/tmrowco/electricitymap-contrib/issues/2499

+GUI is fully functional
+API is fully functional
-It's not functionally integrated into the map ( when changing countries etc. )
-Right now it requires a Github Token exposed in the source code. You have to replace it with your own just like in this demo: https://github.com/GitMatze/githubAPI/blob/master/githubAPI.png

`feedbackform_class_syntax.js` is equivalent to  `feebackform.js`, just written in class syntax


corradio:
Hi @GitMatze ,
Pretty cool! How does it work when someone is not logged into github?
I'm cc'ing my colleagues @Kongkille and @FelixDQ who will be able to follow up.

GitMatze:
Not sure what you mean... This is how I thought this could work: 
You create a github user named website-contributions. Using a token of this account you can make it automatically post messages, either by opening a new issue for every contribution or by commenting on a generic issue created for this purpose. 
This cannot be implemented in the frontend, however, because the secret token would be exposed. So I thought you send the user input to the backend, store the contact data for potential further enquiries and post the message on github using the API.

corradio:
That answers my question - makes sense.
I'm leaving on vacation but I'll let @Kongkille and @FelixDQ take over the discussion and assess whether or not this is something that can be introduced in the product.
In any case, thanks for taking the time pushing this!

Kongkille:
I've done something similar in the past. It uses the Github API to open issues. These issues would then have the description text that the user has posted on the form.

I think it's a cool implementation, however, there are some things that may worry me when issues can be created so fast.

1. Higher chance of duplicate issues. 
2. Easier issue-creation may also lead to lower-quality issues.
3. Reduced engagement. We **want** people to join the Github and follow their issues through, hopefully staying on and contributing to other parts of the map.
4. Is there a cap on the number of issues there can be in a repository?

Right now, we already have such a high amount of issues, and some of them dating back quite a bit. I like the idea of making issue-creation easier for users who are not familiar with Github, but I am afraid that this might lead to an overflow of low-quality issues causing us to overlook more important issues that people have spent a great amount of time on.

GitMatze:
Yep, cluttering the page with auto-generated issues is probably not a good idea. That's why I thought to create a generic issue that collects all website contributions. You could simply browse through them every once in a while and check if there's anything of value. 
I'm not sure what all this would yield but I feel like it could potentially open up quite a useful input stream. There sure  are a lot of people out there that would quickly drop a link but wouldn't take the hurdle to engage on github, be it due to lacking time, english skills, motivation or just because they have no idea what github is.
Anyway, I totally see your points and it might also be quite some work to implement this...

Kongkille:
I absolutely agree with your point that getting input from those people who may not be comfortable with Github would be a great source of input. We just have to find a way in which it does not affect the input stream of those who can and want to contribute directly.

I am going to close this for now, but can I tag you if I open up an issue related to this in the future?

GitMatze:
Sure!

PR #2547, opened 2020-06-11
Raise validation error if netFlow of an exchange is of greater magnitude than the interconnector capacity.

pierresegonne:
## Problem
If the data is incorrect, it can feed the system some exchanges with flows greater than the interconnector capacities. 

## Goal
Raise a Validation error when this is detected so that the issue with the data appears in Kibana and thus makes realising that wrong data emanates from the sources easier.


pierresegonne:
@corradio not sure how I can test it though

utils/config.py
```diff
@@ -34,6 +35,11 @@ def relative_path(script_reference_path, rel_path):
         if zone_name_1 not in ZONE_NEIGHBOURS:
             ZONE_NEIGHBOURS[zone_name_1] = set()
         ZONE_NEIGHBOURS[zone_name_1].add(zone_name_2)
+        if (zone_name_1 not in INTERCONNECTORS) and ('capacity' in v):
+            INTERCONNECTORS[zone_name_1] = {}
+        INTERCONNECTORS[zone_name_1][zone_name_2] = {
```
corradio, line 40: isn't that redundant with EXCHANGES_CONFIG[f'{zone_name1}->{zone_name2}'] ?

pierresegonne: 🤦 Yes indeed that makes everything simpler

parsers/lib/quality.py
```diff
@@ -41,6 +42,18 @@ def validate_exchange(item, k):
...
+    # Verify that the exchange flow is not greater than the interconnector capacity
+    # Use https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/example.py for expected format
+    if item.get('sortedZoneKeys', None) and item.get('netFlow', None):
+        zone_names = item['sortedZoneKeys']
+        if len(zone_names) == 2:
+                if ((zone_names in EXCHANGES_CONFIG) and
+                    ('capacity' in EXCHANGES_CONFIG[zone_names])
+                ):
+                    interconnector_capacities = EXCHANGES_CONFIG[zone_names]['capacity']
+                    if not (min(interconnector_capacities) <= item['netFlow'] <= max(interconnector_capacities)):
```
corradio, line 54: Capacity data might be outdated or imprecise. I would suggest adding a margin of error (10%?). What do you think?

pierresegonne: Good idea indeed, but is there to way to evaluate what's the "expected" level of discrepancy between what data we have and the truth?

corradio: I'm not sure to be honest. We could start with 10% and see if any exchange fail?

q--: >Capacity data might be outdated or imprecise. I would suggest adding a margin of error (10%?). What do you think?

Note that the number might also be inflated due to transmission losses, especially over long distances

PR #2549, opened 2020-06-12
Cordova update & fixes

FelixDQ:
CC @fbarl 

The scrolling issue turned out to be because of this setting 😬 
```
<preference name="DisallowOverscroll" value="true" />
```

The plugins also seem to work now, but we should keep an eye out if we upgrade them later on. Firebase for example breaks with the newest version.


corradio:
Thank you so much Felix! You're da 👑

Issue #78, opened 2016-09-21
ENTSOE: Invalid values for coal and gas Netherlands


ThierryOllivero:
Basic stuff (total production / total consumption / imports / exports) can be found here: http://energieinfo.tennet.org/dataexport/exporteerdatacountry.aspx

Did not find an actual split by generation type, yet.

corradio:
ENTSOE data is lacking production:

![image](https://cloud.githubusercontent.com/assets/1655848/20111464/b4754a60-a5e8-11e6-95c1-a13fc2c0a0be.png)

rustyb:
Hi Folks - just to let you know we're looking into the issue with generation per type for the Netherlands on the ENTSO-E transparency platform.

I can confirm that the 1MW values for coal is the same value delivered to us by the dutch TSO('s). Will keep you posted here.

PS. I'm the web guy at ENTSO-E

corradio:
Updating this issue: invalid values for gas and coal:

![image](https://cloud.githubusercontent.com/assets/1655848/22364405/9d28bd0c-e471-11e6-8261-1205225abb55.png)

corradio:
There's hope:
https://twitter.com/FlorenceKofi/status/839133868322062336

MariusKroon:
The Netherlands is still showing up as grey on the map yesterday afternoon and this morning, although there are values available on ENTSO-e for production per type for the Netherlands. 
The production per type bars are correctly showing the values for the respective fuel types, but somehow the calculation for the emission is not working. 

Btw; The bar graph shows question marks for fuel type that are not used in the Netherlands, there are no power plants that run on e.g. Lignite or Oil. ENTSO-e has a specific report available that returns the reported installed capacity per production type [here](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show) - it might be more graceful to base the overview of total installed capacities on this data instead of the static file in the config folder.

corradio:
Hi @MariusKroon,

[...]

PR #2561, opened 2020-06-28
Smarter NL parser

PaulCornelissen:
Since the ENTSO-E data for the Netherlands is still very much incomplete (#78 #1871), I've tried to make a smarter parser:
* It revives the old energieopwek.nl parser that uses the very up to date estimates for solar and wind made by Martien Visser et. al..
* It combines this with ENTSO-E numbers for nuclear, gas, coal, biomass and unknown.
* It checks whether the sum of all these numbers is at least equal to ENTSO-E consumption minus nett imports.

In my opinion the NL data has these issues:
Only some data is reported. Mainly I think that only data from the main TSO is reported. So anything happening on the DSO grid is missed. That's why wind and solar are structurally very low; because most wind farms almost all solar feed into the DSO grid instead of the TSO grid. Coal and nuclear seem to be fairly accurate because there's 3 resp. 1 and they all feed into the TSO net. A lot of gas generation is also missing because NL has a lot of small CHP's (IIRC 4,5GW) and those feed into the DSO grid (mostly?). Possibly also some smaller gas units do this.
Also the consumption data is only the consumption that flows via the TSO grid. That's why especially when there's a lot of solar generation this number is very low. So we can treat this as a minimum, but actual consumption is higher and sometimes a lot higher. That's why I've chosen to add the difference between total generation + nett imports and this consumption number as "unknown" when it has a positive value. That way we should get closer to the actual amount of generation.


q--:
If I understand correctly your parser will first have the Energieopwek renewable production "eat" the 'unknown' generation part of the ENTSOE mix before adding "on top of" the mix, right?

If so, how do you account for the fact that a lot (most?) of the solar production is (residential) rooftop solar, which from the TSO perspective mostly results in reduced consumption? Should that still 'eat' into the 'unknown' part? Because the other 'unknown'  generators probably won't all have shut down at that point, (gas-fired CHP, biomass plants etc.)

I'm thinking it might make sense to always add solar 'on top' of the mix instead of having it eat into the unknown part first, if that makes sense.

PaulCornelissen:
@q-- It works slightly different. There's two kinds of unknown production:
* Unknown source production reported by ENTSO-E (let's call that entsoe-unknown)
* Unknown source production that's not reported, but what we can calculate, because all ENTSO-E prodcution, energieopwek.nl solar & wind and net imports combined are less than the consumption reported by ENTSO-E. (Let's call this calculated-unknown)

Entsoe-unknown production is never "eaten" by solar and wind. It's just added on top. It's perfectly possible that the sum of all production and net imports is more than the consumption reported by ENTSO-E, because in the real world the actual production is higher than the ENTSO-E reported consumption. 

Calculated-unknown is "eaten" by solar and wind. I don't think we can assume that no significant (or even the majority) part of solar and wind is actually in that calculated-unknown production, so that's why have to let it "eat" that to stay on the conservative side.

q--:
Lots of renewables across much of Europe yesterday. I took a look at the graph for the Netherlands and noticed very little 'unknown' output, which to me suggests that most of the 'unknown' output would be (gas-fired) cogen (especially on days with much 'unknown' output).

Screenshotted ElectricityMap snipped showing very little 'unknown' output when there are high levels of renewables

PaulCornelissen:
@q-- That's not unlikely. However that mostly matters when we want to calculate a custom CO2 intensity for unknown. That is a good idea, but not part of this PR.

@corradio Who can give some feedback on this or possible approve and merge?

corradio:
I will look at it together with @FelixDQ .
Sorry for the delay - I have been swamped lately! Thanks for looking at it.

FelixDQ:
Hi @PaulCornelissen
I'm very sorry for the delay. Really nice work on this!

Before we merge this I just want to get some insights from ENTSOE to make sure that our assumptions are correct. I have reached out to them and I will update here as soon as I get a response.

We can hopefully get this merged soon :-)

PaulCornelissen:
Hi @FelixDQ,

Good to hear! 

Yes, that around the time that the DK->NL interconnector was added. I believe it was added a bit later officially (IIRC), but maybe that's the date they updated the statistics. So it makes sense to hard-code to zero before that.

FelixDQ:
Awesome! Would you be able to do that in this parser? :-)

PaulCornelissen:
Yes probably, time allowing.

FelixDQ:
Hey @PaulCornelissen, do you think that you have time to do the final change or should I go ahead and finish it up? :-)

tth4vb:
Hey @PaulCornelissen :) Checking to see if you had time to make the final change, otherwise we'll see if we can find time in our dev cycle to finish!

PaulCornelissen:
Sorry, started a new job recently which had me quite preoccupied. I started working on this 3x al ready, but I found another larger bug. I at least pushed the fix for the DK interconnector not working for very historic dates. But that other bug remains. I just had a good look at it: 

@FelixDQ what is the exact expected behavior of the `--target_datetime` option? Because the ENTSOE parser doesn't actually return any values for the specified date, rather it only returns data for 2 days _preceding_ the specified date. I find this "up to, but not including" interpretation not very intuitive tbh, but if that's the expected behavior, then I'll adjust the NL parser. 

(Currently the e-opwek parser gets values for the specified date also and because the ENTSOE parser doesn't return any values for the specified date, the merging of the two sets goes wrong.)

tth4vb:
@PaulCornelissen no worries! @FelixDQ is on vacation but can get back to you next week :)

PaulCornelissen:
@tth4vb With the last commit the parser now only returns datapoints for which both ENTSOE and energieopwek.nl return data. Because of the way the ENTSOE handles the --target_datetime this means the 26hrs (due to timezones) preceding the specified date are returned. 

If that's ok, then I think this can now be merged.

FelixDQ:
This looks good to me.  🎉 

I need to run some tooling and look at the data when merging this. I'll do this next week at the latest.

FelixDQ (approved):
Let's merge 🎉

Thanks for the great work @PaulCornelissen !

mhooi:
I was wondering wat the reasoning is behind the assumed 700 gCO2eq/kWh carbon intensity for the unknown part of the Netherlands? We know that the majority of the electricity is generated by Gas stations currently so a value around the 490 gCO2eq/kWh would be a better guestimate I think for the Netherlands.

PaulCornelissen:
@mhooi There was non. 700 gCO2eq/kWh is the universal number for unknown.

See also this issue: #738 

You are probably right though. All coal production (there's only 4 plants left) is probably reported as coal in ENTSOE. NL is unlikely to have any significant oil based generation (like the rest of continental Europe), so that means unknown is probably all gas of some kind (CHP, gas turbine and CCGT). You should open a new issue and explain this reasoning a bit further and probably collect some data/sources to back it up and then propose a custom value for unknown.

PR #2563, opened 2020-06-29
Drag and drop states (debug tool)

pierresegonne:
## What

Add a sneaky drag and drop zone on emap front-end where a state can be dropped to display the emap based on any state from the gcloud storage

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -31,6 +31,7 @@
     "moment": "^2.23.0",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
+    "react-dropzone": "^11.0.1",
     "react-map-gl": "^5.2.5",
     "react-motion": "^0.5.2",
     "react-portal": "^4.2.1",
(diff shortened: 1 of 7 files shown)


fbarl:
Wow that seems like a very handy debugging feature @pierresegonne!

I'm probably the last one who's made major changes to the frontend code so please let me know if you need any help proceeding :wink:

pierresegonne:
Hey @fbarl 

Yes you could definitely take over on this one, I left it in the open because we didn't define exactly how to integrate it properly (I am not sure we want it on prod) and because the priority on this is very low.

corradio:
cc @FelixDQ

FelixDQ:
I'll take over this one, and then we'll consider how to integrate it properly.

Kongkille:
@FelixDQ given the time this has been up, I think we should probably close it. We can perhaps still look to do it in the future, but I'm not sure exactly how it should be implemented, so it might be easier to just close it.

FelixDQ:
yea okay. It's a pretty neat debug feature, but we will probably not use it anytime soon. So closing it, for now, makes sense

PR #2565, opened 2020-07-02
Computed fallback emission factor for CA-QC.

### LINKED ISSUE #2507: Current Québec ( Canada ) export intensity impossible
Hi,
While we don't have yet the live data from province of Québec in Canada, the value of its carbone intensity is **301 gCO₂eq/kWh** shown currently on electricitymap.org is simply not possible, the vast majority of the generation of electricity there is hydro and some wind.
We should see more a number like **20** or **25 gCO₂eq/kWh**. Maybe 30.
But 301, it's simply impossible with all the installations of large hydro, and very little biomass and thermal.

pierresegonne:
Resolves #2507

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -1218,6 +1218,24 @@
           "wind": 0.9982474825220405
         }
       },
+      "CA-QC": {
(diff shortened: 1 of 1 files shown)


corradio:
FYI I had created this internal issue https://github.com/tmrowco/tmrow/issues/3210 as the fallback carbon intensity could be computed from the fallback mix

PR #2567, opened 2020-07-02
Switch parser for exchange between GE and RU to GE.

### LINKED ISSUE #2273: Data Available for Exchange RU-1->GE
more data about Power flow than actual map https://www.electricitymap.org/ on http://www.gse.com.ge/for-customers/data-from-the-power-system/power-flow

pierresegonne:
Resolves #2273

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -1236,7 +1236,7 @@
       43.158267
     ],
     "parsers": {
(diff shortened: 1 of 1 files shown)


systemcatch:
Probably best to let @jarek review this as he was involved in the issue discussions.

PR #2576, opened 2020-07-06
Panama: classify biogas as 'biomass'

q--:
The Panama parser currently lists 'biogas' under 'gas', it should be listed as 'biomass' instead.

I looked up the only biogas installation [producing electricity right now](http://sitr.cnd.com.pa/m/pub/gen.html), [Cerro Patacón](https://cdm.unfccc.int/Projects/DB/TUEV-SUED1355992480.72/view?cp=1). This uses landfill gas.

Historical data should probably fixed as well. Since 'biogas' is the only thing the parser stores as 'gas', that could be done by simply moving all the gas production to biomass.


FelixDQ:
Thank you for the PR @q-- ! 💪

PR #2585, opened 2020-07-10
Ps/add checks outliers

pierresegonne:
## What

Added plausibility checks on parsed data. Namely we now check that production (one factor) and consumption (all combined) don't exceed 500GW and exchanges don't exceed 100GW. 

Might need to update these numbers or make an exception if China joins in

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -30,7 +30,7 @@ def validate_consumption(obj, zone_key):
         raise ValidationError('%s: consumption has negative value '
                               '%s' % (zone_key, obj['consumption']))
     # Plausibility Check, no more than 500GW
-    if abs(obj.get('consumption', 0)) > 500000:
+    if obj['consumption'] is not None and abs(obj['consumption']) > 500000:
         raise ValidationError('%s: consumption is not realistic (>500GW) '
                               '%s' % (zone_key, obj['consumption']))
     validate_reasonable_time(obj, zone_key)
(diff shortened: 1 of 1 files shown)


parsers/lib/quality.py
```diff
@@ -29,6 +29,10 @@ def validate_consumption(obj, zone_key):
     if obj['consumption'] is not None and obj['consumption'] < 0:
         raise ValidationError('%s: consumption has negative value '
                               '%s' % (zone_key, obj['consumption']))
+    # Plausibility Check, no more than 500GW
+    if obj['consumption'] is not None and abs(obj['consumption']) > 500000:
```
corradio, line 33: You can do 
```suggestion
    if abs(obj.get('consumption', 0)) > 500000:
```
It basically will fallback to be 0

parsers/lib/quality.py
```diff
@@ -99,4 +108,9 @@ def validate_production(obj, zone_key):
         if v < 0:
             raise ValidationError('%s: key %s has negative value %s' %
                                   (zone_key, k, v))
+        # Plausibility Check, no more than 500GW
+        if abs(v) > 500000:
```
corradio, line 112: No need to check for `abs` as you know `v >= 0`

pierresegonne:
@corradio the suggestion `if obj['consumption'] is not None and abs(obj['consumption']) > 500000:` -> `if abs(obj.get('consumption',0)) > 500000:` fails, as object does not have the get method

Issue #2573, opened 2020-07-05
Exchange arrows don't move when resizing the map horizontally


q--:
Resize the browser window horizontally. The map behaves correctly, but the arrows don't move. (They don't jump to the right spot after you're done resizing either.)

Resizing vertically works fine.

Occurs on both Chrome and Firefox, so it probably happens for all browsers.

Kongkille:
Hi :) I just tried myself and I can't seem to replicate this. I see that it seems they do not move to the right spot while resizing the window, however when I am done resizing it seems to work. Can you perhaps attach a .gif or a video?

I've tested on Chrome & Safari.

q--:
![ElectricityMap horizontal resize bug](https://user-images.githubusercontent.com/2470175/87421625-88ae8280-c5d7-11ea-96ab-d0095ea01019.gif)

Kongkille:
Ah I see. I resized it using the bottom right corner and only dragged horizontally, my bad. Thanks for finding this :)

fbarl:
Good catch @q--!

It seems that the source of this issue is that [react-map-gl](https://github.com/visgl/react-map-gl) `onViewportChange` hook gets called when the map viewport height gets updated, but _not when the width does_. I'm tempted to report an issue on their repo as I didn't find anything similar online, but before I do I'd like to make sure it's not something caused by our CSS :thinking: 

In any case, their is a simple workaround if we want to move fast - we can simply insert our own separate width listener in [layout/map.js](https://github.com/tmrowco/electricitymap-contrib/blob/aa523a67aa821b2005c7c94dc3cf55451c44c1c4/web/src/layout/map.js) - I tried this piece of code and it fixed the issue:

```javascript
const width = useWidthObserver(ref);

useEffect(() => {
  handleViewportChange({ ...viewport, width });
}, [width]);
```

fbarl:
I opened visgl/react-map-gl#1157 to try to address the issue in the library, but in the meantime I addressed it via the above suggested hack in #2589.

PR #2589, opened 2020-07-13
Responsive UI fixes

fbarl:
A couple of surgical fixes to the responsive aspect of the UI based on some recent issues / comments.

#### Changes

* Fix left side panel overflow on mobile - addresses https://github.com/tmrowco/electricitymap-contrib/commit/58a5d495125fecf0a73fd236d1cd906afd54067f#r40174062
* Look at screen width instead of isMobile for `/info` page redirect - the new `useIsSmallScreen` / `useIsMediumUpScreen` React hooks can be used instead of CSS screen size breakpoints when more of the styles is moved to React components #2261
* Update exchange arrows under window width resizing - fixes https://github.com/tmrowco/electricitymap-contrib/issues/2573

### DIFF
--- a/web/src/hooks/viewport.js
+++ b/web/src/hooks/viewport.js
@@ -45,3 +45,36 @@ export function useHeightObserver(ref, offset = 0) {
 
   return height;
 }
+
+function useWindowSize() {
+  const [windowSize, setWindowSize] = useState({});
+
+  // Resize hook
+  useEffect(() => {
+    const updateSize = () => {
+      setWindowSize({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    };
+    // Set initial size
+    updateSize();
+    // Add window listeners
+    window.addEventListener('resize', updateSize);
+    return () => {
+      window.removeEventListener('resize', updateSize);
+    };
+  });
+
+  return windowSize;
+}
+
+export function useIsSmallScreen() {
+  const { width } = useWindowSize();
+  return width < 768;
+}
+
+export function useIsMediumUpScreen() {
+  const { width } = useWindowSize();
+  return width >= 768;
+}
--- a/web/src/hooks/viewport.js
+++ b/web/src/hooks/viewport.js
@@ -46,7 +46,7 @@ export function useHeightObserver(ref, offset = 0) {
   return height;
 }
 
-function useWindowSize() {
+export function useWindowSize() {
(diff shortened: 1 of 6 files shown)


web/src/layout/map.js
```diff
@@ -146,16 +147,37 @@ export default () => {
...
 
+  // Update the map viewport whenever the window width or height changes.
+  // TODO: This is a hack that should ideally not be needed, but without it
+  // the exchange arrows don't seem to update well under resizing the width.
+  // See https://github.com/tmrowco/electricitymap-contrib/issues/2573#issuecomment-658242753
+  const { width, height } = useWindowSize();
+  useEffect(() => {
+    handleViewportChange({ ...viewport, width, height });
+  }, [width, height]);
+
```
Kongkille, line 180: Do you know if this affects performance? I know we have had a lot of issues with the exchange arrows and the performance in the past.

Even still this is probably still the best solution.

Just curious, in the bug found by q--, it seemed to only occur when dragging horizontally, that is when the width changes and the height stay constant. Do we then need to use this "hack" for both height and width changes, or could we just use it for width changes only?

fbarl: > Do you know if this affects performance?

I tested it quickly and it seems like the answer is not really. Also, I think it's important to note that window resizing doesn't happen that often and the performance in fixed viewport should be completely unaffected by this code.

> Do we then need to use this "hack" for both height and width changes, or could we just use it for width changes only?

You're right, in this particular case I think we can remove the height listener but I decided to keep it as I wasn't entirely sure if not tracking the width is a bug or rather tracking the height is an accidental _feature_ so I added both here to be on the safe side regarding the future changes :slightly_smiling_face:

Kongkille: Good points, thanks :)

web/src/hooks/viewport.js
```diff
@@ -45,3 +45,36 @@ export function useHeightObserver(ref, offset = 0) {
...
+
+export function useIsSmallScreen() {
+  const { width } = useWindowSize();
+  return width < 768;
+}
+
+export function useIsMediumUpScreen() {
+  const { width } = useWindowSize();
+  return width >= 768;
+}
```
FelixDQ, line 80: NIce hooks :-)

fbarl: Thanks :wink: And thank you for the review! :pray:

Issue #541, opened 2017-05-11
Add Japan


corradio:
Seems like real-time data source is available here:
http://agora.ex.nii.ac.jp/earthquake/201103-eastjapan/energy/electrical-japan/

jarek:
From brief exploration on 2017-05-11, I didn't see a real-time generation breakdown by type.

From http://agora.ex.nii.ac.jp/earthquake/201103-eastjapan/energy/electrical-japan/stat/01/ (aggregate, not real time) it looks majority thermal/fossil.

http://agora.ex.nii.ac.jp/earthquake/201103-eastjapan/energy/electrical-japan/stat/25/ (again aggregate over past few years) says thermal gen is ~55% gas, 40% coal.

Maybe we can ask via Twitter if they know of real-time breakdowns: https://twitter.com/electricaljapan

corradio:
Great. I'll ask from the Tomorrow account!

corradio:
Twitter thread: https://twitter.com/electricalJapan/status/864240082764546048

corradio:
I've sent an email to one of the professors at NIL.

alixunderplatz:
I may have found a data source for **Kyushu**, the southern main island of Japan:
It seems that _data is mainly displayed in "ten thousands" kilowatts_ in the graphs and charts.

http://www.kyuden.co.jp/power_usages/pc.html
Here is demand, solar generation and other lines (forecast?).
![image](https://user-images.githubusercontent.com/25743609/36150677-b3f6bbb2-10c4-11e8-93d8-6e4b56a96e8e.png)

Some installed capacities for RES are here:
http://www.kyuden.co.jp/effort_renewable-energy_application

I also managed to find some .csv-files that you can download, but we'd need somebody knowing the Japanese language to help us search that site for info for generation by type, exports and imports, etc... 

There are many buttons that cannot be translated that easily, so I OCRed some and had them translated afterwards^^

Maybe @happysalada, who was involved in the Japanese translation, could help?

happysalada:
ok so quick update
on [the page you mentioned](http://www.kyuden.co.jp/power_usages/pc.html)
there is a big button to download the csv of the day data. Next to it is a button with all past data. The only data given though is power consumption.
Past data is given by day. A daily csv includes data by hour and by 5 minutes.

On their graph they show the percentage of power production that is solar, however that data didn't show up on the csv.

[...]

PR #2604, opened 2020-07-24
Fetches nuclear production for JP-KN

Kongkille:
Based on research from #541 (thanks a lot for the help)

It fetches from https://www.kepco.co.jp/energy_supply/energy/nuclear_power/info/monitor/live_unten/, which seems to update every 20 minutes or so. The scraper will most certainly fail if changes are made to the HTML, it's not robust at all.

It tries to match the nuclear production with solar production if it does not find a match within 30 minutes, it will fail.

Let me know if there are any code or methodology improvements I should implement! :)


config/zones.json
```diff
@@ -2937,7 +2937,7 @@
     ],
```
FelixDQ, line 2937: Remember to add you name here 😄

parsers/JP_KN.py
```diff
@@ -0,0 +1,151 @@
...
+    JP_data = JP_fetch_production(zone_key, session, target_datetime, logger)
+    nuclear_mw, nuclear_datetime = get_nuclear_production()
+    latest = JP_data[-1] #latest solar data is the most likely to fit with nuclear production
+    diff = None
+    if nuclear_datetime > latest["datetime"]:
+        diff = nuclear_datetime-latest["datetime"]
+    else:
+        diff = latest["datetime"]-nuclear_datetime
+    if abs(diff.seconds) > 30 * 60:
+        raise Exception("Difference between nuclear datetime and solar data is too large")
```
FelixDQ, line 46: A nitpick, but it's the difference between nuclear data and the rest of the data right (solar + unknown)? Maybe change the comments to be "JP data" or "production data".

parsers/JP_KN.py
```diff
@@ -0,0 +1,151 @@
...
+
+    latest["production"]["nuclear"] = nuclear_mw
+    latest["production"]["unknown"] = latest["production"]["unknown"] - nuclear_mw
+    return latest
+
+# https://stackoverflow.com/questions/50655738/how-do-i-resolve-a-tesseractnotfounderror
```
FelixDQ, line 52: Was this link meant to be here? Maybe move it up to the import

Kongkille:
@FelixDQ 

For installing tesseract (I believe it is already installed, but here they are for linux)
- sudo apt update
- sudo apt install tesseract-ocr
- sudo apt install libtesseract-dev

And for installing the japanese language (I am not sure this works, since I tested on Mac)

- sudo apt-get install tesseract-ocr-jpn

PR #2605, opened 2020-07-24
Monthly update of Poland solar capacity

wojciej:
Monthly update of Poland solar capacity according to newest tweet of PSE
https://twitter.com/pse_pl/status/1282562359375077376

I have also added the following values so we would get rid of "?" in the displayed data:
* geothermal:0
* battery storage: 0
* unknown: 0

Source file has been also updated with newest url to PSE tweet


jarek:
Thanks!

PR #2607, opened 2020-07-27
Nuclear emissions for french nuclear

DrDrebon:
Added a zone override for french nuclear power : 
  - use of low carbon electricity and centrifugation for enrichment  of  uranium fuel allow a lower footprint  (6 instead of 12)

Source ADEME : https://www.bilans-ges.ademe.fr/documentation/UPLOAD_DOC_FR/index.htm?conventionnel.htm


FelixDQ:
Thank you @DrDrebon !

Before we merge this, I just need to find out how much this affects. If this affects our forecast etc. then we might need to wait with merging this until we will run our pipeline.

I will get back to you! Thank you for the contribution :-)

DrDrebon:
Thank you.

config/co2eq_parameters.json
```diff
@@ -326,6 +326,10 @@
         "hydro discharge": {
           "source": "2019 average by Tomorrow",
           "value": 48.14997796838915
+        },
+        "nuclear": {
+        "source": "ADEME",
+        "value": 6
         }
```
FelixDQ, line 333:
```suggestion
        "nuclear": {
          "_url": "https://www.bilans-ges.ademe.fr/documentation/UPLOAD_DOC_FR/index.htm?conventionnel.htm"
          "source": "ADEME 2015",
          "value": 6
        }
```

DrDrebon: Ok, it is not as if there was an emergency on the matter, so it's quite fine.

For my curiosity what do you call "your pipeline" ?

I saw there was a pull request for a regional parser in France, do the regional override work with hereditary (FR_XXX inherits from FR unless an override is defined for FR_XXX) ?

Thanks  best regards

DrDrebon:
> Ok unfortunately we need to wait until we will run the pipeline. It will probably be in few weeks :/
> 
> We are currently rewriting the pipeline, so in the future, this will be much faster.
> 
> Sorry for the delay.
Ok, it is not as if there was an emergency on the matter, so it's quite fine.

For my curiosity what do you call "your pipeline" ?

I saw there was a pull request for a regional parser in France, do the regional override work with hereditary (FR_XXX inherits from FR unless an override is defined for FR_XXX) ?

Thanks best regards

FelixDQ:
Hey @DrDrebon. We had some talks about this, and we are unfortunately not ready to add country-specific overrides. It is very important that countries are comparable, and we don't have a process that can ensure this for country-specific overrides yet. So, unfortunately, I have to close this PR for now, but thank you for your contribution!

Our next priority in this regard is to add the ecoinvent factors which are peer-reviewed and published. They will also give us country-specific emissions. Take a look at https://github.com/tmrowco/electricitymap-contrib/issues/738 if you want to see more.

DrDrebon:
Well, it seems a bit disappointing to me :

   This value is issued from peer review papers collected by ADEME.

   You already have some zone override (a lot of hydro, and even some Coal Power Plants override).

PR #2627, opened 2020-08-23
Sweden Nuclear CO2-eq override based on EPD.

VIKTORVAV99:
I found a EPD that includes the majority of the Sweden nuclear plants (2 out of 3) that has a CO2-eq of 4.1gCO2-eq/kWh and I think it's safe to assume the last nuclear plant has a similar environmental impact.

Link to EPD: [Vattenfall EPD](https://gryphon4.environdec.com/system/data/files/6/17371/S-P-00923%20EPD%20Electricity%20from%20Vattenfall%20Nordic%20Nuclear%20Power%20Plants%202019.pdf)

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -684,6 +684,12 @@
           "source": "2017 average by Tomorrow",
           "value": 46.02408724838775
         },
+        "nuclear": {
+          "_comment": "Assumes the CO2-eq for the nuclear plant not included in the EDP to be the same as the Vattenfall nuclear plants.",
+          "_url": "https://gryphon4.environdec.com/system/data/files/6/17371/S-P-00923%20EPD%20Electricity%20from%20Vattenfall%20Nordic%20Nuclear%20Power%20Plants%202019.pdf",
+          "source": "Vattenfall EPD",
+          "value": 4.1
+        },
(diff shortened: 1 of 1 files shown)


FelixDQ:
Hey @VIKTORVAV99. We had some talks about this, and we are unfortunately not ready to add country-specific overrides. It is very important that countries are comparable, and we don't have a process that can ensure this for country-specific overrides yet. So, unfortunately, I have to close this PR for now, but thank you for your contribution!

Our next priority in this regard is to add the ecoinvent factors which are peer-reviewed and published. They will also give us country-specific emissions. Take a look at https://github.com/tmrowco/electricitymap-contrib/issues/738 if you want to see more.

VIKTORVAV99:
Hi @FelixDQ, while I understand your decision I'd just like to point out that the EPD use Ecoinvent 3.5 and the GaBi Professional database 2019 for generic data.

So even if you wait for Ecoinvent data directly it's unlikely to differ by a significant amount since the EPD covers upstream sources down to end storage.

PS: I added a comment with the same data #738 but haven't seen any activity in the thread since.

FelixDQ:
Yes, I think you are right. We just need to change all our emission factors at the same time.

PR #2635, opened 2020-08-30
Bump arrow to 0.16.0

systemcatch:
The latest version of arrow contains support for PEP 495 (ambiguous datetimes). This will help with any future DST shifts and stop some of the strange behavior currently seen when these occur.


FelixDQ (approved):
Neat! Thank you @systemcatch :-)

Issue #2376, opened 2020-04-14
Data for country Cyprus


alexaristotelous:
Cyprus doesn't seem to be showing data anymore. 
When checking the "List of sources" Cyprus has:
<http://www.dsm.org.cy/en/daily-system-generation-on-the-transmission-system-mw>

A link that returns HTTP 404. 

The new link is here:
<https://tsoc.org.cy/total-daily-system-generation-on-the-transmission-system/>

It's a shame but they seem to have removed pages in English.
If you need assistance translating anything from Greek please let me know.

jarek:
Thank you!

Looks like the data format has changed, but it should be possible for us to use the new one.

As you might have local knowledge, can you take a look at the page and confirm the following:
- which timezone do the hours seem to be in?
- is the blue line total load? green line wind production? purple line thermal/fossil fuel/conventional production? 
- Google translates the label for the yellow line as "Photovoltaics and Biomass". Do you happen to know or can you find how much is solar and how much is biomass? Currently in the middle of the night it appears to show a constant 5 MW yellow line - does that mean 5 MW is biomass?
- is it still the case that all non-solar and non-wind generation is fuelled by oil?

The chart is drawn based on values included in the HTML source, which look to be generated automatically. The HTML source has several parts like this:

```
        			var dateStr = "2020-04-16";
			var hourStr = "03";
			var minutesStr = "45";
			var dateStrFormat = new Date(dateStr);
			dateStrFormat.setHours(hourStr, minutesStr);
						if (synchronizedGenerationView) {
				forecastValues.push(
				[dateStrFormat, 265, 19, 5, 265, 241]
			);
			} else {
				forecastValues.push(
				[dateStrFormat, 19, 5, 265, 241]
			);
			}
        			var dateStr = "2020-04-16";
			var hourStr = "04";
			var minutesStr = "00";
			var dateStrFormat = new Date(dateStr);
			dateStrFormat.setHours(hourStr, minutesStr);
							tempDate = dateStrFormat;
				if (counter % 2 == 0) { 
					ticksString.push(dateStrFormat);
				}
				counter = counter + 1;
						if (synchronizedGenerationView) {
				forecastValues.push(
				[dateStrFormat, 265, null, null, null, null]
			);
			} else {
				forecastValues.push(
				[dateStrFormat, null, null, null, null]
			);
			}
```

[...]

PR #2637, opened 2020-09-01
Reimplement CY parser and update capacities

veqtrus:
Scraping https://tsoc.org.cy/total-daily-system-generation-on-the-transmission-system/ and returning only the latest stats for now.

Biomass and solar are lumped together and are estimates anyway so I report them only as solar. They are labelled as 'distributed' sources so I assume it includes rooftop solar (hence estimates). More discussion in #2376.

I think it will be possible to get historical data but would like to get the basics working for now. Tested with

    $ python3 CY.py 
    fetch_production() ->
    {'zoneKey': 'CY', 'production': {'oil': 635.0, 'solar': 6.0, 'wind': 0.0, 'biomass': None}, 'storage': {}, 'source': 'tsoc.org.cy', 'datetime': datetime.datetime(2020, 9, 2, 1, 45, tzinfo=datetime.timezone(datetime.timedelta(seconds=10800)))}


jarek:
Thanks for the update! I haven't yet had a chance to look into it in depth but on first glance is looks great. Since we're changing the data source URL, can you also update it in the main README in "Real-time electricity data sources" section?

veqtrus:
Done. Is there a way to test the map visualisation locally? Ideally I would like to only generate updated Cypriot data.

veqtrus:
The estimated solar+biomass is flat 6MW during the night so I will report it as 6MW biomass always and subtract that from what is reported by TSOC to get solar.

q--:
Since biomass output will vary depending on which plants are running, assuming it'll always be 6MW will lead to incorrect results. (The installed capacity for biomass is In fact 12.4 MW, so more generation might come online.)




An alternative could be to take the "solar" and "biomass" electricity generation reported by the [IEA](https://www.iea.org/countries/cyprus "Select 'Electricity generation by source'") and calculate what percentage of either contributed in order to derive a static carbon intensity.

The biomass/solar figure could then be reported as "unknown" with a carbon intensity override, like it's done for e.g. [Ukraine](https://github.com/tmrowco/electricitymap-contrib/pull/1812/files)

veqtrus:
According to https://tsoc.org.cy/total-daily-wind-and-solar-farm-generation/ (translated):

> NOTES:
> 1. The Production of Wind Farms is channeled to the Transmission System, while the production of Photovoltaic Parks is distributed to the Distribution System.
> 2. The Production of Photovoltaic Parks in the Distribution System is an estimate

So from what I gather solar+biomass is produced by small operators (like rooftop PV) and hence their exact productions are likely unknown to the TSO. So the value reported in the chart is likely to be flat 6MW for biomass. Solar OTOH is likely an estimate from weather conditions (it gets more variable in winter).

The previous parser attributed solar+biomass to solar alone. IMO subtracting that 6MW is reasonable for solar, while for biomass we could return `None` to indicate that we don't have exact data, if reporting 6MW feels wrong. I don't think we should report solar as unknown given that the share of biomass is so low.

FelixDQ (approved):
Sorry for the delay. I've been a bit swamped lately. 
I think your assumption about biomass/solar is reasonable. It would also be great if we could add historical data at some point, but this is fine for now

Let's merge 🎉 Great job @veqtrus!

veqtrus:
So do you want it 6MW flat like it is now for biomass or `None`?

FelixDQ:
I think reporting it flat as `6MW` like it is now will give us the best result as it would otherwise not be counted at all.

Reporting it as unknown would probably be a bit more accurate, but given as the share of biomass is so low, I think this solution is fine.

Do you guys agree?

PR #2642, opened 2020-09-04
Panama: report fossil plants with known fuel type as coal/oil/gas

q--:
Reports two known coal plants as 'coal' after extracting their generation from the generation by unit breakdown on the page

**Update:** now also reporting a fuel-oil fired plant as 'oil'
**Update 2:** after investigating the Costa Norte plant I've concluded all units are generating electricity from natural gas
**Update 3:** Parser now avoids reporting negative generation

Sources are listed in parser


q--:
For future reference, I've also done some research into other fossil plants but haven't added these to the pull request for various reasons. I'm documenting these here.

The pull request can be merged as it currently is.

#ACP Minaflores 2,5,6,7,8,9,10

A few sources have these as oil, but I don't know how up-to-date and reliable these are.  Also don't seem to cover all units.

https://www.industryabout.com/country-territories-3/2031-panama/fossil-fuels-energy/31047-miraflores-6-8-oil-power-plant
http://globalenergyobservatory.org/geoid/43554

FelixDQ (approved):
Awesome! Nice work @q-- ! 💪

PR #2645, opened 2020-09-06
Update Panama installed capacity

q--:
Also updated the source. To be exact, I used the data from the second sheet of the first [spreadsheet](http://168.77.210.79/energia/wp-content/uploads/sites/2/2020/08/2-CEE-1970-2019-GE-Generaci%C3%B3n-El%C3%A9ctrica.xls) linked on that page.

As I'm working on reporting output from fossil plants by their fuel type in #2642, it is no longer necessary to put all fossil capacity under "unknown".


FelixDQ (approved):
Awesome! Looks good @q--. 

I fixed a small syntax error. (json doesn't allow trailing comma's)

q--:
O, so that's what the test failure was about. I don't have a CircleCI account so I just assumed it would be because installed capacity wasn't rounded.

I'll undo my rounding of the capacity then

q--:
Undid the rounding by going back to the older commit and rebased the branch on top of upstream/master, which also contains your comma fix, then force-pushed.

Should be good to merge now

FelixDQ:
Ok awesome!

PR #2654, opened 2020-09-14
Add Himachal Pradesh (India) parser (#2686)

willbeaufoy:
This is a partial implementation of a parser for the Himachal Pradesh province of India.

It uses the real time data page from the Himachal Pradesh SLDC: https://hpsldc.com/intra-state-power-transaction/ with corroborating information from here: http://meritindia.in/state-data/himachal-pradesh

The data page is a bit hard to interpret, below is my interpretation:
* State power plants are listed in the table GENERATION OF HP(Z)
* ISGS (Inter-State Generating Station) plants (i.e. they are owned by multiple states) are listed in the table (B1)ISGS(HPSLDC WEB PORTAL)
* Total production from state power plants is (as of now) ~1.1GW, while total production from ISGS plants is ~2.3GW. Total demand in Himachal Pradesh is ~1.1GW (see https://hpsldc.com), therefore Himachal Pradesh must be a large electricity exporter. This makes sense given its large hydropower resources, and the populous neighbouring states, e.g. see https://www.downtoearth.org.in/coverage/drowned-in-power-43881
* The page http://meritindia.in/state-data/himachal-pradesh lists what % of the ISGS plants capacity is allocated to Himachal Pradesh state; is is generally around 15%.
* Therefore I believe that, roughly, Himachal Pradesh's electricity production is the total of the two tables, while its consumption is the total of the state plants table + the percentage of the ISGS plants table allocated to the state (given in the meritindia page).
* Therefore we can definitely calculate the total production, and probably the total consumption as well.
* Every known plant in the data source is a hydro plant, therefore this will be good to show the extent of renewable energy in India. I believe that some of the unknown sources may be hydro as well (the microp ones), will need to look more into this.

This draft pull request only does production. If we decide to go ahead with this I will look at implementing price and consumption.


FelixDQ:
Nice job @willbeaufoy! This looks very promising. I agree with your interpretation of the data. I linked this PR in the issue to see if anyone else has input, but I would be fine with continuing with this approach.

willbeaufoy:
> Nice job @willbeaufoy! This looks very promising. I agree with your interpretation of the data. I linked this PR in the issue to see if anyone else has input, but I would be fine with continuing with this approach.

Great thanks for checking it out. I'll continue with my current approach then.

parsers/IN_HP.py
```diff
@@ -0,0 +1,163 @@
...
+        return rows
+    except (AttributeError, ValueError):
+        raise Exception('Error reading table {}'.format(table_name))
+
+
+def get_gen_type(desc):
+    """Gets the generation type for the plant described by the given
+    description string. Raises an error if the description is not recognised."""
+    for plant_type in PLANT_NAMES_TO_TYPES.items():
+        if plant_type[0] in desc.upper():
```
systemcatch, line 145: We want to be careful with this sort of partial matching, it might be better to use `==` and use the exact plant names provided (`BASPA(3x100MW)`) but I understand the full names are kind of annoying.

Also if you're using `.items()` here you can doing the following for readability.

```python
    for plant_name, plant_type in PLANT_NAMES_TO_TYPES.items():
        if plant_name in desc.upper():
            return plant_type
```

parsers/IN_HP.py
```diff
@@ -0,0 +1,163 @@
...
+    Data is from the table titled (B1)ISGS(HPSLDC WEB PORTAL)"""
+    table_name = '(B1)ISGS(HPSLDC WEB PORTAL)'
+    gen = {GenType.HYDRO.value: 0.0, GenType.UNKNOWN.value: 0.0}
+    # Read all rows except the first (headers) and last two (OTHERISGS and Total).
+    for row in get_table_rows(soup, 'table_4', table_name)[1:-2]:
+        try:
+            cols = row.find_all('td')
+            if not cols[0].has_attr('class'):
+                # Ignore COMPANY column.
+                del cols[0]
```
systemcatch, line 120: This won't give us an ordering problem right?

systemcatch (changes requested):
Hey @willbeaufoy I left a few comments but overall I think this is a good parser given how mangled the data is.

willbeaufoy:
> Hey @willbeaufoy I left a few comments but overall I think this is a good parser given how mangled the data is.

Thanks, am addressing your comments now, then I will add additional commits to complete the parser. However I'm a bit confused about which functions I'm meant to be adding, could you clarify?

* The main `README` says to just add `fetch_production()`, which is all I've done so far.
* `parsers/example.py` has `fetch_production()`, `fetch_price()` and `fetch_exchange()`. I could add `fetch_price()` but I can't add `fetch_exchange()` as I haven't found any data for which states Himachal Pradesh's exports go to.
* Many other IN-* parsers have a `fetch_consumption()` function in addition to `fetch_production()` (but no `fetch_exchange()`). However I cannot see any difference either on the map or on the panel for any of the states when switching the toggle between production and consumption. Is `fetch_consumption()` still in use, and if so should I add it? I'm still unsure as to whether my data source provides this info, I thought it might be the 'NR.HPSCHEDULE' column in the '(B1)ISGS(HPSLDC WEB PORTAL)' table, but then that would mean that total production for use within HP (i.e. this column + the GENERATION OF HP(Z) table) is about double current demand (demand as seen on https://hpsldc.com/), which seems wrong (though I don't know much about actual power generation systems).

parsers/IN_HP.py
```diff
@@ -0,0 +1,163 @@
...
+    res = r.get(url)
+    assert res.status_code == 200, 'Exception when fetching production for ' \
+                                   '{}: error when calling url={}'.format(
+                                       zone_key, url)
+    soup = BeautifulSoup(res.text, 'html.parser')
+    return {
+        'zoneKey': ZONE_KEY,
+        'datetime': arrow.now(TZ).datetime,
+        'production': combine_gen(get_state_gen(soup), get_isgs_gen(soup)),
+        'source': url
```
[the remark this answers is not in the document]
willbeaufoy: Done

parsers/IN_HP.py
```diff
@@ -0,0 +1,163 @@
...
+    r = session or requests.session()
+    if target_datetime is None:
+        url = DATA_URL
+    else:
+        raise NotImplementedError(
+            'This parser is not yet able to parse past dates')
+    res = r.get(url)
+    assert res.status_code == 200, 'Exception when fetching production for ' \
+                                   '{}: error when calling url={}'.format(
+                                       zone_key, url)
```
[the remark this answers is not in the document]
willbeaufoy: Done

parsers/IN_HP.py
willbeaufoy: Done, using exact matches actually allows me to simplify the code.

parsers/IN_HP.py
willbeaufoy: Not quite sure what you mean, what this does is ensures we always get the entry in the second column (STATION) and not the first one (COMPANY). For some rows the first `<td>` is for COMPANY (e.g. `<td rowspan="5">NHPC</td>`), but for most it is for STATION.

willbeaufoy:
> > Hey @willbeaufoy I left a few comments but overall I think this is a good parser given how mangled the data is.
> 
> Thanks, am addressing your comments now, then I will add additional commits to complete the parser. However I'm a bit confused about which functions I'm meant to be adding, could you clarify?
> 
>     * The main `README` says to just add `fetch_production()`, which is all I've done so far.
> 
>     * `parsers/example.py` has `fetch_production()`, `fetch_price()` and `fetch_exchange()`. I could add `fetch_price()` but I can't add `fetch_exchange()` as I haven't found any data for which states Himachal Pradesh's exports go to.
> 
>     * Many other IN-* parsers have a `fetch_consumption()` function in addition to `fetch_production()` (but no `fetch_exchange()`). However I cannot see any difference either on the map or on the panel for any of the states when switching the toggle between production and consumption. Is `fetch_consumption()` still in use, and if so should I add it? I'm still unsure as to whether my data source provides this info, I thought it might be the 'NR.HPSCHEDULE' column in the '(B1)ISGS(HPSLDC WEB PORTAL)' table, but then that would mean that total production for use within HP (i.e. this column + the GENERATION OF HP(Z) table) is about double current demand (demand as seen on https://hpsldc.com/), which seems wrong (though I don't know much about actual power generation systems).

Looking at the other Indian parsers again I may have misunderstood the `fetch_consumption()` function. Is it just to get the total current demand? If so I could do that just from the live data on https://hpsldc.com/. However I'm still unsure if this function is still used.

parsers/IN_HP.py
willbeaufoy: Now added some explanatory comments.

willbeaufoy:
> Hey @willbeaufoy I left a few comments but overall I think this is a good parser given how mangled the data is.

I've now added a unit test, and from my perspective this parser is now complete. It still only has `fetch_production()` though, see my previous comment for my question about whether/how to implement `fetch_consumption()`, and I asked a similar question about how to implement `fetch_price()` in [slack](https://tmrow-team.slack.com/archives/C2DQPTHA6/p1600454488006100).

FelixDQ:
Awesome! This looks really good. Thanks for adding a test :-) 

`fetch_price()` is shown on the map like this. It's cool to see the relationship between price and production mix. I would add it if you have a good source, but it is not required.
image


`fetch_consumption()` is a little bit misleading. The consumption view on the map doesn't use this function but instead calculates the consumption by taking into account the exchanges. We currently don't have any exchanges in India which is why there isn't a difference between the two views. I don't think this function is actually used by the live map, so feel free to skip that one. I think that it is a possible parameter for the forecasts.


Does this answer your questions?

willbeaufoy:
> Awesome! This looks really good. Thanks for adding a test :-)
> 
> `fetch_price()` is shown on the map like this. It's cool to see the relationship between price and production mix. I would add it if you have a good source, but it is not required.
> image
> 
> `fetch_consumption()` is a little bit misleading. The consumption view on the map doesn't use this function but instead calculates the consumption by taking into account the exchanges. We currently don't have any exchanges in India which is why there isn't a difference between the two views. I don't think this function is actually used by the live map, so feel free to skip that one. I think that it is a possible parameter for the forecasts.
> 
> Does this answer your questions?

Thanks for the reply, I'll leave `fetch_consumption()` out for now then. For `fetch_price()`, my source only has something called 'DSM rates' (i.e. demand side management rates) (see the right hand side [here](https://hpsldc.com/)). I've just read a bit more about DSM and now I believe (though I'm not sure) that this refers to a payment made by the grid to consumers who reduce their consumption at times of high demand, therefore it is not appropriate as a measure of electricity price. So unless you say differently I'll leave out `fetch_price()`.

I've added the final bits to the pull request so I think it's ready to go, I will change the status from draft.

parsers/IN_HP.py
systemcatch: I was thinking more about the risk of mutating an object while iterating over it but tbh it probably doesn't matter much here.

systemcatch (approved):
All good from my perspective.

willbeaufoy:
> All good from my perspective.

Great, is there anything further I need to do to get this merged? I haven't merged changes from master into it yet as it might become out of date again before it's merged.

FelixDQ (approved):
Let's merge! 🎉

Nice job @willbeaufoy and thanks for reviewing @systemcatch!

PR #2656, opened 2020-09-15
Fix #2568

### LINKED ISSUE #2568: The import/export between Menorca and Mallorca appears reversed
Our data is showing the export going one way, but our data source is showing the reverse according to one user. See screenshots:
IMG_9768
image

q--:
I checked https://demanda.ree.es/visiona/baleares/mallorca/total and the import/export is indeed reversed. This should fix it.


FelixDQ (approved):
Nice catch!

PR #2658, opened 2020-09-15
Updated Capacity of "DE" zone

Bohne13:
updated capacity of "DE" zone. Data is coming from "Frauenhofer ISE" 
https://www.energy-charts.de/power_inst_de.htm?year=2020&period=annual&type=power_inst
last data update was on 1.09.20 at 22:55

PR #2659, opened 2020-09-15
Update Germany production capacity

Bohne13:
updated the source of production capacity from Germany

***
Edit by @FelixDQ:
Ref: https://github.com/tmrowco/electricitymap-contrib/pull/2658


FelixDQ (approved):
Next time, you can do this in the same PR. That makes it a bit easier to reference later on, but this is fine for now.

Thank you @Bohne13 :-)

PR #2664, opened 2020-09-16
Cyprus: dynamically determine the biomass portion of solar+biomass generation

q--:
Instead of a hardcoded value, use the combined solar+biomass generation
at midnight to determine which portion of said generation constitutes
biomass


FelixDQ (approved):
Ah neat! That's a good idea

PR #2668, opened 2020-09-16
List international capacity sources in README

q--:
Having a list of international sources for installed capacities should be useful when adding a new region or capacities for an existing region. This PR adds a few of these above the list of capacities by country.

I went through the capacity sources list to find useful international sources, and added a few others.


FelixDQ (approved):
Good idea! :-)

PR #2670, opened 2020-09-16
Panama: categorize remaining thermal plants by fuel type

q--:
Report the output from the remainining thermal power plants listed at
https://sitr.cnd.com.pa/m/pub/gen.html that were not already categorized
in #2642 by their respective fuel types

Only the fuel type for the BLM 9 plant remains ambiguous


FelixDQ (approved):
Cool! Thank you @q--! :-)

PR #2671, opened 2020-09-17
Added IT-SO Renewable Capacities

nessie2013:
Regions included in IT-SO are: "Calabraia, Apulia, Basilicata, Molise"


FelixDQ (approved):
Thank you @nessie2013 !

PR #2672, opened 2020-09-17
Added IT-CSO Renewable Capacities

nessie2013:
(hydro is not split into normal and storage, as this source doesn't have that data)
Regions included in IT-SO are: "Campania, Lazio, Abruzzo"


FelixDQ (approved):
Thank you @nessie2013 ! :-)

PR #2673, opened 2020-09-17
Give battery storage a distinctive color

q--:
Currently, battery storage has the same color as "unknown".

It would be nice if it had its own color.

I went with this color because it's distinct enough from the other production type colors.

The color #af5bca


Kongkille (approved):
LGTM,

since we have one for hydro storage, we might as well have one for battery as well. I also like that it doesn't really relate to a specific color, as we don't know the exact footprint of the stored electricity

Issue #2638, opened 2020-09-03
ENTSOE data retrieval for the Netherlands keeps failing due to negative values for  "unknown"


q--:
[Kibana](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:extra.key,negate:!f,params:(query:NL,type:phrase),type:phrase,value:NL),query:(match:(extra.key:(query:NL,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))

The Netherlands have been gray most of the time for the last few days.

Values are roughly in the -100 to -250 range

FelixDQ:
That is weird actually. When looking at the [transparency platform](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=false&viewType=TABLE&areaType=CTY&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=03.09.2020+00:00|UTC|DAYTIMERANGE&dateTime.endDateTime=03.09.2020+00:00|UTC|DAYTIMERANGE&area.values=CTY|10YNL----------L!CTY|10YNL----------L&productionType.values=B20&productionType.values=B15&dateTime.timezone=UTC&dateTime.timezone_input=UTC) the values doesn't seem to be negative. Am I looking at the wrong thing, or do we have a bug in the parser?

q--:
I don't know, I'm just going by what's logged in Kibana.

I don't have an ENTSOE API key so I can't help with debugging

JHuiz:
Would this PR fix this issue? Sounds like the PR will be merged soon. #2561

[...]

PR #2676, opened 2020-09-17
Fix negative unknown production in NL

PaulCornelissen:
Proposed fix for #2638 

This ignores consumption that is reported when fetching production values.


FelixDQ:
Awesome! I just sent an email to ENTSO-E to be 100% sure that this is the correct way to interpret the data :-)

PaulCornelissen:
I also did that this morning after your message, but hadn't reported back here since I'm at work now... So maybe we get 2 answers 😀

PaulCornelissen:
I just got an answer back which confirms my thoughts:

> Dear Paul,
> 
> Thank you for your email.
> Regarding the consumption of generation units, the manual of procedures for the Actual Generation per Generation Unit says:
> "Average of all available instantaneous net power output values in each Market Time Unit.
> A generation unit could never be considered as a consumption unit"
> For some units, in addition to their generation activity they may consume energy. One explanation could be the fact it is an industrial site and consume energy for their activity. However, if they also sell electricity, they are seen as a generation unit and since a generation unit can never be considered as a consumption unit, their consumption is reported in the Actual Generation section.
> If other countries have values close to 0, it must means that their generation units are not conjugated with a consumption source of energy.
> 
> I hope the above helps.
> Best regards,

So consumption reported under actual production is consumption at the site of a producer, so it should not be subtracted from that production value. Rather it is consumption that is not reported as consumption when using fetch_consumption, so it should really be added there to get a good view of consumption in a country. In the case of NL (where unknown is calculated as the difference between consumption and known production) this should actually be added to unknown production.

PaulCornelissen:
@FelixDQ Can this be merged?

FelixDQ (approved):
Yes, sorry. Let's merge :-)

@PaulCornelissen would you mind opening an issue stating the remaining TODO's regarding adding the self-consumption to `fetch_consumption`, and the special case for NL?

PR #2677, opened 2020-09-17
Remove issue template

FelixDQ:
The current template isn't very useful.

We should try to make some better templates instead of using this one.


Kongkille (approved):
Ok, perhaps we can make some additional templates in the future? Not sure if it's worth it to make the issues now?

Issue #1972, opened 2019-09-27
Costa Rica data, official source


Mario500X:
Hello Guys, 
Just want to point out theres an official source for data from Costa Rica at 

https://apps.grupoice.com/CenceWeb/CencePosdespachoNacional.jsf

where data is published daily.  Its from our electricity dispatch center and lists all the generation plants operating in Costa Rica, daily. 

I downloaded it to excel and attached plant type (hydro, wind, etc. ) and on "emission factors" tab, put the emission factors from last year.  So country emission factor can be programmed easily.  Hope that works for you. Daily generation information can be reached at the link above. 

There's a few cases where one reservoir serves to cascading plants. I point it out in the excel where appropriate. 

[2019-09-26 Generacion_Real_MWh.xlsx](https://github.com/tmrowco/electricitymap-contrib/files/3663546/2019-09-26.Generacion_Real_MWh.xlsx)

We plant to publish this online later next year making it even easier for you to compare Costa Rica with others in your map.  electricitymap, btw,  is very handy for comparisons. 

Best regards, 

Mario Montero

martincollignon:
Nice @Mario500X! Do you work for the Costa Rican government? If yes, do you have contacts in other countries? Feel free to DM me on our slack (slack.tmrow.com)

jarek:
We actually have a Costa Rica parser https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/CR.py which uses that very page, however it seems to be out of date right now - possibly a new power plant has been added that needs handling on our side

madsnedergaard:
The parser for Costa Rica is running and should be fine, so closing this issue - feel free to reopen if anyone finds it can be improved further :)

PR #2678, opened 2020-09-17
Costa Rica: Classify new plants as geothermal and hydro

q--:
CR.py logs warnings for two unknown plants. This plant was listed by @Mario500X as geothermal in #1972. There's another plant not mapped to a generation type, _Jorge Manuel Dengo_, but it was not on the list provided by Mario Montero.

I suspect it's hydro because there's also a hydro plant named _Dengo_, but according to [Wikipedia](https://es.wikipedia.org/wiki/Jorge_Manuel_Dengo_Obreg%C3%B3n) Mr. Dengo is the founder of the Instituto Costarricense de Electricidad so it would make sense if they'd name a different kind of power plant after him too.


Mario500X:
Dengo is hydro and is the same cited above.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

q--:
Thanks, I've added that one as well

FelixDQ (approved):

PR #2679, opened 2020-09-18
Panama: Add fuel type for last plant, warn about plants with unknown fuel type

q--:
Every plant is categorized now


FelixDQ (approved):
Nice! 💪

PR #2680, opened 2020-09-18
Added IT-CNO Renewable Capacities

nessie2013:
Added Biomass, hydro, solar, geothermal, and wind for the region from the listed source: https://www.terna.it/en/electric-system/statistical-data-forecast/evolution-electricity-market (no source update required)
Regions included in IT-CNO are: "Marche, Umbria, Tuscany"


FelixDQ (approved):
👍 Nice! 

It would be nice if next time you make PR, try to minimize the diff. I.e. not move stuff around and only insert the lines needed. This makes it a bit easier to trace the changes if we for example want to find out when a line has been changed.

But this is fine for now :-) Good job

PR #2681, opened 2020-09-18
Added IT-NO Renewable Capacities

nessie2013:
(hydro is not split into normal and storage, as this source doesn't have that data)
Regions included in IT-CNO are: "Liguria, Emilia-Romagna, Veneto, Friuil Venezia Giulia, Trentino-South Tyrol, Lombardy, Piedmont, Aosta Valley"


FelixDQ (approved):
Thanks @nessie2013! :-)

PR #2684, opened 2020-09-18
Added EE->RU-1 and LV->RU-1 Interchange Capacities

nessie2013:
source from #2410


FelixDQ (commented):
Thanks @nessie2013. Can you also add the source to the [README](https://github.com/tmrowco/electricitymap-contrib#cross-border-transmission-capacity-data-sources)?

Kongkille (approved):
Thanks! Looks good

Issue #2687, opened 2020-09-21
Broken link


jmdesmon:
The link http://www.rte-france.com/en/eco2mix/eco2mix-mix-energetique-en is broken in section Electricity prices (day-ahead) data sources
France: RTE
on page https://github.com/tmrowco/electricitymap-contrib#real-time-electricity-data-sources

PR #2690, opened 2020-09-24
update market link for France

Oneill38:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2687

I'm not sure what the original link displayed (that's now broken), so please let me know if the new link is incorrect. The new link shows the current market data (maybe not the day ahead? unclear to me)


Kongkille (approved):
Yes, I can't seem to find any other relevant sources on there either. So I think it's the right one.

PR #2693, opened 2020-09-27
Fix internal link in the README

ajoga:
In the big refactor 20228916b87284ff1449bf73ec25f0364d0c5941 some links where broken


ajoga:
By the way I played a bit with github actions and `remark-validate-links` to detect internal broken links.

The config is [here](https://github.com/ajoga/electricitymap-contrib/blob/master/.github/workflows/validate-local-links-in-md.yml) and you can see what it does [there](https://github.com/ajoga/electricitymap-contrib/pull/1). Would you like me to PR that ?

If yes, I'll also have to move the badges from the heading of the README to bellow it because a [bug](https://github.com/remarkjs/remark-validate-links/issues/59) in `remark-validate-links` wrongfully marks that link as broken

FelixDQ (approved):
Cool, thank you! :-) 

Yes, it would be awesome if you also added the validator.

PR #2695, opened 2020-09-27
Add an example to test exchange parsers

ajoga:


FelixDQ (approved):
Nice, thank you!

As a sidenote: We are also working on improving our documentation, and we will probably add a wiki to the repo soon, so this is very helpful :-)

PR #2711, opened 2020-10-05
Detection of broken links between md

ajoga:
After doing the work manually of fixing links, and as suggested in https://github.com/tmrowco/electricitymap-contrib/pull/2693#pullrequestreview-501867595 I'm submitting this to avoid future toil.

The modifications of the heading and links in the README are a workaround for a bug in the tool that checks broken links where images in heading are making the check fail. Bug is opened but not yet fixed: remarkjs/remark-validate-links#59

### DIFF
--- a/.github/workflows/validate-local-links-in-md.yml
+++ b/.github/workflows/validate-local-links-in-md.yml
@@ -1,0 +1,36 @@
+name: CI
+
+# Controls when the action will run. Triggers the workflow on all PR where md files are modified
+# Disclaimer : this will fail one's PR for one's others broken internal link that went into master.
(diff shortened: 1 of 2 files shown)


ajoga:
paging @FelixDQ since you're aware of the context of this work :)

FelixDQ (approved):
LGTM! I moved the badges to their own line, so they don't push the text.

Thanks for adding this :-)

ajoga:
> LGTM! I moved the badges to their own line, so they don't push the text.

Whoops, good catch. Thanks!

ajoga:
Thank you for the fast feedback loop @FelixDQ  :)

PR #2718, opened 2020-10-08
Change gradient in colorblind-friendly themes to have higher luminance range

veqtrus:
The color blind mode has a gradient with low luminance (brightness) range, which is what would help colour-blind people differentiate values. Even as a non-colour-blind person, the current mode provides less information than the colourful version.

I changed the steps to `#F0E030` at 0g (0%), `#F02030` at 240g (30%), `#002030` at 800g (100%).

This is what it would look like using test data:

light-mode
dark-mode


Kongkille:
This is super interesting @veqtrus! Did you consider the arrows as well? As @willwillgit also pointed out. They can be almost invisible at times, this might also help a bit either way but perhaps something can be done with the arrows independently?
image

veqtrus:
I think the arrows should have some sort of shadow or border. I can't find willwillgit's comment though.

veqtrus:
What would you say about replacing GIFs with something like <https://jsfiddle.net/y63xdbz4/121/>?

veqtrus:
If we simplify the animation we can specify the colour with CSS but use GIF transparency: https://jsfiddle.net/y63xdbz4/123/

Each GIF would have only four frames so it should be quite lightweight. We would need one for each speed (just change the frame delay).

Kongkille:
Yeah, a lot of different things have been tried in terms of the arrows. Right now they are, I believe the most resource-intensive part of the website aside from the map file. 

Simplifying it down to an animation like that might work, but of course, it doesn't look as great. For now, I would suggest you create a new [discussion](https://github.com/tmrowco/electricitymap-contrib/discussions) on whether something should be done about the arrows. There are also other things that could be done.

As for this PR, I think we can accept that the arrows are still not super visible. Generally, it's still a lot better IMO.

Kongkille (approved):

fbarl:
Great change @veqtrus, makes me wonder how come none of us did it earlier! :slightly_smiling_face:  

> What would you say about replacing GIFs with something like https://jsfiddle.net/y63xdbz4/121/?
> @fbarl you might be interested in this.

You're right! I should have joined this discussion earlier because I've actually tried doing a similar optimization in https://github.com/tmrowco/electricitymap-contrib/pull/2481, but unfortunately I found out that it very much depends on the browser which rendering technique is more performant..

> Yeah, a lot of different things have been tried in terms of the arrows. Right now they are, I believe the most resource-intensive part of the website aside from the map file.

I personally feel that we should either drop the arrows animation or cut down on the number of frames, the aesthetic benefit doesn't quite justify the performance degradation at this point IMO :thinking: But that's definitely something to be discussed separately and it shouldn't be a blocker on this PR! :wink:

fbarl (approved):

veqtrus:
If only the gradient colour change should be merged for now, someone from the team can regenerate the arrows with the script to match the new gradient.

FelixDQ:
I generated the new arrows with the script
```
yarn generate-arrows
```

Is this ready to merge now?

veqtrus:
Tested ACK

PR #2726, opened 2020-10-11
Greek translation

veqtrus:
FAQs untranslated for now.

~~When I'm running the local website I don't get the option to switch to Greek (tested by modifying `en.json`). Not sure if there's something more I should do.~~ Solved!


web/locales/gr.json
```diff
@@ -0,0 +1,1788 @@
+{
+    "panel-initial-text": {
+        "thisproject": "Αυτό το έργο είναι",
+        "opensource": "ανοικτού κώδικα",
+        "see": "δείτε τις",
+        "datasources": "πηγές δεδομένων",
+        "contribute": "Συνεισφέρετε <a href=\"%s\" target=\"_blank\">προσθέτοντας την περιοχή σας</a>"
+    },
+    "mobile-main-menu": {
+        "map": "electricityMap",
```
veqtrus, line 10: Should I translate this as "Map"?

veqtrus:
The README for localisation suggests using ISO 3166-1 Alpha-2 codes but wouldn't ISO 639-1 codes be more accurate? In that case Greek would be `el`.

veqtrus:
> When I'm running the local website I don't get the option to switch to Greek (tested by modifying `en.json`). Not sure if there's something more I should do.

I see I will need to modify [locales-config.json](https://github.com/tmrowco/electricitymap-contrib/blob/master/web/locales-config.json) as well. Should probably be added to the README.

Kongkille:
Hey @veqtrus what's the status of this? I know we've talked a bit about the point of using ISO 639-1 language codes instead, but does that constrain the current PR?

veqtrus:
@Kongkille it can be merged as is and I can make a new PR when I get to translating the rest.

FelixDQ (approved):
Thank you @veqtrus! :-)

Issue #2707, opened 2020-10-04
Misleading name "hydro storage" for pumped storage


Gazer75:
The line saying "Hydro Storage" is kind of misleading as this is actually pumped hydro storage.
Maybe change the name of this one to clarify this?

ENTSO-E is displaying this as "Hydro Pumped Storage"

Kongkille:
Thanks @Gazer75, I'm closing this issue, you can see our discussion on the linked PR!

Issue #2663, opened 2020-09-16
change "gas" to natural gas so people don't think it's gasoline


ecadvocate:
change "gas" to natural gas so people don't think it's gasoline (yes, they do)

PR #2730, opened 2020-10-11
Clarify "gas" as "natural gas" in English

veqtrus:
~~"hydro storage" -> "pumped hydro storage",~~
"gas" -> "natural gas"

Suggested in #2707 and #2663.


Kongkille (commented):
I do think both changes make sense but for pumped hydro storage, I am a bit afraid that the text will be too long to be properly displayed?

tth4vb:
Agreed on pumped hydro - I think it's fine to use hydro storage. @veqtrus do you mind removing that change? then I think it's ok to merge

veqtrus:
Would perhaps "pumped hydro" make sense? Otherwise we can just change gas.

Kongkille:
I would argue that we should just change gas for now :)

Kongkille (approved):

Gazer75:
Hydro storage to me is power stations with big reservoirs, so it was kind of confusing before I looked at the code and the data at ENTSOE.

PR #2734, opened 2020-10-12
Update Belgian Capacities

Fred-3000:
Gas, Nuclear from ENTSO-E
Wind from Elia (https://www.elia.be/en/grid-data/power-generation/wind-power-generation)

First time doing this, so forgive any possible mistakes :)

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -505,18 +505,19 @@
     "capacity": {
       "biomass": 1173,
       "coal": 0,
-      "gas": 6777,
+      "gas": 6792,
       "geothermal": 0,
       "hydro": 125,
       "hydro storage": 1310,
-      "nuclear": 5919,
+      "nuclear": 5943,
       "oil": 341,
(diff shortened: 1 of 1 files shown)


FelixDQ (commented):
Thank you! Can you also update the source in DATA_SOURCES.md? 

https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md

Fred-3000:
I've managed to do it in a separate pull request. Github is a bit overwhelming as a newbie...

veqtrus:
> I've managed to do it in a separate pull request. GitHub is a bit overwhelming as a newbie...

I think it's because you didn't create a new branch for this change and tried to edit it directly so GitHub created an 'invisible' branch. In the other PR there seems to be a proper branch.

Kongkille:
Moved to https://github.com/tmrowco/electricitymap-contrib/pull/2738

PR #2735, opened 2020-10-12
Added autosquash

FelixDQ:
This should make it a bit faster to merge PRs. 

I added a personal token for now, but I will replace it with @tmrow-bot when I have the token.


Kongkille (approved):
niiice

PR #2738, opened 2020-10-12
Update Belgian capacity and source

Fred-3000:
I wasn't immediately able to find how to add this commit to the existing pull request 😅.


Kongkille (approved):
I've moved the changes from https://github.com/tmrowco/electricitymap-contrib/pull/2734 here to keep it aligned

Fred-3000:
Thank you!

Issue #2731, opened 2020-10-11
Kosovo as a state


betim:
Please update your maps to reflect Kosovo as a state.

jarek:
Is Kosovo's electrical grid operated separately from Serbia? Do you know what their operator is named?

betim:
Yes it is. Operator is called KOSTT. Couple of months ago it withdrew from SMM bloc. KEK which runs the thermo centrals uses coal to generate 800MW (A) + 678MW (B).

jarek:
Thanks for the confirmation!

Looks like this should be pretty easy - edit generate-geometries.js to remove `KOS` from `RS` and then add it separately (looks like `XK` is used as the two-letter code for Kosovo?):

https://github.com/tmrowco/electricitymap-contrib/blob/57884bd70cc44448861259000f81a47b4ebadd60/web/generate-geometries.js#L557

veqtrus:
They have transparency data: https://www.kostt.com/Transparency/BasicMarketDataOnGeneration
Could be possible to write a parser.

jarek:
however currently the "Current production (actual generation) of electricity" spreadsheet contains data for October 11th (all 24 hours) while it's just after midnight on October 13th in Pristina, so not quite realtime

veqtrus:
We have capacities at least: https://www.kostt.com/Content/ViewFiles/Transparency/BasicMarketDataOnGeneration/EN/Installed%20capacity%20of%20production%20units.pdf

jarek:
One more question, if anyone knows!

Currently we have exchanges configured for "Serbia to North Macedonia" and "Albania to Serbia". Geographically, at least the Albanian border is exclusively with Kosovo, and Kosovo also borders North Macedonia notably close to Skopje which I'm guessing might be an infrastructure centre.

![Serbia - North Macedonia exchange on Electricity Map](https://user-images.githubusercontent.com/47415/96006646-b5a74700-0e0b-11eb-9cd5-53a1bd41c7c5.png)

Currently the data for the exchanges comes from ENTSOE, where Serbia is configured to be `10YCS-SERBIATSOV`, Albania `10YAL-KESH-----5`, and North Macedonia `10YMK-MEPSO----8`, and exchanges appear to be plain ENTSOE exchanges between those.

Does anyone know what's the present electrical reality, particularly for the North Macedonia exchange? Should either or both of these be reclassified as Kosovo? How does this all look in ENTSOE?

[...]

PR #2739, opened 2020-10-12
update map to show Serbia and Kosovo separately

jarek:
ref: https://github.com/tmrowco/electricitymap-contrib/issues/2731

Not tested at all, but might work?


FelixDQ (approved):
LGTM!

I just generated the geometries with

```
docker-compose run --rm web ./topogen.sh
```

PR #2744, opened 2020-10-15
Create electricty.py

ayushigoyal2840:


FelixDQ:
I'm not sure that I understand this. Please reopen with a description if you think that this is something that should be added

PR #2754, opened 2020-10-19
Refactor CY parser

veqtrus:
- Refactor parser as a class
- Add type annotations to functions
- Compile regexes to constants


parsers/CY.py
```diff
@@ -12,64 +12,102 @@
...
-    return data
+EXPECTED_COLUMNS = ['Συνολική Προβλεπόμενη Ζήτηση', 'Αιολική Παραγωγή',
+    'Εκτίμηση Διεσπαρμένης Παραγωγής (Φωτοβολταϊκά και Βιομάζα)',
+    'Συνολική Ζήτηση', 'Συμβατική Παραγωγή']
+
+COLUMNS_REGEX = re.compile(r'data\.addColumn\("number", "([^"]*)"\);')
+TIMES_REGEX = re.compile(
+    r'var dateStr = "([^"]*)";\s+var hourStr = "(\d+)";\s+var minutesStr = "(\d+)";')
+PRODUCTIONS_REGEX = re.compile(
+    r'\[dateStrFormat,\s*(\d+|null),\s*(\d+|null),\s*(\d+|null),\s*(\d+|null)\]')
```
systemcatch, line 23: Good idea to pull these constants out

parsers/CY.py
```diff
@@ -12,64 +12,102 @@
...
+
+class CyprusParser:
+    session = None
+    logger: logging.Logger = None
+
+    def __init__(self, session, logger: logging.Logger):
+        self.session = session
+        self.logger = logger
```
systemcatch, line 31: There is no real benefit to making this parser a class (the class is holding functions but the object doesn't use them).

parsers/CY.py
```diff
@@ -12,64 +12,102 @@
...
+            'zoneKey': 'CY',
+            'production': prods_value,
+            'storage': {},
+            'source': 'tsoc.org.cy',
+            'datetime': time_value
+        })
+
+
+    def parse_html(self, html: str) -> list:
+        global EXPECTED_COLUMNS, COLUMNS_REGEX, TIMES_REGEX, PRODUCTIONS_REGEX
```
systemcatch, line 69: In general it's best to avoid globals, these could either be class attributes or straight constants if the parser stays as a set of functions.

parsers/CY.py
```diff
@@ -12,64 +12,102 @@
...
+
+
+    def fetch_production(self, target_datetime: datetime.datetime) -> list:
+        if target_datetime is None:
+            url = 'https://tsoc.org.cy/total-daily-system-generation-on-the-transmission-system/'
+        else:
+            # convert target datetime to local datetime
+            url_date = arrow.get(target_datetime).to('Asia/Nicosia')
+            url = 'https://tsoc.org.cy/archive-total-daily-system-generation-on-the-transmission-system/?startdt={}&enddt=%2B1days'.format(
+                url_date.format('DD-MM-YYYY'))
```
systemcatch, line 95: Minor nitpick, you could format `url_date` on L93 to save a longer line below.

parsers/CY.py
```diff
@@ -12,64 +12,102 @@
...
+
+        res = self.session.get(url)
+        assert res.status_code == 200, 'CY parser: GET {} returned {}'.format(url, res.status_code)
+
+        data = self.parse_html(res.text)
+
+        if len(data) == 0:
+            self.logger.warning('No data returned')
```
systemcatch, line 103:
```suggestion
        if len(data) == 0:
            self.logger.warning(f'No production data returned for Cyprus on {url_date}', extra={'key': 'CY'})
```

systemcatch (changes requested):
Hey @veqtrus I've left some comments.

parsers/CY.py
veqtrus: The idea is that if we were to add logging in more places or make more requests in the future, we wouldn't need to pass these objects as function parameters.

parsers/CY.py
veqtrus: I removed `url_date` from the message as it's not set when fetching live data.

parsers/CY.py
systemcatch: That is true but we trade off by having to use `self` everywhere and making the structure of the parser more complicated. However it's not a huge deal, let's see what @FelixDQ thinks.

FelixDQ: I don't have a strong opinion for or against classes. I think that the implemented solution using a class is neat, so I'm fine with this solution.

I also don't want to be very strict about how a parser should be implemented because I want to be easy for new contributors to create a parser. We obviously have to draw the line somewhere, but I think classes are fine.

systemcatch: Alright let's roll with it then.

systemcatch (approved):

PR #2755, opened 2020-10-19
BO: Use static carbon intensity for "unknown"

### LINKED ISSUE #2747: Bolivia: calculate static carbon intensity for "unknown"
According to [2019 IEA data](https://www.iea.org/data-and-statistics?country=BOLIVIA&fuel=Energy%20supply&indicator=ElecGenByFuel), thermal electricity generation in Bolivia is mostly natural gas (and a bit of oil and biomass), so the default carbon intensity ElectricityMap uses for "unknown" generation is too high for the country.

Someone should calculate a static carbon intensity for the unknown part. (Note that `unknown` is only thermal generation.)

q--:
Closes #2747

By the way, should I update `contributors` for changes like this, or is that only for updates to the parser itself?


config/co2eq_parameters.json
```diff
@@ -147,6 +147,11 @@
         "hydro discharge": {
           "source": "2017 average by Tomorrow",
           "value": 558.083871451602
+        },
+        "unknown": {
+          "_url": "https://www.iea.org/data-and-statistics?country=BOLIVIA&fuel=Energy%20supply&indicator=ElecGenByFuel",
+          "source": "IEA 2019, assumes 1.4% oil, 94.5% gas, 4.0% biomass",
+          "value": 481.2
```
veqtrus, line 154: Percentages don't add up to 100%. You can do 1.4% oil, 94.5% gas, 4.1% biomass for 481.58g avg.

veqtrus: https://www.iea.org/data-and-statistics/data-tables?country=BOLIVIA&energy=Electricity&year=2019 has a table with the productions, which is what I used. But whichever page is fine.

q--: >Percentages don't add up to 100%. 

The usual rounding error. I've done the "round to the nearest integer" (whatever it's called when you round to a decimal number instead of an integer), which I think is pretty standard for statistics and certainly less unexpected than rounding one of the numbers in a different way than the others.

>https://www.iea.org/data-and-statistics/data-tables?country=BOLIVIA&energy=Electricity&year=2019 has a table with the productions, which is what I used. But whichever page is fine.

That's neat, I'll use that next time I need these figures from the IEA. I'll keep the current link though, as you've already noticed the same data is present in both.

FelixDQ (approved):
Cool! Yes, go ahead and update the contributor list :-)

PR #2756, opened 2020-10-19
Add Kosovo parser

veqtrus:
From #2731
Only coal data available in [Current production (actual generation) of electricity](https://www.kostt.com/Content/ViewFiles/Transparency/BasicMarketDataOnGeneration/Prodhimi%20aktual%20gjenerimi%20faktik%20i%20energjise%20elektrike.xlsx) spreadsheet. It has data for the previous day.


veqtrus:
Based on the apparent discontinuity in production and the order in the spreadsheet it seems that the last data-point belongs to the next day.

TC Kosova B2 production with 00:00 last:
TC Kosova B2 production

veqtrus:
It might be that the times are the ends of the periods, so 00:00 would mean 23:00-00:00. In that case we would need to shift all times by one hour back.

FelixDQ (commented):
Nice job @veqtrus! Do you want to reach out to KOSTT to understand how the timestamps should be interpreted?

veqtrus:
@FelixDQ I tried to contact them but they don't accept emails from private addresses: https://kostt.com/News/New/11
I could in principle use another domain, however I'm not sure if my email would be welcome.

veqtrus:
I'm fairly certain however that 00:00 is indeed the last data point as I found another spreadsheet that has the date more explicitly.

However now that I think about it, I'm not sure if shifting the time by an hour would make sense. There are probably more sources that return the end of time periods (is period the right terminology?). It's certainly that way for Cyprus.

FelixDQ:
I'm not sure that I understand. Can you explain a bit more why it wouldn't make sense to shift the values? The backend expects the start datetime of a "period".

veqtrus:
There are parsers that only return the latest data. Those data typically apply to a period before 'now', but the parser itself might not know how long that period is so it can't calculate the start datetime. Some sources might not calculate an average at all, so the returned value is only valid for that particular time.

If you would like to have start datetimes, then I suspect quite a few parsers would need fixing.

FelixDQ:
It's okay if the parser doesn't calculate an average. The backend will automatically calculate that if there is more than one value in an hour.

The only assumption that we do is that if there are a start and end time, then we pick the start time. In the future, we will probably add a config option to specify exactly how long each event is valid.

But it please open an issue if you find a bug in another parser :-)

veqtrus:
Ok, I made it so the parser returns the starts of periods.

FelixDQ (commented):
Awesome! It should be almost ready to go then. I'm currently getting an error when running the parser:
```
> python test_parser.py XK
  File "test_parser.py", line 92, in <module>
    print(test_parser())
  File "/Users/felix/.pyenv/versions/3.7.4/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/felix/.pyenv/versions/3.7.4/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/felix/.pyenv/versions/3.7.4/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/felix/.pyenv/versions/3.7.4/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "test_parser.py", line 52, in test_parser
    res = parser(*args, target_datetime=target_datetime)
  File "/Users/felix/Projects/tmrow/tmrowapp/src/electricitymap/contrib/parsers/XK.py", line 90, in fetch_production
    time_str = time.strftime('%H:%M')
AttributeError: 'float' object has no attribute 'strftime'
```

Do you think something changed in the data source or am I doing something wrong?

veqtrus:
It works for me... Tablib chooses a Python datatype based on the Excel datatype of the cell. Perhaps the different version of the library does it somewhat differently. Will have to look into it and possibly check the type of the returned value. Although float seems weird for a time cell...

FelixDQ (approved):
LGTM! Nice job @veqtrus 🎉 

I quickly added the data delay to zones.json (see https://github.com/tmrowco/electricitymap-contrib/issues/2783 for more context).

PR #2760, opened 2020-10-22
Update NL capacities

IgorYbema:
According to https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show

Solar production in https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf seems much too high (maybe end of 2020?)


Kongkille:
Thank you, did any data sources change? 

https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md

IgorYbema:
No, only solar was overrated in the IRENA source which is probably suspecting a growth by the end of 2020 but doesn't match what ENTSOE is telling with a update just a few month ago.

Kongkille (approved):
Alright, it looks good to me :)

PaulCornelissen:
I don't know why ENTSO-E has the numbers so horribly wrong, but it's much more than that.

The national bureau of statistics reported a 2019 eoy capacity of 6874MW:
https://www.cbs.nl/nl-nl/nieuws/2020/25/vermogen-zonnepanelen-voor-het-eerst-groter-bij-bedrijven-dan-bij-woningen

ATM it's already over 8GW. But in reality nobody knows exactly, because PV systems are not exhaustively registered anywhere. There's different partial registries, but they have overlap and even combined do not nearly cover all of the market. How do we know it's over 8GW? Because CBS does pretty extensive estimations based on many different data sources. Then we have some public registries with monthly numbers. Add those to the last official numbers and maybe do some conservative extrapolation. Follow https://twitter.com/Polder_PV

IgorYbema:
That solar power will then never be recorded als solar power anyways so why bother setting the max higher?
I think it is better to match the sources (procuded and max) to be the same. This will show at least the percentage correctly.

Kongkille:
Yes I figured you had some insights here @PaulCornelissen. The capacity is mainly there for the user to understand whether the region is actually utilizing what it could. So if we set it higher, it will also seem as if NL is not utilizing solar power as well as it could have (while the reality may be that we are just not able to display it)

PaulCornelissen:
@IgorYbema The solar power generation for NL is sourced from energieopwek.nl which does use a accurate estimation of the PV capacity in NL. IIRC they have a model of PV installations happening per day per region. So they estimate the PV capacity installed per region per day and combine that with reference plants and weather data to calculate the actual production.

@Kongkille if that's your argument, then why do we bother manually looking up sources instead of just automatically using the highest measured production in some rolling time window? 😉

Kongkille:
@PaulCornelissen hmm yeah, no I definitely see your point. I haven't seen NL reach the capacity of gas either. And a lot of other countries have capacities a lot higher than their max consumption. For me, I think it's also about having a source wherever we can, and usually, we don't see these big gaps between good sources.

I'm just looping in @FelixDQ, maybe we've had similar discussions in the past.

FelixDQ:
Ok, I see your points. I don't think that we should display a capacity that we know isn't accurate. I suggest either:
1. Add the best estimate that we can find.
2. Remove the solar capacity.

@PaulCornelissen do you have a link to the best recent estimate?

PaulCornelissen:
@FelixDQ I think it's probably best to stick with the numbers from CBS (National Statistics Bureau). They published 2019 EOY numbers. Same link I posted above:
https://www.cbs.nl/nl-nl/nieuws/2020/25/vermogen-zonnepanelen-voor-het-eerst-groter-bij-bedrijven-dan-bij-woningen

6874MW at the end of 2019

FelixDQ:
👍 Ok cool. I'll create a new PR

PR #2761, opened 2020-10-24
Added Kosovo in Italian

Raffox97:


jarek:
This is not strictly needed because in absence of Italian translation, the interface falls back to the English which in this case is the same. See https://www.electricitymap.org/zone/XK?lang=it

jarek (approved):

Issue #2758, opened 2020-10-21
IN-MH parser data is coming in spotty


tth4vb:
Looking at the last few months of data coming into IN-MH, we're seeing very scarce data, a couple of hourly data points every 24 hours, meaning it's not very useful when looking at the recent history graphs on eMap for example:
<img width="444" alt="Screen Shot 2020-10-21 at 10 50 57 AM" src="https://user-images.githubusercontent.com/8129157/96696703-59a06d80-138b-11eb-9924-b73cef71cf02.png">

Here's a chart of the data coming in for 2020:
<img width="677" alt="Screen Shot 2020-10-21 at 10 57 38 AM" src="https://user-images.githubusercontent.com/8129157/96697547-55c11b00-138c-11eb-9e39-ca475f14d4eb.png">

It would be awesome if someone has the chance to dig in a bit here as to why this is happening, particularly @GitMatze who I believe built the original parser. 

If we cannot identify the issue and/or a potential solution, I think this also would bring up a conversation previously started in #1982 by @systemcatch about switching to all of India as a parser instead given some of the challenges with regional parsers. My concern there is that there is still a decent proportion of data coming into that parser as "unknown production" where we're assigning some static values (based on skimming through #304, after which we had concluded on sticking to regional parsers). But let's first see if there's some solution to this problem in IN-MH.

GitMatze:
This is a known issue caused by a timeout. Since the data is parsed from a dashbord image using character recognition this parser is quite computation-heavy. I've already tried to optimize it, but apparently it's still not fast enough. See https://github.com/tmrowco/electricitymap-contrib/issues/2343

tth4vb:
Ah thanks for the clarification @GitMatze! Good to know the limitations here

Kongkille:
@GitMatze just looked at the parser. It seems you've thought of a lot of ways to improve it. Did you try to crop the image to specifically only get the values that are relevant? I'm wondering which aspects of the computation takes a long time, it could also be the original loading of the image as it's a bit heavier than a lot of other images we use tesseract on.

GitMatze:
Just did some profiling. On my system fetching the image accounts for 62% and running tesseract for 32% of the time. However, with the cpu power resctricted  to a minimum, fetching the image takes only 34% and tesseract 70% of the time. Tesseract definitely is the cpu-sensitive process here.

ssmssam:
[...]

PR #2768, opened 2020-10-26
speedup IN-MH parser

ssmssam:
Using pytessy sped up the pytesseract portion of the code by ~10% on my machine. This is linked to issue #2758
I haven't incorporated it into the other parsers that use pytesseract yet as it was causing an error in JP_KN (OSError: [WinError 127] The specified procedure could not be found). Narrowed this down to the importing of beautifulsoup4

@GitMatze

On my machine:

Tesseract:
%timeit pytesseract.image_to_string(imgs_line, lang='digits_comma', config='--psm 7')
521 ms ± 7.66 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

pytessy:
%timeit digit_reader.read(imgs_line.tobytes(), imgs_line.size[0], imgs_line.size[1], 1)
484 ms ± 20.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

and for a file taken at an earler time:
Tesseract:
%timeit pytesseract.image_to_string(imgs_line, lang='digits_comma', config='--psm 7')
1.02 s ± 3.97 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

pytessy:
%timeit digit_reader.read(imgs_line.tobytes(), imgs_line.size[0], imgs_line.size[1], 1)
939 ms ± 5.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


GitMatze:
@ssmssam  Seems like you can't pass the 'psm 7' argument with pytessy? It advises tesseract to read the image as a single text line. I will do some logging to see if there are any differences in prediction reliability.

ssmssam:
Note, I have also made a change to the parser SG.py to help with issue #2779 . I can seperate this into a different pull request if that is easier @GitMatze  .

parsers/IN_MH.py
```diff
@@ -219,9 +219,9 @@ def fetch_production(zone_key='IN-MH', session=None, target_datetime = None,
     # string together all image sections and recognize resulting line
     imgs_line = np.hstack( list(np.asarray(i) for i in imgs[:]) )  
     imgs_line = Image.fromarray( imgs_line)
-    text = pytesseract.image_to_string(imgs_line, lang='digits_comma', config='--psm 7')
+    digit_reader = pytessy.PyTessy(language='digits_comma')
```
GitMatze, line 222: On my linux system I get a `NotImplementedError('PyTessy: Library search on Linux is not implemented yet.')`
To avoid this you have to provide pytessy with the exact paths. For me `digit_reader = pytessy.PyTessy(lib_path='/usr/bin/tesseract', data_path='/usr/share/tesseract-ocr/4.00/tessdata', language='digits_comma')` works. 
@Kongkille  how would this work for the backend?
For more information on the pytessy class: https://github.com/hyperrixel/pytessy/blob/master/source/pytessy.py

Kongkille: Sorry for the late reply @GitMatze! The backend is on Linux, so we'll need to specify the explicit library paths, I'm not entirely sure what we would do, but if you could get it working on your machine, I'm sure we can tweak it to the backend?

ssmssam: Is there anything I can do to help this?

Kongkille: @FelixDQ would you know which changes need to be made for it to work on our backend?

FelixDQ: It's running using the [python:3.6 docker image](https://hub.docker.com/_/python). We have a list of setup commands that we run when building the container. If you can provide the command(s), that we should add, then it would be perfect.

Otherwise, if you have the steps needed to install the library / find the path on your own Linux system, then we can probably get it working on the server.

GitMatze: You only need to adjust the paths passed to the pytessy.Pytessy() object in order for it to access the tesseract library. Depending on where tesseract is installed on the backend the paths should look pretty similar..

corradio: Hi @GitMatze ,
In our linux container, the binary `/usr/bin/tesseract` exists but `find_library` in https://github.com/hyperrixel/pytessy/blob/master/source/pytessy.py#L154 doesn't return anything.
Reading https://docs.python.org/3/library/ctypes.html, and running from our python instance in our linux container, I get:

```bash
>>> find_library('tesseract')
'libtesseract.so.4'
```
but `find_library('/usr/bin/tesseract')` returns None
Note that libtesseract.so.4 is located in /usr/lib/x86_64-linux-gnu/ (using the path didn't help either.
Passing just `libtesseract` as `lib_path` doesn't work, due to this check: https://github.com/hyperrixel/pytessy/blob/master/source/pytessy.py#L265-L269

Is this supposed to work on Linux? Seems like `find_library` returns the full path on MacOS, but not on Linux.

corradio: @GitMatze are you able to help here?

skovhus: I've updated the pyproject setup, so now we also get the error on CI.

> NotImplementedError: PyTessy: Library search on Linux is not implemented yet.

skovhus (commented):
Thank you so much for trying this out @ssmssam!

I've looked into the Linux support here and the issues.

pytessy seems rather immature currently – especially if you are not on Windows. There is also no activity on the project, and several bugs.

I fixed the library path issue by hardcoding the path in the source code of pytessy. After doing so I got a Segmentation fault when running the parser:  
```
!strcmp(locale, "C"):Error:Assert failed:in file baseapi.cpp, line 209 
Segmentation fault
```

Based on the learnings here, I do not think pytessy is a viable option for us right now...

But I'm curious if there is anything else we can do to speed up the OCR. I tried to upgrade pytesseract to the latest version, but it doesn't seem like it gives any performance increase.

Any other ideas @ssmssam?

ssmssam:
That's ok @skovhus  . Sorry, I didn't think to try it out on Linux before. Will try that next time round. I haven't got any off the top of my head but will have a look around and see if I can find something!

PR #2771, opened 2020-10-27
Fix incorrect data source url in Moldova parser

lgrigoryeva1:
There was an issue with the Moldova parser MD.py, where it failed to fetch data from the specified URL. 
Turns out the data source url was missing `.php` at the end. 
The correct url is: http://www.moldelectrica.md/utils/load4.php


lgrigoryeva1:
Thinking if I should have added some error message if the parser can't access the data in case this happens again in the future?

FelixDQ (commented):
Thanks @lgrigoryeva1! Yes adding a better error message is probably a good idea if you want to do that. Otherwise, the only thing left is adding yourself to the contributor list

lgrigoryeva1:
@FelixDQ Done. Let me know if there's something else :)

FelixDQ (approved):
Let's merge 🎉

Issue #2764, opened 2020-10-25
Data for Isle of Wight (England)


q--:
https://www.ssen.co.uk/anm/iow/

The page is almost identical to the one used by the `GB_ORK.py` parser

Kongkille:
In the previous PR there were some concerns of the quality of the data. I am also having some trouble understanding the graphs properly. It seems the connection between actual generation and capacity is almost non-existing. 


We also need some kind of static breakdown for production to color it on the map! :)

VIKTORVAV99:
It should be possible for us to create a production parser mapping it all to unknown like we have done for several other zones, but there wouldn't be a major difference in the data without either a interchange with another zone or an estimation method (due to the size and if it lacks an interconnection this is unlikely to happen)

PR #2772, opened 2020-10-27
Add GB-IOW parser (Isle of Wight)

danstan5:
As suggested in #2764 the data available is basically identical to the Orkney Islands - both on Active Network Management (ANM) with live data accessed from ssen.co.uk.
I've refactored GB_ORK to simplify/ generalise so the added GB_IOW is more or less a copy of this parser. Outputs remain the same.


q--:
Would it be possible to handle the Isle of Wright and the Orkney Islands in the same parser (just like e.g. most of Europe uses ENTSOE.py)?

Could perhaps be named GB_SSEN.py

danstan5:
Yeah, makes sense - will look into updating that.

danstan5:
I've made these suggested changes, so final commit now adds the info for the GB-IOW to the SSEN parser. 
I've also updated the zones/exchanges.json with bounding boxes etc, and just let me know if there is anything else need to do for this !  :)

danstan5:
Also comment on Isle of Wight data.. like Orkney there is no breakdown available generation mix from SSEN. 
Web searching it appears they still have gas power plants and renewable capacity but haven't found any figures on this yet. SSEN claims "Total Renewable Capacity 280 MW", so I imagine the carbon intensity would be alot lower than default for 'unknown' source.

FelixDQ (commented):
Nice job @danstan5

From my side, we are missing the following:
- [ ] Separate the geometry from GB. (See how it was done for GB-ORK [here](https://github.com/tmrowco/electricitymap-contrib/pull/1615) and read [this readme](https://github.com/tmrowco/electricitymap-contrib/blob/master/web/README.md))
- [ ] Can we make any assumptions about the power breakdown?

FelixDQ:
Hi @danstan5. What is the status on this? Let us know if you need help with this

danstan5:
Hi Felix. Unfortunately the Live AMN generation for IOW is not reporting correctly. I reached out to SSEN about it a while back and was informed while works were underway on a connected site the generation from wouldn't be correctly relayed. It looks like this still hasn't be fixed (and been in this state for at least 2 months now)..
There doesn't seem to be any more info on when it might be back up, so I'd put this on pause.

Manu1400:
From status page : https://www.ssen.co.uk/ANMGeneration/

Zone | ANM Operation | SHEPD Equipment | Generator Site Issues
-- | -- | -- | --
Core Zone |  warning | OK  | KO 

and Core Zone mean Isle of Wight, source : https://www.ssen.co.uk/ANM/

FelixDQ:
Thanks for the update! I converted this PR to draft while we wait. Let us know when there are any updates on this :-)

gwpicard:
@danstan5 are you still working on this? Keep us in the loop if you need help :)

FelixDQ:
Hi @danstan5,

I think that we'll close this issue for now. Please do reopen it if the situation changes.

Thanks you for the work on this :-)

PR #2773, opened 2020-10-27
Fix broken get_prices() in Turkey parser

lgrigoryeva1:
As of now the Turkey parser crashes when running `get_prices`. Looks like the data structure has changed, and now the prices are also available in EUR and USD. 
I fixed the issue and kept the prices in original currency, e.g. TRY. 
If needed, I can add a functionality to retrieve the EUR and USD prices.


lgrigoryeva1:
I forgot to branch out when making this change, so don't be confused why there's some file modifications not related to `TR.py`.

FelixDQ (commented):
Cool! Thank you @lgrigoryeva1 

It looks like the file MD.py is still present in the PR (it's empty). Can you remove the file completely? Otherwise, it should be good to go :-)

(Also remember to add yourself to the contributor list in zones.json)

lgrigoryeva1:
@FelixDQ Done :+1:

FelixDQ (approved):
LGTM!

PR #2784, opened 2020-10-30
CAISO: ignore nuclear self-consumption

q--:
Test-parser.py gave an error, so I also added the missing logging.

As an aside, I get

>Validation failed: datetime was not returned for US-CA  

when running test-parser. I don't think this is caused by any of my changes.

The lines preceding that one are 

>min returned datetime: 2020-10-30T07:00:00+00:00 UTC  
>max returned datetime: 2020-10-30T15:55:00+00:00 UTC  -- OK, <2h from now :) (now=2020-10-30T15:56:21.527212+00:00 UTC)   

so not all the datetime logic is broken.

### DIFF
--- a/parsers/US_CA.py
+++ b/parsers/US_CA.py
@@ -5,13 +5,14 @@ import pandas
 import requests
 from bs4 import BeautifulSoup
 from collections import defaultdict
+import logging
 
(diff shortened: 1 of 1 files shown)


FelixDQ (approved):
Thank you @q-- !

Yes, I introduced a bug in test_parser.py, that has now been fixed (https://github.com/tmrowco/electricitymap-contrib/pull/2782)

PR #2787, opened 2020-11-01
Update wind capacity for Norway

Gazer75:


tth4vb:
@Gazer75 thanks for this :) could you just reference which source you're using for these capacity updates?

tth4vb:
@Gazer75 thanks for this :) could you just reference which source you're using for these capacity updates?

Gazer75:
Same as before. NVE should be listed

tth4vb (approved):

tth4vb:
Got it, thanks for clarifying :)

PR #2788, opened 2020-11-04
Prettyprint result of test_parser.py

FelixDQ:
Before:
image

After:
image


Kongkille:
a really good idea!

Kongkille (approved):
nice :-))

idk if it's just me, but i feel like the datetime is hard to read, would a possible extension of this be to output the datetime in a more readable format?

PR #2791, opened 2020-11-06
Fix broken IN-HP parser

willbeaufoy:
The parser was [broken](https://kibana.electricitymap.org/app/kibana#/doc/96f67170-0c49-11e9-85c1-1d63df8c862c/logstash-2020.11.05/doc?id=Q7HQmnUBBKMWQLREuz1I&_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))) because the data source page had added a new power plant, and the parser is designed to fail if it encounters any data it does not recognise (in order to ensure accuracy).


systemcatch:
@willbeaufoy it would be nice if we logged this failure but then continued with the parser, else this breakage will happen everytime a new plant is added.

willbeaufoy:
> @willbeaufoy it would be nice if we logged this failure but then continued with the parser, else this breakage will happen everytime a new plant is added.

I could do this but then the data would not be completely accurate - but I'm happy to do it if that is the electricityMap policy (i.e. show best effort data). I actually asked in [slack](https://tmrow-team.slack.com/archives/C2DQPTHA6/p1604621499245900) whether there is a way that I could be notified when the parser hits errors?

Kongkille:
> > @willbeaufoy it would be nice if we logged this failure but then continued with the parser, else this breakage will happen everytime a new plant is added.
> 
> I could do this but then the data would not be completely accurate - but I'm happy to do it if that is the electricityMap policy (i.e. show best effort data). I actually asked in [slack](https://tmrow-team.slack.com/archives/C2DQPTHA6/p1604621499245900) whether there is a way that I could be notified when the parser hits errors?

Sorry for the late reply here @willbeaufoy, I think logging and continuing is fine. If we do not catch it, then I think it's more of a responsibility on our side not checking the logs.

Regarding your message in Slack, I haven't seen any way to notify on error, but the logs are queryable, so perhaps there is a way to setup some kind of notification system elsewhere, or perhaps we could consider adding it to the frontend somehow, where a warning icon appears if there's been several errors on the logs?

willbeaufoy:
> > > @willbeaufoy it would be nice if we logged this failure but then continued with the parser, else this breakage will happen everytime a new plant is added.
> > 
> > 
> > I could do this but then the data would not be completely accurate - but I'm happy to do it if that is the electricityMap policy (i.e. show best effort data). I actually asked in [slack](https://tmrow-team.slack.com/archives/C2DQPTHA6/p1604621499245900) whether there is a way that I could be notified when the parser hits errors?
> 
> Sorry for the late reply here @willbeaufoy, I think logging and continuing is fine. If we do not catch it, then I think it's more of a responsibility on our side not checking the logs.
> 
> Regarding your message in Slack, I haven't seen any way to notify on error, but the logs are queryable, so perhaps there is a way to setup some kind of notification system elsewhere, or perhaps we could consider adding it to the frontend somehow, where a warning icon appears if there's been several errors on the logs?

Great thanks for clarifying, I've now made the modifications to log and continue, and I'll investigate further re: getting notifications from kibana.

willbeaufoy:
@Kongkille @systemcatch is there anything more I need to do to get this merged? It's done from my perspective.

systemcatch (approved):

Issue #2705, opened 2020-10-02
Improve front-end to make it easier for visitors to contribute


martincollignon:
We had a hack day with the Tomorrow team today, and started working with @Kongkille on a project to improve front-end to make it easier for visitors to contribute. I'm putting this as an issue to allow anyone to come with feedback and help out.

The idea would be that each area is in a state of development, and that we should make it very easy for visitors (regardless if they're visitors or not) to know exactly how they can help.

## A zone has no real-time data and no contact details.

- [ ] Label when hovering over the zone should change from "No data available" to "Help by adding contact details. Click to learn more"
- [ ] Left tab would change to something that looks like [this](https://docs.google.com/drawings/d/19UvvBT23QD9uV8AjPWeNTQkOrXlAF9mNKrctS_q2Djg/edit?usp=sharing).
- [ ] A link to create a new issue with this [format](https://github.com/tmrowco/electricitymap-contrib/pull/2702)
- [ ] Picture to use: https://undraw.co/search , name: Searching, hex color: #000000

## A zone has no real-time data but contact details

- [ ] A new array for each zone should be created in zones.json with contact details (@Kongkille please link to example in your branch)
- [ ] Label when hovering over the zone should change from "No data available" to "Help by contacting relevant organisations. Click to learn more"
- [ ] Left tab would change to something that looks like [this](https://docs.google.com/drawings/d/1IKFsRgxXDFcdyS1m7ZHdiDvE0FsgNrWSFiAqYIQPLf8/edit?usp=sharing).
- [ ] Picture to use: https://undraw.co/search , name: Contact Us, hex color: #000000

## A zone has a real-time parser that is available 
- [ ] Label when hovering over the zone should change from "No data available" to "Help by building a parser. Click to learn more"
- [ ] Left tab would change to something that looks like [this](https://docs.google.com/drawings/d/1IYEgAXXVAIc7GUZUCAvzL8syh_0eF58JWBAuaBF2Fvc/edit?usp=sharing).
- [ ] Github issues that should be fetched would have the "parser buildable!" label, as well as a label that follows this format; _zone:FR_ 
- [ ] Picture to use: https://undraw.co/search , name: Collecting, hex color: #000000

## A zone has a real-time parser but there is a bug

- [ ] Label when hovering over the zone should change from "No live data available" to "Help by fixing the parser. Click to learn more"
[...]

PR #2792, opened 2020-11-06
Actionable data overview [DRAFT]

Kongkille:
Based on #2705


Kongkille:
I'm closing this for now. I'm not sure this is exactly how we'd want to implement the issue, so I think taking a step back would be good.

PR #2793, opened 2020-11-09
Update Estonian Production Capacities From Renovable Sources

Vaikevimbu:
Updated Estonian production capacities according to The Estonian Renewable Energy Association (EREA) 2019 report. http://www.taastuvenergeetika.ee/wp-content/uploads/2020/10/ETEK_aastaraamat_A4_2019_veeb.pdf It is in Estonian, foreigners can use Google Translate. Wind energy (tuuleenergia) stats are on page 20, sun energy (päikeseenergia) on page 22, hydro on page 26 and other renewable sources (mostly sun energy from mircro-producers) on page 27. ENTSO-E data is similar, but EREA should have the most precise data.


tth4vb:
@Vaikevimbu thanks for this! Should we be updating hydro as well?

Vaikevimbu:
There is no change in hydro electricity capacity in Estonia for years now, because it is not favoured. Decimal number is 7,3MW, I would round it to 7MW, as it was already.

tth4vb:
ah, that's my bad for trying to read Estonian and choosing the wrong number ;)

tth4vb (approved):

Vaikevimbu:
Do I have to do something more or when will the changes take effect? At the moment the electricity Map is still showing old capacities for Estonia and I am wondering if something more needs to be done. Thnx.

PR #2795, opened 2020-11-10
Update fr.json: Shetland & french oversea flag

JeanBaptisteScellier:
GB-SHI is Shetland not Orkney. + typo in "Grèce". Added countryName for all Overseas departments and regions of France which official flag is the french one: Guadeloupe, Martinique, Guyane française, La Réunion, Mayotte, Wallis-et-Futuna, Saint-Martin, Saint-Pierre-et-Miquelon.
GB-SHI est Shetland pas "Orkney". Ajout du countryName pour toutes les Régions-départements d'outre-mer qui ont pour drapeau le drapeau tricolore : Guadeloupe, Martinique, Guyane française, La Réunion, Mayotte, Wallis-et-Futuna, Saint-Martin, Saint-Pierre-et-Miquelon. + faute d'orthographe Grèce.

### DIFF
--- a/web/locales/fr.json
+++ b/web/locales/fr.json
@@ -584,7 +584,7 @@
         },
         "GB-SHI": {
             "countryName": "Grande Bretagne",
-            "zoneName": "Orkney"
+            "zoneName": "Shetland"
         },
         "GB-ZET": {
             "countryName": "Grande Bretagne",
@@ -597,6 +597,7 @@
             "zoneName": "Géorgie"
         },
         "GF": {
+            "countryName": "France",
             "zoneName": "Guyane Française"
         },
         "GG": {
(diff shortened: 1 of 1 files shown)


tth4vb (approved):

jarek:
For what it's worth, if you wanted to also change the flags shown in Electricity Map, I don't think this change will be sufficient, since the flags are loaded based on the two-character ISO key. You'd have to change the files in https://github.com/tmrowco/electricitymap-contrib/tree/master/web/public/images/flag-icons/flags_iso

jarek:
(Actually the above was incorrect: there is a `flag_file_name` field in https://github.com/tmrowco/electricitymap-contrib/blob/master/config/zones.json which you can use to override the flag if you wanted)

PR #2805, opened 2020-11-17
Fix missing ENTSOE hydro storage (fixes #2800)

### LINKED ISSUE #2800: ENTSO-E Pumped Hydro consumption missing
It seems that consumption values for pumped hydro (storage) are missing for countries based on ENTSO-E data. 
They are reported on their website, so I guess something is wrong in the parser.

### LINKED ISSUE #2802: GB pumped hydro consumption missing
It appears pumped hydro consumption in GB is missing. Last night there were super low power prices and GB was exporting to FR despite that resulting in nuclear being curtailed there, and yet there was zero pumped hydro consumption in GB.

willbeaufoy:
97865b8 introduced this bug by removing all consumption data from the ENTSOE parser, however it was not intended to remove hydro storage, only power plant self-consumption.

This pull request reverts the earlier change, and adds a check so that only hydro storage consumption data will be included, rather than no consumption data at all.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -620,18 +620,22 @@ def parse_production(xml_text):
         is_production = len(timeseries.find_all('inBiddingZone_Domain.mRID'.lower())) > 0
         psr_type = timeseries.find_all('mktpsrtype')[0].find_all('psrtype')[0].contents[0]
 
-        if len(timeseries.find_all('inBiddingZone_Domain.mRID'.lower())) > 0:
-            for entry in timeseries.find_all('point'):
-                quantity = float(entry.find_all('quantity')[0].contents[0])
-                position = int(entry.find_all('position')[0].contents[0])
-                datetime = datetime_from_position(datetime_start, position, resolution)
-                try:
-                    i = datetimes.index(datetime)
+        for entry in timeseries.find_all('point'):
+            quantity = float(entry.find_all('quantity')[0].contents[0])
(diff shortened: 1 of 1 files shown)


Kongkille (approved):
Nice, makes sense!

parsers/ENTSOE.py
```diff
@@ -620,18 +620,22 @@ def parse_production(xml_text):
...
+            datetime = datetime_from_position(datetime_start, position, resolution)
+            try:
+                i = datetimes.index(datetime)
+                if is_production:
                     productions[i][psr_type] += quantity
-                except ValueError:  # Not in list
-                    datetimes.append(datetime)
-                    productions.append(defaultdict(lambda: 0))
-                    productions[-1][psr_type] = quantity if is_production else -1 * quantity
+                elif psr_type in ENTSOE_PARAMETER_GROUPS['storage']['hydro storage']:
```
q--, line 631: Shouldn't we ignore all types of storage? Otherwise, if another storage type is added to ENTSO-E (e.g. battery storage) we'll run into the same problem again

Kongkille: Hmm, I guess it makes sense that it would also apply for other types of storage from reading through some of the other issues. What do you think @willbeaufoy?

willbeaufoy: Good idea, I've made a new pull request to do that here:

https://github.com/tmrowco/electricitymap-contrib/pull/2812

PR #2808, opened 2020-11-19
Update solar capacity for Cyprus

veqtrus:
https://tsoc.org.cy/total-daily-system-generation-on-the-transmission-system/ (Φωτοβολταϊκή Εγκατεστημένη Ισχύς)


Kongkille (approved):

Kongkille:
Hey quick question @veqtrus, when you decide to take a look at the capacities, are there any indicators that a source might need an update on capacities? I'm wondering whether it would be worth it to add a "last_updated" or something like that to the source, so it's easy to see whether it might be worth it to take another look?

veqtrus:
TBH I only looked at the Cypriot one. Most of solar in Cyprus is rooftop which is continuously expanded, so it's expected that once in a while the stats will be updated.

A "last_updated" field would mostly be useful if it would be displayed on the website as a reminder. It would be nice however to be able to express capacity updates as a parser rather than hardcoded values, though those would be updated less frequently than the live stats. Speaking of update frequency, it would be nice to not run parsers every few minutes for sources that only update once an hour. The Kosovo one is updated only once a day even.

Kongkille:
> TBH I only looked at the Cypriot one. Most of solar in Cyprus is rooftop which is continuously expanded, so it's expected that once in a while the stats will be updated.
> 
> A "last_updated" field would mostly be useful if it would be displayed on the website as a reminder. It would be nice however to be able to express capacity updates as a parser rather than hardcoded values, though those would be updated less frequently than the live stats. Speaking of update frequency, it would be nice to not run parsers every few minutes for sources that only update once an hour. The Kosovo one is updated only once a day even.

Okay thanks. I agree that it can seem a bit much to run it every 15 minutes for parsers we know only update once a day, but sometimes the parser will fail, and in those cases the next request can fetch the unresolved data. I realize it's unlikely for Kosovo that it will fail 100 times, but that is our current reasoning. 

But i like the idea, we should research it further to see if there are some parser that don't need to be activated on each run!

PR #2814, opened 2020-11-21
Make ENTSOE fetch_consumption() include self-consumption (fix #2710)

### LINKED ISSUE #2710: ENTSOE fetch_consumption must include consumption by production units
While fixing a issue with negative values for "unknown" production (#2638) it was found out (#2676) that ENTSOE reports consumption at the site of a generator under production as a negative production. For most power plants the electricity consumption is neglible (a couple of MW at most), but in NL we have a lot of CHP plants at greenhouses and such. They are very large electricity consumer and at times this can be several 100's of MW. This consumption is _probably_ not included in the consumption values we fetch from ENTSOE. So the ENTSOE parser for fetch_consumption needs to also fetch production information from ENTSOE and add the consumption of production plants.

willbeaufoy:
This adds power plant self-consumption (which is found in the production data)
to the total consumption value for a given timezone.

It has a limitation in that consumption data is often available a few minutes before
production data, so self-consumption will not be included in total consumption for the
most recent timezone. E.g. running the parser:

python test_parser.py NL consumption

at 17:32 UTC returns consumption data for the most recent datetime of 16:45 (total
value 14780), but does not include self-consumption as at this point production data is only
available until 16:30. However if you run it again at 17:40, production data for 16:45 is
now available, so the total consumption for the most recent timezone of 16:45 is now 15611.

I wasn't sure the best way to handle, this, is this the right approach to take? If needed I could easily change it to remove any timezones from the results that did not have self-consumption data available yet.


willbeaufoy:
@Kongkille is there anything more I need to do to get this merged?

Kongkille:
Sorry for the late turnaround, @FelixDQ would you mind taking a look at this?

FelixDQ (commented):
Nice! Sorry for the late reply.

I would be hesitant to remove values where we don't have self-consumption because the ENTSOE parser is used so much, but it would probably be good to add a warning. (`logger.warning`)

Otherwise, this looks good to me :-)

willbeaufoy:
> Nice! Sorry for the late reply.
> 
> I would be hesitant to remove values where we don't have self-consumption because the ENTSOE parser is used so much, but it would probably be good to add a warning. (`logger.warning`)
> 
> Otherwise, this looks good to me :-)

Ok good idea, done.

FelixDQ (approved):
Awesome! LGTM :-)

q--:
I'm seeing errors in [Kibana](https://kibana.electricitymap.org/app/kibana#/doc/96f67170-0c49-11e9-85c1-1d63df8c862c/logstash-2021.02.04/doc?id=ihaKbXcBipTrw-FaMDU0&_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))) for the NL parser related to [this part of the code](https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTSOE.py#L831), is this expected or should I open an issue?

```

  File "/home/src/electricitymap/feeder/lib/fetch_data.py", line 180, in launch_parsers
    objs = res.get(timeout) or []
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
  File "/home/src/electricitymap/feeder/lib/fetch_data.py", line 178, in <lambda>
    **parser_kwargs))
  File "/home/src/electricitymap/contrib/parsers/NL.py", line 25, in fetch_production
    logger=logger)
  File "/home/src/electricitymap/contrib/parsers/ENTSOE.py", line 831, in fetch_consumption
    i = datetimes.index(k)
ValueError: <Arrow [2021-02-04T09:30:00+00:00]> is not in list
```

FelixDQ:
That seems like an error. @willbeaufoy have you seen this before?

willbeaufoy:
> That seems like an error. @willbeaufoy have you seen this before?

Yeah looks like an error. Can look this evening.

willbeaufoy:
Sorry no time to look this eve, but will def be able to check on the weekend.

PR #2819, opened 2020-11-25
Extend co2eq_parameters.json with information from co2eq_parameters.js

FelixDQ:
This is the first step of deprecating `co2eq_parameters.js`.
The information is now explicit in co2eq_parameters.json. This allows us to import information from this file in other languages than js.

I also added the test from our internal repo.


corradio (approved):
great! 💪

PR #2823, opened 2020-12-01
Ratelimit XK parser

veqtrus:
Data source is updated only once a day, so I limited fetching updates to once an hour.

### DIFF
--- a/parsers/XK.py
+++ b/parsers/XK.py
@@ -13,6 +13,7 @@ import requests
 # please try to write PEP8 compliant code (use a linter). One of PEP8's
 # requirement is to limit your line length to 79 characters.
 
+last_updated = None
 
 def fetch_production(zone_key='XK', session=None,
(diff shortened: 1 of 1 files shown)


Kongkille:
Hey @veqtrus, we talked about this in a thread at some point. I do believe this is worth looking into, however, I don't think that we can do it on a parser level. Instead, we would need to look at changes to the broader infrastructure. Perhaps we could define it in zones.json in the same way we introduced delayed data. Also a side note, I am actually not sure the global would actually work when it runs on the server

Perhaps you can open an issue where we discuss potential implementation?

veqtrus:
@Kongkille I also added some extra logging with the latest timestamp. Would that change be alright?

Kongkille:
Ah yes @veqtrus let's keep that. Should I reopen or will you create a new one?

PR #2825, opened 2020-12-02
Use new design for the desktop header

fbarl:
This PR introduces a new header bar in desktop view, keeping small / mobile screens unchanged.

Screenshot_2020-12-10 Live CO₂ emissions of electricity consumption

The new `SharedHeader` component is implemented in a portable way so that we can reuse it in https://api.electricitymap.org/.


veqtrus:
Doesn't that just reduce the size of the map? I'm not sure what this design improves.

fbarl:
Hey @veqtrus, the idea here is to unify the headers used in https://api.electricitymap.org and https://www.electricitymap.org and this PR is a step in that direction.

You're right that this reduces some vertical space for the map, but (for now) the new header will only appear on desktop screens and I think it's a worthwhile compromise for the electricityMap branding & smoother navigation between its parts.

Does this make sense? :slightly_smiling_face:

@Kongkille Feel free to provide some more context if I missed something.

veqtrus:
On the API page the header is also transparent initially (before scrolling). In terms of branding, Google Maps also has a transparent header. If you would like to make the headers more similar, I could suggest making the header links turn bold instead of blue on the map page, just like on the API one.

fbarl:
@veqtrus There is one more effect we want to achieve here and that is to have the header be hierarchically above the map / API website / blog so that the effect would be as if we were selecting different tabs of the same product (regardless of the fact that they all currently live on different domains).

But I think you make a good point about the API page header initially being transparent which takes away from the above mentioned tabs experience and makes me think that maybe we should remove the transparency "feature". WDYT @Kongkille?

Kongkille:
@veqtrus yes so this is one part of a change we have been working on. It's not finished yet! The idea as @fbarl mentions is to unify headers across the electricityMap domain so we can make it clearer what we are working on, which data we actually have and which actions you can take. I'll share more on this later as well

I'll take another look at the transparency. You may be right, I think it still works regardless, but it might, as you say, take a bit away from it. 

On another note, I was thinking we could also add an URL parameter header=[false|true], so powerusers (and people using it for dashboards etc can hide it. Would you mind adding that @fbarl?

fbarl:
I had a call with @Kongkille and we agreed to make a couple more design tweaks to the header before merging. I'll address these comments briefly:

- Switch to using underline instead of text shadow for active header links
- Add more horizontal padding to the header content from both sides - this will break the horizontal alignment with some elements but will make the experience of switching to other the API page smoother (and will look better there)
- Shrink horizontal paddings on small screens - this will not make a difference here as we're hiding it on small screens but it will look better when reused on the API page

fbarl:
@Kongkille PTAL - I addressed all 3 points above, the only differences from the design being that I lifted the underline closer to the text and that I kept the text shadow for selected links because I thought it looked better across the pages.

Some screenshots:

Screenshot_2020-12-10 Live CO₂ emissions of electricity consumption
Screenshot from 2020-12-10 13-38-22
Screenshot from 2020-12-10 13-38-35
Screenshot from 2020-12-10 13-38-51
Screenshot from 2020-12-10 13-38-59

Kongkille:
Thanks @fbarl, I think that it's a bit better UX if the underline is at the bottom of the header because it then signifies more that you're on the website and that changing would mean you would end up on a "different" site

cc @tth4vb

fbarl:
> I think that it's a bit better UX if the underline is at the bottom of the header because it then signifies more that you're on the website and that changing would mean you would end up on a "different" site

That's an interesting point I haven't thought of @Kongkille!

I dropped the underline to the bottom in https://github.com/tmrowco/electricitymap-contrib/pull/2825/commits/f8e15da5c9d10278e3c44e5957d9d9238bc83e8d and updated the screenshots I posted in https://github.com/tmrowco/electricitymap-contrib/pull/2825#issuecomment-742497097. Please take one more look :slightly_smiling_face:

Kongkille (approved):

PR #2831, opened 2020-12-07
Switch GE<->AM exchange to different source

q--:
AM has been down for some time now, so switch to GE which also has exchange information and is still available

(There's no power being [exchanged](http://gse.com.ge/home) right now, but the other exchanges in the parser output appear to work correctly.)

By the way, it seems the exchange with Russia isn't being shown on the map because the GE parser uses `RU` instead of `RU-1`

Edit: in  #1877 it was noted that the Georgian data has some minor deficiencies, so if the AM parser can be repaired we should probably switch it back again.


Kongkille:
So the idea, if I understand correctly would be to switch exchanges, for now, so we still potentially get the exchanges from AM, and then if we fix the AM parser we can switch it back?

q--:
@Kongkille That is correct.

However, after making this PR I noticed (https://github.com/tmrowco/electricitymap-contrib/issues/2462#issuecomment-749598769) that the entire website used by the Armenia parser is gone, so it's unlikely we'll be able to change it back to the Armenian side anytime soon.

FelixDQ (approved):
LGTM. Let's use this for now

PR #2833, opened 2020-12-08
Fix typo in ES pumped hydro name, update capacity

q--:
Pumped hydro capacity had a hyphen in the key so it wasn't showing up on ElectricityMap. Also updated the capacity to that from ENTSO-E.

By the way, shouldn't there be some kind of unit test to catch this?


Kongkille (approved):
Great. Yes I'll create an issue for testing zones.json

PR #2845, opened 2020-12-11
Add link to Open Source page

fbarl:
Adds a header link to the new [Open Source](https://api.electricitymap.org/open-source/) page.

Screenshot from 2020-12-11 11-53-48

### DIFF
--- a/web/src/layout/header.js
+++ b/web/src/layout/header.js
@@ -11,6 +11,10 @@ const headerLinks = [
     active: true,
   },
   {
+    label: 'Open Source',
+    href: 'https://api.electricitymap.org/open-source?utm_source=electricitymap.org&utm_medium=referral',
+  },
+  {
     label: 'Blog',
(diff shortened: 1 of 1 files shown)


Kongkille (approved):

veqtrus:
On the Open Source page the nav link that is underlined is API.

Also IMO the header could be made smaller.

Kongkille:
Yes, we are working on a fix on the Open source underline.

I do think we can tweak the header height yes. 

It's an interesting idea, I'll try to make a design for it then we can see how it looks. I was also thinking this might be a good idea to move further with dark theme being a bit more dominant (eg changing the entire overlay)

veqtrus:
Were there really 1700+ contributors?

fbarl:
> On the Open Source page the nav link that is underlined is API.

FYI This has been fixed now :wink:

Issue #2848, opened 2020-12-14
New additions to the Faroe Islands grid


Bjarkigd:
There have been added a bit of new renewables in the Faroes, and the website is finally showing the production. 

https://www.sev.fo/framleidsla/el-orka-i-foroyum/ 

Vindorka - Wind has had +6.3 MW installed since last update. 18+MW will be added next year if all permits etc go the right way.
Biogasorka - Biogas is a new thing with with 1.5 MW installed capacaty
Sólorka - Solar, is new as well, not much with 0.24MW installed capacity at one solar farm
There are also two solar roof installations(that I know of, there are probably more out there) with public data, two 12kW installations that can be seen here http://os.fo/ out to the right "sólórka á.....". Don't know if solar is even going to make a blip on the graph even in summer, but there's absolutely growing interest for it.

In addition to this there's a new trial tidal power thing, which is supplying energy to the grid, but it is small 100kW and doesn't show live generation, but it's planned for more next year.

If anyone wants to update it then feel free.

q--:
Thanks for the heads-up.

I had a look at the FO parser and it's currently using https://w3.sev.fo/hagtol/xml/xkiefjSDKFjeijgjdkjf3847tgfjlkfdgnlsnfvm.xml as its data source. I can't find solar production in it (Ctrl+F Sol gives me nothing).

The production page you linked is also fetching its data from a different source than that, namely https://www.sev.fo/api/realtimemap/now. So now I'm wondering if it's still being updated with new sources. (Output on the SEV site is pretty close to that on ElectricityMap and could easily be explained by a bit of lag, so I'm unsure. However, the new URL *does* have results for Ctrl+F Sol, so it's at least missing solar.)

Is the solar actually showing up on Sev.fo or should it be fetched from http://os.fo? It's night there now so I can't check

q--:
And to return to the subject of installed capacity, I was wondering if you've found a new source for the capacity data that can be listed in [DATA_SOURCES](https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md)? 

The current source is a thesis paper that's obviously not getting updated (in fact, it's gone from the website) so if you've found a new source for this data that would be great 👍

Bjarkigd:
[...]

PR #2849, opened 2020-12-15
Switch FO parser to new data source

q--:
FO now also has solar, biomass and tidal generation (#2848). The old URL used by the parser doesn't have these, so I've switched  the parser to the new URL and updated it to handle these new generation types.


config/co2eq_parameters.js
```diff
@@ -49,7 +49,7 @@ exports.fossilFuelAccessor = (zoneKey, k, v) => {
   return (k == 'coal' ||
           k == 'gas' ||
           k == 'oil' ||
-          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey !== 'SG' && zoneKey !== 'PR')) ||
+          (k === 'unknown' && (zoneKey !== 'GB-ORK' && zoneKey !== 'UA' && zoneKey !== 'SG' && zoneKey !== 'PR' && zoneKey !== 'FO')) ||
```
FelixDQ, line 52: We are actually in the progress of deprecating this file. So for now we also have to add this to the bottom of co2eq_parameters.json. (https://github.com/tmrowco/electricitymap-contrib/blob/9550a1ff35f2fad6e363c1afb096970f96fec687/config/co2eq_parameters.json#L2949-L3012)

FelixDQ (commented):
Awesome. Only one comment. Otherwise this looks good to me :-)

config/co2eq_parameters.js
q--: Done

q--: Also removed a now-unused import

jarek (approved):
:tada:

Issue #2570, opened 2020-07-03
Data for Nigeria


q--:
Real-time consumption: https://www.sogrid.org/DisCoLoadProfile

Generation by hour: https://www.sogrid.org/GenerationProfile (includes historical)  
Generation seems quite low compared to consumption for the current time, a brief look through the historical data suggests this might be due to some generation companies being slower with providing data

The dashboard also has some other stuff like grid frequency, [generation/consumption by day for the past seven days et cetera](https://www.sogrid.org/EnergyTrend)

Found via https://nsong.org/

The regulatory commision also has some [data](http://www.nercng.org/index.php/library/industry-statistics), but it looks older/less detailed/broken.

Nigeria has interconnectors to neighbouring countries. According to the graph at http://icc.ecowapp.org/ (scroll down) it exports  a lot. Imports for 2019 were zero.

https://www.nbet.com.ng/ is the electricity trading company. Installed capacity breakdown on the homepage, unclear if it's for all of Nigeria

systemcatch:
Nice find, looks like it uses aspx behind the scenes so we should be able to get the data out but it might end up being painful.

Kongkille:
Hi guys, I just looked at the data briefly but would you say we can set the labels "Parser buildable" and "Help wanted" for this issue? Or do we need further investigation?

q--:
@Kongkille It should be possible to build a parser with the data that's available. Carbon intensity can be calculated from [IEA data](https://www.iea.org/countries/nigeria), the country almost exclusively uses natural gas and hydro

sashavor:
So essentially each of the sources listed at [https://www.sogrid.org/GenerationProfile](https://www.sogrid.org/GenerationProfile) needs to be linked with a type of production (natural gas/hydro) and then the parser can be built?

q--:
@sashavor In theory, yes, but the GenerationProfile seems to be incomplete (compare the lastest generation total and the demand at the DisCoLoadProfile page - older generation profile data appears to be more complete however), so for real-time it might be better to grab the consumption from the DisCoLoadProfile page and calculate a static carbon intensity from IEA data.

corradio:
@q-- if in case you deem the GenerationProfile data of insufficient quality, then I don't think we should qualify this parser as buildable.
I'm not sure that just having the load profile with a static intensity is of great value here?
However it might be good to show the incomplete mix at this stage, as this might motivate Nigerian officials to do something about their data once they see it on the map!

[...]

PR #2851, opened 2020-12-15
feat: add Nigeria live parser

byronwilliams:
@q-- I've attempted a parser for Nigeria for ticket - https://github.com/tmrowco/electricitymap-contrib/issues/2570

I have taken the data from the Generation Load Profile page as I feel it is better to be able to report the actual values of the sources as these are available, even if they are incomplete.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -2864,8 +2864,7 @@
       "productionPerModeForecast": "ENTSOE.fetch_wind_solar_forecasts",
       "price": "ENTSOE.fetch_price",
       "production": "ENTSOE.fetch_production"
-    },
-    "timezone": "Europe/Rome"
+    }
   },
   "IT-SAR": {
     "bounding_box": [
@@ -3544,6 +3543,25 @@
       [
         25.759780721000112,
         -16.451057230999922
+      ]
+    ]
+  },
+  "NG": {
+    "contributors": [
+      "https://github.com/byronwilliams"
+    ],
+    "parsers": {
+      "production": "NG.fetch_production"
+    },
+    "timezone": "Africa/Lagos",
+    "bounding_box": [
+      [
+        2.1710819900001184,
+        3.7721621770000695
+      ],
+      [
+        15.16993615700008,
+        14.380290832000114
       ]
     ]
   },
--- a/parsers/NG.py
+++ b/parsers/NG.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+
+"""Parser for the electricity grid of Nigeria"""
+
+import arrow
+import logging
+import requests
+from bs4 import BeautifulSoup
+
+
+LIVE_PRODUCTION_API_URL = "https://www.niggrid.org/GenerationLoadProfileBinary?readingDate={0}&readingTime={1}"
+TYPE_MAPPING = {"hydro": "hydro", "gas": "gas", "gas/steam": "gas", "steam": "gas"}
+
+
+def extract_name_tech(company):
+    parts = company.split("(")
+    tech = parts[1].strip(")").lower()
+
+    return parts[0], TYPE_MAPPING[tech]
+
+
+def template_response(zone_key, datetime, source):
+    return {
+        "zoneKey": zone_key,
+        "datetime": datetime,
+        "production": {
+            "gas": 0.0,
+            "hydro": 0.0,
+        },
+        "storage": {},
+        "source": source,
+    }
+
+
+def fetch_production(
+    zone_key=None,
+    session=None,
+    target_datetime=None,
+    logger=logging.getLogger(__name__),
+):
+    """Requests the last known production mix (in MW) of a given zone
+    Arguments:
+    zone_key (optional) -- used in case a parser is able to fetch multiple zones
+    session (optional) -- request session passed in order to re-use an existing session
+    target_datetime (optional) -- used if parser can fetch data for a specific day, a string in the form YYYYMMDD
+    logger (optional) -- handles logging when parser is run
+    Return:
+    A list of dictionaries in the form:
+    {
+      'zoneKey': 'FR',
+      'datetime': '2017-01-01T00:00:00Z',
+      'production': {
+          'biomass': 0.0,
+          'coal': 0.0,
+          'gas': 0.0,
+          'hydro': 0.0,
+          'nuclear': null,
+          'oil': 0.0,
+          'solar': 0.0,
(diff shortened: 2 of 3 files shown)


parsers/NG.py
```diff
@@ -0,0 +1,114 @@
...
+      'source': 'mysource.com'
+    }
+    """
+
+    if target_datetime is not None:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+    now = arrow.now(tz="Africa/Lagos")
+    dt_day = now.format("DD/MM/YYYY")
+    dt_hm = now.format("HH:00")
```
q--, line 85: Should probably round it down to the hour...

parsers/NG.py
```diff
@@ -0,0 +1,114 @@
...
+    resp = r.get(fullUrl)
+
+    try:
+        soup = BeautifulSoup(resp.text, "html.parser")
+        table = soup.find("table", {"id": "MainContent_gvGencoLoadProfiles"})
+        rows = table.find_all("tr")[1:-1]  # ignore header and footer rows
+    except AttributeError:
+        raise LookupError('No data currently available for Nigeria.')
+
+    result = template_response(zone_key, now, "niggrid.org")
```
q--, line 98: ...otherwise, the `datetime` on the result is going to be different than that of the website. (E.g. fetching at 12:34 retrieves data for 12:00, but gets logged as data for 12:34 to the ElectricityMap database.)

q-- (changes requested):

parsers/NG.py
byronwilliams: I've updated this such that the `datetime` returned is the current timestamp `now = arrow.now(tz="Africa/Lagos")` - this seems to be the pattern in the others parsers

q--: What I meant is that in this function call  `now` should be (a variable with the the result) of  `now.replace(minute=0)` so the timestamp returned by the parser is the same one as the timestamp given on the website

q--: I'd replace line 83-98 with
```


    timestamp = arrow.now(tz="Africa/Lagos").replace(minute=0)
    dt_day = timestamp.format("DD/MM/YYYY")
    dt_hm = timestamp .format("HH:mm")
    fullUrl = LIVE_PRODUCTION_API_URL.format(dt_day, dt_hm)

    resp = r.get(fullUrl)

    try:
        soup = BeautifulSoup(resp.text, "html.parser")
        table = soup.find("table", {"id": "MainContent_gvGencoLoadProfiles"})
        rows = table.find_all("tr")[1:-1]  # ignore header and footer rows
    except AttributeError:
        raise LookupError("No data currently available for Nigeria.")

    result = template_response(zone_key, timestamp, "niggrid.org")
```

Line 83, 84, 85 and 98 are changed. (Variable name changed only for the sake of clarity)

q-- (changes requested):

parsers/NG.py
byronwilliams: Ohhh.... I understand now. I've updated the PR to reflect your comment

q-- (approved):
LGTM. Still a bit unsure about having initially incomplete data (and in that light, extending the parser with fetching historical data would be helpful), but no dealbreakers that would have to be fixed before going live IMO.

As an aside, here's a bit of config for `config/zones.json` that was required to make `python3 test_parser.py NG` work (the test reported no errors🎉):

```
  "NG": {
    "contributors": [
      "https://github.com/byronwilliams"
    ],
    "parsers": {
      "production": "NG.fetch_production"
    },
    "timezone": "Africa/Lagos"
  },
```
This also needs `"bounding_box":` metadata for things to show up on the map, but since I still haven't figured out how to generate those I'll leave that to someone else.

parsers/NG.py
```diff
@@ -0,0 +1,112 @@
...
+        "zoneKey": zone_key,
+        "datetime": datetime,
+        "production": {
+            "biomass": 0.0,
+            "coal": 0.0,
+            "gas": 0.0,
+            "hydro": 0.0,
+            "nuclear": 0.0,
+            "oil": 0.0,
+            "solar": 0.0,
```
q--, line 33: Given that the IEA reports [solar generation for Nigeria](https://www.iea.org/data-and-statistics?country=NIGERIA&fuel=Energy%20supply&indicator=ElecGenByFuel "chart is broken, but the Download chart data button works") but it's missing from the generation output, it might be better to omit this or set it to None. That way it'll show up as a question mark on the ElectricityMap site

byronwilliams: * I've removed all of the extra production types which were not used
* Zones updated
* DATA_SOURCES updated

FelixDQ:
Awesome! Is this ready to merge @q-- @byronwilliams ?

byronwilliams:
Fine by me! If someone can review I'll rebase it and merge it

parsers/NG.py
```diff
@@ -0,0 +1,102 @@
...
+      'storage': {
+          'hydro': -10.0,
+      },
+      'source': 'mysource.com'
+    }
+    """
+    if target_datetime is not None:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    timestamp = arrow.now(tz="Africa/Lagos").replace(minute=0)
```
FelixDQ, line 73: You can update it to fetch historical data using 
```suggestion
    if target_datetime is not None:
        timestamp = arrow.get(target_datetime).to("Africa/Lagos").replace(minute=0)
    else:
        timestamp = arrow.now(tz="Africa/Lagos").replace(minute=0)
```

FelixDQ (changes requested):
Hi @byronwilliams. I'm very sorry for the delay on my side. I have 2 comments, and then we should be ready to merge

config/zones.json
```diff
@@ -2864,8 +2864,7 @@
       "productionPerModeForecast": "ENTSOE.fetch_wind_solar_forecasts",
       "price": "ENTSOE.fetch_price",
       "production": "ENTSOE.fetch_production"
-    },
-    "timezone": "Europe/Rome"
```
byronwilliams, line 2868: This was a duplicate - I thought I'd tidy it as I went - https://github.com/tmrowco/electricitymap-contrib/pull/2851/files#diff-5fdb410dda63034a171c7bacd44c070de0c6a2d0ce5c0d5436b3626877b6a583R2837

FelixDQ (approved):
Nice job @byronwilliams! Let's merge 🎉

PR #2854, opened 2020-12-17
Fix King Island battery key

### LINKED ISSUE #2801: King Island: battery never seems to discharge
I've been keeping an eye on AUS-TAS-KI since it was added to the map. Charging works fine, discharging doesn't show up.

q--:
Also changed the URL returned to the page containing the dashboard in an iframe, instead of the current one which is just the dashboard without context


jarek (approved):
Looks good to me, added @corradio for double verification :+1:

corradio (approved):

Issue #2700, opened 2020-09-30
Wrong carbon intensity for cross-border tooltip


GitMatze:
Probably due to power being zero..

![TooltipError](https://user-images.githubusercontent.com/45139485/94737383-09c31d80-036e-11eb-9f8f-dc038f4b77d5.png)

corradio:
This is due to a discrepancy between the two conditions
https://github.com/tmrowco/electricitymap-contrib/blob/52db4e03fac4ee5d85b59b08754899b4418bf37e/web/src/helpers/zonedata.js#L41
and https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/components/tooltips/countrypanelexchangetooltip.js#L35

GitMatze:
@corradio Seems to be still wrong

corradio:
@GitMatze can you post a screenshot?

GitMatze:
![image](https://user-images.githubusercontent.com/45139485/104908050-31261080-5986-11eb-977a-e04f1528aff7.png)

corradio:
What would you see as the expected behaviour?

GitMatze:
I think the last line in the tooltip is wrong. It says the carbon intensity of Southwest Norway is 383 g/kwh, but  that's the carbon intensity of Germany.

corradio:
Indeed that's a bug! Thanks for following up

PR #2856, opened 2020-12-22
Update zonedata.js

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2700


fbarl (approved):
Good catch @corradio!

Gotta love those one character PRs :wink:

Issue #2560, opened 2020-06-25
min/max of left panel should take into account history


corradio:
Impact: high (this happens as soon as one scroll through historical values)

Currently, the horizontal scale is based on the min/max production (or capacity) of a given datapoint.
A better way to do it would be to take the min/max over **the whole history of that zone**, which would avoid rescaling the horizontal bars when sliding through the past.
cc @fbarl 

![image](https://user-images.githubusercontent.com/1655848/85779383-d431e700-b723-11ea-99e2-8baf0bb4a682.png)

PR #2859, opened 2020-12-22
Compute min and max based on history

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2560


web/src/components/countrytable.js
```diff
@@ -455,6 +443,29 @@ const CountryTable = ({
   const [productionTooltip, setProductionTooltip] = useState(null);
   const [exchangeTooltip, setExchangeTooltip] = useState(null);
 
+  // Use the whole history to determine the min/max,
+  // fallback on current data
+  const history = useCurrentZoneHistory() || [data];
+  const [minPower, maxPower] = useMemo(
```
fbarl, line 449: Is there a reason why this part is not being calculated in here instead of deeper in `CountryElectricityProductionTable` (which is the only place where it's being used)?

If we moved it there, we wouldn't need to pass any extra props and it wouldn't be calculated also when `CountryCarbonEmissionsTable` is rendered instead.

corradio: you're absolutely right! Thanks for flagging this!

web/src/components/countrytable.js
```diff
@@ -327,24 +328,32 @@ const CountryElectricityProductionTable = React.memo(({
 
   const { productionY, exchangeFlagX, exchangeY } = getDataBlockPositions(productionData, exchangeData);
 
+  // Use the whole history to determine the min/max,
+  // fallback on current data
+  const history = useCurrentZoneHistory() || [data];
```
fbarl, line 333: Ah I missed this earlier but if `useCurrentZoneHistory` is undefined here, this will always recreate a new array and therefore the `useMemo` below won't really be memoizing properly but will be recalculated whenever any of the props change.

For a further optimization, I'd recommend changing this line to `const history = useCurrentZoneHistory()` and then adding `data` dependency to `useMemo` below and handling it as a fallback separately within its function's body.

But I imagine this is not such a big deal anyway as most of the time `useCurrentZoneHistory()` won't be undefined so up to you whether you want to address it here :wink:

fbarl (approved):
Thanks for this change, a small code diff with a big UX improvement! :tada: 

Didn't check the behaviour here either but the code looks good too (keeping in mind the new comment I added) :slightly_smiling_face:

web/src/components/countrytable.js
corradio: Thanks for catching this. Consistency is key so I'll edit it!

PR #2863, opened 2020-12-23
Consistent co2eq parameters

corradio:
Fixes https://github.com/tmrowco/tmrow/issues/3210

Probably requires mirroring this behaviour in the new pipeline


config/co2eq_parameters.test.js
```diff
@@ -5,6 +5,8 @@ describe('Each fallbackZoneMixes sums to one', () => {
     ['defaults', defaultCO2Params.fallbackZoneMixes.defaults],
     ...Object.entries(defaultCO2Params.fallbackZoneMixes.zoneOverrides),
   ])('for zone %s', (zone, fallbackZoneMix) => {
+    // Verify the key exists
+    expect(fallbackZoneMix.powerOriginRatios).toBeTruthy();
```
corradio, line 9: Moved the python test to JS by adding this line.
I did this as the python test was part of the "parser" tests, which didn't make sense

FelixDQ (approved):
Makes sense! I'll be sure to update the new pipeline.

We should probably also update the [update-co2eq_parameters](https://github.com/tmrowco/tmrow/blob/master/src/electricitymap/update-co2eq-parameters.py#L61) script so that we don't accidentally add them again.

Issue #2674, opened 2020-09-17
update_state.py is broken


PaulCornelissen:
The update state script is broken because for the exchanges the `fetch_exchange()` method of the parsers return an element called `sortedZoneKeys` where the front-end expect that element in the ``state`` file to be called `sortedCountryCodes`. Don't know if the front-end is wrong here or if the parsers are wrong or they are both right but different and the update script should rename the element?

PaulCornelissen:
The state file also doesn't contain exchange capacities. So you can't test locally whether exchange capacities are displayed correctly.

corradio:
Hi @PaulCornelissen ,

I agree with the exchange capacities. However, I couldn't reproduce the issue you faced with `sortedCountryCodes` / `sortedZoneKeys`, although I agree the naming should be unified.
Could you let us know how to reproduce the issue?

PaulCornelissen:
I'm not near my computer currently, but from memory:
If you look at the state file (https://raw.githubusercontent.com/tmrowco/electricitymap-contrib/master/mockserver/public/v3/state) you see some exchanges with a key called `sortedCountyCodes` in there and some with the key called `sortedZoneKeys`. Now if you run the mock server and look at the map, then the exchanges `sortedCountryCodes` show up whilst the ones with `sortedZoneKeys` as key name do not. Hence I infer that the front-end expects the key to be called `sortedCountryCodes`. 

The `update_state.py` scripts currently puts `sortedZoneKeys` if you update an exchange, so that's wrong.

corradio:
Hi @PaulCornelissen ,
Thank you for bringing this to our attention. It comes from the fact that we allow both `sortedZoneKeys` and `sortedCountryCodes` in our parsers (cf https://github.com/tmrowco/tmrow/blob/master/src/electricitymap/feeder/lib/fetch_data.py#L114-L115 in our backend, which is not open sourced).
Note that we internally map `sortedZoneKeys` to `sortedCountryCodes`.

We shouldn't have these type of behaviours. In the meantime I've therefore made a PR to ensure `update_state` correctly works.

PR #2866, opened 2020-12-24
Fix update state

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/2674


mockserver/update_state.py
```diff
@@ -156,5 +167,5 @@
 
 # Save
 with open('mockserver/public/v3/state', 'w') as f:
-    json.dump({'data': obj}, f)
+    json.dump({'data': obj}, f, indent=4, sort_keys=True)
```
corradio, line 170: will make it easier to see diffs in github

FelixDQ: Great idea!

FelixDQ (approved):
LGTM!

PR #2870, opened 2020-12-26
Fix error reporting missing data

corradio:
I noticed a lot of Sentry errors as we were reporting empty data returned from the server when calling /history, see e.g. https://sentry.io/organizations/tomorrow/issues/1625179506/?project=1295430&query=is%3Aunresolved
I've made sure these errors are not reported anymore.
Tested locally.


FelixDQ (approved):
Nice

PR #2871, opened 2020-12-26
Add sunrise/sunset times

corradio:
Hi all,

I've played around with another visualisation where we showcase the sunrise/sunset times on the graphs.
The times are calculated using the lonlat of the center of the zone polygon, using [this](https://github.com/udivankin/sunrise-sunset) library.
I'm not completely sure about the UI here. Maybe it's worth showing small icons of sunrise (sun?) and sunset (moon?) on the axis to make it clearer.

image


q--:
>  I'm not completely sure about the UI here. Maybe it's worth showing small icons of sunrise (sun?) and sunset (moon?) on the axis to make it clearer.


sun moon on graph

Something like this perhaps?

Kongkille:
It's a good idea. One thing in terms of UX is that it can be a bit confusing perhaps that the day/night cycle is only visible on the graph, when the data on the entire page is related to that one specific time (CI, %renewables also changes when scrolling in time). 

If we add the icons we also probably can't guarantee it will look as good on all graphs.

How about something like:

A:

image

B:
image


I like A because it's quite noninvasive. The icon could also be moved down next to the date if we wanted. "A" also potentially gives us the opportunity to develop further on this and maybe add weather indicators, eg if the wind is blowing a lot of there is a lot of sun the icon could change slightly. As such it would perhaps make it more intuitive for users to see the correlation between weather and renewable prod.


I tried to make B a bit artsy to show that it's a potential route (it's done in paint, so it's still not very good). The thing with B is mainly that the time scale is the place where you change the time and as such, it would be very intuitive that changing the time would change the day/night cycle.

corradio:
@madsnedergaard @skovhus feel free to play around with this PR if you find interesting UI solutions here

q--:
On a related note, a possible alternative might be to show local times for the zone instead of using the browser time zone.

Kongkille:
@corradio I would suggest that we perhaps close this and mention it in a new discussion on how to add sunset/sunrise. I think the current PR adds too much additional complexity to the graph, which is already quite complex. Especially for someone viewing the graph for the first time.

I think that the tech is quite cool in terms of finding the local sunset/sunrise and we should still find a way to show it on the map, but perhaps taking a step back would be good?

corradio:
> @corradio I would suggest that we perhaps close this and mention it in a new discussion on how to add sunset/sunrise. I think the current PR adds too much additional complexity to the graph, which is already quite complex. Especially for someone viewing the graph for the first time.
> 
> I think that the tech is quite cool in terms of finding the local sunset/sunrise and we should still find a way to show it on the map, but perhaps taking a step back would be good?

Hello @Kongkille ,
I agree with you: A step back would be good at this stage.

However I'd like not to loose the code, and it'd be great if this didn't get completely forgotten.
I'd suggest having an extremely simple solution out so we at least have the core functionality, as I do agree with you that the graph is too complex already, what about simply colouring the slider based on night/day? or adding a subtle axis marker to show start of day/night?

I know this is not a perfect solution, but at least it enables us to get the ball rolling by having something minimally invasive.
Alternatively I'd suggest changing the PR to just add the information to the data model without actually doing any UX changes.

Kongkille:
Yes, let's do that. I think coloring the slider just going from light grey to dark grey would work. Then it's a bit of a hidden feature, which I think is fine and then we can still create a discussion around making it more clear because I do think it should be more visible to the user whether it's day or night in a specific zone

web/src/components/graph/areagraph.js
```diff
@@ -193,6 +182,17 @@ const AreaGraph = React.memo(({
         isMobile={isMobile}
         svgNode={node}
       />
+      <TimeAxis
```
corradio, line 185: move the time/value axis on top of graph so it looks better

corradio:
@Kongkille what do you think about this solution to start with?
Having only the icons looked weird.

image

madsnedergaard:
Hmm, I think it's a bit sad that it's taking more vertical space. I'm also thinking it's adding more things with same "importance", which makes it all a bit cluttered in my opinion. How about something like this, where there's a gray bar indication night and a small icon in the "bubble" that could change depending on where it's at? I'd be happy to give it a try this week :)

Screenshot 2021-03-09 at 16 11 27

corradio:
> Hmm, I think it's a bit sad that it's taking more vertical space. I'm also thinking it's adding more things with same "importance", which makes it all a bit cluttered in my opinion. How about something like this, where there's a gray bar indication night and a small icon in the "bubble" that could change depending on where it's at? I'd be happy to give it a try this week :)
> 
> Screenshot 2021-03-09 at 16 11 27

I agree! Do you know of a good way to style input bars? Feel free to give it a try 🎩 !

madsnedergaard:
I had some fun with a codepen just to see if it's technically feasible :)

https://codepen.io/MadsNedergaard/pen/qBqQyOj

The colors, icons, etc. are just placeholders - but seems like it should be doable!
This works for Safari, Chrome, Firefox and should also work for latest Edge (not tested).

skovhus:
> I had some fun with a codepen just to see if it's technically feasible :)

Wow @madsnedergaard! This is awesome!

madsnedergaard:
Alright, I finally got around to work on this again! 😄 

### Previews

Kapture 2021-07-16 at 08 10 20
Kapture 2021-07-16 at 08 07 43

madsnedergaard:
This is ready for review, @pierresegonne and @gwpicard - would you mind taking a look? :) 

Also, a nice feature that I first thought was a bug: It shows the different lengths of night across the world 🌍 

Brazil:
Screenshot 2021-07-20 at 08 57 29

Russia:
Screenshot 2021-07-20 at 09 00 29

web/src/components/TimeSliderInput.js
```diff
@@ -0,0 +1,40 @@
...
+const createGradient = sets => sortBy(sets, 'end')
+  .map(
+    ({ start, end }) => `${COLORS.DAY} ${start}%, ${COLORS.NIGHT} ${start}%, ${COLORS.NIGHT} ${end}%, ${COLORS.DAY} ${end}%`,
+  )
+  .join(',\n');
+
+export const StyledInput = styled.input`
+  --time-gradient: linear-gradient(
+    90deg,
+    ${props => createGradient(props.nightSets)}
```
pierresegonne, line 19: not sure to understand what sets are in the context of `nightSets`.

Not sure if it's worth changing appellation, but just to let you know that the name does not obviously explicits what it refers to

web/src/components/TimeSliderInput.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const TimeSliderInput = ({
+  onChange, value, nightTimeSets, isValueAtNight, min, max,
+}) => {
+  const timeClass = isValueAtNight ? 'night' : 'day';
+  return (
+    <StyledInput
+      type="range"
+      className={`time-slider-input ${timeClass}`}
+      nightSets={nightTimeSets}
```
pierresegonne, line 32: Also curious, when looking at your preview, it does not seem like there are any gradients, so what's their use here ?

web/src/hooks/redux.js
```diff
@@ -102,3 +107,48 @@ export function useSmallLoaderVisible() {
...
+    }
+    const { latitude, longitude } = getCenteredZoneViewport(zone);
+    const nightTimes = [];
+    let baseDatetime = moment(datetimeStr).startOf('day').toDate();
+    const earliest = history && history[0] && new Date(history[0].stateDatetime);
+    const latest = new Date((history && history[history.length - 1])
+      ? history[history.length - 1].stateDatetime
+      : datetimeStr);
+    do {
+      console.log('got it')
```
pierresegonne, line 129: remove ?

web/src/hooks/redux.js
```diff
@@ -102,3 +107,48 @@ export function useSmallLoaderVisible() {
...
+      // Get last nightTime
+      const nightStart = getSunset(latitude, longitude, baseDatetime);
+      let nightEnd = getSunrise(latitude, longitude, baseDatetime);
+      // Due to some bug in the library, sometimes we get nightStart > nightEnd
+      if (nightStart.getTime() > nightEnd.getTime()) {
```
pierresegonne, line 134: Had exactly the same issue in python :) 

https://github.com/SatAgro/suntime/issues/17

web/src/hooks/redux.js
```diff
@@ -102,3 +107,48 @@ export function useSmallLoaderVisible() {
...
+    const nightTimes = [];
+    let baseDatetime = moment(datetimeStr).startOf('day').toDate();
+    const earliest = history && history[0] && new Date(history[0].stateDatetime);
+    const latest = new Date((history && history[history.length - 1])
+      ? history[history.length - 1].stateDatetime
+      : datetimeStr);
+    do {
+      console.log('got it')
+      // Get last nightTime
+      const nightStart = getSunset(latitude, longitude, baseDatetime);
```
pierresegonne, line 131: No need to handle timezones ?

web/src/hooks/redux.js
```diff
@@ -102,3 +107,48 @@ export function useSmallLoaderVisible() {
...
+      let nightEnd = getSunrise(latitude, longitude, baseDatetime);
+      // Due to some bug in the library, sometimes we get nightStart > nightEnd
+      if (nightStart.getTime() > nightEnd.getTime()) {
+        nightEnd = moment(nightEnd).add(1, 'day').toDate();
+      }
+      // Ignore nights that start after the latest time we have
+      // or that finish before the earliest time we have
+      if (nightStart.getTime() < latest.getTime() && nightEnd.getTime() > earliest.getTime()) {
+        nightTimes.push([nightStart, nightEnd]);
```
pierresegonne, line 140: Why push these dates if you want to ignore them ?

web/src/components/TimeSliderInput.js
madsnedergaard: The gradient is what sets the gray colors on the bar (the steps are "hard", so there's no fluid transition between the colors). It's a (kinda hacky) way to change the background color of an input field without having multiple elements :)

web/src/hooks/redux.js
madsnedergaard: Haha, could be that one was "inspired" by the other 😄

web/src/components/TimeSliderInput.js
madsnedergaard: I have now changed it to `nightTimeSets` to be consistent - but yeah, the naming isn't great.. Any suggestions?

web/src/hooks/redux.js
madsnedergaard: Good question! 🤔 
We probably should take it into consideration, I'll look into it now

web/src/hooks/redux.js
madsnedergaard: Hmm, Oli wrote the comment, but I think the point is that we ignore nights that doesn't match the conditional (so we only push the dates we **don't** ignore). I have rewritten the comment now

web/src/components/TimeSliderInput.js
pierresegonne: already a bit better ;)

web/src/hooks/redux.js
madsnedergaard: Turns out, no - I spent a lot of time trying to adjust it, only to realise that the timeslider is in local time and since the sunset-sunrise library returns dates in UTC it shouldn't actually match the slider 😓 

Think this can be verified by changing to different timezones locally and checking that the sunrise/sunset times match reality:
Screenshot 2021-07-21 at 02 12 29

Screenshot 2021-07-21 at 12 13 14

pierresegonne (approved):

PR #2872, opened 2020-12-26
Add Morocco power origin ratios

q--:
Needed for Spanish imports

By the way, I used the percentages from the Our World in Data percentage view and the numbers add up to 1.0002 (rounding error I guess), is this a problem?


FelixDQ (approved):
Thank you @q-- !
The rounding error shouldn't be a problem.

PR #2875, opened 2020-12-27
Fix SV parser

q--:
The parser for El Salvador is broken. ut.com.sv now [reports wind generation](http://estadistico.ut.com.sv/OperacionDiaria.aspx), and has done this by inserting it in the output as the second element, after biomass but before everything else.

As the result of this, every reported generation other than biomass at ElectricityMap is wrong because production values are mapped to the wrong type.

The fact that there is still some output on ElectricityMap is because when there's no solar output at night, it is absent from the object so the number of keys appears to be correct during that time. But when day dawns, the object has "solar" data so there is one more key than the parser is programmed to handle and it errors.

This commit fixes things by mapping wind (and consequently everything else) correctly.

# Fixing historical data
Should you want to fix historical data, this should get you started.

Here is some nocturnal parser output (again, by daylight it erred so that data wasn't even logged). See "Actual output" for which is which.

(I've obtained the 'raw' output by inserting     `print(point)` on the line below `for point in joined_data:`)

```

Raw output:
{'datetime': 5, 0: 156.93, 1: 27.93, 2: 153.9, 3: 26.4, 4: 108.86, 5: 10.91}

Correct output:

 {'datetime': datetime.datetime(2020, 12, 27, 5, 0, tzinfo=tzstr('UTC-6')),
  'production': {'biomass': 156.93,
                 'coal': 0.0,
                 'gas': 0.0,
                 'geothermal': 153.9,
                 'hydro': 26.4,
                 'nuclear': 0.0,
                 'oil': 10.91,
                 'solar': 0.0,
                 'unknown': 0.0,
                 'wind': 27.93},
                 
Actual output (from the map itself, so rounded):
geothermal: 27.9MW (actually: wind)
biomass: 157MW (actually biomass, the only correct one because it came before where wind was inserted)
solar: 10.9MW (actually: oil)
hydro: 154MW (actually: geothermal)
oil: 109MW (actually: interconnection)

```

You can find the days with incorrect output because they only have solar generation at night; historical data before wind was added to the output was of course correct (other than wind data simply not being known).


corradio (approved):
This is terrific! Do you confirm the historical parser also works, and that we can simply rerun it for the entire period?

q--:
SV.py currently can't fetch historical data. However, I recall seeing a historical data source mentioned in a different GitHub issue, so I suppose someone could try to implement it. Haven't looked into it, so I don't know if that source is still available.

corradio:
Ok thanks. Let's merge it!

PR #2886, opened 2020-12-30
Update to CO2 values for Slovenia

Lightning-5:
I've added a custom value for coal in Slovenia. Only full time coal power plant in Slovenia is TEŠ6. Official CO2 value for is TEŠ6 869g/KW. When TEŠ6 is out of service, old TEŠ5 power plant is turned on but that is rare and an exception. I believe that CO2 of TEŠ6 most accurately represents coal CO2 value.

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -704,6 +704,10 @@
         "hydro discharge": {
           "source": "2017 average by Tomorrow",
           "value": 246.34741810042195
+        },
+        "coal": {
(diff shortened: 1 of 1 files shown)


FelixDQ:
Thanks for the contribution @Lightning-5. Unfortunately, we don't currently add individual emission factors as it would be too difficult for us to validate all the different sources as they might be calculated using different criterias.  We currently rely on IPC peer reviewed emission factors, but we also want to refine those. Look at https://github.com/tmrowco/electricitymap-contrib/issues/738 to see the progress here.

PR #2887, opened 2020-12-31
ja.json updated

hisa-shimoji:
Japanese locale file updated. Mainly words for FAQ section are translated.


Kongkille (approved):
Thank you :)

PR #2888, opened 2020-12-31
Set NL unknown CO2 intensity to 468. Fixes #2804.

### LINKED ISSUE #2804: Over estimation of NL CO2/kWh emissions due to high unknown CO2 emission factor
I was wondering what the reasoning is for the assumed output of 700 g CO2eq/kWh for the part of electricity consumption where the source is unknown. 

elecmap

We know that the power generation from Coal fired power plants is very limited these days and is reported separately and quite accurate I think. So the unknown part in the Netherlands could be generated by Gas  (490 g CO2eq/kWh) or by even lower CO2 emitting sources.    [https://pbs.twimg.com/media/EnAsK0-XcAA1IUY?format=jpg&name=900x900](url)

For other countries this doesn't really matter but the Netherlands has often a fair bit of "unknown" (20-30%) so the total CO2 emission per kWh is overestimated by a fair amount! 

What the best value for the unknown part should be, I don't really know. By my opnion a max value of 490 should be reasonable. 

Maybe @MartienVisser or @PaulCornelissen  can comment?

PaulCornelissen:
To fix #2804 I made a best estimate for the CO2 intensity of NL unknown production (see discussion there).


mhooi (approved):
For the current Netherlands electricity production mix this assumption is OK. Might need some update in the future.

mhooi (approved):

mhooi:
Ok

Kongkille (approved):

PR #2869, opened 2020-12-26
Deploy electricityMap static assets to GCS

corradio:
Internal PR to simplify deployment on our side:
- upload assets to GCS
- use a consistent working directory such that Dockerfile and BUILD.yaml produce the same folder structure

~Blocked by https://github.com/tmrowco/brick/pull/74~

PR #2890, opened 2021-01-02
Mobile app CI build and deploy

corradio:
This PR adds support for continuously deploying the cordova app in our CI.
@FelixDQ @Kongkille FYI

Outstanding tasks:
- [x] Add google-services.json and GoogleService-Info.plist secret files
- [x] Test staging build
- [x] Remove all TODOs
- [x] Blocked by https://github.com/tmrowco/brick/pull/78
- [x] Blocked by https://github.com/tmrowco/electricitymap-contrib/pull/2869
- [x] Add secrets to drone config file (https://github.com/tmrowco/electricitymap/pull/47)

Now also shows current app version (when it's on mobile):
image


web/src/layout/leftpanel/mobileinfotab.js
```diff
@@ -20,6 +20,27 @@ const MobileInfoTab = () => {
...
+        appVersion, // The native version of the application this package update is intended for.
+        description, // same as given during deployment
+        // isFirstRun, // flag indicating if the current application run is the first one after the package was applied.
+        label, // The internal label automatically given to the update by the CodePush server, such as v5. This value uniquely identifies the update within it's deployment
+      } = localPackage;
+
+      setMobileAppVersion(`${appVersion} ${label} (${description})`);
+    }, err => console.error(err));
+  }
+
```
FelixDQ, line 43: Would this be slightly better with a `useEffect`? 

i.e. instead of
```js
if (!mobileAppVersion && window.isCordova) { 
  ...
}
```
It can be
```js
useEffect(() => {
  if (window.isCordova) { 
    ...
  }
}, []);
```

web/BUILD.yaml
```diff
@@ -11,7 +11,7 @@ steps:
 
   build:
     environment:
-      ELECTRICITYMAP_PUBLIC_TOKEN: ${BRICK_ELECTRICITYMAP_PUBLIC_TOKEN}
+      ELECTRICITYMAP_PUBLIC_TOKEN: ${BRICK_ELECTRICITYMAP_PUBLIC_TOKEN:-unknown}
```
FelixDQ, line 14: What does this `:-unknown` mean?

FelixDQ (approved):
This is awesome!

I added a minor comment, but it looks good to me 🎉

web/BUILD.yaml
corradio: it's a fallback value for when the env variable is not defined.
It's due to https://github.com/tmrowco/brick/issues/79

web/src/layout/leftpanel/mobileinfotab.js
corradio: I'm not that familiar with `useEffect` - what difference would it have?
Happy to do the change either way ;-)

FelixDQ: I don't think that it makes a big difference here, but the current if-statement gets evaluated on every render, which means that if `codePush.getCurrentPackage` is slow, then it could potentially get triggered multiple times. `useEffect` with an empty array will make sure that it only runs on the first render.

My brain also reads the `useEffect` with an empty array like the `onComponentDidMount`, so it's a bit easier for me to understand when it's supposed to run. But that's just a personal preference :-)

corradio: understood! Did the changes.

FelixDQ (approved):
LGTM

PR #2895, opened 2021-01-04
Add Kazakhstan power origin ratios

q--:
Needed for Russian imports

I calculated the percentages from the raw numbers, unlike in #2872 where I copied them from Our World in Data, for a slight increase in precision.

"unknown" is "other renewables", should it be marked as renewable somehow? (It's only a very small part of the generation mix")


corradio:
Hello! Thank you for doing this. If unknown is actually mostly renewables, then I suggest changing the emission factor for unknown for this particular country?
Thanks 💪

corradio:
Hello @q-- 

I'll merge this right now and we can refine the `unknown` part later if needed. Thanks!

Olivier

corradio (approved):

PR #2896, opened 2021-01-04
Add Thailand power origin ratios

q--:
Needed for Peninsular Malaysia.

From IEA 2019 (used because Our World in Data has a lot of 'other renewables' generation not broken down by type).
Biomass includes waste.


FelixDQ (approved):
LGTM

PR #2898, opened 2021-01-05
Implementation of enhanced parser for estimating the Swiss electricity carbon footprint

Kikof2:
The carbon footprint of the Swiss electricity consumption is currently unavailable on electricitymap.org. The main reason has been in large discrepancies of input data and official, national values reported for Switzerland. Now, we present our approach that drastically decreased initial deviations (>30%) towards an acceptable level (<5%).

During the [Opendata Energy Data Hackdays 2020](https://opendata.ch/projects/energy-data-hackdays-2020/) in August we (a team of some energy specialist in Switzerland and developers) sat together to analyze the current situation in Switzerland. We focused on two topics:

1. Which data is missing in the ENTSOE values?
2. Improve live data quality of Switzerland.

### Which data is missing?

In general the Swiss production values are listed in the [ENTSOE transparency platform](https://transparency.entsoe.eu/), but as was pointed out in the past there is data missing from these values. So we compared for 2019 the aggregates of the ENTSOE transparency platform with the [annual report of the Swiss government](https://www.bfe.admin.ch/bfe/de/home/versorgung/statistik-und-geodaten/energiestatistiken/gesamtenergiestatistik.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvMTAxMzg=.html) (Table 24 on page 39).

image
*Swiss electricity production in TWh in 2019*

For most production categories the values (in TWh) align. However, for run-of-river less than 2 of the 17.7 TWh are included in the ENTSOE values. This was confirmed by the energy specialists in our team. Switzerland has a lot of small run-of-river plants that are so small that they do not have to report live data, so that this data is not available. Conventional thermal is the other production category that is missing in the data. Those are mainly waste incineration plants. We are currently also undertaking efforts to improve the quality of the data regarding solar power with better national data.

Knowing the missing production categories we estimated an average CO2 emission value of **131 gCo2/kWh**. The value is the average of the missing categories derived by the IPPC report.

### Improve live data quality of Switzerland.

As described above no live data for certain categories (run-of-river, waste incineration) is available. Therefore, we decided for a solution similar to the one that has been developed for the Netherlands that faced similar challenges in the past: The total production of Switzerland is calculated as the difference between the consumption and the exchanges with neighbouring countries. The difference do the known electricity production (sum of all categories) will be assigned to the category "unkown" and the emission value **131 gCo2/kWh** (see above) is applied.

image
*Schema of the solution and validation of the solution with Swissgrid values*

Swissgrid (the electricity grid operator in Switzerland) publishes on their website [the total hourly electricity production in Switzerland](https://www.swissgrid.ch/de/home/operation/grid-data/generation.html). Unfortunately, this data is not live and is not separated in production categories. However, it can be used to verify the approach.

image
*Comparison of hourly gap between Swissgrid <-> ENTSOE values (red) and Swissgrid <-> improved approach for 2019*

The red curve shows the difference between the Swissgrid values and the aggregated known production of ENTSOE (sum of all production categories). The black curve shows the difference between the Swissgrid values and the total production after applying our solution. (see diagram above) Thus, by applying the solution the missing electricity production in the Swiss values is reduced from ~30% to <5%.

With the strong improvements brought by the presented approach, that led to a significant reduction in deviation of values, we are now confident that Swiss can be put back on the electricity map again


corradio (commented):
Hello!
Thank you so much for the great write-up! I've gone through the PR and it looks good.
For validation purposes I'd like to make sure the new parser aligns better with the Swissgrid data (for which you provided a plot).
Are you able to use this code to generate a month of data, calculate the total load and compare it to the Swiss grid data (just like the plot you showed)?
Then we'll be able to merge right after.

Also I think you should add yourselves as contributors:
https://github.com/tmrowco/electricitymap-contrib/pull/2898/files#diff-5fdb410dda63034a171c7bacd44c070de0c6a2d0ce5c0d5436b3626877b6a583R1116-R1119

Let me know if you have any questions!

Kikof2:
@corradio Thanks a lot for your feedback. I added myself as contributor for Switzerland.

I validated the results of the new approach. As you can see below, the new approach (blue) matches the Swissgrid data (green) significantly better than the raw ENTSOE values (orange).


image
*Comparison of Swiss Electricity production for Swissgrid (green), ENTSOE (orange) and the newly implemented approach (blue)*

I also attach a pdf created from the Jupyter Notebook I used to validate the approach so you can understand and reproduce the steps. Unfortunately, I cannot attach the original notebook. Text me if you need the notebook or if any further validation is required.

[ValidationSwissElectricityProduction.pdf](https://github.com/tmrowco/electricitymap-contrib/files/5807471/ValidationSwissElectricityProduction.pdf)

corradio (approved):
Amazing! Let's get this into production

PaulCornelissen:
This is a great development! 👍

But I think you forgot to add the custom calculated CO2 intensity for unknown to [co2eq_parameters.json](https://github.com/tmrowco/electricitymap-contrib/blob/master/config/co2eq_parameters.json) 

image

Kikof2:
@PaulCornelissen That is a great catch. Thanks a lot. I will fix it tonight and open a second pull request to fix the issue.

PR #2899, opened 2021-01-05
Update mapbox

corradio:
Update to latest non-paying version of mapbox.
I did this to see if some performance benefits could be obtained.
Tested locally, works!


web/src/reducers/dataReducer.js
```diff
@@ -134,9 +134,6 @@ module.exports = (state = initialDataState, action) => {
             console.warn(`${key} produces more than its capacity of ${mode}`);
           }
         });
-        if (!zone.exchange || !Object.keys(zone.exchange).length) {
```
corradio, line 137: removed those confusing logs

FelixDQ (approved):

PR #2900, opened 2021-01-05
Fix viewport hook bug

corradio:
I saw this error in the console:

Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in Tooltip (created by MapCountryTooltip)
    in MapCountryTooltip (created by ConnectFunction)
    in ConnectFunction (created by CountryHistoryCarbonGraph)
    in CountryHistoryCarbonGraph (created by ConnectFunction)
    in ConnectFunction (created by CountryPanel)
    in div (created by CountryPanel)
    in div (created by CountryPanel)
    in div (created by CountryPanel)
    in CountryPanel (created by ConnectFunction)
    in ConnectFunction (created by ZoneDetailsPanel)
    in div (created by ZoneDetailsPanel)
    in ZoneDetailsPanel (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in Route (created by LeftPanel)
    in Switch (created by LeftPanel)
    in div (created by LeftPanel)
    in LeftPanel (created by ConnectFunction)
    in ConnectFunction (created by Main)
    in div (created by Main)
    in div (created by Main)
    in Main (created by ConnectFunction)
    in ConnectFunction
    in Router
    in Provider

After some investigation, it happened because of `useWidthObserver` which was called without the empty array argument, meaning that it should only run once.
After reading https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node, I noticed there was a better way to read the dimensions, and to obtain the DOM node.


web/src/components/circulargauge.js
```diff
@@ -6,6 +6,12 @@ import { arc } from 'd3-shape';
 /* eslint-disable jsx-a11y/mouse-events-have-key-events */
 // TODO: re-enable rule
 
+/*
```
corradio, line 9: a note here related to another bug that I saw sometimes:

Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in Motion
    in g
    in g
    in svg
    in div
    in Unknown (created by TooltipContent)
    in div (created by TooltipContent)
    in div (created by TooltipContent)
    in div (created by TooltipContent)
    in div (created by TooltipContent)
    in TooltipContent (created by MapCountryTooltip)
    in div (created by Tooltip)
    in Portal (created by Tooltip)
    in Tooltip (created by MapCountryTooltip)
    in MapCountryTooltip (created by ConnectFunction)
    in ConnectFunction (created by CountryHistoryCarbonGraph)

web/src/components/graph/graphhoverline.js
```diff
@@ -10,7 +10,7 @@ const GraphHoverLine = React.memo(({
   markerHideHandler,
   selectedLayerIndex,
   selectedTimeIndex,
-  svgRef,
+  svgNode,
```
corradio, line 13: a DOM node is now passed instead of a ref

web/src/components/tooltip.js
```diff
@@ -24,38 +24,46 @@ const Tooltip = ({
...
-    // TODO(olc): Once the tooltip has taken 100% width, it's width will always be 100%
-    // as we base our decision to revert based on the current width
-    style.width = '100%';
-  } else {
-    x = position.x + MARGIN;
-    // Check that tooltip does not go over the right bound
-    if (width + x >= screenWidth) {
-      // Put it on the left side
-      x = position.x - width - MARGIN;
+  if (hasDimensions) {
```
corradio, line 43: I wrapped all the resizing logic inside this wrapper in order to make sure this doesn't happen before it should. This helped a bit with performance and fixed some weird layout bugs I could see at some point

web/src/hooks/viewport.js
```diff
@@ -1,49 +1,33 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useCallback } from 'react';
 
-export function useWidthObserver(ref, offset = 0) {
+export function useRefWidthHeightObserver(offsetX = 0, offsetY = 0) {
```
corradio, line 3: this is the core of the change

web/src/hooks/viewport.js
```diff
@@ -1,49 +1,35 @@
...
     };
     // Set initial width
-    updateWidth();
+    update();
     // Update container width on every resize
-    window.addEventListener('resize', updateWidth);
-    return () => {
-      window.removeEventListener('resize', updateWidth);
-    };
-  });
```
fbarl, line 21: I really don't know what was going through my mind when I didn't list the dependencies here :confused: I'm glad you caught this! :+1:

web/src/hooks/viewport.js
```diff
@@ -1,49 +1,35 @@
...
-    updateHeight();
-    // Update height on every resize
-    window.addEventListener('resize', updateHeight);
+    window.addEventListener('resize', update);
     return () => {
-      window.removeEventListener('resize', updateHeight);
+      window.removeEventListener('resize', update);
     };
-  });
+  }, [offsetX, offsetY]);
```
fbarl, line 28: I feel we should probably add `width`, `height` and `node` to the dependencies here as long as we're doing the inequality checks `if (var !== newVar)`, otherwise we might potentially run into inaccuracies comparing to the old values.

Alternatively, maybe all 3 checks could just be dropped to keep it simple as in the [docs example](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node)?

corradio: Wouldn't that mean that `useCallback` is re-executed when the `width` changes, potentially returning a new `ref`?

fbarl: Yes, but the new ref wouldn't make any changes then, right? It's a bit clunky, I agree. So maybe it's just better to keep it simple and drop the `if` conditions :thinking: 

Just to illustrate why I think the current code is problematic, imagine we start with both `width` and `newWidth` set to 500 and then `newWidth` gets changed to 600, which triggers a call to `setWidth` which supposedly updates the `width` externally to 600 as well. But if my understanding is correct, since `width` is not in the dependencies, the cached callback function will not be changed and it will still be using the old value for `width` which is 500. So when `newWidth` gets changed back to 500, the if-condition will prevent `setWidth` to be called so `useRefWidthHeightObserver` will keep returning 600 instead of the correct width of 500 now. Does that make sense?

web/src/components/tooltip.js
```diff
@@ -24,42 +24,50 @@ const Tooltip = ({
...
+    height,
+  } = useRefWidthHeightObserver();
 
   const screenWidth = window.innerWidth;
   const screenHeight = window.innerHeight;
 
   if (!position) return null;
 
+  // Note: at first render, width and height will be undefined
+  // They will only be set once the DOM node has been created
```
fbarl, line 39: Wait, according to `useRefWidthHeightObserver` below, they actually start with a value `0`. Maybe we should indeed change it to `null` or `undefined` or update this comment and put a more explicit condition below (e.g. `width > 0 && height > 0`) to avoid new corner cases and keep it simple.

fbarl (approved):
Alright, I reviewed the code and all LGTM except that `useCallback` method for `ref` that I commented on - I'm approving the PR nevertheless to unblock you if the behaviour is still correct and my concerns apply only to marginal cases (it'd be cool to add a comment in that case)!

Glad you addressed this though, these changes were much needed and it's a clear improvement of the code quality! :ok_hand: :pray:

web/src/hooks/viewport.js
corradio: @fbarl I think I understand. Are you saying that `width` inside the closure of `update = () { ... }` points to the *initial*  value of the `width` state, and thus, if the `width` state is not part of the dependencies, then the callback will still compare the initial `width`.
That seems sensible to me.

One thing to note though is that the returned `ref` is actually a callback function `(node) =>  { ... }` which gets executed once the `ref` (function) is attached to the DOM.
Changing the `ref` by instantiating a new `callback` will recreate a new callback that will never be called, as it will never be re-attached (DOM element is already created).

The simplest solution is to get rid of the `if` conditions.

web/src/components/tooltip.js
corradio: good point

fbarl (approved):
Thanks for addressing my feedback @corradio - I think the `useRefWidthHeightObserver` hook's a bit easier to read and understand this way and if the behaviour is good feel free to merge it! :star:

PR #2906, opened 2021-01-08
(French locale) Some improvements

Situphen:
Just some modifications to improve the french locale.


veqtrus:
What is the rationale for putting spaces before `: ! ?`?

Situphen:
> What is the rationale for putting spaces before `: ! ?`?

Consistency. At the moment we have a mix of sentences with spaces before `: ! ?` and others without. I chose to follow the convention in France which is to put spaces before `: ! ?` but we can choose to use another convention if we want (I should have specified this before, sorry). It would be even better if those spaces where non-breaking spaces but I wasn't able to do so with my text editor (I don't why and I may try again later).

veqtrus:
> I wasn't able to do so with my text editor (I don't why and I may try again later).

Could you try copying <code>&nbsp;</code>?

veqtrus:
Now that I think about it perhaps it would be better to write it explicitly as `&nbsp;` so that someone else's editor doesn't convert them back to normal spaces.

Situphen:
I have found a way of inserting a non-breaking space by doing Ctrl+Shift+U then typing the hexacode 00A0. And I just found the right shortcut on my keyboard to do it (Shift-Alt Gr-Space). The issue wasn't my text editor but rather that I can't copy a non-breaking space from Firefox (when I paste it, it is a regular space).

corradio (approved):
Looks good to me! Ill let you merge

Situphen:
I have rebased it so anyone with write access can merge it.

PR #2907, opened 2021-01-10
:art: Bundle parsers in a python package

wizmer:
- Requirements are now in the `setup.py` and the requirements file has been removed
- no more relative import in the parser package (but still in the test files)
- test file are not included in the package but can be run by `python setup.py test`
- in tests use `pathlib` module to resolve relative paths
- moved `utils/config/` into new package folder `parsers/parsers/utils/config`
- The `test_parser.py` file is moved into the package and the CLI `test-parser`
is registered in `setup.py`.
- Use `tox` to run the tests
So that parsers can more easily be tested locally. The user simply have to
`pip install tox` and then run `tox` to test both the unit tests and the linting.
- top level `__init__.py` is now unnecessary and has been removed


corradio:
cc @FelixDQ @skovhus as we thought about it as well

skovhus:
@wizmer thank you so much for the pull request. 

We are currently looking into using https://python-poetry.org/, which should help with some of the issues that you raise. Are you familiar with Poetry?

wizmer:
No, I have never used poetry. I'm having a look right now but in all honesty, I'm not sure that it brings much to the table compared to simply maintaining a proper `setup.py`.

skovhus:
Thanks for your feedback, I appreciate the honesty. I'm aware that tooling and architecture decisions are subjective preferences.

Internally Poetry has several advantages for us at Tomorrow:
- lock files for all dependencies (reproducibility and reliability)
- automatically setup of virtual environments (isolation)
- great support for mono repositories (which we love at Tomorrow)
- covering equivalent functionality of both Pipenv and setup.py and an enjoyable CLI

wizmer:
Ok, I'll try to make it work with poetry then. It'll be a good occasion to learn something new.

skovhus:
> Ok, I'll try to make it work with poetry then. It'll be a good occasion to learn something new.

Note that we are already working on this – I expect to have some ready for review tomorrow or later this week. I’ll ping you once I have a PR ready. I would like to understand if it solves the requirements you had.

Thanks again for your work here and for pushing for a better development environment! Much appreciated.

PR #2909, opened 2021-01-11
 :memo: Point to the documentation page when a token is not found 

wizmer:
https://github.com/tmrowco/electricitymap-contrib/wiki/Create-tokens


FelixDQ (approved):
Great idea! Thank you @wizmer! 💪

Issue #2339, opened 2020-03-30
Investigate switching data source for Ireland


Lobidu:
Hi,
the data for Ireland has been missing for a while now. As it has been working before, is there anything broken that I could help resolve?
Thanks
Janis

AyrtonB:
I believe that Ireland is currently using ENTSOE data, bizarrely ENTSOE has no data for Ireland the country and only provides information on the SEM market which became ISEM back in 2018.

Northern Ireland & the Republic of Ireland are both part of the same electricity market so it also doesn't make sense that the map currently shows Northern Ireland only. 

There are two sites we could get Irish data from apart from ENTSOE, the official route through https://www.sem-o.com/ or scraping the dashboard from http://smartgriddashboard.eirgrid.com/. The EirGrid dashboard is at 15min res whilst SEM-O is at 30mins.

AyrtonB:
```
""" Caller """
class Caller(object):
    ## Core Query Caller
    def query_API(self, start_date:datetime, end_date:datetime, area:str, region:str):## Formatting & Forming Query
        ## Formatting & Checking Parameters
        start_date, end_date = self.format_date_inputs(start_date, end_date) # Format datetimes
                
        ## Checking stream is an EI endpoint
        assert area in self.possible_areas, f"Area must be one of {''.join([stream+', ' for stream in self.possible_areas])[:-2]}"
        assert region in self.possible_regions, f"Region must be one of {''.join([time_group+', ' for time_group in self.possible_regions])[:-2]}"

        ## Creating Query
        query_params_dict = self.create_query_params_dict(area, region, start_date, end_date) # create dictionary of parameters and their values
        
        if area == 'marketdata':
            data_type = 'marketdata'
        else:
            data_type = 'data'
            
        query = self.params_dict_2_query(query_params_dict, data_type=data_type) # create the url query string

        ## Getting & Parsing Response
        response = requests.get(query)
        r_json = response.json()

        ## Checking Response is OK
        response_metadata = self.get_metadata(r_json)
        error = response_metadata['error_message']

        assert isinstance(error, type(None)), f'Error with API call: {error}'
        assert response_metadata['num_rows'] > 0, 'No rows were returned'
        
        return r_json
    
    ## Helper Functions
    def format_date_inputs(self, start_date, end_date):
        format_dt = lambda dt: datetime.strftime(dt, '%d-%b-%Y+%H:%M').replace(':', '%3A') if isinstance(dt, date) else dt

[...]

PR #2916, opened 2021-01-11
POC for the Eirgrid parser for Ireland

wizmer:
Based on the work of @AyrtonB in issue #2339

Eirgrid provides data from northern Ireland, Republic Of Ireland and both
combined. So the parser can be used for both regions if needed.

## fetch_production 
The granularity of the mix is not very well defined so I don't know if this is usable.
See here: http://smartgriddashboard.eirgrid.com/#roi/generation

Data is simply divided into:

- coal
- gas
- net export
- other
- renewable

According to the legend of the plot:
```
The Fuel Mix Graph is broken down into Gas, Coal, Renewables, Oil, Net import
and Other. The “Other” category includes Peat, Combined Heat and Power (CHP),
Aggregated Generating Units (AGUs), Demand Side Units (DSUs), Distillate and
Waste.
```

The time granularity is 15 min.

## fetch_price

I haven't found the data. There is something called `Imbalance Settlement Price`
at http://smartgriddashboard.eirgrid.com/#roi/market-pricing but I don't think
that's what we want

## fetch_exchange

I haven't found this information either. However the request in `fetch_production` returns
something called `FUEL_NET_IMPORT` but I don't know if this can be used.


daviessm:
The dashboard also has wind data separately - does this match the renewable percentage? http://smartgriddashboard.eirgrid.com/#roi/wind

parsers/ROI.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+from datetime import date, datetime, time, timedelta
+
+import pandas as pd
+
+import arrow
+import requests
+
+
+QUERY_ROOT = f'http://smartgriddashboard.eirgrid.com/DashboardService.svc/data?'
```
veqtrus, line 11: This doesn't have any parameters so it doesn't need to be a formatting string.

parsers/ROI.py
```diff
@@ -0,0 +1,52 @@
...
+
+MAPPING = {'FUEL_COAL': 'coal',
+           'FUEL_GAS': 'gas',
+           'FUEL_RENEW': 'renew',
```
veqtrus, line 15: I'm not sure if the backend accepts 'renew'.

parsers/ROI.py
```diff
@@ -0,0 +1,52 @@
...
+def fetch_production(zone_key='ROI', session=None,
+                     target_datetime: datetime = None,
+                     logger: logging.Logger = logging.getLogger(__name__)):
+
+    r = session or requests.session()
+
+    if target_datetime is None:
+        target_datetime = datetime.now()
+
+    url = QUERY_ROOT + ''.join(f'&{query_key}={query_value}' for query_key, query_value
```
veqtrus, line 33: The first argument has an extra `&`. I would rewrite it as:
```python3
datefrom = format_date_inputs(target_datetime - timedelta(minutes=15))
dateto = format_date_inputs(target_datetime)
url = QUERY_ROOT + f'area=fuelmix&region={zone_key}&datefrom={datefrom}&dateto={dateto}'
```

parsers/ROI.py
```diff
@@ -0,0 +1,52 @@
...
+    assert obj['Status'] == 'Success', 'Exception when fetching production for ' \
+        '{}: error when calling url={}'.format(zone_key, url)
+
+    data = {
+        'zoneKey': zone_key,
+        'production': {MAPPING[row['FieldName']]: float(row['Value']) for row in obj['Rows']
+                       if row['FieldName'] in MAPPING},
+        'storage': {},
+        'source': 'smartgriddashboard.eirgrid.com',
+        'datetme': datetime.strptime(obj['Rows'][0]['EffectiveTime'], '%d-%b-%Y %H:%M:%S').datetime,
```
veqtrus, line 49: This might not be setting the right timezone. Most parsers use `arrow` which is a nicer library.
Also typo: 'datetme' should be 'datetime'.

parsers/ROI.py
```diff
@@ -0,0 +1,52 @@
...
+
+    url = QUERY_ROOT + ''.join(f'&{query_key}={query_value}' for query_key, query_value
+                               in {'area': 'fuelmix', 'region': zone_key,
+                                   'datefrom': format_date_inputs(target_datetime - timedelta(minutes=15)),
+                                   'dateto': format_date_inputs(target_datetime)}.items())
+
+    res = r.get(url)
+    obj = res.json()
+    assert obj['Status'] == 'Success', 'Exception when fetching production for ' \
+        '{}: error when calling url={}'.format(zone_key, url)
```
veqtrus, line 41: You could use a format string here like `f'{zone_key}: error when calling url={url}'`.

veqtrus (changes requested):

veqtrus:
BTW country-specific parsers use the two-letter country code.

Manu1400:
> ## fetch_exchange
> I haven't found this information either. However the request in `fetch_production` returns
> something called `FUEL_NET_IMPORT` but I don't know if this can be used.

CSV : GET https://smartgriddashboard.eirgrid.com/DashboardService.svc/csv?area=interconnection&region=ALL&datefrom=20-Jan-2021%2000:00&dateto=20-Jan-2021%2023:59
or
JSON : GET https://smartgriddashboard.eirgrid.com/DashboardService.svc/data?area=interconnection&region=ALL&datefrom=20-Jan-2021%2000:00&dateto=20-Jan-2021%2023:59

Manu1400:
Actual zones : 
* https://www.electricitymap.org/zone/IE
* https://www.electricitymap.org/zone/GB-NIR

but the proposal filename is `ROI.py` ?

FelixDQ:
Nice job everyone! Regarding the production data:

We would need to try and make an assumption regarding the mix of the renewable as we don't have a renewable type. Looking at the historical data from ENTSOE, the primary renewable source is wind (and a little bit of hydro):
image

Would one of you try and dive into the data and investigate the possibility?

It would also be good to validate this data against the ENTSOE data, to see if there are any discepencies.

FelixDQ:
I think that there might be something wrong with the scaling of the production values
```
20-Jan-2021 23:00:00: {'coal': 42672.21, 'gas': 199958.21, 'unknown': 22195.2, 'renew': 76653.09}
```
which I can't really match with the wind generation on [the dashboard](http://smartgriddashboard.eirgrid.com/#roi/wind) which reads 1284 MW at the same point in time

veqtrus:
Interestingly ENTSO-E seems to have data now.

parsers/ROI.py
wizmer: Indeed, should it go as 'unknown' then ?

parsers/ROI.py
wizmer: Cool, I did not know arrow, that's much better !

wizmer:
> Interestingly ENTSO-E seems to have data now. 

Should I keep going for the Eirgrid parser then or directly look at the ENTSO-E data ?

parsers/ROI.py
veqtrus: ENTSO-E separates renewables. https://www.electricitymap.org/zone/IE

wizmer:
> but the proposal filename is ROI.py ?

I have renamed to IE.py

> I think that there might be something wrong with the scaling of the production values

Yes, I don't know what is the absolute unit. But we can compute the scaling factor based on the total amount of produced energy.

> It would also be good to validate this data against the ENTSOE data, to see if there are any discrepancies.

I'll try to find the time for it this week

corradio:
Hello @wizmer,

Thank you for helping here! Have you had time to compare against ENTSOE? We're curious to see how this new parser compares in terms of data quality.

Olivier

corradio:
Hi again,

I recently opened https://github.com/tmrowco/electricitymap-contrib/issues/3060 as GB-NIR is down and thought the EIRGRID parser built here could be useful.
I therefore suggest changing the focus from IE to GB-NIR in this PR.

This PR would be useful for GB-NIR.
I suggest:
- making sure that we double check that the returned datetime matches the target datetime (maybe check it's the same day at least?). I suggest raising an exception if the data doesn't match. I found out that changing the url's datetime doesn't change the result :/
- making sure that the parser works with both IE and GB-NIR as input (the `region` parameter is respectively ROI and NI)
- rename the parser to be called EIRGRID.py (it should describe the data provider and not the zone)
- update zones.json and exchanges.json to make sure GB-NIR uses EIRGRID.py, and that interconnectors and prices work.

Does that make sense?

Manu1400:
GET https://www.smartgriddashboard.com/DashboardService.svc/data?area=fuelmix&region=ALL&datefrom=31-Jul-2021+00%3A00&dateto=31-Jul-2021+23%3A59&_=1627689659160 (used by https://www.eirgridgroup.com/how-the-grid-works/system-information/ ) give live data :

```json
{
  "ErrorMessage":null,
  "LastUpdated":"31-Jul-2021 01:04:18",
  "Rows":[{"EffectiveTime":"31-Jul-2021 01:00:00","FieldName":"FUEL_COAL","Region":"ALL","Value":6578.45},{"EffectiveTime":"31-Jul-2021 01:00:00","FieldName":"FUEL_GAS","Region":"ALL","Value":35848.51},{"EffectiveTime":"31-Jul-2021 01:00:00","FieldName":"FUEL_NET_IMPORT","Region":"ALL","Value":14218.85},{"EffectiveTime":"31-Jul-2021 01:00:00","FieldName":"FUEL_OTHER_FOSSIL","Region":"ALL","Value":2962.96},{"EffectiveTime":"31-Jul-2021 01:00:00","FieldName":"FUEL_RENEW","Region":"ALL","Value":32456.63}],
  "Status":"Success"
}
```

wizmer:
Hi ! I have so time to work on this one. Would it be still relevant ? @veqtrus @Manu1400 
Thanks !

madsnedergaard:
Closing this again as Eirgrid only provides breakdown on a daily level (https://github.com/electricitymap/electricitymap-contrib/issues/3383#issuecomment-1003087224) and ENTSO-E is working fine for Ireland.
We'll probably revisit this topic again in the future, but for now I don't think switching parser will improve the data.

wizmer:
Okay, should I keep my branch alive just in case or can I delete it ? Thanks

madsnedergaard:
Let's keep it around in case the data provider becomes more usable later :) Also thanks for the work you put into this!

PR #2919, opened 2021-01-12
Fix node ref for wind/solar layers

corradio:
Forgot that `ref.current` is not defined anymore, but that we need to use the `node` returned by the hook.


fbarl:
Related to https://github.com/tmrowco/electricitymap-contrib/pull/2900.

fbarl (approved):
Good catch, code LGTM! :+1:

PR #2921, opened 2021-01-12
Reduces height of header and adjusts font-size

### LINKED ISSUE #2876: Reduce header size
Tested on multiple screens. I think we can reduce header height by about 8 px and then reduce font size accordingly. Prob from about 16 to 14. 

- [ ] Reduce header height
- [ ] Reduce header font size

madsnedergaard:
## Description

Reduces header size (from 66px to 58px).

Also decreases font-size in header (from 16px to 14px). I think 16px might actually be fine, it's not the best font for smaller text sizes :)

### Preview

Dunno if this helps much, but I did a gif with the different font-sizes 😅 

What do you think, @Kongkille?

fontsize

Fixes #2876 

PS. Let me know if I should do the same for the other sites (API/open-source, etc.)

### DIFF
--- a/web/src/components/sharedheader.js
+++ b/web/src/components/sharedheader.js
@@ -9,10 +9,10 @@ const Wrapper = styled.header`
   color: black;
   display: flex;
   font-family: 'Euclid Triangle';
-  font-size: 16px;
-  height: 66px;
+  font-size: 14px;
+  height: 58px;
   justify-content: space-between;
-  min-height: 66px; /* required for old Safari */
+  min-height: 58px; /* required for old Safari */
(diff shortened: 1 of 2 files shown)


Kongkille:
Very cool @madsnedergaard ! And yes we also need to make sure this looks good on API and Open source! :D

Kongkille:
And I think maybe 15px then?

madsnedergaard:
> Very cool @madsnedergaard ! And yes we also need to make sure this looks good on API and Open source! :D

Ah, didn't realise the API site was using the same component - just tested it there as well and it still looks good 👌

Kongkille (approved):

Issue #2533, opened 2020-05-31
Separate Flinders Island from Tasmania


chuqtas:
(first submission, please be nice if I've done it wrong!)

This request is similar to the King Island request:  https://github.com/tmrowco/electricitymap-contrib/issues/636 "split Australia (Tasmania) to show King island off grid system"

Basically it relates to another island with many similarities = it is in the same area of the world, also with it's own off-grid power system, also built and operated by Hydro Tasmania, also with a web & app dashboard for live info.

However - it's dashboard is a totally different design and therefore parsing the data might be easier (or maybe no different - no idea).

There are two parts to this issue:

1) Re-arranging regions:
- Remove Flinders Island and nearby island from the "Tasmania" region
- Create new "Flinders Island (Australia)" region with just the large one
- Leave the small one blank - if it needs a name, call it "Cape Barren Island (Australia)".
- The island at the bottom of Tasmania should stay as part of the "Tasmania" region, no changes needed

![image](https://user-images.githubusercontent.com/51102336/83344283-f1071480-a347-11ea-99b3-6417e5ecc770.png)

2) Attempt to extract data from the dashboard to apply to this region.

The website is - https://www.hydro.com.au/clean-energy/hybrid-energy-solutions/success-stories/flinders-island for anyone who is able to make sense on whether this data can be extracted or not.

Hope I've provided enough info - can't be of much use with the data extraction/coding though :(

q--:
I'm seeing requests to `data.ajenti.com.au/live/signalr/` on that page in my browser console, which is also present in https://github.com/tmrowco/electricitymap-contrib/pull/2456/files#diff-c3b7bb94e70c63d689a577143fcab2a9R12

So the user interface might look different, but under the hood they're probably very similar. Might even make sense to try to combine the parser for this and King Island

pierresegonne:
Hey @chuqtas! Thanks for the thorough investigation and breakdown of the tasks. This would be a good place to start contributing! 

1.  

This should be a first PR, few elements to help you

- _Remove Flinders Island and nearby island from the "Tasmania" region_ -> look into https://github.com/tmrowco/electricitymap-contrib/blob/master/web/README.md to understand how the geometries of the zones are generated
- _Create new "Flinders Island (Australia)" region with just the large one_ -> This needs creating a new zone in `zones.json` and add its geometry accordingly



2. 

This should be a second PR, or could be embedded in #2534 for example


Let me know if you need more details :)

[...]

PR #2925, opened 2021-01-14
Separate Flinders Island and Cape Barren Island regions from Tasmania

chuqtas:
This change alters geographic regions only.

- Creates a new region "AUS-TAS-FI" for Flinders Island
- Creates a new region "AUS-TAS-CBI" for Cape Barren Island
   - Note: It does not include providing data for either of these regions.
- Modifies the region "AUS-TAS" to remove the above islands (also a small number of other nearby islands which are not part of the state power grid)

These changes are detailed in part 1 of [issue #2533](https://github.com/tmrowco/electricitymap-contrib/issues/2533)


chuqtas:
After reviewing the King Island change I've just noticed I hadn't updated /web/locales/en.json or config/zones.json.
I've just added a commit to update en.json, I don't *think* I need to edit zones.json yet because its just geography, no data yet.

FelixDQ (changes requested):
Awesome! Only some small comments and then we should be ready to merge:

1. The system doesn't like geojson FeatureCollections for some reason, so you would need to extract the feature. I.e. Instead of writing:
```json
{
  "type": "FeatureCollection",
  "features": [
    {"type":"Feature", ...}
  ]
}
```
It should just be:
```json
{"type":"Feature", ...}
```



2. And then we need to generate a new `world.json`.  This can be done with the following command (you need docker):
```
docker-compose run --rm web ./topogen.sh
```
See more [here](https://github.com/tmrowco/electricitymap-contrib/tree/master/web)


You can also validate that it looks good by running the map using
```
docker-compose up
```


Let me know if you need help or if I should try to generate the new `world.json`

chuqtas (commented):
Remove featureCollection references from geojson files.

chuqtas:
Thanks @FelixDQ!  I have made the necessary changes to the geojson files.  Regarding building the new world.json, I don't have access to a suitable dev environment.  Hopefully you or someone else can help with that?

FelixDQ (approved):
Thank you @chuqtas! I regenerated world.json :-)

chuqtas:
Thanks for the merge and the earlier assistance with this @FelixDQ  :)

PR #2933, opened 2021-01-17
Show emissions in tooltip

corradio:
Hello folks!
Hereby a PR with a quick idea I wanted to prototype.
Some countries have a minor part of their import/production contributing to a large part of their footprint.
I wanted to highlight that in the tooltip.
Below is a screenshot of Norway importing from the Netherlands at a time where NL power is quite dirty.

Tell me thumbs up / down depending on what you think!

image


corradio:
@madsnedergaard @Kongkille let me know what you think here and if it makes sense to review/merge!

madsnedergaard:
Seems pretty good to me!

Do you have a screenshot of the tooltip for production as well? (the one seen here is for the exchange). 

I'm a bit unsure if it's nice to have it for both or just the exports, but I'm lacking the necessary knowledge to weight in on that - so I'll leave that up to @Kongkille :)

corradio:
One nice screenshot for production (8% of power represents 40% of emissions):
image

It's also there for exports, but not that % of power will always be equal to % of emissions (because one can't decide to export only the good or bad stuff):
image

Kongkille (approved):
Super cool idea! The tooltip is already quite technical and dense, so I don't think one additional sentence makes much of a UX difference.

One reflection, if I understand correctly, from this is also that this feature makes the assumption that users are not looking that much at the emission tab, as this insight is represented there already.

I would love to get some data on how many users are clicking on the emission tab. As I think it's a somewhat underused feature and I would want to understand why (or proven wrong)

corradio:
> Super cool idea! The tooltip is already quite technical and dense, so I don't think one additional sentence makes much of a UX difference.
> 
> One reflection, if I understand correctly, from this is also that this feature makes the assumption that users are not looking that much at the emission tab, as this insight is represented there already.
> 
> I would love to get some data on how many users are clicking on the emission tab. As I think it's a somewhat underused feature and I would want to understand why (or proven wrong)

I think we have it in analytics. The event should be called `displayByEmissions`

PR #2940, opened 2021-01-19
Expose config as a Poetry package

skovhus:
This creates a new `electricitymap.contrib.config` package using [Poetry](https://python-poetry.org/). The package is exposing the `config/*.json` files through an interface identical to `utils/config.py` (e.g. `ZONE_BOUNDING_BOXES`, `ZONES_CONFIG`, `ZONE_NEIGHBOURS`).

This was done without changing the current folder structure to make the changes backward compatible. 

Note that:
- the `config` folder is now a bit convoluted (we can clean that up once everything is migrated)
- `utils/config.py` was duplicated into the `config/electricitymap` folder (eventually it will only be part of `electricitymap.contrib.config` package and removed from the `utils` folder)
- we now have a few unit tests for `electricitymap.contrib.config` :)
- a Poetry package for the parsers will be created once the config package battle-tested 

Thanks to @wizmer for the inspiration to use `pathlib` in https://github.com/tmrowco/electricitymap-contrib/pull/2907 and for pushing for improving the development setup. 

https://github.com/tmrowco/electricitymap/issues/5


corradio (commented):
Thanks for this @skovhus 🎉 
Do we need a BUILD.yaml for `config` so it builds/tests properly on our internal monorepo?
Or do we assume this is the responsibility of this repo?
I'm inclined to say the latter.

skovhus:
> Do we need a BUILD.yaml for config so it builds/tests properly on our internal monorepo? Or do we assume this is the responsibility of this repo? I'm inclined to say the latter.

Good question. I would also say that this should be guaranteed by this repository. :)

corradio (approved):

skovhus:
As this shouldn't have any effect on the workflow of the contributors, we do not need to change any documentation.

PR #2941, opened 2021-01-19
Update/Add solar capacities in Australia

Manu1400:
From source Australian PV Institute (APVI):
https://pv-map.apvi.org.au/historical => https://d3hysioybks0ok.cloudfront.net/attachments/analysis/7/states_65c6.csv

PR #2937, opened 2021-01-19
Fix hydro capacity in France

Manu1400:
Goal : Switch ENTSOE source to French opendata (national register of all installations)

Sources : (`puisMaxInstallee` = capacity installed in kW at 30 september 2020 !)
[...]

PR #2913, opened 2021-01-11
Add value for AD 🇦🇩

Manu1400:
"La production interne est globalement issue d’une centrale hydroélectrique (3x15 MW, moyenne de 85 GW.h (...)". in https://unfccc.int/sites/default/files/resource/AND_BUR1_Definitiu.pdf

PR #2952, opened 2021-01-23
Update DK capacity (at 2020-12-31) from opendata

Manu1400:
Good dataset : https://www.energidataservice.dk/tso-electricity/capacitypermunicipality


veqtrus:
@Manu1400 can you add the source to `DATA_SOURCES.md`? You should also do that in #2941, #2937, #2911, #2913.

madsnedergaard (approved):
I have added the data source :)

PR #2957, opened 2021-01-29
update Yukon timezone

jarek:
We're getting error logs in Kibana:

`invalid point: {'datetime': datetime.datetime(2021, 1, 28, 21, 30, tzinfo=tzfile('/usr/share/zoneinfo/Canada/Pacific')), 'zoneKey': 'CA-YT', 'production': {'unknown': 23.5, 'hydro': 63.42, 'coal': 0, 'nuclear': 0, 'geothermal': 0}, 'storage': {}, 'source': 'www.yukonenergy.ca', 'schemaVersion': 3}, reason:Data from CA-YT can't be in the future, data was 2021-01-28T21:30:00-08:00, now is 2021-01-29T04:37:55.615266+00:00`

This is because the parser currently specifies Pacific timezone:

https://github.com/tmrowco/electricitymap-contrib/blob/97813abdb36514955ae6eafec79a4ca7cc4aa046/parsers/CA_YT.py#L8

Pacific is currently UTC-8. However, Yukon decided to use UTC-7 year-round starting this fall, and got a new tzinfo. See https://time.is/Whitehorse or https://timezonedb.com/time-zones/America/Whitehorse


skovhus (approved):

skovhus:
@jarek thanks for fixing this!

@FelixDQ anything I should be aware of here before merging?

FelixDQ:
I don't think there's anything we have to be aware of in this case.
This parser can't parse historical data, but otherwise, I would check if this fix also worked on historical data and determine if we also have to backfill the data.

Issue #2963, opened 2021-02-03
Consider using jsonlint-mod to catch duplicate keys


veqtrus:
One such issue found in #2851.

skovhus:
@veqtrus thanks, that is a great idea!

skovhus:
I'm trying this out. 

https://github.com/tmrowco/electricitymap-contrib/pull/2862 also introduced a duplicate entry for `CA-QC` in `coeq_parameters.json`

PR #2964, opened 2021-02-03
Introduce jsonlint-mod and fix CA-QC duplicate key

skovhus:
Use [jsonlint-mod](https://github.com/circlecell/jsonlint-mod) as suggested by @veqtrus. We use it to check for duplicated keys. 

It found one issue in `co2eq_parameters.json`, which seems to be a regression introduced when bringing back fallback mixes #2862.

Ref: https://github.com/tmrowco/electricitymap-contrib/issues/2963

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
       - checkout
       - run:
           command: |
-            sudo npm install -g jsonlint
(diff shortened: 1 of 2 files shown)


corradio (approved):
Good catch!

FelixDQ (approved):

PR #2966, opened 2021-02-03
Fix url for PE parser

martin-laurent:
The old URL (withotu https) would redirect to another HTML page


skovhus:
@martin-laurent thanks for fixing this! 

I'm curious to hear how you discover the issue.

skovhus (approved):

PR #2967, opened 2021-02-03
Fix KR requests and MX signature

martin-laurent:
KR parsers was raising a HTTPCOnnectionError because of the missing certificate.
MX fetch_production signature was not expecting `logger`


skovhus:
@martin-laurent awesome!

I just noticed the MX logger issue. I was considering adding a unit tests that would verify the interface of the parsers. It could be done by running `inspect.getfullargspec(someParser.fetch_production)`.

skovhus (approved):

Issue #2882, opened 2020-12-29
Bolivia parser down


q--:
The page mentioned in DATA_SOURCES gives error 403 Forbidden, so perhaps the API URLs have moved.

[A (new?) page with generation breakdown is here](https://www.cndc.bo/gene/index.php), it appears to have the data the parser used. Haven't investigated if it's in the same format.

---
[Kibana](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now%2Fw,mode:quick,to:now%2Fw))&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:extra.key,negate:!f,params:(query:BO,type:phrase),type:phrase,value:BO),query:(match:(extra.key:(query:BO,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))

```
Traceback (most recent call last):
  File "/home/src/electricitymap/feeder/lib/fetch_data.py", line 178, in launch_parsers
    objs = res.get(timeout) or []
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/src/electricitymap/feeder/lib/fetch_data.py", line 176, in <lambda>
    **parser_kwargs))
  File "/home/src/electricitymap/contrib/parsers/BO.py", line 204, in fetch_generation_forecast
    obj = webparser(response)
  File "/home/src/electricitymap/contrib/parsers/BO.py", line 43, in webparser
    obj.hour = hours
  File "/usr/local/lib/python3.6/site-packages/pandas/core/generic.py", line 4405, in __setattr__
    self[name] = value
[...]

PR #2969, opened 2021-02-03
fix(parsers): bolivia now uses json

byronwilliams:
@q-- I've updated the Bolivia parser to use JSON as per your comment in #2882


skovhus:
@q-- let me know if you have some feedback here, as I see that you already worked a bit on [fixing this issue](https://github.com/q--/electricitymap-contrib/tree/Fix_BO_parser).

skovhus:
@byronwilliams thanks for fixing this.

I tried to validate the parser here by running `test-parser.py BO`. It does fail with a `json.decoder.JSONDecodeError`. I tried inspecting the `resp.text` and see that it contains some odd characters:

```
>>> resp = r.get(url_init.format(formatted_date))
>>> resp.text
'ï»¿{"data":[]}'
```

Kongkille:
@byronwilliams are you still investigating this? :)

corradio (approved):
I took the liberty of finalising some changes.

PR #2973, opened 2021-02-04
Switzerland CO2eq parameters

KabelWlan:
Hi everyone
I'm relativley new here and I don't have much experience but I really want to support this. If I did something wrong, I'm sorry, I'm always happy to get feedback.

I added Switzerland since the CO2 intensity isn't calculated and exports are shown with a intensity of 456gCO2eq/KWh which is way too high for Switzerland. I got the data from the Bundesamt für Statistik (https://www.bfe.admin.ch/bfe/de/home/versorgung/statistik-und-geodaten/energiestatistiken/gesamtenergiestatistik.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvNzUxOQ==.html) from the sheet T24.
Is it fine or did I forget something?


corradio:
Hi @KabelWlan !
Thanks for the PR. This makes a lot of sense as it will make sure to use a default CO2e intensity for CH when the data quality is not sufficient to be shown on the map.

corradio (approved):

PR #2975, opened 2021-02-06
Update Ukrainian exchange capacities for bidirectional transfer

veqtrus:
Ukraine both exports and imports electricity.


ajoga:
Can you back up your claim that flows are bidirectional ? #2071 introduced these flows and referenced https://ua.energy/transmission-and-dispatching/dispatch-information/transborder-flows/ as source. 

It still states : 
> Ukrenergo transmits electricity via interstate power lines in the **following directions**:

then lists the interconnects. The one from Burshtynska TPP Island to Hungary, Slovakia and Romania of 650MW is explicitly stated as Directional.

veqtrus:
@ajoga https://www.electricitymap.org/zone/UA shows transfers in both directions all the time.

corradio (approved):

PR #2976, opened 2021-02-06
Fix Japan consumptionForecast parsers

martin-laurent:
I noticed the urls used for the consumption forecast Japan parsers were not updated which led to many 404.

Also made the exchange parsers faster with only 1 day's worth of data (1 query instead of 2) which reduces the time from ~11 secs to 5 secs.
I noticed the exchanges/imports are used in the `fetch_consumption` logic for the japan parsers. I have not seen that pattern anywhere else in parsers. 

The Japanese parsers fetches the consumption then substracts the exchanges. Is electricity-map backend smart enough to do that computation if we were to only provide fetch_consumption and fetch_exchange?

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -20,17 +20,18 @@ from parsers import occtonet
 # JP-CG  : Chūgoku
 
 sources = {
-        'JP-HKD':'denkiyoho.hepco.co.jp',
-        'JP-TH':'setsuden.nw.tohoku-epco.co.jp',
-        'JP-TK':'www.tepco.co.jp',
-        'JP-CB':'denki-yoho.chuden.jp',
-        'JP-HR':'www.rikuden.co.jp/denki-yoho',
-        'JP-KN':'www.kepco.co.jp',
-        'JP-SK':'www.yonden.co.jp',
-        'JP-CG':'www.energia.co.jp',
-        'JP-KY':'www.kyuden.co.jp/power_usages/pc.html',
-        'JP-ON':'www.okiden.co.jp/denki/'
+        'JP-HKD': 'denkiyoho.hepco.co.jp',
+        'JP-TH': 'setsuden.nw.tohoku-epco.co.jp',
+        'JP-TK': 'www.tepco.co.jp',
+        'JP-CB': 'denki-yoho.chuden.jp',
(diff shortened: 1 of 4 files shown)


parsers/JP.py
```diff
@@ -70,9 +70,6 @@ def fetch_production_df(zone_key='JP-TK', session=None, target_datetime=None,
     Calculates production from consumption and imports for a given area
     All production is mapped to unknown
     """
-    if target_datetime:
```
corradio, line 73: Is the parser now able to parse historical data?

parsers/JP.py
```diff
@@ -99,51 +96,52 @@ def fetch_production_df(zone_key='JP-TK', session=None, target_datetime=None,
...
-    if zone_key in ['JP-CB', 'JP-CG', 'JP-TK', 'JP-HKD', 'JP-HR', 'JP-KN', 'JP-SK', 'JP-KY']:
-        df.columns = ['Date', 'Time', 'cons', 'solar']
-    elif zone_key in ['JP-TH']:
+        startrow = 54
+
+    try:
+        df = pd.read_csv(consumption_url[zone_key], skiprows=startrow,
+                         encoding='shift-jis')
+    except pd.errors.EmptyDataError as e:
+        logger.error("Data not available yet")
```
corradio, line 138:
```suggestion
        logger.exception("Data not available yet")
```

corradio: this will capture the exception as well

corradio:
> The Japanese parsers fetches the consumption then substracts the exchanges. Is electricity-map backend smart enough to do that computation if we were to only provide fetch_consumption and fetch_exchange?

Hello @martin-laurent,
Unfortunately not. We use `fetch_production` to assess the production *mix*. `fetch_consumption` only fetches the *total* consumption (and not the mix).

parsers/JP.py
martin-laurent: I have just checked that they can parse up to 2020-05-01 except the Hokkaido one which is excluded now.

parsers/JP.py
martin-laurent: addressed

corradio (approved):

corradio:
Thanks for the help here!

PR #2867, opened 2020-12-24
Genezalize the kings island parser so it can parse all of adjenti's islands

ajoga:
This commit supports flinders island (#2533) and rottnest island (#2534), AUS (bounding_boxes are missing though, I'm gonna try to understand #2857 to see if I need to add something in this PR about that. inputs welcome in the meantime)

Issue #2534, opened 2020-05-31
Data for Rottnest Island (Western Australia)


q--:
After another energy dashboard was found in  #2533, I decided to have a [look](https://www.hydro.com.au/clean-energy/hybrid-energy-solutions/success-stories/) at Hydro Tasmania's website to see what else I could find.

Turns out there's also a [dashboard](https://www.hydro.com.au/clean-energy/hybrid-energy-solutions/success-stories/rottnest-island) for Rottnest island, [off the coast of Western Australia](https://www.google.nl/maps/place/Rottnest+Island+WA+6161,+Australia)

Judging by the browser console logs, this one [also](https://github.com/tmrowco/electricitymap-contrib/pull/2456/files#diff-c3b7bb94e70c63d689a577143fcab2a9R12) uses Ajenti and a SignalR WebSocket connection under the hood. It will likely be possible to make this work with the parser from #2456

To Do

- [ ] Add Rottnest island to the map (I don't know how to do this. It's off the coast of Perth and not currently on the map)
- [ ] See if we can adapt the King Island parser to also work for this island

q--:
Changing "taghub" to "hogshub" (with different capitalization?)  should do

Found in a request setting up the connection:

    GET | https://data.ajenti.com.au/live/signalr/negotiate?clientProtocol=1.5&connectionData=[{"name":"hogshub"}]&_=1603470672758

madsnedergaard:
Closing this as it's been done in #2977 and #2867

PR #2977, opened 2021-02-07
Add Geodata for Rottnest Island (Australia)

ajoga:
This enable that island to have its own data collected and published online. The PR for the parser is dependant on this PR ([see here](https://github.com/tmrowco/electricitymap-contrib/pull/2867#issuecomment-771591341)). Both of these PR will solve #2534 .

I took inspiration from #2533 

To core reviewer, please see my inline comment about one change I do not fully understand that I had to make.


web/generate-geometries.js
```diff
@@ -947,6 +949,6 @@ topoMoreDetails = topojson.simplify(topoMoreDetails, 0.001);
 // Merge topoMoreDetails into topo
 mergeTopoJsonSingleZone(topo, topoMoreDetails);
 
-topo = topojson.filter(topo, topojson.filterWeight(topo, 0.009));
+//topo = topojson.filter(topo, topojson.filterWeight(topo, 0.009));
```
ajoga, line 952: The island is very small (19km2).

With the line commented, I get the island to render in the browser, with `"AUS-WA-RI":{"type":"MultiPolygon","arcs":[[[9507]]]}` in `/web/src/world.json` (size=641K)

With the line uncommented, the island does not appear in the browser and gets referred to as `"AUS-WA-RI":{"type":null}` in `/web/src/world.json` (size=547K)

I guess this code serves some purpose to reduce the size of the map to be made available to the browser, but I don't know what to do with it so this island can be rendered. You'll note that `moreDetails: true` doesn't help.

madsnedergaard:
Hey @ajoga, thanks for the work here!

You're right that it's there to reduce the file-size, but it should only apply to zones that don't have the `moreDetails` flag :)

I have a fix for this in #2987, and using that the geometries for Rottnest Island are generated:

Screenshot 2021-02-09 at 16 49 33

It's not the most pretty shape, but to fair it is a really small island! 😄 

Once the other PR is un-merged, you should be able to revert the uncommenting in `generate-geometries` and we can get this merged :)

web/generate-geometries.js
ajoga: addressed by @madsnedergaard in #2987 :)

madsnedergaard (approved):
Awesome! 💪

PR #2979, opened 2021-02-07
Ignore ENTSOE self-consumption values with no corresponding consumption

### LINKED ISSUE #2710: ENTSOE fetch_consumption must include consumption by production units

willbeaufoy:
We were getting errors where self-consumption values were found that had no
corresponding consumption value, during fetch_consumption() for ENTSOE:

https://github.com/tmrowco/electricitymap-contrib/pull/2814#issuecomment-773376731

This was not expected when the original code was written to fix #2710 so there was no
handling for it, hence the errors.

As self-consumption values are not useful for a given datetime unless there is an existing
consumption value to add them to, it makes most sense just to ignore these values. I'm not
sure exactly why we are getting these values, but probably at certain points ENTSOE is reporting
production values (from which self-consumption is derived) for datetimes that do not yet have
consumption data available.

Note, I can't remember why I originally ignored values for the first datetime in the list,
but it seems to make no sense, so I also removed this code.


Kongkille (approved):
LGTM 👍

PR #2981, opened 2021-02-08
Updates text-color of the carbon intensity "square" based on contrast to background

### LINKED ISSUE #1886: Use better text color for CO2 color rectangle
When on yellow or light green, white text for CO2 value is hard to see.

Maybee you could use W3CAG for better readeability.

Here is a codepen that show how to implement it : https://codepen.io/phiphou/pen/YBeMxw

madsnedergaard:
## Issues

Fixes #1886 (thanks for the code and suggestions!)

Might also partially help with #2304

### Description

The text is now changing color based on the background-color - so if it's light, we change text color to black.

While at it, I also noticed that `<TooltipContent />` was being re-rendered on every mouse-move event, which makes no sense if it's within the same zone. The component has been memoized here :)

### Preview

Kapture 2021-02-08 at 13 52 41

This has been fixed now: 👇 
Screenshot 2021-02-08 at 14 02 57

### DIFF
--- a/web/src/components/carbonintensitysquare.js
+++ b/web/src/components/carbonintensitysquare.js
@@ -18,9 +18,8 @@ const getTextColor = (rgbColor) => {
   return contrastRatio > 128 ? 'black' : 'white';
 };
 
-
 const Value = styled.span`
-font-weight: bold;
+  font-weight: bold;
 `;
 
 const Box = styled.div`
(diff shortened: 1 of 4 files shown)


corradio:
Good catch @madsnedergaard !! This makes me so happy.
@skovhus can I let you review the frontend code?

skovhus (approved):
This is super nice! 👏  

And I'll happily review the frontend PRs.

PR #2982, opened 2021-02-08
Added autosquash

FelixDQ:
Let's see if this works now with the `pull_request_target` trigger.


skovhus (approved):
Nitpick: would make sense to call the file automerge.  :)

PR #2983, opened 2021-02-08
Document how to generate bounding boxes for zones

skovhus:
- Document how to generate bounding boxes for zones using `generate-zone-bounding-boxes.js`.
- Update `generate-geometries.js` to ensure `public/dist` folder exists

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -34,3 +34,10 @@ TopoJSON format, which is a more compressed format than geoJSON. It only stores 
 on a grid. All together, this allows to convert a ~`24MB` file to a ~`1MB` one.
 
 The final file is named `world.json` and is the one sent to the client.
+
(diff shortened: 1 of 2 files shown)


FelixDQ (approved):

FelixDQ:
Ok, one limitation of the new automerge label. It won't trigger when approving the pr. The triggers are
```
      - labeled
      - unlabeled
      - synchronize
      - opened
      - edited
      - ready_for_review
      - reopened
      - unlocked
```
So the workflow would be adding the label after approving the PR. The [pull_request_review](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review) trigger won't work because it runs on the fork and the secrets aren't available on forks.

skovhus:
Ah, good to know. 👍

PR #2986, opened 2021-02-09
Adds geodata for Stewart Island (New Zealand)

### LINKED ISSUE #2741: New Zealand: Stewart Island is incorrectly shown as part of the South Island electricity network
The [electricityMap](https://www.electricitymap.org/zone/NZ-NZS) incorrectly shows New Zealand's Stewart Island as being part of the South Island electricity distribution network.
image

However, there is no electricity distribution link between the two islands, and Stewart Island has it's own independent diesel generator setup to provide electricity to the community on the island.

A description of Stewart Island's electricity infrastructure can be found at: https://www.southlanddc.govt.nz/my-southland/siesa-2/what-sieasa-does/

madsnedergaard:
## Issue

Fixes #2741

## Description

Separate Stewart Island from the New Zealand South Island zone.

### Progress 

Here's what I have so far:

- I found the coordinates for South Island (New Zealand) in the generated `tmp_countries.json` file
- I took the coordinates to [geojson.io/](https://geojson.io/) and
	- Found the specific coordinates for Stewart Island and added to a new `NZ-NZST.geojson`
	- Took the remaining coordinates and added to `NZ-NZS.geojson` (with a subid of `"NZS-without-nzst"` to ensure we don't use the default)
- Updated `generate-geometries.js` to take in the two new files
- Updated `zones.json` with new zone + updated boundaries

### Preview

Screenshot 2021-02-09 at 16 00 47


madsnedergaard:
Just for fun I also added capacities (https://www.southlanddc.govt.nz/my-southland/siesa-2/what-sieasa-does/), but since it isn't showing anything it's not really too critical ;)

web/Dockerfile
```diff
@@ -9,7 +9,6 @@ RUN yarn
 
 # Generate map
 RUN mkdir -p public/dist/
-ADD web/build ./build
```
skovhus, line 12: I'm curious why this was here 🧐

web/generate-geometries.js
```diff
@@ -531,13 +533,14 @@ const zoneDefinitions = [
   { zoneName: 'NO-NO4', type: 'subZone', id: 'NO-NO4' },
   { zoneName: 'NO-NO5', type: 'subZone', id: 'NO-NO5' },
   { zoneName: 'NP', type: 'country', id: 'NPL'},
-  // { zoneName: 'NZ', type: 'country', id: 'NZL'},
   { zoneName: 'NU', type: 'country', id: 'NIU'},
   { zoneName: 'NR', type: 'country', id: 'NRU'},
+  // { zoneName: 'NZ', type: 'country', id: 'NZL'},
```
skovhus, line 538: Any reason for keeping this? Can it be useful in the future?

skovhus (approved):
Looks good! Thanks for documenting your steps here. Would be great if you can check if the wiki should be updated. :)

web/generate-geometries.js
madsnedergaard: As discussed, this is what we've generally done for all other zones to indicate that we're using subzones instead of the country :)

web/Dockerfile
madsnedergaard: Probably left-over, it works without it :)

madsnedergaard:
> Looks good! Thanks for documenting your steps here. Would be great if you can check if the wiki should be updated. :)

There's nothing in the wiki about this, only the README. I'm not sure if it's worth adding it to the wiki as well. I found it hard to generalise the steps and learnings I had in this PR, since it really depends on what data is available - and I'm not sure how often we need to perform an exercise like this. 

Will see if it occurs again and if there's generalised learnings coming up later :)

PR #2987, opened 2021-02-09
Make it possible to generate geometries for smaller zones

madsnedergaard:
## Description

As discovered by @ajoga in #2977, we currently can't generate geometries for very small areas.

This PR ensures that we only applies the `topojson.filter` function for areas without `moreDetails: true`.

The increase to file size of `world.json` is extremely small:

```
Before: 560.009 bytes (561 KB on disk)
After:  560.077 bytes (561 KB on disk)
```


madsnedergaard:
CC/FYI @corradio and @FelixDQ in case I have missed something 🙃

corradio (approved):
Well done!

skovhus (approved):
👍

PR #2993, opened 2021-02-10
update DK-1 and DK-2 capacities

KabelWlan:
I updated the capacities from DK-1 and DK-2, source is Entsoe.


madsnedergaard (approved):
Looks good to me! Yay for more solar and less coal 🌞 👏

PR #2994, opened 2021-02-11
Update sv.json

magol:
Added missing strings from #2986, #2977, #2933, #2925 to Swedish localization.


skovhus (approved):
Tack så mycket!

Issue #2950, opened 2021-01-21
No calculation of CO2 intensity for Switzerland


KabelWlan:
Last week the CO2 intensity of Switzerland was calculated and showed on the map after the changes were merged. A few days ago it stopped showing and calculating the CO2 intensity despite the data being available. 

Does anyone have an idea for what reason it won’t work anymore?

corradio:
Hello @KabelWlan ,
Indeed we have disabled it temporarily as we have experts reviewing the data quality. We'll let you know once it's back online!

PaulCornelissen:
I'm very interested in following that. Is this review public somewhere?

But regarding data quality in general:
I must add that I find it inconsistent that #2973 gets approved, while this obviously introduces a huge error because (it neglects imports and that makes up >50% of electricity at times), while #2898 that has fairly thorough analysis behind it, but nonetheless calculations gets disabled. Enabling the calculations for Switzerland now would significantly reduce the error, compared to the static CO2 intensity of 29gr/CO2 that's currently reported.

In general I find it unclear what the position of electricitymap is regarding accuracy. E.g. why is Switzerland under review while the Netherlands is not? The Netherlands is now much more accurate now than it has been for years, but it still has obvious flaws. Yet Switzerland gets disabled for review? I believe the Swiss CO2 intensity is more reliable than the Dutch. Heck, probably even more reliable than the German or Polish estimate, because those hugely underestimate brown coal emissions.

KabelWlan:
@PaulCornelissen
You're obviously right that I made a mistake with the power origin ratios. I'm sorry for that. I noticed it when I first saw the 29gr on the map which is clearly too low. I did some research the last few days and I'll creat a pull request later with the power origin ratios which include the imports. It should be more accurate then but I guess #2898 would still do a better job.

madsnedergaard:
Hey @PaulCornelissen, thanks for pointing this out - data quality is of course very important to us!

Regarding the parser for Switzerland: 
We're working with the Swiss energy ministry on this, and they are doing an internal review of it currently. 
It was not supposed to go live before they were ready, but the PR was accidentally merged too early - so that's why we disabled it again temporarily. 

Hope that answers some of your concerns :)

corradio:
FYI we're still awaiting an answer. Hopefully this will be solved soon.

PR #2996, opened 2021-02-15
Switzerland powerOriginRatios

KabelWlan:
As in #2950 mentioned I added the import data to the Swiss powerOriginRatios which I forgot in the beginning... I'm sorry. Sources didn't change. It should be more accurate now.


madsnedergaard (approved):
No worries, and thanks for this update :)

PR #2997, opened 2021-02-16
Fix King Island country name

chuqtas:
Fix King Island country name


chuqtas:
Extremely simple merge, review welcomed :)

madsnedergaard (approved):
Ah good point, Tasmania isn't an independent country yet ;)

PR #3008, opened 2021-02-23
Adds Danish translations

madsnedergaard:
## Description

We're missing a bunch of Danish translations - and given most of the team is native Danish speakers, I guess we should be able to do those translations :)﻿


skovhus (approved):
🇩🇰!

PR #3010, opened 2021-02-24
Update CY parser to return dynamic capacities

veqtrus:
I noticed that the source page has a hidden (by default) table which also includes the available oil capacity at any time.

The source URL changed which broke fetching historical data.


veqtrus:
BTW other parsers tend to use `html.parser` instead of the faster `lxml` as the `BeautifulSoup` backend.

skovhus:
Thanks for fixing this! I've verified that it works using `test-parser.py CY`. 👏

skovhus (approved):

PR #3013, opened 2021-02-27
CY: Return static capacities if dynamic unavailable

veqtrus:
Ideally the backend would do that automatically when the dynamic capacities are not returned, or perhaps even extrapolate from the last available.

### DIFF
--- a/parsers/CY.py
+++ b/parsers/CY.py
@@ -17,16 +17,12 @@ class CyprusParser:
         'Εγκατεστημένη Ισχύς Βιομάζας': 'biomass'
     }
     STATIC_CAPACITIES = {
-        'biomass': 12.4,
         'coal': 0,
         'gas': 0,
         'geothermal': 0,
         'hydro': 0,
         'hydro storage': 0,
-        'nuclear': 0,
-        'oil': 1478,
-        'solar': 229.3,
-        'wind': 157.5
+        'nuclear': 0
     }
 
     session = None
--- a/parsers/CY.py
+++ b/parsers/CY.py
@@ -16,6 +16,18 @@ class CyprusParser:
         'Φωτοβολταϊκή Εγκατεστημένη Ισχύς': 'solar',
         'Εγκατεστημένη Ισχύς Βιομάζας': 'biomass'
     }
+    STATIC_CAPACITIES = {
+        'biomass': 12.4,
+        'coal': 0,
+        'gas': 0,
+        'geothermal': 0,
+        'hydro': 0,
+        'hydro storage': 0,
+        'nuclear': 0,
+        'oil': 1478,
+        'solar': 229.3,
+        'wind': 157.5
+    }
 
     session = None
     logger: logging.Logger = None
@@ -36,10 +48,11 @@ class CyprusParser:
             values = [td.string for td in row.find_all('td')]
             if None in values or '' in values:
                 break
+            production = {}
             datum = {
                 'zoneKey': 'CY',
-                'production': {},
-                'capacity': {},
+                'production': production,
+                'capacity': dict(self.STATIC_CAPACITIES),
                 'storage': {},
                 'source': 'tsoc.org.cy'
             }
@@ -49,9 +62,9 @@ class CyprusParser:
                 elif col == 'Συνολική Διαθέσιμη Συμβατική Ικανότητα Παραγωγής':
                     datum['capacity']['oil'] = float(val)
                 elif col == 'Αιολική Παραγωγή':
-                    datum['production']['wind'] = float(val)
+                    production['wind'] = float(val)
                 elif col == 'Συμβατική Παραγωγή':
-                    datum['production']['oil'] = float(val)
+                    production['oil'] = float(val)
                 elif col == 'Εκτίμηση Διεσπαρμένης Παραγωγής (Φωτοβολταϊκά και Βιομάζα)':
                     # Because solar is explicitly listed as "Solar PV" (so no thermal with energy storage) and there
                     # is zero sunlight in the middle of the night (https://www.timeanddate.com/sun/cyprus/nicosia),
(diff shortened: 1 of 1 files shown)


veqtrus:
Just to clarify: What I'm reporting as oil capacity is what is currently available, not the total installed capacity. Is that what is expected? I figured that would be the case based on the fluctuation of the exchange capacities in https://www.electricitymap.org/zone/AUS-VIC.

parsers/CY.py
```diff
@@ -16,6 +16,18 @@ class CyprusParser:
         'Φωτοβολταϊκή Εγκατεστημένη Ισχύς': 'solar',
         'Εγκατεστημένη Ισχύς Βιομάζας': 'biomass'
     }
+    STATIC_CAPACITIES = {
```
gwpicard, line 19: I'm not sure if `STATIC_CAPACITIES` are necessary, and if they are it might be better to default to capacity values in `zones.json` as opposed to from the parser.

@FelixDQ do you know what static values are used as fallbacks when capacities are pulled dynamically from a parser?

veqtrus: @gwpicard they currently aren't https://www.electricitymap.org/zone/CY. Alternatively I could only set the ones that are zero.

corradio: Hello,
In theory the static capacities from zones.json should be automatically used as a fallback (so they are not needed in the parser).
If this fallback mechanism doesn't work, then I suggest opening an issue so we can look into it!

gwpicard: @corradio If that is the case, then there shouldn't be a need to have `STATIC_CAPACITIES` included in the parser. 

From what I'm seeing at [https://www.electricitymap.org/zone/CY](https://www.electricitymap.org/zone/CY), the values for biomass, wind and solar are the same as what is specified in `zones.json` even though they aren't currently specified in the parser. The '?' are because we don't currently have any capacity values for the other production modes.

So it looks as though the `zones.json` values are used as a fallback unless I'm missing something @veqtrus?

veqtrus: The values for biomass, wind, solar, and oil don't have '?' as they are returned by the parser. What the parser doesn't currently return is the zero capacities, as the parser looks for which capacities are mentioned in the source. Therefore, no, the backend does not correctly fallback to zones.json, as that file has zero values for other production types.

gwpicard: I see. Perhaps the best way to do it so that it is simpler would be to have a way for the backend to fall back to values in `zones.json` even in the case where the parser only returns some capacity values.

Essentially, we only use dynamic capacities if we are able to return all of them from the parser, and if not, we fall back to `zones.json`. How does that sound? :)

veqtrus: @gwpicard I may have understood you wrong, but if you fall back to zones.json if some values are missing from the parser output, the parser will never be used as it doesn't return the zero capacities. What should happen is that for those values that are missing from the parser output, values returned recently or the ones from zones.json should be used.

gwpicard: >What should happen is that for those values that are missing from the parser output, values returned recently or the ones from zones.json should be used

Yes, this is also an acceptable approach! In essence, I think whatever would avoid having capacities stored in the parser (like `STATIC_CAPACITIES`) is preferable. And if we can do that using your above approach that would be great.

madsnedergaard:
Hey @gwpicard, would you mind driving this one forward seeing as you have the context? :)

gwpicard:
>What should happen is that for those values that are missing from the parser output, values returned recently or the ones from zones.json should be used.

@veqtrus were you able to work on a fix for this suggestion or open an issue for it?

veqtrus:
> @veqtrus were you able to work on a fix for this suggestion or open an issue for it?

@gwpicard the backend is not open-source AFAIK, so I can't fix it there. I changed `STATIC_CAPACITIES` to only have the zero ones.

gwpicard:
> @gwpicard the backend is not open-source AFAIK, so I can't fix it there. I changed STATIC_CAPACITIES to only have the zero ones.

@veqtrus good point... @skovhus @FelixDQ any ideas on how best to proceed? Basically, CY has a parser that returns dynamic capacities and we're trying to find the best way to deal with when the parser has missing capacity values, either by defaulting to the most recent values or using the values in `zones.json`. This may be something for other zones that have dynamic capacities as well.

corradio:
Hello there!
This should be documented better, but parsers should never return static capacities (even if they are 0).
I do no therefore believe we need this PR.
However, we currently have a bug where we are not able to "merge" the dynamic capacities returned by a parser with the static capacities given in the configuration file, hence the behaviour you're experiencing.
I have therefore opened an issue on our backend.

Let me know if you agree with this.

veqtrus:
> However, we currently have a bug where we are not able to "merge" the dynamic capacities returned by a parser with the static capacities given in the configuration file, hence the behaviour you're experiencing.
> I have therefore opened an issue on our backend.
> Let me know if you agree with this.

I agree, this PR was meant to be a workaround until the backend is fixed.

> Just to clarify: What I'm reporting as oil capacity is what is currently available, not the total installed capacity. Is that what is expected? I figured that would be the case based on the fluctuation of the exchange capacities in https://www.electricitymap.org/zone/AUS-VIC.

Could you confirm this?

corradio:
> > Just to clarify: What I'm reporting as oil capacity is what is currently available, not the total installed capacity. Is that what is expected? I figured that would be the case based on the fluctuation of the exchange capacities in https://www.electricitymap.org/zone/AUS-VIC.
> 
> Could you confirm this?

We will need to provide some clearer guidelines as this isn't totally clear on electricityMap.
The idea is to provide installed capacity (that's how the UI is showing it and it mentions the term "installed capacity").
Do you have more insights into what is the number you're reporting if it's not installed capacity? Is it the installed capacity with outages subtracted?

veqtrus:
> Do you have more insights into what is the number you're reporting if it's not installed capacity? Is it the installed capacity with outages subtracted?

The installed capacity is also available if that's what you want. However in the case of Cyprus it is much bigger than the peak load, so some units may be offline. The demand will then be met by varying the output of online units. The maximum output of online units is what I'm reporting.

corradio:
> > Do you have more insights into what is the number you're reporting if it's not installed capacity? Is it the installed capacity with outages subtracted?
> 
> The installed capacity is also available if that's what you want. However in the case of Cyprus it is much bigger than the peak load, so some units may be offline. The demand will then be met by varying the output of online units. The maximum output of online units is what I'm reporting.

Got it. Thanks for clarifying. Then it's better to report installed capacity in order to stay consistent.

veqtrus:
Since an issue has been opened for the backend I'm closing this PR.

PR #3016, opened 2021-03-02
Bumps Pillow version to fix issues with Homebrew

madsnedergaard:
## Description

Currently there's a problem when running `poetry install` and installing `Pillow`. See details in https://github.com/python-pillow/Pillow/issues/3438.

In v8 of Pillow this was fixed ([in this PR](https://github.com/python-pillow/Pillow/pull/4842)), so bumping it makes it work out of the box for me on MacOS Big Sur.

I checked [the release notes](https://pillow.readthedocs.io/en/stable/releasenotes/8.0.0.html) and didn't find anything that would be breaking, but would be nice with a second set of eyes on this :)

---

Using Pillow requires having `zlib` and `libjpeg` installed on your machine, which isn't documented anywhere. 

So I also propose the following update to the wiki:

Screenshot 2021-03-02 at 09 03 30

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -322,11 +322,11 @@ python-versions = ">=2.6"
 
 [[package]]
 name = "pillow"
-version = "7.1.0"
+version = "8.1.0"
 description = "Python Imaging Library (Fork)"
 category = "main"
 optional = true
-python-versions = ">=3.5"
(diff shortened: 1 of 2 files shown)


skovhus (approved):
Thanks for fixing this! The wiki change looks good. 👏

PR #3025, opened 2021-03-04
Updates editor settings

madsnedergaard:
## Description

- Adds eol setting to `.editorconfig` so IDEs know which style to use
- Updates `.gitattributes` so git knows which style to use (so it doesn't clone with CRLF)
- Adds `prettierrc.js` so editors with Prettier can pick it up

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,6 +2,7 @@ root = true
 
 [*]
 indent_style = space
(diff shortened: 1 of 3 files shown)


madsnedergaard:
cc @corradio

skovhus (approved):

PR #3026, opened 2021-03-05
Various bugfixes

corradio:
Various bugfixes discovered with @Kongkille 
See PR comments.


web/server.js
```diff
@@ -137,7 +137,7 @@ app.use('/', (req, res) => {
   const isTmrowCo = req.get('host').indexOf('electricitymap.tmrow') !== -1;
   const isNonWWW = req.get('host') === 'electricitymap.org'
     || req.get('host') === 'live.electricitymap.org';
-  const isStaging = req.get('host') === 'staging.electricitymap.org';
+  const isStaging = req.get('host').includes('staging');
```
corradio, line 140: Required to make sure our staging environment is better detected (when e.g. called by cloud run)

web/src/components/countrytable.js
```diff
@@ -329,10 +329,12 @@ const CountryElectricityProductionTable = React.memo(({
 
   // Use the whole history to determine the min/max,
   // fallback on current data
-  const history = useCurrentZoneHistory();
+  const history = useCurrentZoneHistory() || [];
```
corradio, line 332: Required as `useCurrentZoneHistory` currently falls back to returning `[]`.
However, we don't want to have this function assume that it always fallsback to `[]`, so we added a redundant fullback here to always make sure it returns `[]` in case of missing data.

web/src/components/countrytable.js
```diff
@@ -329,10 +329,12 @@ const CountryElectricityProductionTable = React.memo(({
...
   const [minPower, maxPower] = useMemo(
     () => {
-      const historyOrCurrent = history || [data];
+      const historyOrCurrent = history.length
```
corradio, line 335: Required as `useCurrentZoneHistory` currently falls back to returning `[]`.

web/src/components/loadingoverlay.js
```diff
@@ -21,6 +21,6 @@ const Overlay = styled.div`
 
 export default ({ fadeTimeout = 500, visible }) => (
   <CSSTransition in={visible} timeout={fadeTimeout} classNames="fade">
-    <Overlay fadeTimeout={fadeTimeout} />
+    <Overlay id="loading" fadeTimeout={fadeTimeout} />
```
corradio, line 24: Added an `id` so we can refer to id when making the timelapse.

web/src/globalstyle.js
```diff
@@ -6,7 +6,7 @@ const FadeAnimationDefinition = css`
   .fade-exit { opacity: 1; }
   .fade-exit-active { opacity: 0; }
   /* Don't set display to 'none' to keep the same width and height */
-  .fade-exit-done { opacity: 0; pointer-events: none; }
+  .fade-exit-done { opacity: 0; pointer-events: none; visibility: hidden; }
```
corradio, line 9: Make sure the element is hidden so we can use `await page.waitForSelector('#loading', { hidden: true });` in puppeteer

madsnedergaard (approved):

PR #3027, opened 2021-03-06
Improved Spanish translations

CeleminHero:


web/locales/es.json
```diff
@@ -104,28 +104,28 @@
...
-    "hydro": "hidráulica",
+    "wind": "energía eólica",
+    "solar": "energía solar",
+    "hydro": "energía hidráulica",
     "hydro storage": "almac. hidrául.",
     "battery storage": "almac. en baterías",
     "biomass": "biomasa",
     "nuclear": "nuclear",
-    "geothermal": "geotérmica",
+    "geothermal": "geotermia",
```
madsnedergaard, line 114: Would you mind explaining this one? I've seen both terms used a lot, but haven't been able to tell the difference :)

web/locales/es.json
```diff
@@ -104,28 +104,28 @@
         "legend": "Leyenda",
         "faq": "Preguntas frecuentes"
     },
-    "wind": "eólica",
-    "solar": "solar",
-    "hydro": "hidráulica",
+    "wind": "energía eólica",
+    "solar": "energía solar",
+    "hydro": "energía hidráulica",
```
madsnedergaard, line 109: Not sure if we want the full title here, we also haven't done this for any other languages :)

CeleminHero: Yeah, I should have explained this :P

I added "energía" because it goes better with "X% of electricity is produced by...", which I took from the French translation because it was the formula that sounded best in Spanish compared to the Italian and Portuguese translations of the line. In French, though, they didn't add "energy" to the energy sources, and I don't know how it sounds in French, but in Spanish it sounds wrong to say ""X% of electricity is produced by solar". Your mind just finds something missing there.

That's why I added "energía", which I considered shortening to "en." but in the end decided against it (I counted the letters and saw it was no longer than battery storage ("almac. en baterías"), for example. But it could perfectly be shortened, if that's a problem :)

web/locales/es.json
CeleminHero: Yeah, they're pretty much synonyms. I would say that "geotermia" is used more often in the context of geothermal heat pumps, but it also applies to geothermal energy for electricity production. I made the change here so that I didn't have to add "energy" to "geotérmica", because "X% of electricity is produced by geotermia" sounds good on its own. Hope I explained myself :P

web/locales/es.json
madsnedergaard: Makes good sense for the tooltips, I'm mostly just wondering about the labels here:

Screenshot 2021-03-08 at 13 17 14

Wouldn't you have to also prefix nuclear, etc. with `energía` in that case?

What's your opinion on this, @Kongkille? :)

web/locales/es.json
madsnedergaard: Thanks for the explanation, I'm happy to go with your suggestion in that case :)

web/locales/es.json
CeleminHero: Yup, seems like I missed nuclear. The rest would be good as is.

veqtrus: Would it make sense to add 'energía' to the tooltip text instead?

Kongkille: I would also prefer it without energia, but of course, I don't have the same context as @CeleminHero. I like @veqtrus idea of potentially putting it in the tooltip? wdyt?

CeleminHero: Hm... nope, I don't think it works if we put it in the tooltip.

Portuguese has it in the tooltip, but I would say it doesn't make too much sense in Portuguese either (energia goes well with solar, wind, hydro, geothermal and nuclear, which in Spanish and Portuguese are adjectives, but not with nouns such as coal or battery storage). That's why, to avoid having to say "geothermal energy", I chose a noun (geotermia) instead of the adjective (geotérmica).

I think the way I did it is the one that sounds best, but if there are problems I'd say either shorten energia to "en.", or go back to the previous translation, which was pretty understandable too.

veqtrus: In Greek I also used adjectives but it's not a problem in the tooltip. The text is "**X** % of electrical energy available in **Y**  comes from **Z**". If **Z** is an adjective it goes with the noun 'energy', which is feminine in Greek. 'Source' is also feminine so it goes nicely together.

CeleminHero: Hmm yeah that's a clever solution. I'm still not too sure what the problem with "energía" is, but changing "electricidad" into "energía eléctrica" in the tooltip would certainly make it sound better.

Kongkille: I think we are mainly trying to optimize so the labels aren't too large horizontally. If there's a way you think we can avoid having energia in label while still making it understandable, I think we should go for that.

Kongkille (approved):
I'll approve, we also have other long labels, if it's a problem, we can fix it in the future :)

PR #3029, opened 2021-03-07
Add theMap section in zh-cn locales

SunweiWang:
Add theMap section in the faq part in zh-cn locales file.


madsnedergaard (approved):
Hey @SunweiWang, thanks for adding these translations!
We're very happy to see more of the app being translated to Chinese 💪

PR #3030, opened 2021-03-08
Update GB capacity 

q--:
Source is ENTSO-E like before

For some reason [PRIS](https://pris.iaea.org/PRIS/CountryStatistics/CountryDetails.aspx?current=GB) reports more nuclear than ENTSO-E. I went with the ENTSO-E value.


corradio (approved):
Thanks @q-- !

PR #3036, opened 2021-03-11
AT Capacities update

KabelWlan:
I updated the generation capacities in Austria according to the Entsoe data (source didn't change).
I don't know why "unknown" is so much more than before. Anyone have a clue?


madsnedergaard (approved):
I'm unsure why the "unknown" has increased so much, but I see the numbers are the same here: https://www.apg.at/en/markt/Markttransparenz/erzeugung/installierte-leistung

KabelWlan:
Production reported from unknown is still very low so I guess it should be fine.

PR #3037, opened 2021-03-11
update BG generation capacities

KabelWlan:
Updated the generation capacities in Bulgaria according to Entsoe.

If I go through all the Entsoe countries and update their capacities do you guys prefere one PR for all of them or shall I create a new PR for every country? Which is the easiest for you?


madsnedergaard:
> Updated the generation capacities in Bulgaria according to Entsoe.
> 
> If I go through all the Entsoe countries and update their capacities do you guys prefere one PR for all of them or shall I create a new PR for every country? Which is the easiest for you?

Hey @KabelWlan, thanks for doing this :)
I'd be okay with grouping it into one PR as long as it's the same data source and there's not any major changes (#3036 for example is nice to have as a separate so we can discuss "unknown").

madsnedergaard (approved):

PR #3040, opened 2021-03-12
CY: Return oil installed capacity instead of availability

veqtrus:
As clarified in https://github.com/tmrowco/electricitymap-contrib/issues/3013#issuecomment-797564631


madsnedergaard:
Parser runs fine and changes looks good to me, leaving final review to @corradio :)

Screenshot 2021-03-13 at 21 55 10

corradio (approved):

Issue #3015, opened 2021-02-28
NL Down


Fred-3000:
ENTSO-E data seems fine, so is there a problem with energieopwek.nl?

Kongkille:
I'm fairly certain it's our parsers. Based on the logs, it seems it's hitting an error

madsnedergaard:
Here's the error by the way: https://kibana.electricitymap.org/app/kibana#/doc/96f67170-0c49-11e9-85c1-1d63df8c862c/logstash-2021.03.01/doc?id=teX77HcBipTrw-Fajpub&_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))

smiba:
This is because energieopwek.nl does no longer supply us with json data, the response is simply empty:

https://energieopwek.nl/jsonData.php?sid=2ecde3&Day=2021-03-01

As their website describes that we need to contact one of their partners like netanders.nl I think they may have closed this down, I don't think their partners offer this for free, so they will probably demand a fat stack of cash :)

Their regular data.php is encrypted (afaik), but as this gets decoded in the client we could probably reverse engineer this. Not sure if I like that solution though!

skovhus:
FYI @MartienVisser 

We are having trouble loading JSON data from https://energieopwek.nl/jsonData.php?sid=2ecde3&Day=2021-03-01

Let me know if you can help.

MartienVisser:
Dear Kenneth,

I know that a new release was done last week. I forward your E-mail to Sander Huizinga from NetAnders. I hope he can help.

Martien

________________________________

smiba:
Received message back from them. I was indeed right with my guess that they made scraping more difficult and obfuscated on purpose

They also told me they're also in contact with Tomorrow themselves about it, so I'll let them handle it further.

MartienVisser:
Thanks for the info. I hope you will resolve it. Martien

________________________________

PaulCornelissen:
Do we want a parser that works and just uses the ENTSO-E solar and wind (which is much to low) and just puts the rest as unknown?

That can be done fairly easily as a stop-gap until we have a better data source.

Kongkille:
> Do we want a parser that works and just uses the ENTSO-E solar and wind (which is much to low) and just puts the rest as unknown?
> 
> That can be done fairly easily as a stop-gap until we have a better data source.

[...]

PR #3044, opened 2021-03-16
URL changed to https and another php script for NL energieopwek

IgorYbema:


jarek:
Does this relate to or fix #3015?

IgorYbema:
> Does this relate to or fix #3015?

Wasn't aware of the issue. But it seems this should fix the issue. Website uses same new url.

PR #3046, opened 2021-03-17
Ignore flaky code push errors

skovhus:
The code push job is flaky and blocks our deployment pipeline... We might be able to stabilize this by upgrading to https://github.com/microsoft/appcenter-cli (note that the current version of the code push CLI is 3 years old).

Silently failing is not nice, but for now we should monitor that code is being shipped to the Android and iOS apps.

Ref: https://github.com/tmrowco/electricitymap/issues/394


madsnedergaard (approved):
Makes sense, LGTM!

PR #3048, opened 2021-03-17
Update pyproject.toml

srinathjayaraman:
Change pyproject.toml to include python version 3.6 and 3.7 as acceptable, based on a suggestion by Kenneth Skovhus.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,8 @@ packages = [
 include = ["config/*.json"]
 
 [tool.poetry.dependencies]
-python = "^3.6"
+#python = "^3.6"
+python = '>= 3.6, < 3.8'
 
 arrow = {version="0.16.0", optional=true}
 beautifulsoup4 = {version="4.6.0", optional=true}


pyproject.toml
```diff
@@ -9,7 +9,8 @@ packages = [
 include = ["config/*.json"]
 
 [tool.poetry.dependencies]
-python = "^3.6"
+#python = "^3.6"
```
skovhus, line 12: We don't need to preserve the old value here.

```suggestion
#python = "^3.6"
```

skovhus: In general we avoid commented out code. This can always be restored from version control.

skovhus:
@srinathjayaraman can you [Allow edits from maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? In general that is nice. Then we can adjust the PR if we have any additions.

skovhus:
Replace by #3050. Thank you so much for your contribution!

srinathjayaraman:
I just saw this and PR #3050, thank you Kenneth!

pyproject.toml
srinathjayaraman: Thank you I will make a note of this for the future!

PR #3055, opened 2021-03-18
Upgrade ESLint + fix react-hook errors

skovhus:
We found an issue where the map would crash. This was caused by a misuse of React hooks. This should have been caught by our linting setup, but it wasn't. 

Here, we are using a more modern linting setup (with the bloated airbnb config). This also catches the issue mentioned above, and caught a bunch of other potential issues. 

I went over the code and fixed the trivial cases. 

Before merging this, then I/we would need **some more manual testing**.


madsnedergaard:
If you don't want to test all the `useEffect` dependencies (I see a lot of those changes and it might be tricky to manually test all edge cases for them), I'd be fine with adding  `// eslint-disable-next-line react-hooks/exhaustive-deps` comments for now, and then we can hopefully fix them as they are used/changed and as new ones are added :)

skovhus:
> If you don't want to test all the useEffect dependencies (I see a lot of those changes and it might be tricky to manually test all edge cases for them), I'd be fine with adding // eslint-disable-next-line react-hooks/exhaustive-deps comments for now, and then we can hopefully fix them as they are used/changed and as new ones are added :)

Good idea. That might be the best solution for now. 😅

madsnedergaard (approved):
LGTM! 👏

PR #3058, opened 2021-03-22
Update German translations

jld3103:
I'm not entirely sure about the wording, because it would look like "macht 55% der Emissionen aus", which is totally correct, but the verb for the `representing` key is split between the two values. Maybe there is a better wording for this that doesn't have this problem.


FelixDQ (approved):
Thanks @jld3103!

PR #3064, opened 2021-03-24
Fixed IT-SIC->IT-SO exchange

### LINKED ISSUE #3062: IT-South <-> IT-Sicily transmission
Since we changed IT-RO to IT-CA in #3053 we'd need to change the transmission to IT-Calabria - IT-Sicily too.

FelixDQ:
Fixes #3062


gwpicard (approved):
LGTM!

PR #3066, opened 2021-03-25
OpenNem historical

corradio:
Adds ability to refetch historical data
Tested locally with production + interconnectors + price (realtime and historical)


parsers/OPENNEM.py
```diff
@@ -75,48 +85,70 @@ def dataset_to_df(dataset):
...
-            for obj in filtered_datasets if obj['region'] == ZONE_KEY_TO_REGION[zone_key]
+            for obj in filtered_datasets if obj['region'] == region
         ])
         return df, pd.Series(capacities)
     else:
         return df
 
 
-def sum_vector(pd_series, keys, transform=lambda x: x):
+def sum_vector(pd_series, keys, transform=None):
```
corradio, line 142: Made this function a bit faster using pands

parsers/OPENNEM.py
```diff
@@ -163,9 +196,10 @@ def fetch_production(zone_key=None, session=None, target_datetime=None, logger=l
         },
         'source': SOURCE,
         'zoneKey': zone_key,
-    } for dt, row in df.iterrows()]
+    } for dt, row in tqdm(df.iterrows())]
```
corradio, line 199: Used tqdm to print the speed of this iteration (it's currently the bottleneck in the parser)

test_parser.py
```diff
@@ -21,6 +21,7 @@
 )
 
 logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
```
corradio, line 24: set default logging to DEBUG in test_parser so we can see all debug logs

pyproject.toml
```diff
@@ -29,6 +29,7 @@ opencv-python = {version="4.2.0.32", optional=true}
 xlrd = {version="1.2.0", optional=true}
 freezegun = {version="0.3.15", optional=true}
 signalr-client-threads = {version="0.0.12", optional=true}
+tqdm = "^4.59.0"
```
skovhus, line 32: As this is only used by the parsers, we should add the `optional` flag here. I also lowered the version here to support the version we are using internally.

```suggestion
tqdm = {version="^4.19.5", optional=true}
```

We should also add `tqdm` to the list to `parsers` in the `[tool.poetry.extras]` section.

skovhus (changes requested):

pyproject.toml
skovhus: > We should also add tqdm to the parsers list in the [tool.poetry.extras] section.

Reminder to update the pyproject.toml file.

skovhus: @corradio

corradio: Let me know if https://github.com/tmrowco/electricitymap-contrib/pull/3066/commits/407f0117520b9381178042311a4942b733299efe fixes it

skovhus (approved):

PR #3069, opened 2021-03-25
update renewable capacities India

KabelWlan:
I updated the renewable energy generation capacities in India according to the following sources: 
wind: https://www.mnre.gov.in/wind/current-status/
solar: https://www.mnre.gov.in/solar/current-status/
biomass: https://www.mnre.gov.in/bio-energy/current-status

Also I add this sources to data_source.md

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -2505,8 +2505,8 @@
   },
   "IN": {
     "capacity": {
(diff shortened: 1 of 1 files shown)


KabelWlan:
I'm sorry for accidently opening the same PR twice...

Issue #2541, opened 2020-06-06
Data for region Labrador (Canada)


q--:
[Current generation](https://nlhydro.com/system-information/supply-and-demand/)

>Current Island System Generation

so for Newfoundland only

[It seems there is an interconnector](https://www.thetelegram.com/news/local/no-one-but-locals-looking-to-wheel-power-through-newfoundland-and-labrador-333035/), yet, curiously the page linked at the beginning says "Current system generation = current demand on the system". This may or may not mean imports are included in the figure

---
[Generation + forecast data](http://www.pub.nl.ca/applications/IslandInterconnectedSystem/DemandStatusReports.htm) of only a few days old is available in PDF format here.

---

[Installed capacity](https://www.oasis.oati.com/woa/docs/NLSO/NLSOdocs/Network_Resource_List_03132019.pdf), from the TSO [website](https://www.oasis.oati.com/NLSO/index.html)
(Via [here](https://nlhydro.com/system-information/)  - "The [Newfoundland and Labrador System Operator](https://www.oasis.oati.com/NLSO/index.html) (NLSO) is responsible for the safe and reliable operation of the bulk electric system in Newfoundland and Labrador, including the administration and provision of transmission on the NL Transmission System pursuant to the Multi-Party Pooling Agreement (MPPA).")

This one seems to be for Labrador as well, then.

[Page 14 (16 in the PDF)](https://www.oasis.oati.com/woa/docs/NLSO/NLSOdocs/Methodology_for_the_Development_of_Rates_for_Transmission_Service_FINAL_02052018.pdf) of this has a map with installed generating facilities, including their type. Coupled with the installed capacity document linked above, it should be possible to determine what type of generation the facilities are and where they are located.

[This page](https://nlhydro.com/about-hydro/corporate-overview/) also mentions installed capacity. The grid map it links to is dated 2011 instead of the one from 2017 in the previous link.

---
Possibly interesting: [quarterly reports](https://nlhydro.com/about-hydro/publications/), haven't looked through these yet

jarek:
Still no fuel data, eh?

The Maritime Link to Nova Scotia is fairly new. Kind of sad if they didn't manage to export much on it.

pierresegonne:
Hey @q-- Thanks for the investigation and the breakdown,

[...]

PR #3077, opened 2021-03-29
feat(zone): add newfoundland

byronwilliams:
Hi!

I was browsing the Newfoundland Hydro website - https://nlhydro.com/system-information/supply-and-demand/ - as you do... and came across the current island generation. It's not that great since it lumps together a load of different production methods. But it's not on the map.

When you get a moment can you generate the bounding box for it?

I've called it CA-NL-NL because it's not the whole of Newfoundland and Labrador, feel free to change

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1009,7 +1009,7 @@
     "timezone": null
   },
   "CA-NL-NF": {
-    "_comment": "5428 MW of hydro is Churchill Falls Generating Station, electricity from which is mostly sold to Quebec and might not appear on balance sheets",
+    "_comment": "https://nlhydro.com/system-information/supply-and-demand/ This is just the island of Newfoundland",
     "flag_file_name": "ca.png",
     "timezone": "America/St_Johns",
     "bounding_box": [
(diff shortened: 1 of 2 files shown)


madsnedergaard:
Nice, is there any of the suggested tasks in #2541 that needs to be done here? Looping you in here as well, @pierresegonne :)

pierresegonne:
Thanks @byronwilliams !

We'll be spliting up the newfoundland and labrador zones as a first step :)

For reference, (See #2541), even with your parser, we still need to figure out a way to estimate a static generation breakdown for aggregating everything in the `unknown`, otherwise Electricity there will appear way dirtier than it actually is :)

pierresegonne:
Hey @byronwilliams took the liberty to make a few changes to adapt your great work on the parser to the slitting of Newfoundland-Labrador into both Newfoundland and Labrador

madsnedergaard:
What's the status on this, could we maybe just merge this even though it's all grouped into "unknown" and then it can be improved further afterwards? @pierresegonne

gwpicard:
@madsnedergaard @pierresegonne any updates on what we should do with this PR?

pierresegonne:
The status for this is tracked in #2541 
Basically, we need to compute the emission factors for the large "unknown" this parser will provide us.

PR #3085, opened 2021-04-12
Update NL.py

leohuisman:
Setting energieopwek_nl to False in order to use the ENTSOE data instead of the energieopwek-data (which has not been available for quite some time). The issue was discussed at https://github.com/tmrowco/electricitymap-contrib/issues/3015

### DIFF
--- a/parsers/NL.py
+++ b/parsers/NL.py
@@ -12,7 +12,7 @@ import requests
 
 
 def fetch_production(zone_key='NL', session=None, target_datetime=None,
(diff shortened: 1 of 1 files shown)


FelixDQ:
Thanks @leohuisman, I gave an update on the issue here https://github.com/tmrowco/electricitymap-contrib/issues/3015#issuecomment-817692119. 

We are going to try to implement a proper solution instead of using the fallback for now.

PR #3086, opened 2021-04-12
Pylance - PEP8 and code cleanup in Parsers [WIP]

yuvalabou:
- Clean unused imports
- Sort imports
- Remove docstrings that are copied from "example"
- Remove "return" from docstrings and replace with a type hint

### DIFF
--- a/parsers/AX.py
+++ b/parsers/AX.py
@@ -1,11 +1,7 @@
-#!/usr/bin/env python3
-# The arrow library is used to handle datetimes
-import arrow
-# The request library is used to fetch content through HTTP
-import requests
 
-# Numpy and PIL are used to process the image
+import arrow
 import numpy as np
+import requests
 from PIL import Image
 
 
--- a/parsers/AX.py
+++ b/parsers/AX.py
@@ -262,34 +262,8 @@ def _fetch_data(session=None):
 
 
 def fetch_production(zone_key='AX', session=None, target_datetime=None, logger=None):
-    """Requests the last known production mix (in MW) of a given country
-
-    Arguments:
-    zone_key (optional) -- used in case a parser is able to fetch multiple countries
-    session (optional)      -- request session passed in order to re-use an existing session
-
-    Return:
-    A dictionary in the form:
-    {
-      'zoneKey': 'FR',
-      'datetime': '2017-01-01T00:00:00Z',
-      'production': {
-          'biomass': 0.0,
-          'coal': 0.0,
-          'gas': 0.0,
-          'hydro': 0.0,
-          'nuclear': null,
-          'oil': 0.0,
-          'solar': 0.0,
-          'wind': 0.0,
-          'geothermal': 0.0,
-          'unknown': 0.0
-      },
-      'storage': {
-          'hydro': -10.0,
-      },
-      'source': 'mysource.com'
-    }
+    """
+    Requests the last known production mix (in MW) of a given country.
     """
     if target_datetime:
         raise NotImplementedError('This parser is not yet able to parse past dates')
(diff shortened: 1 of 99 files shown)


parsers/AX.py
```diff
@@ -1,11 +1,7 @@
-#!/usr/bin/env python3
-# The arrow library is used to handle datetimes
-import arrow
-# The request library is used to fetch content through HTTP
-import requests
 
-# Numpy and PIL are used to process the image
```
jarek, line 7: Keeping comments like these is useful IMO. Maybe not `requests`, but that numpy is for image processing

parsers/BD.py
```diff
@@ -75,9 +73,10 @@ def new_format_converter(df, logger):
     return df
 
 
-def production_processer(df, target_datetime, old_format=False):
-    """Takes dataframe and extracts all production data and timestamps.
-    Returns a list of 2 element tuples in form (dict, arrow object)."""
```
jarek, line 80: The second line of this comment is not fully captured in the return type annotation. Same below for exchange_processer

parsers/CL.py
```diff
@@ -32,10 +32,9 @@ def get_data_live(session, logger):
     return json_total, json_ren
 
 
-def production_processor_live(json_tot, json_ren):
+def production_processor_live(json_tot, json_ren) -> list:
     """
     Extracts generation data and timestamp into dictionary.
-    Returns a list of dictionaries for all of the available "live" data, usually that day.
```
jarek, line 38: Also potentially useful information removed

parsers/ES_IB.py
```diff
@@ -1,18 +1,13 @@
...
 from requests import Session
-from ree import (Formentera, Ibiza,
-                 Mallorca, Menorca,
-                 BalearicIslands)
-# package "ree" is used to parse data from www.ree.es // maintained on github by @hectorespert
 
 from .lib.exceptions import ParserException
 from .lib.validation import validate, validate_production_diffs
 
-## Guess we'll need to figure these out later?! Adapted from ES-CN:
```
jarek, line 14: This comment should stay, minimum zone demands are specified to be 0, that is clearly a placeholder. Maybe add a `TODO:` to make it clearer? Otherwise, remove the use of FLOORS in this script altogether, it's optional in `validate`.

parsers/IL.py
```diff
@@ -7,37 +6,86 @@
...
+        },
+        "source": IEC_URL,
+        "price": PRICE,
     }
 
-    return datapoint
+    return data
+
 
+# def fetch_consumption(zone_key='IL', session=None, target_datetime=None, logger=None):
```
jarek, line 71: what happened here? rebase needed?

parsers/IL.py
```diff
@@ -18,29 +18,46 @@
...
+                flat_list.append(item)
+        else:
+            flat_list.append(element)
+    return flat_list
 
 def fetch_production(zone_key="IL", session=None, logger=None):
     first = requests.get(IEC_PRODUCTION)
     first.cookies
-    second = requests.get(IEC_PRODUCTION, cookies=first.cookies)
+    with requests.get(IEC_PRODUCTION, cookies=first.cookies) as second:
```
jarek, line 36: Does using a long context manager here add something? Also immediately above there's a statement with no effect.

parsers/CL.py
```diff
@@ -118,34 +117,8 @@ def production_processor_historical(raw_data):
...
-          'geothermal': 0.0,
-          'unknown': 0.0
-      },
-      'storage': {
-          'hydro': -10.0,
-      },
-      'source': 'mysource.com'
-    }
+    """
+    Requests the last known production mix (in MW) of a given zone.
```
jarek, line 121: All other instances of this comment don't end with a full-stop, only this one does

jarek (commented):
Hey, thanks for going through this and cleaning up! Looks pretty good, just noted some things as I scroll down through the diff as it is now (I do note that it's a draft).

It would also be interesting to set up a PR-time check with `black` to maintain style in future PRs.

parsers/IL.py
yuvalabou: Yes, will do. Noticed that right after I made that PR

parsers/BD.py
yuvalabou: I think that it can be done using __future__
Needs testing

parsers/IL.py
yuvalabou: This is part of IL parser which do not belong here. will rebase

parsers/CL.py
yuvalabou: This is part of IL parser which do not belong here. will rebase

yuvalabou:
Messed up my repo, and this is why you don't make changes to master 🤦
I think I will start over if you don't mind.

PR #3087, opened 2021-04-12
Fixes OPENNEM parser when there's no solar_rooftop data available

madsnedergaard:
## Description

The parser currently fails when there's no `SOLAR_ROOFTOP` data available.
This PR should fix that.

### Issue
[Kibana log entry](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:extra.key,negate:!f,params:(query:AUS-NSW,type:phrase),type:phrase,value:AUS-NSW),query:(match:(extra.key:(query:AUS-NSW,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))

Screenshot 2021-04-12 at 13 45 01

### DIFF
--- a/parsers/OPENNEM.py
+++ b/parsers/OPENNEM.py
@@ -100,9 +100,7 @@ def fetch_main_df(data_type, zone_key=None, sorted_zone_keys=None, session=None,
     df = pd.concat([dataset_to_df(ds) for ds in filtered_datasets], axis=1)
     if data_type == 'power' and zone_key:
         # SOLAR_ROOFTOP is only given at 30 min interval, so let's interpolate it
-        df["SOLAR_ROOFTOP"] = (
-            df["SOLAR_ROOFTOP"].interpolate(limit=5) if "SOLAR_ROOFTOP" in df else 0
-        )
+        df['SOLAR_ROOFTOP'] = df['SOLAR_ROOFTOP'].interpolate(limit=5) if 'SOLAR_ROOFTOP' in df else 0
         # Parse capacity data
         capacities = dict([
             (obj['id'].split('.')[-2].upper(), obj.get('x_capacity_at_present'))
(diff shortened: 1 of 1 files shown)


corradio (approved):

parsers/OPENNEM.py
```diff
@@ -100,7 +100,7 @@ def fetch_main_df(data_type, zone_key=None, sorted_zone_keys=None, session=None,
     df = pd.concat([dataset_to_df(ds) for ds in filtered_datasets], axis=1)
     if data_type == 'power' and zone_key:
         # SOLAR_ROOFTOP is only given at 30 min interval, so let's interpolate it
-        df['SOLAR_ROOFTOP'] = df['SOLAR_ROOFTOP'].interpolate(limit=5)
+        df['SOLAR_ROOFTOP'] = df['SOLAR_ROOFTOP'].interpolate(limit=5) if 'SOLAR_ROOFTOP' in df else 0
```
corradio, line 103: We could also just do 
```python
if 'SOLAR_ROOFTOP' in df:
  df['SOLAR_ROOFTOP'] = df['SOLAR_ROOFTOP'].interpolate(limit=5)
```

madsnedergaard: In that case `df['SOLAR_ROOFTOP']` will be undefined, is that okay or will it mess with other calculations? :)

madsnedergaard: Seems like the result is the same, so going with your suggestion.

Issue #2495, opened 2020-05-11
Capacity data for US zones should be added back


robertahunt:
As pointed out by https://mobile.twitter.com/thefrackingguy/status/1258225824567705601, would be nice to have capacities for US zones again, seems like some of this data was removed when we switched to the new view.

I am not yet sure if there is a central repository for this data for not, or if we would have to comb each individual BA's website to get the data.

robertahunt:
It looks like these forms might be able to be used to extract the capacity data:
https://www.eia.gov/electricity/data/eia860/

At first glance, it looks like the Plant data can be used to map each plant to a balancing authority, and the Generation data can be used to get the capacity and fuel type for each plant.

robertahunt:
https://www.eia.gov/electricity/data/eia860M/ Looks like it has even more recent data we could use

corradio:
Closing due to inactivity

PR #3093, opened 2021-04-14
add US capacities

KabelWlan:
I added the generation capacites to the US zones. Got the idea from the issue #2495. 

I used the following source, which I'll add to data_sources.md: https://www.eia.gov/electricity/data/eia860M/

(It shows me in the preview that there are some colones missing in the zones AVRN, BPAT and PSEI. I don't know why because they're there in the normal file.) -> figured it out, it was the commas...


FelixDQ (changes requested):
Thank you @KabelWlan!

Looks like there is a syntax error somewhere in the zones.json. We should be ready to merge if you can fix the issue :-)

KabelWlan:
Finally found my mistake... I hope it’s fine now.

martincollignon:
@KabelWlan you should probably add yourself as a contributor as well :)

madsnedergaard (approved):

FelixDQ (approved):
Awesome!

PR #3095, opened 2021-04-15
IL refactor

yuvalabou:
Hope this is the last try..

Refactor IL code:
Black, sorted and cleaned

### DIFF
--- a/parsers/IL.py
+++ b/parsers/IL.py
@@ -1,43 +1,69 @@
-#!/usr/bin/env python3
-# coding=utf-8
+"""
+Israel's electricity system load parser.
 
-# This parser returns Israel's electricity system load (assumed to be equal to electricity production)
-# Source: Israel Electric Corporation
-# URL: https://www.iec.co.il/en/pages/default.aspx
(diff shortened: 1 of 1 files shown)


FelixDQ (approved):
LGTM! I took the liberty of adding the target_datetime parameter, so that the parser can be called through `test_parser.py`

parsers/IL.py
```diff
@@ -1,43 +1,76 @@
...
+
+import arrow
+from requests import get
 from bs4 import BeautifulSoup
 
+IEC_URL = "www.iec.co.il"
+IEC_PRODUCTION = (
+    "https://www.iec.co.il/_layouts/iec/applicationpages/lackmanagment.aspx"
+)
+PRICE = "50.66"  # Price is determined yearly
```
q--, line 25: Maybe a check for the current year somewhere would be a good idea, in case nobody remembers to update the price for 2022?

yuvalabou: I can probably scrape the site while I'm at it, consider this a placeholder.

PR #3098, opened 2021-04-16
Disable Chile live parser

FelixDQ:
The current Chile live parser doesn't provide breakdown values for Hydro and Geothermal. Hydro is a fairly large part of the unknown production mix, so the current CFE% and carbon intensity is a bit misleading.

It's annoying that we have to disable the live parser, but we currently don't have the infrastructure to ingest both data sources. We are working on a solution for this, and it hopefully won't be long before we can get Chile back on the map.

**For reviewer:** I also formatted the code in a second commit. Look at the first commit for the changes

### DIFF
--- a/parsers/CL.py
+++ b/parsers/CL.py
@@ -16,6 +16,11 @@ API_BASE_URL = "https://sipub.coordinador.cl/api/v1/recursos/generacion_centrale
 API_BASE_URL_LIVE_TOT = "http://panelapp.coordinadorelectrico.cl/api/chart/demanda"
 API_BASE_URL_LIVE_REN = "http://panelapp.coordinadorelectrico.cl/api/chart/ernc"  # ERNC = energias renovables no convencionales
 
+# Live parser disabled because of a insuffcient breakdown of Unknown.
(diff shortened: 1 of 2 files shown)


parsers/CL.py
```diff
@@ -102,53 +102,16 @@ def production_processor_historical(raw_data):
         combined[elem['datetime']].update(elem)
 
     ordered_data = sorted(combined.values(), key=itemgetter("datetime"))
-
-    # For consistency with live API, hydro and geothermal must be squeezed into unknown
```
skovhus, line 106: Dont' we still need to do this if we enable the live parser again? In that case we could move `ENABLE_LIVE_PARSER` to the outer scope and check it here.

FelixDQ: I was thinking that if we enable the live parser again, we would have a system to keep both datasets. I.e. we don't want to reenable the live parser if that means we have to lump hydro and geothermal into unknown.

But, I can see how it would be confusing in the meantime. I've added the check

skovhus (approved):

PR #3099, opened 2021-04-18
Add Israel consumption data

yuvalabou:
Add IL consumption info.
Add myself as contributor

### DIFF
--- a/parsers/IL.py
+++ b/parsers/IL.py
@@ -89,7 +89,7 @@ def fetch_consumption(
     return {
         "zoneKey": zone_key,
         "datetime": arrow.now(TZ).datetime,
-        "consumption": {"unknown": consumption[0]},
+        "consumption": consumption[0],
         "source": IEC_URL,
     }
 
--- a/parsers/IL.py
+++ b/parsers/IL.py
@@ -26,21 +26,8 @@ PRICE = "50.66"  # Price is determined yearly
 TZ = "Asia/Jerusalem"
 
 
-def flatten_list(_2d_list) -> list:
-    flat_list = []
-    for element in _2d_list:
-        if type(element) is list:
-            for item in element:
-                flat_list.append(item)
-        else:
-            flat_list.append(element)
-    return flat_list
-
-
-def fetch_production(zone_key="IL", session=None, target_datetime=None, logger=None) -> dict:
-    if target_datetime:
-        raise NotImplementedError('This parser is not yet able to parse past dates')
-        
+def fetch_all() -> list:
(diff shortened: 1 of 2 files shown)


parsers/IL.py
```diff
@@ -26,21 +26,8 @@
...
-        else:
-            flat_list.append(element)
-    return flat_list
-
-
-def fetch_production(zone_key="IL", session=None, target_datetime=None, logger=None) -> dict:
-    if target_datetime:
-        raise NotImplementedError('This parser is not yet able to parse past dates')
-        
+def fetch_all() -> list:
```
madsnedergaard, line 29: Suggestion for improvement (although it's a bit out of scope):
Would you mind changing the selectors from finding all `.statusVal`s to use the more stable selectors (`#lblActualSupply`,`#lblCurrentSupply`)? That way we're not relying on the order (and have to remove an entry in the middle) and the function could then return a dictionary with `{production: X, consumption: Y}` instead of a list :)

parsers/IL.py
```diff
@@ -54,23 +41,62 @@ def fetch_production(zone_key="IL", session=None, target_datetime=None, logger=N
...
+
+    data = fetch_all()
+
+    consumption = [float(item) for item in data]
+
+    # all mapped to unknown as there is no available breakdown
+    return {
+        "zoneKey": zone_key,
+        "datetime": arrow.now(TZ).datetime,
+        "consumption": {"unknown": consumption[0]},
```
madsnedergaard, line 92: The value of `"consumption"` should just be the number and not a breakdown-dictionary, like in this example:

```js
    {
      'zoneKey': 'FR',
      'datetime': '2017-01-01T00:00:00Z',
      'consumption': 10.0,
      'source': 'mysource.com'
    }
```

It currently fails with this error:

Screenshot 2021-04-28 at 12 05 03

madsnedergaard (changes requested):
Hey @yuvalabou, thanks for the work here!
I have two minor comments - one is just a suggestion that would be nice, but not strictly required :)

parsers/IL.py
yuvalabou: Maybe I can do it, but as sayed it is out of the scope of this PR.
will open a seperate one for it soon.. 👍

skovhus (approved):
It would be great if you could fix the suggestion of using a more specific selector in another PR. Thanks!

madsnedergaard:
Thanks for continuing this @skovhus, and thanks again @yuvalabou 🙏

PR #3100, opened 2021-04-19
Clean docstring left from example

yuvalabou:
added annotations where neccessary


parsers/NI.py
```diff
@@ -88,16 +88,14 @@ def get_time_from_system_map(text):
...
+def get_production_from_map(requests_obj) -> tuple:
     """
     Get frequently-updated information on MAP_URL.
     This page is programmed to refresh every 10 seconds, and the timestamp
     in its text indicates that the information is updated every 10 to 30 seconds.
 
     However, it seems to bundle in solar generation with generation at another plant.
     get_production_from_summary() includes solar explictly.
-
-    :return: tuple(production, datetime_datetime)
```
madsnedergaard, line 100: Worth keeping.

parsers/TR.py
```diff
@@ -49,12 +49,8 @@ def as_float(prod):
     return prod
 
 
-def get_last_data_idx(productions):
-    """
-    Find index of the last production
-    :param productions: list of 24 production dict objects
-    :return: (int) index of the newest data or -1 if no data (empty day)
```
madsnedergaard, line 56: Worth keeping.

parsers/US_NY.py
```diff
@@ -54,11 +50,10 @@ def timestamp_converter(timestamp_string):
     return dt_aware
 
 
-def data_parser(df):
+def data_parser(df) -> list:
     """
     Takes dataframe and loops over rows to form dictionaries consisting of datetime and generation type.
     Merges these dictionaries using datetime key.
-    Maps to type and returns a list of tuples containing datetime string and production.
```
madsnedergaard, line 61: Worth keeping.

parsers/BD.py
```diff
@@ -75,9 +75,8 @@ def new_format_converter(df, logger):
     return df
 
 
-def production_processer(df, target_datetime, old_format=False):
-    """Takes dataframe and extracts all production data and timestamps.
-    Returns a list of 2 element tuples in form (dict, arrow object)."""
```
madsnedergaard, line 80: Worth keeping.

madsnedergaard (changes requested):
Hi @yuvalabou, thanks for doing this - it's pretty nice to get this cleaned up and we really appreciate the thorough work! 💪 

I think it would be nice to keep some of the comments about returned values. Even though it can often be inferred from the return type (and thanks for adding those!), there are some cases where the comments give additional context that is nice. I have added comments for some of those, but probably missed some (it's a pretty big PR 😉).

So I'd suggest the following:
- [x] Remove all the copied examples 
- [x] Remove useless comments about returns (e.g. `Returns a dictionary.`)
- [ ] Keep comments that are more detailed than the return type (e.g. `:return: (int) index of the newest data or -1 if no data (empty day)`) (you have to read through the code to find the "-1" part)

skovhus:
This PR is not getting easier to review after all the unrelated formatting changes...

It would be great if we could limit the scope to only be about removal of redundant doc strings and some type annotations. 

About the formatting changes: Dis you used a formatter, like Black, or manually formatted the code?

yuvalabou:
I didn't use any formatter as it would result in massive changes to all files, every change I did by hand and indeed tried not to get carried away.

But I do agree, I will revert some of them and make a different PR for some of the more format related changes.

skovhus:
> But I do agree, I will revert some of them and make a different PR for some of the more format related changes.

Thanks! :)

parsers/KR.py
```diff
@@ -58,10 +59,8 @@ def timestamp_processor(timestamps, with_tz=False, check_delta=False):
     return unified_timestamp
 
 
-def check_hydro_capacity(plant_name, value, logger):
-    """Makes sure that generation for each hydro plant isn't above listed capacity.
-    Returns True or raises ValueError.
-    """
+def check_hydro_capacity(plant_name, value, logger) -> True | ValueError:
```
skovhus, line 62:
```suggestion
def check_hydro_capacity(plant_name, value, logger) -> bool | ValueError:
```

skovhus (approved):
I've looked through the changes. This took a bit of time as the PR is fairly big. :D 

@madsnedergaard when you have time, it would be great if you could double check.

madsnedergaard (approved):
Went through everything again and it looks good to me! 
It's really nice to get all the parsers cleaned up, so thanks for doing this! 🙏

Issue #3002, opened 2021-02-18
Show error message when wind or solar is unavailable


corradio:
Show an error message if the wind and/or solar data is not available when clicking the wind/solar layer buttons. 

Right now, there's no feedback when the enabling those layers if we're missing weather data (you click a button and nothing happens).

![image](https://user-images.githubusercontent.com/3296643/115720493-6b37f180-a37d-11eb-89e3-2cd99e397f9c.png)

evcheng1:
I am interested in taking this issue. Thanks!

PR #3104, opened 2021-04-22
Wind solar data error

sidneythekidney:
This pull request was inspired by issue #3002. I created a small front end feature that displays a simple error message on the country tooltip if that country/zone is missing solar and/or wind data. The error does not display for countries with no available data.

Sample Image:
image

### DIFF
--- a/web/locales/en.json
+++ b/web/locales/en.json
@@ -95,7 +95,8 @@
         "cpinfo": "<b>consumption</b> takes imports and exports into account, <b>production</b> ignores them",
         "selectLanguage": "Select language",
         "lowCarbDescription": "Includes renewables and nuclear",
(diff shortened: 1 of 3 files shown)


madsnedergaard:
Hey @sidneythekidney. Thanks for your work here! 
With the proposed change here, we would show the error message if a zone does not have any production of solar or wind. 

However, I think the original issue (#3002) was meant as "show an error message if the wind and/or solar layer is not available when clicking the wind/solar buttons". Right now, there's no feedback when the enabling those layers and we're missing weather data (you click a button and nothing happens).

image

Sorry for the lack of description in original issue, we'll try to do better! :)

sidneythekidney:
No problem, appreciate the review anyways!

PR #3106, opened 2021-04-22
Update JP-TK and KR capacities.

pierresegonne:
# Methodology

__KR__: 
Pull out the capacity data from [here](http://epsis.kpx.or.kr/epsisnew/selectEkpoBftChart.do?menuId=020100) and take the figures for `2020-03`
* What is translated to "Positive Number" by google trad is actually pumped hydro.
* Coal is accounted as: "Bituminous coal" + "Hard coal" + "Coal gasification" + "Etc" ("Etc" = residual heating)
* Gas is accounted as: "LNG" + "Fuell cell"

__JP-TK__:
Pulled out the map of all the TEPCO power plants from [here](http://agora.ex.nii.ac.jp/earthquake/201103-eastjapan/energy/electrical-japan/operator/3.html.ja), extracted the html table and parsed it into an excel file.
Then I went out to check each thermal's plant fuel and computed the total capacity for each production factor.
The final csv can be found here:

[JP-TK_plants.csv](https://github.com/tmrowco/electricitymap-contrib/files/6360428/JP-TK_plants.csv)


gwpicard (approved):
LGTM!

PR #3109, opened 2021-04-23
Fix iOS layout

### LINKED ISSUE #3034: Look into potentially broken navigation bar for the iOS app on iPhone 12?
## Description

A user reported that the navigation bar isn't working properly on the iOS app on iPhone 12.

Can anyone with an iPhone 12 and the electricityMap app installed help with testing this? :)

Screenshot 2021-03-11 at 15 09 59

hybridcattt:
Fixes #3034 

- uses [safe area insets](https://webkit.org/blog/7929/designing-websites-for-iphone-x/) instead of hardcoded values
- added extra padding on the bottom for the tab bar
- fixes unnecessary padding on map elements and flash messages on iPad
- on iPad the issue with the status bar still stands, as `#header` doesn't apply to anything. I didn't address it here, keeping it manageable from my side :) 

Not sure if adding a style tag to head like that is a good way to go in this case - maybe there's a better place for styles with css selectors. I thought that checking for device type and iOS version with JS would be too messy - there's a lot of combinations and things change every year. 

PS: regarding the invisible status bar (white on white). Something is off with Cordova it seems, it's configured as black, but renders white only on latest iOS. Might need to update Cordova if you decide to fix that. 
PS#2: getting Cordova stuff to work was a bit tricky - had to disable Sentry and add a dummy google-info plist. The app was also getting some 500's when run via Cordova - hope that didn't trigger any alerts :)


corradio:
So cool to see you here @hybridcattt !

hybridcattt:
@corradio happy to help ☺️

web/src/cordova.js
```diff
@@ -28,25 +28,56 @@ export const cordovaApp = {
...
+      };
+
+      select("head")
+        .append("style")
+        .text(`
+            /* iOS 10 */
+            ${styles('20px', '0px')}
+            /* iOS 11.0 */
+            @supports(padding-top: constant(safe-area-inset-top)) {
+              ${styles('constant(safe-area-inset-top), 20px)', 'constant(safe-area-inset-bottom, 0px)')}
```
skovhus, line 74:
```suggestion
              ${styles('constant(safe-area-inset-top, 20px)', 'constant(safe-area-inset-bottom, 0px)')}
```

web/src/cordova.js
```diff
@@ -28,25 +28,56 @@ export const cordovaApp = {
...
-
-      select('.flash-message .inner')
-        .style('padding-top', `${extraPadding}px`);
-
-      select('.mapboxgl-zoom-controls')
-        .style('transform', `translate(0,${extraPadding}px)`);
-      select('.layer-buttons-container')
-        .style('transform', `translate(0,${extraPadding}px)`);
+
+      var styles = function (top, bottom) {
```
skovhus, line 32:
```suggestion
      const styles = function (top, bottom) {
```

skovhus (commented):
Thank you so much for your contribution! I'm very happy to see you here. :) 

I'll test this out next week.

hybridcattt:
Thanks @skovhus 😊 and thanks for taking care of the fixes! Weird how that paren got past me at last moment, hehe. 
Looking forward to hear how the testing goes :)

skovhus:
@hybridcattt I'm unfortunately having a hard time with my Xcode/Cordova setup. Did you manage to get it up and running with Xcode 12?

hybridcattt:
@skovhus yes, though it is a bit tricky indeed. Here's the order of things that worked for me: 
```
docker-compose build
cd mobileapp
./build.sh
cordova platform remove ios && cordova platform add ios
```

After that open `mobileapp/platforms/ios/electricityMap.xcworkspace` with Xcode 12. 
You also might want to disable codepush (in `cordova.js`), otherwise the app restores to the production state when you reopen it.

skovhus:
Thank you! 

With `cordova platform remove ios && cordova platform add ios` I also got it building now! Did you skip `cordova prepare`?

I'm getting a white screen though, which seems to be related to a missing `cordova_plugins.js` file (shown in the Safari debugger). This might be caused as I'm skipping `cordova prepare`. 

I'll look more into this next week! 😅

hybridcattt:
@skovhus 

The very first time I followed the instructions from [the readme](https://github.com/tmrowco/electricitymap-contrib/blob/master/mobileapp/README.md), including `cordova prepare`. `run` and `build` didn't work for me though, I ran the app from Xcode directly. Then for rebuilding after making source changes I ended up on what I shared above. 

Good luck :)

skovhus (approved):

skovhus:
Thank you so much for fixing this!

web/src/cordova.js
```diff
@@ -28,25 +28,56 @@ export const cordovaApp = {
...
+        /* TODO: this selects nothing, header on iPad still overlaps with the status bar */
+        #header {
+          padding-top: ${top};
+        }
+
+        #mobile-header {
+          padding-top: ${top};
+        }
+
+        @include respond-to('small') {
```
skovhus, line 43: I haven't seen the `include respond-to` syntax before. Is this valid?

skovhus: Removing `@include respond-to('small') {` solves the issue where all the CSS rules below are not applied.

hybridcattt: There's 30 occurrences of that in the files (well, 29 without this one), so yeah, it works. It's needed because padding does not need to be applied when there's a header (on iPad). 
It worked on my machine (c) but perhaps it's missing an import? I'm checking locally as well :)

skovhus: It is a scss/sass construct that doesn't work in plain CSS (which we are using here). The occurences are in .scss files that are preprocessed.

hybridcattt: right 🤦 then it's an easy fix, just replacing that with a media selector. I can do it, but later - otherwise go ahead ☺️

hybridcattt: I tested on a web version a lot because I couldn't stand waiting 3 min of rebuild after every change. and there it obviously worked 😅

skovhus: I'll add the media query to https://github.com/tmrowco/electricitymap-contrib/pull/3132 :)

hybridcattt: thanks 👍

skovhus: > waiting 3 min of rebuild after every change. and there it obviously worked 😅

I hope to get time looking into a better local development flow – the current one is awful. 😅

hybridcattt: would be an awesome improvement 😄

PR #3118, opened 2021-04-28
Upgrade tqdm to ensure it works in worker threads

skovhus:
@jarek from an issue in https://github.com/tmrowco/electricitymap-contrib/pull/3114 where tqdm wouldn't work in worker threads.


madsnedergaard (approved):
LGTM!

PR #3122, opened 2021-04-29
Format .lib folder

yuvalabou:
- Format .lib folder using Black
- Remove unused and sort imports
- Remove docstrings from self describing functions


skovhus:
Thanks for your contribution. 

Internally we do use Black as a formatter – and we are pretty fond of it. But we don't enforce any formatting here on contrib, as it raises the barrier of entry for new contributors. As your formatting changes here will not be enforced, then I'm not sure about the value this provides. New contributions will not follow the conventions here.

I do like the cleaning up of the doc strings though. 👏

yuvalabou:
> But we don't enforce any formatting here on contrib, as it raises the barrier of entry for new contributors.

I don't really agree here if I may.
The upper most reason to format a code is to *lower* the barrier and allow more begginers like me to join and collab.
Internally you are right, as it is not change this often - but it is nice to have a mor readable code across.

skovhus:
Note that I talk about _enforcing_ strict code styling. E.g. failing the build if you are quoting strings using `'` instead of `"`. Doing so will raise the barrier of entry in my opinion.

yuvalabou:
Okay, well I don't pushing to that direction, I agree that things like 'pre-commit' and others are really headache for new users when you forced to use them. But this is not really the case here.. I am just opening PRs for things that I know and prefer, taste if you will.. I do like those conversations though, Learning a lot about open-source

skovhus (approved):
Looks good to me. Thanks for only doing formatting in this PR, it might it much easier to review.

PR #3123, opened 2021-04-29
Add price scraping to IL

yuvalabou:


corradio (approved):
Thank you! Much better than the static price.
This probably shouldn't have made its way to the parser in the first place.

parsers/IL.py
```diff
@@ -71,7 +83,7 @@ def fetch_production(
         "datetime": arrow.now(TZ).datetime,
         "production": {"unknown": production[0] + production[1]},
         "source": IEC_URL,
-        "price": PRICE,
+        "price": price,
```
corradio, line 86: This key is actually ignored.
The data should be returned in the `fetch_price` method, which should have the signature depicted here:
https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/example.py#L110-L111

corradio (changes requested):

parsers/IL.py
yuvalabou: Oh didn't knew that. will fix.

yuvalabou:
> This probably shouldn't have made its way to the parser in the first place.

You are probably right, however Im still early on my learning curve, and every new function is new to me.
Thank you for understanding

corradio:
> > This probably shouldn't have made its way to the parser in the first place.
> 
> You are probably right, however Im still early on my learning curve, and every new function is new to me.
> Thank you for understanding

Absolutely no problem :) It's actually our fault as I realise we should have documented it better.

parsers/IL.py
```diff
@@ -55,14 +55,29 @@ def flatten_list(_2d_list) -> list:
     return flatten_list(cleaned_list)
 
 
+def fetch_price(zone_key="IL", session=None, target_datetime=None, logger=None) -> dict:
+    """Fetch price from IEC table."""
+    response = get(IEC_PRICE)
```
corradio, line 60: Are you able to add a check for the presence of `target_datetime`?
```python
    raise NotImplementedError('This parser is not yet able to parse past dates')
```

parsers/IL.py
```diff
@@ -55,14 +55,29 @@ def flatten_list(_2d_list) -> list:
...
+    soup = BeautifulSoup(response.content, "lxml")
+
+    price = soup.find("td", class_="ms-rteTableEvenCol-6")
+
+    return {
+        "zoneKey": zone_key,
+        "currency": "NIS",
+        "datetime": arrow.now(TZ).datetime,
```
corradio, line 68: Is it possible to retrieve the date at which the price was set?
This would make the underlying data more robust.

parsers/IL.py
```diff
@@ -55,14 +55,29 @@ def flatten_list(_2d_list) -> list:
...
+        "price": price.p.text,
```
corradio, line 69: Could you cast this value to a float?
I think `float(price.p.text)` should suffice.

corradio (commented):
Thanks again for working on this!

parsers/IL.py
yuvalabou: No asthe price determined yearly, and not by "date"

corradio: I see. Would it be possible to extract the "valid from" date?
That would be the most correct way to model the data

image

parsers/IL.py
```diff
@@ -55,14 +55,32 @@ def flatten_list(_2d_list) -> list:
...
 
+def fetch_price(zone_key="IL", session=None, target_datetime=None, logger=None) -> dict:
+    """Fetch price from IEC table."""
+    if target_datetime is not None:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+    response = get(IEC_PRICE)
+    soup = BeautifulSoup(response.content, "lxml")
+
+    price = soup.find("td", class_="ms-rteTableEvenCol-6")
```
skovhus, line 66: Can't we find a more stable selector than this? This is very fragile. As mentioned in the other PR, it would be great if we could stabilize the selectors.

parsers/IL.py
yuvalabou: Probably, and insert it in place of the arrow? How do you suggest implementing it?

skovhus: You can use `arrow.get` to parse the date. See more in their documentation: https://arrow.readthedocs.io/en/latest/

parsers/IL.py
yuvalabou: This is actually a very good selector, but as in any web scraping apllications its as stable as the site itself..
I am using it in my HomeAssistant as my source of price in the past years and it hasn't been changed

skovhus: I've looked into the page, and I see that we cannot do much unfortunately.

There are 54 elements on the page with the class `ms-rteTableEvenCol-6`. Are we just interested in the first one?

yuvalabou: Well, it depends..

The first one is what most people will pay (e.g. households, agricultural)
however - in israel the tariff is decided by both the consumer - if he wants a load based or fixed, and also by type (e.g. factories vs housholds)

I have picked first one as it seems the most logical choice for me. you can however vote against :)

parsers/IL.py
yuvalabou: Done

parsers/IL.py
```diff
@@ -55,14 +55,46 @@ def flatten_list(_2d_list) -> list:
     return flatten_list(cleaned_list)
 
 
+def fetch_price_date():
+    """Fetch updated price date."""
+    response = get(IEC_PRICE)
+    soup = BeautifulSoup(response.content, "lxml")
+
+    date_str = soup.find("span", lang="HE").text
+    date_str = date_str.split(sep=" - ")
```
corradio, line 64: It seems like this method does another call to the URL `IEC_PRICE`.
Why not have it pass the `soup` object from L78?
Then you only do a single URL call.
```suggestion
def parse_price_date_from_soup(soup):
    """Fetch updated price date."""

    date_str = soup.find("span", lang="HE").text
    date_str = date_str.split(sep=" - ")
```

corradio (commented):
Minor idea to make the parser slightly faster by having only one HTTP call instead of two.

parsers/IL.py
yuvalabou: Did something a bit different but I hope that is fine. did not figure out how to pass that argument and still call the function once..

corradio (approved):
I did some last minor fixes here:
https://github.com/tmrowco/electricitymap-contrib/pull/3123/commits/e79841e10b6e0546b454d99163bb08d80a962e97

Looking good to me!

PR #3132, opened 2021-05-06
Fix iOS layout regression

skovhus:
#3034

https://github.com/tmrowco/electricitymap-contrib/pull/3109 introduced a regression. I believe this issue is related to trying to avoid the padding on larger screens. but the `@include respond-to('small')` selector is scss and needs a preprocessor.

I'll look into adding a proper media query, but it could be done in another PR.

### Before this PR

Screenshot 2021-05-06 at 21 18 24

### After this PR

For some reason we see a lot of 500 internal server errors when developing locally for `v3/state`. So I couldn't get a nice screenshot 🤔 

Screenshot 2021-05-06 at 21 29 01
 Screenshot 2021-05-06 at 21 27 21

### DIFF
--- a/mobileapp/.gitignore
+++ b/mobileapp/.gitignore
@@ -6,3 +6,9 @@ plugins
 platforms
 
 www/electricitymap
+
+src
+ul_web_hooks/
+GoogleService-Info.plist
+google-services.json
+locales-config.json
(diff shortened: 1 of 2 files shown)


corradio:
The v3/state error comes from the fact that one needs to build the JS using
```bash
docker-compose build --build-arg ELECTRICITYMAP_PUBLIC_TOKEN='......' web && ./build.sh
```
as it runs v3/state against our production server.

madsnedergaard (approved):

PR #3136, opened 2021-05-07
Update CH discharge values

### LINKED ISSUE #2932: Calculate Swiss hydro storage CO2 intensity
Switzerland uses the global average CO2 intensity. But since #2931 Switzerland has a reliable CO2 estimation, so the infamous "Average by Tomorrow" needs to be calculated.

image

martincollignon:
fixes #2932
Based on computed annual averages

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -214,6 +214,16 @@
           "source": "assumes 64% hydro, 13% hydro storage, 12% unknown thermal, 7% solar, 3% unknown renewable",
           "value": 131
         }
+        ,
+        "hydro discharge": {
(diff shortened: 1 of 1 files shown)


FelixDQ (approved):

veqtrus:
Wouldn't the intensity of 'unknown' also need to be updated?

matthme:
> 
> 
> Wouldn't the intensity of 'unknown' also need to be updated?

I agree. Although it is not clear on what emission factors for pumped storage #2898 relies upon but it would probably be nice to repeat the procedure and calculate the emission factor of "unknown" with the new pumped storage emission factor of 87 g/kWh @Kikof2.

PR #3141, opened 2021-05-09
Split AR parser

yuvalabou:
The AR parser had a lot of values which made scrolling to the actual code very long.
this is also a suggenstion to move more constants like this to the separate folder.


skovhus:
Thanks @yuvalabou. Sorry for the late reply. Your proposal makes sense. We will likely reconsider how to hide away parser constants and use your proposal as inspiration.

But for now we would rather stick with one file per parser to be consistent with the rest of the codebase. Consistency makes a lot of things easier for us right now.

PR #3143, opened 2021-05-10
refactor CR Parser

yuvalabou:
Updated capacities
Split file to const and parser


skovhus:
Thanks @yuvalabou. Sorry for the late reply. Your proposal makes sense. We will likely reconsider how to hide away parser constants and use your proposal as inspiration.

But for now we would rather stick with one file per parser to be consistent with the rest of the codebase. Consistency makes a lot of things easier for us right now. 

It would be great if you could create a self-contained PR for the capacity changes.

PR #3144, opened 2021-05-10
Change "API" to "Get our data"

martincollignon:
Very few people among our potential stakeholders actually understand what an API is, and we also provide data through CSVs.


madsnedergaard (commented):
Makes sense to me, did you talk/clear with @Kongkille as well? Also remember to update it on our internal pages (shared link in DM on Slack)

PR #3146, opened 2021-05-10
Require `solar` to be present in DE

corradio:
Missing solar data causes some discontinuities in the data.
I therefore suggest that we make solar required, which will reject points missing solar, and thus will not show the data while it lacks solar.

image


jarek:
When it's missing, is it 0 or null? That is, can we distinguish between missing and nighttime?

madsnedergaard (approved):
LGTM, waiting with merging this until next week so we can monitor it and ensure it looks correct :)

PR #3149, opened 2021-05-12
update generation capacities Argentina

KabelWlan:
I updated the generation capacities. I used the following source: https://cammesaweb.cammesa.com/informe-sintesis-mensual/

Also, I added the source to data_sourcesmd.


madsnedergaard (approved):
Hurray, oil going down and renewables going up!

yay renewables

PR #3150, opened 2021-05-12
Use ISO 639-1 language codes

veqtrus:
Closes #2737

It turns out that only Korean had to be renamed from `kr` to `ko`. The `locales` variable in the mobile app was outdated. I'm not sure if it's actually used.


madsnedergaard (approved):
Nice, thanks for doing this :)

Issue #3129, opened 2021-05-04
Wrong installed capacity for wind power in Sweden


swakter:
Installed capacity for wind power in Sweden should be 10 GW (or more). Today it shows as 8.98 GW which is outdated and means the capacity factor is not correct.

Information about 10 GW is from Swedish wind energy organisation, last updated February 2021:
https://svenskvindenergi.org/wp-content/uploads/2021/02/Q4-2020-Statistics-and-forecast-Svensk-Vindenergi-2020-02-08-FINAL.pdf

KabelWlan:
Hi @swakter 
Thanks for bringing this up. The Q1 2021 report should be released any day. Normally they release it 5-7 weeks after the end of the quarter. I‘ll change the capacity as soon as the newest data is available.

PR #3154, opened 2021-05-13
update Sweden wind capacity

KabelWlan:
solves #3129 

I updated the wind power capacity according to the newest report and I updated the data source. (https://svenskvindenergi.org/wp-content/uploads/2021/05/Q1-2021-Statistics-and-forecast-Svensk-Vindenergi-210511-1.pdf)


madsnedergaard (approved):
Brilliant, thanks for following up on that! 💪

PR #3155, opened 2021-05-14
CY: Improve biomass estimate

veqtrus:
Instead of just taking the reported generation at midnight, we now consider biomass+solar generation between 10pm and 3am as entirely biomass, and use the latest estimate for the rest of the day.


veqtrus:
@madsnedergaard could you rerun the checks? The labeler got stuck.

parsers/CY.py
```diff
@@ -63,17 +63,16 @@ def parse_production(self, html, capacity: dict) -> list:
...
-                    else:
-                        production['biomass'] = midnight_biomass
-                        production['solar'] = float(val) - midnight_biomass
+                    # Because solar is explicitly listed as "Solar PV" (so no thermal with energy storage)
+                    # and there is no sunlight between 10pm and 3am (https://www.timeanddate.com/sun/cyprus/nicosia),
+                    # we use the nightly biomass+solar generation reported to determine the portion of biomass+solar
+                    # which constitutes biomass.
+                    value = float(val)
+                    hour = datum['datetime'].hour
+                    if hour < 3 or hour >= 22:
```
madsnedergaard, line 72: Do you know how the source handles daylight saving time? :)

madsnedergaard (commented):
I have one minor comment about DST, but otherwise it looks good!
When running the test-parser it seemed like I got identical data from before this change, did I miss something? :)

veqtrus:
> When running the test-parser it seemed like I got identical data from before this change, did I miss something? :)

Most of the time biomass is static. However occasionally it is higher at midnight which can cause the solar estimate to go negative.

parsers/CY.py
veqtrus: The source has table cells like `2021-05-26 00:00:00` which I suppose include DST. We set the timezone to Asia/Nicosia so I suppose `arrow` will take care of it.

madsnedergaard (approved):
Thanks for the explanation! Let's merge it 💪

PR #3156, opened 2021-05-16
update Norway capacities

KabelWlan:
I updated the generation capacities of Norway.  Also I updated the data source. I used the same source for all the changes. (https://www.nve.no/energiforsyning/kraftproduksjon/?ref=mainmenu)

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -292,10 +292,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Nicaragua: [Climatescope](http://global-climatescope.org/en/country/nicaragua/)
 - Nigeria: [SO Grid](https://www.niggrid.org)
 - North Macedonia: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
-- Norway
-  - Hydro: [NVE](https://www.nve.no/energiforsyning/kraftproduksjon/vannkraft/vannkraftdatabase/)
-  - Wind: [NVE](https://www.nve.no/energiforsyning/kraftproduksjon/vindkraft/vindkraftdata/)
-  - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
+- Norway: [NVE](https://www.nve.no/energiforsyning/kraftproduksjon/?ref=mainmenu)
 - Northern Ireland: [ENTSO-E](https://m-transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show)
 - Oman: [IRENA](https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf)
 - Panama: [Secretaría de Energía de Panamá](http://www.energia.gob.pa/mercado-energetico/?tag=84#documents-list)
--- a/config/zones.json
+++ b/config/zones.json
@@ -3878,14 +3878,17 @@
   },
   "NO": {
     "capacity": {
+      "biomass": 113.3,
       "coal": 0,
-      "gas": 634,
+      "gas": 473.4,
       "geothermal": 0,
-      "hydro": 32440,
+      "hydro": 33033,
+      "hydro storage": 1438,
       "nuclear": 0,
       "oil": 0,
-      "solar": 0,
-      "wind": 3644
+      "solar": 160,
+      "unknown": 112.6,
+      "wind": 3977
     },
     "contributors": [
       "https://github.com/corradio"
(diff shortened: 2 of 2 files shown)


madsnedergaard:
Where did you find regional data on biomass? Maybe it's just me, but I find that website pretty confusing :D

KabelWlan:
I found them on this page (https://www.nve.no/energiforsyning/kraftproduksjon/termisk-kraft/?ref=mainmenu) on the right hand side. It's an Excel file called "Oversikt over termiske kraftverk"

I sorted them the following way:
Avfallsforbrenning and Biogass -> biomass
Naturgass -> gas
CO gass and Varmegjenvinning -> unknown

madsnedergaard (approved):
Nice!

madsnedergaard:
> I found them on this page ([nve.no/energiforsyning/kraftproduksjon/termisk-kraft/?ref=mainmenu](https://www.nve.no/energiforsyning/kraftproduksjon/termisk-kraft/?ref=mainmenu)) on the right hand side. It's an Excel file called "Oversikt over termiske kraftverk"
> 
> I sorted them the following way:
> Avfallsforbrenning and Biogass -> biomass
> Naturgass -> gas
> CO gass and Varmegjenvinning -> unknown

Ah great, thanks for sharing - I missed that link on the right :)

Gazer75:
How was pumped storage calculated for this? Did you go through the list and manually add up the ones with reversible turbines?

KabelWlan:
I exported the data as an excel sheet and then I sorted it by hydro and pumped hydro. In NO3 I found that there are 9 pumped hydro stations listed with a capacity of 87.4 MW in total. 

You can check it for youself. If I missed or missunderstood something in that excel sheet, please let me know.

Gazer75:
You must have included the pure pumping stations ("type" column as "pumpe") that never produce anything, but rather consume power to pump if needed.
Those stations will have a negative value in the "Maks ytelse [MW]" column.

The "type" you have to look at is "pumpekraftverk" only.

KabelWlan:
Yes I did because they're part of the pumped hydro category in my opinion.

Why would you have a power station that pumps water uphill and never produces anything? To me this does not make much sense. 
I guess that the difference between "pumpe" and "pumpekraftverk" is that the facilites listed as "pumpe" don't have a natural water inflow because they don't produce any power, they have negative production values. The stations in the category "pumpekraftverk" acutally produce power over the year which means the need to get that water from somewhere like a bigger river. 

You seem to be more familiar with the system in Norway so correct me if I'm wrong.

Let me know what you think and keep me updated on the messae you sent to Stanett and ENTSOE.

Gazer75:
They are not power stations, they are pumping stations, They never produce any power. They have no generators.
The category "pumpekraftverk" have reversible Francis turbines.

Hydro pumped storage plants in Norway are not like the more common artificial ones where they pump at night and produce during the day. They mostly produce, and might pump if a lower, and often smaller, reservoir is full.
Reason for this is the price difference between day and night is usually not big enough.
The price variation is more tied to the seasons here.

The pumping stations basically pump water up from one reservoir to another so that it can be used by a power station. They can be several kilometers away. It's a way to get more water to the main reservoir basically.

Example:
Breive pumping station is the largest one with 8 2.5MW pumps spending on average 44GWh per year pumping water from Breievatn/Hartevatn at ~758m ASL to Store Førsvatn at 836-843m ASL, it then flows through a tunnel down to Vatndalsvatnet at 700-840m ASL.
This water is then used 25km away from Breive by Holen power station turbine #3 which is a 160MW Francis and producing on average 246GWh per year.

The whole stored hydro power system in Norway is a big web of reservoirs and tunnels, with often a string of power stations using the water several times before it reaches the ocean.

KabelWlan:
Thanks for your clarification and explanation. I obviously don't know the system well enough. I'm used to the "conventional" pumped hydro stations like you said pump at night and produce during the day. 

However, I need to correct all the zones in Norway since I accounted "pumpe" in the others as well.

NO1: 0 MW
NO2: 840 MW
NO3: 60.3MW
NO4: 0 MW
NO5: 444 MW
Norway total: 1344.3

This should be the correct capacities with only "pumpekraftverk" included.

I'm still curious what ENTSOE and Statnett say to this issue that NO3 (and btw NO2 as well) report more generation than capacity available. 
ENTSOE hasn't listed any pumped hydro capacity in Norways zones anyway so there must be something wrong.

Gazer75:
There are quite a few errors on what is reported to ENTSOE and the actual value. I think it was indicated to me that power stations below 10MW are not listed in total capacity to ENTSOE as it is not required or something like that.
From what I've understood there are limitations in the data systems they use that prevents them from doing more detailed reporting, and since its not required its far down the list of priority.

And to complicate it even more, Saurdal power station have 4 160MW Francis turbines, but only two of them are reversible.
Not sure if the code support different negative value for the pumped hydro section.
For NO2 it should be +840MW and -640MW.
Doesn't really matter anyway as I've not found the consumption of pumping being reported to ENTSOE.

KabelWlan:
Do you have any news from ENTSOE or Statnett yet?

Gazer75:
No news, and I do not expect much tbh. Last time I reported errors to ENTSOE I only got a reply that they would look into it and never heard back. That reply came many weeks after I sent the email.

PR #3157, opened 2021-05-16
Fix IN-HP parser

willbeaufoy:
The parser was broken because the page has changed. It now loads the actual
data from a separate URL. Therefore I have fixed the parser by changing it
to point to this new URL instead of the original page.

Also did some minor formatting changes.

### DIFF
--- a/parsers/IN_HP.py
+++ b/parsers/IN_HP.py
@@ -10,16 +10,20 @@ import arrow
 import requests
 from bs4 import BeautifulSoup
 
-
-DATA_URL = 'https://hpsldc.com/intra-state-power-transaction/'
+# This URL is called from within the
+# https://hpsldc.com/intra-state-power-transaction/
+# page to load the data.
+DATA_URL = 'https://hpsldc.com/wp-admin/admin-ajax.php'
 ZONE_KEY = 'IN-HP'
 TZ = 'Asia/Kolkata'
 
 
 class GenType(Enum):
     """Enum for plant generation types found in the data.
+
     Enum values are the keys for the production dictionary returned
-    from fetch_production()."""
+    from fetch_production().
+    """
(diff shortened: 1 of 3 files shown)


madsnedergaard:
Nice! Does running the `test-parser` script work for you? It's timing out for me, but calling the `DATA_URL` directly seems to work 🤔 
I have pretty flaky internet right now though, so I'll retry later today.

Screenshot 2021-05-17 at 08 06 12


Screenshot 2021-05-17 at 08 07 47

willbeaufoy:
> Nice! Does running the `test-parser` script work for you? It's timing out for me, but calling the `DATA_URL` directly seems to work thinking
> I have pretty flaky internet right now though, so I'll retry later today.
> 
> Screenshot 2021-05-17 at 08 06 12
> 
> Screenshot 2021-05-17 at 08 07 47

Hi, yes it does work for me, but the request does take about 15 seconds to return. I use the command `poetry run test-parser IN-HP production`.

Kongkille:
The data page also takes approximately 30 seconds or so to load for me. As such, it might time out sometimes, however, we are looking into our infrastructure to allow some parser additional time :)

willbeaufoy:
> The data page also takes approximately 30 seconds or so to load for me. As such, it might time out sometimes, however, we are looking into our infrastructure to allow some parser additional time :)

Oh ok, does this mean this solution will work generally, but fail sometimes? Or is there a larger problem.

madsnedergaard (approved):
Works for me now (think it was just a bad connection) 👍 

> Oh ok, does this mean this solution will work generally, but fail sometimes? Or is there a larger problem.

Yes, it will generally work - and this should be within limits most of the time, as I think the limit is a few minutes :)

Issue #3158, opened 2021-05-17
US-SW-PNM breaks EIA parser with None hydro


KabelWlan:
US-SW-PNM does not display any data. 
I guess the problem is the negative value for solar which is reported from EIA during some time of the night. I think this is going on since a few days already because yesterday was not the only day the value for solar was negative.

![negative solar](https://user-images.githubusercontent.com/49598085/118535736-1f455600-b74b-11eb-9eed-21376a0b5938.png)

link to EIA: https://www.eia.gov/electricity/gridmonitor/dashboard/electric_overview/balancing_authority/PNM
link to the zone: https://www.electricitymap.org/zone/US-SW-PNM

jarek:
Looks like we need to amend the parser:

https://kibana.electricitymap.org/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-4h,mode:quick,to:now))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:pnm),sort:!('@timestamp',desc))

```
Traceback (most recent call last):
  File "/home/projects/feeder-electricity/src/lib/fetch_data.py", line 203, in launch_parsers
    objs = res.get(timeout) or []
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/projects/feeder-electricity/src/lib/fetch_data.py", line 199, in <lambda>
    *keys, session, logger=public_logger, **parser_kwargs
  File "/home/src/electricitymap/contrib/parsers/EIA.py", line 421, in fetch_production_mix
    if type == 'hydro' and point['value'] < 0:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

jakedorne:
Hiya, am I able to pick this one up? 😄

PR #3165, opened 2021-05-24
Handle None hydro value for EIA parser

jakedorne:
Simple check for None values to stop the EIA parser from breaking.

It looks like there aren't any existing tests for this parser so I wasn't sure if I should be adding a regression test?

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -418,7 +418,7 @@ def fetch_production_mix(zone_key, session=None, target_datetime=None, logger=No
                 0 > point['value'] >= negative_threshold:
                 point['value'] = 0
 
-            if type == 'hydro' and point['value'] < 0:
+            if type == 'hydro' and point['value'] and point['value'] < 0:
                 point.update({
                     'production': {},# required by merge_production_outputs()
                     'storage': {type: point.pop('value')},


parsers/EIA.py
```diff
@@ -418,7 +418,7 @@ def fetch_production_mix(zone_key, session=None, target_datetime=None, logger=No
                 0 > point['value'] >= negative_threshold:
                 point['value'] = 0
 
-            if type == 'hydro' and point['value'] < 0:
+            if type == 'hydro' and point['value'] and point['value'] < 0:
```
corradio, line 421: Is it possible to distinguish, in the raw data, between pumped hydro and run-of-the-river hydro?
Right now it seems that we interpret positive values as run-of-the-river, and negative values as storage, which causes us to never register the times at which pumped hydro generates electricity (we incorrectly interpret that situation as non-pumped hydro generation instead)

jakedorne: Hey, sorry for the late reply! To be honest, I'm super new to the project (this is my first PR) so I have no idea. I'll look into it when I get some time this week. 👍

jakedorne: Hey @corradio , as far as I can tell it's not possible to distinguish in the raw data, which just contains the "Net generation from hydro for x region".

Here's an example request:
https://api.eia.gov/series/?api_key=<EIA_API_KEY>&series_id=EBA.TEN-ALL.NG.WAT.H&num=24

Does this project distinguish between run-of-the-river and pumped hydro anywhere? From what I can see, negative values are interpreted as storage and positive values are interpreted as general "hydro".

corradio: Hi @jakedorne,
Thanks for investigating! This project indeed distinguishes between pumped hydro (which in electricityMap world is the `hydro` key under `storage`) and run-of-the-river (which is the `hydro` key under `production`).
If we only have access to the net generation, then I think we should put the net under `storage` (as it can become negative).
Does the EIA API allow us to separate run-of-the-river hydro and pumped hydro?

jakedorne: Hey @corradio, sorry, I think I misunderstood your original comment - I thought you meant there were a total of 3 distinctions (pumped, run-of-the-river, and storage).

I'm struggling to find any way to distinguish between pumped and run-of-the-river generation.

If there is no way to distinguish, I'm happy to edit the code to add the net generation under storage, but does that not mean we'd have the opposite problem and be interpreting all generation as as pumped?

Just to confirm: If I can find a way to distinguish, should all pumped generation be added to production and all run-of-the-river be added to storage?

Again, sorry for the late reply. Busy times. 😞

corradio: > If there is no way to distinguish, I'm happy to edit the code to add the net generation under storage, but does that not mean we'd have the opposite problem and be interpreting all generation as as pumped?

Correct. I suggest we leave it like that for now.

> Just to confirm: If I can find a way to distinguish, should all pumped generation be added to production and all run-of-the-river be added to storage?

If we can find a way to distinguish, pumped generation (which can become negative) should be added to storage, and run-of-the-river (which can never be negative) added to production.

> Again, sorry for the late reply. Busy times. 😞

No worries! Thanks again

corradio (approved):
Thanks!

corradio:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/3158

PR #3166, opened 2021-05-26
update USA capacites Q1

KabelWlan:
I updated the capacities of the USA with the same source as before. EIA publishes the new data on a monthly basis with a 3 months delay. I'll do an update every quarter since it is a lot of work going through all the balacing authorities and most of the USA isn't colored at the moment anyway. 

Also I added Puerto Ricos capacities. They're included in the same report.


brandongalbraith:
@KabelWlan You might consider updating /DATA_SOURCES.md as well to reflect the data source URL referenced for updating US balancing authority/ISO production capacity data.

KabelWlan:
I updated the source now. When I first implemented that report as a data source I listed it under "Federal". I changed it to "Balancing Authorities" now which hopefully makes things clearer.

skovhus (approved):
Thank you so much for updating this

PR #3167, opened 2021-05-27
Update id.json

Jironah:


web/locales/id.json
```diff
@@ -562,31 +562,31 @@
...
-            "countryName": "France",
+            "countryName": "Perancis",
             "zoneName": "Corsica"
         },
         "GA": {
             "zoneName": "Gabon"
         },
         "GB": {
-            "zoneName": "Britania Raya"
+            "zoneName": "Inggris Raya"
```
jarek, line 575: Hello, thank you for the changes! This seems to reference "England" more than "Britain" - is that the right reference? The wikipedia article https://id.wikipedia.org/wiki/Britania_Raya uses the Britania Raya name and seems to list Inggris, Irlandia Utara, Skotlandia, and Wales as constituents, so Inggris seems to be England?

Jironah: Yup. I think that's right. 
Thank's for the correction.

skovhus: @jarek thanks for the careful review. 👏

skovhus (approved):
Thanks for fixing this!

PR #3169, opened 2021-06-02
Removes old CL-SIC zone

madsnedergaard:
## Description

This zone is no longer in use and was replaced a while ago by a new parser source in https://github.com/tmrowco/electricitymap-contrib/pull/2454.

Therefore we're removing it to reduce confusing and simplify code :)

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1292,20 +1292,6 @@
     ],
     "timezone": "America/Punta_Arenas"
   },
-  "CL-SIC": {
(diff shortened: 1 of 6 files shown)


config/zones.json
```diff
@@ -1292,20 +1292,6 @@
     ],
     "timezone": "America/Punta_Arenas"
   },
-  "CL-SIC": {
```
pierresegonne, line 1295: como dicen ahi, Adios!

pierresegonne (approved):

skovhus (approved):
Tak for nu. 👋

PR #3173, opened 2021-06-05
change CA-NB unknown emission factor

KabelWlan:
I added/changed the emission factor of "unknown" for New Brunswick to 223. 
The data for the calculation is from the New Brunswicks Provincial and Territorial Energy Profile (https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-new-brunswick.html).
I hope it can be colored on the map now too.

### DIFF
--- a/config/co2eq_parameters.json
+++ b/config/co2eq_parameters.json
@@ -184,6 +184,13 @@
           "value": 597.2654373454017
         }
       },
+      "CA-NB": {
+        "unknown": {
+          "_comment": "Source: https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-new-brunswick.html",
+          "source": "2018 Provincial and Territorial Energy Profiles - New Brunswick",
+          "value": 223
+        }
+      },
       "CA-NS": {
         "hydro discharge": {
           "source": "2017 average by Tomorrow",


jarek:
Just to note, the Canada Energy Regulator data (previously NEB) is already used for exchange intensity:

https://github.com/tmrowco/electricitymap-contrib/blob/df4f6fac3a6fa237857a3a29c3cbac252532500a/config/co2eq_parameters.json#L1230-L1240

The Electricity Map server calculates the intensity as 223.79000000000002 and the website shows 224 g CO2eq/kWh, so rounding up to 224 g here might be worth it for consistency of display.

> I hope it can be colored on the map now too.

Hm, I know Electricity Map has started doing this for other regions, but I'm not sure it's appropriate.

It's one thing for places like Singapore to almost always be the same colour -- we have real time data there and it says that their generation is basically all from one source. For New Brunswick, it would be somewhat misleading since New Brunswick's fossil fuel production varies on an hourly or daily scale while the nuclear production doesn't (weekly/monthly can differ when Point Lepreau has [a "small fire"](https://www.nbpower.com/en/about-us/news-media-centre/news/2021/point-lepreau-nuclear-generating-station-update-february-2021/) I guess).

You can take a look at [Statcan data table 25-10-0015-01](https://www150.statcan.gc.ca/t1/tbl1/en/tv.action?pid=2510001501&pickMembers%5B0%5D=1.5&pickMembers%5B1%5D=2.1&cubeTimeFrame.startMonth=01&cubeTimeFrame.startYear=2018&cubeTimeFrame.endMonth=12&cubeTimeFrame.endYear=2021&referencePeriods=20180101%2C20211201) to get an idea how much the production types jump around month to month, even though that data doesn't break down fossil fuel type. Compare April 2020: hydro 518 GWh, nuclear 478 GWh, fossil 155 GWh, wind 86 GWh with October 2020: hydro 248 GWh, nuclear 0, fossil 443 GWh, wind 82 GWh, plus likely import/export changes.

Historically, between 2013 and 2016, monthly average carbon intensity jumped around from usual high of ~400 g CO2eq/kWh to usual low of ~300, peaking above 500 a few times when Point Lepreau was down. That was with Point Lepreau being less reliable than now, but even so, putting in one figure of 223 or 224 feels a bit misleadingly precise. The data isn't there to give a number with three significant digits. It's useful to provide an estimate for exchanges since otherwise the exchange data would be pretty wrong, but I'm not sure it warrants being on the map, let alone with 223 g.

(For historical record, in 2018 we set the NB export intensity to 300 g in #1258 based on my calculation from 2017 detailed in [my post about then-available data in the region](https://piorkowski.ca/rev/2017/06/canadian-electricity-co2-intensities/))

KabelWlan:
Thank you for your comprehensive feedback @jarek. I thought, if Israel or South Corea with so little information in fule type breakdown given, are colored on the map, New Brunswick could be too. 

It's best to close the PR at this point.

PR #3175, opened 2021-06-06
add North Sea Link capacity

KabelWlan:
I added the capacity and source of the North Sea Link interconnector.


skovhus:
Thanks! 

From what I can read, it is still under construction:
- https://northsealink.com/
- https://en.wikipedia.org/wiki/North_Sea_Link

KabelWlan:
Yes you're right, the interconnector isn't commissioned yet. 

I thought since the arrow is already showing an electricity flow on the map, I could add the capacity. We can also wait with merging this PR until commissioning or the test operation started.

skovhus:
Adding the capacity makes sense. 👍

skovhus (approved):

PR #3182, opened 2021-06-14
Ensure all zones are defined in zones.json

skovhus:
This ensures that:
- all zones defined in exchanges.json has an entry in zones.json
- all sub zones also have a top level entry in zones.json

I made one exception in the tests for the old US zones.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -545,6 +545,9 @@
     },
     "timezone": "Europe/Mariehamn"
   },
+  "AZ": {
+    "timezone": null
+  },
   "BA": {
     "bounding_box": [
       [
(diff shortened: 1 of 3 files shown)


FelixDQ (approved):
👍 Makes sense!

I'll just link this issue https://github.com/tmrowco/electricitymap/issues/34 so that we remember to remove the special US case when we remove the old us zones.


Also, it might be worth running [generate-zone-bounding-boxes.js](https://github.com/tmrowco/electricitymap-contrib/blob/master/web/generate-zone-bounding-boxes.js) to add bounding boxes to the zones.

PR #3185, opened 2021-06-19
Updated FI Nuclear

nessie2013:
Finland has just had another nuclear plant go online, so I found an updated capacity. Updated source


veqtrus:
Commercial production is scheduled for February 2022.

skovhus (approved):
Thanks for updating this!

PR #3187, opened 2021-06-20
Change colorblind-friendly gradient to less intense colors

veqtrus:
In #2718 I changed the colours to have a higher luminance range, however I realised that the shade of red I chose is too intense. Additionally since then the carbon intensity text has been made to switch from white to black depending on the background, which allows using lighter colours without affecting readability. After trying a lot of combinations I came up with the following gradient:

`#FFFFB0` at 0g (0%), `#E0B040` at 200g (25%), `#A06030` at 400g (50%), `#602020` at 600g (75%), `#000010` at 800g (100%)

This is what it would look like using test data:

light-mode

dark-mode


madsnedergaard (approved):
Awesome! 💪 
As a non-color-blind it's kinda hard to test this, but using http://colororacle.org the colors still looks very "readable" :)

PR #3189, opened 2021-06-22
update Alberta to new categories available

KabelWlan:
Today, AESO, Albertas grid operator published their updated "Supply and Demand" page. They added solar, energy storage and dual fuel to the existing categories.
What I tried is to implement those changes to the praser. I hope I did it correct, it's my first time.

Explanation to what I did:
I added Energy storage and solar to the existing categories. 
I changed the categorie "other" to biomass since all the facilities are biomass or waste heat power plants.
Dual Fuel is under "unknown" so that we can change the emission factor according to the mix of coal and gas which the power plants use. Battle River 4 and 5 use 50% coal and gas. (https://en.wikipedia.org/wiki/Battle_River_Generating_Station)
I didn't find anything about Sheerness 2 but according to some sources (https://www.gem.wiki/Sheerness_power_station#cite_note-3) work is still ongoing to change to natural gas. 
I'd suggest using a emission factor of 655 for "unknown" which would represent 50% gas and 50% coal.

Also I'll add the emission factor for battery storage of 597. I used the following data for my calculation: https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-alberta.html

Maybe you can calculate the "2020 average by Tomorrow"? Would probably be more accurate than the numbers from 2018.


brandongalbraith:
@KabelWlan Might be a good time to check if the capacity for this zone in zones.conf needs to be updated as well.

jarek:
> Might be a good time to check if the capacity for this zone in zones.conf needs to be updated as well.

Alberta does not have a hardcoded capacity in zones.json, the capacities are reported by the parser based on the "maximum capability" data in the source page.

https://github.com/tmrowco/electricitymap-contrib/blob/a611d4dfd89f32de54de1d9f70f47215957bf1ac/config/zones.json#L909-L929

skovhus (commented):
I've double checked that ENERGY STORAGE doesn't include other types of storage. From this report is seems to be only battery: https://www.aeso.ca/assets/Uploads/Energy-Storage-Roadmap-Report.pdf

@corradio can you check the emission factors and biomass/unknown change?

config/co2eq_parameters.json
```diff
@@ -179,9 +179,14 @@
...
-        "hydro discharge": {
-          "source": "2017 average by Tomorrow",
-          "value": 597.2654373454017
+        "unknown": {
+          "_comment": "dual fuel power plants: assumption of 50% coal and 50% natural gas",
+          "value": 655
+        }
+        ,
+        "battery discharge": {
+          "source": "https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-alberta.html",
```
corradio, line 188: Isn't the source for the 597 figure still the "2017 average by Tomorrow"?

KabelWlan: Yes it is. Is there any chance that you could calculate the 2020 average? Since 2017 a few coal power plants switched to gas or were shut down and I guess it‘s lower now. Probably more like 500.

corradio: > Yes it is. Is there any chance that you could calculate the 2020 average? Since 2017 a few coal power plants switched to gas or were shut down and I guess it‘s lower now. Probably more like 500.
@skovhus should we create an internal issue on our side?

Aside from that, I suggest we change the source so it properly reflects the origin of the 597 figure:
```suggestion
          "source": "2017 average by Tomorrow",
```

If we do any updates, we will probably make a new PR

jasonrwang:
FYI @skovhus there is an announced 100 MW pumped hydro project for Alberta, which will still take a few years to build and commission. Just wanted to flag so current implementation can be robust :)

skovhus (approved):

pbubik:
Can we please make the following changes in CA-AB?

1. Unknown should describe source as "Dual fuel, coal/gas assumes 50/50"
2. Hydro Storage (this alone is an unfortunate name - should be Pumped Storage - to differentiate from storage in large hydro reservoirs that is not pumped) - should read 0MW (rather than "?") until Canyon Creek is commissioned. It IS 0MW in Alberta at this time.
3. Battery Storage of 597g/kWh isn't a good estimate. The mix has changed significantly since 2017. Coal was reduced, wind and solar was added. Can we use 2020 average GHGs? They could be teased out of this report: https://aeso.ca/assets/Uploads/2020-Annual-Market-Stats-Final.pdf page 16, figure 16 available here: https://aeso.ca/market/market-and-system-reporting/annual-market-statistic-reports/ 
- Coal 39% of energy, gas 46%, wind, 11%, hydro+solar 3%... ergo 2020 average is 547g/kWh
4. Nuclear shoud read 0MW rather than "?" - there is 0MW capacity. Source is http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet?beginDate=&endDate=&contentType=html
5. Geothermal should read 0MW rather than  "?" - there is 0MW capacity. Source is http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet?beginDate=&endDate=&contentType=html
6. Oil should read 0MW rather than  "?" - there is 0MW capacity. Source is http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet?beginDate=&endDate=&contentType=html

Thank  you!

skovhus:
@pbubik it would be awesome if you could create a new GitHub issue for this so we can track this. Thanks. :)

PR #3192, opened 2021-06-24
Strict configuration system using Pydantic

skovhus:
JSON files are great for contributions, but while consuming the data we could benefit from a more type safe system that would:
- catch errors before they reach production (e.g. validate that capacities should be a non negative int)
- help us while developing (e.g. editor autocompletion)
- add stronger guarantees when refactoring code

This PR is an experiment to [Pydantic](https://pydantic-docs.helpmanual.io/), a very solid and "Data validation and settings management library using python type annotations". The model/interface that we expose here is optional and doesn't affect contributors. 

But I believe the Pydantic models will:
- replace the JSON schemas (unless they help someone when editing the JSON files – in that case they can be generated by Pydantic)
- eventually will be the main interface for the data

A quick experiment shows that it takes 20 ms to load all the configuration, validate them and expose them as a class based model (e.g. `config_model.zones["US-NW-PSCO"].parsers.get_function("production")`). 

### Experiments

<details>
  <summary>Some benchmarks</summary>
  
```python

if __name__ == "__main__":
    # a few tests to validate how fast it is to read from Pydantic compared
    # to a simple JSON file.
    import time
    import pydantic

    assert pydantic.compiled

    class Timer(object):
        def __init__(self, name: str):
            self.name = name

        def __enter__(self):
            self.t_start = time.time()

        def __exit__(self, exc_type, exc_value, tb):
            assert exc_type is None
            assert exc_value is None
            print(f"{self.name} time", 1000 * (time.time() - self.t_start), "ms")

    with open("config/zones.json") as f:
        zone_json_str = f.read()

    with open("config/exchanges.json") as f:
        exchanges_json_str = f.read()

    print("\n1) JSON dict")
    with Timer("load time"):
        zones_json_dict = json.loads(zone_json_str)
        exchanges_json_dict = json.loads(exchanges_json_str)

    with Timer("access time"):
        bounding_boxes = [
            zone.get("bounding_box") for zone_key, zone in zones_json_dict.items()
        ]

    print("\n2) Pydantic object from dict")
    with Timer("load time"):
        config_model = ConfigModel(exchanges=exchanges_json_dict, zones=zones_json_dict)
    dk_dk1 = config_model.zones["DK-DK1"]
    with Timer("access time"):
        bounding_boxes = [
            zone.bounding_box for zone_key, zone in config_model.zones.items()
        ]

    config_model = load_model()

    print("\nSanity and testing")
    print("US-NW-PSCO", config_model.zones["US-NW-PSCO"])
    print(">>>", config_model.zones["US-NW-PSCO"].parsers.get_function("production"))

```
</details>

### DIFF
--- a/electricitymap/contrib/config/model.py
+++ b/electricitymap/contrib/config/model.py
@@ -1,0 +1,109 @@
+import json
+from typing import Callable, Dict, List, NewType, Tuple, Optional
+
+from pydantic import BaseModel, Field, NonNegativeInt, PositiveInt
+from pydantic.utils import import_string
+
+from electricitymap.contrib.config import EXCHANGES_CONFIG, ZONES_CONFIG
+
+# we could cast this to a NamedTuple with x/y accessors
+Point = NewType("Point", Tuple[float, float])
+
+
+class StrictBaseModel(BaseModel):
+    class Config:
+        extra = "forbid"
+
+
+class Capacity(StrictBaseModel):
+    # TODO: if zone.json used underscores for keys we didn't need the Field()
+    battery_storage: Optional[NonNegativeInt] = Field(None, alias="battery storage")
+    biomass: Optional[NonNegativeInt]
+    coal: Optional[NonNegativeInt]
+    gas: Optional[NonNegativeInt]
+    geothermal: Optional[NonNegativeInt]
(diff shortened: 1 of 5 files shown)


FelixDQ (approved):

madsnedergaard (approved):
Looks nice :)

Issue #3201, opened 2021-07-02
fix CA-AB


KabelWlan:
I must have made some mistake in the changes in #3189 because Alberta is down now.

This is the error message:
invalid point: {'datetime': datetime.datetime(2021, 7, 2, 9, 28, tzinfo=<DstTzInfo 'Canada/Mountain' MDT-1 day, 18:00:00 DST>), 'zoneKey': 'CA-AB', 'production': {'gas': 5378.0, 'hydro': 396.0, 'solar': 189.0, 'wind': 536.0, 'biomass': 272.0, 'unknown': 387.0, 'coal': 2530.0}, 'storage': {'battery storage': 0.0}, 'capacity': {'gas': 8378.0, 'hydro': 894.0, 'battery storage': 30.0, 'solar': 290.0, 'wind': 1988.0, 'biomass': 423.0, 'unknown': 940.0, 'coal': 3731.0}, 'source': 'ets.aeso.ca', 'schemaVersion': 3}, reason:unexpected keys in storage: {'battery storage'}

I'm sorry for the mistake...

jarek:
Should hopefully be fixed by https://github.com/tmrowco/electricitymap-contrib/pull/3202. Thank you for checking on it! The different key names between `storage` and `capacity` are a bit unintuitive, even though I get why they're different...

KabelWlan:
Thanks a lot! So I should have used „battery“ instead of „battery storage“. I‘ll keep learning and adding to my knowledge.

PR #3202, opened 2021-07-02
fix CA_AB battery key #3201

jarek:
Fixes https://github.com/tmrowco/electricitymap-contrib/issues/3201

Hopefully correct key name taken from https://github.com/tmrowco/electricitymap-contrib/blob/e608afbf1ae6bdfc978b738e175964aeb2cad8b0/parsers/OPENNEM.py#L188-L191

Also tiny indent fix


madsnedergaard (approved):
Thanks for following up on this and fixing it :)

PR #3215, opened 2021-07-13
Show error message when wind or solar is unavailable 

nickvallee:
## Introduction

Hi! This is my first time contributing to this project, so feedback and change requests are more than welcome! This PR is in reference to the [this issue](https://github.com/tmrowco/electricitymap-contrib/issues/3002). Since the issue's description is not verbose, I decided to just try something, but if we want to handle errors in a specific way, just let me know!

## What's Changed

- Wind and solar icons become red when data fetch returns an error.
- This is accompanied by a static error message in the tooltip, notifying that the data is unavailable.
- error message is stored in the reducer and erased when a new request is sent for the same data.
- added english error messages to the translation file.

Screen Shot 2021-07-13 at 5 06 13 PM

## Questions

- do we want a specific pattern for handling these errors? 
- should the error message be static, or should it display a specific message returned from the http request?


veqtrus:
Isn't the solar layer broken anyway?

solar layer

web/locales/en.json
```diff
@@ -109,7 +109,9 @@
         "faq": "Frequently Asked Questions"
     },
     "wind": "wind",
+    "windDataError": "Wind data is unavailable",
     "solar": "solar",
+    "solarDataError": "Solar data is unavailable",
```
madsnedergaard, line 114: Maybe emphasise that this is only temporarily unavailable:

```suggestion
    "windDataError": "Wind data is currently unavailable",
    "solar": "solar",
    "solarDataError": "Solar data is currently unavailable",
```

web/public/images/weather/sun_active_error.svg
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
+<!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Capa_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="612px" height="612px" viewBox="0 0 612 612" style="enable-background:new 0 0 612 612;" xml:space="preserve">
+<g>
+	<g id="Sun">
+		<g>
+			<path fill="red" d="M104.914,306c0-14.478-11.75-26.229-26.229-26.229H26.229C11.75,279.771,0,291.522,0,306
```
madsnedergaard, line 9: I think the red icon here is a bit intense and potentially distracting for people that don't want to use these layers - maybe it would be enough to just make the button disabled, so you only see the tooltip/"problem" if you actually want to enable the solar layer? :)

madsnedergaard (commented):
Hey Nick, thanks for the contributions, this is great! 💪 
I think this is overall pretty cool - I have added a few comments/suggestions below.

Regarding your questions:
> do we want a specific pattern for handling these errors?
I think this is fine for now :)

> should the error message be static, or should it display a specific message returned from the http request?
Static is also fine here, as it probably doesn't matter much to the user why the data is unavailable.

madsnedergaard:
> Isn't the solar layer broken anyway?

I don't think it is broken (at least not right now). Here's a comparison with https://en.sat24.com/en/eu/visual right now:

Screenshot 2021-07-15 at 16 49 02

veqtrus:
> > Isn't the solar layer broken anyway?
> 
> I don't think it is broken (at least not right now). Here's a comparison with https://en.sat24.com/en/eu/visual which shows pretty much the same for Europe:

It breaks when you zoom or move the map.

madsnedergaard:
> > > Isn't the solar layer broken anyway?
> > 
> > 
> > I don't think it is broken (at least not right now). Here's a comparison with [en.sat24.com/en/eu/visual](https://en.sat24.com/en/eu/visual) which shows pretty much the same for Europe:
> 
> It breaks when you zoom or move the map.

I have created a new issue for this: https://github.com/tmrowco/electricitymap-contrib/issues/3217

So we can keep this PR about the error messages :)

web/public/images/weather/sun_active_error.svg
nickvallee: Yeah,  that's a fair point 👍

web/src/layout/layerbuttons.js
```diff
@@ -35,13 +37,15 @@ export default () => {
...
         />
       </Link>
       <Link to={solarToggledLocation}>
         <ButtonToggle
           active={solarEnabled}
           tooltip={__(solarEnabled ? 'tooltips.hideSolarLayer' : 'tooltips.showSolarLayer')}
+          errorMessage={solarDataError}
           icon="weather/sun"
         />
       </Link>
```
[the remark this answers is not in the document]
nickvallee: Thanks for the feedback! I will give this a look some time next week most likely. Got busy with work :).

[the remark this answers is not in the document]
nickvallee: I also removed the "hide tool tip" message when an error is present, since the button is disabled now, so it has no action

web/src/components/buttontoggle.js
```diff
@@ -40,11 +41,17 @@ const ButtonToggle = ({
...
       />
       {tooltipVisible && !isEmpty(tooltip) && (
         <div className="layer-button-tooltip">
           <div className="tooltip-container">
-            <div className="tooltip-text">{tooltip}</div>
+            <div className="tooltip-text">
+              { !errorMessage && (<div>{tooltip}</div>)}
+              { active && errorMessage && (
+                <div className='tooltip-error'>{errorMessage}</div>
+              )}
```
madsnedergaard, line 53: Right now the error message is only showing when the button is active, so we're missing the "it's unavailable" message. 

Screenshot 2021-08-11 at 14 17 24

This change will make it appear no matter of it's active or not :)

```suggestion
              { errorMessage && (
                <div className='tooltip-error'>{errorMessage}</div>
              )}
```

madsnedergaard:
Great work here, only thing left is fixing the merge conflict 💪

madsnedergaard (approved):
🙌

PR #3216, opened 2021-07-14
Read from new cache

FelixDQ:
https://github.com/tmrowco/electricitymap/issues/854

### TODOs
- [x] make time slicer hourly (this depends on the number of datapoints, but `.nice()` + 25 datapoints should work`)
- [x] remove now from right hand side of the graphs 
- [x] Ensure that charts and time slider use the same timescale.
image
- [x] ~~Fix power breakdown chart when data is delayed by 1 hour~~ I think this is actually expected behavior because imports/exports are available, so it is those that you are seeing. It still looks a bit weird though..
image


web/src/hooks/redux.js
```diff
@@ -30,11 +30,13 @@ export function useCurrentZoneHistoryDatetimes() {
 // the graph (when not inferable from historyData timestamps).
 export function useCurrentZoneHistoryEndTime() {
   const customDatetime = useCustomDatetime();
-  const gridDatetime = useSelector(state => (state.data.grid || {}).datetime);
+  const gridDatetime =
+    useSelector((state) => (state.data.grid || {}).datetime) ||
+    moment().subtract(2, 'hour').startOf('hour');
```
madsnedergaard, line 35: Would be nice if we could get the latest timestamp from data if possible :)

FelixDQ: The issue is that if the state becomes outdated, I would think that we want it to show a gap in the graph
(see this comment: https://github.com/tmrowco/electricitymap-contrib/pull/3216/files/4a5a80045c8ea4492c228b46747a750808e09ec9#diff-5754e0648333a5b23fb97471161519bbe522f06ed95da9668d10c2a734559642R28-R30)

What do you think is the best solution here?

madsnedergaard: Ah, I guess that makes sense - let's stick with this solution then :)

skovhus (approved):

PR #3220, opened 2021-07-18
AM parser formatting

yuvalabou:
PEP8 Compatible
sort imports
general cleanup


madsnedergaard (approved):
Formatting is nice, but the parser is not actually working..

Screenshot 2021-07-19 at 13 55 17
Screenshot 2021-07-19 at 13 55 37

I'll approve anyway, but would be great if we could look into finding a new source which potentially could bring the country back on the map :)

solevi:
Hello,
I will try to find contact person to solve the accessibility/authentication problem. 
I guess whitelisting the host from which query is performed will be sufficient. Please let me know from where the query will be performed.

PR #3227, opened 2021-07-19
Edit russia

JeanBaptisteScellier:
Hello,
I propose these updates on Russia's data:
- add exchange data and arrow between european Russia and Crimea
- edit some wrong direction of exchanges between Russia and some other states. It was assumed that the direction was from states based on their alphabetical order, which is not the case. I checked consistency on https://br.so-ups.ru/ to get the right one.
- add some exchange informations between Russia and Finland and Lituania even if it already existed with other sources but it can be useful
- changed the source for exchange between Russia and Georgia because the current one doesn't look like it is working, so I put the source to RU.fetch_exchange

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -1308,7 +1308,7 @@
       43.158267
     ],
     "parsers": {
-      "exchange": "GE.fetch_exchange"
+      "exchange": "RU.fetch_exchange"
     },
     "rotation": 0
   },
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -2370,6 +2370,16 @@
     },
     "rotation": -135
   },
+  "RU-1->UA-CR": {
+    "lonlat": [
+      36.60,
+      45.25
+    ],
+    "parsers": {
+      "exchange": "RU.fetch_exchange"
+    },
+    "rotation": -90
+  },
   "SE-SE1->SE-SE2": {
     "parsers": {
       "exchangeForecast": "ENTSOE.fetch_exchange_forecast"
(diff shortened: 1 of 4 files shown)


JeanBaptisteScellier:
I've added some french names for Russia and correct some mistakes in names about Ural and Crimea

config/exchanges.json
```diff
@@ -1308,7 +1308,7 @@
       43.158267
     ],
     "parsers": {
-      "exchange": "GE.fetch_exchange"
+      "exchange": "RU.fetch_exchange"
```
madsnedergaard, line 1311: Did you check that the `RU.py` parser returns this exchange data? :)

web/locales/fr.json
```diff
@@ -1232,17 +1232,19 @@
...
         },
         "RU-2": {
             "countryName": "Russie",
             "zoneName": "Sibérie"
         },
         "RU-EU": {
-            "zoneName": "Russie"
+            "countryName": "Russie",
+            "zoneName": "Arctique"
         },
```
madsnedergaard, line 1244: Would you mind also doing this for the English language file? It's nice to have actual zones and not just the country name for all of them :)

madsnedergaard (commented):
Nice work here! 💪 
A few minor comments

config/exchanges.json
JeanBaptisteScellier: Yes I did!

web/locales/fr.json
JeanBaptisteScellier: No problem

madsnedergaard (approved):
Nice! 👏

Issue #2173, opened 2020-01-18
Capacities for Italian BZNs


Juma7C9:
Hi,
I found that Terna, the Italian TSO publishes the production capacities divided by region and type (link: [1]), so it is possible to add them to the map simply summing them together for each bidding zone.

This is the, albeit partial, result:

![immagine](https://user-images.githubusercontent.com/5333677/72853076-aba74c00-3cb0-11ea-95c8-3c54d6be3852.png)

[Update 2: Added Rossano BZN data to South BZN]
[Update 3: Manually checked first ~250 entries (sorted by installed capacity) in the ENTSO-E data, casting entries incorrectly classified as _Other_ into the correct categories. Data and changes made in the attached .zip file]

Unfortunately the classification is slightly different than that used by the ENTSO-E, in particular concerning hydro and pumped storage, and thermal sources.

For hydro, and in particular pumped storage, the data distinguishes "pure storage" and "mixed storage", the former having <5% of the energy generated from natural inflows, the latter >5%.
As I don't know where the 'intermediate' case falls into the ENTSO-E data, I added in the table both values, so it is possible to choose from them.

Apart from this, the main issue remains the thermal capacity, as the detail per region does not distinguish the fuel used, while the fuel detail is not divided by region/macro area.

I also cross-checked the data with those available on the 'Installed capacity per production unit', summing all the units under the same BZN.
The result is that the ENTSO-E data underestimates, from slightly (e.g. for Wind) to severely (for Solar) the data reported by Terna, and it is consistent with missing all production units < ~10MW (e.g. for Hydro+Storage, Terna reports a total of 22.9GW, while ENTSO-E reports 19.0GW. The difference is consistent within ~400MW with the capacity of all the units below 10MW [3.5GW, Terna page 68]).

This unfortunately means that the aggregated capacities for thermal/fossil units from the ENTSO-E data are quite inaccurate, as almost 20% of the capacity is missing (only 51.1GW of the 62.3GW total are accounted for), assuming that the ENTSO-E 'Other' units (~9.3GW) are all thermal.
One way I see to solve this discrepancy is to suppose that the missing capacity is due to smaller units, and make an educated guess of their main fuel source(s) (e.g. by comparing ~for each BZN~ which type is mostly missing) -- or simply to add them to the 'Other/Unknown data'.

Additional data:

[Capacities.zip](https://github.com/tmrowco/electricitymap-contrib/files/4094461/Capacities.zip)

[...]

PR #3228, opened 2021-07-19
update capacities Italy

KabelWlan:
I updated Italys capacities (which also solves #2173). 

Sources: 
Entsoe: https://transparency.entsoe.eu/generation/r2/installedCapacityPerProductionUnit/show
Terna: https://www.terna.it/it/sistema-elettrico/statistiche/pubblicazioni-statistiche

I used Entsoe data for coal and oil. The file from Terna reports all the renewables and the total thermal capacity. I then used the total thermal capacity number to subtract the oil and coal capacity to get the gas capacity. 

It is possible that I did't account for all the oil power plants since (probably) not all of them are big enough to be listed on Entsoe. However, the coal plants are big enough and I doubble checked with other websites. 
In the end I think the methode using oil and coal numbers from Entsoe to get the gas values (insted of using the Entsoe values for gas generation capacity) is more accurate.

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -255,9 +255,8 @@ For many European countries, data is available from [ENTSO-E](https://transparen
   - [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)
   - [LNRG capacities data](https://www.lnrg.technology/app/download/12390579357/IL+electricity+capacities.json)
 - Italy
-  - Other: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedCapacityPerProductionUnit/show)
-  - Per Region Renewable: [Terna](https://www.terna.it/en/electric-system/statistical-data-forecast/evolution-electricity-market)
-  - Wind & Solar: [IRENA](http://resourceirena.irena.org/gateway/countrySearch/?countryCode=ITA)
+  - Coal and Oil: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedCapacityPerProductionUnit/show)
+  - Other: [Terna](https://www.terna.it/it/sistema-elettrico/statistiche/pubblicazioni-statistiche)
 - India:
   - solar: [MNRE.GOV.IN](https://www.mnre.gov.in/solar/current-status/)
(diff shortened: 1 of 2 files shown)


madsnedergaard:
Nice work and apparently a much needed update to the capacities! :)

Any idea why ENTSO-E reports such low gas numbers? South Italy is only reported as 4753 MW of fossil gas, while you found it to be 9135 🤔

KabelWlan:
The main reason it that not all power plants (below a certain size) must report their capacity to Entsoe. 

Quote from the linked issue:
„The result is that the ENTSO-E data underestimates, from slightly (e.g. for Wind) to severely (for Solar) the data reported by Terna, and it is consistent with missing all production units < ~10MW (e.g. for Hydro+Storage, Terna reports a total of 22.9GW, while ENTSO-E reports 19.0GW. The difference is consistent within ~400MW with the capacity of all the units below 10MW [3.5GW, Terna page 68]).

This unfortunately means that the aggregated capacities for thermal/fossil units from the ENTSO-E data are quite inaccurate, as almost 20% of the capacity is missing (only 51.1GW of the 62.3GW total are accounted for), assuming that the ENTSO-E 'Other' units (~9.3GW) are all thermal.“

I hope this explains the difference.

madsnedergaard:
> The main reason it that not all power plants (below a certain size) must report their capacity to Entsoe.
> 
> Quote from the linked issue:
> „The result is that the ENTSO-E data underestimates, from slightly (e.g. for Wind) to severely (for Solar) the data reported by Terna, and it is consistent with missing all production units < ~10MW (e.g. for Hydro+Storage, Terna reports a total of 22.9GW, while ENTSO-E reports 19.0GW. The difference is consistent within ~400MW with the capacity of all the units below 10MW [3.5GW, Terna page 68]).
> 
> This unfortunately means that the aggregated capacities for thermal/fossil units from the ENTSO-E data are quite inaccurate, as almost 20% of the capacity is missing (only 51.1GW of the 62.3GW total are accounted for), assuming that the ENTSO-E 'Other' units (~9.3GW) are all thermal.“
> 
> I hope this explains the difference.

It does indeed, thanks for sharing :)

madsnedergaard (approved):

PR #3234, opened 2021-07-24
Update data-source and add archive for moldova

### LINKED ISSUE #973: Moldova MD / emission factor for gas

Impelon:
I noticed Moldova's biggest power plant was still listed as *unknown*, despite it being classified as *gas* in #973.
The problem is that #1380 just modified a value in an unused constant, so the change had no effect.

Regarding dafc431:
I've noticed that there is a new URL for the data-source (http://www.moldelectrica.md/utils/load5.php) and I redesigned the parser using it.
**The old URL seems to just be stuck at some point in time, hence why the data for Moldova is not updated at all!**
I've also added links to some additional documents explaining how the *CERS Moldovenească* power plant operates.
Notably I can confirm that in 2019 and 2020 the power plant still uses mostly gas (over 99%) as fuel:
http://moldgres.com/wp-content/uploads/2021/02/proizvodstvennye-pokazateli-zao-moldavskaja-gres-za-2020-god.pdf
(By searching for *Производственные показатели* aka. *Performance Indicators* on http://moldgres.com/ all available documents can be found. They are written in Russian, but I've been able to easily translate them using Google.)

Additionally this commit takes the renewable energy generation (solar + wind + biogas) in Moldova into consideration. With a capacity of around 3 MW it is quite low and not distinctly mentioned on the map or the `data_url`.
But it can easily be calculated as the difference between the total production and all other forms of production.
The values seem to match up with the values from https://moldelectrica.md/ro/activity/operative_info.
Unfortunately it is not mentioned how this renewable energy mix is composed exactly, so I attributed it all to *biomass*.
(The [presentation](http://en.fas.gov.ru/upload/other/National%20Agency%20for%20Energy%20Regulation%20(G.%20Pyrzy).pdf) mentioned in #926 would indicate that this is mostly biomass, but those figures do not add up with the current data from Moldelectrica anyways.)

In my fork I also added 0903c614d90f8ee530cbb175cd921dffb3b7501e, which represents the code I used to check what the csv-data from `data_url` actually represents. This uses the contents of the `display_url`, where the contents of the data-url are directly mapped to the represented map. In case the data fetched changes slightly in the future, but `display_url` and `data_url` work the same, the code from 0903c614d90f8ee530cbb175cd921dffb3b7501e could be used to update the parser again.
But I do not know how much it fits with the rest of the parser, so I chose not to include it in this PR. Do you have any recommendations on where to put this *maintanence*-code? 

Regarding beeb6ab2:
I added the archive-data used at https://moldelectrica.md/ro/activity/operative_info to the parser as well,
meaning the parser does now support `target_datetime` for past dates.
The data-url for the archive is https://moldelectrica.md/utils/archive2.php?id=table&type=html; various formats are supported and additionally other time-series with higher time-resolution (f.e. for the electricity mix) are supported by passing other numerical values for the id. (I only implemented fetching the table archive data, not for the higher resolution timeseries.)
The archive-data contains data from the end of 2018 to 2021 and ongoing.

Notably the archive also contains `planned` values for production, exchange and consumption. 
You can take a look at https://moldelectrica.md/ro/activity/operative_info to see how the planned values compare to the actual values.
The archive does not contain `planned` values for future dates, so I'm not sure if these values are what the `*_forecast`-methods should return. 
If desired, I could also quickly add implementations for `fetch_consumption_forecast`, `fetch_generation_forecast`, `fetch_exchange_forecast` using the archive to this PR.

Keep in mind, I do not know if the data for the correct timerange is returned when `target_datetime` is used; right now it returns all data for the given date, the time-component is ignored.

----------

This fixes #973.

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -274,7 +274,10 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Malaysia: [GSO](https://www.gso.org.my/SystemData/PowerStation.aspx)
 - Moldova
   - Renewable: [IRENA](https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf)
-  - Other: [FAS](<http://en.fas.gov.ru/upload/other/National%20Agency%20for%20Energy%20Regulation%20(G.%20Pyrzy).pdf>)
+  - Gas (CET-1 & CET-2 Chişinău): [Termoelectrica](https://www.termoelectrica.md/ro_RO/despre/informatii-tehnice/)
+  - Gas (CET-Nord): [CET-Nord](https://www.cet-nord.md/en/transparent/tep)
+  - Gas and coal (CERS Moldovenească): [MOLDGRES](http://moldgres.com/o-predpriyatii/equipment)
+  - Summary as permanent document (slightly outdated): [FAS](<http://en.fas.gov.ru/upload/other/National%20Agency%20for%20Energy%20Regulation%20(G.%20Pyrzy).pdf>)
 - Montenegro
   - Solar, Hydro & Wind: [IRENA](https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2020/Mar/IRENA_RE_Capacity_Statistics_2020.pdf)
   - Other: [EPCG](http://www.epcg.com/en/about-us/production-facilities)
--- a/config/zones.json
+++ b/config/zones.json
@@ -3678,19 +3678,22 @@
     ],
     "capacity": {
       "biomass": 6,
-      "gas": 335,
+      "gas": 1263,
       "hydro": 64,
       "solar": 4,
       "wind": 29,
-      "unknown": 2520
+      "coal": 1600
     },
     "contributors": [
       "https://github.com/systemcatch",
       "https://github.com/alixunderplatz",
       "https://github.com/nessie2013",
-      "https://github.com/lgrigoryeva1"
+      "https://github.com/lgrigoryeva1",
+      "https://github.com/Impelon"
     ],
     "parsers": {
+      "consumption": "MD.fetch_consumption",
+      "price": "MD.fetch_price",
       "production": "MD.fetch_production"
     },
     "timezone": "Europe/Chisinau"
--- a/parsers/MD.py
+++ b/parsers/MD.py
@@ -4,35 +4,59 @@
 """Parser for Moldova."""
 
 import arrow
-from operator import itemgetter
 import requests
 
-TYPE_MAPPING = {
-    u'tmva476': 'hydro',  # NHE Costeşti (run-of-river) #2 index
-    u'tmva112': 'hydro',  # NHE Dubăsari (run-of-river) #4 index
-    u'tmva367': 'gas',  # CET Nord (CHPP) #3 index
-    u'tmva42': 'gas',  # CET-1 Chişinău (CHPP) #6 index
-    u'tmva378': 'gas',  # CET-2 Chişinău (CHPP) #5 index
-    u'tmva1024': 'gas',  # CERS Moldovenească (fuel mix 2017 99.92% gas, 0.08% oil) #7 index
-}
 
 display_url = 'http://www.moldelectrica.md/ro/activity/system_state'
-data_url = 'http://www.moldelectrica.md/utils/load4.php'
+data_url = 'http://www.moldelectrica.md/utils/load5.php'
+
+# To match the fields with the respective index and meaning,
+# I used the following code which may be used in the future for maintenance:
+# https://gist.github.com/Impelon/09407f739cdff05134f8c77cb7a92ada
+
+# Fields that can be fetched from data_url linked to production.
+# Types of production and translations have been added manually.
+production_fields = (
+    {'index': 2, 'name': 'NHE Costeşti', 'type': 'hydro'},  # run-of-river
+    {'index': 3, 'name': 'CET Nord', 'type': 'gas'},  # CHPP
+    {'index': 4, 'name': 'NHE Dubăsari', 'type': 'hydro'},
+    {'index': 5, 'name': 'CET-2 Chişinău', 'type': 'gas'},  # CHPP
+    {'index': 6, 'name': 'CET-1 Chişinău', 'type': 'gas'},  # CHPP
+    {'index': 7, 'name': 'CERS Moldovenească', 'type': 'gas'},  # fuel mix: 99.94% gas, 0.01% coal, 0.05% oil (2020)
+)
+# Other relevant fields that can be fetched from data_url.
+other_fields = (
+    {'index': 26, 'name': 'consumption'},
+    {'index': 27, 'name': 'generation'},
+    {'index': 28, 'name': 'exchange UA->MD'},
+    {'index': 29, 'name': 'exchange RO->MD'},
+    {'index': 30, 'name': 'utility frequency'},
+)
+
+# Further information on the equipment used at CERS Moldovenească can be found at:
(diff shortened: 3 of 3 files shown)


Impelon:
Forgot to change some variables, as I've made some last minute changes before. Should be ready for review now!

Impelon:
The price could also be returned from https://moldelectrica.md/ro/activity/tariff, 
but it doesn't change, it is specified to be 0.145 RON / kWh by this law:
http://lex.justice.md/viewdoc.php?action=view&view=doc&id=360109&lang=1
So I don't think it would be worth writing a parser for it, since it may break before/when the price actually changes.

In their [2020 annual report Moldelectrica (page 51)](https://moldelectrica.md/files/docs/raport/RAPORT%20ANUAL%202020.pdf) also mention approved capacities for renewables and the de-facto capacities for additional power-sources connected to their grid, which are as follows:
* solar: 0.5 MW
* wind: 4.14 MW
* biogas: 3.6 MW
* combined heat and power: 13.4 MW
Unfortunately Moldelectrica does not seem to provide live-data on these.

madsnedergaard:
This is awesome, thanks for doing this extensive research and work! 💪 
And great with historical data!

> But I do not know how much it fits with the rest of the parser, so I chose not to include it in this PR. Do you have any recommendations on where to put this maintanence-code?

We currently don't have a unified way of doing this. But maybe you could add a link to the commit or move the code to a Github Gist and add that link as a comment to the parser code? It's not ideal, but I think it could work for now :)

> The archive does not contain planned values for future dates, so I'm not sure if these values are what the *_forecast-methods should return.
> If desired, I could also quickly add implementations for fetch_consumption_forecast, fetch_generation_forecast, fetch_exchange_forecast using the archive to this PR.

The values are meant to be forecast for future dates, and it seems like that's not possible here - so no need to do that.

> price

It would actually be nice to save the price, so this graph has data:

Screenshot 2021-07-26 at 12 12 15

I think your suggestion of a function returning the static value (and a comment about why) is perfect for this use case - then it can easily be updated later if the price changes, instead of having a price-parser that might break.

---

And then one additional comment: 

You mention that coal production capacity of coal is 1600 MW, but the parser doesn't return any coal - is that correct, and are we potentially missing out on something? If the data isn't available, there's not much we can do - I'm just curious :)

Impelon:
> You mention that coal production capacity of coal is 1600 MW, but the parser doesn't return any coal - is that correct, and are we potentially missing out on something? If the data isn't available, there's not much we can do - I'm just curious :)

I believe the coal-powered units are mostly unused, as the "Performance Indicators"-reports ([the one for 2020 is also linked above](http://moldgres.com/wp-content/uploads/2021/02/proizvodstvennye-pokazateli-zao-moldavskaja-gres-za-2020-god.pdf)) for the past few years indicate coal is rarely used as fuel. For example the exact fuel-ballance for 2020 was:
* gas: 99.94%
* coal: 0.01%
* oil: 0.05% 

Although it is not explicitly mentioned how the ratio is determined, and I can imagine that if it was by volume, pulverized coal would have a low ratio.

But considering that the CCGT and gas+oil units are virtually enough to meet Moldova's planned peak power demand of 905 MW today (and CERS Moldovenească currently produces slightly less than 800 MW), I do think that it is possible the coal units are mostly used as a reserve only.

Unfortunately there is no real way to tell, Moldelectrica's archive simply aggregates CERS Moldovenească's production together with the thermal power plants. And CERS Moldovenească only provides the "Performance Indicators"-reports.

On that note I also just noticed there does not seem to be a source mentioned in #926 that explicitly states the other smaller power-plants to be natural-gas powered.
But at least CET-1 and CET-2 do seem to be mostly gas-powered as indicated here: https://www.termoelectrica.md/en_US/despre/informatii-tehnice/
Should I add that as a source in README.md?

I think it will be hard to get the electricity-mix right for Moldova, I found some sources that specify that around 20% of Moldova's energy share is renewable, mostly from biogas. But Moldelectrica simply does not provide the data for this, maybe because it is not supplied to their grid, or because they are not the producer and hence they do not report on it.
I found this chart for Moldova's energy share (including heating) in 2020 from [here](https://www.enerdata.net/estore/energy-market/moldova.html):
image

> This is awesome, thanks for doing this extensive research and work! :muscle:
> And great with historical data!

Glad I could help! I'll implement all the other proposed changes and update the PR later today!

madsnedergaard:
Thanks for the explanation!

> Should I add that as a source in DATA_SOURCES.md?
No need to update `DATA_SOURCES.md` :)


> I think it will be hard to get the electricity-mix right for Moldova, I found some sources that specify that around 20% of Moldova's energy share is renewable, mostly from biogas

We might not get it 100% accurate, but I think your work here is more than good enough - especially when the mix is mostly gas and some biomass as you found.

Impelon:
> No need to update `DATA_SOURCES.md` :)

Oops, I saw that a bit late. 62d3664 can safely be ignored then!

According to https://www.cet-nord.md/en/transparent/tep, CET-Nord has a new capacity of 37 MW (previously 28,5 MW), so I also added that to the gas category, for a total of 66 (CET-1) + 240 (CET-2) + 37 (CET Nord) + 920 (CERS Moldovenească) = 1263 MW.

I also added `fetch_consumption`. Consumption was previously returned in `fetch_production`, but I figured it should be in the proper function, right?

Otherwise everything should be as discussed.

Impelon:
@madsnedergaard If you wish I can remove 62d3664 via a rebase!

madsnedergaard (approved):
The data-source update is fine, this is overall great! 🥳

Impelon:
Glad I could help 🎉

madsnedergaard:
It's live! 💯 
Screenshot 2021-07-29 at 08 46 50

Impelon:
Looks great! :smile: 
Although it seems the price does not work. It just says ? / MWh... 

Also I guess I forgot to scale up the price to MWh, I assume the price should be returned in MWh, although the [wiki-page](https://github.com/tmrowco/electricitymap-contrib/wiki/Building-a-new-parser#fetch_price) does not mention it.

So the price returned by `fetch_price` should be 145 and probably not a string, I guess that's why that does not work :sweat_smile:

madsnedergaard:
> Looks great! 😄
> Although it seems the price does not work. It just says ? / MWh...
> 
> Also I guess I forgot to scale up the price to MWh, I assume the price should be returned in MWh, although the [wiki-page](https://github.com/tmrowco/electricitymap-contrib/wiki/Building-a-new-parser#fetch_price) does not mention it.

Good point - I have updated the wiki now to mention that price is per MWh :)

> So the price returned by `fetch_price` should be 145 and probably not a string, I guess that's why that does not work 😅

Whoops yeah, it should be a float it seems 😄 
Screenshot 2021-07-29 at 10 33 58

PR #3240, opened 2021-07-28
Update timezone for NL

gwpicard:
Updating the timezone property in `zones.json` for the NL

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -4103,7 +4103,7 @@
       "price": "ENTSOE.fetch_price",
       "production": "NL.fetch_production"
     },
-    "timezone": null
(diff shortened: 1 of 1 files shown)


FelixDQ (approved):

pierresegonne:
We'll have to update the config with the disclaimer as well, should we do both at once ?
I'll approve and let you decide

pierresegonne (approved):

gwpicard:
> We'll have to update the config with the disclaimer as well, should we do both at once ?
> I'll approve and let you decide

Might make more sense to add the disclaimer in a separate PR once we include the updated emission factors. This is to make it easier to have models down the line with the timezone info.

PR #3241, opened 2021-07-28
Enable NL with adjusted emission factor and disclaimer

### LINKED ISSUE #3015: NL Down

FelixDQ:
Fixes #3015 

See explanation for change here: https://github.com/tmrowco/electricitymap-contrib/issues/3015#issuecomment-888154930

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -4095,6 +4095,7 @@
       "https://github.com/PaulCornelissen",
       "https://github.com/nessie2013"
     ],
+    "disclaimer": "The data provider (ENTSO-E) underreports solar and wind production. The unknown emission factor is adjusted to account for this.",
     "parsers": {
       "consumption": "ENTSOE.fetch_consumption",
       "consumptionForecast": "ENTSOE.fetch_consumption_forecast",
--- a/parsers/NL.py
+++ b/parsers/NL.py
@@ -12,7 +12,7 @@ import requests
 
 
 def fetch_production(zone_key='NL', session=None, target_datetime=None,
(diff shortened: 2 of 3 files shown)


config/zones.json
```diff
@@ -4095,6 +4095,7 @@
       "https://github.com/PaulCornelissen",
       "https://github.com/nessie2013"
     ],
+    "disclaimer": "The data provider (ENTSO-E) underreports solar and wind production. The unknown emission factor is adjusted to account for this.",
```
FelixDQ, line 4098: @pierresegonne does this properly explain the problem? Do you want to add something?

pierresegonne: Sounds good!

parsers/NL.py
```diff
@@ -12,7 +12,7 @@
 
 
 def fetch_production(zone_key='NL', session=None, target_datetime=None,
-                     logger=logging.getLogger(__name__), energieopwek_nl=True):
```
pierresegonne, line 15: do we still get the unknown part from total production, total consumption and exchanges without the flag ?

pierresegonne (approved):

parsers/NL.py
FelixDQ: Yes, that happens here:
 https://github.com/tmrowco/electricitymap-contrib/blob/cf34cb7730d6355530c6bda70c397135126cba59/parsers/NL.py#L138-L148

Issue #3218, opened 2021-07-17
CA-QC data is now available!


bouchecl:
Hydro-Québec has finally released a JSON file with the breakdown of generation in their balancing area by source. Data is updated hourly.
https://www.hydroquebec.com/data/documents-donnees/donnees-ouvertes/json/production.json

Although the fields are in French, the utility provides some documentation in English.
https://www.hydroquebec.com/documents-data/open-data/electricity-generation-quebec/

jarek:
@pjakobsen commented yesterday in https://github.com/tmrowco/electricitymap-contrib/issues/1521 that he will try to create a parser, I assume from this data

bouchecl:
>This is 6 hours :/. It would refer to a 411MW gas turbine https://en.wikipedia.org/wiki/Gentilly_Nuclear_Generating_Station#Gentilly-2 Maybe it doesn't run very frequently?

Bécancour is a combustion turbine. It's rarely used as a peaker, mostly in winter.

> Not sure how to classify those - might be good if someone can dig into statistical historical data if any is available, to see if any of these categories prevail. Electricity Map doesn't currently have a category for "biogas", we'd have to check what exactly that consists of.

Biogas = Landfill waste. It's a reference to is a 4.8 MW cogen unit running in Montreal http://eolectric.com/en/news/2017-10-16

jarek:
> Biogas = Landfill. It's a reference to a 4.8 MW cogen unit running in Montreal http://eolectric.com/en/news/2017-10-16

Ah thanks! 4.8 MW isn't all that large compared to the ~500 MW total values being reported in the "other renewable" field (and certainly not large compared to the ~20 GW hydro generation).

Do you know roughly how much of "other renewable" is biomass and how much is independent hydro?

If we can't figure it out, or if hydro is not a large part of it, we can probably just use Electricity Map's default biomass value of 230 g CO2eq/kWh for this category. Hydro's CO2eq intensity is a lot lower so if anything it would only improve the total, while the biogas will be a tiny part of the total, so just assuming the biomass value of 230 is a decent approximation.

bouchecl:
> Do you know roughly how much of "other renewable" is biomass and how much is independent hydro?

[...]

PR #3254, opened 2021-08-01
handle thermal/gas generation in Québec

jarek:
Per https://github.com/tmrowco/electricitymap-contrib/issues/3218, Québec's thermal generation is at Bécancour gas turbine. It is reported with a delay, and data source returning 0.0 can indicate either no generation or not-yet-reported generation. To handle this, if reported value is 0.0, overwrite it to None, so that backend can know this is not entirely reliable and might be updated later.

Set geothermal to 0 as there are no geothermal electricity generation stations in Québec (and all of Canada for that matter).

I didn't do the unknown/biomass change discussed in #3218 since @pjakobsen was looking into that.

### DIFF
--- a/parsers/CA_QC.py
+++ b/parsers/CA_QC.py
@@ -44,13 +44,20 @@ def fetch_production(
                 "production": {
                     "biomass": 0.0,
                     "coal": 0.0,
-                    "gas": 0.0,
+
+                    # per https://github.com/tmrowco/electricitymap-contrib/issues/3218 , thermal generation
+                    # is at Bécancour gas turbine. It is reported with a delay, and data source returning 0.0
+                    # can indicate either no generation or not-yet-reported generation.
+                    # To handle this, if reported value is 0.0, overwrite it to None, so that backend can know
+                    # this is not entirely reliable and might be updated later.
+                    "gas": if_exists(elem, "thermal") or None,
+
                     "hydro": if_exists(elem, "hydro"),
                     "nuclear": 0.0,
                     "oil": 0.0,
                     "solar": if_exists(elem, "solar"),
                     "wind": if_exists(elem, "wind"),
-                    "geothermal": if_exists(elem, "geothermal"),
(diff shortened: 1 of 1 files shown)


pjakobsen:
> Per #3218, Québec's thermal generation is at Bécancour gas turbine. It is reported with a delay, and data source returning 0.0 can indicate either no generation or not-yet-reported generation. To handle this, if reported value is 0.0, overwrite it to None, so that backend can know this is not entirely reliable and might be updated later.
> 

Returning None for thermal from the parser produces the following error (is this what you expected @jarek? - if so, I will commit suggested changes)

```
{'datetime': datetime.datetime(2021, 8, 1, 19, 30, tzinfo=tzfile('/usr/share/zoneinfo/America/Montreal')),
 'production': {'biomass': 569.0,
                'coal': 0.0,
                'gas': 0.0,
                'geothermal': 0.0,
                'hydro': 20369.0,
                'nuclear': 0.0,
                'oil': 0.0,
                'solar': 0.0,
                'thermal': None,
                'unknown': 0.0,
                'wind': 534.0},
 'source': 'hydroquebec.com',
 'zoneKey': 'CA-QC'}
---------------------
took 0.27s
min returned datetime: 2021-08-01T23:30:00+00:00 UTC
max returned datetime: 2021-08-01T23:30:00+00:00 UTC  :( >2h from now !!!
2021-08-01 22:02:10,727 WARNING  __main__                       Validation failed @ 2021-08-01 19:30:00-04:00: Couldn't find emission factor for 'thermal' in 'CA-QC'. Maybe you misspelled one of the production keys?
```

jarek:
`thermal` is not a key recognized by Electricity Map. We've concluded that everything connected to Hydro-Québec grid described as "thermal" is fueled by natural gas (Bécancour station), so we should use `gas` as Electricity Map key for what Hydro-Québec calls "thermal".

skovhus (approved):
Thanks @jarek!

PR #3261, opened 2021-08-02
Give accessible names to buttons.

steren:
Accessibility Lighthouse score goes from 76 to 93


skovhus (approved):

skovhus:
Thanks for fixing this!

PR #3263, opened 2021-08-02
Preconnect to API endpoint

steren:
Sending a request to api.electricitymap.org takes time because it is a CORS request, and therefore needs to send first a Preflight request. By pre-connecting to the domain immediately, the Preflight request should be sent as soon as possible. Potentially saving ~500ms to retrievethe map data.

image


skovhus (approved):
Thanks for fixing this! 👏

PR #3274, opened 2021-08-03
Consistently use window so we can render in node.js/Gatsby

skovhus:
Consistently use window so we can render in node.js/Gatsby in the translation helper.

I would love to revamp the translation system (likely with [LinguiJS](https://lingui.js.org/) or [react-intl](https://formatjs.io/docs/getting-started/installation/)) which would enable us to:
- get rid of node.js for preparing the language files (they can be loaded dynamically)
- have the default english text in the code (less moving parts + easier to grep for)
- leverage automatic ID generation 
- easier share components with subsystems by not relying on the global object to provide locales and locale

### DIFF
--- a/web/src/helpers/translation.js
+++ b/web/src/helpers/translation.js
@@ -1,6 +1,8 @@
 /* eslint-disable prefer-rest-params */
 /* eslint-disable prefer-spread */
 // TODO: re-enable rules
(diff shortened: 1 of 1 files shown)


skovhus:
Merging this now as it blocks the release of contrib after https://github.com/tmrowco/electricitymap-contrib/pull/3261 was introduced.

PR #3276, opened 2021-08-04
Add Russia East (RU-AS) parser

JeanBaptisteScellier:
Hello,
I'd like to add a parser for Eastern Russia. Source is the same than 2 other existing zones in RU except that API link and frequency are different.
As it doesn't cover all East Russia, I've splited it into 2 zones: RU-AS stays itself but lost its most eastern part now called RU-FE for which there is no data.
I also updated the CO2 emissions with 2021 data based on already used source.
I've generated the frontend on local, it looks fine. I've edited zones.json manually for bounding box because if I understand, Russia cannot be handled by 'generate-zone-bounding-boxes.js'.


parsers/RU.py
```diff
@@ -51,23 +57,66 @@
...
+
     if target_datetime:
-        raise NotImplementedError('This parser is not yet able to parse past dates')
+        target_datetime_tz = arrow.get(target_datetime).to(tz)
+    else:
+        target_datetime_tz = arrow.now(tz)
+    # Query at t gives production from t to t+1
+    # I need to shift 1 to get the last value at t
+    datetime_to_fetch = target_datetime_tz.shift(hours=-1)
+    date = datetime_to_fetch.format('YYYY.MM.DD')
```
FelixDQ, line 113: All dates are now shifted one hour from before. Is this correct?

FelixDQ (commented):
Thank you @JeanBaptisteScellier!

Looks great! I took the liberty to update the bounding boxes a little bit. The reason why the script fails is because the zone crosses the 180th meridian, so your solution of doing it manually is fine. `RU-AS` no longer crosses it, so I regenerated the script and only excluded the new `RU-FE`

I have one question, and then we should be ready to merge 🎉

parsers/RU.py
JeanBaptisteScellier: Exactly. Now consumption at hour h means consumption from h-1 to h (or from h-30min to h for East Russia). It seemed more intuitive to me.

JeanBaptisteScellier: Actually, you commented here but it doesn't take place here. At this place, I am just saying that if it 00:30 we should fetch the data of yesterday, not today because data cannot exist yet.
The shift takes place on line 143.

JeanBaptisteScellier: But my commented code was here, it is a bit misleading, I should move it.

FelixDQ: Ah, I see. Thanks for explaining.

I know using the end datetime is the standard in other places, but electricityMap uses the start datetime due to the way we standardize data across different frequencies. i.e. a datetime at hour `13:00` mean `13:00` - `14:00` and NOT `12:00-13:00`.  So we would have to shift it back again.

FelixDQ (changes requested):

JeanBaptisteScellier (commented):
solved by 2 last commits

FelixDQ (approved):
Awesome! Thank you @JeanBaptisteScellier 🎉 

LGTM!

madsnedergaard:
Hey @JeanBaptisteScellier, thanks again RU-AS is now live! 🥳 
Screenshot 2021-08-12 at 16 56 05

PR #3282, opened 2021-08-06
Set battery storage and hydro capacity in CA-YT, Yukon (Canada) 🇨🇦 

Manu1400:
### Set battery storage and hydro capacity in CA-YT

source: Yukon Energy
* https://yukonenergy.ca/energy-in-yukon/projects-facilities/solar (40 * 260W = 10400 W)
* https://yukonenergy.ca/energy-in-yukon/projects-facilities/battery-storage (status: project, 2022)

### Update/set capacities in CA-YT

Sources: 
* https://en.wikipedia.org/wiki/List_of_generating_stations_in_Yukon
> Yukon Energy facilities have a total capacity of 116 MW
* https://en.wikipedia.org/wiki/Yukon_Energy

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1206,7 +1206,8 @@
   "CA-YT": {
     "contributors": [
       "https://github.com/jarek",
-      "https://github.com/nessie2013"
+      "https://github.com/nessie2013",
+      "https://github.com/Manu1400"
     ],
     "flag_file_name": "ca.png",
     "parsers": {
@@ -1224,7 +1225,9 @@
       ]
     ],
     "capacity": {
-      "hydro": 92
+      "battery storage": 0,
+      "hydro": 92,
+      "solar": 0.0104
     }
   },
   "CH": {
--- a/config/zones.json
+++ b/config/zones.json
@@ -1227,7 +1227,10 @@
     "capacity": {
       "battery storage": 0,
       "hydro": 92,
-      "solar": 0.0104
+      "nuclear": 0,
+      "solar": 0.0104,
+      "wind": 0.81,
+      "unknown": 23.1796
     }
   },
   "CH": {


config/zones.json
```diff
@@ -1224,7 +1225,12 @@
       ]
     ],
     "capacity": {
-      "hydro": 92
+      "battery storage": 0,
+      "hydro": 92,
+      "nuclear": 0,
+      "solar": 0.0104,
+      "wind": 0.81,
+      "unknown": 23.1796
```
jarek, line 1233: What is this value?

jarek: Ah I see, it's in the commit message:
> * https://en.wikipedia.org/wiki/List_of_generating_stations_in_Yukon
> Yukon Energy facilities have a total capacity of 116 MW
> * https://en.wikipedia.org/wiki/Yukon_Energy

So this value `23.1796` is `116 - 92 - 0.81 - 0.0104`.

But the various numbers quoted don't really line up, so I don't think we should make this calculation. This isn't based on figures we can add and subtract like this.

The citation for "Yukon Energy facilities have a total capacity of 116 MW" in https://en.wikipedia.org/wiki/List_of_generating_stations_in_Yukon is https://web.archive.org/web/20100719030413/http://yukonenergy.ca/about/profile/ which says

> Yukon Energy has the capacity to generate 116 megawatts of power. Seventy five megawatts of that are provided by our hydro facilities in Whitehorse, Mayo and Aishihik Lake (40 megawatts at Whitehorse, 30 megawatts at Aishihik and 5 megawatts at Mayo), 39 megawatts by diesel generators (which we currently only use as back-up) and 0.8 megawatts by two wind turbines located on Haeckel Hill near Whitehorse.

First, that is an archive as of 2010. Second, It lists 75 MW of hydro whereas you list 92 MW, but I'm not really sure where the 92 MW comes from (possibly from Yukon Energy's self-reported max hydro capacity, which they report in their current status page when not generating thermal?). Third, the values quoted add up to 114.8 MW, not 116 MW. Lastly, related to the first point, since 2010 Yukon has gained a 12.6 MW LNG peaker/backup plant in Whitehorse that might have replaced some of the diesel back-up, but LNG generation is reported together with diesel in "thermal" category, which goes into our unknown. Per https://yukonenergy.ca/media/site_documents/2_BESS_Part_3_Application_-_FINAL.pdf section "3.1.1  Existing Facilities and Project Components", just the installed LNG+diesel capacity at Whitehorse is 24 MW (more than we'd be adding here), and I think there's some elsewhere as well.

BTW, you can see some details from the 2017 parser implementation in the comment in https://github.com/tmrowco/electricitymap-contrib/blob/2e87c0871b0d9a6779633a3e344f17ffe7ca1523/parsers/CA_YT.py.

config/zones.json
```diff
@@ -1224,7 +1225,12 @@
...
     }
```
jarek, line 1234:
```suggestion
      "wind": 0.81
    }
```

Suggest to remove the `unknown` as per comment above.

gwpicard:

PR #3284, opened 2021-08-06
Update CA-NS capacities

Manu1400:
source: https://www.nspower.ca/about-us/electricity/producing

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1045,10 +1045,11 @@
     ],
     "capacity": {
       "biomass": 90.1,
-      "coal": 1232,
+      "coal": 1252,
       "gas": 637,
       "hydro": 418,
-      "wind": 567
+      "wind": 567,
+      "oil": 222
     },
     "contributors": [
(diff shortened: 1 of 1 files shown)


jarek:
Per the NS Power page, should natural gas be updated to 500 MW to match Tufts Cove capacity? Noting that this also matches https://en.wikipedia.org/wiki/Tufts_Cove_Generating_Station so probably our current Wikipedia source, https://en.wikipedia.org/wiki/List_of_generating_stations_in_Nova_Scotia is out of date. 

The generation page https://www.nspower.ca/cleanandgreen/todays-energy-stats lists natural gas and oil together ( :| ) but apparently the oil stations don't run very frequently, so it should be okay to keep it as a close-enough approximation.

Could you also update https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md to include the NS Power URL in addition to the Wikipedia lists which are used for biomass, hydro, and wind?

gwpicard:

PR #3289, opened 2021-08-08
Update exchanges.json

dhilbig:
Added capacities to several Australian interstate exchanges


skovhus:
@dhilbig thanks for adding these. Can you add the source to https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md? Or does it already cover where you found these capacities?

skovhus (approved):
Thanks! I see you updated DATA_SOURCES in https://github.com/tmrowco/electricitymap-contrib/pull/3288

PR #3290, opened 2021-08-08
Update OPENNEM.py

dhilbig:
Fetch exchange between NSW/VIC and SA/VIC not working. Docs suggest the entries should be alphabetical so have reversed the entries. I have made decision about the direction based on AEM dashboard. Note that I have never programmed in python, so just making educated guess.


corradio:
Tested locally. It works now. Thanks!!

parsers/OPENNEM.py
```diff
@@ -29,11 +29,11 @@
         'direction': 1,
     },
     'AUS-NSW->AUS-VIC': {
-        'region_id': 'VIC1->NSW1',
-        'direction': 1,
+        'region_id': 'NSW1->VIC1',
+        'direction': -1,
```
corradio, line 33:
```suggestion
        'direction': 1,
```
The reason is that the direction of the key (AUS-NSW->AUS-VIC) is the same as the region_id (NSW1->VIC1)

parsers/OPENNEM.py
```diff
@@ -29,11 +29,11 @@
...
     },
     'AUS-SA->AUS-VIC': {
-        'region_id': 'VIC1->SA1',
+        'region_id': 'SA1->VIC1',
         'direction': -1,
```
corradio, line 37:
```suggestion
        'direction': 1,
```

corradio (changes requested):

corradio (approved):

PR #3291, opened 2021-08-08
add real-time data sources Australia

KabelWlan:
I added the real-time data source for Flinders Island, Rottnest Island and OpenNEM.

Theorecially we could delete the other sources 
- AREMI National Map
- AEMO
- Australia (distributed solar generation)
- Australia (South Battery)

Those are not in use anymore if I see that correct.

What do you guys think?


FelixDQ (commented):
Thank you @KabelWlan :-)

Yes, it would be great if you also removed the data sources that are no longer in use.

Thanks!

KabelWlan:
Thank you for your feedback. I deleted the old ones.

FelixDQ (approved):
🙌

bch920:
@KabelWlan Should this enable real time access for WA as well ? Currently WA doesn't seem to have live data. I've tried to look through the Parser /logs ( I'm not to technical)  but couldn't see anything obviously wrong.

FelixDQ:
@bch920 this PR is only a documentation change :-) 

I think the problem with WA is that we get delayed data from OpenEM.

PR #3293, opened 2021-08-09
Add installed capacity of Angola 🇦🇴 

Manu1400:
source: http://www.sapp.co.zw/demand-and-supply

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -104,6 +104,12 @@
     },
     "timezone": "Asia/Yerevan"
   },
+  "AO": {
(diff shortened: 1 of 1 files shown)


madsnedergaard:
This number seems to be pretty low compared to data from here (where we also get the full breakdown):

https://www.usaid.gov/powerafrica/angola

and 

https://www.irena.org/IRENADocuments/Statistical_Profiles/Africa/Angola_Africa_RE_SP.pdf

gwpicard:

PR #3294, opened 2021-08-09
Add Russia East (RU-AS) parser

JeanBaptisteScellier:
Sorry, I've created a bug which wasn't detected. It happens when getting the whole RU zone data. I just solved it.


FelixDQ (approved):
Woops. Good catch :-)

PR #3296, opened 2021-08-09
Changed kptclwebsite domain from .com to .in

zuron7:
The website TLD has been changed to a .in from the previous .com 
Checked the basics, seems like the rest of the website is still the same, so the parser should work with no problems.


Manu1400:
update https://github.com/tmrowco/electricitymap-contrib/blob/master/DATA_SOURCES.md#real-time-electricity-data-sources

Manu1400:
@zuron7 please update URL in the parser test file: https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/test/test_IN_KA.py

skovhus (approved):
I took the liberty of updated the PR. Thanks for fixing this... 

IN-KA has been down since 2021-04-29.

PR #3301, opened 2021-08-11
Cache updater fixes

skovhus:
Follow up on the changes in https://github.com/tmrowco/electricitymap-contrib/pull/3216.

This PR fixes the emission graphs and the history mix graph. 

I recommend reviewing with the w=1 query parameter set: https://github.com/tmrowco/electricitymap-contrib/pull/3301/files?w=1

Ref: https://github.com/tmrowco/electricitymap/issues/1225

### DIFF
--- a/web/src/reducers/dataReducer.js
+++ b/web/src/reducers/dataReducer.js
@@ -182,8 +182,6 @@ module.exports = (state = initialDataState, action) => {
           [action.zoneId]: action.payload.map(datapoint => ({
             ...datapoint,
             hasParser: true,
-            // Exchange information is not shown in history observations without production data, as the percentages are incorrect
-            exchange: isEmpty(datapoint.production) ? {} : datapoint.exchange,
           })),
         },
       };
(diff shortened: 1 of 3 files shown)


web/src/reducers/dataReducer.js
```diff
@@ -182,8 +181,6 @@ module.exports = (state = initialDataState, action) => {
           [action.zoneId]: action.payload.map(datapoint => ({
             ...datapoint,
             hasParser: true,
-            // Exchange information is not shown in history observations without production data, as the percentages are incorrect
-            exchange: isEmpty(datapoint.production) ? {} : datapoint.exchange,
```
madsnedergaard, line 186: What's the reason for removing `exchange` here altogether? Where did it show? :)

madsnedergaard: Ah nevermind, I now realise we already do `...datapoint` above!

FelixDQ (approved):
🙌

madsnedergaard (approved):

web/src/reducers/dataReducer.js
skovhus: `exchange` is still there, but now we do not modify it in the reducer. Less remapping.

PR #3303, opened 2021-08-12
Fix slider styling

skovhus:
Ref: https://github.com/tmrowco/electricitymap/issues/854

### Before
Screenshot 2021-08-12 at 09 08 12

### Now
Screenshot 2021-08-12 at 09 07 44

### DIFF
--- a/web/src/scss/content/panel/_time-slider.scss
+++ b/web/src/scss/content/panel/_time-slider.scss
@@ -41,6 +41,7 @@
     padding: 0;
     overflow: visible;
     border: none;
+    margin: 5px 0;
 
     &:focus {
     	outline: none;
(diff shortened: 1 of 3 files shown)


web/src/scss/content/panel/_time-slider.scss
```diff
@@ -40,6 +41,7 @@
     padding: 0;
     overflow: visible;
     border: none;
+    margin: 5px 0;
```
skovhus, line 44: 5px is very much a magic variable... This makes the vertical margin look the same as before. 🤔

web/src/scss/content/panel/_time-slider.scss
```diff
@@ -27,7 +27,8 @@
 }
 
 .time-slider-axis .tick {
-    font-size: 12px;
+    font-size: .85em;
```
skovhus, line 30: Same font size as most of the other small text in the panel

madsnedergaard (approved):
🙌

PR #3308, opened 2021-08-13
add Alberta interconnector capacities

KabelWlan:
Added the interconnector capacites for Alberta.
Source: https://www.aeso.ca/assets/Information-Documents/2011-001R-ATC-and-Transfer-Path-Management.pdf


FelixDQ (approved):
Thanks @KabelWlan ! 🎉

PR #3323, opened 2021-08-20
Added electricity prices for New Zealand

davidcole1340:
Added ability to pull electricity prices for New Zealand. Data comes from [Energy Market Services](https://em6live.co.nz) which is the [commercial entity](https://www.transpower.co.nz/news/ems-launches-new-em6-website-enhanced-services) of Transpower, the New Zealand transmission company.

EMS provides price data for each region of NZ however ElectricityMap only breaks the power usage down into each island, so for the price I have averaged all the regions in each island for the price.


skovhus (commented):
Thanks for adding this!

It seems like something went wrong when you forced pushed the branch. If you are rebasing, you should probably update the main branch first.

davidcole1340:
Yup looks like I rebased from the wrong remote... Will fix now

skovhus (approved):
Thanks for adding this!

And welcome to the eMap community. 👋

PR #3326, opened 2021-08-21
Fix US-CA parser

mutantmonkey:
It seems that the CSV has changed capitalization, so the parser needs to
be updated accordingly.


corradio (approved):
Thanks!

PR #3329, opened 2021-08-23
Remove old US zones

skovhus:
We finally we got around to deprecating the old US zones from the map.

Ref: https://github.com/tmrowco/electricitymap/issues/34 / https://github.com/tmrowco/electricitymap/issues/1310

### Before
Screenshot 2021-08-23 at 15 41 08

### After
Screenshot 2021-08-23 at 15 41 06


madsnedergaard (approved):
Nice! And it works as expected locally? :)

skovhus:
@madsnedergaard everything seems to work, but it would be great if you had a look... 

Unrelated: topogen.sh fails on CI (and locally when doing a clean build): `curl: (7) Failed to connect to naciscdn.org port 443: Connection refused`. It has been failing since yesterday. Example of a call we do that times out: `curl -z ne_10m_admin_0_map_subunits.zip -o x.zip https://naciscdn.org/naturalearth/10m/cultural/ne_10m_admin_0_map_subunits.zip`

I wanted to deploy this to staging to make it easier to do some validation.

skovhus:
@madsnedergaard it is now deployed to https://staging.electricitymap.org/map

Some of the exchanges could probably be placed better now as we don't have to make room for multiple exchanges, but I'm not planning on fixing that right now.

Screenshot 2021-08-24 at 08 34 56

skovhus:
FYI @madsnedergaard I had a few issues related to the optimization in https://github.com/tmrowco/electricitymap-contrib/pull/3233 – I've reverted all changes related to the alpine docker image as we need Python for node-sass and other dependencies. I guess this wasn't noticed before now as the image was somehow cached. 🤔

madsnedergaard:
> FYI @madsnedergaard I had a few issues related to the optimization in #3233 – I've reverted all changes related to the alpine docker image as we need Python for node-sass and other dependencies. I guess this wasn't noticed before now as the image was somehow cached. 🤔

I am puzzled why this is a problem now, but haven't been so far - as far as I can tell it's not because of cache, as we are re-running these steps all the time 🤔

skovhus:
> I am puzzled why this is a problem now, but haven't been so far - as far as I can tell it's not because of cache, as we are re-running these steps all the time 🤔

Me too. But we had similar issues with some other images that I rolled back: https://github.com/tmrowco/electricitymap-contrib/pull/3265

madsnedergaard:
> > I am puzzled why this is a problem now, but haven't been so far - as far as I can tell it's not because of cache, as we are re-running these steps all the time 🤔
> 
> Me too. But we had similar issues with some other images that I rolled back: #3265
> 
> I don't think it makes sense to spent much time on though. :)

But the CI is already so sloooow 😭 (not that this makes that much of a difference)

config/exchanges.json
```diff
@@ -2434,76 +2434,6 @@
     },
```
madsnedergaard, line 2434: I think this case with position of exchange arrows is worth fixing now:


1. New York -> New England

Screenshot 2021-08-24 at 09 47 23

The nicely placed exchange in the middle is the old one that is removed, so would be great if the new one could take it's place instead of being far down at the bottom.

_Old exchange_
Screenshot 2021-08-24 at 09 51 15

_New exchange_
Screenshot 2021-08-24 at 09 51 21

skovhus: Do you know what the best workflow for updating these? 🤔

madsnedergaard: Copy-paste position and rotation? 😅 Don't think it's scripted

skovhus: #sad... I'll try. :D

skovhus: Should be fixed now. Much nicer.

Before (prod): 
Screenshot 2021-08-24 at 11 36 49

After:
Screenshot 2021-08-24 at 11 36 51

madsnedergaard (approved):
Cool, let's see how it goes!

PR #3331, opened 2021-08-25
Reorganizes and cleans up all geo scripts

madsnedergaard:
## Description

The geo scripts were quite messy and all over the place, so I have:
- rearranged everything into a `geo/` folder to keep it together
- Cleaned up scripts and rewritten some code to make it cleaner
- Moved all constants and json imports from `generate-geometries.js` to a new `geometries-config.js`

### TODO

- [x] Figure out why world.json is regenerated and what has changed -> turns out there was a mistake in the geojson imports which meant we missed some maps :)


skovhus (approved):
A much needed refactor! Looks solid. 👏

PR #3347, opened 2021-08-30
Languages update

Raffox97:
English language update:
* Added missing JEA string (Source: https://en.wikipedia.org/wiki/JEA)
* Sorted zone names alphabetically
 master

Italian language update:
* Added missing strings
* Fixed formatting for easy future translations


skovhus (approved):
Thanks for fixing this!

PR #3348, opened 2021-08-30
Updated unknown co2eq value for Sweden with 2019 data

VIKTORVAV99:
Eurostat has released new data for 2019 and this PR just updates the previous value using the same methodology with the new data.


madsnedergaard (approved):
Nice 👍

PR #3357, opened 2021-09-06
Use step curve in graphs

corradio:
Points shown in the history graphs are now aggregates (averages) over the next step (instead of samples that represent an instantaneous value).
As each point represents the average over the next step, it therefore doesn't make sense anymore to use a line interpolated between points, as it misleads to think the value is instantaneous.
A step curve in this case depicts the data more accurately:
https://github.com/d3/d3-shape#curveStepAfter

Before:
image

After:
image


gwpicard (approved):
Nice change :) LGTM! 🎉

PR #3360, opened 2021-09-06
Adds delays for all US zones using EIA parsers

madsnedergaard:
This adds an expected delay to all the US zones that are using the EIA parser. This should remove warnings about these zones from https://github.com/tmrowco/electricitymap-contrib/wiki/Parser-issues


skovhus (commented):
Nice! I suggest you get rid of the outdated JSON schema (which is failing). Note that the Pydantic model will validate the changes to zones.json.

If we want, then the JSON schema can be auto-generated from the Pydantic model: https://github.com/tmrowco/electricitymap-contrib/blob/master/electricitymap/contrib/config/model.py#L63

add-delays-to-eia-zones.js
```diff
@@ -0,0 +1,51 @@
+const path = require("path");
```
skovhus, line 1: Do you want to commit this? :) In that case we can put it in `scripts`.

madsnedergaard: Whoops no, that was not supposed to be added 🙈

skovhus (approved):

Issue #3363, opened 2021-09-07
Update canonical link to exclude query parameters


madsnedergaard:
## Description

Currently, when searching for "electricitymap france" we get 5+ results for the same zone, but with a mix of query parameters enabled:

<img width="1359" alt="Screenshot 2021-09-07 at 10 54 47" src="https://user-images.githubusercontent.com/3296643/132316367-8fd34ba8-3c2f-4210-a891-279f516857de.png">

This makes it harder to know which one you need, and the links are essentially duplicates.

### Proposed solution

Unless we for some reason want to specifically have "zone with solar layer enabled" show up as a search result, I suggest that we update the canonical and alternate links in `<head />` to exclude query parameters.

<img width="583" alt="Screenshot 2021-09-07 at 10 51 20" src="https://user-images.githubusercontent.com/3296643/132316787-01c26d9f-daba-44ee-abcd-53b0092d4099.png">

<img width="566" alt="Screenshot 2021-09-07 at 10 54 06" src="https://user-images.githubusercontent.com/3296643/132316782-9b899fb3-e936-4700-9669-19298f900345.png">

#### Implications

The following parameters will no longer create unique search results:

- `?lang=XX` - so the same page will not show up in different languages like it currently does
- `?solar=X`
- `?wind=X`
- `?remote=X` (this is only used for local development, so it should never have been here in the first place)
- `?fbclid=XXX`
- other random parameters used by others when linking to the map (like UTM and diverse tracking)

madsnedergaard:
If anyone can name reasons for not doing this, then I'd love to hear why as I might have missed something :)

corradio:
> If anyone can name reasons for not doing this, then I'd love to hear why as I might have missed something :)

This makes a lot of sense! The only query string I would keep is the language one.

skovhus:
Makes sense Mads. +1 on keeping the language.

hboisgibault:
Hi guys, I opened a pull request to eliminate duplicate links in the metadata.

gwpicard:
Thanks so much for the PR @hboisgibault !!

corradio:
May I suggest keeping the issue open until the changes are caught by the search engines? Currently, this issue doesn't seem to be resolved (it might just take a bit of time)

corradio:
It seems like this is not caught by search engines anymore! Feel free to re-open if you see it appearing again.

PR #3366, opened 2021-09-08
Modified canonical and alternate URLs in web

hboisgibault:
Solving issue  #3363 

Canonical URL and alternate URLs should not contain query params, otherwise search engines will see duplicate pages.

In the index template _baseUrl_ and _path_ are fetched from the request instead of _originalUrl_ to create the canonical URL, as they do not contain the query string.

For alternate URLs, we just need to loop through locales and add the language as a query param to the canonical URL.

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -158,7 +158,10 @@ app.use('/', (req, res) => {
       res.setLocale(lr[0]);
     }
     const { locale } = res;
-    const fullUrl = `https://www.electricitymap.org${req.baseUrl + req.path}`;
+    let canonicalUrl = `https://www.electricitymap.org${req.baseUrl + req.path}`;
+    if(req.query.lang) {
+      canonicalUrl += `?lang=${req.query.lang}`;
+    }
 
     // basic auth for premium access
     if (process.env.BASIC_AUTH_CREDENTIALS) {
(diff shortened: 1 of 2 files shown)


web/server.js
```diff
@@ -158,7 +158,7 @@ app.use('/', (req, res) => {
       res.setLocale(lr[0]);
     }
     const { locale } = res;
-    const fullUrl = `https://www.electricitymap.org${req.originalUrl}`;
+    const fullUrl = `https://www.electricitymap.org${req.baseUrl + req.path}`;
```
madsnedergaard, line 161: Can we keep the `lang` query parameter in canonical, like suggested by @corradio and @skovhus in the issue? :) 
Could make a new variable for this, so we keep `fullUrl` like this (although maybe it should also be renamed now that it's not the full url 😄)

madsnedergaard (commented):
Awesome, thanks for doing this! 💪 
One comment about lang in canonical

web/server.js
hboisgibault: Noted, I will do add the lang parameter !

hboisgibault: Updated the branch

gwpicard (approved):
LGTM! Thanks for making this change 🎉

PR #3379, opened 2021-09-10
Enable serving live map on app.electricitymap.org

corradio:
This PR disables the redirect to www from app.electricitymap.org so we can test serving it on both domains.
@skovhus I think it would be good to add some redirect tests here. Do you have any idea of how we could go about this? (also feel free to add the scaffolding directly to the PR)

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -137,13 +137,14 @@ app.use('/', (req, res) => {
   const isTmrowCo = req.get('host').indexOf('electricitymap.tmrow') !== -1;
   const isNonWWW = req.get('host').includes('electricitymap.org')
     || req.get('host') !== 'www.electricitymap.org';
-  const isExcludedFromRedirect = req.get('host') === 'app.electricitymap.org'; // Temporary
   const isStaging = req.get('host').includes('staging');
+  const isExcludedFromRedirect = isStaging
+    || req.get('host') === 'app.electricitymap.org'; // Temporary
   const isHTTPS = req.secure;
   const isLocalhost = req.hostname === 'localhost'; // hostname is without port
 
   // Redirect all non-facebook, non-staging, non-(www.* or *.tmrow.co)
-  if (!isStaging && (isNonWWW || isTmrowCo) && (req.headers['user-agent'] || '').indexOf('facebookexternalhit') == -1) {
+  if (!isExcludedFromRedirect && (isNonWWW || isTmrowCo) && (req.headers['user-agent'] || '').indexOf('facebookexternalhit') == -1) {
     res.redirect(301, `https://www.electricitymap.org${req.originalUrl}`);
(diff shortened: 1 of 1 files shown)


web/server.js
```diff
@@ -135,14 +135,16 @@ app.use('/', (req, res) => {
   // redirect everyone except the Facebook crawler,
   // else, we will lose all likes
   const isTmrowCo = req.get('host').indexOf('electricitymap.tmrow') !== -1;
-  const isNonWWW = req.get('host') === 'electricitymap.org'
-    || req.get('host') === 'live.electricitymap.org';
+  const isNonWWW = req.get('host').includes('electricitymap.org')
+    || req.get('host') !== 'www.electricitymap.org';
```
madsnedergaard, line 139: Hmm, will this change from `=== live.e` to `!== www.e` affect other subdomains like api.electricitymap.org? 🤔

corradio: I don't think so as this node server doesn't serve api.electricitymap.org?

madsnedergaard: Okay, let's try it then ¯\\\_(ツ)\_/¯

madsnedergaard (approved):

corradio:
I'm puzzled as https://app.electricitymap.org/ is currently NOT redirected..
Any idea why? Maybe it's time to introduce some tests.

madsnedergaard:
> I'm puzzled as [app.electricitymap.org](https://app.electricitymap.org/) is currently NOT redirected..
> Any idea why?

It's because the redirect logic is currently kinda broken due to `isNonWWW`. That constants says if you're `electricitymap.org` or `live.electricitymap.org`, then you're not `wwww`... So it never captures "app" or anything else.

Here's a quick demo:

Screenshot 2021-09-15 at 08 36 48

>  Maybe it's time to introduce some tests.

I've been avoiding that as it would be nice to rethink this whole server approach as we've talked about before, but maybe it's better to have a few tests until then :)

web/server.js
madsnedergaard: Actually, I think the "includes" usage here will mean it's always evaluating to true!

madsnedergaard: Couldn't we just do `const isNonWWW = req.get('host') !== 'www.electricitymap.org';` or would that have other consequences? 🤔

madsnedergaard (changes requested):
See above comments

web/server.js
corradio: > Actually, I think the "includes" usage here will mean it's always evaluating to true!

Good catch. Why don't we simply remove the redirect for this case then (in order to keep the current behaviour)?
I can just remove the `isNonWWW` from the condition.

corradio:
Closing this PR as the map can already be served on app.electricitymap.org

PR #3380, opened 2021-09-10
Fix JP occtonet exchange parser 

pierresegonne:
## Issue 

#3368 #3369 #3370 #3371 #3372 #3373 #3374 #3375 #3376

## Description

The occtonet exchange parser has gone out of service due to an url change from occtonet. 

This PR fixes the url, and improves code readability of the parser.

## Status

The parser seems to be up and running again, but I want to finish refactoring it before merging


parsers/occtonet.py
```diff
@@ -22,225 +22,270 @@
...
-    headers=r.text
-    headers = eval(headers.replace('false', 'False').replace('null', 'None'))
-    if headers['root']['errMessage']:
-        raise RuntimeError('Headers not available due to {}'.format(headers['root']['errMessage']))
+    r = session.post(
+        "https://occtonet3.occto.or.jp/public/dfw/RP11/OCCTO/SD/CA01S070C",
+        data=form_data,
+    )
+    headers = r.text
+    headers = eval(headers.replace("false", "False").replace("null", "None"))
```
skovhus, line 211: We should get rid of the eval here and use `json.parse`.

parsers/occtonet.py
```diff
@@ -22,225 +22,270 @@
...
+    session: requests.Session,
+    form_data: Dict[str, str],
+):
+    form_data["fwExtention.actionSubType"] = "ok"
+    r = session.post(
+        "https://occtonet3.occto.or.jp/public/dfw/RP11/OCCTO/SD/CA01S070C",
+        data=form_data,
+    )
+    headers = r.text
+    headers = eval(headers.replace("false", "False").replace("null", "None"))
```
skovhus, line 240: We should try getting rid of the eval here and use `json.load`.

pierresegonne:
Just wondering if we should add logging ?

skovhus:
> Just wondering if we should add logging ?

In cases you think would be worth logging? I'm not sure we need additional logging.

parsers/occtonet.py
```diff
@@ -22,225 +23,230 @@
...
+        "tgtRklHdn": "01,北海道・本州間電力連系設備,02,相馬双葉幹線,03,周波数変換設備,04,三重東近江線,05,南福光連系所・南福光変電所の連系設備,06,越前嶺南線,07,西播東岡山線・山崎智頭線,08,阿南紀北直流幹線,09,本四連系線,10,関門連系線,11,北陸フェンス",
+        "transitionContextKey": "DEFAULT",
+        "updDaytime": "",
+    }
+
+    r = session.post(
+        "https://occtonet3.occto.or.jp/public/dfw/RP11/OCCTO/SD/CA01S070C",
+        data=form_data,
+    )
+    response_content = json.loads(r.text)
```
skovhus, line 167:
```suggestion
    response_content = r.json()
```

parsers/occtonet.py
```diff
@@ -22,225 +23,230 @@
...
+        form_data["updDaytime"] = response_content["root"]["bizRoot"]["header"][
+            "updDaytime"
+        ]["value"]
+
+    form_data["fwExtention.actionSubType"] = "ok"
+    r = session.post(
+        "https://occtonet3.occto.or.jp/public/dfw/RP11/OCCTO/SD/CA01S070C",
+        data=form_data,
+    )
+    response_content = json.loads(r.text)
```
skovhus, line 194:
```suggestion
    response_content = r.json()
```

skovhus (approved):
Looks good + and it works using `poetry run test_parser 'JP-TK' production`! 👏 

Small suggestion to use https://docs.python-requests.org/en/master/user/quickstart/#json-response-content instead using the `json` library directly.

PR #3381, opened 2021-09-11
change text color function for carbon intensity square

### LINKED ISSUE #3365: Improve text color on CO2 color rectangle
Hi,

Two years ago, I proposed you to use a background based color text for CO2 color rectangle. (#1886)

[You accepted](https://github.com/tmrowco/electricitymap-contrib/pull/2981) my proposition, and I was kind of proud of it. But, after some new tests, I realized it was not the best solution.

I said you to use the YIQ contrast method, what you did. But in some cases, when background color is brown, it doesn't produce a good result, especially on mobile with very low luminosity.

Instead of YIQ method, I tested the W3CAG method more in depth. With the default settings, result is close to the YIQ method and it doesn't solve the problem for these colors. So I modified it a bit. The idea is to switch to white text earlier on the color ramp. The switch should, in my opinion, starts around 400g/kWh.

You can play with this CodePen to see the difference between YIQ, W3CAG standard and my modified version : [https://codepen.io/phiphou/pen/YBeMxw](https://codepen.io/phiphou/pen/YBeMxw)

You can find more informations about W3CAG color contrast here : [https://www.w3.org/TR/WCAG20-TECHS/G17.html](https://www.w3.org/TR/WCAG20-TECHS/G17.html)

It is not standard, but it meets the needs well.

Here is how to implement my solution in [web/src/components/carbonintensitysquare.js](https://github.com/tmrowco/electricitymap-contrib/blob/edb856322ff308316a3994c3f8bd5a27d34fbbdf/web/src/components/carbonintensitysquare.js) : 

```javascript
const getTextColor = rgbColor=> {
  const colors = rgbColor.replace(/[^\d,.]/g, '').split(',');
  const r = parseInt(colors[0], 10);
  const g = parseInt(colors[1], 10);
  const b = parseInt(colors[2], 10)
  const rgbArr = [r, g, b];
  const computedArr = rgbArr.map (c => { 
    const s = c / 255;
    return s <= 0.03928 ? s / 12.92 : Math.pow((s + 0.055) / 1.055, 2.4);
  })
  const luminosity = 0.2126 * computedArr[0] + 0.7152 * computedArr[1] + 0.0722 * computedArr[2];
  // Here is the modification, I changed (1.05 * 0.05) whith (1.05 * 0.16)
  // Math.sqrt(1.05 * 0.16) - 0.05 is a constant, it can be replaced with it's 
  // approached value (0.35987803063838397) to avoid useless computations 
  return luminosity > Math.sqrt(1.05 * 0.16) - 0.05 ? "black" : "white";
}; 
```

phiphou:
Fix #3365. 

The idea is to switch to white text starting around 400g/kWh rather than 500. 

This fix use a modified version of W3CAG algorithm instead of YIQ method. You can find more informations on this method [here](https://www.w3.org/TR/WCAG20-TECHS/G17.html).

See function comment for more informations about the changes.

### DIFF
--- a/web/src/components/carbonintensitysquare.js
+++ b/web/src/components/carbonintensitysquare.js
@@ -13,11 +13,11 @@ import { __ } from '../helpers/translation';
  * See https://github.com/tmrowco/electricitymap-contrib/issues/3365 for more informations.
  * @param {string} rgbColor a string with the background color (e.g. "rgb(0,5,4)")
  */
- const getTextColor = rgbColor=> {
+ const getTextColor = (rgbColor) => {
   const colors = rgbColor.replace(/[^\d,.]/g, '').split(',');
   const r = parseInt(colors[0], 10);
   const g = parseInt(colors[1], 10);
-  const b = parseInt(colors[2], 10)
+  const b = parseInt(colors[2], 10);
   const rgb = [r, g, b].map (c => { 
     const s = c / 255;
     return s <= 0.03928 ? s / 12.92 : Math.pow((s + 0.055) / 1.055, 2.4);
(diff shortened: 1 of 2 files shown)


skovhus (approved):
Thanks for updating this! 👏

PR #3401, opened 2021-09-22
Use constants for AM

yuvalabou:


parsers/AM.py
```diff
@@ -54,10 +56,14 @@
...
+
 
-def fetch_production(zone_key="AM", session=None, target_datetime=None, logger=None):
+def fetch_production(
+    zone_key="AM", session=None, target_datetime=None, logger=None
+) -> dict:
     r = session or requests.session()
-    url = "http://epso.am/poweren.htm"
+    url = SOURCE
     response = r.get(url)
```
jarek, line 67: `url` isn't used further below

```suggestion
    response = r.get(SOURCE)
```

parsers/AM.py
```diff
@@ -119,7 +117,7 @@ def fetch_exchange(
     sorted_keys = "->".join(sorted([zone_key1, zone_key2]))
 
     r = session or requests.session()
-    url = "http://epso.am/poweren.htm"
+    url = SOURCE
     response = r.get(url)
```
jarek, line 121: same as above

```suggestion
    response = r.get(SOURCE)
```

jarek (commented):
nice! just something i noticed:

parsers/AM.py
```diff
@@ -68,26 +73,18 @@ def fetch_production(zone_key="AM", session=None, target_datetime=None, logger=N
     data_string = soup.find(text=re.compile("var"))
```
pierresegonne, line 73:
```suggestion
    data_string = soup.find(text=re.compile("var"))
    
    if data_string is None:
        logger.warning(f"Could not parse {html_doc}")
        raise ValueError("Empty data object scraped, cannot be parsed.")
```

parsers/AM.py
```diff
@@ -54,11 +56,14 @@
     "[28]": "sparum2",
 }
 
+REGEX = "[-+]?[.]?[\d]+(?:,\d\d\d)*[\.]?\d*(?:[eE][-+]?\d+)?"
+
 
-def fetch_production(zone_key="AM", session=None, target_datetime=None, logger=None):
+def fetch_production(
```
pierresegonne, line 62:
```suggestion
def fetch_production(
    zone_key=ZONE_KEY,
    session=None,
    target_datetime=None,
    logger: logging.Logger = logging.getLogger(__name__),
) -> dict:
```

parsers/AM.py
```diff
@@ -1,7 +1,6 @@
 import re
```
pierresegonne, line 1:
```suggestion
import logging
import re
import sys
from datetime import datetime

import requests
from bs4 import BeautifulSoup
from dateutil import parser as dparser
from dateutil import tz
```

pierresegonne (approved):
Nice!

Suggestion, 

As the `data_string` can be `None`, the suggested code will improve our ability to debug when that happens

CloCkWeRX:
Is it worth adopting this given #3337 ? Or is it better to maybe archive the parser?

madsnedergaard:
> Is it worth adopting this given #3337 ? Or is it better to maybe archive the parser?

I think it's worth merging this in case it becomes available later, but I've also created a discussion about archiving parsers: https://github.com/tmrowco/electricitymap-contrib/discussions/3473

PR #3402, opened 2021-09-22
AX parser cleanup

yuvalabou:
- isort
- black
- pythonic variables naming
- remove whitespaces
- adjust comments


pierresegonne (approved):
Thanks for the improvement!

PR #3405, opened 2021-09-23
Ps/investigate jp refetch

pierresegonne:
### Issue

Refetching JP zones seemed to fail

### Description

* Faulty function call that would give a `datetime` to the expected `session` argument for `fetch_exchange` of `occtonet.py`
* Some zones appear to only have real time data. Added an exception if attempt to refetch these zones.
* Run reformatting tools on the files

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
 # coding=utf-8
+import datetime as dt
 import logging
+
 # The arrow library is used to handle datetimes
 import arrow
-import datetime as dt
 import pandas as pd
+
 from parsers import occtonet
 
 # Abbreviations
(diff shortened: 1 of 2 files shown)


pierresegonne:
Commit by commit review is recommended here

parsers/JP.py
```diff
@@ -129,7 +129,10 @@ def fetch_consumption_df(
     Returns the consumption for an area as a pandas DataFrame.
     For JP-CB the consumption file includes solar production.
     """
-    datestamp = arrow.get(target_datetime).to('Asia/Tokyo').strftime('%Y%m%d')
+    ZONES_ONLY_LIVE = ["JP-TK", "JP-CB", "JP-SK"]
```
skovhus, line 132: Would be nice to move this constant outside this function

skovhus (approved):
Looks good. One nitpick 👍

PR #3407, opened 2021-09-23
feat(parsers): add Sri Lanka parser

byronwilliams:
This is unfortunately not live live and not quite historical. I thought I'd add it for reference


skovhus:
Hi @byronwilliams, thanks for trying to add Sri Lanka. Unfortunately this doesn't live up to our [requirements for a new parser](https://github.com/tmrowco/electricitymap-contrib/wiki/Technical-requirements-for-parser-data), as this does not contain a full breakdown of the production mix or live (as you mention).

I hope we can find other data providers in the future. And thanks for sharing your findings here – it can now be found in the PR history if someone wants to work further on this.

PR #3410, opened 2021-09-24
feat(parsers): add Thailand

byronwilliams:
Thailand's data comes from here - https://www.sothailand.com/sysgen/egat/

They seem only to track current and peak load and only a total. There are some negative values, so if somebody could translate them, we might find they are exports or hydro storage.

### DIFF
--- a/parsers/TH.py
+++ b/parsers/TH.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+"""Parser for the electricity grid of Nigeria"""
+
+import arrow
+import logging
+import requests
+
+
+LIVE_PRODUCTION_API_URL = "https://www.sothailand.com/sysgen/ws/sysgen/actual"
+
+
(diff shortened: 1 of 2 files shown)


parsers/TH.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+
+"""Parser for the electricity grid of Nigeria"""
```
CloCkWeRX, line 3: Country name

skovhus, line 3:
```suggestion
"""Parser for the electricity grid of Thailand"""
```

skovhus:
Hi @byronwilliams, thanks for trying to add Thailand. Unfortunately this doesn't live up to our [requirements for a new parser](https://github.com/tmrowco/electricitymap-contrib/wiki/Technical-requirements-for-parser-data), as this does not contain a full breakdown of the production mix.

I hope we can find other data providers in the future.

PR #3420, opened 2021-09-29
BR refactor

yuvalabou:
Use constants
Black formatting

### DIFF
--- a/parsers/BR.py
+++ b/parsers/BR.py
@@ -1,6 +1,5 @@
-#!/usr/bin/env python3
-
 from collections import defaultdict
+from parsers.BD import GENERATION_MAPPING
 
 import arrow
 import requests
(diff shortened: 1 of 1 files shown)


parsers/BR.py
```diff
@@ -1,191 +1,184 @@
-#!/usr/bin/env python3
-
 from collections import defaultdict
+from parsers.BD import GENERATION_MAPPING
```
byronwilliams, line 2: :-1: - unnecessary coupling to save 8 lines - or is this just a mistake as you have `GENERATION_MAPPINGE` below

parsers/BR.py
```diff
@@ -1,191 +1,184 @@
...
-        'flow': -1
-    },
-    'PY': {
-        'name': u'paraguai',
-        'flow': -1
-    }
+URL = "http://tr.ons.org.br/Content/GetBalancoEnergetico/null"
+SOURCE = "ons.org.br"
+
+GENERATION_MAPPINGE = {
```
byronwilliams, line 13: spelling mistake?

byronwilliams (changes requested):

parsers/BR.py
```diff
@@ -1,191 +1,184 @@
...
+}
+
+COUNTRIES_EXCHANGE = {
+    "UY": {"name": "uruguai", "flow": 1},
+    "AR": {"name": "argentina", "flow": -1},
+    "PY": {"name": "paraguai", "flow": -1},
 }
 
 
 def get_data(session, logger):
```
jarek, line 50: If you're doing refactoring, i'm not sure this function is worth keeping:
- it's literally two lines, which are a very standard use of a very standard library
- it's called only three times
- it doesn't even use the `logger` param

At the very least, let's remove the `logger` param. Can probably also inline the return statement, no need for `json_data` variable.

parsers/BR.py
jarek: importing the Bangladeshi GENERATION_MAPPING seems to be a mistake

parsers/BR.py
```diff
@@ -1,191 +1,184 @@
...
-    entriesToRemove = (u'hidraulica', u'itaipu50HzBrasil', u'itaipu60Hz', u'total')
-    for k in entriesToRemove:
+    totals["hydro"] = (
+        totals.get("hidraulica", 0.0)
+        + totals.get("itaipu50HzBrasil", 0.0)
+        + totals.get("itaipu60Hz", 0.0)
+    )
+    entries_to_remove = ("hidraulica", "itaipu50HzBrasil", "itaipu60Hz", "total")
+    for k in entries_to_remove:
         totals.pop(k, None)
```
jarek, line 78: Refactor to be more Pythonic, using a set and a dict comprehension

```suggestion
    entries_to_remove = {"hidraulica", "itaipu50HzBrasil", "itaipu60Hz", "total"}
    totals = {k, v for k, v in totals.items() if k not in entries_to_remove}
```

parsers/BR.py
```diff
@@ -1,191 +1,184 @@
...
 
-    mapped_totals = {generation_mapping.get(name, 'unknown'): val for name, val
-                     in totals.items()}
+    mapped_totals = {
+        GENERATION_MAPPING.get(name, "unknown"): val for name, val in totals.items()
+    }
```
jarek, line 82: refactor to be more Pythonic, using a set and including the filter in the dict comprehension, something like:

```python
    entries_to_remove = {"hidraulica", "itaipu50HzBrasil", "itaipu60Hz", "total"}

    mapped_totals = {
        GENERATION_MAPPING.get(name, "unknown"): val for name, val in totals.items()
        if name not in entries_to_remove}
    }
```

parsers/BR.py
```diff
@@ -1,191 +1,184 @@
...
-    data = {
-        'datetime': arrow.get(gd['Data']).datetime,
-        'sortedZoneKeys': '->'.join(sorted([zone_key1, zone_key2])),
-        'netFlow': gd['internacional'][country_exchange['name']] * country_exchange['flow'],
-        'source': 'ons.org.br'
+    return {
+        "datetime": arrow.get(gd["Data"]).datetime,
+        "sortedZoneKeys": "->".join(sorted([zone_key1, zone_key2])),
+        "netFlow": gd["internacional"][country_exchange["name"]]
+        * country_exchange["flow"],
```
jarek, line 131: This formatting looks weird. Can we pull it out into a variable, like in `fetch_region_exchange`?

parsers/BR.py
```diff
@@ -1,191 +1,179 @@
...
 
 
 def fetch_production(zone_key, session=None, target_datetime=None, logger=None) -> dict:
     """Requests the last known production mix (in MW) of a given country."""
     if target_datetime:
-        raise NotImplementedError('This parser is not yet able to parse past dates')
+        raise NotImplementedError("This parser is not yet able to parse past dates")
 
     gd = get_data(session, logger)
     generation = production_processor(gd, zone_key)
```
jarek, line 90: Oh, one more thing. This could be a little nicer as:

```suggestion
    timestamp, production = production_processor(gd, zone_key)
```

Then below

```python
        "datetime": timestamp.datetime,
        "production": production,
```

Rather than using somewhat nondescript and mysterious `generation[0]` and `generation[1]`

parsers/BR.py
```diff
@@ -1,191 +1,179 @@
...
 
-    if zone_key1 in countries_exchange.keys():
-        country_exchange = countries_exchange[zone_key1]
+    if zone_key1 in COUNTRIES_EXCHANGE.keys():
+        country_exchange = COUNTRIES_EXCHANGE[zone_key1]
 
-    if zone_key2 in countries_exchange.keys():
-        country_exchange = countries_exchange[zone_key2]
+    if zone_key2 in COUNTRIES_EXCHANGE.keys():
+        country_exchange = COUNTRIES_EXCHANGE[zone_key2]
```
jarek, line 124: `keys()` is not needed here because `in <dict>` automatically checks the keys.

In fact this whole thing could be written more concisely as:

```python
    country_exchange = COUNTRIES_EXCHANGE.get(zone_key1) or COUNTRIES_EXCHANGE.get(zone_key2)
```

parsers/BR.py
```diff
@@ -1,191 +1,179 @@
...
+    zone_key1, zone_key2, session=None, target_datetime=None, logger=None
+) -> dict:
     """Requests the last known power exchange (in MW) between two regions."""
     if target_datetime:
-        raise NotImplementedError('This parser is not yet able to parse past dates')
+        raise NotImplementedError("This parser is not yet able to parse past dates")
 
     gd = get_data(session, logger)
+    dt = arrow.get(gd["Data"]).datetime
+    scc = "->".join(sorted([zone_key1, zone_key2]))
```
jarek, line 118: would rather see `sorted_country_codes` here rather than `scc`, this isn't C++ :P

same with `gd` and `nf`...

jarek (commented):
This looks nice! Just a few more thoughts from Python nitpicker :pick:

PR #3422, opened 2021-09-30
Change backend for the live map

skovhus:
https://github.com/tmrowco/electricitymap/issues/1353

The live map now has a new dedicated backend. Locally this now runs on port 8001 instead of 9000.


corradio (approved):
Drumroll!

PR #3425, opened 2021-10-01
Fix version detection

skovhus:
The new version detection always used the version from the web/package.json file (which is fixed). It also only did a check if the is a new version when the web app started up (it should do that continuously).

While doing this I migrated from Redux to https://swr.vercel.app/ (great little library, that we should start using a lot more).

### Preview
Screenshot 2021-10-01 at 10 33 57
Screenshot 2021-10-01 at 10 35 37


corradio (approved):
🎉

PR #3429, opened 2021-10-03
IN-CT

yuvalabou:
use constants for IN-CT
cleanup


madsnedergaard (approved):
Refactor looks good, the parser is down - but I'll approve in case it comes back up :)

PR #3433, opened 2021-10-06
add mobile responsive view for language translation selector

oscarmrom:
See comment thread here: https://github.com/tmrowco/electricitymap-contrib/issues/1841#issuecomment-935235883

I've added a few SCSS changes to the button container with an additional button toggle similar to what activates the language selector. 

The icon can be altered for a different one that may be more fitting. At this time, I could not make material-icons work properly. I may have been referencing them incorrectly. I was trying to use the `Close` icon: https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Aclose%3A

Happy to take feedback and make adjustments.

### DIFF
--- a/web/src/components/languageselect.js
+++ b/web/src/components/languageselect.js
@@ -35,6 +35,14 @@ const LanguageSelect = () => {
       />
       {languagesVisible && (
         <div id="language-select-container">
+          <div id="close-language-select-container">
+            <ButtonToggle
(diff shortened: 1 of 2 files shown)


oscarmrom:
@skovhus , I've taken a quick initial try at simply extracting the Language Select scss over into the component via `styled-components`. Let me know if this is enough for now, or any tips on how to break apart the rest of the styles if you have them.

PR #3460, opened 2021-10-25
Add test case for equal zone keys

pierresegonne:
Example of a test to verify matching between `zones.json` and `world.geojson`

For example `world.geojson` check https://raw.githubusercontent.com/tmrowco/electricitymap-contrib/mk-improve-geometry-process/web/geo/new_world.geojson

This fails and reveals that we still have deprecated / not used zones in `zones.json`.


pierresegonne:
Status: 

This PR can only be merged once the new geometries process has been completed.

madsnedergaard:
Does this still make sense after the changes or is it not needed any longer? 🤔

pierresegonne:
This should be unblocked by the new geographies!

Let's pick it up during the week - it's just a test to verify that all zones with geographies are defined in zones.json and vice versa

pierresegonne:
@madsnedergaard let's discuss IRL this, as it has important consequences for how we define zones :)

pierresegonne:
I messed up the commit history, I'll just recreate a branch and make this cleaner

madsnedergaard:
Oh okay, fair 😄 
I'm also leaning towards Oli's idea of not deleting the "unused" zones for now, as I'd rather do it when we're 100% sure of how to do regional emission factors :)

pierresegonne:
Yes, I'll update the test to make sure that the 1 way relation works (all geometries are defined in zones.json) + make a subsequent PR to remove useless zones (some in US + NZ north/south) and update some others

Issue #3475, opened 2021-11-07
East Denmark wind capacity is too low


sandos:
As can be seen in the attached screenshot, the total wind capacity is either too low or the reported wind figures are too high for East Denmark, since it is currently reported was 123%

![image](https://user-images.githubusercontent.com/60831/140651656-9680f2ef-1bfb-4ba7-856b-014cbb0afbb9.png)

KabelWlan:
Yeah I know… Still waiting for Entsoe to update their capacity list. However, in the meantime we could just add the capacity from the Kriegers Flak offshore wind farm. It came online in September and added more than 600 MW.

KabelWlan:
I think this can be closed for now. I'll update the capacities again as soon as Entsoe releases new data.

skovhus:
@KabelWlan thanks!

PR #3478, opened 2021-11-08
update wind capacity DK2

KabelWlan:
Like suggested in #3475 I added the 604 MW of capacity from the Kriegers Flak offshore wind farm.


madsnedergaard (approved):
Nice, thanks!

The DATA_SOURCES.md entry doesn't reflect this right now, but since ENTSO-E will eventually update it I think it's fine :)

PR #3480, opened 2021-11-08
Adds zone_key to a bunch of logger.warning calls

madsnedergaard:
By adding a `zone_key` field to the log entires, we can more easily filter these messages by zone instead of lumping them all together.

Some of the error log entries that should now have a zone key attached:
```
- Warning: toplam (30136) is missing in mapping!
- Sparkle Power 5 is not mapped to generation type
- Error importing data from row:
- No corresponding consumption value found for self-consumption at ...
- Self-consumption data not yet available for X at ...
- Required generation type coal is missing from *
```

### DIFF
--- a/parsers/CR.py
+++ b/parsers/CR.py
@@ -162,9 +162,12 @@ def df_to_data(zone_key, day, df, logger):
         results.append(data)
 
     for plant in unknown_plants:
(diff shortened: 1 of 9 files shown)


madsnedergaard:
Closing this, we should do this in a smarter, automated way on the `logger` itself! :)

PR #3544, opened 2021-11-15
Make EIA parser raise errors

pierresegonne:
EIA parsers have been down for a while.
The EIA endpoint returns and error, explaining that the API key is not valid anymore, but the error is not captured and does not lead to an exception being raised. 

This PR addresses this lack.

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -17,331 +17,331 @@ from .ENTSOE import merge_production_outputs
 from .lib.utils import get_token
 from .lib.validation import validate
 
-#Reverse exchanges need to be multiplied by -1, since they are reported in the opposite direction
+# Reverse exchanges need to be multiplied by -1, since they are reported in the opposite direction
 REVERSE_EXCHANGES = [
-    'US-CA->MX-BC',
-    'MX-BC->US-CAL-CISO',
-    'CA-SK->US-CENT-SWPP',
-    'CA-MB->US-MIDW-MISO',
-    'CA-ON->US-MIDW-MISO',
-    'CA-QC->US-NE-ISNE',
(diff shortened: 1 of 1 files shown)


parsers/EIA.py
```diff
@@ -361,133 +367,163 @@ def fetch_production_mix(zone_key, session=None, target_datetime=None, logger=No
...
 
+    eia_error_message = None
+    # check nested key with forgiveness
+    try:
+        eia_error_message = raw_data["data"]["error"]
+    except KeyError:
+        pass
+    if eia_error_message:
+        logger.error(f"EIA error: {eia_error_message}")
+        return []
```
madsnedergaard, line 498: Hmm, couldn't you do it like this instead?

```python
    try:
        eia_error_message = raw_data["data"]["error"]
        logger.error(f"EIA error: {eia_error_message}")
        return []
    except KeyError:
        pass
```

Or maybe even this? Not sure what's clearer:
```python
    eia_error_message = raw_data.get('data', {}).get('error')
    if eia_error_message: 
        logger.error(f"EIA error: {eia_error_message}")
        return []


```

pierresegonne: Good suggestion! Didn't know about the `.get` trick ;)

madsnedergaard (approved):
Sorry, thought I approved it again yesterday!

parsers/EIA.py
```diff
@@ -9,6 +9,7 @@
 https://www.eia.gov/opendata/register.php
 """
 import datetime
+from logging import error
```
skovhus, line 12: Do we use this import?

pierresegonne: removed it :)

PR #3545, opened 2021-11-15
[Geometries] Add new geometry process

Kongkille:
## Description
This PR changes the way in which geometries are edited. Currently the system is quite complicated and has a lot of logic from automatically regenerating the world.json from a set of sources, some public and other internal. This is really hard to edit as a contributor and there's very little visibility into what's going on.

This PR reduces the complexity drastically, in that we instead work with a single world.geojson. Due to the high amount of simplification we do on the app, there's no reason for us to have high original complexity, as such we should be fine with just one single simplified world.geojson that users can edit in directly and additionally there are some validation scripts that can help us keep the map valid.

image

- [ ] Create wiki for how to edit


madsnedergaard:
Let me know when this is ready for a review :)

Kongkille:
> Let me know when this is ready for a review :)

I just marked it ready. I think the validation.js code could be simplified and use some more JS standards. I can also look a bit more into it, but I think it's worth starting the review.

web/geo/detect-changes.js
```diff
@@ -0,0 +1,90 @@
...
+
+    return modified;
+}
+
+
+function getCombinedFeature(fc, id) {
+    // returns polygon or multipolygon
+    const polygons = fc.features.filter(x => x.properties.zoneName === id);
+    if (polygons.length > 1) {
+        return multiPolygon(polygons.map(x => getCoords(x)), { zoneName: id }); // TODO remove id
```
Kongkille, line 47: Seems there's a todo left here

web/geo/README.md
```diff
@@ -0,0 +1,12 @@
+# Overview
+
+The world.geojson file is a geojson file containing all geographies and metadata for all zones visible on the electricityMap app. The zones are also used in the backend of electricityMap.
+
+# How to update world.geojson
+
+To update geographies on the app
+
+1. Create manual changes on world.geojson
+2. Run updateworld.js
```
FelixDQ, line 10: Should we make this into an npm script?

web/package.json
```diff
@@ -44,6 +44,7 @@
     "sunrise-sunset-js": "^2.2.1",
     "swr": "^1.0.1",
     "topojson": "^3.0.2",
+    "topojson-server": "^3.0.1",
```
FelixDQ, line 47: Could this also be a devDependency?

web/geo/README.md
```diff
@@ -0,0 +1,12 @@
+# Overview
```
FelixDQ, line 1: Maybe we should add some of this to the wiki after it has been merged?

FelixDQ (commented):
Looks great! I added some small comments :-)

madsnedergaard:
Some general comments about the code :)

1. Would be nice with some linting here - we should ideally have something running automatically, but until then you should install the Prettier extension and run it (`shift+opt+F` runs `Format document` in VSCode).

2. In general it's considered bad practice to modify arguments in a function. A clean function always takes inputs and return outputs without any side effects - see examples and why here: https://github.com/ryanmcdermott/clean-code-javascript#avoid-side-effects-part-1

3. `.filter()` and `.map()` returns new arrays with the matching result, so instead of pushing objects to an array and returning that, you should take advantage of that.

web/geo/detect-changes.js
```diff
@@ -0,0 +1,90 @@
+const topojsonClient = require("topojson-client");
+const {  getCoords, featureCollection, area, multiPolygon } = require("@turf/turf")
+const fs = require("fs");
+const { getPolygons } = require("./utilities");
+
+const getJSON = (fileName, encoding = "utf8", callBack = () => { }) =>
+    typeof fileName === "string" ?
+        JSON.parse(fs.readFileSync(fileName, encoding, () => callBack())) :
+        fileName;
```
madsnedergaard, line 9: This function exists in utilities

web/geo/utilities.js
```diff
@@ -0,0 +1,64 @@
...
+    /* Transform the feature collection of polygons and multi-polygons into a feature collection of polygons only */
+    /* all helper functions should rely on its output */
+     const handlePolygon = (feature, props) => polygon(getCoords(feature), props)
+     const handleMultiPolygon = (feature, props) => getCoords(feature).map(coord => polygon(coord, props));
+ 
+     const polygons = [];
+ 
+     const dataGeojsonType = getType(data);
+     if (dataGeojsonType !== "FeatureCollection") {
+         // eslint-disable-next-line no-param-reassign
```
madsnedergaard, line 14: It's usually a bad sign if you have to override eslint rules - they are there for a reason 😉

web/geo/README.md
Kongkille: Yes, I will create a complete wiki for how to edit. Added a checkbox to the PR

web/geo/detect-changes.js
Kongkille: Just wanted to sneak in some extra code lines

web/geo/utilities.js
```diff
@@ -0,0 +1,64 @@
...
+     });
+     return featureCollection(holes);
+ }
+
+
+const newFileCallback = fileName => console.log(`Created new file: ${  fileName}`);
+const getJSON = (fileName, encoding = "utf8", callBack = () => { }) =>
+    typeof fileName === "string" ?
+        JSON.parse(fs.readFileSync(fileName, encoding, () => callBack())) :
+        fileName;
```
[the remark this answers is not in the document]
Kongkille: Nice agreed. Yes it also seems that they weren't necessary in this context

web/geo/utilities.js
skovhus: @Kongkille I see you marked this as resolved, did you push new code fixing this?

web/geo/README.md
skovhus, line 10:
```suggestion
2. Run `node update-world.js`
```

web/geo/utilities.js
Kongkille: Uh no, but I have it on my end. I decided to do a bit of a larger revamp cleaning up the functions, and it was just to keep track of these things. I expect to push it later today.

madsnedergaard (commented):
A few minor things

skovhus:
@madsnedergaard I now made updating `world.json` required, by checking that it is up-to-date on CI. This removed the need for all the detect changes scripts. Eventually the file could be removed from version control – but I'll leave that for now.

madsnedergaard (approved):
Very nice changes 👏

PR #3550, opened 2021-11-18
Improve OPENNEM parser

pierresegonne:
Related to https://github.com/opennem/opennem/issues/40

This PR makes the propagation of "null" values for solar rooftop from opennem to None solar production impossible.

### DIFF
--- a/parsers/OPENNEM.py
+++ b/parsers/OPENNEM.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from typing import Dict, List, Mapping, Tuple
 
 import arrow
 import pandas as pd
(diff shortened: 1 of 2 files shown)


parsers/OPENNEM.py
```diff
@@ -80,181 +81,267 @@ def dataset_to_df(dataset):
...
-            (zone_key and ds.get('region') == region)
-            or (sorted_zone_keys and ds.get('id').split('.')[-2] == EXCHANGE_MAPPING_DICTIONARY['->'.join(sorted_zone_keys)]['region_id'])
-        )
-    ]
-    logger.debug('Concatenating datasets..')
+    logger.debug("Parsing JSON..")
+    datasets = r.json()["data"]
+    logger.debug("Filtering datasets..")
+
+    def filter_dataset(ds: dict) -> bool:
```
pierresegonne, line 133: Extracted this function outside of list comprehension to make code easier to read

parsers/OPENNEM.py
```diff
@@ -80,181 +81,267 @@ def dataset_to_df(dataset):
...
-    if data_type == 'power' and zone_key:
+    if data_type == "power" and zone_key:
         # SOLAR_ROOFTOP is only given at 30 min interval, so let's interpolate it
-        if 'SOLAR_ROOFTOP' in df:
-            df['SOLAR_ROOFTOP'] = df['SOLAR_ROOFTOP'].interpolate(limit=5)
+        if "SOLAR_ROOFTOP" in df:
+            # TODO decide 5 or 10
+            # I set it to 10 to make sure to minimise impact of NaNs we receive from opennem
+            # https://github.com/opennem/opennem/issues/40
+            df["SOLAR_ROOFTOP"] = df["SOLAR_ROOFTOP"].interpolate(limit=10)
```
pierresegonne, line 163: Changed this to allow for a hole (and interpolate)

parsers/OPENNEM.py
```diff
@@ -80,181 +81,267 @@ def dataset_to_df(dataset):
...
+                if obj["region"] == region
+            ]
+        )
         return df, pd.Series(capacities)
     else:
         return df
 
 
-def sum_vector(pd_series, keys, transform=None):
+def sum_vector(pd_series, keys, ignore_nans=False):
```
pierresegonne, line 178: `transform` was dead code. 

Added `ignore_nans` to allow specifying if we choose to allow nans to be counted as 0 or not.

skovhus (commented):
Great stuff! It would be great to try out https://github.com/opennem/opennempy, maybe we can make an issue for this?

madsnedergaard:
As me and Pierre discussed IRL, it might make more sense to:
- Drop the entry if `solar_rooftop` or another value is `null` - we'd rather have a hole (that can later be estimated) and show a grey zone on the map, than showing wrong information
- Use the lowest common denominator for the interval instead of doing interpolating - so switch to 30 minutes interval instead of 5 minutes interval

pierresegonne:
@madsnedergaard I've pushed the changes we discussed - and added some tests. I suggest you review commit by commit  to not be polluted with black and reformatting

madsnedergaard (approved):
Awesome! 👏

PR #3553, opened 2021-11-18
Fix Bolivia parser that now includes solar production

skovhus:
#3463 

As seen here https://storage.googleapis.com/electricitymap-parser-logs/BO.html, the parser fails while trying to unpack the following list: `[hour, forecast, _total, _thermo, _hydro, _wind, _unknown] = hour_row `

It turns out that https://www.cndc.bo/gene/index.php now includes solar! 🌞 

While fixing this issue I discovered that the old parser was setting `unknown = thermo`... But I believe this should have been `unknown = total - (the_production_modes_we_extract)`. This seems to be a regression that was introduced in https://github.com/tmrowco/electricitymap-contrib/pull/2969

I recommend reviewing commit by commit.

### DIFF
--- a/parsers/BO.py
+++ b/parsers/BO.py
@@ -9,33 +9,12 @@ import requests
 
 tz_bo = "America/La_Paz"
 
+SOURCE = "cndc.bo"
+
 
 def extract_xsrf_token(html):
     """Extracts XSRF token from the source code of the generation graph page."""
     return re.search(r'var ttoken = "([a-f0-9]+)";', html).group(1)
-
-
-def template_response(zone_key, datetime, source):
-    return {
-        "zoneKey": zone_key,
-        "datetime": datetime,
-        "production": {
-            "hydro": 0.0,
-            "unknown": 0,  # Gas + Oil are mixed, so unknown for now
-            "wind": 0,
-        },
-        "storage": {},
-        "source": source,
-    }
-
-
-def template_forecast_response(zone_key, datetime, source):
-    return {
-        "zoneKey": zone_key,
-        "datetime": datetime,
-        "value": None,
-        "source": source,
-    }
 
 
 def fetch_production(
@@ -64,7 +43,7 @@ def fetch_production(
     payload = []
 
     for hour_row in hour_rows:
-        [hour, forecast, _total, thermo, hydro, wind, _unknown] = hour_row
+        [hour, _forecast, total, thermo, hydro, wind, solar, _bagasse] = hour_row
 
         if target_datetime is None and hour > now.hour:
             continue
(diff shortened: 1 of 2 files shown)


parsers/BO.py
```diff
@@ -74,64 +66,79 @@ def fetch_production(zone_key='BO', session=None, target_datetime=None, logger=N
...
+    zone_key="BO", session=None, target_datetime=None, logger=None
+) -> list:
+    """Requests the last known production mix (in MW) of a given country."""
+    payload = []
 
+    for row in fetch_data(
+        session=session, target_datetime=target_datetime, logger=logger
+    ):
+        # NOTE: thermo includes gas + oil mixed, so we set these as unknown for now
+        modes_extracted = [row.hydro, row.solar, row.wind]
```
jarek, line 99: any particular reason we don't want to report bagasse as biomass?

jarek (commented):
Nice cleanup and refactor, looking through the commits it makes sense to me (disclaimer I didn't try to run it)

parsers/BO.py
```diff
@@ -77,10 +56,24 @@ def fetch_production(
         if target_datetime is not None and hour < 24:
             timestamp = timestamp.replace(hour=hour - 1)
 
-        hour_resp = template_response(zone_key, timestamp.datetime, "cndc.bo")
-        hour_resp["production"]["unknown"] = thermo
-        hour_resp["production"]["hydro"] = hydro
-        hour_resp["production"]["wind"] = wind
+        # NOTE: that thermo includes gas + oil are mixed, so this is unknown for now
+        modes_extracted = [hydro, solar, wind]
```
pierresegonne: `_bagasse` should be fed into `biomass` (It's leftovers from sugar cane production)

parsers/BO.py
skovhus: Good catch. Will fix. :)

parsers/BO.py
```diff
@@ -12,15 +13,25 @@
 SOURCE = "cndc.bo"
 
 
+class HourlyProduction(NamedTuple):
```
pierresegonne, line 16: 🤠

parsers/BO.py
```diff
@@ -54,76 +66,73 @@ def fetch_production(
             timestamp = now
 
         if target_datetime is not None and hour < 24:
```
pierresegonne, line 68: This whole timestamp handling is hard to read. Maybe extract to a fx?

pierresegonne (commented):
Once the bagasse -> biomass is fixed, we're good to go :)

parsers/BO.py
skovhus: Fixed in https://github.com/tmrowco/electricitymap-contrib/pull/3553/commits/cbb21eed8d3483a3874278324565b01f11292a9c

parsers/BO.py
skovhus: Yes. After testing it it also turns out it had two flaws:
- hour 23 was always skipped
- when using target_datetime the timezone information was not provided, so the returned data was wrongly in UTC

This should be fixed now. The part about skipping if we are in the future is already handled in the fetch_production function as part of the `if row.total is None or None in modes_extracted` check.

pierresegonne (approved):

PR #3568, opened 2021-11-24
BG fix parser

MindFreeze:
Switched from deprecated HTML endpoint to new JSON API

### DIFF
--- a/parsers/BG.py
+++ b/parsers/BG.py
@@ -15,17 +15,9 @@ TYPE_MAPPING = {                        # Real values around midnight
     u'ВяЕЦ': 'wind',                    # 488
     u'ФЕЦ': 'solar',                    # 0
     u'Био ТЕЦ': 'biomass',              # 29
+    u'Био ЕЦ': 'biomass',               # 29
     u'Товар РБ': 'consumption',         # 3175
 }
-
-
-def time_string_converter(ts):
-    """Converts time strings into aware datetime objects."""
-
-    dt_naive = arrow.get(ts, 'DD.MM.YYYY HH:mm:ss')
-    dt_aware = dt_naive.replace(tzinfo='Europe/Sofia').datetime
-
-    return dt_aware
 
 
 def fetch_production(zone_key='BG', session=None, target_datetime=None, logger=None) -> dict:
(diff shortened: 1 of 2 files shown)


pierresegonne (approved):
Hey, thanks for your contribution.

Super nice! 

Just a few points: 

* There is an unused import (Beautiful soup) left
* you can add yourself in the contributors for the `BG` zone as defined in `config/zones.json`.

It seems like the nuclear production we get from the parser is slightly larger than the known capacity, I'll look into it

pierresegonne:
From different sources ([World nuclear](https://world-nuclear.org/information-library/country-profiles/countries-a-f/bulgaria.aspx), [ENTSOE](https://transparency.entsoe.eu/generation/r2/installedGenerationCapacityAggregation/show?name=&defaultValue=false&viewType=TABLE&areaType=BZN&atch=false&dateTime.dateTime=01.01.2021+00:00|UTC|YEAR&dateTime.endDateTime=01.01.2021+00:00|UTC|YEAR&area.values=CTY|10YCA-BULGARIA-R!BZN|10YCA-BULGARIA-R&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19), [Wikipedia](https://en.wikipedia.org/wiki/Kozloduy_Nuclear_Power_Plant) it does seem that the capacity is still at 2000MW.

MindFreeze:
I cleaned up the import.
I know the nuclear plant in BG has been working at around 105% capacity for a few months at least. This might explain the difference.

pierresegonne:
Thanks! Last nitpick: 

Could you replace 

```python
url = 'http://www.eso.bg/api/rabota_na_EEC_json.php'
response = r.get(url).json()
```

with 

```python
url = "http://www.eso.bg/api/rabota_na_EEC_json.php"
res = r.get(url)

assert (
    res.status_code == 200
), f"Exception when fetching production for {zone_key}: error when calling url={url}"

response = res.json()

logger.debug(f"Raw generation breakdown: {response}", extra={"key": zone_key})
```

after setting the `logger` argument to `logger: logging.Logger = logging.getLogger(__name__)`, with importing `import logging`?


I have a gist here with these changes if it's unclear what I mean https://gist.github.com/pierresegonne/951d10597d4284b54448b50c14aebe7b

MindFreeze:
Done. Thanks for the feedback. Python is not my native lang :)

parsers/BG.py
```diff
@@ -19,14 +21,27 @@
...
-    response = r.get(url).json()
+    res = r.get(url)
+
+    assert (
+        res.status_code == 200
+    ), f"Exception when fetching production for {zone_key}: error when calling url={url}"
+
+    response = res.json()
+
+    logger.debug(f"Raw generation breakdown: {response}", extra={"key": zone_key})
```
skovhus, line 44:
```suggestion
    logger.debug(f"Raw generation breakdown: {response}")
```

@pierresegonne FYI the extra is not needed here as the logger we parse in already has this information.

pierresegonne:
Well done :)

PR #3574, opened 2021-11-29
Fix iOS mobile app build

corradio:
- Updated cordova-ios to 6.2.1
- Update Sentry and Code push
- Updated README with better instructions
- Updated index.html (this was the culprit) based on the HelloWorld cordova example.

Works like a charm now on iOS. Didn't test for Android.
I also didn't test the new deeplinks, but rather I made sure the app runs smoothly in the simulator.


madsnedergaard (approved):
Nice! 🥳 
Let's roll it out, will you get someone at the office with an Android device to test it when it's out?

corradio:
I'll follow up on our internal issue.

skovhus:
Yay 👏

PR #3578, opened 2021-12-02
Ensure US_HI.py uses real-time data; add unit tests

danielmatsuda:
Hi, this is my first PR for electricityMap! I hope to make more contributions in the future (like fixing broken parsers or adding more tests).

For this PR, I added mock data and unit tests for the US_HI.py parser. Also, I noticed that the parser doesn’t check whether the data is within 2 hours old before returning it, so I added some code to handle this. If the data source eventually serves historical data, I believe this new code will still be useful.

Running the commands “poetry run test_parser US-HI-OA production” and “poetry run test” appear to be successful.


kepiej:
Hi @danielmatsuda! As the original author of the HI parser I would like to clarify that the datasource **does** provide historical data (up to 2013 I believe) ! When I wrote the original parser my Python (and Pandas) skills at the time were very basic and I just prioritized current data over fetching historical data. In fact, I hinted at it in a comment of the original code. For example, you can fetch all data on 05/12/2013 using https://www.islandpulse.org/api/mix?date=2013-12-05.

Having said this: maybe you (or someone else) could also look at adding historical data parsing to the parser?

danielmatsuda:
@kepiej Thanks for pointing this out! Sorry I missed this, I should have checked your discussion in #814 about the API.
I'll update my PR with support for historical data.

danielmatsuda:
I should be all set for a review now. Thanks again to @kepiej for writing the initial parser and giving initial feedback for my PR!

madsnedergaard (approved):
Great, thanks for the work here @danielmatsuda 🙌 
And welcome to electricityMap!

PR #3581, opened 2021-12-03
Updated czech word for oil from 'olej' (which means oil for cooking) …

darryl-revok:
…to 'ropa' (which means oil, as in crude oil/petroleum)


FelixDQ (approved):
Thanks!

PR #3587, opened 2021-12-07
Draft: Added current consumption parser for Canada (Saskatchewan)

r24mille:
Hello! I'm interested in bringing a Saskatchewan, Canada parser up for discussion. It's limited only to the current consumption, which I know isn't much.

**Background**
The following page shows a System Load widget in megawatts: 
https://www.saskpower.com/Our-Power-Future/Our-Electricity/Electrical-System/System-Map

It makes an AJAX call to: 
https://www.saskpower.com/ignitionapi/Content/GetNetLoad

That endpoint simply returns a number as a string. I've interpreted this to be the current consumption based solely on the website's language surrounding this number, "system load" and "net load".

**Questions**
1. Is implementing only `fetch_consumption` sufficient to get a parser started and merged? Does it provide any value? Of all functions described on the [Building a new parser](https://github.com/electricityMap/electricitymap-contrib/wiki/Building-a-new-parser) page, `fetch_consumption` is the only one I could implement from SaskPower. For that reason, I realize parser falls far short of the [Technical requirements for parser data](https://github.com/electricityMap/electricitymap-contrib/wiki/Technical-requirements-for-parser-data).
2. The [Verify data sources](https://github.com/electricityMap/electricitymap-contrib/wiki/Verify-data-sources) page gives good background. SaskPower is certainly the most  authoritative data source for this region. However, the fact that it's a single number on a system summary page doesn't provide a lot of context about how exactly the number was derived. What other data source verification steps are typically taken?

Thanks!


parsers/CA_SK.py
```diff
@@ -0,0 +1,39 @@
...
+    if target_datetime:
+        raise NotImplementedError('This parser is not yet able to parse past dates')
+
+    request = session or requests.session()
+    response = request.get(url=CONSUMPTION_URL)
+    if response.ok:
+        decoded_content = response.content.decode(response.encoding)
+        consumption = float(decoded_content.strip('"'))
+        date_header = response.headers.get('date')
+        consumption_date = arrow.get(date_header, 'ddd, DD MMM YYYY HH:mm:ss ZZZ').datetime
```
r24mille, line 23: I realize that the date of the HTTP response isn't really a great value to be used for, "the date of the data." I'm going to leave my computer up and polling for a while to sample how often this number changes. From what I've seen during development it's fairly frequently.

However, even if those results show it's fresh, I don't know how we could programmatically return a better date to describe when the consumption value was observed.

parsers/CA_SK.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return {
+            'zoneKey': zone_key,
+            'datetime': consumption_date,
+            'consumption': consumption,
+            'source': SOURCE
+        }
+    else:
+        raise ParserException('CA_SK.py', 'HTTP request for current consumption failed', zone_key)
```
r24mille, line 32: What value is intended to be used for the `parser` arg?  I noticed some other parsers are re-using the `zone_key` value. I assumed that would be better left to the `zone_key` arg.

parsers/CA_SK.py
r24mille: As promised, I set up a cron job locally to poll `CA_SK.fetch_consumption(...)` once per minute over the past day. I can share the results if requested, but this graph from a spreadsheet I quickly put together summarizes things well.

saskpower_consumption_2021-12-10

The value is updated every ~5 minutes. It's a little jumpy but seems reasonable.

parsers/CA_SK.py
```diff
@@ -0,0 +1,39 @@
...
+import arrow
+import requests
+
+from .lib.exceptions import ParserException
+
+CONSUMPTION_URL = 'https://www.saskpower.com/ignitionapi/Content/GetNetLoad'
+SOURCE = 'saskpower.com'
+ZONE_KEY = 'CA-SK'
+
+
```
r24mille, line 12: Since data update frequency is [known to be ~ 5 minutes](https://github.com/electricityMap/electricitymap-contrib/pull/3587/files#r766983812) would adding the `refetch_frequency` decorator provide any value? Aside from its docstring description, how is the decorator used by Electricity Map's ingestion pipeline?

```suggestion

@refetch_frequency(datetime.timedelta(minutes=5))
```

r24mille:
Fixes #3633.

madsnedergaard:
Hey @r24mille, sorry for never getting back to you!
The consumption number without a breakdown of production types is not ideal - especially for a zone with a high mix of energy sources: 

Screenshot 2022-01-27 at 14 39 34

While I appreciate the work you've done here, I'm afraid the data is unfortunately not up to the required standard of the app 😕

PR #3589, opened 2021-12-07
Shortens very long zone names

madsnedergaard:
### Description

We have some zones with very long names, that makes the tooltips super wide.

So here we will truncate and add ellipsis (`...`) if the title is too long.
I also changes "United States of America" to "USA" as this is mostly an issue in combination with that full country name.

**Before:**
Screenshot 2021-12-07 at 12 54 58
Screenshot 2021-12-07 at 12 56 31

**After:**
Screenshot 2021-12-07 at 12 55 30
Screenshot 2021-12-07 at 12 56 16


web/src/helpers/translation.js
```diff
@@ -39,4 +39,21 @@ export function getFullZoneName(zoneCode) {
...
+  const countryName = translate(`zoneShortName.${zoneCode}.countryName`);
+  if (!countryName) {
+    return zoneName;
+  }
+
+  if (zoneName.length > limit) {
+    return `${zoneName.substring(0, limit)}... (${countryName})`;
+  }
+
+  return `${zoneName} (${countryName})`;
```
skovhus, line 56: Wouldn’t we just do this with css and a max width?

skovhus (approved):

web/src/helpers/translation.js
madsnedergaard: Nope, the reason is that I want to shorten the zoneName, but not the countryName - so fx `Somethi... (USA)` instead of `Something (U...`.
 I guess we could split the HTML into two elements and do some work there, but I think this is much simpler for now :)

PR #3593, opened 2021-12-08
Fix build: topogen.sh was removed

skovhus:
topogen.sh was removed in https://github.com/electricityMap/electricitymap-contrib/pull/3576

This should also be removed from the Brick file and README.


skovhus:
@FelixDQ merging this now as it breaks the build. Let me know if you have any comments.

PR #3595, opened 2021-12-08
Brazil capacity update

KabelWlan:
I updated the capacities for Brazil. The north-east zone reports more solar than capacity for a few weeks already, so it's about time. I used the same source as before.


madsnedergaard (approved):
Perfect 👌

PR #3598, opened 2021-12-09
Adds zonegeometries.json file

madsnedergaard:
This should fix the problems :)


Kongkille (approved):
Nice, and then when we merge the other PR, we can deprecate this file for good and replace it with the new simplified world.geojson.

Note that if we do not want this large file in dist, I can also create a simplified version of the file which should work exactly similarly

madsnedergaard:
Reverted https://github.com/electricityMap/electricitymap-contrib/pull/3599 instead

PR #3604, opened 2021-12-17
Update world.json

skovhus:
During our build process the world.json file changes. As we have been messing with the geography scripts, then I assume this is caused by it.

Here I'm adding the world.json file that is generated. The file is much smaller than the old one. The old one contained the entire world, but the new one only contains some zones... What I don't like is that Alberta is gone. 🤔 

### Before 

Screenshot 2021-12-17 at 10 54 01

https://github.com/electricityMap/electricitymap-contrib/blob/master/web/src/world.json

### After

Screenshot 2021-12-17 at 10 53 51

https://github.com/electricityMap/electricitymap-contrib/blob/72da3e38f1e3bc66037460ab4d82dee683267262/web/src/world.json


skovhus:
The missing zones in the new world.json is a problem:

Screenshot 2021-12-17 at 15 11 02

I've tried checking out https://github.com/electricityMap/electricitymap-contrib/commit/ff24f6685bb351f3d45c1283ac57d4af7945f06d, and the issue also shows up here when rebuilding the world.json.

madsnedergaard:
Closing this again, as this will not help unfortunately - I have created https://github.com/electricityMap/electricitymap-contrib/pull/3610 which should make it possible to make releases again, and then #3545 will hopefully fix things for good :)

PR #3609, opened 2021-12-20
Fix GB wind being None

pierresegonne:
### Issue

Quite regularly, the wind production for edge hours (23:00 UTC which is 00:00 Europe/London) would be converted to None. This happens because there is a temporal gap between the data original fetched, and the wind fetched ad-hoc, meaning that we loose data for that half hour. 

Reproduce:

`poetry run python test_parser.py GB production --target_datetime 2019-09-20`

### Description

Makes sure to fetch enough data to match wind production against the other production modes, and add fail safe fallback by making wind required.

### DIFF
--- a/parsers/ELEXON.py
+++ b/parsers/ELEXON.py
@@ -238,7 +238,7 @@ def _fetch_wind(target_datetime=None):
 
     # line up with B1620 (main production report) search range
     d = target_datetime.date()
-    start = d - dt.timedelta(hours=24)
+    start = d - dt.timedelta(hours=48)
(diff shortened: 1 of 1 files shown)


madsnedergaard (approved):
Nice! 🙌

parsers/ELEXON.py
```diff
@@ -238,7 +238,7 @@ def _fetch_wind(target_datetime=None):
 
     # line up with B1620 (main production report) search range
     d = target_datetime.date()
-    start = d - dt.timedelta(hours=24)
```
corradio, line 241: Nice!

PR #3610, opened 2021-12-20
Skips building world.json during build step

madsnedergaard:
Adds a `--ignore-world` flag to the script, which we then use during build. 
Since `world.json` is committed and in the code, building it all the time (and breaking builds if it changes) does not make sense :)

Also,  this is currently blocking new releases..


pierresegonne:
Let's try 🤞

PR #3614, opened 2021-12-21
added Norwegian (bokmål) translation

haraldgroven:


web/locales/no-nb.json
```diff
@@ -0,0 +1,275 @@
...
+        "consumption": "forbruk",
+        "cpinfo": "<b>forbruk</b> inkluderer import og eksport, <b>produksjon</b> tar ikke med import/eksport",
+        "selectLanguage": "Bytt språk",
+        "lowCarbDescription": "Inkluderer fornybar og kjernekraft",
+        "dataIsDelayed": "Data er forsinket"
+    },
+    "misc": {
+        "maintitle": "Sanntids CO₂-utslipp fra strømforbruket",
+        "legend": "Forklaring",
+        "oops": "Ka farsken! Vi får ikke kontakt med serverne. Vi prøver igjen om noen sekund!",
```
madsnedergaard, line 77: Haha, is this a common expression in Norway? 😄 
Only thing I could find about it is this link from 2004: https://forum.doktoronline.no/topic/137557-dialekter/page/4/

madsnedergaard (commented):
Dette er kjempefint! :raised_hands:
One very minor comment, otherwise super cool that you added this!

web/locales/no-nb.json
haraldgroven: Haha, skidegod linguistic debugging :-)   
It's Northern-Norwegian dialect, and what linguist would regard as a "mild expletive". So the users would be more offended by the server downtime than the use of expletive

madsnedergaard: Haha fair enough, let's keep it like this then :)

madsnedergaard (approved):

haraldgroven:
@madsnedergaard If its possible to generate a list of untranslated `no_NB` strings in the new version of `electricitymaps` which is missing, I can bring the Norwegian translation to the app back.

madsnedergaard:
> @madsnedergaard If its possible to generate a list of untranslated `no_NB` strings in the new version of `electricitymaps` which is missing, I can bring the Norwegian translation to the app back.

Hey @haraldgroven, you can do that with this command :)

```
cd web 
# We forgot to rename this file earlier - if you create a PR, it would be great to include this change
mv scripts/translation-helper.js scripts/translation-helper.cjs
node scripts/translation-helper.cjs
```

PR #3622, opened 2021-12-24
Fixed PA, RU, RU-AS parsers and updated German capacities

IV1T3:
This PR resolves issue #3608. Instead of `Biogás`, `Biogas` was used as a key for translation. Thus, resulting in a KeyError.

This PR resolves the issues #3583 and #3584. 
The API at [br.so-ups.ru](https://br.so-ups.ru/webapi/api/CommonInfo/GenEquipOptions_Z2?oesTerritory[]=540000&startDate=2021.12.24) returns the key `fHour` as a string without a leading 0 in case fHour < 10 as shown below. 

```python
[   
    {   
        ...
        'fHour': '0',
        ...
    },
    {   ...
        'fHour': '1',
        ...
    },
    ...,
]
```

Thus,`arrow.get('%s %s' % (date, hour), 'YYYY.MM.DD HH', tzinfo=tz)` failed due to a ParserMatchError:
`arrow.parser.ParserMatchError: Failed to match 'YYYY.MM.DD HH:mm' when parsing '2021.12.24 0'`

Additionally, I updated the German capacities.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -1563,23 +1563,24 @@
     ],
     "capacity": {
       "battery storage": 280,
-      "biomass": 8560,
-      "coal": 43950,
+      "biomass": 8570,
+      "coal": 43960,
       "gas": 30500,
       "geothermal": 47,
-      "hydro": 4850,
+      "hydro": 4860,
       "hydro storage": 9810,
       "nuclear": 8114,
       "oil": 4380,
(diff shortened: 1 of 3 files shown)


madsnedergaard (approved):
You're on fire, amazing 🔥 🏆

Issue #3629, opened 2021-12-30
PA parser: implement fetch_consumption(...)


r24mille:
Parsing the consumption value from the _Demanda Total_ widget on https://sitr.cnd.com.pa/m/pub/sin.html aught to be fairly straightforward.

r24mille:
Fixed by #3630.

Issue #3628, opened 2021-12-30
PA parser: Implement fetch_exchange(...)


r24mille:
The implementation of `PA.fetch_exchange(...)` aught to be relatively straightforward from https://sitr.cnd.com.pa/m/pub/int.html

r24mille:
Fixed by #3630.

PR #3630, opened 2021-12-30
PA parser: Added fetch_consumption() and fetch_exchange()

IV1T3:
This PR resolves the issues #3628 and #3629.

The exchange data get parsed from https://sitr.cnd.com.pa/m/pub/int.html.
The consumption data from https://sitr.cnd.com.pa/m/pub/sin.html.

Thank you @r24mille, for pointing out the existing resources!

Since this is my first fetch_exchange() implementation I would like to draw your attention to reviewing the correct flow direction. As far as I interpreted the data on the website, a positive value on the site indicates an export from the specified country to Panama.

Additionally, I found that the reported (by Costa Rica) amount of exported energy from Costa Rica to Panama on electricityMap differs from the amount of imported energy reported by Panama. Currently CR reports to export 131MW, whereas Panama reports to import 225.86MW.


r24mille:
Awesome, thanks for the mention! So that the PR auto-closes the issues if maintainers merge it: Fixes #3628. Fixes #3629.

r24mille:
Fixes #3628

r24mille:
Looks great! Thanks for the mention. I was happily surprised to see your implementation so quickly.

parsers/PA.py
```diff
@@ -174,9 +180,92 @@ def fetch_production(zone_key='PA', session=None, target_datetime=None, logger:
...
+    net_flows = {
+      "CR->PA": net_flow_cr, # Costa Rica to Panama
+      "GT->PA": net_flow_gt, # Guatemala to Panama
+      "HN->PA": net_flow_hn, # Honduras to Panama
+      "NI->PA": net_flow_ni, # Nicaragua to Panama
+      "PA->SV": net_flow_sv, # Panama to El Salvador
+    }
+
+    # Invert sign of flows to account for correct flow direction
+    net_flows["PA->SV"] = -1 * net_flows["PA->SV"]
```
jarek, line 220: I would suggest to do this reversal earlier on, so that value of `net_flows["PA->SV"]` is correct at all points of execution. Something like this:

```suggestion
      "PA->SV": -1 * net_flow_sv, # Panama to El Salvador - invert sign to account for direction in alphabetical order
    }
```

parsers/PA.py
```diff
@@ -174,9 +180,92 @@ def fetch_production(zone_key='PA', session=None, target_datetime=None, logger:
...
 
+def fetch_exchange(zone_key1='CR', zone_key2='PA', session=None,
+                   target_datetime=None, logger=logging.getLogger(__name__)) -> dict:
+    """
+    Requests the last known power exchange (in MW) between two countries.
+    """
+
+    if target_datetime:
+        raise NotImplementedError(
+            'This parser is not yet able to parse past dates')
```
jarek, line 191: Suggest to match format in fetch_consumption:

```suggestion
    if target_datetime:
        raise NotImplementedError("This parser is not yet able to parse past dates")
```

This is longer than 80 characters, but many other lines here are also much longer than 80

parsers/PA.py
```diff
@@ -174,9 +180,92 @@ def fetch_production(zone_key='PA', session=None, target_datetime=None, logger:
...
+      "NI->PA": net_flow_ni, # Nicaragua to Panama
+      "PA->SV": net_flow_sv, # Panama to El Salvador
+    }
+
+    # Invert sign of flows to account for correct flow direction
+    net_flows["PA->SV"] = -1 * net_flows["PA->SV"]
+
+    if sorted_zone_keys not in net_flows:
+        raise NotImplementedError(
+            'This exchange pair is not implemented: {}'.format(sorted_zone_keys))
```
jarek, line 224: Similarly to above, this can be on same line. Also, since the repository `.python-version` specifies Python 3.7.9, we can use an f-string to make it shorter:

```suggestion
    if sorted_zone_keys not in net_flows:
        raise NotImplementedError(f"This exchange pair is not implemented: {sorted_zone_keys}")
```

parsers/PA.py
IV1T3: Decided to invert the sign prior to adding the value to the net_flows dict. This allows to have a cleaner dict assignment.

parsers/PA.py
IV1T3: It now matches format in fetch_consumption as of commit 2b9097e.

parsers/PA.py
IV1T3: As of commit 2b9097e it uses an f-string.

parsers/PA.py
```diff
@@ -174,9 +180,88 @@ def fetch_production(zone_key='PA', session=None, target_datetime=None, logger:
...
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    r = session or requests.session()
+    url = CONSUMPTION_URL
+
+    response = r.get(url)
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(response.text, 'html.parser')
+    consumption_div = soup.find_all('div', {'class': 'widget-small info coloured-icon'})[0]
```
madsnedergaard, line 247: I think this part of using the classes used for presentational values is a bit fragile - if they decide to change color of the box, this will break the parser. Could we use the `h5` title of "Demanda Total" to locate the relevant place instead? 

Think something like this could work, but maybe it can be done in a better way? :)

```python
    consumption_title = soup.find("h5", string=re.compile(r"\s*Demanda Total\s*"))
    consumption_val = consumption_title.find_next_sibling().text.split()[0]
```

config/zones.json
```diff
@@ -4894,7 +4894,8 @@
       "https://github.com/systemcatch",
       "https://github.com/7hibault",
       "https://github.com/nessie2013",
-      "https://github.com/q--"
+      "https://github.com/q--",
+      "https://github.com/IV1T3"
     ],
     "parsers": {
       "production": "PA.fetch_production"
```
madsnedergaard, line 4901: You should also add the consumption parser here: `"consumption": "PA.fetch_consumption"` :)

madsnedergaard (commented):
This is awesome! 🥳 
A few minor comments

parsers/PA.py
IV1T3: Yes indeed, that would likely improve the resilience of that parser! Thank you for highlighting!

parsers/PA.py
```diff
@@ -174,9 +180,88 @@ def fetch_production(zone_key='PA', session=None, target_datetime=None, logger:
...
+
+    response = r.get(url)
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(response.text, 'html.parser')
+    consumption_title = soup.find("h5", string=re.compile(r"\s*Demanda Total\s*"))
+    consumption_val = consumption_title.find_next_sibling().text.split()[0]
+
+    data = {
+        'consumption': consumption_val,
```
madsnedergaard, line 251: Got this error when running `poetry run test_parser PA consumption`:

Screenshot 2022-01-03 at 20 35 31

Which should be fixable :)

```suggestion
        'consumption': float(consumption_val),
```

madsnedergaard:
Very very last thing is a merge conflict with `zones.json` - could you update your branch? 😄

madsnedergaard (approved):
LGTM! 👏

PR #3643, opened 2022-01-04
Fix name of maskable icon

steren:
I noticed an issue with the name of the maskable icon.


steren:
There was a typo in the name. 
the manifest points at `maskable.svg` but the file was `markable.svg`
https://github.com/electricitymap/electricitymap-contrib/blob/master/web/public/manifest.json#L23

skovhus (approved):
Thanks for fixing this!

Issue #3088, opened 2021-04-13
US-CAL-LDWP is down


madsnedergaard:
The zone is down (and has been for a while).

The parser fails with the following error message:

```
invalid point: {'datetime': datetime.datetime(2021, 4, 11, 9, 0, tzinfo=tzlocal()), 'production': {'coal': -25, 'gas': 1082, 'hydro': 370, 'unknown': 192, 'solar': 3, 'wind': 52}, 'storage': {}, 'source': 'eia.gov', 'zoneKey': 'US-CAL-LDWP', 'schemaVersion': 3}, reason:US-CAL-LDWP: key coal has negative value -25
```

Anyone knows why coal is reported as negative? 🤔 

### Links

- [Zone on emap](https://www.electricitymap.org/zone/US-CAL-LDWP)
- [Kibana logs](https://kibana.electricitymap.org/app/kibana#/discover/10af54f0-0c4a-11e9-85c1-1d63df8c862c?_g=(refreshInterval:('$$hashKey':'object:232',display:'5%20minutes',pause:!f,section:2,value:300000),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!(message,extra.key,level),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:level,negate:!f,params:(query:ERROR,type:phrase),type:phrase,value:ERROR),query:(match:(level:(query:ERROR,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'96f67170-0c49-11e9-85c1-1d63df8c862c',key:extra.key,negate:!f,params:(query:US-CAL-LDWP,type:phrase),type:phrase,value:US-CAL-LDWP),query:(match:(extra.key:(query:US-CAL-LDWP,type:phrase))))),index:'96f67170-0c49-11e9-85c1-1d63df8c862c',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))

FelixDQ:
It is most likely self-consumption. We've had this happening for some zones before. The usual fix (if the value is small enough) is to set the value to 0 if it's negative.

We could do the same for this zone, but we might also want to create a general fix at some point where we always clamp the values to 0 if they are negative (within a certain threshold). Or there might be smarter ways to handle self-consumption  🤔

PR #3645, opened 2022-01-04
EIA parser: Improve negative production resilience

IV1T3:
Sometimes, EIA reports negative production values. This led to multiple @tmrow-bot "US-X-X production parser down" issues in the recent past (e.g., #3570, #3625, #3631 and #3634). Thus, the resilience of the EIA parser had to be improved.

In April 2021, @FelixDQ already stated that a general fix should be implemented (#3088).
He added that a possible solution might be to  "clamp the values to 0 if they are negative (within a certain threshold)".

In this PR I propose to not differentiate the production threshold by zone and type, but rather just the type.
Since negative values often come through self-consumption (possibly at startup), it seems reasonable to me to not seperate by zone as, generally speaking, a coal power plant, for example, should roughly consume similar levels of energy at startup independent of the zone.

There might be smarter ways to handle self-consumption. However, this fix will at least improve the resilience of the parser overall.


pierresegonne:
Super nice initiative!

One thing before we merge this one, do you know if it's physically possible to have self-consumption for renewable modes ? e.g solar & wind ? I guess for those we don't want any negative values

r24mille:
> In this PR I propose to not differentiate the production threshold by zone and type, but rather just the type.

Generally, I would agree with a generalized approach. Though, it is curious why there were only a handful of allowed negative thresholds. I have no experience with the EIA parser, but via `git blame` was able to find #2425 where they were originally added. It seems purely empirical, based on what @robertahunt saw. So, given that, generalizing the tolerances seems reasonable.

> One thing before we merge this one, do you know if it's physically possible to have self-consumption for renewable modes ? e.g solar & wind ? I guess for those we don't want any negative values
> 
> And for the other modes, have you researched if the values we already had are coherent ?

I too was curious about this. Why are these negative values present in the first place. The [EIA _Real-time Operating Grid_ documentation](https://www.eia.gov/electricity/gridmonitor/about) provides a bit of insight.

> EIA imputes for anomalous values for total net generation and net generation by energy source if the value is missing or reported as negative, zero, or at least 1.5 times greater than the maximum of past total net generation values reported by that BA with the following exceptions. Small negative net generation and zero net generation values are accepted for generation-only balancing authorities (See the Generation-only balancing authorities section above) and for City of Homestead (HST), Duke Energy Progress West (CPLW), and the Utilities Commission of New Smyrna Beach (NSB) since they may not be running their generators during all hours. Hydro net generation may also have negative values due to pumped storage, although EIA has asked balancing authorities to count pumped storage negative net generation as demand instead.

So it sounds like whenever the generation-only facility draws more power than it generates, it can have a negative value. That's likely what we're seeing.

The last sentence is particularly interesting and brought to mind the way ElectricityMap handles hydro storage and battery storage. It gets plotted as negative values. If this PR is an incremental fix, to be sure EIA parsers don't go down sporadically, is a generalized feature to reflect valid, negative generation the long-term fix?

IV1T3:
> One thing before we merge this one, do you know if it's physically possible to have self-consumption for renewable modes ? e.g solar & wind ? I guess for those we don't want any negative values

As @r24mille pointed out, it is indeed possible for selected renewable modes, i.e. hydro and battery consumption, to report negative values as they would then consume power/discharge. I am not sure whether solar or wind can report negative values. However, at this point I don't believe they do. 

> And for the other modes, have you researched if the values we already had are coherent ? 

All of the negative values that I was able to observe were coal values in the integer range [-16, -18]. The data had no distinct pattern but rather randomly fluctuated in this range. Thus, a negative threshold of -50 for coal does indeed seem a bit high to me. A small buffer of like 1.5x, i.e -18 * 1.5 = -27, would seem reasonable based on my data so far. Although there might be instances with an even higher negative value, which I just haven't been able to capture so far.

Unfortunately, I don't have any data on negative solar, wind, gas or unknown values for which I could perform the consistency check. 

> The best would be to have a distribution of all negative values, for all zones in the US per production type.

That would be awesome! :) 

>  If this PR is an incremental fix, to be sure EIA parsers don't go down sporadically, is a generalized feature to reflect valid, negative generation the long-term fix?

This PR is intended as an incremental fix improving the parser's resilience against intermittent negative values.
Regarding the long-term fix, I believe it depends on how negative production values, e.g. coal values, are currently calculated into the final production/consumption data.

madsnedergaard:
@pierresegonne, anything blocking from merging this? :)

madsnedergaard:
> @pierresegonne, anything blocking from merging this? :)

We talked and agreed on merging this as it definitely improves things!

madsnedergaard (approved):

PR #3657, opened 2022-01-11
Adds an error boundary to avoid showing a blank page

madsnedergaard:
### Description

This catches errors and shows them in a nicer way to users (instead of a blank page), while allowing users to return to the map.

#### Preview

Screenshot 2022-01-11 at 10 04 43

### DIFF
--- a/web/src/components/errorboundary.js
+++ b/web/src/components/errorboundary.js
@@ -57,7 +57,7 @@ class ErrorBoundary extends React.Component {
           <h1>Oh no, something went terribly wrong...</h1>
           <p>
             Please let us know{' '}
-            <a href="https://github.com/tmrowco/electricitymap-contrib/">on Github</a> so we can fix
+            <a href="https://github.com/electricitymap/electricitymap-contrib">on Github</a> so we can fix
             this!
           </p>
           <ErrorMessage>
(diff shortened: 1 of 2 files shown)


Kongkille (approved):
This is a great idea!

web/src/components/errorboundary.js
```diff
@@ -0,0 +1,77 @@
...
+    // TODO: Send this to Sentry
+    console.error(error, errorInfo);
+  }
+
+  render() {
+    if (this.state.hasError) {
+      const url = window.location.href;
+      return (
+        <ErrorBox>
+          <h1>Oh no, something went terribly wrong...</h1>
```
Kongkille, line 57: Perhaps just "Oh no, something went wrong ..." so it comes across less catastrophic

web/src/components/errorboundary.js
```diff
@@ -0,0 +1,77 @@
...
+          <p>
+            Please let us know{' '}
+            <a href="https://github.com/tmrowco/electricitymap-contrib/">on Github</a> so we can fix
```
skovhus, line 60:
```suggestion
            <a href="https://github.com/electricitymap/electricitymap-contrib">on Github</a> so we can fix
```

web/src/components/errorboundary.js
skovhus:
```suggestion
          <h1>Oh no, something went wrong...</h1>
```

PR #3658, opened 2022-01-11
Fixes crash on specific, small zones 

madsnedergaard:
### Description

After the new geojson setup, clicking on some zones crashes the app: https://app.electricitymap.org/zone/US-SW-SRP
Screenshot 2022-01-11 at 10 21 53

This happens due to the structure we're expecting, which apparently now may differ depending on the zone. I've only experienced the issue on certain US zones, but hopefully this will catch all of these errors :)

### DIFF
--- a/web/src/helpers/map.js
+++ b/web/src/helpers/map.js
@@ -1,6 +1,6 @@
 import { max, min, mean, isArray } from 'lodash';
 
-export function getCenteredLocationViewport([longitude, latitude]) {''
+export function getCenteredLocationViewport([longitude, latitude]) {
   return {
     width: window.innerWidth,
     height: window.innerHeight,
--- a/web/src/helpers/map.js
+++ b/web/src/helpers/map.js
@@ -1,6 +1,6 @@
-import { max, min, mean } from 'lodash';
+import { max, min, mean, isArray } from 'lodash';
 
-export function getCenteredLocationViewport([longitude, latitude]) {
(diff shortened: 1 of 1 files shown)


web/src/helpers/map.js
```diff
@@ -1,6 +1,6 @@
-import { max, min, mean } from 'lodash';
+import { max, min, mean, isArray } from 'lodash';
 
-export function getCenteredLocationViewport([longitude, latitude]) {
+export function getCenteredLocationViewport([longitude, latitude]) {''
```
skovhus, line 3: This looks a bit funky... `''`

skovhus (changes requested):

pierresegonne (approved):
Is is normal that the format is not consistent? 
I guess it's because now a zone can consist of several polygons? 

LGTM anyhow

web/src/helpers/map.js
skovhus: Why is the linter not catching this?

madsnedergaard: Dunno, but thanks for catching it! Fixed now

skovhus (approved):

skovhus:
I'm pretty sure this is caused by the arc logic here that is specific to US: https://github.com/electricitymap/electricitymap-contrib/blob/master/web/src/helpers/topos.js#L10

PR #3659, opened 2022-01-11
Fix Bahamas geometry

skovhus:
Bahamas is currently not shown correctly as it contains duplicate features in the geojson file. 

In a follow up PR I'll be introducing checks that guards against this.

### Preview 

Ignore the coloring.

#### Before
Screenshot 2022-01-11 at 20 55 11

#### After
Screenshot 2022-01-11 at 20 55 20


madsnedergaard (approved):
Good catch!

Kongkille (changes requested):
This is quite curious.

This seems to be the island of New Providence.

Instead of removing it, I believe the correct solution would be to merge the two bahamas features, as otherwise we are missing New Providence.

This would be the full bahamas feature:
[bahamas.json.zip](https://github.com/electricitymap/electricitymap-contrib/files/7846971/bahamas.json.zip)

PR #3660, opened 2022-01-11
Pre-compute zone center compile time

skovhus:
Avoid doing unnecessary work runtime by moving the computation of geometry centers to compile time. This also means that https://github.com/electricitymap/electricitymap-contrib/pull/3658 would have failed on CI.

Here we also remove the US specific arc workaround, that I cannot see does anything, but likely caused the issue in https://github.com/electricitymap/electricitymap-contrib/pull/3658


web/src/helpers/topos.js
```diff
@@ -5,13 +5,10 @@ const constructTopos = () => {
...
-      geo = topojson.feature(topo, topo.objects[k]);
-    } else {
-      geo = { geometry: topojson.merge(topo, [topo.objects[k]]) };
-    }
+    const geo = {
+      geometry: topojson.merge(topo, [topo.objects[k]]),
+      properties: topo.objects[k].properties
+    };
     // Exclude zones with null geometries.
     if (geo.geometry) {
```
Kongkille, line 13: This check would also not be necessary any longer. If this function excludes any geometries, we might also not want it to exclude silently as this suggests the world.json is corrupted

Kongkille (approved):
I ended up only commenting about the other things in the PR, sorry about that. I think it makes sense, it's nice that we continue to use the turf methods :)

web/src/helpers/topos.js
```diff
@@ -5,13 +5,10 @@ const constructTopos = () => {
...
-    // Do merge inner arcs for those
-    if (['US'].indexOf(k.split('-')[0]) !== -1) {
-      geo = topojson.feature(topo, topo.objects[k]);
-    } else {
-      geo = { geometry: topojson.merge(topo, [topo.objects[k]]) };
-    }
+    const geo = {
+      geometry: topojson.merge(topo, [topo.objects[k]]),
+      properties: topo.objects[k].properties
+    };
```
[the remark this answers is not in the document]
skovhus: I'll leave it up to a follow up PR to clean this further up. But thanks for explaining this. 👏

web/src/helpers/topos.js
skovhus: Nice! 👏

PR #3661, opened 2022-01-11
Enable geojson clean up in the zone removal script

skovhus:
We currently have a script for removing zones (`python scripts/remove_zone.py`). With the new geometry process we can now help out removing a zone from the geojson file.

To make the output generated reviewable, then I've used Prettier to consistently format the geojson file.

### DIFF
--- a/scripts/remove_zone.py
+++ b/scripts/remove_zone.py
@@ -5,7 +5,7 @@ poetry run python scripts/remove_zone.py DK-DK1
 
 import argparse
 
-from utils import JsonFilePatcher, ROOT_PATH, LOCALE_FILE_PATHS
+from utils import LOCALE_FILE_PATHS, ROOT_PATH, JsonFilePatcher, run_shell_command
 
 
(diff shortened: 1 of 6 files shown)


Kongkille (approved):

scripts/remove_zone.py
```diff
@@ -34,6 +34,15 @@ def remove_zone(zone: str):
             if zone in zone_short_name:
                 del zone_short_name[zone]
 
+    geo_json_path = ROOT_PATH / "web/geo/world.geojson"
```
FelixDQ, line 37: Does this work?
```suggestion
    geo_json_path = ROOT_PATH + "/web/geo/world.geojson"
```

FelixDQ: nvm it does :D

skovhus: Yeah https://docs.python.org/3/library/pathlib.html is pretty cool. ;)

Issue #3575, opened 2021-11-29
Feature development: Estimated data on the map 🗺️ 🔨


madsnedergaard:
## Description

tl;dr: **It's finally time to put a color on the delayed zones (including US)!**

Over the years we have gathered a lot of historical data for zones that are currently grey on the map, because the data is too delayed. 

In the last year we've been working hard on the backend to put infrastructure and modelling in place to infer a high-quality real-time value and finally do justice to the hard work that's been done on on verifying and implementing data sources for these zones.

### Current status

First prototype is live here: https://app.electricitymap.org/map?feature=estimations - please check it out and write your thoughts/ideas as comments here :)

---

### An open process

electricityMap is build by you and us together, and we want to take that further 💪 

_Therefore the design and implementation of this feature will be open and transparent for everyone in the community!_

In this issue we will share how we're going to do it, why we're doing it and share progress as well as ask for feedback and testing of early prototypes from you all. We sincerely hope you'll help us by sharing your inputs, ideas and help test the early iterations 💚 

## Thoughts behind this feature

In the next decade, the transition to a low-carbon electricity system will require every organisation and individual to understand and reduce their real electricity footprint. We believe our data and the information from the app is highly relevant to support this transition. 

However, it's not all zones in which we are currently able to provide this information at every hour of the day. For some cases, the data provider has an outage, other data providers may have long delays of several days (or in some cases several months). Then there are those zones in which there simply aren't any hourly data at all.

All of this makes it more difficult, or even impossible, to understand and reduce the real electricity footprint in all parts of the world without further action. 

In response, the electricityMap team has over the last year worked on providing high quality estimations in the API that is able to cover the gaps and delays from data providers, and in some cases, even estimate the carbon intensity for countries with aggregated or consolidated data. 

So far, the results have been really good and we are super excited to finally bring this data to the app.

madsnedergaard:
## The roadmap

#### Version 0.1

**Goal:** Show estimated carbon-intensity for zones that have delayed data

_Features:_

- Delayed countries will be colored on the map!
- Estimated carbon intensity, low-carbon and renewable percentages will be shown
- Production breakdown, imports and exports will be disabled for this first iteration
- The estimation model will be very simple in the first iteration, then switched to more advanced models later

#### Version 0.2

**Goal:** Expand to show all estimated data + better estimation models

_Features:_
- Enabled graphs and production breakdown with estimated data
- Implements feedback for contributors
- Improved estimation models
[...]

PR #3664, opened 2022-01-12
Adds "estimated" label and text below breakdown

madsnedergaard:
### Description

Based on feedback in #3575, here's a suggestion for adding an "estimated" label. 
There's also room for calling it "estimated data", at least in English - but not sure if it's required/adds value :)

#### Preview
Screenshot 2022-01-12 at 16 15 45
Screenshot 2022-01-12 at 16 15 58


web/src/components/countrytableoverlayifnodata.js
```diff
@@ -37,11 +33,12 @@ const CountryTableOverlayIfNoData = ({ zoneTimeIndex }) => {
 
   // TODO: Shouldn't be hardcoded
   const zonesThatCanHaveZeroProduction = ['AX', 'DK-BHM', 'CA-PE', 'ES-IB-FO'];
-  const zoneHasProductionData = (zoneData.production && Object.values(zoneData.production).every(v => v !== null)) || zonesThatCanHaveZeroProduction.includes(zoneId);
-  const isEstimated = zoneData.estimationMethod !== null;
+  const zoneHasProductionValues = zoneData.production && !Object.values(zoneData.production).every(v => v === null);
+  const zoneHasProductionData = zoneHasProductionValues || zonesThatCanHaveZeroProduction.includes(zoneId);
+  const isEstimated = !isNil(zoneData.estimationMethod);
```
skovhus, line 38: Nitpick: but shouldn't we avoid lodash here? 
```suggestion
  const isEstimated = zoneData.estimationMethod != null;
```

skovhus (approved):
Looks good to me. 👏

web/src/components/countrytableoverlayifnodata.js
madsnedergaard: We need to check for both `undefined` and `null` :)
Alternatively I considered checking if it's a string, but that also feels error-prone.

skovhus: > We need to check for both undefined and null :)

Hence `!=` and not `!==`. ;)

But why do we actually need to check both, the backend will return `null`?

Kongkille (approved):
Looks great!

web/src/components/countrytableoverlayifnodata.js
madsnedergaard: Ah, missed that part :D
`zoneData` is initially empty, so that's why it can be `undefined`... We could probably improve this, but I think that would be part of a larger revamp :)
I've changed to `!=` now.

PR #3666, opened 2022-01-13
Remove US-MIDA-OVEC

skovhus:
- Closes https://github.com/electricitymap/electricitymap-contrib/issues/3343
- Replaces some previous work in https://github.com/electricitymap/electricitymap-contrib/pull/3359

With the new geometry process this was pretty easy. The zone removal script is now also handling updating the mockserver state files.

#### Before
Screenshot 2022-01-13 at 08 52 31

#### After
Screenshot 2022-01-13 at 08 52 45


mockserver/public/v4/state
```diff
@@ -10048,8 +10048,8 @@
         "source": "demanda.ree.es",
         "stateDatetime": "2022-01-04T08:00:00Z",
         "storage": {
-          "battery": -0,
-          "hydro": -0
+          "battery": 0,
+          "hydro": 0
```
madsnedergaard, line 10052: While these changes makes sense, it's not what our app-backend serves - so I'm not sure if we really want to do this "cleaning"?

madsnedergaard (commented):
One small comment, otherwise LGTM 🥳 
And nice that removal is now easier!

mockserver/public/v4/state
skovhus: This is caused by Python not respecting these negative 0 values when formatting the JSON string. The negative 0 doesn't have any semantic meaning in our system, but is an effect of the app-backend `negate_if_not_none` function.

madsnedergaard (approved):

PR #3667, opened 2022-01-13
Releases the estimations-feature publicly

madsnedergaard:
### Description

_Part of #3575_

We've decided to remove the feature-flag and make the initial version available for everyone 🥳 
Some changes have been made based on feedback and we think it makes more sense to just open this up for everyone now, instead of waiting until we have finished the improved estimations model.

For now we're keeping graphs and historical data as-is, as switching this in some cases introduce misleading visual graphs.


skovhus (approved):
👏

PR #3678, opened 2022-01-17
[AW] Fix invalid point due to negative value of "unknown"

Nico1320:
Suggestion and fix for "unknown" value

`invalid point: {'zoneKey': 'AW', 'datetime': datetime.datetime(2022, 1, 17, 17, 30, 3, 441000, tzinfo=tzfile('/usr/share/zoneinfo/America/Aruba')), 'production': {'oil': 98.6903305053711, 'wind': 16.2064266204834, 'solar': 0.3709530830383301, 'biomass': 0.0, 'unknown': -0.8897500038146973}, 'storage': {}, 'source': 'webaruba.com', 'schemaVersion': 3}, reason:AW: key unknown has negative value -0.8897500038146973     `


jarek (commented):
consider only doing this for small values of `unknown`

Nico1320:
> consider only doing this for small values of `unknown`

What do you suggest doing with "unknown?" When "unknown" is negative it means with all power sources combined it surpasses current_total. inversely, if it's a positive number, it doesn't reach current_total. weird situation.

jarek:
Some of it might be due to rounding or data being reported from different sources at slightly different times.

A value of -0.8898 MW is not likely to be an issue, because it represents about a 0.78% error over entire generation. I think it might be better to just report 0 rather than inverting the sign, considering that the value is calculated.

But I would suggest to check how large the calculated `unknown` is and raise an error if it is, say, 5% of total generation, to avoid cases where we might have a -100 MW or +100 MW unknown being accepted - that would be a case of some more serious data problem.

Nico1320:
@jarek This is what i would purpose... agreed?
`    unknwon_check = float(total['value']) - float(fossil['value']) - float(wind['value']) - float(solar['value']) - float(biogas['value'])
    if unknwon_check > 0:
        unknown = unknwon_check
    else:
        unknown = float(0)`

Nico1320:
@jarek This is what I would purpose... agreed?
```
    unknwon_check = float(total['value']) - float(fossil['value']) - float(wind['value']) - float(solar['value']) - float(biogas['value'])
    if unknwon_check > 0:
        unknown = unknwon_check
    else:
        unknown = float(0)
```

jarek (commented):
That change would accept the datapoint if the source reported that total generation was 100 MW, generation from oil was 200 MW, and from wind 50 MW. A suggestion on how to check for that:

parsers/AW.py
```diff
@@ -24,7 +24,12 @@ def fetch_production(zone_key='AW', session=None, target_datetime=None, logger=N
     # biogas live value is 0 MW all the time (2021)
     biogas = top_data['total_bio_gas']
     total = top_data['TotalPower']
-
+    # if "unknown" exceeds total power, it becomes 0
+    unknwon_check = float(total['value']) - float(fossil['value']) - float(wind['value']) - float(solar['value']) - float(biogas['value'])
+    if unknwon_check > 0:
+        unknown = unknwon_check
+    else:
+        unknown = float(0)
```
jarek, line 32:
```suggestion
    # "unknown" is when data reported in the categories above is less than total reported.
    # If categories sum up to more than total, accept the datapoint, but only if it's less than 2% of total.
    # This helps avoid missing data when it's a little bit off, due to rounding or reporting
    reported_total = float(total['value'])
    sources_total = float(fossil['value']) + float(wind['value']) + float(solar['value']) + float(biogas['value'])

    if (sources_total / reported_total) > 1.02:
        raise RuntimeError(f'AW parser reports fuel sources add up to {sources_total} but total generation {reported_total} is lower')

    missing_from_total = reported_total - sources_total
    unknown = missing_from_total if missing_from_total > 0 else 0
```

jarek (commented):
That change would accept the datapoint if the source reported that total generation was 100 MW, generation from oil was 200 MW, and from wind 50 MW. A suggestion on how to check for that:

parsers/AW.py
Nico1320: Brilliant!

madsnedergaard (approved):
Looks great now, let's merge it! 💪

PR #3683, opened 2022-01-18
Detail Russia's production and update french names

JeanBaptisteScellier:
- distribute unknown between oil, gas, coal and unknown. Distribute renewables between wind and solar. Update capacities.
remove co2 calculation for unknown of russia because it is now distributed. Unknown is really unknown now.
- Updated some french names


gwpicard:
Hi @JeanBaptisteScellier, thanks for opening this PR!

We want to avoid distributing production manually in parsers using ratios, as this goes against our approach to parser data. Parsers should only return the true raw data that is available, and we would only break down something like unknown downstream using an estimation model.

The only thing we break down using ratios are the emission factors for unknown production in a zone, as these are typically based on yearly consumption mix averages in the zone.

I would recommend reverting the changes to the `RU.py` parser `co2eq_parameters`, and only keeping the capacity updates and french names. Hope that's ok!

gwpicard:
That sounds good for the solar into unknown.

We should still keep the new capacities as these are data points we have (as opposed to estimated). We still show capacity on the map even if we don't have specific production values for the production mode

gwpicard:
@JeanBaptisteScellier Would you also be able to link the source of the capacity data you found here? So that we can double check it

alixunderplatz:
@gwpicard @JeanBaptisteScellier in the past, the "renewables" category of the RU data was clearly reporting "solar" only. 0 MW at night and a characteristic solar output curve during daytime. Did it change significantly because it really does contain wind data now? Judging by the data on the map it seems like it's still solar only.

gwpicard:
If we can confirm that it indeed _only_ contains solar, then I think that is fine. 
As I said before, even if wind values are negligible in RU renewables, we want to avoid making assumptions or splitting things up. In this case, for example, they decide to suddenly report wind in the renewables category. I haven't personally into this parser before so I can't comment--you both know best for this parser! :)

JeanBaptisteScellier:
@alixunderplatz Currently there are no wind electricity capacities in Siberia. However in European Russia there are almost as much capacity for wind than for solar. You can see a little production of renewable electricity during last nights which must be wind.
@gwpicard my source is here: https://minenergo.gov.ru/node/20706 especially p 76 and 77.

JeanBaptisteScellier:
@alixunderplatz you can see here the detail per region of renewable energy capacity per hour: https://br.so-ups.ru/VSVGO/PowerVIEByRegions
You can see for example in Kalmykia some capacity at night. Solar seems more dominating that what I expected though.
So in the modifications I did, I let solar for Siberia and put renewable in unknown for Europe-Volga.

JeanBaptisteScellier:
You made me think more. I just checked the profile of the consumption we get for 'P_REN' and the profile is clearly solar. Strange. I need to investigate.

JeanBaptisteScellier:
I think the key is the БС or BS. I finally found a definition of what it is.
"Block station: a power plant operating in the energy system and operationally controlled by its dispatch service, but not included in the number of enterprises of the system by departmental affiliation"
So the mix of a БС is unknown and may not only be a thermal power plant, and obviously all wind production seems to be in it.

JeanBaptisteScellier:
So indeed, in the past, P_REN had always zero values in the past. But as its definition is renewable energy, it may contain wind in the future. It is a good coincidence that yesterday was the first day that we could see a production of renewable electricity during the night. Very very tiny but not null.
As the capacity of wind is far more than what we can see in P_REN, it is obvious that it was in BS before.
But I am reverting my changes for renewables.
 
So now there is a little question about how we calculate CO2 emissions of unknown, because maybe BS is not only fossil fuel. But I did a little calculation based on p 69 of my source and in European Russia where we have the wind capacity, wind is not more than 0.5% of electricity produced from (fossil fuel + wind) so we can keep them as it is.

To conclude, my contribution of this PR is that I added some more comments in the RU.py file and updated capacities.

parsers/RU.py
```diff
@@ -18,15 +18,24 @@
...
+    'P_tes': 'fossil fuel' # Fossil fuels
+}
+RENEWABLES_RATIO = {
+    'RU-1': {'solar': 0.5, 'wind': 0.5},
+    'RU-2': {'solar': 1.0, 'wind': 0.0}
+}
+FOSSIL_FUEL_RATIO = {
+    'RU-1': {'coal': 0.060, 'gas': 0.892, 'oil': 0.004, 'unknown': 0.044},
+    'RU-2': {'coal': 0.864, 'gas': 0.080, 'oil': 0.004, 'unknown': 0.052},
+    'RU-AS': {'coal': 0.611, 'gas': 0.384, 'oil': 0.005, 'unknown': 0.00}
```
gwpicard, line 38: How come you still have these ratios in the parser?

madsnedergaard:
Closing due to inactivity - feel free to reopen later :)

PR #3690, opened 2022-01-20
Updated packages and added support for python 3.8.12

FelixDQ:
Upgraded `pandas` and `opencv-python` to now support python 3.8.12.

3.7.9 still seems to work fine, but I still upgraded CI scripts to use 3.8.12 instead.


skovhus (approved):
Did you check the change logs for any potentially breaking changes? Not sure if these packages follows Semver. :)

FelixDQ:
Yea, the only notable change is that the minimum numpy version has been changed from `1.15.4` to `1.17.3`

PR #3691, opened 2022-01-20
KR parser: Integrate new real-time data sources

### LINKED ISSUE #3676: KR production parser down
## Description

This is an automatic error report generated for South Korea (KR).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser KR production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/KR.html">Explore the runtime logs</a>

IV1T3:
Updated the parser, since a new website had been released: https://new.kpx.or.kr.

The goal of this PR is to fix #3676 and improve its functionality.


IV1T3:
Current ToDo list:

- [ ] Implement fetch_consumption from same source as fetch_production
- [ ] Analyze whether hydro might be in etc or renewables
- [ ] Check for negative values

madsnedergaard:
This is awesome, looking forward to getting KR back up in an even better shape! 🚀

IV1T3:
This PR is now feature complete.

Unfortunately, the price is not updated on an hourly basis through the night - resulting in nightly delays. Don't know the exact time yet when they release the prices of the night before. Is it still possible to use the currently implemented fetch_price() and somehow fetch the data afterwards?

parsers/KR.py
```diff
@@ -1,212 +1,295 @@
...
+          'biomass': 0.0,
+          'coal':0.0,
+          'gas': 0.0,
+          'hydro': 0.0,
+          'nuclear': 0.0,
+          'oil': 0.0,
+          'solar': 0.0,
+          'wind': 0.0,
+          'geothermal': 0.0,
+          'unknown': 0.0,
```
madsnedergaard, line 192: I'm wondering if it would be better to not set these values to 0 by default, as it means we'll show `solar=0` in the app instead of solar=`?` :)

madsnedergaard (commented):
Yay!! 🥳 
I only have one pretty minor comment, otherwise it looks great!

> Unfortunately, the price is not updated on an hourly basis through the night - resulting in nightly delays. Don't know the exact time yet when they release the prices of the night before. Is it still possible to use the currently implemented fetch_price() and somehow fetch the data for the night in the next morning on your end?

Hmm, not 100% sure about this, but I think it will now save `0.0` until there's data which should then overwrite with actual values - guess we can merge this and see :)

IV1T3:
> Hmm, not 100% sure about this, but I think it will now save 0.0 until there's data which should then overwrite with actual values

As of now, during the night the price parser will always return the last available price (!= 0.0) at midnight, since that's the last one published before the night. Thus, should I rather then return the time-wise hourly correct value of 0.0 through the night?

parsers/KR.py
IV1T3: Ok! Now, I just set production and storage to an empty dict. :)

madsnedergaard (approved):
Brilliant, I guess everything is great now - so let's get it merged 😄

PR #3696, opened 2022-01-23
Update DO.py fuel type mapping

alixunderplatz:
Some of the newer power plants operate either on natural gas (GN) or fuel oil, which is reflected by the endings of the power plant names.
This will help reducing the "unknown" category back to 0, as some changes were made to the original power plant names :)


madsnedergaard (approved):
Nice!

Issue #3408, opened 2021-09-23
Change MD parser to ENTSO-E


alixunderplatz:
Hi all 😃 
it's been a while since my last contribution to this project, but here we go:
Looks like we could use the ENTSOE.py parser for two more countries :)

-----------------------
**UKRAINE**
[Transparency data for Ukraine](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=false&viewType=TABLE&areaType=CTY&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=23.09.2021+00:00|CET|DAYTIMERANGE&dateTime.endDateTime=23.09.2021+00:00|CET|DAYTIMERANGE&area.values=CTY|10Y1001C--00003F!CTY|10Y1001C--00003F&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2))'s electricity production is disclosed via the ENTSOE transparency platform since 01 September 2021.
Looks like the data is coherent.

Advantages:
- Split of Solar and Wind data. 
- Split of hydro and pumped storage production.
electricityMap could use it as backup data or for the live parser, I haven't compared to the recent data source.

Country code for the zone is "CTY|10Y1001C--00003F"

-----------------------
**KOSOVO**
[...]

PR #3702, opened 2022-01-23
Add cross-border flows with Kosovo

alixunderplatz:
The zone Kosovo has exchanges with Albania (AL->XK; formerly displayed as AL->RS, which I still left in the code in case you run historical data), with Montenegro (ME->XK), with North Macedonia (MK->XK) and with Serbia (RS->XK). Serbias parallel interconnections with Montenegro and North Macedonia remain unaffected and show the remaining power flows on ENTSOE transparency. Only Serbia<->Albania now shows N/A, as the power lines run through Kosovo B substation in Kosovo.

I added the new interconnections to exchanges.json, but didn't test. However, I think it should work, as your ENTSOE.py already fetches the exchanges with Kosovo and its neighboring zones? Edit: I added the zone_key to ENTSOE.py.


alixunderplatz:
I added the zone key of Kosovo 10Y1001C--00100H to the list in ENTSOE.py, so the transmission data should be fetched now.
You could consider switching the production data source for XK to ENTSOE as well, as it seems more reliable :) see #3408

madsnedergaard (approved):
Nice, looks good to me 🥳 
And I quickly tested with `poetry run test-parser "ME->XK" exchange` which gives valid data:
Screenshot 2022-01-27 at 15 51 47

Issue #3562, opened 2021-11-21
New Zealand CO2 Emissions Data Source has been updated...


ChirpyTurnip:
Hi,

The NZ source of the CO2 data has been updated. The news release is here: https://www.ems.co.nz/news/carbon-tracker/

The new data source is here: https://app.em6.co.nz/

Is it possible to update the scraper so that the data once again appears on the map? 

Cheers!

CP.

drc38:
It looks like the parser uses transpower's live data as the main source https://www.transpower.co.nz/power-system-live-data
and that it no longer provides North and South Island data, but rather "New Zealand" combined.
```
"},"soPgenGraph":{"data":{"New Zealand":{"Battery":{"generation":-0,"capacity":1},"Co-Gen":{"generation":138,"capacity":216},"Coal":{"generation":0,"capacity":750},"Gas":{"generation":385,"capacity":1280},"Geothermal":{"generation":876,"capacity":1062},"Hydro":{"generation":3735,"capacity":5415},"Liquid":{"generation":0,"capacity":156},"Wind":{"generation":116,"capacity":1040},"total":5249.406}},"date":"24 Nov 2021 10:00","timestamp":1637701200,"update":true},"soRenewablePgenGauge":{"data":[90.04],"date":"24 Nov 2021 10:00","timestamp":1637701200,"update":true}
```
I will have a go modifying the parser based on the above.

madsnedergaard:
Would it be possible to use some of the other API endpoints in order to maintain the difference between South and North? 

Merging the two zones will be hard due to our infrastructure and existing data, so would be great if we could find another way to do it :)

https://www.ems.co.nz/em6-api-integration-guide/

drc38:
The em6 service is a paid for api, would the project be able to cover the cost?

ChirpyTurnip:
Do we really need the difference between north island and south island? I'm
not even sure if the integration asked me what island I was on when it was
added to home assistant. And, given that the HVDC link could be be flowing
in either direction, or in both direction simultaneously (yes, it does,
it's called 'round power') the difference is mostly academic. And having an
indication of overall renewable content is better than none.

Alternatively we could ask Transpower for a feed specifically for what we
need. I have a contact there....there's no harm in asking. If so, someone
will need to provide me with a technical blurb to pass on to them...

CP.

jarek:
> Do we really need the difference between north island and south island?

[...]

PR #3708, opened 2022-01-25
Merges South and North New Zealand

madsnedergaard:
### Issue

#3562 

### Description
This merges the geometries of the two islands together into one zone.

Depends on #3567 - CI will fail until that is merged!


Kongkille (approved):
Nice

The GH geojson-viewer is actually able to display the world.geojson quite well, which is a nice way to quickly get a look when someone is updating the geometries.

PR #3709, opened 2022-01-25
Add fetch_price method in GB parsers based on fetch_price FR parser

AlexandreCouderc:
GB prices are available in the data used by the FR parser to fetch prices for FR.


madsnedergaard (approved):
Haha, nice "hack" to use French data for getting GB prices 😄

madsnedergaard (approved):
Okay, I was a bit too fast here - there's a few chances required before we can merge this :)

1) The parsers you have changed here are for Northern Ireland and the Orkney Islands. We should probably create a new file for this :)
2) The `zones.json` file needs to be updated to use this new parser for `price` :)

madsnedergaard (changes requested):
Okay, I was a bit too fast here - there's a few chances required before we can merge this :)

1) The parsers you have changed here are for Northern Ireland and the Orkney Islands. We should probably create a new file for this :)
2) The `zones.json` file needs to be updated to use this new parser for `price` :)

madsnedergaard (approved):
Brilliant, thanks for the very quick iteration 🙌

madsnedergaard:
Closes https://github.com/electricitymap/electricitymap-contrib/discussions/3706 and https://github.com/electricitymap/electricitymap-contrib/issues/2922 🥳

parsers/GB.py
```diff
@@ -0,0 +1,66 @@
...
+          'teDeb={}&dateFin={}&mode=NORM'.format(formatted_from, formatted_to)
+    response = r.get(url)
+    obj = ET.fromstring(response.content)
+    datas = {}
+
+    for donnesMarche in obj:
+        if donnesMarche.tag != 'donneesMarche':
+            continue
+
+        start_date = arrow.get(arrow.get(donnesMarche.attrib['date']).datetime, 'Europe/London')
```
AlexandreCouderc, line 42: I have a doubt here if I should put 'Europe/London' or 'Europe/Paris' for the timezone.

madsnedergaard: That is a good question 😄 
Let's see how the data looks now and then we can follow-up on it if it's clear that it's off :)

PR #3715, opened 2022-01-27
Fix OPENNEM test_process_solar_rooftop

pierresegonne:
## Issue

Test failing on CI

## Description

I could replicate the failure after updating to python 3.8.12. It seems like it was simply a numerical precision issue.


madsnedergaard (approved):
Great! 🥳

Issue #3699, opened 2022-01-23
Get AR.py working again - use other data sources


alixunderplatz:
Hi folks,

seems like our old AR.py parser for Argentina isn't working anymore.
However, there is an option to get the data back running, with more granularity than ever and to simplify the parser in the same run. No more need for detailed plant mappings.

[As I mentioned about 1 year ago in this comment](https://github.com/electricitymap/electricitymap-contrib/issues/2378#issuecomment-751355231), there are 2 websites that can be used to get the total generation mix and the renewable mix.
From this general website, we can get nuclear, large hydro, and thermal (mostly gas, very little oil and even less coal, so we'll put this to the "unknown" category or entirely to "gas"). 
Discard "renewables", as we will get the mix from the second source.
https://microfe.cammesa.com/demandaregionchart/#/operaciones
From the renewable website, we can get small hydro, biomass, solar and wind.
Adding both in a new parser should do the job.

<img width="1432" alt="image" src="https://user-images.githubusercontent.com/25743609/150689478-4035f612-4904-4c06-8b2c-9b216b69b6a8.png">

Renewables:
https://cdsrenovables.cammesa.com/renovableschart/#/totalesLineAndPie

<img width="1432" alt="image" src="https://user-images.githubusercontent.com/25743609/150689502-77eb97df-69f5-465d-ab78-5a3921709080.png">

Regarding fixing the exchanges with its neighbouring zones, I'd neglect this for the moment, until we have the data back on the map. Let's come back to that at a later point.
See the other two open AR-related issues #2378 for what else we could implement for AR and its national control areas and #565 for an older general issue with some more insights on the grid topology.

alixunderplatz:
@aemartinez @madsnedergaard happy to see Argentina back on the map! :) The data looks great in the chart!
Fetching the exports/imports with Chile, Paraguay, Uruguay and Brasil will be the next step to implement, then rewriting will be done 👍 I'll take a look at the data sources this weekend, will keep you updated ;)

![grafik](https://user-images.githubusercontent.com/25743609/153183969-06a5b841-8a8e-4778-9eaf-c8cdfbaf3177.png)

alixunderplatz:
hm, fetching the data doesn't seem to work since yesterday:
![grafik](https://user-images.githubusercontent.com/25743609/153354777-31516026-8296-4320-9031-1ae53ebf83d7.png)

madsnedergaard:
[...]

PR #3723, opened 2022-01-30
Reimplementing AR parser from scratch.

aemartinez:
The goal of this PR is to address #3699 . I was able to implement `fetch_production` from scratch consuming CAMMESA's APIs. I deleted old code for `fetch_price` and `fetch_exchange` since I understand they were not working, this cleans up the code and will make it easier to work on future improvements.

### DIFF
--- a/parsers/AR.py
+++ b/parsers/AR.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
 
-import itertools
-import re
-import string
+import datetime
 import logging
 import arrow
 import requests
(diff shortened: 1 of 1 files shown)


alixunderplatz:
Hi @aemartinez!
Awesome job and nice clean-up 👍
I just tested the output, looks good to me production-wise, so the fix could go live quickly :)

fetch_production() ->
{'zoneKey': 'AR', 'production': {'biomass': 137.28, 'hydro': 1755.58, 'solar': 0.0, 'wind': 1391.17, 'nuclear': 1460.5, 'unknown': 10062.8}, 'capacity': {}, 'storage': {}, 'source': 'api.cammesa.com', 'datetime': datetime.datetime(2022, 1, 31, 8, 55, tzinfo=tzutc())}

One remark from my side: Your parser is fetching the latest output for the most recent timestamp, which is fine as long as the API on the other side is responding. It is possible/easy to fetch the output for all the available matching 5-min timestamps of the raw JSONs to have the most recent "daily history" of the available data?

aemartinez:
Hi @alixunderplatz ! Thanks for the review. 

Yes, we could fetch the output for all the available matching 5-min timestamps. Should that be done by matching the parameter `target_datetime`? Should we only return the available data if the timestamp from CAMMESA exactly matches `target_datetime` or should we return the closest data?

gvillo:
For the records I sent an email to Cammesa to request some additional information to have a detailed Fossil consumption. We will see how it goes

gvillo (approved):
LGTM, I don't have write access, if @madsnedergaard could take a look on this one it would be great

madsnedergaard:
As has already been said this is awesome! 💪 

Regarding fetching more data, you can see an example of how we do this in other parsers - for example in [FR.py](https://github.com/electricitymap/electricitymap-contrib/blob/master/parsers/FR.py). If anyone remembers another parser that is more similar or easier to use for inspiration, please do share :)

aemartinez:
Great! Thanks @madsnedergaard for the clarification. After reading other parsers now I see that `fetch_production` can return a `list` of production mixes and not just one. Is that correct? (I hadn't understood that from [this doc](https://github.com/electricitymap/electricitymap-contrib/wiki/Building-a-new-parser#fetch_production).)

If that is the case (`fetch_production` can return a list) then I can work on it so we can have that code in this PR too. Thanks!

parsers/AR.py
```diff
@@ -23,957 +22,111 @@
...
-    try:
-        price_text = find_price.getText()
+    current_session = session or requests.session()
+
+    non_renewables_production: Dict[str, dict] = non_renewables_production_mix(zone_key, current_session)
+    renewables_production: Dict[str, dict] = renewables_production_mix(zone_key, current_session)
+    
+    full_production_list = [
+        {
+            'datetime': arrow.get(datetime_tz_ar).to('UTC').datetime,
```
aemartinez, line 44: Is `.to('UTC')` necessary here? I'm not sure if the `datetime` expects the time to be in a specific timezone.

parsers/AR.py
```diff
@@ -23,957 +22,111 @@
...
+
+    non_renewables_production: Dict[str, dict] = {
+        production_info['fecha']:
+        {
+            'hydro': production_info['hidraulico'],
+            'nuclear': production_info['nuclear'],
+            # As of 2022 thermal energy is mostly natural gas but 
+            # the data is not split. We put it into unknown for now.
+            # More info: see page 21 in https://microfe.cammesa.com/static-content/CammesaWeb/download-manager-files/Sintesis%20Mensual/Informe%20Mensual_2021-12.pdf
+            'unknown': production_info['termico']
```
aemartinez, line 114: Should we just put all of the thermal into 'gas' or leave it to 'unknown'? It's a pity to put that into unknown because it accounts for most of the energy production.

aemartinez:
Hi! I managed to return the list of all available timestamps. Please take a look at the last version. I also made two questions about some decisions I had to make in the code. Thanks! 😸

parsers/AR.py
madsnedergaard: It should be in UTC, so this is good :)

parsers/AR.py
madsnedergaard: For now I think it's better to keep it as unknown, even though it would be great to attribute it more appropriately. Would be great to reach out to the source and ask if there's any way they could consider splitting it up in their API :)
(But that's a next step we can do afterwards)

madsnedergaard:
> After reading other parsers now I see that fetch_production can return a list of production mixes and not just one. Is that correct? (I hadn't understood that from [this doc](https://github.com/electricitymap/electricitymap-contrib/wiki/Building-a-new-parser#fetch_production).)

Yes exactly, although I see it's not very clear from the documentation - I have tried updating the wiki to better reflect it now :)

madsnedergaard (approved):
Awesome! 💪 
Screenshot 2022-02-04 at 07 30 32

PR #3784, opened 2022-02-08
Update hydro, wind and solar capacity in Chile-SEN

Macquaria:
Source : https://www.cne.cl/normativas/electrica/consulta-publica/electricidad/ (updated 2021)


madsnedergaard (approved):
Nice! And quite a significant increase in renewables 💪

Macquaria:
Indeed ! The large increase is quite obvious when we regularly see "solar utilizing 160% or so of installed capacity" ^^

PR #3792, opened 2022-02-11
Change Kosovo/XK parser to ENTSOE

alixunderplatz:
Changes the parser for Kosovo to ENTSOE.py for all available data (= all but fetch_price).
I left the old parser XK.py untouched, just in case it is needed in the future.

grafik


FelixDQ (approved):
Awesome!

PR #3810, opened 2022-02-18
Timing delay for manifest generation race condition

dogversioning:
The watcher webpack-based manifest generation can cause the web nodemon to rapidly restart, which may result in it reading a partially written manifest and crashing, and then not restarting until another file system change is made. This adds a slight delay to nodemon to prevent this from occurring.


madsnedergaard (approved):

madsnedergaard:
Hey @dogversioning, thanks for creating this PR!
We've been stalling a bit on reviewing this, as I'm working on replacing Webpack for local development - but it would probably be fine to get this merged now anyway, in case it drags out :)

PR #3813, opened 2022-02-22
Use Paris' time zone for GB's price parser

AlexandreCouderc:
GB's price parser uses data from rte-france so the time zone should be Paris' one. It can be verified with Portugal's price, which is also available in rte-france's data.


AlexandreCouderc:
Portugal's price is correct:
image
image

GB's price is not correct:
image
image

AlexandreCouderc:
I used Portugal as an example because its time zone is also different from Paris'.

madsnedergaard (approved):
Great, and thanks for following up from #3709 :)

PR #3817, opened 2022-02-23
Add test case to verify that all zones in geometries are in zones.json

pierresegonne:
_Reviewing should be done commit per commit_

This PR adds a test case to verify that all zones with geometries are defined in `zones.json`.

Figured out that we should delete the following zones:

- `NZ-NZN` and `NZ-NZS`, deprecated since last NZ update
- `AUS-ACT` not used
- `US-NW-SCL` and `US-SW-DEAA` small US zones whose geometries were deleted in the geography improvement process
- `US-NW-AVRN` other small US zone that has never "existed" geographically

And replace `US-MIDW-EEI` with `US-MIDW-GLHB`. 

Note that for the US zones, all parsers are EIA so backfilling would be possible if we revert some decision.


Kongkille (commented):
Should we also remove US-SW-DEAA from EIA.py?

pierresegonne:
Good catch, removed it

KabelWlan:
> Should we also remove US-SW-DEAA from EIA.py?

I would not remove that zone since it's still in use. There are times where there's no generation reported (like right now). This zone consists of only one generating station, the Arlington Valley natural gas power plant. As you see on [this site]( https://www.eia.gov/electricity/data/browser/#/plant/55282) it is normal that they don't generate electricity in some months. 

Another option which is probably better:
US-SW-DEAA is a "generation-only" zone, there are a few more of those in the USA (HGMA, GRIF,...). Since they send all the power produced to US-SW-SRP we could think about including DEAA generation in SRP since they're the ones using the electricity.

pierresegonne:
> US-SW-DEAA is a "generation-only" zone, there are a few more of those in the USA (HGMA, GRIF,...). Since they send all the power produced to US-SW-SRP we could think about including DEAA generation in SRP since they're the ones using the electricity.

Interesting! 

Would you have a mapping of all of these zones by any chance? 

If yes, then let's make a subsequent PR where we add the generation only zones to the bigger grids they feed into. 

Otherwise the idea behind deleting some of these zones is that they don't currently "really" exist on the map. They are simply too small to display correctly on the map.

KabelWlan:
Here is a map of all the zones we have for the USA. The yellow dots are "generation-only" zones. The ones connecting to only one "consumption and generating zone" could be included quite easily (don't mind their coloring fo all the consumption and generating zones, this just shows if the consumption is increasing or decreasing). However, there are some which interconnect to multiple zones. Those are more difficult to split up.
EIA

To be more specific, this is a list of which zones I'd include into other ones:

- Include US-SW-DEAA into US-SW-SRP
- Include US-SW-HGMA into US-SW-SRP
- Include US-SW-GRIF into US-SW-WALC
- Include US-NW-GWA into US-NW-NWMT
- Include US-NW-WWA into US-NW-NWMT
- Include US-NW-AVRN gas generation into US-NW-PACW and wind generation into US-NW-BPAT

Then there's US-NW-GRID which makes things more complicated. Wind generation is only transmitted to US-SW-PNM but it's not clear where solar, coal and gas really go, either to US-SW-SRP or US-NW-BPAT.
It's also not clear where US-MIDW-GLHB, US-CAR-SEPA and US-CAR-YAD generating is going to. So we either delete them because they're too small or we leave them as they are right now.

I agree, that less small zones make the map easier to understand. Also it would make things more transparent since some  emissionfactors could change quite a bit.

You can doubble check all the information [here](https://www.eia.gov/electricity/gridmonitor/dashboard/electric_overview/US48/US48) and if something is unclear, reach out to me again.

pierresegonne:
Great! Thanks for the details, will include these in a following commit

madsnedergaard:
Would be good to wrap this up - let me know if we should take over on it :)

corradio:
Might be worth taking a look (and possibly closing https://github.com/electricitymap/electricitymap-contrib/pull/3359)

Kongkille (approved):
Alright, I've tried to update with latest changes, and it seems to look good so lets get it in :)

pierresegonne:
yihouuuuu

PR #3822, opened 2022-03-02
Updates mockserver to node.js and remove /v3

Kongkille:
Updates the old python to node

Also fixes a bug where the mockserver expected v3 requests and so history did not work properly.


docker-compose.yml
```diff
@@ -16,7 +10,6 @@ services:
         - eu.gcr.io/tmrow-152415/electricitymap_web:latest
     image: eu.gcr.io/tmrow-152415/electricitymap_web:latest
     command: npm run server-dev
-    depends_on: ['mockserver']
```
neilfulwiler, line 19: could i ask why we don't need this anymore?

Kongkille: We want to simplify things to speed up and eventually get away from having docker as a prerequisite, so this is taking one step further as we are able to run the mockserver without docker (and running the map without the mockserver) :)

madsnedergaard:
We need to update `.circleci/config.yml` to get build running :)

madsnedergaard (approved):
Awesome! 👏

PR #3823, opened 2022-03-02
Updated German Capacities

Bohne13:
Updated german capacities from Fraunhofer ISE which is already mention as data source. 
Data source was updated on 28.02.22


madsnedergaard (approved):

madsnedergaard:
Awesome! 🥳

PR #3825, opened 2022-03-04
Changes localisation setup

madsnedergaard:
### Description

This changes how we do localisation/translation from a semi-custom-built setup to using `i18next` and `react-i18next`. 

This simplifies things a lot and gives the following benefits:
- Changing language is now instant and doesn't require reloading the page
- Language preference is now stored in cookies and saved so users won't have to select it again all the time
- Code is simpler!
- This change makes it easier to later remove the NodeJS server entirely (I'm working on this)


madsnedergaard:
This PR is doing too much and makes it hard to review and test - I'll split it into smaller, self-contained PRs :)

PR #3834, opened 2022-03-07
allow 10% unknown values (up from 2%) for AW

neilfulwiler:
should help mitigate https://github.com/electricitymap/electricitymap-contrib/issues/3726, I'm not sure exactly why the discrepancy exists, but it seems it's been accounted for before, our threshold for failure just has increased (see https://github.com/electricitymap/electricitymap-contrib/commit/34625aae764facb648b7ac5e21035e2604e7356d)


madsnedergaard (approved):
Nice, thanks for fixing it! :)

PR #3836, opened 2022-03-08
Add direct emission values to config

gwpicard:
Fixes ELE-947
Fixes ELE-954

- adding direct emission factors per production mode and sources
- discharge and unknown values TBD (need to be calculated)


pierresegonne (changes requested):
I suggest we rename the files to remove the `_global` tag, as we currently don't know how the solution for regional emission factors will look like

pierresegonne:
Nice let's include unknown and we'll be good 👏

pierresegonne:
Also fixes ELE-954

config/co2eq_parameters_direct.json
```diff
@@ -0,0 +1,69 @@
...
+        "source": "EIA 2020/BEIS 2021",
+        "value": 406
+      },
+      "solar": {
+        "source": "IPCC 2014",
+        "value": 0
+      },
+      "unknown": {
+        "_comment": "assume conventional",
+        "source": "assumes thermal (coal, gas, oil or biomass)",
```
gwpicard, line 59: @pierresegonne do you agree with the phrasing here (and in the _comment)?

pierresegonne: Let's just remove biomass

pierresegonne (approved):
Nice one

PR #3838, opened 2022-03-08
Moves client-version check from server to static files

madsnedergaard:
_Part of #3837_

Changes the `clientVersion` functionality from being served by the server to a static file.


web/webpack.config.js
```diff
@@ -48,9 +48,11 @@ module.exports = {
     new webpack.optimize.OccurrenceOrderPlugin(),
     function () {
       this.plugin('done', (stats) => {
+        // Avoid dumping everything (~30mb)
+        const output = {...stats.toJson({all: false, assets: true, groupAssetsByChunk: true})};
         fs.writeFileSync(
           `${__dirname}/public/dist/manifest.json`,
-          JSON.stringify(stats.toJson())
+          JSON.stringify(output)
```
madsnedergaard, line 55: Unrelated change, but it's a bit insane that we're using a 30mb file when we only need a tiny part of the content.

Issue #3837, opened 2022-03-08
Remove the NodeJS server used in frontend


madsnedergaard:
### Description

The server complicates the build and local development, while impacting performance and SEO.
It mainly exists because of localisation, which should/could be improved and replaced on client-side anyway :)

See further discussions here: https://github.com/electricitymap/electricitymap-contrib/issues/3262

I've started the work here, but realised there's several areas needed to be addressed here. See list below.

#### Tasks in this

- Replace `clientVersion` usage
- Replace localisation approach (of inlining all translations in `<head />`)
- Replace translated meta-tags generation
- Create static index.html during build step
- Deploy everything to a bucket with another subdomain, transfer when it all works!

VIKTORVAV99:
Not sure if this is helpful or not but the [HtmlWebpackPlugin](https://webpack.js.org/plugins/html-webpack-plugin/#root) can (and should) be used for generating the index.html, it also supports ejs templates so we can mostly use the current template without huge changes, it's just the facebook specific stuff that will need to have a solution. 

We should also change how we handle the page title as it currently using the index.ejs to be generated, we should switch this to a static `electricityMap` title and then update it using client side JS. (this can be done already if we want)

madsnedergaard:
Let's try replacing this issue with a Project: https://github.com/orgs/electricitymap/projects/29/
:)

PR #3839, opened 2022-03-08
Simplifies functionality for getting a zone/country name

madsnedergaard:
_Part of #3837, and splitting up my huge PR in #3825_

### Description

Merges the two functions and makes them simpler to use. Also replaces all usage of the old functions.


web/src/components/tooltips/common.js
```diff
@@ -27,6 +27,6 @@ const Flag = styled.img``;
 export const ZoneName = ({ zone, ellipsify }) => (
   <React.Fragment>
     <Flag className="flag" alt="" src={flagUri(zone)} />
-    {ellipsify ? getShortZoneName(zone) : getFullZoneName(zone)}
+    {getZoneNameWithCountry(zone, ellipsify ? 40 : 0)}
```
Kongkille, line 30: Nitpick: For readability, we might want a number reference for the limit instead of passing '40'. Also we can avoid passing a limit and instead passing undefined, which I also think reads slightly better as passing 0 is the same is saying there's no limit.

So something like this: 

const LIMIT = 40;

getZoneNameWithCountry(zone, ellipsify ? LIMIT : undefined)

web/src/helpers/translation.js
```diff
@@ -27,29 +27,28 @@ export function translate() {
...
-export function getShortZoneName(zoneCode, limit = 40) {
-  const zoneName = translate(`zoneShortName.${zoneCode}.zoneName`);
+export const getZoneName = (zoneCode) => translate(`zoneShortName.${zoneCode}.zoneName`);
+export const getCountryName = (zoneCode) => translate(`zoneShortName.${zoneCode}.countryName`);
+
+/**
+ * Gets the full name of a zone with the country name in parentheses.
+ * Zone name can be ellipsified with the limit parameter.
+ * @param {string} zoneCode 
+ * @param {number} limit
```
Kongkille, line 38: @param {number} [limit=0] if we want to show that there is a default value and that it's optional. However, the description also indicates that it's optional

Kongkille (approved):
Nice, just some small comments, which you can resolve if you want to.

I personally find the function a bit confusing as it can return a lot of different things depending on what's available, eg zoneCode, zoneName, full name, and with a limit or not. However it's prob not worth digging much more into it atm

Kongkille (approved):

PR #3840, opened 2022-03-09
Moves translation files to public folder

madsnedergaard:

### Description

This PR simply moves `locales` from root to `public/locales` and updates all references.
We're doing this to make it easier to switch to https://github.com/i18next/i18next-http-backend and serve the translation files dynamically, instead of doing it on the server level.


Kongkille (commented):
After we merge, we should also quickly update references on the wiki page https://github.com/electricityMap/electricitymap-contrib/wiki/Translating-electricitymap.org

I see we also move the translation-helper script. Any thoughts on having it in the public folder alongside the json files?

madsnedergaard:
> After we merge, we should also quickly update references on the wiki page [electricityMap/electricitymap-contrib/wiki/Translating-electricitymap.org](https://github.com/electricityMap/electricitymap-contrib/wiki/Translating-electricitymap.org)

Good point, I'll make sure to do that!

> I see we also move the translation-helper script. Any thoughts on having it in the public folder alongside the json files?
I just moved the whole thing, but when thinking about it I think it makes to move it to the `web` root along with `translation-status.js` :)

Kongkille (approved):

PR #3842, opened 2022-03-09
Adds new translation setup

madsnedergaard:

### Description

Adds [i18next](http://i18next.com/) and sets up initialisation with relevant plugins.
Also moves `moment` locale change and the `hideLanguageSearchParam` function to based on language change so it in future is done dynamically.

**Note:** I have intentionally not replaced the old translation approach in this PR, so we're running both in parallel for a short while for now. This should make testing and reviewing much easier :)


web/src/helpers/i18n.js
```diff
@@ -0,0 +1,83 @@
...
+    interpolation: {
+      escapeValue: false, // not needed for react as it escapes by default
+    },
+  })
+  .then(() => {
+    hideLanguageSearchParam();
+  });
+
+i18n.on('languageChanged', function (lng) {
+  moment.locale(window.locale.toLowerCase());
```
madsnedergaard, line 79:
```suggestion
  moment.locale(lng);
```

Kongkille (approved):

PR #3848, opened 2022-03-10
Removes translation from server

madsnedergaard:

### Description

Removes all the old translation-related pieces from server and generate-index script.
This should be the last step in splitting the old PR! 🥳 

⚠️ Depends on #3840, #3841 & #3842 - not ready for review before those are merged! ⚠️


Kongkille (approved):

web/server.js
```diff
@@ -174,6 +131,7 @@ app.use('/', (req, res) => {
       res.cookie('electricitymap-token', process.env.ELECTRICITYMAP_TOKEN);
     }
     res.render('pages/index', {
+      maintitle: localeConfigs[locale || 'en'].misc.maintitle,
```
Kongkille, line 134: Do we want to store 'en' as a DEFAULT_LANG const?

Issue #2407, opened 2020-04-24
Should Sweden's unknown co2eq be recalculated with no coal power?


VIKTORVAV99:
Sweden recently closed it's last coal power plant. But when I was going to edit it I noticed that Sweden did not have a entry for coal power so I added it.
So now I wonder if I should just change the coal co2eq entry with 0 too?
Does unknown co2eq need to be recalculated?

Source: https://ieefa.org/sweden-becomes-third-european-country-to-complete-coal-plant-phaseout/ 

EDIT: To be clear I mean the co2eq equivalent of coal and unknown.

robertahunt:
It is a good question - We currently assume 72% biomass, 28% conventional thermal for the unknown makeup. As far as I understand Sweden has quite a few combined heat and power plants (which I think are mainly gas/waste powered) - you can see more info here https://pxexternal.energimyndigheten.se/pxweb/sv/%c3%85rlig%20energibalans/%c3%85rlig%20energibalans__El-%20och%20fj%c3%a4rrv%c3%a4rmeproduktion/EN0202_25.px/table/tableViewLayout2/?loadedQueryId=17d13f92-2720-46c9-9727-463aa7d01cfa&timeType=from&timeValue=0.

And I don't think the coal plant they shut down has been in operation much in the past couple years, so I doubt it's capacity change would effect much. But if you/anyone else has time to do some more digging and find a better approximation for the unknown co2 factor, that would be great!

VIKTORVAV99:
I found pretty detailed statistics for the power production of 2018 on Eurostat.
Here is a link to the zip containing the excel tables: [Energy Balances in the MS Excel file format (2020 edition)](https://ec.europa.eu/eurostat/documents/38154/4956218/Energy-Balances-April-2020-edition.zip/69da6e9f-bf8f-cd8e-f4ad-50b52f8ce616)

If I understand this data correct the ratio for 2018 should be:

- Coal: 2.15%
- Manufactured gases: 4.93%
- Peat: 1.95%
- Oil and Petroleum: 1.97%
- Natural gas: 2.41%
- Biofuel: 65.48%
- Biogas: 0.07% (This one can probably be skipped)
- Liquid biofuel: 0.34% (Possibly this one too)
- Renewable Municipal Waste: 9.66%
-  Non-renewable Waste: 10.08%

[...]

PR #3849, opened 2022-03-10
Updated CO2eq numbers with the latest data (2020).

VIKTORVAV99:
Updated Sweden's CO2eq for unknown production with the latest data from Eurostat (2020 data, 2022 edition) using the same methodology as in #2407.

_PS:_ Looks like there slipped in a formatting change that was not intended but hopefully you can still merge it.


madsnedergaard (approved):
Looks good :raised_hands:

PR #3854, opened 2022-03-11
Replaces all outdated CJS imports/exports

madsnedergaard:
### Description

Removes usage of CommonJS and replaces it with ES modules syntax.

Getting rid of these will aligns things and is a requirement for switching to a modern module-based build tool like https://www.snowpack.dev/ 🏔️

### DIFF
--- a/web/src/components/countryhistorymixgraph.js
+++ b/web/src/components/countryhistorymixgraph.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { max as d3Max } from 'd3-array';
 import { forEach } from 'lodash';
 
-import formatting from '../helpers/formatting';
+import { scalePower } from '../helpers/formatting';
 import { useCo2ColorScale } from '../hooks/theme';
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -49,6 +49,7 @@ jobs:
       - run:
           command: |
             sudo npm install -g jest
+            sed -i -e 's/export {/module.exports = {/' web/src/helpers/constants.js
```
madsnedergaard, line 52: I couldn't get Jest to work without doing a lot of extra work (like creating a package for this one test file), so instead we hack it by turning this file back into CommonJS. It's not ideal, but I think it's an acceptable workaround for now :)

madsnedergaard, line 52: I couldn't get Jest to work without doing a lot of extra work (like creating a package for this one test file), so instead we hack it by turning this file back into CommonJS. It's not ideal, but I think it's an acceptable workaround for now :)

PR #3862, opened 2022-03-14
Adds Snowpack for speedy local development ⚡ 

madsnedergaard:
### Description

Local development is currently really painful and slow. Instead of trying to improve Webpack setup and continue through Docker, I have decided to use a modern ESM-based build tool (https://www.snowpack.dev/) to make it a way more pleasant experience!

- It is now possible to run `yarn develop` for a speedy and amazing experience
- It is still possible to run the old setup with `docker-compose` (for backwards compatibility and to test things closer to how it's currently running in production)

See comparison here (top window is old, bottom is new):

https://user-images.githubusercontent.com/3296643/158170093-fbfc3ebe-7946-4b74-9207-615d1e7b164a.mp4


web/src/scss/styles.scss
```diff
@@ -1,18 +1,18 @@
...
+@import "./content/panel/time-slider";
+@import "./content/panel/zone-list";
 
-@import "components/faqs";
-@import "components/messaging";
-@import "components/modal";
-@import "components/meta";
\ No newline at end of file
+
+@import "./components/messaging";
```
madsnedergaard, line 16: Whoops, accidentally removed this!
```suggestion
@import "./components/faqs";
@import "./components/messaging";
```

web/src/components/zonemap.js
```diff
@@ -126,7 +126,7 @@ const ZoneMap = ({
 
   const handleClick = useMemo(
     () => (e) => {
-      if (ref.current && !ref.current.state.isDragging) {
+      if (ref.current && ref.current.state && !ref.current.state.isDragging) {
```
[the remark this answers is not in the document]
Kongkille: Nvm just saw the commit msg

Kongkille (approved):
Amazing! Just tried it - purring like a kitten haha

I could not get it up and running immediately myself, I had to also run "yarn global add snowpack" before it actually worked. 

We should also add a section on this on 
https://github.com/electricityMap/electricitymap-contrib/wiki/Set-up-local-environment

web/src/helpers/constants.js
```diff
@@ -35,11 +35,72 @@ const DEFAULT_FLAG_SIZE = 16;
...
+  "pt-br": "Português (Brasileiro)",
+  "ro": "Română",
+  "ru": "Русский язык",
+  "sk": "Slovenčina",
+  "sv": "Svenska",
+  "vi": "Tiếng Việt",
+  "zh-cn": "中文 (Mainland China)",
+  "zh-hk": "中文 (Hong Kong)",
+  "zh-tw": "中文 (Taiwan)"
+}
```
VIKTORVAV99, line 66: Shouldn't these language codes match the casing used in #3948?
Same with the Facebook codes.

Kongkille: Good point, I am actually a bit uncertain about the necessity of these files. It seems like we have a lot of dependencies on different language constants, eg this one, the FB ones, and the one in web/locales-config. Perhaps this is necessary to also keep docker-compose

cc @madsnedergaard

madsnedergaard:
> Amazing! Just tried it - purring like a kitten haha
> 
> I could not get it up and running immediately myself, I had to also run "yarn global add snowpack" before it actually worked.
> 
> We should also add a section on this on [electricityMap/electricitymap-contrib/wiki/Set-up-local-environment](https://github.com/electricityMap/electricitymap-contrib/wiki/Set-up-local-environment)

Hmm, that should not be required for using `yarn develop` 🤔 
Did you run it directly? (like `snowpack dev`)

web/src/helpers/constants.js
madsnedergaard: Thanks and good catch! Updated now, I'll address the need and duplication of these constants in a separate PR later on instead of making this one more complicated :)

Kongkille:
> > Amazing! Just tried it - purring like a kitten haha
> > I could not get it up and running immediately myself, I had to also run "yarn global add snowpack" before it actually worked.
> > We should also add a section on this on [electricityMap/electricitymap-contrib/wiki/Set-up-local-environment](https://github.com/electricityMap/electricitymap-contrib/wiki/Set-up-local-environment)
> 
> Hmm, that should not be required for using `yarn develop` 🤔 Did you run it directly? (like `snowpack dev`)

No I just ran yarn and then yarn develop and it didn't seem to work. Could be something else, but it may be worth trying to uninstall snowpack and then rerunning yarn

madsnedergaard:
Weird, I don't have `snowpack` installed globally and it works fine 🤔 
Can you try removing `node_modules` and reinstalling next time you're working on something locally?

VIKTORVAV99:
I just tested it myself and it worked (although it threw some errors but I don't think there is a real problem) but I had to change the version of node-sass before cause it wouldn't install without python2... (I'll open a PR for a bump).

I really like how speedy it is, now that I don't have to mess with docker working with the frontend should be a lot easier! 😃

PR #3865, opened 2022-03-15
Change to Plausible

Kongkille:
Changes analytics to Plausible, which is a lightweight GA alternative.

Note that Plausible strips all url query parameters, meaning wind/solar layers won't show up.

Furthermore, it seemed GA also tracked some kind of execution/load time, which I have removed, as I don't think it was used, and I am not sure Plausible supports the same thing.


web/views/pages/index.ejs
```diff
@@ -99,8 +99,8 @@
         <script defer src="<%= resolvePath('dist/vendor.' + vendorHash + '.js') %>"></script>
         <script defer src="<%= resolvePath('dist/bundle.' + bundleHash + '.js') %>"></script>
 
-        <!-- Google Analytics -->
-        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-79729918-1"></script>
+        <!-- Plausible -->
+        <script defer data-domain="app.electricitymap.org" src="https://plausible.io/js/plausible.local.js"></script>
```
madsnedergaard, line 103: Probably shouldn't still be local? :)

madsnedergaard:
I made some changes to the thirdparty code in https://github.com/electricitymap/electricitymap-contrib/pull/3854 which are now causing a merge conflict :)

web/views/pages/index.ejs
madsnedergaard: Added this comment about testing on mobile apps @Kongkille

Kongkille: Ah great point. We should test if the app can send on localhost

Kongkille:
Just tested on the mobileapp, and it also seems to work. We just have to keep the localhost setting as well. However, it's disabled unless it's in production, so it should not fire when developing locally on the web app.

madsnedergaard (approved):
Awesome! 🥳

madsnedergaard:
FYI There's some cleanup on mobileapps of removing analytics, but we should address that in a separate PR :)

PR #3871, opened 2022-03-17
Update co2eq parameters with yearly zone overrides

pierresegonne:


pierresegonne:
@FelixDQ please verify that the format for the fallback mixes override is indeed the expected one.

FelixDQ:
Can you try to test using this function?https://github.com/electricitymap/electricitymap/blob/master/libs/flowtracing/electricitymap/flowtracing/utils.py#L46

We should probably also update the test in contrib to check if the new format is correct.

pierresegonne:
`poetry run test` within `libs/forecasting` seems to work. Looking into why the tests here fail

FelixDQ:
Okay I just tried it out. The format should be like this:
```json
      "AR": {
        "powerOriginRatios": [
          {
            "_source": "electricityMap, 2019 average",
            "datetime": "2019-01-01",
            "battery charge": 0,
            "battery discharge": 0,
            "biomass": 0.017509,
            "coal": 0.38,
            "gas": 0.232,
            "geothermal": 0.003081,
            "hydro": 0.158,
            "hydro charge": 0,
            "hydro discharge": 0,
            "nuclear": 0.102,
            "oil": 0.03,
            "solar": 0.021966,
            "unknown": 0.007728,
            "wind": 0.047716
          },
          {
            "_source": "electricityMap, 2020 average",
            "datetime": "2020-01-01",
            "battery discharge": 0.0,
            "biomass": 0.0010942423798809989,
            "coal": 0.0076602812925308066,
            "gas": 0.5671125375099301,
            "geothermal": 1.213687697353592e-10,
            "hydro": 0.22328402680254944,
            "hydro discharge": 0.0053804895936092004,
            "nuclear": 0.08030772836986415,
            "oil": 0.013403454510627244,
            "solar": 0.0064906458465778995,
            "unknown": 0.03438412334314057,
            "wind": 0.06088053087834923
          }
        ]
      }
```

gwpicard:
@FelixDQ just tried regenerating using this format but it's still failing... :/ I think the format might still be wrong.

config/co2eq_parameters_all.json
```diff
@@ -1385,344 +1778,424 @@
...
+            "hydro discharge": 0.0,
+            "nuclear": 0.0,
+            "oil": 0.0007836425136503577,
+            "solar": 0.0,
+            "unknown": 0.04927664985922736,
+            "wind": 0.06878744399718065
+          }
+        ]
       },
       "NZ-NZS": {
```
madsnedergaard, line 1920: How come the script is only creating "NZ-NZS" and not a "NZ", when we have CSVs for both? 🤔 
And can I safely remove the NZS and NZN files here? (Sorry offtopic discussion, figured it made sense when you're already working on the script 😄)
Screenshot 2022-03-18 at 07 22 09

pierresegonne: There is a requirement for having a certain number of timestamps with valid data to generate a yearly aggregate. + the last year we are looking at is 2021. I think this explains why NZ does not get its yearly aggregate. 
For NZ-NZN and NZ-NZS, yes we can move the csvs away from there, I'll do it :)

madsnedergaard: Ah okay, makes sense - thanks! 🙌

pierresegonne:
Won't have time to look at it today :(

gwpicard:
I'll take a look

madsnedergaard:
Please remember to update `emission_factors()` to make sure it still works after this, and please also write a test for it :)
See https://github.com/electricitymap/electricitymap-contrib/pull/3895

PR #3873, opened 2022-03-18
Creates shared SocialButtons component

madsnedergaard:
### Description

Cleans up code by having just one social-buttons component, instead of duplicating code in many places.
Stumbled upon this while cleaning up some other code in a future PR :)


Kongkille (approved):
Nice :)

PR #3874, opened 2022-03-18
Adds loading indicator while fetching history

madsnedergaard:
## Description

Makes it clear to users that data is being fetched, instead of first showing "no data is available" until data for a zone is loaded.

### Before
Screenshot 2022-03-18 at 11 47 25

### After
Screenshot 2022-03-18 at 11 46 54


Kongkille (approved):
This is great.

I am wondering if we'd want to still show the name of the zone while loading? I think it's nice small feedback that the users know they have clicked the right zone and is now awaiting the detailed data. 

(I realize it's less than a second usually) 

Adds a bit of extra work, so regardless we could address that in another PR

madsnedergaard:
Good point @Kongkille, I have changed it to this now:

Screenshot 2022-03-25 at 12 51 39

I also fixed the back-arrow button being white on hover so you couldn't see it 🤔

Kongkille (approved):
Excellent!

Issue #3300, opened 2021-08-11
frontend: optimize contributors avatars size


Manu1400:
In webpage of a zone [zone](https://www.electricitymap.org/zone/FR) in ElectricityMap, my browser download contributors avatar with two HTTP requests by contributor:

ie: https://github.com/lorrieq.png => Status Code: 302 => https://avatars.githubusercontent.com/u/6525086?v=4 => avatar in large size

### proposal
get avatar from avatars.githubusercontent.com (to avoid redirection) and then use `?s=20` param (for size 20x20)

=> https://avatars.githubusercontent.com/u/28250254?s=60&v=4

madsnedergaard:
That would require knowing the user id of contributors instead of their handle, right? I think that makes it too complicated, but the size paramter is a good point - feel free to update the url here: https://github.com/tmrowco/electricitymap-contrib/blob/master/web/src/components/contributorlist.js

Maybe use `s=40` to ensure we have a retina-friendly size :)

Manu1400:
I assign myself this dev task

URL: https://avatars.githubusercontent.com/lorrieq?s=40 ✔️

gwpicard:
@Manu1400 How is this going? 😄

PR #3875, opened 2022-03-18
improve: Optimize contributors avatar size

dougfabris:
It replaces the contributor's avatar `URL` to get the avatar image in a smaller size when it's possible and avoid the redirection, as mentioned in the issue attached.

Closes: #3300

### DIFF
--- a/web/src/components/contributorlist.js
+++ b/web/src/components/contributorlist.js
@@ -3,17 +3,26 @@ import { useParams } from 'react-router-dom';
 
 import zonesConfig from '../../../config/zones.json';
 
+const getContributorId = (contributor) => {
+  const contributorSplitted = contributor.split('/');
+  return contributorSplitted[contributorSplitted.length-1];
+}
+
 const ContributorList = () => {
   const { zoneId } = useParams();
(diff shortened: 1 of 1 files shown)


web/src/components/contributorlist.js
```diff
@@ -3,17 +3,26 @@ import { useParams } from 'react-router-dom';
 
 import zonesConfig from '../../../config/zones.json';
 
+const getContributorId = (contributor) => {
+  const contributorSplited = contributor.split('/');
+  return contributorSplited[contributorSplited.length-1];
```
madsnedergaard, line 8: Minor detail 😉 
```suggestion
  const contributorSplitted = contributor.split('/');
  return contributorSplitted[contributorSplitted.length-1];
```

madsnedergaard (commented):
This is great, thanks for fixing it 💪 
One very minor grammar comment

web/src/components/contributorlist.js
dougfabris: fixed :)

madsnedergaard (approved):

PR #3898, opened 2022-03-22
fix(parser): fix the NG parser

kruschk:
A server-side API change broke the NG parser. Before, data could be
requested with a simple GET request, but now it is necessary to send a
GET request, scrape some magic values from the response, then send a
POST request with those magic values.

- Fix issue #3651.
- Overhaul the parser to accomodate the new API.
- Update data sources and zones accordingly.

Refs: #3651

### DIFF
--- a/parsers/NG.py
+++ b/parsers/NG.py
@@ -1,76 +1,88 @@
 #!/usr/bin/env python3
 
-"""Parser for the electricity grid of Nigeria"""
+"""Parser for the Nigerian electricity grid"""
 
-import arrow
+# Standard library imports
+import datetime
 import logging
+import re
+import urllib.parse
+
+# Third-party library imports
+import arrow
+import bs4
 import requests
-from bs4 import BeautifulSoup
+
+# Local library imports
+from parsers.lib import config, validation
+
+API_URL = urllib.parse.urlparse("https://niggrid.org/GenerationProfile")
+API_URL_STRING = API_URL.geturl()
+NORMALISE = {
+    "gas": "gas",
+    "gas/steam": "gas",
+    "hydro": "hydro",
+    # Per the "Electricity generation by source" plot at
+    # https://www.iea.org/countries/nigeria, the majority of Nigeria's
+    # generation comes from natural gas, so "steam" most likely implies "gas";
+    # therefore, we do not map "steam" to "unknown" (coal/gas/oil). See issues
+    # #2570 and #3651 for more information.
+    "steam": "gas",
+}
+PATTERN = re.compile(r"\((.*)\)")
 
 
-LIVE_PRODUCTION_API_URL = "https://www.niggrid.org/GenerationLoadProfileBinary?readingDate={0}&readingTime={1}"
-TYPE_MAPPING = {"hydro": "hydro", "gas": "gas", "gas/steam": "gas", "steam": "gas"}
-
-
-def extract_name_tech(company):
-    parts = company.split("(")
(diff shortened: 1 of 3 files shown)


parsers/NG.py
```diff
@@ -1,76 +1,89 @@
...
+    for row in rows:
+        _, source, power, _ = (tag.text for tag in row.find_all("td"))
+        try:
+            technology = NORMALISE[PATTERN.search(source).group(1).casefold()]
+        except (AttributeError, KeyError) as error:
+            logger.warning(f"Unexpected source '{source.strip()}' encountered")
+            continue
+        production_mix[technology] += float(power)
+    # If we have no production from any source, assume something went wrong.
+    if all(value == 0.0 for value in production_mix.values()):
```
jarek, line 74: Could also word this as `if sum(production_mix.values()) == 0.0:` which is a tiny bit more computationally expensive, but might read more accurately (in a "code should say what you mean" sense).

As an aside, `0 == 0.0` in Python, so `value == 0` and `value == 0.0` are equivalent. `0` is type `int` and `0.0` is type `float`, but they compare the same.

If you would like to test that the value is not too low without being 0, you can also use the centralized validation function: `parsers.lib.validation.validate(datapoint, floor=10)` for minimum 10 MW. See https://github.com/electricitymap/electricitymap-contrib/blob/master/parsers/lib/validation.py

jarek (approved):
This looks really good, thank you!

parsers/NG.py
```diff
@@ -1,76 +1,89 @@
...
+# tell, the table is always populated within 15 min of the turn of the hour).
+@parsers.lib.config.refetch_frequency(datetime.timedelta(minutes=45))
+def fetch_production(zone_key="NG",
+                     session=None,
+                     target_datetime=None,
+                     logger=logging.getLogger(__name__)) -> dict:
+    """Requests the last known production mix (in MW) of a given zone."""
+    timestamp = (arrow.get(target_datetime).to("Africa/Lagos")
+                 if target_datetime is not None
+                 else arrow.now(tz="Africa/Lagos"))
```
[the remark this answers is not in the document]
kruschk: Ah, I didn't realise! That would allow the next line to be chained onto this assignment, which would be nice and tidy. I'm all for it!

parsers/NG.py
kruschk: To be honest, I'm not sure which one better expresses the intent. I'm somewhat partial to mine, although I may be biased! I should probably use `parsers.lib.validation.validate` instead, so I suppose it doesn't make much difference either way.

madsnedergaard (commented):
This is awesome! 💪 
I don't have any specific comments, but I think @jarek raised some good points, so would be good to address those :)

parsers/NG.py
jarek: I'm fine with any of these - probably leaning towards using the `validation` lib for consistency with other parsers :+1:

kruschk:
Thanks, @jarek and @madsnedergaard, for reviewing! I implemented those two suggestions and squash-rebased the two commits onto `master`. In hindsight, it might have been better to just push a new commit and let you decide whether to merge or rebase; please let me know what you prefer and I'll keep it in mind for next time! :slightly_smiling_face:

jarek (approved):
Looks good!

My personal preference on Github PRs is to make a second commit since that makes it easier to see the changes from when I reviewed last. The commits can then be squashed when merging in Github. But not a big deal by any means.

PR #3899, opened 2022-03-22
fix: Maranhao should be part of to North-East zone

dougfabris:
the state of Maranhão is part of the North subsystem instead of North-East: [source](https://pt.wikipedia.org/wiki/Sistema_Interligado_Nacional) 

### before
image

### after
image


VIKTORVAV99:
Don't forget to also move the exchange arrow between North Brazil and North-East Brazil so it match the new border location.

madsnedergaard:
Good point @VIKTORVAV99!
I'll try running it and adjusting arrows today :)

madsnedergaard (approved):
Thanks for fixing this, I have moved the exchange arrow so it's on the border :)

Screenshot 2022-03-31 at 17 08 15

PR #3900, opened 2022-03-22
Fixes OPENNEM (Australia) exchanges

madsnedergaard:
Fixes a bug with exchanges that was introduced in https://github.com/electricitymap/electricitymap-contrib/pull/3550.

Closes https://github.com/electricitymap/electricitymap-contrib/issues/3689

### DIFF
--- a/parsers/OPENNEM.py
+++ b/parsers/OPENNEM.py
@@ -1,6 +1,6 @@
 import datetime
 import logging
-from typing import Dict, List, Mapping, Tuple, Union
+from typing import Dict, List, Mapping, Tuple
 
 import arrow
 import pandas as pd
@@ -191,8 +191,8 @@ def fetch_main_power_df(
     session=None,
     target_datetime=None,
     logger=logging.getLogger(__name__),
-) -> Union[Tuple[pd.DataFrame, pd.Series], pd.DataFrame]:
-    df, region, filtered_datasets = _fetch_main_df(
+) -> Tuple[pd.DataFrame, list]:
+    df, filtered_datasets = _fetch_main_df(
         "power",
         zone_key=zone_key,
         sorted_zone_keys=sorted_zone_keys,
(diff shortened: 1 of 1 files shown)


parsers/OPENNEM.py
```diff
@@ -203,8 +203,11 @@ def fetch_main_power_df(
     # Solar rooftop is a special case
     df = process_solar_rooftop(df)
     logger.debug("Preparing capacities..")
-    capacities = get_capacities(filtered_datasets, region)
-    return df, capacities
+    if region:
+        capacities = get_capacities(filtered_datasets, region)
+        return df, capacities
+    else:
+        return df
```
jarek, line 210: At first glance this looks confusing - how would the function's callers know if it will return 1 or 2 things? Having looked at it I see that `fetch_exchange` calls `df = fetch_main_power_df(...)` while `fetch_production` calls `df, capacities = fetch_main_power_df(...)`.

I would suggest to have this function always return a tuple:

```suggestion
    if region:
        capacities = get_capacities(filtered_datasets, region)
        return df, capacities
    else:
        return df, None
```

function return value type annotation: `Tuple[pd.DataFrame, Optional[pd.Series]]`

change the call in `fetch_exchange` to throw away the second value: `df, _ = fetch_main_power_df(...)`

jarek (commented):
Oh that turned out to be simpler than I thought! I do have a suggestion below

parsers/OPENNEM.py
madsnedergaard: Thanks for the feedback, and I definitely agree that it's a bit confusing right now :D 
I tried improving it a bit without going too deep, what do you think?

madsnedergaard:
> Thanks @jarek! True that I did not clean up that part of the parser. Looks more intelligible now, but if we start looking at enforcing better typing here we might want to also improve the signature of `_fetch_main_df` no?

I've added a return type for `_fetch_main_df`, but I'd prefer not to spend too much time cleaning up this code, but rather get exchanges working as they have been down for a long time :)

pierresegonne (approved):

PR #3901, opened 2022-03-23
Removes NZ-NZN and NZ-NZS

madsnedergaard:
Removing these two outdated zones everywhere!

See https://github.com/electricitymap/electricitymap-contrib/pull/3567 for the initial work.

Maybe this will help fix the problems in #3894, but if nothing else it will at least simplify and clean up things :)


pierresegonne:
FYI there is also some context here https://github.com/electricitymap/electricitymap-contrib/pull/3817
otherwise LGTM

pierresegonne:
Makes sense

I restarted the test as it was not clear why it failed

madsnedergaard:
I found the issue and fixed it (commas in JSON). 
Ready for review @pierresegonne!

pierresegonne (approved):

PR #3902, opened 2022-03-23
Adds a small welcome message with clientVersion details to console

madsnedergaard:
I wanted to log the clientVersion, as I saw the "new client available" and couldn't figure out why. While at it, I also added a small "welcome" message for curious users :)

Screenshot 2022-03-23 at 17 15 59


corradio:
Love it!

Kongkille (approved):

PR #3904, opened 2022-03-24
Update Slovenia Solar to 504,4MW(31.12.2021)

DavidKarlas:
Source is http://pv.fe.uni-lj.si/SEvSLO.aspx which is owned by Faculty of Electrical Engineering and keeps track of solar installations and updates numbers every year.

Existing number is from 2018.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -5651,7 +5651,7 @@
       "hydro storage": 180,
       "nuclear": 696,
       "oil": 0,
-      "solar": 289,
+      "solar": 504.4,
       "wind": 3
     },
     "contributors": [


jarek:
Thank you for the update! Could you also add the source to https://github.com/electricitymap/electricitymap-contrib/blob/master/DATA_SOURCES.md in "Production capacity data sources" section?

DavidKarlas:
After inspecting data in more details it seems like there are too many installations with 16kW and 17kW, I'm suspecting they list inverter listen power instead of panels power, will open new PR when this is cleared up, I will also include source.

PR #3908, opened 2022-03-24
Update Belgian solar capacity

Fred-3000:
Elia updated its monitored solar capacity yesterday, resulting in an increase of more than 1 GW 🥳
Same sources.


config/zones.json
```diff
@@ -650,7 +650,7 @@
       "hydro storage": 1308,
       "nuclear": 5943,
       "oil": 280,
-      "solar": 4788,
+      "solar": 5.898,
```
jarek, line 653: In JSON this means a little under 6 megawatts, not 6 gigawatts :sweat_smile: 

```suggestion
      "solar": 5898,
```

jarek (changes requested):

config/zones.json
Fred-3000: Oops, that's a very stupid mistake I made there! Thanks for pointing it out.

madsnedergaard (approved):

jarek (approved):

PR #3913, opened 2022-03-27
update capacities USA January 2022

KabelWlan:
I updated all capacities in the USA zones to January 2022. The source is still the same.

Lots of new solar and wind capacity and EIA finally updated some of the power plants to the correct zones. The coal capacity of US-NW-PACW was missing so far but that's fixed now.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -5972,7 +5972,7 @@
       "hydro storage": 0,
       "nuclear": 0,
       "oil": 72.2,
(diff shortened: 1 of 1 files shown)


brandongalbraith:
@KabelWlan Super excited to see this capacity update, thanks for putting the time in! <3

pierresegonne (approved):
Great! Thanks for updating this :)

PR #3915, opened 2022-03-28
Historical unknown co2eq values for Sweden

VIKTORVAV99:
Added historical co2eq unknown values for Sweden back to 2015.
Since I had previously removed coal from the latest value because the last coal plant closed it resulted in a marginal increase in the co2eq value from 280.2 to 280.3 as well.


VIKTORVAV99:
You guys might also want to re-run the generation you did for #3871 after merging this for more accurate values.

pierresegonne:
Super! 
I think that if it's possible, we'll go with the more accurate version, and we can come back to this issue when we split up Sweden to use ENTSO-E data consistently.

pierresegonne (approved):

VIKTORVAV99:
@pierresegonne sounds like a plan!
Also when switching to the ENTSOE parser and split bidding zones we will get gas and solar reported separately so they can be directly removed from these calculations and then it should just be biomass, peat and oil left to calculate.

PR #3916, opened 2022-03-28
Chile: update and add capacities

KabelWlan:
I updated the capacities for CL-SEN because solar was already exceeding 100% again. It's crazy how many solar and wind farms they've built in the last year.
I also added capacities for CL-SEA and CL-SEM. They're not on the map yet but it didn't take a lot of time to just add the capacities as well. 
The sources are now updated as well.


madsnedergaard (approved):
Amazing update, Chile is on fire 🔥🌞

Issue #3919, opened 2022-03-29
New data sweden


Henrik575:
Hi,

The following link leads to new data for sweden

https://pxexternal.energimyndigheten.se/pxweb/sv/N%c3%a4tanslutna%20solcellsanl%c3%a4ggningar/N%c3%a4tanslutna%20solcellsanl%c3%a4ggningar/EN0123_1.px/table/tableViewLayout2/?loadedQueryId=1f5d3f27-504d-4668-b762-9b7c9e4d6c7c&timeType=from&timeValue=0

VIKTORVAV99:
Oh I was waiting for this but expected it to be published tomorrow.

I'll get a PR ready asap.

Henrik575:
Thanks!

ons 30 mars 2022 kl. 09:54 skrev Viktor Andersson ***@***.***

PR #3920, opened 2022-03-30
SE solar and wind capacity update.

VIKTORVAV99:
Updated the capacities for solar and wind with the same sources as before.

Closes: #3919


madsnedergaard (approved):
Nice change! 🥳

VIKTORVAV99:
@madsnedergaard I rebased the PR as it was out of date but you should be able to merge it now.

PR #3922, opened 2022-03-30
Sv language updates

VIKTORVAV99:
Updated some grammar related to #3910 and added missing translations.

I also changed one translation from "Vet var din el kommer från" (Know where your electricity comes from" to "Kolla var din el kommer ifrån" (Check where your electricity comes from" as the first one don't really make much sense in Swedish.


madsnedergaard (approved):
Awesome, thanks! 🙌

PR #3923, opened 2022-03-30
Updates Plausible events

Kongkille:
Adds events for clicking on the links in the header and also for the ranking.

Prettier decided to format the sharedHeader code, but the changes are minimal.


web/src/components/zonelist.js
```diff
@@ -93,6 +95,7 @@ const ZoneList = ({
 
   const enterZone = (zone) => {
     dispatchApplication('mapViewport', getCenteredZoneViewport(zone));
+    trackEvent('zoneEntered', {zoneEntered: zone.countryCode});
```
madsnedergaard, line 98: Hmm, I'd suggest either using the same event for selecting a zone (and then have an attribute for determining the source), or having two events clearly named (e.g. "zoneSelectedFromRanking", "zoneSelectedFromMap"). I think the first option is best, but don't know how well that works in Plausible :)

madsnedergaard (commented):
One minor comment, otherwise great :)

web/src/components/zonelist.js
Kongkille: Ok so how about

trackEvent('zoneSelected', {zoneSelected: zone.countryCode, zoneSelectedSource: "ranking"});

Note that the way our event system is setup, we always send all attributes in the context alongside, so all events in plausible has like 10-15 attributes.

web/src/components/sharedheader.js
```diff
@@ -146,24 +157,28 @@ const ResponsiveMenu = ({ children, collapsed }) => {
...
+      <a href="https://app.electricitymap.org/map">
+        <Logo src={logo} alt="logo" />
+      </a>
+      <ResponsiveMenu collapsed={collapsed}>
+        {links.map(({ label, href, active, id }) => (
+          <Link
+            key={id}
+            href={href}
+            active={active}
+            onClick={() => trackEvent('Link Clicked', { linkId: id })}
```
madsnedergaard, line 174: Should we rename to something header-related as talked about?
```suggestion
            onClick={() => trackEvent('HeaderLink Clicked', { linkId: id })}
```

web/src/components/zonelist.js
```diff
@@ -93,6 +95,7 @@ const ZoneList = ({
 
   const enterZone = (zone) => {
     dispatchApplication('mapViewport', getCenteredZoneViewport(zone));
+    trackEvent('ZoneInRanking Clicked', {zone: zone.countryCode});
```
madsnedergaard, line 98: Do you still believe having separate events for this is the best way forward? I'm a bit unsure what we came to a conclusion about 🤔 

Suggestion for alternative naming if we use the same event:  
`'Zone Viewed', {zone: zone.countryCode, source: 'ranking_panel'}`
`'Zone Viewed', {zone: zone.countryCode, source: 'map'}`

(As I think the goal is to view a zone, not to click a link - I think 😅)

web/src/layout/map.js
```diff
@@ -113,11 +111,10 @@ export default () => {
 
   const handleZoneClick = useMemo(
     () => (id) => {
-      trackEvent('countryClick');
```
madsnedergaard, line 116: Why remove this, don't we still want to keep it (but renamed to match conventions?)

web/src/services/thirdparty.js
```diff
@@ -21,7 +20,6 @@ class ConnectionsService {
     this.connections = [];
     if (isProduction()) {
       this.addConnection(new twitterConnection());
-      this.addConnection(new mixpanelConnection());
```
madsnedergaard, line 24: We should also remove mixpanel stuff from index.ejs :)

madsnedergaard (commented):
Some comments :)

web/src/components/zonelist.js
Kongkille: Well, I think for consistency tracking the actions that a user can take and not the resulting experience. I.e. if we add an event for the wind panel, we would track "WindButton Clicked" and not the resulting experience (that the wind layer becomes enabled)

This means that all events would follow "Item + userAction". When that is said, I do like the idea of your events for this case though, I am not sure how important (or even possible) it is to maintain a 100 % consistent events

web/src/layout/map.js
Kongkille: My thoughts here were that it is firing a lot and a lot of the same insights can be seen by the URLs.

We can keep it if we want and rename either to the same event as the one in ranking or something like ZoneOnMap Clicked

web/src/services/thirdparty.js
Kongkille: I don't think there is anything? I've tried searching for things from the docs

https://developer.mixpanel.com/docs/javascript-quickstart

madsnedergaard: Oh you're right, I don't know where I saw this 🤔

web/src/components/zonelist.js
madsnedergaard: Ah, now I read [the Plausible documentation](https://plausible.io/docs/custom-event-goals) and realised that `events !== goals` (we manually set those up). Also seems like I'm a bit stuck on how Amplitude is doing it/recommending it :)
Then I think your solution is fine, and I guess we can just track this specific link and based on that figure out how many people end on `/zone/DE` via this event or other means.

web/src/layout/map.js
madsnedergaard: Thanks for the response, I agree now 🙃

madsnedergaard (approved):
After reading docs and restructuring my thinking, I'm good with merging this PR as it is now 🥳

Issue #1397, opened 2018-05-10
Split Sweden to bidding zones


tmslaine:
This is not exactly compliant with the 'keep calculations to minimum' policy, but it might be possible to show a generation mix for Swedish bidding zones.

- Statnett/Svenska kraftnät have real-time flow data
- Statnett/Svenska kraftnät have a rough fuel mix for all of Sweden
- ENTSO-E has wind generation for Swedish bidding zones
- ENTSO-E/Svenska kraftnät have consumption for the bidding zones

Moreover, based on statistics by Svenska kraftnät at https://www.svk.se/aktorsportalen/elmarknad/statistik/ we can deduce:
- All nuclear capacity is located in the bidding zone SE3
- Over 98% of annual non-wind power generation in SE1 comes from hydro
- Over 97% of annual non-wind power generation in SE2 comes from hydro
- SE3 has annually 5 to 9 times more hydro generation than SE4 (on average 2010-2016 7.4 times)

Then to calculating hydro and thermal generation for each zone:
- Consumption and flow data allow calculation of total generation for each bidding zone
- We can subtract wind generation(ENTSO-E, all areas) and nuclear generation(statnett/SE3) from the total generation figures to get (hydro+unknown) generation for all areas
- By assuming that hydro makes up let's say 98% and 97% of (hydro+unknown) generation in SE1, and SE2, respectively, we have the fuel mix down for SE1 and SE2
- Subtracting SE1 and SE2 generation from the Statnett generation figures for all of Sweden, we have an estimate for the fuel mix of (SE3+SE4) and the total generation for both SE3 and SE4

The remaining problem is to solve the hydro-to-unknown ratios for SE3 and SE4:
- We could fix that something like 12% of hydro generation in (SE3+SE4) occurs in SE4, or use some MW lower/upper bounds for hydro/thermal to get the fuel mix for SE3 and SE4
- Alternatively, we could lump SE3 and SE4 together

maxbellec:
This is awesome, thanks a lot for this. We really want to push regional consumptions as much as possible, even if it means a bit more calculations.

I'll submit a PR soon. ENTSOE publishes onshore wind per zone, but do you know if SE has any offshore wind that might not be published?

Notes for me: 
- [x] OK for exchanges, only a small modification in `statnett.py`
- [x] already done for all SE production
- [ ] ENTSOE wind production per zones: offshore wind?

tmslaine:
[...]

PR #3924, opened 2022-03-31
Added and fixed capacities for SE exchanges

VIKTORVAV99:
I have added or fixed current exchanges with Sweden as a whole and added and fixed the capacities for Sweden's bidding zones in preparation for #1397.

A noticeable difference that will show up right away is the added capacities for FI->SE and NO-NO4->SE that where completely missing before.

I used this image as the source from a SVK power balance [report.](https://www.svk.se/siteassets/om-oss/rapporter/2021/kraftbalansen-pa-den-svenska-elmarknaden-rapport-2021.pdf)
image

### DIFF
--- a/config/exchanges.json
+++ b/config/exchanges.json
@@ -909,14 +909,8 @@
     "rotation": 90
   },
   "DE->SE": {
-    "capacity": [
-      -600,
-      600
-    ],
-    "lonlat": [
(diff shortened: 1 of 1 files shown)


madsnedergaard (approved):
Awesome, it's super cool with the new exchanges! 🥳

PR #3926, opened 2022-03-31
Update co2eq parameters with generalised yearly aggregates

pierresegonne:
## Description

1. Implement the new format for fallback mixes:

either:

```
"powerOriginRatios": [
  {
    "datetime": "xxx",
    "_source": "xxx",
    "value": {
       "oil": 0.1,
       "wind": 0.9 
    }
  }, ...
]
```

or 

```
"powerOriginRatios": {
    "_source": "xxx",
    "value": {
       "oil": 0.1,
       "wind": 0.9   
    }
  }
```

2. Generate yearly fallback mixes for zone overrides
3. Generate yearly emission factors for direct zone overrides


madsnedergaard:
_(Slightly unrelated, but it would be great to replace the JS test with a Python one as we already have Python tests for other config files)_

pierresegonne:
> (Slightly unrelated, but it would be great to replace the JS test with a Python one as we already have Python tests for other config files)
I'll create an issue for that

pierresegonne:
Note that I realised we also need to test that all elements of yearly lists have a datetime key.

gwpicard (approved):

PR #3929, opened 2022-04-01
Moves message so it's not repeated and improves it

madsnedergaard:
The message was rendered multiple times and was a bit messy. I tried cleaning it up now :)

### Before

Screenshot 2022-04-01 at 13 06 01

### After
Screenshot 2022-04-01 at 13 40 22

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -24,7 +24,7 @@ console.log(
 🌍 %cReady to work on fixing the climate full-time?
   https://electricitymap.org/jobs
 🐙 Got comments or want to contribute?
-  https://github.com/tmrowco/electricitymap-contrib`,
+  https://github.com/electricitymap/electricitymap-contrib`,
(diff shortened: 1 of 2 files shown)


web/src/main.js
```diff
@@ -18,6 +18,17 @@ import GlobalStyle from './globalstyle';
 // init localisation
 import './helpers/i18n';
 
+// Initial greeting message for curious people
+console.log(
+  `%cWelcome to electricityMap!
+🌍 %cReady to work on fixing the climate full-time?
+  https://electricitymap.org/jobs
+🐙 Got comments or want to contribute?
+  https://github.com/tmrowco/electricitymap-contrib`,
```
madsnedergaard, line 27:
```suggestion
  https://github.com/electricitymap/electricitymap-contrib`,
```

Kongkille (approved):

PR #3930, opened 2022-04-02
Makes translation path absolute (to fix issues when changing language on a zone)

madsnedergaard:
I made a mistake and made the locales-path relative, which means it only works when language is changed on the root path (/map) 😬 

This will fix it!


madsnedergaard:
Self-merged this as it's currently causing problems for users.

PR #3934, opened 2022-04-04
Map works with minimized state files

Kongkille:
This PR decouples the global grid state from the history state.

The hook "useCurrentZoneHistory" currently grabs the initial data used on the countrypanel from the state file. We now instead fetch the initial data from the latest zonehistory.

This allows us to have much less data within the global state file and instead have all complex data required to produce the graphs and so forth be fetched when a user clicks on the country panel.


madsnedergaard:
Looks great, how thoroughly did you test it? Are there any edge cases which we don't catch here? (e.g. could state/history files be out of sync on the datetime?) 🤔

Kongkille:
Yes the biggest potential issue would indeed be them being out of sync. I tested quite a few things, so far it seems to work great. 

The functionality change is also minimal as there is the zonetimeselector. It is only when the zonetimeselector is null. I.e. the first time a user clicks on any country this could be the case. Otherwise the map works as previously

madsnedergaard:
Seems like CircleCI is [currently down](https://status.circleci.com/) :(

web/src/hooks/redux.js
```diff
@@ -55,7 +55,10 @@ export function useCurrentZoneData() {
       return null;
     }
     if (zoneTimeIndex === null) {
-      return grid.zones[zoneId];
+      // Return latest history data unless latest history does not correspond to the current grid datetime
+      return (zoneHistory[zoneHistory.length - 1].stateDatetime === grid.datetime
+        ? zoneHistory[zoneHistory.length - 1]
+        : null);
```
madsnedergaard, line 61: This change also broke how the mockserver is implemented.. We should either in the mockserver manually override stateDatetime in the response, or consider if we can avoid having this check? 🤔

Kongkille: Ah good point, I only added this check while finally working with production data.

I believe the mockserver also has the issue with the state being somewhere in 2017, whereas the history file is recent. So we could perhaps also just update the mockserver files, as it is a fair assertion that the state and history files are somewhat related.

On the other hand, it is not like returning null really fixes anything if the files do become out of sync, so I am down with removing the check as well, lmk which one you'd prefer

PR #3935, opened 2022-04-04
Fixes generate-bounding-boxes script

madsnedergaard:
It now supports recent changes to world.geojson.
It should work with both polygons and multipolygons. I also changed the script to require a zone parameter, as we most of the time will only require changing/adding boundingBoxes for a single zone.

I have on purpose not gone deeper into cleaning up this script, but we could very well do so in future :)

### Preview

Screenshot 2022-04-04 at 16 56 15


web/geo/generate-zone-bounding-boxes.js
```diff
@@ -6,15 +6,20 @@ const { getJSON } = require('./utilities');
 const { getZonesJson, saveZonesJson } = require('./files');
 
 let zones = getJSON(path.resolve(__dirname, './world.geojson'));
-if (args.length > 0) {
-  zones = zones.filter((d) => d.properties.zoneName === args[0]);
+
+
+if (args.length <= 0) {
+  console.log('ERROR: Please add a zoneName parameter ("node generate-zone-bounding-boxes.js DE"')
```
q--, line 12: Missing closing parenthesis in the logged text

madsnedergaard, line 12:
```suggestion
  console.log('ERROR: Please add a zoneName parameter ("node generate-zone-bounding-boxes.js DE")')
```

madsnedergaard: Good catch!

Kongkille (approved):
Makes sense, it's odd, I ran it when I did the geometry changes, but I guess there were only multipolygons present at that time

In any case, yeah I think we could clean this up next time we look at it. However, there is also a bigger issue with our bounding box approach as it doesn't work to take the bounding box of a multipolygon crossing 180 meridian, and a few other things

Issue #3927, opened 2022-03-31
Replace co2eq parameters .js test to python test for config files


pierresegonne:
## Description

Currently, the co2eq parameters are being tested by [co2eq_parameters.test.js](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters.test.js). In order to improve consistency of the codebase, it would be nice to port these tests in python and remove the js code.

kruschk:
I could use some practical experience with unit testing; could I take a shot at this?

pierresegonne:
Of course!

kruschk:
I've almost got a PR ready! I had to invoke a sketchy regex to extract the `modeOrder` array from [constants.js](https://github.com/electricitymap/electricitymap-contrib/blob/master/web/src/helpers/constants.js), but I think it should work well enough. Should the [co2eq_parameters_direct.json](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_direct.json) file be included in the tests as well? At the moment, only [co2eq_parameters_all.json](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_all.json) and [co2eq_parameters_lifecycle.json](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_lifecycle.json) are included.

pierresegonne:
Super!
Yes if you can, please do it :)

kruschk:
I noticed the [co2eq_parameters_direct.json](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_direct.json) and [co2eq_parameters_lifecycle.json](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_lifecycle.json) files are similar enough that if they are combined into one object (similarly to the current approach), whichever comes second will clobber some of the keys in the first. This could easily lead to a situation where all tests pass even though one of the files isn't really valid. I think that a more robust approach would be to test each of the three files separately, applying only the relevant tests from the original test suite to each. Does this sound reasonable?

pierresegonne:
Yes indeed. I think you can separate the tests for each file as they each represent different information. For the distinction lifecycle VS direct, you can probably write a single test that can be executed independently on either file :)

CorentinBrtx:
Hi, looks like this issue can be closed ?

VIKTORVAV99:
[...]

PR #3937, opened 2022-04-05
test: port CO2eq parameters tests to Python

kruschk:
Refs: #3927


madsnedergaard:
Awesome! With this, we should also remove the `test_js` step in `.circleci/config.yaml` (plus remove it from `workflows` at the bottom):
https://github.com/electricitymap/electricitymap-contrib/blob/master/.circleci/config.yml#L45-L53

pierresegonne:
Note that in https://github.com/electricitymap/electricitymap-contrib/pull/3926 I added another set of tests - verify that all zoneOverrides that are provided as a list of yearly values have a datetime associated :)

kruschk:
I tried to leave a comment last night regarding the failing Jest and Pylint tests, but it seems to have gotten lost somehow! Anyway, thanks @madsnedergaard for reading my mind and recommending to remove `test_js`. I also managed to make Pylint happy by appending "Mixin" to my `CO2eqParametersDirectAndLifecycle` class, because that's exactly what it is; I just didn't know the right terminology (or naming convention) for it.

@pierresegonne I'll modify the tests to reflect the changes to the JSON files! I might abandon the idea of checking the existence of keys in a dedicated test and do those checks within the other tests, as needed, instead. It's getting a bit unwieldy and repetitive to do a dedicated test.

kruschk:
@pierresegonne, I pushed some more changes, but I don't think I'm able to request a review without being a repository collaborator. Could you please review the PR?

kruschk:
@pierresegonne, I pushed some more changes, but I don't think I'm able to request a review without being a repository collaborator. Could you please review the PR?

tests/test_co2eq_parameters.py
```diff
@@ -0,0 +1,188 @@
...
+class CO2eqParametersAll(unittest.TestCase):
+    """A test case for the CO2eq parameters."""
+    modes = get_possible_modes()
+    parameters = parse_json_file("config/co2eq_parameters_all.json")
+
+    @classmethod
+    def check_power_origin_ratios(cls, callback1, callback2):
+        """Apply the callback to each ratio in the 'powerOriginRatios' objects.
+        """
+        # It is useful to keep `test_power_origin_modes_are_valid` and
```
kruschk, line 43: This comment is already out of date, so I'd better remove the first sentence and keep the second.

tests/test_co2eq_parameters.py
```diff
@@ -0,0 +1,188 @@
...
+            self.assertAlmostEqual(
+                1.0,
+                sum(ratio for ratio in ratio["value"].values()
+                    if isinstance(ratio, numbers.Number)),
+                msg=f"zone '{zone}' ratios do not sum to (approximately) 1.0",
+                places=2)
+
+        self.check_power_origin_ratios(callback, callback)
+
+    def test_required_keys_are_present(self):
```
kruschk, line 93: I think the `test_required_keys_are_present` methods are overkill. It would probably be better to just remove them, and they'll exist implicitly in the other tests, relying on `KeyError`s to be raised if an expected key doesn't exist.

pierresegonne (approved):
Awesome work!
Thanks a lot for taking the time to do this, it will enable to make tests even more stricter quite easily. Example of a possible extension would be to verify the values of the different emission factors. But that's outside the scope of this :)

kruschk:
@pierresegonne Thanks! Perhaps a follow-up issue is in order?

PR #3939, opened 2022-04-05
Fix translations on phones - again 🙃 

madsnedergaard:
Seems like [my recent attempt at fixing things on web](https://github.com/electricitymap/electricitymap-contrib/pull/3930) broke mobile... This time it should work for both! 
(I have tested with iOS simulator and web)

### DIFF
--- a/web/src/helpers/i18n.js
+++ b/web/src/helpers/i18n.js
@@ -58,7 +58,7 @@ i18n
   .use(initReactI18next)
   .init({
     fallbackLng: 'en',
-    debug: isProduction() ? true : true,
+    debug: isProduction() ? false : true,
     backend: {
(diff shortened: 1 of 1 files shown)


web/src/helpers/i18n.js
```diff
@@ -56,9 +58,9 @@ i18n
   .use(initReactI18next)
   .init({
     fallbackLng: 'en',
-    debug: isProduction() ? false : true,
+    debug: isProduction() ? true : true,
```
madsnedergaard, line 61: Let's keep debugging off
```suggestion
    debug: isProduction() ? false : true,
```

pierresegonne (approved):

PR #3945, opened 2022-04-06
update capacities Lithuania

KabelWlan:
Entsoe finally updated Lithuania's capacities. The source is still the same.


madsnedergaard (approved):
Great!

PR #3947, opened 2022-04-07
Update solar & wind capacity for Netherlands

Macquaria:
2021 data from CBS : https://www.cbs.nl/en-gb/news/2022/10/more-electricity-from-renewable-sources-less-from-fossil-sources

Replacing inaccurate ENTSO-E data.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -4563,8 +4563,8 @@
       "hydro storage": 0,
       "nuclear": 485,
       "oil": 0,
-      "solar": 16074,
-      "wind": 7303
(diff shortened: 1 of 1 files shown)


madsnedergaard (approved):
Good find, thanks for updating these values!
Now we shouldn't be over 100% capacity 🥳 
Screenshot 2022-04-07 at 10 18 28

Macquaria:
Indeed, today's 100,04% was precisely the reason why I looked for new data haha

Thank you for your impressive reactivity ! 😃

PR #3951, opened 2022-04-09
Additional ENTSO-E parsers for GE and switched `GE->RU-1` from ru parser to ENTSO-E.

VIKTORVAV99:
In this PR:

- Added ENTSO-E mappings for Georgia and Azerbaijan.

- Changed `GE->RU-1` exchange from RU parser to ENTSO-E since the RU parser is currently down.

- Added ENTSO-E forecasts to the exchanges. 

- Additional ENTSO-E parsers for Georgia.

ref: #3408


madsnedergaard (approved):
Awesome! 🙌

Issue #2885, opened 2020-12-30
Change Croatia source to ENTSOE


Lightning-5:
Current source for Croatia is lacking information on type of electricity production. Since information on Croatia is also available on ENTSOE I think that ENTSOE should be uses as a source

[ENTSOE Croatia](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=true&viewType=TABLE&areaType=BZN&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=30.12.2020+00:00|CET|DAYTIMERANGE&dateTime.endDateTime=30.12.2020+00:00|CET|DAYTIMERANGE&area.values=CTY|10YHR-HEP------M!BZN|10YHR-HEP------M&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2))

PaulCornelissen:
This is very good news!

However if I look at today, they seem to be >4hrs behind, so no good for live info. At the moment the data is 7hrs behind.

If this turns out to be just a fluke and the data is normally reliable, I'd be happy to do a PR to fix this.

Otherwise we can combine the current parser with the ENTSOE parser for older data and use this data to update the unknown intensity?

FelixDQ:
The ENTSOE data is still lagging with ~4 hours, so I don't think that we should switch data source right now.

I'm going to close this issue for now, and then we can revisit it if the lag on ENTSOE changes in the future.

PR #3953, opened 2022-04-09
Switch Croatia (HR) to ENTSO-E for production data. (Significantly reduces unknown data)

VIKTORVAV99:
In this PR:

- Switches Croatia (HR) to the ENTSO-E parser to significantly reduce the amount of unknown production data. (Roughly from 45% to 2% at the time I'm writing this.)
- Adds ENTSO-E forecasts to generation and exchanges.
- Removes now unused `HOPS.py` parser, `test_HOPS.py` test file and files associated with the test file.

ref: #3408


q--:
This was proposed earlier, see  #2885

Back then the problem was that ENTSO-E data lagged significantly compared to HOPS. I don't know if things are better now, but it's something to consider

VIKTORVAV99:
> This was proposed earlier, see  #2885
> 
> Back then the problem was that ENTSO-E data lagged significantly compared to HOPS. I don't know if things are better now, but it's something to consider

I did a quick check before making the PR and then the data only lagged 1-2 hours so that shouldn't be a problem.

What might be a problem is they seems to report 0 production in the future and later correcting it with the correct value. For example right now it's 14:30 local time but Croatia has 0 value production for installed sources reported till midnight.

VIKTORVAV99:
Hmm looks like the data reported to ENTSO-E is very irregular, ranging from a 1-hour to delay to a 5-hour delay at the time of writing, we might need to hold off on this PR until further notice. 😐

VIKTORVAV99:
@Kongkille now that historical view has launched I feel like this is worth revisiting.
Do you think the additional source breakdown (and therefore more accurate co2eq/kWh) is worth the longer data delay?

VIKTORVAV99:
@pierresegonne pinging you in here as well since I think it would be a net improvement (especially for historical data) to switch to the ENTSO-E parser for Croatia even if it has significant delays (I have observed up to 8 hours delays but could be longer) in the data and then fill the gap with a estimation model.

I'm not sure what data or how much data is needed to create a estimation model but ENTSOE-E seem to have almost full production data (no geothermal or waste) since 2019-01-01. On 2019-02-01 they added geothermal data and on 2019-07-26 they added waste data.

Kongkille:
Hmm it's hard to factor in the value of not having real-time data tbh.

@pierresegonne let's discuss

pierresegonne:
Hey, sorry for the delay in coming back to you guys. 

I'd say that it depends on how easily predictable the Croatian grid is.
From your experience, what have you seen @VIKTORVAV99 ? What are the main production modes? 

If there are not too many variations, then we know that the model will do an OK job at estimating the real time data, and in that case, I think the trade-off is worth it :)

VIKTORVAV99:
> Hey, sorry for the delay in coming back to you guys. 
> 
> I'd say that it depends on how easily predictable the Croatian grid is.
> From your experience, what have you seen @VIKTORVAV99 ? What are the main production modes? 
> 
> If there are not too many variations, then we know that the model will do an OK job at estimating the real time data, and in that case, I think the trade-off is worth it :)

The main production modes seem to be:
- Biomass - pretty consistent at around 66-68mw the last few days (at least)
- Gas - ranging from 350 MW to 450MW
- Coal - 200±1MW
- Geothermal - 6±1MW
- Hydro - 50 to 500 MW
- Solar - 0 to 42MW
- Unknown - 32±1MW
- Wind - 4 to 200 MW

So everything except hydro, solar and wind seem to be pretty consistent and solar at least should be easy enough to estimate.

It also has solar and wind forecast available, is that something we use in the estimation models?

pierresegonne:
> It also has solar and wind forecast available, is that something we use in the estimation models?

Not yet unfortunately 

---

I think the tradeoff is worth it. I'm not going to merge it straight away - as I'm still doing some post-processing following the merging of SE but will do unless @Kongkille or @tonypls or @mathilde-daugy have strong opinion against

VIKTORVAV99:
> > It also has solar and wind forecast available, is that something we use in the estimation models?
> 
> Not yet unfortunately
> 
> I think the tradeoff is worth it. I'm not going to merge it straight away - as I'm still doing some post-processing following the merging of SE but will do unless @Kongkille or @tonypls or @mathilde-daugy have strong opinion against

I just added a small edit to my comment, please check it as it might impact this.

VIKTORVAV99:
> > It also has solar and wind forecast available, is that something we use in the estimation models?
> 
> Not yet unfortunately
> 
> I think the tradeoff is worth it. I'm not going to merge it straight away - as I'm still doing some post-processing following the merging of SE but will do unless @Kongkille or @tonypls or @mathilde-daugy have strong opinion against

There is also a lot of conflicts and a few things have changed since I created this PR originally so even if we want to merge it I have some cleanup to do first.

VIKTORVAV99:
Looks like the data for HR is being reported hourly now to ENTSO-E so this should have no issues being merged as the previous blockers have been resolved.

I'll fix all the merge conflicts and mark this PR as ready tomorrow (hopefully).

VIKTORVAV99:
I'm running into some issues with lint-staged and eslint and prettier failing cause the stings are too long??? But since the actual changes are pretty small I think I'll just re-create the PR on a clean branch.

PR #3955, opened 2022-04-11
Fix race condition where map doesnt always center

corradio:
Summary of changes:
- Fixes https://github.com/electricitymap/electricitymap-contrib/issues/3024
- Clean up a few logs

### DIFF
--- a/web/src/layout/map.js
+++ b/web/src/layout/map.js
@@ -46,18 +46,23 @@ export default () => {
 
   const [tooltipPosition, setTooltipPosition] = useState(null);
   const [tooltipZoneData, setTooltipZoneData] = useState(null);
+  const [hasCentered, setHasCentered] = useState(false);
+
+  // Center the map initially based on the focused zone and the user geolocation.
+  if (!hasCentered) {
+    if (zoneId) {
+      console.log(`Centering on zone ${zoneId}`);
+      dispatchApplication('mapViewport', getCenteredZoneViewport(zones[zoneId]));
+      setHasCentered(true);
+    } else if (callerLocation) {
+      console.log(`Centering on browser location (${callerLocation})`);
+      dispatchApplication('mapViewport', getCenteredLocationViewport(callerLocation));
+      setHasCentered(true);
+    }
+  }
 
   const handleMapLoaded = useMemo(
     () => () => {
-      // Center the map initially based on the focused zone and the user geolocation.
-      if (zoneId) {
-        console.log(`Centering on zone ${zoneId}`);
-        dispatchApplication('mapViewport', getCenteredZoneViewport(zones[zoneId]));
-      } else if (callerLocation) {
-        console.log(`Centering on browser location (${callerLocation})`);
-        dispatchApplication('mapViewport', getCenteredLocationViewport(callerLocation));
-      }
-
       // Map loading is finished, lower the overlay shield with
       // a bit of delay to allow the background to render first.
       setTimeout(() => {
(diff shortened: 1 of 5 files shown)


web/src/reducers/dataReducer.js
```diff
@@ -132,32 +132,14 @@ const reducer = (state = initialDataState, action) => {
...
-          if (zone.production[mode] !== undefined && zone.production[mode] < 0) {
-            console.warn(`${key} has negative production of ${mode}`);
-          }
-          // Check load factors > 1
-          if (zone.production[mode] !== undefined
-            && (zone.capacity || {})[mode] !== undefined
-            && zone.production[mode] > zone.capacity[mode]) {
-            console.warn(`${key} produces more than its capacity of ${mode}`);
-          }
-        });
```
corradio, line 152: This validation block is either redundant with what we already do in the backend, or simply not useful. It produces a lot of logs, and I recommend removing them to reduce the clutter.

web/src/sagas/index.js
```diff
@@ -44,7 +44,6 @@ function* fetchGridData(action) {
     const payload = yield call(protectedJsonRequest, endpoint);
     yield put({ type: 'TRACK_EVENT', payload: { eventName: 'pageview' } });
     yield put({ type: 'APPLICATION_STATE_UPDATE', key: 'callerLocation', value: payload.callerLocation });
-    yield put({ type: 'APPLICATION_STATE_UPDATE', key: 'callerZone', value: payload.callerZone });
```
corradio, line 47: This is never used

web/src/services/thirdparty.js
```diff
@@ -34,7 +34,6 @@ class ConnectionsService {
   }
 
   trackEvent(eventName, context) {
-    console.log("tracking totally");
```
corradio, line 37: appears a lot and seems verbose.

web/src/layout/map.js
```diff
@@ -46,18 +46,23 @@ export default () => {
...
+    if (zoneId) {
+      console.log(`Centering on zone ${zoneId}`);
+      dispatchApplication('mapViewport', getCenteredZoneViewport(zones[zoneId]));
+      setHasCentered(true);
+    } else if (callerLocation) {
+      console.log(`Centering on browser location (${callerLocation})`);
+      dispatchApplication('mapViewport', getCenteredLocationViewport(callerLocation));
+      setHasCentered(true);
+    }
+  }
```
madsnedergaard, line 62: Hmm, shouldn't this ideally happen in a `useEffect()` block with `hasCentered` as dependency instead? 🤔

web/src/reducers/dataReducer.js
madsnedergaard: One could argue that it might be helpful for users/contributors to identify zones that needs updated capacities... But not sure if anyone ever uses it, so I'm fine with removing it :)

corradio: good point.

web/src/layout/map.js
corradio: of course you're right 🤦

web/src/layout/map.js
```diff
@@ -46,29 +46,33 @@ export default () => {
...
       }
+    }
+  }, [zones, zoneId, callerLocation, hasCentered]);
 
-      // Map loading is finished, lower the overlay shield with
-      // a bit of delay to allow the background to render first.
-      setTimeout(() => {
-        dispatchApplication('isLoadingMap', false);
-      }, 100);
+  const handleMapLoaded = () => {
```
corradio, line 66: this used a `useMemo` but now it doesn't need to anymore!

madsnedergaard (approved):
LGTM!

PR #3961, opened 2022-04-13
Font optimizations

VIKTORVAV99:
In this PR:

- Added woff2 versions of the OpenSans font files using [google/woff2](https://github.com/google/woff2). (size reductions)

- Moved the font css files to the scss bundle (and renamed them with the scss extension) to reduce the amount of http requests on page load.

- Added the woff2 versions of OpenSans to the `@font-face`

- Added font preloading to `index.ejs` so we don't have to wait for the css bundle to start loading the fonts.

- Added `font-display: swap` to the font faces to reduce CLS.

I have tested this locally and found no problems but please test it out first.


web/views/pages/index.ejs
```diff
@@ -39,14 +39,17 @@
         <!-- Pre-connect to improve performances -->
         <link rel="preconnect" href="https://api.electricitymap.org/">
```
VIKTORVAV99, line 40: On a slightly off topic note, is this URL actually used anywhere in the frontend?
I can't find a reference to it or see any requests to it when using dev tools to inspect the webpage (both locally and public version)

madsnedergaard: You're right, it's not - it should be changed to `https://app-backend.electricitymap.org` :)

madsnedergaard:
I just took the liberty of updating `master` and fixing a merge conflict while I was testing it locally :)

madsnedergaard (approved):

Issue #3626, opened 2021-12-29
CA-AB price parser: prefer fetch CSV


Manu1400:
# URL
GET http://ets.aeso.ca/ets_web/ip/Market/Reports/SMPriceReportServlet?contentType=csv

# CSV as text

```csv
Pool Price

""

Date (HE),Price ($),30Ravg ($),AIL Demand (MW)
"12/29/2021 08","-","-","-"
"12/29/2021 07","75.57","122.30","10735.0"
"12/29/2021 06","84.35","122.31","10480.0"
"12/29/2021 05","69.26","122.28","10360.0"
"12/29/2021 04","55.04","122.25","10336.0"
"12/29/2021 03","63.58","122.24","10359.0"
"12/29/2021 02","60.62","122.22","10414.0"
"12/29/2021 01","44.46","122.21","10473.0"
"12/28/2021 24","41.92","122.22","10687.0"
"12/28/2021 23","38.09","122.24","10996.0"
"12/28/2021 22","45.69","122.27","11220.0"
"12/28/2021 21","67.97","122.29","11357.0"
"12/28/2021 20","148.61","122.28","11365.0"
"12/28/2021 19","372.00","122.16","11467.0"
"12/28/2021 18","646.74","121.74","11606.0"
"12/28/2021 17","421.32","120.95","11469.0"
"12/28/2021 16","113.79","120.47","11236.0"
"12/28/2021 15","284.36","120.40","11206.0"
"12/28/2021 14","148.98","120.09","11152.0"
"12/28/2021 13","637.64","119.97","11133.0"
"12/28/2021 12","667.24","119.17","11182.0"
"12/28/2021 11","262.40","118.33","11160.0"
"12/28/2021 10","95.87","118.05","11004.0"
"12/28/2021 09","154.78","118.01","10922.0"


```

# Now
In parser https://github.com/electricityMap/electricitymap-contrib/blob/master/parsers/CA_AB.py#L82
GET http://ets.aeso.ca/ets_web/ip/Market/Reports/SMPriceReportServlet?contentType=html/

# Checklist
- [X] Same values into CSV

IV1T3:
fetch_production() and fetch_exchange() can also be implemented using CSV data:
http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet?contentType=csv

kruschk:
@madsnedergaard I believe this issue can be closed now!

PR #3962, opened 2022-04-14
chore(parser): fetch CSV instead of HTML in CA-AB

kruschk:
The AESO ETS website now exposes a CSV interface in addition to HTML;
modify the CA-AB parser to take advantage of the simpler API.

Refs: #3626


parsers/CA_AB.py
```diff
@@ -1,161 +1,153 @@
...
-    return [data[k] for k in sorted(data.keys())]
-
-
-def fetch_exchange(zone_key1='CA-AB', zone_key2='CA-BC', session=None, target_datetime=None, logger=None) -> dict:
-    """Requests the last known power exchange (in MW) between two countries."""
+        raise NotImplementedError('Currently unable to scrape historical data')
+    session = session or requests.Session()
+    response = session.get(f'{URL_STRING}/SMPriceReportServlet',
+                           params={'contentType': 'csv'})
+    return [
```
kruschk, line 67: The API serves the hourly prices in reverse-chronological order; I'm not sure if they should be reversed, or whether the order matters at all.

parsers/CA_AB.py
```diff
@@ -1,161 +1,153 @@
...
+        'source': URL.netloc,
     }
 
 
-def fetch_price(zone_key='CA-AB', session=None, target_datetime=None, logger=None) -> dict:
-    """Requests the last known power price of a given country."""
+def fetch_price(zone_key=DEFAULT_ZONE_KEY,
+                session=None,
+                target_datetime=None,
+                logger=None) -> dict:
```
kruschk, line 60: The return type should be `list<dict>`, not `dict`.

parsers/CA_AB.py
```diff
@@ -1,161 +1,153 @@
...
+            'zoneKey': zone_key,
+        } for row in csv.reader(response.text
+                                        .split('\r\n\r\n')[2]
+                                        .splitlines()[1:]) if row[1] != '-']
+
+
+def fetch_production(zone_key=DEFAULT_ZONE_KEY,
+                     session=None,
+                     target_datetime=None,
+                     logger=None) -> dict:
```
kruschk, line 83: `validation.validate` expects a valid `logger` argument, but it defaults to `None` here. I should change this to `logger=logging.getLogger(__name__)`.

parsers/CA_AB.py
madsnedergaard: I couldn't tell from looking at the backend either, did you check how other parsers does it? I guess the ENTSO-E or FR should be trustworthy 🤔

kruschk: ENTSO-E and FR appear to return values in chronological (i.e., most recent last) order; however, neither parser explicitly sorts the results; the order is just a consequence of how the API presents the data. I tried shuffling the list with `random.shuffle`, and `poetry run test_parser CA-AB price` was successful no matter what order was presented to it. This leads me to believe that the order doesn't matter.

madsnedergaard: It probably doesn't, but thanks for looking into it :)

madsnedergaard (approved):
LGTM! 🥳

PR #3963, opened 2022-04-14
Arrow optimizations

VIKTORVAV99:
In this PR:

- Moved arrows into the arrows folder.

- Added webp versions of arrows using gif2webp from googles libwebp using default settings.

- Modified `exchangearrows.js` to return a picture element with the webp version of the arrows and gif versions as a fallback.

This should cut the size of the arrows served to all clients that support webp by roughly half. 

**Note:** As the webp files are converted from the GIF files there will be some quality loss but in all honesty I can't tell the difference when testing locally but please test it yourselves and see if this is acceptable.


Kongkille (approved):
Sorry for the late review, this works great. Nice find!

VIKTORVAV99:
@Kongkille hehe don't worry about it, as you may have noticed I got busy finding other areas in the code to optimize. 😉

PR #3965, opened 2022-04-14
improve: Cyprus geometries

### LINKED ISSUE #2724: Improve geometries of Cyprus
Most parts of the world map are correctly charted as such, but a few of the regions if you zoom in a bit have got inaccurate borders. 
For example, Northern Cyprus and Cyprus seem to have body of water between which is incorrect - they are part of the same island.
I figure the map might not be intended to scale since its used to track energy consumption and all. Though it would be great to improve these parts.

image

dougfabris:
It improves the Cyprus Island geometries
Closes #2724

### before
Screen Shot 2022-04-14 at 16 53 35

### after
Screen Shot 2022-04-14 at 16 54 08

**EDIT: Simplified version**


veqtrus:
I think that is too detailed for the map's needs. I don't know how the geometries are generated though.

Perhaps we should somehow annotate that the British bases are also included. Or if there is a way to have them separate but colour them the same as Cyprus. I'm not aware of a public feed of consumption there.

dougfabris:
@veqtrus The map is too detailed now because I spend my time detailing it as much as I could, I don't see why it's a problem since is so much closer to the real geography of the island. Also as the proposal is about the electricity of these regions I don't see why separate or highlight the British base if they're part of the South or North.

veqtrus:
The bases are part of the British Overseas Territories. They are not controlled by the Republic of Cyprus.

dougfabris:
IMO we should split into a separate area/zone if they produced their own energy. Let's see the maintainer's opinion here!

VIKTORVAV99:
> @veqtrus The map is too detailed now because I spent my time detailing it as much as I could (it's a manual job), I don't see why it's a problem since is so much closer to the real geography of the island. Also as the proposal is about the electricity of these regions I don't see why separate or highlight the British base if the area is controlled by the Republic of Cyprus.

The problem is that if it is too detailed it will generate a significantly bigger file that will take longer to load and might cause problems on lower end devices. 
This would impact the whole app as the world.json file is currently bundleded with everything else.

Now this is probably not a issue with Cyprus alone being so detailed but the goal should be that every part of the map has the same quality and then it becomes a real problem with having to much detail.

Kongkille (approved):
Thanks @dougfabris! Which tool are you using to draw with?

I also think that this is slightly higher fidelity than what we can afford all places, but it is certainly quite an upgrade to the current version of the Cypern geometries.

I am not sure how/if to incorporate the british areas, so I suggest we just start by merging this PR

dougfabris:
@Kongkille I'm using the recommended, QGIS! Thinking about the comments I agree with u and @VIKTORVAV99 about the map being so much detailed could increase significantly the bundle size. I'm going to edit again correcting the conflict and trying to reduce the details, sound good?

dougfabris:
@Kongkille Conflicts fixed and the new geometries ready!

Kongkille:
@dougfabris Excellent, I'll get it merged! :)

PR #3970, opened 2022-04-15
Update Turkey Renewable Capacities

nessie2013:


KabelWlan:
I‘d suggest using the even more recent data from [Teias](https://www.teias.gov.tr/tr-TR/kurulu-guc-raporlari) which is already listed in data sources.

madsnedergaard (approved):
Awesome! 🙌

PR #3971, opened 2022-04-15
Update Cyprus Renewable Capacities

nessie2013:


madsnedergaard (approved):
Great! 🙌

config/zones.json
```diff
@@ -1521,21 +1521,22 @@
...
       "coal": 0,
       "gas": 0,
       "geothermal": 0,
       "hydro": 0,
       "hydro storage": 0,
       "nuclear": 0,
       "oil": 1478,
-      "solar": 262.7,
-      "wind": 157.5
+      "solar": 316,
```
veqtrus, line 1532: This is overridden by the live parser. Perhaps we should disable it if the IRENA values are more accurate.

madsnedergaard: Interesting, here's the capacities returned by the parser - seems like they are quite behind, so I think your idea of removing them from the parser makes good sense :)

PR #3972, opened 2022-04-15
Added Indonesia Renewable Capacities

nessie2013:


madsnedergaard:
Could we find a source that also has non-renewable capacities?

nessie2013:
So I found a source for non renewable capacities that looks decent, although its from 2020 its the most recent data I could find. Values come from page 56 in the PDF and this image 
image
The only problem is PLT EBT (which according to google might be solar) not matching with IRENA values if it is in-fact solar, also Diesel fuel has been marked into oil, is this correct categorisation?

madsnedergaard (approved):
Nice, thanks! For the renewables I think it makes sense to go with latest numbers, and then the report you found for the rest :)

PR #3974, opened 2022-04-16
fix: Language selector overlapping

dougfabris:
It fixes the language selector overlapping when labels are displayed and some UX improves

### before
Screen Shot 2022-04-16 at 11 00 08

### after
Screen Shot 2022-04-16 at 10 59 34


Kongkille (commented):
This seems like good changes, are you able to submit a screenshot of the mobile version as well?

dougfabris:
@Kongkille There are no major changes as you can see in the new image I attached in the description. 
I can't see why we're fixing the language selector occupying the entire screen. With these new changes, we could have a result like the screenshot below. Let me know what you think!

Kongkille (approved):
Nice, this looks and feels much better :)

dougfabris:
@Kongkille Fixed 😃

PR #3975, opened 2022-04-16
chore: Flag missing space between header on country tooltip

dougfabris:
It improves the country's tooltip header by adding a margin between the flag and the country title. Also adds missing alt on the flag ``.
### before
Screen Shot 2022-04-16 at 11 10 53

### after
Screen Shot 2022-04-16 at 11 10 04


madsnedergaard (approved):
Awesome! 🙌

PR #3976, opened 2022-04-16
Enable dependabot to keep dependencies up to date

VIKTORVAV99:
This PR enables Dependabot to keep dependencies updated beyond just security updates.

I made this pr because I noticed a lot of the dependencies in the web subfolder are out of date and this will allow Dependabot to maintain most of them by creating PRs automatically on every Monday.

If you have pre-defined reviewers for the different products (mobile, web, parsers, mockserver, etc) I suggest you add them with the following code:
```yml
# Specify reviewers for pull requests

version: 2
updates:
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "daily"
    # Add reviewers
    reviewers:
      - "octocat"
      - "my-username"
      - "my-org/python-team"
```

All the different options can be found [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file).

**NOTE:** If this PR gets merged Dependabot may create a lot of PR:s the following weeks as it's trying to catch up but once it has the PR rate should be significantly lower.


madsnedergaard:
I'm a bit scared of this being something we won't be able to maintain and test - but maybe we can change schedule to monthly so it's less often we have to test everything? :)

VIKTORVAV99:
@madsnedergaard yeah sure we can totally do that, we could even limit it to minor and patch versions if that help?

madsnedergaard:
> @madsnedergaard yeah sure we can totally do that, we could even limit it to minor and patch versions if that help?

I assume we've pinned all dependencies to minor and patch, but would be nice to also ensure dependabot doesn't do major updates :)

VIKTORVAV99:
@madsnedergaard switched it to monthly (except github actions) and ignoring semver:major updates.
I also took the liberty to increase the max open pull requests for `/web` and `/mobileapp` npm dependencies to 10 so we don't get in a position where dependabot can't keep up do to the pull request limits.

madsnedergaard (approved):
Awesome, let's try it! 🙈

PR #3977, opened 2022-04-16
feat(parser): enable GT historical data retrieval

kruschk:
- Implement historical data retrieval for Guatemala (GT).
- Fix the broken parser by switching to a different API endpoint which
  comes from the same source (AMM), provides the same data, and is more
  user-friendly.
- Fix an off-by-one error in the reported timestamps. The API specifies
  hours using 1-indexing in the range [1, 24], so each reported hour
  refers to the past hour (e.g., hour 1 refers to the time period
  [00:00, 01:00)). This was not reflected in the prior implementation,
  which reported hour 1 as the time period [01:00, 02:00).

Refs: #3339


madsnedergaard (approved):
Hurray, amazing! 🚀

kruschk:
@madsnedergaard Thanks! :blush:

PR #3982, opened 2022-04-19
Switch `node-sass` to `sass` in `/web`

VIKTORVAV99:
Simply bumps `node-sass` to the latest version so we don't have to have python2 installed to test the frontend. (and get all the advantages between v4.13.1 and v7.0.1)


madsnedergaard:
This is nice and a good catch, but I actually wanted to switch from `node-sass` to `sass` everywhere, as that is the latest, official Dart-based compiler - which works better and faster in my experience :)
How would you feel about switching to that instead?

madsnedergaard (approved):
👏

madsnedergaard:
> @madsnedergaard I was hoping we could switch to the new [sass-embedded](https://sass-lang.com/blog/embedded-sass-is-live) package soon, it's should be closer in performance to node-sass but I think the change would require us to upgrade to webpack 5 and when I tried to do that a little quick it opened up a whole bunch of problems...
> 
> There is nothing preventing us from switching to just sass in the meantime tho except build performance as far as I can tell.

Switching would be cool, but I think we got bigger fish to fry (in terms of what's taking time during builds) 😄

VIKTORVAV99:
@madsnedergaard yes we do!
Next on my to do list is replacing all the lodash functions with native ES6 functions.

Lodash is currently one of the biggest dependencies we have and we don't really need it as we use babel.

madsnedergaard:
You are on fire, huge thanks from the entire team! 👏👏👏

VIKTORVAV99:
@madsnedergaard thanks for the kind words, they are very much appreciated!

PR #3987, opened 2022-04-20
update remaining renewable capacities (IRENA)

KabelWlan:
It was not super necessary but I updated the remaining renewable capacities with the new IRENA report and I also updated the sources.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -5523,7 +5523,7 @@
       "hydro": 16334,
       "nuclear": 6871,
       "oil": 2251,
-      "solar": 1577,
+      "solar": 1586.32,
       "wind": 12200
     },
     "contributors": [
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
You should use the original source and value for the Sweden solar data as it's the official government source. (It also reports a higher installed value for solar)

KabelWlan:
I changed it back. Thanks for your feedback.

config/zones.json
```diff
@@ -33,8 +33,8 @@
     "capacity": {
       "biomass": 1,
       "hydro": 0,
-      "solar": 1883,
-      "wind": 1
+      "solar": 2705,
+      "wind": 0
```
madsnedergaard, line 37: Oh no, what happened to the [one wind turbine](https://u.ae/en/information-and-services/environment-and-energy/water-and-energy/types-of-energy-sources/wind-energy)? 😢

madsnedergaard (approved):
Nice, thanks for doing this! 💪

PR #3994, opened 2022-04-21
Made the labeler config a little more robust and added a few labels

VIKTORVAV99:


.github/labeler.yml
```diff
@@ -1,13 +1,31 @@
...
 translations 🗣:
   - web/public/locales/**/*
 
+tests:
+  - tests/**/*
+  - parsers/test/**/*
+
+docker:
+  - docker-compose.yml
+  - web/Dockerfile
```
madsnedergaard, line 27: These files are rarely updated and not worth it's own label - maybe put them under infrastructure? :)

VIKTORVAV99: I thought it made sense split it out as it's another "language" as python and js but I can do that if you prefer.

madsnedergaard: I was thinking it would be nice to clean up our labels and I think it's nicer to group things by "functionality" than language :)

VIKTORVAV99: That makes sense. I'll push the changes soon.

It appears that dependabot already created a python label though, sorry about that. 😅

madsnedergaard: Haha no worries 😄

madsnedergaard (approved):

madsnedergaard:
Thanks again @VIKTORVAV99! 🙌

PR #3996, opened 2022-04-21
Switch Material Icon font to SVG sprites

VIKTORVAV99:
This PR removes the Material Icons webfont and replaces it with a SVG sprite for the icons we use.
This have several advantages such as lower overall size, less CLS and better scalability.
Everything should match the current layout and style but there might be some left over css (scss) now since the font related code is no longer needed.


VIKTORVAV99:
BTW is the back button on the country detail panel supposed to be blue? I replicated it as it is blue currently but it makes more sense to fade it like we currently do in the mobile nav bar.

madsnedergaard (commented):
Thanks for doing this, it's great! 
Don't think the blue back arrow makes sense, it's probably just a left-over thing.

Hmm, I think this sprite approach makes good sense, but I'm a bit unsure how we would go about updating this when adding new icons... Could you maybe write a very small guide on how to update the sprite? It could either live in a README somewhere or next to the code (maybe in the suggested Icon component?)

VIKTORVAV99:
@madsnedergaard changes made! 

I didn't add any verification of the icon names as I noticed the component rendered a lot more often than I initially thought and I didn't want to slow anything down by just checking the names are valid.

If it turns out we need it in the future it can easily be added.

madsnedergaard (approved):
Hurray, let's merge it! 🙌

Issue #2675, opened 2020-09-17
Additional data for Peninsular Malaysia - WIP I'm still filling this out


q--:
I found some useful data at https://www.singlebuyer.com.my/

>Single Buyer is the entity authorised by the Minister pursuant to the Electricity Supply Act (ESA) 1990 to conduct electricity planning and manage electricity procurement services for Peninsular Malaysia. Single Buyer plays a key role in promoting efficiency and facilitating competition in the generation sector, hence enhancing the development of the Malaysia Electricity Supply Industry (MESI). [source](https://www.singlebuyer.com.my/about.php?id=1)

# At the website [home page](https://www.singlebuyer.com.my/index.php)

- Current demand (we already have production)
- Forecast demand (appears broken. ***we don't have this yet)
- Installed capacity as of early this year (current source was last updated 2016. Sadly, no breakdown by type so not very useful)
- Forecast System Marginal Price (***don't have this yet***)

And some things not really useful for electricitymap: today's peak demand,  peak demand ever (or this year), and the same figures for energy consumption (MWh)

Kongkille:
Thanks for the overview, this data would be quite useful for any modeling we'd want to do.  Right now, I believe the most concrete action item would be to add the price forecast in the parser. I am uncertain about the forecasted demand if we use a separate source than the current source of production.

In that case, I might rename the issue "Price data for Peninsular Malaysia" - what do you think?

q--:
Personally, I prefer my titles as descriptive as possible. If implementing only a single data source is desired, I'd suggest either splitting the issue in multiple separate issues or adding a list with checkboxes for things that have/haven't been implemented.

Kongkille:
Makes sense, I am just trying to make sure that each issue we have is directly actionable. We have quite a lot of issues and sometimes it can be difficult to understand what exactly to do on all our issues, especially for newcomers.

And so for this issue, my point is that right now the most actionable thing would be to implement fetch_price with this source as I am not too sure we can use the other data types atm

PR #4001, opened 2022-04-23
Updated Malaysia Capacities

nessie2013:
The data is taken from the same source, however instead of just taking the number off the top (from 2016) I added each registered generation plant to get the capacities. Some of the gas plants are supposed to of been shut down already, however since they are still listed and nothing is saying they have actually been closed down I added them into the total


Kongkille (approved):
Sounds good. We'll see if a new source appears with any other updates on the coal plants

q--:
For reference, #2675 has a link to a map with various generating sites. The data is a bit difficult to extract so I don't know if it's better than your source

PR #4008, opened 2022-04-25
Added Japan (Kyūshū) Capacities

nessie2013:
Found a database of power plants and added up the capacities. All thermal plants were lumped under "thermal" and not seperated to coal, gas and such so the "thermal" total has been put in unknown for now


madsnedergaard (approved):
Great! 🙌

PR #4009, opened 2022-04-25
Makes the "no data available" work again

### LINKED ISSUE #3988: "No data source available" no longer shows
All zones without real-time data say that the data is temporarily unavailable instead of missing completely

image

This might be a result of #3934

madsnedergaard:
Fixes issue where left panel would always show "data temporarily unavailable" no matter what.

Closes #3988

We now have all 3 states back:
Screenshot 2022-04-25 at 20 13 07
Screenshot 2022-04-25 at 20 13 17
Screenshot 2022-04-25 at 20 13 27


Kongkille (approved):
Nice!

PR #4031, opened 2022-04-25
Updates /web and /mobileapp dependencies at once

madsnedergaard:
Dont want to wait for CI to run on 20 different PRs and manually update-master and merge every single one of them... So doing a few bulk updates instead!

### DIFF
--- a/mockserver/yarn.lock
+++ b/mockserver/yarn.lock
@@ -15,26 +15,36 @@ array-flatten@1.1.1:
   resolved "https://registry.yarnpkg.com/array-flatten/-/array-flatten-1.1.1.tgz#9a5f699051b1e7073328f2a008968b64ea2955d2"
   integrity sha1-ml9pkFGx5wczKPKgCJaLZOopVdI=
 
-body-parser@1.19.2:
-  version "1.19.2"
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
@madsnedergaard for the record dependabot will automatically close PR that aren't relevant anymore so you don't have to close them all one by one. Merging this would have closed them all. :)

Kongkille (approved):

madsnedergaard:
> @madsnedergaard for the record dependabot will automatically close PR that aren't relevant anymore so you don't have to close them all one by one. Merging this would have closed them all. :)

Ah cool, thanks :D

VIKTORVAV99:
@madsnedergaard @Kongkille I just noticed something, this only updates the lockfiles and not the package.json as well, was this intended?

madsnedergaard:
> @madsnedergaard @Kongkille I just noticed something, this only updates the lockfiles and not the package.json as well, was this intended?

Yup, there's no easy way to update package.json dependencies with yarn :/ 
But as far as I can tell it's also not that important - as long as we're using semver range specifiers (tilde and caret), then yarn will only look at the lockfile.

VIKTORVAV99:
You are right it don't seem very important, I expected yarn to overwrite the lockfile when running `yarn install` (like npm) but it didn't so it looks like this will work just fine.

PR #4010, opened 2022-04-25
Changes CircleCI to use newer Ubuntu image

madsnedergaard:
We currently use Ubuntu v14 and it's deprecated and will soon stop working in CircleCI.

Screenshot 2022-04-25 at 20 44 40

This upgrades the machine to a newer version (v20) :)

PR #4036, opened 2022-04-26
Update babel and necessary configurations

VIKTORVAV99:
This PR updates Babel to the latest versions and update the configuration to match.

I also removed some unused old dependencies and switched a return statement outside a function to process.exit(1) so ESLint would pass.

(Second attempt, the first one got so many merge conflicts it was easier to just make a new PR)


VIKTORVAV99:
hmm looks like the CI is stuck, CircleCI is saying the image is unavailable, is this perhaps related to #4010?

VIKTORVAV99:
I'm aware that there is a conflict with the yarn lockfile but I will wait to resolve it until it has been reviewed to avoid further conflicts with other PRs that target the lockfile.

madsnedergaard:
Seems solid, we're you able to run `yarn develop` locally? I'm seeing this error, but I've also had the same problem with `prop-types` before (which was fixed by adding it to dependencies

Screenshot 2022-04-28 at 23 02 10
)

VIKTORVAV99:
@madsnedergaard hmm, I can't remember I got a issue like that but I mostly used the docker method to test it as it changes the webpack build process. (only used by the docker setup)

Might just be a issue with the snowpack cache?

VIKTORVAV99:
It's funny you mentioned PropTypes...
[10:35:57] [snowpack] Error: ENOENT: no such file or directory, open 'C:\...\electricitymap-contrib\web\node_modules\.cache\snowpack\build\prop-types@15.8.1\prop-types\checkPropTypes.js'

But I am also getting errors for Redux as well as CoreJS.

All I know it works with webpack and docker so maybe it is just a snowpack problem? I am getting a lot of warnings for "Unscanable package import found" with snowpack that seems to indicate a deeper issue?

VIKTORVAV99:
I went down a rabbit hole trying to figure this out and I found this discussion: https://github.com/FredKSchott/snowpack/discussions/3883

Looks like snowpack isn't being actively developed anymore?

VIKTORVAV99:
@madsnedergaard I found a solution to the snowpack babel problem, if we rename all the React files from `filename.js` to `filename.jsx` we can remove the babel plugin from the Snowpack config all together as Snowpack understands jsx files. (using jsx syntax in js files throws an error).

I think this is the best way to solve this as it directly bypasses all the corejs polyfills and should make Snowpack even faster, without affecting any styles or functionality. This means it would be completely safe for local development with the downside is that you would have to use a browser that support all the modern syntax in the code but I think this is more or less a non-issue.

If you want to go down this path I can add a separate PR with the changes as they don't directly relate to these changes. (In either case we should rename the react files to jsx for clarity.)

madsnedergaard:
Thanks for the extensive research and for looking into this! 💚 
I was not aware that Snowpack is less actively maintained now, quite annoying after I already spent time switching like a month ago 😅 But I'm pretty sure most of the changes would have been required for Vite or similar anyway.

I specifically chose Snowpack over other tools as they didn't seem very mature yet. But maybe it's worth switching anyway?
I like your idea of renaming to `jsx`, I think that's a good approach - would be awesome if you can create a PR for doing it!

VIKTORVAV99:
A PR to rename the files to JSX will come soon. 👍🏻

I think it is fine to use snowpack for now, after #3837 is done we could even switch to webpack for local development using the same config as production. It won't be as fast as snowpack but it will be considerable faster than docker and it would match the production environment when you need to test stuff like bundling or minification.

VIKTORVAV99:
Well this is fun, yarn failed to resolve the merge conflict correctly and now `yarn.lock` is a mess...

Any suggestions on how to fix this?

VIKTORVAV99:
I just made it worse... 😅

VIKTORVAV99:
Well this PR is completely broken now...
@madsnedergaard what do you recommend?

I'd be happy to recreate it again on a fresh branch?

*I hate merge conflicts in lockfiles...*

VIKTORVAV99:
Sounds like a plan! 😅

I don't think cherry picking is necessary tho, it's only 5 files that have small changes (not counting the lock file so it will be quicker and simpler to just edit those directly and create a new commit.

PR #4037, opened 2022-04-26
Update CH emission factor

pierresegonne:
To roll out the estimation model for CH we need to update the emission factor of the unknown mode in CH. The estimation model indeeds splits out most of the unknown production into adequate production mode, changing the makeup of the remaining unknown production.

### DIFF
--- a/config/co2eq_parameters_direct.json
+++ b/config/co2eq_parameters_direct.json
@@ -1565,7 +1565,12 @@
             "source": "electricityMap, 2021 average",
             "value": 205.82553856340994
           }
-        ]
+        ],
+        "unknown": {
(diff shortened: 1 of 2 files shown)


gwpicard (approved):
💧 ⛰️ 🌱

jayaddison:
> can 47% of production mix in CH really be unknown?

A mistake in my understanding: the relevant spreadsheet cells are used to estimate _of the `unknown` data_ what percentage is thermal.  Even so, I question the figure.

(nitpick: it's also 48% rather than 47%)

pierresegonne:
Hey @jayaddison FYI we're currently having a fresh look at these emission factors - we're in contact with people at the Swiss Federal Office of Energy to improve them. Will give updates next week

jayaddison:
Great to hear; thank you!

PR #4038, opened 2022-04-26
fix(parser): improve GT production breakdown

kruschk:
Use a different endpoint from the Guatemalan GSO which provides a more
detailed production breakdown, generating better estimates of
Guatemala's emissions.

Refs: #3339


parsers/GT.py
```diff
@@ -57,19 +57,27 @@ def fetch_production(zone_key=DEFAULT_ZONE_KEY,
...
-                'gas':        row.get('TURBINA DE GAS'),
-                'hydro':      row.get('HIDROELÉCTRICA'),
-                'oil':        row.get('MOTOR RECIPROCANTE'),
-                'solar':      row.get('FOTOVOLTAICA'),
-                'wind':       row.get('EÓLICO'),
-                'geothermal': row.get('GEOTÉRMICA'),
+                'biomass': row['BIOGAS']
+                           + row['BIOMASA']
+                           + 0.5*row['BIOMASA/CARBÓN'],
+                'coal': 0.5*row['BIOMASA/CARBÓN']
```
kruschk, line 63: I'm not sure whether splitting the values here is appropriate or if they should instead be lumped into "unknown" and the ratios configured elsewhere.

EHHS15: Hello 
I believe that using the unknown category will cause more confusion about energy production and a better breakdown of biomass and carbon would be better.

VIKTORVAV99: they should be mapped to unknown and then you can add a co2eq override for unknown co2eq in co2eq_parameters_lifecycle.json with the average co2eq of all the unknown sourced based on their share of the unknown production. check out this [issue](https://github.com/electricitymap/electricitymap-contrib/issues/2407#issuecomment-620609844) for reference

EHHS15: I consider that it would be better to leave a division between both types, because if it is placed in unknown, it will cause more doubts on the subject and I consider that the sense of the map is to facilitate the knowledge of energy production, as the case of Bolivia is almost 50% of production is unknown, there is where people have that doubt that will be that production?

VIKTORVAV99: @EHHS15 I see where you are coming from but the parser should follow the same structure as the other parsers and report mixed production as unknown.
We also can't be sure that biomass/coal is split equally 50/50 now or in the future if they are reported together, this can cause a massive inaccuracy of the co2eq value generated for the region.

madsnedergaard: I very much agree with what @VIKTORVAV99 is saying here - as we have no way of knowing how it's split, it should go into unknown and the co2eq params be adjusted :)

kruschk: Makes sense! Any tips for estimating those CO2eq parameters? I was thinking I could look at, say, the last 5 years' average data from https://www.iea.org/countries/guatemala and use the relative proportions of biofuels, coal, and oil, since our "unknown" will be composed entirely of "BIOMASA/CARBÓN", "CARBÓN/PETCOKE", and "SYNGAN" (syngas?).

madsnedergaard: cc @pierresegonne :)

VIKTORVAV99: @kruschk I think the best and easiest would be if you had yearly production data from the same source as the API and then you could use that to subtract all the "pure" production from iea.org for example. That should leave you with a pretty accurate representation of the unknown share.

I know it is some work and can take some time but it's probably the best way to do it (it don't need to be 100% accurate IMO just as close as an approximation as we can get).

EHHS15: Before going any further I found this information on Biomass energy in 2021 where it explains that it constitutes 30 % of Guatemala's energy. I think it may be helpful. https://www.azucar.com.gt/2021/06/30/ingenios-aportaron-hasta-el-46-de-la-energia-electrica-consumida/

EHHS15: And contrasting with the data provided by the Comisión Nacional de Energía Electrica de Guatemala its proportion is not more than 30, it is always oscillating between 30 and 20 percent, so I think it is convenient to allocate 25% of its energy use from biomass/carbon and the rest to be added with coal.
https://www.cnee.gob.gt/wp/3296-2/

kruschk: @VIKTORVAV99 Do you mean integrating all the power production over one year (using the plot at https://www.amm.org.gt/portal), and then comparing that to the IEA data? I think that makes sense, but it looks like there isn't enough data available; I can only go as far back as May 2021. I would need one full calendar year to compare with the IEA's numbers (2020 at the latest).

@EHHS15 Looking at https://www.amm.org.gt/portal, "BIOGAS" and "BIOMASA" currently account for roughly 19 % of Guatemala's electricity production. This figure is missing some of the biomass in "BIOMASA/CARBÓN", but I'm sure that extra contribution would increase the percentage enough to be within the expected range. Since the AMM doesn't tell us precisely the ratio of biomass:coal in that figure, we have to lump it into "unknown", but we can do our best to estimate the composition of this overall "unknown" value so that the reported emission factors are as accurate and fair as possible.

pierresegonne: @kruschk I suggest you write a little scraping script to get data for [2021-05-01, 2022-05-01[. As you pointed out that won't give you the full calendar year to be able to compare, but you can either:

1. Take the sum of production over the year-to-date, i.e [2021-05-01, 2022-05-01[ (I'm not sure there is much inter-year variation for total production values)
2. Extrapolate the total production for [2021-05-01, 2022-01-01[ to [2021-01-01, 2022-01-01[

This way you'll have approximate total production values for each mode reported by the API, and will be able to compare them to what the IEA reports. 

With that knowledge you can then compute the makeup of what will be aggregated into unknown and compute an adequate emission factor for both lifecycle and direct with a weighted average of the relevant emission factors. See [here](https://docs.google.com/spreadsheets/d/16j5BVc4L34BDqVE0GeDBv0P3zQupT6zU77Kj5UK7GjE/edit?usp=sharing) for an example. 

Let me know if you need help or have any question :)

kruschk: I was wrong when I said that AMM historical data only goes as far back as May of 2021. The reality is actually worse: it turns out that data from more than a year ago becomes unavailable; it just disappears from the API. This isn't the end of the world, though, because we still have (roughly) a year of data available to us.

@pierresegonne I did as you suggested and came up with this [analysis](https://docs.google.com/spreadsheets/d/1CegROfej9HqRZTfihpjPpgZTYPUHrTNgTdHWbbP3w74/edit?usp=sharing). Could you please review my work? In particular, I had to compute absolute differences when comparing to the IEA data, or I'd end up with negatives. I'm not sure if that's right.

pierresegonne: Yeah indeed, they only release values for the last running year. Weird but we'll have to live with it.

The analysis looks great, gives a good idea of the mixup of unknown generation, and the associated emission factors. 

For the `powerOriginRatios` the goal is to provide a fallback for the average generation mix if no data is available. So to compute them, just take the values for the running year, and compute the respective contribution of each mode to the total production (+ unknown as the gap with IEA).

pierresegonne:
Looking great, thanks @kruschk for the implementation and @VIKTORVAV99 @EHHS15 for pitching in 👏

pierresegonne (approved):

Issue #4039, opened 2022-04-26
Can't click on regions on the map


VIKTORVAV99:
Looks like v0.255.0 introduced a bug that made countries un-clickable (no country detail panel show up).

It's probably smart to revert the code/dependency updates in `/web` until we know the cause.

Ping @madsnedergaard.

madsnedergaard:
Thanks @VIKTORVAV99! 😞 
I'm reverting the dependencies-update PR, super annoying that it's breaking things...

madsnedergaard:
(I guess I should also have tested this locally, assumed minor changes wouldn't be breaking 😬)

VIKTORVAV99:
Yep unfortunately that's how it is 🙃

I should have some time tomorrow to test a few dependancy upgrades but I will take some time to check them all.

VIKTORVAV99:
They shouldn't have been, but if some sub dependancy didn't follow semver or we where using something incorrectly that got fixed it can happen.

Can also just be a case of miss-matched dependencies where one got updated and the other was blocked for some reason.

madsnedergaard:
Release in progress, should be out within an hour! 
I also went through all the changed dependencies and found the cause (which seems obvious looking back): `react-map-gl` :)

madsnedergaard:
And it's back, thanks again @VIKTORVAV99 😍

PR #4041, opened 2022-04-26
Updates /web dependencies (without breaking change)

madsnedergaard:
Turns out #4031 included updating `react-map-gl` from `5.2.x` to `5.3.x` which made the zones non-clickable (#4039).
In this PR I'm upgrading all dependencies again, this time I have fixed `react-map-gl` to avoid this problem though :)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -33,7 +33,7 @@
     "lodash": "^4.17.21",
     "mapbox-gl": "^1.13.0",
     "moment": "^2.29.2",
-    "prop-types": "15.6.2",
+    "prop-types": "15.8.1",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-i18next": "^11.15.6",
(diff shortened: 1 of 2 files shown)


madsnedergaard:
cc @Kongkille, can't remember if this actually fixes the "Error: ENOENT: no such file or directory, ... prop-types" problem with running `yarn develop` :)

VIKTORVAV99:
> cc @Kongkille, can't remember if this actually fixes the "Error: ENOENT: no such file or directory, ... prop-types" problem with running `yarn develop` :)

I am getting it fore more that prop types but it's seemingly random, sometimes it can't find the react-router-dom sometimes it's prop-types. Usually it can be fixed by removing the snowpack cache and running yarn develop twice... 
I am starting to really look forward to removing the frontend server so we can use webpack locally too (could probably already be done if we used two different configurations).

VIKTORVAV99:
@madsnedergaard what do you want to do with this PR?
Should we try and update it or just close it?
(Or close it and create a new one?)

PR #4043, opened 2022-04-27
Reduce expected range for Slovenia down to 800MW

DavidKarlas:
With TEŠ 6(coal power plant) going offline for 12 days according to https://iip.remitor.eu/#/message/view/8780
Slovenia production can drop below 1GW.
~690MW coming from Nuclear, ~7MW Biomass, ~45MW Gas and Hydro can go down to ~50MW(currently with rain and snow melting its around ~100MW).
I think it will never drop under 800MW, 900MW might also work, but not always...

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -375,7 +375,7 @@ VALIDATIONS = {
     'SI': {
         # own total generation capacity is around 4 GW
         'required': ['nuclear'],
-        'expected_range': (1000, 5000),
+        'expected_range': (800, 5000),
     },
     'SK': {
         'required': ['nuclear']


DavidKarlas:
Validation is failing again: https://storage.googleapis.com/electricitymap-parser-logs/SI.html

madsnedergaard (approved):
Thanks for updating this! 💪

q--:
> `         'required': ['nuclear'],`

Does this need fixing as well? The country only has a single NPP which sometimes is unavailable due to refueling.

DavidKarlas:
I found that surprising as well, I just assumed it means that even if 0 is reported is passing the validation...

Issue #3938, opened 2022-04-05
[feature in development] Historical view 🗺️ 🔨


madsnedergaard:
<div align="center">
<img width="589" alt="Historical View" src="https://user-images.githubusercontent.com/3296643/161724357-4c085909-0816-407e-8cf0-76267ac7fd18.png">
</div>

### Description

**tl;dr: It's finally time to show historical data!**

You (and we) have been asking for it since the dawn of electricityMap, and now is the time to take the map to a whole new level by looking further into the past.

Thanks to all the amazing contributions over time(!) from so many amazing people(!), we have the data available and we have decided to make it publicly available in the name of the climate 🌱 

With this new feature users will be able to not only see the real-time/recent status of the grid, but also...
- deep-dive into the evolution of the grid
- uncover patterns
- compare zones and countries over time
- have more data-driven discussions

### Current status

We are currently shaping the feature, looking into required data structure changes and have started early design prototyping.

We hope you’ll help test this new feature and provide feedback when we have the early prototypes ready to be shared :)

steren:
My use case for historical view is to be able to determine, for a given location, the time of the day with the lowest carbon intensity based on a monthly or yearly average

madsnedergaard:
Oh hey, we forgot to share that the first prototype is actually live now: https://app.electricitymap.org/?feature=history 😄 
It's currently only showing hourly and only working on desktop, but any feedback is very welcome! 💚

VIKTORVAV99:
> Oh hey, we forgot to share that the first prototype is actually live now: https://app.electricitymap.org/?feature=history 😄 It's currently only showing hourly and only working on desktop, but any feedback is very welcome! 💚

Very cool! 🎉
Just one thing I spotted right away that can be problematic, all the exchange arrows are gone. 🙈

Kongkille:
> > Oh hey, we forgot to share that the first prototype is actually live now: https://app.electricitymap.org/?feature=history 😄 It's currently only showing hourly and only working on desktop, but any feedback is very welcome! 💚
> 
> Very cool! 🎉 Just one thing I spotted right away that can be problematic, all the exchange arrows are gone. 🙈

That's true! This initial version will be launched without exchange arrows, as we haven't quite gotten the data ready yet :D

VIKTORVAV99:
[...]

PR #4044, opened 2022-04-27
Adds mockserver data for v5

Kongkille:
Adds preliminary mock data for v5 enabling local development for #3938


VIKTORVAV99:
@Kongkille this isn't directly related to your change but I just had a question/suggestion related to these data structures if you don't mind.

I noticed that a lot of the keys both in production and for local development have null values. This is completely fine but I was thinking that it could be possible to omit these from the file/data itself and just assume all missing keys are null client side.

What do you think?

madsnedergaard:
I the biggest step of reducing the size of these files is to stop returning all zones with no parsers or data - it's a ton of empty objects (now x24) that we could easily remove.

It's a separate topic from this PR though, so I'll just put it on my backlog unless someone else is able to do it :)

VIKTORVAV99:
Sorry I hijacked the PR have created issue #4049 for further discussion if that is okay.

madsnedergaard:
I think we need to prioritise making these files smaller - I'll have a look at removing countries without data, but meanwhile I think it's okay to get this one merged in so we can start working on it :)
Screenshot 2022-04-28 at 10 37 49

madsnedergaard (approved):

PR #3957, opened 2022-04-11
GE: retrieve historical exchange data via ENTSOE

q--:
The GSE data source has real-time data, so you're not dealing with larger intervals like ENTSO-E. However, there is no historical exchange data from this data source. This pull request will obtain that data from the ENTSOE parser when a target datetime is specified.

Related: #3408

----
Not really tested other than checking for syntax errors, please test this before merging.

PR #4046, opened 2022-04-27
use ENTSOE for production history in the HOPS parser

VIKTORVAV99:
I included a check to make sure the requested date in older than 12 hours as the ENTSOE data for Croatia (HR) is lagging behind (12 hours should be safe) and is 0 filled for the rest of the day until it's updated.

Otherwise this should be a simple enough change.

_Note:_ I haven't tested this with live data, just a few logic checks, so please test it before merging.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -2699,7 +2699,8 @@
       "wind": 925
     },
     "contributors": [
-      "https://github.com/PaulCornelissen"
+      "https://github.com/PaulCornelissen",
+      "https://github.com/VIKTORVAV99"
     ],
     "parsers": {
       "consumption": "ENTSOE.fetch_consumption",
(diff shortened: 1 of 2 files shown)


Kongkille:
Hey @VIKTORVAV99,

Unfortunately we can't exactly do this as we don't want to mix data sources. The data is stored in our backend, and we use it in cases such as the historical-view, and for the commercial side of electricityMap. As such, when we mix the data sources it could taint the historical data of the zone. 

We have been looking a bit into some options where we can store multiple values for the same hour, but we'd also don't want to mislead people on the map, where some parts of the day is covered by 'x' source and another by 'y'

Hope that makes sense!

VIKTORVAV99:
I completely understand and it makes sense! 👍🏻

Does the same go for exchange data? If so you might want to revert the changes in #3957 as it was the inspiration for this PR.

Issue #4034, opened 2022-04-26
Update browserslist config and drop IE 11 support


VIKTORVAV99:
We should simplify and update this browserslist config and drop support for [IE 11 as it will soon no longer be supported by Microsoft for general users](https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support).
https://github.com/electricitymap/electricitymap-contrib/blob/e9c08be134ac78b3c05ac81744f7d164353a2acf/web/package.json#L118-L125

I suggest we change the config to this:
```json5
"browserslist": [ 
   "defaults", // defaults are: "> 0.5%, last 2 versions, Firefox ESR, not dead"
   "not ie 11",
   "not opera_mini all" // Not supported by Babel anyway because it lacks compatibility data.
 ] 
```
Old config targets:
```
and_chr 100
and_ff 99
and_qq 10.4
and_uc 12.12
android 100
baidu 7.12
bb 10
bb 7
chrome 100
chrome 99
chrome 98
chrome 97
edge 100
edge 99
edge 98
firefox 99
firefox 98
firefox 97
firefox 91
ie 11
ie 10
ie_mob 11
ie_mob 10
ios_saf 15.4
ios_saf 15.2-15.3
ios_saf 15.0-15.1
ios_saf 14.5-14.8
ios_saf 14.0-14.4
ios_saf 12.2-12.5
ios_saf 7.0-7.1
kaios 2.5
op_mini all
op_mob 64
op_mob 12.1
opera 83
opera 82
safari 15.4
safari 15.2-15.3
safari 14.1
samsung 16.0
samsung 15.0
```
_Due to order of the rules we actually target dead browsers such as IE 10 and bb_

And the new config targets:
```
and_chr 100
and_ff 99
and_qq 10.4
and_uc 12.12
android 100
chrome 100
chrome 99
chrome 98
chrome 97
edge 100
edge 99
edge 98
firefox 99
firefox 98
firefox 97
firefox 91
ios_saf 15.4
ios_saf 15.2-15.3
ios_saf 15.0-15.1
ios_saf 14.5-14.8
ios_saf 14.0-14.4
ios_saf 12.2-12.5
kaios 2.5
op_mob 64
opera 83
opera 82
safari 15.4
safari 15.2-15.3
safari 14.1
samsung 16.0
samsung 15.0
```
Together with #4036 this will lead to a substantially lower bundle size.

[...]

PR #4070, opened 2022-05-02
Remove momentjs

VIKTORVAV99:
This PR removes momentjs and replaces it with date-fns helper functions and native code.
This alone shaves almost 250kb of the vendor bundle (before gzip).

Most of the changes are pretty simple but I decided to use `Intl.DateTimeFormat` and `Intl.RelativeTimeFormat` to localize string, this is somewhat a breaking change as it no longer automatically pick a unit (hour, days, months, etc) and only use the specified unit (hour in this case). It should not be a problem but might look a bit weird to see 2000+ hours when doing local development. 😅 

~~### NO IE 11 SUPPORT!~~
~~As `Intl.RelativeTimeFormat` is not supported by IE 11 this means we would no longer support IE 11, I personally think this is the right way forward as Microsoft themselves will stop supporting it next month as mentioned in #4034.~~ With the merger of #4082 this is no longer relevant. 🙂


VIKTORVAV99:
In local development I got build failure due to babel being outdated and I think this is the same error so this should be fixed after #4036 is merged.

madsnedergaard (approved):
This is awesome! It's really nice to actually get localised date formats at last 😍 
Screenshot 2022-05-08 at 20 05 42
Screenshot 2022-05-08 at 20 05 35
Screenshot 2022-05-08 at 20 05 26

VIKTORVAV99:
> This is awesome! It's really nice to actually get localised date formats at last 😍 
> Screenshot 2022-05-08 at 20 05 42
> Screenshot 2022-05-08 at 20 05 35
> Screenshot 2022-05-08 at 20 05 26

Yeah I saw someone mentioning it in a old issue (can't find it now) and figured it would be a really nice benefit!

PR #4071, opened 2022-05-02
Rename react js files to jsx

VIKTORVAV99:
This pr is pretty simple, it renames the react js files to jsx as talked about in #4036, updates webpack so it can find the new files and removes the babel extension for snowpack as it's not needed for local development anymore.

It also includes a unrelated change, it adds a import for 'process' in store.js to resolve a console error and let the map render.


web/.eslintrc.js
```diff
@@ -75,7 +75,7 @@ module.exports = {
     'no-use-before-define': ['error', { variables: true, functions: false, classes: true }],
     'prefer-const': 'error',
     'prefer-template': 'error',
-    'react/jsx-filename-extension': ['error', { 'extensions': ['.js', '.jsx'] }],
+    'react/jsx-filename-extension': ['error'],
```
madsnedergaard, line 78: I updated this rule to ensure we enforce correct file extensions going forward :)

web/src/store.js
```diff
@@ -3,10 +3,11 @@ import { createStore, applyMiddleware } from 'redux';
 import { logger } from 'redux-logger';
 
 import reducer from './reducers';
+import { isProduction } from './helpers/environment';
 
 export const sagaMiddleware = createSagaMiddleware();
 
-export const store = process.env.NODE_ENV === 'production'
+export const store = isProduction()
   ? createStore(
```
madsnedergaard, line 11: I removed the `process` import and used an existing helper for this.

madsnedergaard (approved):
Awesome, I took the liberty of making a few changes directly :)

PR #4072, opened 2022-05-02
Disables dependabot for normal dependencies

madsnedergaard:
While the intention of #3976 was great, we just can't keep up with manually testing and merging all of these dependency updates.

I'm disabling it for now, but hopefully we can enable it again in the future when we have more automated tests and a faster CI pipeline :)

Screenshot 2022-05-02 at 14 25 30


VIKTORVAV99:
It sure is a bit more persistent than I expected. 😅

I'll start looking into what's needed for doing unit testing with jest as a starting point.

Kongkille (approved):

PR #4078, opened 2022-05-03
Update babel and configs

VIKTORVAV99:
Updates babel and the necessary configs.
See #4036 for reference.

*Attempt 3*

PR #4079, opened 2022-05-03
Remove url-search-params-polyfill (included in babel now)

VIKTORVAV99:
With the merger of #4078 this feature is already polyfilled by babel and corejs3 and we no longer need the separate polyfill.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -51,8 +51,7 @@
     "styled-components": "^5.0.1",
     "sunrise-sunset-js": "^2.2.1",
     "swr": "^1.0.1",
-    "topojson": "^3.0.2",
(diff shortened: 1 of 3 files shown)


madsnedergaard (approved):
Bye IE11 🥳 
I just looked at our analytics, and it's quite interesting how usage of IE11 totally dropped a few years back:

Screenshot 2022-05-03 at 13 56 17

VIKTORVAV99:
> Bye IE11 🥳 
> I just looked at our analytics, and it's quite interesting how usage of IE11 totally dropped a few years back:
> 
> Screenshot 2022-05-03 at 13 56 17
> 
> 

Does that mean we can drop support for IE11? 👀

That would make it considerably easier to use new features.

PR #4080, opened 2022-05-03
[historical-view] Changes colors on the map when timeslider is used

madsnedergaard:
### Description

_part of #3938_

This will make the colors on the map change for all countries where we have loaded historical data, whenever the time-slider is dragged.

For now this will only happen when a user has already clicked a zone, but that will change when we starting using `v5` state :)

### Preview
Kapture 2022-05-03 at 13 09 38


web/src/components/zonemap.jsx
```diff
@@ -107,22 +106,62 @@ const ZoneMap = ({
...
       ocean: { 'background-color': theme.oceanColor },
       zonesBorder: { 'line-color': theme.strokeColor, 'line-width': theme.strokeWidth },
-      zonesClickable: { 'fill-color': ['case', ['has', 'color'], ['get', 'color'], theme.clickableFill] },
+      zonesClickable: {
+        'fill-color': [
+          'coalesce', // // https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#coalesce
+          ['feature-state', 'color'],
+          ['get', 'color'],
+          theme.clickableFill,
+        ],
```
madsnedergaard, line 115: I wasn't aware that mapbox was so advanced, but the coalesce feature is perfect for this use case!
Here it's basically saying:
- use feature-state if available
- else use color property
- fallback to gray color from theme

web/src/components/zonemap.jsx
```diff
@@ -107,22 +106,62 @@ const ZoneMap = ({
...
+    }
+  }, []);
+
+  // TODO: Consider moving the calculation to a useMemo function
+  // change color of zones if timeslider is changed
+  useEffect(() => {
+    if (isHistoryFeatureEnabled && isLoaded && co2ColorScale) {
+      // TODO: This will only change RENDERED zones, so if you change the time in Europe and zoom out, go to US, it will not be updated!
+      // TODO: Consider using isdragging or similar to update this when new zones are rendered
+      const features = ref.current.queryRenderedFeatures();
```
Kongkille, line 136: Perhaps we should just initially start by querying all features here. I did so in my prototype with quite good performance.

Then we can optimize it through dragging if we'd like later on, I'm worried this will cause some odd experiences.

Kongkille (commented):
Nice :) 

I am having some issues testing this. Can you describe how to get the featureToggle working?

Also, I think we'd also need to update the tooltip for this feature to really work in itself. We can handle it in a seperate PR since this is behind a featureToggle if you'd prefer, but I think it ties into this quite well.

web/src/components/zonemap.jsx
Kongkille: Makes sense!

madsnedergaard:
Thanks for the review! You can test with `http://localhost:8080?feature=history` :)
And good point with the tooltips, I've created a new issue for it and will work on it in a separate PR.

web/src/components/zonemap.jsx
madsnedergaard: Hmm, in your prototype (on `mk/map-color-feature-states`) we also change only color of the zones that are rendered as far as I can tell :)

https://user-images.githubusercontent.com/3296643/166935649-98080222-b448-4b13-bc81-5898c32385a6.mp4

Kongkille (approved):
Just approving, but I think we could add isDragging to the useEffect and see if that doesn't just fix everything.

web/src/components/zonemap.jsx
madsnedergaard: Thanks for testing and good point, that definitely solves some of the problems - let's roll with it! :)

madsnedergaard:
Turns out fixing the tooltip was [very simple](https://github.com/electricitymap/electricitymap-contrib/pull/4080/commits/d17fd4cfa9f1f8771e97b8618c8b1c9be3955a5e), so I went ahead and did it in this PR too :)

PR #4081, opened 2022-05-03
Improve contributorlist.jsx and zones.json

VIKTORVAV99:
In this PR:
* removes the `https://github.com/` prefix from the contributors in `zone.json`.
* changes the `contributorlist.jsx` component to handle the new `zone.json` changes. 
* makes some improvements to `contributorlist.jsx` like lazy loading and adding srcset.

This will reduce the main bundle size and give users a better overall experience.


web/src/components/contributorlist.jsx
```diff
@@ -3,25 +3,29 @@ import { useParams } from 'react-router-dom';
...
+            href={`https://github.com/${contributor}`}
+            rel="noopener noreferrer"
+            target="_blank"
+          >
+            <img
+              src={`https://avatars.githubusercontent.com/${contributor}?s=20`} // loads the avatar image at a default size of 20px
+              srcSet={`https://avatars.githubusercontent.com/${contributor}?s=40 2x`} // loads the avatar image at a default size of 40px for high resolution displays
+              alt={contributor}
+              height="20px"
+              width="20px"
```
q--, line 24: Are you sure using `px` in these attributes is allowed?

VIKTORVAV99: Oh crap you are right, according to MDN they are supposed to be without units...
It works just fine with for me with units in Firefox and Chrome though but might be best to change it.

Issue #4000, opened 2022-04-22
Could we use ENTSOE Control Area production data for Sweden?


VIKTORVAV99:
I just noticed that the ENTSOE Control Area [production data for Sweden](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=false&viewType=TABLE&areaType=CTA&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=22.04.2022+00:00|CET|DAYTIMERANGE&dateTime.endDateTime=22.04.2022+00:00|CET|DAYTIMERANGE&area.values=CTY|10YSE-1--------K!CTA|10YSE-1--------K&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2)) includes Gas, Hydro, Nuclear, Other (Unknown), Solar and Wind.
This would be a improvement over the current SVK data we use that only reports Hydro, Nuclear, Heat (Unknown), Unclassified (Unknown) and Wind, it would also increase the accuracy of the unknown gCO2eq/kwh as it splits out gas and solar.

Now I don't have a ENTSOE API key or have had time to manually check the data at length but but a few spot check show it is under a 2% difference (take this with a huge grain of salt). Maybe someone else could do a more extensive comparison?

This would probably require some code changes to the ENTSOE parser and I am not familiar enough with the parser code and ENTSOE API to do it myself so I was hoping some of you could look into it if you have the time?

**NOTE:** This would only be temporary until #3936 gets unblocked and can be merged!

VIKTORVAV99:
Oh I just noticed there is a the ZONE_KEY_AGGEGATES in the ENTSOE parser? Maybe we can use that and aggregate all the SE subzones to the SE zone like for this Italy zone?
```python
ZONE_KEY_AGGREGATES = {
    'IT-SO': ['IT-CA', 'IT-SO'],
}
```

PR #3936, opened 2022-04-04
Split Sweden into subzones.

VIKTORVAV99:
This PR splits Sweden to the subzones SE-SE1, SE-SE2, SE-SE3 and SE-SE4 but should NOT be merge until #1397 is unblocked.

TODO:
- [x] Add subzone capacities. (Looks like the source includes small scale gasoline and diesel backups in gas but it's the only one I could find that splits it per bidding zone. We should try and find a better source in the future) 
- [x] Update exchanges.
- [x] Update parsers.
- [x] Change parsers in `zones.json` and add `subZoneNames` to `SE`.
- [x] Add gCO2eq/kWh for unknown sources. See #3915 for reference.
- [x] Update `world.geojson` with SE-SE1, SE-SE2, SE-SE3, SE-SE4 and remove SE. Help Needed!
- [x] Update translation files with new names for SE subzones.
- [x] Update tests.

I know there is already a PR that aims to do the same thing but it has become outdated and need a lot of edits to get in the right shape to be merged, this PR aims to fix that.

PR:s and commits to this PR are welcome!

Fixes: #1397, #2699

PR #4083, opened 2022-05-03
Change Sweden (SE) to use ENTSOE.fetch_production_aggregate parser for production

VIKTORVAV99:
As #3936 is still blocked I thought we could use the ENTSOE.fetch_production_aggregate for Swedens (SE) production data as it has a more detailed breakdown of the production sources. (It adds gas and solar)

**Note:** ENTSOE only have useful data AFTER 2021-12-14 (before that it's only wind)
 Now I don't know how the backend works when switching parsers so I'm not sure from what date when to add the new co2eq override for unknown production (remove gas and solar from the mix) so this will be a draft for now until I have that info. 🙂 
 
 closes: #4000

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -5573,7 +5573,7 @@
       "consumptionForecast": "ENTSOE.fetch_consumption_forecast",
       "generationForecast": "ENTSOE.fetch_generation_forecast",
       "price": "ENTSOE.fetch_price",
-      "production": "SE.fetch_production",
+      "production": "ENTSOE.fetch_production_aggregate",
       "productionPerModeForecast": "ENTSOE.fetch_wind_solar_forecasts",
       "productionPerUnit": "ENTSOE.fetch_production_per_units"
     },
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -146,6 +146,7 @@ ENTSOE_DOMAIN_MAPPINGS = {
     'UA': '10YUA-WEPS-----0',
     'XK': '10Y1001C--00100H'
 }
+
 # Generation per unit can only be obtained at EIC (Control Area) level
 ENTSOE_EIC_MAPPING = {
     'DK-DK1': '10Y1001A1001A796',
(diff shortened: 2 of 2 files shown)


pierresegonne:
Hey @VIKTORVAV99 for the co2eq overrides, if data comes in only after December 2021, I suggest that we keep the old one for years prior and just add it for 2022

so the format would be:


```
"unknown": [
  {
    "_comment": "XXX",
    "datetime": "2022-01-01",
    "source": "XXXX",
    "value": XXXXX
  }
]
```

On our side we interpret this as:

"this co2eq parameters only applies after 2022-01-01"

Let me know if you have additional questions.

VIKTORVAV99:
> Hey @VIKTORVAV99 for the co2eq overrides, if data comes in only after December 2021, I suggest that we keep the old one for years prior and just add it for 2022
> 
> so the format would be:
> 
> ```
> "unknown": [
>   {
>     "_comment": "XXX",
>     "datetime": "2022-01-01",
>     "source": "XXXX",
>     "value": XXXXX
>   }
> ]
> ```
> 
> On our side we interpret this as:
> 
> "this co2eq parameters only applies after 2022-01-01"
> 
> Let me know if you have additional questions.

I know that, I added the other co2eq overrides. 😅
I am just wondering if we switch the parser will the new parser be used from now or from say 2022-01-01?

pierresegonne:
> we talked about if non-renewable waste should be counted as biomass or not and we decided that when we switched to ENTSOE that it should for consistency.

Yes indeed, for consistency we should count it as biomass. 

>  I am just wondering if we switch the parser will the new parser be used from now or from say 2022-01-01?

No these co2eq parameters don't affect how the parser works. Basically there is only one true version of the parser at any given moment, which is the one on master. When refetching beyond 2021-12, the parser will then just return empty values.

pierresegonne:
Do you need to make any updates regarding biomass? Else it looks good to me :)

VIKTORVAV99:
> > we talked about if non-renewable waste should be counted as biomass or not and we decided that when we switched to ENTSOE that it should for consistency.
> 
> Yes indeed, for consistency we should count it as biomass.
> 
> > I am just wondering if we switch the parser will the new parser be used from now or from say 2022-01-01?
> 
> No these co2eq parameters don't affect how the parser works. Basically there is only one true version of the parser at any given moment, which is the one on master. When refetching beyond 2021-12, the parser will then just return empty values.

I was just a bit worried it might "taint" the existing data if it where to refetch before 2022-01-01 as it still contained wind data so it wouldn't be completely empty if I understand it correctly.

But if it is just used from 2022-01-01 this won't be a problem (might not have been one before either).

VIKTORVAV99:
> Do you need to make any updates regarding biomass? Else it looks good to me :)

All updates regarding biomass is already done so it's safe to merge this :)

VIKTORVAV99:
@pierresegonne could we try and get this merged soon?

I would love to see some solar generation for Sweden. 😉

pierresegonne (approved):

PR #4085, opened 2022-05-04
Add test case for non-empty annual co2eq zone overrides

pierresegonne:
## Issue

It seems like a zone with an empty list of annual overrides for emission factors can make our data processing fail

## Description

Add a test case to verify that none of these cases are present in the co2eq config files


FelixDQ (approved):
Sidenote, but maybe we should look into a way to enforce black on contrib. Ideally, we don't want to require contributors to install pre-commit, but maybe we can add a github action to format files before the PRs are merged. I wonder what the best practice is here.

pierresegonne:
Yeah good idea, it shouldn't be really hard as a github action. 
Can I let you create a slack week friendly issue for it?

Issue #3105, opened 2021-04-22
Support for `Ocean` Emission Factor


pierresegonne:
Some countries have experimented with tidal power and report it in their production data (e.g Korea as reported in these [capacities](http://epsis.kpx.or.kr/epsisnew/selectEkpoBftChart.do?menuId=020100) ) and the [IPCC 2014 report](https://archive.ipcc.ch/pdf/assessment-report/ar5/wg3/ipcc_wg3_ar5_annex-iii.pdf#page=7) that we use currently for the base of some of our emission factors have an estimate for `ocean` .

Based on that, we could add it to our supported production modes. Nevertheless considering the limited scale of tidal power use, I'd like others' opinion on whether it's even useful to include this production factor or not.

q--:
In the FO parser I added it as 'other' with the IPCC emission factor as an override.


Obviously that's only viable if no other electricity source is already using the `other` category

Kongkille:
When we address #738 we'll also look into how to handle new and other production modes

PR #4086, opened 2022-05-04
Filter out modes with 0 production and 0 capacity from the graph

VIKTORVAV99:
This PR filters out modes with 0 production AND 0 capacity from the graphs to reduce the amount of unnecessary whitespace they use.

This will allow us to add more (new) kinds of production modes like requested in #3105 without making the graphs look empty and it will make first glance comparisons easier.

(works for both consumption, production and emissions)

### Before:
image
### After:
image

**NOTE:** I have only tested this with East Denmark but it should not cause any problems with other regions. ~~Only thing I can think of that might cause issues are if a region has 0 capacity (not unknown) for a mode yet is still reporting production values that fluctuate between 0 and >0.~~

_PS: It's pretty cool we use SVG elements directly for the graph!_


madsnedergaard:
Hey @VIKTORVAV99, thanks for the PR here! I think this raises some larger UX questions that we need to discuss first:
- how does this affect the ability to compare countries? Is it harder to compare when things "shuffle" around between different countries?
- How will the "missing" fields be perceived by users, will they all understand that if nuclear is not on the list for a country, that means that there is no nuclear (or could it be seen as the amount is currently 0 or maybe even a "they don't know/have data for that production mode")?
- if we do this, should we go even further and rank production mix by amount currently being produced?

We'll take some time to discuss this internally next week, but we'd also love to hear if anyone else has thoughts on this :)

VIKTORVAV99:
> * how does this affect the ability to compare countries? Is it harder to compare when things "shuffle" around between different countries?
> * How will the "missing" fields be perceived by users, will they all understand that if nuclear is not on the list for a country, that means that there is no nuclear (or could it be seen as the amount is currently 0 or maybe even a "they don't know/have data for that production mode")?

@madsnedergaard Maybe a "show all" toggle/button would remove these concerns?
Please bring it up as a compromise/improvement for the internal meeting.

> * if we do this, should we go even further and rank production mix by amount currently being produced?

As for the last point I think that will be really disorienting if people use the time slider and the bars start "jumping" around. But we could consider sorting them after co2eq values instead of just using a arbitrary order?

corradio:
> As for the last point I think that will be really disorienting if people use the time slider and the bars start "jumping" around. But we could consider sorting them after co2eq values instead of just using a arbitrary order?

What you could do is do the same as exchanges: if a mode (or exchange) is missing *for all hours*, then you can remove it.
That way you won't have any jumping. I think we could extend this reasoning to a lot of the info on the left panel (e.g. the estimation disclaimer) to avoid any jumping.

VIKTORVAV99:
> > As for the last point I think that will be really disorienting if people use the time slider and the bars start "jumping" around. But we could consider sorting them after co2eq values instead of just using a arbitrary order?
> 
> What you could do is do the same as exchanges: if a mode (or exchange) is missing _for all hours_, then you can remove it. That way you won't have any jumping. I think we could extend this reasoning to a lot of the info on the left panel (e.g. the estimation disclaimer) to avoid any jumping.

Great idea!
I'll look into it but I have less free time now so might take some time before I have anything ready to merge.

VIKTORVAV99:
> What you could do is do the same as exchanges: if a mode (or exchange) is missing _for all hours_, then you can remove it. That way you won't have any jumping. I think we could extend this reasoning to a lot of the info on the left panel (e.g. the estimation disclaimer) to avoid any jumping.

Hi @corradio, I tried to look into what would be needed a little quick but I can't seem to find the code that filter the exchanges, could you just point me in the right direction and I'll see if I can make the changes the coming days.

corradio:
Hi @VIKTORVAV99 ,
Code-wise, the bar chart usage of `exchangeKeys` is here:
https://github.com/electricitymap/electricitymap-contrib/blob/master/web/src/components/countrytable.jsx#L428
The calculation of keys is done here:
https://github.com/electricitymap/electricitymap-contrib/blob/master/web/src/hooks/redux.js#L80

I'd probably double check with @Kongkille and @rasmusjeppsen however that this makes sense from a UI/UX perspective, as I just floated this idea around but haven't thought it through ;-)

Thanks for the help!

Olivier

VIKTORVAV99:
Thanks for the code locations!

I'll play around with it a little locally the coming days and see if I find a nice clean way to only filter out a mode if all datapoints for it are 0 while the capacity is also 0.

As for the exchanges I don't think we need to actually filter those as they will never have 0 capacity (they should just be removed or updated if they do).

VIKTORVAV99:
> 

I just realized that jumping here referred to if we where sorting the production modes by the amount of power they generate.

But your idea is still relevant and would be a good improvement over the current implementation in this PR.

Kongkille:
We tried to test it a bit but did not get any conclusive feedback either way (we tested a couple of different versions)

I do like the condensed view, however I think there are still some issues with this current iteration

1. If we begin filtering out non-relevant fuel sources we should make sure all remaining are valid 

Running this on production, there are still some remaining items

image

Here we have geothermal and battery storage even though both of them are 0 for all 24 hours.

2. Graph changes on aggregate change

(I know you made this before historical view so this was not part of the initial scope!)

DK-DK1 hourly:
image


DK-DK1 daily:

image

Here we see there are two different graph elements. Sometimes this will make sense, especially for the yearly view where something may have been deprecated, but otherwise I think it can be confusing.

cc @rasmusjeppsen if you have any more thoughts

VIKTORVAV99:
> We tried to test it a bit but did not get any conclusive feedback either way (we tested a couple of different versions)
> 
> I do like the condensed view, however I think there are still some issues with this current iteration
> 
> 1. If we begin filtering out non-relevant fuel sources we should make sure all remaining are valid
> 
> Running this on production, there are still some remaining items
> 
> Here we have geothermal and battery storage even though both of them are 0 for all 24 hours.

The reason for this is quite simple actually, I wanted to still show modes that have a installed capacity or missing capacity and in the case of Hungary it has a installed capacity of 3 MW for geothermal and the installed capacity for battery storage is missing.

This way we don't remove a production mode just because it has 0 production one day and should help with long term comparability.

> 2. Graph changes on aggregate change
> 
> (I know you made this before historical view so this was not part of the initial scope!)
> 
> Here we see there are two different graph elements. Sometimes this will make sense, especially for the yearly view where something may have been deprecated, but otherwise I think it can be confusing.

Ah must have missed this when historical view launched, since the historical values don't have any capacities they are all included. I'll fix this right away.

VIKTORVAV99:
@Kongkille updated the code to only take the capacity into account if there is a capacity object for the data.
This way if we add capacities to the historical view in the future the transition will be seamless and no code update will be necessary to get the same functionality as the hourly view.

VIKTORVAV99:
@madsnedergaard @Kongkille any new updates on this?
I'd love to see it merged and would be happy to explain some of the choices in more detail if needed. 🙂

madsnedergaard:
I'm going to close this PR for now, but we will revisit this topic again in the somewhat near future where we're going to rethink the left panel as a whole :)

PR #3985, opened 2022-04-20
Remove lodash

VIKTORVAV99:
Lodash is currently one of our biggest dependencies when it comes to size but we don't really need it anymore.
A lot of the Lodash functions we use now have native JS replacements and since we use babel to transpile and polyfill the code during build we can safely use newer features of JS without impacting browser compatibility.

PR #4087, opened 2022-05-05
[historical-view] Adds new timecontroller component

Kongkille:
TBD


web/src/hooks/redux.js
```diff
@@ -16,10 +16,23 @@ export function useCurrentZoneHistory() {
...
 
+export function useCurrentDatetimes() {
+  // TODO: should use V5 state here and v5 should tell the state datetimes
+  const histories = useSelector((state) => state.data.histories);
+  if (histories && Object.keys(histories).length) {
+    return histories[Object.keys(histories)[0]].map((h) => moment(h.stateDatetime).toDate());
+  } else {
+    return [];
+  }
+}
```
Kongkille, line 27: This function assumes all histories have the same datetimes. As the comment suggests, V5/state should keep track of the relevant datetimes instead of the individual zones

I don't believe we have any cases where it makes sense for some zones to have a different set of datetimes than another.

web/src/reducers/index.js
```diff
@@ -49,6 +49,7 @@ const initialApplicationState = {
   solarColorbarValue: null,
   webGLSupported: true,
   windColorbarValue: null,
+  selectedTimeAggregate: 'day',
```
Kongkille, line 52: Arguably this could be called 'selectedTimeRange' since the aggregate is 'hourly'. We could wait for another PR as we will touch this again soon.

web/src/components/graph/timeaxis.js
```diff
@@ -20,15 +36,24 @@ const getTicksValuesFromTimeScale = (scale, count) => {
   const startTime = scale.domain()[0].valueOf();
   const endTime = scale.domain()[1].valueOf();
 
-  const precision = moment.duration(15, 'minutes').valueOf();
+  const precision = moment.duration(60, 'minutes').valueOf();
```
Kongkille, line 39: This change should not affect anything, as we only have hourly values now

Kongkille:
> > @madsnedergaard marking this ready for review. I'll clear out the conflicts, I believe they are mostly related to lodash and moment
> 
> Perhaps it's time to get rid of both of them then! #4070 and #3985 will do that for you! 😉

Haha yes, you are right. I actually thought those were in! I'll get them merged next :)

web/public/locales/en.json
```diff
@@ -49,6 +49,13 @@
         "noParserInfo": "We do not yet have any information about this area.<br/> Want to help fix this? You can contribute by <a href=\"%s\" target=\"_blank\">adding area data</a>.",
         "now": "Now"
     },
+    "time-controller": {
+        "title": "Display data from the last",
+        "day": "Day",
+        "month": "Month",
+        "year": "Year",
+        "5year": "5 years"
```
VIKTORVAV99, line 57: Consider using `Intl.DisplayNames()` and Intl.NumberFormat() for these instead then we don't need to include them in the translation files.

Code needed:
```js
new Intl.DisplayNames(document.documentElement.lang, {type: 'dateTimeField'}).of('day')
```
and:
```js
new Intl.NumberFormat(document.documentElement.lang,  { style: 'unit', unit: 'year', unitDisplay: 'long' }).format(5);
```
[MDN | Intl.DisplayNames() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/DisplayNames#using_type_datetimefield)
[MDN | Intl.NumberFormat() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat)

web/src/components/graph/timeaxis.js
madsnedergaard: FYI: This has changed in #4070

web/src/layout/timeController.js
```diff
@@ -0,0 +1,86 @@
...
+  background: white;
+  left: 10px;
+  bottom: 10px;
+  border-radius: 15px;
+  text-align: center;
+  overflow-y: visible;
+  width: calc((14vw + 16rem) - 70px); // Ensures it is smaller than countrypanel
+  z-index: 999; // TODO: find the needed one
+  position: fixed;
+  box-shadow: rgba(0, 0, 0, 0.66) 0px 10px 36px 0px, rgba(0, 0, 0, 0.06) 0px 0px 0px 1px;
```
madsnedergaard, line 36: It's a bit intense box-shadow, maybe reuse the one from `<Legend />`?

web/src/reducers/index.js
madsnedergaard: Yeah, I think this is fine for now and then we'll think more about it later on

web/public/locales/en.json
```diff
@@ -49,6 +49,13 @@
         "noParserInfo": "We do not yet have any information about this area.<br/> Want to help fix this? You can contribute by <a href=\"%s\" target=\"_blank\">adding area data</a>.",
         "now": "Now"
     },
+    "time-controller": {
+        "title": "Display data from the last",
```
madsnedergaard, line 53:
```suggestion
        "title": "Display data from the past",
```

web/public/locales/en.json
madsnedergaard: I agree, that would be a nice approach :)

madsnedergaard (commented):
Overall this is super cool 😍 
Some comments and ideas, but I think it works pretty well!

web/src/components/graph/timeaxis.js
```diff
@@ -4,13 +4,29 @@ import { range } from 'lodash';
...
 
-// If the tick represents a timestamp not more than 15 minutes in the past,
-// render it as "Now", otherwise render as localized time, i.e. "8:30 PM".
-const renderTickValue = (v, __) => (
-  moment().diff(moment(v), 'minutes') <= 15
-    ? __('country-panel.now')
-    : moment(v).format('LT')
-);
+const TOTAL_TICK_COUNT = 25; // total number of ticks to be displayed
+const TICK_VALUE_FREQUENCY = 6; // Frequency at which values are displayed for a tick
```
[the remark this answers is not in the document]
rasmusjeppsen: Yes, +1! Love this change @madsnedergaard 🎉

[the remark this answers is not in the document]
Kongkille: I've done that now!

We can also reconsider when looking into some other timeranges where there are even more ticks. We might need to filter some of them off :)

web/public/locales/en.json
VIKTORVAV99: @Kongkille that is why I used `document.documentElement.lang` we already set the document language with i18n in `i18n.js`.
https://github.com/electricitymap/electricitymap-contrib/blob/346d621810809ce2f7359212fb092ebeedb2a052/web/src/helpers/i18n.js#L80-L81

I have used `document.documentElement.lang` in other places but we could change those to match, either way it would be good to decide how we want to do it so the codebase is consistent.

Kongkille: Well if I understand correctly, I don't think we want to rely on a document attribute in all places.

In my PR i remove this line so we no longer use this hook. This means that (because TimeAxis is a memo component) we don't actually re-render the component onLanguageChange.
https://github.com/electricitymap/electricitymap-contrib/blob/346d621810809ce2f7359212fb092ebeedb2a052/web/src/components/graph/timeaxis.jsx#L31

Essentially this means that language specific changes are not going through unless the component is guaranteed to rerender. Right now it works on master because they all somehow use the hook, but it can lead to some odd behavior.

As for the Memo I think you're right, I am not sure what the cost of it is but unless it's really expensive we should probably avoid it, if not for readability

Kongkille: You can see here that the TimeAxis actually don't re-render on this PR


https://user-images.githubusercontent.com/45588799/167462682-1837d652-fa3d-4ea9-9920-4d515dec03b3.mov

So I think that we should use the "useTranslation" hook to get the language as well (or create a new hook specifically to get the languageCode perhaps)

VIKTORVAV99: As we aren't passing this global object to the function/component but rather using it directly don't this make memo unnecessary? (I haven't used react much so please correct me if I'm wrong).

Originally I figured it would be best to use a global object over function calls/imports as they would have lower performance overhead but it probably don't matter much when we only do a few calls each time. 

However if we do need memo I think it does make sense to create a languageCode hook for clarity and consistency.

Kongkille: I think certainly some of our Memo components/functions are not necessary but right now we have them in the codebase, so my point is moreso that we risk the components not rerendering.

It seems to be that the most 'react' approach would be to have a languageCode hook so we absolutely guarantee that the component rerenders when language changes.

In the future we could perhaps look into some of these memos to determine how useful they are :)

web/src/components/timeslider_new.jsx
```diff
@@ -0,0 +1,173 @@
+import React, { useMemo, useState } from 'react';
+import { first, last, sortedIndex, isNumber } from 'lodash';
```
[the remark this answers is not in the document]
Kongkille: Excellent, just updated :)

Kongkille:
Thanks for the great review @VIKTORVAV99 and @madsnedergaard 🥇 

I think I've addressed your points now. As for the language items, I've just used the hook for now and then we can maybe look into more in another issue or PR?

VIKTORVAV99:
It's great to see these features coming along and I'm happy to help anyway I can!

As for the discussion about hooks, yeah they aren't apart of this PR really and I think it make sense to address it at a later point.

VIKTORVAV99:
looks like the build is failing because webpack has no loader configured for svg, I can add that in a separate PR if want (best option IMO) otherwise I think you need to change how you reference the svg.

web/src/components/timeslider_new.jsx
```diff
@@ -0,0 +1,173 @@
...
+    background: #ffffff;
+    cursor: pointer;
+    margin-top: -8px;
+    transition: box-shadow 0.4s;
+    box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.15);
+    border: 1px solid lightgray;
+    background-position: center center;
+    background-size: 12px;
+    background-repeat: no-repeat;
+    background-image: url(${thumbImg});
```
madsnedergaard, line 44: Does it fail if we import them like we do in the stylesheets?
Screenshot 2022-05-10 at 12 11 12

If yes, then @VIKTORVAV99's suggestion is probably best - I guess it will in future be nice to able to use SVGs anyway :)

web/src/components/timeControls.js
```diff
@@ -63,15 +64,36 @@ const DateOptionWrapper = styled.div`
...
 `;
 
-const options = [
-  { key: 'day', label: 'time-controller.day' },
-  { key: 'month', label: 'time-controller.month' },
-  { key: 'year', label: 'time-controller.year' },
-  { key: '5year', label: 'time-controller.5year' },
-];
+const getOptions = (language) =>
+  useMemo(
```
madsnedergaard: I don't think it makes sense to useMemo here - in general it should only be used for heavy calculations, as the performance overhead of using the functionality is otherwise higher than not using it :)

web/src/layout/leftpanel/zonelistpanel.js
```diff
@@ -42,10 +42,11 @@ const StyledWrapper = styled.div`
 
 const ZoneListPanel = () => {
   const { __ } = useTranslation();
-  const features = useFeatureToggle();
+  const isHistoryFeatureEnabled = useFeatureToggle('history');
+  console.log(isHistoryFeatureEnabled);
```
madsnedergaard: Console.log :)

web/src/components/graph/timeaxis.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import { useTranslation } from 'react-i18next';
```
madsnedergaard, line 2: Please use our own hook instead (sorry that the naming is identical, I couldn't come up with a better name 😞)
```
```
It's not critical for this use case, but it'll be annoying if we later want to use __ and it's not working as expected (with sprintf) :)

web/src/components/timeslider_new.jsx
VIKTORVAV99: No it won't (I think), the difference here is that webpack is trying to import the contents of the SVG file (and inline it) and use that while in the stylesheet we just do "use this file" basically.

web/src/components/timeControls.js
madsnedergaard: Nevermind, you already fixed this!

web/src/components/graph/timeaxis.jsx
VIKTORVAV99: Maybe we should just use this?
https://react.i18next.com/latest/translation-render-prop

web/src/components/timeslider_new.jsx
Kongkille: Just in the spirit of finalizing this PR I've used the .scss file. We will have to look at this very soon again, as we'll need to remove the entire file.

Then we can handle the SVG loader in a separate PR

Kongkille:
Seems build is passing now :)

web/src/components/timeslider_new.jsx
VIKTORVAV99: The scss files are getting removed? 👀

Last I checked the postcss don't support styled components so we would lose autoprefixer and quite a few optimizations doing that?

web/src/components/graph/timeaxis.jsx
madsnedergaard: Hmm, I personally feel like the pattern of render props is a bit out-of-date - I'd rather just either find a better name for our custom hook or add an eslint rule that ensures we only import the right one :)

web/src/components/graph/timeaxis.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useTranslation } from 'react-i18next';
+import { useTranslation } from '../../helpers/translation';
```
madsnedergaard, line 2: This is also a thing in all the other places - sorry, this is the last thing before merging! 😬

web/src/components/graph/timeaxis.jsx
VIKTORVAV99: Maybe but it would come with benefits such as inbuilt suspense for when the language file is loading and would ge guaranteed to re-render on language change.

Might be worth considering at least.

madsnedergaard (approved):
🚀

PR #4088, opened 2022-05-05
Add overrides for isLowCarbon and isRenewable contributions for CH

pierresegonne:
## Description

Adds zoneOverrides for the considering the contribution of `unknown`to the renewable and carbon-free percentages for CH.
The rationale is detailed in the source provided, but the TLDR; is that 52% of the unknown is hydro discharge, for a zone where more than 80% of the production is carbon free.


corradio (approved):
Great stuff! Maybe it's worth rounding the values to two decimals so we avoid having a file that is so large (and a diff that is quite big)

pierresegonne:
@FelixDQ let's have a look together before merging, I'd like to discuss data format for these parameters (TLDR they don't respect the same format as the other parameters)

FelixDQ (commented):
I would probably also update the other values, so that we only use numbers and not a combination of booleans and numbers.

Also remember that we need to update the code here when we merge this:
https://github.com/electricitymap/electricitymap/blob/master/libs/flowtracing/electricitymap/flowtracing/flowtracing.py#L326-L334

and here
https://github.com/electricitymap/electricitymap/blob/7c11216a4cb241b5f9ff483900381c3eab6450fe/libs/flowtracing/electricitymap/flowtracing/flowtracing.py#L375-L383

FelixDQ (approved):

tests/test_co2eq_parameters.py
```diff
@@ -138,6 +160,80 @@ def callback2(ratio, _zone):
             )
         self.check_power_origin_ratios(callback1, callback2)
 
+    def check_contribution_object(self, contribution, zone, mode, contribution_name):
```
pierresegonne, line 162: testing for two things mainly:
* datetimes - if the contributions for a mode are given as a list with datetimes, we verify that each element of the list have a valid datetime + that the list is ordered by these datetimes
* values for the contributions - check that all contributions have a key "value" and that the value is between 0 and 1

PR #4089, opened 2022-05-05
Webpack 5 migration prep (update css loaders and webpack config)

VIKTORVAV99:
This PR is the start of a long process to get us ready to upgrade to webpack 5.

It:
 * updates all the css loaders to the highest version that still support webpack 4. 
 * switches the use of autoprefixer to postcss-env (it includes autoprefixer among other plugins)
 * removes the custom function to output `mainfest.json` as it's not supported by webpack 5
 * adds the BundleAnalyzerPlugin to output the `manifest.json` file again
 * clean up the `webpack-config.js` file by removing options that won't be supported in webpack 5 or changing them to supported versions.


VIKTORVAV99:
Turns out there won't be a long process!

I have a stable version that use Webpack 5 locally and can push it here or add a separate PR after this is one merged.

Let me know what you think is best!

madsnedergaard:
Lets do a separate PR to make testing easier :)

madsnedergaard (approved):
Tested and AFAIK it works just fine - so let's try it! 🙃

VIKTORVAV99:
Hehe dev dependencies are the "safe" ones. 

It's when we need to update the production dependencies I start to worry. 😅

PR #4094, opened 2022-05-07
Webpack config duplicate bundle fix

VIKTORVAV99:
We accidentally created duplicate bundles of the CSS styles, this PR fixes that.


madsnedergaard (approved):
Good catch!

PR #4103, opened 2022-05-08
Zone overrides for SE unknown production

VIKTORVAV99:
This PR adds a zone override for SE unknown data with the sum of renewable unknown production (biomass and solar).


pierresegonne:
Nice to see that people are already adopting this new parameter format :) 

Just a nitpick, could you add the year you used for computing this in the comment? It does not seem to change very much year-to-year, but would make it easy to find that info again in the future

VIKTORVAV99:
> Nice to see that people are already adopting this new parameter format :)
> 
> Just a nitpick, could you add the year you used for computing this in the comment? It does not seem to change very much year-to-year, but would make it easy to find that info again in the future

Sure thing, in the comment I suppose or do these have a datetime property like co2eq overrides?

pierresegonne:
It's actually supported but considering the low amount of yearly variation I wouldn't impose you to compute each year ahah

VIKTORVAV99:
> It's actually supported but considering the low amount of yearly variation I wouldn't impose you to compute each year ahah

I actually already have the numbers from when I did the co2eq overrides so that would be trivial and I could easily add it if you want.

pierresegonne:
Up to you :)

VIKTORVAV99:
Then I'll add them! 🙂
Just have a errand to run first but they should be added in a 1-2 hours.

VIKTORVAV99:
@pierresegonne Should be all done now, I hope I got the structure right!

pierresegonne (approved):
👏 well done!

PR #4105, opened 2022-05-08
Webpack 5 update

### LINKED ISSUE #4090: Make sure nothing depends on anything lower than @babel/runtime@7.12.0 for Webpack 5
- [ ] Update react-router-dom and history to v6+ and v5+ respectively.
- [ ] Update react-map-gl (Unknown version at this point)
- [ ] Update redux-sage (Unknown version at this point)
- [ ] Update react-transission-group (Unknown version at this point)

VIKTORVAV99:
This is the main PR for Webpack 5.

I reinstalled all dependencies that used anything lower than @babel/runtime@7.12.0. 

Updated Webpack to v5 and Webpack-Cli to v4.

Changed `optimize-css-assets-webpack-plugin` to `css-minimizer-webpack-plugin` as per the [optimize-css-assets-webpack-plugin readme](https://github.com/NMFR/optimize-css-assets-webpack-plugin#optimize-css-assets-webpack-plugin)

Added the filesystem cache to the config.

Closes #4090


web/package.json
```diff
@@ -38,7 +38,7 @@
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-i18next": "^11.15.6",
-    "react-map-gl": "^5.2.11",
+    "react-map-gl": "~5.2.11",
```
VIKTORVAV99, line 41: Pinned this one to patches as it has caused problems in the past.

web/package.json
```diff
@@ -26,7 +26,7 @@
     "date-fns": "^2.28.0",
     "ejs": "^2.5.7",
     "express": "^4.16.2",
-    "history": "^4.9.0",
+    "history": "^4.10.1",
```
VIKTORVAV99, line 29: The lockfile already used this version.

web/webpack.config.js
```diff
@@ -83,4 +87,7 @@ module.exports = {
     path: `${__dirname}/public/dist`,
     pathinfo: false,
   },
+  cache: {
+    type: 'filesystem',
+  }
```
VIKTORVAV99, line 91: We probably need to change the docker config to cache this file.
Default location is `node_modules/.cache/webpack` but we can change that if needed.

madsnedergaard:
`yarn develop` isn't working for me :)

web/webpack.config.js
madsnedergaard: Hmm, I'm unsure what consequences this have - can we start simple to get this PR in (as you suggested)? :)

VIKTORVAV99: The only consequences it should have is that the first run (no cachefile) will be slower but it will speed up subsequent builds.

Since we use docker things get a little tricker though so I'll remove it from this PR and then we can evaluate it at a later date, perhaps after we remove the nodejs frontend server; After that it would be very simple to cache this file for the CI and for local development it wouldn't need any caching at all.

VIKTORVAV99:
@madsnedergaard updated and disabled the cache (should match webpack 4) so this should be ready to merge. 🙂

web/webpack.config.js
madsnedergaard: I think that makes good sense, let's start small :D

Issue #3949, opened 2022-04-07
Missing data panel for regions with data


VIKTORVAV99:
There seems to be a bug where several regions that have valid data show the missing data panel.

![image](https://user-images.githubusercontent.com/30777521/162268928-3d08d215-e77e-4b0f-ad86-c9c498bf1d4e.png)
_Screenshot taken at 18:05 UTC_

Some of the affected regions are:

- Southeast Norway
- Middle Norway
- North Norway
- West Norway
- Sweden
- Poland
- Great Britan
- Central South Italy
- South Italy

**Note:** While writing this issue the regions recovered at about 18:13 UTC.

brandongalbraith:
@VIKTORVAV99 Try quitting and reloading the browser tab or mobile app? As of this comment, the data shows as available for me in both a browser and the mobile iOS app:

![image](https://user-images.githubusercontent.com/350893/162270235-c19e11ff-cd71-41cb-a236-8740ade30e9b.png)

VIKTORVAV99:
> @VIKTORVAV99 Try quitting and reloading the browser tab or mobile app? As of this comment, the data shows as available for me in both a browser and the mobile iOS app:
> 
> ![image](https://user-images.githubusercontent.com/350893/162270235-c19e11ff-cd71-41cb-a236-8740ade30e9b.png)
> 

It recovered on its on at roughly 18:13 UTC but I still thought it was worth filing a issue.
(Last line in my issue mentioned this)

madsnedergaard:
cc @Kongkille, could this be due to the change in #3934? 🤔 
I can't reproduce it, so please do keep us updated if this happens again for anyone!

VIKTORVAV99:
@madsnedergaard might I ask when you tried to reproduce it? (exact time)

I can reproduce it but only just after the time changes to a new hour (8:59 to 9:00 for example).

It also seems to affect regions at random.

Kongkille:
Hmm this could very well be a bug by #3934. 

Essentially that PR decouples the state file (used to show the global map) and the history file (the details)

During testing I did not experience these issues, I am wondering whether this could be an issue with our caching logic. I'll investigate in the app-backend.

Gazer75:
What region might be missing seem quite random. A reload can fix one region, but then another is not showing.
Clearing cache/cookie or using Ctrl+F5 doesn't seem to matter much.

madsnedergaard:
[...]

PR #4107, opened 2022-05-09
Always load latest history data in countryPanel

Kongkille:
Attempt at fixing #3949 

It seems that we have fixed the caching issues, but there are still some issues which seems to be related to the app not realizing that there is newer historical data available.

This PR ensures that we initialize with using the latest zoneTimeIndex always which may fix the issue.

### DIFF
--- a/web/src/hooks/redux.js
+++ b/web/src/hooks/redux.js
@@ -53,12 +53,7 @@ export function useCurrentZoneData() {
     if (!zoneId || !grid || !zoneHistory) {
       return null;
     }
-    if (zoneTimeIndex === null) {
-      // Return latest history data unless latest history does not correspond to the current grid datetime
-      return (zoneHistory[zoneHistory.length - 1].stateDatetime === grid.datetime
-        ? zoneHistory[zoneHistory.length - 1]
-        : null);
-    }
+
     return zoneHistory[zoneTimeIndex];
   }, [zoneId, zoneHistory, zoneTimeIndex, grid]);
 }
--- a/web/src/reducers/index.js
+++ b/web/src/reducers/index.js
@@ -45,7 +45,7 @@ const initialApplicationState = {
   },
   onboardingSeen: cookieGetBool('onboardingSeen', false),
   searchQuery: null,
-  selectedZoneTimeIndex: null,
+  selectedZoneTimeIndex: 24,
   solarColorbarValue: null,
   webGLSupported: true,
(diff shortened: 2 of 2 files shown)


web/src/reducers/index.js
```diff
@@ -45,7 +45,7 @@ const initialApplicationState = {
   },
   onboardingSeen: cookieGetBool('onboardingSeen', false),
   searchQuery: null,
-  selectedZoneTimeIndex: null,
+  selectedZoneTimeIndex: 24,
```
madsnedergaard, line 48: Should we revert this change as discussed IRL?

web/src/hooks/redux.js
```diff
@@ -53,12 +53,7 @@ export function useCurrentZoneData() {
     if (!zoneId || !grid || !zoneHistory) {
       return null;
     }
-    if (zoneTimeIndex === null) {
-      // Return latest history data unless latest history does not correspond to the current grid datetime
-      return (zoneHistory[zoneHistory.length - 1].stateDatetime === grid.datetime
-        ? zoneHistory[zoneHistory.length - 1]
-        : null);
-    }
+
```
madsnedergaard, line 56: What will happen if if `zoneTimeIndex` is `null` now?

An alternative could be return the latest one and ignoring that it might be out of date (as that will be handles with historical view soon anyway)
```suggestion
if (zoneTimeIndex === null) {
      return zoneHistory.at(-1);
    }
```

madsnedergaard, line 56: What will happen if if `zoneTimeIndex` is `null` now?

An alternative could be return the latest one and ignoring that it might be out of date (as that will be handles with historical view soon anyway)
```suggestion
if (zoneTimeIndex === null) {
      return zoneHistory.at(-1);
    }
```

madsnedergaard (approved):

madsnedergaard:
While looking at it again now (while switching to new data format), I realise that we have quite a problem here in what data we're expecting to be available. Previously this hook returned data from `state.grid.zones[zoneId]` if no time-index is selected, but now we always return from `state.histories` where different properties are available.

TypeScript would have been amazing so we actually now what data is available where 😅 
I'll try to reorganise the redux state, just wanted to flag that things are semi-broken in weird places right now because components expects different data from this hook :)

Kongkille:
I am not sure I fully understand, happy to discuss IRL?

PR #4108, opened 2022-05-09
Fix app wrongfully saying app is updated.

### LINKED ISSUE #3995: App wrongfully says a new version is available
The app says a new version is available, but nothing happens when I press "here".

image

It seems like the file was cached on disk and might thus cause some differences:

image

clientVersion.json right now: b4cee58186df3a5bb9903a40520a1d6243d8ea34
Not sure how to get the version running in the browser.

Kongkille:
Fixes #3995

- Adds button to force hide outdated version
- Remove cache on client-version.json

### DIFF
--- a/web/src/layout/main.jsx
+++ b/web/src/layout/main.jsx
@@ -2,7 +2,7 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
 // TODO(olc): re-enable this rule
 
-import React from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import { connect, useDispatch, useSelector } from 'react-redux';
 import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


web/src/layout/main.jsx
```diff
@@ -134,10 +148,11 @@ const Main = ({
               .
             </div>
           </div>
-          <div id="new-version" className={`flash-message ${isClientVersionOutdated ? 'active' : ''}`}>
-            <div className="inner">
+          <div id="new-version" className={`flash-message ${isClientVersionOutdated && (!isClientVersionForceHidden) ? 'active' : ''}`}>
```
madsnedergaard, line 151:
```suggestion
          <div id="new-version" className={`flash-message ${isClientVersionOutdated && !isClientVersionForceHidden ? 'active' : ''}`}>
```

madsnedergaard (approved):
nitty nitty picky picky 🐱 
But approving nonetheless and thanks for fixing it!

Kongkille:
cc @corradio as I know you've spotted this a few times, lmk if it gives you any trouble now!

PR #4112, opened 2022-05-10
Moves to using localStorage instead of cookies

madsnedergaard:
Doing this means we don't need to have an annoying banner as these are the only cookies we're using :)


q--:
>we don't need to have an annoying banner 

I don't think it works that way. Despite the nickname, the 'cookie law' covers any kind of storage, not just cookies
 ([See here for example](https://www.cookielaw.org/the-cookie-law/).)

madsnedergaard:
> Directives set out common goals that EU member states must achieve. Member states must devise their own national laws on how to reach those goals.

I guess it really depends on how each member state have decided to implement the ePrivacy Directive and thus it's super hard to figure out a general approach :/

In the Danish version (and [maybe UK too?](https://law.stackexchange.com/questions/30739/do-the-gdpr-and-cookie-law-regulations-apply-to-localstorage)) at least, it states that "functional" cookies does not require consent (as far as I can tell) - although it's very arguable what is defined as functional or not (e.g. the onboarding modal) :)

PR #4114, opened 2022-05-10
Remove units from height and width in `contributorlist.jsx`

VIKTORVAV99:
These values should not have units.

Thank you @q-- for pointing it out!


sin (approved):
LGTM

madsnedergaard (approved):
Thanks for the fix and thanks for the review 🥳

PR #4117, opened 2022-05-11
Update exchange capacities for Nord Pool zones

VIKTORVAV99:
Added missing capacity data for:
-   DE->DK-DK1
-   EE->LV
-   LT->LV

I also modified all existing capacities to match the Nord Pool data.

Source: https://www.nordpoolgroup.com/en/Market-data1/Dayahead/Capacities1/Capacities/KEY/Norway/?view=table

_Note: This also adds the proper source for the capacities I updated a while back but forgot to include in DATA_SOURCES.md_ 😅


config/exchanges.json
```diff
@@ -972,8 +976,8 @@
   },
   "DK-DK1->NO-NO2": {
     "capacity": [
-      -1700,
-      1700
+      -715,
```
KabelWlan, line 979: I think this value is too low. The exchange values were above 1000 MW in both directions today.

VIKTORVAV99: Nice catch!
The map I was using is a bit messy so probably just picked the wrong number...
image

I'll update it right away.

PR #4118, opened 2022-05-12
[historical-view] Disable date range options and change text to 24 hours

Kongkille:
Two things in one:

Since we only support hourly values still, I'm disabling the other options. A

Edit: woops put in the wrong video, here is the right one:

https://user-images.githubusercontent.com/45588799/168052377-7a0011e3-9bc6-410f-bdc0-2345563d0cd2.mov

Also based on our initial user feedback, it seems "24 hours" is a better text to display so we're trying that.

### DIFF
--- a/web/src/components/timeControls.jsx
+++ b/web/src/components/timeControls.jsx
@@ -67,7 +67,11 @@ const DateOptionWrapper = styled.div`
 const getOptions = (language) => [
   {
     key: 'day',
-    label: new Intl.DisplayNames(language, { type: 'dateTimeField' }).of('day'),
+    label: new Intl.NumberFormat(language, {
+      style: 'unit',
+      unit: 'hour',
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Should the others also be changed to match?

As in:
Month to 1 Month?
And Year to 1 Year?

Kongkille:
> Should the others also be changed to match?
> 
> As in: Month to 1 Month? And Year to 1 Year?

No I don't think we should do that. Specifically the issue with the "Day" label was that users thought it was referring to the last day, eg yesterday, whereas it really is a rolling 24 hour window.

However for month and year it would actually be the last month, eg for now it would show April and not a rolling 30 day window

VIKTORVAV99:
Ah, I assumed the month and year was rolling too. Since it's not the current labels make sense. 👍🏻

PR #4121, opened 2022-05-12
Adds moment back to fix broken wind and solar layers

### LINKED ISSUE #4120: ElectricityMap hangs while loading when wind layer is enabled
Tried on desktop and mobile Firefox and desktop, as well as desktop Chrome, both with private browsing and without.

Tried URLs with  the parser down (https://app.electricitymap.org/zone/PR?wind=true, https://app.electricitymap.org/zone/NL?wind=true) and without (https://app.electricitymap.org/zone/GE?wind=true)

Firefox stays blank, Chrome shows the 'page is not responding' dialog after a while, and Firefox mobile shows the 'no area data' text after a while (for countries that definitely have data).

Turning on my (work) VPN doesn't change anything either.

The title on the browser tab shows up, though.

madsnedergaard:
The change away from moment apparently broke layers - when enabling them the browser crashes hard, so there's something not great going on... Instead of debugging further I'm reverting the specific changes from #4070 to make it work again :)

Fixes #4120

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -35,6 +35,7 @@
     "lodash.mapvalues": "^4.6.0",
     "lodash.sortedindex": "^4.1.0",
     "mapbox-gl": "^1.13.0",
+    "moment": "^2.29.3",
     "prop-types": "15.6.2",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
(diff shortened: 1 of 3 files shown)


VIKTORVAV99:
Well this sucks...
Would there be any way to get solar and wind mockdata so we can test this locally first in the future?

Issue #3807, opened 2022-02-18
"Estimated" note under energy sources can cause flickering


jarek:
1. Go to a zone that has estimated data right now and real data in recent past. We're looking for the "Some data for this region is estimated. Read more about our estimation methodology here." info box.
2. Hover over the 24-hours chart below the fuel source list
3. If you do it right, the display will keep on flickering between most recent data (with "estimated" box) and past data (without  "estimated" box) because the "estimated" box pushes the 24-hours charts down, which makes the chart gain focus, which makes the data shown to set to past when we hove over a past hour, which makes the "estimated" box disappear, which makes the chart lose focus again, which makes the data reset to current, which makes the "estimated" box appear again, and so on.

I've tried to record this happening and attached at the end of the report, hopefully it works.

Browser: Firefox 97.0 on Linux.

![est](https://user-images.githubusercontent.com/47415/154705624-849d7f6a-1761-4dc2-bbe1-e11bb9871a1f.png)

https://user-images.githubusercontent.com/47415/154705256-03f6829e-e352-4eab-a9bb-e73554b5f504.mp4

atjohans:
The cause seems to have been when switching between hovering over data that is estimated, and data that is not estimated. Estimated data caused the new div to appear with the "Some of the data for this region is estimated..." message. which caused everything to shift downward, making it no longer hovered-over, which caused it to move back up which created an endless cycle causing the flickering.

I have a simple solution in place that simply creates a blank "buffer" region that prevents everything from moving when hovering over estimated data. This fixes the flickering.

I could also envision a solution where the system checks if any of the data for the country is estimated and just displays the message if any of the data is estimated, and doesn't display the message if none of the data is estimated.
I am wondering if there is a preferred solution to this issue? (edited) 




Original issue: 

https://user-images.githubusercontent.com/70723718/164061751-4d0539d0-a2be-4370-a5f0-71e63430382b.mp4


The current solution: 



https://user-images.githubusercontent.com/70723718/164061778-04b8b829-1873-435c-90a9-f5448c82bc3d.mp4





I am currently partial to the "buffer" solution, as I have some experience with web dev, and in my experience users tend to prefer when things in the window move around as little as possible

madsnedergaard:
cc @rasmusjeppsen, what are your thoughts on this? And will it still be an issue with historical view? :)

VIKTORVAV99:
[...]

PR #4122, opened 2022-05-12
[POC] New info layout on country panel.

VIKTORVAV99:
This is a proof of concept of a new layout for text information on the country panel.

Copied from issue #3807:
Rough draft when the data is from a API:
Expanded (default):
image
Collapsed:
image

Rough draft of when it's estimated:
Expanded (default):
image
Collapsed:
image

@rasmusjeppsen and @madsnedergaard what do you guys think?

The icons is just placeholders for now until I find more appropriate ones and the same goes for the text but I feel like with this approach we can better explain what a API means vs what estimation means.

Maybe the expanded/collapsed functionality is a bit over the top?

Do we want to split up API and parser explanation/labels?

### Alternative:
We could split these up and put the estimated label in a expandable box under the gauges like this:
image
<sub>_Paint mockup_</sub>

With this option we could also remove the estimated label from here:
image

### Implementation
In both cases the box would be expanded by default and we can store the state in `localStorage` so if users collapse it we can show them their selected state when they visit the page again.

### DIFF
--- a/web/public/images/material-icon-sprite.svg
+++ b/web/public/images/material-icon-sprite.svg
@@ -1,65 +1,19 @@
 <svg>
-  <symbol xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 24 24"  viewBox="0 0 24 24" id="view_list">
-    <rect fill="none" height="24" width="24"></rect>
-    <path d="M3,14h4v-4H3V14z M3,19h4v-4H3V19z M3,9h4V5H3V9z M8,14h13v-4H8V14z M8,19h13v-4H8V19z M8,5v4h13V5H8z"></path>
-  </symbol>
-  <symbol xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" id="info">
-    <path d="M0 0h24v24H0z" fill="none"></path>
-    <path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"></path>
-  </symbol>
-  <symbol xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" id="arrow_right">
-    <path d="M0 0h24v24H0V0z" fill="none"></path>
-    <path d="M10 17l5-5-5-5v10z"></path>
-  </symbol>
-  <symbol xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" id="arrow_left">
-    <path d="M24 0v24H0V0h24z" fill="none" opacity=".87"></path>
-    <path d="M14 7l-5 5 5 5V7z"></path>
-  </symbol>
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
@madsnedergaard @rasmusjeppsen what do you guys think?

This is far from done but should I move forward with this or do you guys want to go in another direction?

Any input would be appreciated!

rasmusjeppsen:
@VIKTORVAV99 Thank you for exploring this option and providing a draft! It is super helpful to see. I like the thoughts you've put into this. After seeing this draft, I believe simply showing the text at the bottom of the side panel once data is estimated, is a more simple solution for users. 

Also, since the data source is either from the "API" or "estimated", I think just showing the label when the data is estimated (like we currently do) is a simpler solution.

Regarding the expanded/collapsed functionality, this is also interesting. Here I think simply displaying the "expanded content" when the data is estimated, is a simpler solution for users. In that way, we make the choice of expanding the text for the user, and if we bring the text box to the bottom of the side panel, I believe it won't disturb the rest of the content.

Thank you again for all your contributions, Viktor. You have been praised here at the electricityMap HQ lately!

This is the design mockup from https://github.com/electricitymap/electricitymap-contrib/issues/3807, where I moved the text box to the bottom of the side panel, that I refer to:
image

VIKTORVAV99:
@rasmusjeppsen thanks for the feedback!
Just want to explain a few of the choices before going forward.

> Also, since the data source is either from the "API" or "estimated", I think just showing the label when the data is estimated (like we currently do) is a simpler solution.

The reason I added the API label (and API text) is so we can keep a static div that isn't empty and don't disappear when the data isn't estimated preventing content to "jump" around. Putting it on the bottom eliminates most of it but there is still some content below it like the source text and contributor icons that would be affected.

> Regarding the expanded/collapsed functionality, this is also interesting. Here I think simply displaying the "expanded content" when the data is estimated, is a simpler solution for users.

Hehe this was more for myself and others that use electricityMap very often, it would be nice to hide/collapse that content as we already know what it says and the estimated label (and API label) makes it easy to tell where the data is coming from anyway.

> Thank you again for all your contributions, Viktor. You have been praised here at the electricityMap HQ lately!

Glad to hear it! ❤️
I recently got a more free time for a little while and figured I should put it to use towards a worthy project!

rasmusjeppsen:
Thank you for the elaboration, @VIKTORVAV99!

Interesting insights and thoughts, and thank you again for the experimentation you did. I think moving forward with the simple version where the text box is moved to the bottom of the side panel is the right move for us:

image

If you feel for it, you are very welcome to give it a go if you want! However, it is by no means expected.

Your contributions are much appreciated!

VIKTORVAV99:
@rasmusjeppsen I'd be happy to create a new PR to just move the info box to the bottom, it would be simple enough.

I would however like to request you and the team consider the other changes for the future (they are by no means a priority) as I think it would be a good way to explain what it means when the data comes from a parser and be a little cleaner UI wise IMO.

PR #4123, opened 2022-05-12
Fix log messages in layers.js

VIKTORVAV99:
This _should_ fix the log messages in `layers.js`.

@madsnedergaard looks like I missed these when I removed moment, this should fix it but I can't test it locally due to the lack of solar and wind layer mockdata.


VIKTORVAV99:
Looks like there is a problem with circleci again, perhaps someone can trigger another CI run?

madsnedergaard (approved):

madsnedergaard:
Thanks for fixing it! 💚

PR #4131, opened 2022-05-16
Try template for GH action

pierresegonne:


madsnedergaard:
(I like the idea here! Changed to draft to avoid notifications 😄)

pierresegonne:
@madsnedergaard FYI because the PR from contributors originate from forks, the github-actions bot does not have the authorisation to push a cleaning commit to their branch. 

Tell me if you know a way around, else what I propose:

- have an action that creates a cleaning PR, if cleaning is needed, everytime a new commit appears on master

wdyt?

madsnedergaard:
> @madsnedergaard FYI because the PR from contributors originate from forks, the github-actions bot does not have the authorisation to push a cleaning commit to their branch.
> 
> Tell me if you know a way around, else what I propose:
> 
> * have an action that creates a cleaning PR, if cleaning is needed, everytime a new commit appears on master
> 
> wdyt?

Isn't there some way we can ensure people automatically run formatting before committing code? I don't like the idea of creating a new PR each afterwards, as that will increase the workload of managing this repository :)

pierresegonne:
That's a possible alternative - but that means requiring contributors to install pre-commits, which I'm not sure is a super nice experience for them. 

To limit the workload for us, we could enable the bot to directly push the formatting to master, but I don't really like that idea.

Else we could do this periodically, automatically reformat the code every week for example .

My preference for now is the idea of automatically creating a cleaning PR when needed and making it easy for contributors to setup autoformatting so that it's rarely necessary to create such a PR.

madsnedergaard:
Would it make sense to have the action check if formatting is correct and then fail if it's not (maybe with a message/guide on how to run format locally)? We could maybe make it easy by having a `poetry run format` command

pierresegonne:
Alright let's try it as you suggested @madsnedergaard and I guess if people don't understand / are really not happy with having to run a CLI command they'll let us know

.github/workflows/python-format.yml
```diff
@@ -2,7 +2,7 @@ name: autoblack
 on: [pull_request]
 
 jobs:
-  autoblack:
+  setup:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
```
VIKTORVAV99, line 8: You need to check out the code and run setup-python in each job, they can't run in a setup job as they are all running in their own containers.

PR #4128, opened 2022-05-16
Adds prettier packages and updates eslint dependencies

madsnedergaard:
### Description

We want to improve linting and make the codebase more standardized.

This PR adds new packages for linting - we'll lint the existing codebase in a separate PR afterwards :)

Steps:
1. Add config (this PR)
2. Lint everything
3. Setup husky and lint-staged for keeping things linted going forward

PR #4132, opened 2022-05-16
Makes linting run before code is committed

madsnedergaard:
### Description

We want to improve linting and make the codebase more standardized.

This PR makes linting and formatting happen automatically before code is committed.

Steps:

- Add config (#4128)
- Lint everything (#4134 )
- Setup husky and lint-staged for keeping things linted going forward (this PR)


VIKTORVAV99:
Consider running prettier on the `/.circleci`, `/.github` and `/config` directories/folders as well.

Probably won't do much on the first two but ruining it on the last would be nice so the config files all have the same formatting throughout. (mostly affecting zones.json and exchanges.json)

madsnedergaard:
Oh yes, that's a great point! Since those are outside of `web` and are often edited through the browser, maybe running it in a GH action or CI like you suggested would be ideal? :)

VIKTORVAV99:
Sounds like we have a plan then!
I'll create a PR after the others are merged.

madsnedergaard:
> Sounds like we have a plan then! I'll create a PR after the others are merged.

Awesome!
We are actually already running [a "lint JSON" step in CircleCI](https://app.circleci.com/pipelines/github/electricitymap/electricitymap-contrib/4438/workflows/0d505c5c-58ea-4845-b5d6-75584d75d0ad/jobs/24675):
```
sudo npm install -g jsonlint-mod
jsonlint -q config/*.json web/public/locales/*.json
```

VIKTORVAV99:
> > Sounds like we have a plan then! I'll create a PR after the others are merged.
> 
> Awesome! We are actually already running [a "lint JSON" step in CircleCI](https://app.circleci.com/pipelines/github/electricitymap/electricitymap-contrib/4438/workflows/0d505c5c-58ea-4845-b5d6-75584d75d0ad/jobs/24675):
> 
> ```
> sudo npm install -g jsonlint-mod
> jsonlint -q config/*.json web/public/locales/*.json
> ```

This would differ somewhat though, prettier formatting can fail even if linting passes.
An example would be:
```json
"capacity": [
      -533,
      533
    ]
```
and
```json
"capacity": [-533, 533]
```
both will pass linting but only the second one will pass prettier.

Now just adding prettier to check if the formatting is right is as simple as it gets but if we want prettier for format it too it gets a little more complicated as it will create a new commit but still very doable (I am personally running it this way in a private repo with a github actions workflow).
I can't say I know off hand how to do it in CircleCI but it can't be that hard to figure out.

madsnedergaard:
> Now just adding prettier to check if the formatting is right is as simple as it gets but if we want prettier for format it too it gets a little more complicated as it will create a new commit but still very doable (I am personally running it this way in a private repo with a github actions workflow). I can't say I know off hand how to do it in CircleCI but it can't be that hard to figure out.

Hmm, the committing changes part is actually a problem with forks (see discussion here: https://github.com/electricitymap/electricitymap-contrib/pull/4131) 🤔 
Not sure how to best approach it, but maybe just checking (and failing on changes) is an okay first step? We could reuse/change the `lint_json` step in CircleCI to do this.

VIKTORVAV99:
I left a comment on the other PR but yeah just doing a check seems like the only option for now and it should be fine for most scenarios.

While reusing existing jobs are a option I think we want to do as much in parallel as we can to reduce CI run durations.
So it might make sense to create a separate job for it, would be easier to tell if it was linting or formatting that failed too.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   js-exec:
     docker:
-      - image: circleci/node:12.22
+      - image: cimg/node:14.19
```
VIKTORVAV99, line 6: Consider bumping this (and the others) to node 16 as it's the active LTS branch.

web/package.json
```diff
@@ -78,10 +79,11 @@
     "eslint-plugin-react-hooks": "4.5.0",
     "eslint-plugin-you-dont-need-lodash-underscore": "^6.12.0",
     "esm": "^3.2.25",
+    "husky": "^8.0.0",
     "lodash.flatmap": "^4.5.0",
     "mini-css-extract-plugin": "1.6.2",
     "ndjson-cli": "^0.3.1",
-    "nodemon": "^1.17.1",
+    "nodemon": "^2.0.16",
```
madsnedergaard, line 86: This fixed the annoying error when running `yarn install` and seeing a bunch of errors from `fsevents`/`node-pre-gyp` :)

.circleci/config.yml
madsnedergaard: Good point, might as well now that I'm bumping it anyway :)

web/package.json
VIKTORVAV99: It also fixes a bunch of security vulnerabilities. 👀

I was actually planning on submitting a PR for it after the current PRs that target the lock file was merged.

Kongkille (approved):
Excellent 👌 

Edit: I'm just going to go ahead and fix the merge conflicts

madsnedergaard:
Thanks for fixing and merging ❤️

PR #4138, opened 2022-05-17
Setup format tooling

pierresegonne:
* Added `black` and `isort` as dependencies
* Centralise tooling scripts within `scripts/tooling.py`
* Have a script to run formatting
* Improve lint script with isort and black checks.


scripts/tooling.py
```diff
@@ -14,10 +14,17 @@ def _run(cmd: str):
         sys.exit(r)
 
 
+def format():
+    _run(f"isort --profile=black .")
+    _run("black .")
+
+
 def lint():
```
madsnedergaard, line 22: Just to ensure I understand it correctly, you envision `format` to be called locally and `lint`/`check` to be called from CI, right? (hence the `--check` flags)

madsnedergaard (approved):
Awesome!

Issue #2838, opened 2020-12-09
More data for Georgia (Country): forecast + historical


q--:
http://gse.com.ge/for-customers/data-from-the-power-system/production-and-consumption

The download button lets you download an XLS file with more detail than the graph; it also has a generation breakdown. (Check the two tabs, "Plan" and "Fact", for forecast and actual production.)

![Planned Georgia Generation Breakdown Excel Sample](https://user-images.githubusercontent.com/2470175/101661757-f658c080-3a48-11eb-9890-9ce74ab2c1e9.png)

Solar production appears to be zero in the XLS unfortunately (it's missing completely from the real-time data we're using now), so either they don't have utility-scale solar plants yet or the data is missing.

madsnedergaard:
The historical data part was addressed in #2847 so closing this again :)

PR #4141, opened 2022-05-17
GE parser: production forecast support  (and a few questions before this is ready to merge)

q--:
Implements the production forecast data source mentioned in #2838

A few questions / things that might still be broken:

- Haven't tested except for syntax errors
- No logic for dealing with no target datetime given yet; I saw in the ENTSOE parser that the corresponding method there returns 'generation forecast for 48 hours ahead and previous 24 hours' in that scenario, but is that standard or ENTSOE-specific? How should this parser handle it?
- I return a `dict`, ENTSOE returns a `list` for this - are both allowed?
- I changed the forecast parser from ENTSOE to this one - is it desirable, or should this parser just exist as a fallback that can be switched to if necessary?
- The ENTSOE parser has a `@refetch_frequency` defined in a few places, should this parser? Is there any documentation on how this works?

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -2435,12 +2435,13 @@
     "contributors": [
       "alixunderplatz",
       "kruschk",
-      "tmslaine"
+      "tmslaine",
+      "q--"
     ],
     "parsers": {
       "consumption": "ENTSOE.fetch_consumption",
       "consumptionForecast": "ENTSOE.fetch_consumption_forecast",
-      "generationForecast": "ENTSOE.fetch_generation_forecast",
+      "generationForecast": "GE.fetch_generation_forecast",
       "production": "GE.fetch_production",
       "productionPerModeForecast": "ENTSOE.fetch_wind_solar_forecasts"
     },
--- a/parsers/GE.py
+++ b/parsers/GE.py
@@ -19,35 +19,13 @@ TIMEZONE = 'Asia/Tbilisi'
 URL = urllib.parse.urlsplit('https://gse.com.ge/apps/gsebackend/rest')
 URL_STRING = URL.geturl()
 
-
-@config.refetch_frequency(datetime.timedelta(hours=1))
-def fetch_production(zone_key='GE',
+# Get the (forecast) production mix for every hour on the day of interest.
+def fetch_production_or_production_forecast(zone_key='GE',
                      session=None,
                      target_datetime=None,
-                     logger=logging.getLogger(__name__)) -> dict:
-    """Request the last known production mix (in MW) of a given country."""
-    session = session or requests.session()
-    if target_datetime is None: # Get the current production mix.
-        # TODO: remove `verify=False` ASAP.
-        production_mix = session.get(f'{URL_STRING}/map', verify=False) \
-                                .json()['typeSum']
-        return validation.validate(
-            {
-                'datetime': arrow.now(TIMEZONE).floor('minute').datetime,
-                'production': {
-                    'gas': production_mix['thermalData'],
-                    'hydro': production_mix['hydroData'],
-                    'solar': production_mix['solarData'],
-                    'wind': production_mix['windPowerData'],
-                },
-                'source': URL.netloc,
-                'zoneKey': 'GE',
-            },
-            logger,
-            remove_negative=True,
-            floor=MINIMUM_PRODUCTION_THRESHOLD)
-    else:
-        # Get the production mix for every hour on the day of interest.
+                     logger=logging.getLogger(__name__),
+                     forecast=False) -> dict:
+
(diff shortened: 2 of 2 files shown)


pierresegonne:
> No logic for dealing with no target datetime given yet; I saw in the ENTSOE parser that the corresponding method there returns 'generation forecast for 48 hours ahead and previous 24 hours' in that scenario, but is that standard or ENTSOE-specific? How should this parser handle it?

If no target_datetime is given, you can assume that the time for which we're fetching is now. And the time window returned will depend on the output of the parser. In the case of ENTSO-E, I believe it says this because the ENTSO-E forecasts, queried with now as a datetime will return data from now - 24h to now + 48h.

> I return a dict, ENTSOE returns a list for this - are both allowed?

The supported formats are either a dict (that represents a parser event, datetime, production values etc) or a list of dicts (list of parser events for different datetimes). I noticed yesterday that the documentation around the expected return types for the parsers is not super clear, so I'll try to make it clearer today.

> I changed the forecast parser from ENTSOE to this one - is it desirable, or should this parser just exist as a fallback that can be switched to if necessary?

We can only support a single parser for a given mode (e.g productionForecast) for a given zone. So if we replace the ENTSO-E parser, this will become the only official parser.
What could be done is to build a wrapper around the ENTSO-E parser that only fetches data with this source when ENTSO-E is not returning what we expect, and declare this new parser as the official one. 
If we do that though, I'd make sure that there is consistency between the two different sources. If the forecasts were to change drastically when going from one source to the other, then we would have to decide to only keep one. 

> The ENTSOE parser has a @refetch_frequency defined in a few places, should this parser? Is there any documentation on how this works?

If you know the refetch frequency, then yes please add it. That frequency is defined as the time period that separates two refetches to keep complete data. So for example if it's 00:00 and the parser returns data from yesterday 00:00 to 23:00 today, then the refetch frequency can be set to 47 hours.

q--:
> If no target_datetime is given, you can assume that the time for which we're fetching is now. And the time window returned will depend on the output of the parser. In the case of ENTSO-E, I believe it says this because the ENTSO-E forecasts, queried with now as a datetime will return data from now - 24h to now + 48h.

The interesting thing with this data source is that the desired time range can be set, so it can be data for a single hour or, say, all data between September 3 10:00 last year and now. But IIRC ElecticityMap only gives a single timestamp as an argument for the relevant parser method, so I'm not sure how much data should be fetched.

>If you know the refetch frequency, then yes please add it. That frequency is defined as the time period that separates two refetches to keep complete data. So for example if it's 00:00 and the parser returns data from yesterday 00:00 to 23:00 today, then the refetch frequency can be set to 47 hours.

Maybe I'm overlooking something, but I'm actually not sure what refetch _is_. Is it when to re-attempt when no data could be fetched? Is it in case the data fetched originally fetched is incomplete, and should be re-fetched after a certain amount of time?

q--:
> We can only support a single parser for a given mode (e.g productionForecast) for a given zone. So if we replace the ENTSO-E parser, this will become the only official parser.
> What could be done is to build a wrapper around the ENTSO-E parser that only fetches data with this source when ENTSO-E is not returning what we expect, and declare this new parser as the official one.
> If we do that though, I'd make sure that there is consistency between the two different sources. If the forecasts were to change drastically when going from one source to the other, then we would have to decide to only keep one.


Now that I think of it, the GSO data source has data by hour
GSO forecast data by hour (Excel screenshot)
and ENTSOE has it in something like 15-minute intervals, right?

In that case it might be better to keep the config pointing to ENTSOE and only have this parser as a fallback that can be enabled with a simple config change when it's needed.

madsnedergaard:
What's the status on this, should we just close it in favor of the ENTSO-E parser? Adding support for having both is something we don't really have the capacity to do these days :)

q--:
Yeah, this has kinda fallen by the wayside. Busy with my job and other things.

Anyway, I believe it should still work and as GE.py exists anyway, it might be worth adding even if it isn't used directly. But last time I had trouble setting up the full ElectricityMap environment locally, so I still haven't been able to test it properly. I don't know if and when I will give doing so another try, so if someone else wants to test it, or if the PR is closed for now, I understand.

madsnedergaard:
> Yeah, this has kinda fallen by the wayside. Busy with my job and other things.

That is very fair! :)


> Anyway, I believe it should still work and as GE.py exists anyway, it might be worth adding even if it isn't used directly. But last time I had trouble setting up the full ElectricityMap environment locally, so I still haven't been able to test it properly. I don't know if and when I will give doing so another try, so if someone else wants to test it, or if the PR is closed (for now), I understand.

I tried continuing work on this PR, but it requires a significant amount of work and testing to get it working again, so we'll unfortunately have to close it down for now. If anyone wants to pick it back up they are very welcome to do so :)

Issue #4142, opened 2022-05-18
UK Nuclear capacity can be updated


madsnedergaard:
Got the following via email from a person at https://niauk.org:

> I just wanted to note that UK nuclear capacity is now down to about 6.9 GW as Hunterston B retired in January. The map still shows 7.89 GW capacity, but we've shrunk unfortunately. 

> You can use the World Nuclear Association’s UK country page: https://world-nuclear.org/information-library/country-profiles/countries-t-z/united-kingdom.aspx, which has a graph of the UK’s total capacity (and further down the page, it goes line by line through each of the reactors and define their nameplate capacity.

> Note that unless something changes, Hinkley Point B nuclear power station will retire in Mid-July 2022, removing a further 965 MW of nuclear capacity.

> Also, you can use the EDF’s page here: https://www.edfenergy.com/energy/power-station/daily-statuses to verify which reactors it views as still operable as well.

KabelWlan:
Can be closed since #4147 is merged :)

PR #4147, opened 2022-05-18
update UK nuclear capacity

KabelWlan:
Update UK nuclear capacity, resolving #4142.


madsnedergaard (approved):
Nice, same [source](https://www.edfenergy.com/energy/power-station/daily-statuses)? :)

KabelWlan:
He listed EDF as source which is already listed. I could add the World Nuclear Association’s UK country page as well because there you see the capacity right away without the need of adding all the reactors up.

madsnedergaard:
It's probably fine with EDF listed :)

PR #4148, opened 2022-05-18
Reduce center coordinate accuracy from 111 pm to 11.1 km

VIKTORVAV99:
This reduces the center coordinate accuracy in the `world.json` file from 111 picometers (15 decimal places) (atomic scale 😅) to 11.1 km (1 decimal place) in a effort to reduce the size of the `world.json` file and the frontend bundle.

Since this is only used for centering the map over a specific region I think it's okay to use 1 decimal place (11.1 km accuracy), it won't be a perfect centering but at the zoom level we use you can't really tell the difference with the naked eye.

Here is a table of decimal places vs accuracy for reference:
| decimal places | degrees    | accuracy           |
|----------------|------------|--------------------|
| 0              | 1.0        | 111 km ± 55.5 km   |
| 1              | 0.1        | 11.1 km ± 5.55 km  |
| 2              | 0.01       | 1.11 km ± 0.555 km |
| 3              | 0.001      | 111 m ± 55.5 m     |
| 4              | 0.0001     | 11.1 m ± 5.55 m    |
| 5              | 0.00001    | 1.11 m ± 0.555 m   |
| 6              | 0.000001   | 1.11 dm ± 0.555 dm |
| 7              | 0.0000001  | 1.11 cm ± 0.555 cm |
| 8              | 0.00000001 | 1.11 mm ± 0.555 mm |


Kongkille (commented):
Hey @VIKTORVAV99 agreed

I'm thinking since we have a lot of our computations around turf functionality. Perhaps we could use turf.truncate instead?

We actually also use it here 
https://github.com/electricitymap/electricitymap-contrib/blob/760a66158735fbb72c53c6e706568e674269ffa6/web/geo/utilities.js#L32

VIKTORVAV99:
> Hey @VIKTORVAV99 agreed
> 
> I'm thinking since we have a lot of our computations around turf functionality. Perhaps we could use turf.truncate instead?
> 
> We actually also use it here
> 
> https://github.com/electricitymap/electricitymap-contrib/blob/760a66158735fbb72c53c6e706568e674269ffa6/web/geo/utilities.js#L32
> 
> Edit: here is the docs by the way https://turfjs.org/docs/#truncate

Wouldn't truncate just cut of the residual values instead of rounding them? (if the function name is anything to go by, I haven't looked at the documentation yet.)

Kongkille:
Oh and fyi, it doesn't seem the cypress test failing has anything to do with this PR, so I'll just skip the test

VIKTORVAV99:
I think it is a no go on using turf for this particular scenario.
It generates the following code:
```json
"center": { "type": "Feature", "properties": {}, "geometry": { "type": "Point", "coordinates": [33.6, 35.3] } }
```
vs my simple rounding:
```json
"center": [33.6, 35.3]
```
vs original:
```json
"center": [33.644854714990075, 35.34916520661007]
```

q--:
Not sure I completely understand this, but would this also work for tiny island regions? (E.g. Rottnest Island is only 19 km^2)

VIKTORVAV99:
> Not sure I completely understand this, but would this also work for tiny island regions? (E.g. Rottnest Island is only 19 km2)

It is just used when centering the map over a location when you click on the zone in the ranking list for example. This means it should have no impact on how big the zone is question is as it's just 1 coordinate point that point to the center of the zone. And since the map is zoomed out the lower accuracy won't even be noticed.

Just tested it with Rottnest Island and it works.

I could easily change it to two decimals though if it's a concern.

Kongkille (approved):
Yeah turf will generate the entire point feature. We could just extract the coordinate array from the point for consistency since performance wouldn't be relevant, but at that point we might as well just use your round function. 

And yeah the size of Island should not matter for the viewport move :)

Nice fix!

PR #4152, opened 2022-05-18
Adds cypress workflow to run on pull request with 'frontend' label

Kongkille:
This should trigger a Github Action on pull requests with label = frontend that executes all our cypress tests (see #4136)

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,7 +1,6 @@
-name: Cypress tests
+name: Cypress
 on:
-  pull_request:
-    types: [ labeled ]
+  pull_request
 env:
   SNOWPACK_PUBLIC_ELECTRICITYMAP_PUBLIC_TOKEN: '${{ secrets.SNOWPACK_PUBLIC_ELECTRICITYMAP_TOKEN }}' # token has to be escaped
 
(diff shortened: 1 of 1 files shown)


Kongkille:
> Instead of using labels I think you might want to use something like this action: [Changed Files](https://github.com/marketplace/actions/changed-files)
> 
> If this is used with a setup action we can reuse it for other things later like this:
> 
> ```yaml
> name: CI
> 
> on:
>   push:
>     branches:
>       - main
>   pull_request:
>     branches:
>       - main
> 
> jobs:
>   setup:
>     runs-on: ubuntu-latest  # windows-latest | macos-latest
>     name: Test changed-files
>     outputs:
>       frontend: steps.changed-files.outputs.any_changed
>     steps:
>       - uses: actions/checkout@v3
>         with:
>           fetch-depth: 0  # OR "2" -> To retrieve the preceding commit.
> 
>       - name: Get changed files
>         id: changed-files
>         uses: tj-actions/changed-files@v20
>         with:
>           path: web
>           
>   test:
>     needs: setup
>     if: needs.setup.outputs.frontend == 'true'
>     # rest of test job
> ```

Makes sense, right now I'm just trying to get something out working. The frontend label might be ok for our first version, since it's automatically added to most things that we'd want to test with Cypress, but you're right. It would likely miss something

PR #4157, opened 2022-05-19
Adds some Cypress integration tests

Kongkille:
- Remove test for countryPanel. It can potentially fail for other reasons than what we are testing for.
- Add history and state API tests to make sure the app requests the correct data succesfully

### DIFF
--- a/web/cypress/integration/api/history.spec.js
+++ b/web/cypress/integration/api/history.spec.js
@@ -1,0 +1,7 @@
+describe('History', () => {
+  it('Asserts visiting zone requests history data succesfully', () => {
+    cy.intercept('https://app-backend.electricitymap.org/v4/history?countryCode=FR').as('fetchHistory');
+    cy.visit('/zone/FR?remote=true');
+    cy.wait('@fetchHistory').its('response.statusCode').should('eq', 200);
+  });
+});
--- a/web/cypress/integration/api/state.spec.js
+++ b/web/cypress/integration/api/state.spec.js
@@ -1,0 +1,7 @@
+describe('State', () => {
+  it('Asserts visiting map requests state data correctly', () => {
+    cy.visit('/map?remote=true');
+    cy.intercept('https://app-backend.electricitymap.org/v4/state').as('getState');
+    cy.wait('@getState').its('response.statusCode').should('eq', 200);
+  });
+});
--- a/web/cypress/integration/countrypanel.spec.js
+++ b/web/cypress/integration/countrypanel.spec.js
@@ -1,9 +1,4 @@
 describe('Country Panel', () => {
-  it('asserts countryPanel contains carbon intensity when available', () => {
-    cy.visit('/zone/SE?remote=true');
-    cy.get('[data-test-id=carbon-intensity').should('exist');
-  });
-
   it('asserts countryPanel contains no parser message', () => {
     cy.visit('/zone/CN?remote=true');
     cy.get('[data-test-id=no-parser-message]').should('exist');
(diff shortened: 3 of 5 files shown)


web/cypress/integration/api/history.spec.js
```diff
@@ -0,0 +1,7 @@
+describe('History', () => {
+  it('Asserts visiting zone requests history data succesfully', () => {
+    cy.intercept('https://app-backend.electricitymap.org/v4/history?countryCode=FR').as('fetchHistory');
+    cy.visit('/zone/FR?remote=true');
+    cy.wait('@fetchHistory').its('response.statusCode').should('eq', 200);
```
gwpicard, line 5: Should you also do a check for what is returned? Or are you just looking for a 200 response?

web/cypress/integration/api/history.spec.js
```diff
@@ -0,0 +1,7 @@
+describe('History', () => {
```
gwpicard, line 1: In the longer term, should this testing go into the app backend testing directly? Since here we are testing the output of something that we should trust (the app should assume that the API works and is returning things correctly).

web/cypress/integration/countrypanel.spec.js
```diff
@@ -1,9 +1,4 @@
 describe('Country Panel', () => {
-  it('asserts countryPanel contains carbon intensity when available', () => {
```
gwpicard, line 2: Maybe worth adding a TODO here so we don't forget to reintroduce a test later

web/cypress/integration/api/state.spec.js
```diff
@@ -0,0 +1,7 @@
+describe('State', () => {
+  it('Asserts visiting map requests state data correctly', () => {
+    cy.visit('/map?remote=true');
+    cy.intercept('https://app-backend.electricitymap.org/v4/state').as('getState');
+    cy.wait('@getState').its('response.statusCode').should('eq', 200);
```
gwpicard, line 5: Same comment as above—only for a 200 test or should we also test output?

web/cypress/integration/api/history.spec.js
Kongkille: The idea is to have both types of testing. This should not be considered as our main API tests, it just verified that the app is able to query the backend correctly :)

gwpicard: Is the way the app queries the data different from a regular API call? I'm trying to understand the difference and why this test is needed on top of the API tests.

web/cypress/integration/countrypanel.spec.js
Kongkille: Hmm I don't think that's necessary I just added the test initially yesterday, and I am not sure we'd want to handle the test like this

web/cypress/integration/api/history.spec.js
Kongkille: Yes, well the app queries the app-backend which again queries our standard API, so there's a middlestep which is the app-backend.

Ideally we want a bunch of unit tests for the app and the app-backend but now I am just testing the **integration** step, i.e. whether the app and the app-backend is able to communicate properly

So to put some words on these tests, it essentially tests that we're able to spin up the app and create authorized requests to the production-app-backend based on user actions, which may seem simple, but is a super nice sanity check for us before merging things

gwpicard: Good to know, makes sense!

gwpicard (approved):

PR #4159, opened 2022-05-19
Remove duplicate keys from `zones.json`

VIKTORVAV99:
Not sure when these duplicates snuck into the code but this PR fixes that.

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -4391,7 +4391,6 @@
     "contributors": [
       "JeanBaptisteScellier"
     ],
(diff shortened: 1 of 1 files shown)


madsnedergaard (approved):
Hmm, interesting that `jsonlint` didn't complain about this 🤔 
But thanks for fixing it!

VIKTORVAV99:
> Hmm, interesting that `jsonlint` didn't complain about this 🤔 But thanks for fixing it!

I found it strange too but luckily VS Code screamed it as loud as it could and it was a easy fix. 😅

PR #4160, opened 2022-05-19
Update CA_QC.py

EarthlyAlien:
Updated CA_QC.py parser so that it can return multiple datapoints instead of just single recent datapoint.

### DIFF
--- a/parsers/CA_QC.py
+++ b/parsers/CA_QC.py
@@ -41,6 +41,7 @@ def fetch_production(
     list_res = []
     for elem in reversed(data["details"]):
         if elem["valeurs"]["total"] != 0:
+            
             list_res.append({
                 "zoneKey": zone_key,
                 "datetime": arrow.get(elem["date"], tzinfo=timezone_id).datetime,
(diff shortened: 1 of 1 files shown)


parsers/CA_QC.py
```diff
@@ -38,13 +38,13 @@ def if_exists(elem: dict, etype: str):
             return 0.0
 
     data = _fetch_quebec_production()
+    list_res = []
     for elem in reversed(data["details"]):
         if elem["valeurs"]["total"] != 0:
-
-            return {
+            list_res.append(dict({
```
jarek, line 44: `dict({})` is redundant - just `{}` also creates a dict (this is called a "dict literal"). Same on line 47 below

EarthlyAlien: Thanks @jarek. I've made changes accordingly.

madsnedergaard (approved):
Nice! 🙌

VIKTORVAV99:
Looks like the python formatting check is failing.
If you run `poetry run format` that should resolve the issue.

EarthlyAlien:
@VIKTORVAV99 I ran `poetry run format` but nothing changed.

Here's the screenshot after running the above command.

image

VIKTORVAV99:
> @VIKTORVAV99 I ran `poetry run format` but nothing changed.
> 
> Here's the screenshot after running the above command.
> 
> image

It looks like black isn't installed yet. Try running `poetry install` first then `poetry run format` again.

EarthlyAlien (commented):
Formatting done.

PR #4162, opened 2022-05-20
Update Sweden real-time data source

VIKTORVAV99:
I realized I forgot to do this in #4083 but better late than never right?


madsnedergaard (approved):
Makes sense, thanks for following up :)

Issue #4149, opened 2022-05-18
[a11y] Work without JavaScript, and with screen readers


Tjoppen:
Currently if a user visits the site with JS disabled then they are greeted with a white screen, even if they have CSS disabled. This makes using the site with a screen reader nearly impossible. Render the HTML serverside instead, delivering updates via JS for clients that have it enabled. ```<meta>``` refresh could be used to enable JS-free refresh.

This is also one of the times where a ```<table>``` is appropriate for blind users, each country having their own row. Use rowspan to deal with multiple regions in one country.

VIKTORVAV99:
Some of these concerns are definitely valid and should be looked at as soon as possible but I believe making the map/application work with JS disabled and using server-side rendering is outside the scope of what is possible, it would require big changes in both code and infrastructure. Not even Google let's you use Google Maps with JS disabled.

We should probably add a fallback message telling users that JavaScript is required if they have it disabled though.

madsnedergaard:
Closing this as it's not something we can do at the moment. Thanks for the JS fallback message which at least makes it clear to users :)

PR #4165, opened 2022-05-21
Add noscript fallback and necessary changes

VIKTORVAV99:
This PR adds a barebones fallback when JavaScript is disabled telling users it's required as disused in #4149

### Preview:
Web capture_21-5-2022_1733_localhost

In this PR:
- Added `<noscript>` fallback
- Added necessary CSS to `<style>` tags since we usually load CSS with js.
- Added `lang="en"` to the ejs template since we can't set it with JS (gets overwritten by i18n when JS in enabled).
- All ejs changes was mirrored to to snowpacks `index.html` for parity.


web/views/pages/index.ejs
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
-<html xmlns="http://www.w3.org/1999/xhtml" xmlns:fb="http://ogp.me/ns/fb#">
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:fb="http://ogp.me/ns/fb#" land="en">
     <head>
-        <meta http-equiv="content-type" content="text/html; charset=UTF8" />
+        <meta charset="utf-8" />
```
VIKTORVAV99, line 4: This is just the short form introduced with html5, functionality is the same.

web/public/index.html
```diff
@@ -1,44 +1,84 @@
...
+    </script>
+    <style>
+      html,
+      body {
+        margin: 0 !important;
+        padding: 0 !important;
+        font-family: 'Open Sans', sans-serif;
+        transition: background-color 800ms;
+        overflow: hidden; /* Disable dragging that can occur on mobile */
+      }
```
madsnedergaard, line 56: I don't think we need to implement this styling directly here, in the build it's loading CSS fine with JS disabled :)

Screenshot 2022-05-23 at 13 47 20
Screenshot 2022-05-23 at 13 47 12

VIKTORVAV99: Hmm it didn't when I tried it last time.

I'm not at my computer right so can't check right now but I'll test it on Edge and Firefox when I get back. (might be browser specific)

madsnedergaard: It makes sense if it fails during local development, but it should be the same on production no matter the browser? 🤔

VIKTORVAV99: Found the problem! We don't actually load the CSS bundles in html when using snowpack but rather snowpack use js to inject them (for HMR?).

I'll change back the ejs template when I get back and keep the styles in the snowpack index file when I get back.

VIKTORVAV99:
@madsnedergaard should be all ready now!

(prettier also decided to format the entire scss file 😅)

web/views/pages/index.ejs
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html xmlns="http://www.w3.org/1999/xhtml" xmlns:fb="http://ogp.me/ns/fb#" land="en">
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:fb="http://ogp.me/ns/fb#" lang="en">
```
madsnedergaard, line 2: Good catch!

madsnedergaard (approved):

web/views/pages/index.ejs
VIKTORVAV99: I will attribute this one to [webhint](https://marketplace.visualstudio.com/items?itemName=webhint.vscode-webhint) 😅

Issue #4166, opened 2022-05-21
Manual overwrite 'Low Carbon' & 'Renewable' percentage (KW/IL/TH)


PPsyrius:
Right now, **Kuwait (KW)**, **Israel (IL)** and **Thailand (TH)** already have a manual overwrite for their own *Carbon Intensity* data as for their total real-time production/consumption data.

Thus, a manual overwrite for *Low Carbon* and *Renewable* percentage seems like the next logical step for these three countries until a proper breakdown data source is made publicly available.

![image](https://user-images.githubusercontent.com/19505219/169662245-dd572b7f-1f49-41eb-8b78-146228588edd.png)

VIKTORVAV99:
Kuwait is sadly 0% on both anyway so no point in changing that one but the other two have some renewable production, I'll see if I have the time to create a PR for those later today.

q--:
Kuwait has [a tiny amount of wind and solar power](https://www.iea.org/data-and-statistics/data-tables/?country=KUWAIT&energy=Electricity&year=2019). Probably disappears due to rounding in the country panel, but might be worth adding if this data is ever going to be used for anything other than the map.

PR #4167, opened 2022-05-22
Renewable and low-carbon overrides for IL and TH

VIKTORVAV99:
Uses the share of renewable sources in unknown production for each country.

Closes: #4166


pierresegonne (approved):
👏

Issue #4161, opened 2022-05-20
[RFC] Use UN Life Cycle Assessment of Electricity Generation Options gCO2-eq/kWh values for EU28.


VIKTORVAV99:
## Consider using the UN Life Cycle Assessment of Electricity Generation Options reports gCO2-eq/kWh values for EU28

The UN has published a report that contains LCAs for several modes of electricity production with specific values for EU28, I would propose we use these values for EU28 countries (current EU countries + UK) as an interim solution until #738 is finalized.

Link to the report: https://unece.org/sites/default/files/2021-10/LCA-2.pdf

Table with current value and the proposed UN value we should use:
| PER KWH     | Type                                   | Current | UN EU28 value                    |
|-------------|----------------------------------------|---------|----------------------------------|
| Hard Coal   | PC, without CSS                        | 820     | 1000 (highest coal value)        |
| Natural Gas | NGCC, without CCS                      | 490     | 430 (highest gas value)          |
| Hydro       | 660 MW                                 | 24      | 150                              |
| Hydro       | 360 MW                                 | 24      | 11                               |
| Nuclear     | average                                | 12      | 5.1                              |
| Solar       | poly-Si, ground-mounted / roof-mounted | 45      | 37 (highest photovoltaic value)  |
| Wind        | onshore                                | 11      | 12                               |

Full details of UN EU28 values on page 74 of the report (PDF page 84)

I realize hydro is probably not an option to implement but the others should be pretty straight forward, especially Gas, Nuclear, Solar and Wind.

q--:
If the highest possible value from the report is lower than the IPPC 2014 average ElectricityMap is using now, it makes sense to use the lower value. However, if the value from the report is higher we probably shouldn't switch. After all, we don't know if a plant is using Pulverized Coal technology (the worst option, worse than current coal carbon intensity value) or Supercritical (the best option, lower than current coal carbon intensity value)

VIKTORVAV99:
> If the highest possible value from the report is lower than the IPPC 2014 average ElectricityMap is using now, it makes sense to use the lower value. However, if the value from the report is higher we probably shouldn't switch. After all, we don't know if a plant is using Pulverized Coal technology (the worst option, worse than current coal carbon intensity value) or Supercritical (the best option, lower than current coal carbon intensity value)

Makes sense, that would mean just the values for Gas, Nuclear and Solar.

[...]

Issue #4143, opened 2022-05-18
Replace outdated animation libraries


madsnedergaard:
We're using two animation libraries. They are both outdated and blocks updating to react@18.
Maybe consider using [react-spring](https://react-spring.io/) which is more popular and frequently updated?

---

**[react-motion](https://github.com/chenglou/react-motion)**
It hasn't been updated since 2017 and it's blocking for updating to react@17.

It's only used for `<CircularGauge />`, so it shouldn't be hard - would be nice to keep the animation with another library (or maybe styled-components is enough?), but might also be fine to just remove the animation entirely at first :)

**[react-transition-group](https://github.com/reactjs/react-transition-group)**
It's not working with react@18 according to [this issue](https://github.com/reactjs/react-transition-group/issues/816).

Only used for CSSTransition, maybe we can just use styled-components functionality for this?

VIKTORVAV99:
PR #4168 should take care of react-motion. 🙂

VIKTORVAV99:
Since we changed the whole graphing stack in the refractor we can consider this one closed.

PR #4168, opened 2022-05-22
Switch react-motion to react-spring

VIKTORVAV99:


VIKTORVAV99:
Found a bug, will remove draft status once I find a fix.

VIKTORVAV99:
I have looked into this and looks like react-spring might not be the package we want to use to animate SVG paths, it's interpolation metod seem to create some weird results especially on arcs (like in the gauges use).

You guys can try it yourselves as it technically works but it morphs the shape in a weird way (very noticeable when you drag the time slider back and forth fast).

If anyone has a solution for it I'd be happy to hear it otherwise it might just be better to remove the animation for now as mentioned in #4143 so we can upgrade to later versions of react and then we can find a way to re-implement the current behavior and use on the graph itself for consistency.

VIKTORVAV99:
> Ugh, annoying that it jumps weirdly around :/
> 
>  Kapture.2022-05-25.at.16.32.16.mp4 
> Dunno if it would make sense to switch to some package for this like https://github.com/naikus/svg-gauge? 🤔 (just first result when googling, haven't looked further into it except [the demo](https://codepen.io/naikus/pen/BzZoLL))

Looks like that package would work just fine. I also found this package, [reaviz](https://reaviz.io/?path=/story/charts-gauge-radial--custom-arc&knob-Animate%20background%20arc=true&knob-Arc%20corner%20radius=12.5&knob-Arc%20width=25&knob-Color=#38e52c&knob-Color%20Scheme=#CE003E,#DF8D03,#00ECB1,#9FA9B1&knob-Data=[{%22key%22:%22Austin,%20TX%22,%22data%22:24}]&knob-End%20Angle=6.283185307179586&knob-Height=300&knob-Start%20Angle=0&knob-Width=700), that could work, it also contain a lot of other visualization options if we want to use it in the future. It was made for react so there might be some benefit to that. [NPM link](https://www.npmjs.com/package/reaviz)

But generally I am in favor of keeping it as simple as possible.

madsnedergaard:
Hmm, good point.. We've also been discussing switching the graphs to a more standardized React library (instead of the heavily customized d3 setup we current have), so maybe worth parking this until that happens? 🤔

VIKTORVAV99:
> Hmm, good point.. We've also been discussing switching the graphs to a more standardized React library (instead of the heavily customized d3 setup we current have), so maybe worth parking this until that happens (to make sure we spend time on finding and implementing the right thing)? 🤔

If the backend was opened up for more production modes (i.e. uncombine the combined ones as suggested in #4161 albeit for a different reason) [this wiz](https://reaviz.io/?path=/story/charts-bar-chart-horizontal-multi-series--stacked&knob-Animate%20background%20arc=true&knob-Arc%20corner%20radius=12.5&knob-Arc%20width=25) would look pretty good (if we change the styling) 😉

madsnedergaard:
Too bad this wasn't working well enough, I suggest we close this PR for now and revisit this at a later stage by looking into all visualisations :)

PR #4170, opened 2022-05-23
Update tests to use mockserver

Kongkille:


VIKTORVAV99:
FYI we have a race condition between the cypress tests and the labeler where the cypress tests are skipped before the labeler are finished.

We might want to go ahead and stat using a setup job or something similar as I suggested before.
(I can do it if you want.)

Kongkille:
> FYI we have a race condition between the cypress tests and the labeler where the cypress tests are skipped before the labeler are finished.
> 
> We might want to go ahead and stat using a setup job or something similar as I suggested before. (I can do it if you want.)

@VIKTORVAV99 That would be amazing yeah, however, I don't get why cypress tests aren't running when I force re-run after the label has been set manually.

If you are able to create some kind of process that would be super helpful.

madsnedergaard (approved):
LGTM! :)

.github/workflows/cypress.yml
```diff
@@ -15,6 +15,6 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           working-directory: ./web
-          start: yarn develop
+          start: yarn develop, yarn run mockserver
           record: true
           wait-on: "http://localhost:8080"
```
madsnedergaard, line 20: Do we need another wait-on, or is it fine? (The mockserver should always be available before the dev env is spinning up though) :)

Kongkille: That was my thought as well, so I don't think it's necessary, but if it ever fails we can add it?

VIKTORVAV99:
eeh before running the mockserver did we run yarn install in that directory? 🙈

Kongkille:
> eeh before running the mockserver did we run yarn install in that directory? 🙈

Hmm no I guess not. It will automatically install in ./web I am not sure what the best way to install in multiple directories would be.

We can override the install command, maybe there's something there? 

https://github.com/marketplace/actions/cypress-io#custom-install-command

PR #4171, opened 2022-05-23
CI setup job

VIKTORVAV99:
Makes the CI more robust by using a setup job to detect changed files and run jobs conditionally based on that.


VIKTORVAV99:
@Kongkille here is the basic setup, I'll add the cypress tests too when #4170 is merged.

Kongkille (approved):
This looks really good. Thanks for digging into this!

PR #4175, opened 2022-05-23
Add custom install script in cypress

Kongkille:
Install both mockserver and web dependencies

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -5,6 +5,7 @@ env:
   CYPRESS_RECORD_KEY: '${{ secrets.CYPRESS_RECORD_KEY }}'
 jobs:
   cypress-test:
+    if: contains(github.event.pull_request.labels.*.name, 'frontend 🎨')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout


VIKTORVAV99:
can you just do a like whitespace change in a frontend file so the cypress tests run?

VIKTORVAV99:
Looks like we can't chain commands `error Command "&&" not found.`

VIKTORVAV99:
I would probably just add this step:
```yml
- name: Install mockserver
        run: |
          cd /mockserver
          yarn install
```
And skip the custom install script.

Kongkille:
> Looks like we can't chain commands `error Command "&&" not found.`

Yeah, I'll just try a few things otherwise I'll disable the tests for now

VIKTORVAV99:
Looks like the install worked but...
```
[FAILED] The Cypress App could not be downloaded.
[FAILED] 
[FAILED] Does your workplace require a proxy to be used to access the Internet? If so, you must configure the HTTP_PROXY environment variable before downloading Cypress. Read more: https://on.cypress.io/proxy-configuration
```

VIKTORVAV99:
oh no don't change that, it's working now 😅

Kongkille:
Ok it looks like it is finally working! 👯

VIKTORVAV99:
> Ok it looks like it is finally working! 👯

Yep, looks like the old one got stuck tho so I'll just cancel that.

Kongkille:
Thanks for the assist! Hopefully this will unblock the checks for now

VIKTORVAV99:
No problem I'm happy to help!

If nothing changes it should stay stable so no more red crosses unless they are supposed to be there!

PR #4177, opened 2022-05-23
Add subzones to ZONE_NEIGHBOURS

corradio:
This PR ensures that if a zone has subzones, then the parent will have as neighbors all the neighbors of the subzones.
This is required in order to be able to make spatial aggregates (as the exchanges will be aggregated across subzones).


electricitymap/contrib/config/__init__.py
```diff
@@ -38,6 +38,20 @@
         if zone_name_1 not in ZONE_NEIGHBOURS:
             ZONE_NEIGHBOURS[zone_name_1] = set()
         ZONE_NEIGHBOURS[zone_name_1].add(zone_name_2)
+# Manage subzones
+for zone in ZONES_CONFIG.keys():
```
madsnedergaard, line 42: This is quite a deep `for if for for if` thing going on, eh? 😅

PR #4178, opened 2022-05-23
Add prettier to the whole repo

VIKTORVAV99:
Waiting for #4171 to be merged before I can finalize this.


.prettierrc.js
```diff
@@ -0,0 +1,14 @@
...
+  trailingComma: 'es5',
+  printWidth: 120,
+  overrides: [
+    {
+      files: '*.md',
+      options: {
+        printWidth: 80,
+      },
+    },
+  ],
```
VIKTORVAV99, line 13: I have found it to be easier to read markdown files if they are formatted to 80 width.
This don't affect the text rendering but keeps code blocks from getting too wide.

VIKTORVAV99:
This should be ready for review now and once it has been reviewed I will add a commit that formats all the files.

.prettierrc.js
Kongkille: That might be ok however I am wondering if this is a good experience for editing 

https://github.com/electricitymap/electricitymap-contrib/edit/master/DATA_SOURCES.md

which is a file often edited by non-technical contributors, likely editing directly in the github web editor

VIKTORVAV99: I ran prettier on `DATA_SOURCES.md` locally and it don't change any of the text. The only changes where to whitespace, quotes and some html codes (like turning `&#32;` into spaces in link tooltips).

package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "electricitymap-contrib",
+  "license": "MIT",
+  "devDependencies": {
+    "prettier": "^2.6.2"
+  },
+  "scripts": {
+    "prettier-fix": "prettier --write .",
+    "prettier-check": "prettier --check ."
+  }
```
madsnedergaard, line 10: How about using `npx prettier` here, then we don't need dependencies at all at root? :)

madsnedergaard (commented):
Hey @VIKTORVAV99, thanks for the work here - and sorry for the slow review, I've been a bit unsure about this PR in general as I fear editing through the Github web interface might introduce style issues that contributors can't fix on their own... 
But maybe it won't be a problem and unified styling would be great, so let's try it :)

Some comments and ideas below

web/package.json
```diff
@@ -105,6 +105,9 @@
     "url": "https://github.com/electricitymap/electricitymap-contrib.git"
   },
   "lint-staged": {
+    "*": [
+      "prettier --write ."
+    ],
```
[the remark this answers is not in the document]
VIKTORVAV99: Good point!
I'll check if that syntax work (it should).

.editorconfig
```diff
@@ -9,8 +9,5 @@ charset = utf-8
 [*.{js,json}]
 indent_size = 2
```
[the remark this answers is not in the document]
VIKTORVAV99: I think moving it under `*` makes more sense as it would run on yaml files too (the github workflows).

package.json
VIKTORVAV99: Yeah I was conflicted about this, but I opted for this implementation so use the same version locally and in the CI. It's possible to run npx with a version specifier but it might get harder to maintain long term.

It's an easy change though so it is up to you and the team.

web/package.json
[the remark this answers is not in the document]
VIKTORVAV99: The syntax did work locally (with a slight change) but failed in lint-staged because it explicitly pass files to prettier. But changing the lint-staged matching to the following seems to work as expected:
```json
"!(*.{js,jsx,ts,tsx})": [
      "prettier --write"
    ],
```

madsnedergaard:
> > Hey @VIKTORVAV99, thanks for the work here - and sorry for the slow review, I've been a bit unsure about this PR in general as I fear editing through the Github web interface might introduce style issues that contributors can't fix on their own... But maybe it won't be a problem and unified styling would be great, so let's try it :)
> > Some comments and ideas below
> 
> Hey! Don't worry about the slow review!
> 
> I have addressed all but the last point (top level `package.json`), but I left a comment on that one. I won't be able change it (if needed) or run prettier on everything until tomorrow through.

Nice, I've responded on the last one. And that's perfectly fine, no stress 😄

VIKTORVAV99:
@madsnedergaard This should be all ready to merge now! 🚀

madsnedergaard (approved):
Aaaawesome, let's gooo! 🥳

PR #4182, opened 2022-05-24
Fix capacity parser NL

pierresegonne:
The NL capacity parser was crashing for a while - making the whole parser crash.

### DIFF
--- a/parsers/NL.py
+++ b/parsers/NL.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 from asyncio.log import logger
-from email.policy import default
+
 import logging
 import math
 from copy import copy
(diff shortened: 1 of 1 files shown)


gwpicard (approved):
🌬️ nice! Thanks for fixing 😄

pierresegonne:
Hey @q-- :)

The NL data will always be estimated, we have a model on top of the ENTSO-E data, as it is missing a lot of solar and gas production.

mhooi:
> Hey @q-- :)
> 
> The NL data will always be estimated, we have a model on top of the ENTSO-E data, as it is missing a lot of solar and gas production.

Do you have a link to the location where is explained how the NL data is estimated?

pierresegonne:
Hey @mhooi, we are currently using a model of that class: https://github.com/electricitymap/electricitymap-contrib/wiki/Reconstruct-Breakdown

I recognise though that it would be nicer to propagate production for modes we know are reliable as is done for CH

mhooi:
> Hey @mhooi, we are currently using a model of that class: https://github.com/electricitymap/electricitymap-contrib/wiki/Reconstruct-Breakdown
> 
> I recognise though that it would be nicer to propagate production for modes we know are reliable as is done for CH

Thanks for the quick reply. I did find the theoretical framework already. 

But where can I find the model and understand that for example the model estimates that the current solar production is 3.63 GW while in reality the actual number is more like 10 GW?

Issue #4181, opened 2022-05-24
Icons disappeared from mobile app.


VIKTORVAV99:
It looks like many of the icons have disappeared in the mobile app/build.

It's probable this is related to the changes in #3996 but it appears to be affecting only the mobile app.

(I noticed this on Android but it's likely that it's affecting iOS too.)

PR #4184, opened 2022-05-25
Fix icons in mobile app.

VIKTORVAV99:
This should fix the icons in the mobile app if the problem is what I think it is while keeping the svg sprite for the web.

Now I am not really sure how to test this in a app setting (on a phone/emulator) but I did make sure all styles are the same when possible where it was applicable.

Closes: #4181


madsnedergaard:
Awesome, I'll test it today! :)

madsnedergaard:
There's unfortunately still an issue, I'll try to dive deeper into why:

Screenshot 2022-05-25 at 16 42 08


You can also run the iOS simulator like this by the way:
```bash
cd mobileapp
docker-compose build web  && ./build.sh
cordova run ios
```

madsnedergaard:
Oh, turns out the fix was extremely simple 😅

VIKTORVAV99:
> There's unfortunately still an issue, I'll try to dive deeper into why:
> 
> Screenshot 2022-05-25 at 16 42 08
> 
> You can also run the iOS simulator like this by the way:
> 
> ```shell
> cd mobileapp
> docker-compose build web  && ./build.sh
> cordova run ios
> ```

Thanks for the build steps!

At least there are question marks now, that's some improvement at least? 😅

PR #4189, opened 2022-05-26
Update webmanifest

VIKTORVAV99:
Updates the webmanifest so the `name` and `short_name` match the names used on the app stores and adds a `description` (also matching the app stores).

Other changes:
- Changes the file extension from `.json` to `.webmanifest` so it matches the Web Application Manifest spec.
- Adds iOS (iTunes) to `related_applications`.
- Changes the size of the svg icons to any.
- Use short hex codes for `background_color` and `theme_color`.


VIKTORVAV99:
We should also strive to support [example screenshots](https://developer.mozilla.org/en-US/docs/Web/Manifest/screenshots), the onbording images are a good start but we should get both narrow (mobile) and wide (desktop) examples. (and maybe update the onbording images?)

madsnedergaard (approved):
Brilliant, thanks! 🙌 
The screenshot examples makes sense in theory, are they actually used anywhere in reality (in "progressive web app stores")?

VIKTORVAV99:
> Brilliant, thanks! 🙌 The screenshot examples makes sense in theory, are they actually used anywhere in reality (in "progressive web app stores")?

The only "PWA app store" I know them uses them is Misrosoft Store together with https://www.pwabuilder.com/ under the recommended section.

image
> **Note**
> This image uses the old PWA manifest.


It's far from a priority though (unless you want to get listed on the ms store but then there are some other stuff to be done first as well if we want to use the PWA).

VIKTORVAV99:
Oh and if you got the "iarc rating id" handy from google play we can use that so the same rating is used across stores that support PWAs

web/public/electricityMap.webmanifest
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "electricityMap",
+  "short_name": "electricityMap",
+  "description": "electricityMap is a live visualization of where your electricity comes from and how much CO2 was emitted to produce it.",
+  "start_url": "/",
+  "background_color": "#fff",
+  "theme_color": "#fff",
+  "display": "standalone",
+  "prefer_related_applications": true,
```
madsnedergaard, line 9: Found it!
```suggestion
  "prefer_related_applications": true,
  "iarc_rating_id": "194a8347-3f9e-4525-9e04-9969d2db0f56",
```

madsnedergaard:
> > Brilliant, thanks! 🙌 The screenshot examples makes sense in theory, are they actually used anywhere in reality (in "progressive web app stores")?
> 
> The only "PWA app store" I know them uses them is Misrosoft Store together with [pwabuilder.com](https://www.pwabuilder.com/) under the recommended section.
> 
> It's far from a priority though (unless you want to get listed on the ms store but then there are some other stuff to be done first as well if we want to use the PWA).

Fair, then I think we should merge this for now and keep the screenshots part in mind for later when we work harder on going PWA (after node server is removed) :)

VIKTORVAV99:
> > > Brilliant, thanks! 🙌 The screenshot examples makes sense in theory, are they actually used anywhere in reality (in "progressive web app stores")?
> > 
> > 
> > The only "PWA app store" I know them uses them is Misrosoft Store together with [pwabuilder.com](https://www.pwabuilder.com/) under the recommended section.
> > It's far from a priority though (unless you want to get listed on the ms store but then there are some other stuff to be done first as well if we want to use the PWA).
> 
> Fair, then I think we should merge this for now and keep the screenshots part in mind for later when we work harder on going PWA (after node server is removed) :)

Yeah that sounds like a plan. I also added a large png icon as pwa builder requires it. Maybe not strictly needed yet but can't hurt to have it there.

VIKTORVAV99:
> > > Brilliant, thanks! 🙌 The screenshot examples makes sense in theory, are they actually used anywhere in reality (in "progressive web app stores")?
> > 
> > 
> > The only "PWA app store" I know them uses them is Misrosoft Store together with [pwabuilder.com](https://www.pwabuilder.com/) under the recommended section.
> > It's far from a priority though (unless you want to get listed on the ms store but then there are some other stuff to be done first as well if we want to use the PWA).
> 
> Fair, then I think we should merge this for now and keep the screenshots part in mind for later when we work harder on going PWA (after node server is removed) :)

Might I suggest creating a public project on this repo where we can add all the nodeJS server stuff and all the stuff we should clean up afterwards?

madsnedergaard:
Good point, I created one here: https://github.com/orgs/electricitymap/projects/29/views/1?query=is%3Aopen+sort%3Aupdated-desc
Do you have permission to add to it? :)

VIKTORVAV99:
I do have access to the other you created and then closed tho 😅

VIKTORVAV99:
Thanks! Should be a little easier to organize things with that!

PR #4192, opened 2022-05-28
Update Slovakia capacities

nessie2013:


jarek (commented):
On the PDF on page 8, I'm seeing 2544 MW hydro. We have 1621 MW regular hydro and 916 MW hydro storage, which add up to 2537 MW. That's not a large difference, but do you know the current split?

I'm curious about the 1 MW hard coal (estimate for electricity grid contribution of some industrial or cogenerating station perhaps?) but that value is in the source report and it doesn't make a real difference to us, I guess :man_shrugging:

nessie2013:
> On the PDF on page 8, I'm seeing 2544 MW hydro. We have 1621 MW regular hydro and 916 MW hydro storage, which add up to 2537 MW. That's not a large difference, but do you know the current split?
> 
> I'm curious about the 1 MW hard coal (estimate for electricity grid contribution of some industrial or cogenerating station perhaps?) but that value is in the source report and it doesn't make a real difference to us, I guess 🤷‍♂️

I found more up to date info for the renewable capacities on the IRENA database which includes the hydro storage split, the 1MW hard coal in the report is indeed strange but yeah it might be co-generation or something so I included it

madsnedergaard (approved):
Looks good to me now :)

Issue #3869, opened 2022-03-16
Grammar for localizations


LaosPL:
Error message: [sprintf] expecting number but found string
Url: https://app.electricitymap.org/zone/PL

That is an error I get quite often

madsnedergaard:
Hey @LaosPL, thanks for reporting this :)
Could you provide some more context, like the following:
- Device
- OS
- Browser
- Language selected

theundefined:
Same issue here.
device - linux ubuntu 22.04
browser: Wersja 100.0.4896.127 (Oficjalna wersja) (64-bitowa)
language: pl.

After refresh sometimes is ok, but after few minutes of "clicking" i still have:
![image](https://user-images.githubusercontent.com/7404069/168030475-c0ea19ac-ed9c-4a4d-b07d-04b0bb9437b5.png)

In console there are more details:
breadcrumbs.ts:117 TypeError: [sprintf] expecting number but found string
    at sprintf.js:62:27
    at o (sprintf.js:26:16)
    at a (sprintf.js:30:24)
    at j (translation.js:9:12)
    at __ (translation.js:20:32)
    at countrypanelexchangetooltip.jsx:45:18
    at da (react-dom.production.min.js:154:98)
    at As (react-dom.production.min.js:259:456)
    at Mu (react-dom.production.min.js:230:22)
    at Ou (react-dom.production.min.js:229:362)
    at xu (react-dom.production.min.js:223:216)
    at react-dom.production.min.js:121:115
    at e.unstable_runWithPriority (scheduler.production.min.js:18:437)
    at Xi (react-dom.production.min.js:120:325)
    at Ji (react-dom.production.min.js:121:61)
    at Yi (react-dom.production.min.js:120:496)
    at lt (react-dom.production.min.js:285:209)
    at On (react-dom.production.min.js:83:253)
    at Mn (react-dom.production.min.js:84:470)
    at An (react-dom.production.min.js:84:98)
    at e.unstable_runWithPriority (scheduler.production.min.js:18:437)
    at Cn (react-dom.production.min.js:82:330)
    at HTMLDocument.sentryWrapped (helpers.ts:90:17) Object

mobuchowski:
I have pretty similar error. 

Firefox 100.0.1
Mac OS 12.04 (MBP 14' 2021)
Language PL

This generally happens after changing few zones.
```
Error message: [sprintf] expecting number but found string
Url: https://app.electricitymap.org/zone/FR
```


```
[...]

PR #4201, opened 2022-05-30
Memoizes Icon component and fixes Polish translation

madsnedergaard:
This fixes the bug in #3869 (it would still be nice to update the translations though, so leaving the issue open for now)


VIKTORVAV99 (approved):
No fuss about memo and I trust you this fixes the bug about the polish translations.

LGTM!

Issue #4202, opened 2022-05-30
Frontend missing icons


VIKTORVAV99:
While #4187 *should* fix the icons on the mobile apps it appears it broke the icons on the web when runing in production mode...

![image](https://user-images.githubusercontent.com/30777521/170959910-6e606bfe-2e67-48e6-821c-17dddaf7f61d.png)

VIKTORVAV99:
@madsnedergaard looks like we fix one platform and then breaks the next in the process. 😅
Any idea how to fix this particular issue?

madsnedergaard:
Oh, great 😅 
I guess we should start writing some more Cypress tests soon 😄

VIKTORVAV99:
> Oh, great 😅 I guess we should start writing some more Cypress tests soon 😄

Sounds like a very good idea!

PR #4187, opened 2022-05-25
Fixes icons on mobile app (in a simpler way)

madsnedergaard:
Fixes #4181 
Replaces #4184 as the extensive rework wasn't required :)

PR #4203, opened 2022-05-30
Fix frontend (web) icons

VIKTORVAV99:
While #4187 fixes the icons on mobile it introduces a CORS(?) error when loading on web.
This pr changes the loading based on platform to use direct vs relative links (Mobile vs Web respectively) to keep the same functionality on both.

fixes: #4202


madsnedergaard (approved):
Nice fix! I thought the resolvePath was smart enough to handle both, but let's try this ¯\\\_(ツ)\_/¯

PR #4204, opened 2022-05-30
[Historical view] Moves TimeSlider component into a bottom "sheet" on mobile

madsnedergaard:
### Description

This makes the new TimeSlider component work on mobile too :)
I have disabled Tabs and Legend for now, as we will revisit how these appear later on.
We'll explore putting the countrypanel in another sheet too later as well!

### Preview

Kapture 2022-05-30 at 14 08 49

### DIFF
--- a/web/src/index.jsx
+++ b/web/src/index.jsx
@@ -11,11 +11,13 @@ import { store, sagaMiddleware } from './store';
 import { cordovaApp } from './cordova';
 import sagas from './sagas';
 
-import Main from './layout/main';
+import Main from './layout/main_old';
+import MainForHistoricalView from './layout/main';
 import GlobalStyle from './globalstyle';
 
 // init styling
 import './scss/styles.scss';
+import 'react-spring-bottom-sheet/dist/style.css';
 
 // init localisation
 import './helpers/i18n';
@@ -34,6 +36,11 @@ console.log(
 // Plug in the sagas
 sagaMiddleware.run(sagas);
 
+// Switches Main component if historical view feature google is used
+const urlSearchParams = new URLSearchParams(window.location.search);
+const isHistoryFeatureEnabled = Object.fromEntries(urlSearchParams.entries())?.feature === 'history';
+const MainComponent = isHistoryFeatureEnabled ? MainForHistoricalView : Main;
+
 // Render DOM
 ReactDOM.render(
   <React.Suspense fallback={<div />}>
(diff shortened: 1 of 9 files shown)


web/src/index.jsx
```diff
@@ -34,6 +36,11 @@ console.log(
 // Plug in the sagas
 sagaMiddleware.run(sagas);
 
+// Switches Main component if historical view feature google is used
+const urlSearchParams = new URLSearchParams(window.location.search);
+const isHistoryFeatureEnabled = Object.fromEntries(urlSearchParams.entries())?.feature === 'history';
```
Kongkille, line 41: I assume we need this because it's hasn't booted up the feature flag yet. I am wondering if there is a way for us to have a central feature flag function that always works?

Or at least sort of similar to how translations work where we have one function to be used inside of react components and one outside if that makes sense.

Kongkille (approved):
Very cool, I added some reflections

web/src/index.jsx
madsnedergaard: It's because we can't use the hook outside of a component. But yeah, I guess the hook could actually use import this functionality from somewhere..
I'd prefer to address this next time we need it though :)

web/src/layout/main.jsx
```diff
@@ -131,6 +141,14 @@ const Main = ({ electricityMixMode, hasConnectionWarning }) => {
               </div>
               <LayerButtons />
             </MapContainer>
+            {/* // TODO: Get CountryPanel shown here in a separate BottomSheet behind the other one */}
+            {isMobile ? (
+              <StyledBottomSheet open snapPoints={() => [60, 160]} blocking={false}>
```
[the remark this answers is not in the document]
madsnedergaard: For now the snappoints are just to match the content, will probably change when moving country panel into a sheet. I also think you have a good idea here, I'll do that when working on this again.

web/src/scss/layout/_left-panel.scss
```diff
@@ -1,9 +1,27 @@
...
+.left-panel div::-webkit-scrollbar-track-piece {
+    background: $light-gray;
+    border-radius: 3px;
+  }
+  
+.left-panel div::-webkit-scrollbar-thumb {
+    background: lightgray;
+    border-radius: 3px;
+  }
+
```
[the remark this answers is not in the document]
madsnedergaard: Good point, done.

web/src/index.jsx
madsnedergaard: The hook is relying on [useLocation from react-router](https://v5.reactrouter.com/web/api/Hooks/uselocation), so rewriting this will actually be a bit tricky I think..

PR #4206, opened 2022-05-31
Simplify Icons

VIKTORVAV99:
This brings over some of the non fix changes in #4184.

In this PR:
- Use `size` property to set height and width instead of CSS.
- Use `color` property to set fill color instead of CSS.
- Remove styled components from the `<Icon />` component as width and height can be set directly in the svg element.
- Include electricitymap icon in the icon sprite.
- Minimized and sorted the icon sprite.
- Renamed the icon sprite from `material-icon-sprite.svg` to `icon-sprite.svg` as it now contains non material icons as well.
- Change the image element in mobile footer to `<Icon />` for the electricityMap icon.
- Cleans up all unneeded an unused CSS related to the changes above.


web/src/components/icon.jsx
```diff
@@ -1,16 +1,10 @@
...
-`;
-
-const path = window.isCordova ? resolvePath('images/material-icon-sprite.svg') : '/images/material-icon-sprite.svg';
+const path = window.isCordova ? resolvePath('images/icon-sprite.svg') : '/images/icon-sprite.svg';
 
 /**
  * To add a new icon to the SVG sprite, open the icons SVG file and copy the the <svg> element.
- * Then paste the the <svg> element and rename it to <symbol> in the file 'material-icon-sprite.svg' and give it a unique id.
+ * Then paste the the <svg> element and rename it to <symbol> in the file 'icon-sprite.svg' and give it a unique id.
  *
```
madsnedergaard, line 8: I wanted to move the `weather/wind` icons in here as well, but had some challenges due to the viewBox size - how should one go about resizing icons to match this new unified size? Just dropping it in here didn't work and it was rendered outside of the svg

web/src/components/icon.jsx
```diff
@@ -19,12 +13,12 @@ const path = window.isCordova ? resolvePath('images/material-icon-sprite.svg') :
  *   </symbol>
  */
 
-const Icon = ({ iconName }) => {
+const Icon = memo(({ iconName, size = 24, color }) => {
   return (
-    <IconContainer>
+    <svg height={size} width={size} fill={color}>
```
madsnedergaard, line 18: Nice, and overriding with css (e.g. on `hover`) will still have higher specificity, right?

madsnedergaard (commented):
Really great! 🙌 
A few minor questions/comments

web/src/components/icon.jsx
VIKTORVAV99: I assume you mean these icons?
image
They are a little bit special as they are injecting the image/icon as a CSS background-image so I think they need a little extra CSS to work properly. But I'll look into it.

Another option would be trying to switch them to using the `<Icon />` component directly, that should make them render just fine.

web/src/components/icon.jsx
madsnedergaard: Ah great! 🥳

web/src/components/icon.jsx
madsnedergaard: I'm addressing it in an upcoming PR, so it was mostly to hear your thoughts on how to resize svgs so they match in the sprite :)

VIKTORVAV99: Oh then I would probably just download the icons from https://fonts.google.com/icons?icon.query=sun&icon.style=Outlined&icon.set=Material+Icons with the correct default size like I did the others and just copy paste the code. I ran the others through https://jakearchibald.github.io/svgomg/ first but that isn't needed (need to be careful with those settings though)

madsnedergaard: Yes, exactly. Will do that for Material icons and then probably replace the ones that are slightly different (wind and "bright mode"/contrast icon) :)

madsnedergaard (approved):

web/src/components/icon.jsx
VIKTORVAV99: For any other icons/svg it should be enough to adjust the viewbox and hight/widht (or remove them as they are not needed with a viewbox) unless they have some inline styling.

PR #4211, opened 2022-05-31
[historical view] Enable country panel for other time aggregates part 4

Kongkille:
_Since #4207 got quite big, I am splitting it out into some different milestone PRs on a temporary main branch_

**Important - Requires #4210  to be merged beforehand!**
### Description - Adjust graphs to production data

This PR updates the fetcher to call and use V5 history data and adjusts the graphs accordingly. So far this has only been tested on hourly and daily data. This should be the final PR for the temp_main branch.

(Note: there are some missing values in the tooltip and other places, which I have some pending changes on the app-backend to fix)

I had to slighly go into the datareducer, but just barely and there's no optimizations on how much it will call the backend, as we'll address that separately.

### Preview

https://user-images.githubusercontent.com/45588799/171193759-61672924-ddee-4010-848c-347aae48e5f7.mov


web/src/reducers/dataReducerForHistoryFeature.js
```diff
@@ -183,10 +183,11 @@ const reducer = (state = initialDataState, action) => {
         isLoadingHistories: false,
         histories: {
           ...state.histories,
-          [action.zoneId]: action.payload.map((datapoint) => ({
+          [action.zoneId]: action.payload.zoneStates.map((datapoint) => ({
```
madsnedergaard, line 186: Dunno if I missed this on the `app-backend`, but did we change the response? :)

web/src/layout/timeController.jsx
```diff
@@ -6,13 +6,16 @@ import { useCurrentDatetimes } from '../hooks/redux';
...
+import { TIME_TO_RANGE } from '../helpers/constants';
 
 const handleZoneTimeIndexChange = (timeIndex) => {
   dispatchApplication('selectedZoneTimeIndex', timeIndex);
 };
 
 const handleTimeAggregationChange = (aggregate) => {
   dispatchApplication('selectedTimeAggregate', aggregate);
+  // set index to the max of the range of selected time aggregate
+  dispatchApplication('selectedZoneTimeIndex', TIME_TO_RANGE[aggregate.toUpperCase()] - 1);
```
madsnedergaard, line 18: Nice! :)

web/src/reducers/dataReducerForHistoryFeature.js
madsnedergaard: Also, should the mockserver be updated to match this?

madsnedergaard (commented):
One minor question :)

web/src/reducers/dataReducerForHistoryFeature.js
Kongkille: The mockserver data also serves the zonesStates for history

Kongkille:
@madsnedergaard ok so i've already found a few bugs, but I'll merge this one for now so we can merge this temp_branch

PR #4214, opened 2022-06-02
Virtualize zonelist

VIKTORVAV99:
This PR uses [react-window](https://github.com/bvaughn/react-window) to virtualize the zone list reducing the number of dom elements on the zone list page from 1221 elements to 312 elements when using mock data.
> **Note**
> This will differ based on viewport size.

The reason for doing this is so that style calculations will be faster and to avoid [layout reflows](https://developers.google.com/speed/docs/insights/browser-reflow).

> **Note**
> This will have a even bigger impact when using production data as there are even more zones in the list.

This PR also includes some codes changes around this feature and some general code cleanup.


web/src/reducers/dataReducerForHistoryFeature.js
```diff
@@ -19,7 +19,6 @@ Object.entries(zonesConfig).forEach((d) => {
   zone.capacity = zoneConfig.capacity;
   zone.contributors = zoneConfig.contributors;
   zone.timezone = zoneConfig.timezone;
-  zone.shortname = translation.getZoneNameWithCountry(key);
```
VIKTORVAV99, line 22: As far as I can tell this was only used in `zonelist.jsx` and was always identical to the countryKey so I removed it in favor of using the countryKey.

madsnedergaard (approved):
Nice! 🥳 

Before: 
Screenshot 2022-06-07 at 15 14 21

After:
Screenshot 2022-06-07 at 15 14 33

PR #4215, opened 2022-06-02
Move estimated data info box

VIKTORVAV99:
This PR moves the estimated data info box to the bottom of the screen (above source and contributors) as discussed in #4122 and #3807.


madsnedergaard (approved):
Thanks for fixing it ❤️

PR #4218, opened 2022-06-03
Adds legal links to FAQ section

madsnedergaard:
### Description

Adds links to legal notice and privacy policy in the FAQ section.
I added some shadow (on desktop) in an attempt to make it clearer that the FAQ section above is scrollable :)

### Preview

**Desktop**
Screenshot 2022-06-03 at 14 10 55

**Mobile**
Screenshot 2022-06-03 at 14 01 28


VIKTORVAV99:
Looks great IMO, the only problem I have (not related to anything in this PR) is that it's pretty hard to notice we have a FAQ on desktop as it just looks like a normal link. We should probably try and find a way to fix that, maybe include it in the top header in some way?

web/src/components/faq.jsx
```diff
@@ -86,24 +101,30 @@ const FAQ = ({ className }) => {
...
+                />
+              ))}
+            </div>
+          ))}
+        </div>
       </div>
-    </div>
+      <TermsAndPrivacyContainer>
+        <a href="https://www.electricitymap.org/privacy-policy/">{__('misc.privacyPolicy')}</a>
+        <a href="https://www.electricitymap.org/legal-notice/">{__('misc.legalNotice')}</a>
```
VIKTORVAV99, line 125: Would it make sense to add icons like these here?
image

Kongkille:
> Looks great IMO, the only problem I have (not related to anything in this PR) is that it's pretty hard to notice we have a FAQ on desktop as it just looks like a normal link. We should probably try and find a way to fix that, maybe include it in the top header in some way?

Agreed. We're restructuring some things on mobile but we could also use some of this for desktop, potentially making it a bit clearer.

image

Kongkille (approved):

VIKTORVAV99:
> > Looks great IMO, the only problem I have (not related to anything in this PR) is that it's pretty hard to notice we have a FAQ on desktop as it just looks like a normal link. We should probably try and find a way to fix that, maybe include it in the top header in some way?
> 
> Agreed. We're restructuring some things on mobile but we could also use some of this for desktop, potentially making it a bit clearer.
> 
> image
> 

Nice! These look very clean and modern, a substantial improvement if you ask me.

My only real issue with the current navbar is that it doesn't really do anything in the app (all links are to electricitymap.org), while I understand it's part of the branding it is a bit counter intuitive for users experience, at least in my opinion.

PR #4221, opened 2022-06-03
Add react-is and babel-plugin-styled-components

VIKTORVAV99:
This PR adds react-is as it was a missing peer dependency of styled components as well as babel-plugin-styled-components as recommended by styled-components for a better developer experience and code minimization.


madsnedergaard (approved):
👏

PR #4228, opened 2022-06-09
[historical view] Creates new shared modal component

madsnedergaard:
And uses it for the onboarding modal.

Will be reused for Settings modal in an upcoming PR :)


web/src/scss/components/_modal.scss
```diff
@@ -1,273 +1,250 @@
-.modal-background-overlay{
```
madsnedergaard, line 1: See https://github.com/electricitymap/electricitymap-contrib/pull/4228/commits/ea0faf756e6686e2c2e79373a46ad2b9f8b09e90 for the actual changes :)

web/src/components/modal.jsx
```diff
@@ -1,10 +1,20 @@
...
 import { useTranslation } from '../helpers/translation';
 
 import Icon from './icon';
 
-const Modal = ({ views, visible, onModalShown, onDismiss }) => {
+const ModalTitle = styled.h2`
+  color: #000;
+  text-align: center;
+  font-weight: bold;
+  font-family: 'Euclid Triangle', 'Open Sans', sans-serif;
```
Kongkille, line 12: We could also perhaps define this more globally?

Kongkille (approved):
Nice that we're cleaning up some stuff :)

PR #4229, opened 2022-06-09
Fix update contrib 

Kongkille:
We use the zonemap elsewhere as well where we can't have the redux running, so I am passing it as a prop instead.


madsnedergaard (approved):
Nice that you fixed this, I was about to duplicate the damn file so we don't have this problem all the time :D

Issue #4183, opened 2022-05-25
Åland wind capacity data source missing


q--:
- Right now Åland wind production is 36.8 MW, 222.22% of the configured installed capacity of 16.6 MW. This needs to be updated with whatever the current installed capacity is. 
  - This was ***partially*** fixed in https://github.com/electricitymap/electricitymap-contrib/pull/4234 by adding a new wind farm to the current capacity, but without a complete data source we don't know if we're still missing some. Oil and interconnectors are now correct.
- `DATA_SOURCES.md` doesn't have a a capacity data source for Åland. I looked in the various Åland-related PRs/commits/issues and didn't notice any either. ~~The [production map](http://www.kraftnat.ax/text2.con?iPage=28&iLan=1) has wind power production as a percentage of (presumably) installed capacity though, so if rounding errors aren't important that could be used to derive the installed capacity.~~ As [VIKTORVAV99](https://github.com/VIKTORVAV99) points out, it's probably the share of total production+imports, not capacity; so, not useful for this purpose.
  - Oil and interconnectors got a data source in https://github.com/electricitymap/electricitymap-contrib/pull/4234, wind still needs one.

Don't have time to look any further right now, so opening this issue with what I know now.

**Edit:** possible wind data source: https://tuulivoimayhdistys.fi/en/wind-power-in-finland/map

VIKTORVAV99:
> The [production map](http://www.kraftnat.ax/text2.con?iPage=28&iLan=1) has wind power production as a percentage of (presumably) installed capacity though, so if rounding errors aren't important that could be used to derive the installed capacity.

I don't think it is percentage of installed capacity but rather percentage of current consumption including imports.

If you take the total consumption (total förbrukning) and multiply it with the percentage you get (almost) the exact figure for wind.

I guess we could use the highest recorded wind production as capacity if we can't find a source.

jvarho:
I don't have a source for the current total capacity, but the increase is from a recently commissioned 42 MW wind park, as described at https://kraftnat.ax/

I think it would be more or less correct to add that capacity to the current number.

VIKTORVAV99:
I think the best option for now is to increase the current wind capacity by 42 megawatt as you suggest @jvarho.

But I think this issue should remain open until we can find a good capacity source for Åland wind (and other) production capacities.

q--:
[...]

PR #4234, opened 2022-06-11
Åland (AX) capacity update

VIKTORVAV99:
This PR updates Åland (AX) production capacities for wind and oil as well as the exchange capacities between Åland and Finland.

Reference: #4183

EDIT: This PR also changes the source domain of the AX parser to the new domain (kraftnat.ax). It was already used in `DATA_SOURCES.md` but leads to a 404 page there, that should be updated when the changes for #4235 are implemented.


madsnedergaard (approved):
Great! 🙌

PR #4238, opened 2022-06-11
[BUGFIX] Fix "of installed capacity" on production tooltip

VIKTORVAV99:
Super small PR to fix the production tooltip.

Before:
image

After:
image


madsnedergaard (approved):
Good catch! :D

PR #4233, opened 2022-06-10
[historical view] Implements a Settings modal

madsnedergaard:
### Description

Yet another modal! 😄 
This time, it's one for settings where we plan to move all the buttons currently covering the map despite rarely being used.

I'm not too happy about the button states (e.g. active, not active, disabled, etc.) as it's hard to understand. We talked about using a "switch" instead to make it more clear, but I decided not to do it in this PR and instead focus on functionality first :)

_depends on #4232 and #4228_

### Preview

Screenshot 2022-06-11 at 00 29 52

PR #4250, opened 2022-06-14
[historical view] Map buttons for info and settings modal

madsnedergaard:
### Description

Removes layer buttons (zoom and settings), and replaces by two buttons: one for opening an info modal and one for a settings modal (see PRs below)

_Depends on #4228, #4232 & #4233_

### Preview

Screenshot 2022-06-14 at 16 10 47


madsnedergaard:
Once this is merged in (after #4233) I'll create an e2e test for it that goes through the different modals :)

web/src/layout/main.jsx
```diff
@@ -62,6 +63,12 @@ const NewVersionButton = styled.button`
   cursor: pointer;
 `;
 
+const ToggleWrapper = styled.div`
```
Kongkille, line 66: Creative idea: instead of calling them wrappers we could call them something like 'HiddenOnMobile' so it's more easily readable

web/src/layout/main_old.jsx
```diff
@@ -163,6 +167,9 @@ const Main = ({ electricityMixMode, hasConnectionWarning }) => {
         <Tabs />
       </div>
       <OnboardingModal />
+      <InfoModal />
+      <FAQModal />
+      <SettingsModal />
```
Kongkille, line 172: Any thoughts for a 'better' structure to incorporate any number of modals?

Kongkille (approved):

Kongkille:
So what's the plan for merging this with the changes on 'historical-view-v1'? I think the earlier the better

madsnedergaard:
> So what's the plan for merging this with the changes on 'historical-view-v1'? I think the earlier the better

I'll merge this to master, then rebase the new branch afterwards :)

web/src/layout/main_old.jsx
madsnedergaard: We could do something more fancy, but don't think it's worth trying to make it too fancy and abstract 😄

web/src/layout/main.jsx
Kongkille: Yeah I just remember Material UI had some functionality like this but it seems it is not deprecated

https://mui.com/material-ui/migration/v5-component-changes/#hidden

madsnedergaard: Ah, I see - I guess that could make sense, hope we can simplify some of these "mobile" vs "desktop" things later on :)

PR #4256, opened 2022-06-16
Drop `electricitymap.contrib` from import (resolves isort issues)

VIKTORVAV99:
@madsnedergaard I am creating this minimal PR mostly so we have a space to talk without cluttering something else up.

I had some issues with isort in #3936 was told to ask you about it.
Iracked down the issue to the `electricitymap.contrib.parsers.lib.parsers` import in `test_parsers.py`.
It seems like isort is recognizing this as a third party import on my system and groups it with the other third party imports but it don't do this on the CI.

This makes isort pass locally but then it fails on the CI.
I'm not really sure why it does this but I suspect it might be OS related since I'm on Windows and the CI runs on Linux (Ubuntu). (I suspect it's a issue with symbolic links.)

Now the changes in this PR seems to stop that from happening so I'm just wondering if it's safe to merge or if I need to make any other changes.


madsnedergaard (approved):
Hmm good question, I can't remember why/how Python/Poetry modules are done this way - but since we have the `.quality` import below in the other format, I can't see how this could hurt anything 😅

VIKTORVAV99:
> Hmm good question, I can't remember why/how Python/Poetry modules are done this way - but since we have the `.quality` import below in the other format, I can't see how this could hurt anything 😅 

I'm going to merge this then and if everything suddenly breaks you know who to blame. 😅

PR #4264, opened 2022-06-17
[historical-view] use history as default

Kongkille:
This removes all things related to the history flag and unused files


web/src/components/timeslider.jsx
```diff
@@ -1,100 +0,0 @@
-import React, { useMemo, useState } from 'react';
```
madsnedergaard, line 1: Could we use `timeslider` as the real one instead of `timeslider_new`? :D

madsnedergaard (commented):
One minor thing, otherwise great!

PR #4267, opened 2022-06-20
[E2E] Tests map, country panel and ranking panel

madsnedergaard:
### Description

More tests!

- Tests different functionality of the map (e.g. pressing all map-related buttons)
- Tests country panel and the timeslider
- Tests searching in ranking panel

### Preview
_(gif of the map test as example)_
Kapture 2022-06-20 at 07 47 40


web/jsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
madsnedergaard, line 1: This enables VSCode IntelliSense for Cypress:

Screenshot 2022-06-20 at 07 54 01

web/cypress/integration/map.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.contains('English').click();
+    cy.contains('consumption');
+
+    // test layers
+    cy.get('[href="/map?wind=true"]').click();
+    cy.contains('Wind power potential');
+    // TODO: For some reason the link is not "mounted in DOM" until after a while...
+    // Dunno how to solve it without avoid adding this delay
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(200);
```
Kongkille, line 36: Well there may be some ways to wait specifically for it or fix the original code, but I am not sure if it's worth spending the time.

We can see if this causes any issues and then change it if that's the case

web/cypress/integration/map.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.get('a[href*="solar=true"]').click();
+    cy.contains('Solar power potential');
+
+    // test dark mode
+    cy.get('[aria-label="Toggle dark-mode"]').click().click();
```
Kongkille, line 41: Maybe we should get a test here, for example that would have actually caught my comment on 

https://github.com/electricitymap/electricitymap-contrib/pull/4233#pullrequestreview-1012371645

Kongkille (commented):
Really nice to get some more tests in

We might not want to rely too much on classNames or IDs as we are likely going to toss that with styled components as we go through the code. 

This will of course cause the test to fail so we'll fix it, but we could future-proof the tests by using data-test-id in all places we want cypress to listen

madsnedergaard:
> Really nice to get some more tests in
> 
> We might not want to rely too much on classNames or IDs as we are likely going to toss that with styled components as we go through the code.
> 
> This will of course cause the test to fail so we'll fix it, but we could future-proof the tests by using data-test-id in all places we want cypress to listen

Yeah, ideally we shouldn't rely on classnames and ids - but I'd personally rather fix that as we convert things instead of adding a bunch of data attributes to elements now. WDYT? At least some of the things tested here are going to change with historical view too :)

web/cypress/integration/map.spec.js
Kongkille: How about instead of checking the map colors we check the color of the countryPanel carbon intensity square? Comparing that to before and after.

Kongkille (approved):

PR #4268, opened 2022-06-20
Added tests for validators

FelixDQ:
I also added a few misc changes
* I added pytest 
* I inverted the expected return values from the validators so that 1=valid and 0=invalid
* Added tests for the current sanity check validators


tests/validators/lib/load_fixture.py
```diff
@@ -0,0 +1,10 @@
+import os
```
pierresegonne, line 1: why not name this file `fixture.py`?

pierresegonne: or `fixtures.py`?

tests/validators/test_sanity_checks.py
```diff
@@ -0,0 +1,24 @@
...
+    validate_production_one_non_nan_value,
+)
+
+from .lib.load_fixture import load_fixture
+
+
+def test_validate_positive_production():
+    events = load_fixture("production_negative_values")
+    res = validate_positive_production(events)
+    assert res["2022-01-01 00:00:00+00:00"] == 1
```
pierresegonne, line 14: could simplify by checking that the column is [1, 0, 1]

pierresegonne (approved):

scripts/tooling.py
```diff
@@ -32,8 +32,7 @@ def lint():
 
 
 def test():
-    _run("python -u -m unittest discover tests")
-    _run("python -u -m unittest discover parsers/test")
+    _run("pytest")
```
pierresegonne, line 35: Ideally we would have added pytest in a separate PR.

Does that still cover all of the other tests?

FelixDQ: > Ideally we would have added pytest in a separate PR.

Yea, I know 🙈

>Does that still cover all of the other tests?

Yup!

tests/validators/test_sanity_checks.py
FelixDQ: 👍 I'll keep it in mind for other tests. I think for these I'll keep the multiple asserts as they are a bit more readable.

PR #4269, opened 2022-06-20
Added all old quality checks to new validator system

FelixDQ:
This adds all rules from https://github.com/electricitymap/electricitymap-contrib/blob/master/parsers/lib/quality.py to the validators. All of the new validators have unit tests as well.

_Sorry about the messy commit log. I based this commit from my previous branch_


tests/validators/test_sanity_checks.py
```diff
@@ -22,3 +28,45 @@ def test_validate_production_one_non_nan_value():
...
+
+def test_validate_exchange_netflow_is_plausible():
+    events = load_fixture("exchange_with_unplausible_netflows")
+    res = validate_exchange_netflow_is_plausible(events)
+    assert res["2022-01-01 00:00:00+00:00"] == 0
+    assert res["2022-01-01 01:00:00+00:00"] == 0
+    assert res["2022-01-01 02:00:00+00:00"] == 1
+
+
+def test_validate_exchange_netflow_doesnt_exceed_capacity():
```
pierresegonne, line 65: Why can't this work with a fixture as well? Is it because of the need to specify by hand the capacity?

pierresegonne (approved):
Brilliant

validators/zone_specific_checks.py
```diff
@@ -0,0 +1,43 @@
...
+import pandas as pd
+
+from electricitymap.contrib.validators.lib.config import validator
+
+
+@validator(
+    kind="production",
+    not_zone_keys=[  # zone_keys where we allow fossil fuel production to be 0
+        "CH",
+        "NO",
```
VIKTORVAV99, line 12: Should these include the NO subzones as well? (NO-NO1, NO-NO2, etc)

validators/zone_specific_checks.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime
+
+import pandas as pd
+
+from electricitymap.contrib.validators.lib.config import validator
+
+
+@validator(
+    kind="production",
+    not_zone_keys=[  # zone_keys where we allow fossil fuel production to be 0
```
VIKTORVAV99, line 10: This array should also include SE as it don't have any fossil fuel production for the vast majority (almost all) of the time.

pierresegonne:
Hey @VIKTORVAV99 thanks for the review :)
For now Felix has just adapted the validation checks we had internally and that were hidden from the world. We're building a parser event validation engine where all the rules will live in here. We'll communicate soon about it, and the possibility for contributors to suggest new validation rules.

So for now, and I hope it's ok with you, I'll ignore your suggestions and merge this as is, as we're still at the stage where we're piecing it all together and don't want to introduce potential changes vs what we have today.

VIKTORVAV99:
> Hey @VIKTORVAV99 thanks for the review :)
> For now Felix has just adapted the validation checks we had internally and that were hidden from the world. We're building a parser event validation engine where all the rules will live in here. We'll communicate soon about it, and the possibility for contributors to suggest new validation rules.
> 
> So for now, and I hope it's ok with you, I'll ignore your suggestions and merge this as is, as we're still at the stage where we're piecing it all together and don't want to introduce potential changes vs what we have today. 

All good! 👍🏻

I'll submit a PR with the relevant changes once everything is done and ready!

PR #4271, opened 2022-06-20
[BUGFIX] `arrow_back` background fix

VIKTORVAV99:
It just removes the colored background for the back arrow that was accidentally introduced in #4232

BEFORE:
image

AFTER:
image


madsnedergaard:
Thanks for fixing this - I actually already fixed it in https://github.com/electricitymap/electricitymap-contrib/pull/4233#discussion_r900185460, just haven't released it yet :)

PR #4274, opened 2022-06-21
[Historical-view] Enable exchanges

Kongkille:
### Description

Enables exchanges on the following conditions
- Time is 'HOURLY'
- TimeIndex is 24, eg it is latest data
- Consumption

In the future, we'd also want to incorporate more historical exchanges, but I think this is the most appropriate solution in the meantime.

### Preview
https://user-images.githubusercontent.com/45588799/174770408-76a707eb-a5df-4e44-bb08-3f3f7793ad07.mov

### DIFF
--- a/web/src/hooks/layers.js
+++ b/web/src/hooks/layers.js
@@ -4,14 +4,20 @@ import { interpolate } from 'd3-interpolate';
 import { formatDistance } from 'date-fns';
 
 import { getRefTime, getTargetTime } from '../helpers/grib';
+import { TIME } from '../helpers/constants';
 
 export function useExchangeArrowsData() {
   const isConsumption = useSelector((state) => state.application.electricityMixMode === 'consumption');
+  const isHourly = useSelector((state) => state.application.selectedTimeAggregate === TIME.HOURLY);
+  const isLatestTimeIndex = useSelector((state) => state.application.selectedZoneTimeIndex === 24);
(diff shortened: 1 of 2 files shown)


web/src/hooks/layers.js
```diff
@@ -4,14 +4,20 @@ import { interpolate } from 'd3-interpolate';
 import { formatDistance } from 'date-fns';
 
 import { getRefTime, getTargetTime } from '../helpers/grib';
+import { TIME } from '../helpers/constants';
 
 export function useExchangeArrowsData() {
   const isConsumption = useSelector((state) => state.application.electricityMixMode === 'consumption');
+  const isHourly = useSelector((state) => state.application.selectedTimeAggregate === TIME.HOURLY);
```
madsnedergaard, line 11: Wondering if it's matters if the equality check happens inside the useSelector or outside (like below)? 🤔 
Just thinking it might be that the variables can't be trusted in useMemo below due to this, but I have no clue...
If not, then just ignore this!
```suggestion
  const isHourly = useSelector((state) => state.application.selectedTimeAggregate) === TIME.HOURLY;
```

madsnedergaard (approved):
Nice! Again only minor comment that I don't know if matters, so I'll approve anyway :)

madsnedergaard:
Failing tests are fixed in #4276, so once that is merged you can rebase this branch.

PR #4276, opened 2022-06-21
Fixes tests

madsnedergaard:
### Description

- Rebases `master`
- Renames some hide-on-mobile wrappers as suggested in https://github.com/electricitymap/electricitymap-contrib/pull/4250#discussion_r902418031


web/src/reducers/dataReducer.js
```diff
@@ -40,7 +40,7 @@ const reducer = createReducer(initialState, (builder) => {
         }
         // Assign all data
         Object.keys(value).forEach((k) => {
-          state.exchanges[k] = value[k];
+          exchange[k] = value[k];
```
Kongkille, line 43: I'm not sure this is correct, but I fix it later in my PR regardless

Kongkille (approved):

Kongkille:
Do you want any help on fixing the tests?

web/src/reducers/dataReducer.js
madsnedergaard: It is, I actually screwed up on the refactor we did together - it should be `state.exchanges[key][k] = value[k]` and not `state.exchanges[k] = value[k]` :)

PR #4281, opened 2022-06-22
Cover edge case for validate_exchange_netflow_doesnt_exceed_capacity

pierresegonne:
### Description

Covers the edge case where the exchange has no declared capacity for `validate_exchange_netflow_doesnt_exceed_capacity`


madsnedergaard (approved):
Makes sense :)

FelixDQ:
Just a thought, but we could also have the validators return `None` when they don't have the information required to do validations. This would allow the backend to know the difference between "OK" and "We don't actually know".

pierresegonne:
Good point, that would make it easier for us to monitor in the backend - let's take it at the next iteration of the validators :) 
Else feel free to just make the change

PR #4283, opened 2022-06-22
[Historical view] Adds more e2e testing of the new feature

madsnedergaard:
### Description

Covers the timeslider component in tests!
Also moves testing of API requests here, while removing the requirement of the mockserver by using files directly as fixtures.

_Depends on #4282_


VIKTORVAV99:
After this is merged I thought we could migrate to [Cypress 10](https://www.cypress.io/blog/2022/06/01/cypress-10-release/) so we can start using component testing with cypress as well. (not a requirement but would be nice to have them combined in one tool)

We could do this right away but there are some migration steps that will cause merge conflicts with this PR so I think it would be best for everyone if I just waited until this PR is merged before I upgrade it Cypress.

madsnedergaard:
> After this is merged I thought we could migrate to [Cypress 10](https://www.cypress.io/blog/2022/06/01/cypress-10-release/) so we can start using component testing with cypress as well. (not a requirement but would be nice to have them combined in one tool)
> 
> We could do this right away but there are some migration steps that will cause merge conflicts with this PR so I think it would be best for everyone if I just waited until this PR is merged before I upgrade it Cypress.

Yeah, I tried it too - migration is pretty simple (but moves files and will cause some conflicts), I personally found using it a bit more annoying to use (additional clicks required to run tests, but maybe that can be scripted) :)

VIKTORVAV99:
> > After this is merged I thought we could migrate to [Cypress 10](https://www.cypress.io/blog/2022/06/01/cypress-10-release/) so we can start using component testing with cypress as well. (not a requirement but would be nice to have them combined in one tool)
> > We could do this right away but there are some migration steps that will cause merge conflicts with this PR so I think it would be best for everyone if I just waited until this PR is merged before I upgrade it Cypress.
> 
> Yeah, I tried it too - migration is pretty simple (but moves files and will cause some conflicts), I personally found using it a bit more annoying to use (additional clicks required to run tests, but maybe that can be scripted) :)

I just usually run the tests with the CLI so no more steps there.
(just use `yarn run cypress-tests` when the mockserver and snowpack is running)

Didn't even know there was a GUI until I started looking into upgrading it to v10. 😅

In either case I think the benefits outweigh the negativ sides of upgrading.

web/package.json
```diff
@@ -133,7 +133,8 @@
     "verify": "VERIFY_NO_UPDATES=1 yarn update-world",
     "mockserver": "cd ../mockserver && node server.js",
     "prepare": "cd .. && husky install web/.husky",
-    "cypress-test": "cypress run",
+    "cypress-test": "TZ=Europe/Copenhagen cypress run",
+    "cypress-open": "TZ=Europe/Copenhagen cypress open",
```
VIKTORVAV99, line 137: Hmm, could we avoid setting the environmental variable here?
It won't work on Windows and will make it harder for people to contribute.

madsnedergaard: Hmm, I guess we could remove any tests related to time-sensitive data, but it's really annoying that it's not easier setting a specific timezone for Cypress :(
(See https://github.com/cypress-io/cypress/issues/1043)

madsnedergaard: An alternative is mocking `new Date()` with https://docs.cypress.io/api/commands/clock, but I tried it without success and it also feels like this should be fixed on a higher level if possible :)

VIKTORVAV99: I haven't really tried this at all yet but some quick searches led me to find this.

Maybe we can set the ENV variables like this:
```js
process.env['NODE_ENV'] = 'production';
```

It's completely backwards but it might work?

Kongkille (approved):

web/package.json
madsnedergaard: I'll address this in a new PR to avoid messing up this one further 😄

PR #4291, opened 2022-06-26
Bump prettier, eslint and webpack development dependencies

VIKTORVAV99:
As the title says this bumps the minor and patch versions of the prettier, eslint and webpack development dependencies but it also updates the prettier ignore files so lint-staged + prettier ignores on `web/yarn.lock` as well (will fail otherwise as prettier don't recognize the format of `yarn.lock` but lint-staged passes it directly to prettier).


madsnedergaard (approved):
Nice, makes sense :)

PR #4290, opened 2022-06-26
Update zones.json

gopikrishna1793:
Updated installed capacity for India as of May 22.

PR #4292, opened 2022-06-26
Update DATA_SOURCES.md

gopikrishna1793:
Data source added for India.


VIKTORVAV99:
If you could move these changes into #4290 that would be great!

PR #4295, opened 2022-06-27
Changes graphs to no update global "selected time" when hovering

madsnedergaard:
### Description

This ensure that when hovering over graphs, we show correct data for the selected "time" on the x-axis, but we don't update globally the selected time any longer. 
This felt too unnatural with the new detached time slider :) 

### Preview
(I guess you cannot really see the timeslider here, but trust me it works 😝) 
Kapture 2022-06-27 at 09 34 13


Kongkille (approved):
What are your thoughts on having all these mouse handlers as well as the graphindex state inside of the areagraph component. 

It feels like we're maintaining 4+ graphs with very similar functionality?

Kongkille (approved):
Nice, it's a much smoother experience IMO.

Would it be difficult, you think, to still show the vertical marker (without tooltip) when the user is dragging on the timeController? Similar to the functionality on production.

I think a user can quickly get a bit lost when scrolling through time and not knowing exactly where they are on the graphs

madsnedergaard:
> Nice, it's a much smoother experience IMO.
> 
> Would it be difficult, you think, to still show the vertical marker (without tooltip) when the user is dragging on the timeController? Similar to the functionality on production.
> 
> I think a user can quickly get a bit lost when scrolling through time and not knowing exactly where they are on the graphs

I was about to say that it's too much work, but then I tried at it was really simple 😄

Kongkille (approved):

PR #4298, opened 2022-06-27
Add neighbors from subzones to parent zone

FelixDQ:
From https://github.com/electricitymap/electricitymap-contrib/pull/4177

Needed to compute spatial aggregates.

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -29,6 +29,13 @@ for zone_id, zone_config in ZONES_CONFIG.items():
     if "bounding_box" in zone_config:
         ZONE_BOUNDING_BOXES[zone_id] = zone_config["bounding_box"]
 
+# Add link from subzone to the full zone
+ZONE_PARENT: Dict[ZoneKey, ZoneKey] = {}
+for zone_id, zone_config in ZONES_CONFIG.items():
+    if "subZoneNames" in zone_config:
+        for sub_zone_id in zone_config["subZoneNames"]:
+            ZONE_PARENT[sub_zone_id] = zone_id
+
 # Prepare zone neighbours
 ZONE_NEIGHBOURS: Dict[ZoneKey, List[ZoneKey]] = {}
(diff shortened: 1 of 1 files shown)


electricitymap/contrib/config/__init__.py
```diff
@@ -38,6 +45,26 @@
...
         ZONE_NEIGHBOURS[zone_name_1].add(zone_name_2)
+
+# Find neighbors to subzones and add them to parent zone.
+for zone in ZONES_CONFIG.keys():
+    subzones = ZONES_CONFIG[zone].get("subZoneNames", [])
+    if not subzones:
+        continue
+
+    for subzone in subzones:
+        for subzone_neighbor in ZONE_NEIGHBOURS[subzone]:
```
pierresegonne, line 56: isn't there a risk of iterating on `None`? 
Better to do `ZONE_NEIGHBOURS.get(subzone, [])`?

electricitymap/contrib/config/__init__.py
```diff
@@ -38,6 +45,26 @@
...
+            if subzone_neighbor in subzones:
+                # ignore the neighbours that are within the zone
+                continue
+
+            if zone not in ZONE_NEIGHBOURS:
+                ZONE_NEIGHBOURS[zone] = set()
+
+            # If neighbor zone is a subzone, we add the exchange to the parent zone
```
pierresegonne, line 64: Image having `A-1`, `A-2`, `A-3`, `B-1`, `B-2`, `C` zones with exchanges between `A-1->A-2`, `A-2->B-1` and `A-2->C`. What would be the resulting neighbour relations?

FelixDQ: `A->B` and `A->C`

electricitymap/contrib/config/__init__.py
FelixDQ: Yes, I think you are right. I changed it

pierresegonne (approved):
Nice

PR #4300, opened 2022-06-28
Ensures sync between history and state

Kongkille:
We want to ensure history and state is synced

They can become out of sync if:

1. State request fails
2. History request fails
3. State fetches newer data than stored state data
4. History fetches newer data than stored history data
5. Either file is corrupt, which can occur if
    1. State and details has a different number of entries
    2. Datetime entries in state or details does not match indexes

For now, I'll focus on 1-4 in frontend. Then ideally the app-backend would ensure the files are never corrupt.

### DIFF
--- a/web/src/components/timeslider.jsx
+++ b/web/src/components/timeslider.jsx
@@ -9,7 +9,8 @@ import TimeControls from './timeControls';
 import styled from 'styled-components';
 import { useTranslation } from '../helpers/translation';
 import { TIME } from '../helpers/constants';
-import { useSelector } from 'react-redux';
+import { useDataStatus } from '../hooks/redux';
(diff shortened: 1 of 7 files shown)


Kongkille:
CC @madsnedergaard if you have any thoughts so far. 

My core idea here is to introduce a data status for both the grid and for each zone separately. Both the grid and the history fetch can expire each other which would trigger a refetch. I will need to think about how not to get into some kind of infinity loop, but that should be solveable

So far I've only added a bit of testing on the datareducer side of things.

madsnedergaard:
This is awesome, I think this approach makes good sense! :)
Let me know if there's anything I can help with or when it's ready for review

PR #4307, opened 2022-06-30
Fix missing countrypanel elements

Kongkille:
The top-level of countryPanel was missing. This seems to work as intended.

I've added a note to ensure the hardcoded zones are taken care off in the backend if relevant.


madsnedergaard (approved):
Great! E2E should have caught this, shouldn't it? 🤔

Kongkille:
> Great! E2E should have caught this, shouldn't it? 🤔

Yes, I believe the issue is that hasData will always be true for the mockserver as it redirects to the default, so it would have passed the tests

PR #4309, opened 2022-07-01
Enable exchanges in 24 hour view

Kongkille:
Slightly change the exchanges data structure and allow it to be controlled by the selectedZoneTimeIndex


web/cypress/integration/timeslider.spec.js
```diff
@@ -17,29 +20,41 @@ describe('TimeController', () => {
...
+    cy.get('[data-test-id=date-display').should(
+      'have.text',
+      formatDate(new Date(hourlyData.data.zoneStates[5].stateDatetime), 'en-GB', 'hourly')
+    );
+    cy.get('[data-test-id=co2-square-value').should(
+      'have.text',
+      Math.round(hourlyData.data.zoneStates[5].co2intensity)
+    );
 
     // Monthly
```
madsnedergaard, line 41:
```suggestion
    // TODO: Switch away from fixed numbers for the remaining of the assertions here, so we can remove the timezone environment variable
    // Monthly
```

PR #4318, opened 2022-07-02
Add Ceuta and Melilla and split Faroe Islands

VIKTORVAV99:
This PR adds Ceuta and Melilla to the zone config.

However we still need to create geographies for them, unfortunately I can't do this myself at the moment so I would appreciate if someone could help with that. Either create a PR that target this branch, if you got write access feel free to just commit directly or just create a separate PR that should be merged together with this one.

>**Note**
>This PR also contain all the changes from #4436 in order to keep all the geographies in sync.


Kongkille:
Thanks Viktor, I'll look into creating the geographies next week after we've launched historical view!

VIKTORVAV99:
> Thanks Viktor, I'll look into creating the geographies next week after we've launched historical view!

Thanks you as well!
If you got the time could you look into splitting Sweden for #3936 as well? Not a priority by any means but then that PR would be more or less ready as well for when we support flowtracing multiple granularities.

Thanks again for helping out with this PR.

Kongkille:
Apologies for the late return here, we had a few production issues this last week. I'll create the geographies later today :)

VIKTORVAV99:
> Apologies for the late return here, we had a few production issues this last week. I'll create the geographies later today :)

No worries!
I figured it was something like that. 👍🏻

Kongkille:
I've got here an updated world.geojson file

This contains both these geographies as well as the ones in #4436 so we'd need to merge those together

I did not commit directly because we still need to run 'update-world' which will only pass when all required zone keys are in zones.json

[world.zip](https://github.com/electricitymap/electricitymap-contrib/files/9349853/world.zip)

VIKTORVAV99:
> I've got here an updated world.geojson file
> 
> This contains both these geographies as well as the ones in #4436 so we'd need to merge those together
> 
> I did not commit directly because we still need to run 'update-world' which will only pass when all required zone keys are in zones.json
> 
> [world.zip](https://github.com/electricitymap/electricitymap-contrib/files/9349853/world.zip)
> 

I can merge the other PR into this one and then we can commit the geofiles in here, that should make the CI and backend happy.

VIKTORVAV99:
@pierresegonne @Kongkille This should be ready to merge now :)

@Kongkille I tried to add you as a co-author for the geographies but I think I messed up, sorry about that.

pierresegonne:
I'll let you approve @Kongkille but LGTM

Kongkille (approved):
Seems to be working. Nice!

Issue #2912, opened 2021-01-11
Add AD 🇦🇩 exchanges


Manu1400:
Add link (without realtime value) : 
* ES<->AD exchange
* FR<->AD exchange

source : 
"L'électricité consommée en Andorre (de moyenne pour les années 2010 et 2011, 599.000 MW.h) est principalement importée
de France et d’Espagne." in https://unfccc.int/sites/default/files/resource/AND_BUR1_Definitiu.pdf

alixunderplatz:
Hi @Manu1400 
technical exchange capacities in Megawatts are required for the interconnectors on the map (not the voltages of the power lines). 
Without real-time data, it is not necessary to include them in the code for the map, as they won't appear.
Could you find any real-time exchange data for Andorra with France and Spain? 
If not, I'd close the issue.

gwpicard:
@Manu1400 any updates on real-time exchange data between FR/ES <--> AD? If not as @alixunderplatz we would close this issue because it won't change much in terms of our data just listing the existence of the exchange without data

silimotion:
I've found that the exchange between Spain and Andorra is published in the [REE website](https://www.esios.ree.es/en/border-balance):

![Screenshot_20230723_174434](https://github.com/electricitymaps/electricitymaps-contrib/assets/84802276/4922cf81-6deb-4b4f-bd11-b30a1726deab)

Would this be useful? It seems ESIOS is also used in the Spain-Morocco connection.

VIKTORVAV99:
> I've found that the exchange between Spain and Andorra is published in the [REE website](https://www.esios.ree.es/en/border-balance):
> 
> 
> Would this be useful? It seems ESIOS is also used in the Spain-Morocco connection.

Nice! Looks like we should be able to build a parser with this info and get the exchange added.

VIKTORVAV99:
I found a slight problem, the dataset that should contain the aggregated final import export for the Spain -> Andorra exchange is empty...

Here is the Andorra exchange dataset:
https://www.esios.ree.es/en/analysis/10210?vis=1&start_date=24-07-2023T00%3A00&end_date=24-07-2023T23%3A55&compare_start_date=23-07-2023T00%3A00&groupby=hour

Here is the Morocco exchange dataset (that we currently use):
[...]

PR #4320, opened 2022-07-03
Add Andorra capacity

autipial:
This PR adds:
- Andorra generation capacity;
- Andorra ⇄ Spain exchange capacity, extrapolated from a graph published by REE in its 2020 annual report.

Fixes partly #2912.

Unfortunately there is no live data yet regarding generation for Andorra, besides very limited data about hydro production on [FEDA's website](https://www.feda.ad/energia-i-meteo/energia/actual) on some days.


DATA_SOURCES.md
```diff
@@ -389,6 +389,7 @@ Cross-border transmission capacities between the zones are centralized in the [e
 - Åland ⇄ Sweden: ["Sverigekabeln": 80 MW](https://kraftnat.ax/stamnatet/)
 - Åland ⇄ Finland: ["Gustavskablen": 10 MW + "ÅL-Link": 100 MW](https://kraftnat.ax/stamnatet/)
 - Albania ⇄ Greece: [533 MW](https://ec.europa.eu/energy/sites/ener/files/documents/2nd_report_ic_with_neighbouring_countries_b5.pdf)
+- Andorra ⇄ Spain: [125 MW](https://www.ree.es/sites/default/files/publication/2022/05/downloadable/inf_sis_elec_ree_2020_0EN_0.pdf)
```
VIKTORVAV99, line 392: Looks good to me but you should add the actual exchange (with capacity) to `exchanges.json` as well.

config/zones.json
```diff
@@ -5,11 +5,13 @@
       [2.2650907800001505, 43.14936167400003]
     ],
     "capacity": {
-      "hydro": 45,
-      "nuclear": 0
+      "biomass": 5,
+      "gas": 1.7,
+      "hydro": 46,
+      "solar": 0.231
```
VIKTORVAV99, line 11: When reviewing #4321 I noticed IRENA lists 4 MW of installed solar, if you added that it would be great.

VIKTORVAV99:
If you add the exchange information to `exchanges.json` this will partially fix #2912

config/zones.json
VIKTORVAV99: Does FEDA include residential and commercial rooftop solar?

I think most if not all of our other sources do that so if they don't it might be better to go with IRENA.

As for when there is conflicting information we try and go with the official ones but if there is another source with more comprehensive data (better breakdown per mode for example) that can be used as a complement or even fully.

I don't think we have had two fully conflicting (official) sources yet so not sure what we'd do if we ran into that problem.

madsnedergaard:
I have taken the liberty of finish this PR up so we can get it merged :) 
The exchange source was removed, as we're not showing this data on the map without a parser.

madsnedergaard (approved):

PR #4327, opened 2022-07-05
Historical view fix width graph

corradio:
Description will be added here soon.

Turns out it also fixes https://github.com/electricitymap/electricitymap-contrib/issues/4144

image
image

### DIFF
--- a/web/src/components/graph/areagraph.jsx
+++ b/web/src/components/graph/areagraph.jsx
@@ -118,7 +118,11 @@ const AreaGraph = React.memo(
     const valueScale = useMemo(() => getValueScale(containerHeight, totalValues), [containerHeight, totalValues]);
     const datetimes = useCurrentZoneHistoryDatetimes();
     const startTime = datetimes.at(0);
-    const endTime = datetimes.at(-1);
+    const lastTime = datetimes.at(-1);
+    const intervalMs = datetimes.length > 1 ? lastTime.getTime() - datetimes.at(-2).getTime() : undefined;
+    // The endTime needs to include the last interval so it can be shown
+    const endTime = useMemo(() => new Date(lastTime.getTime() + intervalMs), [lastTime, intervalMs]);
+    const datetimesWithNext = [...datetimes, endTime];
 
     const timeScale = useMemo(
       () => getTimeScale(containerWidth, startTime, endTime),
@@ -173,7 +177,7 @@ const AreaGraph = React.memo(
         />
         <AreaGraphLayers
           layers={layers}
-          datetimes={datetimes}
+          datetimes={datetimesWithNext}
           timeScale={timeScale}
           valueScale={valueScale}
           mouseMoveHandler={mouseMoveHandler}
@@ -186,7 +190,7 @@ const AreaGraph = React.memo(
           transform={`translate(-1 ${containerHeight - 1})`}
           className="x axis"
           selectedTimeAggregate={selectedTimeAggregate}
-          datetimes={datetimes}
+          datetimes={datetimesWithNext}
         />
         <ValueAxis scale={valueScale} label={valueAxisLabel} width={containerWidth} height={containerHeight} />
         <GraphHoverLine
--- a/web/src/components/graph/areagraphlayers.jsx
+++ b/web/src/components/graph/areagraphlayers.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { noop } from '../../helpers/noop';
-import { area, curveStep } from 'd3-shape';
+import { area, curveStepAfter } from 'd3-shape';
 
 import { detectHoveredDatapointIndex } from '../../helpers/graph';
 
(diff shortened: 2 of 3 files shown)


web/src/components/countryhistorycarbongraph.jsx
```diff
@@ -10,7 +10,8 @@ import AreaGraph from './graph/areagraph';
 import { getCO2IntensityByMode } from '../helpers/zonedata';
 
 const prepareGraphData = (historyData, co2ColorScale, electricityMixMode) => {
-  if (!historyData || !historyData[0]) {
+  if (!historyData || !historyData[0] || !historyData[0].stateDatetime) {
+    // Incomplete data
```
corradio, line 14: This was added because missing `stateDatetime` caused some rendering errors. Turns out it's due to the initial data (before data is loaded from the backend)

web/src/components/graph/areagraph.jsx
```diff
@@ -118,7 +118,11 @@ const AreaGraph = React.memo(
     const valueScale = useMemo(() => getValueScale(containerHeight, totalValues), [containerHeight, totalValues]);
     const datetimes = useCurrentZoneHistoryDatetimes();
     const startTime = datetimes.at(0);
-    const endTime = datetimes.at(-1);
+    const lastTime = datetimes.at(-1);
+    const intervalMs = datetimes.length > 1 ? lastTime.getTime() - datetimes.at(-2).getTime() : undefined;
+    // The endTime needs to include the last interval so it can be shown
+    const endTime = useMemo(() => new Date(lastTime.getTime() + intervalMs), [lastTime, intervalMs]);
+    const datetimesWithNext = [...datetimes, endTime];
```
corradio, line 125: Because we render points at time T by drawing curves from T to T+1, we need to make sure the time scale goes to T+1, else the last curve won't be visible

web/src/components/graph/areagraph.jsx
```diff
@@ -186,7 +190,7 @@ const AreaGraph = React.memo(
           transform={`translate(-1 ${containerHeight - 1})`}
           className="x axis"
           selectedTimeAggregate={selectedTimeAggregate}
-          datetimes={datetimes}
+          datetimes={datetimesWithNext}
```
corradio, line 193: the time axis is passed the "augmented" datetimes array so it renders the last point (i.e. if the last datapoint is 2021, we still need to render the label "2022" so we can show data from 2021 to 2022)

web/src/components/graph/areagraphlayers.jsx
```diff
@@ -14,7 +14,8 @@ const AreaGraphLayers = React.memo(
 
     // Generate layer paths
     const layerArea = area()
-      .curve(curveStep)
+      // see https://github.com/d3/d3-shape#curveStep
+      .curve(curveStepAfter)
```
corradio, line 18: Uses curveStepAfter to make sure the curve starts at T and ends at T+1.

web/src/components/graph/graphhoverline.jsx
```diff
@@ -15,8 +15,11 @@ const GraphHoverLine = React.memo(
     const layer = layers && layers[selectedLayerIndex];
     const fill = layer && layer.markerFill;
     const datapoint = layer && layer.datapoints && layer.datapoints[selectedTimeIndex];
-
-    const x = datetimes && datetimes[selectedTimeIndex] && timeScale(datetimes[selectedTimeIndex]);
+    const interval = datetimes && datetimes.length > 1 ? timeScale(datetimes[1]) - timeScale(datetimes[0]) : undefined;
+    let x = datetimes && datetimes[selectedTimeIndex] && timeScale(datetimes[selectedTimeIndex]);
+    if (interval) {
+      x += 0.5 * interval;
+    }
```
corradio, line 22: The vertical line is just shifted by half an interval so it appears in the middle of it.

Kongkille:
Thanks for the help Oli, it makes sense to me - I'll fix the cypress tests and take it from here!

web/src/components/countryhistorycarbongraph.jsx
Kongkille: Yeah there are some odd states in between full data. This is where typescript would be very helpful

I'm changing to check for the .hasData property instead :)

web/src/components/graph/areagraph.jsx
```diff
@@ -118,7 +118,11 @@ const AreaGraph = React.memo(
     const valueScale = useMemo(() => getValueScale(containerHeight, totalValues), [containerHeight, totalValues]);
     const datetimes = useCurrentZoneHistoryDatetimes();
     const startTime = datetimes.at(0);
-    const endTime = datetimes.at(-1);
+    const lastTime = datetimes.at(-1);
+    const intervalMs = datetimes.length > 1 ? lastTime.getTime() - datetimes.at(-2).getTime() : undefined;
```
Kongkille, line 122: Arguably this could have been hard coded for each time range to make it a bit easier to know where things break in the future, but it's also nice that it's shared across time ranges.

Issue #4326, opened 2022-07-04
Cannot run parsers on Github Codespaces


steren:
I have followed the [local dev setup instructions](https://github.com/electricitymap/electricitymap-contrib/wiki/Set-up-local-environment)

After running `poetry install -E parsers`, when I run `poetry run test_parser FR production`, I get: 

```
$ poetry run test_parser FR production
/workspaces/electricitymap-contrib/.venv/lib/python3.10/site-packages/pandas/compat/__init__.py:124: UserWarning: Could not import the lzma module. Your installed Python is incomplete. Attempting to use lzma compression will result in a RuntimeError.
  warnings.warn(msg)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/python/latest/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/workspaces/electricitymap-contrib/test_parser.py", line 14, in <module>
    from parsers.lib.parsers import PARSER_KEY_TO_DICT
  File "/workspaces/electricitymap-contrib/parsers/lib/parsers.py", line 32, in <module>
    mod = importlib.import_module("parsers.%s" % mod_name)
  File "/opt/python/latest/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/workspaces/electricitymap-contrib/parsers/IN_MH.py", line 4, in <module>
    import cv2
  File "/workspaces/electricitymap-contrib/.venv/lib/python3.10/site-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/workspaces/electricitymap-contrib/.venv/lib/python3.10/site-packages/cv2/__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "/opt/python/latest/lib/python3.10/importlib/__init__.py", line 126, in import_module
[...]

PR #4328, opened 2022-07-05
Add Dockerfile to run test_parser

steren:
This will be helpful to develop new parser or run parser tests in an environment that might not have all dependencies (see #4326 as an example)


parsers.dockerfile
```diff
@@ -0,0 +1,12 @@
+# Container image that executes the parser tests 
+# Build with "$ docker build -f parsers.dockerfile -t parsers . "
+# Run with  "$ docker run parsers US-CAL-CISO production"
+# If needed, pass any token env var using "docker run -e TOKEN=<token>"
+
+FROM python:3.10
```
VIKTORVAV99, line 6:
```suggestion
FROM python:3.8
```
This should be python 3.8

VIKTORVAV99 (commented):
I won't take a stance on if this should be merged or not, that is up to the team but I think the plan is to move away from docker, at least for the frontend.

But in either case it should use python 3.8 or resolve the `.python-version` file.

Kongkille:
Thanks, yeah we're intending to move away from docker on the frontend, but it might still be useful for the parsers.

Adding @FelixDQ

FelixDQ (commented):
The backend for the parsers still uses docker (In Cloud Run), and we don't intend to move away from that. The environment can be quite annoying to set up, so a docker file for contributors seems like a great idea! We just need to remember to update the documentation in the wiki as well.

@steren can you add these to the apt-get install as well: `apt-get install -y tesseract-ocr tesseract-ocr-jpn libgl1`. Some parsers depend on tesseract which also needs to be installed manually.

Otherwise it LGTM :-)

parsers.dockerfile
```diff
@@ -0,0 +1,12 @@
+# Container image that executes the parser tests 
+# Build with "$ docker build -f parsers.dockerfile -t parsers . "
+# Run with  "$ docker run parsers US-CAL-CISO production"
+# If needed, pass any token env var using "docker run -e TOKEN=<token>"
+
+FROM python:3.8
+WORKDIR /workspace
+RUN apt-get update && apt-get install -y python3-opencv
```
FelixDQ, line 8: Ah, I missed that one. Thanks, @VIKTORVAV99 

```suggestion
RUN apt-get update && apt-get install -y python3-opencv tesseract-ocr tesseract-ocr-jpn tesseract-ocr-eng libgl1
```

steren:
Sounds good, I added these additional modules

FelixDQ (approved):

PR #4330, opened 2022-07-06
Update assets/text

Kongkille:
We're changing our name to Electricity Maps, this is the first PR to set up the initial logo and texts.

Will share more soon on why/how we're doing the change :)


VIKTORVAV99:
Will the "company name" also change for the app stores?
If so these lines also need to be updated:
https://github.com/electricitymap/electricitymap-contrib/blob/e01f89daa8fef978574094333e28d411c4d80a6b/web/public/electricityMap.webmanifest#L12-L21

web/public/images/electricitymap-icon.svg
```diff
@@ -1,4 +1,11 @@
...
+<path d="M419.859 111.232C430.035 111.232 438.285 102.685 438.285 92.1408C438.285 81.597 430.035 73.0496 419.859 73.0496C409.683 73.0496 401.434 81.597 401.434 92.1408C401.434 102.685 409.683 111.232 419.859 111.232Z" fill="black"/>
+<path d="M256 111.232C266.176 111.232 274.425 102.685 274.425 92.1408C274.425 81.597 266.176 73.0496 256 73.0496C245.824 73.0496 237.574 81.597 237.574 92.1408C237.574 102.685 245.824 111.232 256 111.232Z" fill="black"/>
+<path d="M92.1407 111.232C102.317 111.232 110.566 102.685 110.566 92.1408C110.566 81.597 102.317 73.0496 92.1407 73.0496C81.9645 73.0496 73.7151 81.597 73.7151 92.1408C73.7151 102.685 81.9645 111.232 92.1407 111.232Z" fill="black"/>
+<path d="M344.736 227.405H267.066V159.846C267.066 152.384 257.933 149.152 253.53 155.059L161.28 272C157.44 277.12 160.979 284.614 167.258 284.614H244.934V352.154C244.934 359.616 254.067 362.848 258.47 356.941L350.72 240C354.56 234.88 351.014 227.405 344.736 227.405Z" fill="black"/>
```
VIKTORVAV99, line 10: Looks like there are 8 perfect circles that could be simplified with [`<circle>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle) here.
Not blocking by any means but it would make the file a bit smaller.

Kongkille: Good point also the paths could be joined as there's no reason to be able to color them individually. However, I am not certain it works with our icon-sprite https://github.com/electricitymap/electricitymap-contrib/blob/e01f89daa8fef978574094333e28d411c4d80a6b/web/public/images/icon-sprite.svg?short_path=83ed875#L29

Kongkille:
Oh thanks for the reviewwe wanted to launch this last week but had to postpone due to sickness.  Anyways, our plan is to get this live tomorrow so I'll just get this merged in :)

Kongkille:
> Will the "company name" also change for the app stores? If so these lines also need to be updated:
> 
> https://github.com/electricitymap/electricitymap-contrib/blob/e01f89daa8fef978574094333e28d411c4d80a6b/web/public/electricityMap.webmanifest#L12-L21
> 
> PS: Might want to rename the entire file too to stay consistent.



> Will the "company name" also change for the app stores? If so these lines also need to be updated:
> 
> https://github.com/electricitymap/electricitymap-contrib/blob/e01f89daa8fef978574094333e28d411c4d80a6b/web/public/electricityMap.webmanifest#L12-L21
> 
> PS: Might want to rename the entire file too to stay consistent.

Yeah there are so many places we need to change this. I've tried in this PR to change it in all the non-breaking places and then we can gradually switch it out over time on these more complex places

web/public/images/electricitymap-icon.svg
VIKTORVAV99: Everything that is valid SVG should work in the sprite but I can take a look at it tomorrow if you want.

Kongkille: Yeah that's a fun idea :)

PR #4335, opened 2022-07-10
Improve Python typings

VIKTORVAV99:
This PR updates the function types, return types and fixes some unbound errors.

This is in a effort to make our code base more type safe and easier to work with in the future.


FelixDQ:
I think that this option would be best:
> Another option would be to create a custom type that should be used as a return type for all parsers.

I would generally lean towards the option that makes it easier for contributors, and I would think that it might be nicer to return a dict in some cases. I'm also open to the other option as well if you think that is better.



> parsers/lib/validation

We will be deprecating this soon after moving everything to the `validators`, so I think that any quickfix that works here is fine.

FelixDQ:
Also, this change is really nice! 🎉

VIKTORVAV99:
@FelixDQ Updated the PR with Union types as valid return types as I figured that would be the easiest way forward for now and this error popped up:
```log
   self.assertEqual(
                [a for a in args],
                EXPECTED_MODE_FUNCTION_ARGS[_mode],
                f"invalid args for {function_name}, arg_spec={arg_spec}",
            )
E           AssertionError: Lists differ: ['zone_key2', 'zone_key1', 'session', 'target_datetime', 'logger'] != ['zone_key1', 'zone_key2', 'session', 'target_datetime', 'logger']
E           
E           First differing element 0:
E           'zone_key2'
E           'zone_key1'
E           
E           - ['zone_key2', 'zone_key1', 'session', 'target_datetime', 'logger']
E           ?               -------------
E           
E           + ['zone_key1', 'zone_key2', 'session', 'target_datetime', 'logger']
E           ?  +++++++++++++
E            : invalid args for BD.fetch_exchange, arg_spec=FullArgSpec(args=['zone_key2', 'zone_key1', 'session', 'target_datetime', 'logger'], varargs=None, varkw=None, defaults=(None, None, <Logger electricitymap.contrib.parsers.BD (ERROR)>), kwonlyargs=[], kwonlydefaults=None, annotations={'return': <class 'list'>, 'zone_key2': <class 'str'>, 'zone_key1': <class 'str'>, 'session': typing.Union[requests.sessions.Session, NoneType], 'target_datetime': typing.Union[datetime.datetime, NoneType], 'logger': <class 'logging.Logger'>})
```
I don't really get why it happens (or what it means really), the only thing this PR should change is the types so this shouldn't be happening. (Check the CI logs for more details if needed)

If you could have a look into why I'd really appreciate it but it seems like it only affects the `BD.py` parser.

parsers/AR.py
```diff
@@ -28,16 +23,16 @@
 
 def fetch_production(
     zone_key="AR",
-    session=None,
-    target_datetime: datetime.datetime = None,
-    logger: logging.Logger = logging.getLogger(__name__),
+    session: Union[Session, None] = None,
+    target_datetime: Union[datetime, None] = None,
```
jarek, line 27: As mentioned in the other PR: could we use [typing.Optional](https://docs.python.org/3/library/typing.html#typing.Optional) here? I find `Optional[datetime]` cleaner than `Union[datetime, None]`.

VIKTORVAV99: Yeah that works, I should actually update the other PR with that too..

VIKTORVAV99:
@FelixDQ I know you are busy but it would be great if we can get this merged before it starts getting a bunch of merge conflicts. Since the PR changes 100 files it is bound to happen sooner or later.

validators/zone_specific_checks.py
```diff
@@ -1,8 +1,6 @@
-from datetime import datetime
-
 import pandas as pd
 
-from electricitymap.contrib.validators.lib.config import validator
+from validators.lib.config import validator
```
FelixDQ, line 3: I think that these needs to remain as `electricitymap.contrib.validators` for us to be able to import them in the backend. There's a symlink from validators to `electricitymap.contrib.validators`.

FelixDQ (commented):
Great! I have a single comment, but otherwise this LGTM :-)

validators/zone_specific_checks.py
VIKTORVAV99: Ah I was hoping it would work with the direkt link since the symlinks used by poetry doesn't work on windows.

I have changed some of them in the past without issue (I think), see #4256 for more details.

So could you double check if this breaks the backend? Obviously if it does I'll revert it otherwise I'd like to keep the changes.

FelixDQ: Ah, I see. Let's try to merge this then. I'll create a follow-up PR reverting this if it breaks the tests.

FelixDQ (approved):

validators/zone_specific_checks.py
VIKTORVAV99: Thanks!

PR #4350, opened 2022-07-17
fix(JP-TK): fix "No recent data found for consumptionForecast parser" warning

shuuji3:
TEPCO's public CSV data URL has been updated twice in the past every time its spec changed.

1. http://www.tepco.co.jp/forecast/html/images/juyo-j.csv
2. http://www.tepco.co.jp/forecast/html/images/juyo-d-j.csv
3. http://www.tepco.co.jp/forecast/html/images/juyo-d1-j.csv (latest)

The current parser function for consumption forecast uses the first one and it seems that the company stopped updating at least the consumption forcast in the old CSVs. Those CSVs always return `0` as forcast.

This PR changes to use he latest one, which is always updated now.


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for the fix.

PR #4354, opened 2022-07-18
fix(KR): correct the price value unit from Won/kWh to Won/MWh

### LINKED ISSUE #4084: South Korean electricity price is off by a factor of 1000x
Someone reached out to us on email with the following feedback:

Just want to mention, that the electricity price in South Korea is probably off by factor of 1000x.
I believe this is due to different way to use decimal point. Great Britain and the United States are two of the few places in the world that use a period to indicate the decimal place. Many other countries use a comma instead. The decimal separator is also called the radix character. Likewise, while the U.K. and U.S. use a comma to separate groups of thousands, many other countries use a period instead, and some countries separate thousands groups with a thin space.

shuuji3:
resolves #4084 

The data source (https://new.kpx.or.kr/smpInland.es?mid=a10606080100&device=pc) shows the value as Won/KWh but the parser treated it as Won/MWh.


shuuji3:
I'll fix the format error.

VIKTORVAV99:
Nice fix. This should be ready to merge when the formatting is done.

VIKTORVAV99 (approved):
LGTM! 🎉

shuuji3:
OK, I applied black formatting.

shuuji3:
Thanks!

PR #4360, opened 2022-07-22
Average by source

VIKTORVAV99:
Small UI change that makes it clearer that the aggregate time periods (day, month and year) is a average by source.

Before:
image
<sup>Note I used the live website for this screenshot.</sup>
After:
image

PS: The huge diff in the `sv.json` file is probably because I used the `translation-helper.js` instead of editing it directly but it only contains one new string. (same as `en.json`)

### DIFF
--- a/web/src/layout/leftpanel/countrypanel.jsx
+++ b/web/src/layout/leftpanel/countrypanel.jsx
@@ -123,6 +123,7 @@ const BySource = styled.div`
   font-size: smaller;
   position: relative;
   top: 0.8rem;
+  margin-bottom: 5px;
 `;
 
 const LoadingText = styled.p`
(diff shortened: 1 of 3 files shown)


web/src/layout/leftpanel/countrypanel.jsx
```diff
@@ -123,6 +123,7 @@ const BySource = styled.div`
   font-size: smaller;
   position: relative;
   top: 0.8rem;
+  margin-bottom: 5px;
```
VIKTORVAV99, line 126: To prevent text overlap.

web/src/layout/leftpanel/countrypanel.jsx
```diff
@@ -323,7 +324,9 @@ const CountryPanel = ({ electricityMixMode, isMobile, tableDisplayEmissions, zon
       <CountryPanelWrap>
         {hasParser ? (
           <React.Fragment>
-            <BySource>{__('country-panel.bysource')}</BySource>
+            <BySource>
+              {__(timeAggregate != TIME.HOURLY ? 'country-panel.averagebysource' : 'country-panel.bysource')}
```
Kongkille, line 328:
```suggestion
              {__(timeAggregate !== TIME.HOURLY ? 'country-panel.averagebysource' : 'country-panel.bysource')}
```

Kongkille: I dont think it changes functionality in any case but I just find that odd things happens when using !=

Kongkille (approved):

PR #4363, opened 2022-07-23
New production parser for South Africa

IV1T3:
This PR introduces the new parser for South Africa based on two Eskom data sources.
The complete discussion can be seen in issue #2881.

### DIFF
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -711,7 +711,7 @@ def get_power_bi_values():
 
 
 def fetch_production(
-    zone_key="SA",
+    zone_key="ZA",
     session=None,
     target_datetime: datetime.datetime = None,
     logger: logging.Logger = logging.getLogger(__name__),
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 from collections import OrderedDict
 from datetime import datetime
 from logging import Logger, getLogger
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 import logging
 import pprint
+import typing
 from collections import OrderedDict
 
 import arrow
@@ -714,7 +715,7 @@ def fetch_production(
     session=None,
     target_datetime: datetime.datetime = None,
     logger: logging.Logger = logging.getLogger(__name__),
-) -> dict:
+) -> typing.List[dict]:
 
     if target_datetime is not None:
         local_target_datetime = arrow.get(target_datetime).to(TIMEZONE)
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -935,7 +935,12 @@ def fetch_production(
         cleansed_csv_values = value["cleansed_csv"].split(",")
 
         storage_inversion_idcs = [3, 4, 11, 12]
-        production_idcs = [0, 1, 2, 6, 7, 8, 9, 10, 16, 17, 18, 19]
+        production_idcs = [0, 1, 2, 6, 8, 9, 10, 16, 17, 18, 19]
+
+        # Omitted values
+        # 5, 7, 13, 14, 15
+        # TODO:
+        # - 7 (international imports) can be further implemented in exchange function.
 
         for j, cleansed_csv_value in enumerate(cleansed_csv_values):
             if j in storage_inversion_idcs:
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
 
-from arrow import now, get
 from collections import OrderedDict
 from datetime import datetime
 from logging import Logger, getLogger
 from pprint import PrettyPrinter
-from requests import Response, post, Session
-from requests import session as req_session
 from typing import List, Optional
 
+from arrow import get, now
+from requests import Response, Session, post
 
 pp = PrettyPrinter(indent=4)
 
@@ -712,8 +711,8 @@ def get_power_bi_values():
 
 
 def fetch_production(
-    zone_key: str="ZA",
-    session: Optional[Session]=None,
+    zone_key: str = "ZA",
+    session: Optional[Session] = None,
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> List[dict]:
@@ -727,7 +726,7 @@ def fetch_production(
                 f"No production data is available for {local_target_datetime}."
             )
 
-    r = session or req_session()
+    r = session or Session()
 
     res: Response = r.get(PRODUCTION_URL)
     assert res.status_code == 200, (
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -882,21 +882,21 @@ def fetch_production(
     # Helpful: https://www.eskom.co.za/dataportal/glossary/
     column_mapping = {
         0: "coal",  # Thermal_Gen_Excl_Pumping_and_SCO
-        1: "unknown",  # Eskom_OCGT_SCO_Pumping             changed to unknown since negative oil is not possible
-        2: "unknown",  # Eskom_Gas_SCO_Pumping              changed to unknown since negative gas is not possible
-        3: "hydro",  # Hydro_Water_SCO_Pumping
+        1: "ignored",  # Eskom_OCGT_SCO_Pumping             changed to ignored since negative oil is not possible, usually [-6, 0]
+        2: "ignored",  # Eskom_Gas_SCO_Pumping              changed to ignored since negative gas is not possible, usually -1 or 0
+        3: "ignored",  # Hydro_Water_SCO_Pumping            Probably electricity consumed by the plant itself (even) when not generating power. Can be ignored.
         4: "hydro",  # Pumped_Water_SCO_Pumping
-        5: "unknown",  # Thermal_Generation                 sum of 0, 1, 2, 3, 4. This value is not used.
+        5: "ignored",  # Thermal_Generation                 sum of 0, 1, 2, 3, 4. Can be ignored.
         6: "nuclear",  # Nuclear_Generation
-        7: "unknown",  # International_Imports
+        7: "ignored",  # International_Imports
         8: "oil",  # Eskom_OCGT_Generation
         9: "gas",  # Eskom_Gas_Generation
         10: "oil",  # Dispatchable_IPP_OCGT
         11: "hydro",  # Hydro_Water_Generation
         12: "hydro",  # Pumped_Water_Generation
-        13: "unknown",  # IOS_Excl_ILS_and_MLR              Interruption of Supply. Not used.
-        14: "unknown",  # ILS_Usage                         Interruptible Load Shed = companies paid not to consume electricity. Not used.
-        15: "unknown",  # Manual_Load_Reduction_MLR         MLS = forced load shedding. Not used.
+        13: "ignored",  # IOS_Excl_ILS_and_MLR              Interruption of Supply. Can be ignored.
+        14: "ignored",  # ILS_Usage                         Interruptible Load Shed = companies paid not to consume electricity. Can be ignored.
+        15: "ignored",  # Manual_Load_Reduction_MLR         MLS = forced load shedding. Can be ignored.
         16: "wind",  # Wind
         17: "solar",  # PV
         18: "solar",  # CSP
@@ -916,7 +916,6 @@ def fetch_production(
             "zoneKey": zone_key,
             "datetime": datetime_key.datetime,
             "production": {
-                "biomass": None,
                 "coal": 0.0,
                 "gas": 0.0,
                 "hydro": 0.0,
@@ -924,7 +923,6 @@ def fetch_production(
                 "oil": 0.0,
                 "solar": 0.0,
                 "wind": 0.0,
-                "geothermal": None,
                 "unknown": 0.0,
             },
             "storage": {"hydro": 0.0},
@@ -933,11 +931,11 @@ def fetch_production(
 
         cleansed_csv_values = value["cleansed_csv"].split(",")
 
-        storage_inversion_idcs = [3, 4, 12]
-        production_idcs = [0, 1, 2, 6, 8, 9, 10, 11, 16, 17, 18, 19]
+        storage_inversion_idcs = [4, 12]
+        production_idcs = [0, 6, 8, 9, 10, 11, 16, 17, 18, 19]
 
-        # Omitted values
-        # 5, 7, 13, 14, 15
+        # Ignored values
+        # 1, 2, 3, 5, 7, 13, 14, 15
         # TODO:
         # - 7 (international imports) can be further implemented in exchange function.
 
@@ -945,12 +943,12 @@ def fetch_production(
             if j in storage_inversion_idcs:
                 data["storage"][column_mapping[j]] = round(
                     data["storage"][column_mapping[j]] + float(cleansed_csv_value) * -1,
-                    2,
+                    13,
                 )
             elif j in production_idcs:
                 data["production"][column_mapping[j]] = round(
                     data["production"][column_mapping[j]] + float(cleansed_csv_value),
-                    2,
+                    13,
                 )
 
         all_data.append(data)
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -932,8 +932,8 @@ def fetch_production(
 
         cleansed_csv_values = value["cleansed_csv"].split(",")
 
-        storage_inversion_idcs = [3, 4, 11, 12]
-        production_idcs = [0, 1, 2, 6, 8, 9, 10, 16, 17, 18, 19]
+        storage_inversion_idcs = [3, 4, 12]
+        production_idcs = [0, 1, 2, 6, 8, 9, 10, 11, 16, 17, 18, 19]
 
         # Omitted values
         # 5, 7, 13, 14, 15
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -919,6 +919,7 @@ def fetch_production(
                 "biomass": None,
                 "coal": 0.0,
                 "gas": 0.0,
+                "hydro": 0.0,
                 "nuclear": 0.0,
                 "oil": 0.0,
                 "solar": 0.0,
--- a/parsers/ZA.py
+++ b/parsers/ZA.py
@@ -1,19 +1,20 @@
 from __future__ import annotations
 
-import datetime
-import logging
-import pprint
-import typing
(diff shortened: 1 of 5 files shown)


IV1T3:
Unfortunately, I don't really understand the error message from the failed test:

` =========================== short test summary info ============================
FAILED tests/test_parser_interface.py::ParserInterfaceTestcase::test_interface
================== 1 failed, 116 passed, 15 warnings in 6.14s ==================`

`AssertionError: False is not true : expected annotation for ZA.fetch_production to be in [<class 'dict'>, <class 'list'>, typing.List[dict]] not {'return': 'typing.List[dict]', 'target_datetime': 'datetime.datetime', 'logger': 'logging.Logger'}`

jarek (commented):
The failed test looks to be a test problem for me. The type annotation looks fine to me.

I can't actually replicate the problem, although I have Python 3.6.9 and the test suite is using Python 3.8. I get `annotations` to be `{'return': typing.List[dict], 'target_datetime': <class 'datetime.datetime'>, 'logger': <class 'logging.Logger'>}` which is what's expected for `return`. I'm not seeing anything in the documentation that would explain why the test is expecting `'typing.List[dict]'` specifically as string rather than type.

VIKTORVAV99:
I'll see if I can give it a look later today or tomorrow at the latest.

parsers/ZA.py
```diff
@@ -0,0 +1,961 @@
+from __future__ import annotations
+
+import datetime
+import logging
+import pprint
+import typing
+from collections import OrderedDict
+
+import arrow
+import requests
```
VIKTORVAV99, line 10: Please import and use only the classes, methods and types needed. It makes things easier to understand.
For example:
```python
from requests import Session, Response
```

parsers/ZA.py
```diff
@@ -0,0 +1,961 @@
...
+    assert len(cleaned_power_bi_values) > 0
+
+    return cleaned_power_bi_values
+
+
+def fetch_production(
+    zone_key="ZA",
+    session=None,
+    target_datetime: datetime.datetime = None,
+    logger: logging.Logger = logging.getLogger(__name__),
```
VIKTORVAV99, line 717:
```suggestion
    zone_key: str="ZA",
    session: Union[Session, None]=None,
    target_datetime: Union[datetime, None] = None,
    logger: Logger = getLogger(__name__),
```

These would be the correct types (if the other above review comment is followed).

parsers/ZA.py
```diff
@@ -0,0 +1,961 @@
...
+        local_one_week_ago = arrow.now(TIMEZONE).shift(days=-7)
+
+        if local_target_datetime < local_one_week_ago:
+            raise NotImplementedError(
+                f"No production data is available for {local_target_datetime}."
+            )
+
+    r = session or requests.session()
+
+    res = r.get(PRODUCTION_URL)
```
VIKTORVAV99, line 731:
```suggestion
    res: Response = r.get(PRODUCTION_URL)
```

parsers/ZA.py
```diff
@@ -0,0 +1,961 @@
...
+            if j in storage_inversion_idcs:
+                data["storage"][column_mapping[j]] = round(
+                    data["storage"][column_mapping[j]] + float(cleansed_csv_value) * -1,
+                    13,
+                )
+            elif j in production_idcs:
+                data["production"][column_mapping[j]] = round(
+                    data["production"][column_mapping[j]] + float(cleansed_csv_value),
+                    13,
+                )
```
VIKTORVAV99, line 950:
```suggestion
        for j, cleansed_csv_value in enumerate(cleansed_csv_values):
            if j in storage_inversion_idcs:
                data["storage"][column_mapping[j]] = round(
                    data["storage"][column_mapping[j]] + float(cleansed_csv_value) * -1,
                    2,
                )
            elif j in production_idcs:
                data["production"][column_mapping[j]] = round(
                    data["production"][column_mapping[j]] + float(cleansed_csv_value),
                    2,
                )
```
2 decimals should be enough but at most it should be 4, we don't need 13 decimals or microwatt scale (if my math checks out).

parsers/ZA.py
```diff
@@ -0,0 +1,961 @@
...
+    # Helpful: https://www.eskom.co.za/dataportal/glossary/
+    column_mapping = {
+        0: "coal",  # Thermal_Gen_Excl_Pumping_and_SCO
+        1: "unknown",  # Eskom_OCGT_SCO_Pumping             changed to unknown since negative oil is not possible
+        2: "unknown",  # Eskom_Gas_SCO_Pumping              changed to unknown since negative gas is not possible
+        3: "hydro",  # Hydro_Water_SCO_Pumping
+        4: "hydro",  # Pumped_Water_SCO_Pumping
+        5: "unknown",  # Thermal_Generation                 sum of 0, 1, 2, 3, 4. This value is not used.
+        6: "nuclear",  # Nuclear_Generation
+        7: "unknown",  # International_Imports
```
VIKTORVAV99, line 893: I don't think this should be in the production parser, it should be in a exchange parser or omitted.

VIKTORVAV99 (changes requested):
I did not find the reason for the failed tests but I did find a few things that should be addressed, mainly typings and imports.

I am aware that some of the suggested changes will differ from what is in the example parser but that file is getting updated in another PR and the suggestions and feedback match that.

Overall great work on this parser though! 🎉

parsers/ZA.py
jarek: Do we use `Union[blah, None]` rather than `Optional[blah]`?

parsers/ZA.py
IV1T3: Tbh I am not really in favor of this convention, but I changed my imports in commit 201c005b280d56fa05a757e8b1d3963e138f05a3.

parsers/ZA.py
IV1T3: Done in commit 201c005b280d56fa05a757e8b1d3963e138f05a3

parsers/ZA.py
IV1T3: Done in commit 201c005b280d56fa05a757e8b1d3963e138f05a3

parsers/ZA.py
```diff
@@ -0,0 +1,962 @@
+from __future__ import annotations
+
+from arrow import now, get
+from collections import OrderedDict
+from datetime import datetime
+from logging import Logger, getLogger
+from pprint import PrettyPrinter
+from requests import Response, post, Session
+from requests import session as req_session
```
VIKTORVAV99, line 9:
```suggestion
```

parsers/ZA.py
```diff
@@ -0,0 +1,962 @@
...
+    if target_datetime is not None:
+        local_target_datetime = get(target_datetime).to(TIMEZONE)
+        local_one_week_ago = now(TIMEZONE).shift(days=-7)
+
+        if local_target_datetime < local_one_week_ago:
+            raise NotImplementedError(
+                f"No production data is available for {local_target_datetime}."
+            )
+
+    r = session or req_session()
```
VIKTORVAV99, line 730:
```suggestion
    r = session or Session()
```

VIKTORVAV99 (changes requested):
Just two more nitpicks otherwise I think everything looks good code wise.

don't forget to run `poetry run format` to fix the black formatting error though.

parsers/ZA.py
IV1T3: Done in commit 355a3467b0951490eb00d5d08ff2c31f6be1dd74

parsers/ZA.py
IV1T3: Done in commit 355a3467b0951490eb00d5d08ff2c31f6be1dd74

parsers/ZA.py
VIKTORVAV99: Yes I'm aware.
But 7, which is the international imports, are included in the production id list.
I suggest you remove it and or add it to a exchange parser if possible.

IV1T3: Ah ok, will simply remove it for now then! Thanks @VIKTORVAV99 for your feedback!

VIKTORVAV99: Happy to help! I'll try and dig into why the tests are failing later (for no apparent reason).

IV1T3: Done in commit 465aeb398a9f44c39ba682155a0f0d84bcc824dd

parsers/ZA.py
IV1T3: Done in commit 201c005b280d56fa05a757e8b1d3963e138f05a3

AndBerna:
Hello, amazing job! 🤩 

We have also been working on this internally so we will also review it in the following days  😊

VIKTORVAV99 (approved):

parsers/ZA.py
```diff
@@ -0,0 +1,966 @@
+from __future__ import annotations
+
```
VIKTORVAV99, line 2:
```suggestion
```
Not sure why but this breaks pytest (and the CI) so let's remove it for now.

pierresegonne: This is only needed if you're going to reference a type that you define in this file

VIKTORVAV99:
@AndBerna and @Kongkille I think this is ready to merge (now that we got the CI sorted) and I have approved it but I'll leave it up to the team to merge it when you feel ready.

pierresegonne:
I've pulled out data to compare historical (that we got from the TSO) and current data from the parser:
[ZA_parser_data.csv](https://github.com/electricitymap/electricitymap-contrib/files/9201489/ZA_parser_data.csv)
[parser.csv](https://github.com/electricitymap/electricitymap-contrib/files/9201491/parser.csv)

Don't have time to make a proper comparison, but from a quick glance it seems like it matches. I'll let @AndBerna make the final analysis and merge what seems to look good to me 👏 Great job guys!!

parsers/ZA.py
q--: > 2 decimals should be enough but at most it should be 4, we don't need 13 decimals or microwatt scale (if my math checks out).

This is only done to eliminate floating point errors,  it should not result in numbers with 13 actual decimal places unless those are actually in the data source. 13 is due to the size of Python's floats, 14 doesn't get rid of the floating point errors.

Anyway, I don't think  rounding it that much is a good idea. On a low production grid it'll actually make a significant difference (e.g. solar on the [Faroe Islands](https://app.electricitymaps.com/zone/FO?wind=true) is less than 100kW, so you see 10kW, 20kW, 70kW - more precision would be actually welcome there.

Furthermore, I think it would be better to leave the precision of the stored data to ElectricytMap's back-end systems, not the parser. That makes it easier to increase the precision if desired, without having to edit multiple individual parsers.

pierresegonne: I confirm, in general the closer the data we receive from the parsers is from the data source the best it is so we can then have the freedom to design a consistent set of operations in the backend

parsers/ZA.py
```diff
@@ -0,0 +1,964 @@
...
+                "unknown": 0.0,
+            },
+            "storage": {"hydro": 0.0},
+            "source": " https://www.eskom.co.za",
+        }
+
+        cleansed_csv_values = value["cleansed_csv"].split(",")
+
+        storage_inversion_idcs = [3, 4, 11, 12]
+        production_idcs = [0, 1, 2, 6, 8, 9, 10, 16, 17, 18, 19]
```
AndBerna, line 936: Amazing job guys! 🤩. Just a small nitpick

At [Eskom Glossary](https://www.eskom.co.za/dataportal/glossary/) they make the difference between  

- Hydro Generation – Generation from large hydro power stations, and sent-out onto the Transmission network.

- Pumped Water Generation – Generation from pumped storage power stations, and sent-out onto the Transmission network.

That will also go in line with the historical data we got from the TSO.

So I suggest including `# Hydro_Water_Generation` in production. 

```suggestion
        storage_inversion_idcs = [3, 4, 12]
        production_idcs = [0, 1, 2, 6, 8, 9, 10, 11, 16, 17, 18, 19]
```

parsers/ZA.py
```diff
@@ -0,0 +1,964 @@
...
+        # storage values can be both positive (when storing energy) or negative (when the
+        # storage is discharged).
+
+        data = {
+            "zoneKey": zone_key,
+            "datetime": datetime_key.datetime,
+            "production": {
+                "biomass": None,
+                "coal": 0.0,
+                "gas": 0.0,
```
AndBerna, line 921:
```suggestion
                "gas": 0.0,
                "hydro": 0.0,
```

parsers/ZA.py
IV1T3: Ah ok, tbh I find their notation quite confusing. 

We have the following four hydro values:
```
3: "hydro",  # Hydro_Water_SCO_Pumping
4: "hydro",  # Pumped_Water_SCO_Pumping
---
11: "hydro",  # Hydro_Water_Generation
12: "hydro",  # Pumped_Water_Generation
```

As far as I understood, 4 and 12 are complimentary (Pumped_Water_SCO_Pumping vs. Pumped_Water_Generation). Given your explanation for 11 still keeps me confused regarding 3. What is meant by "Hydro_Water_SCO_Pumping"?

parsers/ZA.py
```diff
@@ -0,0 +1,965 @@
...
+                data["storage"][column_mapping[j]] = round(
+                    data["storage"][column_mapping[j]] + float(cleansed_csv_value) * -1,
+                    2,
+                )
+            elif j in production_idcs:
+                data["production"][column_mapping[j]] = round(
+                    data["production"][column_mapping[j]] + float(cleansed_csv_value),
+                    2,
+                )
+
```
VIKTORVAV99, line 955:
```suggestion
        for j, cleansed_csv_value in enumerate(cleansed_csv_values):
            if j in storage_inversion_idcs:
                data["storage"][column_mapping[j]] = round(
                    data["storage"][column_mapping[j]] + float(cleansed_csv_value) * -1,
                    13,
                )
            elif j in production_idcs:
                data["production"][column_mapping[j]] = round(
                    data["production"][column_mapping[j]] + float(cleansed_csv_value),
                    13,
                )

```
Reverting this change as it reduces the data resolution, I didn't think 13 decimals was needed but the general consensus seem to be that we should keep the data as close to the source as possible.

parsers/ZA.py
q--: In the light of SCO = Synchronous Condenser Operation, and the way they use pumping for non-charging things (that is, pumped hydro; there are no batteries in the data IIRC),  I strongly suspect that in these cases they use 'pumping' for power plant self-consumption. Basically the power drawn from the grid for keeping the generators spinning and the lights on even if the plant isn't generating any power itself.

IIRC @electricitymap policy was to ignore self-consumption data. (Arguably it should be included in total consumption, but that raises the problem of how to account for it when a power plant *is* generating power and provides for its own self-consumption, so the power draw for that isn't reported because it never 'enters' or 'leaves' the grid. Therefore,  it makes sense to ignore it altogether for the sake of consistency.)

parsers/ZA.py
```diff
@@ -0,0 +1,965 @@
...
+        #     print("-----------------------------------------------------")
+
+        last_cleansed_csv_len = curr_cleansed_csv_len
+        cleansed_csv_value_updates.append(curr_update_dict)
+
+    # Mapping columns to keys
+    # Helpful: https://www.eskom.co.za/dataportal/glossary/
+    column_mapping = {
+        0: "coal",  # Thermal_Gen_Excl_Pumping_and_SCO
+        1: "unknown",  # Eskom_OCGT_SCO_Pumping             changed to unknown since negative oil is not possible
```
q--, line 885: Suggestion: perhaps you could replace "unknown" with "ignored" to make it more intuitive these fields aren't being used? Except `19` of course, as that one is actually being used.

q--:
Given that `unknown` is 'other renewables', I think you also need to add zone overrides for 'isLowCarbon' and 'isRenewable' [here](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_all.json#L11117) and [here](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_all.json#L11271)

IV1T3:
@q-- What would you suggest regarding 1 (Eskom_OCGT_SCO_Pumping) and 2 (Eskom_Gas_SCO_Pumping)? I put them into unknown, since they are both usually negative and oil and gas production don't allow negative values. Implying from your previous comment I guess you would simply ignore them as well, since they only make a small percentage of the overall mix anyways.

parsers/ZA.py
IV1T3: Done in commit 024737e2f1e4c96773dd810f06a60fd479b69ecb

parsers/ZA.py
IV1T3: Done in commit 024737e2f1e4c96773dd810f06a60fd479b69ecb

IV1T3:
Apparently the Prettier test fails for config/co2eq_parameters_all.json. Not sure why. Ran Prettier locally as well as the poetry format.

VIKTORVAV99:
> Apparently the Prettier test fails for config/co2eq_parameters_all.json. Not sure why. Ran Prettier locally as well as the poetry format. 

I'll take care of it. 👍🏻

IV1T3:
lol thx ^^

q--:
> @q-- What would you suggest regarding 1 (Eskom_OCGT_SCO_Pumping) and 2 (Eskom_Gas_SCO_Pumping)? I put them into unknown, since they are both usually negative and oil and gas production don't allow negative values. 

If my assumptions are correct it makes sense to ignore these as well

AndBerna:
> > @q-- What would you suggest regarding 1 (Eskom_OCGT_SCO_Pumping) and 2 (Eskom_Gas_SCO_Pumping)? I put them into unknon, since they are both usually negative and oil and gas production don't allow negative values.
> 
> If my assumptions are correct it makes sense to ignore these as well

Hi, yes, you are correct. We do not currently  account for self-consumption and why we agreed on the changes at [024737e](https://github.com/electricitymap/electricitymap-contrib/pull/4363/commits/024737e2f1e4c96773dd810f06a60fd479b69ecb). 

A small note in that regard, the magnitude and the correlation between `Pumped_Water_SCO_Pumping`  and `Pumped_Water_Generation` reveals that `Pumped_Water_SCO_Pumping`   is the actual pumped hydro consumption. Hence,  that is why it is included. I would agree that labelling as self-consumption (SCO)  it is somehow confusing. 

On the other hand, `Hydro_Water_SCO_Pumping` seems to account for some internal self-consumption. We run a small analysis on the historical data we got from Eskom and I can confirm that it is **always**  `0` throughout the whole time series available.  

It is ignored, for the sake of consistency, but its magnitude it is also negligible. @IV1T3  I hope this somehow clarifies your question at https://github.com/electricitymap/electricitymap-contrib/pull/4363#discussion_r932424776

PR #4364, opened 2022-07-24
fix(GCCIA): fix consumption data fetch for AE, BH, OM, QA, and SA

shuuji3:
These countries using `parsers/GCCIA.py` had the same errors:

```py
'<' not supported between instances of 'dict' and 'int'
  File "/home/contrib/electricitymap/contrib/parsers/lib/quality.py", line 33, in validate_consumption
    if (obj.get("consumption") or 0) < 0:
```

ref. https://storage.googleapis.com/electricitymap-parser-logs/AE.html

This was because `fetch_consumption()` returned dictionary as a consumption (i.e. `{"unknown": 123}`). I guess it was confused with `fetch_production()`. This PR fixes the function to return int value as a consumption value.

Also, SA is currently showing  " - " instead of a value like "123 MW". I added a check if the value is currently available and made it raise `RuntimeError`. Is it OK to raise `RuntimeError` to handle a case like this?

ref. https://storage.googleapis.com/electricitymap-parser-logs/SA.html

### DIFF
--- a/parsers/GCCIA.py
+++ b/parsers/GCCIA.py
@@ -48,7 +48,7 @@ def fetch_consumption(zone_key, session=None, target_datetime=None, logger=None)
     url = "https://www.gccia.com.sa/"
     response = r.get(url)
 
-    pattern = COUNTRY_CODE_MAPPING[zone_key] + '-mw-val">\s*(\d+)'
+    pattern = COUNTRY_CODE_MAPPING[zone_key] + r'-mw-val">\s*(\d+)'
 
     match = re.findall(pattern, response.text)
     if not match:
--- a/parsers/GCCIA.py
+++ b/parsers/GCCIA.py
@@ -18,6 +18,8 @@ from sys import stderr
 import arrow
 import requests
 
+from .lib.exceptions import ParserException
+
 COUNTRY_CODE_MAPPING = {
     "AE": "uae",
     "BH": "bah",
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
I believe a ParserException would be a better fit than a runtime error.
Import it with:
```python 
```

Besides that I'll try and do a full review tomorrow.

shuuji3:
Thank you. I replaced it with `ParserException`.

shuuji3:
Thank you. I replaced it with ParserException.

VIKTORVAV99 (approved):
Changes in here works great.
And since the data source have not changed it should be perfectly safe to merge it.

shuuji3:
@VIKTORVAV99 Thanks for reviewing!

By the way, I saw that several parsers use `RuntimeError`: https://github.com/electricitymap/electricitymap-contrib/search?q=RuntimeError. Do you think it's helpful to use `ParserException` in those parsers (for example, it makes it easy to understand issues)? If so, I'd like to improve those parsers using `ParserException`.

VIKTORVAV99:
> @VIKTORVAV99 Thanks for reviewing!
> 
> By the way, I saw that several parsers use `RuntimeError`: https://github.com/electricitymap/electricitymap-contrib/search?q=RuntimeError. Do you think it's helpful to use `ParserException` in those parsers (for example, it makes it easy to understand issues)? If so, I'd like to improve those parsers using `ParserException`.

ParserException is probably better since it provides more information. So if you want to update them go for it! 🎉

shuuji3:
OK, I'll do it! 😀

PR #4368, opened 2022-07-25
Ignore ISSUE_TEMPLATE in prettier check

Kongkille:
I think it's OK if we ignore the issue templates. I would often like to use the web editor for small changes

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,3 +16,5 @@ web/public/dist/
 web/public/images/
 
 mockserver/public
+
+.github/ISSUE_TEMPLATE/


Kongkille:
CC @VIKTORVAV99 just in case you have any thoughts on this :)

VIKTORVAV99:
BTW @Kongkille would you mind if I updated these to [Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) instead?

Kongkille:
That would be cool, I haven't seen those yet

PR #4370, opened 2022-07-25
added ZONE_SHORTNAME_CONFIG

mathilde-daugy:
updated __init__ in config to include ZONE_SHORTNAME_CONFIG 
This config includes all zones with zone keys, country names, zone names. It will be useful for the zone overview project
image

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -8,10 +8,12 @@ Point = NewType("Point", Tuple[float, float])
 BoundingBox = NewType("BoundingBox", List[Point])
 
 CONFIG_DIR = Path(__file__).parent.parent.parent.parent.joinpath("config").resolve()
+LOCALES_DIR =Path(__file__).absolute().parent.parent.parent.parent.joinpath("web/public/locales").resolve()
 
 # Read JOSN files
 ZONES_CONFIG = json.load(open(CONFIG_DIR.joinpath("zones.json")))
 EXCHANGES_CONFIG = json.load(open(CONFIG_DIR.joinpath("exchanges.json")))
+ZONE_SHORTNAME_CONFIG = json.load(open(LOCALES_DIR.joinpath("en.json")))["zoneShortName"]
 CO2EQ_PARAMETERS_ALL = json.load(open(CONFIG_DIR.joinpath("co2eq_parameters_all.json")))
 CO2EQ_PARAMETERS_LIFECYCLE = {
     **CO2EQ_PARAMETERS_ALL,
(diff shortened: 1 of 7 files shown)


electricitymap/contrib/config/__init__.py
```diff
@@ -8,10 +8,12 @@
 BoundingBox = NewType("BoundingBox", List[Point])
 
 CONFIG_DIR = Path(__file__).parent.parent.parent.parent.joinpath("config").resolve()
+LOCALES_DIR =Path(__file__).absolute().parent.parent.parent.parent.joinpath("web/public/locales").resolve()
 
 # Read JOSN files
```
Kongkille, line 13:
```suggestion
# Read JSON files
```

Kongkille (approved):
Sounds good to me

VIKTORVAV99:
Not exactly sure why but the changes in this PR seems to cause issues for me when running the test parser. 
Particularly this part (I think):
```python
ZONE_SHORTNAME_CONFIG = json.load(open(LOCALES_DIR.joinpath("en.json")))[
    "zoneShortName"
]
```
It's causing me to get a `UnicodeDecodeError`:
```error
File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 10250: character maps to <undefined>
```

System stats to help determine if this is a local issue:
OS: Windows 11 build 25158
Python Version: Python 3.8.10 64-bit (windows store)

@Kongkille or @mathilde-daugy if you can't reproduce it could I recommend reverting the changes in this PR until we know why it fails?

Kongkille:
@VIKTORVAV99 It's working on Mac, and I am having some difficulties getting my Windows setup up and running. We have an internal data debugger running that is now using this export, so it would be great if we could avoid reverting it altogether

Based on Google search it could be something to do with the lack of specified encoding. If you get a chance could you try to add encoding='utf-8' to the open function?

VIKTORVAV99:
> @VIKTORVAV99 It's working on Mac, and I am having some difficulties getting my Windows setup up and running. We have an internal data debugger running that is now using this export, so it would be great if we could avoid reverting it altogether
> 
> Based on Google search it could be something to do with the lack of specified encoding. If you get a chance could you try to add encoding='utf-8' to the open function?

Okay no worries.

I'll give it a try in and report back once I have. 👍🏻

PR #4375, opened 2022-07-26
Update IL capacities

pierresegonne:
### Description

Updated capacities for IL based on Ember's 2020 data.

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -264,6 +264,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Israel
   - [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)
   - [LNRG capacities data](https://www.lnrg.technology/app/download/12390579357/IL+electricity+capacities.json)
+  - [Ember](https://ember-climate.org/data/data-explorer/)
 - Italy
   - Coal and Oil: [ENTSO-E](https://transparency.entsoe.eu/generation/r2/installedCapacityPerProductionUnit/show)
   - Other: [Terna](https://www.terna.it/it/sistema-elettrico/statistiche/pubblicazioni-statistiche)
--- a/config/zones.json
+++ b/config/zones.json
@@ -1991,13 +1991,13 @@
     "capacity": {
       "biomass": 27,
       "coal": 4840,
-      "gas": 12776,
+      "gas": 11010,
       "geothermal": 0,
       "hydro": 6.72,
       "hydro storage": 150,
       "nuclear": 0,
       "oil": 44,
-      "solar": 1962,
+      "solar": 2470,
       "unknown": 175,
       "wind": 28
     },


DATA_SOURCES.md
```diff
@@ -264,6 +264,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Israel
   - [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)
   - [LNRG capacities data](https://www.lnrg.technology/app/download/12390579357/IL+electricity+capacities.json)
+  - [Ember](https://ember-climate.org/data/data-explorer/)
```
AndBerna, line 267:
```suggestion
  - [Ember](https://ember-climate.org/data/data-explorer/)
  - [Irena] (https://pxweb.irena.org/pxweb/en/IRENASTAT)
  - [GEM] (https://globalenergymonitor.org/projects/global-gas-plant-tracker/tracker-map/)
```

config/zones.json
```diff
@@ -1991,13 +1991,13 @@
     "capacity": {
       "biomass": 27,
       "coal": 4840,
-      "gas": 12776,
+      "gas": 11010,
```
AndBerna, line 1994:
```suggestion
      "gas": 12143.00,
```

AndBerna: Based on the active Gas power plants at Israel https://globalenergymonitor.org/projects/global-gas-plant-tracker/tracker-map/

AndBerna (changes requested):

config/zones.json
```diff
@@ -1991,13 +1991,13 @@
...
       "geothermal": 0,
       "hydro": 6.72,
       "hydro storage": 150,
       "nuclear": 0,
       "oil": 44,
-      "solar": 1962,
+      "solar": 2470,
```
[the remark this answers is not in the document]
AndBerna: Based on Renewable Energy Stats available at https://pxweb.irena.org/pxweb/en/IRENASTAT/IRENASTAT__Power%20Capacity%20and%20Generation/ELECCAP_2022_cycle2.px/table/tableViewLayout2/

DATA_SOURCES.md
```diff
@@ -264,6 +264,9 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Israel
   - [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)
```
AndBerna, line 265:
```suggestion
```

DATA_SOURCES.md
```diff
@@ -264,6 +264,9 @@ For many European countries, data is available from [ENTSO-E](https://transparen
...
   - [LNRG capacities data](https://www.lnrg.technology/app/download/12390579357/IL+electricity+capacities.json)
+  - [Ember](https://ember-climate.org/data/data-explorer/)
```
AndBerna, line 267:
```suggestion
```

DATA_SOURCES.md
```diff
@@ -264,6 +264,9 @@ For many European countries, data is available from [ENTSO-E](https://transparen
 - Israel
   - [Global Power Plant Database](http://datasets.wri.org/dataset/globalpowerplantdatabase)
   - [LNRG capacities data](https://www.lnrg.technology/app/download/12390579357/IL+electricity+capacities.json)
```
AndBerna, line 266:
```suggestion
```

AndBerna (changes requested):
Not used data sources should be removed

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
+            irena_url=IRENA_CAPACITY_URL, irena_query=IRENA_SOLAR_THERMAL_QUERY
+        )
+    except AssertionError:
+        solar_thermal = pd.DataFrame()
+
+    solar = pd.concat([solar_pv, solar_thermal]).groupby("year").sum()
+
+    # load solar capacity to data
+    try:
+        data["capacity"] = {"solar": solar.loc[data["datetime"].year, "capacity"]}
```
pierresegonne, line 180: That's going to fail if the provided `target_datetime` is not a `datetime` compatible object (ex: `str`)

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
-        "datetime": arrow.now(TZ).datetime,
-        "production": {"unknown": production[0] + production[1]},
-        "source": IEC_URL,
-    }
+    data = {"zoneKey": zone_key, "datetime": arrow.now(TZ).datetime}
+
+    # fetch pv capacity from IRENA
+
+    try:
+        assert (
```
pierresegonne, line 150: Let's extract the parsing of the capacity to its own function.

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
+            post(url=IRENA_CAPACITY_URL, json=IRENA_SOLAR_PV_QUERY).status_code == 200
```
pierresegonne, line 151: By doing this you're calling the URL twice. Let's put this in the `fetch_IRENA_capacity` base function

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
+        data["capacity"] = {"solar": solar.loc[data["datetime"].year, "capacity"]}
+    except KeyError:
+        logger.warning(
+            f"Solar capacity for {zone_key} at {data['datetime']} not available"
+        )
+
+    # fetch production from IEC
+
+    try:
+        assert (
```
pierresegonne, line 189: Same as above here, this will call that URL twice, which is wasteful. Do the check directly on what is received when trying to fetch the production

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
+            get(IEC_PRODUCTION).status_code == 200
+        ), "IEC production url is not accessible"
+        # current production parser is not able to parse past dates
+        if target_datetime:
```
pierresegonne, line 193: I think there is an issue here, we don't want the parser to crash if there is a provided `target_datetime`. We want to fetch the capacity data if there is some, even if the production parser does not support historical values

parsers/IL.py
```diff
@@ -88,19 +141,66 @@ def extract_price_date(soup):
...
+        assert (
+            post(url=IRENA_CAPACITY_URL, json=IRENA_SOLAR_PV_QUERY).status_code == 200
+        ), "IRENA solar PV url is not accessible"
+        if target_datetime:
+            data["datetime"] = arrow.get(target_datetime).datetime
+        solar_pv = fetch_IRENA_capacity(
+            irena_url=IRENA_CAPACITY_URL, irena_query=IRENA_SOLAR_PV_QUERY
+        )
+
+    except AssertionError:
```
pierresegonne, line 159: This whole try catch structure makes the code complicated. 

What I would suggest: 

1. In the functions (whether for production or capacity), you check whether there is data that can be fetched. If yes then return that data if not return an empty data object, while logging an Error (and not warning) saying that that data point can't be fetched.
2. Then assemble these data objects within the `fetch_production` function.
3. (Optional) if the concatenated data objects are empty it means all individual parsers (production, capacity, ..) have failed, you can raise an exception to make the parser fail. It's Optional as we would already be returning empty data while logging an error

pierresegonne (commented):
Few comments, let me know if it's not clear

pierresegonne:
@AndBerna are you taking care of cleaning this up or should I?

AndBerna:
> @AndBerna are you taking care of cleaning this up or should I?

I am currently on it

pierresegonne:
It's my own PR so I can't approve but LGTM

AndBerna:
> It's my own PR so I can't approve but LGTM, so feel free to approve and merge

Noted 👍 , just one thing, there is one test failing `CI / Prettier / Check (pull_request) Failing after 18s — Prettier / Check`. It is actually the one I mentioned the other day. If you could mention how to fix it, it would be of great help. Tak in advance!

AndBerna (approved):

PR #4376, opened 2022-07-26
Add estimation method to ZA

Kongkille:
South Africa was not properly displayed because there was no parser and no estimation method available. 

For now, I've added time slicer as the estimation method

image

### DIFF
--- a/config/zones.json
+++ b/config/zones.json
@@ -6809,7 +6809,8 @@
       [15.969981316000087, -47.46575286299992],
       [38.47779381600009, -21.626451924999927]
     ],
-    "timezone": null
+    "timezone": null,
+    "estimation_method": "RECONSTRUCT_BREAKDOWN"
   },
   "ZM": {
     "bounding_box": [
--- a/config/zones.json
+++ b/config/zones.json
@@ -6809,7 +6809,7 @@
       [15.969981316000087, -47.46575286299992],
       [38.47779381600009, -21.626451924999927]
     ],
-    "timezone": null,
+    "timezone": "Africa/Johannesburg",
     "estimation_method": "ESTIMATED_TIME_SLICER_AVERAGE"
   },
   "ZM": {
--- a/config/zones.json
+++ b/config/zones.json
@@ -6810,7 +6810,7 @@
       [38.47779381600009, -21.626451924999927]
     ],
     "timezone": null,
-    "estimation_method": "RECONSTRUCT_BREAKDOWN"
+    "estimation_method": "ESTIMATED_TIME_SLICER_AVERAGE"
   },
   "ZM": {
     "bounding_box": [


VIKTORVAV99:
Is this needed if we get #4363 merged?

config/zones.json
```diff
@@ -6809,7 +6809,8 @@
       [15.969981316000087, -47.46575286299992],
       [38.47779381600009, -21.626451924999927]
     ],
-    "timezone": null
+    "timezone": null,
```
AndBerna, line 6812:
```suggestion
    "timezone": "Africa/Johannesburg",
```

AndBerna (changes requested):

config/zones.json
AndBerna: Otherwise everything looks nice 😄

Kongkille:
> Is this needed if we get #4363 merged?

Good point, no perhaps I went a bit too fast here. Depends a bit on the timeline of the other PR as right now it's a bit confusing to users that there is a color on ZA but no data. 

My plan is to update the app with a new version on Friday, so if we can get the new parser in before then we don't need this one. 

I'll mark this as blocked for now!

VIKTORVAV99:
> > Is this needed if we get #4363 merged?
> 
> Good point, no perhaps I went a bit too fast here. Depends a bit on the timeline of the other PR as right now it's a bit confusing to users that there is a color on ZA but no data.
> 
> My plan is to update the app with a new version on Friday, so if we can get the new parser in before then we don't need this one.
> 
> I'll mark this as blocked for now!

The other PR should be good to go but the CI is failing for no obvious reason we haven't figured out yet. But everything works locally.

VIKTORVAV99:
> > > Is this needed if we get #4363 merged?
> > 
> > 
> > Good point, no perhaps I went a bit too fast here. Depends a bit on the timeline of the other PR as right now it's a bit confusing to users that there is a color on ZA but no data.
> > My plan is to update the app with a new version on Friday, so if we can get the new parser in before then we don't need this one.
> > I'll mark this as blocked for now!
> 
> The other PR should be good to go but the CI is failing for no obvious reason we haven't figured out yet. But everything works locally.

Figured out what was causing the issue (but not why) so that PR should be ready to merge in my view, I have already reviewed it and didn't see any issues after the last revision. So if you guys want to review it as well we can have it merged today (just apply my suggested fix if IV1T3 haven't already).

pierresegonne:
I've not followed this change, but why do we now have to specify "estimation_method" in `zones.json`?

VIKTORVAV99:
> I've not followed this change, but why do we now have to specify "estimation_method" in `zones.json`?

I believe it's due to a change in the frontend so it only loads the "data view" if the zone has either a parser or a estimation method, currently South Africa has neither. I don't know why this was needed though.

Kongkille:
> > I've not followed this change, but why do we now have to specify "estimation_method" in `zones.json`?
> 
> I believe it's due to a change in the frontend so it only loads the "data view" if the zone has either a parser or a estimation method, currently South Africa has neither. I don't know why this was needed though.

We have many different data modes, before we had estimation zones our way of knowing the status of a zone was if there was a parser. However, with the estimation zones we now have zones that have no parsers but still have data. As such, this is a way for the frontend to know whether or not to show data for the zone.

I think there's some slight improvements to this, as the current case of ZA should not happen

pierresegonne:
In any case, we'll activate time slicer average on ZA so I guess we can merge this one no?

Kongkille:
> In any case, we'll activate time slicer average on ZA so I guess we can merge this one no?

We don't actually need the estimation_method flag for time slicer as there is usually a parser for those and it would be a bit cumbersome to add the flag for all zones that can default to time slicer.

When I initially had the idea for this PR I had thought the estimation method for ZA was a reconstruct breakdown and so it would have made more sense to add it in the config.

I think I am just going to close this PR, as we probably don't want to start adding unnecessary flags to the config

pierresegonne:
Makes sense!

VIKTORVAV99:
I think the preferable method would be if the backend added these fields to the state file based on what method was used. 

I don't know how it works right now but I assume it could use various different estimation methods for a single zone during different periods?

I.e. one to fill in missing values and one to predict "real time" values?

Kongkille:
> I think the preferable method would be if the backend added these fields to the state file based on what method was used.
> 
> I don't know how it works right now but I assume it could use various different estimation methods for a single zone during different periods?
> 
> I.e. one to fill in missing values and one to predict "real time" values?

Yes perhaps the state file could contain this. I initially wanted the zones.json to contain enough information to know whether or we should expect data from a zone, so we were able to not fetch for a certain zone and immediately display the data status.

However, realistically the state file has almost always been queried at the time a user clicks on a country so it would probably be just as fine to send it through there.

On the other hand, this would put more functionality away from open source, which is also a bit unfortunate.

VIKTORVAV99:
> However, realistically the state file has almost always been queried at the time a user clicks on a country so it would probably be just as fine to send it through there.

Yeah and I don't think people would be very likely to click on grey countries and they wouldn't be colored unless the state file is loaded.

> On the other hand, this would put more functionality away from open source, which is also a bit unfortunate.

While true and I agree it's unfortunate there is also nothing else people that contribute in the frontend can (currently) do to work on the estimation models/system so I think it might be better to keep it separate anyway.

Kongkille:
Good points, I've added this as an internal issue :)

PR #4377, opened 2022-07-26
add `encoding='utf-8'` to open methods in `__init__.py`

VIKTORVAV99:
The fix suggested in #4370 works great so this updates the open methods to include `encoding='utf-8'`.

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -16,19 +16,27 @@ LOCALES_DIR = (
 )
 
 # Read JSON files
-ZONES_CONFIG = json.load(open(CONFIG_DIR.joinpath("zones.json")))
-EXCHANGES_CONFIG = json.load(open(CONFIG_DIR.joinpath("exchanges.json")))
-ZONE_SHORTNAME_CONFIG = json.load(open(LOCALES_DIR.joinpath("en.json")))[
-    "zoneShortName"
-]
-CO2EQ_PARAMETERS_ALL = json.load(open(CONFIG_DIR.joinpath("co2eq_parameters_all.json")))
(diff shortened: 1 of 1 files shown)


pierresegonne:
I'll just check that this also fixes the internal problem we've been facing

Kongkille (approved):

Kongkille:
Oh sorry @pierresegonne I had just run it, should I revert?

pierresegonne:
lol

pierresegonne:
I'll revert if I see something that does not work

pierresegonne:
Didn't seem to create an issue, the issue I see on our end is related to imports and filesystem, but now it raises the warning

```
 /usr/local/lib/python3.8/json/__init__.py:293: DeprecationWarning: 'encoding' is ignored and deprecated. It will be removed in Python 3.9
#9 4.542     return loads(fp.read()
```

It seems that in Python 3.9 `encoding` will be deprecated https://bugs.python.org/issue39377

VIKTORVAV99:
> Didn't seem to create an issue, the issue I see on our end is related to imports and filesystem, but now it raises the warning
> 
> ```
>  /usr/local/lib/python3.8/json/__init__.py:293: DeprecationWarning: 'encoding' is ignored and deprecated. It will be removed in Python 3.9
> #9 4.542     return loads(fp.read()
> ```
> 
> It seems that in Python 3.9 `encoding` will be deprecated https://bugs.python.org/issue39377

That issue seems strange, it says it should be ignored from 3.1 but it fixes my problems locally on 3.8 so it's obviously not ignored. But perhaps we can remove it after we upgrade to 3.9 (or 3.10 preferably).

pierresegonne:
Hmm I think I've found a way to silence the warnings. Can you check @VIKTORVAV99 that adding the `encoding` kwarg within `open` rather than `load` still works for you?

pierresegonne:
My bad it was already the case. All good then!

PR #4378, opened 2022-07-26
Remove ZONE_SHORTNAME_CONFIG

pierresegonne:
### Description

Removing the `ZONE_SHORTNAME_CONFIG` from the init file of the contrib config as it forces all libs/projects depending on contrib to load it, which in turn requires loading the `en.json` file to the filesystems of our different images. 

As we're only using it within the debugger I suggest loading it only there

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -8,21 +8,12 @@ Point = NewType("Point", Tuple[float, float])
 BoundingBox = NewType("BoundingBox", List[Point])
 
 CONFIG_DIR = Path(__file__).parent.parent.parent.parent.joinpath("config").resolve()
-LOCALES_DIR = (
-    Path(__file__)
-    .absolute()
-    .parent.parent.parent.parent.joinpath("web/public/locales")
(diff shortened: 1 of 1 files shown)


gwpicard (approved):

Kongkille:
I'll let you decide on this one but allowing other projects to use metadata around the names of zones could be quite useful. It's not like all projects use all configs regardless. If we start loading various files around in contrib without explicitly declaring the dependency we could end up with bugs and inconsistencies later on


Didn't earthly also allow us to specify dependencies in more details?

pierresegonne:
Yes indeed, but we don't need this per se to do it. The problem with that first approach is that it forces all projects to load these files even though it only concerns the debugger, a completely internal project. The less useless dependencies the better. 

>Didn't earthly also allow us to specify dependencies in more details?

Yes as I demoed, you can make a mini build that only exposes e.g config files from contrib and import that within a project.
But in any case, the less amount of unused files we have to load within other projects the better. If we see that we need that attribute to be exposed multiple times, then we could make it default to be exposed by contrib.

PR #4392, opened 2022-07-28
fix(TH): fix `fetch_consumption()` return type error

shuuji3:
This PR fixes the error where `fetch_consumption()` returns a wrong data structure. Essentially, this is the same error as #4364.

**execution log**

```py
'<' not supported between instances of 'dict' and 'int'

  File "/home/contrib/electricitymap/contrib/parsers/lib/quality.py", line 33, in validate_consumption
    if (obj.get("consumption") or 0) < 0:       
```

This will fix the issue itself. However, I wonder if the consumption value is correct. Because the comment in `fetch_EGAT()` says "Consumption = Production + Import" but the current implementation only uses the same value as the production and ignores import value.


VIKTORVAV99 (approved):
Nice you got this working!

PR #4397, opened 2022-07-29
Update python + lowered image footprint

xmichele:
Hello, congrats for the interesting project.
Here you find the mod that include:
- upgrade to python 3.10 (https://github.com/electricitymap/electricitymap-contrib/issues/4337)
- upgraded debian slim version + instructions to reduce image footprint (lowered to 1.42GB quite 1GB less than the current one!)
- more security and less surface attack thanks to the upgraded & slim version of the base image


VIKTORVAV99:
It's sadly not as simple as this, the electricity maps backed run custom images and the python version need to be bumped there first.

That's why that issue is assigned to @Kongkille.

xmichele:
> It's sadly not as simple as this, the electricity maps backed run custom images and the python version need to be bumped there first.
> 
> That's why that issue is assigned to @Kongkille.

Hello, I changed to gitlab yaml accordingly  as well.
poetry.lock seems the same.

VIKTORVAV99:
> > It's sadly not as simple as this, the electricity maps backed run custom images and the python version need to be bumped there first.
> > 
> > That's why that issue is assigned to @Kongkille.
> 
> Hello, I changed to gitlab yaml accordingly  as well.
> poetry.lock seems the same.
> 

I didn't mean the the circleCI images. I was referring to the backend images that are outside this repository.

Until those are updated this can't be merged cause it will cause a version mismatch between the parsers and backend.

.circleci/config.yml
```diff
@@ -8,20 +8,17 @@ executors:
 jobs:
   verify_python:
     docker:
-      - image: circleci/python:3.8.12
+      - image: cimg/python:3.10.5
```
jarek, line 11: for anyone else wondering about this change:

https://circleci.com/developer/images/image/cimg/python

> This image is designed to supercede the legacy CircleCI Python image, circleci/python.

PR #4399, opened 2022-07-29
Pyhon 3.10.5 update + image footprint reduction

xmichele:
Hello, congrats for the interesting project.
Here you find the mod that include:

upgrade to python 3.10 (https://github.com/electricitymap/electricitymap-contrib/issues/4337)
upgraded debian slim version + instructions to reduce image footprint (lowered to 1.42GB quite 1GB less than the current one!)
more security and less surface attack thanks to the upgraded & slim version of the base image


xmichele:
@VIKTORVAV99 I read about external references outside this repo.
If you can mention them, you can do it :)

parsers/BR.py
```diff
@@ -1,4 +1,4 @@
-from collections import defaultdict
+from collections.abc import defaultdict
```
jarek, line 1: why these changes? it seems to break tests 

```
parsers/BR.py:1: in <module>
    from collections.abc import defaultdict
E   ImportError: cannot import name 'defaultdict' from 'collections.abc' (/home/circleci/.pyenv/versions/3.10.5/lib/python3.10/collections/abc.py)
```

and I don't see a mention of defaultdict in https://docs.python.org/3.10/library/collections.abc.html

VIKTORVAV99: I believe it's due to a depreciation in python 3.9, there are warnings about it in the "normal" ci runs. Haven't had time to check it yet though.

VIKTORVAV99:
> @VIKTORVAV99 I read about external references outside this repo.
> If you can mention them, you can do it :)

Sadly I can't update them either, someone from the electricity maps team has to do it.

Issue #4337, opened 2022-07-11
Update backend to use Python 3.10


VIKTORVAV99:
We currently use Python 3.8 in the backend and therefore also in the parsers. It would be nice to update this to Python 3.10 so we can take advantage of new features and syntax. 

Tasks:
- [x] #4405
- [ ] #5866 

>**Note**
> This Issue can only be solved by one of the Electricity Maps team members as it involves backend changes outside this repo.

xmichele:
Hello,
congrats for the interesting project!
Here yo can find the fix + other perks (https://github.com/electricitymap/electricitymap-contrib/pull/4397)
[Success Run log](https://github.com/electricitymap/electricitymap-contrib/files/9218786/updated-electricitymap-contrib-run.log)
@Kongkille I'll appreciate your fb as well :)
Please consider me if any devOps, cloud topic or also something related to bash script, java, python, dart languages.
Thanks :)

Kongkille:
Hi @xmichele, thanks a lot for looking into this. 

I'm frankly not the best person in terms of upgrading our python version. I'm adding @pierresegonne and @FelixDQ here 

They are both quite busy on other projects however, so it might just take some time for us to get this through.

madsnedergaard:
Copying the response from #4405 to here for visibility:

> Note that it will probably still take a little while before we can update to python 3.10. We lock the same python version between all our systems, so we need to do the same exercise on all our backend components before we can update the python version here. The last time I checked, there still were some dependencies that didn't support python 3.10 yet, but I'll give it another look when I have time.

VIKTORVAV99:
Maybe we should skip 3.10 and go straight to 3.11, python 3.11 have some substantial performance improvements:
> CPython 3.11 is on average 25% faster than CPython 3.10 when measured with the pyperformance benchmark suite, and compiled with GCC on Ubuntu Linux. Depending on your workload, the speedup could be up to 10-60% faster.

Link to release notes: https://docs.python.org/release/3.11.0/whatsnew/3.11.html

Performance specific section: https://docs.python.org/release/3.11.0/whatsnew/3.11.html#whatsnew311-faster-cpython

PR #4405, opened 2022-08-01
Bump dependencies for python 3.10 compat

VIKTORVAV99:
This PR makes it possible to update to Python 3.10 by updating two dependencies to remove deprecation warnings that would be break in Python 3.10.

PS: Someone take a look at the lockfile, the changes in there looks weird to me.

Reference: #4337


FelixDQ (approved):
LGTM!

Note that it will probably still take a little while before we can update to python 3.10. We lock the same python version between all our systems, so we need to do the same exercise on all our backend components before we can update the python version here. The last time I checked, there still were some dependencies that didn't support python 3.10 yet, but I'll give it another look when I have time.

VIKTORVAV99:
> LGTM!
> 
> Note that it will probably still take a little while before we can update to python 3.10. We lock the same python version between all our systems, so we need to do the same exercise on all our backend components before we can update the python version here. The last time I checked, there still were some dependencies that didn't support python 3.10 yet, but I'll give it another look when I have time.

No worries, I just want to have everything ready here for when you do decide to upgrade!

PR #4408, opened 2022-08-01
Add ZA non-renewable capacities

AndBerna:
Adds ZA non-renewable production capacity based on last available [Eskom report
](https://www.eskom.co.za/wp-content/uploads/2021/08/2021IntegratedReport.pdf) (page 6)


config/zones.json
```diff
@@ -6810,6 +6810,13 @@
       [15.969981316000087, -47.46575286299992],
       [38.47779381600009, -21.626451924999927]
     ],
+    "capacity": {
+      "coal": 38.773,
```
pierresegonne, line 6814: Didn't you mistake thousands delimiters with decimals?

AndBerna: No, but I put the info in GW, as it was stated in the front-end.  Good catch. I will correct it now to MW

pierresegonne (approved):

PR #4411, opened 2022-08-03
Updated DE capacity

Bohne13:
Data is from 01.08.22 from Fraunhofer ISE.
For battery storage I have used the max. installed power output and not the total installed GWh of capacity.


VIKTORVAV99 (approved):
Since it's the same source as before this looks good to me.

And using the max output from the batteries is indeed correct.

PR #4412, opened 2022-08-04
Fix link formatting for Israel and South Africa

j-jayes:
Hyperlinks in the section called [sources by region](https://github.com/electricitymap/electricitymap-contrib/blob/master/DATA_SOURCES.md#sources-by-region) had [source] (url). Changed to [source](url).

Thank you for the super interesting map!


j-jayes:
This is my first pull request so apologies if I haven't done the right thing!

Kongkille (approved):
Ah nice, thanks for the fix!

PR #4418, opened 2022-08-05
Fix ZA production format error

IV1T3:
This PR fixes the format issues seen in #4416.


parsers/ZA.py
```diff
@@ -708,6 +708,27 @@ def get_power_bi_values():
     return cleaned_power_bi_values
 
 
+def valid_substitution(original, subs_key):
```
VIKTORVAV99, line 711:
```suggestion
def valid_substitution(original: str, subs_key: str):
```
Adding type hints here will make sure the `original` (csv) and `sub_key` is always a string and it will have cascading effects in the function.

VIKTORVAV99 (commented):
It's a little hard to test this as what it tires to fix seems to be a intermittent issue but as far as I can tell there is no regression at least.

I just had a few comments regarding type hints but beyond that it looks good to me.

>**Note**
>There are some other missing type hints in the file but I think those are outside this PR.

VIKTORVAV99 (approved):
LGTM 🎉

PR #4422, opened 2022-08-07
Properly re-render map upon updates

corradio:
I noticed a bug where the map colors of zones outside of the viewport would only be updated once a dragging operation finished. This caused some zones to appear gray during dragging, and only be coloured once the dragging finished, leaving the user to think that these zones had no data.

The issues comes from the fact that we previously only updated zones in viewport using `queryRenderedFeatures`. It made sense when the colour change was a temporary color change, meaning that it was impossible to drag the map around while changing colors.
Now, with the historical view, it is possible to change colors, drag the map around, and thus we need to change our approach and re-render all zones.

This PR fixes that. I've added inline comments to ease review.


web/src/components/zonemap.jsx
```diff
@@ -94,7 +102,10 @@ const ZoneMap = ({
         features,
       },
     };
-  }, [zones, selectedTimeAggregate]);
+    // TODO: `zoneValues` will change even in cases where the geometry doesn't change.
+    // This will cause this memo to re-update although it should only update when the
+    // geometry changes. This will slow down the map render..
+  }, [zoneValues, selectedTimeAggregate]);
```
corradio, line 108: A performance optimisation is possible, see https://github.com/electricitymap/electricitymap-contrib/issues/4423

web/src/components/zonemap.jsx
```diff
@@ -70,15 +71,22 @@ const ZoneMap = ({
...
+    const features = zoneValues.map((zone, i) => {
       const length = (coordinate) => (coordinate ? coordinate.length : 0);
+      const zoneId = zone.config.countryCode;
       return {
         type: 'Feature',
         geometry: {
           ...zone.geography.geometry,
           coordinates: zone.geography.geometry.coordinates.filter(length), // Remove empty geometries
         },
+        id: i, // assign an integer id so the feature can be updated later on
```
corradio, line 89: by setting an id (which needs to be an integer) we can later have a lightweight operation that updates the features by querying by id.

web/src/components/zonemap.jsx
```diff
@@ -126,34 +137,21 @@ const ZoneMap = ({
     }
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
-  // TODO: Consider moving the calculation to a useMemo function
-  // change color of zones if timeslider is changed
-  useEffect(() => {
-    if (isLoaded && co2ColorScale) {
-      // TODO: This will only change RENDERED zones, so if you change the time in Europe and zoom out, go to US, it will not be updated!
-      // TODO: Consider using isdragging or similar to update this when new zones are rendered
-      const features = ref.current.queryRenderedFeatures();
```
corradio, line 135: It doesn't make sense to only re-render visible zones as the user might drag. We need to update them all instead

web/src/components/zonemap.jsx
```diff
@@ -126,34 +137,21 @@ const ZoneMap = ({
...
+  useMemo(() => {
+    if (isLoaded) {
       const map = ref.current.getMap();
-      features.forEach((feature) => {
-        const { color, zoneId } = feature.properties;
-        let fillColor = color;
-        const zoneData = zones[zoneId]?.[selectedTimeAggregate].overviews[selectedZoneTimeIndex];
-
+      zoneValues.forEach((zone, i) => {
```
corradio, line 143: Iterate over zones instead of features, and then query features by id.

web/src/components/zonemap.jsx
```diff
@@ -126,34 +137,21 @@ const ZoneMap = ({
...
-        if (selectedZoneTimeIndex !== null && co2intensity) {
-          fillColor = co2ColorScale(co2intensity);
-        }
-        const existingColor = feature.id
-          ? map.getFeatureState({ source: 'zones-clickable', id: feature.id }, 'color')?.color
-          : color;
+        const existingColor = map.getFeatureState({ source: 'zones-clickable', id: i }, 'color')?.color;
 
-        if (feature.id && fillColor !== existingColor) {
+        if (fillColor !== existingColor) {
```
corradio, line 150: This might not even be required. If we omit this, then we don't need to call `map.getFeatureState` L148. Didn't test which version was fastest.

web/src/hooks/theme.js
```diff
@@ -21,7 +21,8 @@ export function useCo2ColorScale() {
   const theme = useTheme();
 
   return useMemo(
-    () => scaleLinear().domain(theme.co2Scale.steps).range(theme.co2Scale.colors).unknown('gray').clamp(true),
```
corradio, line 24: The `gray` unknown color actually doesn't represent the gray we use on the map. This is now fixed. It enables to simplify the code in `ZoneMap` by simply calling the colorscale and not bothering to check if the input value is null.

corradio:
@tonypls @Kongkille I just opened the app and it remained gray until I dragged it.
I believe this is caused by the underlying issue fixed by this PR, where the first map render is gray.

Kongkille (commented):
Overall this seems to be a better approach. It keeps renders very minimal as far as I could see.

web/src/components/zonemap.jsx
```diff
@@ -43,6 +43,7 @@ const ZoneMap = ({
   const selectedTimeAggregate = useSelector((state) => state.application.selectedTimeAggregate);
   const electricityMixMode = useSelector((state) => state.application.electricityMixMode);
   const zones = useSelector((state) => state.data.zones);
+  const zoneValues = useMemo(() => Object.values(zones), [zones]);
```
Kongkille, line 46: What are your thoughts on having a separate object for the zoneValues?

corradio: Only impact is memory use, as the object will only be reconstructed when needed.
I think we should reconsider the datastructure here, see https://github.com/electricitymap/electricitymap-contrib/issues/4423.

Kongkille: Yeah, perhaps it's a nitpick but I'd expect the selector to handle this logic, anyways perhaps it's not easily possible. It's probably not worth spending a lot of time on.

As for an update in the datastructure, I'd rather not change it completely until we know how we're going to handle runtime changes in geometries for the spatial aggregations

Kongkille (approved):

web/src/components/zonemap.jsx
corradio: > Yeah, perhaps it's a nitpick but I'd expect the selector to handle this logic, anyways perhaps it's not easily possible. It's probably not worth spending a lot of time on.

I'm not sure I follow. What logic do you think should be handled by the selector?

> As for an update in the datastructure, I'd rather not change it completely until we know how we're going to handle runtime changes in geometries for the spatial aggregations

Fully aligned!

Kongkille: Well so i'm thinking that the we would not need to do a UseMemo on the zoneValue but instead be able to use to zones object. The selector as far as I understand also does a comparison check and only triggers a rerender on change

corradio: > Well so i'm thinking that the we would not need to do a UseMemo on the zoneValue but instead be able to use to zones object

The issue is that `Object.values(zones)` isn't guaranteed to yield results in the same order each time it is called, so even if `zones` doesn't change, the values object could have a different order. That's why we can't use `Object.values(zones)` multiple times: it needs to only be called once.
Also, `Object.values` always returns a *new* object, so any memo that has `zoneValue` as dependency would re-render all the time. That's why a `useMemo` is used. It guarantees that if `zones` doesn't change, then `zoneValue` will not change either.
Hope it makes sense!

> The selector as far as I understand also does a comparison check and only triggers a rerender on change

Correct, it would only recompute if the `zones` object changes.

Kongkille: Ah ok that makes sense. Let's do it like this then :)

PR #4425, opened 2022-08-07
Style price graph

corradio:
The co2 graph is coloured based on the y-value:
image

Therefore, I thought it would make sense to do the same with the price graph:
image


VIKTORVAV99:
What happens if the price is negative? I don't think the current scaling take that into account?

Perhaps changing this:
```js
const priceColorScale = scaleLinear().domain([0, priceMaxValue]).range(['lightgray', '#616161']);
```

to this:
```js
const priceColorScale = scaleLinear().domain([0, Math.Max(priceMaxValue, Math.abs(priceMinValue))]).range(['lightgray', '#616161']);
```

This was just a quick thought and I haven't tested it so some other changes might be required.

corradio:
Good point! I think the scale should actually start at the minimum price, and end at the maximum price.

web/src/components/countryhistorypricesgraph.jsx
```diff
@@ -19,7 +19,8 @@ const prepareGraphData = (historyData) => {
   const valueAxisLabel = `${currencySymbol || '?'} / MWh`;
 
   const priceMaxValue = d3Max(historyData.map((d) => (d.price || {}).value));
```
VIKTORVAV99, line 21:
```suggestion
  const priceMaxValue = d3Max(historyData.map((d) => d.price?.value));
```
GitHub wouldn't let me add a suggestion on the other lines for some reason but I think it would be nice if we cleaned up the code a bit here (both the new and existing code)

Optional Chaining should also be used on line 18 and 22.

corradio: Good points. I'll await review from @Kongkille before making all the changes in one batch if that's fine on your side.

web/src/components/countryhistorypricesgraph.jsx
```diff
@@ -29,8 +30,8 @@ const prepareGraphData = (historyData) => {
   }));
 
   const layerKeys = ['price'];
-  const layerStroke = () => 'darkgray';
-  const layerFill = () => '#616161';
+  const layerStroke = undefined;
```
Kongkille, line 33: It seems odd to have a const undefined, does everything break if you just gently remove it?

Kongkille (commented):
Looks really good,

We might want to deprecate d3 in the future for more modern chart libraries but I don't know exactly when that'd be.

Just waiting for the final changes before approving :)

web/src/components/countryhistorypricesgraph.jsx
```diff
@@ -18,8 +18,9 @@ const prepareGraphData = (historyData) => {
   const currencySymbol = getSymbolFromCurrency(((historyData.at(0) || {}).price || {}).currency);
```
VIKTORVAV99, line 18:
```suggestion
  const currencySymbol = getSymbolFromCurrency(historyData.at(0)?.price?.currency);
```

VIKTORVAV99 (commented):
I haven't really been able to test it with negative prices as the mockdata we have don't contain any, but the code looks good except one minor suggestion I have.

Kongkille (approved):

PR #4431, opened 2022-08-08
Handle grid and zone request failure and retry

tonypls:
## Issue
Failed zone requests do not fail gracefully

## Description

This PR enables users to get feedback when grid or zone api requests fail and allow them to retry


web/src/layout/main.jsx
```diff
@@ -150,13 +152,20 @@ const Main = ({ electricityMixMode, hasConnectionWarning }) => {
...
                 onClick={(e) => {
-                  dispatch(GRID_DATA_FETCH_REQUESTED());
+                  if (failedRequestType === 'grid') {
+                    dispatch(GRID_DATA_FETCH_REQUESTED({ features, selectedTimeAggregate }));
+                  }
+                  if (failedRequestType === 'zone') {
+                    dispatch(
+                      ZONE_HISTORY_FETCH_REQUESTED({ zoneId: failedRequestZoneId, features, selectedTimeAggregate })
+                    );
+                  }
```
Kongkille, line 168: Since we are adding a bit more logic to this component, I'd perhaps move it to its own component eg <ConnectionWarning />

web/src/helpers/redux.js
```diff
@@ -63,7 +63,8 @@ function initDataState() {
   });
 
   return {
-    hasConnectionWarning: false,
+    failedRequestType: null,
+    failedRequestZoneId: null,
```
Kongkille, line 67: I'm wondering whether it is necessary to have the specific failedRequestZoneId stored in state. In other parts of the application we are retrieving the current zoneId through

Since the user is only able to see one zone at any given time, the failedRequestZoneId should also always be the same as the one in params.

web/src/reducers/dataReducer.js
```diff
@@ -52,20 +52,22 @@ const reducer = createReducer(initialState, (builder) => {
...
       state.hasInitializedGrid = true;
       state.isGridExpired[stateAggregation] = false;
     })
     .addCase(GRID_DATA_FETCH_REQUESTED, (state) => {
       state.isLoadingGrid = true;
-      state.hasConnectionWarning = false;
     })
     .addCase(GRID_DATA_FETCH_FAILED, (state) => {
-      state.hasConnectionWarning = true;
+      state.failedRequestType = 'grid';
```
Kongkille, line 63: What do you think about storing the the names as constants, we have contants.js file where some of these things could be stored.

Eg:

failedRequestType : {
GRID: 'grid',
ZONE: 'zone',
}

And then we access it through failedRequestType.GRID

Kongkille (commented):
Looks great. Added a few comments

I think it's a nice fix to reuse the current connectionwarning and get something live sooner rather than later! :)

web/src/layout/leftpanel/countrypanel.jsx
```diff
@@ -193,7 +194,7 @@ const CountryHeader = ({ parentPage, zoneId, data, isMobile, isDataAggregated })
 const CountryPanel = ({ electricityMixMode, isMobile, tableDisplayEmissions, zones }) => {
   const [tooltip, setTooltip] = useState(null);
   const { __ } = useTranslation();
-
+  const { failedRequestType } = useSelector((state) => state.data);
```
Kongkille, line 197: For performance, as far as I understand you'd want to avoid destructuring in the redux selector and instead go directly to the object you need

https://stackoverflow.com/questions/59072200/useselector-destructuring-vs-multiple-calls

Kongkille (approved):

Issue #4396, opened 2022-07-28
[RFC] Split out capacities from zones.json to facilitate historical capacities


VIKTORVAV99:
We now have historical production information in the map but I found myself missing seeing the installed capacities on these new time periods.

So in this RFC I request we split out the capacities from the zones.json file into a new capacity.json file.

This new file would only contain capacities (either just zone capacities or zone + exchange capacities) but would allow for historical values.

I propose we adopt the same structure as for the co2eq values and allow the capacity object for each zone to be either a single capacity object or a array of objects containing the capacity object and a datetime (and possibly source).

VIKTORVAV99:
Another small benefit from this would be auto labeling for PRs that affects the capacity 😅

Kongkille:
I like this idea,

For context, when we added historical view we disabled capacity because the live one of course is not valid for the full 5 year period. Having multiple capacity values seems like a good idea however there are some things we'd need to consider

Off the top of my head:

- Which time periods are capacity valid for, eg for a full year, or less?
- Do we accept multiple different sources for capacity
- How do we update this retrospectively with the current capacities, i.e. some of them may be quite recent whereas others may be from a few years ago

Also adding this discussion for some context on splitting out zones.json #2962

VIKTORVAV99:
> I like this idea,
> 
> For context, when we added historical view we disabled capacity because the live one of course is not valid for the full 5 year period. Having multiple capacity values seems like a good idea however there are some things we'd need to consider
> 
> Off the top of my head:
> 
> * Which time periods are capacity valid for, eg for a full year, or less?
> * Do we accept multiple different sources for capacity
> * How do we update this retrospectively with the current capacities, i.e. some of them may be quite recent whereas others may be from a few years ago
> 
> Also adding this discussion for some context on splitting out zones.json #2962

My ide was to use the same structure as the co2eq intensities so:
* A capacity would be valid until a newer entry is found, this would also allow os to backfill capacities without much of a problem.
```json5
"capacities": {
  "DK-DK1": [
    {
      "datetime": "2020-01-01" // This would be valid from 2020-01-01 to indefinitely.
      "capacity": "capacityObject"
    }
  ],
  "DK-DK2": [
    {
      "datetime": "2020-01-01" // This would be valid from 2020-01-01 to 2020-12-31.
      "capacity": "capacityObject"
    },
    {
      "datetime": "2021-01-01" // This would be valid from 2021-01-01 to indefinitely.
[...]

PR #4435, opened 2022-08-09
Create config per zone

pierresegonne:
## Description

This PR changes the configuration system to support zone configuration as a file per zone. 

The configuration is in `.yaml` for better readability and those yaml files are parsed and merged to generate the same config objects that we are currently using. 

I'd suggest doing some subsequent cleaning work to make the configuration objects be more intuitive. 

I also suggest removing the current config file in a subsequent PR to minimise the risk of breaking changes.

## Preview

Example for a zone for which the configuration is pretty simple: `BD.yaml`

```yaml
bounding_box:
- - 87.52178959200009
  - 20.23871491100003
- - 93.14285119700017
  - 27.123544007000064
contributors:
- systemcatch
- alixunderplatz
parsers:
  production: BD.fetch_production
timezone: Asia/Dhaka
```

And for an exchange, `AT->HU.yaml`

```yaml
lonlat:
- 16.605363
- 47.444264
parsers:
  exchange: ENTSOE.fetch_exchange
  exchangeForecast: ENTSOE.fetch_exchange_forecast
rotation: 120
```


VIKTORVAV99:
Jeez that is a lot of changed files... 
Although I think this is a good change overall I'm just curious if this could implement the capacity changes I suggested in #4396 but modified for individual files instead?

Also don't forget to run prettier, thats a lot of warnings... 😅

VIKTORVAV99:
I noticed a problem with the file names, Windows will not let you put `>` in the name and honestly I have no idea what will happen if you try and open a file with it.
image
So while I am in favor of this change this needs to be addressed before it's merged from the codebase otherwise I'm afraid it will exclude a lot of people from contributing to the repo.

pierresegonne:
> Although I think this is a good change overall I'm just curious if this could implement the capacity changes I suggested in https://github.com/electricitymap/electricitymap-contrib/issues/4396 but modified for individual files instead?

Yeah I can take that into account, we could even split up `zones`, `exchanges`, `capacities`, `emission_factors`, `fallback_zone_mixes`

pierresegonne:
> Windows will not let you put > in the name and honestly I have no idea what will happen if you try and open a file with it.

Is it because of the `>` or because you didn't provide an extension file? 

Otherwise I'll find a way around

VIKTORVAV99:
> > Windows will not let you put > in the name and honestly I have no idea what will happen if you try and open a file with it.
> 
> Is it because of the `>` or because you didn't provide an extension file? 
> 
> Otherwise I'll find a way around

It's cause of the `>` it's a reserved character in windows filenames.

> Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
>
> The following reserved characters:
>
> < (less than)
> > (greater than)
> : (colon)
> " (double quote)
> / (forward slash)
> \ (backslash)
> | (vertical bar or pipe)
> ? (question mark)
> * (asterisk)

VIKTORVAV99:
> > Although I think this is a good change overall I'm just curious if this could implement the capacity changes I suggested in https://github.com/electricitymap/electricitymap-contrib/issues/4396 but modified for individual files instead?
> 
> Yeah I can take that into account, we could even split up `zones`, `exchanges`, `capacities`, `emission_factors`, `fallback_zone_mixes`

It would be nice to have everything related to a zone in its 'zone config' file. Go for it if you have the time. 👍🏻

pierresegonne:
> It would be nice to have everything related to a zone in its 'zone config' file. Go for it if you have the time. 👍🏻

You mean not splitting up the config then? I thought that the other issue was suggesting to split away capacity config into its own config file.

VIKTORVAV99:
> > It would be nice to have everything related to a zone in its 'zone config' file. Go for it if you have the time. 👍🏻
> 
> You mean not splitting up the config then? I thought that the other issue was suggesting to split away capacity config into its own config file.

I don't think we need capacities separate if we have one config file for each zone. Like in this PR, one of the motivations to split it out was that if we supported historical capacities the zone.json file would be huge and hard to navigate. But splitting it up so there is one file per zone already solves that.

VIKTORVAV99:
> I like this a lot for readability and smaller files will also help on version control / merge conflicts.
> 
> I'd like to investigate a bit the following
> 
> - Any issues reading in the frontend? I haven't tried to import yaml files before.
> - Any performance considerations when reading this many files? 
> - Any performance considerations when working with this many files (eg file search in VSCODE)
> 
> 
> I am also wondering if we'd actually need the exchanges in seperate files or if maybe we can just lump all those together? We are not changing those as much so the benefit of separating them out may not be as much.
> 

I haven't checked if it's needed yet but there is  https://www.npmjs.com/package/yaml-loader to load and bundle yaml.

It should just function as JSON after that.

However I'm not totally sure yaml is better than JSON for this case, it might be better from a pure readability perspective but then you'll have to take whitespace into account when editing. There are pros and cons to both though.

If they are bundled there should be no major preformance issues in the front-end at least.

As for searching with vscode it will definitely have a overhead but I think it might be worth it cause it will be easier to find things manually. At least if you know what you are looking for.

web/src/helpers/zonedata.js
```diff
@@ -1,3 +1,46 @@
+const yaml = require('js-yaml');
+const path = require('path');
+const fs = require('fs');
```
VIKTORVAV99, line 3: Sadly we can't use `fs` in the frontend as it is a `nodejs` specific module.

Any and all "yaml bundling" has to happen in the build step.

Kongkille: I am actually not sure how to do this. Any thoughts on how to import all yaml files in the directory?

VIKTORVAV99: I know how to import the YAML files, that is easy with something like [yaml-loader](https://www.npmjs.com/package/yaml-loader).

The problem I haven't figured out yet is how to combine the yaml files before webpack bundles them. But I think there is a way to run custom scripts as a part of the webpack build system but I think that would break snowpack. (We could just use webpack for local development as well with a separate webpack config instead tho.)

>**Note**
>This would be considerable easier if it was just JSON files as we could skip the whole YAML conversion step.

pierresegonne: found a workaround, see [#7e58bb23ad2b7c35f9bff8d167752fde27c46104](https://github.com/electricitymap/electricitymap-contrib/pull/4435/commits/7e58bb23ad2b7c35f9bff8d167752fde27c46104)

pierresegonne: The workaround forces the user to manually run the `yarn generate-zones-config` command, but it's required for local development, so that's the reason I did not extract it to automatically run at build time. Same thing as for the geo world update

VIKTORVAV99: > > > Windows will not let you put > in the name and honestly I have no idea what will happen if you try and open a file with it.
> > 
> > 
> > Is it because of the `>` or because you didn't provide an extension file?
> > Otherwise I'll find a way around
> 
> It's cause of the `>` it's a reserved character in windows filenames.
> 
> > Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
> > The following reserved characters:
> 
> ```
> < (less than)
> > (greater than)
> : (colon)
> " (double quote)
> / (forward slash)
> \ (backslash)
> | (vertical bar or pipe)
> ? (question mark)
> * (asterisk)
> ```
> 
> Fun fact: Windows work fine without file extensions when manually open a file with a program (that supports it) like vs code.

I think the easiest would be to just change the exchange filenames from `zoneKey1->zoneKey2` to `zoneKey1_to_zoneKey2`, what do you think?

VIKTORVAV99:
> > > Windows will not let you put > in the name and honestly I have no idea what will happen if you try and open a file with it.
> > 
> > 
> > Is it because of the `>` or because you didn't provide an extension file?
> > Otherwise I'll find a way around
> 
> It's cause of the `>` it's a reserved character in windows filenames.
> 
> > Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
> > The following reserved characters:
> 
> ```
> < (less than)
> > (greater than)
> : (colon)
> " (double quote)
> / (forward slash)
> \ (backslash)
> | (vertical bar or pipe)
> ? (question mark)
> * (asterisk)
> ```
> 
> Fun fact: Windows work fine without file extensions when manually open a file with a program (that supports it) like vs code.

I think the easiest would be to change the exchange filenames from `zoneKey1->zoneKey2` to `zoneKey1_to_zoneKey2`, what do you think?

pierresegonne:
> I think the easiest would be to change the exchange filenames from zoneKey1->zoneKey2 to zoneKey1_to_zoneKey2, what do you think?

I'm going for unicode right arrow instead: "→" It's more graphic but could be a bit confusing

VIKTORVAV99:
> > I think the easiest would be to change the exchange filenames from zoneKey1->zoneKey2 to zoneKey1_to_zoneKey2, what do you think?
> 
> I'm going for unicode right arrow instead: "→" It's more graphic but could be a bit confusing

Should be fine, otherwise I have a feeling we'll find out...

pierresegonne:
Ok should be good now
FYI with this current version, we're locking the new config per zone to the old zones.json (and same for exchanges and co2eq parameters)
I'm doing this to minimise the risk of a breaking change, but that means that I'll have to follow-up (next week) with a follow up which consists in cleaning all old references to `zones.json` that were not addressed here, and keeping the new config files as the single truth

web/src/helpers/zonedata.js
Kongkille: Hmm I am slightly worried about this process. if I understand correctly

1. This would mean all changes to any zone config, eg capacity changes requires the contributor to run yarn etc. Unless we set up some kind of automatic job, this would make it very hard for contributors who do not have a full development setup.
2. We duplicate all zone configs (in single file and the bundled version)
3. During local development, we'd also have to rerun the script all the time if we're working on something related to config.

The geometry process is in my opinion a bit different because it requires a lot of processing and validation and is not something we do that often.

Maybe there's a way for us to avoid bundling it, perhaps there's some kind of configuration within Webpack/snowpack that allows us to import it in separate files?

VIKTORVAV99: > Hmm I am slightly worried about this process. if I understand correctly
> 
> 1. This would mean all changes to any zone config, eg capacity changes requires the contributor to run yarn etc. Unless we set up some kind of automatic job, this would make it very hard for contributors who do not have a full development setup.
> 2. We duplicate all zone configs (in single file and the bundled version)
> 3. During local development, we'd also have to rerun the script all the time if we're working on something related to config.
> 
> The geometry process is in my opinion a bit different because it requires a lot of processing and validation and is not something we do that often.
> 
> Maybe there's a way for us to avoid bundling it, perhaps there's some kind of configuration within Webpack/snowpack that allows us to import it in separate files?
> 

We could git ignore the merged files and create them locally with a postinstall script.

That way it would be safe to edit them in the browser without adding an extra manual step locally but it would create some extra overhead on local development.

pierresegonne: Yep was thinking of something similar, I have I can find a good solution to this, stay tuned

VIKTORVAV99:
I think we need a postinstall script to the files are generated when the dependencies are installed for the CI to work properly (or we need to change the CI but the first option would be better).

pierresegonne:
So now, we reduce the need for contributors to worry about the aggregation of the configs:
* It's run officially at build time
* For any other command that require it (develop, build, test etc) the command is included in the script]
The files are ignored by the version control, so we only keep the version built at build time as the official one 

The potential weakness I see is if people try to circumvent the yarn commands, and end up with a confusing error message. Might thus be worth adding a piece of documentation that explains the whole setup, but I'm not sure yet where is the most appropriate place to do that

VIKTORVAV99:
@pierresegonne I don't think the script need to be added to that many places, I think it would be enough to add a prebuild, predevelop and postinstall script. But this works for now. 🙂

VIKTORVAV99:
Postinstall is working great!
image

VIKTORVAV99:
> > I think it would be enough to add a prebuild, predevelop and postinstall script
> 
> Wasn't aware that such a thing existed 😅

It's super handy, just take any exisiting script and add a new script with pre or post and they will chain. You can add both if you just need something temporary.

I think we can simplify a lot of our scripts this way but that's something for another PR. 🙂

VIKTORVAV99:
@pierresegonne I think this line in the dockerfile is causing the CI to fail:
```dockerfile
    volumes:
      - './config:/home/config'
```
It does not match the path for the web directories at:
```dockerfile
    volumes:
      - './config:/home/config'
      - './web/.eslintrc:/home/src/electricitymap/contrib/web/.eslintrc'
      - './web/geo:/home/src/electricitymap/contrib/web/geo'
      - './web/public/locales:/home/src/electricitymap/contrib/web/public/locales'
      - './web/locales-config.json:/home/src/electricitymap/contrib/web/locales-config.json'
      - './web/package.json:/home/src/electricitymap/contrib/web/package.json'
      - './web/server.js:/home/src/electricitymap/contrib/web/server.js'
      - './web/src:/home/src/electricitymap/contrib/web/src'
      - './web/views:/home/src/electricitymap/contrib/web/views'
      - './web/webpack.config.js:/home/src/electricitymap/contrib/web/webpack.config.js'
```

VIKTORVAV99:
@pierresegonne did you ever get around to changing the `->` in the file name to the unicode arrow?

pierresegonne:
> @pierresegonne did you ever get around to changing the -> in the file name to the unicode arrow?

Yes it's the case now :)

VIKTORVAV99:
> > @pierresegonne did you ever get around to changing the -> in the file name to the unicode arrow?
> 
> Yes it's the case now :)

Great! 🎉
Thanks for making the change. 🙂

pierresegonne:
@FelixDQ or @Kongkille can I get your review please? 

FYI @tonypls @madsnedergaard @mathilde-daugy

madsnedergaard:
This PR breaks Chrome when trying to review it 😄 
I'd like us to map out the different user journeys for working/changing these configurations to better understand the impact of the proposed changes here - unless you've already done some thoughts on this we could sit down with a whiteboard together? :)

pierresegonne:
> I'd like us to map out the different user journeys for working/changing these configurations to better understand the impact of the proposed changes here - unless you've already done some thoughts on this we could sit down with a whiteboard together? :)

Sure let's discuss today :)

pierresegonne:
@VIKTORVAV99 have you tested with the unicode arrows?

VIKTORVAV99:
> @VIKTORVAV99 have you tested with the unicode arrows?

Not much more than I can actually check out the pr now without errors, I've been pretty busy the last few days but I should have time to test it more in depth during the weekend.

madsnedergaard (approved):

VIKTORVAV99:
Before we merge this have we changed the bounding box helper to generate the bounding boxes in the new yaml files?

VIKTORVAV99:
I also have a slight concern about the usage of the Unicode arrows, visually they are great and easy to understand but it might create problems when trying to add new exchanges as it will be hard for users to find the correct arrow.

Now I don't believe this it's a blocker by itself but I do think it's something we should take a closer look at.

pierresegonne:
> Now I don't believe this it's a blocker by itself but I do think it's something we should take a closer look at.

It would be fairly quick to change, like changing 2 lines. I would recommend though to let it be like this, and make the switch if we see that contributors are confused or complain

web/package.json
```diff
@@ -28,16 +28,20 @@
     "date-fns": "^2.28.0",
     "ejs": "^2.5.7",
     "express": "^4.16.2",
+    "fs": "^0.0.1-security",
     "history": "^4.10.1",
     "i18next": "^21.6.13",
     "i18next-browser-languagedetector": "^6.1.3",
     "i18next-http-backend": "^1.3.2",
     "js-cookie": "^2.2.0",
+    "js-sha256": "^0.9.0",
```
tonypls, line 37: This is not longer used and can be removed

tonypls (commented):
I have tested this web app and was running well. I've tested it with master merged in ( containing the spatial aggregations branch 🎉 ) and it looks good. 

The only thing is there are some changes to the zones.json and exchanges.json that are not yet included, can be seen here #4627

pierresegonne:
> The only thing is there are some changes to the zones.json and exchanges.json that are not yet included, can be seen here https://github.com/electricitymaps/electricitymaps-contrib/pull/4627

I'll include those

electricitymap/contrib/config/__init__.py
```diff
@@ -4,34 +4,77 @@
...
+            if k in zone_config["emissionFactors"]:
+                if k == "direct":
+                    co2eq_parameters_direct["emissionFactors"]["zoneOverrides"][
+                        zone_key
+                    ] = zone_config["emissionFactors"][k]
+                elif k == "lifecycle":
+                    co2eq_parameters_lifecycle["emissionFactors"]["zoneOverrides"][
+                        zone_key
+                    ] = zone_config["emissionFactors"][k]
+        del zone_config["emissionFactors"]
```
jayaddison, line 69: @pierresegonne is there an important reason to delete the `emissionFactors` key from the zone configs after they're loaded?

As context: I'm looking into writing a test/validator that asserts on the equality of the `sources` values between the top-level zone config and individual emissions factor entries, and that caused me to notice that this key gets removed.

pierresegonne: Hey, 
It's purely for memory reasons. In the backend we use the `co2eq_parameters_**` config objects to interact with the emission factors, not the `zone_config`, so we remove that field to reduce clutter. 
That could be challenged :)

PR #4436, opened 2022-08-09
Split FO (Faroe Islands) into two subzones

VIKTORVAV99:
This PR modifies the FO parser to support two subzones, FO-NI (Northen Islands) and FO-SI (Southern Islands).

Zone keys and names are just the first thing I came up with since there is no ISO subdevision zones for FO.

Things left to do:
- [x] Add names for the zones in the translation file.
- [ ] Create geographies for the subzones.
- [x] Add capacities for the new zones.


Kongkille:
@VIKTORVAV99 I am doing some geometry work tomorrow, want me to add a split version of FO as well?

VIKTORVAV99:
That would be great! 🎉

Are you doing Ceuta and Mellila too?

Kongkille:
> That would be great! 🎉
> 
> Are you doing Ceuta and Mellila too?

Yes exactly

VIKTORVAV99:
Anyone have a good translation for the names on the [SEV website](https://www.sev.fo/framleidsla/el-orka-i-foroyum/)?

I tried using a translator for them and, well, one of the zones translated to South Korea...

If we don't find any good translations I think we should go with Northern Islands and Southern Island.

web/public/locales/en.json
```diff
@@ -634,6 +634,14 @@
     "FO": {
       "zoneName": "Faroe Islands"
     },
+    "FO-NI": {
+      "countryName": "Faroe Islands",
+      "zoneName": "Northern Islands"
```
VIKTORVAV99, line 639: I'm not totally sold on this name but I haven't found a better alternative yet.

VIKTORVAV99:
@Kongkille This should be ready to merge after the geometries are done but sadly there is no historical data available from the parser so it would take a day until it shows up on the map.

@pierresegonne requested a review from you to just look over the config changes, but should be pretty straight forward.

web/public/locales/en.json
pierresegonne: For the south it's easy: https://en.wikipedia.org/wiki/Su%C3%B0uroy -> Should translate to South Island
For the north it's a bit more tricky: meginøkið seems to mean something closer to "capital" (https://hagstova.fo/en/node/509 + google translate), so maybe it means like capital territory. But I guess Northern islands is fairer as https://en.wikipedia.org/wiki/Regions_of_the_Faroe_Islands shows that it groups all regons that are not the south island

config/co2eq_parameters_lifecycle.json
```diff
@@ -2937,6 +2937,12 @@
           "value": 17
         }
       },
+      "FO-NI": {
```
pierresegonne, line 2940: You can also put 0 for direct, otherwise it will fallback to the default direct unknown.

VIKTORVAV99: Will do!

web/public/locales/en.json
VIKTORVAV99: Isn't Northern Island just these 6 though?
image
That is what I got from my research at least so it would be better to have a more distinctive name.
But I wouldn't say Northern Islands are completely wrong either.

pierresegonne: Ok I've asked an icelandic friend of mine. Even though icelandic != feroese, it's pretty close. His take was that it means "main islands"

VIKTORVAV99: Okay then I think we should switch to Main Islands. I'll update the translations and zone keys to match later today.

VIKTORVAV99:
@pierresegonne I will close this PR and merge all the changes into #4318 to create a combined PR so we  get the geographies in sync. Just so you are aware.

Issue #4429, opened 2022-08-07
Dynamic loading screen


corradio:
**Is your feature request related to a problem? Please describe.**
The loading screen seen when the app is opened is currently static. I've seen users wonder what is happening.
An indication that this is a loading indicator would be nice

**Describe the solution you'd like**
We could simply have it blink? Many creative directions can be investigated, but having something quick and dirty would already help!

VIKTORVAV99:
Making it blink would probably be a good immediate fix but I and @Kongkille have talked a little about making the dots in the new logo a loading "spinner" (does not need to actually spin) in #4330 but I don't think either of us has had to look into that yet.

We would do this by animating the SVG itself, that way it will scale perfectly just as it currently does and it will keep the file size down.

waltonzt:
@VIKTORVAV99 @corradio Hey there, my name is Zack and I wanted to take a shot at contributing to the project. I created a pull request that animates the SVG logo and hooked it up where I saw fit. Please let me know what you think!

corradio:
Amazing! Someone from the Electricity Maps team will review soon :)

VIKTORVAV99:
I left a few comments on the PR but it looks good so far!

PR #4438, opened 2022-08-10
Dynamic Loading Screen

waltonzt:
Hello, I wanted to take a shot at #4429.

Please review and let me know what you think.


VIKTORVAV99:
The spinner looks good I suppose but I had imagined something more along these lines:
Spin-1s-200px
But I'll leave that up to the team to decide.

Besides that the it would be nice if we had a loading icon (just the icon no text) as well and that they where both was minified. If you need any help with the minification let me know.

Kongkille:
This looks great, I will just await review from our designer @rasmusjeppsen!

waltonzt:
@VIKTORVAV99 I can take a shot at creating something like that. Are you still thinking it moves in the same path as the one I provided?

Also what method would you recommend to minify?

VIKTORVAV99:
> @VIKTORVAV99 I can take a shot at creating something like that. Are you still thinking it moves in the same path as the one I provided?
> 
> Also what method would you recommend to minify?

Either spinning in the path you created or just fading in and out would work pretty well I think.

For minifying it I would run it through svgo (online [interface](https://jakearchibald.github.io/svgomg/ ) here) but you need to be really careful with the options as a lot of transforms are enabled by default.

waltonzt:
@VIKTORVAV99 What about these?

electricitymap-loading-icon
electricitymap-loading

VIKTORVAV99:
> @VIKTORVAV99 What about these?
> 
> electricitymap-loading-icon
> electricitymap-loading
> 
> 

Looks great! 🎉

Thanks for making these!

waltonzt:
@VIKTORVAV99 Awesome! I think I figured out the minify tool you shared so I will commit those changes.

rasmusjeppsen:
> This looks great, I will just await review from our designer @rasmusjeppsen!

@waltonzt I have nothing to add other than these are great! Thank you for creating these, Zack!

VIKTORVAV99:
@Kongkille Do you know if there is any other places we should use this "loading logo" or loading icon?

Otherwise I think this is ready to merge.

Kongkille (approved):
Sorry for the late review!

I think this is all the places at the moment, otherwise, we can just add it.

web/src/components/loadingplaceholder.jsx
```diff
@@ -1,8 +1,8 @@
 import styled from 'styled-components';
 
 export default styled.div`
-  background: transparent url(${resolvePath('images/electricitymap-icon.svg')}) no-repeat center center;
-  background-size: 2rem;
+  background: transparent url(${resolvePath('images/electricitymap-loading.svg')}) no-repeat center center;
+  background-size: 10rem;
```
VIKTORVAV99, line 5:
```suggestion
  background: transparent url(${resolvePath('images/electricitymap-loading-icon.svg')}) no-repeat center center;
  background-size: 10rem;
```
Are the background-size changes needed?

VIKTORVAV99 (changes requested):
Just noticed this isn't using the icon version.

web/src/components/loadingplaceholder.jsx
Kongkille: No those would be breaking, I've fixed it another commit

Kongkille: image

VIKTORVAV99 (approved):

PR #4443, opened 2022-08-11
Spatial Aggregations 🗺️ 

tonypls:
## Issue
Spatial Aggregations

## Description
This PR adds aggregation of zones to enable users to choose between an aggregated country view and a high fidelity view with detailed zones. This will make it easier to evaluate and compare countries.

This PR is version 1 of the spatial aggregations features to view the aggregated map at ```aggregatedViewFF=true``` to your local app.

Removes internal exchanges of aggregated zones
Implements a UI to toggle between the aggregated and detailed view

Associated issues:
https://github.com/electricitymap/electricitymap-contrib/discussions/3309
https://github.com/electricitymap/electricitymap-contrib/issues/4419

### Preview

image


VIKTORVAV99:
Hmm, would this create a secondary `world.json` or similar that contains the aggregated zones?

If so think we need to consider splitting the geographies from the main bundle and bundle them in separate bundles. (should just need a few config changes to webpack)

tonypls:
> 

Markus and I decided it would be a good idea to combine them into one world.json to make switching between the views a bit easier. The plan is to filter by an "aggregated" flag. Do you think this is a good approach?

VIKTORVAV99:
> > 
> 
> Markus and I decided it would be a good idea to combine them into one world.json to make switching between the views a bit easier. The plan is to filter by an "aggregated" flag. Do you think this is a good approach?

It really depends on how big the file will get, I think it's already something like 1.4 MB and 400 kb compressed.
So if it's going to be substantially larger it might be better to split them into two (or more) files and allow the client to request them separately to speed up the first page load and then fetch the rest later (part of the Static Web Project is adding a web worker that will do this automatically in the background).

But it does not need to be a part of this PR, we can evaluate the size changes and decide on a course of action after that.

Kongkille:
> > > 
> > 
> > 
> > Markus and I decided it would be a good idea to combine them into one world.json to make switching between the views a bit easier. The plan is to filter by an "aggregated" flag. Do you think this is a good approach?
> 
> It really depends on how big the file will get, I think it's already something like 1.4 MB and 400 kb compressed. So if it's going to be substantially larger it might be better to split them into two (or more) files and allow the client to request them separately to speed up the first page load and then fetch the rest later (part of the Static Web Project is adding a web worker that will do this automatically in the background).
> 
> But it does not need to be a part of this PR, we can evaluate the size changes and decide on a course of action after that.

Yes we considered both options, I think the increase in file size should be fairly minimal, but let's keep it in mind

tonypls:
Too see aggregated view add ```?aggregatedView=true``` to the url

VIKTORVAV99 (commented):
The aggregate geometries look good just some questions:
- The internal exchange arrows for countries with subzones are still visible, should they be hidden in this PR or will it be fixed in the backend?
- Would it be possible to switch to a different URL path instead of using Params? i.e. `/map/aggregated` instead of `/map?aggregatedView=true` and `zone/aggregated/DK` instead of `/zone/DK?aggregatedView=true`. It wouldn't change the functionality but it would make the URL's look a little cleaner.
- How will we display data when one or more of the subzones are missing data? Shetland Islands and Northern Territory in Australia comes to mind.
- Maybe not tied to this PR but how will we handle overseas territory like Réunion? (FR overseas territory)

Beyond those questions everything look good so far from my side, but I haven't dug into the details of the code yet.

tonypls:
> The aggregate geometries look good just some questions:
> 
> * The internal exchange arrows for countries with subzones are still visible, should they be hidden in this PR or will it be fixed in the backend?
> * Would it be possible to switch to a different URL path instead of using Params? i.e. `/map/aggregated` instead of `/map?aggregatedView=true` and `zone/aggregated/DK` instead of `/zone/DK?aggregatedView=true`. It wouldn't change the functionality but it would make the URL's look a little cleaner.
> * How will we display data when one or more of the subzones are missing data? Shetland Islands and Northern Territory in Australia comes to mind.
> * Maybe not tied to this PR but how will we handle overseas territory like Réunion? (FR overseas territory)
> 
> Beyond those questions everything look good so far from my side, but I haven't dug into the details of the code yet.

Hey Viktor!

Yes nice spotting :) Too keep the reviewing to manageable size this is a MVP of sorts of the spartial aggregations hidden behind the feature flag. The next version will have the internal exchange arrows and a way to toggle between the aggregated and detailed views. I'll update the PR description to reflect this.

I see what you mean and I'll look into it, currently we have a pattern ```https://app.electricitymaps.com/map?solar=true``` like this for the layers so it's nice to keep some consistency and makes it easy to read the code when we have one way of doing things.

At the moment we can't aggregate zones with missing data because we must provide accurate data and missing zones could skew the data.  

I will have a look into the overseas territories for V2 of this feature 👍

VIKTORVAV99:
> > The aggregate geometries look good just some questions:
> > 
> > * The internal exchange arrows for countries with subzones are still visible, should they be hidden in this PR or will it be fixed in the backend?
> > * Would it be possible to switch to a different URL path instead of using Params? i.e. `/map/aggregated` instead of `/map?aggregatedView=true` and `zone/aggregated/DK` instead of `/zone/DK?aggregatedView=true`. It wouldn't change the functionality but it would make the URL's look a little cleaner.
> > * How will we display data when one or more of the subzones are missing data? Shetland Islands and Northern Territory in Australia comes to mind.
> > * Maybe not tied to this PR but how will we handle overseas territory like Réunion? (FR overseas territory)
> > 
> > Beyond those questions everything look good so far from my side, but I haven't dug into the details of the code yet.
> 
> Hey Viktor!
> 
> Yes nice spotting :) Too keep the reviewing to manageable size this is a MVP of sorts of the spartial aggregations hidden behind the feature flag. The next version will have the internal exchange arrows and a way to toggle between the aggregated and detailed views. I'll update the PR description to reflect this.
> 
> I see what you mean and I'll look into it, currently we have a pattern `https://app.electricitymaps.com/map?solar=true` like this for the layers so it's nice to keep some consistency and makes it easy to read the code when we have one way of doing things.
> 
> At the moment we can't aggregate zones with missing data because we must provide accurate data and missing zones could skew the data.
> 
> I will have a look into the overseas territories for V2 of this feature 👍

Nice then this looks good to me and I look forward to the other PRs!

About the URL patterns:
I think it makes more sense of that the wind and solar use params since they can both be used at the same time but aggregations would be a binary choice (either aggregated or not) so a path make more sense for that.

And while this is definitely not required it would just look a little nicer when sharing links, but as long as we decide on something before it launches so we don't get a bunch of broken links I'm happy.

config/zones.json
```diff
@@ -1,4 +1,11 @@
 {
+  "AU": {},
+  "BR": {},
+  "CA": {},
+  "CL": {},
+  "JP": {},
+  "MY": {},
+  "US": {},
```
Kongkille, line 8: Would be cool if we could get the generate-bounding-boxes script generate bounding boxes here as well

web/geo/generate-aggregates.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const combinedZoneKeys = countryKeys.filter((x) => !unCombinedCountries.includes(x));
+
+  const zonesNotToAggregate = unCombinedCountries
+    .map((countryKey) => {
+      const unCombinedZones = features.filter((feature) => feature.properties.countryKey === countryKey);
+
+      return unCombinedZones;
+    })
+    .flatMap((feautures) => {
```
Kongkille, line 17: nitpick: typo here

web/geo/generate-aggregates.js
```diff
@@ -0,0 +1,51 @@
+const { union } = require('@turf/turf');
+
+const generateAggregates = (geojson) => {
+  const { features } = geojson;
```
Kongkille, line 4: Ideally we really want to use the 'getPolygons' method in utilites in the start so we ensure we are consistent.

In general when working with geojson data, it's almost always best to split everything out into polygons when working with it so you don't have a bunch of edge cases from multipart features

Then you can combine it again once you're done

web/geo/generate-aggregates.js
```diff
@@ -0,0 +1,51 @@
+const { union } = require('@turf/turf');
```
Kongkille, line 1: On the overall process:

Your process may be better, in my head I was thinking of an approach where we'd

1. use 'reduce' to get a zoneToPolygon map
2. Iterate through map and union all those where length > 1

I think we'd be able to shave a few JS lines off but it might not be worth it.

web/geo/generate-topojson.js
```diff
@@ -1,10 +1,10 @@
 const { topology } = require('topojson-server');
-const { getJSON, writeJSON, round } = require('./utilities');
+const { getJSON, writeJSON, round, fileExists } = require('./utilities');
 const turf = require('@turf/turf');
 
 function getCenter(geojson, zoneName) {
   const geojsonFeatures = geojson.features.filter((f) => f.properties.zoneName === zoneName);
-  if (geojsonFeatures.length !== 1) {
+  if (geojsonFeatures.length !== 1 && geojsonFeatures.length !== 2) {
```
Kongkille, line 7: Can you elaborate on this change? :)

web/geo/validate-v2.js
```diff
@@ -0,0 +1,187 @@
...
+// TODO: Improve this function so each check returns error messages,
+// so we can show all errors instead of taking them one at a time.
+function validateGeometryV2(fc, config) {
+  const fcHighFidelity = {
+    ...fc,
+    features: fc.features.filter((feature) => {
+      try {
+        return !feature.properties.aggregatedView;
+      } catch (e) {
+        console.log('Error: ', e, 'Feature: ', feature); // eslint-disable-line no-console
```
Kongkille, line 25: Hmm, I am wondering why this try/catch is required?

web/geo/validate-v2.js
```diff
@@ -0,0 +1,187 @@
...
+  if (fcAggregated.features.length) {
+    console.log('Validating geometries aggregated view..'); // eslint-disable-line no-console
+    zeroNullGeometries(fcAggregated);
+    containsRequiredProperties(fcAggregated);
+    zeroComplexPolygons(fcAggregated, config);
+    zeroNeighboringIds(fcAggregated);
+    zeroGaps(fcAggregated, config);
+    zeroOverlaps(fcAggregated, config);
+    matchesZonesConfig(fcAggregated, config);
+  }
```
Kongkille, line 58: Here we could create an inserter-function/wrapper so we don't have to duplicate all lines.

web/src/helpers/featureFlags.js
```diff
@@ -0,0 +1,7 @@
+export const isRemoteParam = () => {
+  return new URLSearchParams(window.location.search).get('remote') === 'true';
+};
+
+export const isAggregatedViewEnabled = () => {
+  return new URLSearchParams(window.location.search).get('aggregatedView') === 'true';
+};
```
Kongkille, line 7: I think we should use the useFeatureToggle hook that is already in place

If we want to ensure that we have everything in one place we can still keep the strings located in this file, eg

useFeatureToggle(AGGREGATED_VIEW)

Note: the remote flag is not considered a feature-flag at least in the way we've worked with the flag in the past.

config/zones.json
tonypls: TODO :)

web/geo/generate-aggregates.js
tonypls: fixed 👍

web/geo/generate-topojson.js
tonypls: Originally there was duplicated features for an aggregated and unaggregated map. This design has been updated to only have new features for the aggregated zones so this check has been removed.

web/geo/validate-v2.js
tonypls: It's not required but it's useful for debugging if the input data has any issues

web/geo/validate-v2.js
tonypls: ✂️

web/src/helpers/featureFlags.js
tonypls: TODO remove the feature flag hook, hooks can only be used within components so when setting redux data the hook doesn't have context

VIKTORVAV99:
Hi @tonypls, would you mind if I added the Swedish translations for the new strings in `en.json` directly in this PR?

tonypls:
> Hi @tonypls, would you mind if I added the Swedish translations for the new strings in `en.json` directly in this PR?

Yes that would be good 👍

config/zones.json
```diff
@@ -1,4 +1,11 @@
 {
+  "AU": {},
```
pierresegonne, line 2: Are these needed? If not, then I suggest not adding them to the config as it's used throughout the systems and might have unforeseen consequences

web/.prettierignore
```diff
@@ -5,5 +5,8 @@
 yarn.lock
 
 world.json
+world-aggregated.json
+excluded-aggregated-exchanges.json
```
pierresegonne, line 9: nitpick but maybe `exchanges-excluding-aggregated` is easier to understand and fits better with the rest of our naming conventions.

web/.prettierignore
```diff
@@ -5,5 +5,8 @@
...
+Dockerfile
```
pierresegonne, line 10: Why ignore the Dockerfile? (or rather why didn't we do that before if it's necessary?)

web/geo/update-world.js
```diff
@@ -1,20 +1,46 @@
...
   ERROR_PATH: path.resolve(__dirname, '.'),
-  MIN_AREA_HOLES: 600000000,
+  MIN_AREA_HOLES: 5000000,
   MAX_CONVEX_DEVIATION: 0.708,
   MIN_AREA_INTERSECTION: 500000,
-  SLIVER_RATIO: 0.01, // ratio of length and area to determine if the polygon is a sliver and should be ignored
+  SLIVER_RATIO: 0.0001, // ratio of length and area to determine if the polygon is a sliver and should be ignored
+  verifyNoUpdates: process.env.VERIFY_NO_UPDATES !== undefined,
+};
+const configV2 = {
```
pierresegonne, line 21: Will that V2 become the default version once merged officially?

pierresegonne (commented):
Did not review in depth the front end code but the rest looks good. Let me know whether you think the little comments I made make sense and then I can approve :)

config/zones.json
tonypls: Good spotting, it's an artefact from before I knew we explicitly defined sub-zones. I shall remove it.

web/.prettierignore
tonypls: Hmm, it's a list of exchanges to exclude when showing the aggregated view, maybe ````exchanges-to-exclude-from-aggregated-view``` but it's very long.

web/.prettierignore
tonypls: The precommit hook runs prettier locally on modified files and can't identify the correct parser for Dockerfile. I'm unsure how it passed in CI. It's a small file, hopefully we can keep it pretty :)

web/geo/update-world.js
tonypls: Yes, this will be released with the featureflag ```?aggregatedViewFF=true```  to use the json files built from this config. Once we remove the feature flag this will be the default config

web/.prettierignore
tonypls: I added a dev dependency to support dockerfile with prettier

VIKTORVAV99: > I added a dev dependency to support dockerfile with prettier

I don't think that will work with the current CI system since we just run prettier on the whole repo with NPX.

We would have to create a top level package.json and add it there too for it to work (and change the CI to install them).

tonypls: I think the NPX prettier must be able to figure out what parser to use for Dockerfile as it's been running on it previously, this dev dep was to allow us to run the husky precommit prettier hook locally.

pierresegonne (approved):

pierresegonne:
Let's make Sweden Sweden again

Issue #4137, opened 2022-05-17
Corsica is exporting 100% or more of its generation to Italy because of missing Corsica-Sardinia interconnector data


q--:
Corsica was added in #3978, but for some reason the carbon-intensity isn't calculated during most of the day. (See the carbon intensity in the top left and the carbon intensity graph in the screenshot below.)

The likely cause of this is that ElectricityMap thinks it's exporting more than 100% of its generation to Italy during those times (and even when intensity *is* available, it's 'exporting' close to 100%).

![Screenshot of Corsica electricity generation in ElectricityMap](https://user-images.githubusercontent.com/2470175/168850000-7c127741-1d48-4175-8695-ce5d9b3ff223.png)

madsnedergaard:
cc @pierresegonne, any idea why?

VIKTORVAV99:
This appears to be the same problem as in #3925.

Not sure what is causing it since I don't know how the backend code works but I haven't seen any issue with the regions affected that time since the fix was applied.

Perhaps it's something that should be run each time a new zone is added?

pierresegonne:
I don't have a clear answer, but I think that it probably comes down to the data not passing the sanity checks we have in the backend. This means that the data is not flowtraced for these hours, and thus no carbon intensity is produced. I'll dive a bit deeper into it.

q--:
Hmm, I just noticed it's exporting approximately 100% of generation to Italy. Even more during  times of the day when data is grayed out. That cannot be right.

VIKTORVAV99:
> Hmm, I just noticed it's exporting approximately 100% of generation to Italy right now. Even more (say, 138%) during times of the day when carbon intensity is grayed out. That cannot be right.

There is a missing exchange between Corsica, Sardinia and Central North Italy, that is probably why it can go above 100% at times. https://en.wikipedia.org/wiki/HVDC_Italy%E2%80%93Corsica%E2%80%93Sardinia

pierresegonne:
Indeed we have no parser for the Corsica-Sardinia exchange.
Plus, running a quick analysis, it's quite clear that it's the fact that total exports > total production that result in no carbon intensity. 
We could: 
1. Deactivate the FR-COR->IT-CNO exchange. That would ensure that we show data for COR, but I don't like the idea of showing data that we know if false
[...]

PR #4450, opened 2022-08-12
Create parser wrapper for `FR-COR->IT-SAR`

VIKTORVAV99:
This PR adds a wrapper that combines the data from the IT-SACOAC and IT-SACODC exchanges.

Fixes: #4137


VIKTORVAV99:
@pierresegonne This is a pretty crude solution to the problem but it works, if you have any suggestions on how to simplify things I'll be happy to implement them.

parsers/FR-COR_IT-SAR.py
```diff
@@ -0,0 +1,129 @@
...
+to produce a single data list for the FR-COR->IT-SAR exchange.
+"""
+
+EXCHANGE_FUNCTION_MAP: Dict[str, Callable] = {
+    "exchange": fetch_ENTSOE_exchange,
+    "exchange_forecast": fetch_ENTSOE_exchange_forecast,
+}
+
+
+def fetch_data(
```
pierresegonne, line 24: Clean! 

I guess the best in theory would be to modify the ENTSO-E parser to accept multiple mappings, but as it's a much bigger project, we should only do that if we another case where that happens.

parsers/ENTSOE.py
```diff
@@ -173,9 +175,13 @@
     "BY->UA": [ENTSOE_DOMAIN_MAPPINGS["BY"], "10Y1001C--00003F"],
     "DE->DK-DK1": [ENTSOE_DOMAIN_MAPPINGS["DE-LU"], ENTSOE_DOMAIN_MAPPINGS["DK-DK1"]],
     "DE->DK-DK2": [ENTSOE_DOMAIN_MAPPINGS["DE-LU"], ENTSOE_DOMAIN_MAPPINGS["DK-DK2"]],
-    "DE->NO-NO2": [ENTSOE_DOMAIN_MAPPINGS["DE-LU"], ENTSOE_DOMAIN_MAPPINGS["NO-NO2"]],
     "DE->SE-SE4": [ENTSOE_DOMAIN_MAPPINGS["DE-LU"], ENTSOE_DOMAIN_MAPPINGS["SE-SE4"]],
-    "FR-COR->IT-CNO": ["10Y1001A1001A893", ENTSOE_DOMAIN_MAPPINGS["IT-CNO"]],
+    "DK-DK2->SE": [ENTSOE_DOMAIN_MAPPINGS["DK-DK2"], ENTSOE_DOMAIN_MAPPINGS["SE-SE4"]],
```
pierresegonne, line 179: Not sure to understand why some of these changes are here?
Erroneous commit ?

pierresegonne (commented):
I'll approve once I understand why some other domain mappings have been modified :)

parsers/ENTSOE.py
VIKTORVAV99: Ah yeah those seem to have slipped in when I merged the master branch into this, I'll clean it up.

parsers/FR-COR_IT-SAR.py
VIKTORVAV99: I don't know how you guys will handle the historical data for aggregated region but one other motivation to add it to the ENTSO-E parser would be to fix the historical exchange between Sweden and NO-NO4.

parsers/ENTSOE.py
VIKTORVAV99: Should be fine now.

VIKTORVAV99:
@pierresegonne Don't really know what happened when I merged in the master branch (git is hard...)

But it looks good now.

pierresegonne (approved):
LGTM!

PR #4456, opened 2022-08-15
Add PL locales for country panel and time controller

cieciurm:
Hi,
Thanks for doing an awesome job by delivering electricity map!

## Issue

When browsing through the app, I have found couple of missing locales for PL.

## Description

This PR adds couple of missing translations for country panel (when it is estimated values) and adds header for time controller panel.


jarek (approved):
Thanks! Dziękujemy!

PR #4457, opened 2022-08-15
Use ENTSO-E for Nordic and Baltic exchanges

VIKTORVAV99:
## Issue
From the NL parser:
```python
# add NO data, fetch once for every hour
# This introduces an error, because it doesn't use the average power flow
# during the hour, but rather only the value during the first minute of the
# hour!
```
## Description

This PR changes all the Nordic and Baltic exchanges that used `statnett.py` to use `ENTSOE.py` instead so we get the full exchange data, this PR also adds exchange forecasts.

I have verified that all exchanges have data and all but the `RU-1` exchanges have exchange forecasts.

>**Note**
>This will also enable historical parsing for the affected exchanges.


VIKTORVAV99:
@pierresegonne I think the underlying issue here could also account for the exchange difference between Norway and Sweden and Finland and Sweden we saw in #3936.

VIKTORVAV99:
Looks good to me at first glance but should I create a new PR with a aggregate parser for NO-NO4->SE to fix the missing exchange volume? (this would just be a interim solution).

I could also update #4457 after this is merged so we still can use ENTSO-E for exchanges. 

I think the mockserver crashes are due to the now outdated world.json file, but I'll look into it and get back to you here.

VIKTORVAV99:
Converting this to a draft PR until #4469 is merged/resolved.

VIKTORVAV99:
@pierresegonne this should be ready now.
It also contains an aggregate parser for `NO-NO4->SE` beyond the simple mapping switches and exchange updates.

VIKTORVAV99:
If this is merged I suggest you guys refetch all the exchange data in the Nordic and Baltic from 2015-01-01 or earlier since it's all basically wrong now...

pierresegonne:
Great, thanks :)

Looks good to me, but @mathilde-daugy could you have a look tomorrow as part of the contrib hat?

config/exchanges.json
```diff
@@ -1544,15 +1551,16 @@
   "NO-NO4->RU-1": {
     "lonlat": [30.1437, 69.52],
     "parsers": {
-      "exchange": "statnett.fetch_exchange"
+      "exchange": "ENTSOE.fetch_exchange",
+      "exchangeForecast": "ENTSOE.fetch_exchange_forecast"
     },
     "rotation": 120
   },
```
VIKTORVAV99, line 1558: I did a final check and noticed this exchange is not in ENTSO-E at all. 
Curiously enough it is always zero at Statnett as well?

Should we just remove it?

mathilde-daugy: Looks like data on Statnett is 0 since 2020/10/24 and before that we have small values. 

it makes sense that the exchanges are 0 at the moment as all flows are stopped. And from what I've read in this article it's also normal that the values are so small. (https://thebarentsobserver.com/en/industry-and-energy/2022/04/nordic-grid-cuts-russian-power)

I think we should unarchive the parser and use it just for the exchange and monitor

mathilde-daugy: sounds good!

mathilde-daugy (approved):

tonypls:
Hey @VIKTORVAV99 

@mathilde-daugy and I have looked through the code and tested the branch, everything looks good and Mathilde will be able to refetch the exchange data after this is merged

VIKTORVAV99:
> Hey @VIKTORVAV99 
> 
> @mathilde-daugy and I have looked through the code and tested the branch, everything looks good and Mathilde will be able to refetch the exchange data after this is merged

Great! 🎉

Then I'll go ahead and merge this!

Gazer75:
What was missing from Statnett API? AFAIK that was live data no?
At least for border flow I believe its minute by minute data. Same as is visible on the map their website.
https://www.statnett.no/en/for-stakeholders-in-the-power-industry/data-from-the-power-system/

ENTSOE is just a forecast and then you'll have to wait for the hour to be over before final data is available.

VIKTORVAV99:
> What was missing from Statnett API? AFAIK that was live data no? At least for border flow I believe its minute by minute data. Same as is visible on the map their website. https://www.statnett.no/en/for-stakeholders-in-the-power-industry/data-from-the-power-system/
> 
> ENTSOE is just a forecast and then you'll have to wait for the hour to be over before final data is available.

The statnett data was "live" but it was not using the average for the time period (at least according to the issue in the NL parser), yes we will have to wait for the ENTSO-E data but it will be a average of the total flow for the hour giving us a better representation of the actual exchange.
(There where some small differences between the statnett parser and ENTSO-E when I compared them as well so I'm not just going by that issue.)

The parser also has a re-fetch frequency of 1 hour meaning if we where to re-fetch data it would require 48 requests for the same amount of data the ENTSO-E parser (re-fetch frequency of 2 days) returns in 1 request.

And since we use ENTSO-E elsewhere it will lead to less data mismatches due to methodology.

And lastly, ENTSO-E is not forecasted (although we have a forecast parser too), the whole map is delayed by 2 hours and if all TSOs follow the ENTSO-E standards all data should be in the Transparency Platform by then.

PR #4461, opened 2022-08-16
Update the generation of zone neighbours

pierresegonne:
## Issue

With the introduction of several zone levels, we have to redefine how we define the zone_neighbours

## Description

For test cases, see unit tests

@Kongkille does that config affect the app in any way?

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -1,7 +1,8 @@
 import json
 from copy import deepcopy
+from itertools import product
 from pathlib import Path
-from typing import Dict, List, NewType, Tuple
+from typing import Any, Dict, List, NewType, Set, Tuple
 
 ZoneKey = NewType("ZoneKey", str)
 Point = NewType("Point", Tuple[float, float])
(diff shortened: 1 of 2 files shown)


electricitymap/contrib/config/__init__.py
```diff
@@ -45,14 +46,39 @@
...
+        all_zone_names.reverse()
+        pairs.extend(list(product(*all_zone_names)))
+        for zone_name_1, zone_name_2 in pairs:
+            if zone_name_1 not in zone_neighbours:
+                zone_neighbours[zone_name_1] = set()
+            zone_neighbours[zone_name_1].add(zone_name_2)
+    return zone_neighbours
+
+
+ZONE_NEIGHBOURS = generate_zone_neighbours(EXCHANGES_CONFIG, ZONES_CONFIG)
```
corradio, line 81: `ZONE_NEIGHBOURS` is actually also changed in the next `for` block L84-104 where neighbours are added to parent zone.
Aren't we now doing the work twice now?

corradio (approved):

Kongkille:
It might affect some things, how does the final config look like?

PR #4465, opened 2022-08-18
Add generate exchanges for the config

pierresegonne:
## Issue

The exchanges static config file does not offer support for the exchanges that are at different geographical levels.

## Description

This PR generates at run time a complete exchanges dictionary with e.g both `NO-NO4->SE-SE1` & `NO->SE`.

It also opens a dictionary that matches zones with all the exchanges they are part of.


corradio:
Makes sense to me, but I'll let @FelixDQ review!

tonypls (approved):

PR #4478, opened 2022-08-22
Update neighbours

corradio:
PR required to be able to merge the split of zones.

The `ZONE_NEIGHBOURS` object now defines the graph over which zones can be flowtraced.
It therefore only includes zone pairs that:
- have an *exchange* parser (not just any parser!)
- do not have subzones (as parents are excluded from the flowtracing)


VIKTORVAV99:
This might be slightly off topic but is the parent zones (zones with subkeys) used in the backend?
I haven't really seen a place where they are used in the frontend so instead of testing if they have subzone keys it might be simpler to just remove the parent zones?

If that is not possible can we at least clean them up by removing the capacities and/or parsers?

electricitymap/contrib/config/__init__.py
```diff
@@ -173,6 +44,27 @@ def _can_zones_be_neighbours(
         for sub_zone_id in zone_config["subZoneNames"]:
             ZONE_PARENT[sub_zone_id] = zone_id
 
+# This object represents the edges of the flow-tracing graph
+ZONE_NEIGHBOURS: Dict[ZoneKey, List[ZoneKey]] = {}
```
pierresegonne, line 48: Can you replicate the structure I laid out with a generation function? 

This allows to unit test it and document more easily what the expected zone_neighbours object we get depending on the provided config file.

corradio:
> This might be slightly off topic but is the parent zones (zones with subkeys) used in the backend? I haven't really seen a place where they are used in the frontend so instead of testing if they have subzone keys it might be simpler to just remove the parent zones?

Yes it is used in the backend in order to be able to make country aggregates. I also foresee it to be used by the frontend when we give the ability to toggle between country and regional view, but the team team has more knowledge than me :)

> If that is not possible can we at least clean them up by removing the capacities and/or parsers?

We have seen cases where we want to collect both parent and subzone data for experimentation. For example, we might want to have collected SE subzone data alongside SE for a while, before making the final switch.
For capacities that's another story - I'll let the team comment here (or consider cleaning things up in another PR).

electricitymap/contrib/config/__init__.py
corradio: Done. I did not unit test it though.

corradio: Done

pierresegonne (commented):
Could you add a simple unit test for the function? I can take over if you don't have tiem

corradio:
> Could you add a simple unit test for the function? I can take over if you don't have tiem

would be terrific if you could add a few tests!

pierresegonne:
> would be terrific if you could add a few tests!

Will do

pierresegonne:
Should be tested now :)

pierresegonne (approved):
Ready to go!¬

PR #4479, opened 2022-08-22
electricitymap_social_image.png

rasmusjeppsen:
## Issue

The current social image does not feature the new Electricity Maps logo as well as newly added zones.

## Description

Updating social image with new logo and feature of South Africa.

## Preview

Skærmbillede 2022-08-22 kl  09 48 31


VIKTORVAV99:
Maybe no big deal but you might want to wait for the next release to get even more zones colored on the map. There are a few fixes and updates for parsers and exchanges that have not gone live yet that should color: 
- 2 existing zones
- add 2 new zones (with production data)
- split 1 zone into 2 (FO).

Would be nice to get those included as well.

Kongkille (approved):

Kongkille:
> Maybe no big deal but you might want to wait for the next release to get even more zones colored on the map. There are a few fixes and updates for parsers and exchanges that have not gone live yet that should color:
> 
> * 2 existing zones
> * add 2 new zones (with production data)
> * split 1 zone into 2 (FO).
> 
> Would be nice to get those included as well.

Good point, we're just getting this one in today because we wanted to do a post on ZA then we can just make more when we get those in again!

Issue #4387, opened 2022-07-27
Confusing output with 5 year historical view at El Hierro (Spain)


AndBerna:
@Kongkille minor issue on the historical view. Apart from that, great feature! 😍

**Describe the bug**
 When looking to the historical  ( 5 years) at el Hierro the results are confusing:
- No results provided ( despite there is at least, one year of data as it can be seen in the 12 months view)
- Empty production chart 
- Source outputs: ? 

**To Reproduce**

Steps to reproduce the behaviour:
1. Go to 'El Hierro (Spain)'
2. Click on '5 Years' 
3. Scroll down to 'Electricity production in the last 5 years'

**Expected behaviour**
- Yearly graph from the available years
- Carbon intensity from the available years
- No graph if there is no data displayed
-  Same source as 24h/30days/12 months ( if there is data displayed)

**Screenshots**
<img width="448" alt="Screenshot 2022-07-27 at 16 33 41" src="https://user-images.githubusercontent.com/45770697/181275134-adc0e99d-4398-435c-be44-ac5a12c0dcc0.png">

**Desktop:**
 - OS: MacOS
 - Browser: Chrome

strassburger:
New Info.
first date of data for El Hiero in "demanda.ree.es" is 2012-02-29
works:
https://demanda.ree.es/visiona/canarias/el_hierro5m/tablas/2012-02-29/2
does nor work:
https://demanda.ree.es/visiona/canarias/el_hierro5m/tablas/2012-02-28/2
Hope helps

Kongkille:
Yes so the reason this is happening is the way we are handlings time windows. Right now we will only show data if there's a value for all expected hours. I.e for 24 hours, we expect all 24 hours.

I've created a new issue to keep track of this

PR #4491, opened 2022-08-24
Allow missing entries in data

### LINKED ISSUE #4398: Allow missing entries in time window
## Description

Cause of #4387. 

The app-backend currently either returns nothing or all expected values. For example for 24 hours, if one hour is missing data completely we will not return any values.

This was implemented to ensure we do not become out of sync with history and state (#3949) and also to minimize the file sizes.

I.e if the length of the array is 24 and the other is 25 but the index is shared, the state and history will become out of sync.

## Solution

The main cause of complexity here is that we use an index to keep track of time. We use 'selectedZoneTimeIndex' which is an index that can both be used in history and state.

I think the optimal solution would be to deprecate 'selectedZoneTimeIndex' and instead simply have a 'selectedDatetime' and change the data structure to map from datetime to zoneData instead of using an index to map to zoneData.

I believe this is more intuitive to work with and guarantees that we do not become out of sync.

Alternative solution:

A potentially less demanding solution would just be to fill in missing values in the array. I.e. we always ensure that there is 24 entries for 24 hours and 5 values for 5 years.

Kongkille:
Fixes #4398 and #4387

I was touching the countryPanel, and decided to split it up to not add even more functionality to a crowded component. Arguably that could have been another PR but I think there's a lot of tech debt around in our components and it's OK that we clean up a bit while adding functionality.

Bonus: also enabled the hiring RSS which had been broken with the domain change

This requires some changes to the backend, but we can still merge this one and I'll make sure to release the app-backend and this one in the same update.


Kongkille:
Cypress failling because I updated mockdata, i'll fix it

Kongkille:
> I'm curious as to what will happen is there is just one or few co2intensity values for a period and how this will display in a graph.
> 
> Also there are still "hasData" variables in helpers/redux.js and dataReducer.js, these are meant to remain right?

Yes it won't look super good, however this is almost never the case. At least I haven't seen a case of this. 

And yes, so the idea was that 'hasData' defines whether there is any data for the zone and the isValid is per entry.

tonypls:
lgtm 👍

strassburger:
@Kongkille 
Thanks, now the 5year data is shown for El Hierro, at least last 4 years data are shown.
 2017 data still missing.
Checked first possible data  2012-03-01, Source:
https://demanda.ree.es/visiona/canarias/el_hierro5m/total/2012-03-01
Checked every single day, some days looked like data was generated just in the moment i asked for.

Looks like same for all canary Island data.
Also learned that El Hierro did not get really better in Co2 results, but spanish mainland is really getting better in CO2 last five years.

Kongkille:
That's great. Yeah so we don't have 2017 yet, but now we're able to show the other years!

Issue #4441, opened 2022-08-10
Frontend breaks at split screen (767 px width)


VIKTORVAV99:
**Describe the bug**
Frontend styling breaks when using the app split screen.

**To Reproduce**
1. Go to app.electricitymaps.com
2. Set width to 767 px.
3. See error

**Expected behavior**
A responsive style.

**Screenshots**
![image](https://user-images.githubusercontent.com/30777521/183989473-8859b38a-0e84-4323-b454-cc795e3a6f0c.png)

Kongkille:
Yikes that looks chaotic. Which browser were you using?

On chrome it seems to work as expected

![image](https://user-images.githubusercontent.com/45588799/184089371-ee96ad6b-90e9-4e41-8aa3-c2ae611a9c3d.png)
![image](https://user-images.githubusercontent.com/45588799/184089551-bb66ce2e-8125-444b-ab2d-2806e0ad18b5.png)

VIKTORVAV99:
I'm using Edge 104.0.1293.47 on Windows 11 with 125% scaling (if that matters).

VIKTORVAV99:
Tried it on another device that uses Brave `Version 1.42.95 Chromium: 104.0.5112.81` as well on Windows 11 Home Insider Preview build 25174.1010 with 150% scale.
Here is the result:
<img width="603" alt="image" src="https://user-images.githubusercontent.com/30777521/184828983-a2458761-1bce-4294-8b99-2d039f44fb69.png">

tonypls:
I've tested this on Mac with Chrome, Firefox and Safari and the layout is working as expected. Because this is quite an edge case, I don't think there is priority right now. @VIKTORVAV99 Feel free to close the issue and we can open it again if we have and more reports of this becoming a problem.

VIKTORVAV99:
> I've tested this on Mac with Chrome, Firefox and Safari and the layout is working as expected. Because this is quite an edge case, I don't think there is priority right now. @VIKTORVAV99 Feel free to close the issue and we can open it again if we have and more reports of this becoming a problem.

I would prefer to leave it open but I don't (and didn't) expect this to be a priority.
I'll see if I can try and narrow down the issue in the coming days but I understand if you want yo close it. 🙂
(Leaving it open would allow other contributors to take a crack at solving it.)

VIKTORVAV99:
[...]

PR #4494, opened 2022-08-24
Align CSS breakpoints

VIKTORVAV99:
Fixes: #4441


VIKTORVAV99:
@tonypls this works for me locally but the code doesn't scale very well when resizing windows in the dev console, so I had some trouble testing it. So please double check so everything works as expected when running it locally.

tonypls:
Hey Viktor excuse the lack of action on this PR, I think it would be good to have a helper or function to reliably determine if mobile or desktop. 

I feel like when we're checking .1 of a px it's a bit of a code smell where we should consider a more robust way to solve this problem across devices / browsers.

VIKTORVAV99:
> Hey Viktor excuse the lack of action on this PR, I think it would be good to have a helper or function to reliably determine if mobile or desktop.
> 
> I feel like when we're checking .1 of a px it's a bit of a code smell where we should consider a more robust way to solve this problem across devices / browsers.

I agree and this was mostly a "hotfix" on my side due to lack of time (and focus on other things) hopefully we can bring a more robust system in place in the future. If you want to close this PR and wait for said more robust system I completely understand but it's up to you and the rest of the team.

VIKTORVAV99:
The issue seems to have resolved itself (for some unknown reason?). So I'll close this PR.

PR #4496, opened 2022-08-25
Add Norway (NO) solar capacity

VIKTORVAV99:
Just adds solar capacity to Norway.

This will also get the added benefit of reduced data missrepresentation because it will show the solar production from Norway subzones as `?` instead of `0` (the production data has been missing from ENTSO-E so far).


VIKTORVAV99:
Would be nice if we could include this in the release today as well. 🙂

Kongkille (approved):

PR #4497, opened 2022-08-25
Return if source not loaded yet

Kongkille:
## Issue
There's an ongoing issue when visiting the app via Safari

image

## Description

It's a bit tricky. I believe this is some kind of race condition which happens because we initially load in the mapstyle with an empty sources object. On our first call to the useMemo we then don't have the sources yet and the code fails.

There might be a more elegant way to fix this but not that i can think of right now.

It was also a bit extreme IMO that every error is treated as webGLnotsupported, so I've changed that.

### DIFF
--- a/web/src/layout/map.jsx
+++ b/web/src/layout/map.jsx
@@ -85,8 +85,10 @@ export default () => {
     dispatchApplication('isLoadingMap', false);
 
     // Disable the map and redirect to zones ranking.
-    dispatchApplication('webGLSupported', false);
-    history.push({ pathname: '/ranking', search: location.search });
+    if (e.error === 'WebGL not supported') {
+      dispatchApplication('webGLSupported', false);
+      history.push({ pathname: '/ranking', search: location.search });
+    }
   };
 
   const handleMouseMove = useMemo(
--- a/web/src/components/zonemap.jsx
+++ b/web/src/components/zonemap.jsx
@@ -133,20 +133,23 @@ const ZoneMap = ({
   useEffect(() => {
     if (!ReactMapGL.supported()) {
       setIsSupported(false);
-      onMapError('WebGL not supported');
+      onMapError({ error: 'WebGL not supported' });
     }
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
   useMemo(() => {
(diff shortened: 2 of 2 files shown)


web/src/layout/map.jsx
```diff
@@ -85,8 +85,10 @@ export default () => {
     dispatchApplication('isLoadingMap', false);
 
     // Disable the map and redirect to zones ranking.
-    dispatchApplication('webGLSupported', false);
-    history.push({ pathname: '/ranking', search: location.search });
+    if (e.error === 'WebGL not supported') {
+      dispatchApplication('webGLSupported', false);
+      history.push({ pathname: '/ranking', search: location.search });
```
VIKTORVAV99, line 90: This should probably have been changed outside this PR already but unless I'm mistaken there is no way to navigate away from the `/ranking` page on mobile.

Now here it actually works out in our favor somewhat since the map won't show if WebGL isn't supported but still, this should probably be addressed in a more elegant manner in another PR.

VIKTORVAV99 (approved):
Not exactly sure how to test this since I have never run into this particular problem on the web or Android (and I don't have iOS devices) but code wise it looks good.

I left another comment regarding navigation but as I mention this should probably be addressed in another PR.

web/src/layout/map.jsx
VIKTORVAV99: Oh you do? 👀

I had some ideas around that and was thinking of doing some mockups for it that supported historical data and aggregates at Uni. (I am going to study Digital Design and Innovation (basically UX and graphic design)). But it would probably take a few weeks if not months.

Kongkille: > Oh you do? 👀

Yes, we found that people are less inclined to use and instead navigate using the map.

> I had some ideas around that and was thinking of doing some mockups for it that supported historical data and aggregates at Uni. (I am going to study Digital Design and Innovation (basically UX and graphic design)). But it would probably take a few weeks if not months.

That sounds cool - would love to see some thoughts around it. We're still not completely settled on it, it is however a slightly complex component to keep alive because we have to consider it on every UI change as well as any larger data changes.

Issue #4488, opened 2022-08-24
Add timestamps to all graphs in the sidebar


rasmusjeppsen:
**Issue**
As a user, I would like to see the exact timestamp when exploring the graphs in the sidepanel.

**Is your feature request related to a problem? Please describe.**
To see the exact time in the graph one must use the historical view slider to explore the graphs as the timestamp does not feature when hovering in most of the graphs: **[Click here to see demo](https://www.loom.com/share/599e89b816764f129493c8d1a14c37ee)**

**Describe the solution you'd like**
Currently, the timestamp is only featured when hovering on the "Carbon intensity in the last 24 hours" graph. A solution for this issue would be to add a similar timestamp to the top of all other graph modals in the sidebar:
- Electricity sources
<img width="564" alt="image" src="https://user-images.githubusercontent.com/58597172/186373678-0450b474-1790-47e1-ad09-9e53c37ac338.png">

- Origin of electricity in the last 24 hours (graph)
<img width="644" alt="image" src="https://user-images.githubusercontent.com/58597172/186373871-1e76aa33-db02-4285-9835-12a91346bf4a.png">

- Electricity prices in the last 24 hours (graph)
<img width="491" alt="image" src="https://user-images.githubusercontent.com/58597172/186373968-74e8513f-0456-4720-91d5-6cd29f23c9ba.png">

_I propose the position of the timestamp to be at the top of the modals, marked with a red line in the screenshots above._

Raader:
I can work on it

rasmusjeppsen:
> I can work on it

@Raader That would be super cool!

Raader:
I'll open a PR when it's done.

PR #4506, opened 2022-08-27
Add timestamps to tooltips

Raader:
## Issue
#4488 
## Description
Adding timestamps to all tooltips
### Preview
Screen Shot 2022-08-27 at 18 25 53
Screen Shot 2022-08-27 at 18 25 57
Screen Shot 2022-08-27 at 18 26 05
Screen Shot 2022-08-27 at 18 26 09
Screen Shot 2022-08-27 at 21 06 48
Screen Shot 2022-08-27 at 22 53 42


web/src/components/tooltips/countrypanelexchangetooltip.jsx
```diff
@@ -62,6 +69,7 @@ const CountryPanelExchangeTooltip = ({
 
   return (
     <Tooltip id="countrypanel-exchange-tooltip" position={position} onClose={onClose}>
+      <StyledTimeDisplay date={zoneData.stateDatetime ? new Date(zoneData.stateDatetime) : null} />
```
VIKTORVAV99, line 72:
```suggestion
      <StyledTimeDisplay date={zoneData.stateDatetime} />
```

web/src/components/tooltips/countrypanelproductiontooltip.jsx
```diff
@@ -69,6 +76,7 @@ const CountryPanelProductionTooltip = ({ displayByEmissions, mode, position, zon
 
   return (
     <Tooltip id="countrypanel-production-tooltip" position={position} onClose={onClose}>
+      <StyledTimeDisplay date={zoneData.stateDatetime ? new Date(zoneData.stateDatetime) : null} />
```
VIKTORVAV99, line 79:
```suggestion
      <StyledTimeDisplay date={zoneData.stateDatetime} />
```

web/src/components/tooltips/pricetooltip.jsx
```diff
@@ -14,6 +21,7 @@ const PriceTooltip = ({ position, zoneData, onClose }) => {
 
   return (
     <Tooltip id="price-tooltip" position={position} onClose={onClose}>
+      <StyledTimeDisplay date={zoneData.stateDatetime ? new Date(zoneData.stateDatetime) : null} />
```
VIKTORVAV99, line 24:
```suggestion
      <StyledTimeDisplay date={zoneData.stateDatetime} />
```

VIKTORVAV99 (changes requested):
Overall this looks good but I added a few suggestions that should be addressed. Since we know the zone data will always exist then these views are shown and that it should always have a datetime we can remove the redundant checks for it.

I also think it would be great if you could update the carbon intensity graph in this PR as well so it also has the date time on top like the other graphs.

Currently it is below the main title as seen here:
image

Raader:
@VIKTORVAV99 updated carbon intensity graph.

web/src/components/tooltips/mapcountrytooltip.jsx
```diff
@@ -92,9 +92,9 @@ const MapCountryTooltip = ({ electricityMixMode, position, zoneData, onClose })
 
   return (
     <Tooltip id="country-tooltip" position={position} onClose={onClose}>
+      <StyledTimeDisplay date={zoneData.stateDatetime ? new Date(zoneData.stateDatetime) : null} />
```
VIKTORVAV99, line 95:
```suggestion
      <StyledTimeDisplay date={zoneData.stateDatetime} />
```

VIKTORVAV99 (changes requested):
Just one more suggestion then I looks ready to me.

VIKTORVAV99 (approved):
LGTM!

@rasmusjeppsen what do you think, any design feedback that needs to be addressed?

@Kongkille am I right in assuming every datapoint will always have a datetime object and the checks aren't needed?

rasmusjeppsen:
> @rasmusjeppsen what do you think, any design feedback that needs to be addressed?

Thank you for providing a PR for this, @Raader! Looks great. And thank you for your input/review @VIKTORVAV99. Much appreciated!

In terms of the design, I think increasing the visibility of the timestamp by making the text bold would be a nice detail to add.
Here are some design mockups based on the screenshots @Raader provided in the PR to illustrate:

timestamp_mockup
timestamp_mockup2

> I also think it would be great if you could update the carbon intensity graph in this PR as well so it also has the date time on top like the other graphs.

Nice catch, Viktor! I agree - this would be great for consistency.

One thing to note here is that we show the country first and then the timestamp second when hovering on the map zones (outside the sidepanel graphs), which then arguably leads to some inconsistency of the sidepanel vs. the map.

zone_hover_example

I do, however, think that in terms of the hierarchy of information displayed, the country is the essential information when navigating/browsing the map, whereas the timestamp is the essential information for the graphs on the sidepanel.
So, **TLDR;** I agree with updating the carbon intensity graph, but think we should keep the layout of having the country first and timestamp second when hovering over the zones on the map. We could address this in a future PR to improve overall consistency if need be.

Great work!

Raader:
@rasmusjeppsen even though I think making timestamps bold will make it look better, wouldn't it introduce even greater inconsistency between the map tooltips and the sidebar tooltips. Updating the map tooltips appearance is might be out of scope for this PR.

rasmusjeppsen:
> @rasmusjeppsen even though I think making timestamps bold will make it look better, wouldn't it introduce even greater inconsistency between the map tooltips and the sidebar tooltips. Updating the map tooltips appearance might be out of scope for this PR.

@Raader I understand the concern regarding inconsistency, however, I think making the timestamps bold in the tooltips for the sidepanel is the right step forward here whilst keeping the following position of the timestamp for the sidepanel as per your preview:
image

And this position for the map as it currently is:
image

I also think that making the timestamps bold in the side panel will further differentiate the layout of tooltips in the sidepanel vs. when hovering on zones on the map which IMO makes the two layouts prioritize the display of information differently (timestamp vs. country first), rather than making it inconsistent in a bad/unthoughtful way.

I suggest we add the bold text to the timestamps in the sidepanel and then everything LGTM!

Raader:
@rasmusjeppsen 
> I also think that making the timestamps bold in the side panel will further differentiate the layout of tooltips in the sidepanel vs. when hovering on zones on the map which IMO makes the two layouts prioritize the display of information differently (timestamp vs. country first), rather than making it inconsistent in a bad/unthoughtful way.
> 
Makes sense. Made them bold.

VIKTORVAV99:
> > @Kongkille am I right in assuming every datapoint will always have a datetime object and the checks aren't needed?
> 
> @VIKTORVAV99 @Kongkille Map country tooltip(the one which had timestamps already) originally had a date time check. Maybe we should've kept those checks just in case.

I think it should be safe as the backend fill the dates for empty values and the graphs should not be visible at all without this data but we'll see what Kongkille says. And even if there is checks needed we could simplify them with optional chaining anyway.

Kongkille:
> @Kongkille am I right in assuming every datapoint will always have a datetime object and the checks aren't needed?

Yes that should always be the case

Kongkille (commented):
Looks great!

On the code side of things, I'm seeing a bit of duplication in the styling, it's not a big deal but the timedisplay could have a default styling corresponding to the shared styling and then we can override for edge cases.

You could do that by styling the <p> tag in timeDisplay.jsx

On UI, I think it looks great and is intuitive. Have you considered adding some kind of clock icon next to the date? 

image

VIKTORVAV99:
> Looks great!
> 
> On the code side of things, I'm seeing a bit of duplication in the styling, it's not a big deal but the timedisplay could have a default styling corresponding to the shared styling and then we can override for edge cases.
> 
> You could do that by styling the
> 
> tag in timeDisplay.jsx
> 
> On UI, I think it looks great and is intuitive. Have you considered adding some kind of clock icon next to the date?
> 
> image

Adding a icon is a good idea, but I think maybe a calendar icon would be a better fit?
Or switching between a clock icon for "live" view and a calendar icon for historical view.
Calendar month icon:
calendar_month_FILL0_wght400_GRAD0_opsz48

Or perhaps the date range or today icon:
date_range_FILL0_wght400_GRAD0_opsz48
today_FILL0_wght400_GRAD0_opsz48

Or a combination of date range/calendar month, today, and the clock depending on the selected time resolution? (If this is believed to be outside the scope of the PR I can create a subsequent PR with the icon changes.)

Raader:
> Or a combination of date range/calendar month, today, and the clock depending on the selected time resolution? (If this is believed to be outside the scope of the PR I can create a subsequent PR with the icon changes.)

@VIKTORVAV99 I think you should open a subsequent PR.

Kongkille (approved):
I think this is good to go!

Thanks a lot @Raader for driving this through

Raader:
@Kongkille is there something that keeps this from being merged?

VIKTORVAV99:
> @Kongkille is there something that keeps this from being merged?

Kongkille is on vacation but I don't think there is anything blocking it.
@tonypls what do you think, we should be able to merge this right?

tonypls:
> > @Kongkille is there something that keeps this from being merged?
> 
> Kongkille is on vacation but I don't think there is anything blocking it. @tonypls what do you think, we should be able to merge this right?

Yes this is good to go! 🚀 

@Raader Feel free to hit the merge button when the tests have passed 👌

tonypls:
Ah okay that makes sense 😄  I'll merge it for ya. Nice work!

PR #4522, opened 2022-09-01
First earthly setup

FelixDQ:
First version of earthly setup. 

This setup is not very optimized yet as it needs to install all dependencies before you can get the artifact. I will try to create a better structure in a followup PR.


pierresegonne (approved):
Nice! 

I guess yeah we could have an artifact that holds the dependencies? This way it could be fetched from cache / registry without having to reinstall anything right?

FelixDQ:
Yes indeed, that would make sense

PR #4523, opened 2022-09-01
Fix "new version available" cache issue

corradio:
## Issue
This screen was often visible when a new deployment was made

image

It turns out the HTTP call used to check the new version is actually cached in the browser:

image

Cache duration: 24h

image

There are two ways to fix this, as it turns out the client-version.json file can be served by two systems:
1. call the static.electricitymap.org version, which has the right cache headers
2. update server.js to override cache values

I opted for option 2 for now.

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -77,6 +77,11 @@ app.all('/dist/*.map', (req, res, next) => {
   return next();
 });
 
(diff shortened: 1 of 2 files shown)


tonypls (approved):
Excellent!

PR #4526, opened 2022-09-02
Update US-HI-OA generation capacities

brandongalbraith:
Oahu [1] (Hawaii) has decommissioned its last coal fired power plant [2]; battery storage has been increased by size specified by the project [3] intended to replace the coal fired generator.

[1] https://app.electricitymaps.com/zone/US-HI-OA?wind=false&solar=false

[2] https://www.theguardian.com/us-news/2022/aug/31/hawaii-close-coal-power-plant-renewable-energy

[3] https://www.kapoleienergystorage.com/


VIKTORVAV99:
Nice to see that you followed up with this!

The capacity updates look good but could you also add the sources to `DATA_SOURCES.md`?

PR #4528, opened 2022-09-02
Clean up type errors in `ENTSOE.py`

VIKTORVAV99:
This is more or less a follow up to the big type PR I did a while back.
It should take care of all type errors except one in the `merge_production_outputs` function but that has a todo of getting split out and reworked so I thought it might be better to do it then.


parsers/ENTSOE.py
```diff
@@ -866,13 +913,14 @@ def validate_production(
...
+    zone_key: str = datapoint["zoneKey"]
 
     validation_criteria = VALIDATIONS.get(zone_key, {})
 
     if validation_criteria:
         return validate(datapoint, logger=logger, **validation_criteria)
 
+    # NOTE: Why are there sepcial checks for these zones?
     if zone_key.startswith("DK-"):
         return validate(datapoint, logger=logger, required=["coal", "solar", "wind"])
```
VIKTORVAV99, line 926: Slightly off topic but why do the DK and NO subzones have special checks here?

mathilde-daugy (approved):

VIKTORVAV99:
@mathilde-daugy I'll clean up the conflicts caused by #4546 and then merge this if that's okay.

Issue #4510, opened 2022-08-29
Improve Page Speed


tonypls:
## Description
**This is a master issue please link sub issues below**

The current app has poor performance for based on Google's PageSpeed Insights. 

Check the results here:
[**Desktop**](https://pagespeed.web.dev/report?url=https%3A%2F%2Fapp.electricitymaps.com%2F&form_factor=desktop)
[**Mobile**](https://pagespeed.web.dev/report?url=https%3A%2F%2Fapp.electricitymaps.com%2F&form_factor=mobile)

We should improve the web app performance to provide users with a better experience and help improve our SEO.

The Page Speed results give actionable opportunities and help us identify the lowest hanging fruit. We should investigate each of these opportunities to get verifiable pieces of work to complete and improve the performance of the app.

tonypls:
Here's a good issue to start with feel free to add any more issues, not everything will be identified from the PageSpeed Insights so all performance improvements are welcome.

https://github.com/electricitymap/electricitymap-contrib/issues/4115

VIKTORVAV99:
This also relates to the [static web project](https://github.com/orgs/electricitymap/projects/29) that should bring some nice overall improvements. So please have a look in there as well if you need more information.

VIKTORVAV99:
This should in theory be "fixed" in the refractor.

VIKTORVAV99:
While there are still a few things we can do in to improve the page speed even further (or perceived page speed) I'll consider this one closed for now as the rewrite is a substantial improvement that will get even faster with #5027.

PR #4529, opened 2022-09-02
remove moment.js

VIKTORVAV99:
I tried removing this once before and well let's just say it didn't go according to plan...

Hopefully it goes better this time.

Removing moment.js should help reduce the initial bundle size pretty significantly which will increase initial loading performance and help with #4510.


VIKTORVAV99:
Looks like cypress is timing out at 10 min (it usually takes around 3 min) so I'll bump it a little and see if the tests pass.

VIKTORVAV99:
@tonypls I don't know why the CI jobs are timing out but the code should be fine.

Do you have any idea what it could be?

If not I think it's bet just to wait a little and try running the CI suite again.

VIKTORVAV99:
It looks like it stalls here:
image

tonypls:
Seems super slow, the server instances seem fine. I'll push to a different branch soon and see if it's suffering the same.

tonypls:
Running cypress locally is also very hit and miss. We should look into this.

VIKTORVAV99:
> Seems super slow, the server instances seem fine. I'll push to a different branch soon and see if it's suffering the same.

> Running cypress locally is also very hit and miss. We should look into this.

Yeah and removing moment.js itself should have no effect on cypress so I really have no idea why it stalls. We should try and push it to a local branch in this repo instead of my fork and see how that behaves but it's really strange.

I had some timeout issue for the python tests in #4528 as well so it may be indicative of a bigger issue or infrastructure change on GitHub's side.

VIKTORVAV99:
@tonypls I commented out the lines that caused the stall both in the CI and locally I hope this is acceptable for now at least. I also set the language explicitly on each visit cause it was trying to compare the Swedish language strings with the expected English strings locally. 😅

This should be ready to be reviewed but could you just try and open up the wind/solar layers locally with live data when reviewing this, that is caused the issues in the past and I can't test it as there is no wind and solar mockdata.

tonypls:
I'm still confused why this was failing, I don't think commenting out the tests really solves much 😆 

Could we try on a new branch, push a single line comment to trigger CI to run cypress then add just the moment changes and see if the tests still pass?

VIKTORVAV99:
> I'm still confused why this was failing, I don't think commenting out the tests really solves much 😆
> 
> Could we try on a new branch, push a single line comment to trigger CI to run cypress then add just the moment changes and see if the tests still pass?

I think it was failing because the DOM elements where never mounted properly (the test code already mentions this) so when trying to click on the element (that did not exist in DOM) the tests simply stalled.

I'm not totally sure if that will work due to the caching in the `cypress-io/github-action` but we can certainly try.

VIKTORVAV99:
> > > I'm still confused why this was failing, I don't think commenting out the tests really solves much 😆
> > > Could we try on a new branch, push a single line comment to trigger CI to run cypress then add just the moment changes and see if the tests still pass?
> > 
> > 
> > I think it was failing because the DOM elements where never mounted properly (the test code already mentions this) so when trying to click on the element (that did not exist in DOM) the tests simply stalled.
> > I'm not totally sure if that will work due to the caching in the `cypress-io/github-action` but we can certainly try.
> 
> Ah okay, so the cypress tests are cached and don't run on each CI push?

Internally the cypress action has at least 2 caches, but I don't know how and what they are set up to cache. I know they cache the `node_modules` (which is probably why it fails in this PR since that got invalidated) but I don't know what else they cache.

VIKTORVAV99:
I can try a PR that just updates cypress to 9.7, that should in theory invalidate all the caches (IMO).

tonypls:
Yeah, that would be a good test 👍 

If that fails then I think we'll need to do an isolated dive into how to make the cypress suit more reliable

tonypls:
Let's soldier on, could you bring the timeout back down though?

.github/workflows/cypress.yml
```diff
@@ -4,12 +4,12 @@ jobs:
   cypress-test:
     name: Cypress tests
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
```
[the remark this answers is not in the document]
VIKTORVAV99: Oh that one... 😆

Yeah sure!

web/src/helpers/gfs.js
```diff
@@ -1,40 +1,40 @@
...
 import { protectedJsonRequest } from './api';
 
 const GFS_STEP_ORIGIN = 6; // hours
 const GFS_STEP_HORIZON = 1; // hours
 
 export function getGfsTargetTimeBefore(datetime) {
-  let horizon = moment(datetime).utc().startOf('hour');
-  while (horizon.hour() % GFS_STEP_HORIZON !== 0) {
-    horizon = horizon.subtract(1, 'hour');
+  let horizon = startOfHour(datetime);
```
tonypls, line 9: I think we need to add
``` let horizon = startOfHour(datetime ? datetime : new Date());```
here

VIKTORVAV99: Wouldn't `let horizon = startOfHour(datetime ?? new Date())` make more sense if that's the case?
Assuming datetime is either undefined or null and not explicitly false.

tonypls: yeah sounds good. I think this could have been something to do with the tests not being happy for the layers.

VIKTORVAV99: I'll add the nullish coalescing and then add back the tests and see if they work now.

PR #4534, opened 2022-09-05
Earthly: Expose src file artifact without installing deps

FelixDQ:
I changed the build steps a bit so that we now have an artifact called `src-files` that exposes the src files, but doesn't install deps until a later step

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,8 +1,8 @@
 VERSION 0.6
+FROM python:3.8
+WORKDIR /contrib
 
-prepare:
-  FROM python:3.8
-  WORKDIR /contrib
+src-files:
   COPY electricitymap ./electricitymap
   COPY parsers ./parsers
   COPY validators ./validators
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -1,27 +1,30 @@
...
+
+prepare:
+  FROM +src-files
   RUN pip install poetry==1.1.12
   RUN apt-get update && apt-get install -y python3-opencv tesseract-ocr tesseract-ocr-jpn tesseract-ocr-eng libgl1
   RUN poetry config virtualenvs.create false
   RUN poetry install -E parsers -E validators
 
-test:
+build:
```
pierresegonne, line 22: does the `SAVE ARTIFACT .` from the src-files get transfered all the way to the build step?

pierresegonne (approved):

Earthfile
FelixDQ: Nope, the artifact is only saved here. For now, we don't really have a need to save artifacts from the other steps. We could add a `SAVE ARTIFACT .` in the build step if need to depend on that in the future. (e.g. if we want to expose built wheels or something like that)

PR #4540, opened 2022-09-06
Fix ENTSOE productionPerUnit

VIKTORVAV99:
It seems like the dict comprehension was broken and honestly it seems like it complicates things way too much for such a simple problem so I just replaced it with a manual mapping.

This should fix the productionPerUnit method in the ENTSOE.py parser.


VIKTORVAV99:
Looks like the CI is timing out for some unknown reason.
I'll bump the timeout and see if it passes the python tests.

@tonypls pinging you since this relates to the ongoing CI issues.

tonypls:
> Looks like the CI is timing out for some unknown reason. I'll bump the timeout and see if it passes the python tests.
> 
> @tonypls pinging you since this relates to the ongoing CI issues.

It's weird because some branches succeed and some fail. @FelixDQ any thoughts on these CI issues?

VIKTORVAV99:
> > Looks like the CI is timing out for some unknown reason. I'll bump the timeout and see if it passes the python tests.
> > @tonypls pinging you since this relates to the ongoing CI issues.
> 
> It's weird because some branches succeed and some fail. @FelixDQ any thoughts on these CI issues?

At least bumping the timeout for the python formatting worked in here. Cypress however is a stranger case as it works in here but not in the other branch (#4529).

parsers/ENTSOE.py
```diff
@@ -69,8 +69,28 @@
     },
     "storage": {"hydro storage": ["B10"]},
 }
+# ENTSOE production type codes mapped to their Electricity Maps production type.
 ENTSOE_PARAMETER_BY_GROUP = {
```
pierresegonne, line 73: Doesn't that make `ENTSOE_PARAMETER_GROUPS` irrelevant then? If yes then we should also remove it

pierresegonne (approved):

VIKTORVAV99:
@pierresegonne I opted for loops as I found them more readable but if you guys want to use dictionary comprehension instead this code should work:
```python
{ENTSOE_key: type for key in ["production", "storage"] for type, groups in ENTSOE_PARAMETER_GROUPS[key].items() for ENTSOE_key in groups }
```

PR #4541, opened 2022-09-07
#4222: if there are multiple decimals in price, ignore all but last

cameronjohnston:
## Description

Proposed fix to avoid conversion error in TR parser


VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
Looks like this PR is affected by the ongoing CI issues we are having so can't merge it just yet but everything looks good as far as I can tell.

PR #4546, opened 2022-09-08
Improve API key selection for `ENTSOE.py`

VIKTORVAV99:
@pierresegonne the first commit contains the changes I mentioned on slack and the subsequent commits are cleanup and restructuring to handle the fact that we don't return data unless it's valid.

The exceptions (messages) might be a little over the top but I basically copied the existing `QueryError` for that.


VIKTORVAV99:
@pierresegonne the CI issues seems to affect this PR as well but it should be safe to merge. The old CI checks pass but I'll leave it up to you.

pierresegonne:
Alright, annoying. We're away from the office for a few days, so I'll have a look next week.

VIKTORVAV99:
> Alright, annoying. We're away from the office for a few days, so I'll have a look next week.

I have half a fix ready at #4564 but it will only fix the new Github workflows. The old circleCI workflows will still fail however (only when a dependency is changes and the cache is invalidated though).

I think the easiest would be to update to poetry 1.2 for everything and migrate the config files (1.1 files work with 1.2 but 1.2 syntax is not supported for 1.1).

PR #4547, opened 2022-09-08
[CI TEST]: Bump Cypress to version 9.7

VIKTORVAV99:


VIKTORVAV99:
@tonypls this apparently works so I'm not sure what is causing it to fail in the other PR.

But should we merge this and then try and do the code changes in one pr and remove moment.js in another separate pr? That should help us find the cause at least.

PR #4548, opened 2022-09-08
[CI TEST]: Remove moment.js from code

VIKTORVAV99:
This removes moment.js from the code only, a subsequent pr will remove it from the package.json if the CI passes on this one.


VIKTORVAV99:
@tonypls this also stalls, I thought it was due to a function I forgot to replace but sadly not.

I'm not sure what is the best course of action but getting moment.js removed will shave off a lot from the bundle, a little more than 250kb according to lighthouse.

PR #4564, opened 2022-09-12
[CI TEST]: Change cache key strategy

VIKTORVAV99:
This PR changes the CI (github workflows) cache key strategy for the python cache in a effort to fix the CI issues we have been having.

### DIFF
--- a/web/cypress/integration/map.spec.js
+++ b/web/cypress/integration/map.spec.js
@@ -29,15 +29,14 @@ describe('Map', () => {
 
     // TODO: Find a way to test these reliably.
     // test layers
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
-    // cy.get('[href*="wind=true"]').click();
-    // cy.contains('Wind power potential');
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
@tonypls I managed to fix the github workflows but I can't seem to figure out how to do it for CirlceCI.

IF you got any tips I'd love to hear them otherwise I think we should start thinking about removing the more "basic" CircleCI tests?

tonypls:
Looks reasonable to me, so we are telling CI to always use the same poetry version in any virtual envs? Did you try without the get poetry version function and just using the --sync flag?

VIKTORVAV99:
> Looks reasonable to me, so we are telling CI to always use the same poetry version in any virtual envs? Did you try without the get poetry version function and just using the --sync flag?

Actually this is just saying that we should invalidate the cache if the poetry version changes.

I did try with just the sync flag but it didn't work as it still used the old cache.

I have however not tried to pin the poetry version with the new cache key strategy but I can do that.

tonypls (approved):

PR #4573, opened 2022-09-15
NZ parser: add placeholder for solar

drc38:
## Description

Solar generation is rapidly being deployed in NZ, at some point Transpower will add it to their production/capacity reports as a separate category.


parsers/NZ.py
```diff
@@ -109,6 +109,7 @@ def fetch_production(
             ],
             "wind": productions.get("Wind", {"generation": 0.0})["generation"],
             "hydro": productions.get("Hydro", {"generation": 0.0})["generation"],
+            "solar": productions.get("Solar", {"generation": 0.0})["generation"],
```
jarek, line 112: It would probably be better to default to `None` rather than `0.0` while we don't know. 0.0 indicates that it's known to be 0. None indicates that we don't know.

drc38: I've defaulted all to None, except Nuclear which for the foreseeable future will always be 0

VIKTORVAV99:
@jarek I think this is good to go, any other input to take into consideration from your side?

jarek:
Do we know if data source returns 0 for an energy source if they know it to be 0? Or do they omit keys they know to be 0?

drc38:
> Do we know if data source returns 0 for an energy source if they know it to be 0? Or do they omit keys they know to be 0?

It will return a 0 if known to be 0 as opposed to omitting.

jarek (approved):
Looks good to me then!

Issue #4567, opened 2022-09-14
Border flows NO-DK gone missing


Gazer75:
As title say it has disappeared from the map. Not even the bars in each area is showing as empty with a question mark.

VIKTORVAV99:
Looks like this is a issue with the `DK.py` parser at first glance. We might be able to simply switch to `ENTSOE.py` for the exchange as they have the data as well.

I'll look into it later today.

VIKTORVAV99:
Looks like the API has moved:
```
"Not found" fetching exchange data for DK-DK1->NO-NO2: ['API has moved. See https://www.energidataservice.dk/news'
```

I'll try and fix it later.

VIKTORVAV99:
I am currently re-writing the parser but I won't be able to get it done until tomorrow at the earliest.

PR #4575, opened 2022-09-16
New dk parser

VIKTORVAV99:
## Issue
The old DK parser used a SQL query API that is no longer supported.
## Description
This PR switches to a new API for the exchanges and removed the production parser capabilities as they used non validated data prone to errors.

Fixes: #4567


VIKTORVAV99:
Ah shit I did not use a clean branch, I'll recreate this pr.

PR #4580, opened 2022-09-19
remove pro badges

tonypls:
## Description

remove pro badges :)


VIKTORVAV99 (approved):
Things looks good me!
We might want to consider shifting the text here a little to the right but that's just a small style change.
image

PR #4586, opened 2022-09-20
use native crypto functions instead of js-sha256

### LINKED ISSUE #4535: Remove js-sha256 in favor of native crypto functions
Snowpack (via esinstall) is throwing warnings about `js-sha256` relating to security risks and minification issues. So we should remove it in favor of [native crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto) functions.

Snowpack error message:
> [15:12:28] [esinstall:js-sha256] /home/runner/work/electricitymap-contrib/electricitymap-contrib/web/node_modules/js-sha256/src/sha256.js:83:17 Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
[15:12:28] [esinstall:js-sha256] /home/runner/work/electricitymap-contrib/electricitymap-contrib/web/node_modules/js-sha256/src/sha256.js:84:17 Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification

This should improve application security and performance (and might therefore help with #4510).

Ugzuzg:
## Issue

Closes #4535

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -33,7 +33,6 @@
     "i18next-browser-languagedetector": "^6.1.3",
     "i18next-http-backend": "^1.3.2",
     "js-cookie": "^2.2.0",
-    "js-sha256": "^0.9.0",
     "lint-staged": "^12.4.1",
     "lodash.mapvalues": "^4.6.0",
     "lodash.sortedindex": "^4.1.0",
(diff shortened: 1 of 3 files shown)


tonypls (approved):
Looks great, thanks for the contribution 🌲 

Our test suite was giving us some trouble so I've made some updates and now this should pass.

PR #4587, opened 2022-09-21
Update to improve cypress reliability

tonypls:
Tests were passing locally but not on CI, this change helps the tests pass on CI.


VIKTORVAV99:
Should we add force to the other click events too?

VIKTORVAV99 (approved):

tonypls:
This should enable the SHA256 update to pass the test suite.

tonypls:
> Should we add force to the other click events too?

I think it overwrites cypress trying to act like a real user so maybe we only force when we have to.

PR #4588, opened 2022-09-21
Use short month for ticks to prevent text overflow

VIKTORVAV99:
## Issue
The current formatting overflows when the month names are long.
image

## Description

This PR simply changes the ticks to use the short form of the months so prevent this overflow.

### Preview

image


tonypls (approved):
Nice, have you checked this in Swedish? 🇸🇪

VIKTORVAV99:
> Nice, have you checked this in Swedish? 🇸🇪 

It's even worse in Swedish 😅

PR #4592, opened 2022-09-22
Tvs/weather layers fix

tonypls:
## Issue
The weather layers aren't working.

The date fns times are zoned times where as the old api calls were made with utc time. I

I tried using date-fns-tz to convert to utc but this stopped the app from loading on my environment.


VIKTORVAV99 (approved):
I really hate time zones.

Guess I'll give it a 3rd shot in a few months...

VIKTORVAV99:
Looks like you need to update the yarn lockfile as well.

tonypls:
Yeah timezones are a pain, moment has a small place in my heart <3

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
VIKTORVAV99: I'm glad it's not you me that accidentally run yarn on the top level. 😅

PR #4601, opened 2022-09-25
Ceuta and Melilla cleanup (bounding boxes, documentation, capacities)

VIKTORVAV99:
## Description

This PR adds the following for both Ceuta and Melilla:
- Capacities for installed sources (assumes 0 for sources not included in REE as they won't get reported by the production parser anyway (same source)).
- Adds missing documentation for the production source and adds a capacity source.
- Adds bounding boxes.

### Side note:
We can probably decrease the accuracy of the bounding boxes a little to save a few bytes in the file size. 5 decimals would be roughly 1.1 meter accuracy at the equator and 0.3 meter at 75° longitude and we currently generate up to 14 decimals it seems. (Lattitude is constant at approximately 1.1 meter for 5 decimals.)
We could probably even go to 4 or 3 decimals from a functionality standpoint but 5 decimals would give us more than enough resolution for even the smallest zones.


VIKTORVAV99:
Okay so it turns out I was a bit tired when I wrote this PR as I used capacities instead of capacity for the JSON key 😅.

But by doing that I found a issue with the new Github CI that we need to address as it didn't catch the issue. I'll open a separate issue for it later today.

pierresegonne:
>We can probably decrease the accuracy of the bounding boxes a little to save a few bytes in the file size. 5 decimals would be roughly 1.1 meter accuracy at the equator and 0.3 meter at 75° longitude and we currently generate up to 14 decimals it seems. (Lattitude is constant at approximately 1.1 meter for 5 decimals.)
We could probably even go to 4 or 3 decimals from a functionality standpoint but 5 decimals would give us more than enough resolution for even the smallest zones.

Good idea, I guess you could open up an issue on this one and ask Markus' opinion on that

VIKTORVAV99:
> >We can probably decrease the accuracy of the bounding boxes a little to save a few bytes in the file size. 5 decimals would be roughly 1.1 meter accuracy at the equator and 0.3 meter at 75° longitude and we currently generate up to 14 decimals it seems. (Lattitude is constant at approximately 1.1 meter for 5 decimals.)
> We could probably even go to 4 or 3 decimals from a functionality standpoint but 5 decimals would give us more than enough resolution for even the smallest zones.
> 
> Good idea, I guess you could open up an issue on this one and ask Markus' opinion on that

Will do!

Issue #4623, opened 2022-09-29
Language selected in web page is different from language shown.


Ankitchaudharyy:
**Is your feature request related to a problem? Please describe.**
For this app, we have local storage. It stores ```i18nextLng``` as key, and ```languageKey``` as value. 
When the tab is closed in browser, and reopens it takes language key from local storage and update and translates the language.

If ```Estonian``` is stored language in local storage, then it will show Estonian on new window.

But the issue is, In language select menu, 'English' is by default selected, even if the language is translated according to localstorage.

<img width="1725" alt="Screenshot 2022-09-29 at 7 19 00 PM" src="https://user-images.githubusercontent.com/46367738/193050394-8c4e4bad-0527-466f-92d9-def219a35013.png">

**Describe the solution you'd like**
Solution to this can be while selecting the language by default, it should fetch it from local storage. If not present there, then default could be English.

Ankitchaudharyy:
Can I work on this issue ahead?

VIKTORVAV99:
Sure 👍🏻

PR #4624, opened 2022-09-29
Checks the preferred language from the local storage.

Ankitchaudharyy:
## Issue
Language selected in web page is different from language shown in the web page.
#4623

## Description
Solution to this can be while selecting the language by default, it should fetch it from local storage.


web/src/helpers/language.js
```diff
@@ -0,0 +1,7 @@
+import { getKey } from '../helpers/storage';
+import { LANGUAGE_NAMES } from '../helpers/constants';
+
+export const getPreferredLanguage = () => {
+  const preferredLanguage = getKey('i18nextLng');
```
tonypls, line 5: Is it maybe better to use the built in function [here](https://www.i18next.com/overview/api#language) incase they change their implementation and the local storage value isn't there?

VIKTORVAV99: I'd also be a fan of using the actual language key (from i18n) instead of the typed out language name. For example (en-GB) for British English.
It should be more robust if we or i18n change anything in the future (We might want to show the language in both the selected language and native language).
However this might require some bigger changes.

Ankitchaudharyy:
@VIKTORVAV99 @tonypls 
- Implemented this using ``` i18next``` inbuilt functions to detect preferred language from local storage / browser.
- Make use of key rather language name.
- Convert ```zh-cn```, ```zh-hk```, and ```zh-tw``` to ```zh-CN```, ```zh-HK```,and ```zh-TW```, as in localstorage it stores the latter ones.

VIKTORVAV99:
Please don't request reviews and then remove requests for reviews again, I don't know about Tony but I get a notification every time I get requested.

web/src/components/languageselect.jsx
```diff
@@ -49,16 +50,16 @@ const LanguageSelectContainer = styled.ul`
 
 const LanguageSelect = () => {
   const [languagesVisible, setLanguagesVisible] = useState(false);
-  const [selectedLanguage, setSelectedLanguage] = useState('English');
+  const [selectedLanguageKey, setSelectedLanguageKey] = useState(getPreferredLanguage());
   const { __, i18n } = useTranslation();
```
VIKTORVAV99, line 54:
```suggestion
  const { __, i18n } = useTranslation();
  const [selectedLanguageKey, setSelectedLanguageKey] = useState(i18n.language);
```

web/src/components/languageselect.jsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 
 import { useTranslation } from '../helpers/translation';
 import { LANGUAGE_NAMES } from '../helpers/constants';
+import { getPreferredLanguage } from '../helpers/language';
```
VIKTORVAV99, line 5:
```suggestion
```

VIKTORVAV99 (changes requested):
We don't need the `language.js` helpers here, we can simply use `i18n.language` directly.

web/src/components/languageselect.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import i18next from 'i18next';
```
VIKTORVAV99, line 2:
```suggestion
```

web/src/components/languageselect.jsx
```diff
@@ -49,16 +50,16 @@ const LanguageSelectContainer = styled.ul`
 
 const LanguageSelect = () => {
   const [languagesVisible, setLanguagesVisible] = useState(false);
-  const [selectedLanguage, setSelectedLanguage] = useState('English');
+  const [selectedLanguageKey, setSelectedLanguageKey] = useState(i18next.language);
   const { __, i18n } = useTranslation();
```
VIKTORVAV99, line 54:
```suggestion
  const [selectedLanguageKey, setSelectedLanguageKey] = useState(i18next.language);
  const { __, i18n } = useTranslation();
```
```suggestion
  const { __, i18n } = useTranslation();
  const [selectedLanguageKey, setSelectedLanguageKey] = useState(i18n.language);
```

VIKTORVAV99 (changes requested):
We already get a i18n object from the `useTranslation()` hook so we don't need to import i18next.

web/src/components/languageselect.jsx
VIKTORVAV99, line 54:
```suggestion
  const { __, i18n } = useTranslation();
  const [selectedLanguageKey, setSelectedLanguageKey] = useState(i18n.language);
```

web/src/helpers/constants.js
```diff
@@ -41,7 +41,7 @@ const LANGUAGE_NAMES = {
   da: 'Dansk',
   de: 'Deutsch',
   el: 'Ελληνικά',
-  en: 'English',
```
VIKTORVAV99, line 44: I'm not sure this change is necessary but if we decide to sitck with it we should add the geography tag to all of the languages.

VIKTORVAV99 (commented):
Everything looks good but the changes in the constants raise a few questions that needs to be discussed with the team first so I'll wait and see what @tonypls says before merging this.

web/src/helpers/constants.js
Ankitchaudharyy: i18n changes the local storages with these constants ```en-GB, zh-TW, zh-CN, zh-HK``` when any of these languages are selected. 
If these are not changed, then it would not be mapped to right language constant, and it would not be selected in the language menu.

tonypls: I guess the only issue is we have a failed network request for en-GB then it falls back to en.

image

VIKTORVAV99: If we decide to add locale tags here we need to update the i18n fallback language to match.

With that said I'm not totally sure we should do this, at least not for English until we have a US English version and a British English version as it will also affect all number and date formating.

VIKTORVAV99 (dismissed):

VIKTORVAV99:
@tonypls do you have anything to add here or can we merge this as is?
(I think changing the language keys/files is outside the scope of this PR).

web/src/helpers/constants.js
```diff
@@ -41,7 +41,7 @@ const LANGUAGE_NAMES = {
...
+  'en-GB': 'English',
```
VIKTORVAV99, line 44: This change is the only thing blocking this from my side as we don't have the other necessary changes for this to work.
(Fallback language, us version for formatting and verified British translations)

What do you think @tonypls?

VIKTORVAV99 (changes requested):

web/src/helpers/constants.js
madsnedergaard: I'd quite prefer if we can avoid this, as it's used in a lot of different places and might cause bugs elsewhere if we're not consistent with the locale names :)

VIKTORVAV99: Agreed, the long term goal should be to support all sub locales of English (GB, US, AU, etc) but for now I think we need to stick with just plain `en` as the language key.

VIKTORVAV99:
@Ankitchaudharyy could you change back the English language key to just `en`?
If you do I think we can merge this without any other changes.

Ankitchaudharyy:
Sure.

VIKTORVAV99 (approved):

PR #4627, opened 2022-09-30
Spatial Aggregations V2

tonypls:
## Issue
Add spatial aggregations to the map, this PR removes the feature flag and includes aggregated exchanges on the map

## Description

Only countries with defined "subZoneNames" are aggregated. A list of exchanges to exclude from each the country and zone view is generated based off the exchange config when running update-world.js


web/geo/generate-exchanges-to-exclude.js
```diff
@@ -20,11 +20,23 @@ const generateExchangesToIgnore = (OUT_PATH, fc) => {
...
-    if (zoneOne === zoneTwo && countryKeysToExclude.includes(key.slice(0, 2))) {
+    if (key.length > 6 && (countryKeysToExclude.includes(zoneOne) || countryKeysToExclude.includes(zoneTwo))) {
       return key;
     }
   });
-  const exchanges = { exchangesToExclude: unCombinedExchanges };
+
+  const countryExchangesWithSubzones = Object.keys(exchangeConfig).filter((key) => {
+    const split = key.split('->');
+    const zoneOne = split[0].slice(0, 2);
```
VIKTORVAV99, line 30: I assume we are slicing here to get the country code?
If so just be aware that for Australia we use AUS as the country code, now the split works here but it won't match zone config keys.

web/src/excluded-aggregated-exchanges.json
```diff
@@ -1,11 +1,37 @@
 {
-  "exchangesToExclude": [
+  "exchangesToExcludeCountryView": [
+    "AT->IT-NO",
+    "AX->SE-SE3",
```
VIKTORVAV99, line 4: We need to add an `AX->SE` exchange in the `exchanges.json` file (couldn't add a comment there).

VIKTORVAV99:
I'll add some more comments/questions later, I need to grab something to eat and test it out a little more first. 🙂

web/geo/generate-exchanges-to-exclude.js
tonypls: Nice catch, I'll need to rethink this solution

web/src/excluded-aggregated-exchanges.json
tonypls: Done. Will push with next commit

web/geo/generate-exchanges-to-exclude.js
VIKTORVAV99: You should be able to split it at `-` and just grab the first part of the array.

tonypls: I think I might re do the logic based on the zones config to make it a bit more of a direct mapping, your catch has highlighted to me why it's a bit flawed to do it this way.

VIKTORVAV99:
I'd like to revisit URL pattern as I think I have a compromise that we are already using for language.
`http://localhost:8080/zone/SE-SE4?aggregated=false` is very declarative and a bit excessive now that we use a state toggle similar to the production/consumption toggle.

I propose we make it funktion the same way as the `?lang=en-GB` parameter for example. In the way that it sets the correct state when included in the URL and then it's removed from the URL. I realize it will be harder to share the correct page this way but if we include a share button somewhere like suggested in #4216 it would make sense.

We could also include a URL parameter for consumption/production the same way.

What do you think?

web/src/hooks/layers.js
```diff
@@ -11,15 +11,21 @@ import { useAggregatesEnabled } from './router';
 export function useExchangeArrowsData() {
   const isConsumption = useSelector((state) => state.application.electricityMixMode === 'consumption');
   const isHourly = useSelector((state) => state.application.selectedTimeAggregate === TIME.HOURLY);
-  const detailedExchanges = useSelector((state) => state.data.exchanges);
+  const allExchanges = useSelector((state) => state.data.exchanges);
+
+  const detailedExchanges = Object.keys(allExchanges)
```
VIKTORVAV99, line 16:
```suggestion
  const zoneViewExchanges = Object.keys(allExchanges)
```

web/src/hooks/layers.js
```diff
@@ -11,15 +11,21 @@ import { useAggregatesEnabled } from './router';
...
+    .reduce((cur, key) => {
+      return Object.assign(cur, { [key]: allExchanges[key] });
+    }, {});
 
   const selectedZoneTimeIndex = useSelector((state) => state.application.selectedZoneTimeIndex);
 
   const isAggregatedToggled = useAggregatesEnabled();
-  const aggregateViewExchanges = Object.keys(detailedExchanges)
-    .filter((key) => !exchangesToExclude.exchangesToExclude.includes(key))
+  const aggregateViewExchanges = Object.keys(allExchanges)
```
VIKTORVAV99, line 25:
```suggestion
  const countryViewExchanges = Object.keys(allExchanges)
```

web/src/hooks/layers.js
```diff
@@ -11,15 +11,21 @@ import { useAggregatesEnabled } from './router';
...
+    .filter((key) => !exchangesToExclude.exchangesToExcludeCountryView.includes(key))
     .reduce((cur, key) => {
-      return Object.assign(cur, { [key]: detailedExchanges[key] });
+      return Object.assign(cur, { [key]: allExchanges[key] });
     }, {});
 
   const exchanges = isAggregatedToggled ? aggregateViewExchanges : detailedExchanges;
```
VIKTORVAV99, line 31:
```suggestion
  const exchanges = isAggregatedToggled ? countryViewExchanges : zoneViewExchanges;
```
We should probably rename the toggle and hook as well to match the new naming scheme (`isAggregatedToggled`).

web/src/layout/leftpanel/countrypanel/index.jsx
```diff
@@ -57,20 +58,42 @@ const CountryPanelStyled = styled.div`
...
   const trackEvent = useTrackEvent();
   const history = useHistory();
   const location = useLocation();
   const { zoneId } = useParams();
-  const timeAggregate = useSelector((state) => state.application.selectedTimeAggregate);
-
-  const data = useCurrentZoneData() || {};
+  //const selectedZoneTimeIndex = useSelector((state) => state.application.selectedZoneTimeIndex);
+  const showAggregated = useAggregatesEnabled();
+  const zoneViewLeftPanelData = useCurrentZoneData() || {};
```
VIKTORVAV99, line 83:
```suggestion
  const zoneViewLeftPanelData = useCurrentZoneData() ?? {};
```
Should be easier to narrow down the type this way when we get typescript up and running.

web/src/layout/leftpanel/countrypanel/index.jsx
```diff
@@ -57,20 +58,42 @@ const CountryPanelStyled = styled.div`
...
+  //const exchangesForZone = getExchangesForZone(zoneId, allExchanges);
+
+  // const countryViewLeftPanelData = {
+  //   ...zoneViewLeftPanelData,
+  //   exchange: zoneForTimePeriod?.exchange,
+  //   exchangeCapacities: zoneForTimePeriod?.exchangeCapacities,
+  //   exchangeCo2Intensities: zoneForTimePeriod?.exchangeCo2Intensities,
+  // };
+
+  const leftPanelData = showAggregated ? zoneViewLeftPanelData : zoneViewLeftPanelData;
```
VIKTORVAV99, line 94: I assume this will show countryViewLeftPanelData when `true` when the PR is ready?

VIKTORVAV99 (commented):
Just a few notes, nothing major found so far that needs to be addressed yet.

But I'll review it continuously when changes are made. 👍🏼

web/geo/generate-exchanges-to-exclude.js
corradio: Note: you might want to look at zones.json which has the `subZones` key which enables you to figure out which zones have subzones.

web/src/layout/leftpanel/countrypanel/index.jsx
tonypls: Yeah, everything in this file is just testing to see what data was available. Markus is coming through with a sql query so we can change the back-end api to get everything we need to show exchange info on the country panel.

electricitymap/contrib/config/__init__.py
```diff
@@ -72,6 +72,22 @@ def generate_zone_neighbours(
     ZONES_CONFIG, EXCHANGES_CONFIG
 )
 
+# This object represents all neights regardless of granularity
```
Kongkille, line 75:
```suggestion
# This object represents all neighbours regardless of granularity
```

web/geo/validate.js
```diff
@@ -132,21 +132,26 @@ function zeroOverlaps(fc, { MIN_AREA_INTERSECTION }) {
...
   const overlaps = features
     .filter(
       (ft1, idx1) =>
         features.filter((ft2, idx2) => {
           if (idx1 !== idx2 && intersect(ft1.bbox, ft2.bbox)) {
             const intersection = intersect(ft1.ft, ft2.ft);
             if (intersection && area(intersection) > MIN_AREA_INTERSECTION) {
+              if (countriesToIgnore.includes(ft1.ft.properties.countryKey)) {
+                return false;
+              }
```
Kongkille, line 146: Shouldn't we have this conditional higher up? I.e is there a reason that we compute stuff related to the countriesToIgnore, or could we just filter them off in the beginning?

web/geo/validate.js
```diff
@@ -132,21 +132,26 @@ function zeroOverlaps(fc, { MIN_AREA_INTERSECTION }) {
...
               return true;
             }
           }
         }).length
     )
     .map(({ ft, _ }) => ft.properties.zoneName);
-
-  if (overlaps.length) {
+  if (overlaps.length > 0) {
+    //With the current number of zones combined there should be 143 overlaps
```
Kongkille, line 154: Do we need this comment?

web/public/locales/en.json
```diff
@@ -127,13 +127,12 @@
...
     "cpinfo": "<b>consumption</b> takes imports and exports into account, <b>production</b> ignores them",
-    "aggregated": "aggregated",
-    "detailed": "detailed",
-    "aggregateinfo": "Countries with missing zone data are unable to be aggregated",
+    "aggregateinfo": "Countries are aggregated when the annual zone capacity is greater than 90% of the total capacity",
     "selectLanguage": "Select language",
     "lowCarbDescription": "Includes renewables and nuclear",
     "dataIsDelayed": "Data is delayed"
   },
+  "buttons": { "country": "country", "zone": "zone" },
```
Kongkille, line 135: Perhaps a more descriptive name, such as aggregateButtons or aggregateToggle?

web/src/components/zonemap.jsx
```diff
@@ -81,46 +79,24 @@ const ZoneMap = ({
...
+          id: i, // assign an integer id so the feature can be updated later on
+          properties: {
+            color: undefined,
+            zoneData: zone[selectedTimeAggregate].overviews,
+            zoneId,
+          },
+        };
       }
-      if (!isAggregatedViewFFEnabled) {
+      if (!isAggregateEnabled && !zone.geography.properties.isCombined) {
```
Kongkille, line 99: Can we put a comment on this functionality perhaps? I am not sure I understand

web/src/hooks/redux.js
```diff
@@ -106,8 +127,10 @@ export function useCurrentZoneExchangeKeys() {
         Object.keys(zoneData.exchange).forEach((k) => exchangeKeys.add(k));
       }
     });
-    return Array.from(exchangeKeys).sort();
-  }, [isConsumption, zoneHistory, currentZoneData]);
+    const exchangeKeysToDisplay = getExchangesToDisplay(zoneId, isAggregated, Array.from(exchangeKeys).sort());
```
Kongkille, line 130: We could just compute exchanges for each zone once and store it somewhere right? I.e instead of having it as part of the hook functionality?

tonypls: Yes but this way it updates when the zone data changes and keeps the list of current zone exchange logic in one file.

Exchange data could be missing for different time periods etc.

web/src/components/zonemap.jsx
tonypls: Done :D

PR #4631, opened 2022-10-01
Improve front end tests

Ugzuzg:
## Issue

https://github.com/electricitymaps/electricitymaps-contrib/issues/4550

## Description

- Upgrade Cypress to 10
- Make sure WebGL is enabled in Chromium browsers


VIKTORVAV99 (approved):
Everything looks good as far as I can tell but I'll wait for @tonypls to do a review as well before merging.

Thanks for taking the time to improve our testing!

web/cypress/plugins/index.js
```diff
@@ -19,4 +19,12 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+  on('before:browser:launch', (browser = {}, launchOptions) => {
+    if (browser.family === 'chromium' && browser.name !== 'electron') {
+      return {
+        ...launchOptions,
+        args: [...launchOptions.args.filter((arg) => arg !== '--disable-gpu'), '--ignore-gpu-blacklist'],
```
tonypls, line 26: So we always want to use GPU's for some reason?

tonypls (approved):

web/cypress/plugins/index.js
Ugzuzg: Judging by the error message I was getting in Chromium launched from Cypress, the map is rendered using WebGL, for which the GPU is required.

madsnedergaard: Would be nice to add this explanation as a comment to the code :)

Ugzuzg: Done.

madsnedergaard (approved):
Nice!

PR #4634, opened 2022-10-02
Improve typing

yuvalabou:
## Description

Improve type hints by using built in types for ```dict```, ```list``` and ```tuple```


VIKTORVAV99:
Sadly this is not supported in Python 3.8 whice is what we use currently. Therefore I am going to close this PR for now but feel free to create another one once we have updated to Python 3.10 or later.

I hope you understand.

PR #4635, opened 2022-10-02
Translated to Hindi

### LINKED ISSUE #4604: Add translations to the historical view slider
## Description
As en Electricity Maps user I want the historical view slider to change language with the language selection menu so that I can understand which time period I am viewing.

Currently the historical view slider mostly shows in English, it would be great to follow the pattern we use for languages accross the map using i18next, more information can be found [here](https://github.com/electricitymaps/electricitymaps-contrib/tree/master/web/public/locales)

image

TheClearSky:
## Description

Translated around 5% to Hindi, including the one in the issue.


VIKTORVAV99 (commented):
I don't see any real issues here but it would be great if we could get another set of eyes to take a look at the translations before merging them.

Thanks for the PR! 🎉

tonypls:
Nice work :D

tonypls (approved):

madsnedergaard:
Very nice indeed!
I quickly checked with Google Translate and these two lines looks weird, but maybe it's just Google translating being silly? Can you verify that these two lines are correct, @TheClearSky? :)

Screenshot 2022-10-06 at 21 10 22

tonypls:
Hey @madsnedergaard I took a look further into this and it looks like it's just a quirk with Google Translate.

After translating each term individually this is the result:

"Measuring the effect of electricity on the climate"

"We color regions of the world according to the carbon intensity of their electricity consumption (all greenhouse gases included). The more green the color, the more electricity that does not harm the climate."

I'll go ahead and merge this in 🚀 

Thanks @TheClearSky

madsnedergaard:
Haha thought so, thanks for testing! 👏

PR #4636, opened 2022-10-02
Fix deadlink to Denmark TSO's dataset list

ThomasF34:
## Issue

On the DATA_SOURCES file, in the parsers list, there is a deadlink for Denmark's parser. 

## Description

This PR fixes the link, leading to Denmark TSO's Dataset list where we can find the `Electricity Production and Exchange 5 min Realtime` dataset that you use


tonypls (approved):
LGTM 👍

PR #4637, opened 2022-10-03
ZA parser: Integrated new CSV file format

IV1T3:
This PR integrates the new CSV file format from Eskom into the ZA parser. This fixes issue https://github.com/electricitymaps/electricitymaps-contrib/issues/4622.

Eskom now finally substituted their value seperator "," with ";", which leaves a correct (simply parseable) CSV file format.
The new file format, however, led to a parser failure, which has been fixed in this PR.

Previous discussion on the broken CSV file: https://github.com/electricitymaps/electricitymaps-contrib/issues/2881


VIKTORVAV99:
Did they change the historical CSV files as well or do we need to keep the old code to parse historical values (from before the change)?

IV1T3:
There are no historical CSV files in that sense. There is always just one CSV containing values going back about 1-2 weeks from the current day.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #4641, opened 2022-10-04
Update Spanish translations

jmunoz94:
## Issue

Spanish translations are slightly outdated, this PR attempts to update the Spanish translation file (these translations also address #4604 which is the Issue I initially intended to address).

## Description

Update Spanish translations to bring them to an updated state. The only part that I did not translate was under `country-history`; in Spanish we currently have `hourly` entries while the English version is using `default`.

Besides not having the full context on the reason for that difference, I also avoided the translation because of gender in Spanish. In line 73 I did "Intensidad de carbono en los últimos %1$s" because I thought masculine is more generic, but I may have err-ed.

Masculine (time-related) nouns are:
* Segundos (seconds).
* Minutos (minutes).
* Días (days).
* Meses (months).
* Años (years).

Feminine nouns are:
* Horas (hours).
* Semanas (weeks).

The feminine version of the sentence above would be "Intensidad de carbono en las últimas %1$s". An "inclusive" approach would also be possible, but it could be considered awful and/or may take extra space.

Other than that, I used the Italian translation as a reference for gender and other potentially ambiguous contexts (I do not speak Italian but it is reasonably similar to Spanish).

I used "tú" for the translations based on what I saw that was already there. I changed the gender of "app" because app is an anglicism for "aplicación" and that word is feminine ([here](https://twitter.com/RAEinforma/status/1371364855521873921) is also a reference from Real Academia de la Lengua Española, RAE, which is the entity that regulates Spanish).

### Preview
No preview available.


tonypls (approved):
Nice stuff!

We have a few wording changes coming up for the aggregated and detailed terms and tooltip, is it okay if I tag you to update those when we've finalised the changes?

jmunoz94:
> We have a few wording changes coming up for the aggregated and detailed terms and tooltip, is it okay if I tag you to update those when we've finalised the changes?

Yes! I will gladly do my best to help with that.

PR #4642, opened 2022-10-04
Test running e2e on webkit browser

madsnedergaard:
**WIP** Tests running Cypress on Webkit now that experimental support has been added

https://www.cypress.io/blog/2022/09/13/cypress-10-8-experimental-run-tests-in-webkit/


VIKTORVAV99:
> ### Conclusion
> It's doable, but adds almost 5 min to CI, so we should consider if this can:
> 
> * run faster
> * run a subset of the tests
> * only run on merge to master
> 
> Closing this PR again as it was just to experiment.

Runing on master only is certainly possible but kind of defeats the purpose as we would only spot any potential issue after it has been merged.

Is it the run itself that adds 5 min or the setup? If it's the setup it should be possible to cache that step and bring it down to just a few seconds.

PR #4644, opened 2022-10-04
Use fetch depth 2 for CI checkout

VIKTORVAV99:


VIKTORVAV99:
@tonypls this should take care of the CI issues in #4627 (and others).

tonypls (approved):
Looks good, depth 0 will get all the history but I think this should do.

VIKTORVAV99:
> Looks good, depth 0 will get all the history but I think this should do.

I figured we should try with just the HEAD first so we don't fetch an unnecessary amount. If it doesn't work we can switch to 0.

PR #4645, opened 2022-10-04
Fixes app breaking when using URL with invalid zoneid

madsnedergaard:
## Issue

https://github.com/electricitymaps/electricitymaps-contrib/issues/4172

## Description

This fixes the problem where everything broke if the zoneId does not exist (invalid or removed).

Going to https://app.electricitymaps.com/zone/GERMANY will just send you back to the map.

Fixing the lower/uppercase requires a bit more work as we consume the parameter in different ways throughout the app, so I decided to skip that for now :)


tonypls (approved):
Looks good :D

Issue #2653, opened 2020-09-14
GT<->MX exchange not showing up on map


q--:
Despite having a [configuration](https://github.com/tmrowco/electricitymap-contrib/blob/c383d6b46466cd9a9e1ee86d39d7ffe739b37811/config/exchanges.json#L1261) and [working parser](https://github.com/tmrowco/electricitymap-contrib/blob/master/parsers/ENTE.py), the exchange between Guatemala and Mexico isn't showing up on the map.

Maybe it's because the exchange is configured to connect to the MX zone instead of one of the two zones in Mexico bordering Guatemala, MX-PN and MX-OR? There is a warning in the console that the MX zone doesn't exist:

>Zone MX from configuration is not found. Ignoring..

If that is indeed the cause, since the carbon intensity of the southern Mexican zones isn't known, it might make the most sense to calculate a static carbon intensity for all of Mexico using e.g. IEA data and creating the `MX` zone definition using that.

Thoughts?

corradio:
Hello @q-- 
Sorry for being so slow to respond here. Indeed you suspected right: it's because the parser returns data for the country (MX) and we're currently displaying the region.
I'll see if we can get a bit of time to look at what the status is with MX so we can make a decision on which level makes sense (national or regional).

q--:
Alright, thank you

corradio:
Related to https://github.com/tmrowco/electricitymap-contrib/issues/2954

VIKTORVAV99:
As far as I can tell there is only one interconnection between México and Guatamala so it should be safe  to remap the exchange to `GT->MX-OR` instead of `GT->MX`. This way it should show up on the map and work for both the zone level and country level in #4627.

However before making such a change I'd love some help confirming there is just one interconnection between the two countries (or rather no interconnection with `MX-PN` as well).

q--:
https://sitr.cnd.com.pa/m/pub/int.html lists only one interconnection

VIKTORVAV99:
> https://sitr.cnd.com.pa/m/pub/int.html lists only one interconnection

Thanks for confirming it. Should be safe to re-map it then. If you want you can do it yourself otherwise I'll submit a PR for it sometime tomorrow.

q--:
If I have any time to work on ElectricityMap it won't be until later this week. Thanks

VIKTORVAV99:
> If I have any time to work on ElectricityMap it won't be until later this week. Thanks 👍

[...]

PR #4654, opened 2022-10-10
Remap `GT->MX` to `GT->MX-OR`

VIKTORVAV99:
## Issue
The exchange did not show up on the map due to a missmatch in mappings.

## Description
This PR remaps the `GT->MX` exchange to `GT->MX-OR` which should resolve the issue.
>**Note**
>This is only safe to do becasue there is only one interconnection between Mexico and Guatamala so we will always know which zones the electricity is coming from.

Closes: #2653


VIKTORVAV99:
@pierresegonne since all this data should aready exist in the database under the `GT->MX` "key" do you think it would be possible to just move this data to `GT->MX-OR` so we get all the historical data?

pierresegonne (approved):

VIKTORVAV99:
> Can you just ping me when you merge so that I remember ?

Sure thing! 🎉

tonypls (approved):
👍

VIKTORVAV99:
@pierresegonne I'm merging this now as it no longer has any conflicts with other open PRs.

PR #4655, opened 2022-10-10
Remove JSON configuration files (zones, exchanges, co2 params)

pierresegonne:
## Issue

Follow up on #4435. This removes definitely the now deprecated global .json config.

## Description

* Fix references
* Remove contrib files
* Update scripts that used directly the config


scripts/ENTSOE_capacity_update.py
```diff
@@ -60,17 +64,18 @@ def parse_args():
...
-def parse_from_entsoe_api(zone, token):
+def parse_from_entsoe_api(zone_key: ZoneKey, token: str) -> dict:
     """Parses installed generation capacities from the ENTSOE API,
     see https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html#_reference_documentation"""
-    if zone not in ENTSOE_DOMAIN_MAPPINGS:
+
+    if zone_key not in ENTSOE_DOMAIN_MAPPINGS:
         print(
-            "Zone {} does not exist in the ENTSOE domain mapping".format(zone),
+            "Zone {} does not exist in the ENTSOE domain mapping".format(zone_key),
```
VIKTORVAV99, line 73:
```suggestion
            f"Zone {zone_key} does not exist in the ENTSOE domain mapping",
```

scripts/ENTSOE_capacity_update.py
```diff
@@ -102,13 +107,13 @@ def parse_from_entsoe_api(zone, token):
             result[generation_type] = int(value)
     except Exception as e:
         raise ValueError(
-            "Data for zone {} could not be retrieved from ENTSOE".format(zone), e
+            "Data for zone {} could not be retrieved from ENTSOE".format(zone_key), e
```
VIKTORVAV99, line 110:
```suggestion
            f"Data for zone {zone_key} could not be retrieved from ENTSOE", e
```

scripts/ENTSOE_capacity_update.py
```diff
@@ -147,14 +147,18 @@ def main():
...
             exit(1)
 
-        data = parse_from_entsoe_api(zone, token)
+        data = parse_from_entsoe_api(zone_key, token)
 
     aggregated_data = aggregate_data(data)
 
     print("Aggregated capacities: {}".format(json.dumps(aggregated_data)))
-    print("Updating zone {}".format(zone))
+    print("Updating zone {}".format(zone_key))
```
VIKTORVAV99, line 155:
```suggestion
    print(f"Updating zone {zone_key}")
```

web/generate-zones-config.js
```diff
@@ -44,6 +44,43 @@ const mergeExchanges = () => {
   return exchanges;
 };
 
+const mergeRatioParameters = () => {
+  // merge the fallbackZoneMixes, isLowCarbon, isRenewable params into a single object
```
VIKTORVAV99, line 48: Is this something we might want to do directly in the config files?
Perhaps following a structure like this:
```yaml
overrides:
  2021:
    fallbackZoneMixes:
    isLowCarbon:
    isRenewable:
  2022:
    fallbackZoneMixes:
    isLowCarbon:
    isRenewable:
# ...
```
I suppose this could become a problem if the values don't change  at the same time though but it could be worth taking a look at at least.

web/src/excluded-aggregated-exchanges.json
```diff
@@ -1,86 +1 @@
...
-    "GR->IT",
-    "IT->SI",
-    "IT->MT",
-    "LT->SE",
-    "NO->SE",
-    "NO->RU-1",
-    "PL->SE"
-  ]
-}
+{ "exchangesToExcludeCountryView": [], "exchangesToExcludeZoneView": [] }
```
VIKTORVAV99, line 1: This should not be empty right?

VIKTORVAV99:
Oh and before I forget myself, we probably need to update the wiki in a few places once this is merged.

web/src/excluded-aggregated-exchanges.json
pierresegonne: @tonypls do you know?

.github/labeler.yml
```diff
@@ -36,8 +36,5 @@ parser:
 archived parser:
   - parsers/archive/**/*
 
-science:
-  - config/co2eq_parameters.json
-
```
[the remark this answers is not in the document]
pierresegonne: That would be cool indeed! Can I let you take care of it?

web/generate-zones-config.js
pierresegonne: I'm not sure to understand what you have in mind here :)

web/src/excluded-aggregated-exchanges.json
tonypls: We need these :)

pierresegonne:
> Oh and before I forget myself, we probably need to update the wiki in a few places once this is merged. 

Yes that's on our radar :)

.github/labeler.yml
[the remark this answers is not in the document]
VIKTORVAV99: I'll get a PR ready later today! 👍🏻

web/generate-zones-config.js
VIKTORVAV99: I might have misunderstood what this function does but if it just combines the fallbackZoneMixes, isLowCarbon and isRenewable that might be something we want to do directly in the config files so we can avoid this step. But I haven't really had the time look into it more in depth yet.

(This should not be a blocker for merging this PR tho.)

.github/labeler.yml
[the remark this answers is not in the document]
VIKTORVAV99: PR ready at #4661.

web/generate-zones-config.js
pierresegonne: It does so for all zonesOverrides, which now live in the respective zone configs. 
So unfortunately we can't really do it directly in a config file. 

I want to have a second look into this once everything is running smoothly to optimise the config flow. Ideally, we might not need that object. I had to add it due to backend-dependencies, but I'm pretty sure we could do things in a smoother way.

VIKTORVAV99: Sounds like a plan then! 👍🏻 
I'll mark this as resolved.

web/geo/generate-exchanges-to-exclude.js
```diff
@@ -1,5 +1,7 @@
+const { mergeExchanges } = require('../generate-zones-config');
 const { getJSON, writeJSON, fileExists } = require('./utilities');
-const exchangeConfig = require('../../config/exchanges.json');
+
+const exchangeConfig = mergeExchanges;
```
tonypls, line 4:
```suggestion
const exchangeConfig = mergeExchanges();
```

web/src/excluded-aggregated-exchanges.json
tonypls: That's a nice idea!

electricitymap/contrib/config/constants.py
```diff
@@ -1,5 +1,7 @@
 # Note: web/src/helpers.js contains some of the same information.
 
+EXCHANGE_FILENAME_ZONE_SEPARATOR = "→"
```
tonypls, line 3: Maybe this could just be "-" style dash, exchanges can go both ways so a directional arrow doesn't really fit. Also having a character that doesn't exist on a keyboard is a bit annoying.

tonypls (approved):
Looks good from a front end perspective, I've tested the config generation and a manual test of the app.

Not sure if you'd want to do it as part of this PR but we could move all the generated config files to the new config directory. Move world-aggregated.json and excluded-aggregated-exchanges.json (world.json can be deleted too) so we have all config files in one directory.

electricitymap/contrib/config/constants.py
pierresegonne: It can't be a dash as zone keys already have dashes in them. We could do underscore though and it was discussed before. 
The change is minimal so I can include it. I agree about the change. 

Aren't the config pieces you're mentioning only used in web?

pierresegonne: I'll let you update the web config pieces you were mentioning in a subsequent PR :)

tonypls: The underscore works well 👍

PR #4660, opened 2022-10-13
Catches errors with Intl.NumberFormat and adds fallback

### LINKED ISSUE #4630: Intl.NumberFormat with style 'unit' doesn't work on Apple devices with pre-2021 OS
We use `Intl.NumberFormat(lang, { style: 'unit' })` to format hour, day, month, and year with localized names/numbers: https://github.com/electricitymaps/electricitymaps-contrib/blob/0cad313aeb4bb3da0ecf2e4b754cf277b1b20644/web/src/helpers/formatting.js#L71-L76 This was introduced for the history view feature in https://github.com/electricitymaps/electricitymaps-contrib/pull/4220 four months ago.

Per https://caniuse.com/mdn-javascript_builtins_intl_numberformat_numberformat_options_unit_parameter , this is not supported by browsers used by about 7-8% of users.

In particular it is not supported by Safari on iOS before version 14.5 (making up about 1.5% of global users), nor on Safari on Mac OS before version 14.1. This seems to have been released in April 2021. Firefox gained support mid-2020, in version 78. Other major browsers supported this before 2020.

I would have hoped that the formatter would just leave anything it doesn't recognize unformatted, but it appears to throw errors instead. Here are a few reports:
* https://github.com/electricitymaps/electricitymaps-contrib/issues/4625
* https://github.com/electricitymaps/electricitymaps-contrib/issues/4533 this mentions Firefox 104 which should support it, but also Android 9 (release 2018), and if using the Play Store app, it might be using an old version of Webkit to actually render the page?
* https://github.com/electricitymaps/electricitymaps-contrib/issues/4570 on Safari 12.1.2 on macos Sierra 10.12.6; reported as working on Safari 15.5 on macos 12.4 which checks out with Safari gaining support in version 14.5

Is it feasible to wrap the formatter in a try-catch and fallback to ISO format or other unlocalized format?

madsnedergaard:
## Issue

Closes #4630

## Description

This catches exceptions where browsers don't support Intl.NumberFormat with unit param.

I considered if we could test in a smarter way, but seeing as [formatJS also does a try-catch to check if it should polyfill](https://github.com/formatjs/formatjs/blob/main/packages/intl-numberformat/should-polyfill.ts), I reckon it should be fine to do the same :)

### Preview

_Only tested by throwing inside the try block, haven't actually tested on an older version of Safari_

Screenshot 2022-10-13 at 16 11 01

### DIFF
--- a/web/src/helpers/formatting.js
+++ b/web/src/helpers/formatting.js
@@ -68,12 +68,20 @@ const formatDate = function (date, lang, time) {
   }
 };
 
-const getLocaleNumberFormat = (lang, { unit, unitDisplay, range }) =>
-  new Intl.NumberFormat(lang, {
-    style: 'unit',
-    unit,
-    unitDisplay: unitDisplay || 'long',
-  }).format(range);
+const getLocaleNumberFormat = (lang, { unit, unitDisplay, range }) => {
+  try {
+    return new Intl.NumberFormat(lang, {
+      style: 'unit',
+      unit,
+      unitDisplay: unitDisplay || 'long',
+    }).format(range);
+  } catch (error) {
+    // As Intl.NumberFormat with custom 'unit' is not supported in all browsers, we fallback to
+    // a simple English based implementation
+    const plural = range !== 1 ? 's' : '';
+    return `${range} ${unit}${plural}`;
+  }
+};
 
 const formatTimeRange = (lang, timeAggregate) => {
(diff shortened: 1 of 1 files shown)


Kongkille (approved):

VIKTORVAV99:
@madsnedergaard don't forget to push a new mobile build for this release.

PR #4664, opened 2022-10-14
add missing exchange and logic for swapping between country and zone …

tonypls:
…views for aggregated zone panels

## Issue
When toggling between country and zone view with the left panel open, it doesn't make sense to show information for a zone that is not longer available, this PR adds some logic to route the app to the appropriate pages.

subzones will be routed to the country, going the opposite way the app will be routed to the map. 

Also adds a missing exchange between DK->NO that was lost somewhere along the way.


VIKTORVAV99:
Maybe we should try and implement a "zoneList" feature in the future (not in this PR) that shows the subzone of a country when switching from country to zone view?

If your interested I can try and make a simple mock up of it.

FelixDQ (approved):
🎉 🎉 

We also considered your idea @VIKTORVAV99, but we weren't sure if it made sense to show a subzone. Which subzone would a user expect to see? Maybe the last one that they had selected? It got a bit convoluted, so we just decided to close the menu for now.

VIKTORVAV99:
> 🎉 🎉 
> 
> We also considered your idea @VIKTORVAV99, but we weren't sure if it made sense to show a subzone. Which subzone would a user expect to see? Maybe the last one that they had selected? It got a bit convoluted, so we just decided to close the menu for now. 

I was just planning on replacing the details in the country details panel with a list of subzones.

I can probably throw together a very basic mockups to demonstrate it if you wish.

web/src/layout/leftpanel/countrypanel/index.jsx
```diff
@@ -95,6 +99,24 @@ const CountryPanel = ({ electricityMixMode, isMobile, tableDisplayEmissions, zon
     return <Redirect to={parentPage} />;
   }
 
+  // Redirect to the parent page if the zone is subzone and country view is enabled
```
Kongkille, line 102: Redirect to country page?

web/src/layout/leftpanel/countrypanel/index.jsx
```diff
@@ -95,6 +99,24 @@ const CountryPanel = ({ electricityMixMode, isMobile, tableDisplayEmissions, zon
...
+  if (
+    zones[zoneId].geography.properties.isHighestGranularity === true &&
+    zones[zoneId].geography.properties.isAggregatedView === false &&
+    isAggregateEnabled === true
```
Kongkille, line 106: It may be more subjective but I personally don't use this explicit boolean typing.

I think it can make sense in some cases without typescript to emphasise that we're working with a boolean but since these booleans are prefixed with 'is' I think it's more readable to just say


```
    zones[zoneId].geography.properties.isHighestGranularity &&
    !zones[zoneId].geography.properties.isAggregatedView &&
    isAggregateEnabled
```

If you wanted to you could destructure the booleans out before so you'd just say

```
    isHighestGranularity &&
    !isAggregatedView &&
    isAggregateEnabled
```

tonypls: Because this is an edge case, I wanted to make these cases as strict as possible. For example is isAggregatedView is undefined or null I wouldn't want this to execute

Kongkille: Hmm okay i didn't read it as such. I guess I would usually have a guard clause specifically for those cases if I didn't want to execute.

Kongkille (approved):

web/src/layout/leftpanel/countrypanel/index.jsx
tonypls: Yeah I think yours is more readable, hopefully we make it to the promised typed land :)

web/src/layout/leftpanel/countrypanel/index.jsx
tonypls: Updated

PR #4675, opened 2022-10-18
Add guard to `generate-aggregates.js`

VIKTORVAV99:
## Issue
If there is a subZone included that has no geometry feature the script will fail with a non-descriptive error.
## Description
This PR adds a simple guard that prints a clear error message if there is no geometry feature for a included subZone.


web/geo/generate-aggregates.js
```diff
@@ -35,8 +35,12 @@ const generateAggregates = (geojson, zones) => {
     const [multiZoneCountry] = unCombinedZones.filter((feature) => feature.properties.zoneName === country[0]);
     for (let i = 0; i < country.length; i++) {
       const [zoneToAdd] = unCombinedZones.filter((feature) => feature.properties.zoneName === country[i]);
-
-      combinedCountry.geometry = union(combinedCountry.geometry, zoneToAdd.geometry).geometry;
+      if (zoneToAdd) {
+        combinedCountry.geometry = union(combinedCountry.geometry, zoneToAdd.geometry).geometry;
+      } else {
+        process.stdout.write(`ERROR: Could not find a feature for ${country[i]}`);
```
VIKTORVAV99, line 41: @madsnedergaard we could probably use this in all the "script" files to get around the ESLint errors for `console`. What do you think?

madsnedergaard: Hmm, it seems like previously `console.log` was just `process.stdout.write()` with `/n` appended, but when I look at the source code now it seems like more is happening: https://github.com/nodejs/node/blob/main/lib/internal/console/constructor.js

It is annoying with the errors though, maybe we can allow logging for any file in a future `/scripts` folder instead? :)

VIKTORVAV99: Part of the current issue is they are a bit scatered right now. However if we move them to one place (or switch to a strict naming scheeme like `<filename>.script.js`) it should be pretty easy to add a ESLint overwrite.

If we decide to do that I'll change it to a simple Console statment and create a seperate PR for the other changes.

madsnedergaard: Hmm yeah, but I guess it also makes sense to have this particular script in `/geo` - so maybe the naming convention is the best option? I'd also be fine with switching to process.stdout/stderr throughout though :)

VIKTORVAV99: Yeah I think the best would be to create a naming convention as it would allow us to add rules that only target the local scripts (or exclude them like in this case).

We'll just have to agree on a naming convention that makes sense.

Kongkille: I think console.error and console.warn is still available no?

VIKTORVAV99: Looks like you are right!
```js
 'no-console': ['error', { allow: ['error', 'warn'] }],
 ```
 If we add info (`console.info()`) to that we could probably use it without the disable line comments in production as well (console.info() is lower priority than log, warn or error). Allthough I guess we should aim to remove all log messages from the frontend and report criticals errors to sentry instead, but that's another discussion all together.

Kongkille: We could consider adding info there. Right now I like that it's a bit strict because it forces you to be intentional but it makes less sense in scripts like these

VIKTORVAV99: So a naming convention and ESLint overrides might still be best?
I.e. allow all console statments in "script" files and only allow warn and error in the frontend?

VIKTORVAV99: We can always add an override and naming convention later if we have any other conditional rules so if @Kongkille don't have any objections I'll create a PR for `.info` and update the console statments tomorrow. 🙂

madsnedergaard:
We should be able to get this merged in now that rewrite is done and linting setup is improved :)

VIKTORVAV99:
> We should be able to get this merged in now that rewrite is done and linting setup is improved :)

I'll try and get this updated tomorrow.

VIKTORVAV99:
@madsnedergaard PR is updated to the new code structure in the file and should handle multiple missing files geometries now. I opted for a `process.exit(1)` to make sure we don't miss it accidentally but it could just as easily just skip them by removing the call to `process.exit()`

madsnedergaard (approved):
Awesome! 💪

PR #4676, opened 2022-10-18
Aggregate Brazil

VIKTORVAV99:
## Description
Aggregates Brazil and removes the unused BR-I zone as it has no geometry, no parser and no useful config information.

>**Note**
> Timezone information is missing from this PR, but all subzones have `timezone: Null` so it shouldn't make a difference. I'd be happy to include it if we know which timezone to use though (Brazil has 3 or 4 depending on daylight savings).

>**Warning**
>This should not be merged until #4677 is resolved.


config/exchanges/AR_BR.yaml
```diff
@@ -0,0 +1,4 @@
+lonlat:
```
unitrium, line 1: Actually I don't think those are necessary.
The aggregation script will keep the exchanges that are need and set them to the parent level when the country level view is selected.

unitrium (changes requested):
Let's restart this now that the US aggregation has been fulfilled :)

I don't think the exchanges are necessary here. And it's probably good to rerun update world.

config/exchanges/AR_BR.yaml
VIKTORVAV99: Are you sure? We manually added the others (and I added the ones of the US).

VIKTORVAV99:
> Let's restart this now that the US aggregation has been fulfilled :)
> 
> I don't think the exchanges are necessary here. And it's probably good to rerun update world.

Isn't this still technically blocked by #4677?

config/exchanges/AR_BR.yaml
unitrium: Actually yes you're right, I read too fast through the config files. So no changes need here :)

VIKTORVAV99:
> We have a fix on the backend side almost ready to be merged, we might have the time to get it merged next week :)

I'll try and get the merge conflicts cleaned up this weekend then, should be pretty straight forward. 👍🏼

unitrium:
Yes, don't worry yourself for it :)

.prettierignore
```diff
@@ -1,6 +1,7 @@
 # Needed for lint staged.
 .prettierignore
 yarn.lock
+.eartlyignore
```
VIKTORVAV99, line 4: This was needed for lint staged to pass as prettier could not infer a parser for it and since lint staged explicitly pass files to prettier it failed and I could not commit without it.

web/public/locales/de.json
```diff
@@ -1420,10 +1423,6 @@
       "countryName": "Großbritannien",
       "zoneName": "Shetland Islands"
     },
-    "GB-ZET": {
-      "countryName": "Großbritannien",
-      "zoneName": "Shetland Islands"
-    },
```
VIKTORVAV99, line 1426: Apparently there was a duplicate here, removed it when I fixed the merge conflicts.

VIKTORVAV99:
I noticed there are a few changed that slipped through the merge conflict that shouldn't have so I'll have to take care of those before this is merged.

VIKTORVAV99:
Should be all good to go now.

unitrium (approved):
Looks good, I will let you know once the fix is merged in the backend side :)

unitrium:
Just merged the latest changes :)

VIKTORVAV99:
Merging this now as the upstream issues blocking it has been resolved! 🎉

Issue #4424, opened 2022-08-07
Historical prices


corradio:
**Is your feature request related to a problem? Please describe.**
As a user, I would be very curious to see the long-term evolution of electricity prices

**Describe the solution you'd like**
Show the price graphs in the aggregated views.

VIKTORVAV99:
Would these be just daily/monthly/yearly averages or would it be averages weighed by the current load for that price period?

Using just the average might give people the wrong idea when electricity usually is very cheap at night and expensive during the day when there is strain on the grid.

corradio:
Yes indeed I believe a consumption-based average would be appropriate.

VIKTORVAV99:
Would this be just the consumption or would it include exports as well?

For some regions exports account for a significant amount of the load.

corradio:
As the unit of prices is a currency per unit of electricity *consumed*, I
believe consumption should be used. Note that consumption takes into
account exports (the *mix* isn't affected by exports, but the absolute
value is).

VIKTORVAV99:
Ah I assumed consumption and exports where calculated separately.

But since it's included I also think it's the best way to calculate the averages for the longer time periods.

PR #4686, opened 2022-10-21
Add historical prices

corradio:
## Issue
Fixes https://github.com/electricitymaps/electricitymaps-contrib/issues/4424

image


madsnedergaard (commented):
LGTM, leaving approval for @Kongkille as I've forgotten too much context around this 🙃

VIKTORVAV99:
Nice, will be awesome to visualize the impact the current energy "crisis" have had.

Next up aggregated prices? We should be able to use the same kind of model discussed in #4424 with some tweaks.

corradio:
> Nice, will be awesome to visualize the impact the current energy "crisis" have had.
> 
> Next up aggregated prices? We should be able to use the same kind of model discussed in #4424 with some tweaks.

These **are** the historical prices!

Kongkille (approved):
Excellent (;

VIKTORVAV99:
Just realized this, we don't explain the method anywhere (in the frontend) right?
Perhaps we should add a note about it in the FAQ?

PR #4691, opened 2022-10-22
Change all current log messages to info, warn or error

VIKTORVAV99:
Changes all current log messages to info, warn or error and adds info to allowed console functions.


VIKTORVAV99:
@madsnedergaard is there some particular reason we only run eslint on these files instead of everything?
```
eslint geo src --ext .js,.jsx
```

We could just do this if there is no particular reason:
```
eslint .
```

Because right now eslint miss some of the files both in the CI and running `yarn run lint`.

madsnedergaard:
> @madsnedergaard is there some particular reason we only run eslint on these files instead of everything?
> 
> ```
> eslint geo src --ext .js,.jsx
> ```
> 
> We could just do this if there is no particular reason:
> 
> ```
> eslint .
> ```
> 
> Because right now eslint miss some of the files both in the CI and running `yarn run lint`.

Good question, I can't remember why it was done like that 🤔 Maybe it's faster because it's skipping some folders/files? If it's not, then let's just change it :D

madsnedergaard (approved):
LGTM!

VIKTORVAV99:
> > @madsnedergaard is there some particular reason we only run eslint on these files instead of everything?
> > ```
> > eslint geo src --ext .js,.jsx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We could just do this if there is no particular reason:
> > ```
> > eslint .
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Because right now eslint miss some of the files both in the CI and running `yarn run lint`.
> 
> Good question, I can't remember why it was done like that 🤔 Maybe it's faster because it's skipping some folders/files? If it's not, then let's just change it :D

Since I added caches to eslint a while ago it should just be a matter of milliseconds 😅
I'll change it and then get this merged.

PR #4704, opened 2022-10-24
update build file web

tonypls:
two build systems 🧱 🌍


VIKTORVAV99 (approved):
LGTM!

VIKTORVAV99:
> two build systems 🧱 🌍

Hopefully we won't even need that in the future if we can get some work done on the static web project 😅

PR #4706, opened 2022-10-24
Update Taiwanese Capacities

nessie2013:
2021 data, fixes wind power almost going over capacity


VIKTORVAV99 (approved):
LGTM! 🎉

On a side note: I'm not sure this will fix the wind issue as capacities are returned by the parser as well and I think those have priority.

Issue #4708, opened 2022-10-25
clicking on a country on the map - error


RobvanderLinden:
When clicking on a country on the map:
Error message: undefined is not an object (evaluating 'g.join')
Url: https://app.electricitymaps.com/zone/NL

VIKTORVAV99:
Could you provide some more information like what browser you used? What zones it happens on and if it happens consistently when you click on a zone?

RobvanderLinden:
I'm using Safari 15.6.1 (15613.3.9.1.16, 15613) on MacOs Catalina 10.15.7
It happens consistently when clicking on a region.

Issue #4705, opened 2022-10-24
Zone access issue on MacOS Safari


gregbarton:
**Describe the bug**
Total failure accessing any zone on MacOS Safari.  Chrome remains unaffected.

![image](https://user-images.githubusercontent.com/1505300/197594450-8c17342e-9dfb-4e11-84ba-a4b5e6dcd983.png)

**To Reproduce**
1. Go to 'https://app.electricitymaps.com/zone/GB' or any zone URL.
2. See error

**Expected behavior**
Display of zone UI.

**Screenshots**
![image](https://user-images.githubusercontent.com/1505300/197595345-a0dd7c13-aa64-4ca4-b733-19564f23a68b.png)

**Desktop (if relevant):**
 - OS: MacOS 10.15.7
 - Browser: Safari
 - Version: 15.6.1

VIKTORVAV99:
Does this happen for all zones or just a select few? Does it happen consistently every time you visit a zone?

gregbarton:
All zones I’ve tried. (10 or so.) I haven’t done an exhaustive test.

PR #4712, opened 2022-10-26
[BUGFIX]: Don't join on undefined or null

VIKTORVAV99:
## Issue
If `dataSources` are `undefined` or `null` on a system that don't support `Intl.ListFormat` the app crashes.
## Description
This adds optional chaining to make sure we don't call `.join()` on `undefined` or `null` values wo we avoid a hard crash and simply display a ? as the source instead.

Fixes: #4705
Fixes: #4708


tonypls (approved):
Will be great to get this fixed!
LGTM 🐛 ✅

PR #4713, opened 2022-10-26
Streamline node.js CI jobs

VIKTORVAV99:
This PR does a few things for different reasons, but they are all interconnected so I thought it would be best to do it in one PR.

First it uses the web binary for prettier instead of using NPX as I noticed we had a prettier plugin installed that wouldn't be detected otherwise.

Secondly since we now install all node modules in the prettier job it caches them to reduce the time impact it will have on the job runtime. Should just be a few seconds longer than before.

Thirdly, before we installed the node modules at these 3 jobs independently: eslint, verify and jest. But since we now install and cache them in prettier, we can reuse the cache without worrying about race-conditions between the jobs.

Fourthly, this also adds the dependencies we installed in the CI to the package.json so they are always in sync with what we use locally.

Overall, this should cut down the total CI duration for the node.js jobs by 30 seconds and should save us a total CI time of 90 seconds (parallel savings of 30 sec each).

>**Note**
>There are some unrelated changes as a result of prettier picking up the new plugin, these should be pretty straight forward and easy to spot but still I felt it was worth mentioning.

PS: There could be some additional saving to be had if we used a custom install script (basically the same as for the other jobs) for Cypress but I have left it as is for now. Additionally with these changes it would be easy for us to add the prettier-toml plugin if we wish but that is outside the scope of this PR.


madsnedergaard (approved):
This is really cool! 🥳

PR #4716, opened 2022-10-26
add capacitor

tonypls:
Add capacitor in a separate dir, building of the apk / archive still has to be done in android studio / xcode.

Generated icons and splash screens but they are not very good so we will have to do them through android studio and xcode


madsnedergaard:
Closing this, moving it to https://github.com/electricitymaps/electricitymaps-contrib-rewrite

PR #4724, opened 2022-10-28
Create helper script to sum subZone capacities

VIKTORVAV99:
This helper script sum a countries sub zone capacities and writes them to the country level capacity object.

>**Note**
>Right now this only sum one zone at a time but it can easily be modified to update all zones and could be made in such a way it easily plugs into other scripts.


madsnedergaard:
cc @unitrium :)

unitrium:
I already created a python script to handle this (not merged yet). It uses the python classes we have for capacity which I think might be better, so that if we do a change there it will be reflected. It will be merged with the US aggregation ;)

VIKTORVAV99:
> I already created a python script to handle this (not merged yet). It uses the python classes we have for capacity which I think might be better, so that if we do a change there it will be reflected. It will be merged with the US aggregation ;)

Would it be possible to add this to the backend instead as a computational step, so we don't even need to sum the capacities for the countries with subzones?

unitrium:
Should be doable, I'm also thinking it could be nice to keep a trace of the evolution of production capabilities somewhere. I will check that with the team, if we can add it in our backlog. For the short term, we will stick with the YAML files so that we can ship quickly the US aggregation.

VIKTORVAV99:
I already summed the US capacities with my own script, so you don't have to keep that in mind 😅.
So have a chat with the team and see what they say, if we want to discuss it more in depth just hit me up on slack.

I also think it would be nice with historical capacities, I actually created an RFC for it a while back at #4396 but the historical part of it got lost when we split the files.

VIKTORVAV99:
Closed in favor of #4741.

PR #4725, opened 2022-10-28
Update French translations

nicolasca:
## Issue

https://github.com/electricitymaps/electricitymaps-contrib/issues/4604

## Description

I used the script `translation-helper.js` to generate the new json file 

While I'm a native french speaker, this exercice is quite complicated. Here are some thoughts about the translations:

- I translated "Open Source" by..."Open Source". One of the official translation in French is "Logiciel libre", but it is very common to keep the English wording "open source". I even found it on some French official website so I kept this way

- The word `zone`, as I don't have always the context, I was not sure to translate by `zone` or `region`. I used `zone` but this is not definitive

- Multiple wordings with "in the last" ask for french to choose a gender. I picked masculine just because it is more common, for example if the word `month` (`mois` in French, masculine) is after, but it could be wrong for another word like `year` (`année`, feminine). 

- The most difficult was the foreign agencies or companies. Sometimes I felt it was better to translate (like Electricity Company) because it would give a clue to a french non-english reader. But sometimes I didn't translated because it would have been meaningless (for example for `Naturener Wind Watch`).
However the decisions were based on my intuition, so there might be multiple improvements as I'm not a translator 😄 

Don't hesitate for changing requests !

NB: in the diff we can observe a lot of changes due to code styling.
I checked with `npx prettier --check .` but I didn't see any issue with it. Do I need to so anything about it ?

### Preview

No preview available.

### DIFF
--- a/web/public/locales/fr.json
+++ b/web/public/locales/fr.json
@@ -6,9 +6,15 @@
     "datasources": "source des données",
     "contribute": "Contribuez en <a href=\"%s\" target=\"_blank\">ajoutant votre pays</a>"
   },
-  "mobile-main-menu": { "map": "Electricity Maps", "areas": "Zones", "about": "À propos" },
+  "mobile-main-menu": {
+    "map": "Electricity Maps",
+    "areas": "Zones",
+    "about": "À propos"
+  },
   "onboarding-modal": {
-    "view1": { "subtitle": "Cartographie de l’impact de l’électricité sur le climat" },
+    "view1": {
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
Despite my _very_ limited French this looks good to me (with the help of online translators 😅)

VIKTORVAV99:
@pierresegonne could you just have a quick look at these so I didn't miss anything?

pierresegonne (approved):
LGTM

PR #4726, opened 2022-10-28
Set the selected language at first position

nicolasca:
## Issue

The selected language is not always visible in the select (if the one chosen is not in the first positions)
https://github.com/electricitymaps/electricitymaps-contrib/issues/4667

## Description

In the `LanguageSelect` component we put in first position the selected language, so it should be always visible to the user.

### Preview
Capture d’écran 2022-10-28 à 14 05 40


nicolasca:
I couldn't find, but is there a place I should add a unit/integration test for this behaviour ?

VIKTORVAV99:
> I couldn't find, but is there a place I should add a unit/integration test for this behaviour ?

There is not, at least not yet. 😅

However this PR is similar to #4624 and tries to fix the same issue. We can't merge both so we will have to evaluate which is best both performance wise and which makes most sense UI/UX wise.

nicolasca:
Ah okey sorry I didn't see it was already treated. My mistake !

VIKTORVAV99:
> Ah okey sorry I didn't see it was already treated. My mistake !

Don't worry about it!

There are some things I like about both, so we might be able to combine them!

Ankitchaudharyy:
This can be acheived without changing the languageAsArray, by changing CSS only. 
Have a look. #4879

VIKTORVAV99:
I am going to close this particular PR as the whole frontend has changed with the rewrite and there are conflicts. But if you still want to take a crack at it feel free to open a new PR that works with the current code structure.

PR #4727, opened 2022-10-28
Fix two warnings in the console

nicolasca:
## Issue

None

## Description

As I installed the local environement and created my first PRs, I saw some warnings in the console.
Here is a suggestion to correct two of them. 
I'm not sure you are interested in it, but my day was dedicated for this today 😃 

### Preview

Capture d’écran 2022-10-28 à 15 13 07

Capture d’écran 2022-10-28 à 15 13 23

### DIFF
--- a/web/src/components/timeslider.jsx
+++ b/web/src/components/timeslider.jsx
@@ -73,7 +73,7 @@ const TimeSlider = ({
     [width, datetimes, startTime, endTime]
   );
 
-  const startTimeValue = timeScale.domain()[0].valueOf();
+  const startTimeValue = timeScale.domain()[0].valueOf() || 0;
   const endTimeValue = timeScale.domain()[1].valueOf() || 1; // Ensures timeslider thumb is at the end at all times
 
   const handleChangeAndInput = useMemo(
--- a/web/src/components/timesliderheader.jsx
+++ b/web/src/components/timesliderheader.jsx
@@ -46,7 +46,7 @@ const TimeSliderHeader = () => {
   return (
     <Wrapper>
       <Title>{__('time-controller.title')}</Title>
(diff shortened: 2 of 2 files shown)


web/src/components/timesliderheader.jsx
```diff
@@ -46,7 +46,7 @@ const TimeSliderHeader = () => {
   return (
     <Wrapper>
       <Title>{__('time-controller.title')}</Title>
-      <DateDisplay data-test-id="date-display" loading={isLoading}>
```
nicolasca, line 49: Use of transcient props with styled-components since v5.1

web/src/components/timeslider.jsx
```diff
@@ -73,7 +73,7 @@ const TimeSlider = ({
     [width, datetimes, startTime, endTime]
   );
 
-  const startTimeValue = timeScale.domain()[0].valueOf();
+  const startTimeValue = timeScale.domain()[0].valueOf() || 0;
```
nicolasca, line 76: Here we have a Nan. Is that a problem we set a default value to 0 ?

web/src/components/timesliderheader.jsx
```diff
@@ -46,7 +46,7 @@ const TimeSliderHeader = () => {
...
+      <DateDisplay data-test-id="date-display" $loading={isLoading}>
```
nicolasca, line 49: Use of transcient props with styled-components since v5.1

VIKTORVAV99:
I haven't had time to test the code yet locally so I can't say if it works or not but I am curious on how these errors occurred in the first place.
So if you wouldn't mind could you describe how they happened?

madsnedergaard:
Thanks for the contribution here 💪 
While working on #4731 we took your changes here into consideration and ensured those errors don't happen anymore :)

PR #4732, opened 2022-11-01
Add US-NW-SCL geometry

unitrium:
## Issue
The Seattle City Light zone was missing a geometry, which would cause troubles in the future aggregation pipeline for the US.

## Description

Changes a bit the geometries around Seattle and add the US-NW-SCL geometry for Seattle.

### Preview
image


madsnedergaard (approved):
LGTM! :)

PR #4736, opened 2022-11-01
Migrate web from Brick to Earthly

FelixDQ:

PR #4738, opened 2022-11-02
Migrate web

pierresegonne:
## Issue

Duplicate of #4736


FelixDQ (commented):
Cool. Let's just go with this PR as you have tested it with our API. It would probably be a good idea to clean this up in the future, so that web isn't both a library and a project at the same time.

After some thinking, I think that we should keep the brick file until we have made the switch . This way we can merge this now without breaking anything.

FelixDQ (approved):
I readded the brick file.

PR #4739, opened 2022-11-02
Add regional emission factors

### LINKED ISSUE #4737: [feature in development] Regional Emission Factors 🗺️ 🔨
<div align="center">
	Regional emission factors
</div>

### Description
**tl;dr: We're currently implementing regional emission factors!**

Since the dawn of Electricity Maps (or very close to it at least), we have had one grand issue that kept reappearing. It has taken a while, but now we finally have the infrastructure and team in place to work on the this issue: [#738 Refine GHG emission factors by regions](https://github.com/electricitymaps/electricitymaps-contrib/issues/738) 🌍 

Today we mostly use emission factors ([wiki page documenting usage](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Default-emission-factors)) established by the IPCC in 2014, which is not ideal because they are both outdated and do not convey regional differences for the same production modes. With this new feature in place, we will be able to use regional data from reliable sources to give a more accurate estimate for the CO2 impact of different production modes.

All calculations and sources for these regional emission factors will be open and transparent - we'll very soon implement the first few in this repository, so stay tuned! 🔜 

### Timeline

- Right now, the team is working behind the scenes on making this possible - we're aiming to have a first version out before end of November 2022
- At first we will focus on US and a few EU zones - but we're building an open, scalable system that means everyone can contribute with new data for new zones!
- Once it's launched and working, we will document how to contribute and work on making it as easy as possible to gather data, update regions and build the best possible map together with you 💚 

**We will share updates on this issue whenever there's anything new.**

_Lastly, we decided to keep conversations about this topic in a new issue here instead of continuing in #738, as the conversations there were focused on specific data sources, while we here want to talk about the functionality itself_

pierresegonne:
## Issue

Fixes #4737 #738 

## Description

Adds all regional emission factors to our config.

* The methodology is documented in our [wiki](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Emission-factors)
* The data sources are regrouped in the [emission factors data source](https://github.com/electricitymaps/electricitymaps-contrib/blob/e8fd0e356574c96033ada97e6232213d1b47be70/EMISSION_FACTORS_SOURCES.md)


config/zones/SE.yaml
```diff
@@ -157,6 +161,14 @@ emissionFactors:
...
         source: electricityMap, 2021 average
         value: 42.07081808517848
+    nuclear:
+      datetime: '2020-01-01'
+      source: UNECE, https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf#page=56
+      value: 5.13
+    solar:
+      datetime: '2021-01-01'
+      source: INCER ACV, https://docs.google.com/spreadsheets/d/1w5DJ7sPen6axIHU8TCVcuzNCjlct4I6JAbhUlw-ZXu8/edit?usp=sharing
+      value: 35.76666667
```
VIKTORVAV99, line 171: I know this is a draft but these seems to be missing from the zone level SE zones, is this just an oversight?

At least for nuclear it should be pretty straight forward as SE-SE3 is the only zone which has nuclear power plants right now.

config/zones/AT.yaml
```diff
@@ -42,6 +42,18 @@ emissionFactors:
       - datetime: '2021-01-01'
         source: electricityMap, 2021 average
         value: 189.1555066532277
+    biomass:
+      datetime: '2021-01-01'
+      source: EU-ETS, ENTSO-E 2021
```
madsnedergaard, line 47: Would it make sense to have a `source_link` attribute here? We could then use it to auto-generate a markdown file similar to DATA_SOURCES (or just added to the same), and it would also make it easier to find information :)

VIKTORVAV99: Or perhaps a top-level object to map the source to the link so we don't have to replicate it for every time the source is used?
Example if DK-DK1 file with sources object:
```yaml
bounding_box:
  - - 10.535889
    - 54.380557
  - - 12.788086
    - 56.36525
capacity:
  battery storage: 0
  biomass: 1015
  coal: 1943
  gas: 1018
  geothermal: 0
  hydro: 7
  hydro storage: 0
  nuclear: 0
  oil: 196
  solar: 1086
  wind: 5165
contributors:
  - corradio
  - tmslaine
sources:
  ENTSO-E 2021:
    link: <link>
    doi: <doi number>
...
```
This way we don't create duplicates for each production mode for each year it's used (if it's reused).

config/zones/SE.yaml
pierresegonne: We've only added the configuration at zone level. That's a good point, I'll add the overrides for the country levels as well.

VIKTORVAV99: This is at the county level, but it's missing from the zone level. 😅
This is probably due to the fact that SE used to be a zone level zone as well.

config/zones/AT.yaml
pierresegonne: Great idea @VIKTORVAV99! Just integrated the change, and I'll follow up with a script to automatically update data sources

VIKTORVAV99:
Some of the values in here differ quite a lot from the IPCC values we used in the past. I realize this is partly because some sources have improved over time and some use more local factors in the calculations. But would it make sense to create let's say a linear average from the 2014 IPCC emission factors to the new ones that seem to be applicable from 2020 and 2021? (To avoid a sudden spike or decline in the emission factors).

This is just a rough idea, so I don't know if this is something we can actually do or if it would cause more issues, but I thought I'd put it out there at least.

PS: If we do this these calculations should happen on the backend and not be stored in the zone files in my opinion and then dynamically update as soon as a source changes or a new one is added.

pierresegonne:
>This is just a rough idea, so I don't know if this is something we can actually do or if it would cause more issues, but I thought I'd put it out there at least. 

Thanks for sharing!

Due to transparency reasons, we want to do the least amount of computations on our side though, and stick to using as much as possible public and trustworthy data sources. That means that unless we find intermediate emission factors for the years 2014 - 2020 we don't have much of a choice here :/

VIKTORVAV99:
> >This is just a rough idea, so I don't know if this is something we can actually do or if it would cause more issues, but I thought I'd put it out there at least. 
> 
> Thanks for sharing!
> 
> Due to transparency reasons, we want to do the least amount of computations on our side though, and stick to using as much as possible public and trustworthy data sources. That means that unless we find intermediate emission factors for the years 2014 - 2020 we don't have much of a choice here :/

Understood, I suppose a bit of weirdness is preferred over baseless estimates. 👍🏻

VIKTORVAV99 (commented):
Solar, Hydro, Wind and Nuclear is missing from SE-SE1, SE-SE2, SE-SE3 and SE-SE4. But beyond that things have looked okay, (I've only taken a close look at the Nordic and some other Europeans countries).

EDIT: As for Sweden there is also the question if we want to use Vattenfalls Nuclear emission factors instead of the EU wide emission factors.

pierresegonne:
> EDIT: As for Sweden there is also the question if we want to use Vattenfalls Nuclear emission factors instead of the EU wide emission factors? Or a combination of the two where the plants operated by Vattenfall get the Vattenfall factor and the others get the EU wide one then we calculate a weighted average based on the annual production for all plants.

The same applies to e.g. nuclear in France. I'm personally not too much a fan of relying on emission factors reported by power producers themselves, and prefer relying on 3rd party sources.

pierresegonne:
> Solar, Hydro, Wind and Nuclear is missing from SE-SE1, SE-SE2, SE-SE3 and SE-SE4. But beyond that things have looked okay, (I've only taken a close look at the Nordic and some other Europeans countries).

I had forgotten to rerun the script with the final data, it should be good now

VIKTORVAV99:
> > Solar, Hydro, Wind and Nuclear is missing from SE-SE1, SE-SE2, SE-SE3 and SE-SE4. But beyond that things have looked okay, (I've only taken a close look at the Nordic and some other Europeans countries).
> 
> I had forgotten to rerun the script with the final data, it should be good now

Everything except Solar look good for Sweden, shouldn't solar differ based on the lon/lat of the zones?

Also looks like I missed that Norway have the same issue with missing production modes, Solar should be easy to add, and I think we could use the EU averages for Norway as well even though they are technically not in the EU?

VIKTORVAV99:
> > Also looks like I missed that Norway have the same issue with missing production modes, Solar should be easy to add, and I think we could use the EU averages for Norway as well even though they are technically not in the EU?
> 
> NO is not in the EU so we did not include them, and won't in this first iteration.

All good. 👍🏻 

> For solar in SE, the difference would be really small between the zones, so we kept them equal at the country level

This does not match how it's done for Denmark though; it has different emission factors for each zone?
I would prefer the per zone factors but being consistent is more important.

VIKTORVAV99:
> > his does not match how it's done for Denmark though; it has different emission factors for each zone?
> > I would prefer the per zone factors but being consistent is more important.
> 
> Good point, forgot about this as we split up SE after the solar factors were computed. Just fixed it.

SE-SE1 and SE-SE2 is using the IPCC numbers but the generation there is so low it probably doesn't matter for now.
Same for Norway as it's not reported at all to ENTSO-E yet. So, I think this is good to go for the Nordic at least, I'll try and look over the rest of Europe later today.

pierresegonne:
> SE-SE1 and SE-SE2 is using the IPCC numbers but the generation there is so low it probably doesn't matter for now.

Yeah it's so far up north that I could not find proper potential PV output ahah

mathilde-daugy (approved):

VIKTORVAV99:
FR-COR is also missing solar, a big portion of the generation comes from solar so would be great to get it added.

Does the INCER ACV tool have an API we could use to automate solar factor calculations based on our bounding box geometries?

PR #4679, opened 2022-10-19
Aggregate USA

VIKTORVAV99:
There are still some things left to do in here but the basics are there.
Left to do:
- [x] Sum all capacities
- [ ] Add translations / language strings
- [x] Add country level exchanges
- [ ] Figure out what to do with `US-SW-DEAA` 
- [x] Figure out what to do with `US-NW-SCL`
- [ ] Figure out what to do with `US-NW-AVRN`

>**Note**
>There is 75 US zones by my count and this just aggregates the 72 vissible ones on the map as adding zones without any gemoetries is currently not possible.

>**Warning**
>This should not be merged until #4677 is resolved.

Issue #3358, opened 2021-09-06
Remove deprecated US zones


madsnedergaard:
### Description

Some of the US zones we have are deprecated by EIA.

List:

- https://github.com/tmrowco/electricitymap-contrib/issues/3345
- https://github.com/tmrowco/electricitymap-contrib/issues/3344
- https://github.com/tmrowco/electricitymap-contrib/issues/3343
- https://github.com/tmrowco/electricitymap-contrib/issues/3342

KabelWlan:
The zone US-MIDW-GLHB retired their last coal power plant at the beginning of September 2022 which means that this zone can be removed as well.

Issue #4173, opened 2022-05-23
Clean up small and generation only US zones


pierresegonne:
### Description

As explained in https://github.com/electricitymap/electricitymap-contrib/pull/3817#issuecomment-1050852974, there are multiple small US zones that should be re-arranged or deleted. 

Some of the US zones were too small to show on the map, so to stay consistent, we should remove them from the config altogether.

Others are generation only zones that feed into larger zones. The parsers for these should be merged with those where they feed in.

pierresegonne:
Hey @KabelWlan FYI the TODO was for me not, not really because your overview was not clear 😅. It is very clear thanks, I've updated the issue description.

unitrium:
-     Include all US-SW-DEAA generation into US-SW-SRP
-     Include all US-SW-HGMA generation into US-SW-SRP
-     Include all US-SW-GRIF generation into US-SW-WALC
-     Include all US-NW-GWA generation into US-NW-NWMT
-     Include all US-NW-WWA generation into US-NW-NWMT
-     Include US-NW-AVRN gas generation into US-NW-PACW and wind generation into US-NW-BPAT

Will be done through #4741 by adding the producing zones directly in the parser to the supplying zones.
For the unclear zones we leave as it is for now.

KabelWlan:
US-NW-GRID which transmits wind generation to US-SW-PNM only might be one to be added as well as long as you're okay with splitting up zones because it's not clear where solar, coal and gas really is sent to.

unitrium:
@KabelWlan just a small follow up on US-NW-NWMT. Do you know where I could find info about the power supply of US-NW-NWMT by US-NW-GWA and US-NW-NWA?

unitrium:
Thanks the clarification!

unitrium:
I'm revisiting this thread a bit because US-SE-SEPA has been a bit annoying lately:
![image](https://github.com/electricitymaps/electricitymaps-contrib/assets/38283096/6f80fdd6-b99b-49cf-9adc-091d01ec64b1)
It's a zone we often can't flowtrace because it has no exchanges and it's production is 0. Would be great if we could figure out where the electricity flows.

unitrium:
Quickly looking at the EIA exchanges it sends electricity to DUK, SC, SOCO, SCEG. So it might be tricky to figure out who is getting the electricity.

unitrium:
[...]

PR #4741, opened 2022-11-03
Enable US-Aggregation view

unitrium:
## Issue

To be merged once backend has been updated to support the change.
close https://github.com/electricitymaps/electricitymaps-contrib/issues/3358
close https://github.com/electricitymaps/electricitymaps-contrib/issues/3345
close https://github.com/electricitymaps/electricitymaps-contrib/issues/3342

## Description

- Add python script to generate aggregation files.
- Clean US zones according to #4173 by integrating the production only zones directly into the supplied zones using the EIA parser.
- Add tests for EIA parser (with mock files).
- Add US.yaml and update world accordingly with cleaned US Zones.

### Preview

image


VIKTORVAV99:
Should I close #4679 in favour of this one and add the translations in a separate PR?

unitrium:
@VIKTORVAV99 yes that would be perfect

parsers/EIA.py
```diff
@@ -549,6 +575,25 @@ def _fetch(
     ]
 
 
+def _order_by_timestamp(datapoints: List[dict]) -> Dict[str, dict]:
+    indexed_data = {}
+    for datapoint in datapoints:
+        indexed_data[datapoint["datetime"]] = datapoint
```
pierresegonne, line 581: I think that the naming can be improved here. It seems to me that the function does more of an indexing operation than an ordering one.

parsers/EIA.py
```diff
@@ -549,6 +575,25 @@ def _fetch(
     ]
 
 
+def _order_by_timestamp(datapoints: List[dict]) -> Dict[str, dict]:
```
pierresegonne, line 578:
```suggestion
def _index_by_timestamp(datapoints: List[dict]) -> Dict[str, dict]:
```

parsers/EIA.py
```diff
@@ -549,6 +575,25 @@ def _fetch(
...
+    indexed_data = {}
+    for datapoint in datapoints:
+        indexed_data[datapoint["datetime"]] = datapoint
+    return indexed_data
+
+
+def _merge_production_mix(mixes: List[List[Dict[str, Any]]]) -> List[Dict[str, Any]]:
+    merged_data = {}
+    for mix in mixes:
+        indexed_mix = _order_by_timestamp(mix)
```
pierresegonne, line 588:
```suggestion
        indexed_mix = _index_by_timestamp(mix)
```

parsers/EIA.py
```diff
@@ -549,6 +575,25 @@ def _fetch(
...
 
 
+def _order_by_timestamp(datapoints: List[dict]) -> Dict[str, dict]:
+    indexed_data = {}
+    for datapoint in datapoints:
+        indexed_data[datapoint["datetime"]] = datapoint
+    return indexed_data
+
+
+def _merge_production_mix(mixes: List[List[Dict[str, Any]]]) -> List[Dict[str, Any]]:
```
pierresegonne, line 585: Otherwise, great the code is really clear

parsers/test/test_EIA.py
```diff
@@ -0,0 +1,189 @@
+import os
```
pierresegonne, line 1: Awesome that you added tests for this. 

On contrib, the guidelines for mocking are quite loose, but did you have a look at how other parsers might be mocked (if we even do that?). 

If yes, then it would be nice to stay consistent in between the different parsers.

scripts/create_aggregated_zone_config.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
+"""
+poetry run python scripts/create_aggregated_zone_config US US-Central
```
pierresegonne, line 3: Can you maybe add to the top comment the exact purpose of that script? For now as the idea is fresh in our minds, it's quite clear, but I don't think we'll be running that script every 2nd week, and we might forget what it's used for.

pierresegonne (commented):
Small things / comments.

parsers/EIA.py
```diff
@@ -254,14 +255,14 @@
     "US-NW-CHPD->US-NW-PSEI": "&facets[fromba][]=CHPD&facets[toba][]=PSEI",
     "US-NW-GCPD->US-NW-PACW": "&facets[fromba][]=GCPD&facets[toba][]=PACW",
     "US-NW-GCPD->US-NW-PSEI": "&facets[fromba][]=GCPD&facets[toba][]=PSEI",
-    "US-NW-GWA->US-NW-NWMT": "&facets[fromba][]=GWA&facets[toba][]=NWMT",
+    #    "US-NW-GWA->US-NW-NWMT": "&facets[fromba][]=GWA&facets[toba][]=NWMT", integrated directly with US-NW-NWMT
```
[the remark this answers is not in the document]
unitrium: I'm thinking, we should keep a trace of this. I've removed the exchanges from the yaml configs so it's the only remaining trace.
The reason is that in one year someone might think, we forgot to include those exchanges and adds them.
We would then start counting the electricity production from the supplying zones twice, once through the direct integration and a second time through the exchanges.

parsers/test/test_EIA.py
unitrium: To this I've taken some inspiration from the ENTSOE tests and extended them, so they are in line with what we have (though we could improve what we have a bit).

KabelWlan:
Regarding the cleaning of small zones, #3358 might as well be an issue related to the changes done (and might be included?).

VIKTORVAV99:
Should the US Virgin islands and other outlying areas also be included in the US?
We normally don't include zones that have their own ISO code but if we take the Virgin Island as a example againISO use both VI and US-VI for its zone keys so this one is up to us.

I'm in favor of keeping it seperate otherwise we would have to include all of these in the US:
image

unitrium:
> Regarding the cleaning of small zones, #3358 might as well be an issue related to the changes done (and might be included?).

I will make sure to include those changes as well.

parsers/test/test_EIA.py
pierresegonne: alright goodonya

parsers/EIA.py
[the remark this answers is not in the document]
pierresegonne: Ok I think it makes sense

unitrium:
Removed additional depreciated zones.
Close #3358 
Close #3345
Close #3342

unitrium:
@VIKTORVAV99 regarding the US's other territories we will keep them separated.

unitrium:
@KabelWlan  do you have any idea with which zone US-SE-AEC should be merged with ? The EIA retired this zone in 2021. So another balancing authority should have taken over their reporting but I can't find which one.

electricitymap/contrib/config/model.py
```diff
@@ -104,7 +104,7 @@ class Zone(StrictBaseModelWithAlias):
     disclaimer: Optional[str]
     flag_file_name: Optional[str]
     parsers: Parsers = Parsers()
-    sub_zone_names: Optional[List[str]] = Field(None, alias="subZoneNames")
+    sub_zone_names: Optional[List[ZoneKey]] = Field(None, alias="bypassedSubZones")
```
pierresegonne, line 107: copy paste gone wrong?

unitrium: No it's intended, since we have a ZoneKey type alias, I might as well use it.

unitrium: Oups!

unitrium: Fixed

config/zones/US.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - US-SW-EPE
+  #- US-SW-GRIF Production only zone, all production integrated with US-SW-WALC
+  #- US-SW-HGMA Production only zone, all production integrated with US-SW-SRP
+  - US-SW-PNM
+  - US-SW-SRP
+  - US-SW-TEPC
+  - US-SW-WALC
+  - US-TEN-TVA
+  - US-TEX-ERCO
+bypassedSubZones:
```
VIKTORVAV99, line 101: I assume this is for zones with no parsers?
Just for my own future reference.

KabelWlan:
> @KabelWlan do you have any idea with which zone US-SE-AEC should be merged with ? The EIA retired this zone in 2021. So another balancing authority should have taken over their reporting but I can't find which one.

I'm not sure which zone took over US-SE-AEC. Howerver there are hardly any zones in that region so I'd let US-SE-SOCO fill up the space on the map.

config/zones/US.yaml
unitrium: Exactly, or zones that have been depreciated so no data comes in from the parser but we don't know yet with which we should merge it. (US-MIDW-GLHB) in this case.

pierresegonne (approved):

scripts/create_aggregated_zone_config.py
```diff
@@ -0,0 +1,87 @@
...
+
+from electricitymap.contrib.config.model import Capacity, Zone
+
+NAME_MAPPING = {"hydro storage": "hydro_storage", "battery storage": "battery_storage"}
+
+
+def create_aggregated_config(zoneKey: str, timezone: str):
+    zone = Zone(
+        key=zoneKey,
+        sub_zone_names=[],
```
VIKTORVAV99, line 21: Perhaps it would be better to stick with the `subZoneNames` naming scheme.

It also look like pylance (python type checking tool) expects it to be `subZoneNames`
image

scripts/create_aggregated_zone_config.py
```diff
@@ -0,0 +1,87 @@
...
+        timezone=timezone,
+        contributors=[],
+        capacity=Capacity(),
+    )
+    zone.parsers = None
+    for path in Path("config/zones").glob(f"{zoneKey}-*.yaml"):
+        if path.is_file():
+            with open(path, "r") as file:
+                subzone = yaml.safe_load(file.read())
+                zone.sub_zone_names.append(path.stem)
```
VIKTORVAV99, line 31:
```suggestion
if instanceof(zone.sub_zone_names, list):
    zone.sub_zone_names.append(path.stem)
```
Improves type saftey as yaml.safe_load can return None or a Dict.

However it looks like it's used as a dict on the next line?

scripts/create_aggregated_zone_config.py
```diff
@@ -0,0 +1,87 @@
...
+                if "capacity" in subzone.keys():
+                    capacities: dict = subzone["capacity"]
+                    for key, capacity in capacities.items():
+                        mapped_key = (
+                            key if not key in NAME_MAPPING else NAME_MAPPING[key]
+                        )
+                        if capacity is not None:
+                            current_capacity = zone.capacity.__getattribute__(
+                                mapped_key
+                            )
```
VIKTORVAV99, line 45: `__getattribute__` expects a secondary argument as well.

scripts/create_aggregated_zone_config.py
```diff
@@ -0,0 +1,87 @@
...
+                            if current_capacity is None:
+                                current_capacity = 0
+                            zone.capacity.__setattr__(
+                                mapped_key, current_capacity + capacity
+                            )
```
VIKTORVAV99, line 50: `__setattr__` also expects a secondary argument.

VIKTORVAV99 (commented):
Almost everything looks good but there seem to be a few type issues in the aggregation script. 

I think I identified most of them.

It also looks like the aggregation script removes some of the existing information from the file like bypassedSubZones, bounding box and country level only contributors.

Overall it think it looks good though but we should probably address this, perhaps in a separate PR.

config/zones/US.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - US-HI-HA
+  - US-HI-KA
+  - US-HI-KH
+  - US-HI-LA
+  - US-HI-MA
+  - US-HI-MO
+  - US-HI-NI
+  - US-HI-OA
+  - US-MIDW-GLHB
+timezone: US-Central
```
VIKTORVAV99, line 110: I think this should be in the following format: `US/Central`

PR #3559, opened 2021-11-20
Rewrite Overseas France parser (#3461)

RobinFrcd:
- Currently, La Réunion is the only territory with live data
  (https://opendata-reunion.edf.fr/)
- Historical data is fetched from https://opendata-corse-outremer.edf.fr
- No price data available after 2019
- Simplified parser. @PETILLON-Sebastien did a great job but it wasn't
  compatible with current data/projet structure anymore

PR #4742, opened 2022-11-03
New parser for French overseas territories

VIKTORVAV99:
This PR rewrites the FR_O.py parser to conform to our current methods as well as add more regions to the parser.

>**Note**
>This parser returns estimated data together with validated data and flags them with `estimated: True` or `estimated: False` accordingly. I can modify the PR to only return Validated numbers if needed.

Closes: #3559


parsers/FR_O_old.py
```diff
@@ -0,0 +1,351 @@
+#!/usr/bin/env python3
```
unitrium, line 1: If FR_O_old.py is in the parser folder it will be run when we schedule the parsers. I'm not sure this is the intended behavior. Better to place it under parsers/archived if it's no longer going to be used.

VIKTORVAV99: Oh right, that was accidental. I think I'll just remove it as it has too many "estimation" features and tries to split unknown into other modes.

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
-            datapoints.append(result)
 
-    return datapoints
+def fetch_data(
+    zone_key: str = "FR-COR",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger=getLogger(__name__),
+) -> Tuple[Any, str]:
+    ses = session or Session()
```
unitrium, line 103: It would be more clear to have Session as a default value for the function.
So:
def fetch_data(zone_key:str, session = Session(),....)

unitrium: I know that other languages will require a compile constant but hey, let's take advantage that Python is interpreted :)

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
-                        multiple = current_source[type_name]
-                        result["production"][type_name] += value * multiple
+def generate_url(zone_key, target_datetime):
+    return f"{DOMAIN_MAPPING[zone_key]}/api/v2/catalog/datasets/{HISTORICAL_DATASETS[zone_key] if target_datetime else LIVE_DATASETS[zone_key]}/exports/json"
 
-            datapoints.append(result)
 
-    return datapoints
+def fetch_data(
+    zone_key: str = "FR-COR",
```
unitrium, line 98: I'm thinking it might be dangerous to have a default zone key. It might be better to have it as a required parameter, so in case someone forgot to supply it, they wouldn't end up with Corsica's data by default but get a clear error.

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
+    if data == []:
+        raise ParserException(
+            "FR_O.py",
+            f"No data available for {zone_key} for {target_datetime.strftime('%Y')}"
+            if target_datetime
+            else f"No live data available for {zone_key}.",
+            zone_key,
+        )
+    return_list: List[Dict[str, Any]] = []
+    for object in data:
```
unitrium, line 158: Maybe improve the naming here?
Data = production_mixes?
object = production_mix?
The idea is that we can understand this line of code straight away without looking at the surrounding context.

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
+        production: Dict[str, Union[float, int]] = {}
+        storage: Dict[str, Union[float, int]] = {}
+        for key in object:
```
unitrium, line 161: Same here, key could be production_type ?

unitrium: Ah but I see that since it's mixed with storage it might be hard to find a better word.

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
+            "solaire_mw",
+        ],
+        "wind": [
+            "eolien",
+            "eolien_mw",
+            "eolien_mwh",
+        ],
+        "unknown": [
+            "bagasse_charbon_mwh",
+            "charbon_bagasse_mw",
```
unitrium, line 69: charbon is coal, so I think we can categorize those as coal for now and adjust with regional emission factors later when it's released

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
+            "thermique_mw",
+            "thermique_mwh",
```
unitrium, line 71: I think thermique and turbines a combustion would count as gas

unitrium (changes requested):
Thanks for the work! There are a few thing that we could improve :)

unitrium:
I'm also thinking it could be nice to add a unit test.
I wrote a small one with mock data I took from the API to try out your code.
It seems to work perfectly though I didn't try edge cases yet.
Let me know if you want I can push the unit test to your branch.
```
import unittest
from json import loads

from pkg_resources import resource_string
from requests import Session
from requests_mock import ANY, Adapter

from parsers import FR_O


class TestFR_O(unittest.TestCase):
    def setUp(self):
        self.session = Session()
        self.adapter = Adapter()
        self.session.mount("https://", self.adapter)

    def test_fetch_exchange(self):
        json_data = resource_string("parsers.test.mocks.FR_O", "FR_GP.json")
        self.adapter.register_uri(ANY, ANY, json=loads(json_data.decode("utf-8")))
        data_list = FR_O.fetch_production("GP", self.session)
        self.assertIsNotNone(data_list)
        expected_data = [
            {
                "zoneKey": "GP",
                "production": {
                    "unknown": 1,
                    "coal": 2,
                    "oil": 3,
                    "hydro": 4,
                    "geothermal": 5,
                    "wind": 6,
                    "solar": 7,
                    "biomass": 8,
                },
                "storage": {},
            },
            {
                "zoneKey": "GP",
                "production": {
                    "unknown": 10,
                    "coal": 11,
                    "oil": 12,
                    "hydro": 13,
                    "geothermal": 14,
                    "wind": 15,
                    "solar": 16,
                    "biomass": 17,
                },
                "storage": {},
            },
        ]
        self.assertEqual(len(data_list), len(expected_data))
        for index, actual in enumerate(data_list):
            self.assertEqual(actual["zoneKey"], "GP")
            self.assertEqual(actual["source"], "edf.fr")
            for production_type, production in actual["production"].items():
                self.assertEqual(
                    production, expected_data[index]["production"][production_type]
                )


if __name__ == "__main__":
    unittest.main()
```

parsers/FR_O.py
```diff
@@ -1,351 +1,188 @@
...
+from typing import Any, Dict, List, Optional, Tuple, Union
 
-def get_source(zone_key: str, target_datetime: datetime):
-    if target_datetime is None:
-        return REAL_TIME_SOURCES[zone_key]
-    else:
-        return HISTORICAL_SOURCE
+from requests import Response, Session
 
+from .lib.exceptions import ParserException
```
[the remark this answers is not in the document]
VIKTORVAV99: the reason I didn't use the symlinked path is well, symlinks don't work on windows... 😅

I have actually removed them in from the other files in the past so we can run on all operating systems.

parsers/FR_O.py
VIKTORVAV99: Agreed but this is the way we do it in the other parsers so I felt consistensy is more important than "correctness".

I'd be happy to change it though.

parsers/FR_O.py
VIKTORVAV99: Hmm, good point, I'll make the change but as previously mentioned it's like this in several parsers by "convention", mostly cause some only parse one zone I suppose.

parsers/FR_O.py
VIKTORVAV99: The object structure looks like this:
```json
[
  {
    "territoire": "Corse",
    "statut": "Estim\u00e9",
    "date_heure": "2021-01-01T00:00:00+00:00",
    "production_totale_mw": 333.2,
    "thermique_mw": 90.5,
    "hydraulique_mw": 86.2,
    "micro_hydraulique_mw": 11.7,
    "photovoltaique_mw": -0.7,
    "eolien_mw": 3.7,
    "bioenergies_mw": 0.8,
    "importations_mw": 141.0,
    "cout_moyen_de_production_eur_mwh": null
  },
  {
    "territoire": "Corse",
    "statut": "Estim\u00e9",
    "date_heure": "2021-01-01T03:00:00+00:00",
    "production_totale_mw": 275.6,
    "thermique_mw": 71.2,
    "hydraulique_mw": 72.8,
    "micro_hydraulique_mw": 11.7,
    "photovoltaique_mw": -1.0,
    "eolien_mw": 2.4,
    "bioenergies_mw": 0.8,
    "importations_mw": 117.7,
    "cout_moyen_de_production_eur_mwh": null
  }
]
So data is the top level array, and object is well the objects in that array. Perhaps adding a comment would enough?

File: parsers/FR_O.py (hunk 5 above)
Comment (VIKTORVAV99) [reply parsers/FR_O.py:161]: There isn't actally any storage in the parser yet but perhaps mode_key would be better?

File: parsers/FR_O.py (hunk 6 above)
Comment (VIKTORVAV99) [reply parsers/FR_O.py:69]: In the past we have put all mixed modes in unknown and then created a adjusted regional carbon intensity for unknown. I'd like to stick with this method if possible as it makes it clear that all biomass is biomass and all coal is coal etc.

This is how we (I) did it for Sweden for example:
```yaml
    unknown:
      - _comment: Uses the 2015 (2022 edition) eurostat production values for the sources
          in unknown
        _url: https://ec.europa.eu/eurostat/documents/38154/4956218/Energy-Balances-January-2022-edition.zip/2dc71f1b-806a-b9ed-e77d-67b7c705197b?t=1643660477829
        datetime: '2015-01-01'
        source: assumes 85.46% biomass, 7.98% gas, 2.77% coal, 1.8% oil, 1.3% peat,
          0.69% solar
        value: 67.7
```

parsers/FR_O.py
VIKTORVAV99: I'll re-assign them 👍🏻

VIKTORVAV99:
> I'm also thinking it could be nice to add a unit test. I wrote a small one with mock data I took from the API to try out your code. It seems to work perfectly though I didn't try edge cases yet. Let me know if you want I can push the unit test to your branch.
> 
> ```
> import unittest
> from json import loads
> 
> from pkg_resources import resource_string
> from requests import Session
> from requests_mock import ANY, Adapter
> 
> from parsers import FR_O
> 
> 
> class TestFR_O(unittest.TestCase):
>     def setUp(self):
>         self.session = Session()
>         self.adapter = Adapter()
>         self.session.mount("https://", self.adapter)
> 
>     def test_fetch_exchange(self):
>         json_data = resource_string("parsers.test.mocks.FR_O", "FR_GP.json")
>         self.adapter.register_uri(ANY, ANY, json=loads(json_data.decode("utf-8")))
>         data_list = FR_O.fetch_production("GP", self.session)
>         self.assertIsNotNone(data_list)
>         expected_data = [
>             {
>                 "zoneKey": "GP",
>                 "production": {
>                     "unknown": 1,
>                     "coal": 2,
>                     "oil": 3,
>                     "hydro": 4,
>                     "geothermal": 5,
>                     "wind": 6,
>                     "solar": 7,
>                     "biomass": 8,
>                 },
>                 "storage": {},
>             },
>             {
>                 "zoneKey": "GP",
>                 "production": {
>                     "unknown": 10,
>                     "coal": 11,
>                     "oil": 12,
>                     "hydro": 13,
>                     "geothermal": 14,
>                     "wind": 15,
>                     "solar": 16,
>                     "biomass": 17,
>                 },
>                 "storage": {},
>             },
>         ]
>         self.assertEqual(len(data_list), len(expected_data))
>         for index, actual in enumerate(data_list):
>             self.assertEqual(actual["zoneKey"], "GP")
>             self.assertEqual(actual["source"], "edf.fr")
>             for production_type, production in actual["production"].items():
>                 self.assertEqual(
>                     production, expected_data[index]["production"][production_type]
>                 )
> 
> 
> if __name__ == "__main__":
>     unittest.main()
> ```

Feel free to push it directly, you already made it so you deserve the credit for it. 👍🏼

parsers/FR_O.py
VIKTORVAV99: In theory it could also return
```json
{
  "message": "ODSQL query is malformed: invalid_function() Clause(s) containing the error(s): select.",
  "error_code": "ODSQLError"
}
```
and
```json
{
  "errorcode": 10002,
  "reset_time": "2021-01-26T00:00:00Z",
  "limit_time_unit": "day",
  "call_limit": 10000,
  "error": "Too many requests on the domain. Please contact the domain administrator."
}
```
I could add specific checks for these.

VIKTORVAV99:
@unitrium explained some of the reasoning behind the choices for the comments you made, but I'll address the most obvious ones right away.

VIKTORVAV99:
I have no idea why CodeQL is marking these as failed since it don't introduce new dependencies but it should be safe to ignore those alerts.

parsers/FR_O.py
unitrium: I think we could update that at a later stage, it would simplify stuff.

unitrium (approved):
Look good to me, we can have the french overseas territory data back.

unitrium:
@VIKTORVAV99, maybe you can try to merge the latest changes from master. It will rerun the CI, if those errors show up again from code scanning I can by pass them.

VIKTORVAV99:
> @VIKTORVAV99, maybe you can try to merge the latest changes from master. It will rerun the CI, if those errors show up again from code scanning I can by pass them.

So can I 😉

PR #4743, opened 2022-11-03
US translation strings

VIKTORVAV99:
This adds US translation strings for (almost all) the languages that already had US translations.


web/public/locales/sv.json
```diff
@@ -1850,6 +1857,14 @@
     },
     "ZW": {
       "zoneName": "Zimbabwe"
+    },
+
+    "US": {
+      "zoneName": "Amerikas f�renta stater"
```
madsnedergaard, line 1863: Is it just Github that cannot show the `ö` or is something else going on? 🤔

VIKTORVAV99: Strange, I'll look into if the coding changed on the file. Or if there is something else going on.

VIKTORVAV99: I think I found the issue, it's neither the coding or github. It's the translation helper...

I'll update the translations manually and then look into that script to see what's going on under the hood.

unitrium (approved):

madsnedergaard:
Why did you remove all the DE zone-name translations, was that on purpose? @VIKTORVAV99

VIKTORVAV99:
> ~~Why did you remove all the DE zone-name translations, was that on purpose? @VIKTORVAV99~~
> Nevermind, I didn't notice the problem in merge commits 😅 

Yep that was on purpose, cause I accidentally included them twice 😅

I think I removed the correct "duplicate".

PR #4746, opened 2022-11-04
run CI only on master pushes

### LINKED ISSUE #4745: All CI jobs are sometimes run twice
Looks like sometimes the CI jobs are run twice if there is a push in a pull request branch in some cases.

I think we can change it so it only runs on push to master and still keep the pull request trigger for all branches.

unitrium:
## Issue
CI runs on every push. Restrict it to run only on master. Close #4745 
## Description

Restrict CI runs to only run on pushes for master. Keep CI run on PRs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,6 +1,8 @@
 name: CI
 on:
   push:
+    branches:
+      - master
   pull_request:
 
 jobs:


VIKTORVAV99 (approved):
LGTM! 🎉

The issue was more a note to myself so I wouldn't forget it while bowling but it's great you fixed it. 😅

PR #4748, opened 2022-11-05
Fix areagraph for missing data

corradio:
## Issue
In this screenshot, we see that the last interconnector is missing (seen by the change in carbon intensity only for the last hour). However, the mix graph shows two missing datapoints. This is because of the way we plot using stepCurves.

Screenshot 2022-11-05 at 10 03 56

The solution is to add a point for each start and end bin, to make sure missing values are plotted correctly.
Here's the solution, where I introduced another missing point in the middle of the series in order to check the fix works in all cases:
Screenshot 2022-11-05 at 10 27 58

This will help better debug our data.


madsnedergaard (approved):
Nice! 🙌

Kongkille:
Heyo Oli, a (sideeffect?) of this is that the gradient is not continous anymore eg

Before:
image


Now:
image

Do you have any thoughts on which version we'd prefer?

VIKTORVAV99:
> Heyo Oli, a (sideeffect?) of this is that the gradient is not continous anymore eg
> 
> Before (color scale is different than the other img): image
> 
> Now: image
> 
> Do you have any thoughts on which version we'd prefer?

I kind of like the new version better you can tell the time intervall apart better.

corradio:
@Kongkille I must admit I favour the new change as it more accurately represents the data.
The co2 value between t and t+1 represents the average of all the observations within that interval, therefore, in my opinion, it makes more sense to use the same color from t to t+1.
The gradient made sense at a time where the value at t and t+1 were instant values and not averages, and thus, we were interpolating between two instant values (as opposed to an average between two times).

Oli

Kongkille:
Yes that makes sense. I still think the graph looks great and the data becomes more readable without the gradient, so let's have a solid colid for each bar.

PR #4749, opened 2022-11-05
Fix negative prices display

### LINKED ISSUE #4729: Price graph cannot handle extreme negatives (-999$ MWh)
**Describe the bug**
The bars in the price graph overflows when there's a very high negative price. See screenshot for example.
I'm not sure how frequent this really is across the world, I just noticed by accident while looking at Australia.

I have verified that data is actually correct ([see AEMO graph here](https://aemo.com.au/energy-systems/electricity/wholesale-electricity-market-wem/data-wem/data-dashboard#market-balancing)), so this is only a frontend error.

**To Reproduce**
1. Go to https://app.electricitymaps.com/zone/AUS-WA
2. Scroll down to price graph

**Expected behavior**
The graph should not overflow if bars are extreme in any end.

**Screenshots**
Screenshot 2022-10-30 at 20 51 05

**Desktop:**
 - OS: MacOS
 - Browser: Chrome

**Smartphone:**
 - Device: iPhone 12 Mini
 - OS: iOS 16.0
 - Browser: Native app

corradio:
## Issue
Fixes #4729 

## Description
- Graph now uses as minimum the minimum price
- Coloured negative prices a bit differently

image


madsnedergaard (approved):
LGTM, thanks for fixing it! 🙌

PR #4750, opened 2022-11-05
Minimize generated `zones.json` and `exchanges.json` by excluding unused objects and keys

VIKTORVAV99:
This PR aims to reduce the frontend bundle by excluding more unused objects and keys from the generated `zones.json` file we bundle.

### TODO:
- [x] Exclude unused zone objects from `zones.json`
- [x] Do the same for exchanges.
- [x] Update tests to match the new structure.

### Analysis:
| File             | Size before  | Size after  | Result      |
|------------------|--------------|-------------|-------------|
| `zones.json`     | 131 420 byte | 68 606 byte | 48% smaller |
| `exchanges.json` | 48 204 byte  | 25 134 byte | 48% smaller |

This also lays the groundwork for what I mentioned in Slack, by removing the unused objects and keys we get fewer changes to the generated files allowing them to be cached for longer by client browsers if we bundle them independently in the refractor.


web/src/helpers/redux.js
```diff
@@ -35,12 +35,10 @@ function initDataState() {
     Object.keys(TIME).forEach((agg) => {
       zone[TIME[agg]] = { details: [], overviews: [], isExpired: true };
     });
-
-    zone.config.capacity = zoneConfig.capacity;
```
VIKTORVAV99, line 39: As far as I can tell we don't use this at all in the frontend and get all the capacities from the state files.

web/generate-zones-config.js
```diff
@@ -14,14 +14,31 @@ const mergeZones = () => {
...
       }
     }
+    /*
+     * The parsers object is only used to check if there is a production parser in the frontend.
+     * This moves this check to the build step, so we can minimize the size of the frontend bundle.
+     */
+    if (zoneConfig?.parsers?.production?.length > 0) {
+      zoneConfig.parsers = true;
+    } else {
+      zoneConfig.parsers = false;
```
VIKTORVAV99, line 40: We could remove the parser object entirely for some additional size savings if we want.

madsnedergaard (approved):
Thanks for doing this work Viktor! 💪 
We've also been very inspired by this in the rewrite and will rethink how we manage all of this not-used data in the frontend :)

I'll merge this PR soon, we currently have an internal bottleneck on releases due to the removal of some US zones, so I'll wait for that before releasing this to keep it isolated to better verify that it works.

VIKTORVAV99:
Figured it was something like that with 10+ days without a release. But take your time, if we included the information for this lone we can do it for a few more days as well.

VIKTORVAV99:
With the merger of #4739 we now have the sources object in these files as well and it's quite large. Now the question is do we want to remove it entirely, or do we want to map it to a top-level object and de-duplicate it (or split it into its own file)?

madsnedergaard:
Hmm, given that we're likely going to change what the backend and frontend exchanges, maybe we should first merge this and then look into the sources during/after rewrite project? :)

VIKTORVAV99:
Should we just exclude the sources for now then since we don't use them in the front-end yet?

madsnedergaard:
> Should we just exclude the sources for now then since we don't use them in the frontend yet?

Ah now I see, because we're only using the sources from app-backend anyway, right?

Screenshot 2022-11-21 at 21 44 56

Screenshot 2022-11-21 at 21 43 06

VIKTORVAV99:
> > Should we just exclude the sources for now then since we don't use them in the frontend yet?
> 
> Ah now I see, because we're only using the sources from app-backend anyway, right?

Correct we get the "parser sources" from the backend directly as these can change (when doing estimations and aggregations) and with the merger of #4739 we also have a top-level source object in the yaml file that get added to the generated json file. These source objects only contain the source for the co2eq calculations/sources.

The latter we don't use att all in the frontend due to them being so new and they are huge size wise as some of the links are massive and get replicated for each zone using them, so for now I think we should exclude them and if we wish to use them later, we should re-map them and deuplicated them in a new top-level object in the generated json.

madsnedergaard:
Sounds great, let's merge it! 🥳

PR #4754, opened 2022-11-08
Adjust SE-SE3 and SE-SE4 border

VIKTORVAV99:
Adjust the SE-SE3 and SE-SE4 border to better match the one shown on https://www.svk.se/kontrollrummet

BEFORE:
image

AFTER:
image

The geometry is a little "smoother" or "simpler" after but I don't think that will matter.


madsnedergaard (approved):
LGTM :)

PR #4755, opened 2022-11-08
Round geopoints to 10 meter accuracy

VIKTORVAV99:
This pr adds a small function to round the geopoints in the `world.geojson` to 4 decimals (10 meter accuracy) before it's used by `update-world.js` to generate the `world.json` file.

This significantly reduce the file size of the `world.json` file from `1 472 010 bytes` to `752 221 bytes` cutting the file size in almost half without compromising the generated map in any noticeable way.


VIKTORVAV99:
@Kongkille not sure if this is what you had in mind when we discussed this on slack but I thought this was an elegant solution that is pretty much plug and play.

Kongkille (commented):
Yes this is indeed what I had in mind :)

It seems we already had a truncate step I had forgotten about. 

I am wondering if there's any difference between truncating the coords vs rounding them?

VIKTORVAV99:
> Yes this is indeed what I had in mind :)
> 
> It seems we already had a truncate step I had forgotten about.
> 
> I am wondering if there's any difference between truncating the coords vs rounding them?

I don't know what that step was doing but it's for sure not rounding/truncating the existing coordinates. Perhaps it only truncate the generated ones?

Usually the difference between rounding is that well rounding rounds the number and truncate cuts it of but I'm not sure how it's implemented here.

madsnedergaard (approved):
LGTM! :)

VIKTORVAV99:
@Kongkille unless you have any objections I'll merge this.

We can look into what the truncate step actually does later but it would be nice to get the size reduction live in production. 🙂

Kongkille:
Feel free to merge @VIKTORVAV99 :)

PR #4757, opened 2022-11-08
Add `US-SE-SOCO` to `US` subZones

VIKTORVAV99:
Looks like we missed to add the newly created zone.


config/zones/US.yaml
```diff
@@ -96,6 +96,7 @@ subZoneNames:
   - US-SW-WALC
   - US-TEN-TVA
   - US-TEX-ERCO
+  - US-SE-SOCO
```
unitrium, line 99: It might be better to keep the ordering, so have it right after US-SE-SOCO.

unitrium (changes requested):

config/zones/US.yaml
VIKTORVAV99: Ah yeah, I'll move it.

unitrium (approved):

PR #4758, opened 2022-11-08
Mounts secret during build

madsnedergaard:
## Description

This makes sure the secret token is used for builds with Earthly.
While testing this on a temporary staging server, I couldn't figure it why it wasn't sending the signature to the app-backend - turns out [it only works on a secure origin (https)](https://stackoverflow.com/a/57170494), so now we throw an error message if this happens.


FelixDQ (approved):
🙌

PR #4765, opened 2022-11-14
SE unknown regional emission factors

VIKTORVAV99:
Updated the emission factors with the new regional values using the SE County level values and averages as we don't yet have zone level data.

### DIFF
--- a/config/zones/SE-SE3.yaml
+++ b/config/zones/SE-SE3.yaml
@@ -97,8 +97,8 @@ sources:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
   UNECE 2022:
     link: https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf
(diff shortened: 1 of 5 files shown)


config/zones/SE-SE3.yaml
```diff
@@ -91,8 +97,8 @@ sources:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
   UNECE 2022:
     link: https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf
-  ? UNECE 2022, WindEurope "Wind energy in Europe, 2021 Statistics and the outlook
-    for 2022-2026" Wind Europe Proceedings (2021)
-  : link: https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf#page=37,
```
VIKTORVAV99, line 96: This object looked a little weird, so I thought I'd take the opportunity to ask if it was intentional or if it should look like the my changes?

PS: `>-` allows for multi line text.

pierresegonne (approved):

config/zones/SE-SE3.yaml
VIKTORVAV99: I'll revert the changes here and see if I can find a way to format them all in a nicer way.

PR #4766, opened 2022-11-15
Md/ele 1406 development zone in west

mathilde-daugy:
## Description

world files updated to reflect new IN zones
existing IN zone config moved to "retired_zones" and replaced by new configs

IN-WE parser, includes production, consumption and exchanges
for **exchanges**, it is hard to tell which value to take. I chose "Current loading" (see below) because PX(import)/PX(exports) seem to be contracted on the Power Exchange. 

image

### Preview

image

@VIKTORVAV99: FYI. we have started the development of IN and we have updated the zones. India is now split according to its power grids and not its states. :)


VIKTORVAV99:
> @VIKTORVAV99: FYI. we have started the development of IN and we have updated the zones. India is now split according to its power grids and not its states. :)

Smart move! It will align India closer to the rest of our zones and will help when we have enough data to aggregate it! 🎉

VIKTORVAV99 (commented):
I will begin doing a full review, but I suggest moving all existing IN parsers (that don't support the new zones) into the archived parsers folder (along with the tests if relevant).

config/zones/IN-EA.yaml
```diff
@@ -0,0 +1,20 @@
+bounding_box:
+  - - 67.434082
+    - 11.393879
+  - - 100.327148
+    - 27.664069
+capacity:
+  battery storage: 0
```
VIKTORVAV99, line 7: Have we confirmed 0 storage capacity? If not it's better to leave it as Null or ommit it in my oppioion?

parsers/IN_WE.py
```diff
@@ -0,0 +1,318 @@
...
+KIND_MAPPING = {
+    "production": {"url": PRODUCTION_URL, "datetime_column": "lastUpdate"},
+    "exchange": {"url": EXCHANGE_URL, "datetime_column": "lastUpdate"},
+    "consumption": {"url": CONSUMPTION_URL, "datetime_column": "current_datetime"},
+}
+
+
+def fetch_data(
+    zone_key: str = "IN-WE",
+    kind: str = None,
```
VIKTORVAV99, line 63:
```suggestion
    kind: Optional[str] = None,
```
Otherwise we can't set it to None without breaking the type.

parsers/IN_WE.py
```diff
@@ -0,0 +1,318 @@
...
+    logger: Logger = getLogger(__name__),
+) -> dict:
+    """- get production data from wrldc.in
+    - filter all rows with same hour as target_datetime"""
+    r = session or Session()
+
+    dt_12_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %I:%M")).datetime
+    payload = {"date": target_datetime.strftime("%Y-%m-%d")}
+
+    resp = r.post(url=KIND_MAPPING[kind]["url"], json=payload)
```
VIKTORVAV99, line 75:
```suggestion
    resp: Response = r.post(url=KIND_MAPPING[kind]["url"], json=payload)
```
Improves type saftey later when calling resp.json()

Response needs to be imported from requests.

parsers/IN_WE.py
```diff
@@ -0,0 +1,318 @@
...
+}
+
+
+def fetch_data(
+    zone_key: str = "IN-WE",
+    kind: str = None,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 67:
```suggestion
) -> pd.DataFrame:
```
So we have the correct return type based on what we actually return.

parsers/IN_WE.py
```diff
@@ -0,0 +1,318 @@
...
+    )  # for older datetimes datetimes are not the same in the AM and the PM which makes sorting harder
+    if len(data) == len(POWER_PLANT_MAPPING):
+        df_production = data.copy()
+    else:
+        for plant in set(data.State_Name):
+            df_plant = data.loc[data.State_Name == plant].copy()
+            for dt in set(df_plant.lastUpdate):
+                df_dt = df_plant.loc[df_plant.lastUpdate == dt]
+                if len(df_dt) > 1:
+                    if target_datetime.hour >= 12:
```
VIKTORVAV99, line 120: Since target datetime can be None we should make sure target_datetime is not None before accesing it's properties.

VIKTORVAV99 (changes requested):
There are a few changes that should be made in regards to type safety but beyond that I don't see any major issues and everything works as intended as far as I can tell.

I'll do a second review when we have cleaned up the types a little.

config/zones/IN-EA.yaml
mathilde-daugy: good point

unitrium (approved):
Looks good to me once @VIKTORVAV99 's feedback has been implemented

mathilde-daugy:
> I will begin doing a full review, but I suggest moving all existing IN parsers (that don't support the new zones) into the archived parsers folder (along with the tests if relevant).

We're not going to archive the parsers because we want to keep collecting data at the most granular level even though it won't show on the map.

VIKTORVAV99:
> > I will begin doing a full review, but I suggest moving all existing IN parsers (that don't support the new zones) into the archived parsers folder (along with the tests if relevant).
> 
> We're not going to archive the parsers because we want to keep collecting data at the most granular level even though it won't show on the map. 

I see. It could become a problem and cause confusion in the future but I suppose it's something that we could address then, but one option would be to move them into a legacy_parsers folder. But that would require other changes as well.

config/retired_zones/README
```diff
@@ -0,0 +1,2 @@
+Those zones have been retired for now to allow for a higher level overview of certain zones.
+They can be reused once an intermediate/state view has been implemented.
```
VIKTORVAV99, line 2: Might want to add the .md extension to this file so it properly reads as a markdown file for all tools and systems.

VIKTORVAV99 (approved):
Once the tests pass this looks good to me. 👍🏼

Kongkille:
Ok i think it should pass now

PR #4778, opened 2022-11-17
add US-FLA-HST to byPassedZones

unitrium:
## Description

Fix bounding box for the US and time zone.
Add US-FLA-HST to the bypassed aggregation subzones as it seems to be often lacking data.


VIKTORVAV99 (approved):
I assume the bounding box changed due to Alaska and Hawaii?

But the changes look good to me 🎉.

PR #4784, opened 2022-11-18
[Fix] Merge again zones that were demerged from the us

unitrium:
## Issue

Some decomissioned/ production only zones where merged out of their parent zones during a merge conflict.

## Description

Remerge some US zones together.


Kongkille (approved):

unitrium:
> As far as I can tell by looking at the map, all of those merged zones dropped out of their partent zone.
> 
> Furthermore, US-MIDW-GLHB was retired as well. There's no need for it to be on the map anymore.

Yes, so far we haven't been able to identify with which BA took over for US-MIDW-GLHB.

PR #4785, opened 2022-11-18
Improve US-CENT-SWPP load parser

pierresegonne:
## Issue

The update of the url for the production per mode forecast does not provide reliable forecasts


unitrium (approved):
LGTM

PR #4786, opened 2022-11-18
Remove decomissioned US-MIDW-GLHB zone

unitrium:
## Issue

US-MIDW-GLHB's only power plant (Joppa Power Plant) has been retired and therefore the zone has been retired by the EIA.

## Description
Remove the zone and fill the remaining space with US-MIDW-MISO.

PR #4787, opened 2022-11-18
restore SE-SE3 SE-SE4 border change

VIKTORVAV99:
Looks like this change got reverted in the merge conflict we had.
~~I'll run the update world script once #4786 is merged to avoid further conflicts.~~ Done!


VIKTORVAV99:
After this is merged we should publish a new mobile app build to get the gemoetry changes into the apps.

madsnedergaard:
> After this is merged, we should publish a new mobile app build to get the geometry changes into the apps.

Good point, will do!

madsnedergaard:
Just noticed the exchange arrow positioning between these two zones is wrong on production, should we fix this here? :)

VIKTORVAV99:
> Just noticed the exchange arrow positioning between these two zones is wrong on production, should we fix this here? :) Screenshot 2022-11-23 at 10 47 27

It's the border that's wrong! 😅

In #4754 I moved both the border and exchange arrow but then there was a merge conflict in another PR and the geometry changes where lost, this PR restores that, so the border is under the arrow again.

Kongkille (approved):

VIKTORVAV99:
@madsnedergaard should be good to go for a new mobile build now.

PR #4795, opened 2022-11-21
fix_in_parser

mathilde-daugy:
## Issue
in the historical refetch there was only one datetime per date so the following changes were made: 
- updated in parser to get all datetimes in one day 
- added refetch frequency (day=1)


parsers/IN_WE.py
```diff
@@ -57,6 +58,12 @@
     "consumption": {"url": CONSUMPTION_URL, "datetime_column": "current_datetime"},
 }
 
+get_date_range = lambda dt: pd.date_range(
```
unitrium, line 61: I'm not a super fan of using lambda functions like this, we're using it 3 times in this module, I think a regular function definition would be just fine. It would add more consistency and if we hover over the function we would get type hints and definition.

So something like this:
```
def get_date_range(dt: pd.DateRange) -> PyDatetime:
your magic
```

parsers/IN_WE.py
```diff
@@ -93,17 +98,46 @@ def fetch_data(
                 item[datetime_col] = dt.shift(minutes=1).floor("minute").datetime
             else:
                 item[datetime_col] = dt.floor("minute").datetime
-        df_data = pd.DataFrame(
-            [item for item in data if item[datetime_col].hour == dt_12_hour.hour]
-        )
-        return df_data
+        return data
```
unitrium, line 101: I'm thinking we can write the exception catching a little more concise:
```
try:
except json.decoder.JSONDecodeError:
        raise ParserException(
            parser="IN_WE.py",
            message=f"{target_datetime}: {kind} data is not available",
        )
 datetime_col = KIND_MAPPING[kind]["datetime_column"]
 for item in data:
    ... do your stuff
 return data
    
```
This way no need for ```is_response_content_empty = False``` and the extra check at the end :)

unitrium, line 101: Are we still returning a data frame ? For me it looks like we're returning a dictionary. Either we should update the signature of the function or make sure to return a data frame. Besides I think that now all the code above is not used anymore because we return data. All the operations done on items do not seem to have an effect?
```
        return data
```

parsers/IN_WE.py
```diff
@@ -93,17 +98,46 @@ def fetch_data(
...
 
 
+def format_raw_data(
+    kind: str,
+    data: json,
+    zone_key: str,
+    target_datetime: datetime,
+    zone_key2: Optional[str] = None,
+) -> dict:
+    assert data is not None
```
unitrium, line 116: Nothing to change there, it's just a small comment.
I know we discussed whether we should include checks in our functions to make user users adhere to the function signature.

Usually functions should be self contained we should not assume other functions to take care of sanitization but it could be acceptable that since we supply type hints, if the user supplies a None they should expect stuff will go wrong, at least in other languages a None wouldn't be accepted :) So I would say it's also acceptable to not put the asserts.

parsers/IN_WE.py
```diff
@@ -93,17 +98,46 @@ def fetch_data(
...
+    assert kind is not None
+
+    dt_12_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %I:%M")).datetime
+    datetime_col = KIND_MAPPING[kind]["datetime_column"]
+    df_data = pd.DataFrame(
+        [item for item in data if item[datetime_col].hour == dt_12_hour.hour]
+    )
+    if kind == "production":
```
unitrium, line 124: I'm thinking we could do a mapping here but I'm not sure it would improve the readability of the code.
Something like:
```
FORMAT_FUNCTIONS = {
"production": format_production_data,
"consumption": format_consumption_data,
"exchange": format_exchanges_data
}
formatted_data_point = FORMAT_FUNCTIONS[kind](**kwargs)
```
But might make the code more complex to read and there's still some logic to do for exchange anyways.

unitrium (approved):
Small code improvement comments

unitrium (approved):
Looks great! Can't wait for India-WE to be on the map!

PR #4800, opened 2022-11-23
Italian interconnectors disappear when toggling country aggregation #4767

purplepaste5:
## Issue
Italy and Montenegro country interconnector was not displaying due to a missing yaml file

## Description

Added a yaml file for IT to ME interconnection

PR #4801, opened 2022-11-23
Update CI to run verify on zone & exchange changes

VIKTORVAV99:
The changes should be pretty self explanatory but the gist of it is that this PR makes our CI more robust by checking if `excluded-aggregate-exchanges.json` should update or not (matching `world.json` behavior) and changes the verify triggers to run on `/geo/`, `/config/zones/` and `/config/exchanges/` changes.

Also I added the updated `excluded-aggregate-exchanges.json` I forgot to check in #4800.


.github/workflows/CI.yml
```diff
@@ -99,7 +111,7 @@ jobs:
   verify:
     name: Verify
     needs: [setup, prettier]
-    if: needs.setup.outputs.frontend != 'false' || needs.setup.outputs.workflows != 'false'
+    if: needs.setup.outputs.geometries != 'false' || needs.setup.outputs.workflows != 'false'
```
madsnedergaard, line 114: Is this change on purpose - shouldn't it still run on frontend changes? 🤔

madsnedergaard (approved):
One minor question, otherwise LGTM!

.github/workflows/CI.yml
VIKTORVAV99: It now runs on all the web/geo changes and zone + exchange changes. There should be no need to check if the map has updated if we are just changing "pure" frontend components like say the translations.

VIKTORVAV99: @madsnedergaard merging this now, but if there are any issues, we should spot them pretty easily as we run this and the CirlceCI jobs in parallel.

madsnedergaard: Ah okay, I thought we would need either `.frontend` or `.geometries` to be not false, but makes sense :)

VIKTORVAV99: Hmm, good point. It might be needed for cypress since it needs verify. I'll monitor the situation and add `.frontend` too if it's needed.

Issue #4434, opened 2022-08-09
Click and drag on cross border export arrow not panning map


cshuenss:
**Describe the bug**
Clicking and dragging when the mouse cursor is on a cross border exprort arrow drags the image of the cross border export arrow.

**To Reproduce**
At [https://app.electricitymaps.com/map](https://app.electricitymaps.com/map), hover the mouse above a cross border export such that the info box appears. Click, hold and drag.

**Expected behavior**
The map pans.

**Screenshots**
![Screenshot 2022-08-09 153201](https://user-images.githubusercontent.com/71438287/183677200-7f6cdf38-3e2e-468a-aae3-0e69746378ef.png)
The large arrow pointing straight down above Poland is the dragged image.

**Desktop (if relevant):**
Bug appears to be Firefox specific (Firefox 103.0). Works as intended on Chrome 104.0, Edge 104.0.

nicolasca:
I can reproduce the issue with Chrome 107
<img width="1336" alt="Capture d’écran 2022-10-28 à 09 14 50" src="https://user-images.githubusercontent.com/2886734/198526504-aa7458bc-d199-4077-8486-f4e2fbc5d2c7.png">

BStephenBB:
Going to take this one if that's okay

PR #4803, opened 2022-11-23
Fix arrow drag

BStephenBB:
## Issue
If a user happens to click on an export arrow while they're dragging the map around, they will drag the arrow image rather than move the map. See #4434 
## Description
This PR fixes this issue by simply adding `draggable={false}` to the arrow images, which prevents this interaction. This could also have been solved by adding a `onDragStart` event handler that returns false, or by turning off pointer events, but those solutions are more complex or mess up the tooltip interaction, so I went with just setting draggable to false. 

### Preview
Example of the issue before the fix:
bad-drag
And after the fix:
good-drag


VIKTORVAV99:
I'm going to test it tomorrow but should be pretty straight forward so I see no reason why it can't be merged.

Also CC: @madsnedergaard, @Kongkille, @tonypls you guys should be able to add this to the rewrite as well without any problems.

Kongkille:
That's nice! I didn't know this property existed

VIKTORVAV99 (approved):
LGTM! 🎉

PR #4805, opened 2022-11-25
Adds hardcoded env for now

madsnedergaard:


tonypls (approved):

VIKTORVAV99:
Is this something we need to add to the CI as well or is it purely for the backend?

madsnedergaard:
Closing again, as it was not the root cause of the problems.

madsnedergaard:
> Is this something we need to add to the CI as well or is it purely for the backend?

Nope, it was just an attempt at fixing a bug as we migrated to deploy the app with Earthly - but turns out the problem was #4806 instead :D

PR #4806, opened 2022-11-25
Ignore index.html

FelixDQ:
Ignore index.html. This was previously ignore in .dockerignore, but earthly can't read that file.


madsnedergaard (approved):
Yay 🥳

PR #4807, opened 2022-11-25
Change CI to use config + frontend instead of geometries for verify

VIKTORVAV99:


VIKTORVAV99:
Looks like you where right @madsnedergaard we do need both `.frontend` and `.geometries` (now `.config) otherwise cypress won't run on frontend changes. 😅

madsnedergaard (approved):
Haha fair, thanks for the quick response 👏

PR #4810, opened 2022-11-28
add US-SE-SEPA to bypassed subzones for the US

unitrium:
## Issue

US-SE-SEPA is frequently without data lately. Add it to the bypassed subzones until we activate an estimation model for it.


VIKTORVAV99 (approved):
LGTM

PR #4814, opened 2022-11-28
update biomass lca factors for europe

mathilde-daugy:
## Description

reverted biomass lca emission factors for Europe to IPCC 2014 (230 g/kwh)

### Preview

NA


VIKTORVAV99:
Why are these getting reverted?

mathilde-daugy:
> Why are these getting reverted?

Hi Viktor, we've decided that the sample used to compute these factors was not representative enough. For instance for France we only have data for one PP on Entsoe when there is at least 2 GW of installed capacity on the territory. (this is the case in most of the EU) Computing a factor for the whole zone using data from only one PP is not very accurate. Also, it seems like depending on the fuel input, the EF will vary a lot and we do not have this information at this point.

We think it is better to use the IPCC 2014 factors for now. In the mean time, we will keep working on finding a better estimate and more precise methodology for biomass in Europe :)

VIKTORVAV99:
> > Why are these getting reverted?
> 
> Hi Viktor, we've decided that the sample used to compute these factors was not representative enough. For instance for France we only have data for one PP on Entsoe when there is at least 2 GW of installed capacity on the territory. (this is the case in most of the EU) Computing a factor for the whole zone using data from only one PP is not very accurate. Also, it seems like depending on the fuel input, the EF will vary a lot and we do not have this information at this point.
> 
> We think it is better to use the IPCC 2014 factors for now. In the mean time, we will keep working on finding a better estimate and more precise methodology for biomass in Europe :)

Ah okay, then it does indeed make sense to revert these and look for a better source or wait until the ENTSO-E database contains more individual power plants.

I'll submit a PR that fixes Sweden unknown EF after this is merged.

asfi30:
The biomass really depends on the fuel used and the Power Plant efficiency. For example Bioethanol (E100) has specific = 230g/KWh, Biomethanol (M100) =250g/KWh, Biodiesel (B100)=252g/KWh, Wood=410g/KWh. It really matters if you directly burn the biomass that is similar to the burning of CHP plant, or you make Cogeneration  Power Plant where the steam is used not for heating homes but for the cooking and drying of the biomass. In this way you have 2 products - Electricity and Fertilizers.

config/zones/AT.yaml
```diff
@@ -105,16 +105,16 @@ emissionFactors:
         value: 189.1555066532277
     biomass:
       datetime: '2021-01-01'
-      source: EU-ETS, ENTSO-E 2021
-      value: 439.145806
+      source: IPCC 2014; EU-ETS 2021
```
pierresegonne, line 108: Why do we still have EU-ETS as source? Shouldn't we only have IPCC?

config/zones/SK.yaml
```diff
@@ -142,21 +142,20 @@ emissionFactors:
...
       value: 1124.903938
     solar:
       datetime: '2021-01-01'
       source: INCER ACV
       value: 33.0
     wind:
       datetime: '2021-01-01'
-      source: UNECE 2022, WindEurope "Wind energy in Europe, 2021 Statistics and the
-        outlook for 2022-2026" Wind Europe Proceedings (2021)
+      source: INCER ACV
```
pierresegonne, line 158: Das er problem?

config/zones/SI.yaml
```diff
@@ -280,10 +279,4 @@ sources:
     link: https://docs.google.com/spreadsheets/d/1w5DJ7sPen6axIHU8TCVcuzNCjlct4I6JAbhUlw-ZXu8/edit?usp=sharing
   IPCC 2014:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
-  UNECE 2022:
```
pierresegonne, line 283: Why have the wind things disappeared/

config/zones/BE.yaml
```diff
@@ -46,10 +46,10 @@ emissionFactors:
...
     biomass:
       datetime: '2021-01-01'
-      source: EU-ETS, ENTSO-E 2021
+      source: BEIS 2021
       value: 0.0
     coal:
-      datetime: '2020-01-01'
+      datetime: '2021-01-01'
       source: EU-ETS, ENTSO-E 2021
       value: 1057.262766
```
VIKTORVAV99, line 54: Not feedback related to this pr in particular but shouldn't it be:
```suggestion
    - datetime: '2014-01-01'
      source: IPCC 2014
      value: 820
    - datetime: '2021-01-01'
      source: EU-ETS, ENTSO-E 2021
      value: 1057.262766
```
It would increase the transparency of the sources (even if these are the defaults) and would make it more clear when looking directly at the files.

Just an idea though so don't let it be a blocker for this PR.

config/zones/ES-CN-FVLZ.yaml
```diff
@@ -23,10 +23,10 @@ emissionFactors:
         value: 345.4548264323977
     biomass:
       datetime: '2021-01-01'
-      source: EU-ETS, ENTSO-E 2021
+      source: BEIS 2021
       value: 0.0
     coal:
-      datetime: '2020-01-01'
```
pierresegonne, line 29: Wasn't EU-ETS data for 2020?

pierresegonne (commented):
Just unsure about the comment I left otherwise LGTM

config/zones/ES-CN-FVLZ.yaml
mathilde-daugy: no EU-ETS is 2021

config/zones/BE.yaml
mathilde-daugy: Thanks for your comment Viktor. We'll see how we incorporate factors for previous years. We want to compute them for EU as the data is available

pierresegonne (approved):

Issue #3656, opened 2022-01-10
Iso3166 GB-SHI seems to be not valid according to ISO3166-2


Fastaxx:
For Shetland island, the iso-code `GB-SHI` seems not to be correct. Indeed

https://www.iso.org/obp/ui/fr/#iso:code:3166:GB

Should it be renamed as `GB-ZET` as proposed in the ISO code?

https://en.wikipedia.org/wiki/ISO_3166-2:GB

PR #4817, opened 2022-11-29
Migrate Shetland Island to use `GB-ZET` as the zone code to match ISO

VIKTORVAV99:
## Issue
The current zone code doesn't match the ISO subdivision code for the Shetland Islands.
## Description
Changes Shetlands zone key from `GB-SHI` to `GB-ZET` to match ISO.

Closes: #3656

>**Note**
> Since we don't have any parser or data for the Shetland Islands this should be a straightforward and safe migration.


unitrium (approved):
Looks good to me!

PR #4818, opened 2022-11-29
Update US config

Kongkille:
## Issue
We have data for the US but the frontend is not rendering it because it doesn't know whether there's a parser or estimation.

Note: this is something we should ensure doesn't happen with the rewrite @madsnedergaard @tonypls 

## Description

Adding an estimation method to the config should do the trick

### Preview
image


tonypls:
LGTM

VIKTORVAV99:
Don't forget to push a new build (and mobile build) as well.

PR #4819, opened 2022-11-29
updated the occurence of electricityMap in config files

SaumyaBhushan:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=mIBu0oz)
## Issue
https://github.com/electricitymaps/electricitymaps-contrib/issues/4816
## Description
We are mentioning electricityMap multiple times in the config files. It would be awesome if someone could replace all of these mentions to Electricity Maps

### DIFF
--- a/config/zones/AM.yaml
+++ b/config/zones/AM.yaml
@@ -19,32 +19,32 @@ emissionFactors:
   direct:
     battery discharge:
       - datetime: '2019-01-01'
-        source: electricityMap, 2019 average
+        source: Electricity Maps, 2019 average
         value: 165.87320597981062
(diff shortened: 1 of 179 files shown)


VIKTORVAV99 (changes requested):
That was a lot of changes 😅

Looks like two links where broken in the process but I added suggestions to fix them that you can just commit.

SaumyaBhushan:
Thanks @VIKTORVAV99 . I have commited those changes

VIKTORVAV99 (approved):
LGTM! 🎉

PR #4824, opened 2022-12-01
Switch to contiguous US

unitrium:
## Issue

Colorizing Alaska and Hawaii can be misleading for users as they might think we have data for those zones when looking at the aggregation.

## Description

Exclude Alaska and Hawaii from the US aggregation and rename the US zone to contigous united states to avoid any confusion.

image


VIKTORVAV99 (commented):
I think that's fine but then we definitely need to change the zone key so it's free when we do have for all of the US later and don't run inte to same issue as with Spain and France.

Do you know if there is a ISO code for "mainland" United States?

unitrium:
This is intended to be temporary, we don't intend on keeping the denomination contiguous us.
Changing the zone key would require some changes in the backend, essentially moving the existing data to the new zone key. So I would rather be in favor of not doing that.

unitrium:
There's no ISO code for contiguous US I believe.

unitrium:
Also it would mean that later once we have the data for Alaska and Hawaii we would need to rename the zone key and change the data again.

unitrium:
Though we already have data for an island of Hawaii that we would have to reintegrate to the US at some point. So we would have to rerun the flowtracing anyways when we reintegrate Hawaii and Alaska.

web/public/locales/sv.json
```diff
@@ -1843,7 +1843,7 @@
       "zoneName": "Zimbabwe"
     },
     "US": {
-      "zoneName": "Amerikas förenta stater"
+      "zoneName": "Sammanhängande Amerikas förenta stater"
```
VIKTORVAV99, line 1846: Not sure if people actually write it this way and it might be a little long so I suggest we just use "Sammanhängande USA" to make it a bit shorter.
```suggestion
      "zoneName": "Sammanhängande USA"
```

config/zones/US.yaml
```diff
@@ -114,14 +114,14 @@ subZoneNames:
...
-  - US-HI-NI
-  - US-HI-OA
+  # - US-HI-HA
+  # - US-HI-KA
+  # - US-HI-KH
+  # - US-HI-LA
+  # - US-HI-MA
+  # - US-HI-MO
+  # - US-HI-NI
+  # - US-HI-OA
```
VIKTORVAV99, line 124: Maybe we should add a small note why they are commented out?

VIKTORVAV99 (approved):
Discussed the first issue I raised on Slack and we decided this is not really a problem for the US zone so this looks good to me.

I did add two small suggestions though.

VIKTORVAV99:
Oh you had auto-merge on 😅
Good thing the suggestions where not that important...

PR #4839, opened 2022-12-05
Follow up on e2e testing

Kongkille:
## Issue
Follows up on madsnedergaard/ele-1363-duplicate-current-e2e-tests

## Description

This makes our test pass. I have not moved it into component testing yet and mostly just commented things out as a lot of functionality is not working.

This is quite a mess and a fairly big regression compared to our production e2e tests.

I also ended up changing the structure of the mockserver to a more elegant structure now that we've moved away from a url parameter countrycode.


Kongkille:
Looks like this opened in the wrong repo woops

PR #4840, opened 2022-12-05
Update datasource for UY production parser

ZachEichen:
## Description

The old UY parser used a webpage on UTY (Uruguay's national energy producer)'s website that had been taken down. Because of it, the parser was no longer functional. To remedy this, I changed the data source to pull from [UTE's energy generated infopage](https://ute.com.uy/energia-generada-intercambios-demanda).

### Changes: 
* Fixes broken UY production  parser
* Update UY production parser datasource to `https://ute.com.uy/energia-generada-intercambios-demanda`
*  adds parser for current total consumption, grabbed from the same website.


parsers/UY.py
```diff
@@ -46,59 +64,98 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+        hour_recs.append(datum)
+    return hour_recs
+
+
+def fetch_consumption(
+    zone_key: str = "UY",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 142:
```suggestion
def fetch_consumption(
    zone_key: str = "UY",
    session: Optional[Session] = Session(),
    target_datetime: Optional[datetime] = None,
    logger: Logger = getLogger(__name__),
) -> dict:
```
Then you can remove the optional sessions on `parse_page`, `correct_for_salto_grande` and `get_salto_grande` and just the session type.

parsers/UY.py
```diff
@@ -46,59 +64,98 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+        "zoneKey": zone_key,
+        "datetime": entry["time"].datetime,
+        "consumption": entry["demand"],
+        "source": "ute.com.uy",
+    }
+
+    return data
 
 
 def fetch_production(
```
VIKTORVAV99, line 161: Same as above, if you add `Session()` as the default then we can remove the optional types and just use required types.

VIKTORVAV99 (changes requested):
Thanks for making this pull request!

Everything works locally but there are a few changes that should be made and a few that are optional. I've clearly stated which are not required.

I can also see that the graph on the source you linked has data for the last 24 hours. It would be nice if the parser could return all of this data at once. So if it fails for 1 hour for any reason the backend can fill that hour gracefully the next time the parser is run.

PS: It's nice to see that you collaborated with someone on this!

parsers/UY.py
```diff
@@ -15,27 +14,46 @@
...
-    "Biomasa": "biomass",
-    "Térmica": "oil",
+    "hydro": "r",
+    "wind": "eolica",
+    "solar": "fotovoltaica",
+    "biomass": "biomasa",
+    "unknown": "termica",
+    "trade": "intercambios",
+    "demand": "demanda",
+    "comprassgu": "comprassgu",
```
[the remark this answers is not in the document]
VIKTORVAV99: I was thinking a more descriptive key for `comprassgu` so it's clear to everyone what it actually is (including me 😅).

parsers/UY.py
```diff
@@ -14,28 +13,47 @@
...
+UTE_URL = url = "https://ute.com.uy/energia-generada-intercambios-demanda"
 
 SALTO_GRANDE_URL = "http://www.cammesa.com/uflujpot.nsf/FlujoW?OpenAgent&Tensiones y Flujos de Potencia&"
 
 
-def get_salto_grande(session: Optional[Session]) -> float:
+def get_salto_grande(session: Session, targ_time: Optional[datetime] = None) -> float:
     """Finds the current generation from the Salto Grande Dam that is allocated to Uruguay."""
+    # handle optional arguments
+    s = session or Session()
```
VIKTORVAV99, line 40: This can be removed since we require session to be defined in this method now.

parsers/UY.py
```diff
@@ -14,28 +13,47 @@
...
     current_time = arrow.now("UTC-3")
-    if current_time.minute < 30:
-        # Data for current hour seems to be available after 30mins.
+    if (
+        current_time.floor("hour") == targ_time.floor("hour")
+        and current_time.minute < 30
+    ):
+        print(
+            "Looking at previous hour's data for salto grande, because it is too soon after the hour"
+        )
```
VIKTORVAV99, line 52: I think we can remove this print statment.
Or alternatively pass the logger to this function and then use that, but I don't think we need to log this as it's not a error or unexpected behavior.

parsers/UY.py
```diff
@@ -14,28 +13,47 @@
...
         current_time = current_time.shift(hours=-1)
-    lookup_time = current_time.floor("hour").format("DD/MM/YYYY HH:mm")
 
-    s = session or Session()
+    lookup_time = lookup_time.floor("hour").format("DD/MM/YYYY HH:mm")
```
VIKTORVAV99, line 55:
```suggestion
    lookup_time: str = lookup_time.floor("hour").format("DD/MM/YYYY HH:mm")
```

VIKTORVAV99 (commented):
There are still some changes that need to be addressed since the last review but it's going in the right direction so nice work!

It looks like `parse_page` now fetches all the data for the past hour but `fetch_production` itself only returns one datapoint instead of a list of datapoints.

ZachEichen:
> It looks like `parse_page` now fetches all the data for the past hour but `fetch_production` itself only returns one datapoint instead of a list of datapoints.

What is the desired behavior for returning a list of datapoints? Should this just happen if no `target_datetime` argument is passed?

VIKTORVAV99:
> > It looks like `parse_page` now fetches all the data for the past hour but `fetch_production` itself only returns one datapoint instead of a list of datapoints.
> 
> What is the desired behavior for returning a list of datapoints? Should this just happen if no `target_datetime` argument is passed?

The desired behavior is to always return a list of data if possible and to be consistent about it's timespan as we can use this to set a refetch frequency and because it increases the reliability of the parser over time.

For example the Entso-E parser always ruturn the last 48 hours worth of data if no datetime is specified and if a datetime is specified it returns the data for the datetime and the 48 hours before it.

This means it has a refetch frequency of 48 hours or 2 days.

Now in this case it looks like the dashboard we are parsing always have data available for the last 24 hours so it should return all that data as a list of datapoints and if it supports fetching historical data it should return that datetime and the 24 hours before that if possible.

Hope this adds some clarity on the topic.

PS: If you could I would greatly appreciate it if you marked this PR as a draft until it's ready for a final review but do feel free to ping me if you have any questions.

VIKTORVAV99:
> So the data source we're using returns data from the last 24 hours, but not anything before that, so the parser is only able to get the full 24 hr data stream for the current time. If the parser is called with a specified time within the last 24 hours, we will have historical data, but it will be for the range between 24 hours ago and the time specified. In that case, would it be better for the parser to return a list of all available time data, or to just return the most applicable data entry?

I think the best option would be to just forget about the `target_datetime` and raise a parser exception saying the parser don't support fetching historical data if it passed to the parser functions, like we have done for the other parsers that only have real-time data.

Parser exception can be imported with:
```python
```

And then always return all available data when the `fetch_production` and `fetch_consumption` methods are called without a `target_datetime`.

parsers/UY.py
```diff
@@ -1,41 +1,57 @@
...
+    s = session or Session()
+    lookup_time = datetime.now(tz=tz) if targ_time is None else targ_time
 
+    # Data for current hour seems to be available after 30mins.
+    # if we're too soon into the hour, check the hour before
     current_time = arrow.now("UTC-3")
-    if current_time.minute < 30:
-        # Data for current hour seems to be available after 30mins.
+    if (
+        current_time.floor("hour") == targ_time.floor("hour")
```
VIKTORVAV99, line 47: Since `targ_time` can be undefined in it's current form calling `.floor()` on it generates a type error.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+    Queries the url in UTE_URL, and parses hourly production and trade data
+    and retruns the results as a list of dictionary objects
+    """
+    # load page
+    resp = session.get(UTE_URL)
+
+    # get the encoded XML with all the data we want
+    soup = BeautifulSoup(resp.text, "html.parser")
+    xml_tag = soup.find(id="valoresParaGraficar")
+    xml_string = f"<{xml_tag.contents[0]}>{xml_tag.contents[1]}"
```
VIKTORVAV99, line 101: You need to check that the `xml_tag` is defined before calling it's properties, current behavior results in a type error.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+            for key, spanish in MAP_GENERATION.items()
+        }
+        # solar can sometimes return -0.1 at night, round up to 0
+        datum["solar"] = max(datum["solar"], 0)
+
+        # ingest date field
+        datefield = hour.find("hora").contents[0]
+        datestr = re.findall("\d\d/\d\d/\d\d\d\d \d+:\d\d", str(datefield))[0]
+        date = arrow.get(datestr, "DD/MM/YYYY h:mm").replace(
+            tzinfo=dateutil.tz.gettz(tz)
```
VIKTORVAV99, line 123: `tz` is not a property/member of the dateutil module.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+
+def fetch_consumption(
+    zone_key: str = "UY",
+    session: Optional[Session] = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> Union[dict, List[dict]]:
+
+    if target_datetime is not None:
+        raise ParserException("This parser is unnable to parse dates in the past")
```
VIKTORVAV99, line 164: `ParserException` needs two arguments, parser and message.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+            raw_entry = correct_for_salto_grande(raw_entry, session)
+        entries.append(make_output(raw_entry))
 
-    return obj
+    return entries
+
+
+def fetch_consumption(
+    zone_key: str = "UY",
+    session: Optional[Session] = Session(),
```
VIKTORVAV99, line 158: I think this one is on me but session should not be optional.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
-) -> dict:
-    if target_datetime:
-        raise NotImplementedError("This parser is not yet able to parse past dates")
+) -> Union[dict, List[dict]]:
 
-    obj = parse_page(session)
+    if target_datetime is not None:
+        raise ParserException("This parser is unnable to parse dates in the past")
```
VIKTORVAV99, line 187: Same as the other comment, two arguments are needed.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
 def fetch_production(
     zone_key: str = "UY",
-    session: Optional[Session] = None,
+    session: Optional[Session] = Session(),
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
-) -> dict:
-    if target_datetime:
-        raise NotImplementedError("This parser is not yet able to parse past dates")
+) -> Union[dict, List[dict]]:
```
VIKTORVAV99, line 184: The return type should only be `List[dict]` as we should always return lists now.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+        }
+
+    # then delegate all the actual work to get_entry_list
+    return get_entry_list(session, make_datum, correct_hydro=False)
 
 
 def fetch_production(
     zone_key: str = "UY",
-    session: Optional[Session] = None,
+    session: Optional[Session] = Session(),
```
VIKTORVAV99, line 181: Optional can be removed here.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
-    return data
+    # then delegate all the actual work to get_entry_list
+    return get_entry_list(session, make_datum, correct_hydro=True)
 
 
 def fetch_exchange(
     zone_key1: str = "UY",
     zone_key2: str = "BR-S",
-    session: Optional[Session] = None,
+    session: Optional[Session] = Session(),
```
VIKTORVAV99, line 205: Optional can be removed here as well.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
 
-    session = session or Session()
-    if target_datetime:
-        raise NotImplementedError("This parser is not yet able to parse past dates")
+    if target_datetime is not None:
+        raise ParserException("This parser is unnable to parse dates in the past")
 
     # set comparison
     if {zone_key1, zone_key2} != {"UY", "BR"}:
         return None
```
VIKTORVAV99, line 216: it should raise an ParserException instead of returning None.

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
 
 
 def fetch_exchange(
     zone_key1: str = "UY",
     zone_key2: str = "BR-S",
-    session: Optional[Session] = None,
+    session: Optional[Session] = Session(),
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> dict:
```
VIKTORVAV99, line 208: Return type should be `List[dict]`

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+    xml_string = f"<{xml_tag.contents[0]}>{xml_tag.contents[1]}"
+
+    # Use beautiful soup  to actually parse that XML string, and get information
+    xml_doc = BeautifulSoup(xml_string, "lxml")
+
+    # Data is encoded hourly... get that subsection of the document
+    # then get the data for each encoded hour, as a json-like object
+    hourly = xml_doc.find("fuentesporhora")
+    hour_recs = []
+    for hour in hourly.find_all("nodo"):
```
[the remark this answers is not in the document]
ZachEichen: what should I return if the web fetch fails?

parsers/UY.py
ZachEichen: should this also be the case for `fetch_production`? I'm confused cause this optional Session seems to be in line with the format defined in the wiki.

parsers/UY.py
ZachEichen: Should this also be the case for `fetch_consumption`? I'm confused cause this optional Session seems to be in line with the format defined in the wiki for `fetch_production`.

parsers/UY.py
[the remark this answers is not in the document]
VIKTORVAV99: Yeah a `ParserException` would be appropriate.

parsers/UY.py
VIKTORVAV99: Technically it is in like with the Wiki but we will be moving away from it, I just haven't had the time to update the examples and wiki yet. The new method makes it more clear and generate less code and type errors.

parsers/UY.py
VIKTORVAV99: Copy pasting my previous reply here:

Technically it is in like with the Wiki but we will be moving away from it, I just haven't had the time to update the examples and wiki yet. The new method makes it more clear and generate less code and type errors.

parsers/UY.py
ZachEichen: Same as above, I'm confused because the optional `session` argument seems to be in line with what's recommended in the format defined in the wiki for `fetch_exchange`

parsers/UY.py
ZachEichen: should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: Should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
```diff
@@ -46,118 +62,182 @@ def get_salto_grande(session: Optional[Session]) -> float:
...
+        datum["time"] = date
+
+        # add to list
+        hour_recs.append(datum)
+    return hour_recs
+
+
+def get_entry_list(
+    session: Session, make_output: Callable[[dict], dict], correct_hydro=True
+) -> dict:
```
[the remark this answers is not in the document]
ZachEichen: should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: Are you referring to `dateutil.tz.gettz`?  if so, I seem to be able to able to find that in the [documentation](https://dateutil.readthedocs.io/en/2.8.2/tz.html#dateutil.tz.gettz)  for the version of `dateutil` that is in the poetry.lock file 

If you're referring to the argument, I've updated the name of the variable to `TIME_ZONE` for clarity in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
```diff
@@ -1,41 +1,57 @@
...
+UTE_URL = url = "https://ute.com.uy/energia-generada-intercambios-demanda"
 
 SALTO_GRANDE_URL = "http://www.cammesa.com/uflujpot.nsf/FlujoW?OpenAgent&Tensiones y Flujos de Potencia&"
 
 
-def get_salto_grande(session: Optional[Session]) -> float:
+def get_salto_grande(session: Session, targ_time: Optional[datetime] = None) -> float:
     """Finds the current generation from the Salto Grande Dam that is allocated to Uruguay."""
+    # handle optional arguments
+    s = session or Session()
```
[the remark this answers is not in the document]
ZachEichen: Should be fixed in 950e3dad8ab32c0cea9d86c7fee46a794ccc36f5

parsers/UY.py
ZachEichen: So, when I update that the parser starts failing the interface test on CI. Is this okay, or is there some way I should fix that error

ZachEichen:
I'm in a spot of of confusion when it comes to removing the `Optional` session arguments for the `fetch_` methods for the parser. When I make the session non-optional, the parser no longer aligns with the spec. I think I might have misunderstood what you meant when you said to remove the `Optional` arguments. I can only think of three ways that this can be done. 

1. Remove the `Optional` specifier for the session argument but change nothing else. This seems confusing because it means that the type hints are not in line with the actual status of the method, but the parser passes the tests from the test suite
2. Leave the methods with the `Optional` type hints, making a note that the optional argument shouldn't be used. This is a variation on (1), but keeps the type hints up to date with how the functions actually work
3. Remove the `Optional` hint alltogether and make `session` a positional arugment rather than a keyword argument. 

Because it felt most consistent with what you said, I implemented (3). Did you actually mean something more like (1) or (2)?

VIKTORVAV99:
> I'm in a spot of of confusion when it comes to removing the `Optional` session arguments for the `fetch_` methods for the parser. When I make the session non-optional, the parser no longer aligns with the spec. I think I might have misunderstood what you meant when you said to remove the `Optional` arguments. I can only think of three ways that this can be done.
> 
> 1. Remove the `Optional` specifier for the session argument but change nothing else. This seems confusing because it means that the type hints are not in line with the actual status of the method, but the parser passes the tests from the test suite
> 2. Leave the methods with the `Optional` type hints, making a note that the optional argument shouldn't be used. This is a variation on (1), but keeps the type hints up to date with how the functions actually work
> 3. Remove the `Optional` hint alltogether and make `session` a positional arugment rather than a keyword argument.
> 
> Because it felt most consistent with what you said, I implemented (3). Did you actually mean something more like (1) or (2)?

You can declare it like this:
```python
def fetch_consumption(
    zone_key: str = "UY",
    session: Session = Session(),
) -> List[dict]:
```

That way if there is no session argument provided when calling the function it will use the default, Session(), and if it's called with the argument it will just use that. This means we don't have to do any null checks, that it will be more type safe and just produce less code in general.

ZachEichen:
Okay, the changes to typing should be done in commit 080d17bef89ecd36a941c1e752ef3368892836d7. Let me know if that's good.

parsers/UY.py
```diff
@@ -1,163 +1,264 @@
...
-    if current_time.minute < 30:
-        # Data for current hour seems to be available after 30mins.
-        current_time = current_time.shift(hours=-1)
-    lookup_time = current_time.floor("hour").format("DD/MM/YYYY HH:mm")
+    if (
+        current_time.floor("hour") == targ_time.floor("hour")
+        and current_time.minute < 30
+    ):
+        # Looking at previous hour's data for salto grande, because it is too soon after the hour
+        targ_time = targ_time.shift(hours=-1)
```
VIKTORVAV99, line 48: Isn't there a risk we get data for the previous hour of salto grande and the current hour for the test with this?

You probably need to add some validation somewhere to make sure both datetimes match.

parsers/UY.py
```diff
@@ -1,163 +1,264 @@
...
+        raise ParserException(
+            parser="UY", message="This parser is unnable to parse dates in the past"
+        )
 
     # set comparison
     if {zone_key1, zone_key2} != {"UY", "BR"}:
-        return None
+        raise ParserException(
+            parser="UY", message="This parser is unnable to parse dates in the past"
+        )
```
VIKTORVAV99, line 236: This parser message don't seem to match what the if statement actually does?

parsers/UY.py
```diff
@@ -1,163 +1,264 @@
...
 
-    obj = parse_page(session)
-    netFlow = obj["Interconexión con Brasil"]  # this represents BR->UY (imports)
-    if zone_key1 != "BR":
-        netFlow *= -1
+    # flip directions if needed
+    direction_coeff = -1 if zone_key1 == "UY" else 1
```
VIKTORVAV99, line 239: This is not needed since the parser only support `BR->UY` exchanges and the keys are always sorted alphabetically.

VIKTORVAV99 (commented):
Things are looking better but there are still a few old comments that haven't been addressed and some other stuff that needs to be addressed. Mostly regarding the potential mismatch of data that can occur with the current implementation.

parsers/UY.py
ZachEichen: I think this whole function in the parser is never used, cause the BR->UY transfer is handled on the brazil side/ Should I just delete this method for clarity?

parsers/UY.py
ZachEichen: fixed in [cb4aec7](https://github.com/electricitymaps/electricitymaps-contrib/pull/4840/commits/cb4aec7502c9d09a03148e912e5e29fbe2860044)

parsers/UY.py
ZachEichen: this comparison is more or less verbatim what was done from commit [63bd25e](https://github.com/electricitymaps/electricitymaps-contrib/pull/1334/commits/63bd25e234f491bcf65107c04beef034295eecbc) when this method was introduced.

parsers/UY.py
ZachEichen: This should be fixed in cb4aec7502c9d09a03148e912e5e29fbe2860044

parsers/UY.py
```diff
@@ -1,163 +1,264 @@
...
+
+AVALIABLE_KEYS = ["hydro", "wind", "solar", "biomass", "unknown"]
+
+UTE_URL = url = "https://ute.com.uy/energia-generada-intercambios-demanda"
 
 SALTO_GRANDE_URL = "http://www.cammesa.com/uflujpot.nsf/FlujoW?OpenAgent&Tensiones y Flujos de Potencia&"
 
 
-def get_salto_grande(session: Optional[Session]) -> float:
+def get_salto_grande(session: Session, targ_time: datetime = None) -> float:
```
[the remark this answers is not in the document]
ZachEichen: This should be fixed in cb4aec7502c9d09a03148e912e5e29fbe2860044

parsers/UY.py
VIKTORVAV99: I'd keep it, might be good to have a backup if something happens with the Brazil parser.

parsers/UY.py
VIKTORVAV99: Hmm looking at that presents another problem, do you know if the dates used is forward looking or back looking? i.e. does 10:00 include the production between 10:00 and 11:00 or between 9:00 and 10:00?

Electricity Maps use the first and I suspect this parser use the second so we might have to shift the datetimes around.

In either case the problem with this function in particular is that if the time is say 10:31 salto grande would show the production for 10:00 while the other source might show the production for 9:00. We would need to compare the datetimes in each response to make sure they are reporting production for the same hour so we don't get a mismatch between sources. This should have been done in the previous version of parser as well.

VIKTORVAV99 (commented):
There is still the issue with the possible datetime mismatch between the two different sources that need to be addressed before this can proceed.

parsers/UY.py
```diff
@@ -1,163 +1,313 @@
...
+        # some values can sometimes return -0.1 at night, round up to 0
+        for key, val in datum.items():
+            if key not in ["trade", "demand"]:
+                datum[key] = max(val, 0)
+
+        # ingest date field
+        datefield = hour.find("hora").contents[0]
+        datestr = re.findall("\d\d/\d\d/\d\d\d\d \d+:\d\d", str(datefield))[0]
+        date = arrow.get(datestr, "DD/MM/YYYY h:mm").replace(
+            tzinfo=dateutil.tz.gettz(TIME_ZONE)
```
VIKTORVAV99, line 142:
```suggestion
            tzinfo=tz.gettz(TIME_ZONE)
```

parsers/UY.py
```diff
@@ -1,163 +1,313 @@
...
+            for key, spanish in MAP_GENERATION.items()
+        }
+        # some values can sometimes return -0.1 at night, round up to 0
+        for key, val in datum.items():
+            if key not in ["trade", "demand"]:
+                datum[key] = max(val, 0)
+
+        # ingest date field
+        datefield = hour.find("hora").contents[0]
+        datestr = re.findall("\d\d/\d\d/\d\d\d\d \d+:\d\d", str(datefield))[0]
```
VIKTORVAV99, line 140:
```suggestion
        datestr = re.findall(r"\d\d/\d\d/\d\d\d\d \d+:\d\d", str(datefield))[0]
```

VIKTORVAV99 (commented):
I don't think just shifting the datetimes are safe and we need to make sure we maintain the data integrity between the two sources.

There is also some additional suggestions that fixes some issues pylance is complaining about.

VIKTORVAV99:
Unfortunately we will have to close this PR.
A better source was found that include all generation in one endpoint (including Salto Grande) and it supports historical parsing, this source was implemented in #4925.

Now the contributions in here has not gone unnoticed so I added you to the contributors in the zone config and marked you as co-authors on the commit so you will get the same credit as if this PR was merged.

This was not related to your contributions and I would like to apologize for the whole situation.

Thank you for your understanding and thanks for all the work in here!

PR #4842, opened 2022-12-06
Change validation thresholds for GB parser.

unitrium:
## Issue

At the moment the BMRS API reports 0 for any production mode except for wind.
Our validation threshold makes it so that the data points are dropped.
Therefore the only datapoints are created through a TIME_SLICER_AVERAGE. The last parser datapoint dates back to the 30/11.

## Description

This PR will allow datapoints with only wind generation data to be accepted.
They will be mixed with estimated data points in the flowtracing pipeline and therefore we will have  data for wind.


unitrium:
We should decide on if we want to temporarly report null for modes that are missing at the moment.

unitrium:
One thing to take into consideration, the parsed data might overtake the estimated data and therefore we might start showing a 0 nuclear, gas and coal.

PR #4846, opened 2022-12-08
Italian time period translation/bugfix

tjperr:
## Issue

I noticed the Italian title for energy price wasn't rendering correctly:

Screen Shot 2022-12-08 at 17 21 26

This was due to the `default` locale string incorrectly having the `hourly` key. Fixing that just introduced bad grammar so I've just added all time periods separately (Italian is my second language)

I'm aware you're mid-frontend-refactor so let me know how you want to proceed - Thanks!

### DIFF
--- a/web/public/locales/it.json
+++ b/web/public/locales/it.json
@@ -84,22 +84,46 @@
       "yearly": "Intensità di carbonio negli ultimi 5 anni"
     },
     "emissions": {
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
Thanks for opening this PR! 🎉

I can't say I know Italian very well so I'll trust you on the grammar and such, so for me everything looks good.

Thanks for taking the time to fix it!

Issue #4235, opened 2022-06-11
Missing exchange volume between Åland (AX) and Finland (FI).


VIKTORVAV99:
Currently there is two interconnections between Åland (AX) and Finland (FI), one AC connection with a capacity of 10 MW and a DC connection with a capacity of 100 MW (primarily for backup).

The current parser only appears to read the data from the AC connection and miss a small but almost constant import from the DC interconnection.

All the data we need is available here https://kraftnat.ax/driftcentralen/ but the parser would need to be updated. The new source also contains (text) data for the last 24 hours (in 15 minute increments) vs just real time data so switching data source would make the parser more reliant as well.

As you can clearly see if you compare the old map vs the new map we are missing 0.28 MW (at the time of writing) of imports from Finland, this is roughly 1.1% of Ålands total electricity (imports + production). 

OLD MAP:
<img width="637" alt="image" src="https://user-images.githubusercontent.com/30777521/174145545-97cc13df-32e4-4607-a107-37dbc6504d05.png">

NEW MAP:
<img width="893" alt="image" src="https://user-images.githubusercontent.com/30777521/174145655-07641019-5800-4c44-ba0f-dd7bc0ec8579.png">

PR #4852, opened 2022-12-12
New parser for Åland

VIKTORVAV99:
## Issue
The current parser is only returning one datapoint at a time and don't include low values of export/import at all due to the nature of the source.
Fixes: #4235
## Description

This PR fetches the data from a iFrame at https://kraftnat.ax/driftcentralen/ and then parse that. The  benefit is that we no longer need to use OCR and it fetches the data for the last 24 hours making the whole parser and data more reliable for Åland.

>**Note**
>I am deliberately using a new file name so this will be easier to review, if it get's approved I will rename the file and archive the old parser.


VIKTORVAV99:
I will rename the parser file to `AX.py` once there is an approving review.

parsers/AX_NEW.py
```diff
@@ -0,0 +1,208 @@
...
+                "fossil": float(fossil),
+                "gustavs": float(gustavs),
+                "wind": float(wind),
+                "consumption": float(consumption),
+            }
+        )
+    return data_list
+
+
+def formated_data(
```
unitrium, line 67: This should probably live in the shared lib folder at some point.
It's probably ok for now to have it here, but I feel that all parsers would benefit from sharing such a function.

unitrium (approved):
Looks good!
Just a small thought on the format function that could be shared.

parsers/AX_NEW.py
VIKTORVAV99: This function in particular probably won't work well shared but we could make a shared function that formats data to our different return types (these could also include some basic type checks if we want).

VIKTORVAV99:
Thanks, I'll do the final cleanup and archive the old parser before merging this.

PR #4860, opened 2022-12-14
Add NTESMO parser

unitrium:
## Description

Add a parser for NTESMO to fill AUS-NT + enable aggregation of AUS.


config/zones/AUS.yaml
```diff
@@ -0,0 +1,34 @@
+capacity:
```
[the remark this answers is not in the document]
unitrium: Agree, for now the sub zones will remain the same as they already have data but for this new zone it's fine.

unitrium:
Since there is no data currently for AUS-NT let's take the opportunity to rename it to AU-NT as well.

config/zones/AU.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - AU-NT
+  - AUS-QLD
+  - AUS-SA
+  - AUS-TAS
+  - AUS-TAS-CBI
+  - AUS-TAS-FI
+  - AUS-TAS-KI
+  - AUS-VIC
+  - AUS-WA
+  - AUS-WA-RI
```
VIKTORVAV99, line 38: Could we rename any of these zones as well or do we have data for all of them?
For example `AUS-TAS-CBI`.

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+                        generator_key
+                    ]
+        production_mix.append(data_point)
+    return production_mix
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_consumption(
+    zone_key: str = "AU-NT",
+    session: Optional[Session] = None,
```
VIKTORVAV99, line 179:
```suggestion
    session: Session = Session(),
```
Just like `fetch_production_mix`.

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+    logger: Logger = getLogger(__name__),
+):
+    consumption = get_data(session, target_datetime, extract_demand_price_data, logger)
+    return parse_consumption(consumption, target_datetime, logger)
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_price(
+    zone_key: str = "AU-NT",
+    session: Optional[Session] = None,
```
VIKTORVAV99, line 190:
```suggestion
    session: Session = Session(),
```
Just like `fetch_production_mix`.

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+):
+    production_mix = get_data(session, target_datetime, extract_production_data, logger)
+    return parse_production_mix(production_mix, logger)
+
+
+if __name__ == "__main__":
+    target_datetime = datetime.now() - timedelta(days=2)
```
VIKTORVAV99, line 210: Is this just for testing purposes or is there a 48 hour delay on the data?
If there is a delay on the data we should add the following to the `AU-NT.yaml` file:
```yaml
delays:
  production: 48
  consumption: 48
  # etc
```

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+    except KeyError:
+        raise Exception(
+            f"Cannot find file on the index page for date {target_datetime}"
+        )
+    return extraction_func(data_file)
+
+
+def parse_consumption(
+    raw_consumption: pd.DataFrame,
+    target_datetime: datetime,
```
VIKTORVAV99, line 111:
```suggestion
    target_datetime: Optional[datetime],
```
Since the functions that call these function have optional datetime this means datetime can be  None.

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+    logger: Logger,
+    price: bool = False,
+) -> List[dict]:
+    data_points = []
+    for _, consumption in raw_consumption.iterrows():
+        # Market day starts at 4:30 and reports up until 4:00 the next day.
+        # Therefore timestamps between 0:00 and 4:30 excluded need to have an extra day.
+        raw_timestamp = consumption[0]
+        timestamp = datetime.combine(date=target_datetime.date(), time=raw_timestamp)
```
VIKTORVAV99, line 120: Should probably check so datetime is not None here before accessing it's `.date()` function.

parsers/NTESMO.py
```diff
@@ -0,0 +1,214 @@
...
+    generation_units.remove("Period Start")
+    generation_units.remove("Period End")
+    if not generation_units == PLANT_MAPPING.keys():
+        raise Exception(
+            f"New generator {generation_units - PLANT_MAPPING.keys()} detected in AU-NT, please update the mapping of generators."
+        )
+    for _, production in raw_production_mix.iterrows():
+        data_point = {
+            "zoneKey": "AU-NT",
+            "datetime": production["Period Start"],
```
VIKTORVAV99, line 152: This don't seem to be a native datetime object so the test_parser is throwing an exception on it. So we will have to convert it to a native datetime object first before we return it.

config/zones/AU.yaml
unitrium: The only one that doesn't have data is AUS-TAS-CBI. So I guess I could rename this one as well.

parsers/NTESMO.py
unitrium: Good catch, there's indeed a 24 hours delay for the data to get published.

unitrium:
All done, I also put in the backlog the renaming of remaining AUS zones.

config/zones/AU-NT.yaml
```diff
@@ -0,0 +1,21 @@
...
+  consumption: 30
+  price: 30
+contributors:
+  - unitrium
+parsers:
+  price: NTESMO.fetch_price
+  production: NTESMO.fetch_production_mix
+  consumption: NTESMO.fetch_consumption
+flag_file_name: au.png
+timezone: null
```
VIKTORVAV99, line 21: This should probably not be null, unless it exists outside time and space 😅.

VIKTORVAV99 (approved):
Looks good to me but it looks like there is a missing time zone for AU-NT.

Nice work! 🎉

parsers/NTESMO.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def get_data(
+    session: Session,
+    target_datetime: Optional[datetime],
+    extraction_func: Callable[[bytes], pd.DataFrame],
+    logger: Logger,
+) -> pd.DataFrame:
+    if target_datetime is None:
+        target_datetime = datetime.utcnow()
```
pierresegonne, line 97: When testing with `test_parser.py` the parser raises an exception when the target datetime is not specified. I guess it's because the data is delayed. That's fine, except that when running our feeder, we won't be specifying a datetime so the parser would only work if we do re-fetching.
Why not then set it by default as today - delay (30h)? 

See suggestion above ^

pierresegonne, line 97: I don't know with certainty, but I think we should stop using `utcnow` https://blog.ganssle.io/articles/2019/11/utcnow.html we can use instead `arrow.utc().to("UTC")`

parsers/NTESMO.py
```diff
@@ -0,0 +1,220 @@
...
+
+def extract_demand_price_data(file: bytes) -> pd.DataFrame:
+    return pd.read_excel(
+        file, "System Demand and Market Price", skiprows=4, header=0, usecols="A:C"
+    )
+
+
+def get_data(
+    session: Session,
+    target_datetime: Optional[datetime],
```
pierresegonne, line 92:
```suggestion
    target_datetime: datetime = arrow.now().to("utc").shift(hours=-DELAY),
```

OR something like that

pierresegonne: actually it should probably be done at the `fetch_*` level

parsers/NTESMO.py
```diff
@@ -0,0 +1,220 @@
...
+    generation_units.remove("Period End")
+    if not generation_units == PLANT_MAPPING.keys():
+        raise ParserException(
+            "NTESMO.py",
+            f"New generator {generation_units - PLANT_MAPPING.keys()} detected in AU-NT, please update the mapping of generators.",
+        )
+    for _, production in raw_production_mix.iterrows():
+        data_point = {
+            "zoneKey": "AU-NT",
+            "datetime": production["Period Start"],
```
pierresegonne, line 157:
```suggestion
            "datetime": production["Period Start"].to_pydatetime(),
```

parsers/NTESMO.py
```diff
@@ -0,0 +1,220 @@
...
+                    ]
+        production_mix.append(data_point)
+    return production_mix
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_consumption(
+    zone_key: str = "AU-NT",
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
```
pierresegonne, line 186: So yeah my suggestion for datetimes would be to set them here by default to `now_with_delay` where `now_with_delay = arrow.now().to("utc").shift(days=-1).datetime`

parsers/NTESMO.py
```diff
@@ -0,0 +1,220 @@
...
+    production_mix = []
+    generation_units = set(raw_production_mix.columns)
+    generation_units.remove("Period Start")
+    generation_units.remove("Period End")
+    if not generation_units == PLANT_MAPPING.keys():
+        raise ParserException(
+            "NTESMO.py",
+            f"New generator {generation_units - PLANT_MAPPING.keys()} detected in AU-NT, please update the mapping of generators.",
+        )
+    for _, production in raw_production_mix.iterrows():
```
pierresegonne, line 154:
```suggestion
    raw_production_mix["Period Start"] = raw_production_mix["Period Start"].dt.tz_localize("Australia/Darwin")
    for _, production in raw_production_mix.iterrows():
```

pierresegonne: We need to localise these datetimes. They are most likely given in local time

pierresegonne: Screenshot 2022-12-15 at 10 27 07
That's the gas production when setting the datetimes to be local

pierresegonne: There is 9h30 of difference with UTC so that would put the peak at unreasonable times if it was not local time

pierresegonne:
Quite a lot of comments in the end, let me know if something is not clear

parsers/NTESMO.py
unitrium: Agree, actually it would be better at fetch_ level.

VIKTORVAV99: I would suggest using `datetime.now() - timedelta(hours=30)` but otherwise it makes sense.

parsers/NTESMO.py
unitrium: Yes, I've added that now. Though I think the delay should rather be in hours (30 to be sure) as they might not publish data right away at midnight.

madsnedergaard:
I'm late to the party, but amazing that data is available 🥰

pierresegonne:
https://storage.googleapis.com/electricitymap-parser-logs/AU-NT.html I think we overlooked something when providing the default now with delay value. We have to specify that we're requesting that target datetime in UTC

pierresegonne (commented):
Oh and sorry one of my comments was missing. I'm pretty sure that we must localise the returned timestamp in parse consumption as well

PR #4866, opened 2022-12-15
Prevent edge case in AX parser

VIKTORVAV99:
## Issue
I realized there could be a edge case that would corrupt the data if we fetched just after midnight (local time) before the iFrame have been updated.

This PR adds a small check to prevent that.


parsers/AX.py
```diff
@@ -75,15 +75,22 @@ def formated_data(
...
         second=0,
         microsecond=0,
     )
     return_list = []
     for data in data_list:
-        corrected_date = date - timedelta(minutes=15 * data_list.index(data))
+        index = data_list.index(data)
+        corrected_date = date - timedelta(minutes=15 * index)
+        if corrected_date > (date_time - timedelta(minutes=15 * index)):
+            raise ParserException(
```
q--, line 90: Looking at this code, instead of raising a `ParserExcepion`, wouldn't it be better to correct the date to yesterday? 

Probably something along the lines of 

    date = date_time.replace(
        day=<yesterday>

VIKTORVAV99:
@q-- is this what you had in mind?

q--:
Yes, if I'm not misreading the code

parsers/AX.py
```diff
@@ -75,12 +75,15 @@ def formated_data(
...
     data_list.reverse()
-    date = datetime.now(timezone(TIME_ZONE)).replace(
+    date_time = datetime.now(timezone(TIME_ZONE))
+    date = date_time.replace(
         hour=int(data_list[0]["time"].split(":")[0]),
         minute=int(data_list[0]["time"].split(":")[1]),
         second=0,
         microsecond=0,
     )
+    if date > date_time:
```
unitrium, line 85: I'm wondering about the effect of this. Will this mean that at midnight on a day, the datapoint will be attributed to yesterday ?

VIKTORVAV99: Since we are just getting the hours and minutes back from the iFrame we run the risk of assigning the past values to a future date at midnight (local time) if we fetch data after midnight but before the iFrame has updated.

So for example if we fetch data December 5 00:01 and the iFrame updates at 00:02 we would assign the data for the past hour starting December 5 23:45 and then backwards 24 hours. This is obviously not what we want.

So if the date we get from the parser is grater than the date when we run the parser we shift the  date back one day (which should be the correct date).

unitrium: Ok makes sense then to do this!

unitrium (approved):

PR #4871, opened 2022-12-15
Upgrade PJM production parser

mathilde-daugy:
## Issue

see Linear issue ELE-1281

## Description

In the last data quality report, we noticed that there were many missing production data points for PJM. It seems like the current parser cannot collect data points at 4AM each day and sometimes for multiple hours.
I investigated and found that our production parser is a live parser so there is no way to get data for the missing points. There is an API however and it would be better to use it to get production (we get consumption forecast this way). This will enable us to backfill and get a more consistent dataset over time and potentially capture any data revisions.
This API can also be used to get LMP prices but would require a bit more work.

### Preview
NA


parsers/US_PJM.py
```diff
@@ -4,17 +4,21 @@
 
 import json
```
VIKTORVAV99, line 5:
```suggestion
```
This is unused.

parsers/US_PJM.py
```diff
@@ -4,17 +4,21 @@
...
 import re
-from datetime import datetime
+from datetime import datetime, timedelta
 from logging import Logger, getLogger
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
```
VIKTORVAV99, line 9:
```suggestion
from typing import List, Optional, Union
```
Also unused.

parsers/US_PJM.py
```diff
@@ -48,81 +52,39 @@
...
 
-    return production
+    headers = {
+        "Host": "api.pjm.com",
+        "Ocp-Apim-Subscription-Key": get_token("PJM_TOKEN"),
+        "Origin": "http://dataminer2.pjm.com",
+        "Referer": "http://dataminer2.pjm.com/",
+    }
+    url = API_ENDPOINT + kind
+    resp = get(url, params, headers=headers)
```
VIKTORVAV99, line 79:
```suggestion
    resp: Response = session.get(url, params, headers=headers)
```

parsers/US_PJM.py
```diff
@@ -4,17 +4,21 @@
...
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
 
 import arrow
 import demjson3 as demjson
+import pandas as pd
+import pytz
 from bs4 import BeautifulSoup
 from dateutil import parser, tz
 from requests import Session, get
```
VIKTORVAV99, line 17:
```suggestion
from requests import Session, Response
```
Let's just the `Session.get()` function directly instead of importing get (unless there is a reason?).

parsers/US_PJM.py
```diff
@@ -48,81 +52,39 @@
...
-
-def data_processer(data) -> dict:
-    """Takes a list of dictionaries and extracts generation type and value from each."""
 
-    production = {}
-    for point in data:
-        gen_type = mapping[point["name"]]
-        gen_value = float(point["y"])
-        production[gen_type] = production.get(gen_type, 0.0) + gen_value
+def fetch_api_data(kind: str, params: dict) -> list:
```
VIKTORVAV99, line 70:
```suggestion
def fetch_api_data(kind: str, params: dict, session: Session) -> list:
```

parsers/US_PJM.py
```diff
@@ -145,9 +107,7 @@ def fetch_consumption_forecast_7_days(
     params = {"download": True, "startRow": 1, "forecast_area": "RTO_COMBINED"}
 
     # query API
-    url = API_ENDPOINT + "load_frcstd_7_day"
-    resp = get(url, params, headers=headers)
-    data = json.loads(resp.content)
+    data = fetch_api_data(kind="load_frcstd_7_day", params=params)
```
VIKTORVAV99, line 110:
```suggestion
    data = fetch_api_data(kind="load_frcstd_7_day", params=params, session=session)
```

parsers/US_PJM.py
```diff
@@ -163,28 +123,55 @@ def fetch_consumption_forecast_7_days(
     return data_points
 
 
+@refetch_frequency(timedelta(days=1))
 def fetch_production(
     zone_key: str = "US-PJM",
     session: Optional[Session] = None,
```
VIKTORVAV99, line 129:
```suggestion
    session: Session = Session(),
```

parsers/US_PJM.py
```diff
@@ -163,28 +123,55 @@ def fetch_consumption_forecast_7_days(
...
-        raise NotImplementedError("This parser is not yet able to parse past dates")
-
-    extracted = extract_data(session=None)
-    production = data_processer(extracted[0])
+) -> list:
+    """uses PJM API to get generation  by fuel. we assume that storage is battery storage (see https://learn.pjm.com/energy-innovations/energy-storage)"""
+    if target_datetime is None:
+        target_datetime = arrow.utcnow().datetime
+    if not session:
+        session = Session()
```
VIKTORVAV99, line 137:
```suggestion
```
With the above change this can be removed.

parsers/US_PJM.py
```diff
@@ -163,28 +123,55 @@ def fetch_consumption_forecast_7_days(
...
+    all_data_points = []
+    for dt in set(data.datetime_beginning_ept):
+        production = {}
+        storage = {}
+        data_dt = data.loc[data.datetime_beginning_ept == dt]
+        for mode in set(data.fuel_type):
+            if mode == "battery":
+                storage["battery"] = data_dt.loc[
+                    data_dt.fuel_type == mode, "mw"
+                ].values[0]
```
VIKTORVAV99, line 159: Not sure why but accessing `values` like this gives all kinds of type errors.

VIKTORVAV99 (commented):
Added some comments on the new parts of the parser but I would like to take the oppertunity to clean up the old parts of the parser as well if you don't mind. I can get this done on friday.

VIKTORVAV99:
@mathilde-daugy I cleaned up the usage of session a bit (I hope that's okay) but feel free to revert it if needed.

mathilde-daugy:
> @mathilde-daugy I cleaned up the usage of session a bit (I hope that's okay) but feel free to revert it if needed.

Thanks Viktor :) 
I'll take a look on Monday, we're going to make a few adjustments as well before deploying

unitrium (approved):
Small code feedbacks but otherwise looks great!

PR #4874, opened 2022-12-15
Add fetch price to FR_O regions

meiyiniu:
## Issue
https://github.com/electricitymaps/electricitymaps-contrib/issues/4747

## Description

Fetches prices for the FR_O zones, which are available only for historical data. The price is no longer reported since the beginning of 2020, but historical data can still be meaningful to know.

Testing: poetry run test_parser RE price --target_datetime 2016-01-01

### DIFF
--- a/parsers/FR_O.py
+++ b/parsers/FR_O.py
@@ -72,6 +72,9 @@ API_PARAMETER_GROUPS = {
         "unknown": ["bagasse_charbon_mwh", "charbon_bagasse_mw"],
     },
     "storage": {},
+    "price": {
+        "price": ["cout_moyen_de_production_eur_mwh"],
+    },
 }
 
 PRODUCTION_MAPPING = {
@@ -88,6 +91,12 @@ STORAGE_MAPPING = {
     for API_TYPE in groups
 }
 
+PRICE_MAPPING = {
+    API_TYPE: type
+    for key in ["price"]
+    for type, groups in API_PARAMETER_GROUPS[key].items()
+    for API_TYPE in groups
+}
 
 def generate_url(zone_key, target_datetime):
     return f"{DOMAIN_MAPPING[zone_key]}/api/v2/catalog/datasets/{HISTORICAL_DATASETS[zone_key] if target_datetime else LIVE_DATASETS[zone_key]}/exports/json"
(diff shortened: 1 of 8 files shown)


VIKTORVAV99:
@meiyiniu could you fix the formatting before I review this?

You should be able to do it with `poetry run format`.

meiyiniu:
@VIKTORVAV99 done! please take a look now :)

parsers/FR_O.py
```diff
@@ -203,3 +213,32 @@ def fetch_production(
...
     return return_list
+
+
+def fetch_price(
+    zone_key: str,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger=getLogger(__name__),
+):
+    production_objects, date_string = fetch_data(
```
VIKTORVAV99, line 224: I would suggest renaming `production_objects` to something more destinctive so it's not confused with the production modes.

parsers/FR_O.py
```diff
@@ -203,3 +213,32 @@ def fetch_production(
...
+        zone_key, session, target_datetime, logger
+    )
+
+    return_list: List[Dict[str, Any]] = []
+    for production_object in production_objects:
+        price: Union[float, int] = 0
+        for mode_key in production_object:
+            if mode_key in PRICE_MAPPING:
+                price += production_object[mode_key]
```
VIKTORVAV99, line 233: This code makes the price 0 even if any price data is missing. If the price data is missing it should return `None` and not `0`.

VIKTORVAV99 (changes requested):
Most of it looks good but if a value is missing it should return `None` and not be set to a arbitrary value (`0` in this case).

meiyiniu:
@VIKTORVAV99 Thank you for the review! Addressed your comments - can you please give it a re-review?

parsers/FR_O.py
```diff
@@ -203,3 +213,33 @@ def fetch_production(
...
+
+    return_list: List[Dict[str, Any]] = []
+    for data_object in data_objects:
+        price: Union[float, int, None] = None
+        for mode_key in data_object:
+            if mode_key in PRICE_MAPPING:
+                if price == None:
+                    price = data_object[mode_key]
+                else:
+                    price += data_object[mode_key]
```
VIKTORVAV99, line 234:
```suggestion
                price = data_object[mode_key]
```
There should only be one instance of price data per datapoint if I remember correctly so we can just directly assign the data without any issue.

VIKTORVAV99 (commented):
Everything looks good, I just have one small comment to save us a few lines of code.

Nice work!

meiyiniu:
@VIKTORVAV99 Please take another look! Thank you :)

VIKTORVAV99:
> @VIKTORVAV99 Please take another look! Thank you :)

Will do so tomorrow. 🙂

VIKTORVAV99 (approved):
LGTM! 🎉

PR #4878, opened 2022-12-19
Add a temporary hardcoded fix for AUS exchanges

unitrium:
## Issue

AU internal exchanges are not hidden when using the aggregated country view. This is due to the fact that two namings are coexisting for austalian zones at the moment: AU and AUS.

## Description

Hardcode generate-exchanges to exclude to consider AUS exchanges until we completly rename AUS zones.


tonypls (commented):
I wonder if it's better to have the combined zone as AUS for now and have it all part of the migration to AU when that happens? Feels a bit like this leaves us in a halfway state with potential for an edge case to be forgotten about.

Open to discuss :)

pierresegonne (approved):

pierresegonne:
> I wonder if it's better to have the combined zone as AUS for now and have it all part of the migration to AU when that happens? Feels a bit like this leaves us in a halfway state with potential for an edge case to be forgotten about.


We're planning to make the final transition to AU-* in the first week of January, as we don't want to introduce potentially breaking changes now

Issue #4667, opened 2022-10-16
Selected language is not known to user.


Ankitchaudharyy:
**Describe the bug**
When Select language menu is opened, selected language like Italiano (which occurs when scroll) is not known, as it is in the drop down below.

**Expected behavior**
Selected language can come to top when selected, and rest other in the same order can come below it.
It can go to its respective position when deselected.

**Screenshots**
Currently, Italiano is selected here, and it is not known from the menu.
<img width="614" alt="Screenshot 2022-10-16 at 11 04 35 PM" src="https://user-images.githubusercontent.com/46367738/196049958-14f92b42-000c-404e-b656-c3778d2df8c9.png">

Ankitchaudharyy:
It can be this way:
<img width="568" alt="Screenshot 2022-10-16 at 11 18 18 PM" src="https://user-images.githubusercontent.com/46367738/196050316-b95d7185-419f-4be9-9f25-12eb5ffa801c.png">

Ankitchaudharyy:
I have submitted a PR #4879 using CSS only.

PR #4879, opened 2022-12-19
Add. preferred language to the top of language menu.

Ankitchaudharyy:
## Issue
#4667
Selected language is not shown at the top.

## Description
Screenshot 2022-12-19 at 4 40 21 PM
This is acheived by changing the CSS of the `preferred-language` and `other-language` class.

### DIFF
--- a/web/src/components/languageselect.jsx
+++ b/web/src/components/languageselect.jsx
@@ -43,6 +43,12 @@ const LanguageSelectContainer = styled.ul`
     }
     &.preferred-language {
       background-color: rgba(0, 0, 0, 0.1);
+      position: absolute;
+      top: 0;
+    }
+    &.other-language {
+      position: relative;
+      top: 30px;
     }
   }
 `;
@@ -76,7 +82,7 @@ const LanguageSelect = () => {
             <li key={key}>
               <button
                 onClick={() => handleLanguageSelect(key, language)}
-                className={selectedLanguage === language ? 'preferred-language' : null}
+                className={selectedLanguage === language ? 'preferred-language' : 'other-language'}
               >
                 {language}
               </button>


web/src/components/languageselect.jsx
```diff
@@ -43,6 +43,12 @@ const LanguageSelectContainer = styled.ul`
     }
     &.preferred-language {
       background-color: rgba(0, 0, 0, 0.1);
+      position: absolute;
+      top: 0;
+    }
+    &.other-language {
+      position: relative;
+      top: 30px;
```
VIKTORVAV99, line 51: For some reason this produce a small gap, my local testing found that 27px would remove the gap while still not overlapping.

VIKTORVAV99 (commented):
Thanks for the PR. This seems to work fine when in desktop mode but don't work at all when scaled to mobile. Could you look into that?

web/src/components/languageselect.jsx
Ankitchaudharyy: You are right.
27px removes the unnecessary gap.

Ankitchaudharyy:
> Thanks for the PR. This seems to work fine when in desktop mode but don't work at all when scaled to mobile. Could you look into that?

I have updated the language menu for mobile mode. 

Reason I shifted the `StyledLanguageSelectWrapper` component inside the`SettingsModal` is due to new state `selectedLanguage`. Lot of states and setState methods are being passed as arguements, otherwise.
Also with earlier approach. `i18n` also needs to be initialised again.

> const LanguageSelect = ({ isOpen, closeLanguageSelect, selectedLanguage, setSelectedLanguage }) => {
  if (!isOpen) {
    return null;
  }
  const { __, i18n } = useTranslation();
  const handleLanguageChange = (languageKey, preferredLanguage) => {
    i18n.changeLanguage(languageKey);
    setSelectedLanguage(preferredLanguage);
    closeLanguageSelect(false);
  };
  return (
    <StyledLanguageSelectWrapper>
      {Object.entries(LANGUAGE_NAMES).map(([key, language]) => (
        <li key={key}>
          <button 
            onClick={() => handleLanguageChange(key, language)}
            className={selectedLanguage === language ? 'preferred-language' : 'other-language'}
          >
            {language}
          </button>
        </li>
      ))}
    </StyledLanguageSelectWrapper>
  );
};

web/src/components/settingsmodal.jsx
```diff
@@ -76,25 +87,10 @@ const SettingButton = styled(Button).attrs({
...
-        </li>
-      ))}
-    </StyledLanguageSelectWrapper>
-  );
-};
-
 const SettingsView = () => {
   const { __, i18n } = useTranslation();
   const [languageSelectOpen, setLanguageSelectOpen] = useState(false);
+  const [selectedLanguage, setSelectedLanguage] = useState('English');
```
VIKTORVAV99, line 93: We can't use english as the defualt here becuase then it will always be set to english before anyone changes the language (regardless of the language shown).

We need to get the language from i18n or local storage.

VIKTORVAV99 (changes requested):
Things looks mostly good now as it works for both mobile and desktop but the default language is always english when it should be the language shown the the user at the time.

Ankitchaudharyy:
@VIKTORVAV99 
Added the default language from i18.

This also resolves issue #4623, for which PR #4624 was made.

VIKTORVAV99 (approved):
LGTM! 🎉

Ankitchaudharyy:
Hi @VIKTORVAV99, these changes works fine in my local development, but lacks in the production site.
I am not sure, why is it so.

Can you help?

VIKTORVAV99:
A new build has to be released before any changes show up in the web app (mobile apps also need a new update).

Honestly I'm not sure about what the plan is right now for deployment as we are also planning on deploying the rewrite very soon but I'll see if we can get a new version out before then.

PR #4887, opened 2022-12-20
Add estimation method to Brazil

unitrium:
## Issue

Aggregated zones need to have an estimation method attached otherwise, the app-backend will not return results for the zone.

## Description
Add estimation method to the aggregated Brazil

### DIFF
--- a/config/zones/BR.yaml
+++ b/config/zones/BR.yaml
@@ -15,3 +15,4 @@ subZoneNames:
   - BR-N
   - BR-NE
   - BR-S
+estimation_method: RECONSTRUCT_BREAKDOWN


unitrium:
@VIKTORVAV99 I forgot this needs to be enabled for the backend to properly process the aggregated zone.

VIKTORVAV99 (approved):

VIKTORVAV99:
> @VIKTORVAV99 I forgot this needs to be enabled for the backend to properly process the aggregated zone.

Ah right, every zone needs either a parser or estimation method.

@madsnedergaard this is something we should look at after the holidays, maybe we can exclude this requirement for aggregated zones as they only be colored if the subzones have data anyhow?

madsnedergaard:
> > @VIKTORVAV99 I forgot this needs to be enabled for the backend to properly process the aggregated zone.
> 
> Ah right, every zone needs either a parser or estimation method.
> 
> @madsnedergaard this is something we should look at after the holidays, maybe we can exclude this requirement for aggregated zones as they only be colored if the subzones have data anyhow?

Yes, that would make good sense - I'll add it to the backlog :)

PR #4890, opened 2022-12-20
Create MK ENTSOE wrapper

VIKTORVAV99:
## Issue
The data for North Macedonia have very intermittent publishing times so sometimes we miss production due to the fact it has not been published yet (but so far it always have been, up to 3 days behind).

So this wrapper wraps the ENTSO-E parser and modify the `target_datetime` to be 24 hours in the past when run without a `target_datetime` passed to it.

This should give us a 48+24 hour window to catch the "live" data so we don't have to manually re-run it. (This can be bumped if needed)

### DIFF
--- a/parsers/MK_wrapper.py
+++ b/parsers/MK_wrapper.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 from logging import Logger, getLogger
 from typing import List, Optional, Union
 
+from pytz import utc
 from requests import Session
 
 from . import ENTSOE
@@ -16,7 +17,7 @@ If no target_datetime is provided, the wrapper will provide a target datetime th
 
 def modify_target_datetime(target_datetime: Optional[datetime]) -> datetime:
     if target_datetime is None:
-        return datetime.utcnow() - timedelta(hours=24)
+        return datetime.now(tz=utc) - timedelta(hours=24)
     else:
         return target_datetime
 
--- a/parsers/MK_wrapper.py
+++ b/parsers/MK_wrapper.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
-from typing import List, Optional, Union
+from typing import Optional
 
 from pytz import utc
 from requests import Session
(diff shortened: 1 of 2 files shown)


parsers/MK_wrapper.py
```diff
@@ -0,0 +1,117 @@
+from datetime import datetime, timedelta
+from logging import Logger, getLogger
+from typing import List, Optional, Union
+
+from requests import Session
+
```
pierresegonne, line 6: we'll have to import a timezone module (pytz or datetime.timezone)

parsers/MK_wrapper.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+This wrapper is used to fetch data from ENTSOE for the country of North Macedonia.
+If no target_datetime is provided, the wrapper will provide a target datetime that is
+24 hours in the past due to intermittent data availability.
+"""
+
+
+def modify_target_datetime(target_datetime: Optional[datetime]) -> datetime:
+    if target_datetime is None:
+        return datetime.utcnow() - timedelta(hours=24)
```
pierresegonne, line 19: https://blog.ganssle.io/articles/2019/11/utcnow.html 
```suggestion
        return datetime.now(tz=utc) - timedelta(hours=24)
```

the utc timezone mus either come from datetime.timezone or pytz.utc

pierresegonne (approved):
Nice! Just a little nitpick around the use of `utcnow`

pierresegonne (changes requested):
Did not mean to approve huhu

parsers/MK_wrapper.py
VIKTORVAV99: Maybe we can update to Python 3.9+ soon and use https://docs.python.org/3.11/library/zoneinfo.html#module-zoneinfo

I'll update this in the meantime tho, but I assumed that utcnow() simply returned now() with the UTC datetime but it appears it don't... 😅

VIKTORVAV99:
> Did not mean to approve huhu

Updated but the returned object looks exactly the same 😅

parsers/MK_wrapper.py
pierresegonne: yeah it's really weird ahah

pierresegonne (approved):

PR #4891, opened 2022-12-20
Fix datetime query in Production mix for ELEXON parser

unitrium:
## Issue
The current datetime was not taken into account when fetching data.

## Description
Correctly create the time window to include the current day.

# Preview

image

### DIFF
--- a/parsers/ELEXON.py
+++ b/parsers/ELEXON.py
@@ -120,7 +120,9 @@ def query_production(
     }
     if report == "FUELINST":
         params = {
-            "FromDateTime": (target_datetime).date().strftime("%Y-%m-%d %H:%M:%S"),
+            "FromDateTime": (target_datetime - timedelta(days=1))
+            .date()
+            .strftime("%Y-%m-%d %H:%M:%S"),
             "ToDateTime": (target_datetime + timedelta(days=1))
             .date()
             .strftime("%Y-%m-%d %H:%M:%S"),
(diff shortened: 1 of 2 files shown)


parsers/ELEXON.py
```diff
@@ -120,10 +120,10 @@ def query_production(
     }
     if report == "FUELINST":
         params = {
-            "FromDateTime": (settlement_date - timedelta(hours=24)).strftime(
-                "%Y-%m-%d %H:%M:%S"
-            ),
-            "ToDateTime": settlement_date.strftime("%Y-%m-%d %H:%M:%S"),
+            "FromDateTime": (target_datetime).date().strftime("%Y-%m-%d %H:%M:%S"),
+            "ToDateTime": (target_datetime + timedelta(days=1))
+            .date()
```
VIKTORVAV99, line 125: Shouldn't this be the opposite?
I.e. from datetime should be -1 day and to datetime should be the current datetime?

unitrium: The source allows to put a datetime in the future.
So if we are the 2022-12-20 at 14:00 and we run this, 
FromDateTime would be 2022-12-20 00:00 as we are taking only the date part
ToDateTime would be 2022-12-21 00:00
And therefore we can encompass our entire day.

VIKTORVAV99: Ah that makes sense, we could still do -1 day though to get a 2 day re-fetch frequency like the ENTSOE parser. But  it's not required.

mathilde-daugy (approved):

parsers/ELEXON.py
unitrium: Actually that would be nice to do, it would cover some potential gaps.

PR #4893, opened 2022-12-20
Add Argentina exchanges

meiyiniu:
## Issue
https://github.com/electricitymaps/electricitymaps-contrib/issues/3699

## Description

Adds exchanges between Argentina and neighbouring countries (Chile, Uruguay, Paraguay). Exchange data to Brazil exist, but the app is currently using the Brazil data source for that exchange, so we left it out of this PR.


VIKTORVAV99 (changes requested):
This currently fails when running it locally due to time zone issues so until that is resolved I can't fully review this.

Nice to see someone taking a crack at fixing the issue though, so thanks for that!

parsers/AR.py
```diff
@@ -143,8 +158,53 @@ def fetch_exchange(
     logger: Logger = getLogger(__name__),
 ) -> dict:
     """Requests the last known power exchange (in MW) between two zones."""
+    if target_datetime:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    sorted_zone_keys = sorted([zone_key1, zone_key2])
+    sorted_codes = "->".join(sorted_zone_keys)
+    flow: Union[float, None] = None
+    target_datetime = arrow.now(tz="America/Argentina/Buenos Aires").format(
```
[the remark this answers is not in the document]
elisaluan: Added an underscore (`America/Argentina/Buenos_Aires`) to be consistent with the [tz database timezones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

Please let me know if this works for you locally, thanks!

parsers/AR.py
```diff
@@ -18,8 +18,23 @@
...
+CAMMESA_EXCHANGE_ENDPOINT = (
+    "https://api.cammesa.com/demanda-svc/demanda/IntercambioCorredoresGeo/"
+)
 CAMMESA_RENEWABLES_ENDPOINT = "https://cdsrenovables.cammesa.com/exhisto/RenovablesService/GetChartTotalTRDataSource/"
 
+SUPPORTED_EXCHANGES = {
+    "AR->CL-SEN": "ARG-CHI",
+    "AR->PY": "ARG-PAR",
+    "AR->UY": "ARG-URU",
+}
```
VIKTORVAV99, line 30: I noticed the endpoint support a lot more exchanges than just these. Could you add those as well?
They will be needed in #4894.

parsers/AR.py
```diff
@@ -143,8 +158,50 @@ def fetch_exchange(
...
+    if sorted_codes in SUPPORTED_EXCHANGES:
+        current_session = session or Session()
+
+        api_cammesa_response = current_session.get(CAMMESA_EXCHANGE_ENDPOINT)
+        assert api_cammesa_response.status_code == 200, (
+            "Exception when fetching exchange for "
+            "{}: error when calling url={}".format(
+                sorted_codes, CAMMESA_EXCHANGE_ENDPOINT
+            )
+        )
```
VIKTORVAV99, line 178: This should be using `f"string {object}"` instead both for readability and preformance reasons.

parsers/AR.py
```diff
@@ -143,8 +158,50 @@ def fetch_exchange(
...
+        exchange_list = api_cammesa_response.json()
+        for exchange in exchange_list["features"]:
+            properties = exchange["properties"]
+            if properties["nombre"] == exchange_name:
+                angle_config = EXCHANGE_DIRECTIONS[sorted_codes]
+                given_angle = int(properties["url"][6:])
+                flow = int(properties["text"])
+                if angle_config != given_angle:
+                    flow = -flow
+                target_datetime = (
```
VIKTORVAV99, line 190: The target datetime is not the same as the returned datetime so it should probably not be reassigned, just adding a `returned_datetime` variable and assinging the date to that would probably be the best way forward.

VIKTORVAV99, line 190: I would use `fromisoformat()` here instead of doing custom stuff.
It can be imported from the datetime library.

parsers/AR.py
```diff
@@ -143,8 +158,50 @@ def fetch_exchange(
     logger: Logger = getLogger(__name__),
 ) -> dict:
     """Requests the last known power exchange (in MW) between two zones."""
+    if target_datetime:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    sorted_zone_keys = sorted([zone_key1, zone_key2])
+    sorted_codes = "->".join(sorted_zone_keys)
+    flow: Union[float, None] = None
+    target_datetime = arrow.now(tz="America/Argentina/Buenos_Aires").datetime
```
VIKTORVAV99, line 167: This is actally never used an therefore not needed.

parsers/AR.py
```diff
@@ -143,8 +158,50 @@ def fetch_exchange(
...
+                target_datetime = (
+                    properties["fecha"][:-2] + ":" + properties["fecha"][-2:]
+                )
+                break
+    else:
+        raise NotImplementedError("This exchange is not currently implemented")
+
+    exchange = {
+        "sortedZoneKeys": sorted_codes,
+        "datetime": arrow.get(target_datetime).datetime,
```
VIKTORVAV99, line 199: With the above changes you can just map `returned_datetime` to `"datetime"` here instead.

VIKTORVAV99 (commented):
Looks good and everything is working!

I do have some requests and suggestions for improvements though.

parsers/AR.py
meiyiniu: I added this in the failure case where the exchange isn't returned in the API response (i.e. `if properties["nombre"] == exchange_name` never evaluates to true), in which case the flow returned would be `None`.

VIKTORVAV99: If the exchange is not returned we should raise a `ParserException` instead of returning None.
I was mostly referring to assigning a date to `target_datetime` though that is never used later (and not needed as we get the returned datetime from the API).

parsers/AR.py
meiyiniu: I've added them to these dicts here. Should I add the exchange config files in this PR or the #4894?

VIKTORVAV99: I would add the configs in the other PR as adding it in here would produce a lot of errors if when this is merged with exchanges that don't have the corresponding zones.

VIKTORVAV99:
I won't be able to review this until Monday as I will be celebrating Christmas but if you have any questions just ask and I'll reply when I have the time.

parsers/AR.py
```diff
@@ -18,8 +20,51 @@
...
+    "AR-CEN->AR-CUY": 45,
+    "AR-LIT->AR-NEA": 225,
+    "AR-BAS->AR-LIT": 270,
+    "AR-CUY->AR-NOA": 45,
+    "AR-LIT->AR-CEN": 45,
+    "AR-LIT->AR-NOA": 135,
+    "AR-BAS->AR-CEN": 135,
+    "AR-COM->AR-PAT": 90,
+    "AR-NEA->AR-NOA": 180,
+}
```
VIKTORVAV99, line 66:
```suggestion
EXCHANGE_DIRECTIONS = {  # directions are from second region -> first region
    "AR->CL-SEN": 0,
    "AR->PY": 225,
    "AR->UY": 180,
    "AR-NEA->BR-S": 225,
    "AR-BAS->AR-COM": 45,
    "AR-CEN->AR-COM": 90,
    "AR-CEN->AR-NOA": 315,
    "AR-CUY->AR-COM": 135,
    "AR-CEN->AR-CUY": 45,
    "AR-LIT->AR-NEA": 225,
    "AR-BAS->AR-LIT": 270,
    "AR-CUY->AR-NOA": 45,
    "AR-LIT->AR-CEN": 45,
    "AR-LIT->AR-NOA": 135,
    "AR-BAS->AR-CEN": 135,
    "AR-COM->AR-PAT": 90,
    "AR-NEA->AR-NOA": 180,
}
```
Small spelling fix.

parsers/AR.py
```diff
@@ -143,8 +188,52 @@ def fetch_exchange(
     logger: Logger = getLogger(__name__),
 ) -> dict:
     """Requests the last known power exchange (in MW) between two zones."""
+    if target_datetime:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    sorted_zone_keys = sorted([zone_key1, zone_key2])
+    sorted_codes = "->".join(sorted_zone_keys)
+    flow: Union[float, None] = None
```
VIKTORVAV99, line 196:
```suggestion
    sorted_zone_keys = sorted([zone_key1, zone_key2])
    sorted_codes = "->".join(sorted_zone_keys)
    flow: Optional[float] = None
    returned_datetime: datetime
```
Let's use optional instead + adding returned_datetime here so it's never unbound later.

VIKTORVAV99 (changes requested):
Things look good, I just have two tiny suggestions then this can get merged.

parsers/AR.py
```diff
@@ -3,11 +3,13 @@
 
 from datetime import datetime
 from logging import Logger, getLogger
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
```
VIKTORVAV99, line 6:
```suggestion
from typing import Dict, List, Optional
```
Union is unused.

VIKTORVAV99 (approved):
Just have one tiny cleanup comment but things are looking good to me. I'll go ahead and commit said cleanup and then merge this!

Nice work!

PR #4894, opened 2022-12-21
Add production for AR regions

meiyiniu:
## Issue

None

## Description

Adds regional production data for Argentina (real-time renewable and non-renewable), pulling from the Cammesa data source which is currently used for Argentina national data.

### Preview

image


VIKTORVAV99:
> Thanks for the great work, looking forward to see this on the map! TLDR: try to have only changes related to your feature on this PR, here there are also some stuff related to FR_O. Running `git rebase master` should do the trick.
> 
> Long version: Please just make sure to rebase your branch on the latest master branch to avoid taking commits from your other branches. The reasoning is that we should have more or less 1 PR = 1 branch = 1 feature. Otherwise it gets a bit mixed up when reviewing. I see here that we have modifications related to FR_O which is not the same feature as the Argentinian regions :) Your changes for FR_O will be merged in with #4501 .

Agree with everything in here but I just wanted to add since you are using a fork it is probably better to use `upstream/master` (`git rebase upstream/master`) as the rebase target to avoid rebasing on a outdated master branch.

meiyiniu:
@unitrium @VIKTORVAV99 Thanks for letting me know! This is our first time contributing to open-source so we're still trying to gain familiarity with workflows - your advice is much appreciated :) I've rebased the branch onto `upstream/master` - marking this PR as ready to review now!

config/zones/AR.yaml
```diff
@@ -162,3 +162,12 @@ fallbackZoneMixes:
 parsers:
   production: AR.fetch_production
 timezone: America/Argentina/Buenos_Aires
+subZoneNames:
```
unitrium, line 165: I think we will not compute Argentina as an aggregated zone. As we have direct data for it, it's probably better to not have it as an aggregated zones. Though, this will require some changes on our backend to interpret it as a country view.

parsers/AR.py
```diff
@@ -37,9 +55,15 @@ def fetch_production(
...
-    renewables_production: Dict[str, dict] = renewables_production_mix(
-        zone_key, current_session
-    )
+
+    if zone_key == "AR":
+        renewables_production: Dict[str, dict] = renewables_production_mix(
+            zone_key, current_session
+        )
+    else:
+        renewables_production: Dict[str, dict] = regional_renewables_production_mix(
```
unitrium, line 64:
```suggestion
        renewables_production = regional_renewables_production_mix(
```
Since you added a typed return the type can be inferred from the function, so there's no need to put the type again.

parsers/AR.py
```diff
@@ -37,9 +55,15 @@ def fetch_production(
     non_renewables_production: Dict[str, dict] = non_renewables_production_mix(
         zone_key, current_session
     )
-    renewables_production: Dict[str, dict] = renewables_production_mix(
-        zone_key, current_session
-    )
+
+    if zone_key == "AR":
+        renewables_production: Dict[str, dict] = renewables_production_mix(
```
unitrium, line 60:
```suggestion
        renewables_production = renewables_production_mix(
```
Same here.

parsers/AR.py
```diff
@@ -37,9 +55,15 @@ def fetch_production(
     non_renewables_production: Dict[str, dict] = non_renewables_production_mix(
         zone_key, current_session
     )
-    renewables_production: Dict[str, dict] = renewables_production_mix(
-        zone_key, current_session
-    )
+
+    if zone_key == "AR":
```
unitrium, line 59: I'm thinking that it would be better to just have one `renewables_production_mix` function. The logic for switching to national or regional view could live in there.

parsers/AR.py
```diff
@@ -101,14 +125,50 @@ def renewables_production_mix(zone_key: str, session: Session) -> Dict[str, dict
         }
         for production_info in sorted_production_list
     }
+    return renewables_production
+
+
+def regional_renewables_production_mix(
```
unitrium, line 131: So looking at the function, most parts could be integrated in the `renewables_production_mix` function, therefore limiting the amount of similar code.

unitrium (changes requested):
It looks promising! A few comments about some improvements that could be done on the code.
Also, we have a few exchanges for Argentina, you will need to create the regional exchanges in the config as well.
So for instance we have AR_BR in config/exchanges, you will need to find out in which region on the argentinian side this exchange lives and create the corresponding exchange config. It's probably something like AR-CEN_BR-S. Also note that Brazil is an aggregated country so make sure to do it on the regional level there.

unitrium:
As I said, we will probably want to keep the AR zone without it becoming an aggregation as we have direct data for it. I will discuss it with the team to decide on how we want to tackle this.
Probably in your config file it will be something like this to add
```
AR.yaml

contryView: true
```

VIKTORVAV99:
> As I said, we will probably want to keep the AR zone without it becoming an aggregation as we have direct data for it. I will discuss it with the team to decide on how we want to tackle this. Probably in your config file it will be something like this to add
> 
> ```
> AR.yaml
> 
> contryView: true
> ```

We could probably just check if the country level zones have any parsers instead of adding additional config keys?

That way we could also use the parser information to infer what parts should be aggregated or not.
Example country level:
```yaml
parser:
    production: "example parser"
```
Here we have a country level config with a production parser so we would only need to aggregate say consumption and price from the subzones (assuming they have it).

unitrium:
True that would be a better way of being more granular. I'm just afraid it might add some hidden mechanics.

> > As I said, we will probably want to keep the AR zone without it becoming an aggregation as we have direct data for it. I will discuss it with the team to decide on how we want to tackle this. Probably in your config file it will be something like this to add
> > ```
> > AR.yaml
> > 
> > contryView: true
> > ```
> 
> We could probably just check if the country level zones have any parsers instead of adding additional config keys?
> 
> That way we could also use the parser information to infer what parts should be aggregated or not. Example country level:
> 
> ```yaml
> parser:
>     production: "example parser"
> ```
> 
> Here we have a country level config with a production parser so we would only need to aggregate say consumption and price from the subzones (assuming they have it).

True that would be a better way of having a more granular control, I'm just afraid it would result in some hidden mechanics people are not aware of.

config/zones/AR.yaml
unitrium: Agree, that would be some changes happening in the backend, I will add a note for us there.

parsers/AR.py
meiyiniu: I combined the two into one function - please take a look again and let me know if it's acceptable!

parsers/AR.py
```diff
@@ -131,10 +165,26 @@ def non_renewables_production_mix(zone_key: str, session: Session) -> Dict[str,
...
 
+def get_region_id(zone_key, session: Session) -> int:
+    """Fetches the region id for the zone that is required to get the production of that zone."""
+    regions_response = session.get(CAMMESA_REGIONS_ENDPOINT)
+
+    assert (
+        regions_response.status_code == 200
+    ), "Exception when fetching regions for AR: " "error when calling url={}".format(
+        CAMMESA_REGIONS_ENDPOINT
+    )
```
VIKTORVAV99, line 179: I would prefer if this used `f"string {object}"` instead, it's a little easier to work with and it's sligltly faster from a preformance point of view.

VIKTORVAV99 (commented):
The code changes looks good but I do have some questions about the parser itself.
First it seems like the regional zones only return one data point at once? Would it be possible to return a list like for the AR zone?

It also seems to be some availability issues? I tried ran though the zones and all of the sudden it stopped working? This would probably be resolved if it's possible to return it as a list as well.

parsers/AR.py
```diff
@@ -20,26 +20,40 @@
...
-        zone_key, current_session
-    )
-    renewables_production: Dict[str, dict] = renewables_production_mix(
-        zone_key, current_session
-    )
+    non_renewables_production = non_renewables_production_mix(zone_key, current_session)
+    renewables_production = renewables_production_mix(zone_key, current_session)
 
     full_production_list = [
         {
```
unitrium, line 59: I'm just thinking that here we will miss datapoints which have only non_renewables.
```
renewables = {
'2022-12-22 01:00': {},
'2022-12-22 02:00': {}
}

non_renewables = {
'2022-12-22 01:00': {},
'2022-12-22 03:00': {}
}
Here the 03:00 timestamp would be ignored for non_renewable.
```
Maybe it would be better to rewrite this a bit to merge them if they exist with the same timestamp

unitrium (commented):
Looks better! I haven't been able to test it locally yet though.
If you are up for it, it would be great if you could write a unit test to demonstrate how it works. For this you could just add some mock data from the API and put it in the tests/mock folder. You can have a look at test/test_NTESMO.py for inspiration.

Otherwise just missing the exchange config and we're almost done!

parsers/AR.py
VIKTORVAV99: I would probably add unique keys for hydro (since it exits in both) and then just merge the dicts with the datetime as the key, that would ensure we get all data in one place grouped by the correct datetime and then format the data in a separate function.

meiyiniu:
> First it seems like the regional zones only return one data point at once? Would it be possible to return a list like for the AR zone?
> It also seems to be some availability issues? I tried ran though the zones and all of the sudden it stopped working? This would probably be resolved if it's possible to return it as a list as well.

@VIKTORVAV99 Both of these stem from the same reason, which is that the renewables endpoint for regional zones only returns live data, i.e. data for the current time, which is only 1 data point. The non-renewables endpoint, which AR uses, returns data in 5-minute intervals over the entire day. I tried looking for renewables data across the entire day like the non-renewables one, but it does not seem to be available. 

To merge the data from the two endpoints, I'm assigning the datetime for the renewables live data to the last 5min mark (e.g. 12:09 -> 12:05), and checking the non-renewables data for the datetime. So, to answer your first question, that's why there's only one data point - we only have both renewables and non-renewable data for the last 5min mark. To answer your second question, sometimes the non-renewables data takes a few minutes to update, so the last 5min mark might not yet be available (e.g. if it's 12:09, the datetime is set to be 12:05, but the non-renewables might have only updated up to 12:00). In this case, no data would be returned since the datetime cannot be found in the non-renewables data. A solution could be to merge the live renewables data with the last updated non-renewables data whatever the time is, but this adds a dependency on the non-renewable API: if the non-renewables data takes longer than just a few minutes to update (it could be the case that it hasn't been updated for 2+ hours) the data returned by the parser would be inaccurate/out of date. 

Not sure what to do about the first problem of there being only 1 data point, aside from just returning all of the non-renewables data even if there is no corresponding renewables data for the older datetimes. But, not sure if that is ideal/meaningful/accurate to have just non-renewables data for past times but both for live data. Any suggestions would be appreciated!

parsers/AR.py
meiyiniu: @unitrium please see my comment here, which addresses your concern:
https://github.com/electricitymaps/electricitymaps-contrib/pull/4894#issuecomment-1364375994

meiyiniu:
@unitrium I've added the regional exchange under `AR-NEA_BR-S` :)

config/exchanges/AR-NEA_BR-S.yaml
```diff
@@ -0,0 +1,6 @@
+lonlat:
+  - -56.407453
+  - -28.880006
+parsers:
+  exchange: BR.fetch_exchange
+rotation: 110
```
VIKTORVAV99, line 6: The `BR.py` parser does not support the `AR-NEA->BR-S` exchange so this will have to be implemented in the `AR.py` parser. I would suggest adding it in #4893 and then we can merge that first.

meiyiniu: Added in: f65e0b9

VIKTORVAV99:
Once #4893 is merged we can proceed with this and create all the new exchange configs in this PR.

VIKTORVAV99:
Since #4893 is merged now we can proceed with this.
What we need to do in here now is first to clean up the merge conflicts, create config files for the exchanges between all zones, and then last of all make sure it's all functioning as expected. 🙂

VIKTORVAV99:
@meiyiniu can you rebase this again? It seems like it's detecting changes from other PR:s as well.

parsers/AR.py
VIKTORVAV99: Like Robin I am unsure how to best handle this (based on the above comment) as there seem to be no way to ensure the two APIs are in sync.

VIKTORVAV99 (commented):
We need to clean up some of the exchanges, the AR->BR-S exchange can be removed. The AR->CL-SEN need to be updated as well, and we need to add a new one between one of the AR subzones and UY. (And some I'm sure I have missed 😅)

meiyiniu:
@VIKTORVAV99 I've made the changes to clean up the exchanges:
- removed AR->BR-S exchange
- updated AR->CL-SEN to be AR-NOA->CL-SEN
- added AR-LIT->UY and AR-NEA->PY
- fixed the lonlat coordinates of the exchanges to be a bit more accurate and fall on the border line

VIKTORVAV99:
I have started a full review but right now I have some other stuff I need to do. Hopefully I'll be done by tonight. 🙂

VIKTORVAV99 (changes requested):
I updated the exchange directions of the arrows since  they where off for most and I added mock data so we can easily see them.

There is also two exchanges that needs to be updated as their keys are not in alphabetical order. The easies way would be to use the search tool in VS code (or another editor) and replace all instances of the key with the updated one. Just make sure to verify the exchange direction in the paser so it don't get inverted.

But the geometries looks good and if @unitrium consider the comments in the parser files we can merge this once the exchanges have been fixed (and verified).

VIKTORVAV99:
@meiyiniu any update on the progress here? It would be great to get this merged and split up AR!
Let me know if you need any assistance.

meiyiniu:
@VIKTORVAV99 Sorry about the wait! I've made the updates to change the order of non-alphabetical exchanges as well as have the UY parser handle AR-LIT as well.

VIKTORVAV99:
> @VIKTORVAV99 Sorry about the wait! I've made the updates to change the order of non-alphabetical exchanges as well as have the UY parser handle AR-LIT as well. 

Thanks! Don't worry about the wait we have had plenty of other things to do! 😅

I'll do a full review later today!

VIKTORVAV99 (commented):
Code wise things look good, I took the liberty fix some unbound issues but for some reason the production parser is not returning any data.

This seems to be because of this part:
```python
    full_production_list = [
        {
            "datetime": arrow.get(datetime_tz_ar).to("UTC").datetime,
            "zoneKey": zone_key,
            "production": merged_production_mix(
                non_renewables_production[datetime_tz_ar],
                renewables_production[datetime_tz_ar],
            ),
            "capacity": {},
            "storage": {},
            "source": "cammesaweb.cammesa.com",
        }
        for datetime_tz_ar in non_renewables_production
        if datetime_tz_ar in renewables_production
    ]
```

So we might have to rethink how we merge these two production dicts. In the worst case we could remove one of them and just use one data source (fossil one returns the most datapoints at once but the renewable is what is used today).

meiyiniu:
@VIKTORVAV99 Thanks for taking another look! In my [comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/4894#issuecomment-1364375994) a few weeks ago I explained the reason behind why the production parser sometimes doesn't return data currently. I also suggested a possible fix to mitigate this issue, but not sure if it's valid:

> A solution could be to merge the live renewables data with the last updated non-renewables data whatever the time is, but this adds a dependency on the non-renewable API: if the non-renewables data takes longer than just a few minutes to update (it could be the case that it hasn't been updated for 2+ hours) the data returned by the parser would be inaccurate/out of date.

Please let me know what you think, thanks!

VIKTORVAV99:
> @VIKTORVAV99 Thanks for taking another look! In my [comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/4894#issuecomment-1364375994) a few weeks ago I explained the reason behind why the production parser sometimes doesn't return data currently. I also suggested a possible fix to mitigate this issue, but not sure if it's valid:
> 
> > A solution could be to merge the live renewables data with the last updated non-renewables data whatever the time is, but this adds a dependency on the non-renewable API: if the non-renewables data takes longer than just a few minutes to update (it could be the case that it hasn't been updated for 2+ hours) the data returned by the parser would be inaccurate/out of date.
> 
> Please let me know what you think, thanks!

Ah yeah, while that could work in theory it would by it's very nature be highly unreliable we wouldn't have very high confidence in the data. Now if we pick the renewable data only we would have the same data as we have to day but better geographical breakdowns and in the future we might be able to estimate the unknown parts of the production. While not ideal I think this is the best way forward.

In the future it might be possible for us to have multiple production parsers for a zone then we could potentially merge this data in the backend instead keeping all datetimes and original data intact but that is just a idea from my side and something the @unitrium and the rest of the data team would have to take a look at. But this is not the only instance where it would be beneficial so it would be a nice general improvement.

So for now, I at least personally feel like we should try and get this merged with just the renewable data. And once we have that in place we can look into ways on how to improve it. Just getting the individual zones are a huge improvment. However I don't want to merge this without all of us being on the same page so I'll wait for both you and Robin to respond and then we take things from there.

meiyiniu:
@VIKTORVAV99 That makes sense! I'm on board with getting this merged just with the production data.

meiyiniu:
@unitrium Hi Robin, what are your thoughts on the discussion above about merging the data from the two endpoints? If it is okay with you then I will go ahead and make the change.

unitrium:
Hi!
I'm just back from vacations and catching up :)
If I understand it correctly, with this merge, we would produce datapoints for the regions only with renewable data for now ?

It would be ok in my opinion to show this on the map with the rest of the production being labelled as unknown.
With the simple approach `rest production = consumption - renewables`.
Then we can use one of our estimation model to breakdown the modes in the unknown production until we get the actual data. We can do that on a later stage, it's something @mathilde-daugy has been working on for IN-NO.

I see one problem with that approach, in our current setting, the most granular data takes precedence over the higher level data in the final state that is displayed on the map.

This means that since the AR region is becoming an aggregated region, the data we would show on the map for the country level view would be the data from the aggregated sub regions of AR, not the direct data from the parser of AR. 
So this would be a problem as we would show a data with less information then what we have. We will still store the data from the national parser because we have two tables in our database, one for parser data, one for the final state of the data.

So my thinking is that in the backend we should introduce an option to skip the aggregation of sub regions if we know the data of the national region. I will discuss it with the team to give you this feature during the week.
Then we can merge your changes, as @VIKTORVAV99 said, it's already a huge improvement to have the sub regions with just renewables!

VIKTORVAV99:
> I see one problem with that approach, in our current setting, the most granular data takes precedence over the higher level data in the final state that is displayed on the map.
>
> This means that since the AR region is becoming an aggregated region, the data we would show on the map for the country level view would be the data from the aggregated sub regions of AR, not the direct data from the parser of AR.
So this would be a problem as we would show a data with less information then what we have. We will still store the data from the national parser because we have two tables in our database, one for parser data, one for the final state of the data.

I'm not sure this is a problem for this particular zone as we currently only have renewable + nuclear and I do think it's using the same endpoint as well. So the sum of the subzones should be identical (minus eventual rounding) to the country level data. Here is the country level data as of right now:
image

unitrium:
Ok thanks for the insights @VIKTORVAV99!
Just to make sure does the regional data include the nuclear and hydro generation ?
If yes, then you're right we are not loosing too much data at the national level and we can proceed with this.

VIKTORVAV99:
> Ok thanks for the insights @VIKTORVAV99! Just to make sure does the regional data include the nuclear and hydro generation ? If yes, then you're right we are not loosing too much data at the national level and we can proceed with this.

I'll double check later today!

VIKTORVAV99:
The new renewable endpoint does not have a contain nuclear production but I think I might have found a way to merge the two data lists. I'm currently building a new parser for HN but I'll report back as soon as I have the time to test it.

VIKTORVAV99:
I just pushed a new commit that should fix the no match issue we had. After some testing I noticed that the non renewable endpoint has a publishing delay of 5 minutes and that we used the current time rounded to the nearest downwards 5 minute mark. This resulted in the renewable endpoint having 0s delay when run at for example 18:05 while the non-renewable endpoint would return with a 5 minute delay (18:00). So I added `timedelta(minutes=-5)` to the renewable endpoint assuming they have the same publishing deadline/delay.

meiyiniu:
@VIKTORVAV99 Hi Viktor, are there any action items for me for this PR?

VIKTORVAV99:
> @VIKTORVAV99 Hi Viktor, are there any action items for me for this PR?

Yes and no, we need a review from @unitrium (as he requested changes) and then we need to determine if the parser is stable enough. I think my latest changes improved things a little (at least during my local testing) but we might have to rework things further.

unitrium (changes requested):
Sorry for taking so much time to come back to this!
A few more stuff about the exchanges, it should be easy to fix, it's just removing code.

Regarding exchanges we should query them at the smallest granularity level, here the regions. We should maintain a single source of truth for the data. The aggregation pipeline will then create the data for the country derived from those.

unitrium:
Regarding the conflicts that are on your branch you probably need to:
- pull the latest version of master. A lot has changed recently.
- overwrite your `web/config/exchanges.json` and `web/config/zones.json` with the version on master.
- rerun npm generate-world in /web.

meiyiniu:
@unitrium I've fixed the things you mentioned, please take another look :)

VIKTORVAV99 (commented):
Things look good from my side so there should be nothing blocking this from approval once the tests are in.

I also did a final cleanup as the fix for AR broke all the AR-* zones but now it should work for both.

meiyiniu:
@unitrium Thanks for making the fixes! I've added a test for regional production and for exchange. Please take a look and let me know what you think.

VIKTORVAV99:
@unitrium what did you and @FelixDQ come up with during your discussion, is it safe to merge this as is assuming the tests check out (I'll give them a review later today) or do we need to make any other changes in here or in the backend? (Like the retry function in the parser itself we discussed really quick a while back?)

It would be great to get this merged as it seems we are way off on the carbon emissions as described in #5196 which this would take care of.

FelixDQ:
> Actually in the backend we do the time realignment. So if you publish two separate production events like this: `[{solar:10, datetime: 11/05/2015 15:05}, {unknown:20, datetime: 11/05/2015 15:35}] ` The backend would reconstruct a coherent dict for the hour 15:00 with
> 
> ```
> {
>  solar: average of all solar events from 15:00 to 16:00,
>  unknown: average of all unknown events from 15:00 to 16:00
> }
> ```


I might be misunderstanding what you mean, but I don't think that is exactly what's happening. The time alignment would essentially decide that `{solar:10, datetime: 11/05/2015 15:05}` is valid from `15:05` to `15:34`, and `{unknown:20, datetime: 11/05/2015 15:35}` would be valid from `15:35` to `15:35+2 hours or until another event comes in`. It wouldn't merge the events.

madsnedergaard:
This was a really nice contribution and we appreciate your work and all the changes you made based on all of our comments 🙏 
While it would be really nice to get this done, it's not something we currently have the time to focus on (we are going deep on expanding coverage and data quality these days), and I apologise that we left this PR hanging for so long - we will definitely pick this up again when we in the future have time to focus on splitting countries into more regions, so it's not forgotten or wasted time :)

Closing this PR for now, but keeping in in our internal backlog.

Issue #4835, opened 2022-12-05
[Data Issue]: Discrepancy w grid ISO for GB


corradio:
### When did this happen?

Sunday 4th @ 8:36 PM

### What zones are affected?

GB

### What is the problem?

Our data:
![image](https://user-images.githubusercontent.com/1655848/205581768-3521711e-5a48-4083-8bc3-ad357ec1a19d.png)

Grid ISO:
![image](https://user-images.githubusercontent.com/1655848/205581778-ec301534-ac5c-47de-8fb7-d86301a1a26f.png)

corradio:
Could be related to https://github.com/electricitymaps/electricitymaps-contrib/issues/4829

unitrium:
The latest parser data we have for GB seems to date from the 30/11.
A quick look on the [ENTSOE website](https://transparency.entsoe.eu/generation/r2/actualGenerationPerProductionType/show?name=&defaultValue=false&viewType=TABLE&areaType=CTY&atch=false&datepicker-day-offset-select-dv-date-from_input=D&dateTime.dateTime=16.11.2022+00:00|CET|DAYTIMERANGE&dateTime.endDateTime=16.11.2022+00:00|CET|DAYTIMERANGE&area.values=CTY|10Y1001A1001A92E!CTY|10Y1001A1001A92E&productionType.values=B01&productionType.values=B02&productionType.values=B03&productionType.values=B04&productionType.values=B05&productionType.values=B06&productionType.values=B07&productionType.values=B08&productionType.values=B09&productionType.values=B10&productionType.values=B11&productionType.values=B12&productionType.values=B13&productionType.values=B14&productionType.values=B20&productionType.values=B15&productionType.values=B16&productionType.values=B17&productionType.values=B18&productionType.values=B19&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2)) seems like there is an issue for UK as there is no available data at all.

unitrium:
The issue is on ELEXON's side.
Most of their power generation api report values of 0.
https://www.bmreports.com/bmrs/?q=actgenration/actualaggregated

unitrium:
We are asking them a status update on this.

kaz911:
UK is not listed as an ENTSOE member

corradio:
@unitrium isn't the issue coming from the estimation model (screenshot shows it's estimated)?

unitrium:
[...]

Issue #4236, opened 2022-06-11
Drastic differences between GB carbonintensity.org.uk API shown Carbon Intensity and GB electricitymap.org shown intensity


OwenQWERTY:
How can the carbon intensity be so different between the two sources, and would their API not be more accurate. I don't know if you replace working sources/parsers, but it also has a regional breakdown with 14 regions across GB

![Screenshot 2022-06-11 160340](https://user-images.githubusercontent.com/67866804/173193277-565f3ee8-9f89-4e5b-9f1c-06a52bf785f7.jpg)
![Screenshot 2022-06-11 155840](https://user-images.githubusercontent.com/67866804/173193280-9de719aa-385e-43da-bb65-616205cabc7a.jpg)
![Screenshot 2022-06-11 155813](https://user-images.githubusercontent.com/67866804/173193281-1044b4ad-3c78-4cbf-adb0-8bb20b2a40bb.jpg)
![Screenshot 2022-06-11 155744](https://user-images.githubusercontent.com/67866804/173193282-42a4b908-8c01-41e9-b234-4f63b9633bc3.jpg)

VIKTORVAV99:
I think the biggest reason is that the emission factors carbonintensity.org.uk use is vastly different from the ones electricityMap use.

carbonintensity.org.uk:
![Screenshot_20220611-173154_Samsung Notes.jpg](https://user-images.githubusercontent.com/30777521/173194504-099c8919-04a1-411c-ad0c-5fec981972b7.jpg)

ElectricityMap:
![Screenshot_20220611-173504_Brave.jpg](https://user-images.githubusercontent.com/30777521/173194568-1e33f52b-b150-4e67-afee-a2c3fec95dc2.jpg)

(In fact the carbonintensity.org.uk figures more closely match the electricityMap direct emissions.)

This seems to be the biggest factor in the difference co2 intensity for the different platforms although it might not account for 100% of the difference.

OwenQWERTY:
Thank you for your response!

> I think the biggest reason is that the emission factors carbonintensity.org.uk use is vastly different from the ones electricityMap use.

I didn't even consider that they would use such extremely different factors, I think the components of ElectricityMap make more sense and are more balanced overall. However the unknown is assumed to be fossil fuels, and this is not representative for nations actually reducing their carbon consumption.

VIKTORVAV99:
> Thank you for your response!
> 
[...]

PR #4899, opened 2022-12-23
Add Embeded wind/solar and pumped hydro storage to ELEXON parser.

unitrium:
## Issue
The data provided by the BMRS FUEL_INST report does not describe the full power generation in the UK. To complete it we need to add embedded wind/solar accounting for around 26% of installed capacity and hydro storage.

Related issue #4236.
## Description

Use the National Grid Eso database to query and add the missing data.


pierresegonne:
Do you have a reference for the issue that you're mentioning?

parsers/ELEXON.py
```diff
@@ -224,11 +271,30 @@ def parse_production_FUELINST(
...
+    df.rename(columns=ESO_FUEL_MAPPING, inplace=True)
+    return df.set_index("datetime")
+
 
+def process_production_events(
+    fuel_inst_data: pd.DataFrame, eso_data: pd.DataFrame
+) -> List[dict]:
+    """Combine FUELINST report and ESO data together to get the full picture and to EM Format."""
+    df = fuel_inst_data.join(eso_data, rsuffix="_eso")
+    df.rename(columns={"wind_eso": "wind", "solar_eso": "solar"}, inplace=True)
```
pierresegonne, line 293: nitpick, but we generally avoid doing inplace operations

VIKTORVAV99:
It should be the last part to fix #4835, at least temporarily.

parsers/ELEXON.py
pierresegonne: So that means we discard completely wind and solar from bnreports in favour of this new source?

pierresegonne (approved):
LGTM but would be nice to get more context when describing the PR

parsers/ELEXON.py
pierresegonne: Elements for why inplace is not really recommended in most cases:

https://towardsdatascience.com/why-you-should-probably-never-use-pandas-inplace-true-9f9f211849e4


TL;DR is that it creates mutability issues

unitrium: @pierresegonne we need to investigate which source we keep, once BNReports comes back online. The source might actually not report embedded wind and solar as well.

PR #4903, opened 2022-12-27
Update ko.json

thefirstdeer:


web/public/locales/ko.json
```diff
@@ -113,17 +113,17 @@
...
+  "hydro storage": "양수",
+  "battery storage": "배터리 저장",
   "biomass": "바이오매스",
   "nuclear": "원자력",
-  "geothermal": "지열의",
+  "geothermal": "지열",
   "gas": "가스",
   "coal": "석탄",
-  "oil": "오일",
+  "oil": "석유",
```
VIKTORVAV99, line 126: Now I have only used a online translator to verify some of these and the new one here seems to translate to petroleum? In that case I think we should go with the old one that translates to oil as that is what we are using in other languages and includes more for example when plants burn raw oil.

web/public/locales/ko.json
```diff
@@ -113,17 +113,17 @@
...
   },
-  "wind": "바람",
-  "solar": "태양",
-  "hydro": "전기",
-  "hydro storage": "전기 저장",
-  "battery storage": "배터리 용량",
+  "wind": "풍력",
+  "solar": "태양광",
+  "hydro": "수력",
+  "hydro storage": "양수",
```
VIKTORVAV99, line 119: Right now this says something like pumping if I got my translations right?

Should it not say hydro storage?

VIKTORVAV99 (commented):
I just need a bit of clarity around a few things.

It looks like some of the translations themselves are questionable when run trough a translator but that could be wrong so could you give some context around those?

It would also be great if you could fill out the PR template so we can se what the issue is this PR is trying to solve.

VIKTORVAV99:
@thefirstdeer I'm going to closet this due to the lack of a response.

If things change feel free to re-open it.

PR #4904, opened 2022-12-27
New parser for GB-NIR

msaynevirta:
## Issue

The parser for Northern Ireland isn't working anymore as UK data isn't available on the ENTSO-E transparency platform.

## Description

A separate parser has been created that uses data from SONI, the TSO of Northern Ireland. Data is collected from their [SmartGrid Dashboard](https://www.smartgriddashboard.com/) that provides almost live 15-minute measurements on demand, total generation and wind generation. Addtionally, the service describes the status of the Moyle interconnector between Nothern Ireland and Scotland.

Power transfers between the Republic of Ireland and Northern Ireland are calculated by substracting generation and Scottish transfers from the total demand of Northern Ireland.

### DIFF
--- a/parsers/GB_NIR.py
+++ b/parsers/GB_NIR.py
@@ -5,6 +5,8 @@ from typing import Optional
 from dateutil import tz
 from requests import ConnectionError, Response, Session
 
+from parsers.lib.config import refetch_frequency
+
 TZ = "Europe/Dublin"
 DATA_URL = "https://www.smartgriddashboard.com/DashboardService.svc/data"
 
@@ -53,6 +55,7 @@ def _fetch_json_data(
         raise ConnectionError
 
 
+@refetch_frequency(timedelta(minutes=15))
 def fetch_consumption(
     zone_key: str = "GB-NIR",
     session: Session = Session(),
@@ -73,6 +76,7 @@ def fetch_consumption(
         print("Failed to retrieve consumption at requested timestamp." + e)
 
 
+@refetch_frequency(timedelta(minutes=15))
 def fetch_production(
     zone_key: str = "GB-NIR",
     session: Session = Session(),
@@ -112,6 +116,7 @@ def fetch_production(
         print("Failed to retrieve production at requested timestamp." + e)
 
 
+@refetch_frequency(timedelta(minutes=15))
 def fetch_exchange(
     zone_key1: str = "GB",
     zone_key2: str = "GB-NIR",
--- a/parsers/GB_NIR.py
+++ b/parsers/GB_NIR.py
@@ -3,20 +3,22 @@ from logging import Logger, getLogger
 from typing import Optional
 
 from dateutil import tz
-from requests import Session, Response, ConnectionError
+from requests import ConnectionError, Response, Session
 
-TZ="Europe/Dublin"
-DATA_URL="https://www.smartgriddashboard.com/DashboardService.svc/data"
+TZ = "Europe/Dublin"
+DATA_URL = "https://www.smartgriddashboard.com/DashboardService.svc/data"
+
 
 def get_previous_quarter(time: datetime):
-    last_quarter_minute = 15*(time.minute//15)
-    return(time.replace(minute=last_quarter_minute))
+    last_quarter_minute = 15 * (time.minute // 15)
+    return time.replace(minute=last_quarter_minute)
+
 
 def fetch_json_data(
     region: str,
     dataset: str,
     session: Session = Session(),
-    target_datetime: Optional[datetime] = None
+    target_datetime: Optional[datetime] = None,
 ) -> dict:
     try:
         if not target_datetime:
@@ -24,12 +26,18 @@ def fetch_json_data(
 
         previous_quarter = get_previous_quarter(target_datetime)
 
-        date=previous_quarter.strftime("%d-%b-%Y+%H:%M")
+        date = previous_quarter.strftime("%d-%b-%Y+%H:%M")
 
-        params = "?area=" + dataset \
-            + "&region=" + region \
-            + "&datefrom=" + date \
-            + "&dateto=" + date
+        params = (
+            "?area="
+            + dataset
+            + "&region="
+            + region
+            + "&datefrom="
+            + date
+            + "&dateto="
+            + date
+        )
 
         response: Response = session.get(DATA_URL + params)
 
@@ -37,6 +45,7 @@ def fetch_json_data(
     except ConnectionError as e:
         print("Failed to connect to SmartGrid Dashboard:" + e)
         raise ConnectionError
+
 
 def fetch_consumption(
     zone_key: str,
@@ -50,13 +59,14 @@ def fetch_consumption(
         demand_mw = float(demand["Value"])
 
         return {
-            'zoneKey': zone_key,
-            'datetime': effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
-            'consumption': demand_mw,
-            'source': 'smartgriddashboard.com'
+            "zoneKey": zone_key,
+            "datetime": effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "consumption": demand_mw,
+            "source": "smartgriddashboard.com",
         }
     except TypeError:
         print("Failed to retrieve consumption at requested timestamp.")
+
 
 def fetch_consumption(
     zone_key: str = "GB-NIR",
@@ -70,50 +80,54 @@ def fetch_consumption(
         demand_mw = float(demand["Value"])
 
         return {
-            'zoneKey': zone_key,
-            'datetime': effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
-            'consumption': demand_mw,
-            'source': 'smartgriddashboard.com'
+            "zoneKey": zone_key,
+            "datetime": effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "consumption": demand_mw,
+            "source": "smartgriddashboard.com",
         }
     except TypeError:
         print("Failed to retrieve consumption at requested timestamp.")
+
 
 def fetch_production(
     zone_key: str = "GB-NIR",
     session: Session = Session(),
     target_datetime: Optional[datetime] = None,
-    logger: Logger =getLogger(__name__),
+    logger: Logger = getLogger(__name__),
 ) -> dict:
     try:
         production = fetch_json_data("NI", "generationactual", session, target_datetime)
         wind = fetch_json_data("NI", "windactual", session, target_datetime)
 
-        effective_time = datetime.strptime(production["EffectiveTime"], "%d-%b-%Y %H:%M:%S")
+        effective_time = datetime.strptime(
+            production["EffectiveTime"], "%d-%b-%Y %H:%M:%S"
+        )
         total_production_mw = float(production["Value"])
         wind_mw = float(wind["Value"])
 
-        unknown_mw = total_production_mw - wind_mw # remaining generation
+        unknown_mw = total_production_mw - wind_mw  # remaining generation
 
         return {
-        'zoneKey': zone_key,
-        'datetime': effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
-        'production': {
-            'biomass': None,
-            'coal': None,
-            'gas': None,
-            'hydro': None,
-            'nuclear': None,
-            'oil': None,
-            'solar': None,
-            'wind': wind_mw,
-            'geothermal': None,
-            'unknown': unknown_mw
-        },
-        'source': 'smartgriddashboard.com'
+            "zoneKey": zone_key,
+            "datetime": effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "production": {
+                "biomass": None,
+                "coal": None,
+                "gas": None,
+                "hydro": None,
+                "nuclear": None,
+                "oil": None,
+                "solar": None,
+                "wind": wind_mw,
+                "geothermal": None,
+                "unknown": unknown_mw,
+            },
+            "source": "smartgriddashboard.com",
         }
 
     except TypeError:
         print("Failed to retrieve production at requested timestamp.")
+
 
 def fetch_exchange(
     zone_key1: str = "GB",
@@ -122,18 +136,20 @@ def fetch_exchange(
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> dict:
-    sortedZoneKeys = '->'.join(sorted([zone_key1, zone_key2]))
+    sortedZoneKeys = "->".join(sorted([zone_key1, zone_key2]))
     interconnection = fetch_json_data("NI", "interconnection", session, target_datetime)
 
-    effective_time = datetime.strptime(interconnection["EffectiveTime"], "%d-%b-%Y %H:%M:%S")
+    effective_time = datetime.strptime(
+        interconnection["EffectiveTime"], "%d-%b-%Y %H:%M:%S"
+    )
     moyle_mw = float(interconnection["Value"])
 
     if sortedZoneKeys == "GB->GB-NIR":
         return {
-                'sortedZoneKeys': sortedZoneKeys,
-                'datetime': effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
-                'netFlow': moyle_mw,
-                'source': 'smartgriddashboard.com'
+            "sortedZoneKeys": sortedZoneKeys,
+            "datetime": effective_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "netFlow": moyle_mw,
+            "source": "smartgriddashboard.com",
         }
     elif sortedZoneKeys == "GB-NIR->IE":
         production = fetch_json_data("NI", "generationactual", session, target_datetime)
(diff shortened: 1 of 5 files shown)


VIKTORVAV99:
Just a small comment before a full review.

The exchange data would be really nice to get and would be pretty straight forward so no issues there.

The other data however I think I need to discuss with the team if they think it's acceptable to go from "full" resolution to just wind and unknown but in my own opinion it's better than nothing.

msaynevirta:
One additional thing that I came across while researching were these the monthly 15-min resolution logs. Newer ones contain additional data for solar generation, but the exact portion of thermal sources (and others methods) remain still unknown. These could be used for the longer term dataviews if using multiple sources is somehow technically possible. The files go back to 2014 and contain solar data from 2018 onwards.

- [2022-2023](https://www.soni.ltd.uk/media/documents/System-Data-Qtr-Hourly-2022-2023.xlsx)
- [2020-2021](https://www.soni.ltd.uk/media/documents/System-Data-Qtr-Hourly-2020-2021.xlsx) 
- [2018-2019](https://www.soni.ltd.uk/media/documents/System-Data-Qtr-Hourly-2018-2019.xlsx)
- [2016-2017](https://www.soni.ltd.uk/media/documents/System-Data-Qtr-Hourly-2016-2017.xlsx)  
- [2014-2015](https://www.soni.ltd.uk/media/documents/System-Data-Qtr-Hourly-2014-2015.xlsx)

The unknown sources could be broken down based on either these and combining them with the known [2021 fuel mix](https://www.soni.ltd.uk/media/documents/Fuel-Mix_2022.png). Another, more challenging option could be to use [the rolling 24-hour generation mix](https://www.smartgriddashboard.com/#ni/generation) of the SmartGrid Dashboard available as a [json file](https://www.smartgriddashboard.com/DashboardService.svc/data?area=fuelmix&region=NI&dateRange=day). It contains a breakdown for thermal sources but would require some math to work the hourly numbers.

VIKTORVAV99:
Getting solar data as well would definitely be an improvement but it looks like those 15 min logs don't update fast enough to be used for live generation.
Using multiple sources in the same parser is possible but currently it's not possible to use data from multiple parsers (there are some ways around this but it's not optimal).

It's possible we can use this data anyway though by training a estimation model on the old data and estimate the unknown part but I will have to discuss this with the team first.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,166 @@
...
+            + dataset
+            + "&region="
+            + region
+            + "&datefrom="
+            + date
+            + "&dateto="
+            + date
+        )
+
+        response: Response = session.get(DATA_URL + params)
```
VIKTORVAV99, line 50:
```suggestion
        params = {
            "area": dataset,
            "region": region,
            "datefrom": date,
            "dateto": date
        }

        response: Response = session.get(DATA_URL, params=params)
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,166 @@
...
+
+        return response.json()["Rows"][0]
+    except ConnectionError as e:
+        print("Failed to connect to SmartGrid Dashboard:" + e)
```
VIKTORVAV99, line 54:
```suggestion
        ParserException("GB_NIR.py", f"Failed to connect to SmartGrid Dashboard: {e}")
```
This uses ParserException instead of just printing the error.

VIKTORVAV99, line 54:
```suggestion
        ParserException("GB_NIR.py", f"Failed to connect to SmartGrid Dashboard: {e}")
```
This uses ParserException instead of just printing the error.

This should be done for a few more later down as well.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,166 @@
+from datetime import datetime, timedelta
+from logging import Logger, getLogger
+from typing import Optional
+
+from dateutil import tz
+from requests import ConnectionError, Response, Session
+
+from parsers.lib.config import refetch_frequency
```
VIKTORVAV99, line 8:
```suggestion
from parsers.lib.config import refetch_frequency
from parsers.lib.exceptions import ParserException
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,166 @@
...
+    try:
+        demand = _fetch_json_data("NI", "demandactual", session, target_datetime)
+        demand_mw = float(demand["Value"])
+
+        return {
+            "zoneKey": zone_key,
+            "datetime": _parse_effective_time(demand["EffectiveTime"]),
+            "consumption": demand_mw,
+            "source": "smartgriddashboard.com",
+        }
```
VIKTORVAV99, line 74: The API returns multiple datapoints, it would be great if we parsed them all and returned a list of datapoints, that way the parser is more robust in case of failure or if we are unable to fetch for one time period.

VIKTORVAV99 (changes requested):
Everything is working so great job!

I do have some code suggestions to improve things though to make it a little more robust.

parsers/GB_NIR.py
msaynevirta: The parser retrieves now a day's worth of data and returns the valid datapoints with number values as a list of dicts. This might need to be changed in future if we want to build a fancier breakdown of thermal generation, as it would require using the real-time and previously retrieved fuelmix data (as these are 24h aggregate values in MWh).

VIKTORVAV99: Nice!

I'll do another review as soon as I have the time but I don't think there should be any issues with it now.

I'll double check with the team about the data but in either case it would be nice to get this merged for the exchanges alone.

msaynevirta:
I calculated the following mix for the unknown generation based on the [2021 fuel mix figures](https://www.soni.ltd.uk/media/documents/Fuel-Mix_2022.png). The figures have been calculated by substracting wind generation and ignoring the energy exports.

|                      | Share, original (%) | Share, final (%) |
|----------------------|----------------|----------------|
| Gas                  | 51.5           | 70.61          |
| Solar                | 2              | 2.74           |
| Hydro                | 0.2            | 0.27           |
| Other renew          | 4.5            | 6.17           |
| Coal                 | 14.4           | 19.74          |
| Oil                  | 0.3            | 0.41           |
| Other non-renewables | 0.04           | 0.05           |
| TOTAL                | 72.94          | 100            |

This could be used as a starting point to calculate a rough emission factor if we end up using the SONI dashboard data in place of the more detailed but unavailable ENTSO-E data.

VIKTORVAV99:
That's great, we can definitely use that as the basis for the unknown co2eq intensity values as we have done for other regions.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+        params = {
+            "area": dataset,
+            "region": region,
+            "datefrom": date_from,
+            "dateto": date_to,
+        }
+
+        response: Response = session.get(DATA_URL, params=params)
+
```
VIKTORVAV99, line 68:
```suggestion

    if target_datetime is None:
        date_to = datetime.now(tz.gettz(name=TZ))
    else:
        date_to = target_datetime.astimezone(tz.gettz(name=TZ))
    # get all available data for the last 48 hours
    date_from = date_to - timedelta(days=2)

        params = {
            "area": dataset,
            "region": region,
            "datefrom": date_from,
            "dateto": date_to,
        }
    try:
        response: Response = session.get(DATA_URL, params=params)
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+        return response.json()["Rows"]
+    except ConnectionError as e:
+        ParserException("GB_NIR.py", f"Failed to connect to SmartGrid Dashboard: {e}")
+        raise ConnectionError
+
+
+@refetch_frequency(timedelta(minutes=15))
```
VIKTORVAV99, line 75:
```suggestion
@refetch_frequency(timedelta(days=2))
```
With the above change this becomes 2 days.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+def fetch_consumption(
+    zone_key: str = "GB-NIR",
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 81:
```suggestion
) -> list:
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+                consumption_list.append(
+                    {
+                        "zoneKey": zone_key,
+                        "datetime": _parse_effective_time(row["EffectiveTime"]),
+                        "consumption": float(row["Value"]),
+                        "source": "smartgriddashboard.com",
+                    }
+                )
+            else:
+                return consumption_list
```
VIKTORVAV99, line 97:
```suggestion
```
We want the loop to run over all data regardless if one is missing a value sinch we check each one anyway.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@refetch_frequency(timedelta(minutes=15))
+def fetch_consumption(
+    zone_key: str = "GB-NIR",
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
+    try:
```
VIKTORVAV99, line 82: This try catch/excpet is very large, I would suggest putting the except statment and it's contents directly after the fetch.

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+                        "zoneKey": zone_key,
+                        "datetime": _parse_effective_time(row["EffectiveTime"]),
+                        "consumption": float(row["Value"]),
+                        "source": "smartgriddashboard.com",
+                    }
+                )
+            else:
+                return consumption_list
+
+        return consumption_list
```
VIKTORVAV99, line 99:
```suggestion
        if consumption_list != []:
            return consumption_list
        else:
            raise ParserException(
            "GB_NIR.py", "No valid data was able to be parsed."
        )
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+    except TypeError as e:
+        ParserException(
+            "GB_NIR.py", f"Failed to retrieve consumption at requested timestamp: {e}"
+        )
+        raise TypeError
```
VIKTORVAV99, line 104:
```suggestion
        raise ParserException(
            "GB_NIR.py", f"Failed to retrieve consumption at requested timestamp: {e}"
        )
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+                            "solar": None,
+                            "wind": wind_generation_mw,
+                            "geothermal": None,
+                            "unknown": unknown_generation_mw,
+                        },
+                        "source": "smartgriddashboard.com",
+                    }
+                )
+            else:
+                return production_list
```
VIKTORVAV99, line 169:
```suggestion
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+        return production_list
```
VIKTORVAV99, line 171:
```suggestion
        if production_list != []:
            return production_list
        else:
            raise ParserException(
            "GB_NIR.py", "No valid data was able to be parsed."
        )
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+    except TypeError as e:
+        ParserException(
+            "GB_NIR.py", f"Failed to retrieve production at requested timestamp: {e}"
+        )
+        raise TypeError
```
VIKTORVAV99, line 177:
```suggestion
    except TypeError as e:
        raise ParserException(
            "GB_NIR.py", f"Failed to retrieve production at requested timestamp: {e}"
        )
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@refetch_frequency(timedelta(minutes=15))
+def fetch_exchange(
+    zone_key1: str = "GB",
+    zone_key2: str = "GB-NIR",
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 187:
```suggestion
) -> list:
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+        raise TypeError
+
+
+@refetch_frequency(timedelta(minutes=15))
+def fetch_production(
+    zone_key: str = "GB-NIR",
+    session: Session = Session(),
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 113:
```suggestion
) -> list:
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+                interconnection_list.append(
+                    {
+                        "sortedZoneKeys": sortedZoneKeys,
+                        "datetime": _parse_effective_time(moyle_row["EffectiveTime"]),
+                        "netFlow": moyle_row["Value"],
+                        "source": "smartgriddashboard.com",
+                    }
+                )
+            else:
+                return interconnection_list
```
VIKTORVAV99, line 205:
```suggestion
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+            production_json, demand_json, moyle_json
+        ):
+            production_mw = production_row["Value"]
+            demand_mw = demand_row["Value"]
+            moyle_mw = moyle_row["Value"]
+
+            if production_mw and demand_mw and moyle_mw:
+                power_flow_mw = (
+                    production_mw + moyle_mw - demand_mw
+                )  # power flow gb_nir -> ie
```
VIKTORVAV99, line 224: I'm not sure we can do this as we don't account for losses here and would get a higher exchange than it should be.
@unitrium could I get your opinion on this?

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+                interconnection_list.append(
+                    {
+                        "sortedZoneKeys": sortedZoneKeys,
+                        "datetime": _parse_effective_time(moyle_row["EffectiveTime"]),
+                        "netFlow": power_flow_mw,
+                        "source": "smartgriddashboard.com",
+                    }
+                )
+            else:
+                return interconnection_list
```
VIKTORVAV99, line 234:
```suggestion
```

parsers/GB_NIR.py
```diff
@@ -0,0 +1,245 @@
...
+
+        return interconnection_list
```
VIKTORVAV99, line 236:
```suggestion
        if interconnection_list != []:
            return interconnection_list
        else:
            raise ParserException(
            "GB_NIR.py", "No valid data was able to be parsed."
        )
```

VIKTORVAV99 (changes requested):
Some final comments, most of them are nitpicks and cleanup from the previous changes.

parsers/GB_NIR.py
msaynevirta: One question, does this push the data to be older than the expected 2 hours? The SONI dashboard and the json data source updates every 15 minutes during the day but would this imply we pull their data only every 2 days?

parsers/GB_NIR.py
msaynevirta: I removed the TypeError checks from the functions as the values are validated now inside the loops and the lists are checked prior to returning them.

parsers/GB_NIR.py
msaynevirta: To my knowledge the SONI / EirGrid dashboard should be taking into account the demand side system losses, but is there something else that should be taken into account to compute these flows? The dashboard has also the nubmers for ROI available (demand, generation and EWIC transfers), so it might be interesting to compare the two and see if there's a more robust way to compute the power flows between them.

VIKTORVAV99:
I have merged this PR into #4941 as that had more parser functions but this had capacity updates, source documentation and exchange updates as well. I hope you understand.

msaynevirta:
Sure, makes sense. The new IE parser implements everything nicely. There seems to be small bugs in it though, the Moyle interconnector feeds currently from IE to GB-NIR insted of GB to GB-NIR as it should. Also, breaking the unknown generation furhter down might also be something of interest in the future. Short term the easiest is probably to just calculate a more accurate average emissons factor but something more detailed might be possible to build out of that live 15 minute generation breakdown on the eirgrid / soni site. IE - GB-NIR flows are also an interseting question but calculating them doesn't seem so straight forward as it first seemed due to certain small inconsistensies in the data.

VIKTORVAV99:
> Sure, makes sense. The new IE parser implements everything nicely. There seems to be small bugs in it though, the Moyle interconnector feeds currently from IE to GB-NIR insted of GB to GB-NIR as it should. Also, breaking the unknown generation furhter down might also be something of interest in the future. Short term the easiest is probably to just calculate a more accurate average emissons factor but something more detailed might be possible to build out of that live 15 minute generation breakdown on the eirgrid / soni site. IE - GB-NIR flows are also an interseting question but calculating them doesn't seem so straight forward as it first seemed due to certain small inconsistensies in the data.

Would you mind implementing a fix for the interconnector?

As for the unknown data, it will be further broken down by a estimation model in the backend.

msaynevirta:
Sure, I'll pull the changes and take a look.

VIKTORVAV99:
Thanks!

PR #4905, opened 2022-12-27
Updates isort to the same version as the backend

VIKTORVAV99:
## Issue
Pins the version to the same version as the backend PR.


unitrium:
Also upgraded black like in the backend (I forgot to mention it on the PR), requires a little reformating.

unitrium (approved):

PR #4909, opened 2022-12-27
CR parser: Added new production source

IV1T3:
## Issue
Fixes issue #4190.

## Description
The parser now uses a new data source since the old one was removed.

### DIFF
--- a/parsers/CR.py
+++ b/parsers/CR.py
@@ -8,115 +8,119 @@ from typing import Optional
 
 import arrow
 import pandas as pd
-from bs4 import BeautifulSoup
 from requests import Session
 
 TIMEZONE = "America/Costa_Rica"
-DATE_FORMAT = "DD/MM/YYYY"
-MONTH_FORMAT = "MM/YYYY"
-POWER_PLANTS = {
-    "Aeroenergía": "wind",
-    "Altamira": "wind",
-    "Angostura": "hydro",
-    "Arenal": "hydro",
-    "Balsa Inferior": "hydro",
-    "Barranca": "oil",
-    "Barro Morado": "geothermal",
-    "Belén": "hydro",
-    "Bijagua": "hydro",
-    "Birris12": "hydro",
-    "Birris3": "hydro",
-    "Boca de Pozo": "hydro",
-    "CNFL": "hydro",
-    "Cachí": "hydro",
-    "Campos Azules": "wind",
-    "Canalete": "hydro",
-    "Cariblanco": "hydro",
-    "Carrillos": "hydro",
-    "Caño Grande": "hydro",
-    "Caño Grande III": "hydro",
-    "Chiripa": "wind",
-    "Chocosuelas": "hydro",
-    "Chucás": "hydro",
-    "Cote": "hydro",
-    "Cubujuquí": "hydro",
-    "Daniel Gutiérrez": "hydro",
-    "Dengo": "hydro",
-    "Don Pedro": "hydro",
-    "Doña Julia": "hydro",
-    "Echandi": "hydro",
-    "Electriona": "hydro",
-    "El Encanto": "hydro",
-    "El Angel": "hydro",
-    "El Angel Ampliación": "hydro",
-    "El Embalse": "hydro",
-    "El General": "hydro",
-    "El Viejo": "biomass",
-    "Garabito": "oil",
-    "Garita": "hydro",
-    "Guápiles": "oil",
-    "Hidrozarcas": "hydro",
-    "Jorge Manuel Dengo": "hydro",
-    "La Esperanza (CoopeL)": "hydro",
-    "La Joya": "hydro",
-    "Las Pailas II": "geothermal",
-    "Los Negros": "hydro",
-    "Los Negros II": "hydro",
-    "Los Santos": "wind",
-    "MOVASA": "wind",
-    "Matamoros": "hydro",
-    "Miravalles I": "geothermal",
-    "Miravalles II": "geothermal",
-    "Miravalles III": "geothermal",
-    "Miravalles V": "geothermal",
-    "Moín I": "oil",
-    "Moín II": "oil",
-    "Moín III": "oil",
-    "Orosí": "wind",
-    "Orotina": "unknown",
-    "Otros": "unknown",
-    "PE Cacao": "wind",
-    "PE Mogote": "wind",
-    "PE Río Naranjo": "hydro",
-    "PEG": "wind",
-    "Pailas": "geothermal",
-    "Parque Solar Juanilama": "solar",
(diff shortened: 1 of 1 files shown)


parsers/CR.py
```diff
@@ -8,115 +8,120 @@
...
+#     "Vara Blanca": "hydro",
+#     "Ventanas": "hydro",
+#     "Ventanas-Garita": "hydro",
+#     "Vientos de La Perla": "wind",
+#     "Vientos de Miramar": "wind",
+#     "Vientos del Este": "wind",
+#     "Volcán": "hydro",
+# }
+
+# Note: Ignoring Intercambio deliberately
```
VIKTORVAV99, line 115: Is Intercambio the exchange data? If so could it be used to create a exchange parser?

parsers/CR.py
```diff
@@ -202,62 +175,65 @@ def fetch_production(
...
-
-    # tell pandas which table to use by providing CHARACTERISTIC_NAME
-    df = pd.read_html(
-        response.text, match=CHARACTERISTIC_NAME, skiprows=1, index_col=0
-    )[0]
-
-    results = df_to_data(zone_key, target_datetime, df, logger)
+    day = target_datetime.format("DD")
+    month = target_datetime.format("MM")
+    year = target_datetime.format("YYYY")
```
VIKTORVAV99, line 180: `target_datetime` have the `datetime` type which don't have a format function but is assigned a arrow object higher up. This causes multiple type errors.

VIKTORVAV99 (commented):
The parser seems to work with the current changes so nice work!

Although I'm a little hesitant to merge this in it's current state due to some of the old pieces of code however I relize this is not part of the current scope for this PR so I won't require any changes in here.

Let me know if you think this should be merged as is and I'll approve it and we can fix the old parts in another PR:

parsers/CR.py
IV1T3: Yes Intercambio is indeed the exchange data and was categorized into North (Nicaragua) and South (Panama). Added a new exchange parser based on the already existing parser https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/ENTE.py. Does fetch_exchange() have to implemented in CR.py or can we just use ENTE.py as an exchange parser using the config/zones/CR.yaml config?

VIKTORVAV99: There is no requirement for a fetch_exchange() method as well but having it would be nice.
I haven't checked which parser would be "best" in terms of data and quality but having a second option is always best in case the first one becomes unavailable.

IV1T3: The fetch_exchange() in CR.py is now basically the same as in ENTE.py using the same data source. My question was rather about whether there has to be an exchange parser for CR in CR.py explicitly or can we just set ENTE.py to be the exchange parser for CR in its respective config file?

VIKTORVAV99: Oh they are both using the same source (and domain)?
If that's the case it's is indeed pretty redundant but I assume it's already been implemented now so lets keep it. I'll try and do a full review of this PR before the weekend.

IV1T3: Any update on this one?

VIKTORVAV99: I have been pretty busy so didn't have time, but I have a lot of free time the rest of the week so I promise I'll get it done this week at the very least.

parsers/CR.py
```diff
@@ -7,115 +7,26 @@
...
 }
 
-CHARACTERISTIC_NAME = "Angostura"
+EXCHANGE_JSON_MAPPING = {
+    "CR->NI": "5SISTEMA.LT230.INTER_NET_CR.CMW.MW",
+    "CR->PA": "6SISTEMA.LT230.INTER_NET_PAN.CMW.MW",
+}
 
 
 def empty_record(zone_key: str):
```
VIKTORVAV99, line 32: We should never return a manually modified datapoint, especially not a 0 filled one as it will break out ability to estimate the zone in the future.

parsers/CR.py
```diff
@@ -179,85 +77,84 @@ def fetch_production(
...
-    # tell pandas which table to use by providing CHARACTERISTIC_NAME
-    df = pd.read_html(
-        response.text, match=CHARACTERISTIC_NAME, skiprows=1, index_col=0
-    )[0]
+    results = {}
+    for hourly_item in resp_data:
+        if hourly_item["fuente"] == "Intercambio":
+            continue
+        if hourly_item["fecha"] not in results:
+            results[hourly_item["fecha"]] = empty_record(zone_key)
```
VIKTORVAV99, line 116:
```suggestion
            raise ParserException("CR.py", f"Failed to fetch data for {zone_key}."
```

parsers/CR.py
```diff
@@ -179,85 +77,84 @@ def fetch_production(
...
+    s = session or Session()
 
-    if sorted_zone_keys == "CR->NI":
-        flow = df["NICR"][0]
-    elif sorted_zone_keys == "CR->PA":
-        flow = -1 * df["CRPA"][0]
-    else:
-        raise NotImplementedError("This exchange pair is not implemented")
+    raw_data = s.get(EXCHANGE_URL).json()
+    flow = round(extract_exchange(raw_data, sorted_zones), 1)
```
VIKTORVAV99, line 147: Since `extract_exchange` can return None we can't call round on it directly. We need to make sure it's a number first.

VIKTORVAV99 (changes requested):
The parser works great but there is one thing inherited from the old  parser we need to fix and one small type error.

After this is done we can merge this.

parsers/CR.py
IV1T3: What do you mean by "manually modified"? All "empty records" are going to be filled later on from line 118 through 123. I first just create the empty dictionary and populate it afterwards.

parsers/CR.py
```diff
@@ -179,85 +66,97 @@ def fetch_production(
...
-    df = pd.read_html(
-        response.text, match=CHARACTERISTIC_NAME, skiprows=1, index_col=0
-    )[0]
+        if (
+            SPANISH_TO_ENGLISH[hourly_item["fuente"]]
+            not in results[hourly_item["fecha"]]["production"]
+        ):
+            results[hourly_item["fecha"]]["production"][
+                SPANISH_TO_ENGLISH[hourly_item["fuente"]]
+            ] = 0.0
```
VIKTORVAV99, line 117: Why are we manually assigning this to 0?
It should be either ignored or raise a exception if it's not available.

Right now it seems to result in hydro being 0 even when the data is unavailabe, we should avoid this so we can run estimations.

VIKTORVAV99 (changes requested):
All but one little thing look good, but we need to change that before it can be merged.

VIKTORVAV99 (approved):
LGTM! 🎉

Issue #4828, opened 2022-12-02
GP parser works locally but fails when run in the backend?


VIKTORVAV99:
## Description
#4742 introduced a new parser that should have enabled Guadalupe (GP) to be colored on the map but this is not the current case.

Currently the parser returns a non standard format that includes a `estimated: <bool>` in the return object to flag if the data is estimated or not since the API endpoint for the parser includes both and without the estimated data the lag would be months.

Now this works locally without any issues but it seems to fail when ran in the backend probably due to the extra key in the return object.

Parser logs:
```log
Expecting value: line 2 column 1 (char 1)
File "/projects/feeder-electricity/src/lib/fetch_data.py", line 172, in launch_parsers
    objs = res.get(timeout) or []
File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
     raise self._value
File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker
     result = (True, func(*args, **kwds))
File "/projects/feeder-electricity/src/lib/fetch_data.py", line 164, in <lambda>
     lambda: parser(    File "/contrib/parsers/FR_O.py", line 166, in fetch_production     production_objects, date_string = fetch_data(
File "/contrib/parsers/FR_O.py", line 137, in fetch_data
     data = response.json()
File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 900, in json
     return complexjson.loads(self. Text, **kwargs)
File "/usr/local/lib/python3.8/json/__init__.py", line 357, in loads
     return _default_decoder.decode(s)
File "/usr/local/lib/python3.8/json/decoder.py", line 337, in decode
     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
File "/usr/local/lib/python3.8/json/decoder.py", line 355, in raw_decode
     raise JSONDecodeError("Expecting value", s, err. Value) from None
```

Now I see two solutions to this:

1. Ignore the estimated data and return a normal parser production object and accepting the months long delays.
2. Modify the backend to accept the new object form and either reject the estimated data or use it in place of Electricity Maps own estimation models. In theory the estimations from the TSO should be higher fidelity as they have access to more data points.

>**Note**
>This applies for all zones using the `FR_O.py` parser.

VIKTORVAV99:
@pierresegonne we talked a little about this already a while back when I re-did the FR_O parser so pinging you in here as well.

pierresegonne:
[...]

PR #4910, opened 2022-12-28
Temp fix for FR_O.py

VIKTORVAV99:
## Issue
The parser currently fails when run in the backend due to the inclusion of an estimated key in the return object.
## Description
Temporarily comments out the additional estimated key until the changes discussed in #4828 are implemented.


unitrium (approved):
I'm not even sure at the moment if the `estimated` key is used in the pipeline so that those data points are routed to the estimated data table. I think they just end up as regular datapoints in parsed_data. Estimated data is only generated through our models.

VIKTORVAV99:
> I'm not even sure at the moment if the `estimated` key is used in the pipeline so that those data points are routed to the estimated data table. I think they just end up as regular datapoints in parsed_data. Estimated data is only generated through our models.

Correct, which is what is causing the issue because the pipeline don't currently support the estimated key. Hopefully this is something we can support soon but until then it would be nice to get the data at least and then we can re-fetch it later with the correct tags once it's implemented.

PR #4911, opened 2022-12-28
update capacities North America (Canada, USA, Puerto Rico)

KabelWlan:
## Description
I updated the capacities in North America (Canada, USA, Puerto Rico) with the latest data available. Still the same sources as already listed in DATA_SOURCES.md.

### DIFF
--- a/config/zones/US-HI-OA.yaml
+++ b/config/zones/US-HI-OA.yaml
@@ -4,14 +4,14 @@ bounding_box:
   - - -157.14716739499988
     - 22.209850429000085
 capacity:
-  battery storage: 186
+  battery storage: 40.9
   biomass: 157.7
   coal: 0
   gas: 0
   geothermal: 0
   hydro: 0
   oil: 1528.9
-  solar: 201.1
+  solar: 240.9
   unknown: 0
   wind: 126.6
 contributors:
--- a/config/zones/US-HI-OA.yaml
+++ b/config/zones/US-HI-OA.yaml
@@ -4,7 +4,7 @@ bounding_box:
   - - -157.14716739499988
     - 22.209850429000085
 capacity:
-  battery storage: 40.9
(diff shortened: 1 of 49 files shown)


config/zones/US-HI-OA.yaml
```diff
@@ -4,14 +4,14 @@ bounding_box:
   - - -157.14716739499988
     - 22.209850429000085
 capacity:
-  battery storage: 186
+  battery storage: 40.9
```
VIKTORVAV99, line 7: This don't look right, any explanation for the sudden decrease?

(It's possible the source was not added to the sources file.)

VIKTORVAV99 (commented):
Everything looks good, I just have one question about one change.

config/zones/US-HI-OA.yaml
KabelWlan: In #4526 was the Kapolein Energy Storage Facility added. This one is probably missing because it is not yet in operation, according to this source: https://www.spglobal.com/marketintelligence/en/news-insights/latest-news-headlines/us-energy-storage-wave-builds-strength-amid-delays-72885152

This makes sense since the facility isn't yet listed in the EIA data set.

VIKTORVAV99: I think we could just add 185 MW with the existing EIA value then and hope it updates correctly next year.

KabelWlan: That's fine with me. The Hawaii data isn't very accurate at the moment anyway. The estimated data is still based on the coal power plant running. However, that's another topic.

VIKTORVAV99 (approved):

VIKTORVAV99:
Thanks for all the updates in here and for updating it so fast! 🎉

KabelWlan:
Thanks from my side as well to both of you

PR #4912, opened 2022-12-29
Fix DK's exchange parser's invalid datetime format

unitrium:
## Issue

NL parser is down at the moment, this is due to the fact that NL calls the DK exchange parser which doesn't handle timezone aware datetimes in the API call.

## Description

If the datetime is passed with a timezone to DK's exchange fetcher then remove it.

### DIFF
--- a/parsers/DK.py
+++ b/parsers/DK.py
@@ -4,8 +4,8 @@ from typing import List, Optional, Union
 
 from requests import Response, Session
 
-from .lib.config import refetch_frequency
-from .lib.exceptions import ParserException
+from electricitymap.contrib.parsers.lib.config import refetch_frequency
+from electricitymap.contrib.parsers.lib.exceptions import ParserException
 
 EXCHANGE_MAPPING = {
     "DE->DK-DK1": {"id": "ExchangeGermany", "direction": 1, "priceArea": "DK1"},
@@ -29,6 +29,11 @@ def fetch_data(
     Helper function to fetch data from the API.
     """
     ses = session or Session()
+
+    if target_datetime.tzinfo:
+        # Data source doesn't support timezone aware
+        # datetimes.
+        target_datetime.replace(tzinfo=None)
 
     params = {
         "limit": 144,
--- a/parsers/DK.py
+++ b/parsers/DK.py
@@ -4,8 +4,8 @@ from typing import List, Optional, Union
 
 from requests import Response, Session
 
-from electricitymap.contrib.parsers.lib.config import refetch_frequency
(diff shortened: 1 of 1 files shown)


parsers/DK.py
```diff
@@ -4,8 +4,8 @@
 
 from requests import Response, Session
 
-from .lib.config import refetch_frequency
-from .lib.exceptions import ParserException
+from electricitymap.contrib.parsers.lib.config import refetch_frequency
+from electricitymap.contrib.parsers.lib.exceptions import ParserException
```
VIKTORVAV99, line 8: Pretty sure this will fail on windows as it don't support symbolic links.

unitrium: In which folder should I be ? Because otherwise the relative path doesn't work if I'm at the root, `poetry run parsers/NL.py` it fails due to the relative imports.

VIKTORVAV99: I think it's best to use the `test_parser` instead of running the file directly.

So try `poetry run test_parser NL`, that should work.

parsers/DK.py
```diff
@@ -30,6 +30,11 @@ def fetch_data(
     """
     ses = session or Session()
 
+    if target_datetime.tzinfo:
+        # Data source doesn't support timezone aware
+        # datetimes.
+        target_datetime.replace(tzinfo=None)
```
pierresegonne, line 36:
```suggestion
        target_datetime = target_datetime.replace(tzinfo=None)
```

See [here](https://www.geeksforgeeks.org/datetime-replace-function-in-python/) for documentation that explains that `.replace` returns a datetine and does not do inplace mutation

Weirdly it seems that it fixes real-time but then it throws an internal server error with historical data:

https://api.energidataservice.dk/dataset/ElectricityProdex5MinRealtime?limit=144&filter=%7B%22PriceArea%22%3A%22DK1%22%7D&start=2022-12-29T00%3A00&end=2022-12-30T00%3A00

pierresegonne (changes requested):

parsers/DK.py
pierresegonne: At the same time the endpoint is called "RealTime" so maybe that's to be expected 😅

unitrium: I'm not sure why but now the historical data link you posted seems to work just fine...

pierresegonne (approved):

pierresegonne:
@unitrium I went ahead to make the fix, `NL` is a priority zone for us, and we are breaking our SLO on data completeness by having it be out.

unitrium:
Thanks! Let's release it as soon as possible.

PR #4919, opened 2023-01-01
Fix SE unknown emission factors

VIKTORVAV99:
## Issue
The current Swedish emission factors are based on the regional emissions calculated for biomass that was later reverted.
## Description
Adjust the emission factors using 230 as the value for biomass.


pierresegonne (approved):
Thanks for keeping such a close eye on Sweden ;) 

And happy new year!

VIKTORVAV99:
> Thanks for keeping such a close eye on Sweden ;) 
> 
> And happy new year!

I felt kind of obligated since I changed them in the first place 😅

Happy New years to you as well!

PR #4921, opened 2023-01-02
Remove SE parser aggregation + general cleanup for DK, NO and SE

VIKTORVAV99:
## Issue
We are making unnecessary API calls for data that will never be available or is already fetched by other means (in the case of the parser level aggregation for SE).

## Description
- Removes the subzone aggregation in the ENTSOE parser for SE. - We can fetch historical data for SE with this removal and it should not impact our "live" data for SE as it is aggregated in the backend as well.
- Removes productionPerUnits from SE. - We only assign these to the subzones anyway so it will never be return any data.
- Removes price parsers for DK, NO, SE. - Will never be available as all the bidding zones are priced individually and there is no national pricing.
- Adds timezones to NO subzones. - They should not be null. 😅


VIKTORVAV99:
@pierresegonne this PR would allow us to fetch country level data for SE that exist since 2016 but it's missing solar and gas. But since the aggregations currently overwrite the fetched country level data this should allow us to get the best of both worlds and not be dependant on (estimated) aggregations for SE data? (I noticed the historical estimations for SE-SE3 are way off due to the closure several nuclear power plants. Should be in the 8-9 GW range not the current 5-6 GW range.)

pierresegonne:
Looks good, except that I don't understand exactly the consequences of the changes for `SE` on the production parser?

VIKTORVAV99:
> Looks good, except that I don't understand exactly the consequences of the changes for `SE` on the production parser?

Hopefully the only difference in the end is that we gain years worth of data.
But I'll attempt to explain it a little more in depth.

Currently the SE subzones have data for gas, solar, wind, hydro, nuclear and unknown but this data is only available from 2021 march or april something.

Sweden as a whole on the other hand only has wind, hydro, nuclear and unknown but data since 2016-01-01 at least.

So before we had the temporal aggregations and had historical data from the SVK parser it made sense to just aggregate the subzones directly in the parser to gain the additional granularity from solar and gas production.

However at this point in time this should be redundant as this is already aggregated by the aggregation modells. So right now we fetch the same data twice and use it once...

Now this removes that again so we get historical data further back in time and still get the solar and gas data when it's available due to the aggregation in the backend.

Hopefully that makes sense and my aim here is that we can fetch SE data directly so we won't have to rely on aggregated estimations for the Sweden as they are a bit off right now (like I mentioned in #4383).

pierresegonne:
Ok thanks for the clarification.

I'm not sure it will work though. I double checked in the backend code, and it looks like that a zone that has subzones won't get flowtraced, but rather expect to be computed as the aggregate from its children.

pierresegonne (approved):

VIKTORVAV99:
> Ok thanks for the clarification.
> 
> I'm not sure it will work though. I double checked in the backend code, and it looks like that a zone that has subzones won't get flowtraced, but rather expect to be computed as the aggregate from its children.

Oh well, in the worst case scenario we save a few API calls then. If it don't work we can remove all the SE parser all together I assume?

PR #4925, opened 2023-01-03
update_UY_parser

mathilde-daugy:
## Description

When doing the DQR for 4Q 2022, i noticed that the UY parser was down and that the link we used was no longer available. I updated the parser to collect data from ADME (electricity market operator). This source has historical data for generation, consumption and exchanges.

Also I removed the part getting the Salto Grande data from Cammesa (AR) because it is included in the new source (see issue [1325
](https://github.com/electricitymaps/electricitymaps-contrib/issues/1325)

Data is available at 10-min intervals

Exchanges are also available but we use the values from the BR and AR parsers

### Preview
data for 01/01/2023

production
image

consumption
image

### DIFF
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -93,7 +93,6 @@ def fetch_data(
 
 def fix_solar_production(dt: datetime, row: pd.Series) -> int:
     """sets solar production to 0 during the night as there is only solar PV in UY"""
-    print(row.get("value"))
     if (5 >= dt.hour or dt.hour >= 20) and row.get("value") != 0:
         return 0
     else:
@@ -218,8 +217,8 @@ def fetch_exchange(
     return all_data_points
 
 
-# if __name__ == "__main__":
-#     print("fetch_production() ->")
-#     print(fetch_production())
-#     print("fetch_exchange(UY, BR) ->")
-#     print(fetch_exchange("UY", "BR"))
+if __name__ == "__main__":
+    print("fetch_production() ->")
+    print(fetch_production())
+    # print("fetch_exchange(UY, BR) ->")
+    # print(fetch_exchange("UY", "BR"))
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -100,7 +100,7 @@ def fix_solar_production(dt: datetime, row: pd.Series) -> int:
 @refetch_frequency(timedelta(days=1))
 def fetch_production(
     zone_key: str = "UY",
-    session: Session = Session()
+    session: Session = Session(),
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> list:
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -99,7 +99,7 @@ def fix_solar_production(dt: datetime, row: pd.Series) -> int:
 @refetch_frequency(timedelta(days=1))
 def fetch_production(
     zone_key: str = "UY",
-    session: Optional[Session] = None,
+    session: Session = Session()
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> list:
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -52,6 +52,7 @@ def get_adme_url(target_datetime: datetime, session: Session) -> str:
     r = session.get(url=link)
     soup = BeautifulSoup(r.content, "html.parser")
     href_tags = soup.find_all("a", href=True)
+    data_url: str = ""
     for tag in href_tags:
         if tag.button is not None:
             if tag.button.string == "Archivo Scada Detalle 10minutal":
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -40,18 +40,15 @@ EXCHANGES_MAPPING = {
 }
 
 
-def get_adme_url(target_datetime: datetime, session: Optional[Session] = None) -> str:
+def get_adme_url(target_datetime: datetime, session: Session) -> str:
     """"""
-    session = session or Session()
     next_day = target_datetime + timedelta(days=1)
-    date_format = "{day}%2F{month}%2F{year}".format(
-        day=target_datetime.day, month=target_datetime.month, year=target_datetime.year
+    date_format = (
+        f"{target_datetime.day}%2F{target_datetime.month}%2F{target_datetime.year}"
     )
-    next_day_format = "{day}%2F{month}%2F{year}".format(
-        day=next_day.day, month=next_day.month, year=next_day.year
-    )
+    next_day_format = f"{next_day.day}%2F{next_day.month}%2F{next_day.year}"
 
-    link = ADME_URL + date_format + "&fecha_fin=" + next_day_format + "&send=MOSTRAR"
+    link = f"{ADME_URL}{date_format}&fecha_fin={next_day_format}&send=MOSTRAR"
     r = session.get(url=link)
     soup = BeautifulSoup(r.content, "html.parser")
     href_tags = soup.find_all("a", href=True)
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -1,13 +1,11 @@
 #!/usr/bin/python3
 
 import io
-import re
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
 from typing import Optional
 
 import arrow
-import dateutil
 import pandas as pd
 import pytz
 
@@ -19,18 +17,7 @@ from electricitymap.contrib.config.constants import PRODUCTION_MODES
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
-tz = "America/Montevideo"
-
-MAP_GENERATION = {
-    "Hidráulica": "hydro",
-    "Eólica": "wind",
-    "Fotovoltaica": "solar",
-    "Biomasa": "biomass",
-    "Térmica": "oil",
-}
-INV_MAP_GENERATION = dict([(v, k) for (k, v) in MAP_GENERATION.items()])
-
-SALTO_GRANDE_URL = "http://www.cammesa.com/uflujpot.nsf/FlujoW?OpenAgent&Tensiones y Flujos de Potencia&"
+UY_TZ = "America/Montevideo"
 
 ADME_URL = "https://pronos.adme.com.uy/gpf.php?fecha_ini="
 PRODUCTION_MODE_MAPPING = {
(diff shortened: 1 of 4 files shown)


VIKTORVAV99:
So eeh #4840 is already open for this 😅
However there are some issues with that PR that need to be addressed before merging so I'll compare the two and give my opinion on which should be closed or merged.

mathilde-daugy:
> So eeh #4840 is already open for this 😅 However there are some issues with that PR that need to be addressed before merging so I'll compare the two and give my opinion on which should be closed or merged.
> 
> Initial looks favor this PR though as it only uses a single source (which was the cause of the issues in the other PR).

Ah my bad, i didn't see that...

> So eeh #4840 is already open for this 😅 However there are some issues with that PR that need to be addressed before merging so I'll compare the two and give my opinion on which should be closed or merged.
> 
> Initial looks favor this PR though as it only uses a single source (which was the cause of the issues in the other PR).

Ah my bad i didn't see that....
I also noticed that the other source doesn't enable us to backfill because it is a live data source. This means that if for some reason we can't get the data point, it is lost forever which is not ideal. 
I've checked and we can get data from at least 2020/01/01 using this source and also access spot prices.

VIKTORVAV99:
> > So eeh #4840 is already open for this 😅 However there are some issues with that PR that need to be addressed before merging so I'll compare the two and give my opinion on which should be closed or merged.
> > Initial looks favor this PR though as it only uses a single source (which was the cause of the issues in the other PR).
> 
> Ah my bad i didn't see that.... I also noticed that the other source doesn't enable us to backfill because it is a live data source. This means that if for some reason we can't get the data point, it is lost forever which is not ideal. I've checked and we can get data from at least 2020/01/01 using this source and also access spot prices.

Yeah it looks like this source is way better quality wise, so don't worry to much about the other parser PR, it would most likely have been replaced by this sooner or later.

I'll try and do a full review tomorrow.

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
 
-def parse_page(session: Optional[Session]):
-    r = session or Session()
-    url = "https://apps.ute.com.uy/SgePublico/ConsPotenciaGeneracionArbolXFuente.aspx"
-    response = r.get(url)
-    soup = BeautifulSoup(response.text, "html.parser")
 
-    datefield = soup.find(
-        "span", attrs={"id": "ctl00_ContentPlaceHolder1_lblUltFecScada"}
+def get_adme_url(target_datetime: datetime, session: Optional[Session] = None) -> str:
```
unitrium, line 43:
```suggestion
def get_adme_url(target_datetime: datetime, session: Session) -> str:
```
Since we are calling this function only from functions that handle the missing session case on their own, I would rather have this function just require a session.
This would add some clarity as the missing session problem would be treated all at the top and everything down just requires the session argument.

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
+    """"""
+    session = session or Session()
```
unitrium, line 45: No longer needed if the above is accepted.

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
-import dateutil
+import pandas as pd
+import pytz
 
 # BeautifulSoup is used to parse HTML to get information
 from bs4 import BeautifulSoup
 from requests import Session
 
-tz = "America/Montevideo"
+from electricitymap.contrib.config.constants import PRODUCTION_MODES
```
pierresegonne, line 16: I think you can just do config.constants the full path does not work on windows if I'm not mistaken

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
+    else:
+        raise ParserException(
+            parser="UY.py",
+            message="no data available for target_dateitme",
+        )
+
+
+def fix_solar_production(dt: datetime, row: pd.Series) -> int:
+    """sets solar production to 0 during the night as there is only solar PV in UY"""
+    if (5 >= dt.hour or dt.hour >= 20) and row.get("value") != 0:
```
pierresegonne, line 96: Isn't that a bit arbitrary ? 😅

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
+@refetch_frequency(timedelta(days=1))
+def fetch_consumption(
+    zone_key: str = "UY",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
+    """collects consumption data from ADME and format all data points for target_datetime"""
+    if target_datetime is None:
+        target_datetime = arrow.utcnow().datetime
```
unitrium, line 155: Shouldn't this be localized here? Because I guess their datasource is running on their timezone ?

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
+        }
+        all_data_points += [data_point]
+    return all_data_points
+
+
+@refetch_frequency(timedelta(days=1))
 def fetch_exchange(
-    zone_key1: str = "UY",
-    zone_key2: str = "BR-S",
+    zone_key1: str,
```
pierresegonne, line 182: If we're fetching some exchanges with different parsers I'd suggest asserting that the given keys don't correspond to the exchange we fetch otherwise, or at least log it

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
-        obj[k] = v
-
-    # https://github.com/tmrowco/electricitymap/issues/1325#issuecomment-380453296
-    salto_grande = get_salto_grande(session)
-    obj["Hidráulica"] = obj.get("Hidráulica", 0.0) + salto_grande
-
-    return obj
-
-
+    link = ADME_URL + date_format + "&fecha_fin=" + next_day_format + "&send=MOSTRAR"
```
unitrium, line 54:
```suggestion
    link = f"ADME_URL{date_format}&fecha_fin={next_day_format}&send=MOSTRAR"
```
It's probably a personal opinion, I tend to find f-strings more readable than concatenation.

parsers/UY.py
VIKTORVAV99: Actually this one will work cause config is the only folder in there and wont work with config.constant as it's the only one that is not a symlink. (and config.constants would point to the other config folder and it don't have constants file).

It's all a little confusing 😅

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
-    response = r.get(url)
-    soup = BeautifulSoup(response.text, "html.parser")
 
-    datefield = soup.find(
-        "span", attrs={"id": "ctl00_ContentPlaceHolder1_lblUltFecScada"}
+def get_adme_url(target_datetime: datetime, session: Optional[Session] = None) -> str:
+    """"""
+    session = session or Session()
+    next_day = target_datetime + timedelta(days=1)
+    date_format = "{day}%2F{month}%2F{year}".format(
```
unitrium, line 47:
```suggestion
    date_format = f"{target_datetime.day}%2F{target_datetime.month}%2F{target_datetime.year}"
```
You can also do this. It's up to you which one you prefer.

unitrium (approved):
Just a small question about the localization of the target_datetime otherwise it looks good to me!

parsers/UY.py
mathilde-daugy: https://sunrise.maplogs.com/uruguay.120.html?year=2022
I had a look at that website for 2022 and thought it was an ok-ish range.
For some reason there is some solar production at night and it's really weird
image

parsers/UY.py
```diff
@@ -1,163 +1,224 @@
...
 
+    data = fetch_data(
+        zone_key=zone_key,
+        session=session,
+        target_datetime=target_datetime,
+        sheet_name="GPF",
+        logger=logger,
+    )
+    data = data.rename(columns=PRODUCTION_MODE_MAPPING)
+    data = data.groupby(data.columns, axis=1).sum()
```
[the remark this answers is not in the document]
mathilde-daugy: after renaming we have multiple hydro colulns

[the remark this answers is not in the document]
unitrium: This will group production modes together. So for instance if you have COALITA and COALITO that are both mapped to COAL they will get summed together. It's a nice trick @mathilde-daugy  showed me on the ELEXON parser.

parsers/UY.py
mathilde-daugy: I didn't localize here because i feel like it's not something that is done in other parser but i may be completely wrong

parsers/UY.py
```diff
@@ -1,163 +1,221 @@
...
+            "datetime": arrow.get(dt).datetime.replace(tzinfo=pytz.timezone(UY_TZ)),
+            "consumption": round(consumption[dt]["consumption"], 3),
+            "source": "pronos.adme.com.uy",
+        }
+        all_data_points += [data_point]
+    return all_data_points
+
+
+@refetch_frequency(timedelta(days=1))
 def fetch_exchange(
```
VIKTORVAV99, line 178: I think it would make sense to use this parser to fetch all or at least the AR exchange as this historical parsing support.
So we should probably update the exchange configs as well.

VIKTORVAV99 (commented):
Pierre and Robin already covered most but I have some (nit-pick) suggestions as well. 😅

parsers/UY.py
mathilde-daugy: I didn't realise that the exchanges from the AR parser were live. In this case, it's better to use this source.

parsers/UY.py
VIKTORVAV99: That is usually dependant on the source data. Most "good" sources are fully iso compliant and use UTC time for the timestamps but this may not be the case here (I have not looked myself).

parsers/UY.py
mathilde-daugy: I will also update the exchanges with BR because it also doesn't parse past dates

VIKTORVAV99:
Before merging this I would like to add the author of #4840 as a contributor in the YAML file and then I'll go ahead an explain the situation and close said PR when this is merged.

VIKTORVAV99:
> Before merging this I would like to add the author of #4840 as a contributor in the YAML file (and do feel free to add yourself as well if you want) and then I'll go ahead an explain the situation and close said PR when this is merged.

mathilde-daugy:
> Before merging this I would like to add the author of #4840 as a contributor in the YAML file (do feel free to add yourself as well if you want) and then I'll go ahead an explain the situation and close said PR when this is merged.

No problem, please do :)

parsers/UY.py
VIKTORVAV99: It's possible this also includes storage from solar generation sites such as battery storage but that's just a guess.

VIKTORVAV99:
> > Before merging this I would like to add the author of #4840 as a contributor in the YAML file (do feel free to add yourself as well if you want) and then I'll go ahead an explain the situation and close said PR when this is merged.
> 
> No problem, please do :)

I added the contributors to the file as marked them as co-authors in my commit so feel free to merge this when it's ready.

parsers/UY.py
mathilde-daugy: I would argue that we should set to 0 if we don't know ( the quantities are small anyways)

pierresegonne: Could it be a timezone issue?

q--: Solar could also include Concentrated Solar Power with energy storage. I found [this 2015 government study](http://www.energiasolar.gub.uy/index.php/investigacion-e-innovacion/investigacion-concentradores) investigating it, but haven't encountered anything they actually went through with it.

Issue #4886, opened 2022-12-20
CA-ON (Ontario) Emission factors


pierresegonne:
**Is your feature request related to a problem? Please describe.**

What appear to be trustworthy emission factors have been computed for Ontario. See [here](https://github.com/electricitymaps/electricitymaps-contrib/discussions/4869) and [here](https://github.com/electricitymaps/electricitymaps-contrib/discussions/4791#discussioncomment-4295383) for discussions.

**Describe the solution you'd like**
Add the identified lifecycle emission factors for CA-ON

* Coal (not applicable, retired in 2014): 1,069gCO2/kWh
* Natural Gas: 497gCO2/kWh
* Hydro: 22gCO2/kWh
* Wind: 10.69gCO2/kWh
* Nuclear: 4.8gCO2/kWh
* Solar: TBD with INCERACV as was done for other zones, see [here](http://viewer.webservice-energy.org/incer-acv/app/)

<img width="705" alt="Screenshot 2022-12-20 at 11 38 42" src="https://user-images.githubusercontent.com/32778266/208647378-ca90315e-ab3a-4f70-bc8a-140da88e2f3e.png">

The source for those should be (except for solar)

`Mallia, Eric, and Geoffrey Lewis. "Life cycle greenhouse gas emissions of electricity generation in the province of Ontario, Canada." The International Journal of Life Cycle Assessment 18.2 (2013): 377-391.`

with a link to https://link.springer.com/article/10.1007/s11367-012-0501-0

MisterClean:
🙋‍♂️ I'll take this one!

VIKTORVAV99:
Sure thing! I'll assign the issue to you and if you have any questions just ping me and I'll see what I can do.

Nil-Cipher:
Does this still need help?

VIKTORVAV99:
Looks like it does as the other PR has gone stale. Feel free to pick this up but please take note of the feedback for the other PR.

PR #4926, opened 2023-01-04
updated lifecycle estimates for CA-ON

MisterClean:
## Issue
https://github.com/electricitymaps/electricitymaps-contrib/issues/4886
## Description
Implements carbon emission intensity estimates for the CA-ON zone, based on this publication
`Mallia, Eric, and Geoffrey Lewis. "Life cycle greenhouse gas emissions of electricity generation in the province of Ontario, Canada." The International Journal of Life Cycle Assessment 18.2 (2013): 377-391.`

- Natural Gas: 497gCO2/kWh
- Hydro: 22gCO2/kWh
- Wind: 10.69gCO2/kWh
- Nuclear: 4.8gCO2/kWh


config/zones/CA-ON.yaml
```diff
@@ -87,33 +87,26 @@ emissionFactors:
...
-      - datetime: '2020-01-01'
-        source: Electricity Maps, 2020 average
-        value: 48.54750352993932
-      - datetime: '2021-01-01'
-        source: Electricity Maps, 2021 average
-        value: 57.47323263502428
+      datetime: '2015-01-01'
+      source: Mallia, E., Lewis, G. "Life cycle greenhouse gas emissions of electricity
+        generation in the province of Ontario, Canada." (2013)
+      value: 22
```
jarek, line 99: `hydro discharge` in Electricity Maps refers to discharge from pumped hydro storage, so I think it's inappropriate to use this number here.

(Extra context: The values Electricity Maps has now are yearly averages for IESO carbon intensity, based on logic that pumped hydro is pumped up with electricity from the grid. In any event, the current Ontario parser doesn't report pumped hydro charging or discharging. If I recall correctly, from available data we might be able to get discharge at the Adam Beck pumped reservoir, but I haven't seen charging data.)

config/zones/CA-ON.yaml
```diff
@@ -87,33 +87,26 @@ emissionFactors:
       - datetime: '2021-01-01'
         source: Electricity Maps, 2021 average
         value: 57.47323263502428
+    gas:
+      datetime: '2015-01-01'
+      source: Mallia, E., Lewis, G. "Life cycle greenhouse gas emissions of electricity
+        generation in the province of Ontario, Canada." (2013)
+      value: 497
```
jarek, line 94: Gas has combined lifecycle and direct emissions. This value is for total (lifecycle + direct) but has been put in the `lifecycle` section. Should this be in `direct` section instead? Or somewhere else altogether? @pierresegonne

config/zones/CA-ON.yaml
```diff
@@ -87,33 +87,26 @@ emissionFactors:
...
-      - datetime: '2021-01-01'
-        source: Electricity Maps, 2021 average
-        value: 57.47323263502428
+      datetime: '2015-01-01'
+      source: Mallia, E., Lewis, G. "Life cycle greenhouse gas emissions of electricity
+        generation in the province of Ontario, Canada." (2013)
+      value: 22
     nuclear:
-      datetime: '2013-01-01'
+      datetime: '2015-01-01'
```
jarek, line 101: The screenshot from the paper in https://github.com/electricitymaps/electricitymaps-contrib/issues/4886#issue-1504346233 says "Summary of 2008 Ontario electricity generation", so shouldn't this be 2008 if anything?

config/zones/CA-ON.yaml
pierresegonne: The reported figure is computed using life-cycle assessment so it's correct to put it in lifecyle. For direct, we could go through the paper, to see if they provide a breakdown of the emission factor, or else remove the upstream emissions from IPCC (120g CO2eq/kWh)

config/zones/CA-ON.yaml
VIKTORVAV99: Judging by the same screenshot you shared it should.

VIKTORVAV99 (commented):
Don't forget to add the source to the top level sources object (close to the bottom of the file).

config/zones/CA-ON.yaml
MisterClean: Good callout - I wasn't sure if the app could accept dates that far in the past, given that electricity data starts after that point

config/zones/CA-ON.yaml
```diff
@@ -87,6 +87,11 @@ emissionFactors:
       - datetime: '2021-01-01'
         source: Electricity Maps, 2021 average
         value: 57.47323263502428
+    gas:
+      datetime: '2008-01-01'
+      source: Mallia, E., Lewis, G. "Life cycle greenhouse gas emissions of electricity \n
+        generation in the province of Ontario, Canada." (2013)
```
VIKTORVAV99, line 93: Why are there newlines in the source? This also don't match the source provided in the top level sources object.

config/zones/CA-ON.yaml
```diff
@@ -211,8 +226,7 @@ parsers:
   price: CA_ON.fetch_price
   production: CA_ON.fetch_production
 sources:
-  ? "Mallia, E., Lewis, G. \"Life cycle greenhouse gas emissions of electricity generation\
-    \ in the province of Ontario, Canada.\" Int J Life Cycle Assess 18, 377\u2013\
-    391 (2013)"
+  ? Mallia, Eric, and Geoffrey Lewis. "Life cycle greenhouse gas emissions of electricity generation in the province of Ontario, Canada."
```
VIKTORVAV99, line 229: This is not formatted correctly, it should follow MLA citation style if I'm not wrong.

Also if you wrap the full key in `''` there is no need to escape double quotes or colons. It should also be on one line as right now it's interpreted as a explicit key with no value and an inplicit key with a explicit value.

VIKTORVAV99 (changes requested):

config/zones/CA-ON.yaml
jarek: I see this has been moved to its own hydro key so this seems fine now

config/zones/CA-ON.yaml
christopheradlam: > Judging by the same screenshot you shared it should.
> The screenshot from the paper in [#4886 (comment)](https://github.com/electricitymaps/electricitymaps-contrib/issues/4886#issue-1504346233) says "Summary of 2008 Ontario electricity generation", so shouldn't this be 2008 if anything?

The date of the paper is 2013 though, they've just used the source mix from 2008 in that table.

VIKTORVAV99:
@MisterClean could you take another look at this PR so we can get it merged?

PR #4929, opened 2023-01-05
Add validation rules to BA in ENTOSE.py

VIKTORVAV99:
## Issue
BA sometimes report 0 for all production modes which seems to break the estimations models.

LT should always have gas production, fixes [ELE-1529](https://linear.app/electricitymaps/issue/ELE-1529/add-validation-rule-for-lithuania)

## Description
Adds validation rules to BA and LT so we only get the data at the quality and granularity we expect.


parsers/ENTSOE.py
```diff
@@ -410,9 +407,7 @@
         "required": ["coal"],
         "expected_range": (5000, 50000),
     },
-    "LT": {
-        "required": ["biomass", "gas", "hydro", "solar", "wind"]
```
FelixDQ: Huh, that is weird. Does this mean that gas was already required? Is there a way to set it so that it must be above 0?

VIKTORVAV99: I think you accidentally just reviewed the formatting commit 😅
It was not required before.

Huh there might be, we our own validate lib and the comment in there says:
```
        expected_range: tuple | dict
            Checks production total against expected range.
            Tuple is in form (low threshold, high threshold), e.g. (1800, 12000).
            If a dict, it should be in the form
            {
            'nuclear': (low, high),
            'coal': (low, high),
            }
```
So I'll change it to the dict based version of the threshold and test it out.
What should the minimum threshold be in your opinion?

VIKTORVAV99:
@FelixDQ I added a expected range for gas, it's currently 1 to 2000 MW. 😅

VIKTORVAV99:
I think we can still keep the required keys as it will only filter out data when it's none (not 0) and just remove the expected range for LT.

FelixDQ:
There are actually 561 events in the database with null gas production. I'm not sure if those are data errors or not 🤔 
I would probably just skip this check for now until we have time to do a deeper dive into the data for Lithuania.

VIKTORVAV99:
If they are after 2015-01-04 I think it is a mistake (all data before are N/A in the entsoe tp)

VIKTORVAV99:
@FelixDQ can we merge this as is or should I remove LT all together?

FelixDQ:
Hey, sorry for the slow response. I was ooo last week. Let's remove LT altogether and investigate it further in the future to ensure that we don't accidentally discard actual data.

VIKTORVAV99:
@FelixDQ I removed the LT validations so it's just the BA ones left now.

FelixDQ (approved):

Issue #4782, opened 2022-11-18
Translations for the new zones in India


VIKTORVAV99:
We recently added new zones for India that represent the power grids instead of the states, this mean we are lacking translations for the new zones.

Language tracker of missing translations:
- [ ] English
- [ ] Arabic
- [ ] Czech
- [ ] Danish
- [ ] German
- [ ] Greek
- [ ] Spanish
- [ ] Estonian
- [ ] Finnish
- [ ] French
- [ ] Hindi
- [ ] Croatian
- [ ] Indonesian
- [ ] Italian
- [ ] Japanese
- [ ] Korean
- [ ] Dutch
- [ ] Norwegian - Bokmaal
- [ ] Polish
- [ ] Portuguese - Brazil
- [ ] Romanian
- [ ] Russian
- [ ] Slovak
- [ ] Swedish
- [ ] Vietnamese
- [ ] Chinese - China
- [ ] Chinese - Hong Kong
- [ ] Chinese - Taiwan

>**Note**
>I will be using this issue as a tracker and won't assign the whole issue to any one individual.

VIKTORVAV99:
> Hi @VIKTORVAV99 ! Please assign it to me
> 
Hi @Cor4zon! 

I'm going to use this issue to track all languages so think I'll assign anyone to the whole issue so if you could just tell me what languages you would to translate I'll add a note to those. 

Just a small note as well:
We strive for the translations to be as authentic as possible so only translate those you feel comfortable doing without translation software like Google translate.

jarek:
What are the names in the native language(s)? Is it like "West India" or "India (West)" or "West Indian Grid" or "West Indian System" or "West Indian ISO" or...?

VIKTORVAV99:
> What are the names in the native language(s)? Is it like "West India" or "India (West)" or "West Indian Grid" or "West Indian System" or "West Indian ISO" or...?

Good question... 😅

@mathilde-daugy do you know the native/original names for these zones?

mathilde-daugy:
> > What are the names in the native language(s)? Is it like "West India" or "India (West)" or "West Indian Grid" or "West Indian System" or "West Indian ISO" or...?
> 
> Good question... 😅
> 
> @mathilde-daugy do you know the native/original names for these zones?

They are called the Northern, Eastern, Western, North Eastern and Southern Grids

VIKTORVAV99:
Then I think we can do Northern India, Southeast India, etc for the English translations.

[...]

PR #4943, opened 2023-01-10
Adds missing translations of zone names

madsnedergaard:
## Description

Just noticed that https://api.electricitymap.org/v3/zones returns a bunch of unknown names because we're still missing English translations for new zones and aggregates. This should ideally not happen, as we're also leaving areas on the map with weird titles :(

Related to #4782 for the Indian zones - I just went ahead and used the names from this Wikipedia: https://en.wikipedia.org/wiki/National_Grid_(India)#:~:text=Individual%20State%20grids%20were%20interconnected,between%20States%20in%20each%20region.

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -288,6 +288,9 @@
     "AT": {
       "zoneName": "Austria"
     },
+    "AU": {
+      "zoneName": "Australia"
+    },
     "AU-NSW": {
       "countryName": "Australia",
       "zoneName": "New South Wales"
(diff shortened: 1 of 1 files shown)


web/public/locales/en.json
```diff
@@ -775,6 +778,24 @@
     "IM": {
       "zoneName": "Isle of Man"
     },
+    "IN": {
+      "zoneName": "India"
+    },
+    "IN-EA": {
+      "zoneName": "Eastern"
```
VIKTORVAV99, line 785: Should we use the style we use for Italy here or the one we use for Norway, Sweden, and Denmark?
I.e. should it be:
```suggestion
      "zoneName": "Eastern India"
```

VIKTORVAV99 (approved):
Just a thought about the naming, we should strive to have a unified naming schemes.

This should be considered non-blocking.

Related issue: [ELE-1574](https://linear.app/electricitymaps/issue/ELE-1574/consistently-name-all-subzones)

VIKTORVAV99 (changes requested):
Oh I just noticed we miss the country name for the India Subzones.

web/public/locales/en.json
madsnedergaard: We're a bit inconsistent here, as Italy for example don't have it:
Screenshot 2023-01-10 at 22 17 50

I personally think it looks better like this (the "North Sweden (Sweden)" part is weird), but we then do need to make sure that the country is always included next to any usage - so I guess that option is at least safer and easier to use :)

VIKTORVAV99 (approved):
Let's get this merged then we can evaluate what format we should use for the subzone names separately.

madsnedergaard:
Thanks Viktor, forgot to push the change :)

PR #4944, opened 2023-01-10
Updates README description and adds link to dashboard

madsnedergaard:
## Description

I think it's time we remove the "built with d3.js and mapbox" as those are by now a tiny piece in the whole infrastructure and stack :) 

Started editing here to add a link to a new [Grafana dashboard](https://electricitymap.grafana.net/public-dashboards/4f775c8021324f5cb8682c4fedd3207e) that shows status of all parsers, to make it easier to know which are working and which are not.


README.md
```diff
@@ -20,6 +23,7 @@ Here are some of the ways you can contribute:
 - [Translating the app](https://github.com/electricitymap/electricitymap-contrib/wiki/Translating-electricitymaps.com)
 - [Updating region capacities](https://github.com/electricitymap/electricitymap-contrib/wiki/Update-region-capacities)
 
+
```
madsnedergaard, line 26:
```suggestion
```

VIKTORVAV99 (approved):
Nice, it's cool we have a dashboard, perhaps we can add exchanges at a later date as well?

VIKTORVAV99:
Should we remove the d3 tag on the repository as well? They could most likely use a update 😅

madsnedergaard:
Good point, would you mind doing so? :)

VIKTORVAV99:
Not at all, any suggestions on what we should add instead (if any)?

madsnedergaard:
> Not at all, any suggestions on what we should add instead (if any)?

Did you do it already? I think these are good enough 😄 

Screenshot 2023-01-17 at 20 02 40

VIKTORVAV99:
> > Not at all, any suggestions on what we should add instead (if any)?
> 
> Did you do it already? I think these are good enough 😄
> 
> Screenshot 2023-01-17 at 20 02 40

I only removed d3, but we could potentially add electricity, map, energy, etc to improve discoverability of the repo but we probably don't want to over do it either.

Fun fact, there is a [electricitymap tag](https://github.com/topics/electricitymap) as well 😅

madsnedergaard:
> I only removed d3, but we could potentially add electricity, map, energy, carbon-emissions, carbon-footprint, etc to improve discoverability of the repo but we probably don't want to over do it either.

Yeah, it's a bit difficult to find the balance here I think. We could also add `python`, `data-science` and `react` - but don't know how useful that really is ¯\\\_(ツ)\_/¯
 
> Fun fact, there is a [electricitymap tag](https://github.com/topics/electricitymap) as well 😅

lol, weird :D

PR #4953, opened 2023-01-18
Add config filename validation to CI

VIKTORVAV99:
## Issue
We are currently not checking if the filenames are valid in the CI.
## Description
Adds some basic validation to make sure all files are uppercase and that the exchange configs use sorted keys.


madsnedergaard (approved):
Nice, this is a great idea! 
Don't know what's best practice or if it's even possible, but maybe we could consider verifying filename/zonekey in the Python test cases itself instead of having a separate script? 🤔 

Don't know if/how `tests/test_config.py` and `tests/test_config_model.py` are working, but would be cool if all requirements/testing were happening in the same way. If it's a bigger task or you don't think it makes sense, I also think this is a great solution and I'll approve it and make it up to you :)

VIKTORVAV99:
> Nice, this is a great idea! Don't know what's best practice or if it's even possible, but maybe we could consider verifying filename/zonekey in the Python test cases itself instead of having a separate script? 🤔
> 
> Don't know if/how `tests/test_config.py` and `tests/test_config_model.py` are working, but would be cool if all requirements/testing were happening in the same way. If it's a bigger task or you don't think it makes sense, I also think this is a great solution and I'll approve it and make it up to you :)

It's not that it don't make sense but the other python jobs comes with a lot of overhead like installing a specific python version and installing all our dependencies. I wanted something quick and lean to just check the filenames as soon as they are modified so I went with this. The execution time for the whole job is just 5 sec compared to the 1 minute or so for the other python jobs.
Reference image:
image

PR #4954, opened 2023-01-18
Add india consumption

mathilde-daugy:
## Description

this update enables us to collect live consumption data from vidyutpravah.in
We will use this data later to create an hourly profile (weekday/weekend) for the in zone to breakdown daily production from CEA/NPP


parsers/IN.py
```diff
@@ -82,6 +119,44 @@ def fetch_production(
...
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    total_consumption = 0
+    for state in STATES_MAPPING[zone_key]:
+        r = session.get(DEMAND_URL.format(state=state))
+        soup = BeautifulSoup(r.content, "html.parser")
+        try:
+            state_consumption = int(
+                soup.find(
+                    "span", attrs={"class": "value_DemandMET_en value_StateDetails_en"}
```
unitrium, line 139: Maybe we should have a specific exception in case we cannot find this attribute in the html. It would mean that we need to update our code to match the new attributes. Otherwise this would be silently ignored and will go under the radar.

unitrium (approved):
A small suggestion to improve the resilience of the parser. Otherwise looks good to me.

parsers/IN.py
```diff
@@ -82,6 +120,47 @@ def fetch_production(
...
+                )
+                .text.strip()
+                .split()[0]
+                .replace(",", "")
+            )
+        except:
+            raise ParserException(
+                parser="IE.py",
```
VIKTORVAV99, line 149:
```suggestion
                parser="IN.py",
```

parsers/IN.py
```diff
@@ -8,10 +8,15 @@
 from typing import Optional
 
 import arrow
+import pytz
 import requests
 from bs4 import BeautifulSoup
 from requests import Session
```
VIKTORVAV99, line 14:
```suggestion
from requests import Response, Session
```

parsers/IN.py
```diff
@@ -82,6 +120,47 @@ def fetch_production(
...
+    logger: Logger = getLogger(__name__),
+) -> dict:
+    """Fetches live consumption from government dashboard. Consumption is available per state and is then aggregated at regional level.
+    Data is not available for the following states: Ladakh (disputed territory), Daman & Diu, Dadra & Nagar Haveli, Lakshadweep"""
+    if target_datetime is not None:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+
+    total_consumption = 0
+    for state in STATES_MAPPING[zone_key]:
+        r = session.get(DEMAND_URL.format(state=state))
```
VIKTORVAV99, line 136:
```suggestion
        r: Response = session.get(DEMAND_URL.format(state=state))
```

VIKTORVAV99 (commented):
Just a few small suggestions and I agree with Robins comment, we should probably check that the HTML element itself exists so we know if we need to change the parser.

PR #4956, opened 2023-01-18
Develop in no zone

mathilde-daugy:
## Description
This IN parser now collects daily values from NPP (conventional thermal, nuclear, hydro) and CEA (VRE + other REN)
The data collected will be stored in one data point (at 00:00). This daily data will then be used in the estimation pipeline to compute a hourly power breakdown from daily totals.

### Preview

image


parsers/IN.py
```diff
@@ -161,6 +219,125 @@ def fetch_consumption(
     return data
```
VIKTORVAV99, line 219:
```suggestion
    if data is None:
        raise ParserException(
            parser="IN.py",
            message=f"{target_datetime}: No valid consumption data found for {zone_key}",
        )
    return data
```
Validate consumption will return None if the consumption is invalid.

parsers/IN.py
```diff
@@ -161,6 +219,125 @@ def fetch_consumption(
...
+        target_datetime = arrow.now(tz=IN_NO_TZ).floor("day").datetime - timedelta(
+            days=2
+        )
+    else:
+        target_datetime = (
+            arrow.get(target_datetime).floor("day").datetime.replace(tzinfo=IN_NO_TZ)
+        )
+
+    renewable_production = fetch_cea_production(
+        zone_key=zone_key, session=session, target_datetime=target_datetime
```
VIKTORVAV99, line 326: There are some type errors flagged here but I think they are safe to ignore based on the logic above. It complains about target_datetime potentially being None but that should not be possible since we replace it if it is.

parsers/IN.py
```diff
@@ -161,6 +219,125 @@ def fetch_consumption(
...
+            production[NPP_MODE_MAPPING[mode]] = round(
+                df_zone.iloc[df_zone.index.get_indexer_for([mode])[0]].get("value")
+                / CONVERSION_MWH_MW,
+                3,
+            )
+        return production
+    else:
+        raise ParserException(
+            parser="IN.py",
+            message=f"{target_datetime}: {zone_key} conventional production data is not available",
```
unitrium, line 268:
```suggestion
            message=f"{target_datetime}: {zone_key} conventional production data is not available : [{r.status_code}]",
```
I would suggest to log the status code, this will allow us a faster troubleshooting.

parsers/IN.py
```diff
@@ -161,6 +219,125 @@ def fetch_consumption(
...
+
+def fetch_cea_production(
+    zone_key: str,
+    target_datetime: datetime,
+    session: Session = Session(),
+    logger: Logger = getLogger(__name__),
+) -> dict:
+    """Gets production data for wind, solar and other renewables
+    Other renewables includes a share of hydro, biomass and others and will categorized as unknown
+    DISCLAIMER: this data is only available since 2020/12/17"""
```
unitrium, line 280: Maybe raise an error if the target time is before this date ?

unitrium (approved):
Looks good to me!

VIKTORVAV99 (commented):
Just have a suggestion and small comment in addition to Robins review.

PR #4961, opened 2023-01-19
update set-output to match the new syntax

VIKTORVAV99:
Github updated how output is set so this PR updates out workflow to match the new syntax.

I also threw in a fix for the CodeQL warnings and fixed the setup step to log the right things.


madsnedergaard (approved):
Nice! 👏

PR #4965, opened 2023-01-22
Fix FR_O.py storage values

VIKTORVAV99:
## Issue
Apparently we consider negative storage values as discharge (production) and positive ones as charge (consumption). This did not match the values provided in the API so this PR inverts the numbers to match our internal structure.

Side note:
Can we look into if we can change this behavior to match the exchange and production structure where positive is discharge (production)?


madsnedergaard (approved):
LGTM!
It would be great to revert it indeed, but I'm afraid it would also be quite some work as all infrastructure right now relies on it being "wrong". Hope it's something we can address at some point, but for now I think there's bigger fish to fry :)

VIKTORVAV99:
> LGTM!
> It would be great to revert it indeed, but I'm afraid it would also be quite some work as all infrastructure right now relies on it being "wrong". Hope it's something we can address at some point, but for now I think there's bigger fish to fry :)

Agreed and if we changed it we would have to go over every parser + the backend + database and change them all at once. Likely freezing development for other features. It should be on the todo list at some point though.

PR #4966, opened 2023-01-23
IN-NO_fix_tz_issue

mathilde-daugy:
## Description

parser didn't run because of time zone awareness


unitrium:
Looks like you have an issue with your commit history, it seems to not be up to date. Could you either do:
- `git pull master` then `git rebase master`
- create a new branch from master and add your fix commit there

PR #4972, opened 2023-01-25
TR_parser_update

mathilde-daugy:
## Description

update TR parser because source is down
this new parser enables us to collect historical data from EPIS

### Preview

image

### DIFF
--- a/parsers/TR.py
+++ b/parsers/TR.py
@@ -55,6 +55,7 @@ def fetch_data(session: Session, target_datetime: datetime, kind: str) -> dict:
 
 def validate_production_data(
     data: list,
+    exclude_last_data_point: bool,
     logger: Logger = getLogger(__name__),
 ) -> list:
     """detects outliers: for real-time data the latest data point can be completely out of the expected range and needs to be excluded"""
@@ -65,7 +66,8 @@ def validate_production_data(
     all_data_points_validated = [
         x for x in data if validate(x, logger, required=required, floor=floor)
     ]
-
+    if exclude_last_data_point:
+        all_data_points_validated = all_data_points_validated[:-1]
     return all_data_points_validated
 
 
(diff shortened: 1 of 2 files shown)


parsers/TR.py
```diff
@@ -1,177 +1,152 @@
...
-                    data["datetime"] = tr_datetime.datetime
-
-                production_by_hour.append(data)
-    else:
-        raise Exception("Extracted data was None")
 
-    return production_by_hour
+    all_data_points = []
+    for item in data:
+        data_point = {}
```
VIKTORVAV99, line 114:
```suggestion
```
I think this is just some duplicate code, either way it's safe to remove.

VIKTORVAV99 (changes requested):
Just a few changes needed to tidy things up but everything is working perfectly!

Nice work!

parsers/TR.py
```diff
@@ -1,177 +1,162 @@
...
-        {
-            "zoneKey": "TR",
-            "currency": "TRY",
-            "datetime": arrow.get(datetime.combine(date, time))
-            .to("Europe/Istanbul")
-            .datetime,
-            "price": price,
+) -> list:
+    if target_datetime is None:
+        target_datetime = datetime.now(tz=TR_TZ) - timedelta(hours=2)
```
pierresegonne, line 82: Is this because there is a delay?

pierresegonne (approved):
Read through the code and looks neat! 
Didn't test though, so I'll leave it to your discretion to do so

parsers/TR.py
mathilde-daugy: actually this was to try to remove the last point of data which was rubbish but it's not necessary since there is data validation in fetch_production so i'll remove

VIKTORVAV99 (approved):
LGTM!
Let's get this merged!

PR #4978, opened 2023-01-27
Update Belgian solar & wind capacities

Fred-3000:
Same sources.


VIKTORVAV99 (approved):
LGTM! Thanks for keeping these updated!

PR #4979, opened 2023-01-28
Fix exchange fetching in parsers/IE.py

msaynevirta:
## Issue

The Ireland parser returns incorrect exchange data for Northern Ireland, as flows in the Moyle interconnector displayed between GB-NIR and IE instead of GB and GB-NIR. 

## Description

Zone key index in fetch_exchange of the IE parser has been corrected. In this case, the second key is the correct one, as the 'GB' source key occupies the first zone key index for both Moyle (GB->GB-NIR) and East-West (GB->IE) interconnectors.

### DIFF
--- a/parsers/IE.py
+++ b/parsers/IE.py
@@ -127,14 +127,17 @@ def fetch_exchange(
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> list:
-    """gets exchanges values for the East-West interconnector (GB->IE)"""
+    """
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

mathilde-daugy:
> I'm going to test this today and try and get it merged as soon as possible.
> 
> @mathilde-daugy would it be possible for us to migrate all the existing data for the `GB-NIR -> IE` exchange to `GB -> GB-NIR` once that is done?

I'll refetch the whole data set once it's deployed :)

PR #4980, opened 2023-01-28
test: fix deprecation warning of escape character in regex

shuuji3:
## Issue

There are several deprecation warnings related to the invalid escape characters during tests like this:
```
DeprecationWarning: invalid escape sequence \d
```

You can see the example here: https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/4031541423/jobs/6930900530

## Description

This expression will be invalid in the future version of Python. This PR fixes those regex-related warnings by using the raw string expression `r'...'`.

Note: I could fix most of the warnings automatically using `pyupgrade` by running the command: `pipx run pyupgrade parsers/JP-KY.py`. This might be useful for other future fixes.

Also, `pyupgrade` suggested using the string format `{:02d}` instead of `{0:02d}`. So I refactored a few expressions related to that in `JP-TK.py` using f-string.

### Preview

N/A


VIKTORVAV99:
I'll try and test this as soon as I have time but at first glance of the code everything looks good.

VIKTORVAV99 (approved):
LGTM! 🎉

shuuji3:
Thanks! 😀

PR #4988, opened 2023-01-30
remove experimental optmisation

tonypls:

### DIFF
--- a/web/tailwind.config.js
+++ b/web/tailwind.config.js
@@ -47,7 +47,6 @@ const config = {
       '2xl': '3rem',
     },
   },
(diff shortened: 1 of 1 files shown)


tonypls:
First PR back on the real repo! 

I believe the tailwind experimental setting is what was causing us the issues that we thought we're chunking related and the netlify previews wouldn't work until I got rid of it.

VIKTORVAV99 (approved):
LGTM! 🎉

If this works we can gradually enable chunking again.

madsnedergaard (approved):

PR #4990, opened 2023-01-30
No more yarn in web

VIKTORVAV99:
Removes the leftover mentions of yarn in /web.


web/README.md
```diff
@@ -11,11 +11,12 @@ Prerequisites:
 - Ensure you have `NodeJS` and `pnpm` installed locally (`brew install pnpm`)
 - Run `pnpm install` in both the web and mockserver directories
 
-1. Start the mockserver: `yarn run mockserver`
-2. Run app in another tab: `yarn develop`
+1. Start the mockserver: `pnpm run mockserver`
+2. Run app in another tab: `pnpm dev`
```
madsnedergaard, line 15: We don't have the command for starting mockerserver from web folder anymore, so let's update this: 
```suggestion
1. Start the mockserver (inside `mockserver` folder): `pnpm run start`
2. Run app in another terminal (inside `web` folder): `pnpm dev`
```

VIKTORVAV99: I updated the scripts so it works mockserver also works from web (like before).

web/package.json
```diff
@@ -11,6 +11,7 @@
     "url": "https://github.com/electricityMaps/electricitymaps-contrib.git"
   },
   "scripts": {
+    "mockserver": "nodemon ../mockserver/server.js",
```
madsnedergaard, line 14: This would require nodemon being installed globally - let's call it `cd ../mockserver && pnpm run start` (but cross-platform friendly?) like before - and maybe convert the mockserver from yarn to pnpm?

web/README.md
madsnedergaard: Oh, didn't see you added it - nevermind then 😄

.github/labeler.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - mockserver/package.json
   - mockserver/yarn.lock
   - web/package.json
-  - web/yarn.lock
+  - web/pnpm-lock.yaml
```
madsnedergaard, line 14: More like this I think:

Screenshot 2023-01-30 at 22 38 47

madsnedergaard:
And one more:
Screenshot 2023-01-30 at 22 40 27

Screenshot 2023-01-30 at 22 40 01

web/package.json
VIKTORVAV99: Hmm I wonder why it worked for me then, I don't have a global version of nodemon installed.
You sure that's needed when it's PNPM doing the command? If it works like NPM it should pick up the local version (could be wrong tho).

VIKTORVAV99:
> And one more: Screenshot 2023-01-30 at 22 40 27
> 
> Screenshot 2023-01-30 at 22 40 01

I didn't add this as prettier will format this file as it's yaml. Should we add ignores to both then?

madsnedergaard:
> VSCode settings screenshot
> I didn't add this as prettier will format this file as it's yaml. Should we add ignores to both then?

It's just hiding files in the explorer on VSCode :)

VIKTORVAV99:
> > VSCode settings screenshot
> > I didn't add this as prettier will format this file as it's yaml. Should we add ignores to both then?
> 
> It's just hiding files in the explorer on VSCode :)

Oh...
It was right under the formatting options so I just assumed it was for those 😅

web/package.json
madsnedergaard: Weird, it actually works even if I delete `node_modules` in `mockserver` - so don't know where/how it's using `nodemon` 🤔 
I think we should switch from yarn to pnpm in mockserver no matter what though, so we're consistent with package managers :)

VIKTORVAV99: I think we can keep it as is then and if it ends up breaking I'll fix it at that point.

VIKTORVAV99: Update to this, Node 19 added a `--watch` flag which can replace nodemon, while we can't and shouldn't update node to v19 now it would solve the issue in the future and allow us to remove nodemon entirely.

madsnedergaard (approved):

PR #4992, opened 2023-01-31
Add validator for hydro in US-CAR-YAD

pierresegonne:
## Issue

ELE-1421 (in our internal Linear)

-> The hydro production in `US-CAR-YAD` reaches unrealistic levels. Only hydro is on the production grid for that zone, which exports a good portion of it. So that creates cases where flowtracing fails

## Description

* Adds a validator that marks all events with hydro production < 5 MW as invalid (see comment in code for why it's very unlikely)


VIKTORVAV99 (approved):
LGTM! 🎉

Should we add upper limits as well or is it only the lower bound that was a issue?

pierresegonne:
> Should we add upper limits as well or is it only the lower bound that was a issue?

Only the lower band was

PR #4993, opened 2023-01-31
Archive IN zones

pierresegonne:
## Issue

We still have traces of all the old `IN` zones.

I propose we do a cleanup of these references.

This comes as I'm looking at the priority zones that are down, and `IN-MH` is triggering the alert even though we should not reference it anywhere

## Description

* Moves zones and exchanges config to their respective `retired_*` folder
* Moves parsers for retired zones to `archived`

Still to do, but wanted to verify with you that we indeed want to remove those zones
* Cleanup locales
* Cleanup mocks 

Funny enough, the old zones are still visible on the map 

Screenshot 2023-01-31 at 10 21 45


mathilde-daugy:
Initially we had kept the parser because we thought it would interesting to keep collecting data from the states but i guess it doesn't make sense if we can stil view it on the app

VIKTORVAV99:
Have we reached any decisions on this?
It would be good to get the parsers archived soon and determine what to do with the configs as they are causing some minor issues with contributors "fixing" the old parsers even though we no longer use them.

madsnedergaard:
Replaced by #5492 that is more extensive :)

PR #4997, opened 2023-01-31
mobile all release 1.5.1

tonypls:
## Issue
Release candidate for iOS and Android.

## Description
Is this the version schema we want to follow?


VIKTORVAV99:
As for the schema I think all bugfixes, capacity and parser updates should be x.x.* releases and all changes to the frontend itself should be x.*.x releases if that makes sense.

madsnedergaard (approved):
Yay! And plus one on versioning, I think that makes good sense for now :)

PR #5001, opened 2023-02-01
Add US-CENT-SPP exchanges 

mathilde-daugy:
## Description
Collects real-time and historical data for exchanges from SPP at 1 min intervals
This will enable us to show the exchanges on the map

### Preview
image


parsers/US_SPP.py
```diff
@@ -297,6 +328,109 @@ def fetch_wind_solar_forecasts(
...
+def fetch_historical_exchange(
+    zone_key1: str,
+    zone_key2: str,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
+
+    filename = target_datetime.strftime("TieFlows_%b%Y.csv")
+    file_url = HISTORICAL_EXCHANGE_URL + filename
```
unitrium, line 361: It's more safe to use either f strings or .format.

parsers/US_SPP.py
```diff
@@ -297,6 +328,109 @@ def fetch_wind_solar_forecasts(
...
+
+    all_data_points = []
+    for dt in data_filtered.index:
+        data_dt = data_filtered.loc[data_filtered.index == dt]
+        data_point = {
+            "sortedZoneKeys": sorted_zone_keys,
+            "netFlow": round(data_dt.values[0], 4),
+            "datetime": arrow.get(dt).datetime,
+        }
+        all_data_points += [data_point]
```
unitrium, line 401:
```suggestion
        all_data_points.append(data_point)
```
The way it's written is confusing, I think this would result in a list of list of datapoints. Is it the intention?
Like it would be all_data_points = [[{dp1}], [{dp2}]] while I think we rather want [{dp1}, {dp2}].

parsers/lib/validation.py
```diff
@@ -129,6 +129,27 @@ def validate_consumption(
...
 
 
+def validate_exchange(
+    datapoint: Dict, logger: Union[Logger, None]
+) -> Union[Dict[str, Any], None]:
+    """
+    Validates a production datapoint based on given constraints.
+    If the datapoint is found to be invalid then None is returned.
+    """
+    if logger is None:
```
unitrium, line 139: We can put this in the function's signature as you did for format_exchange_data.

unitrium (approved):
A couple of suggestion, but it already looks great!

parsers/US_SPP.py
VIKTORVAV99: f strings would be the preferred way as they are slightly more performant.

PR #5003, opened 2023-02-01
Adds earthfile

madsnedergaard:
## Description

Adds Earthly deployment file back in.


unitrium (approved):
After this we won't need geo-files and src-files in the Earthly api file right ? :)

madsnedergaard:
> After this we won't need geo-files and src-files in the Earthly api file right ? :)

Nope, it should only be the `api-files` target :)

PR #5009, opened 2023-02-02
Reverts name change as internal packages rely on it

madsnedergaard:
## Description

Turns out updating all the Poetry lockfiles for internal libraries and projects was harder than anticipated. So for now we change the name back here :)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "electricitymaps-contrib"
+name = "electricitymap-contrib"
```
tonypls, line 2: This is the fix?

tonypls (approved):
LGTM

PR #5010, opened 2023-02-02
Updates inputs to ensure all geo files are available

madsnedergaard:
## Description

Aaaand one more - I'm 90% certain this will be the last change that makes the app deploy 😅


VIKTORVAV99 (approved):
It looks good but if it works is another question 😅

PR #5011, opened 2023-02-02
Add ERCOT real time data

ghost:
## Description

Add real time production data for Texas. 
The generation data is available on the TSOs dashboard. it enables us to fill in the gap between EIA's latest available data point and this way we don't need to estimate production anymore.
This parser collects real time data continuously but is able to fetch historical data from the EIA dataset. As ERCOT publishes their data on the EIA the data should be the same. Looking at the table below, we can see that the difference are very small between the two sources and probably come from data aggregation as EIA publishes only one data point per hour and ERCOT has data available at a more granular level.

It seems like EIA doesn't report battery storage unlike ERCOT. Looking at 2 data sets, i think that storage is aggregated as unknown in the EIA data. 

**Should we put the ERCOT battery storage in unknown to be consistent with the EIA ?**

NB: the table below shows a comparison for 02/02/2023 average between 5 and 23 because 0-4 is not available from ERCOT 

  |   | production.coal | production.gas | production.hydro | production.nuclear | production.unknown | production.solar | production.wind | storage.battery
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
01/02/2023 5-23 | eia.gov | 11345 | 39759 | 17 | 5125 | 70 | 486 | 2819 |  
01/02/2023 5-23 | eroct.com | 11334 | 39750 | 16 | 5125 | 2 | 485 | 2833 | 58

### Preview

image


parsers/US_ERCOT.py
```diff
@@ -0,0 +1,208 @@
...
+        data_dict = {**data_dict, **data_json[key]}
+    all_data_points = []
+    for item in data_dict:
+        production = {}
+        storage = {}
+        dt = arrow.get(item).datetime.replace(tzinfo=TX_TZ)
+        for mode in data_dict[item]:
+            value = data_dict[item][mode]["gen"]
+            if mode == "Power Storage":
+                storage[GENERATION_MAPPING[mode]] = value
```
VIKTORVAV99, line 92: Keep in mind that we treat battery values pretty werid and negative values are discharge and positive ones charge. I fell into this trap in https://github.com/electricitymaps/electricitymaps-contrib/pull/4965

So I think this should be:
```suggestion
                storage[GENERATION_MAPPING[mode]] = value * -1
```

Documentation from the example parser:
```python
    for item in obj["storage"]:
        # Positive storage means energy is stored
        # Negative storage means energy is generated from the storage system
        data["storage"][item["key"]] = item["value"]  # Should be a floating point value
```

VIKTORVAV99 (changes requested):
Code wise this looks good but we treat storage weird so we need to fix that so we get the correct values.

parsers/US_ERCOT.py
mathilde-daugy: Actually i'm wondering if we should aggregate storage here as unknown to be consistent with the EIA data? from analysing the 2 data sources i think that this share goes in unknown in the EIA.

VIKTORVAV99: I would say we should put it in storage because of 3 main reasons:

1. We should strive to have as little unknown data as possible and (battery) storage will only grow.
2. Having it separate would allow us to change how we calculate storage emissions in the future without re-fetching a bunch of data. Right now they are inherently flawed as they use a yearly average. (I have some ideas around this but that's another issue).
3. It would allow us to show that battery storage is actually used to balance the grid in the app.

mathilde-daugy: in this case should we also make sure that when we backfill with EIA data, this battery storage is not double counted in unknown? Not sure how to proceed here though

VIKTORVAV99: Oh right, so for compatibility sake it might be better to just put it in the unknown data anyway as double counting it would invalidate my above points anyway. 😅

The only other solution would be if we where able to "deduct" the battery values from the unknown parts of the EIA data but I thinks that's something we would have to do in the backend and would most likely require bigger changes.

VIKTORVAV99: > Oh right, so for compatibility sake it might be better to just put it in the unknown data anyway as double counting it would invalidate my above points anyway. 😅
> 
> The only other solution would be if we where able to "deduct" the battery values from the unknown parts of the EIA data but I thinks that's something we would have to do in the backend and would most likely require bigger changes. (This would have massive benefitss for other zones as well, such as SE-SE4 where we could use the entso-e per plant data to get oil production but it's published on a 5 day delay)

mathilde-daugy: yes maybe we could apply a very simple estimation after with a simple condition like if source==eia.org then split unknown as unknown + battery storage. I'll discuss this with Robin and we'll see next week :)

unitrium: As discussed with Mathilde, since this is a very little generation, for now we agreed that putting it in unknown is easier. With our current code base, if we refetch historical data, we should introduce a static attribution for this zone that a specific share of the unknown goes towards battery storage. Which might be wrong from time to time. Therefore, it's probably more correct to always put it in unknown for now until we can build something better.

VIKTORVAV99: Sounds good, I'll change my review once the changes have been made.

parsers/US_ERCOT.py
```diff
@@ -35,7 +35,7 @@
     "Natural Gas": "gas",
     "Nuclear": "nuclear",
     "Other": "unknown",
-    "Power Storage": "battery",
+    "Power Storage": "unknown",  # we lose this information in the EIA parser and have no easy way to split it when we run fetch_production for past dates. As it is a minor share of the production mix, it will be categorized as unknown
```
VIKTORVAV99, line 38: Just to be clear here, is this always positive or do we need to make sure it's not negative so it reduces the unknown production? (or how does EIA do it)

mathilde-daugy: looks like they don't account for the storage part and only what is injected on the mix

VIKTORVAV99 (approved):

parsers/test/test_ERCOT.py
```diff
@@ -0,0 +1,74 @@
...
+from arrow import get
+from testfixtures import LogCapture
+
+from parsers import US_ERCOT
+
+
+class TestUSERCOT(unittest.TestCase):
+    """Patches in a fake response from the data source to allow repeatable testing."""
+
+    def test_fetch_production(self):
```
unitrium, line 22: Since you are using the TestCase class, I would recommend using the adaptor and session for passing patches, it's a bit cleaner.
Have a look here for inspiration:
https://github.com/electricitymaps/electricitymaps-contrib/blob/2005e686ed2a27a1e6af815f20afb0abd6981858/parsers/test/test_NTESMO.py#L14

unitrium (approved):

VIKTORVAV99:
Looks like we need to update the tests before we can merge this.

mathilde-daugy:
> Looks like we need to update the tests before we can merge this.

also, we're having some issues with the proxy so we need to find a solution :)

VIKTORVAV99:
> > Looks like we need to update the tests before we can merge this.
> 
> also, we're having some issues with the proxy so we need to find a solution :)

I'll add the blocked label then, feel free to remove it once the proxy situation resolves.

VIKTORVAV99:
> @mathilde-daugy @VIKTORVAV99 Did y'all every resolve the proxy situation here? If not, I am happy to engage ERCOT to see if they'll allowlist your parser IP(s) or host an EM-only US proxy. Let me know when time permits!

Not sure, but I'll make sure we take another look at it!

VIKTORVAV99:
@FelixDQ could you take a look at this? I cleaned up some of the basic changes but it needs another look.

I'm also a bit worried about the fact that the realtime data source returns 5 minute intervals of data while the historical return hourly. If we don't aggregate the 5 minute data to hourly in the parser there is a real risk we will end up with a mixed mess of data.

It's working for me locally as well so I'm not sure we need to go with the proxy route anymore.

FelixDQ:
Yeah, we probably need to ensure that real-time and historical data are consistent with each other. But otherwise, this seems like a nice improvement to the data

FelixDQ:
I see that ERCOT has an [official API](https://developer.ercot.com/). It looks like the endpoint only exposes solar and wind at the moment. I expressed our interest for getting the data through the API [here](https://github.com/ercot/api-specs/discussions/41), as I think that we should use the official API if possible.

VIKTORVAV99:
This has been implemented in https://github.com/electricitymaps/electricitymaps-contrib/pull/7447

PR #5013, opened 2023-02-02
trigger useEffect on map loaded to fill colors

tonypls:
## Issue
Opening the app in a background tab would load a gray map
## Description
This PR adds a isLoadingMap trigger to the useEffect responsible for painting the map, this way when the map is loaded it will then fill the clickable zones with their colors.


madsnedergaard (approved):
Oh, was it really this easy? Hurray 🥳

tonypls:
> Oh, was it really this simple? Hurray 🥳

Almost a 1 liner :)

PR #5018, opened 2023-02-02
Bumps Node version to 18 everywhere

madsnedergaard:
## Description

Bumps Node version. 

Version 14 is harder to install on M1 Macbooks and therefore it's harder and slower to test things locally if we stay on v14.


VIKTORVAV99 (approved):
Let's go!

PR #5021, opened 2023-02-03
Fixes breaking builds!

madsnedergaard:
## Description

The new app is currently disabled because it would occasionally create a build that was broken due to the way modules got bundled.
I also don't know why the boilerplate we used set it up like this, so reverting to the default Vite approach is probably also better :)

Also:
- Enabled chunk mapping for more modules
- Made error message visible in dark mode
- Fixed a bunch of smaller things in the deployment script


VIKTORVAV99 (approved):
Not really sure how to test the Earthfile changes locally but the rest of them look good.

PR #5022, opened 2023-02-03
Fix ESLint part 1

VIKTORVAV99:
First part of actually fixing ESLint errors.


web/package.json
```diff
@@ -187,7 +187,6 @@
   "lint-staged": {
     "*": "prettier -uw --cache",
     "*.{ts,tsx}": [
-      "eslint --cache --fix",
```
VIKTORVAV99, line 190: This prevented me from pushing the code until all ESlint errors are fixed. Will add it back once they are fixed.

web/.prettierignore
```diff
@@ -9,6 +9,7 @@ dist/
 config/
 public/images/
 .husky/_
+pnpm-lock.yaml
```
VIKTORVAV99, line 12: Not ESLint related but figured I might as well include it.

web/scripts/generate-index-files.ts
```diff
@@ -13,13 +13,13 @@ const allPaths = [...generalPaths, ...zonePaths];
 let filesGenerated = 0;
```
madsnedergaard, line 13: This file is deleted :)

madsnedergaard (approved):
LGTM

web/scripts/generate-index-files.ts
VIKTORVAV99: I'll update the PR to reflect the current state 👍🏼

PR #5023, opened 2023-02-03
Migrates from GCP Bucket to Cloudflare Pages for hosting the app

madsnedergaard:
## Description

The ideal hosting setup is not great and we're lacking the features from platforms serving just one purpose (rollback, cache management, etc.). And then there's the weird workaround we had to do in order to not hit 404s for index files.

I've been playing around with migrating to [Cloudflare Pages](https://pages.cloudflare.com/) and after setting things up it's working really well.

So I think we're ready to just switch over, so we can get the new app back up :)

This also means we get both normal hosting and previews builds from the same place - here's one for this PR: https://mn-migrate-to-cloudflare-pag.electricitymaps-app.pages.dev/map

### Preview

The domain https://staging.electricitymaps.com/ is served using Cloudflare Pages, and it should also create preview builds on new PRs.


web/src/features/panels/LeftPanel.tsx
```diff
@@ -17,26 +17,6 @@ import RankingPanel from './ranking-panel/RankingPanel';
...
-          pathname: pathWithoutIndex,
-          search: searchParameters.toString(),
-        }}
-      />
-    );
-  }
-
-  return children;
-}
-
```
VIKTORVAV99, line 39: Honestly I'm glad to see this is gone 😅

VIKTORVAV99 (approved):
This together with #5021 is giving us some clean module loading and a easy way to test it!
image
(Before we loaded 4 in parallel in the beginning (after index)).

web/src/features/panels/LeftPanel.tsx
madsnedergaard: Haha yeah same, is was quite annoying that we had to add this 😬

PR #5026, opened 2023-02-04
WIP: Adds new workflow for running Lighthouse

madsnedergaard:
Just testing if this is feasible and worth it :)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -104,6 +104,14 @@ jobs:
       needs.setup.outputs.workflows != 'false'
     uses: ./.github/workflows/unit_test.yml
 
+  lighthouse:
+    name: Lighthouse
+    needs: [setup]
+    if: >-
+      needs.setup.outputs.frontend != 'false' ||
+      needs.setup.outputs.workflows != 'false'
+    uses: ./.github/workflows/lighthouse.yml
+
   # eslint:
   #   name: ESLint
   #   needs: [setup, prettier]
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -1,0 +1,23 @@
+on:
+  workflow_call:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Await CF Pages
+        id: await-deploy
+        uses: WalshyDev/cf-pages-await@v1
+        with:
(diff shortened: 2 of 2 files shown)


VIKTORVAV99:
Lighthouse actually have their on CI action we could potentially use at https://github.com/GoogleChrome/lighthouse-ci, I've used it in a personal project once and it worked pretty good.

madsnedergaard:
Ah cool, missed that one 😅
It does seem like it requires managing/setting up more infrastructure compared to the unofficial action used here - is that correct? Anyway, it didn't work out of the box and I was a bit unsure if it's nice or just a bit noisy as most PRs will not affect the score. So maybe we should just run a test manually if it matters instead? 😅

VIKTORVAV99:
> Ah cool, missed that one 😅 It does seem like it requires managing/setting up more infrastructure compared to the unofficial action used here - is that correct? Anyway, it didn't work out of the box and I was a bit unsure if it's nice or just a bit noisy as most PRs will not affect the score. So maybe we should just run a test manually if it matters instead? 😅

No not really, that is only for historical comparisons and long term trends but you can definitely run it without it and just use the public temporary storage.

I have been running all the PR's affecting the build systems manually already but it would be nice to have some sort of automated system that is consistent across PR's/environments so we aren't just measuring local performance.(The new preview builds help with that.)

madsnedergaard:
Yeah that would be nice. Can't remember where I documented this idea (probably one of my multiple personal note-taking apps), so sharing it here for you too: I think it would be nice to have a bunch of "checks" (LH score, test coverage, type coverage, etc.) that validate that we are not decreasing any of those numbers. So it would only post a comment if there was a problem ("X is now lower than it was is on `master`"), instead of spamming every PR that might be unrelated with the values :)

VIKTORVAV99:
> Yeah that would be nice. Can't remember where I documented this idea (probably one of my multiple personal note-taking apps), so sharing it here for you too: I think it would be nice to have a bunch of "checks" (LH score, test coverage, type coverage, etc.) that validate that we are not decreasing any of those numbers. So it would only post a comment if there was a problem ("X is now lower than it was is on `master`"), instead of spamming every PR that might be unrelated with the values :)

You did put some of these ideas as draft issues in the project 😅
But yes it would be nice to have a robust regression test suit, GitHub has some nice features for putting it in a summary field in the CI tab we can explore as well, like how the Cypress action does it.

PR #5027, opened 2023-02-05
Enable PWA + icons for everything

VIKTORVAV99:
Enables the Service Worker and updates it's configuration. It also adds every icon you can imagine for iOS, Android and Windows 11.

After some local testing I can see that the first page load (new user) is a little longer (not really noticeable) and download a little more data. But repeat visits are now under 50ms total (the longest task now is waiting for the backend to return any data) and don't download anything except the state data.


madsnedergaard:
Oh wow, that is a lot of files 😅 
I don't think we need all of these extremely specific sizes - when I researched it earlier on, I added the only ones that should be required for "modern" browsers. Where did you see that we need all of these?

VIKTORVAV99:
> Oh wow, that is a lot of files 😅 I don't think we need all of these extremely specific sizes - when I researched it earlier on, I added the only ones that should be required for "modern" browsers. Where did you see that we need all of these?

I just threw our biggest android image into https://www.pwabuilder.com/imageGenerator (By microsoft) and added the output 😅

tonypls:
I think it will be great to get PWA out there and should help reduce some errors we've seen in Sentry but I don't think we're quite ready. Without the manual unsubscribing to the service work it will make rolling back to the previous app take ~24 hours. 

I think it's best if we wait until we've had a few weeks of stable releases until we take this step.

VIKTORVAV99:
> I think it will be great to get PWA out there and should help reduce some errors we've seen in Sentry but I don't think we're quite ready. Without the manual unsubscribing to the service work it will make rolling back to the previous app take ~24 hours.
> 
> I think it's best if we wait until we've had a few weeks of stable releases until we take this step.

I don't see why it would take that long? The service worker itself should detect any changes to it's config and invalidate all affected caches and fetch from the network again. It should in theory take no more than our current CDN cache at 4 hours and could be even quicker as we have must revalidate cache control headers?

madsnedergaard:
> > I think it will be great to get PWA out there and should help reduce some errors we've seen in Sentry but I don't think we're quite ready. Without the manual unsubscribing to the service work it will make rolling back to the previous app take ~24 hours.
> > I think it's best if we wait until we've had a few weeks of stable releases until we take this step.
> 
> I don't see why it would take that long? The service worker itself should detect any changes to it's config and invalidate all affected caches and fetch from the network again. It should in theory take no more than our current CDN cache at 4 hours and could be even quicker as we have must revalidate cache control headers?

The issue here is that if something breaks and clients have a service-worker with local caching, it's going to be harder to switch DNS records back to the old app (as far as I can tell). That's why it would be nice to wait until we're a bit more certain that everything works as expected 😅

tonypls:
We don't want to switch back :) Having a fallback is really nice though!

VIKTORVAV99:
It's been another month now and I think we are comfortable enough with the new app now so can we proceed with this PR?

I can reduce the number of icons if needed (I may have gone a tiny bit overboard with them 😅) but beyond that I don't think there is anything else in the PR itself that should be blocking (except the merge conflicts but I'll take care of those when we go forward with this).

VIKTORVAV99:
> > but beyond that I don't think there is anything else in the PR itself that should be blocking (except the merge conflicts but I'll take care of those when we go forward with this).
> 
> As discussed on our 1:1 today, I still think we should hold back on this until releasing is automated and we're confident that we can easily roll back. We could however remove the caching part of this PR and then get the first version merged in right away, what do you think?

Yeah I think just making it installable is a good first step, I'll look into if we can use it without caching with the current approach or if we need to use a new one. But it's definatly possible.

VIKTORVAV99:
I reduced the amount of icons but keeping all sizes for ios and android as they are our main platforms and most likely to be installed (I think?).

I also looked into what we can do with the service worker about the caching and it looks like using vite-pwa with the injectManifest option and building our own service worker is the way forward (easier than it sounds). Then we can use custom strategies for caching and I propose we use either [Network first, falling back to cache](https://developer.chrome.com/docs/workbox/caching-strategies-overview/#network-first-falling-back-to-cache) or [Stale-while-revalidate](https://developer.chrome.com/docs/workbox/caching-strategies-overview/#stale-while-revalidate)

The first service worker strategy will only serve assets from cache if the network is unavailable. Easy to implement and get the benefits of caching when there is no network.

The second one uses a hybrid approach where it prioritizes performance by serving from cache first and then revalidating the asset at the same time so it's updated the next time it's requested. This one is the most advanced of the two and might be the best option long term providing the most benefits while still keeping the content up to date.

There is a third approach we could create that just use the precache option to serve the static assets from the service worker and passing through all other requests to the network, this should be relatively safe even if it prioritizes the cache since it should automatically update when the app bundle is updated.

All of them use the cache in some way or another (the whole point of a service worker) so there is some downsides to all of them if we need to revert the app. But I think it should be relativly safe to move forward with it as all the precached as they are hashed and the static assets are more or less the same between the versions (icons, images, languages). (if they have different names or locations they will fetch directly from the network).

VIKTORVAV99:
Closing this for now and will create a new one when we want to enable this.

PR #5033, opened 2023-02-06
add field to exclude price data for nordpool / epex zones

tonypls:
## Issue
We are unable to redistribute price data that may have been sourced from Nordpool or Epex
## Description
This PR adds a field to the zone config of zones that price data could have been sourced from Nordpool or Epex via ENTSOE. 

We will add app backend updates to filter price data out of these zones


VIKTORVAV99:
@madsnedergaard what was the response about the aggregated price? Where we still allowed to show that?

madsnedergaard:
> @madsnedergaard what was the response about the aggregated price? Are we allowed to show that?

Unfortunately not, so we will disable the graph on all time aggregates :)

VIKTORVAV99:
> > @madsnedergaard what was the response about the aggregated price? Are we allowed to show that?
> 
> Unfortunately not, so we will disable the graph on all time aggregates :)

That's unfortunate. Hopefully we can find a solution in the future.

madsnedergaard (approved):

PR #5037, opened 2023-02-06
Parser datapoint validation

mathilde-daugy:
## Description
This function enables the validation of each parser datapoint before uploading. It makes sure that the datapoint are compliant with our format.
This will prevent us from merging typos and all parsers should be able to load data.

### Preview

NA

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -172,22 +172,3 @@ def validate_production(obj: Dict[str, Any], zone_key: ZoneKey) -> None:
             )
 
     validate_reasonable_time(obj, zone_key)
-
-
-
-def validate_data_point(datapoint: Dict[str, Any], kind:str, zone_key: ZoneKey):
-    """
-    Checks that a datapoint has the required keys. A parser can only be merged if the datapoints for each function have the correct format.
-    """
-    standard_keys = ["datetime","source"]
-    keys_dict= {"production":["zoneKey", "production"]+standard_keys,
-                "consumption":["zoneKey", "consumption"]+standard_keys,
-                "exchange":["sortedZoneKeys", "netFlow"] +standard_keys,
-                "price":["zoneKey","currency","price"]+standard_keys}
-    if  datapoint.keys() != keys_dict[kind] :
-        raise ValidationError(
-              "{} - data point does not have the required keys:  {} is missing".format(zone_key,
-                [key for key in keys_dict[kind] if key not in datapoint.keys()]
-            ),
-        )
-
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -15,6 +15,27 @@ class ValidationError(ValueError):
     pass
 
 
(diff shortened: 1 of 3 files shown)


FelixDQ (commented):
The naming is a bit weird, but can you instead put this into `parsers/lib/quality` and call it in the different `validate_production`, `validate_exchange` etc. This way it will also be called from feeder-electricity

VIKTORVAV99:
> The naming is a bit weird, but can you instead put this into `parsers/lib/quality` and call it in the different `validate_production`, `validate_exchange` etc. This way it will also be called from feeder-electricity

Would it not be better to have it in `validators/sanity_checks.py`? with a validator decorator function?
Decorator function documentation:
```python
def validator(
    kind: str, zone_keys: List[str] = None, not_zone_keys: List[str] = None
) -> Callable:
    """
    Decorator function to mark a function as a validator.
    The backend will run all functions marked as validators.

    A validator function is expected to return as a pandas Series of values between 0 and 1,
    where 1 means that no problems was found and 0 means that the data point is invalid.
    Floating points between 0 and 1 can be used to indicate suspicious data points.

    Parameters
    ----------
    kind : str
        The kind of data that validator supports. e.g. "production", "exchange"
    zone_keys : List[str]
        If this is included, the validator will only be run for the given zone keys.
    not_zone_keys : List[str]
        If this is included, the validator will run for ALL OTHER zones.
    """
```

FelixDQ:
That will be part of the QA layer. The function is not currently used in production. So for now we can only change `parsers/lib/quality` as that is the function being used in feeder-electricity and test_parser

VIKTORVAV99:
> That will be part of the QA layer. The function is not currently used in production. So for now we can only change `parsers/lib/quality` as that is the function being used in feeder-electricity and test_parser

Oh this is not even used at the moment? Maybe we should slip that into the comment... 😅

VIKTORVAV99:
> Nope, it's WIP. We just defined the interface which we will use.

Oh I already sent some feedback around it to Mathilde and Robin earlier today under the assumption that it was done. I'll just add it here as well real quick (even if it's slightly off topic).

> And I do think a new validation function is the way forward. But as far as I know we are restricted by the kind field right now as we would like to run this particular validation on all parser events (except exchange perhaps) without creating 4 or 5 duplicate functions. Perhaps we can make kind accept an array of values similar to zone_keys? (could be a future improvment)

mathilde-daugy:
> Sorry for the slightly off topic comments in here but the changes looks good to me and it worked great when I tested it.

No worries, it is true that there are things all over the place, we'll let you know when we resume working on the QA layer

madsnedergaard:
I'm a bit late to the party, but was there any reason for this approach vs having stronger typing and a parser model with Pydantic like we have for a zone configuration (assuming that it's possible, I might be wrong on that)?

madsnedergaard:
Ah okay, fair enough - thanks for explaining :)

PR #5038, opened 2023-02-06
Update nuclear capacity for SK to include Mochovce-3

### LINKED ISSUE #5035: Nuclear capacity for Slovakia needs to updated to include Mochovce units 3 and 4
The nuclear capacity for Slovakia needs to be extended to accommodate for the new power units 3 and 4 at Mochovce nuclear power plant. Unit 3 first synchronized with the grid on January 31, 2023 [1], unit 4 is supposed to follow one or two years later.

Currently, Electricity Maps shows the nuclear generation in Slovakia to be at 102% which is a result of the outdated information regarding the installed generation capacity for nuclear power in the country, see screenshot below:

em-bug-20230206

> [1] https://www.world-nuclear-news.org/Articles/Mochovce-3-supplies-first-electricity-to-grid

glaubitz:
Fixes #5035

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -353,6 +353,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
   - Solar: [Energy Market Authority Statistics](https://www.ema.gov.sg/statistic.aspx?sta_sid=20170711hc85chOLVvWp)
   - Other: [Energy Market Authority](https://www.ema.gov.sg/statistic.aspx?sta_sid=20140730XofavKNX5Ti7)
 - Slovakia:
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
> OK, will have a look later today. Wasn't aware of this since this is my first contribution.

All good, take as long as you need!

glaubitz:
Should be good now. I am using IAEA PRIS as the reference now which list a total nuclear net capacity of 2308 MW.

It seems that the previously used source was using the gross capacity which is inaccurate.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for adding the source!

PR #5039, opened 2023-02-06
display correct axis on ff

### LINKED ISSUE #5034: Fix chart labels in Firefox
Image

tonypls:
## Issue
Fixes #5034

## Description
The time axis was not showing correctly in firefox, this PR makes it aligned correctly in chrome and firefox

### Preview
Before:
image

After:
image

### DIFF
--- a/web/src/features/time/TimeAxis.tsx
+++ b/web/src/features/time/TimeAxis.tsx
@@ -94,12 +94,6 @@ function TimeAxis({
     );
   }
 
-  // The svg transform attribute is different from the CSS transform property,
-  // so we need to replace the commas with spaces and remove the 'px' suffix.
-  const svgTransform = transform
-    ? transform.replace(/px/g, '').replace(/,/g, ' ')
-    : undefined;
-
   const scale = getTimeScale(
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (commented):
While it does seem to fix the firefox issue it brings back these errors in the console instead:
image
(they are just warnings in firefox though)

tonypls:
Hmm, I'll take a look and see if there's an error / warning free solution

madsnedergaard:
This works because the transform is now ignored - so it means it apparently works without adding this transform on both Firefox and Chrome, so we should just remove that transform completely it seems 😄

q--:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform

From a cursory glance, it appears the CSS `transform`  requires `, ` and the SVG `transform` attribute ` ` between arguments

tonypls:
image
Damn, both FF and chrome look like this without the transform, I'll try find an error free way to do this.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5044, opened 2023-02-07
Better parser documentation

VIKTORVAV99:
## Description
This PR will update our parser documentation and hopefully make things a little easier for our contributors as well as new team members. The structure of this PR is based on what was discussed in the meeting with Mads and will bring over and update the information from https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser into individual files specific for every parser function and then simplify that page.

TODO:

- [x] Update example.py with new syntax and conventions.

Function specific documentation for:
- [x] Production
- [ ] Consumption
- [ ] Price
- [ ] ProductionPerUnit
- [ ] Exchange
- [ ] GenerationForecast
- [ ] GenerationPerModeForecast
- [ ] ConsumptionForecast
- [ ] ExchangeForecast


madsnedergaard:
Yay, love the new README! 👏

parsers/examples/example_parser.py
```diff
@@ -300,43 +314,44 @@ def fetch_exchange(
...
-    print("fetch_price() ->")
-    print(fetch_price())
-    print("fetch_exchange(DK, NO) ->")
-    print(fetch_exchange("DK", "NO"))
+    print("fetch_production(DK-DK1) ->")
+    print(fetch_production("DK-DK1"))
+    print("fetch_price(DK-DK1) ->")
+    print(fetch_price("DK-DK1"))
+    print("fetch_exchange(DK-DK1, DK-DK2) ->")
+    print(fetch_exchange("DK-DK1", "DK-DK2"))
```
madsnedergaard, line 357: Should we continue with the non-specific values here?

```suggestion
    print("fetch_production(XX) ->")
    print(fetch_production("XX"))
    print("fetch_price(XX) ->")
    print(fetch_price("XX"))
    print("fetch_exchange(XX, YY) ->")
    print(fetch_exchange("XX", "YY"))
```

madsnedergaard (approved):
This is awesome, I think you should consider just merging it now with production and then adding documentation for more types in new PRs :)

madsnedergaard:
cc @lav-julien so you can see the great work Viktor is doing (that we discussed during a meeting today) 💪

parsers/examples/example_parser.py
VIKTORVAV99: I wanted something a little more abstract and general but it might be better to go back to the DK zones as examples as there is really no substitute for the time zones anyway.

VIKTORVAV99:
> This is awesome, I think you should consider just merging it now with production and then adding documentation for more types in new PRs :)

I have some final cleanup to do but we can merge this tomorrow if you wish. Then we could also link this to the contributing.md file and remove some of the duplication in there as well.

lav-julien:
> cc @lav-julien so you can see the great work Viktor is doing (that we discussed during a meeting today) 💪

That looks great! Thank you @VIKTORVAV99 for working on this!

PR #5045, opened 2023-02-07
Tonyvanswet/ele 1693 create list of zones with zones where

tonypls:
## Issue
Remove price data for Croatia


VIKTORVAV99:
Looks like this PR was created of the old branch (should be fine) but I'm curious if we actually need this field in the frontend config files or if we can exclude it from the generated zone.json file?

madsnedergaard:
> Looks like this PR was created of the old branch (should be fine) but I'm curious if we actually need this field in the frontend config files or if we can exclude it from the generated zone.json file?

Good point, we could actually exclude it as well in generate-zones-config :)

madsnedergaard:
We're getting this error on CI that we need to fix:
Screenshot 2023-02-07 at 13 14 59

VIKTORVAV99:
> We're getting this error on CI that we need to fix: Screenshot 2023-02-07 at 13 14 59
> 
> The new key needs to be added in `electricitymap/contrib/config/model.py:L99`

Damit mads beat me too it, but I think we should make timezone required but I'll do that in a separate PR and add time zones where it's needed.

madsnedergaard (approved):
LGTM! 👍

Issue #5040, opened 2023-02-06
fly-to-zone from URL


tonypls:
As a app user I want to have the map center on the zone in the url when the app is loaded so that I can link to a zone or find a zone from the zone key

PR #5046, opened 2023-02-07
change fly to zone based on atom

tonypls:
## Issue
#5040 
## Description
This PR adds a state atom for the selected zone id which is determined from the url zoneId param.

The map will now fly to zone from this atom.

It will be good to give this one a nice test on the preview as there is a chance for unexpected behaviour


web/src/types.ts
```diff
@@ -146,6 +146,7 @@ export interface MapGeometry extends Feature<Polygon | MultiPolygon> {
   properties: {
     zoneId: string;
     color: string;
+    center?: [number, number];
```
VIKTORVAV99, line 149: I don't think this need to be optional, there should always be a center for each zone.

tonypls:
Unfortunately the preview builds don't work when refreshing anything but the root path so it's quite hard to test this on the previews.

VIKTORVAV99:
> Unfortunately the preview builds don't work when refreshing anything but the root path so it's quite hard to test this on the previews.

Huh that worked on the cloudflare ones, maybe we should switch again... 😅

VIKTORVAV99 (approved):
LGTM! 🎉

I just one small comment.

web/src/features/map/mapAtoms.ts
```diff
@@ -11,3 +11,5 @@ export const mousePositionAtom = atom({
 export const hoveredZoneAtom = atom<HoveredZone | null>(null);
 
 export const mapMovingAtom = atom(false);
+
+export const zoneIdAtom = atom<FeatureId>(undefined);
```
madsnedergaard, line 15: Hmm, could we name this so that it is more clear what this is? Perhaps something like "activeZoneId" or "selectedZoneId"?

web/src/features/map/Map.tsx
```diff
@@ -23,8 +23,8 @@ import {
   loadingMapAtom,
   mapMovingAtom,
   mousePositionAtom,
+  zoneIdAtom,
```
madsnedergaard, line 26: Do we actually need an atom for tracking this - couldn't we just use the `useParams` hook from router instead? That way there's only one single source of truth, where here we now need to manually ensure these two are always in sync :)

web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -30,7 +32,7 @@ export default function ZoneDetails(): JSX.Element {
   if (!zoneId || Array.isArray(data)) {
     return <Navigate to="/" replace />;
   }
-
+  setZoneId(zoneId);
```
madsnedergaard, line 35: This should probably live inside a useEffect and not just in root of the component - otherwise it will fire whenever component rerenders :)

madsnedergaard (commented):
Some questions/thoughts to discuss :)

madsnedergaard (changes requested):
(marking as Request changes to ensure we don't accidentally merge before discussions are resolved)

VIKTORVAV99:
> (marking as Request changes to ensure we don't accidentally merge before discussions are resolved)

I think there is a GitHub setting for that 😅
Should be under `settings->branches` if I remember correctly, we could consider turning it on.

web/src/features/map/mapAtoms.ts
tonypls: Have updated this

web/src/features/panels/zone/ZoneDetails.tsx
tonypls: :)

web/src/features/map/Map.tsx
tonypls: I like the params approach and there's currently an issue where the app will crash if the zoneId is invalid. Before I fix the crash we should decide on the approach. 

I went with a global state zoneId because of this comment ```// TODO: Selected feature-id should be stored in a global state instead (and as zoneId).``` 

and currently the routes are initialised in the LeftPanel, to have context of the zoneId in useParams() in map.tsx I'd need to initialize the routes above the map.

Let me know what you think!

tonypls: @madsnedergaard Let me know what you think about lifting the routes to the top level vs using an atom or another option.

madsnedergaard: I think it makes sense to lift the routes to top level, it's already a bit confusing that it happens down in the panels only I think :)

madsnedergaard: I can't remember how to do nested routing with React Router, but it should be doable :D

madsnedergaard (approved):
Let's try it :)

PR #5048, opened 2023-02-07
[ESLint]: Rename files to camelCase

VIKTORVAV99:
## Description
Renames files to camelCase to get rid of 6 more ESLint errors.


web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -11,7 +11,7 @@ const zones = mergeZones();
 
 if (arguments_.length <= 0) {
   console.error(
-    'ERROR: Please add a zoneName parameter ("node generate-zone-bounding-boxes.js DE")'
+    'ERROR: Please add a zoneName parameter ("ts-node generateZoneBoundingBoxes.ts DE")'
```
tonypls, line 14: Will this work without ts-node?

tonypls: Actually nvm :)

tonypls (approved):
LGTM 👍

web/geo/generateZoneBoundingBoxes.ts
VIKTORVAV99: You already resolved it and yes technically it will work without ts-node for now but not when we start fixing the types.

PR #5049, opened 2023-02-07
ESLint: Fix in loops and no process exit

VIKTORVAV99:
Fix more ESLint errors!

Should be just two more PR's after this one!

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -169,7 +169,8 @@
         "project": ["./tsconfig.node.json"]
       },
       "rules": {
-        "unicorn/no-array-reduce": "off"
+        "unicorn/no-array-reduce": "off",
+        "unicorn/no-process-exit": "off"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


web/scripts/generateZonesConfig.ts
```diff
@@ -31,12 +31,14 @@ const mergeZones = () => {
...
-      point[1] = round(point[1], 4);
-    });
+    if (zoneConfig?.bounding_box) {
+      for (const point of zoneConfig.bounding_box) {
+        point[0] = round(point[0], 4);
+        point[1] = round(point[1], 4);
+      }
+    }
 
-    for (const key in zoneConfig) {
```
madsnedergaard, line 39: What's the rule for not allowing `x in y`?

web/.eslintrc.json
```diff
@@ -169,7 +169,8 @@
         "project": ["./tsconfig.node.json"]
       },
       "rules": {
-        "unicorn/no-array-reduce": "off"
+        "unicorn/no-array-reduce": "off",
+        "unicorn/no-process-exit": "off"
```
madsnedergaard, line 173: Thanks for turning these off :D

madsnedergaard (approved):
Not quite sure why `in` is not allowed, but otherwise LGTM

web/scripts/generateZonesConfig.ts
VIKTORVAV99: Something about duplicating the underlying prototype objects and slowing down performance. So `Object.keys(object)` creates a new iterator and then of loops over it as if it was an array.

I didn't dig too much into it though and just fixed it 😅

Issue #5052, opened 2023-02-07
Map tooltip overflows in some languages


madsnedergaard:
**Describe the bug**

The content inside the tooltip overflows on certain languages - for example Nederlands.

**To Reproduce**
1. Change language to Nederlands
2. Hover a zone

**Expected behavior**

It should look good for as many languages as possible.

**Screenshots**

![Screenshot 2023-02-07 at 22 08 46](https://user-images.githubusercontent.com/3296643/217367894-7f5dff3c-6dd7-42a6-89a7-54665d21b6f4.png)

PR #5053, opened 2023-02-07
Updates spacing of MapTooltip to make it look nice in more languages

madsnedergaard:
## Issue

#5052

## Description

This ensures it never overflows (with current translations at least). I also adjusted spacing between the elements to make it look nice and aligned in as many languages as possible. In a few of them (non-main language of our users) it looks a bit off still, but couldn't find a fix that worked for everything.

### Preview

See the preview build for difference - and try it out with different languages.


tonypls:
If the height increases too much then the cursor can overlap the tooltip causing flashing of the tooltip and you move around.

madsnedergaard:
> If the height increases too much then the cursor can overlap the tooltip causing flashing of the tooltip and you move around.

Ah, is that why it had the fixed height? (We should test this in different languages and see if it is a problem - and potentially add a new fixed height that looks good and adjust tooltip position for it)

tonypls:
> 

Yeah, fixed dimensions are there to help with the alignment and when to flip at the screen edges. We could do some container width and height variables but it's a bit complicated.

VIKTORVAV99:
> > If the height increases too much then the cursor can overlap the tooltip causing flashing of the tooltip and you move around.
> 
> Ah, is that why it had the fixed height? (We should test this in different languages and see if it is a problem - and potentially add a new fixed height that looks good and adjust tooltip position for it)

I'm both for and against a fixed height. Maybe we could some something like max high instead and let it flex as needed with a upper limit?

madsnedergaard:
I can try fiddling around with this tomorrow and see if I can strike a balance :)

madsnedergaard:
Hey, finally back from the bed and had time to look at this again. Instead of adding a fixed height (as it was always looking "off" in one language), I added "pointer-events-none" which means that hovering over the tooltip doesn't cause a flash :)

tonypls (approved):
Excellent!

PR #5054, opened 2023-02-08
IE: Fix consumption forecast parser

FelixDQ:
As per the [wiki](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser#fetch_consumption_forecast), the correct key for consumption forecasts is "value", and not "consumption".

We should extend the checks added [here](https://github.com/electricitymaps/electricitymaps-contrib/pull/5037), to also check forecast modes.


VIKTORVAV99:
Could this be harmonised to better match their non forecast counterparts?

The same goes with storage that is having the opposite behavior as production. (Negative is "production" and positive is storage)

mathilde-daugy:
@FelixDQ I saw that the other day but forgot about it, thanks for flagging

mathilde-daugy (approved):

mathilde-daugy:
> Could this be harmonised to better match their non forecast counterparts?
> 
> The same goes with storage that is having the opposite behavior as production. (Negative is "production" and positive is storage)

I agree, looking at the wiki i feel like we don't really have consistency depending on the kind parsed.

PR #5055, opened 2023-02-08
Create VN.py [WIP, help wanted!]

alixunderplatz:
ref #1974 
So far, it's a beginning only 😄
If one from the community or team wants to finish this parser and get data for another 97 million earthlings parsed, go ahead! 😃😃😃

### DIFF
--- a/parsers/VN.py
+++ b/parsers/VN.py
@@ -1,21 +1,26 @@
-#!/usr/bin/env python3
 # WORK-IN-PROGRESS for parsing price + load data of VIETNAM
+#!/usr/bin/env python3
 
 import requests
 import arrow
-import logging
-from datetime import datetime, timedelta
+import pytz
 
+from logging import Logger, getLogger
+from datetime import datetime
+from requests import Session
+from typing import Optional
 
-dt_day = arrow.now(tz='Asia/Ho_Chi_Minh').format('DD/MM/YYYY')
-dt_now = arrow.now(tz='Asia/Ho_Chi_Minh').format('DD/MM/YYYY HH:mm:ss')
-print('Fetching price and load data for:', dt_day, "at", dt_now, "local Vietnamese time.")
+dt_day = arrow.now(tz='Asia/Ho_Chi_Minh').format('YYYY-MM-DD')
+print('Fetching all available price and load data for >', dt_day, '< in Vietnam.')
+tz = pytz.timezone("Asia/Ho_Chi_Minh")
 
 ## Vietnamese National Load Dispatch Center https://www.nldc.evn.vn/
 price_url="https://www.nldc.evn.vn/GiaBienHandle.ashx?d=31/01/2023"
 load_url="https://www.nldc.evn.vn/PhuTaiHandle.ashx?d=31/01/2023"
 
-##fetch_price()
+zone_key_list = [['VN-N'], ['VN-C'], ['VN-S'], ['VN']]
+
+##  fetch_price()
 # [0] = price of Northern price zone
 # [1] = price of Central price zone
 # [2] = price of Southern price zone
@@ -28,6 +33,9 @@ price_north = price_data['data'][0]
 price_central = price_data['data'][1]
 price_south = price_data['data'][2]
 
+load_list = []
+load_values = []
+
 for hours in range(len(price_north)):
     price_values = dict(hour=datetime.time(datetime.strptime(("%d:%02d" % (int(hours/2), (hours*30 % 60))), "%H:%M")),
                        north=price_north[hours],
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
I have not worked on the parser yet but by looking through the code you already have it seems like we could get data for 3 regions in Thailand? Therefore I think we should make the parser handle these zones + a national level for Thailand. This would allow us to split up the country (and aggregate it) if we can find exchange data between these regions as well. Do you know of any such source?

alixunderplatz:
@VIKTORVAV99 correct, the parser gets the consumption and price of the 3 market regions and consumption for Vietnam as a whole.
I'd name the zones VN-N, VN-C, VN-S and VN. 
Outlines of the 3 sepearate zones are likely based on existing geographical regions.
I had not looked for transmission data between the zones, I guess that's harder to find. 

From what I remember (ref #1974), Vietnam had no big import/export balance with other countries, so making Electricity Maps Estimations for its production on the countrywide zone VN, based on the total load, will be quite accurate (like Israel e.g.) as a first step to make sth out of this datasource. Solar boom was ongoing in the last couple of years.

VIKTORVAV99:
@alixunderplatz any progress on this?
If you need help with anything in particular don't hesitate to let me know and I'll assist as much as I can.

alixunderplatz:
> @alixunderplatz any progress on this? If you need help with anything in particular don't hesitate to let me know and I'll assist as much as I can.

Hi @VIKTORVAV99 ! I just added my latest progress from a month ago. I have no capacity atm, unfortunately, so please continue =)

Last changes:
- got the date format for the half-hourly data to match the number of available entries working
- just appending all the 4 zones' data into one "block", but not really in the required format or distinguishing between VN, VN-N, VN-C or VN-S
- price is just being printed for visualization and time is just for vis, but much of the code should be 1:1 copyable from fetch_consumption() once you get that clean and tidy. 3 price zones, so no avg price for VN (unlike sum of load)!

VIKTORVAV99:
> > @alixunderplatz any progress on this? If you need help with anything in particular don't hesitate to let me know and I'll assist as much as I can.
> 
> Hi @VIKTORVAV99 ! I just added my latest progress from a month ago. I have no capacity atm, unfortunately, so please continue =)
> 
> Last changes:
> 
> * got the date format for the half-hourly data to match the number of available entries working
> * just appending all the 4 zones' data into one "block", but not really in the required format or distinguishing between VN, VN-N, VN-C or VN-S
> * price is just being printed for visualization and time is just for vis, but much of the code should be 1:1 copyable from fetch_consumption() once you get that clean and tidy. 3 price zones, so no avg price for VN (unlike sum of load)!

Thanks for pushing the latest changes!
I am pretty busy myself latley but I'll try and take a look this weekend. 👍🏼

I'm not sure we can publish this as 3 seperate zones and aggregate it without exchange data between the zones but in either case we should have enough data to add it on a national level.

As for the price data 1 price for each zone is fine and we could in theory use a weighted average of the cosumption for each zone to create a national price but I'm unsure if the parser is the right place to do that as it would probably be preferable to do it in the backend for all zones with subzones.

MuffinCompiler:
Hey folks,

I decided this weekend to look at the Vietnam data and made @alixunderplatz parser work as intended. I implemented the `fetch_consumption` and `fetch_price` for both current and historical data, for the three sub-zones (north, central, south), and Vietnam in total. Also created `.yaml` files accordingly, but without geo information.

The data source is a bit... weird, especially regarding past data. If you request data for a given past day, you only get data for that day up to the current time. So you should poll past data late in the evening (Vietnam time) to get a complete set ;)
On the upside, the source also has data for the upcoming 2 hours, probably a short-range consumption forecast, these future data points change every 10 minutes or so.

There are a few more issues I encountered:
- The daily graph on the data page does not match perfectly the archive of real time data by a difference of +/- 2% of total load. I guess there is some post-processing going on, the title of the daily chart has some word in it that translates to 'correlated load'. The real time data might also be more of a estimate.
- The price in the daily chart is more significantly off. However, the real time data set is labeled as 'market price', while the daily chart prices translate to something like 'variable price'. The real time price also randomly plummets to 0, 1 or some other random low value. I set a simple threshold for some filtering, but this still gives weird values sometimes.

I also found a nice source of [capacities ](https://www.nldc.evn.vn/FullNewsg/100/Thong-tin-thi-truong-dien/default.aspx) of different power sources. This source seems also being updated from time to time. There is also a next-day forecast of some load, but not with a time, more like a maximum being expected. But this also gives us a nice estimate on the power sources! All in all, this should give us a parser similar to the Kuwait one.

Any chance you could permit access to push to this pull request, or should I just create a new one?

VIKTORVAV99:
> Hey folks,
> 
> I decided this weekend to look at the Vietnam data and made @alixunderplatz parser work as intended. I implemented the `fetch_consumption` and `fetch_price` for both current and historical data, for the three sub-zones (north, central, south), and Vietnam in total. Also created `.yaml` files accordingly, but without geo information.
> 
> The data source is a bit... weird, especially regarding past data. If you request data for a given past day, you only get data for that day up to the current time. So you should poll past data late in the evening (Vietnam time) to get a complete set ;) On the upside, the source also has data for the upcoming 2 hours, probably a short-range consumption forecast, these future data points change every 10 minutes or so.
> 
> There are a few more issues I encountered:
> 
> * The daily graph on the data page does not match perfectly the archive of real time data by a difference of +/- 2% of total load. I guess there is some post-processing going on, the title of the daily chart has some word in it that translates to 'correlated load'. The real time data might also be more of a estimate.
> * The price in the daily chart is more significantly off. However, the real time data set is labeled as 'market price', while the daily chart prices translate to something like 'variable price'. The real time price also randomly plummets to 0, 1 or some other random low value. I set a simple threshold for some filtering, but this still gives weird values sometimes.
> 
> I also found a nice source of [capacities ](https://www.nldc.evn.vn/FullNewsg/100/Thong-tin-thi-truong-dien/default.aspx) of different power sources. This source seems also being updated from time to time. There is also a next-day forecast of some load, but not with a time, more like a maximum being expected. But this also gives us a nice estimate on the power sources! All in all, this should give us a parser similar to the Kuwait one.
> 
> Any chance you could permit access to push to this pull request, or should I just create a new one?

First of all thanks for finishing the parser!

Unfortunately there is no way to grant write access to just one PR so I think the best bet would be to fork the repo and then create a new branch of this branch and then open a new PR, that should preserve all the git history.

We would prefer if there is no filtering or any other data manipulation in the parser and let the backend handle that if needed. For a few different reasons but mainly to make sure all parsers work the same.

As for fetching the data we run the parsers roughly every 15 minutes so we should be getting all data so there is no concern about that in this case.

Lastly, differences between the historical data and the live data is not a concern right now but we are working on a way to account for this in the future but that is not ready yet.

VIKTORVAV99:
There is a continuation of this pr at #5364 which will close this PR once it's merged. It was created directly from this branch so all previous contributions will be included as well.

Thanks for the work in here!

PR #5056, opened 2023-02-08
Fix ESLint disable comments + eslint --fix

VIKTORVAV99:
Opening this as a draft first to make sure the solar layer stuff still works (lots of fix changes by ESLint).

This is the last PR before the big one that fixes all the `__dirname` stuff.


VIKTORVAV99:
Okay so things are looking like they worked with all the ESLint --fix changes so turning this into a normal PR.

web/src/features/weather-layers/solar/stackBlurImageOpacity.ts
```diff
@@ -42,34 +40,39 @@ function BlurStack() {
...
 
-  var stackStart = new BlurStack();
-  var stack = stackStart;
-  for (i = 1; i < div; i++) {
+  const stackStart = new BlurStack();
+  let stack = stackStart;
+  for (index = 1; index < div; index++) {
     stack = stack.next = new BlurStack();
-    if (i == radiusPlus1) var stackEnd = stack;
+    // eslint-disable-next-line no-var
```
VIKTORVAV99, line 63: Could possibly be removed but I didn't want to mess with the layers too much, they break way too easy and figured it would be best to get the Linting CI up and running before making more changes.

web/src/main.tsx
```diff
@@ -52,7 +52,9 @@ if (isProduction) {
 
 // Temporarily disabled to ensure we can more easily rollback
 // Also removes existing service workers to ensure they don't interfer
+
 if (navigator.serviceWorker) {
+  // eslint-disable-next-line unicorn/prefer-top-level-await
```
VIKTORVAV99, line 57: As we plan on removing this code once we feel everything is stable I felt no need to "fix it" now by turning it into an await statment.

VIKTORVAV99:
@madsnedergaard have you reviewed this pr? I see your thumbs up on the comments but that's it.

Would be great to get it merged so I can do the next one without conflicts.

tonypls (approved):
LGTM, always a little gun shy on anything that touches the weather layers but I think we're fine in this case. 👍

VIKTORVAV99:
> LGTM, always a little gun shy on anything that touches the weather layers but I think we're fine in this case. 👍

Tell me about it, I tried to change those 3 times and we had to revert them 3 times.
This is the stuff of nightmares... 😅

Issue #5051, opened 2023-02-07
Mobile app: Cannot open country panel when language is Nederlands


madsnedergaard:
**Describe the bug**

When language is set to Nederlands, the app will not open data panel when clicking on a country.
This was reported via our feedback form - and I am able to reproduce it.

**To Reproduce**
1. Open Apple mobile app
2. Set language to Nederlands
3. Try clicking any country

**Expected behavior**
It should work no matter what language you are using 😅 

**Screenshots**

https://user-images.githubusercontent.com/3296643/217367455-d7b408c7-8dbb-424f-bdca-737f06c26856.mp4

**Smartphone (if relevant):**
 - Device: iPhone 12 Mini
 - OS: iOS 16.2
 - Native App
 - Version 1.5.1

Potentially unrelated, but on desktop the map tooltip also overflows <-- created new issue for this (#5052)!

PR #5058, opened 2023-02-08
hide zonetooltip on smaller devices

tonypls:
## Issue
Unable to open country panel on mobile devices in when useing Nederlands as the language
Fixes: #5051
## Description
When the language protrudes from the zonetooltip and the tooltip is left aligned ( cursor close to right of window ) the tap or click from the user is blocked by the tooltip.

This PR prevents the tooltip from being rendered on small devices as it was not intended to be. This solves the bug on mobile but a deeper look into overflow should be taken at #5053


VIKTORVAV99:
Ugh the Netlify bot seems to have gotten a little crazy and is now posting a new comment every time there is a update to a PR...

madsnedergaard (approved):

PR #5061, opened 2023-02-08
update arrow config

mathilde-daugy:
## Description

Update configs for exchanges for:

- US-CENT-SWPP->US-MIDW-MISO (overlapping and in the wrong place)
- US-CENT-SWPP->US-MIDW-AECI (overlapping and in the wrong place)
image

- AR->PY (not on the border)
- AR-> CL-SEN (in the sea)

image

### Preview

NA


VIKTORVAV99 (commented):
Could we hold of on the changes to AR exchanges?
They are fixed in #4894

If not could we update this to use the same coordinates and rotations so there are no conflicts?

mathilde-daugy:
> Could we hold of on the changes to AR exchanges? They are fixed in #4894
> 
> If not could we update this to use the same coordinates and rotations so there are no conflicts?

Ok thanks for flagging, we'll be using the config in the other PR then

config/exchanges/US-CENT-SWPP_US-SW-EPE.yaml
```diff
@@ -1,6 +1,6 @@
 lonlat:
-  - -103.81455
+  - -1000.4
```
tonypls, line 2: I this meant to be 1000?

tonypls:
These two need to be flipped 180 degrees 
image

tonypls (approved):
LGTM 🦦

Issue #4974, opened 2023-01-26
New mobile app : back button


phiphou:
When I click on a country, I used to tap the back button on my GS7 to go back to the map. With the new app,the back button close the app.

VIKTORVAV99:
Thanks for the feedback!
I do agree it would be great if it worked like the old app.

cc @tonypls @madsnedergaard just so you guys are aware of it too.

tonypls:
Nice catch, yes we should set up our routing to handle the native buttons in app. I'll create a task!

PR #5063, opened 2023-02-08
listen for back button on android devices

tonypls:
## Issue
Enable back button on android, fixes [#4974](https://github.com/electricitymaps/electricitymaps-contrib/issues/4974)
## Description
The back button on android currently closes the app. We should allow the back button to navigate back from the ranking panel and close the app from the map


mobileapp/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@capacitor/android": "^4.6.2",
+    "@capacitor/app": "^4.1.1",
```
VIKTORVAV99, line 20: Should this not use the same minor version as the other `@capacitor/*` packages?

web/package.json
```diff
@@ -36,6 +36,7 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@capacitor/app": "^4.1.1",
```
VIKTORVAV99, line 39: Same here as above.

VIKTORVAV99 (commented):
Just some small comments 🙂

mobileapp/capacitor.config.ts
```diff
@@ -15,7 +15,8 @@ switch (process.env.NODE_ENV) {
     config = {
       ...baseConfig,
       server: {
-        url: 'http://localhost:5173/',
+        url: 'use address from pnpm dev --host',
```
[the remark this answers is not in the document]
tonypls: Yeah, it's intentional. I can't seem to get hot reloading to work on android with localhost and the exposed host from Vite will change each time

mobileapp/package.json
tonypls: No I think they're each their own independent packages

madsnedergaard (approved):
Yay, thanks for fixing it! 🙌

PR #5068, opened 2023-02-09
Fix license

lav-julien:
## Issue
The MIT license (old license) should be kept in the repository and referenced in the readme


README.md
```diff
@@ -34,7 +34,7 @@ We would love your feedback on how to improve the contribution experience, pleas
 
 ## License
 
-Electricity Maps is licensed under GNU-AGPLv3. Find our license [here](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE.md)
+This repository is licensed under GNU-AGPLv3 since v1.5.0 released on 31st January 2023. Find our license [here](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE.md). Prior contributions were licensed under [MIT license](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE_MIT.txt)
```
VIKTORVAV99, line 37: We should specify the exact commit instead of the release version here.

LICENSE_MIT.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Tomorrow
```
VIKTORVAV99, line 3: Should the date here include a time span, i.e. 2017-2022?

VIKTORVAV99 (commented):
Just two small comments.

LICENSE_MIT.txt
lav-julien: Not sure if we should modify the text here as it's the license itself. According to our lawyers the date should be specified in the readme file

README.md
lav-julien: I have a hard time finding which commit we should specify here, do you have a suggestion?

VIKTORVAV99: The exact commit is cb9664f43f0597bedf13e832047c3fc10e67ba4e but I would add it as `[cb9664f](https://github.com/electricitymaps/electricitymaps-contrib/commit/cb9664f43f0597bedf13e832047c3fc10e67ba4e)` in consistency with GitHub commit referencing.

LICENSE_MIT.txt
VIKTORVAV99: Huh that's strange, I don't think I have ever seen a date for a license in a readme before. But I guess that works?

lav-julien: Maybe adding 2020-2022 here since the original data in the license was 2020? and then specify the date in the readme would be a good combination of both options?

VIKTORVAV99: I think that would work but please do double check with @madsnedergaard as well. 👍🏼

lav-julien: I removed the date from the readme and only specified version+commit, let me know what you think about it

README.md
```diff
@@ -34,7 +34,7 @@ We would love your feedback on how to improve the contribution experience, pleas
 
 ## License
 
-Electricity Maps is licensed under GNU-AGPLv3. Find our license [here](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE.md)
+This repository is licensed under GNU-AGPLv3 since v1.5.0, find our license [here](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE.md). Contributions prior to commit cb9664f were licensed under [MIT license](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE_MIT.txt)
```
VIKTORVAV99, line 37:
```suggestion
This repository is licensed under GNU-AGPLv3 since v1.5.0, find our license [here](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE.md). Contributions prior to commit cb9664f43f0597bedf13e832047c3fc10e67ba4e were licensed under [MIT license](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/LICENSE_MIT.txt)
```
Let's use the full commit hash (GitHub will truncate this and link it that way)

LICENSE_MIT.txt
madsnedergaard: I think date span makes sense, to make it clearer for people that this license was active during that time (but not before or after).

lav-julien:
@madsnedergaard @VIKTORVAV99 I updated the date in the MIT license and incorporated the commit number in the readme suggested by Viktor. Looks good to be merged to me

madsnedergaard:
Should it still be Tomorrow in the license file? 😅

lav-julien:
> Should it still be Tomorrow in the license file? 😅

I Think it's fair to say that the copyright should includes the year and the holder name when entering into effect (https://opensource.org/license/mit-0/). Also our company still have Tomorrow as a second name on legal status so I don't think it's a big problem :)

madsnedergaard (approved):

q--:
Glad to see this fixed, thank you

Issue #5032, opened 2023-02-06
Enable map zoom when hovering exchange arrow


tonypls:
**Describe the bug**
Unable to zoom when cursor is over an exchange arrow

**To Reproduce**
Open the app on desktop
Hover an exchange arrow
Use mouse wheel to trackpad to zoom

**Expected behavior**
The map is zoomed in or out as expected

**Screenshots**

**Desktop (if relevant):**
All desktop browsers

PR #5069, opened 2023-02-09
prevent default zoom on exchange arrows and trigger map zoom

tonypls:
## Issue
#5032 
Enable zooming when hovering exchanges
## Description
Enables zooming while hovering on exchange arrows by preventing the global window zoom then using the wheel event to trigger the hiding of exchange arrows which allows the event to pass through to the map controls.


web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -19,11 +21,17 @@ function ExchangeArrow({ data, viewportWidth, viewportHeight, map }: ExchangeArr
...
+  const setIsMoving = useSetAtom(mapMovingAtom);
   if (!lonlat) {
     return null;
   }
 
+  useEffect(() => {
+    const cancelWheel = (event: WheelEvent) => event.preventDefault();
+    document.body.addEventListener('wheel', cancelWheel, { passive: false });
+    return () => document.body.removeEventListener('wheel', cancelWheel);
+  }, []);
```
VIKTORVAV99, line 33: I assume there is no way to skip the `preventDefault()` and make this a passive event listener in our current implementation?

tonypls: Unfortunately passive tells the browser it can still do it's thing without considering the JS so it will lead to a double zoom effect with the whole page being zoomed at the same time as the map.

tonypls: The alternative solution would be like we used to handle this with our own listeners for all of the things

VIKTORVAV99: Let's go with this for now and then add a task about evaluating the performance benefits (if any) of switching to custom event listeners.

madsnedergaard:
Could we potentially do something similar so that it is possible to drag the map while clicking on an arrow? Don't know if it is doable and worth it, just a quick thought as I sometimes accidentally do it and am confused why the map is not moving 😅

tonypls:
> 

It would be quite a bit of work to achieve, the main challenge is the radix tooltip component is listening for a hover. A nice solution is to set the pointer-events of the arrow to none and pass through all the pointer events to the map. The trouble is triggering the tooltips without pointer events.  

I think for now it's nice to have feature but perhaps not a high priority because of the time investment.

madsnedergaard (approved):

Issue #5030, opened 2023-02-06
Display Exchange Tooltip on Mobile


tonypls:
**Describe the bug**
We want to display the exchange tooltip on mobile devices, currently there is not handler for tapping on exchanges, the exchange tooltip is only shown on hover. 

The exchange tooltip is useful information and we should enable users to view it on mobile

**To Reproduce**
Open the app on a mobile device, tap an exchange arrow

**Expected behavior**
As a mobile user when tapping an exchange arrow the exchange tooltip is displayed

**Smartphone (if relevant):**
All mobile devices, web app and mobile app

madsnedergaard:
Duplicate of #5030, but let's keep this one as it has additional context :)

tonypls:
> Duplicate of #5030, but let's keep this one as it has additional context :)

Is this not the #5030 itself? 🤯

PR #5070, opened 2023-02-10
Tonyvanswet/exchange arrow tooltip mobile

tonypls:
## Issue
#5030
Enable exchange tooltips on mobile.

## Description
This PR creates mobile specific tooltips and triggers for the exchange arrows. These tooltips are opened via a click / touch as opposed to the hover trigger of the desktop tooltips. The Layout is also changed to be more vertical to better suit the portrait screen layout on mobile.

### Preview

image


VIKTORVAV99 (commented):
Just a small question.
Could we get the arrow between the zones to point up and down instead of left and right?

VIKTORVAV99 (commented):
I assume the current behavior is not intentional so can this PR be modified to close any existing tooltip when a new arrow is pressed?

Current behavior might result in this otherwise:
image

Besides that everything looks good, there are some overflowing on the zone names but I believe we have a linear issue open about how to (partially) resolve it by excluding the country if it's already included in the zone name so lets ignore in this pr.

tonypls:
> I assume the current behavior is not intentional so can this PR be modified to close any existing tooltip when a new arrow is pressed?
> 
> Current behavior might result in this otherwise: image
> 
> Besides that everything looks good, there are some overflowing on the zone names but I believe we have a linear issue open about how to (partially) resolve it by excluding the country if it's already included in the zone name so lets ignore in this pr.

This was by design / a compromise on the effort required vs value. 

The overlap case isn't great but it does allow users to compare exchanges and the tooltips will be removed as soon as the map is moved or the initial tooltip is touched. 

I think it's good to get this added and look into improving later if we find it's a priority

VIKTORVAV99:
> > I assume the current behavior is not intentional so can this PR be modified to close any existing tooltip when a new arrow is pressed?
> > 
> > Current behavior might result in this otherwise: image
> > 
> > Besides that everything looks good, there are some overflowing on the zone names but I believe we have a linear issue open about how to (partially) resolve it by excluding the country if it's already included in the zone name so lets ignore in this pr.
> 
> This was by design / a compromise on the effort required vs value. 
> 
> The overlap case isn't great but it does allow users to compare exchanges and the tooltips will be removed as soon as the map is moved or the initial tooltip is touched. 
> 
> I think it's good to get this added and look into improving later if we find it's a priority

Oh okay 👍🏻
Then this looks good to me but we should open a new (draft) issue for it so we don't forget.

I'll approve it right away!

VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard:
Hey, just found an issue introduced in this PR: It is no longer possible to scroll with trackpad inside the leftpanel 😬 
I suspect it's because of the changes to `wheel` events, can we somehow reduce scope to only be for arrow components and not on body?

tonypls:
> Hey, just found an issue introduced in this PR: It is no longer possible to scroll with trackpad inside the leftpanel 😬 I suspect it's because of the changes to `wheel` events, can we somehow reduce scope to only be for arrow components and not on body?

Will take a look :)

PR #5074, opened 2023-02-11
Update TR Geothermal Values

SSD1805:
Added Geothermal CO2 energy output as per issue 5019

## Issue
https://github.com/electricitymaps/electricitymaps-contrib/issues/5019#issue-1569478266
## Description
 Verify that the values in the paper are valid.
 Create a PR to update the regional emissions if the above task is done.

### Preview

image

### DIFF
--- a/config/zones/TR.yaml
+++ b/config/zones/TR.yaml
@@ -2,6 +2,9 @@ _comment: https://doi.org/10.1016/j.rser.2017.01.060 says pumped hydro not econo
   viable. http://www.hurriyetdailynews.com/turkey-plans-to-hold-new-big-tenders-for-wind-solar-plants-by-summer-2018-minister-127769
   says battery will be tendered summer 2018 so below battery storage will need to
   be updated when built.
+  Akın, S., Örücü, Y., & Frıdrıksson, T. (2020). Characterizing the declining CO2 emissions from Turkish geothermal power plants. 
+  https://pangea.stanford.edu/ERE/db/GeoConf/papers/SGW/2020/Akin.pdf
+  Value for geothermal: The average emission of these GPPs is 595.92 CO2/kWh respectively, which is significantly lower than 2015 average (887 g/kWh).
 bounding_box:
   - - 25.16325931100002
     - 35.319778545000034
@@ -25,6 +28,7 @@ contributors:
   - lorrieq
   - nessie2013
   - lgrigoryeva1
+  - SSD1805
 emissionFactors:
   direct:
     battery discharge:
@@ -120,6 +124,12 @@ emissionFactors:
       datetime: '2021-01-01'
       source: INCER ACV
       value: 26.15
+    geothermal:
+      datetime: '2020-01-01'
+      source: Akın, Serhat, Yasemin Örücü, and Thrainn Frıdrıksson. "Characterizing the declining CO2 emissions from \n
+      Turkish geothermal power plants." (2020)
+      url: https://pangea.stanford.edu/ERE/db/GeoConf/papers/SGW/2020/Akin.pdf
+      value: 595.92
 fallbackZoneMixes:
   powerOriginRatios:
     - _source: Electricity Maps, 2015 average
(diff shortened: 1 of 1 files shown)


config/zones/TR.yaml
```diff
@@ -120,6 +124,12 @@ emissionFactors:
       datetime: '2021-01-01'
       source: INCER ACV
       value: 26.15
+    geothermal:
+      datetime: '2020-01-01'
+      source: Akın, Serhat, Yasemin Örücü, and Thrainn Frıdrıksson. "Characterizing the declining CO2 emissions from \n
+      Turkish geothermal power plants." (2020)
```
VIKTORVAV99, line 130: This should follow MLA citation style as and should not contain a line break (`\n`). If needed, you can wrap the whole string in single quotes (`''`).

config/zones/TR.yaml
```diff
@@ -120,6 +124,12 @@ emissionFactors:
       datetime: '2021-01-01'
       source: INCER ACV
       value: 26.15
+    geothermal:
+      datetime: '2020-01-01'
```
VIKTORVAV99, line 128: The datetime listed in the paper for the value you added is August 2019, not January 2020.

VIKTORVAV99 (changes requested):
First, I'm not sure it's as easy as this PR makes it out to be. The paper in question uses different values per year so we should as well, preferably going back as far as 2015 till today (or 2024 as the paper suggest)

Secondly, we need to verify these values to make sure they can be trusted.

Thirdly, this PR does not follow the structure that is required for the zone files, which will need to be corrected before this can be merged.

config/zones/TR.yaml
```diff
@@ -120,6 +124,12 @@ emissionFactors:
...
+      source: Akın, Serhat, Yasemin Örücü, and Thrainn Frıdrıksson. "Characterizing the declining CO2 emissions from \n
+      Turkish geothermal power plants." (2020)
+      url: https://pangea.stanford.edu/ERE/db/GeoConf/papers/SGW/2020/Akin.pdf
```
mathilde-daugy, line 131: url is not a valid key for this config, we only use source to link to the research paper

mathilde-daugy: In [KR.yaml](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/config/zones/KR.yaml), there is a url linked for unknown using the key `_url`

config/zones/TR.yaml
```diff
@@ -120,6 +123,15 @@ emissionFactors:
...
       value: 26.15
+    geothermal:
+      datetime: '2019-08-01'
+      source: 'Akın, Serhat, Yasemin Örücü, and Thrainn Frıdrıksson. "Characterizing the declining CO2 emissions from Turkish geothermal power plants." (2020)'
+      value: 595.92
+    unknown:
+      _comment: 'Source: 2020 Characterizing the declining CO2 emissions from Turkish geothermal power plants'
+      _url: https://pangea.stanford.edu/ERE/db/GeoConf/papers/SGW/2020/Akin.pdf
+      source: Akin 2020, Power weighted average and arithmetic average CO2 emissions of these GPP’s are 582.16 and 595.92 g CO2/kWh. 
+      Value for geothermal: The average emission of these GPPs is 595.92 CO2/kWh respectively, which is significantly lower than 2015 average (887 g/kWh).
```
VIKTORVAV99, line 134: Why is this listed under unknown?

VIKTORVAV99 (changes requested):
There are still highly questionable changes in here and this will fail in our backend.

Please do take a look how the previous regional sources are formatted.

We also need to make sure we get a accurate value for each year the paper cover before this can be merged, right now we only have data for the end of 2019. We also need to verify these numbers as we can't just take it at face value.

VIKTORVAV99:
@SSD1805 have you made any progress on this incorporating the feedback here? We can't merge this as is.

VIKTORVAV99:
Closing this PR as there has been no response from the author.

PR #5082, opened 2023-02-12
Ignore generated files

### LINKED ISSUE #5062: Remove generated files from git
The zones.json and similar files that we always generated should probably not be committed, but generated on the fly in CI instead.

Image

I think we can ignore all the files in web/config

VIKTORVAV99:
## Issue
Closes #5062
## Description
Removes the generated files from git and ensures they are built as needed.
The current implementation is rough, and force rebuilds in all environments. In the future we could create a new script that act as a "setup-ci" script that is run on pre{build,dev,test} that only generate the files if the environment variable CI is true (GitHub sets this automatically when run in actions). But I think that should be done in a separate PR.


web/package.json
```diff
@@ -11,20 +11,23 @@
     "url": "https://github.com/electricityMaps/electricitymaps-contrib.git"
   },
   "scripts": {
-    "mockserver": "nodemon ../mockserver/server.js",
```
VIKTORVAV99, line 14: Turns out you were right about this one Mads 😅

I don't know why it worked when I originally tested it locally, but it works now.

.github/workflows/CI.yml
```diff
@@ -136,14 +136,14 @@ jobs:
...
-      needs.setup.outputs.workflows != 'false'
-    uses: ./.github/workflows/validate_generated_files.yml
+  # validate_generated_files:
+  #   name: Validate
+  #   needs: [setup]
+  #   if: >-
+  #     needs.setup.outputs.frontend != 'false' ||
+  #     needs.setup.outputs.config != 'false' ||
+  #     needs.setup.outputs.workflows != 'false'
+  #   uses: ./.github/workflows/validate_generated_files.yml
```
VIKTORVAV99, line 146: This is no longer needed since all the files should be generated now, and it has nothing to verify but let's keep it around for now but commented out.

madsnedergaard:
`world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔 

Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?

VIKTORVAV99:
> `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> 
> Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?



> `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> 
> Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?

I think it was so we could run the "no unexpected update" parts of our CI. But now that we regenerate the file every time it's no longer needed. At least I think that was the reason 😅

madsnedergaard:
> > `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> > Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?
> 
> > `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> > Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?
> 
> I think it was so we could run the "no unexpected update" parts of our CI. But now that we regenerate the file every time it's no longer needed. At least I think that was the reason 😅

Right now the `verify:ci` script is called on our internal CI before updating contrib pointer, so I guess we need to remove that too? I am still slightly unsure about the changes here, but let's try it 🙈

VIKTORVAV99:
> > > `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> > > Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?
> > 
> > 
> > > `world.json` used to still be committed before the rewrite. I cannot remember why we chose to do that originally, but I'm a bit unsure if removing it will cause any problems 🤔
> > > Will try to take some time to review this PR properly soon, alternatively we can split it up and deal with the world.json thing in a separate PR?
> > 
> > 
> > I think it was so we could run the "no unexpected update" parts of our CI. But now that we regenerate the file every time it's no longer needed. At least I think that was the reason 😅
> 
> Right now the `verify:ci` script is called on our internal CI before updating contrib pointer, so I guess we need to remove that too? I am still slightly unsure about the changes here, but let's try it 🙈

Ah I was not aware of that, due to the email thing and the fact that the internal CI use SSO I don't have access to those dashboards and I never really dug into the backend CI. (Haven't really needed it before either TBH).

Let me know how you want to proceed, I think everything should be working as expected (tested most flows locally) but there is still a risk I missed something.

madsnedergaard:
Okay, let's resolve conflicts and get this merged!

madsnedergaard (approved):

VIKTORVAV99:
> Okay, let's resolve conflicts and get this merged!

I'll take care of the conflicts!
Did we remove the `verify:ci` in the internal CI as well (so this is safe to merge)?

VIKTORVAV99:
Before merging this I realized we no longer test if this builds at all, maybe we should add a simple job in the CI that just checks if the files "compile" without errors?

madsnedergaard:
> Before merging this I realized we no longer test if this builds at all, maybe we should add a simple job in the CI that just checks if the files "compile" without errors?

Yeah, that would be great! I had this note on one of the draft issues in the new project around this:

> Testing pnpm run build to make sure the app compiles (could be expanded to mobile builds in future if necessary)

VIKTORVAV99:
> > Before merging this I realized we no longer test if this builds at all, maybe we should add a simple job in the CI that just checks if the files "compile" without errors?
> 
> Yeah, that would be great! I had this note on one of the draft issues in the new project around this:
> 
> > Testing pnpm run build to make sure the app compiles (could be expanded to mobile builds in future if necessary)

While these would technically run on build I don't think we want to wait for that and if we have it as a separate one we can run it on /config changes without running the whole build pipeline, It will be some duplication but I think it might be worth it. (There are even ways we could reuse the files in the build step either with caching or artefacts but that might take longer than just re-running it.

VIKTORVAV99:
Merging this now and I'll update some other branches to make sure it works.

madsnedergaard:
Hey, I just noticed that we now have two files that are somewhat different and only one is in use:
Screenshot 2023-03-08 at 07 44 58

Should the one we use not have the `VERIFY_NO_UPDATES: 1` env var set?

VIKTORVAV99:
> Hey, I just noticed that we now have two files that are somewhat different and only one is in use:
> Screenshot 2023-03-08 at 07 44 58
> 
> Should the one we use not have the `VERIFY_NO_UPDATES: 1` env var set?

Unless I misunderstood the code there is nothing in the repo to verify against now since we no longer commit these files?

But I wanted to keep the old on around if we  needed to switch back (without needing to revert the whole PR).

madsnedergaard:
Ah wait, now I get it - thanks for the explanation 😄

PR #5084, opened 2023-02-12
Fix Aruba timestamp-issue

alixunderplatz:
No data for Aruba at the moment on the map. The problem lies within the timestamp of the data.
We're getting the "fossil" timestamp, which shows the correct hour and minute, but is stuck at 2023-01-01.

I guess your system is full of 2023-01-01 timestamps now? ^^

Only the "biogas" timestamp is correct, though, so I'm proposing to switch to it to get AW back on the map.
I rounded the floats as well, no need for milliwatts^^

grafik

output with fossil timestamp:
{'zoneKey': 'AW', 'datetime': datetime.datetime(**_2023, 1, 1, 13, 15_**, tzinfo=tzfile('/usr/share/zoneinfo/America/Aruba')), 'production': {'oil': 63.4831772, 'wind': 27.21717, 'solar': 1.87696636, 'biomass': 0.0, 'unknown': 0}, 'storage': {}, 'source': 'webaruba.com'}

output with biogas timestamp:
{'zoneKey': 'AW', 'datetime': datetime.datetime(**_2023, 2, 12, 13, 15_**, 4, 935000, tzinfo=tzfile('/usr/share/zoneinfo/America/Aruba')), 'production': {'oil': 63.483, 'wind': 27.217, 'solar': 1.8770, 'biomass': 0.0, 'unknown': 0}, 'storage': {}, 'source': 'webaruba.com'}


parsers/AW.py
```diff
@@ -55,17 +55,17 @@ def fetch_production(
...
         "datetime": zone_date_time.datetime,
         "production": {
-            "oil": float(fossil["value"]),
-            "wind": float(wind["value"]),
-            "solar": float(solar["value"]),
-            "biomass": float(biogas["value"]),
+            "oil": round(float(fossil["value"]), 3),
+            "wind": round(float(wind["value"]), 3),
+            "solar": round(float(solar["value"]), 3),
+            "biomass": round(float(biogas["value"]), 3),
```
VIKTORVAV99, line 68: If we want to round them and assume these are in MW we should round them to six decimals (1 W) not three (1kw). But I really don't see a need to round them. However, we might consider rounding the unknown one to six decimals as it's computed but it's also not really needed.
```suggestion
            "oil": round(float(fossil["value"]), 6),
            "wind": round(float(wind["value"]), 6),
            "solar": round(float(solar["value"]), 6),
            "biomass": round(float(biogas["value"]), 6),
```
or
```suggestion
            "oil": float(fossil["value"]),
            "wind": float(wind["value"]),
            "solar": float(solar["value"]),
            "biomass": float(biogas["value"]),
            "unknown": unknown,
```

VIKTORVAV99 (commented):
Hehe the database should be fine. It should index things by datetime so any duplicates would be overridden. Which mean that 2023-01-01 is more than likely having the wrong data. We should consider checking if all datetimes returned by the API/endpoint is the same but that means that the parser would still not function unfortunately.

parsers/AW.py
```diff
@@ -55,17 +55,17 @@ def fetch_production(
...
 
     data = {
         "zoneKey": zone_key,
         "datetime": zone_date_time.datetime,
         "production": {
-            "oil": float(fossil["value"]),
-            "wind": float(wind["value"]),
-            "solar": float(solar["value"]),
-            "biomass": float(biogas["value"]),
+            "oil": round(float(fossil["value"]), 3),
```
q--, line 65: Never really been fond of discarding data when it's not necessary so I'm biased, but is there a point in doing this? If the backend still stories it as floats, it won't save anyone storage space (assuming no compression), and if it doesn't, it'll do away with the extras precision by itself when converting it to whatever data type the backend uses. 

By the way, if the point is to get rid of floating point errors, rounding to 13 decimal points will suffice.

parsers/AW.py
VIKTORVAV99: > If we want to round them and assume these are in MW, we should round them to six decimals (1 W) not three (1kw). But I really don't see a need to round them. However, we might consider rounding the unknown one to six decimals as it's computed but it's also not really needed or to thirteen decimals to get rid of floating-point errors as suggested by @q-- below.
> 
> or

parsers/AW.py
VIKTORVAV99: Are you using `poetry run test_parser AW` for example? That should give you a pretty printed output.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5092, opened 2023-02-14
[CI]: Run tests in parallel

### LINKED ISSUE #5091: [CI]: Make sure Python Tests run on config changes
Currently some of our config validation is included in the Python Test but we only run these test when there is a python change. This need to be changed.

VIKTORVAV99:
## Issue
fixes #5091 

This should prevent any non valid config files from slipping through.


.github/workflows/python_formatting.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: .venv
+          key: ${{ runner.os }}-venv-${{steps.get-poetry-version.outputs.poetry-version}}-${{ hashFiles('poetry.lock') }}
+          restore-keys: ${{ runner.os }}-venv-${{steps.get-poetry-version.outputs.poetry-version}}
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          poetry install -E parsers --sync
```
madsnedergaard, line 36: Do we need to install all parser dependencies? If it simplifies caching having the same thing it's fine, just wanted to see :)

VIKTORVAV99: Probably not but I didn't want to change things before we move to poetry 1.2 syntax as these are cached anyway. When we do move to poetry 1.2 syntax we can create a "formatting" group and a parser "group" to optimize this.

PR #5093, opened 2023-02-14
Ignore paths instead of query for CodeQL

VIKTORVAV99:
Instead of excluding a query we should ignore the test paths instead. We can expand this to the python tests later if necessary.


madsnedergaard (approved):
Makes much sense, thanks! 🙌

PR #5096, opened 2023-02-15
Fix the last ESLint errors and enable CI

### LINKED ISSUE #5007: Enable ESLint again
## Description
We currently don't have the ESLint CI job enabled due to a lot of errors and warnings that was introduced with TypeScript in combination with the rewrite.

TODO:

- [x] #5017
- [x] #5016
- [x] Fix in loops ##5049
- [x] Fix process.exit() ##5049
- [ ] Get rid of __dirname
- [x] Change the array.reduce() methods.
- [x] #5050
- [ ] Enable ESLint CI job

Bonus:
- [ ] Fix as many ESLint warnings as possible

VIKTORVAV99:
Closes #5007
Closes ELE-1607

So this is the last PR in the ESLint fix quest. It fixes all the eslint issues but the Code Scanning job for ESLint fails because of all the tsc any issues (these really should be warnings but are not) but I think these should be considered non-blocking as they should only error for this PR. And guess what, most of the errors are in the wind and solar layers. 😅


VIKTORVAV99:
The only thing I am unsure about in this PR is if we should continue using the sarif formatter for eslint (it's the one reporting all the errors to github code scanning) or just use one of the builtin ones. There are benefits to both but I don't want to spam the security dashboard with a bunch of errors that really should just be warnings. 🫤

madsnedergaard:
> The only thing I am unsure about in this PR is if we should continue using the sarif formatter for eslint (it's the one reporting all the errors to github code scanning) or just use one of the builtin ones. There are benefits to both but I don't want to spam the security dashboard with a bunch of errors that really should just be warnings (it looks like either the sarif formatter or github is ignoring the disable comments) 🫤

Once we have cleaned up typings and fixed all tsc errors, I think we should ensure that all following PRs are blocked if they introduce type errors. Does that help with the decision? :D 
(Will review this PR soon!)

VIKTORVAV99:
> > The only thing I am unsure about in this PR is if we should continue using the sarif formatter for eslint (it's the one reporting all the errors to github code scanning) or just use one of the builtin ones. There are benefits to both but I don't want to spam the security dashboard with a bunch of errors that really should just be warnings (it looks like either the sarif formatter or github is ignoring the disable comments) 🫤
> 
> Once we have cleaned up typings and fixed all tsc errors, I think we should ensure that all following PRs are blocked if they introduce type errors. Does that help with the decision? :D (Will review this PR soon!)

This will flag all new errors and fail the CI (which is why this PR is failing). If we make this a required check and update the level on type errors from warnings to errors it will prevent any new type errors from entering the codebase. The only one(s) that could merge things at that point would be you or another admin 😅

web/geo/generateExchangesToExclude.ts
```diff
@@ -1,5 +1,5 @@
-import { mergeExchanges } from '../scripts/generateZonesConfig';
-import { fileExists, getJSON, writeJSON } from './utilities';
+import { mergeExchanges } from '../scripts/generateZonesConfig.js';
+import { fileExists, getJSON, writeJSON } from './utilities.js';
```
madsnedergaard, line 2: Why `.js` here when source is in TS? 🤔

web/tsconfig.node.json
```diff
@@ -10,5 +10,5 @@
     "strict": true,
     "target": "ESNext"
   },
-  "include": ["vite.config.ts", "cypress.config.ts", "scripts", "geo"]
+  "include": ["vite.config.cts", "cypress.config.ts", "scripts", "geo"]
```
madsnedergaard, line 13: Ugh, so annoying with all the CommonJS file extensions 😬

madsnedergaard (approved):
A few comments, otherwise LGTM :)

web/geo/generateExchangesToExclude.ts
VIKTORVAV99: This is part of the cascading issue that caused so many seemingly unrelated changes.
First using the meta imports forced the us to use pure js modules and set the flag in the package.json.
This in turn forced us/me to add extensions to the files that aren't used in the bundler.
Since we use typescript and tsc to compile these it would be logical if they where ts files but tsc only changes typescript related code and don't touch the imports (unless they are marked as types) which means that we have to use the js extension because it will be js when it's compiled and run.

It's all a little messy but should only be a one time change. Hopefully 😅

web/tsconfig.node.json
VIKTORVAV99: We could probably convert it to a module and just use .ts but I didn't want to make more changes than absolutely necessary in this PR.

VIKTORVAV99:
Merging this now, let's hope I don't break everything... 😅

madsnedergaard:
Unfortunately it did break Storybook 😄 
Dunno if easily fixable though

Screenshot 2023-02-23 at 15 33 58

VIKTORVAV99:
Oh that's not good, and it's like the only thing I didn't test...

Looks like some sort for internal conflict between CJS loaders and the new module syntax.

I'll take a look and see if I can find a easy solution.

VIKTORVAV99:
https://github.com/electricitymaps/electricitymaps-contrib/pull/5120 should take cake of it.

PR #5098, opened 2023-02-16
update in-no config

mathilde-daugy:
## Description

Update unknown emission factor as the unknown share of production is mainly biomass and hydro. The EF calculation was done using CEA values for 2022
[IN unknown emission calculation ](https://docs.google.com/spreadsheets/d/1on86UoTcljjLztobbPWMMXZZvXON12uKAGIH1ftTypw/edit#gid=0)

### Preview

NA


config/zones/IN-NO.yaml
```diff
@@ -20,7 +20,17 @@ contributors:
...
 timezone: Asia/Kolkata
+emissionFactors:
+  direct:
+    unknown:
+      source: CEA 2022; assumes 56% biomass, 36% hydro, 8% unknown
+      value: 45
+  lifecycle:
+    unknown:
+      source: CEA 2022; assumes 56% biomass, 36% hydro, 8% unknown
+      value: 192
```
VIKTORVAV99, line 33: We should also add isRenewable and isLowCarbon values as well so we get the "correct" renewable and low carbon share.

VIKTORVAV99 (changes requested):
We just need to add isRenewabe and isLowCarbon overrides for unknown as well.

config/zones/IN-NO.yaml
```diff
@@ -20,7 +20,23 @@ contributors:
...
+isLowCarbon:
+  unknown:
+    value: 1
```
VIKTORVAV99, line 36:
```suggestion
    value: 0.92
```
Shouldn't it be the sum of known renewable of the unknown?

config/zones/IN-NO.yaml
```diff
@@ -20,7 +20,23 @@ contributors:
...
+isRenewable:
+  unknown:
+    value: 1
```
VIKTORVAV99, line 39:
```suggestion
    value: 0.92
```

VIKTORVAV99 (approved):
I have a suggestion but otherwise things look good! 👍🏼

config/zones/IN-NO.yaml
mathilde-daugy: i guess but the remainder is also other renewables according to the data source so I put 1 
I think we can put 0.92 to be on the safe side

VIKTORVAV99: Ah in that case I leave it up to you 👍🏼

mathilde-daugy: i'll put 0.92 as we have no idea what the 8% are and are using the default unknown EF

PR #503, opened 2017-04-26
add information for Canadian Atlantic provinces (#166)

jarek:
Added information for PEI and New Brunswick as much as possible. New Brunswick doesn't provide a breakdown, but total production is possible to calculate, and their central location makes it worth it for the different exchanges. Capacity is about 60-70% fossil, so 'unknown' should reflect CO2 within an order of magnitude at least...

I looked into Newfoundland and Labrador's https://www.nlhydro.com/system-information/system-information-center/ listed in #166, but it only contains current load, possibly only for Newfoundland (without Labrador), and has no breakdown by source - so it doesn't seem that useful.

Amended Nova Scotia parser to return the actual load, rather than assume 100 MW.

Added capacities for NB, NS, NL, and PEI by going through Wikipedia lists. They might not be perfect.

It looks like PEI's and NB's reported exchanges with each other don't always match up - maybe they count things differently? Would like to see how it looks in electricitymap before making tweaks, if possible.

Added rough locations of exchanges for the map. I didn't preview them in the front-end, so tweaks might be necessary.

Hope this is workable!

PR #5100, opened 2023-02-16
Create Newfoundland consumption-parser

alixunderplatz:
I've overlooked that we already had a thread for Newfoundland in #2541 and #503 where the source was mentioned back then, and even a parser was written but PR #3077 was closed. So hopefully this is no duplicate.
The parser fetches consumption for CA-NL-NF only, as there is no further breakdown of the fuel types. 5-minute-granularity.

### DIFF
--- a/parsers/CA-NL-NF.py
+++ b/parsers/CA-NL-NF.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+from datetime import datetime
+from logging import Logger, getLogger
+from typing import Optional
+
+from bs4 import BeautifulSoup
+from pytz import timezone
+from requests import Session
+
+# This parser fetches demand data for Canada-Newfoundland. The source-website https://nlhydro.com/ states:
+# "Current Island System Generation", so I assume only generation for Newfoundland island, excluding Labrador, is shown.
+# "Current system generation = current demand on the system" and
+# "Any generation from non-utility generators, including Star Lake, Rattle Brook, Corner Brook Pulp and Paper,
+# and wind generation from the St. Lawrence and Fermeuse wind farms, is included in our Hydro System Generation."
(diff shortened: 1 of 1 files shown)


jarek:
It's not really clear to me how https://nlhydro.com/system-information/supply-and-demand/ still says "Current system generation = current demand on the system" considering they now have the Maritime Link providing some energy from Labrador (Muskrat Falls) and to Nova Scotia.

q--:
> It's not really clear to me how https://nlhydro.com/system-information/supply-and-demand/ still says "Current system generation = current demand on the system" considering they now have the Maritime Link providing some energy from Labrador (Muskrat Falls) and to Nova Scotia.

For reference, @Hurc2000  compared the data from that page to the daily reports https://github.com/electricitymaps/electricitymaps-contrib/issues/2541#issuecomment-1451165792 and confirmed that demand indeed isn't the same as supply

madsnedergaard:
Hey @alixunderplatz, it's been a while, hope you're doing good :D
Do you have any memory on what the state of this PR was and how we can get it merged? Would be nice to get Newfoundland on the map, so I'm also happy to help out :)

q--:
On cursory glance, it appears the page now separately accounts for Muskrat Falls Generation and Churchill Falls Generation, which is great because that was one of the issues with the data.  
(These would still need to be added to the parser. Also, there might be other obstacles, I don't have time to go refresh my memory right now, but #2541 should have the details.)

jarek:
I wrote a bunch at the issue https://github.com/electricitymaps/electricitymaps-contrib/issues/2541#issuecomment-2489960694

But the summary is that we can still only get Newfoundland (CA-NL-NF) consumption as reliable real-time data.

madsnedergaard:
Thanks for the update, I really appreciate that! 🙏 
In that case I suggest we close this PR for now, but keep the issue open :)

PR #5102, opened 2023-02-17
Tonyvanswet/exchange arrow tooltip mobile

tonypls:
## Issue
Enable zooming the map when hovering an exchange arrow 
## Description

This PR was merged but this is an update to ensure the event listener for preventing wheel events is only attached to the exchange layer


VIKTORVAV99 (approved):
Seems to work as expected now and scrolling isn't overwritten in the side panel. Nice work!

The only think that can be improved as far as I can tell is that the tooltip don't show back up after you have scrolled until you move the mouse every so slightly but I don't think that should be any blocking thing, it's just a small note on what we can improve in the future when we have the time. 👍🏼

madsnedergaard (approved):

PR #5103, opened 2023-02-17
include capacity for JP-TH for map visualization

wobniarin:
## Issue

JP-TH zone is still not visible on the map. 

## Description

JP-TH parser has been validated and configuration file has been updated with: 
- capacities 
- disclaimer to state that at the moment only solar and unknown will be shown, but that we are working on the source breakdown.


VIKTORVAV99:
I think the reason it don't show up on the map is because it's excluded from flowtracing in the flowtracing lib for some reason.

I'll send you a direct link to it on slack.

config/zones/JP-TH.yaml
```diff
@@ -3,9 +3,22 @@ bounding_box:
...
+  hydro: 2000
+  hydro storage: 460
+  nuclear: 2750
+  oil: 2500
+  solar: 7000
+  wind: 1800
+  unknown: 0
 contributors:
   - tmslaine
   - lorrieq
```
VIKTORVAV99, line 20: If you want you can include your github username here, but it's not required.

VIKTORVAV99 (approved):
I generated the updated `zones.json` for you and the other changes look good!

pierresegonne (approved):
LGTM, simple and clean!

PR #5105, opened 2023-02-17
Update PR template

VIKTORVAV99:
## Description

Updates the PR templates with checkboxes to remind people to test their changes locally and format their code.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,6 @@
 ## Issue
 
-<!-- If you want to close a issue with your PR use magic words such as closes, fixes or resolves "issue number" to autmatically link a PR to an issue. For example: Closes #0000, Fix: #0000 or Resolves: #0000 -->
-
+<!-- If you want to close an issue automatically when your PR is merged, write "Closes X" where X is the PR number. For example: Closes #000 -->
 ## Description
 
 <!-- Explains the goal of this PR -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,7 @@
 ## Issue
 
 <!-- If you want to close an issue automatically when your PR is merged, write "Closes X" where X is the PR number. For example: Closes #000 -->
+
 ## Description
 
 <!-- Explains the goal of this PR -->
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,9 +1,16 @@
 ## Issue
 
+<!-- If you want to close a issue with your PR use magic words such as closes, fixes or resolves "issue number" to autmatically link a PR to an issue. For example: Closes #0000, Fix: #0000 or Resolves: #0000 -->
+
```
madsnedergaard, line 4: Let's give fewer options for the non-github-savy users :D

```suggestion
<!-- If you want to close an issue automatically when your PR is merged, write "Closes X" where X is the PR number. For example: Closes #000 -->

```

.github/pull_request_template.md
```diff
@@ -1,9 +1,16 @@
...
 <!-- Explains the goal of this PR -->
 
 ### Preview
 
 <!-- Please add screenshots and/or gif that shows visual changes (if applicable) -->
+
+### Double check
+
+- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
+- [ ] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.
```
madsnedergaard, line 16: In an ideal world this should be handled automatically by tools, but until then it's nice to add the check here :)

madsnedergaard (commented):
A small comment, otherwise great changes here!

.github/pull_request_template.md
VIKTORVAV99: Yeah I had a quick discussion with Pierre about it and the last one is already checked by the bots (but people seem to forget it anyway 😅). The first one is a lot trickier, we would have to calculate the changed parsers and then run them (and hope the source isn't down) or require tests to be implemented in each parser.

.github/pull_request_template.md
VIKTORVAV99: Fair enough 😅

pierresegonne:
Nice, thanks for the quick turnaround!

pierresegonne (approved):

.github/pull_request_template.md
madsnedergaard: The second one should ideally be a pre-commit thing so that forgetting it is not an option :D 

Good point about the first one!

VIKTORVAV99: Let's remove the second one when we add the pre-commit 👍🏼

PR #5107, opened 2023-02-19
[TW] add geothermal from taipower.com.tw

Manu1400:
## Issue
Add geothermal in TW zone (🇹🇼) 

## Description

Taiwan have geothermal 5MW installed.

Get live data from taipower.com.tw

### DIFF
--- a/parsers/TW.py
+++ b/parsers/TW.py
@@ -73,9 +73,10 @@ def fetch_production(
     check_values = production.output <= production.capacity
     modes_with_capacity_exceeded = production[~check_values].index.tolist()
     for mode in modes_with_capacity_exceeded:
-        logger.warning(
-            f"Capacity exceeded for {mode} in {zone_key} at {dumpDate.datetime}"
-        )
+        if mode != "Geothermal":
+            logger.warning(
+                f"Capacity exceeded for {mode} in {zone_key} at {dumpDate.datetime}"
+            )
 
     coal_capacity = (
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (changes requested):
Okay so this PR can't be merged as is.
It currently don't work because geothermal is not a fueltype in the returned JSON and it's production and capacity is currently included in "Other Renewable Energy" (which don't seem to be returned by the parser.).

While there looks like there could be a way to extract it, it would require a significant rewrite of the parser. If you feel like you have the time and knowledge to do so feel free otherwise I suggest just updating the parser to include the "Other Renewable Energy" production in the unknown category and then I can look into rewriting the parser in a separate PR.

parsers/TW.py
```diff
@@ -96,6 +97,7 @@ def fetch_production(
         "production": {
             "coal": coal_production,
             "gas": gas_production,
+            "geothermal": production.loc["Geothermal"].capacity,
```
jarek, line 100: Was this meant to be `production.loc["Geothermal"].output` ?

VIKTORVAV99: It may have been but right now that would have been undefined as there is no Geothermal "class" in the dataset.

VIKTORVAV99:
@Manu1400 what is the status here? Would be great to get this added.

If you need any assistance just let me know and I'll help as soon as I can.

PR #5112, opened 2023-02-20
1899/czech parser

Frantisek12:
## Description
replace ENTSOE data provider with the official Czech grid maintainer CEPS

### Double check


VIKTORVAV99 (commented):
Some early feedback before a full review.

It looks like we can replace arrow with just native datetime and timedelta together with the `datetime.isoformat()` function for formatting it. That should help geting rid of some of the type errors.

Some of the functions here seem to be a little more complex than they need to be. But I'll take a look at that tomorrow when I do a full review.

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+    data_tag = xml.find("data")
+    data_list = []
+
+    for values in data_tag:
+        data = {
+            "zoneKey": zone_key,
+            "production": {},
+            "storage": {},
+            "source": url,
+            "datetime": datetime.fromisoformat(values["date"]).isoformat(),
```
VIKTORVAV99, line 106:
```suggestion
            "datetime": datetime.fromisoformat(values["date"]),
```
This should be returned as a datetime object, not a string.

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        for k, v in mapper.items():
+            generator = translate_table_gen[k]
+            if k != "PsPP":
+                data["production"][generator] = float(values[v])
+            else:
+                data["storage"][generator] = float(values[v])
```
VIKTORVAV99, line 114:
```suggestion
            else:
                data["storage"][generator] = float(values[v]) * -1
```
It's a bit counter intuetive but we currently treat negative values as production and positives ones as "charging"

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+    xml = BeautifulSoup(content, "xml")
+    mapper = get_mapper(xml)
+
+    data_tag = xml.find("data")
+    data_list = []
+
+    for values in data_tag:
+        data = {
+            "sortedZoneKeys": f"{zone_key1}->{zone_key2}",
+            "datetime": datetime.fromisoformat(values["date"]).isoformat(),
```
VIKTORVAV99, line 158:
```suggestion
            "datetime": datetime.fromisoformat(values["date"]),
```
Same here as above.

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+    zone_key: str = "CZ",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+):
+
+    if not target_datetime:
+        target_datetime = arrow.now().replace(minute=0, second=0)
+    from_datetime = target_datetime.shift(days=-5)
+
```
VIKTORVAV99, line 189: This still use arrow, if we switch this to datetime like the others when we can remove the arrow import.

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+    content = make_request(session, payload, zone_key).content
+    xml = BeautifulSoup(content, "xml")
+    mapper = get_mapper(xml)
+
+    data_tag = xml.find("data")
+    data_list = []
+
+    for values in data_tag:
```
VIKTORVAV99, line 100: There is a posibility that data_tag is None here.

We should check that data_tag is not none here before running the loop.
Like this:
```python
if data_tag is not None:
  for values in data_tag:
  ...
else:
  ParserException("CZ.py", f"There was no data returned for {zone_key} at {target_datetime}", zone_key)
```

parsers/CZ.py
```diff
@@ -0,0 +1,221 @@
...
+def fetch_production(
+    zone_key: str = "CZ",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> Union[List[dict], dict]:
+
+    if not target_datetime:
+        target_datetime = datetime.now().replace(minute=0, second=0, microsecond=0)
+    from_datetime = target_datetime - timedelta(hours=1)
```
VIKTORVAV99, line 75:
```suggestion
    from_datetime = target_datetime - timedelta(hours=48)
```
Let's get the last 48 hours to match the ENTSO-E parser.

(This should be added to all of the methods.)

VIKTORVAV99 (changes requested):
Everything is working but there are a few improvemnts that need to be made here and I have added suggestions for most.

I'll re-review this when the changes have been addressed. If there are any questions don't hesitate to ping me or reach out on slack.

parsers/CZ.py
Frantisek12: this will be problematic. CEPS is not publishing data about charging. I can write an email to them asking whether it is somewhere in the data set but I couldn't find it. It simply returns positive values once producing nothing more.

VIKTORVAV99: That's fine (I think) but positive production need to be returned as a negative value.

(We really want to change this but fixing this tiny thing would require a lot of changes 😅)

config/exchanges/AT_CZ.yaml
```diff
@@ -2,6 +2,6 @@ lonlat:
   - 15.486554
   - 48.909846
 parsers:
-  exchange: ENTSOE.fetch_exchange
-  exchangeForecast: ENTSOE.fetch_exchange_forecast
+  exchange: CZ.fetch_exchange
+  exchangeForecast: CZ.fetch_exchange_forecast
```
VIKTORVAV99, line 6: This function has not been implemented in the CZ parser. Let's revert it to ENTSOE.
```suggestion
  exchangeForecast: ENTSOE.fetch_exchange_forecast
```

parsers/CZ.py
```diff
@@ -0,0 +1,206 @@
...
+    data_list = []
+
+    if data_tag is not None:
+
+        for values in data_tag:
+            data = {
+                "zoneKey": zone_key,
+                "production": {},
+                "storage": {},
+                "source": url,
```
VIKTORVAV99, line 107: The source should just be the domain i.e. `ceps.cz`.
(This should be implemented in both functions.)

config/exchanges/CZ_DE.yaml
```diff
@@ -2,6 +2,6 @@ lonlat:
   - 12.321836
   - 50.227335
 parsers:
-  exchange: ENTSOE.fetch_exchange
-  exchangeForecast: ENTSOE.fetch_exchange_forecast
+  exchange: CZ.fetch_exchange
+  exchangeForecast: CZ.fetch_exchange_forecast
```
VIKTORVAV99, line 6: Same here:
```suggestion
  exchangeForecast: ENTSOE.fetch_exchange_forecast
```

config/exchanges/CZ_PL.yaml
```diff
@@ -2,6 +2,6 @@ lonlat:
   - 16.496641
   - 50.269487
 parsers:
-  exchange: ENTSOE.fetch_exchange
-  exchangeForecast: ENTSOE.fetch_exchange_forecast
+  exchange: CZ.fetch_exchange
+  exchangeForecast: CZ.fetch_exchange_forecast
```
VIKTORVAV99, line 6: Same here:
```suggestion
  exchangeForecast: ENTSOE.fetch_exchange_forecast
```

config/exchanges/CZ_SK.yaml
```diff
@@ -2,6 +2,6 @@ lonlat:
   - 18.100645
   - 49.089498
 parsers:
-  exchange: ENTSOE.fetch_exchange
-  exchangeForecast: ENTSOE.fetch_exchange_forecast
+  exchange: CZ.fetch_exchange
+  exchangeForecast: CZ.fetch_exchange_forecast
```
VIKTORVAV99, line 6: Same here:
```suggestion
  exchangeForecast: ENTSOE.fetch_exchange_forecast
```

parsers/CZ.py
```diff
@@ -0,0 +1,206 @@
...
+        "Content-Type": "application/soap+xml; charset=utf-8",
+        "Content-Length": "1",
+    }
+
+    r = session or Session()
+    res = r.post(url, headers=headers, data=payload)
+    assert res.status_code == 200, (
+        "Exception when fetching production for "
+        "{}: error when calling url={}".format(zone_key, url)
+    )
```
VIKTORVAV99, line 62:
```suggestion
def make_request(session: Session, payload: str, zone_key: str):
    headers = {
        "Content-Type": "application/soap+xml; charset=utf-8",
        "Content-Length": "1",
    }

    res: Response = session.post(url, headers=headers, data=payload)
    assert res.status_code == 200, (
        f"Exception when fetching production for {zone_key}: error when calling {url}"
    )
```
Let's clean this up a bit and use the newest syntax.

parsers/CZ.py
```diff
@@ -0,0 +1,206 @@
...
+
+    return res
+
+
+def fetch_production(
+    zone_key: str = "CZ",
+    session: Optional[Session] = None,
```
VIKTORVAV99, line 69:
```suggestion
    session: Session = Session(),
```
This way we can get rid of the conditional in the make_request function.

parsers/CZ.py
```diff
@@ -0,0 +1,206 @@
...
+            zone_key,
+        )
+
+    return data_list
+
+
+def fetch_exchange(
+    zone_key1: str = "CZ",
+    zone_key2: str = "DE",
+    session: Optional[Session] = None,
```
VIKTORVAV99, line 133:
```suggestion
    session: Session = Session(),
```
Same here.

VIKTORVAV99 (commented):
This PR is making some nice progress so good work!

I have some more suggestions but after that I think this is more or less ready to be merged. 🙂

VIKTORVAV99 (commented):
Everything looks great! Thanks for implementing this!

Just one final question to double check something then we can merge this.
Does is the datetime forward looking or backwards looking?
I.e. does 15:00 mean the production between 15:00 and 15:14:59 or 14:45:01 and 15:00?

Edit:
Oh and if you wish please do add yourself to the list of contributors in the CZ.yaml file.

Frantisek12:
DateTime is backward looking as you need to provide fromDate and TargetDate to obtain results. The result is in the range of these two parameters.  I've added my GitHub nickname among contributors in the CZ.yaml file.

VIKTORVAV99:
> DateTime is backward looking as you need to provide fromDate and TargetDate to obtain results. The result is in the range of these two parameters. I've added my GitHub nickname among contributors in the CZ.yaml file.

Okay then we need to adjust these datetimes to be forward looking as that is what the backend expects.
If you need assistance with how to do this in a safe way let me know and I'll provide some guidance.

PS: It don't look like you pushed the changes with your nickname to this branch yet.

Frantisek12:
Hi Viktor, sorry for the delay. Yes, please give me some hints on how to do it as I have no idea now. I guess you are right I haven't pushed it yet

VIKTORVAV99:
Okay so I have dug into the code a bit and it does look like it's returning the correct format so that's great!

I did find a place where we can improve things though, if you add the following function we can both improve things and reduce code duplication.

Here is the function we need to add:
```python
def get_target_datetime(dt: Optional[datetime]) -> datetime:
    if dt is None:
        now = datetime.now()
        dt = (now - timedelta(minutes=now.minute % 15)).replace(second=0, microsecond=0)
    return dt
```
This function gets the nearest 15 minute step instead of the full hour like before and can be re-used across the functions.

And here is what we need to replace:
```diff
-    if not target_datetime:
-        target_datetime = datetime.now().replace(minute=0, second=0, microsecond=0)
+    target_datetime = get_target_datetime(target_datetime)
```

parsers/CZ.py
```diff
@@ -0,0 +1,237 @@
...
+          <agregation>{2}</agregation>
+          <function>{3}</function>
+          <version>{4}</version>
+        </CrossborderPowerFlows>
+      </soap12:Body>
+    </soap12:Envelope>""".format(
+        from_datetime.isoformat(), target_datetime.isoformat(), "QH", "AVG", "RT"
+    )
+
+    content = make_request(session, payload, zone_key1).content
```
VIKTORVAV99, line 100:
```suggestion
    content = make_request(session, payload, zone_key1).text
```
Won't change anything functionality wise but should satisfy the the type requirement for BeautifulSoup better.

parsers/CZ.py
```diff
@@ -0,0 +1,237 @@
...
+                  <function>{3}</function>
+                  <version>{4}</version>
+                  <para1>{5}</para1>
+                </Generation>
+              </soap12:Body>
+            </soap12:Envelope>""".format(
+        from_datetime.isoformat(), target_datetime.isoformat(), "QH", "AVG", "RT", "all"
+    )
+
+    content = make_request(session, payload, zone_key).content
```
VIKTORVAV99, line 166:
```suggestion
    content = make_request(session, payload, zone_key).text
```
Same here 🙂

VIKTORVAV99 (commented):
I added a bit more feedback than necessary to get this merged since you requested it but it's only the last comment (about the exchangeForecasts) that need to be resolved before we can merge this.

VIKTORVAV99 (approved):
LGTM! 🎉

Let's get this merged! Thanks for taking the time to create this parser and bearing with me in making the requested changes.

VIKTORVAV99:
For some reason the CI is failing (I don't think it should) but I don't have time to look into that more right now but I'll try and figure out why it's failing this weekend.

PR #5116, opened 2023-02-23
Exchange config fixes for IE & GB-NIR

msaynevirta:
## Issue

Exchanges between the UK, Ireland and Northern Ireland seem to be configured incorrectly. Flows in the Moyle interconnector are reversed compared to the actual ones displayed by the EirGrid / SONI dashboard. GB-IE flows are incorrectly shown a second time between GB-NIR and IE due to an incorrect exchange config. The IE parser doesn't currently support this specific exchange.

## Description

Flow direction returned by the IE parser has been reversed and GB-NIR_IE interconnection has been disabled in the exchange config.


VIKTORVAV99 (commented):
It's great that you spotted this! 
> The IE parser doesn't currently support this specific exchange.

Could you maybe add a check so this raise a parser exception if the exchange is not supported? It should make sure this don't happen again.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing this and adding the test!
@mathilde-daugy can you remove the erroneous data from the database?

mathilde-daugy:
@VIKTORVAV99, @msaynevirta I'm not sure I understand why we are removing the GB-NIR->IE exchange. Could you please explain the logic behind this?
Thanks :)

VIKTORVAV99:
> @VIKTORVAV99, @msaynevirta I'm not sure I understand why we are removing the GB-NIR->IE exchange. Could you please explain the logic behind this? Thanks :)

If I understood, it correctly we are accidentally using the same data for the exchange between GB-NIR->IE and GB->IE (identical flows) when we should only be using it for GB-IE where the Moyle interconnection is physically located. It also seems we were parsing the flow directions wrong.

But it was @msaynevirta that discovered it so it might be best to wait for a full explanation from them.

mathilde-daugy:
> If I understood, it correctly we are accidentally using the same data for the exchange between GB-NIR->IE and GB->IE (identical flows) when we should only be using it for GB-IE where the Moyle interconnection is physically located. It also seems we were parsing the flow directions wrong.
> 
> But it was @msaynevirta that discovered it so it might be best to wait for a full explanation from them.

Ok, I look into it and the North-South interconnector seems to be a project still. If @msaynevirta can confirm then we update and merge this PR.

mathilde-daugy (approved):

msaynevirta:
Sure, I'll try to explain the changes. Looking at the all-island grid, we have three operational interconnections between the three jurisdictions (GB-->IE, GB-->GB-NIR and GB-NIR-->IE). Of these three, EirGrid and SONI provide data for only two of them, i.e. the DC interconnectors EWIC between GB-->IE and Moyle between GB-->GB-NIR.

There is a synchronized AC connection between GB-NIR-->IE, but clean information for it isn't currently available from any data source. This data is likely baked into the generation / interconnection / consumption data in the Eirgrid / SONI smart grid dashboard, but thus far I think no one has been able to clearly separate it.

Regarding the fixes, what was happening previously was that the parser was pulling the GB-->IE data and was showing it as flows between GB-NIR-->IE. This is due to only the first zone id being checked in the IE parser when collecting exchange data (GB-->IE vs GB-NIR-->IE).

Also, the flow direction between GB-->GB-NIR was reversed compared to the actual in the IE parser.

mathilde-daugy:
Thank you very much for the clarification and for taking the time to look into this and review the parser!

PR #5118, opened 2023-02-23
Fixes TypeScript errors

madsnedergaard:
### Preview

**Before:** `Found 67 errors in 26 files.`

**After:** `Found 37 errors in 16 files.`

### Double check


web/src/components/MetricRatio.tsx
```diff
@@ -1,6 +1,15 @@
-export function MetricRatio({ value, total, format }) {
-  const formattedValue = Number.isFinite(value) ? format(value) : '?';
-  const formattedTotal = Number.isFinite(total) ? format(total) : '?';
+import type { Maybe } from 'types';
+
+interface MetricRatioProps {
+  value: number;
+  total: Maybe<number>;
```
VIKTORVAV99, line 5: While it's an improvement shouldn't this always be either null or a number? Would it not be better to mark the it as such instead of maybe number?

web/src/utils/formatting.ts
```diff
@@ -15,7 +15,7 @@ export const formatPower = function (d: number, numberDigits = DEFAULT_NUM_DIGIT
   return power;
 };
 
-export const formatCo2 = function (d, numberDigits = DEFAULT_NUM_DIGITS) {
+export const formatCo2 = function (d: number, numberDigits = DEFAULT_NUM_DIGITS) {
```
VIKTORVAV99, line 18:
```suggestion
export const formatCo2 = function (d: number, numberDigits: number = DEFAULT_NUM_DIGITS) {
```
While it can implicitly infer that this is a number marking it as one will ensure DEAFULT_NUM_DIGITS is always a number and error if it changes for whatever reason.

web/src/features/weather-layers/solar/stackBlurImageOpacity.ts
```diff
@@ -1,3 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+/* eslint-disable unicorn/no-abusive-eslint-disable */
+/* eslint-disable */
```
VIKTORVAV99, line 4: I thought we where getting rid of these 😅

VIKTORVAV99 (commented):
Just some suggestions and general feedback.

web/src/features/weather-layers/solar/stackBlurImageOpacity.ts
madsnedergaard: Not for this nasty old code 😉

web/src/components/MetricRatio.tsx
madsnedergaard: Since `total` can be `capacity` which comes all the way back from `ZoneDetails`, it should either accept the same type or we should add a check somewhere in between that ensures it's only `null` and not `undefined`. I don't think there's much value in enforcing this right now though, so leaving it as `Maybe<number>` and then we can rethink the whole `undefined` vs `null` (vs `NaN`) later on :)

web/src/utils/formatting.ts
madsnedergaard: Good point!

web/src/components/MetricRatio.tsx
VIKTORVAV99: We could add it as `total?: number | null` but I suppose that just does the same as `Maybe<number>`?

madsnedergaard: `Maybe<>` allows both undefined and null. I'd rather be too cautious here and then later on change Maybe type where possible - it means reworking types throughout the whole app, so I don't think we should address it right now ;)

madsnedergaard: I actually just digged further, and the main issue here is that on non-hourly aggregrates we don't get capacity data from the backend. 
So therefore capacity values can be either `undefined`, `null` or a `number`. 
We could write it in either of the ways below, but don't know if it matters much :)
```
capacity?: number | null; 
// capacity: Maybe<number>;
```

VIKTORVAV99: Maybe we should start returning data for these even if it's null in preparation for capacity parsers (some zones should already have de facto capacity parsers as it's returned as a part of the production object but I don't know how this is saved)?

ENTSO-E TP already have yearly installed capacity for most zones so we just need to save this to a database really. (and use the capacity in the config as a fallback.)

madsnedergaard: But then we're back to the problem of trying to reduce backend payload size! :D

VIKTORVAV99: I don't think it's the zone state files that are the issue right now but the global one (that is always downloaded and refetched).

web/src/components/MetricRatio.tsx
```diff
@@ -1,6 +1,14 @@
...
+  format: (value: number) => string | number;
+}
+
+export function MetricRatio({ value, total, format }: MetricRatioProps) {
+  const formattedValue = Number.isFinite(value) && value ? format(value) : '?';
+  const formattedTotal = Number.isFinite(total) && total ? format(total) : '?';
```
VIKTORVAV99, line 11: Why is the added check here needed? Shouldn't isFinite only return true if its a defined number?

web/src/features/weather-layers/solar/stackBlurImageOpacity.ts
VIKTORVAV99:
```suggestion
/* eslint-disable */
// @ts-nocheck
```
Shouldn't this be enough? First disable eslint for the whole file which should supress the following ts-check warning?

web/src/features/weather-layers/wind-layer/windy.ts
```diff
@@ -532,8 +532,8 @@ var Windy = function (params: any) {
     started: any;
     field?: any;
     params?: any;
-    start?: (bounds: any, width: any, height: any, extent: any) => void;
-    stop?: () => void;
+    start: (bounds: any, width: any, height: any, extent: any) => void;
```
VIKTORVAV99, line 535: I really hope width and height are numbers 😅

VIKTORVAV99 (approved):
Looks good but I do have some questions and suggestions. They should be considered non blocking.

I also found a bunch of places where we should be using Enums but that can be changed in another PR.

I'm also not sure why ESLint is failing but it could be related to my seccond comment.

web/src/components/MetricRatio.tsx
madsnedergaard: Unfortunately not :( 

https://stackoverflow.com/questions/66703917/why-number-isfinite-dont-have-type-guard

VIKTORVAV99: Well that's weird... But that answers that question.

web/src/features/weather-layers/solar/stackBlurImageOpacity.ts
madsnedergaard: We still need unicorn/no-abusive-eslint-disable it seems, but the ban-ts-comment can be removed by reordering :)

web/src/features/weather-layers/wind-layer/windy.ts
madsnedergaard: Haha yes, but I don't think it's worth typing this file up without refactoring it completely - and I don't think that's the best use of our time for now :D

web/src/features/map/MapFallback.tsx
```diff
@@ -9,12 +9,6 @@ export function isOldIOSVersion() {
 
   // If version is less than 14.7 Maplibre won't work
   const lowestSupportedVersion = '14_7';
-  console.group('isOldIOSVersion');
-  console.log('isIos', isIos);
-  console.log('!isDevelopmentTools', !isDevelopmentTools);
-  console.log('iosVersion', iosVersion);
-  console.log('iosVersion < lowestSupportedVersio', iosVersion < lowestSupportedVersion);
-  console.groupEnd();
```
tonypls, line 17: My console is clean again <3

web/src/utils/viewport.ts
```diff
@@ -4,11 +4,12 @@ import { useCallback, useState } from 'react';
 export function useReferenceWidthHeightObserver(offsetX = 0, offsetY = 0) {
   const [width, setWidth] = useState(0);
   const [height, setHeight] = useState(0);
-  const [node, setNode] = useState(null); // The DOM node
+  const [node, setNode] = useState<any>(null); // The DOM node
```
tonypls, line 7: could be <Element | null>?

tonypls (approved):
LG, optional change

web/src/utils/viewport.ts
madsnedergaard: Good point - I had to update the weather layers and add some checks for this to work, but seems like they still work 🙈

VIKTORVAV99:
Looks like my potential fix for ESLint didn't fix this particular issue.
I'd merge it anyway and see if it's still there on master, if it is we can temporarily disable the sarif uploader and I'll try and fix it in another PR.

madsnedergaard:
Found the source of problem by looking into the SARIF file - it has duplicate suppressions for some reason, while try to readjust the "eslint disable" comments:
Screenshot 2023-02-24 at 16 45 17

madsnedergaard:
Ah, found it - it was caused by having a global eslint-disable comment + an eslint-disable-next-line comment in the same file! 😄

VIKTORVAV99:
> Ah, found it - it was caused by having a global eslint-disable comment + an eslint-disable-next-line comment in the same file! 😄

Huh I get why it does that but surly ESLint (or the formatter) should handle such things gracefully?

PR #5125, opened 2023-02-24
Changes how the cache key works for ESLint

VIKTORVAV99:
## Issue
There are some weird behavior with ESLint in on of the PR's.

## Description
This changes how the caches are handled (this is probably how it was supposed to be in the first place) 😅

### Preview
Previously it saved keys without the ref such as `Linux-eslint_cache--4261546878`

Now it saves keys with the ref like this: `Linux-eslint_cache-5125/merge-4262063308`


VIKTORVAV99:
Not sure if this fixes the issues you had in #5118 but it fixes something at least 😅

madsnedergaard (approved):
Haha great, thanks for (maybe) fixing it :D

PR #5126, opened 2023-02-24
[Typing Fixes] Fixes TypeScript errors related to charts

madsnedergaard:
## Issue

Part of https://github.com/electricitymaps/electricitymaps-contrib/issues/5122

## Description

More TS error fixes! Can be reviewed per commit :)

### Preview

```
After: Found 2 errors in 2 files.
```

### Double check


VIKTORVAV99:
It's weird that the ESLint job works here (and other PRs) but not in the layers PR, I can't see a obvious reason for it.

madsnedergaard:
> It's weird that the ESLint job works here (and other PRs) but not in the layers PR, I can't see a obvious reason for it.
Commented on the other PR, but the problem was that it had both a global eslint-disable comment + a eslint-disable-next-line comment - and apparently it does not like that 😅

web/src/features/charts/bar-breakdown/BarBreakdownEmissionsChart.tsx
```diff
@@ -26,7 +26,7 @@ interface BarBreakdownEmissionsChartProps {
   ) => void;
   onProductionRowMouseOut: () => void;
   onExchangeRowMouseOver: (
-    mode: ElectricityModeType,
+    mode: ZoneKey,
```
VIKTORVAV99, line 29: Should the mode really be a ZoneKey (which is just a string)?

web/src/features/charts/bar-breakdown/BarElectricityBreakdownChart.tsx
```diff
@@ -33,7 +33,7 @@ interface BarElectricityBreakdownChartProps {
   ) => void;
   onProductionRowMouseOut: () => void;
   onExchangeRowMouseOver: (
-    mode: ElectricityModeType,
+    mode: ZoneKey,
     data: ZoneDetail,
     event: React.MouseEvent<SVGPathElement, MouseEvent>
   ) => void;
```
VIKTORVAV99, line 39: This entire interface seem to be the exact same as the ` BarBreakdownEmissionsChartProps` interface in `BarBreakdownEmissionsChart.tsx` so we can probably deduplicate them.

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -39,15 +38,24 @@ export function getProductionCo2Intensity(
...
     return exchangeCo2Intensity;
   }
 
-  return getCO2IntensityByMode(zoneData, electricityMixMode);
+  // We don't use getCO2IntensityByMode in order to more easily return 0 for invalid numbers
+  if (electricityMixMode === Mode.CONSUMPTION) {
+    return zoneData.co2intensity || 0;
+  }
+
+  return zoneData.co2intensityProduction || 0;
```
VIKTORVAV99, line 58: Assuming these can only be undefined or a number as indicated by the type we should use nullish coalescing.
```suggestion
    return zoneData.co2intensity ?? 0;
  }

  return zoneData.co2intensityProduction ?? 0;
```

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -145,13 +153,14 @@ export interface ExchangeDataType {
...
 export const getExchangeData = (
   data: ZoneDetail,
-  exchangeKeys: string[],
+  exchangeKeys: ZoneKey[],
   electricityMixMode: Mode
 ): ExchangeDataType[] =>
-  exchangeKeys.map((mode) => {
+  exchangeKeys.map((mode: ZoneKey) => {
     // Power in MW
     const exchange = (data.exchange || {})[mode];
```
VIKTORVAV99, line 165:
```suggestion
    const exchange = data.exchange?.[mode];
```
Let's clean this up while we are here 🙂

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -145,13 +153,14 @@ export interface ExchangeDataType {
...
     const exchangeCapacityRange = (data.exchangeCapacities || {})[mode];
```
VIKTORVAV99, line 166:
```suggestion
    const exchangeCapacityRange = data.exchangeCapacities?.[mode];
```
And this 🙂

web/src/features/charts/hooks/useBreakdownChartData.ts
```diff
@@ -124,19 +128,24 @@ export default function useBreakdownChartData() {
...
-  key: string,
+  key: ElectricityStorageType,
   value: ZoneDetail,
   valueFactor: number,
   displayByEmissions: boolean
 ) {
-  const storageKey = key.replace(' storage', '') as ElectricityStorageType;
-  let scaledValue = (-1 * Math.min(0, (value.storage || {})[storageKey])) / valueFactor;
+  const storageKey = key.replace(' storage', '') as ElectricityStorageKeyType;
+  const storageValue = (value.storage || {})[storageKey];
```
VIKTORVAV99, line 137:
```suggestion
  const storageValue = value.storage?.[storageKey];
```
One more.

VIKTORVAV99 (commented):
Have a few comments and suggestions but overall I think it looks good!

web/src/features/charts/bar-breakdown/BarBreakdownEmissionsChart.tsx
madsnedergaard: Yes, but I have just made a new commit that renames it from "mode" to "zoneKey" for exchanges - it was quite confusing that it was called "mode" but always were a zoneKey :)

web/src/features/charts/bar-breakdown/BarElectricityBreakdownChart.tsx
madsnedergaard: They are very close, but with slight differences:
Screenshot 2023-02-24 at 20 21 13

We could extract the common part and extend from that, but I don't think it's worth abstracting it for just these two components :)

web/src/features/charts/bar-breakdown/utils.ts
madsnedergaard: I was considering it, but then we wouldn't catch `NaN` 🤔

VIKTORVAV99: Damn number that is not really a number 😄
Let's keep it as is then until we figure out how to deal with all the undefined, none and NaN.

madsnedergaard: Haha yeah, it is very confusing 😅

web/src/features/charts/bar-breakdown/utils.ts
madsnedergaard: Weirdly, I had to add a fallback if the value is undefined - which makes sense, but don't know why TS wasn't complaining before :D

web/src/features/charts/types.ts
```diff
@@ -1,12 +1,14 @@
...
+  layerData: { [layerKey: LayerKey]: number };
 }
 
 export interface InnerAreaGraphTooltipProps {
   zoneDetail?: ZoneDetail;
-  selectedLayerKey?: string;
+  selectedLayerKey?: LayerKey;
 }
+
+export type LayerKey = ZoneKey | ElectricityModeType;
```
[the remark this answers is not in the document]
madsnedergaard: Makes sense! :)

VIKTORVAV99 (approved):
LGTM! 🎉

Nice work getting rid of those confusing mode keys!

PR #5133, opened 2023-02-27
Md/ele 1807 is add validation to discard fake 0null

mathilde-daugy:
## Description

- added validation to ensure that empty datapoints are not loaded to the DB. For example, in IS the parser collects production data points with only 0 (see image). These are "fake" zeros and should be discarded.
image

### Preview
image
The datapoint will be discarded if all values are None or 0 

### Double check

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -201,4 +201,7 @@ def validate_production(obj: Dict[str, Any], zone_key: ZoneKey) -> None:
                 % (key, zone_key)
             )
 
+    if all((obj["production"][key]==0) or (obj["production"][key] is None) for key in obj.get("production", {})):
+        raise ValidationError(f"{zone_key}, {obj.get('datetime')}: unrealistic datapoint, all production values are 0.0 MW or null")
+
     validate_reasonable_time(obj, zone_key)
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -201,7 +201,13 @@ def validate_production(obj: Dict[str, Any], zone_key: ZoneKey) -> None:
                 % (key, zone_key)
             )
 
-    if all((obj["production"][key]==0) or (obj["production"][key] is None) for key in obj.get("production", {})):
-        raise ValidationError(f"{zone_key}, {obj.get('datetime')}: unrealistic datapoint, all production values are 0.0 MW or null")
+    if all(
+        (obj.get("production").get(key) == 0)
+        or (obj.get("production").get(key) is None)
+        for key in obj.get("production", {})
+    ):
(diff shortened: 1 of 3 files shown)


parsers/lib/quality.py
```diff
@@ -201,4 +201,13 @@ def validate_production(obj: Dict[str, Any], zone_key: ZoneKey) -> None:
                 % (key, zone_key)
             )
 
+    if all(
+        (obj.get("production").get(key) == 0)
+        or (obj.get("production").get(key) is None)
+        for key in obj.get("production", {})
```
unitrium, line 207:
```suggestion
       (value == 0)
        or (value is None)
        for value in obj.get("production", {}).values()
```
Makes it a bit easier to read.

unitrium: Maybe do not commit it like this.

unitrium: As discussed, please just add the check for 'storage only zone'

parsers/lib/quality.py
```diff
@@ -200,5 +200,14 @@ def validate_production(obj: Dict[str, Any], zone_key: ZoneKey) -> None:
...
             )
+    if "production" in obj.keys():
+        if all(
+            (val== 0)
+            or (val is None)
+            for val in obj.get("production",{}).values()
+        ):
+            raise ValidationError(
+                f"{zone_key}, {obj.get('datetime')}: unrealistic datapoint, all production values are 0.0 MW or null"
+            )
```
VIKTORVAV99, line 211: I don't think we can do it like this, we have a few zones where it's expected to be all 0 at times when they only rely on imports, like Åland (AX).

mathilde-daugy: hmmmm annoying. it does make AX fail ...

VIKTORVAV99: There should be a config in the backend config that has the list of allowed 0 production zones. Maybe we could just exclude those (and move the config here)?

mathilde-daugy: @VIKTORVAV99, do you have any idea of other zones that would also be like this?

mathilde-daugy: ok I decided to go about this the other way around and to add validation at the parser level

VIKTORVAV99: > @VIKTORVAV99, do you have any idea of other zones that would also be like this?

There are a few other zones, mostly islands that have a interconnection. But I think the parser level is the way to go for this particular zone.

VIKTORVAV99 (approved):
I see no issues in here but unfortunately like I mentioned in Linear it's not possible to test the IS parser locally. But things *should* work.

PR #5143, opened 2023-03-01
Fixing ZA parser

### LINKED ISSUE #5137: ZA production parser down
## Description

This is an automatic error report generated for South Africa (ZA).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser ZA production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/ZA.html">Explore the runtime logs</a>

E1Cattivo:
…ng to missing upload, and negative nuclear production according to shutdown of power plant)

## Issue

Closes #5137

## Description
 
added detection, if no data for month available jet, to use last month instead

added correction for production values < 0 according to faulty data in this case for example the nuclear power plant was shut down and production should be 0 but was about -35

### Double check


E1Cattivo:
As this is my first edit, I hope everything is correct

VIKTORVAV99:
Hi @E1Cattivo and thanks for opening the PR!
Is this ready for review? If so please mark it as such.

E1Cattivo:
I didn't get the poetry thing running for now, so I can not finish the Tests at the moment.
Should I mark it anyway?

VIKTORVAV99:
> I didn't get the poetry thing running for now, so I can not finish the Tests at the moment. Should I mark it anyway?

If you can it would be best to install poetry and test this locally (even if the changes aren't huge).
I personally found it easiest to install it via pipx. Here is their official guide on how to do it (wit just pipx): https://python-poetry.org/docs/#installing-with-pipx

E1Cattivo:
Finaly did it and updated the code after reformatting.

Issue disappeared on its own for now but this will happen again every month and shutdown of the nuclear power plant.

parsers/ZA.py
```diff
@@ -11,9 +11,11 @@
 TIMEZONE = "Africa/Johannesburg"
 
 
-def get_url() -> str:
+def get_url(
+    date_shift: int = 0
+) -> str:
     """Returns the formatted URL"""
-    date = datetime.utcnow()
+    date = datetime.utcnow() - timedelta(weeks = date_shift * 4)
```
VIKTORVAV99, line 18: 4 weeks seems like a big step for the retry, would it not make more sense to use something like 1 week or what are the publishing deadline for the underlying data/source?

VIKTORVAV99:
> Finaly did it and updated the code after reformatting.
> 
> Issue disappeared on its own for now but this will happen again every month and shutdown of the nuclear power plant.

Great, we should be able to get this merged soon!

parsers/ZA.py
E1Cattivo: Yes this should work as well, but does not change much, as the link is merged of year an month. The exact date does not matter.

This will also just be needed when in the current month there have not been any data uploaded by the source.

VIKTORVAV99: Let's leave it as it is then.

parsers/ZA.py
```diff
@@ -49,7 +53,7 @@ def fetch_production(
             row_data = row.split(";")[1:]
             if row_data[0] != "":
                 production_csv_data[date] = [
-                    float(item.replace(",", ".")) for item in row.split(";")[1:]
+                    float(item.replace(",", ".")) if float(item.replace(",", ".")) >= 0 else 0 for item in row.split(";")[1:]
```
VIKTORVAV99, line 56: Wouldn't this prevent us from getting the hydro storage values? They have both positive and negative values and both are valid.

It would be better to implement the flooring down on line 124, in this part of the code:
```python
            elif j in production_idcs:
                data["production"][column_mapping[j]] = round(
                    data["production"][column_mapping[j]] + float(prod_data_value),
                    13,
                )
```

VIKTORVAV99 (changes requested):
We need to move the rounding function for this to work properly with storage values.

VIKTORVAV99:
@E1Cattivo any update on this PR?

PR #5144, opened 2023-03-01
Replaces the fallback flag with a square

madsnedergaard:
## Description

Turns out the fallback flag I initially added for fun was actually shown on a zone after all... Replacing it with something more appropriate and also removes an edge case for handling AUS zones, which is not needed now that we have renamed the zones! 🥳 

### Preview

Screenshot 2023-03-01 at 14 21 10

### Double check


tonypls (approved):
LGTM but not to the pirates 🏴‍☠️

PR #5150, opened 2023-03-03
Adds typechecking workflow to CI

madsnedergaard:
## Issue

Closes https://github.com/electricitymaps/electricitymaps-contrib/issues/5122

## Description

This adds a TypeScript typechecking step to CI.
Also fixes the last error :)

### Preview

It fails when it should:
Screenshot 2023-03-03 at 08 02 46

And succeeds when it should: _(see checks below)_


.github/workflows/CI.yml
```diff
@@ -112,6 +112,14 @@ jobs:
       needs.setup.outputs.workflows != 'false'
     uses: ./.github/workflows/eslint.yml
 
+  typecheck:
+    name: Typecheck
+    needs: [setup, prettier]
+    if: >-
+      needs.setup.outputs.frontend != 'false' ||
```
VIKTORVAV99, line 119: Since this just checks types we could refine it further and only run it on *.ts *.tsx changes but I'm not if it's worth it as we might have to run it on every frontend change if we add a build step later as this would most likely be a requirement.

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,33 @@
...
+        id: cache-web
+        uses: actions/cache@v3
+        with:
+          path: ./web/node_modules
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('./web/pnpm-lock.yaml') }}
+      - name: Install web dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: pnpm install --frozen-lockfile
+      - name: Run TypeScript typechecking
+        run: pnpm run run-tsc
```
VIKTORVAV99, line 33: hehe we should probably change the names of these commands, run run-tsc looks weird 😅

But we can do that in a separate PR.

VIKTORVAV99 (approved):
LGTM! 🎉

I have added some comments about future improvements but these does need to be addressed it this PR.

.github/workflows/typecheck.yml
madsnedergaard: Haha yeah, it is a bit weird :D

PR #5155, opened 2023-03-05
Zone outage message component

corradio:
**Note: This was done as a part of a hackday (our internal monthly hackathons) and the code here is a proof of concept rather than final implementation. Next steps and prioritisation of this is still to be decided!**

## Description

The goal of this PR is to make it faster for us to report data errors and potential outages.
This PR still needs some work (testing, design, styling implementation), so hopefully someone can take over.

The way it works is that it checks a [publicly readable spreadsheet](https://docs.google.com/spreadsheets/d/1C7zULJugzjFj3zw98WZYzBhEIHNYzTtTLFdTcVdzqzc/edit#gid=0) (only editable by the Electricity Maps team) and displays a message accordingly.

cc @mathilde-daugy @unitrium @madsnedergaard @tonypls 

### Preview

image


web/src/features/panels/zone/ZoneHeader.tsx
```diff
@@ -49,13 +51,56 @@ export function ZoneHeader({
...
+        'message': matching.c[1].v,
+        'issueId': matching.c[2].v,
+      });
+    } else {
+      setOutageData(null);
+    }
+  };
+  useEffect(() => {
+    fetchData();
+  }, [zoneId]);
```
VIKTORVAV99, line 76: Would it not be better to do this in the app-backend and return it as a part of the zone data?

Right now we are introducing an additional dependency on google docs being accessible to every client device and we are making extra network calls. By moving it we can replace thousands of network calls on client device (on a aggregated level) with just one call in the app-backend and there is no requirement for the client device to be able to access google docs.

corradio: Absolutely! This was just a proof of concept. You are absolutely right.

madsnedergaard (commented):
Overall works and looks nice! I have suggested some small design tweaks below :)

tonypls:
> Overall works and looks nice! I have suggested some small design tweaks below - I also have the changes locally and can push them all if you want, but I decided to give you the option of picking between them :) I also added it to the DK-DK2 mock data to make testing locally easier.

Changes look great, want to push em then I'll check them out locally?

madsnedergaard:
> > Overall works and looks nice! I have suggested some small design tweaks below - I also have the changes locally and can push them all if you want, but I decided to give you the option of picking between them :) I also added it to the DK-DK2 mock data to make testing locally easier.
> 
> Changes look great, want to push em then I'll check them out locally?

Done! I also removed the break-all, but feel free to change that if it really needs to be there

tonypls:
> > > Overall works and looks nice! I have suggested some small design tweaks below - I also have the changes locally and can push them all if you want, but I decided to give you the option of picking between them :) I also added it to the DK-DK2 mock data to make testing locally easier.
> > 
> > 
> > Changes look great, want to push em then I'll check them out locally?
> 
> Done! I also removed the break-all, but feel free to change that if it really needs to be there
Here's the case where we want break-all:
image

I'm not sure if it's necessary but it is an edge case I had tested.


I've also updated the overflow, the sizing of the icons and added a div so icons get squished when there is overflow.

image

madsnedergaard (approved):
Awesome, let's go! 🥳

luuu-xu:
Hi, while I was working on issue 5309, I noticed that our ZoneDetail type does not have the required zoneMessage property. Should we fix that?

tonypls:
> Hi, while I was working on issue 5309, I noticed that our ZoneDetail type does not have the required zoneMessage property. Should we fix that?

That would be great 👍

luuu-xu:
How should we approach this small fix? Opening an issue and PR to fix it?

VIKTORVAV99:
> How should we approach this small fix? Opening an issue and PR to fix it?

Just opening a PR is enough. No need to create a issue if you are fixing it at the same time, at least that's my personal opinion.

luuu-xu:
Oh, I am so sorry. I fetched master branch and realized ZoneDetails does have zoneMessage property. I think when I was working my issue branch I did not diverge from the most recent master commit so TypeScript had an error.

VIKTORVAV99:
> Oh, I am so sorry. I fetched master branch and realized ZoneDetails does have zoneMessage property. I think when I was working my issue branch I did not diverge from the most recent master commit so TypeScript had an error.

Don't worry about it, it happens to us all.

I have done such mistakes myself a more than once occasion. 😅

PR #5162, opened 2023-03-06
create config for enabling aggregated view

wobniarin:
## Issue
Japan is not visible under the aggregated view yet. 

## Description
This PR creates the config file for Japan `JP.yaml` so that the aggregated view can be generated. 
At the moment, since the estimation model is not implemented yet, there is no flowtraced data and therefore the mock data is no available at the moment. 

### Preview

Screenshot 2023-03-06 at 15 18 58

### DIFF
--- a/config/zones/JP.yaml
+++ b/config/zones/JP.yaml
@@ -1,0 +1,37 @@
+bounding_box:
+  - - 138.5
+    - 34.52
+  - - 141.16
+    - 37.17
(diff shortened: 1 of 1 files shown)


unitrium (approved):
🇯🇵 Looking forward to it!

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5163, opened 2023-03-06
Tonyvanswet/pre commit

tonypls:
## Issue
Adds pre-commit prettifying and tests.

## Description
We had shifted to running pre push but this is too late if prettier or eslint needs to add and changes. Also now run vitest on any changed files.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -196,7 +196,7 @@
     }
   },
   "lint-staged": {
-    "*": "pretty-quick --staged",
+    "*": "prettier -w --cache",
     "*.{ts,tsx}": [
       "eslint --cache --fix",
       "pnpm exec vitest  --changed --run --coverage=false"
--- a/web/package.json
+++ b/web/package.json
@@ -190,10 +190,7 @@
     "not op_mini all"
   ],
   "lint-staged": {
-    "*": [
-      "echo `console.log(   'hello')` >> geo/validate.ts && prettier -uw --cache",
-      "echo `console.log(   'hello')` >> geo/validate.ts && pretty-quick --staged"
-    ],
+    "*": "prettier -uw --cache",
     "*.{ts,tsx}": [
       "eslint --cache --fix",
       "vitest related --run --coverage=false"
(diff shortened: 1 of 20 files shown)


web/package.json
```diff
@@ -170,6 +170,7 @@
     "postcss": "8.4.18",
     "prettier": "2.8.4",
     "prettier-plugin-tailwindcss": "0.1.13",
+    "pretty-quick": "^3.1.3",
```
VIKTORVAV99, line 173: Why do we need this? Can't we run prettier directly?
I'm pretty sure it only format changed files anyway.

(We did run prettier as a pre commit before I think?)

tonypls: I think pre rewrite we did but it hasn't worked in a while.

Prettier suggested using this [here](https://prettier.io/docs/en/precommit.html) and it's just a dev dep

VIKTORVAV99: Don't we already do option 1 described there?

tonypls: I assumed a library would have some optimisations, I timed the execution twice here are the results:
pretty-quick --staged : 
0.82s user 0.33s system 103% cpu 1.111 total
0.80s user 0.30s system 99% cpu 1.108 total

prettier -w . :
6.20s user 0.60s system 159% cpu 4.263 total
6.21s user 0.58s system 159% cpu 4.247 total

Not sure about the backwards math of the bash time function 🤔  but pretty-quick was pretty quick 🤣

VIKTORVAV99: did you try prettier with the `--cache` option as well?

tonypls: 1.29s user 0.35s system 115% cpu 1.423 total
1.30s user 0.33s system 116% cpu 1.394 total

:D pretty good

VIKTORVAV99: Thanks for testing it!
It does looks like pretty-quick still is faster so I'll leave it up to you but the less dependencies the better in my opinion. 😅

madsnedergaard (approved):
LGTM once merge conflict on version number is solved :)

PR #5165, opened 2023-03-06
Fixes typechecking CI step!

madsnedergaard:
## Description

It currently fails with type errors because the config files are not generated. This fixes it!

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -30,7 +30,6 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: pnpm install --frozen-lockfile
       - name: Generate config files
-        if: steps.cache.outputs.cache-hit == 'true'
         run: pnpm create-generated-files
       - name: Run TypeScript typechecking
         run: pnpm run run-tsc
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -29,5 +29,8 @@ jobs:
       - name: Install web dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: pnpm install --frozen-lockfile
+      - name: Generate config files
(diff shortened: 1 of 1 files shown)


.github/workflows/typecheck.yml
```diff
@@ -29,5 +29,8 @@ jobs:
       - name: Install web dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: pnpm install --frozen-lockfile
+      - name: Generate config files
+        if: steps.cache.outputs.cache-hit == 'true'
```
tonypls, line 33:
```suggestion
        id: generated-files
        if: steps.generated-files.outputs.cache-hit != 'true'
```

VIKTORVAV99, line 33: I think we can remove the conditional here.
(Which should also make this work 😅)
```suggestion
```

tonypls: Not 100% on this but looking at the docs [here](https://github.com/actions/cache#using-a-combination-of-restore-and-save-actions) it seemed reasonable

tonypls: If this works we should update the other workflows that generate these files to share the cache id

VIKTORVAV99: The cache will be shared by the workflows but the ID will not be, so it needs a unique identifier (like hashing the lockfiles) otherwise we can't guarantee it's the latest cache as they don't update if there is a cache hit. There are some ways around this though such as adding the run id which will always be different then using fallback keys for the current branch and master.

tonypls: I've removed it to see what happens 🫣

tonypls (approved):
Removed the caching thing then approved, up to you @madsnedergaard if you think we need some caching here

madsnedergaard:
Ah, thanks for updating it - I just copy-pasted from another workflow and assumed it would work :D

Issue #5141, opened 2023-02-28
Charts don't re-size when the window size changes


VIKTORVAV99:
## Description
If you start with the app open in let's say split screen and then later expand the window to full screen you get a state where the charts are narrower than they should be.

For example:
![image](https://user-images.githubusercontent.com/30777521/221908885-f0ec36b5-57fb-4fe4-a45a-f883b0bd0cab.png)

Then if you refresh the page the charts render correctly:
![image](https://user-images.githubusercontent.com/30777521/221909068-c26ec1ff-a2f5-4905-b66a-62e6e6ff340a.png)

madsnedergaard:
Maybe this could be fixed by replacing our custom ref resize observer with a third-party tool like https://github.com/ZeeCoder/use-resize-observer :)

15Arindam:
Hey, I would like to try this, I'm new in OS contrib., could you please guide me how to setup locally & reproduce this issue

VIKTORVAV99:
> Hey, I would like to try this, I'm new in OS contrib., could you please guide me how to setup locally & reproduce this issue

Hey @15Arindam!

Local setup should be documented in the [CONTRIBUTING.md](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/CONTRIBUTING.md) file. There isn't really much more too it than that for the frontend 🙂.

As for reproducing the issue that's pretty easy so all you need to do is the same as I did in the description of this issue.

If you have any questions feel free to reach out any time!

15Arindam:
Thanks @VIKTORVAV99 
I was able to locally setup, did some changes (need more help on it) but unable to push them.

VIKTORVAV99:
> Thanks @VIKTORVAV99 I was able to locally setup, did some changes (need more help on it) but unable to push them.

Hi again @15Arindam 
If you are trying to push directly to this repo that is expected and by design.
If you haven't already you need to fork the repository and then clone that instead of this one.
Once that is done you can push your changes to your fork and then open a pull request, PR, to this repository. That will also make sure all our automated tests run and give me and other members of the team time to review it before it's merged. (It's actually required.)

You can follow the detailed steps here: https://docs.github.com/en/get-started/quickstart/contributing-to-projects

Just make sure you make a copy of your local changes first.

15Arindam:
Hey @VIKTORVAV99 ,
[...]

PR #5166, opened 2023-03-06
use-resze-observer impl.

15Arindam:
## Issue

Charts don't re-size when the window size changes #5141

## Description

Implemented [use-resize-observer](https://github.com/ZeeCoder/use-resize-observer) hook for this scenario

### Double check

closes: #5184

### DIFF
--- a/web/cypress/e2e/countrypanel.spec.ts
+++ b/web/cypress/e2e/countrypanel.spec.ts
@@ -37,8 +37,11 @@ describe('Country Panel', () => {
       '24'
     );
     // ensure tooltip is shown and changes depending on where on the graph is being hovered
-    cy.get('[data-test-id=details-carbon-graph]').trigger('mousemove', 'left');
-    cy.get('[data-test-id=carbon-chart-tooltip]').should('be.visible');
+    cy.get('[data-test-id=details-carbon-graph]')
+      .trigger('mousemove', 'left')
+      .then(() => {
+        cy.get('[data-test-id=carbon-chart-tooltip]').should('be.visible');
+      });
     cy.get('[data-test-id=carbon-chart-tooltip] ').should('contain.text', '234');
 
     cy.get('[data-test-id=details-carbon-graph]').trigger('mouseout');
@@ -67,7 +70,7 @@ describe('Country Panel', () => {
       .contains('Data is temporarily unavailable for the selected time');
   });
 
-  it('asserts countryPanel contains no parser message when zone has no data', () => {
+  it.skip('asserts countryPanel contains no parser message when zone has no data', () => {
     cy.visit('/zone/CN?lang=en-GB');
     cy.waitForAPISuccess('v6/state/hourly');
     cy.get('[data-test-id=no-parser-message]').should('exist');
--- a/web/cypress/e2e/countrypanel.spec.ts
+++ b/web/cypress/e2e/countrypanel.spec.ts
@@ -7,9 +7,11 @@ describe('Country Panel', () => {
   it('interacts with details', () => {
     cy.interceptAPI('v6/details/hourly/DK-DK2');
 
-    cy.visit('/zone/DK-DK2?skip-onboarding=true&lang=en-GB');
+    cy.visit('/zone/DK-DK2?lang=en-GB');
+    cy.get('[data-test-id=close-modal]').click();
     cy.waitForAPISuccess('v6/state/hourly');
     cy.waitForAPISuccess('v6/details/hourly/DK-DK2');
+    cy.get('[data-test-id=loading-overlay]').should('not.exist');
     cy.contains('East Denmark');
     cy.contains('Carbon Intensity');
     cy.get('[data-test-id=left-panel] [data-test-id=co2-square-value]').contains('232');
--- a/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
+++ b/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
@@ -7,13 +7,13 @@ import { useTranslation } from 'translation/translation';
 import { ElectricityModeType, ZoneDetail, ZoneKey } from 'types';
 import { displayByEmissionsAtom } from 'utils/state/atoms';
 import { useBreakpoint } from 'utils/styling';
-import { useReferenceWidthHeightObserver } from 'utils/viewport';
 import useBarBreakdownChartData from '../hooks/useBarElectricityBreakdownChartData';
 import BreakdownChartTooltip from '../tooltips/BreakdownChartTooltip';
 import BarBreakdownEmissionsChart from './BarBreakdownEmissionsChart';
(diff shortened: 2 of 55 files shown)


VIKTORVAV99 (commented):
I have been testing this locally and while the resizing seem to work now it seems to have broken this code somehow:
https://github.com/electricitymaps/electricitymaps-contrib/blob/d7250e9af451eda922ad4c7aaadb058aff469840/web/src/features/charts/graphUtils.ts#L20-L22

I'm not really sure why it broke and have not had the time to dig into it further but until that is fixed we can't merge this as it breaks hovering over the (some) of the graphs.

VIKTORVAV99:
@15Arindam please do not use force push, it makes it impossible to review changes since the last commit.

VIKTORVAV99 (commented):
As far as I can tell everything is working as expected now, nice work!

In order for this to pass the CI though you need to run `pnpx prettier --write .` in the `/web` folder.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -27,10 +27,10 @@ function BarBreakdownChart() {
     height,
   } = useBarBreakdownChartData();
   const [displayByEmissions] = useAtom(displayByEmissionsAtom);
-  const { ref, width } = useReferenceWidthHeightObserver(X_PADDING);
+  const { ref, width: observerWidth = 0 } = useResizeObserver();
```
VIKTORVAV99, line 30: Why do we need an initial value here?

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -105,12 +105,12 @@ function AreaGraph({
...
   tooltipSize,
 }: AreagraphProps) {
-  const {
-    ref,
-    width: containerWidth,
-    height: containerHeight,
-    node,
-  } = useReferenceWidthHeightObserver(Y_AXIS_WIDTH, X_AXIS_HEIGHT);
+  const reference = useRef(null);
+  const { width: observerWidth = 0, height: observerHeight = 0 } =
```
VIKTORVAV99, line 109: and here?

These will become negative initially and well negative width and height don't really work.

web/src/features/exchanges/ExchangeLayer.tsx
```diff
@@ -3,12 +3,12 @@ import { useExchangeArrowsData } from 'hooks/arrows';
...
 import React from 'react';
 import { MapboxMap } from 'react-map-gl';
-import { useReferenceWidthHeightObserver } from 'utils/viewport';
 import ExchangeArrow from './ExchangeArrow';
+import useResizeObserver from 'use-resize-observer';
 
 function ExchangeLayer({ map }: { map?: MapboxMap }) {
   const [isMapMoving] = useAtom(mapMovingAtom);
-  const { ref, width, height } = useReferenceWidthHeightObserver();
+  const { ref, height = 0, width = 0 } = useResizeObserver();
```
VIKTORVAV99, line 11: Here we also set them to 0 initially. Why?

web/src/features/time/TimeAxis.tsx
```diff
@@ -84,8 +84,8 @@ function TimeAxis({
   className,
 }: TimeAxisProps) {
   const { i18n } = useTranslation();
-  const { ref, width } = useReferenceWidthHeightObserver(24);
-
+  const { ref, width: observerWidth = 0 } = useResizeObserver();
+  const width = observerWidth - 24;
```
VIKTORVAV99, line 88: This will also be negative initially.

web/src/features/weather-layers/solar/SolarLayer.tsx
```diff
@@ -32,13 +32,14 @@ export default function SolarLayer({ map }: { map?: MapboxMap }) {
   });
   const solarData = solarDataArray?.[0];
 
-  const { ref, node, width, height } = useReferenceWidthHeightObserver();
+  const reference = useRef<HTMLCanvasElement>(null);
+  const { width, height } = useResizeObserver({ ref: reference });
```
VIKTORVAV99, line 36: Here we don't set an initial value, this seems correct for the usecase.

web/src/features/weather-layers/wind-layer/WindLayer.tsx
```diff
@@ -29,7 +29,8 @@ const createWindy = async (canvas: HTMLCanvasElement, data: any, map: MapboxMap)
 export default function WindLayer({ map }: { map?: MapboxMap }) {
   const [isMapMoving] = useAtom(mapMovingAtom);
   const [windy, setWindy] = useState<Maybe<WindyType>>(null);
-  const { ref, node, width, height } = useReferenceWidthHeightObserver();
+  const reference = useRef<HTMLCanvasElement>(null);
+  const { height = 0, width = 0 } = useResizeObserver<HTMLCanvasElement>({ ref: reference });
```
VIKTORVAV99, line 33: and here we set it once again? Why the inconsistency and why is the initial values needed?

VIKTORVAV99 (changes requested):
There seems to be some inconsistencies with the initial values that need to be corrected and/or explained, this PR is currently not consistent all the way and these seems to be no real need to set an initial value in the first place.

VIKTORVAV99:
@15Arindam I noticed you created a new PR but I would prefer we keep using this one to preserve he discussions and comments in the original PR.

15Arindam:
Hi @VIKTORVAV99 , The type value of width, height returned from `useResizeObserver` hook is `undefined | number` , methods doesnt' throw typescript error , there I kept as is otherplace I defaulted with zero. But as you pointed out I missed the negative value cases.

VIKTORVAV99 (dismissed):
LGTM! 🎉

Thanks for implementing the change!
I'll merge this later today after #5183 is merged to see that the tests are still passing but they should be as I saw no changes that should have affected them.

VIKTORVAV99:
@tonypls do you know why cypress is failing in here? As far as I can tell there is no direct changes that should affect the tests.

tonypls:
> @tonypls do you know why cypress is failing in here? As far as I can tell there is no direct changes that should affect the tests.

It looks like the resize observer doesn't see the size of the container of the AreaGraph in cypress. The graph just keeps growing so Cypress in unable to hover

web/cypress/e2e/countrypanel.spec.ts
VIKTORVAV99, line 1: @tonypls would you mind reviwing the changes to the tests here?
I'm not super familiar with how cypress works 🙂

tonypls: Damn, looks like there's some sort of loop in the width observer.

https://user-images.githubusercontent.com/18622768/226459378-c5337922-130e-4168-b17c-e85a2937b434.mov

madsnedergaard:
Closing this for now, it was a nice contribution but we need Cypress tests to still work - if anyone wants to pick this up again feel free to do so :)

PR #5168, opened 2023-03-07
Tonyvanswet/component testing ci

tonypls:
## Issue
Adds cypress component test job to CI


.github/workflows/cypress.yml
```diff
@@ -21,3 +21,13 @@ jobs:
...
+      - name: Cypress run component tests
+        uses: cypress-io/github-action@v5
+        env:
+          TZ: Europe/Copenhagen
+        with:
+          working-directory: ./web
+          # we have already installed everything
+          install: false
+          # to run component tests we need to use "component: true"
+          component: true
```
VIKTORVAV99, line 33: Right now this runs as a step in the same job as the e2e tests. I think it would be preferable to split this into it's own job so it will still run if the e2e tests fail.

tonypls: I think it's quite nice to keep the building for cypress testing all in one file / run. Also I'm 90% sure even if the e2e don't pass that component will run. 

We already have 17ish actions which is already quite messy, I'd prefer not to add more especially when these two share concerns.

tonypls:
Closed in favour of #5183

PR #5169, opened 2023-03-07
Include target_time iteration and fix pytz issues 

wobniarin:
**Disclaimer**: This is a **WIP** 

## Issue

When checking the `IN.py` parser the following items would happen:
- `pytz` library would convert from IST to UTC by sustracting 5h 27' instead of 5h 30' , causing an inconsistency in the format when storing the data. 
- The datetime where Renewable Energy Sources data is available 17-12-2020 was hardcoded and since this will be used for other zones, this could be included as a global variable. 
- Due to delays in the data, sometimes production data is not available in real-time but with 2 days delayed. However, some other times we can get data corresponding to yesterday. 

## Description
Some small fixes have been implemented in the IN.py parser with the following objectives: 
- Include global variable for when renewable energy data is available `START_DATE_RENEWABLE_DATA = datetime(2020, 12, 17)`
- Include lookback period to check data from today and up to two days ago to get as fresh data as possible. 
- Change pytz localization to avoid weird time changes when converting it to UTC. 

### Double check

### TODO (pending)
For now production data has only data available for one day. We currently floor the target_time to 00:00 of the current date (in IST local time)
However, when we get the parser data, this data is converted into UTC, being converted to data from the previous day. Our question is: Since we only have one data point for each day, shall we send the request at a specific time (00:00 UTC) so that the production data range covers the entire day? 

Example of the discussion: 
- We ask for data corresponding to today 7-03-2023.
- This data is not available, but there is data corresponding to the entire day of 06-03-2023 in IST
- Since we only get one datapoint for each day, this datapoint is floor to 06-03-2023 00:00:00 IST
- The test parser converts it to UTC, showing and leading to a potential misunderstanding that the data is from the 05-03-2023 at 18:30:00 UTC. 
image


pierresegonne:
Well done!

pierresegonne (approved):

PR #5170, opened 2023-03-07
cypress test all script

tonypls:
## Issue
As part of https://linear.app/electricitymaps/issue/ELE-1832/automate-deployment-to-staging

## Description
Adds a script to package.json to run both cypress test at once


tonypls:
Please review me <3

web/package.json
```diff
@@ -22,6 +22,7 @@
     "prebuild": "pnpm run create-generated-files",
     "cy:component": "cypress open --component -b electron",
     "cy:e2e": "cypress open --e2e -b electron",
+    "cy:test-all": "cypress run && cypress run --component",
```
VIKTORVAV99, line 25: should this also use the electron browser as the cy:component script or does that not matter?

VIKTORVAV99 (approved):
Looking good but I have one question.

web/package.json
tonypls: Electron is the default :)

PR #5171, opened 2023-03-07
publish from earthly to staging

tonypls:
## Issue

This PR points the earthly build outputs to be uploaded to a staging project in cloudflare pages


VIKTORVAV99:
Would it not be easier and simpler to use the Cloudflare action I suggested for the preview links to also deploy to staging if it's run on master?

The benefit of that is that we could use GitHubs native environment integration to "showcase" it on the repo "main" page in the sidebar.

tonypls:
> > > Would it not be easier and simpler to use the Cloudflare action I suggested for the preview links to also deploy to staging if it's run on master?
> > > The benefit of that is that we could use GitHubs native environment integration to "showcase" it on the repo "main" page in the sidebar.
> > 
> > 
> > This PR is to validate the build / test and deploy process through earthly that we will use for production. It's setup to initially deploy to staging.electricitymaps.com on a different cloudflare project. When we've validated this will work we can switch the project to app.electricitymaps.com to automate our production builds.
> > It's a little bit of an unconventional case that we want to deploy from a submodule in our mono-repo so I'm not sure the cloudflare action will make sense in this case.
> > We can definitely use the actions later to deploy previews and staging but for now the focus is on getting robust automated production deployments
> 
> Would just be nice to integrate it with GitHub to get things like this working: image
> 
> But maybe that is still possible?

Yeah, I think a good goal state will be to have actions trigger cloudflare pages for all PR's that change web code and have staging.electricitymaps.com to build from electricitymaps-contrib master branch

madsnedergaard (approved):

PR #5173, opened 2023-03-08
fix FO emission factor

### LINKED ISSUE #5167: Faroe islands data seem wrong ?
Hello !
Curious to what was causing the Faroe Islands to be so black (967g-CO2/kWh), I clicked on them to see the detail and I see it's a mix of oil (650g), biomass (230g), hydro (24g) and wind (11g). I guess that you're actually using the oil coefficient of 1125g (instead of 650g) as in the rest of the continent but just didn't update the label...
Thanks for your attention !

VIKTORVAV99:
## Issue

Fixes #5167 

## Description

Update the FO aggregated zone with the data emission factors from the subzones as they both match which means FO will always have the same emission factors as it's subzone for the respective mode.

### Double check


pierresegonne (approved):
Thanks for taking care of this!

PR #5174, opened 2023-03-08
Adjusts the borders of Japan regions to be more precise

madsnedergaard:
## Description

The placement of the border between `JP-TH` and `JP-TK` is slightly off. This moves the border to be more precise.

Also fixes some invalid paths in the Earthfile unrelated to this 🙈 

### Preview

**BEFORE**
Screenshot 2023-03-08 at 16 28 01

**AFTER**
Screenshot 2023-03-08 at 16 27 45

### Double check


VIKTORVAV99 (changes requested):
👀
```
Validating geometries...
ERROR: Found gap, see gaps.geojson
```

VIKTORVAV99:
> I am not able to succesfully run `pnpm run generate-world` since there are gaps in the geo.geojson file.
> 
> I attach below the gaps.json file I get as an output.
> 
> `{"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[139.8596,37.1311],[139.9502,37.1518],[140.1047,37.1181],[140.1859,37.0575],[140.231859,36.962479],[140.1853,37.0566],[140.1041,37.1177],[139.9505,37.1508],[139.8596,37.1311]]]}}]`
> 
> If I plot the gaps.geojson file I can see the gaps are just the zone between JP-TK and JP-TH
> 
> Could it be the reason why we had these overlaps before?

I think part of the issue is that qgis is way more accurate than we need something like 8 or 12 decimals so it makes it really hard to align the borders properly. There are for sure some setting or tool to simplify this but I have not found it yet 😅.

madsnedergaard:
I have zoomed far in and dragged points around to make them match, so now there's no gaps! 🥳 

Before:
Screenshot 2023-03-09 at 21 26 57

And the decimal changes it resulted in:
Screenshot 2023-03-09 at 21 25 06

VIKTORVAV99 (approved):
The map generation code is very strict about a few things and not so strict about others 😅

wobniarin (approved):
LGTM! No gaps 🥳

PR #5179, opened 2023-03-09
[Geo I] Typing functionality

madsnedergaard:
## Description

First step in improving the geometry-related code we have: Typing it up.
I have a few more PRs coming up as I'm splitting up the work I've done :)


web/scripts/generateZonesConfig.ts
```diff
@@ -124,7 +125,7 @@ const mergeRatioParameters = () => {
   return ratioParameters;
 };
 
-const writeJSON = (fileName: any, object: any) => {
+const writeJSON = (fileName: string, object: unknown) => {
```
VIKTORVAV99, line 128: We should try and define the object type here later on but I relize it's not the simplest of types 😅

VIKTORVAV99 (commented):
This was a big one with many small changes, most of them are pretty straight forward but I do have a few questions.

web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -46,7 +47,7 @@ for (const zone of zonesGeo.features) {
   let maxLon = -200;
 
   if (geometryType == 'MultiPolygon') {
-    for (const coord of allCoords) {
+    for (const coord of allCoords as Position[]) {
```
VIKTORVAV99, line 50: Should this not be a point instead of a position?

VIKTORVAV99 (commented):
Seems like github forgot one of my comments, weird

web/geo/generateAggregates.ts
```diff
@@ -29,31 +42,30 @@ const generateAggregates = (geojson, zones: ZoneConfig) => {
   const combinedZones = countryZonesToCombine
     .map((country) => {
       if (country === undefined) {
-        return null;
+        return emptyFeature;
```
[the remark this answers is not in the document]
madsnedergaard: Because if we change the FeatureCollection type to allow `features: Feature | null` we would have to add null-checks all over the code. And since we actually never want a feature to be `null`, I thought this seemed like the best option - although I also don't think it's great :)
Do you have a better idea?

web/geo/generateWorld.ts
```diff
@@ -7,8 +7,9 @@ import { generateExchangesToIgnore } from './generateExchangesToExclude.js';
 import { generateTopojson } from './generateTopojson.js';
 import { getJSON, round } from './utilities.js';
 import { validateGeometry } from './validate.js';
+import { WorldFeatureCollection } from './types.js';
```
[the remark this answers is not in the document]
madsnedergaard: I think we should add linting to align this later on, as we currently are doing both in different places. 
I was also ready up on it again, and I'd say we should avoid `import type {}` due the points raised in this comment: https://github.com/microsoft/TypeScript/issues/39861#issuecomment-668131921

madsnedergaard:
> This was a big one with many small changes, most of them are pretty straight forward but I do have a few questions.

Thanks for reviewing - and I found it really hard with TS to split it up further, as changes to one functions types affects other places. I attempted to split it up a bit by commits though, but would love to hear your thoughts on how to make reviewing easy next time :)

web/scripts/generateZonesConfig.ts
madsnedergaard: Yeeeeeeah, I started doing that, but it becomes a type of anything that can be JSON.stringify'ed, and when looking in the native NodeJS types and around on the interwebs it seems like this was most common practice. And given the actual impact of this, I did not consider it worth pursuing further - but I'm very open to changing my mind on this 😄

web/geo/generateZoneBoundingBoxes.ts
madsnedergaard: Turf defines it like this

Screenshot 2023-03-10 at 07 29 43

web/geo/generateAggregates.ts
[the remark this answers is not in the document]
VIKTORVAV99: Maybe we can filter the list first?

Somewhat related stackoverflow link: https://stackoverflow.com/questions/24806772/how-to-skip-over-an-element-in-map

web/scripts/generateZonesConfig.ts
VIKTORVAV99: Since we are only writing 4 kinds of JSON objects (at least right now) we could make 4 types and then a union type for this object.

These would be zone, world, exchanges and the aggregated exchanges thing.

We should already have types for the zone at least right?

tonypls (approved):
LGTM 🎉

tonypls:
We could make some improvements but the output is the same as it used to be and having typing will be extra nice when we want to make changes in the future!

web/geo/generateZoneBoundingBoxes.ts
VIKTORVAV99: That feature collection looks weird, it should only be `cords: Position[]` or even more specific `cords1: [number, number]` and not `Position[][]` right? I think `Position[][]` is only for bounding_boxes as they contain 2 pairs of coordinates but I'm not completely sure here.

web/geo/generateWorld.ts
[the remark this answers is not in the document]
VIKTORVAV99: Good points, let's avoid it as much as possible then. I just know there was some issue with I think it was the Angular CLI back in the day where it accidentally included imports only used as types in the bundle but I don't remember the specifics.

VIKTORVAV99 (approved):
LGTM!

I do have a suggestion for making things even faster though but we can adress that later if you wish.

web/scripts/generateZonesConfig.ts
madsnedergaard: Hmm, I started doing this, but now that I think of it I'm not sure we should limit what can be written to a JSON file - it feels counter-intuitive to be typing what output the "excluded exchanges" should be in here, when this function really shouldn't care about this 🤔 Happy to revisit this later though!

web/geo/generateZoneBoundingBoxes.ts
madsnedergaard: Here's the raw data:

Screenshot 2023-03-17 at 09 52 04

So I think this makes sense, it at least passes all typechecks ¯\\\_(ツ)\_/¯

web/scripts/generateZonesConfig.ts
VIKTORVAV99: Makes sense, and in any case the type errors should have been caught before we write things to json.

web/geo/generateZoneBoundingBoxes.ts
VIKTORVAV99: Ugh nested un-named arrays gives me headaches 😅
But it does look like this is correct then and if the typechecking works I say merge it!

PR #5183, opened 2023-03-09
Tvs/unflake tests

tonypls:
## Issue
Occasionally the e2e tests would flake on CI

## Description
This PR seeks to prevent flaky cypress tests by using the cy.click() function over checking for url search params.

Tests usually fail on CI due to the onboarding modal being on top of the component being tested, the onboarding modal should have disappeared due to the search params functionality that was in the component, my hypothesis is that the modal was briefly rendered and blocked cypress during the test window. 

I have removed the search params functionality and will rely on the cy.click() to close the modal

### DIFF
--- a/web/src/components/modals/OnboardingModal.tsx
+++ b/web/src/components/modals/OnboardingModal.tsx
@@ -62,13 +62,10 @@ const views = [
   },
 ];
 
-export function OnboardingModal({ isComponentTest }: { isComponentTest?: boolean }) {
+export function OnboardingModal() {
   const [hasOnboardingBeenSeen, setHasOnboardingBeenSeen] = useAtom(
     hasOnboardingBeenSeenAtom
   );
-  if (import.meta.env.TEST && !isComponentTest) {
-    setHasOnboardingBeenSeen(true);
-  }
   const handleDismiss = () => {
     setHasOnboardingBeenSeen(true);
   };
--- a/web/src/components/modals/OnboardingModal.tsx
+++ b/web/src/components/modals/OnboardingModal.tsx
@@ -1,4 +1,4 @@
-import { resolvePath, useSearchParams } from 'react-router-dom';
+import { resolvePath } from 'react-router-dom';
 import { hasOnboardingBeenSeenAtom } from 'utils/state/atoms';
 
 import { useAtom } from 'jotai';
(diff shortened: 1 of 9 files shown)


madsnedergaard:
Do you have an example where this problem occurs? I've only seen it being flaky because the loading screen was overlapping other content :)

tonypls:
> Do you have an example where this problem occurs? I've only seen it being flaky because the loading screen was overlapping other content :)

Good spotting, yes it was the loading overlay! 

Although it's functionally not very different I can still see using this approach to get cypress to act in the same way as a user as being useful. 

I've also added a check for the overlay to disappear, ideally I'd like the e2e tests to be super repeatable and then leave them relatively static and shift our focus to component testing.

tonypls:
Be great to get this in an see if we still have failures on CI, it's passed the last seven times 🤞

web/src/components/modals/OnboardingModal.tsx
```diff
@@ -66,18 +66,14 @@ export function OnboardingModal() {
   const [hasOnboardingBeenSeen, setHasOnboardingBeenSeen] = useAtom(
     hasOnboardingBeenSeenAtom
   );
-  const [searchParameters] = useSearchParams();
-  const skipOnboarding = searchParameters.get('skip-onboarding') === 'true';
-  const visible = !hasOnboardingBeenSeen && !skipOnboarding;
-
```
VIKTORVAV99, line 72: Does this mean ?skip-onbording won't work anymore? I know some users use this parameter in the wild.

VIKTORVAV99 (commented):
No issues with the tests as far as I can tell but one question about the change to the Onboarding Modal.

web/src/components/modals/OnboardingModal.tsx
tonypls: Yes it will remove this feature! I'm very curious why people would use this in the wild? 

I think it's good to let the components focus on their functionality and the tests focus on testing as opposed to baking test functionality into components.

VIKTORVAV99: Not sure exactly why they use it, but they do 😅

Presumably to avoid the onboarding in either saved links (favorites) and/or as a poor man's kiosk mode?

I remember people asking for it when we re-designed the app (but before we had added the functionality back) so I'd suggest keeping the functionality but removing it from the tests, at least until we have a full-fledged kiosk mode.

tonypls: Hmm okay, I'll add it back in

VIKTORVAV99 (approved):
LGTM! 🎉

web/src/components/modals/OnboardingModal.tsx
madsnedergaard: A simple kiosk mode indeed! A few companies are showing the app on big screens with no peripherals, so the only way to control this is with query-parameters. On my wish-list is making a dedicated "kiosk mode" that does this and more all at once, but until then we should definitely keep the skip-onboarding param :)

PR #5189, opened 2023-03-12
Parser cleanup focusing on hardcoded 0 returns

VIKTORVAV99:
## Issue
There are several parsers that are returning hardcoded 0 values for production.

## Description

This PR takes care of the easy ones that could be removed without changing any of the parser logic.

>**Note**
> There are still parsers that return hardcoded 0 values when they should return None.

### Double check

### DIFF
--- a/parsers/CA_PE.py
+++ b/parsers/CA_PE.py
@@ -9,6 +9,8 @@ from typing import Any, Dict, Optional
 import arrow
 from requests import Session
 
+from .lib.exceptions import ParserException
+
 timezone = "Canada/Atlantic"
 
 
@@ -65,7 +67,7 @@ def _get_pei_info(requests_obj):
 
 def fetch_production(
     zone_key: str = "CA-PE",
-    session: Optional[Session] = None,
+    session: Session = Session(),
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> Dict[str, Any]:
@@ -73,11 +75,10 @@ def fetch_production(
     if target_datetime:
         raise NotImplementedError("This parser is not yet able to parse past dates")
 
-    requests_obj = session or Session()
-    pei_info = _get_pei_info(requests_obj)
+    pei_info = _get_pei_info(session)
 
     if pei_info is None:
-        return None
+        raise ParserException("CA_PE.py, No data returned from API.", zone_key)
 
     data = {
         "datetime": pei_info["datetime"],
@@ -87,14 +88,7 @@ def fetch_production(
             # These are oil-fueled ("heavy fuel oil" and "diesel") generators
             # used as peakers and back-up
             "oil": pei_info["pei_fossil_gen"],
-            # specify some sources that definitely aren't present on PEI as zero,
-            # this allows the analyzer to better estimate CO2eq
-            "coal": 0,
-            "hydro": 0,
-            "nuclear": 0,
-            "geothermal": 0,
         },
-        "storage": {},
         "source": "princeedwardisland.ca",
     }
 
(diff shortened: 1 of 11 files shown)


parsers/CA_QC.py
```diff
@@ -49,18 +49,13 @@ def if_exists(elem: dict, etype: str):
...
-                        "coal": 0.0,
                         "hydro": if_exists(elem, "hydro"),
-                        "nuclear": 0.0,
-                        "oil": 0.0,
                         "solar": if_exists(elem, "solar"),
                         "wind": if_exists(elem, "wind"),
                         # See Github issue #3218, Québec's thermal generation is at Bécancour gas turbine.
                         # It is reported with a delay, and data source returning 0.0 can indicate either no generation or not-yet-reported generation.
                         # Thus, if value is 0.0, overwrite it to None, so that backend can know this is not entirely reliable and might be updated later.
                         "gas": if_exists(elem, "thermal") or None,
```
VIKTORVAV99, line 58: Is this still a concern or can we remove the `or None` part here?

FelixDQ:
Why are we changing this? Doesn't it make sense to explicitly say that production is 0 if we know that it's 0?
In the app it will be shown as a `?` instead of an empty bar. 
image

VIKTORVAV99:
> Why are we changing this? Doesn't it make sense to explicitly say that production is 0 if we know that it's 0? In the app it will be shown as a `?` instead of an empty bar. image

The app will show it as `?` if there is no capacity set or if the capacity is above 0. If the capacity is set to 0 and the production is None/Null it will be shown as 0 production.

It's mostly so we get rid of hardcoded values in the parsers themselves to align with our documentation.

FelixDQ:
Hmm. I think that this makes the data validation job harder as we now no longer have the distinction between:
* We know that this value is 0
* We don't know what this value is, but we assume that it is 0

VIKTORVAV99:
> Hmm. I think that this makes the data validation job harder as we now no longer have the distinction between:
> 
> * We know that this value is 0
> * We don't know what this value is, but we assume that it is 0

But wouldn't it be better for data assurance and auditing if we never manually edit these values with static figures? I'd assume it's best if the parsers consume and return as raw and unmodified data as possible?

parsers/CA_YT.py
```diff
@@ -13,7 +13,7 @@
 
 def fetch_production(
     zone_key: str = "CA-YT",
-    session: Optional[Session] = None,
+    session: Session = Session(),
```
FelixDQ, line 16: Btw this is a python quirk, but it can actually lead to some weird problems if you use mutable objects as default parameters: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

FelixDQ:
> But wouldn't it be better for data assurance and auditing if we never manually edit these values with static figures? I'd assume it's best if the parsers consume and return as raw and unmodified data as possible?

I think that depends a bit on how you view the parsers. I think that we are already making a lot of assumptions in our parsers, and I don't think that explicitly returning known information would be a difference there. 

I guess that we could make the None/0 distinction in another layer, but we don't currently have that layer, and It might also just add more complexity.

@pierresegonne do you have an opinion on this?

parsers/CA_PE.py
```diff
@@ -87,14 +88,7 @@ def fetch_production(
             # These are oil-fueled ("heavy fuel oil" and "diesel") generators
             # used as peakers and back-up
             "oil": pei_info["pei_fossil_gen"],
-            # specify some sources that definitely aren't present on PEI as zero,
-            # this allows the analyzer to better estimate CO2eq
-            "coal": 0,
-            "hydro": 0,
-            "nuclear": 0,
-            "geothermal": 0,
```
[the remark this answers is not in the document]
VIKTORVAV99: Sure thing, I'll update the PR with the capacities if we want to go forward with it. 👍🏼

VIKTORVAV99:
Closing this as it will resolve it self when we gradually upgrade the parsers.

PR #5193, opened 2023-03-13
Update toggles 

tonypls:
## Issue
https://linear.app/electricitymaps/issue/ELE-1620/update-css-styling-of-map-ui-toggles

Fixes ELE-1620

## Description
Update toggles to new design, prevent emission toggle from toggling on selected value click and align the bg colors for tooltips

### DIFF
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -39,10 +39,10 @@ export default function ToggleButton({
     }
   };
   return (
-    <div className="z-10 flex h-9 rounded-full bg-zinc-100  px-[5px] py-1  drop-shadow-lg dark:bg-gray-900">
+    <div className="z-10 flex h-9 rounded-full bg-zinc-100  px-[5px] py-1   dark:bg-gray-900">
       <ToggleGroupRoot
         className={
-          'flex-start flex h-[26px] flex-grow flex-row items-center justify-between self-center rounded-full bg-gray-100 shadow-inner dark:bg-gray-700'
+          'flex-start flex h-[26px] flex-grow flex-row items-center justify-between self-center rounded-full bg-gray-100 dark:bg-gray-700'
         }
         type="single"
         aria-label="Toggle between modes"
@@ -53,14 +53,15 @@ export default function ToggleButton({
             key={`group-item-${key}`}
             value={option.value}
             onClick={() => onToggle(option.value)}
-            className={`
-       inline-flex h-[26px] w-full  items-center whitespace-nowrap rounded-full px-4 ${fontSize} ${
+            className={twMerge(
+              'inline-flex h-[26px] w-full items-center whitespace-nowrap rounded-full px-4',
+              fontSize,
               option.value === selectedOption
-                ? ' bg-white  shadow transition duration-500 ease-in-out dark:bg-gray-500'
+                ? ' bg-white text-brand-green  shadow transition duration-500 ease-in-out dark:bg-gray-500'
                 : ''
-            }`}
+            )}
           >
-            <p className="sans flex-grow select-none  dark:text-white">
+            <p className="sans flex-grow select-none font-inter font-bold capitalize  dark:text-white">
               {__(option.translationKey)}
             </p>
           </ToggleGroupItem>
@@ -76,16 +77,16 @@ export default function ToggleButton({
                 role="button"
                 tabIndex={0}
                 className={twMerge(
-                  'b ml-2 h-6 w-6 select-none justify-center self-center rounded-full bg-white text-center drop-shadow dark:border dark:border-gray-500 dark:bg-gray-900',
+                  'b ml-2 inline-flex h-6 w-6 select-none   justify-center self-center rounded-full bg-white shadow dark:border dark:border-gray-500 dark:bg-gray-900',
                   isToolTipOpen && 'pointer-events-none'
                 )}
               >
-                <p>i</p>
+                <p className="size  self-center font-inter text-sm font-bold ">i</p>
               </div>
             </TooltipTrigger>
             <TooltipPortal>
               <TooltipContent
-                className="relative right-[48px] z-50 max-w-[164px] rounded border bg-zinc-50 p-2  text-center text-sm drop-shadow-sm dark:border-0 dark:bg-gray-900"
+                className="relative right-[48px] z-50 max-w-[164px] rounded border bg-neutral-200 p-2  text-center text-sm  dark:border-0 dark:bg-gray-900"
                 sideOffset={10}
                 side="bottom"
                 onPointerDownOutside={onToolTipClick}
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -18,7 +18,6 @@ interface ToggleButtonProperties {
   selectedOption: string;
   onToggle: (option: string) => void;
   tooltipKey?: string;
-  fontSize?: string;
 }
 
 export default function ToggleButton({
@@ -26,7 +25,6 @@ export default function ToggleButton({
   selectedOption,
   tooltipKey,
   onToggle,
-  fontSize = 'text-sm',
 }: ToggleButtonProperties): ReactElement {
   const { __ } = useTranslation();
   const [isToolTipOpen, setIsToolTipOpen] = useState(false);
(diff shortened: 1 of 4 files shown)


web/src/utils/constants.ts
```diff
@@ -25,6 +25,11 @@ export enum SpatialAggregate {
   ZONE = 'zone',
 }
 
+export enum LeftPanelToggleOptions {
+  ELECTRICITY = 'electricity',
+  EMISSIONS = 'emissions',
+}
+
```
VIKTORVAV99, line 32: Nice!

I ❤️ Enums!

web/src/components/ToggleButton.tsx
```diff
@@ -53,14 +51,14 @@ export default function ToggleButton({
             key={`group-item-${key}`}
             value={option.value}
             onClick={() => onToggle(option.value)}
-            className={`
-       inline-flex h-[26px] w-full  items-center whitespace-nowrap rounded-full px-4 ${fontSize} ${
+            className={twMerge(
+              'inline-flex h-[26px] w-full items-center whitespace-nowrap rounded-full px-4 text-xs',
```
VIKTORVAV99, line 55:
```suggestion
            className={twMerge(
              'inline-flex h-[26px] w-full items-center whitespace-nowrap rounded-full px-4 text-sm',
```
This was sm before, should we keep it as such?
Before:
image
After:
image

web/src/components/ToggleButton.tsx
```diff
@@ -53,14 +51,14 @@ export default function ToggleButton({
...
               option.value === selectedOption
-                ? ' bg-white  shadow transition duration-500 ease-in-out dark:bg-gray-500'
+                ? ' bg-white text-brand-green  shadow transition duration-500 ease-in-out dark:bg-gray-500'
                 : ''
-            }`}
+            )}
           >
-            <p className="sans flex-grow select-none  dark:text-white">
+            <p className="sans flex-grow select-none font-inter font-bold capitalize  dark:text-white">
```
VIKTORVAV99, line 61: If we make this bold we should we make these bold as well?

image

web/src/components/ToggleButton.tsx
```diff
@@ -76,16 +74,16 @@ export default function ToggleButton({
                 role="button"
                 tabIndex={0}
                 className={twMerge(
-                  'b ml-2 h-6 w-6 select-none justify-center self-center rounded-full bg-white text-center drop-shadow dark:border dark:border-gray-500 dark:bg-gray-900',
+                  'b ml-2 inline-flex h-6 w-6 select-none   justify-center self-center rounded-full bg-white shadow dark:border dark:border-gray-500 dark:bg-gray-900',
                   isToolTipOpen && 'pointer-events-none'
                 )}
               >
-                <p>i</p>
+                <p className="size  self-center font-inter text-xs font-bold ">i</p>
```
VIKTORVAV99, line 81:
```suggestion
                <p className="size self-center font-inter text-sm font-bold">i</p>
```
Same here about the font sice

web/src/components/ToggleButton.tsx
```diff
@@ -76,16 +74,16 @@ export default function ToggleButton({
                 role="button"
                 tabIndex={0}
                 className={twMerge(
-                  'b ml-2 h-6 w-6 select-none justify-center self-center rounded-full bg-white text-center drop-shadow dark:border dark:border-gray-500 dark:bg-gray-900',
+                  'b ml-2 inline-flex h-6 w-6 select-none   justify-center self-center rounded-full bg-white shadow dark:border dark:border-gray-500 dark:bg-gray-900',
```
VIKTORVAV99, line 77:
```suggestion
                  'b ml-2 inline-flex h-6 w-6 select-none justify-center self-center rounded-full bg-white shadow dark:border dark:border-gray-500 dark:bg-gray-900',
```
I'm surprised this pass linting, is there a ESLint rule we can enable to remove multiple spaces in classes?

web/src/components/ToggleButton.tsx
```diff
@@ -76,16 +74,16 @@ export default function ToggleButton({
...
                 )}
               >
-                <p>i</p>
+                <p className="size  self-center font-inter text-xs font-bold ">i</p>
               </div>
             </TooltipTrigger>
             <TooltipPortal>
               <TooltipContent
-                className="relative right-[48px] z-50 max-w-[164px] rounded border bg-zinc-50 p-2  text-center text-sm drop-shadow-sm dark:border-0 dark:bg-gray-900"
+                className="relative right-[48px] z-50 max-w-[164px] rounded border  bg-zinc-50 p-2  text-center text-sm dark:border-0 dark:bg-gray-900"
```
VIKTORVAV99, line 86:
```suggestion
                className="relative right-[48px] z-50 max-w-[164px] rounded border bg-zinc-50 p-2 text-center text-sm dark:border-0 dark:bg-gray-900"
```
Be gone spaces!

VIKTORVAV99 (commented):
Overall this looks good but I have some considerations about the use of bold and font-xs we should discuss first.

web/src/components/ToggleButton.tsx
tonypls: The figma file was even smaller, I think xs is pretty good but it's up to @rasmusjeppsen (the netlify preview should help)

web/src/components/ToggleButton.tsx
tonypls: I think so! @rasmusjeppsen what do you think?

web/src/components/ToggleButton.tsx
tonypls: There's a few of these around 🙄

web/src/components/ToggleButton.tsx
VIKTORVAV99: Hmm, I'm not sure it's a great combo to both make these smaller and make them bold. It might work fine for higher resolution screens (but even then, the text is really small) but on lower resolution screens things will become almost merged at that size and would make things hard to read, especially at first glance.

I think just bumping the size to what it was before should take care of it.

web/src/components/ToggleButton.tsx
VIKTORVAV99: Something to add here as well: Don't we use inter for basically everything (with a few exceptions)? Would it not be easier to set the font style on the root of the html tag or the body tag and override it where needed instead?

tonypls: Yeah, it's already our default so I didn't need to specify here at all 🤦

web/src/components/ToggleButton.tsx
rasmusjeppsen: Looks great! 

**A few feedback points:**
Great points of feedback, Viktor. I agree and suggest that:
1. We should make the not-toggled option "regular" and only the toggled option "bold" consistent with the historical view toggles.
2. After seeing it live on the Netlify preview, I think we should bump up the font size to 12px to match the historical view toggles and ensure accessibility is optimal in terms of readability. In other words, bumping the size to what it was before, as Viktor suggests.

PS. I'm aware that my Figma file had some challenging font sizes! Great job Tony ⚡️

VIKTORVAV99 (approved):
Nice work on the changes, looks great now!

tonypls:
I've made a couple more changes here with feedback from Rasmus, @VIKTORVAV99 would you mind taking another quick look?

VIKTORVAV99:
> I've made a couple more changes here with feedback from Rasmus, @VIKTORVAV99 would you mind taking another quick look?

Sure thing!

web/src/components/ToggleButton.tsx
```diff
@@ -39,10 +39,15 @@ export default function ToggleButton({
...
+    <div
+      className={twMerge(
+        'z-10 flex h-9 rounded-full bg-gray-200/80 px-1  shadow    dark:bg-gray-900',
+        transparentBackground ? 'bg-gray-200/80 backdrop-blur-sm' : 'bg-gray-200'
+      )}
+    >
       <ToggleGroupRoot
         className={
-          'flex-start flex h-[26px] flex-grow flex-row items-center justify-between self-center rounded-full bg-gray-100 shadow-inner dark:bg-gray-700'
+          'flex-start flex flex-grow flex-row items-center justify-between self-center  dark:bg-gray-700'
```
VIKTORVAV99, line 50: I think you still need the rounded-full class here, I think it's removal is what made this happen: 😅
image

VIKTORVAV99 (changes requested):
I think most changes look good but something broke with the rounding in the last batch of changes.

Also, the info "i" looks strange, it has the same font size and all but for some reason it just feels off, maybe play around with that a little? But this one can be considered non-blocking.

web/src/components/ToggleButton.tsx
tonypls: Nice spotting :)

VIKTORVAV99 (approved):
Looks like the CI is timing out for some reason, but it should not be affected by the changes here so LGTM! 🎉

Nice work!

web/src/components/ToggleButton.tsx
madsnedergaard: Unfortunately the Tailwind Prettier plugin removed this check for double spaces as it was causing problems in some other edge cases: https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/71

We might be able to find another eslint plugin just for handling duplicate spaces though :)

PR #5195, opened 2023-03-13
on outside click for mobile exchange tooltip

tonypls:
## Issue
Multiple exchange tooltips

## Description
I discovered that the radix tooltip content has a nice onPointerDownOutside function so we can easily fix the multiple tooltips issue.

### DIFF
--- a/web/src/components/tooltips/MobileTooltipWrapper.tsx
+++ b/web/src/components/tooltips/MobileTooltipWrapper.tsx
@@ -41,6 +41,9 @@ export default function MobileTooltipWrapper(
             onClick={() => {
               setIsOpen(false);
             }}
+            onPointerDownOutside={() => {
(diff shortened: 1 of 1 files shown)


web/src/components/tooltips/MobileTooltipWrapper.tsx
```diff
@@ -41,6 +41,9 @@ export default function MobileTooltipWrapper(
             onClick={() => {
               setIsOpen(false);
             }}
+            onPointerDownOutside={() => {
+              setIsOpen(false);
+            }}
```
VIKTORVAV99, line 46: Well that sure was simple! 😅

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing it!

PR #5197, opened 2023-03-13
speed up generate-world by skipping validation if the world file is the same as the current one

VIKTORVAV99:
## Issue
No real issue but will improve DX.

## Description
Speed up the generate-world file by skipping validation if the existing world file is identical to the generated one.


VIKTORVAV99:
Seems like there was no need for any hashing and I overthought the whole thing in the beginning, but this should significantly speed things up when doing local development.

VIKTORVAV99:
Tony, would you mind giving this a look? The changes should be pretty straight forward.

tonypls (approved):
LGTM 🎉

madsnedergaard:
Oh hey, didn't see this PR! I did the exact same thing in https://github.com/electricitymaps/electricitymaps-contrib/pull/5182, so we should probably not double skip the check :D 
Your solution seems simpler though, so I will update my PR to remove the check again!

PR #5199, opened 2023-03-14
add tooltip for zone titles longer than 20 char

tonypls:
## Issue
Some zones have extra long names which are then truncated in the zone title

## Description
This PR adds a tooltip of the zone name to zone titles that are longer than 20 char.

image

### DIFF
--- a/web/src/features/panels/zone/ZoneHeaderTitle.tsx
+++ b/web/src/features/panels/zone/ZoneHeaderTitle.tsx
@@ -44,19 +44,24 @@ export default function ZoneHeaderTitle({
                 size={18}
                 className="shadow-[0_0px_3px_rgba(0,0,0,0.2)]"
               />
-              <div className="ml-2 flex flex-row">
-                <h2
-                  className="max-w-[300px] overflow-hidden truncate text-lg font-medium sm:max-w-[230px] md:max-w-[270px]"
-                  data-test-id="zone-name"
-                >
-                  {title}
-                </h2>
-                {isSubZone && (
-                  <p className="ml-2 flex w-auto items-center whitespace-nowrap rounded-full bg-gray-200 py-0.5 px-2  text-sm dark:bg-gray-900">
-                    {countryName || zoneId}
-                  </p>
-                )}
-              </div>
+              <TooltipWrapper
+                tooltipContent={title.length > 20 ? title : undefined}
+                side="bottom"
+              >
+                <div className="ml-2 flex flex-row">
+                  <h2
+                    className="max-w-[300px] overflow-hidden truncate text-lg font-medium sm:max-w-[230px] md:max-w-[270px]"
+                    data-test-id="zone-name"
+                  >
+                    {title}
+                  </h2>
+                  {isSubZone && (
+                    <p className="ml-2 flex w-auto items-center whitespace-nowrap rounded-full bg-gray-200 py-0.5 px-2  text-sm dark:bg-gray-900">
+                      {countryName || zoneId}
+                    </p>
+                  )}
+                </div>
+              </TooltipWrapper>
             </div>
             {disclaimer && (
               <TooltipWrapper side="bottom" tooltipContent={disclaimer}>


web/src/features/panels/zone/ZoneHeaderTitle.tsx
```diff
@@ -44,19 +44,24 @@ export default function ZoneHeaderTitle({
...
-                </h2>
-                {isSubZone && (
-                  <p className="ml-2 flex w-auto items-center whitespace-nowrap rounded-full bg-gray-200 py-0.5 px-2  text-sm dark:bg-gray-900">
-                    {countryName || zoneId}
-                  </p>
-                )}
-              </div>
+              <TooltipWrapper
+                tooltipContent={title.length > 20 ? title : undefined}
+                side="bottom"
```
VIKTORVAV99, line 49: Looks good but why are we restricting this to just zones with names longer than 20 characters?

I think it would make more sense to apply it to all zone names regardless of length to make the experience more consistent across the zones, so the users always get what they expect.

I may or may not just have spent 3 weeks working with this subject so I might be slightly biased. 😅

web/src/features/panels/zone/ZoneHeaderTitle.tsx
```diff
@@ -44,19 +44,24 @@ export default function ZoneHeaderTitle({
...
+              >
+                <div className="ml-2 flex flex-row">
+                  <h2
+                    className="max-w-[300px] overflow-hidden truncate text-lg font-medium sm:max-w-[230px] md:max-w-[270px]"
```
VIKTORVAV99, line 53: I think it makes sense to increase the max width here to avoid multiline names as much as possible. Making it wider would also avoid making it higher when the name wraps and should look more like you would expect it to. Maybe we can even skip the max width entirely for large screens?

VIKTORVAV99 (commented):
No issues with the code but I do have some comments in regard to the UX for this, feel free to double check with @rasmusjeppsen if he agrees though.

web/src/features/panels/zone/ZoneHeaderTitle.tsx
tonypls: Zone names above 20 characters are truncated.

Maybe it's just me but a tooltip that says exactly what the trigger says seems a little silly 😆 

image

web/src/features/panels/zone/ZoneHeaderTitle.tsx
tonypls: This was just an indenting change. This PR is just about adding the tooltip!

We should improve the header width scaling at some point though, I notice it playing up quite a bit on Arc with the narrower window space.

web/src/features/panels/zone/ZoneHeaderTitle.tsx
VIKTORVAV99: Then we have another problem to take into account as well. 👀
image
Here it is shown but the zone name is not truncated.

web/src/features/panels/zone/ZoneHeaderTitle.tsx
VIKTORVAV99: Oh 😅

Then I'm sorry for dumping all this feedback on a somewhat unrelated PR but we should adress it at some point.

web/src/features/panels/zone/ZoneHeaderTitle.tsx
VIKTORVAV99: Ah okay, so it's not always truncated if they are above 20 characters (if there is available screen space) but the tooltip is always shown if the zone name is above 20 characters. 

(Just summarizing things so far.) 

In that case I think we should go either with:

- only showing the tooltip if the zone name is truncated

or:

- Always show the tooltip regardless of zone name length.

Both have benefits and downsides so I'm not sure which is the best one going forward but the second option is easier to implement.

VIKTORVAV99 (approved):
Approving this as our discussions have been resolved and the changes themselves make sense 👍🏼

tonypls:
> Approving this as our discussions have been resolved and the changes themselves make sense 👍🏼

ty, it's not the perfect solution but it really helps when trying to tell some of the zones that have almost the same names apart

PR #5203, opened 2023-03-15
WIP: exchange capacity parsing for the ENTSO-E parser

VIKTORVAV99:
## Description

Adds exchange capacity parsing for the ENTSO-E parser using the day-ahead capacity information (A61).

This still need some testing and such but it's open for comments.
Unfortunately, less zones than I was initially expecting had live capacity data for exchanges, but it does work well when it is available, and it will gracefully fall back if it's not.

One thing to note is that I needed to add 2 api calls to get this data so we might have to plan for that in the rate limits and wait for [ELE-1595](https://linear.app/electricitymaps/issue/ELE-1595/rotate-ips-for-entso-e-to-avoid-hitting-rate-limits) until we merge this (1 api call if there is no data).

Also, just to note: This is technically forecasted capacity data but it should be orders of magnitudes more accurate than the static factors. 

### Preview
```python
 {'capacity': [-2300.0, 3653.0],
  'datetime': datetime.datetime(2023, 3, 13, 18, 0, tzinfo=tzutc()),
  'netFlow': 2528.0,
  'sortedZoneKeys': 'ES->FR',
  'source': 'entsoe.eu'}
```

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -525,7 +525,6 @@ def query_ENTSOE(
 def query_consumption(
     domain: str, session: Session, target_datetime: Optional[datetime] = None
 ) -> Union[str, None]:
-
     params = {
         "documentType": "A65",
         "processType": "A16",
@@ -562,7 +561,6 @@ def query_production_per_units(
     session: Session,
     target_datetime: Optional[datetime] = None,
 ) -> Union[str, None]:
-
     params = {
         "documentType": "A73",
         "processType": "A16",
@@ -585,7 +583,6 @@ def query_exchange(
     session: Session,
     target_datetime: Optional[datetime] = None,
 ) -> Union[str, None]:
-
     params = {
         "documentType": "A11",
         "in_Domain": in_domain,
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
FYI: @FelixDQ, @pierresegonne, @madsnedergaard

madsnedergaard:
This looks very nice! We should definitely get the rate-limiting fixed, so we can implement this! :)

VIKTORVAV99:
> This looks very nice! We should definitely get the rate-limiting fixed, so we can implement this! :)

I think we should try and split it into its own parser function as well before we actually merge this. While the current implementation works and provides a good proof of concept it has some limitations:

1. There is no way to run it independently from the exchange parser.
2. It's only run for the exchanges using the ENTSO-E parser.
3. It's always run for the exchanges using the ENTOS-E parser regardless of data availability.
4. It does not return future values (would give even more value in regards to guiding estimations.)

(I also think it would make sense to run this just once per hour if implemented and not every 15ish min like the exchange parsers.)

VIKTORVAV99:
@unitrium is this something we want to proceed with now that we have added the historical endpoint for ENTOS-E which free up some API calls?

Or should we add a new "capacity" endpoint to the parsers so capacity can be parsed individually from exchange data first?

VIKTORVAV99:
> Is this still relevant? :)

Somewhat, but not sure where our priorities are as it requires some new database tables to be created and all the work that goes into that + modifying the pipelines to use it.

Would be nice to get this added so it's automatic but not sure of the value vs effort ratio here.

madsnedergaard:
> > Is this still relevant? :)
> 
> Somewhat, but not sure where our priorities are as it requires some new database tables to be created and all the work that goes into that + modifying the pipelines to use it.
> 
> Would be nice to get this added so it's automatic but not sure of the value vs effort ratio here.

Makes sense, let's maybe close this PR then and keep it in mind if this gets prioritised later on? :)

PR #5216, opened 2023-03-21
Integrate Power API into British Columbia parser

juliadu:
# Summary

This PR adds changes to integrate the British Columbia endpoints of Power API into the BC parser. Note that the endpoint is currently localhost:8000. This PR is in drafts since I haven't tested it yet!


juliadu:
Created by mistake!

PR #5218, opened 2023-03-21
Tvs/theme selector mobile

tonypls:
## Issue
Adds theme selector and language selector to mobile menu using new style drop down menu

Switched from pop-over to drop down menu as the pop over wasn't scrolling when used in the settings modal.

Feedback request: Should the menu's close after selecting an item? Currently they close with any action outside the menu content. If yes, should this be the same for desktop and mobile?

[Preview here](https://deploy-preview-5218--phenomenal-syrniki-c5ceaa.netlify.app/)


web/src/features/map-controls/LanguageSelector.tsx
```diff
@@ -22,12 +23,19 @@ export function LanguageSelector() {
...
+        isMobile ? (
+          <Button icon={<HiLanguage size={21} />} className="mb-0">
+            {__('tooltips.selectLanguage')}
+          </Button>
+        ) : (
+          <MapButton
+            icon={<HiLanguage size={20} style={{ strokeWidth: '0.5' }} />}
+            tooltipText={__('tooltips.selectLanguage')}
+          />
+        )
```
VIKTORVAV99, line 35: This piece of code seems and the one below is almost identical, does it make sense to try and combine them?

web/src/features/map-controls/ThemeSelector.tsx
```diff
@@ -30,12 +31,22 @@ export default function ThemeSelector() {
...
+          >
+            {__('tooltips.changeTheme')}
+          </Button>
+        ) : (
+          <MapButton
+            icon={<BsMoonStars size={14} style={{ strokeWidth: '0.2' }} />}
+            tooltipText={__('tooltips.changeTheme')}
+            dataTestId="theme-selector-open-button"
+          />
+        )
```
VIKTORVAV99, line 47: This code 🙂

VIKTORVAV99 (commented):
Seems to work pretty good!

Do we want to add a component test in this PR as well to ensure we don't accidentally regress in the future? (or do you think this is more suited for e2e tests?)

web/src/features/map-controls/ThemeSelector.tsx
tonypls: I think it will end up being more complicated, they have different margin, icons, testId and tooltips

VIKTORVAV99 (approved):
Nice work on the tests!

PR #5219, opened 2023-03-21
close dropdown on outside click

tonypls:
## Issue
Mobile issue where closing the modal while a dropdown menu was open left the app in a frozen state

## Description
Using multiple radix components with portals can cause strange interactions, the outside click for the modal didn't properly close the portal for the dropdown menu leaving the pointer-events: none state on the body element. 

The fix to this was to specify modal={false} in the child portal, in this case the dropdown menu.

Also added closing of the dropdown menu when a value is selected.

### DIFF
--- a/web/src/components/Button.tsx
+++ b/web/src/components/Button.tsx
@@ -15,6 +15,7 @@ interface ButtonProps {
   href?: string;
   className?: string;
   onClick?: () => void;
+  renderAsDiv?: boolean;
 }
 
 export function Button({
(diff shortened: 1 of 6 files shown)


madsnedergaard (approved):
Oh, good catch! And a bit annoying that the now have to "manage" state that radix should be fully handling, but oh well ¯\\\_(ツ)\_/¯

tonypls:
> Oh, good catch! And a bit annoying that the now have to "manage" state that radix should be fully handling, but oh well ¯\_(ツ)_/¯

I found a slack overflow that mentioned using triggers correctly for dialogue components, I tried this and it didn't solve this issue. In the future switching all our modals to use trigger components could be a nice way to tidy things a little.

VIKTORVAV99:
Hehe it looks like the component tests you added are doing what they should 😅

VIKTORVAV99 (approved):
LGTM! 🎉

Issue #5036, opened 2023-02-06
Avoid displaying price data from ENTSO-E


madsnedergaard:
## Description

It has recently come to our attention that we are not allowed to show this information in the app due to licensing terms. This is the case for both for real-time and aggregated views.

If anyone knows of other sources for price data for these zones (that we are allowed to share) we would love to hear more :)

We hope to have pricing data back on the app in the future, but for now we're removing the price graph for the affected zones. 

### Zones affected

- AT
- BE
- CH
- DE
- LU
- DK (DK-DK1, DK-DK2, DK-BHM)
- FR
- GB
- NL
- NO (NO-NO1, NO-NO2, NO-NO3, NO-NO4, NO-NO5)
- PL
- SE (SE-SE1, SE-SE2, SE-SE3, SE-SE4)
- FI
- AX
- EE 
- LT
- LV

madsnedergaard:
Reopening as this has not been merged on backend yet.

Trolli132:
> ## Description
> It has recently come to our attention that we are not allowed to show this information in the app due to licensing terms. This is the case for both for real-time and aggregated views.
> 
> If anyone knows of other sources for price data for these zones (that we are allowed to share) we would love to hear more :)
> 
> We hope to have pricing data back on the app in the future, but for now we're removing the price graph for the affected zones.
> 
> ### Zones affected
> * AT
> * BE
> * CH
> * DE
> * LU
> * DK (DK-DK1, DK-DK2, DK-BHM)
> * FR
> * GB
> * NL
> * NO (NO-NO1, NO-NO2, NO-NO3, NO-NO4, NO-NO5)
> * PL
> * SE (SE-SE1, SE-SE2, SE-SE3, SE-SE4)
> * FI
> * AX
> * EE
> * LT
> * LV

Hi would the following site be a solution. It’s a site from the German government where you can download market data in 15min lot‘s. https://www.smard.de/en/downloadcenter/download-market-data/

VIKTORVAV99:
We need to double check this first but it is licensed under CC BY 4.0 even though the data comes from ENTSO-E so we might be able to use it.

It should contain data for the following regions:

[...]

PR #5220, opened 2023-03-21
Adds disclaimer about removed price data

madsnedergaard:
## Description

We removed price data in #5036, but we made a severe error in the way we approached it. Our understanding of the value and usefulness of seeing this data was wrong, and a lot of people have reached out and asked why the price data was not showing.

Unfortunately we cannot bring price data for ENTSO-E zones back on the app at this stage, but at least we can make it clear that the data is gone and help users understand what happened.

### Preview

Screenshot 2023-03-21 at 13 36 59

### Double check


web/src/features/charts/hooks/usePriceChartData.ts
```diff
@@ -30,6 +30,9 @@ export function usePriceChartData() {
     return { isLoading, isError };
   }
 
+  // We assume that if the first element has price disabled, all of them do
+  const priceDisabledReason = Object.values(zoneData.zoneStates)[0].price?.disabledReason;
+
```
VIKTORVAV99, line 35: Would it not be safer if we checked if any of them has price disabled instead of just the first?

(Might be total overkill)

web/src/features/charts/PriceChart.tsx
```diff
@@ -11,38 +12,61 @@ interface PriceChartProps {
...
-      />
+      <div className="relative">
+        {isPriceDisabled && (
+          <div className="absolute top-0 h-full w-full">
+            <div className=" h-full w-full rounded bg-white opacity-50 dark:bg-gray-700" />
+            <div className="absolute top-[50%] left-[50%] z-10 w-60 -translate-x-1/2 -translate-y-1/2 rounded-sm bg-gray-200 p-2 text-center text-sm shadow-lg dark:bg-gray-900">
+              {__(`country-panel.disabledPriceReasons.${priceDisabledReason}`)}
+            </div>
+          </div>
+        )}
```
VIKTORVAV99, line 52: Is this waiting for a backend change?

It does not show up on the live netlify preview.

tonypls (approved):
LGTM

web/src/features/charts/hooks/usePriceChartData.ts
VIKTORVAV99: This only checks the first element though, which in 99.9% of the situations should be enough but checking all values would catch 100% of the situations.

VIKTORVAV99:
Was this meant to produce this "fake" data for zones without price data?
image

madsnedergaard:
> Was this meant to produce this "fake" data for zones without price data? 

Oh definitely no, thanks for flagging it! I accidentally removed the "if check" while cleaning it up 😅

web/src/features/charts/hooks/usePriceChartData.ts
madsnedergaard: We could and I first had a `.some()`, but then we also need to handle the error message that in theory could differ. So I figured it would not be worth adding some complexity here when we control this value internally :)

VIKTORVAV99 (approved):
LGTM! 🎉

web/src/features/charts/hooks/usePriceChartData.ts
```diff
@@ -59,6 +62,8 @@ export function usePriceChartData() {
     markerFill,
     valueAxisLabel,
     layerStroke: undefined,
+    isPriceDisabled: Boolean(priceDisabledReason),
+    priceDisabledReason,
```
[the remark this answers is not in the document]
tonypls: that we're passing through two bits of data that are essentially the same instead of just passing price disabled reason then using the component to do logic.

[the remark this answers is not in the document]
madsnedergaard: Ah yeah, I was also unsure about that bit - I initially leaned towards having the logic together with the other data manipulation, but I've moved it now :)

mueslo:
The disclaimer is good, but it is still quite frustrating because it doesn't really give any details on why. Maybe link to the Github issue or explicitly mention that it is due to the ENTSO-E terms?

VIKTORVAV99:
> The disclaimer is good, but it is still quite frustrating because it doesn't really give any details on why. Maybe link to the Github issue or explicitly mention that it is due to the ENTSO-E terms?

It's just not due to ENTSO-E terms but also due to NordPool terms. We have found other sources for the price that says we can use it which is in direct contradiction to the NordPool license which causes might cause further confusion. This is also not limited to on particular issue and is spread out on few different ones so can't really just link to one of them for the full picture.

Thise issues where also not made with copywriting in mind and is not really made to distribute this information and is more about working with the community to resolve the issues, but in this case it's not much we can do until we get clarity in the matter.

I'll see if we can do something to make it even more clear as for what the reason is.

PR #5221, opened 2023-03-21
Removes Sentry tracing

madsnedergaard:
## Issue

We're above our plan quota with transactions:

Screenshot 2023-03-21 at 12 08 36

## Description

Since we don't actually use tracing across distributed systems at all, I figured it would be best to just remove it completely for now.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -53,7 +53,6 @@
     "@radix-ui/react-tooltip": "^1.0.2",
     "@react-spring/web": "^9.5.5",
     "@sentry/react": "^7.27.0",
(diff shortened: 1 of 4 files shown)


tonypls (approved):
LGTM 🎉

VIKTORVAV99 (changes requested):
We should set the sentry tracing (env?) variable to false to tree shake this!

https://docs.sentry.io/platforms/javascript/guides/react/configuration/tree-shaking/

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5222, opened 2023-03-21
Fixes net exchanges not always appearing in the graph when changing selected time

### LINKED ISSUE #5066: Fix issue with graphs and net exchanges - they come and go when dragging the time slider instead of always being there
The exchanges are not always staying like they did before the app rewrite.
This issue should have a video showing the bug :)

madsnedergaard:
## Issue

Closes #5066

## Description

Due to the way we looked at what exchanges should appear in the breakdown graph (and the breakdown "table" too), we would only show exchanges if there was data for that specific timestamp - even though the graph shows all 24 hours.

### Preview

See before and after in this video:

https://user-images.githubusercontent.com/3296643/226741749-7b0bda1e-6634-4782-ba2d-9cbd8a7c94ef.mp4

### Double check

### DIFF
--- a/web/src/features/charts/bar-breakdown/utils.test.ts
+++ b/web/src/features/charts/bar-breakdown/utils.test.ts
@@ -303,18 +303,33 @@ describe('getDataBlockPositions', () => {
 
 describe('getExchangesToDisplay', () => {
   it('shows aggregated exchanges only when required', () => {
-    const ZoneExchanges = ['AT', 'BE', 'NO', 'NO-NO2'];
-    const result = getExchangesToDisplay('DE', true, ZoneExchanges);
+    const ZoneStates = {
+      date1: {
+        ...zoneDetailsData,
+        exchange: { AT: -934, BE: 934, NO: -934, 'NO-NO2': -500 },
+      },
+    };
+    const result = getExchangesToDisplay('DE', true, ZoneStates);
     expect(result).toEqual(['AT', 'BE', 'NO']);
   });
   it('shows non-aggregated exchanges only when required', () => {
-    const ZoneExchanges = ['AT', 'BE', 'NO', 'NO-NO2'];
-    const result = getExchangesToDisplay('DE', false, ZoneExchanges);
+    const ZoneStates = {
+      date1: {
+        ...zoneDetailsData,
+        exchange: { AT: -934, BE: 934, NO: -934, 'NO-NO2': -500 },
+      },
+    };
+    const result = getExchangesToDisplay('DE', false, ZoneStates);
     expect(result).toEqual(['AT', 'BE', 'NO-NO2']);
   });
   it('handles empty exchange', () => {
-    const ZoneExchanges = [] as string[];
-    const result = getExchangesToDisplay('DE', false, ZoneExchanges);
+    const ZoneStates = {
+      date1: {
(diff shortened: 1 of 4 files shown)


web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -179,10 +180,16 @@ export const getExchangesToDisplay = (
...
-    : [];
+  // get all exchanges for the given period
+  const allExchangeKeys = new Set<string>();
+  for (const state of Object.values(zoneStates)) {
+    for (const key of Object.keys(state.exchange)) {
+      allExchangeKeys.add(key);
+    }
+  }
+  const uniqueExchangeKeys = [...allExchangeKeys];
+
```
madsnedergaard, line 191: Added this part to find unique exchanges across timestamps. It could probably be done cleaner, but it works ¯\\\_(ツ)\_/¯

VIKTORVAV99: It's weird there is no filter function on a set, if there where we could avoid the secondary constant declaration and the conversion to a array.

web/src/features/charts/hooks/useBarElectricityBreakdownChartData.ts
```diff
@@ -44,7 +44,7 @@ export default function useBarBreakdownChartData() {
   const exchangeKeys = getExchangesToDisplay(
     zoneId,
     isAggregateToggled,
-    currentData.exchange
+    zoneData?.zoneStates
```
VIKTORVAV99, line 47: Will zoneData not always be defined?

web/src/features/charts/hooks/useBreakdownChartData.ts
```diff
@@ -37,20 +37,19 @@ export const getLayerFill = (
...
+  if (isLoading || isError || !zoneData || !zoneId) {
     return { isLoading, isError };
   }
 
   const exchangesForSelectedAggregate = getExchangesToDisplay(
-    currentData.zoneKey,
+    zoneId,
     isAggregateToggled,
-    currentData.exchange
+    zoneData?.zoneStates
```
VIKTORVAV99, line 52: Same goes here, just adding a comment so we don't forget this one if we change the other one.

web/src/features/charts/bar-breakdown/utils.test.ts
VIKTORVAV99, line 1: It's great to see some more tests added!

VIKTORVAV99 (commented):
Nice work!

This was probably our biggest (visual) bug in the frontend!

web/src/features/charts/hooks/useBarElectricityBreakdownChartData.ts
madsnedergaard: Yes, good point! :)

VIKTORVAV99 (approved):
LGTM! 🎉

web/src/features/charts/bar-breakdown/utils.ts
madsnedergaard: Yeah, it is a bit messy. We could alternatively gather everything into one array, send it to a new Set() and back to remove non-unique keys, and then use that. In the end it's more or less the same though, so don't think it matters much :)

PR #5226, opened 2023-03-22
increase date container width

tonypls:
## Issue
Fixes
image


madsnedergaard:
Nice! Did you test with other languages? German tends to always be even longer 😅

web/src/features/charts/tooltips/CarbonChartTooltip.tsx
```diff
@@ -35,7 +35,7 @@ export default function CarbonChartTooltip({ zoneDetail }: InnerAreaGraphTooltip
           ></div>
           <div className="text-base font-bold">{__('tooltips.carbonintensity')}</div>
         </div>
-        <div className="my-1 h-[32px] max-w-[160px] select-none rounded-full bg-brand-green/10 py-2 px-3 text-sm text-brand-green dark:bg-gray-700 dark:text-white">
+        <div className="my-1 h-[32px] max-w-[165px] select-none rounded-full bg-brand-green/10 py-2 px-3 text-sm text-brand-green dark:bg-gray-700 dark:text-white">
```
VIKTORVAV99, line 38: Can't we just set the width to min-content instead and skip the max width?
Or do we want to keep the same width all the time?

madsnedergaard:
Nice! Did you test with other languages? Some might be even longer 😅

tonypls:
> Nice! Did you test with other languages? Some might be even longer 😅

No good deed goes unpunished 😆

madsnedergaard (approved):
👏

PR #5227, opened 2023-03-22
Reduces panel re-rerenders when dragging timeslider

madsnedergaard:
## Description

While working on #5222, I found that a lot of components were rerendering when changing the selected time, even though they didn't need this.

I have solved this here by moving the logic down to the few components that actually need it :)

### Preview

**Before**
Screenshot 2023-03-22 at 21 32 11

**After**
Screenshot 2023-03-22 at 21 32 37

### Double check


VIKTORVAV99 (approved):
The changes look so simple yet does so much, it's simply elegant in its complex simplicity.

So, this looks good to me!

Issue #4689, opened 2022-10-22
Translations needed country-zone toggle


Ankitchaudharyy:
## Description
Translation for `tooltips.aggregateinfo`, `aggregateButtons.country` & `aggregateButtons.zone` are missing in locals/*.json for all languages except English, Espanol, and Svenska.

Edit:
Dutch (DE) has also been translated in https://github.com/electricitymaps/electricitymaps-contrib/pull/5228

madsnedergaard:
This has been fixed :)

PR #5228, opened 2023-03-22
fix aggregateButtons country-zone toggle translation [de.json]

gnzng:
## Issue

translation for country-zone toggle missing as mentioned in issue #4689.

## Description

Added translation for german in the country-zone toggle by using the [web/public/locales/sv.json](https://github.com/electricitymaps/electricitymaps-contrib/blob/ec7d7c8b16e1bb285ea209ff96e56870cc92d03c/web/public/locales/sv.json#L173-L175) template, which works with the current web version. 

### Preview

Did not run it locally.

### DIFF
--- a/web/public/locales/de.json
+++ b/web/public/locales/de.json
@@ -125,6 +125,10 @@
     "zoomOut": "Raus zoomen",
     "aggregateInfo": "Wechsel zwischen spezifischer Zonenansicht und aggregierter Länderansicht"
   },
+  "aggregateButtons": {
+    "country": "Land",
+    "zone": "Gebiet"
+  },
   "themeOptions": {
     "dark": "Dunkel",
     "light": "Hell",


VIKTORVAV99 (approved):
Thanks for opening this PR!

I'll approve and merge this right away but it might take some time for it to show up in the app depending on the platform.

madsnedergaard:
In other German translations we use "Zonen" instead of "Gebiet" - how do they compare? :)

gnzng:
The more I think about it, I would say that 'Zone' actually is a better and more technical term. It describes a more regulated zone/area, also in e.g. mail, traffic, etc...

gnzng:
I just opened a PR (#5231) to change the toggle button name to 'Zone'.

madsnedergaard:
Thanks for the update and for following up 👏

PR #5229, opened 2023-03-23
allow landscape for android and hide header when native

tonypls:
## Issue
We've had some complaints about not being able to use landscape mode on tablets, this PR enables the mobile apps to use landscape mode and hides the header when on landscape in mobile. 

iOS was already allowing landscape

### DIFF
--- a/mobileapp/android/app/src/main/AndroidManifest.xml
+++ b/mobileapp/android/app/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
             android:theme="@style/AppTheme.NoActionBarLaunch"
             android:launchMode="singleTask"
             android:exported="true"
-            android:screenOrientation="portrait">
+            android:screenOrientation="unspecified">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
--- a/mobileapp/ios/App/App.xcodeproj/project.pbxproj
+++ b/mobileapp/ios/App/App.xcodeproj/project.pbxproj
@@ -194,10 +194,10 @@
 			buildActionMask = 2147483647;
 			files = (
 			);
-			inputPaths = (
+			inputFileListPaths = (
 			);
 			name = "[CP] Embed Pods Frameworks";
-			outputPaths = (
+			outputFileListPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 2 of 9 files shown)


mobileapp/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,7 @@
             android:theme="@style/AppTheme.NoActionBarLaunch"
             android:launchMode="singleTask"
             android:exported="true"
-            android:screenOrientation="portrait">
+            android:screenOrientation="unspecified">
```
VIKTORVAV99, line 18: If this is unspecified can't we just remove it or is it still a requirement for capacitor/the APK?

mobileapp/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -194,10 +194,10 @@
 			buildActionMask = 2147483647;
 			files = (
 			);
-			inputPaths = (
+			inputFileListPaths = (
 			);
 			name = "[CP] Embed Pods Frameworks";
-			outputPaths = (
+			outputFileListPaths = (
```
VIKTORVAV99, line 200: I assume this is just a update to use a new syntax/config?

VIKTORVAV99 (approved):
I trust you tested this locally as I don't have Android Studio installed to test things 😅

Two small comments for some clarity but I don't think there is any issues here so I'm approving it.

mobileapp/android/app/src/main/AndroidManifest.xml
tonypls: yeah it is the default so we can remove it, little more explicit and makes it a little easier for searching later if it's there. Can remove if you think it's nicer ✂️

mobileapp/ios/App/App.xcodeproj/project.pbxproj
tonypls: Not 100% what caused this but it was a generated change during the capacitor build process

mobileapp/android/app/src/main/AndroidManifest.xml
VIKTORVAV99: I don't think it matters much but I guess removing it could have unintended consequences if the default change?

mobileapp/ios/App/App.xcodeproj/project.pbxproj
VIKTORVAV99: Let's not mess with the generated files 😅

VIKTORVAV99 (approved):
New changes look good as well 👍🏼

PR #5230, opened 2023-03-23
More helpers and tests + tests for existing functions

VIKTORVAV99:
## Description

Adds more helper functions and add tests for them as well as some of the existing functions.

It should be possible to review this per commit for related changes if I didn't mess anything up.

### ToDo

~~Rewrite `createToWithState()` to be a pure function and add tests for it.~~
- [x] Add tests for `getProductionCo2Intensity()`

### Double check


VIKTORVAV99:
I decided not to create a test for `createWithState()` in this PR as there was no easy way to mock the location object and I wanted to keep this PR as clean as possible.

We can also rewrite it to be a "pure" function that don't rely on location but it's a pretty simple function so I'm not sure if the changes are worth it.

web/src/features/map/MapTooltip.tsx
```diff
@@ -39,11 +39,21 @@ function TooltipInner({
...
+    co2intensityProduction
+  );
+  const fossilFuelPercentage = getFossilFuelRatio(
     isConsumption,
     fossilFuelRatio,
     fossilFuelRatioProduction
   );
+  const renewable = getRenewableRatio(
+    isConsumption,
+    renewableRatio,
```
tonypls, line 54: Would have been nice for all of these if we named the data field renewableRationConsumption here

tonypls:
> 

Cypress works well with the router :)

tonypls (approved):
LGTM

web/src/features/map/MapTooltip.tsx
VIKTORVAV99: Agreed, I can add it to this PR if you wish, should be a simple change of renaming the symbols.

web/src/utils/helpers.test.ts
```diff
@@ -1,59 +1,187 @@
...
+  it('returns 10.7 when the type is hydro', () => {
+    const actual = getProductionCo2Intensity('hydro', zoneDetailMock);
+    expect(actual).toBe(10.7);
+  });
+
+  it('returns 136.442_667_362_438_53 when the type is battery storage', () => {
+    const actual = getProductionCo2Intensity('battery storage', zoneDetailMock);
+    expect(actual).toBe(136.442_667_362_438_53);
+  });
+});
```
madsnedergaard, line 53: I don't think you need to add the correct value to the "title" here, as it just means having to change in two places if the number changes - and when running tests and getting an overview, we shouldn't have to care about what the specific value is :)

Perhaps we could name it something like `it('returns correct value for battery storage')` instead?

madsnedergaard (approved):
Nice! One very minor comment here :)

web/src/utils/helpers.test.ts
VIKTORVAV99: I kind of forgot about this PR but I'll update the tiles and merge this. 😅

PR #5231, opened 2023-03-23
change german term for 'zone'

gnzng:
## Issue

As discussed in the comments of merged PR #5228 from yesterday I would change the german translation from 'Gebiet' to the more technical term 'Zone' in german translation to keep consistency. Sorry!


tonypls (approved):
Sieht für mich gut aus!

PR #5234, opened 2023-03-24
component tests for chart tooltips

tonypls:
## Issue
We love tests 

## Description
Add some simple component tests for the chart tooltips, it might be overkill to do the same for desktop and mobile.

PR #5239, opened 2023-03-27
add istanbul test coverage

tonypls:
## Issue
Add test coverage so we know where to invest effort to increase testing across the app

## Description
We weren't sure on our test coverage so I've added [Istanbul](https://istanbul.js.org/) which works with vite and cypress component tests :)

To view the the report, run cypress component tests using
```
pnpm exec cypress run --component
```
then
Open the static file to view the results:
electricitymap-contrib/web/coverage/lcov-report/index.html

P.S.

This report was generated with the PR included #5234 

### Preview

image

### DIFF
--- a/web/cypress.config.ts
+++ b/web/cypress.config.ts
@@ -1,4 +1,5 @@
 import { defineConfig } from 'cypress';
+import task from '@cypress/code-coverage/task';
 
 export default defineConfig({
   fileServerFolder: 'dist',
@@ -14,6 +15,15 @@ export default defineConfig({
     devServer: {
       framework: 'react',
       bundler: 'vite',
+    },
+    setupNodeEvents(on, config) {
+      task(on, config);
+      return config;
+    },
+  },
+  env: {
+    codeCoverage: {
+      exclude: 'cypress/**/*.*',
     },
   },
 });
--- a/web/cypress/support/component.ts
+++ b/web/cypress/support/component.ts
@@ -18,6 +18,8 @@ import './commands';
 
 import '../../src/index.css';
 
+import '@cypress/code-coverage/support';
+
 import { mount } from 'cypress/react18';
 
 Cypress.Commands.add('mount', mount);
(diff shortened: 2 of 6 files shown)


VIKTORVAV99 (commented):
This seems to only include the cypress component tests and not the vitest unit tests as well, was this intentional?

tonypls:
> This seems to only include the cypress component tests and not the vitest unit tests as well, was this intentional?

Excellent point :D

I might not have time to look at this till next week, feel free to add a commit if it's an easy fix :)

VIKTORVAV99:
> > This seems to only include the cypress component tests and not the vitest unit tests as well, was this intentional?
> 
> Excellent point :D
> 
> I might not have time to look at this till next week, feel free to add a commit if it's an easy fix :)

Should be easy enough to get working but combining the two (vitest and cypress) might be the tricky part.

madsnedergaard:
We are already using Istanbul for vitest coverage. Maybe we should just rename the reports/folders to ensure they don't overwrite each other?
Screenshot 2023-04-03 at 14 08 17

tonypls:
> > We are already using Istanbul for vitest coverage. Maybe we should just rename the reports/folders to ensure they don't overwrite each other? Screenshot 2023-04-03 at 14 08 17
> 
> Yeah I noticed that as well. I was thinking there might be a way to merge the reports into one but maybe that's overkill and more work than needed?

I think as these are used as an ad hoc check and improvement tool it think it's okay to have two reports.

VIKTORVAV99:
> > > We are already using Istanbul for vitest coverage. Maybe we should just rename the reports/folders to ensure they don't overwrite each other? Screenshot 2023-04-03 at 14 08 17
> > 
> > 
> > Yeah I noticed that as well. I was thinking there might be a way to merge the reports into one but maybe that's overkill and more work than needed?
> 
> I think as these are used as an ad hoc check and improvement tool it think it's okay to have two reports.

I was thinking more about the draft issue about adding regression CI checks which this would be a part of. Then it would be nice to have it all as a unified report for the bots or whatnot we decide to use.

But maybe I'm thinking too far ahead again?

tonypls:
Closing as stale

PR #5240, opened 2023-03-27
updated FR parser to enable historical backfill

mathilde-daugy:
## Issue

We noticed that we couldn't refetch data that was older than 9 months using the current dataset. We know that RTE publishes final data at Y+1 and we want to be able to update our dataset with the most up-to-date data.
image

## Description

the aim here is to enable historical refetch. the historical data is available in the dataset `eco2mix-national-cons-def`

### Preview
2023-03-26
image

2021-12-31

image

### Double check

### DIFF
--- a/parsers/FR.py
+++ b/parsers/FR.py
@@ -37,12 +37,14 @@ MAP_HYDRO = [
     "pompage",
 ]
 DATASET_RT = "eco2mix-national-tr"
-DATASET_CONSOLIDATED = "eco2mix-national-cons-def" # API called is Données éCO2mix nationales consolidées et définitives for datetimes older than 9 months
+DATASET_CONSOLIDATED = "eco2mix-national-cons-def"  # API called is Données éCO2mix nationales consolidées et définitives for datetimes older than 9 months
+
 
 def is_not_nan_and_truthy(v) -> bool:
     if isinstance(v, float) and math.isnan(v):
         return False
     return bool(v)
+
 
 def get_dataset_from_datetime(target_datetime: datetime) -> str:
     """Returns the dataset to query based on the target_datetime. The real-time API returns no values for target datetimes older than 9 months and we need to query the consolidated dataset."""
@@ -53,12 +55,14 @@ def get_dataset_from_datetime(target_datetime: datetime) -> str:
         dataset = DATASET_RT
     return dataset
 
-def get_data(session: Optional[Session] = None,
+
+def get_data(
+    session: Optional[Session] = None,
     target_datetime: Optional[datetime] = None,
-    ) -> pd.DataFrame:
+) -> pd.DataFrame:
     """Returns a DataFrame with the data from the API."""
     if target_datetime:
-        target_datetime =arrow.get(target_datetime).replace(tzinfo="Europe/Paris")
+        target_datetime = arrow.get(target_datetime).replace(tzinfo="Europe/Paris")
     else:
         target_datetime = arrow.now(tz="Europe/Paris")
 
@@ -87,6 +91,7 @@ def get_data(session: Optional[Session] = None,
     df = pd.DataFrame(data)
     return df
 
+
 @refetch_frequency(timedelta(days=1))
 def fetch_production(
     zone_key: str = "FR",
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (changes requested):
Note this is not a review of the code itself but the implementation of the new data source. In #4853 there was some concerns about validating the data source and making sure all the data was valid before we proceeded. Is this still the case?

If not might it not be better to proceed with #4853 and split France up in the process when switching the parsers (and aggregating it again together with corsica)?

mathilde-daugy:
> If not might it not be better to proceed with #4853 and split France up in the process when switching the parsers (and aggregating it again together with corsica)?

Hi @VIKTORVAV99, I don't know how much we want to prioritize splitting FR into regions.  We realised we couldn't backfill and that the data can be significantly modified and thought that this was more important to the Sales team. As you can see on the OpenData [page](https://odre.opendatasoft.com/explore/dataset/eco2mix-national-tr/information/?disjunctive.nature), the data we collect is provisional.

VIKTORVAV99:
> > If not might it not be better to proceed with #4853 and split France up in the process when switching the parsers (and aggregating it again together with corsica)?
> 
> Hi @VIKTORVAV99, I don't know how much we want to prioritize splitting FR into regions. We realised we couldn't backfill and that the data can be significantly modified and thought that this was more important to the Sales team. As you can see on the OpenData [page](https://odre.opendatasoft.com/explore/dataset/eco2mix-national-tr/information/?disjunctive.nature), the data we collect is provisional.

Perhaps we should increase the priority of https://github.com/electricitymaps/electricitymaps/pull/3230, since it would allow us to fetch estimated data (separate from validated data)? Seems like yet another scenario where it would be beneficial.

I'll do a full code review this afternoon as well 👍🏻

parsers/FR.py
```diff
@@ -42,25 +46,36 @@ def is_not_nan_and_truthy(v) -> bool:
...
-@refetch_frequency(timedelta(days=1))
-def fetch_production(
-    zone_key: str = "FR",
+def get_dataset_from_datetime(target_datetime: datetime) -> str:
+    """Returns the dataset to query based on the target_datetime. The real-time API returns no values for target datetimes older than 9 months and we need to query the consolidated dataset."""
+    if target_datetime < arrow.now(tz="Europe/Paris").shift(months=-9):
+        # API called is Données éCO2mix régionales consolidées et définitives for datetimes that are more than 9 months in the past
+        dataset = DATASET_CONSOLIDATED
+    else:
+        dataset = DATASET_RT
```
unitrium, line 55:
```suggestion
        dataset = DATASET_REAL_TIME
```
Same as above.

unitrium (approved):
Just some naming suggestions but looks good to me!

parsers/FR.py
```diff
@@ -42,25 +46,36 @@ def is_not_nan_and_truthy(v) -> bool:
...
-    formatted_to = to.format("YYYY-MM-DDTHH:mm")
+    formatted_from = target_datetime.shift(days=-1).format("YYYY-MM-DDTHH:mm")
+    formatted_to = target_datetime.format("YYYY-MM-DDTHH:mm")
 
     params = {
-        "dataset": "eco2mix-national-tr",
+        "dataset": dataset,
         "q": "date_heure >= {} AND date_heure <= {}".format(
             formatted_from, formatted_to
         ),
```
VIKTORVAV99, line 81: I relize you didn't change this but could we make this an f"string" when we are changing things anyway and get rid of the fromat function.

parsers/FR.py
```diff
@@ -42,44 +46,63 @@ def is_not_nan_and_truthy(v) -> bool:
...
-    logger: Logger = getLogger(__name__),
-) -> list:
+) -> pd.DataFrame:
+    """Returns a DataFrame with the data from the API."""
     if target_datetime:
-        to = arrow.get(target_datetime, "Europe/Paris")
+        target_datetime = arrow.get(target_datetime).replace(tzinfo="Europe/Paris")
     else:
-        to = arrow.now(tz="Europe/Paris")
+        target_datetime = arrow.now(tz="Europe/Paris")
```
VIKTORVAV99, line 67: This is technically a type regression. As we define target_datetime to be Optional Datetime we can not assing a arrow object to it. We either need to remove the use of arrow, return the datetime instead of the arrow object or use a temporary variable like before.

VIKTORVAV99 (approved):
Everything looks good to me with the exception of the type issue. If you can take care of that it would be greatly appreciated.

PR #5242, opened 2023-03-28
fix lockfile

tonypls:
## Issue

Deploy issues with lockfile

See here https://github.com/pnpm/action-setup/issues/40


unitrium (approved):
Thanks I was about to write a message to mention this issue :)

PR #5245, opened 2023-03-28
remove DEMAND_URL (twice)

Manu1400:
## Description

remove same variable declared twice


parsers/IN.py
```diff
@@ -53,7 +53,6 @@
     "उत्तर-पूर्वी क्षेत्र  / North-Eastern Region": "IN-NE",
 }
 
-DEMAND_URL = "https://vidyutpravah.in/state-data/{state}"
 STATES_MAPPING = {
```
VIKTORVAV99, line 56: Looks like there is a duplicate `STATES_MAPPING` as well, would you mind removing that as well before we merge this?

VIKTORVAV99 (approved):
I removed the other duplicate mapping so this can now be merged.

PR #5248, opened 2023-03-29
Improve validate_config_filenames.py script

qyearsley:
## Description

I noticed that the validate_config_filenames.py script, unlike the other scripts, has everything at top-level instead of in functions. In addition to changing this, this PR also:

 - Adds a file-level docstring
 - Changes the "if flag == False:" pattern to more idiomatic "if flag:".

### Double check

- [N/A] I have tested my parser changes locally with `poetry run test_parser "zone_key"`


scripts/validate_config_filenames.py
```diff
@@ -1,40 +1,58 @@
+"""
+This script checks if the config filenames are valid.
+
+This may be run as part of CI or locally to ensure that the config filenames
+are valid.
```
VIKTORVAV99, line 5:
```suggestion
This is run as a part of the CI but can be run locally to ensure all filenames are valid before you commit the changes.
```

scripts/validate_config_filenames.py
```diff
@@ -1,40 +1,58 @@
...
+        print("There are errors in the above exchange filenames!")
+    else:
+        print("All exchange filenames are valid!")
+
+    if zone_error or exchange_error:
+        exit(1)
+
+
+if __name__ == "__main__":
+    main()
```
VIKTORVAV99, line 58: I'm not really sure why this is needed, won't it add additional overhead (albeit small).

VIKTORVAV99 (commented):
Just some small comments.

scripts/validate_config_filenames.py
qyearsley: Thanks for review.

This is intended to improve readability (mostly through consistency). Python scripts in general, including all other scripts here, usually have functions at the top level, and the only statement at the top level is `if __name__ == "__main__": main()` at the bottom.

Having a consistent style across scripts (and across code in general) makes them easier to read. Some reasons to organize code in functions and conditionally call main at the bottom are:

1. it makes it so that it's possible to import modules without creating side-effects (https://realpython.com/python-main-function/#best-practices-for-python-main-functions).
2. it's a familiar pattern for Python programs and across programming languages.

This PR also changes a couple of other unusual or un-idiomatic things:

```
    if zone_error == False:
        print('OK')
    elif zone_error == True:
        print('Errors')
```
becomes
```
    if zone_error:
        print("Errors")
    else:
        print("OK")
```

And string literals above statements (which are not file or function docstrings) become regular Python comments.

VIKTORVAV99: hmm I'm not sure if changing the if statments here make sense, for one the first one is more clear as to what exactly we are checking and more type safe as the second one is only checking if zone_error is truthy and printing ok in every other case, even if it would be undefined.

There is also the order of operations to consider here, since we expect all names to be error free 99.9% of the time it makes sense to check and print that first, to avoid the secondary else or elif statment. This is into micro optimization territory but it's similar to an early return pattern that is commonly used.

VIKTORVAV99: Okay let's get rid of the strict Boolean check then, not I don't really get why it would be a anti pattern in a loosely typed language but that's a whole different thing.

But I do think we should change back the order of the if else statement like the last example you provided.

qyearsley: Yep, makes sense, and I see what you're saying. Now changed. I also thought maybe using has_zone_error is slightly more explicit/clear than just zone_error?

Note that this is also a convention other languages. The explanation given is just "comparing booleans with boolean literals is unnecessary, directly using the booleans is more concise".

 - https://typescript-eslint.io/rules/no-unnecessary-boolean-literal-compare/
 - https://dartcodemetrics.dev/docs/rules/common/no-boolean-literal-compare
 - https://progmiscon.org/misconceptions/Java/ComparisonWithBooleanLiteral/

VIKTORVAV99 (approved):
LGTM! 🎉

Let's get this merged!

PR #5255, opened 2023-04-03
prevent exchanges rendering while moving the map

tonypls:
## Issue
When zooming the map it was possible to quickly drag the map which would cause the exchange arrows to render statically while the map was still being moved. This resulted in misaligned exchange arrows

## Description
The PR adds specific functions for the onDragStart/End and onZoomStart/End callbacks within the map, these functions check if the map is still zooming or dragging before flagging the map as stationary.

### Preview
Before:

https://user-images.githubusercontent.com/18622768/229500754-3e3bfaa3-bbed-4c1f-9743-8c1f25e1d18e.mov

After:

https://user-images.githubusercontent.com/18622768/229500794-81b8a568-9cf5-4fc4-9209-fc885f4cee63.mov


VIKTORVAV99 (approved):
LGTM! 🎉

I can no longer replicate this locally and the code changes look clean!

PR #5261, opened 2023-04-03
update fetch_production

mathilde-daugy:
## Issue
the datetime stored in the database are weird due to a pytz issue
image

## Description

We want data to be stored at the start of the hour to have consistency with other zones

### Preview
we do not use the pytz library anymore to avoid this issue
image

### Double check

### DIFF
--- a/parsers/US_PJM.py
+++ b/parsers/US_PJM.py
@@ -169,9 +169,7 @@ def fetch_production(
                 production[mode] = row.get("mw")
         data_point = {
             "zoneKey": zone_key,
-            "datetime": arrow.get(dt).datetime.replace(
-                tzinfo=pytz.timezone("US/Eastern")
-            ),
+            "datetime": arrow.get(dt).replace(tzinfo="US/Eastern").datetime,
             "production": production,
             "storage": storage,
             "source": "pjm.com",


mathilde-daugy:
@VIKTORVAV99 I added a ParserException to fetch_production to understand our logs better. we can't collect data between 23 and 4 which is a bit odd. Let me know if this is still good to merge :)
image

VIKTORVAV99:
> @VIKTORVAV99 I added a ParserException to fetch_production to understand our logs better. we can't collect data between 23 and 4 which is a bit odd. Let me know if this is still good to merge :) image

I'll take another look and update my review right away!

parsers/US_PJM.py
```diff
@@ -387,13 +391,13 @@ def fetch_price(
...
     print("fetch_production() ->")
-    print(fetch_production())
-    print("fetch_exchange(US-NY, US-PJM) ->")
-    print(fetch_exchange("US-NY", "US-PJM"))
-    print("fetch_exchange(US-MISO, US-PJM)")
-    print(fetch_exchange("US-MISO", "US-PJM"))
-    print("fetch_price() ->")
-    print(fetch_price())
+    print(fetch_production(target_datetime=datetime(2022, 2, 4, 23)))
+    # print("fetch_exchange(US-NY, US-PJM) ->")
```
unitrium, line 398: Is there a particular reason all of those have been commented out ?
I don't think it's necessary to introduce this change unless the functions have been depreciated :)

unitrium (approved):
Looks good, I just added a comment for the changes in the main. Those might be unnecessary changes so better to leave them out to keep the PR clean 🫧

parsers/US_PJM.py
mathilde-daugy: forgot to uncomment ...

PR #5264, opened 2023-04-04
md/update gas EF for eu

mathilde-daugy:
## Issue

Fixes issue [1878](https://linear.app/electricitymaps/issue/ELE-1878/suspiciously-high-co2-intensity-for-gas-in-ro)

## Description
Gas emission factor for Romania was way too high because of undetected outliers
We have now set the upper limit to 800 gCO2eq/kWh for gas
This also makes the EU average decrease hence the modification of EF for zones which use the default EU average

### Preview

See this [sheet](https://docs.google.com/spreadsheets/d/1dITlaKmWC5E21M-CTRhQzgNqjPz7YjqW7U6dGEQOf9A/edit?usp=sharing) for updated factors

### Double check

### DIFF
--- a/config/zones/FI.yaml
+++ b/config/zones/FI.yaml
@@ -53,7 +53,7 @@ emissionFactors:
     gas:
       datetime: '2021-01-01'
       source: EU-ETS, ENTSO-E 2021
-      value: 684.2607818
+      value: 295.9885661
     hydro discharge:
       - datetime: '2015-01-01'
         source: Electricity Maps, 2015 average
@@ -115,7 +115,7 @@ emissionFactors:
     gas:
       datetime: '2021-01-01'
       source: EU-ETS, ENTSO-E 2021; IPCC 2014
-      value: 804.2607818
+      value: 415.9885661
     hydro:
       datetime: '2020-01-01'
       source: UNECE 2022
--- a/config/zones/FO-MI.yaml
+++ b/config/zones/FO-MI.yaml
@@ -21,7 +21,7 @@ emissionFactors:
     gas:
       datetime: '2021-01-01'
       source: EU-ETS, ENTSO-E 2021
-      value: 419.8221756
+      value: 336.9719859
     oil:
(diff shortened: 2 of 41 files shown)


config/zones/FI.yaml
```diff
@@ -53,7 +53,7 @@ emissionFactors:
     gas:
       datetime: '2021-01-01'
       source: EU-ETS, ENTSO-E 2021
-      value: 684.2607818
```
VIKTORVAV99, line 56: Any reason for why the change here is so massive?

config/zones/FO-MI.yaml
VIKTORVAV99, line 1: The FO subzone numbers seems to have increased but we are missing the change to the FO.yaml file as well. (will be identical numbers to the subzones but needed for the frontend to represent the new values in aggregated mode)

VIKTORVAV99 (changes requested):
Only thing blocking this from my side is that we need to update the FO.yaml file as well.

config/zones/FO-MI.yaml
mathilde-daugy: We were previously using the gas EF from DK and I think it makes more sense to use the EU average maybe? especially since they do not have gas generation.
What do you think

config/zones/FI.yaml
mathilde-daugy: there was also probably one power plant with a suspiciously high EF which we have now removed from the set

config/zones/FO-MI.yaml
VIKTORVAV99: Yeah I think it does so in that regard it looks great but the FO.yaml file is still using the old DK numbers. So that needs a bump as well.

mathilde-daugy: ah i didn't realise this. I'll just update the FO files with the numbers so they are consistent

VIKTORVAV99 (approved):
LGTM! 🎉

Nice to see we are getting even more accurate emission factors!

Issue #3990, opened 2022-04-20
Check value of co2eq files at test time


pierresegonne:
After the amazing work of @kruschk in #3937, we now have a nice setup for testing the co2eq parameter files ([co2eq_parameters_all](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_all.json), [co2eq_parameters_direct](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_direct.json), [co2eq_parameters_lifecycle](https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_lifecycle.json))

The tests currently ran can be extended to also check the values of the given parameters. 
We for example currently check that origin ratios sum to 1. See: https://github.com/electricitymap/electricitymap-contrib/blob/e73130642064acd16e178a80e035b4d2584c7f8d/tests/test_co2eq_parameters.py#L79-L91

But don't run any checks on the value of the emission factors. 
We could for example:
- Check that all emission factors are >= 0
- Check that all emission factors are <= 2000
- Could also do per mode check, using large uncertainty around the default values
- etc (be creative 😄 )

du-phan:
Hi @pierresegonne, 
First-time contributor here, I can try this out if it's ok for you ?

pierresegonne:
Of course! Everyone is more than welcome to contribute :)
Sorry for the delay, I was on holidays

ZachEichen:
Hi, if this issue is still relevant I'd like to give it a go as well. 
Hi, I'm also a first timer, but I'll have a look into it.

kruschk:
Hey, @ZachEichen! I can't say whether this issue is still relevant because I haven't contributed in a while. I wrote the original test suite, though, so let me know if you have any questions and I'll try to answer them!

qyearsley:
Quick note here, as noted in the above PR there are more potential improvements here, including splitting expected ranges into direct and lifecycle emissions sections. I plan to do that in a follow up.

PR #5269, opened 2023-04-08
Refactor by splitting out functionality from config/__init__.py

qyearsley:
## Issue

Context: I was starting to look at #3990, so I was trying to understand more about the config, and how it is loaded and tested.

## Description

I noticed that electricitymap/contrib/config/__init__.py contained a mix of code and global variables and functions at top-level. I felt that perhaps it might be easier to read if it were reorganized a little so that:

 - `config/__init__.py` contains basically just the top-level global variables that are used elsewhere
 - Helper functions are separated out and tested. Specifically, types that are used throughout the package were put in config_types.py, and reading functions put in reading.py and other functions put in zones.py and co2eq_parameters.py. I'm totally open to renaming any of those or changing how they may be split up.

There are no changes to functionality, this is intended purely as a refactoring change. All tests still pass and a few new tests have been added.

I'm happy un-do any part of this if project owners feel that some parts don't improve organization/readability.

## Checks

- [X] I have run `poetry run format` to format my changes.
- [X] I have run Python tests with `poetry run pytest`.


electricitymap/contrib/config/co2eq_parameters.py
```diff
@@ -0,0 +1,60 @@
+"""Contains a function to make co2eq parameter dicts from
```
qyearsley, line 1: This could potentially put in the same file as the functions in zones.py (and reading.py) and could be called utils.py or something. Similar with config_types.py.

tests/test_config.py
qyearsley, line 1: I could potentially add more tests here.

electricitymap/contrib/config/__init__.py
```diff
@@ -1,137 +1,62 @@
...
-            zone_neighbours[zone_name_1].add(zone_name_2)
-    # Sort
-    return {k: sorted(v) for k, v in zone_neighbours.items()}
-
-
 ALL_NEIGHBOURS: Dict[ZoneKey, List[ZoneKey]] = generate_all_neighbours(EXCHANGES_CONFIG)
 
 
-def emission_factors(zone_key: ZoneKey) -> Dict[str, float]:
+def emission_factors(zone_key):
```
qyearsley, line 55: I kept this here in this change because it's used elsewhere and directly references CO2EQ_PARAMETERS. It could potentially also be moved and refactored to take co2eq_parameters as an argument.

qyearsley (commented):
Note: If owners of this project don't think that any part this improves readability / organization I'm happy to make changes. I'm also happy to add more tests in this PR or another.

tests/test_exchanges_json.py
```diff
@@ -2,18 +2,17 @@
 
 from electricitymap.contrib.config import EXCHANGES_CONFIG, ZONES_CONFIG
 
-ZONE_KEYS = ZONES_CONFIG.keys()
-
 
 class ExchangeJsonTestcase(unittest.TestCase):
     def test_all_zones_in_zones_json(self):
+        zone_keys = ZONES_CONFIG.keys()
```
VIKTORVAV99, line 8: I don't think this should be moved since it's a constant?

tests/test_zones_json.py
```diff
@@ -15,10 +15,11 @@ def test_bounding_boxes(self):
                 self.assertLess(bbox[0][1], bbox[1][1])
 
     def test_sub_zones(self):
+        zone_keys = set(ZONES_CONFIG.keys())
```
VIKTORVAV99, line 18: Same here.

electricitymap/contrib/config/co2eq_parameters.py
```diff
@@ -0,0 +1,60 @@
...
+config read from defaults.yaml and zones/*.yaml.
+"""
+
+from copy import deepcopy
+from typing import Dict
+
+
+def generate_co2eq_parameters(defaults, zones_config):
```
VIKTORVAV99, line 9: It would be great if this had a return signature.

electricitymap/contrib/config/__init__.py
```diff
@@ -1,137 +1,59 @@
...
```
VIKTORVAV99, line 55: This had a return signature before, why was it removed?

electricitymap/contrib/config/__init__.py
```diff
@@ -1,137 +1,59 @@
+"""Global config variables with data read from the config directory."""
+
 from collections import defaultdict
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, List, NewType, Tuple
+from typing import Dict, List
 
 import yaml
```
VIKTORVAV99, line 8: This import is no longer used.

electricitymap/contrib/config/__init__.py
```diff
@@ -1,137 +1,59 @@
+"""Global config variables with data read from the config directory."""
+
 from collections import defaultdict
```
VIKTORVAV99, line 3: Neither is this.

electricitymap/contrib/config/co2eq_parameters.py
```diff
@@ -0,0 +1,60 @@
+"""Contains a function to make co2eq parameter dicts from
+config read from defaults.yaml and zones/*.yaml.
+"""
+
+from copy import deepcopy
```
VIKTORVAV99, line 5: This is also not used.
(Neither is the typing below but it should be.)

electricitymap/contrib/config/__init__.py
```diff
@@ -1,137 +1,59 @@
...
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, List, NewType, Tuple
+from typing import Dict, List
 
 import yaml
 
+from electricitymap.contrib.config.co2eq_parameters import generate_co2eq_parameters
+from electricitymap.contrib.config.config_types import BoundingBox, ZoneKey
 from electricitymap.contrib.config.constants import EXCHANGE_FILENAME_ZONE_SEPARATOR
```
VIKTORVAV99, line 12: This is also not used.

electricitymap/contrib/config/reading.py
```diff
@@ -0,0 +1,35 @@
...
+import yaml
+
+from electricitymap.contrib.config.config_types import ZoneKey
+from electricitymap.contrib.config.constants import EXCHANGE_FILENAME_ZONE_SEPARATOR
+
+
+def read_defaults(config_dir) -> Dict[str, Any]:
+    """Reads the defaults.yaml file."""
+    defaults_path = config_dir.joinpath("defaults.yaml")
+    return yaml.safe_load(open(defaults_path, encoding="utf-8"))
```
VIKTORVAV99, line 12: Here the return type don't match the return signature. We probably need to validate the type first with `isinstance`.

electricitymap/contrib/config/reading.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any, Dict, List, NewType, Tuple
```
VIKTORVAV99, line 1: Here is 3 unused imports.

electricitymap/contrib/config/zones.py
```diff
@@ -0,0 +1,66 @@
+"""This module contains functions for working with zones configs."""
+
+from collections import defaultdict
+from copy import deepcopy
+from pathlib import Path
```
VIKTORVAV99, line 5: These are unused.

electricitymap/contrib/config/zones.py
```diff
@@ -0,0 +1,66 @@
...
+    zone_neighbours = defaultdict(set)
+    for k, v in exchanges_config.items():
+        if not v.get("parsers", {}).get("exchange", None):
+            # Interconnector config has no parser, and will therefore not be part
+            # of the flowtracing graph.
+            continue
+        zone_1, zone_2 = k.split("->")
+        if zones_config[zone_1].get("subZoneNames") or zones_config[zone_2].get(
+            "subZoneNames"
+        ):
```
VIKTORVAV99, line 47: Here is a type error as we can't use the zone_1 and zone_2 strings to access the zone_config that has the type ZoneKey. We need to cast these to correct types first.

tests/test_config.py
```diff
@@ -1,3 +1,13 @@
...
+package.
+
+The values in these configs should be loaded from the config files in the
+config/ directory. The tests in this file ensure that the values are loaded.
+It doesn't check all or most values, but serves as a sanity check that the
+configs are loaded correctly.
+"""
+
+
 import json
```
VIKTORVAV99, line 11: This is unused. Along with DeepDiff further down.

tests/test_config.py
```diff
@@ -10,128 +20,41 @@
...
-        )
-        self.assertIn("DK-DK1", zone_neighbours.keys())
-        dk_neighbours = zone_neighbours["DK-DK1"]
-
+    def test_zones_config(self):
+        # ZONES_CONFIG is a dict mapping zones to their config,
+        # which includes lots of keys with information about the zone.
+        self.assertIn("DK-DK1", config.ZONES_CONFIG.keys())
+        self.assertIn("bounding_box", config.ZONES_CONFIG["DK-DK1"].keys())
+        self.assertIn("capacity", config.ZONES_CONFIG["DK-DK1"].keys())
```
VIKTORVAV99, line 28: We can not use string literals here when the expected type is ZoneKey. The same goes for the functions further down as well.

tests/test_config_zones.py
```diff
@@ -0,0 +1,163 @@
...
+class TestConfigZones(unittest.TestCase):
+    def test_bounding_boxes_basic(self):
+        zones = {
+            "AD": {
+                "bounding_box": [[0.906, 41.928], [2.265, 43.149]],
+            },
+            "XX": {},
+        }
+        self.assertDictEqual(
+            zone_bounding_boxes(zones),
```
VIKTORVAV99, line 26: zones here need to be cast to the correct type if we want to avoid type errors. (Currently `dict[str, unknown]` and it needs to be `Dict[ZonKey, Any]`)

VIKTORVAV99 (commented):
I added a bunch of comments on the existing code but I do think it's generally more readable.

We need to take care to make sure these changes don't break anything in the backend though. I'll try and dig into that once I have more time.

tests/test_exchanges_json.py
qyearsley: Are you concerned about moving ZONE_KEYS, or using ZONES_CONFIG?

If the first: and nothing imports (and nothing _should_ import!) the test to use a global in a test module. If it's a constant used throughout one test case, it could be a class-level attribute in the test case class. It was only used in one test method, which is why I moved it here.

If the second: If we're concerned about tests (or other code elsewhere) potentially mutating ZONES_CONFIG, we should use `frozendict` when creating the config dicts.

electricitymap/contrib/config/__init__.py
qyearsley: That was an accident, I've added it back now.

electricitymap/contrib/config/co2eq_parameters.py
qyearsley: Good point, thanks

electricitymap/contrib/config/__init__.py
qyearsley: Thanks, done. Btw is there a tool that you use/recommend for automatically checking for unused imports?

electricitymap/contrib/config/zones.py
qyearsley: I have a couple questions about this:

- Should I cast by just changing `zone_1` to  `ZoneKey(zone_1)` or using `typing.cast` (https://docs.python.org/3/library/typing.html#typing.cast) and use `cast(zone_1, ZoneKey)`?
- Is there anything I should run to check types? (The tests pass with no warnings. In general, the Python3 interpreter ignores type hints. I've used `mypy` to check type hints before... and I see that `pydantic` is used in config/model.py, which might do type checking?)

tests/test_config.py
qyearsley: Done, fixed.

tests/test_config.py
qyearsley: I have a similar question as above -- should I use typing.cast or something else?

Is this actually checked anywhere? AFAIU ZoneKey is an alias for string, so what is meant exactly by "we can not use string literals here"? (The tests do pass with no warnings.)

electricitymap/contrib/config/zones.py
qyearsley: Done.

qyearsley:
Quick note, I replied to comments last week, and I believe that this PR may be ready for review again (let me know if that's not the case).

madsnedergaard:
It would be really great to get moving on this PR - when working on any project importing a single type from this file, the script takes 4s instead of 0.5s because of all the reading going on. We are often importing the `ZoneKey` type in internal projects :)

I'd be happy to help move the work here forward if you want me to take over, @VIKTORVAV99?

qyearsley:
I’d also be happy to split it up into separate PRs!

I’m wondering though, are you sure that the performance issue is due to
re-reading configs? (As-is, I believe that this PR shouldn’t change/improve
performance)

My open questions on this PR after first round of review were basically:
should I be running any extra automated tools to check Python3 type
annotations and unused imports?

VIKTORVAV99:
> It would be really great to get moving on this PR - when working on any project importing a single type from this file, the script takes 4s instead of 0.5s because of all the reading going on. And we are often importing the `ZoneKey` type in internal projects :)
> 
> I'd be happy to help move the work here forward if you want me to take over, @VIKTORVAV99?

I'll make sure to get this done tomorrow!

electricitymap/contrib/config/zones.py
VIKTORVAV99: We have no set typechecking in the CI yet and this is definitely something we should consider in the future. Personally I am using the Pylance VS code extension powered by https://github.com/microsoft/pyright

As for casting I think you can just do ZoneKey(zone_1).

electricitymap/contrib/config/config_types.py
VIKTORVAV99, line 1: @madsnedergaard we should update the backend to import the types from this file directly, that should bypass any unnecessary imports (or reading) if the type is the only thing that is used.

tests/test_config.py
VIKTORVAV99: More tests are always welcome, if you have the time feel free to do so.

tests/test_config.py
VIKTORVAV99: The ZoneKey is an extension of the string type, which means all ZoneKeys are strings but not all strings are ZoneKeys basically.

Right now it's basically the same thing though but type checking tools will fail on it.

tests/test_exchanges_json.py
VIKTORVAV99: Sounds good!

If you wish you can open a "RFC" issue about using frozendict for the config dicts. I don't think there should be an issue with that but best to discuss it first.

VIKTORVAV99 (commented):
Looking though it again and I think it looks good now, I updated the imports in all the other files to point to the new type path.

There is still two unused imports in the __init__.py file but we should verify nothing imports it from there before removing it.

@madsnedergaard @unitrium do you guys have any opinion on where we should put the new types. I am leaning towards putting it under contrib.libs instead of contrib.config and renaming it to just types but what do you guys think?

madsnedergaard:
Thanks both of you 🙏 
The issue I have is that this file is currently loading all .yaml files (1.5mb) at the root level - so importing a type means it runs all of the loading too. And with this PR we can load the types directly instead, so that's going to be helpful!

Leaving the type location question up to @unitrium :)

tests/test_config_zones.py
qyearsley: Added type casting in this file.

tests/test_exchanges_json.py
qyearsley: Thanks -- and indeed I think using frozendict could make sense here but it would be best to discuss first. Is a RFC issue just an issue (created with https://github.com/electricitymaps/electricitymaps-contrib/issues/new) with "RFC: ..." in the title?

electricitymap/contrib/config/reading.py
qyearsley: So, yaml.safe_load generally should return a dict with str keys, but is the issue here just that it's not guaranteed to? Should I add something like the following?

```
d = yaml.safe_load(open(defaults_path, encoding="utf-8"))
assert isinstance(d, dict)
return d
```

electricitymap/contrib/config/reading.py
qyearsley: This was resolved.

electricitymap/contrib/config/co2eq_parameters.py
qyearsley: Done.

electricitymap/contrib/config/__init__.py
qyearsley: Note, autoflake catches the unused imports; flake8 (which is used in current lint process) also should be able to catch it, but apparently doesn't catch it currently.

qyearsley:
> Thanks both of you 🙏 The issue I have is that this file is currently loading all .yaml files (1.5mb) at the root level - so importing a type means it runs all of the loading too. And with this PR we can load the types directly instead, so that's going to be helpful!
> 
> Leaving the type location question up to @unitrium :)

Thanks -- just to clarify so I understand the performance issue: re-loading configs is not an issue, but there are some cases where we just want the types and we don't need all of the configs. So by splitting the type definitions to a separate file and importing only types when only types are needed, we avoid having to load configs. (It's worth noting that after this PR, loading configs is still all-or-nothing basically, even if you just need a subset of config files you get all of them loaded).

With regard to type definition location, I also have no opinions.

From this review, it sounds like we might want to consider adding some more type validation in the linting process (probably pyright?) but that's something that can be done in the future. Also maybe tweak the way that flake8 is run (e.g. add check for unused imports https://www.flake8rules.com/rules/F401.html).

unitrium:
I was always thinking that it is confusing that the `ZoneKey` type is in config.
I'm thinking it would be better if at least the zoneKey type that is being used extensively everywhere could be in `electricityMap/lib`.

qyearsley:
> > I was always thinking that it is confusing that the `ZoneKey` type is in config. I'm thinking it would be better if at least the zoneKey type that is being used extensively everywhere could be in `electricityMap/lib`.
> 
> You want another lib folder directly under the electricitymap folder or can we put it in the electricitymaps/contrib/libs folder?
> 
> I'm thinking that maybe we want to keep as much as possible under the contrib/libs subfolder but I'm open to the idea of another libs folder as well.

contrib/libs SGTM - the latest commit for this PR moves ZoneKey to electricitymaps/contrib/libs/types.py and leaves BoudingBox in electricitymaps/contrib/config/ and adds comments.

unitrium (approved):
Yes this works for me!

PR #5280, opened 2023-04-11
Update granularity of FR realtime parser

lav-julien:
## Description

Change the granularity of the real time parser from 15 minutes to 30 minutes for consistency with the historical data parser (only available with 30min granularity)

Real-time dataset specs
image

Historical dataset specs
image

### Double check

### DIFF
--- a/parsers/FR.py
+++ b/parsers/FR.py
@@ -196,9 +196,6 @@ def fetch_consumption(
     datapoints = []
     for row in df_consumption.itertuples():
 
-        if arrow.get(row.datetime).datetime.minute not in [0, 30]:
-            continue
-
         datapoints.append(
             {
                 "zoneKey": zone_key,
--- a/parsers/FR.py
+++ b/parsers/FR.py
@@ -156,10 +156,6 @@ def fetch_production(
         if not any([is_not_nan_and_truthy(v) for k, v in production.items()]):
             continue
 
-        # only keep datapoints which minutes are 0 and 30 (ignore 15 and 45)
-        if arrow.get(row[1]["date_heure"]).datetime.minute not in [0, 30]:
-            continue
-
         datapoint = {
             "zoneKey": zone_key,
             "datetime": arrow.get(row[1]["date_heure"])
--- a/parsers/FR.py
+++ b/parsers/FR.py
@@ -182,7 +182,7 @@ def fetch_production(
 
         datapoint = {
             "zoneKey": zone_key,
-            "datetime": arrow.get(row[1]["datetime_30"])
+            "datetime": arrow.get(row[1]["date_heure"])
             .replace(tzinfo="Europe/Paris")
             .datetime,
             "production": production,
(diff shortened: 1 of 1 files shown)


parsers/FR.py
```diff
@@ -156,6 +156,10 @@ def fetch_production(
         if not any([is_not_nan_and_truthy(v) for k, v in production.items()]):
             continue
 
+        # only keep datapoints which minutes are 0 and 30 (ignore 15 and 45)
+        if arrow.get(row[1]["date_heure"]).datetime.minute not in [0,30]:
+            continue
```
mathilde-daugy, line 161:
```suggestion
        datapoint = {
            "zoneKey": zone_key,
            "datetime": arrow.get(row[1]["date_heure"])
            .replace(tzinfo="Europe/Paris")
            .datetime,
            "production": production,
            "storage": storage,
            "source": "opendata.reseaux-energies.fr",
        }
        datapoint = validate(datapoint, logger, required=["nuclear", "hydro", "gas"])
        datapoints.append(datapoint)
```

parsers/FR.py
```diff
@@ -156,6 +156,10 @@ def fetch_production(
...
+
```
VIKTORVAV99, line 162: This would cause our data to become inaccurate.

A better solution would be to collect all data point and average 0 and 15 and then average 30 and 45 into datapoints for 0 and 30 respectively. That would allow us to get accurate but averaged data.

I'm not sure why this is needed though? The backend should average all data points in an hour to a hourly value?

VIKTORVAV99 (changes requested):
The current implementation would mean we loose data.

parsers/FR.py
lav-julien: Right now when we refetch data, only half of the data is overwritten (the ones with minute 0 or minute 30) and the other half kept untouched. So we average the new data with old data

mathilde-daugy: this is needed because in the historical dataset we get a 1/2 hour granularity vs 1/4 hour for real-time data. This means that when we refetch we only correct 1/2 points and our dataset is not the same as the one published by RTE and we lose accuracy

VIKTORVAV99: Ah okay that makes sense, but in order to keep all the new data we need to average it into two data points from four before we return it otherwise if we just skip it we would lose data for 2 datapoints.

For example if there is a sudden spike in say gas production during a 15 minute period that start at 12:15 we would lose that additional gas consumption and would be getting inaccurate values.

If we instead do
```python
datapoint_1 = (datapoint_1 + datapoint_2) / 2
```
we gett all of the data but still only two datapoint for each hour.

lav-julien: See the new changes

lav-julien:
> 

It should be fine now!

VIKTORVAV99 (approved):
Things work great now!
Thank you both!

PR #5285, opened 2023-04-12
Missing aggregateButtons and other Italian labels

daiste:
## Issue

Translation for country-zone toggle (https://github.com/electricitymaps/electricitymaps-contrib/issues/4689) and other labels missing.

## Description

Added some translations that I found comparing with en.json, other than the ones mentioned here: https://github.com/electricitymaps/electricitymaps-contrib/issues/4689

## Preview
I tested the web app without mock server

Screenshot 2023-04-12 at 23 10 31
Screenshot 2023-04-12 at 23 17 49


web/public/locales/it.json
```diff
@@ -163,9 +163,15 @@
     "production": "Produzione",
     "consumption": "Consumo",
     "cpinfo": "Il <b>consumo</b> considera importazioni ed esportazioni, la <b>produzione</b> li ignora",
+    "aggregateInfo": "Visualizza dati per zona o aggregati per nazione",
```
VIKTORVAV99, line 166: I'm not a Italian speaker but does this match the same intention as:
"Toggle between a granular zone view and an aggregated country view"

It seems to miss the granular part?

daiste: It's not the literal translation of the English one, but I think it would sound nicer to a native speaker.
Literally translated would be something like "View data grouped by zone or grouped by country"
I think it's already clear, but if you wanna stress the granular part more, it could be changed to
"Visualizza dettagli per zona o dati aggregati per nazione"

If you want something closer to the English version, another option could be:
"Alterna tra una vista dettagliata per zona e una vista aggregata per nazione"

VIKTORVAV99: I think this will be fine. Let's get it merged!

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5287, opened 2023-04-13
enable automated deployment to point to prod

tonypls:
## Issue
Re-enable automated deployments

## Description
We've had automated deployments running successfully for a couple of weeks against staging, let's switch to production automated deployments that will be triggered after contrib is updated.


tonypls:
> Just to confirm this will only run once contrib is updated in the mono repo right?
> 
> (which means we could use staging for merges to master in here?)

Yes that's correct :D

madsnedergaard (approved):

PR #5288, opened 2023-04-13
Fix US-CAL-CISO battery values

VIKTORVAV99:
Looks like we forgot to invert the values for battery production once again 😅

We really need to change how this works in the future. Let me know when you guys want to take a crack at it and I'll update all the contrib parsers. We just need to co-ordinate it with a database migration and change to the backend.


mathilde-daugy:
oops 😅

mathilde-daugy (approved):

mathilde-daugy:
ok that makes a lot of sense. we should definitely discuss this in the future

PR #5291, opened 2023-04-14
US-CAL-CISO fix negative hydro

mathilde-daugy:
## Issue

image

## Description

update `fetch_prodcution` to collect hydro as production if the value is positive or as storage if the value is negative

### Double check

### DIFF
--- a/parsers/US_CA.py
+++ b/parsers/US_CA.py
@@ -123,7 +123,7 @@ def fetch_production(
         for mode in ["Small hydro", "Large Hydro"]:
             production_value = float(row[mode])
             if production_value < 0:
-                storage["hydro"] += production_value
+                storage["hydro"] += production_value * -1
             else:
                 production = add_production_to_dict(mode, production_value, production)
 
--- a/parsers/US_CA.py
+++ b/parsers/US_CA.py
@@ -55,6 +55,7 @@ def get_target_url(target_datetime: Optional[datetime], kind: str) -> str:
         target_url = f"{CAISO_PROXY}/outlook/SP/History/{target_datetime.strftime('%Y%m%d')}/{HISTORICAL_URL_MAPPING[kind]}.csv?host=https://www.caiso.com"
     return target_url
 
+
 def add_production_to_dict(mode: str, value: float, production_dict: dict) -> dict:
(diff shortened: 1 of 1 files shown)


parsers/US_CA.py
```diff
@@ -56,6 +56,15 @@ def get_target_url(target_datetime: Optional[datetime], kind: str) -> str:
     return target_url
 
 
+def add_production_to_dict(mode: str, value: float, production_dict: dict) -> dict:
+    """Add production to production_dict, if mode is in PRODUCTION_MODES."""
+    if PRODUCTION_MODES_MAPPING[mode] not in production_dict:
+        production_dict[PRODUCTION_MODES_MAPPING[mode]] = value
+    else:
+        production_dict[PRODUCTION_MODES_MAPPING[mode]] += value
+    return production_dict
```
VIKTORVAV99, line 65: I think we should consider making this a lib function as it would be usefull in other parsers as well.

parsers/US_CA.py
```diff
@@ -77,30 +86,48 @@ def fetch_production(
...
                 )
                 production_value = 0.0
 
-            production[mode] = production_value
+            production = add_production_to_dict(mode, production_value, production)
+
+        for mode in ["Small hydro", "Large Hydro"]:
+            production_value = float(row[mode])
+            if production_value < 0:
+                storage["hydro"] += production_value
```
VIKTORVAV99, line 126: don't this need to be multiplied by -1 as well like the battery storage?
```suggestion
                storage["hydro"] += production_value * -1
```

VIKTORVAV99 (commented):
Just some small comments othewise it looks good.

parsers/US_CA.py
mathilde-daugy: yes i was wondering the same thing but i think this makes sense

parsers/US_CA.py
mathilde-daugy: noted, i think i've already seen this logic in a parser as well

mathilde-daugy: will merge for now though so that i can trigger the data refetch

PR #5292, opened 2023-04-14
don't show wind sprite tails on iOS

tonypls:
## Issue
image

The wind layer on older iOS devices was not fading out after the wind animation

## Description
This PR adds the check for iOS devices, then uses an animation without a trail for iOS devices.

### Preview

https://user-images.githubusercontent.com/18622768/232050000-0825ba21-2011-4fdb-a4a6-6fc4d29f0eaf.mov


VIKTORVAV99 (approved):
Looks good, I suppose there is a underlying issue on iOS for why this is needed?

PR #5296, opened 2023-04-16
Update German nuclear capacity after closue of the last three reactors on April 15, 2023

sin:
Use data from IAEA PRIS (Power Reactor Information System): https://pris.iaea.org/PRIS/CountryStatistics/CountryDetails.aspx?current=DE

Before:
Emsland: 1335 MW
Isar-2: 1410 MW
Neckarwestheim-2: 1310 MW
Total: 4055 MW

After:
Total: 0 MW

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -261,7 +261,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
   - Thermal: [service-public.pf](https://www.service-public.pf/sde/documentation-2/rapports-delegataires)
 - Germany:
   - Nuclear: [IAEA](https://pris.iaea.org/PRIS/CountryStatistics/CountryDetails.aspx?current=DE)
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (changes requested):
Before we can merge this we also need to update the DATA_SOURCES.md file as well with the source for the new capacities.

sin:
I've added [PRIS](https://pris.iaea.org/PRIS/CountryStatistics/CountryDetails.aspx?current=DE) and [BASE](https://www.base.bund.de/EN/ns/nuclear-phase-out/nuclear-phase-out_node.html) as sources for Germany's nuclear capacity. Now there's more sources than nuclear power plants in Germany.

VIKTORVAV99 (approved):
As unfortunate as the closure is let's get this PR merged so we accurately represent the capacity state of the German grid!

PR #5301, opened 2023-04-17
mobile release 1.21.0

tonypls:
## Description

This PR contain is the app state for mobile release version 1.21.0 

This contains fixes for the wind and solar layers


madsnedergaard (approved):

madsnedergaard:
> I'm not sure how the bot currently handle mobile releases but would it make sense to add a tag and release for this as well in some way?

Ideally it would be the exact same thing that was released on web - in future we could include these version bumps in the "bump version" commit so it's all done at once though :)

PR #5302, opened 2023-04-17
Parser Validation: Create Event structure

unitrium:
## Issue

The datapoints pushed to the database lack standardization which can lead to rubish datapoints being pushed to the database. Also contributors can have a hard time figuring out all the required fields.

https://linear.app/electricitymaps/issue/ELE-1978/create-data-point-class

## Description

The goal of this PR is to start rolling out a standard data structure with minimal validation procedures on all parsers. For the moment the structure will allow backwards compatibility with the old dictionary structure until all parsers have been updated. This first PR just creates the structure and provides an example. 

We create a wrapper structure around the event creations to handle cases where the validation fails so that contributors don't have to worry about logging those errors themselves.

In the long run this structure should be also used for internally generated events like forecasts and estimates.

### Preview
The NI parser has been reworked to fully integrate those changes as a test.

### Notes:
The pydantic-pylint plugin has been added so that pylint can properly parse the added models.
This had the side effect that old validation functions started raising an issue. Those are false positives and have been suppressed. Anyway most of them would be replaced once all parsers have been refactored.


libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+    biomass: Optional[float]
+    coal: Optional[float]
+    gas: Optional[float]
+    geothermal: Optional[float]
+    hydro: Optional[float]
+    nuclear: Optional[float]
+    oil: Optional[float]
+    solar: Optional[float]
+    unknown: Optional[float]
+    wind: Optional[float]
```
VIKTORVAV99, line 31: Should these also accept int as a type or do we want all values to be floats?

libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+
+class StorageMix(Mix):
+    battery: Optional[float]
+    hydro: Optional[float]
+
+
+class Datapoint(BaseModel, ABC):
+    zoneKey: ZoneKey
+    datetime: datetime
+    source: dict
```
VIKTORVAV99, line 42: If source is the datasource should this not be a string or do we plan on changing the data structure to be able to return multiple sources?

libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+
+    @staticmethod
+    def create(
+        logger: Logger,
+        zone_key: ZoneKey,
+        datetime: datetime,
+        source: dict,
+        value: float,
+        forecasted: bool = False,
+    ) -> Optional["Exchange"]:
```
VIKTORVAV99, line 106: While it's not documented the exchange parsers can also return the capacity. Take a look at https://github.com/electricitymaps/electricitymaps-contrib/pull/5203 for an example.

libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+    @staticmethod
+    def create(
+        logger: Logger,
+        zone_key: ZoneKey,
+        datetime: datetime,
+        source: dict,
+        production: ProductionMix,
+        storage: Optional[StorageMix] = None,
+        forecasted: bool = False,
+    ) -> Optional["ProductionBreakdown"]:
```
VIKTORVAV99, line 180: Unless we are planning on changing things this also support returning dynamic capacities.

libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+        return {
+            "datetime": self.datetime,
+            "zoneKey": self.zoneKey,
+            "currency": self.currency,
+            "price": self.price,
+            "source": self.source,
+        }
+
+
+class DatapointBatch(ABC):
```
VIKTORVAV99, line 301: Could we rename this to something other than batch? I think maybe just naming them as list would make the most sense? Especially since we use the same syntax as list with append.

Or perhaps set as we expect these to be unique?

I think it's a good idea to leverage the already existing naming conventions for clarity, especially since we have a lot of open source contributors.

parsers/EIA.py
```diff
@@ -14,12 +14,13 @@
...
 from dateutil import parser, tz
-from requests import Session
-
 from parsers.ENTSOE import merge_production_outputs
 from parsers.lib.config import refetch_frequency
 from parsers.lib.utils import get_token
 from parsers.lib.validation import validate
+from requests import Session
+
+from electricitymap.contrib.libs.models.datapoints import ConsumptionBatch
```
VIKTORVAV99, line 23: The imports using symlinks will break on windows, I suggest either moving the whole libs folder into contrib or importing it directly from libs instead. 

While it should not be an issue with our backend it might cause unexpected problems for our contributors if they try and run it locally.

parsers/EIA.py
```diff
@@ -372,18 +373,24 @@ def fetch_consumption(
     session: Optional[Session] = None,
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
-):
+) -> List[Dict[str, Any]]:
```
VIKTORVAV99, line 376: I'm not all that familiar with python types but would it be possible to define the exact type here as a part of ConsumptionBatch so we can use ConsumptionBatch.type as the return type for example?

This type would be using a set or sets of [TypedDicts](https://peps.python.org/pep-0589/) to get type hinting on the dicts. It would be too much work to do it on a per parser level but since we can reuse these across all parsers I think it will be worth it.

VIKTORVAV99 (commented):
Great work!
I have added some initial thought as comment and hopefully you won't mind I did so while this is still a draft 😅.

libs/models/datapoints.py
unitrium: Do you have an example of a parser that does that so I could have a look?

parsers/EIA.py
unitrium: The description of the PR is not 100% complete, the goal in the long run would that parsers return those objects. So no need for using a typed dict here, as in the end it will be those objects that will be returned :)

libs/models/datapoints.py
VIKTORVAV99: The CA_AB.py parser returns capacity values like this:
```python
            "capacity": {
                "gas": generation["GAS"]["MC"],
                "hydro": generation["HYDRO"]["MC"],
                "battery storage": generation["ENERGY STORAGE"]["MC"],
                "solar": generation["SOLAR"]["MC"],
                "wind": generation["WIND"]["MC"],
                "biomass": generation["OTHER"]["MC"],
                "unknown": generation["DUAL FUEL"]["MC"],
                "coal": generation["COAL"]["MC"],
            },
            "datetime": get_csd_report_timestamp(response.text),
            "production": {
                "gas": generation["GAS"]["TNG"],
                "hydro": generation["HYDRO"]["TNG"],
                "solar": generation["SOLAR"]["TNG"],
                "wind": generation["WIND"]["TNG"],
                "biomass": generation["OTHER"]["TNG"],
                "unknown": generation["DUAL FUEL"]["TNG"],
                "coal": generation["COAL"]["TNG"],
            },
            "source": URL.netloc,
            "storage": {
                "battery": generation["ENERGY STORAGE"]["TNG"],
            },
            "zoneKey": zone_key,
        },
        ```
In the future I also hope we can have separate capacity parsers for production and exchange if the data is at another endpoint. Might be worth taking into account as well when shaping this?

File: parsers/EIA.py (hunk 7 above)
Comment (VIKTORVAV99) [reply parsers/EIA.py:376]: That's great!
No need for this then!

File: libs/models/datapoints.py (hunk 1 above)
Comment (unitrium) [reply libs/models/datapoints.py:31]: I will double check but I think if you supply an int, it just automatically gets cast to a float. Which is anyways an operation that we do afterwards because in the database they are stored as floats.

File: libs/models/datapoints.py (hunk 4 above)
Comment (unitrium) [reply libs/models/datapoints.py:180]: Thanks for pointing this out, I wasn't aware of this.  I'm trying to figure out if we act on this data in the backend or if the yaml config overrides this.
Anyways, I think we should keep this information in this parser for now and therefore add it to the data point structure. Though in the future it could probably become a data point of its own, with its own fetch function.

Comment (VIKTORVAV99) [reply libs/models/datapoints.py:180]: > Thanks for pointing this out, I wasn't aware of this. I'm trying to figure out if we act on this data in the backend or if the yaml config overrides this. Anyways, I think we should keep this information in this parser for now and therefore add it to the data point structure. Though in the future it could probably become a data point of its own, with its own fetch function.

I think we use it for the capacities in the frontend at least but I'm not 100% sure. I'm not aware of any other use case but they might be used in the estimations as well? If not that is something they could be used for in the future if the capacity is updated fairly regular (or change hourly such as for the exchanges).

File: libs/models/constants.py [hunk 8]
```diff
@@ -0,0 +1,171 @@
...
+    "UYU",
+    "UZS",
+    "VUV",
+    "VEF",
+    "VED",
+    "VND",
+    "YER",
+    "ZMW",
+    "ZWL",
+]
```
mathilde-daugy, line 171: are these the currencies for the zones we have or all available currencies?

libs/models/datapoints.py
mathilde-daugy: do we want to include the source as dict in the first instance of this project?

mathilde-daugy: I'm wondering if this might not fall outside of the scope for the first version of this project, where we want to make sure that data points loaded to `parser_table` are valid.

libs/models/datapoints.py
mathilde-daugy: Why is this part of the data point and not only in the config file like for production parsers?

mathilde-daugy: I feel like this adds a layer of complexity. Is there an exchange parser that actually has capacity in the data point?

libs/models/datapoints.py
mathilde-daugy: where does this capacity data go? I'm really confused that we have multiple ways of collecting capacity data

libs/models/datapoints.py
```diff
@@ -0,0 +1,400 @@
...
+        datetime: datetime,
+        source: dict,
+        value: float,
+        forecasted: bool = False,
+    ):
+        datapoint = Exchange.create(
+            self.logger, zone_key, datetime, source, value, forecasted
+        )
+        if datapoint:
+            self.datapoints.append(datapoint)
```
mathilde-daugy, line 334: should we log which data points were excluded or will this overload our logs?

libs/models/datapoints.py
VIKTORVAV99: It's for parsing live capacities such as when there is unavailability for example. I'm not sure why they are in the same datapoint though.

As for the exchange capacities there are no parsers that use that right now, but #5203 that is a proof of concept do use it and if we want live exchange capacities we will have to include it in the exchange parser until capacity have their own parser functions.

parsers/EIA.py
```diff
@@ -372,18 +373,24 @@ def fetch_consumption(
...
-        point["consumption"] = point.pop("value")
+        batch.append(
+            zone_key=point["zoneKey"],
+            datetime=point["datetime"],
+            source=point["source"],
+            consumption=point["value"],
+        )
 
-    return consumption
+    return batch.to_dict()
```
mathilde-daugy, line 393: as mentioned yesterday, this function should be renamed as we expect to return a list of dict and not a dict. It is a bit confusing

libs/models/datapoints.py
VIKTORVAV99: The capacity that is used should say something like this somewhere in the code (I forgot where):
```python
capacity = parsed_capacity | yaml_capacity
```
So if there is parsed capacity that is what is used and if there is not it falls back to the capacity defined in the config file.

libs/models/datapoints.py
mathilde-daugy: i'm not sure what a live capacity is. Because in the case of an exchange, the capacity is defined by the transmission line and should be immutable. has this got something to do with traded amounts?

VIKTORVAV99: The Maximum net transfer capacity is limited by the transmission lines themselves but there are other factors in the live capacities like bottlenecks, downed power lines, risk of overload, safety margins (like always having 1 power lines capacity in reserve in case of sudden failure to prevent a full blackout), even temperature plays a role. These all together make up the hourly capacity value for exchanges.

libs/models/datapoints.py
unitrium: I'm still unsure about this. Because the app always use the data available in the flowtraced data. So far I have not been able to find where in the flowtracing pipeline we would load the dynamic capacity instead of the static yaml. So I'm starting to think this data might not be used in the end.

libs/models/constants.py
unitrium: All available currencies in the world.

libs/models/datapoints.py
mathilde-daugy: I think that as long as this is not clearly resolved it falls outside the scope of this project. Also I'm not sure if the production capacities change on a hour to hour basis

libs/models/datapoints.py
unitrium: We are doing it, inside the create method of each element errors are logged. I think it is important to do and would not overload us once we add the improved logging that I started.

libs/models/datapoints.py
unitrium: I reverted it back to str for this PR, I put the improved source capabilities on another branch for a subsequent PR.

libs/models/datapoints.py
unitrium: Ok so I got insights on where the dynamic capacities are used. I would suggest that to keep this PR small we ignore this issue for now and address it in a later PR. The parsers that we are going to update are not using dynamic capacities for now.

parsers/EIA.py
unitrium: I removed the symlink and put it into contrib directly.

libs/models/datapoints.py
unitrium: Same as with production, we are going to address this in a subsequent PR to keep this one small. Anyways the parsers that we are going to update do not make use of this capability for now.

libs/models/datapoints.py
unitrium: Renamed to list

parsers/EIA.py
unitrium: Done

libs/models/datapoints.py
madsnedergaard: Would it make sense to add a comment here about this, to ensure anyone working on parsers before "v2" is implemented will know about this?

unitrium:
The PR should be good to be reviewed.

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+class StorageMix(BaseModel):
+    """
+    Containts the storage mix for a zone at a given time.
+    All values are in MW.
+    """
+
+    battery: Optional[float] = None
+    hydro: Optional[float] = None
```
VIKTORVAV99, line 47: This should also have a set_value function as it will most likely be used in tandem with ProductionMix if there is storage.

It will also make sure our "API" surface is the same between the two.

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+
+    @validator("datetime")
+    def _validate_datetime(cls, v, values: Dict[str, Any]):
+        if v.tzinfo is None:
+            raise ValueError(f"Missing timezone: {v}")
+        if v < LOWER_DATETIME_BOUND:
+            raise ValueError(f"Date is before 2000, this is not plausible: {v}")
+        if not values.get("forecasted", False) and v > datetime.now(
+            timezone.utc
+        ) + timedelta(days=1):
```
unitrium, line 81: Adding a delta of one day to add a margin of error. We have seen some flaky test in the past with such procedures.

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+            )
+        except ValueError as e:
+            logger.error(f"Error creating production breakdown Event {datetime}: {e}")
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "datetime": self.datetime,
+            "zoneKey": self.zoneKey,
+            "production": self.production.dict(exclude_none=True),
+            "storage": self.storage.dict(exclude_none=True) if self.storage else None,
```
VIKTORVAV99, line 234: Do we want to invert the storage values in here so we can use normalized values in the parsers themselves?

(Right now storage values are expected to be inverted (negative values are production))

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+class Consumption(Event):
+    consumption: float
+
+    @validator("consumption")
+    def _validate_consumption(cls, v: float):
+        if v < 0:
+            raise ValueError(f"Consumption cannot be negative: {v}")
+        if v > 500000:
```
VIKTORVAV99, line 246: Preferably it would be nice if we could specify this value per zone instead of having one global static value. Perhaps something for v2?

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+
+class Price(Event):
+    price: float
+    currency: str
+
+    @validator("currency")
+    def _validate_currency(cls, v: str):
+        if v not in VALID_CURRENCIES:
+            raise ValueError(f"Unknown currency: {v}")
+        return v
```
VIKTORVAV99, line 294: Right now this is a bit strict, I think we should allow lowercase matches as well and convert them to uppercase in here so it's consistent in the backend.

VIKTORVAV99 (commented):
Some additional comments and suggestions but I do think this is ready to be merged without them so implement them at your own discretion.

electricitymap/contrib/libs/models/datapoints.py
unitrium: Yes, this one is just a baseline check to make sure values are not crazy. A per zone layer will be added at a later stage, as part of the backend.

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+        datetime: datetime,
+        source: str,
+        value: float,
+        forecasted: bool = False,
+    ):
+        event = Exchange.create(
+            self.logger, zoneKey, datetime, source, value, forecasted
+        )
+        if event:
+            self.events.append(event)
```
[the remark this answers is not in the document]
unitrium: We have a part in the backend that checks for duplicates before inserting in the database. In a V2 we could move this check here and use it to remove duplicates. Because at the moment if a batch contains duplicates we raise an error and discard it.

electricitymap/contrib/libs/models/datapoints.py
```diff
@@ -0,0 +1,427 @@
...
+from datetime import datetime, timedelta, timezone
+from logging import Logger
+from typing import Any, Dict, List, Optional
+
+from pydantic import BaseModel, validator
+
+from electricitymap.contrib.config import EXCHANGES_CONFIG, ZONES_CONFIG, ZoneKey
+from electricitymap.contrib.libs.models.constants import VALID_CURRENCIES
+
+LOWER_DATETIME_BOUND = datetime(2000, 1, 1, tzinfo=timezone.utc)
```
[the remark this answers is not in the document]
unitrium: I think it's a constant that has no purpose beyond this file. So it would be better to keep it here.

tests/validators/test_sanity_checks.py
```diff
@@ -11,52 +9,54 @@
...
 
+from electricitymap.contrib.config import EXCHANGES_CONFIG
+
 from .lib.fixtures import load_fixture
 
 
 def test_validate_positive_production():
     events = load_fixture("production_negative_values")
-    res = validate_positive_production(events)
+    res = validate_positive_production(events) # type: ignore[too-many-function-args]
```
unitrium, line 19: Note those ignores are side effects of adding the pydantic pylint plugin. They are false positives. Since in the end those functions would become obsolete.

electricitymap/contrib/libs/models/events.py
```diff
@@ -0,0 +1,427 @@
...
+    An abstract class representing all types of electricity events that can occur in a zone.
+    forecasted: Whether the point is a forecasted point or not.
+    Should be set to True if the point is a forecast provided by a datasource.
+    zoneKey: The zone key of the zone the event is happening in.
+    datetime: The datetime of the event.
+    source: The source of the event.
+    We currently use the root url of the datasource. Ex: edf.fr
+    """
+
+    forecasted: bool = False
```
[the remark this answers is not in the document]
pierresegonne: or simply a type field that can take values from an enum "measured", "forecasted", "estimated"?

electricitymap/contrib/libs/models/events.py
```diff
@@ -0,0 +1,427 @@
...
+    gas: Optional[float] = None
+    geothermal: Optional[float] = None
+    hydro: Optional[float] = None
+    nuclear: Optional[float] = None
+    oil: Optional[float] = None
+    solar: Optional[float] = None
+    unknown: Optional[float] = None
+    wind: Optional[float] = None
+
+    def set_value(self, mode: str, value: float) -> None:
```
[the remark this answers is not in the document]
VIKTORVAV99: It would be nice to have a helper function for it if you have two different kind of say solar in the source data (Thermal and PV) and it would easily sum it up in a tested and documented function.

electricitymap/contrib/libs/models/events.py
```diff
@@ -0,0 +1,427 @@
...
+LOWER_DATETIME_BOUND = datetime(2000, 1, 1, tzinfo=timezone.utc)
+
+
+class ProductionMix(BaseModel):
+    """
+    Containts the production mix for a zone at a given time.
+    All values are in MW.
+    """
+
+    biomass: Optional[float] = None
```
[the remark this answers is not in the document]
unitrium: Those are attributes of a class, they should be available statically so that type checkers and other tools can easily pick up on them.

electricitymap/contrib/libs/models/events.py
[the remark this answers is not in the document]
unitrium: The reason we added this is if someone wants to set them through a loop.
```
for mode, value in production_modes:
   prod_mix.set_value(mode, value)
```
This can't be done directly through the attributes. So we should have a setter here.

electricitymap/contrib/libs/models/events.py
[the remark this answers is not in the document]
unitrium: We have ruled out making children classes for this because it seemed a bit overkilled, since there shouldn't be much difference between the two. Though it might come in handy since we have plans to enhance the source.
An enum could be a nice solution though are we sure that "measured", "forecasted", "estimated" will always be mutually exclusive ?
Could it be that at some point we make an estimate of a forecasted point from an external provider ?

electricitymap/contrib/libs/models/constants.py
```diff
@@ -0,0 +1,171 @@
+VALID_CURRENCIES = [
```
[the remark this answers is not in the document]
unitrium: Hum but internally we named our library, libs 😅

electricitymap/contrib/libs/models/events.py
```diff
@@ -0,0 +1,427 @@
...
+            raise ValueError(f"Not an exchange key: {v}")
+        zone_keys = v.split("->")
+        if zone_keys != sorted(zone_keys):
+            raise ValueError(f"Exchange key not sorted: {v}")
+        if v not in EXCHANGES_CONFIG:
+            raise ValueError(f"Unknown zone: {v}")
+        return v
+
+    @validator("value")
+    def _validate_value(cls, v: float):
```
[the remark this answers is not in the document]
unitrium: Added

electricitymap/contrib/libs/models/datapoints.py
unitrium: Since we are going to spend time updating every parser, I think it can be worth it to standardize and be strict about it.

electricitymap/contrib/libs/models/datapoints.py
unitrium: I think we should rather put a comment on what we intend to have in the storage object.
Like copying the wiki notes about this. Because it will depend on each parser if we have to invert or not. So I would say better to not introduce any logic behind this here.

electricitymap/contrib/libs/models/events.py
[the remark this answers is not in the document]
unitrium: Let's go for an enum.

unitrium:
@pierresegonne  I adressed all the changes we discussed yesterday, let me know if you have other comments or need any help in reviewing this.

electricitymap/contrib/lib/models/events.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def __init__(self, **data: Any):
+        """Overriding the constructor to check for negative values and set them to None."""
+        super().__init__(**data)
+        for attr, value in data.items():
+            if value is not None and value < 0:
+                self._corrected_negative_values.add(attr)
+                self.__setattr__(attr, None)
+
+    def set_value(self, mode: str, value: float) -> None:
```
pierresegonne, line 54: I think it might be better to do something like:

(giving a whole snippet to play around)

```
from typing import Optional

from pydantic import BaseModel, PrivateAttr


class ProductionMix(BaseModel):
    """
    Contains the production mix for a zone at a given time.
    All values should be positives, otherwise they will be set to None
    and a warning will be logged.
    All values are in MW.
    """

    # We use a private attribute to keep track of the modes that have been set to None.
    _corrected_negative_values: set = PrivateAttr(set())
    biomass: Optional[float] = None
    coal: Optional[float] = None
    gas: Optional[float] = None
    geothermal: Optional[float] = None
    hydro: Optional[float] = None
    nuclear: Optional[float] = None
    oil: Optional[float] = None
    solar: Optional[float] = None
    unknown: Optional[float] = None
    wind: Optional[float] = None

    def __setattr__(self, name, value):
        if name in ["biomass"]:
            if value < 0:
                self._corrected_negative_values.add(name)
                print(
                    f"Negative value for {name} production"
                    f"Setting to None."
                )
                value = None
            super().__setattr__(name, value)


if __name__ == "__main__":
    pt = ProductionMix()
    breakpoint()
```

This way there is no way for a contributor to unknowingly circumvet this logic by doing `mix.biomass = -1`, and it makes the setting less verbose

pierresegonne: (still need to also do the check at init)

pierresegonne (commented):
Overall small comments around how to handle setters for negative production 
Once they are adressed I'll approve

electricitymap/contrib/lib/tests/test_events.py
```diff
@@ -0,0 +1,297 @@
...
+            value=-1,
+            source="trust.me",
+        )
+        assert exchange.value == -1
+
+    def test_raises_if_invalid_exchange(self):
+        with self.assertRaises(ValueError):
+            Exchange(
+                zoneKey=ZoneKey("AT->DE"),
+                datetime=datetime(2023, 1, 1),
```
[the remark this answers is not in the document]
unitrium: Yes this one should fail because of missing tz.

electricitymap/contrib/lib/models/events.py
```diff
@@ -0,0 +1,357 @@
...
+    Contains the storage mix for a zone at a given time.
+    All values are in MW.
+    Values can be both positive (when storing energy) or negative (when the storage is discharged).
+    """
+
+    battery: Optional[float] = None
+    hydro: Optional[float] = None
+
+
+class EventObservation(str, Enum):
```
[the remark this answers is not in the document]
unitrium: Done

unitrium:
All done @pierresegonne

electricitymap/contrib/lib/models/events.py
pierresegonne: What about this @unitrium ?

unitrium: I oversaw this comment since it was made separately. I will take a look.

unitrium: We discussed this with @mathilde-daugy  and thought that it might be overwhelming for beginner contributors to use the __setattr__ since some might not know this default method.
However I agree that it introduces the risk of a contributor using it and by passing our verification.

unitrium: So I would rather settle for this.

unitrium: Added this and kept the set_mode for consistency with storage mix.

electricitymap/contrib/lib/models/events.py
```diff
@@ -0,0 +1,376 @@
...
+                self.__setattr__(attr, None)
+
+    def __setattr__(self, name: str, value) -> None:
+        if name in PRODUCTION_MODES:
+            if value is not None and value < 0:
+                self._corrected_negative_values.add(name)
+                return super().__setattr__(name, None)
+        return super().__setattr__(name, value)
+
+    def set_mode(self, mode: str, value: float) -> None:
```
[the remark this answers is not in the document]
unitrium: Yes actually my bad, since it is inherited there's no need to add this one.

pierresegonne (approved):

PR #5308, opened 2023-04-20
Update parser example

unitrium:
## Issue

A new event structure has been made available. Contributors should have access to examples on how to use it.

## Description

Update the parser-example to reflect new best practices.


VIKTORVAV99:
Could you update production.md as well? it should be in the same folder.

parsers/examples/example_parser.py
```diff
@@ -42,71 +51,12 @@ def fetch_production(
...
-        },
-        'source': 'mysource.com'
-      },
-      ...
-    ]
+    A  ProductionBreakdownList should be returned containing all ProductionBreakdown
+    events. Each ProductionBreakdown event should contain a datetime, a zoneKey,
+    a ProductionMix, a source and optionally a StorageMix.
+    The ProductionMix should contain the production breakdown for each fuel type.
+    The StorageMix should contain the storage breakdown for each storage type.
```
mathilde-daugy, line 58: simple language suggestion to be more directive

```suggestion
   A  ProductionBreakdownList contains all ProductionBreakdown
    events. 
    Each ProductionBreakdown event contains a datetime, a zoneKey,
    a ProductionMix, a source and optionally a StorageMix.
    The ProductionMix contains the production breakdown for each fuel type.
    The StorageMix contains the storage breakdown for each storage type.
```

parsers/examples/example_parser.py
```diff
@@ -126,44 +76,61 @@ def fetch_production(
...
     res: Response = session.get(url)
-    assert (
-        res.status_code == 200
-    ), f"Exception when fetching production for {zone_key}: error when calling url={url}"
+    if not res.status_code == 200:
+        raise ParserException(
+            "example_parser.py",
+            f"Exception when fetching production error code: {res.status_code}: {res.text}",
+            zone_key,
+        )
```
mathilde-daugy, line 84: do we not want the target_datetime in the error message to be able to track when the parser fails better?

VIKTORVAV99: Maybe it would be better to update the parser exception to also accept the target datetime?

Right now you have to include it as a text string in the message

mathilde-daugy: @unitrium you had a more long-term idea about this?

unitrium: I am working on an improved logger to have this.
It's outside of the scope of this PR.

mathilde-daugy (approved):

Issue #5140, opened 2023-02-28
Incorrect axis labels in emissions chart (production mode)


tonypls:
**Describe the bug**
When using the emissions chart in production mode the y axis changes to an incorrect scale

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the app
2. Select production mode
3. Open the country panel
4. Select carbon emissions
5. The carbon emissions chart will have incorrect labels for the Y axis

**Expected behavior**
The axis labels match the data

**Screenshots**
<img width="530" alt="image" src="https://user-images.githubusercontent.com/18622768/221870969-9a9658d2-d107-4189-8894-a50191c08190.png">

mrSeoD:
I could look into this!

VIKTORVAV99:
> I could look into this!

Sure thing! I'll assign the issue to you.

Thanks for taking a look at it!

mrSeoD:
My apologies for a delayed look into this, but I wanted to ask about another bug that I ran into as I was looking into this issue. I was able to find a place where I could scale the y-axis, but I noticed another bug:
![image](https://user-images.githubusercontent.com/64814632/233685117-78e601d0-2469-4c8c-8a47-f45ce9c17671.png)

As you can see from the screenshot above, I clicked South Central Sweden, which as 28g Intensity, but the panel on the left shows 696g. And this is the case for every single country. No matter what country I press, the carbon intensity on the left panel is always set to 696g and the graph that needs to be fixed for this issue has the same values as well. 

I checked the official website, and it doesn't have this problem. But this only happens within the mockserver. Do you have an idea why this may be happening?

VIKTORVAV99:
> My apologies for a delayed look into this, but I wanted to ask about another bug that I ran into as I was looking into this issue. I was able to find a place where I could scale the y-axis, but I noticed another bug:
> ![image](https://user-images.githubusercontent.com/64814632/233685117-78e601d0-2469-4c8c-8a47-f45ce9c17671.png)
> 
> As you can see from the screenshot above, I clicked South Central Sweden, which has a correct intensity of 28g, but the panel on the left shows 696g. And this is the case for every single country. No matter what country I press, the carbon intensity on the left panel is always set to 696g and the graph that needs to be fixed for this issue has the same values as well. 
> 
> I checked the official website, and it doesn't have this problem. But this only happens within the mockserver. Do you have an idea why this may be happening?

[...]

PR #5312, opened 2023-04-22
Production mode carbon emission graph's y-axis scale fixed

mrSeoD:
## Issue
Incorrect axis labels in emissions chart (production mode) #5140 

## Description
The goal of this PR is to fix the y-axis scale of Carbon Emission graph under the Production Mode. Previously, as represented in the issue link, the y-axis values was completely off and do not match with actual values of graph. In order to fix it, when calling the function "getTotalElectricity(zoneData, displayByEmissions)" in useEmissionChartData.ts file, I had to change the second parameter to use "displayByEmissionsAtom" instead of mixMode.

### Preview
image
As you can see in the picture, the y-axis is correctly scaled.

### Double check
- [N/A] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [DONE ] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

### DIFF
--- a/web/src/features/charts/hooks/useEmissionChartData.ts
+++ b/web/src/features/charts/hooks/useEmissionChartData.ts
@@ -2,14 +2,13 @@ import useGetZone from 'api/getZone';
 import { max as d3Max } from 'd3-array';
 import { scaleLinear } from 'd3-scale';
 import { useAtom } from 'jotai';
-import { Mode } from 'utils/constants';
-import { productionConsumptionAtom } from 'utils/state/atoms';
+import { displayByEmissionsAtom } from 'utils/state/atoms';
 import { getTotalElectricity, tonsPerHourToGramsPerMinute } from '../graphUtils';
 import { AreaGraphElement } from '../types';
 
 export function useEmissionChartData() {
   const { data, isLoading, isError } = useGetZone();
-  const [mixMode] = useAtom(productionConsumptionAtom);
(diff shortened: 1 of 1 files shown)


tonypls:
Hi @mrSeoD,

Thanks for the work, it looks like there is an existing bug in the "getTotalElectricity" function and it's not named very well either.


With the fix it shows identical carbon emissions in consumption and production mode.

I think we should have two functions here, psuedocode below:

```
export function getTotalElectricity(zoneData: ZoneDetail, isProductionMode: boolean) {
  if (totalProduction == null) {
    return Number.NaN;
  }
  return isProductionMode
    ? productionValue
    : productionValue + zoneData.totalDischarge + zoneData.totalImport - zoneData.totalExport;
} 


export function getTotalEmissions(zoneData: ZoneDetail, isProductionMode: boolean) {
  if (zoneData.totalCo2Production == null) {
    return Number.NaN;
  }

  return isProductionMode
    ? zoneData.totalCo2Production
    : zoneData.totalCo2Production + zoneData.totalCo2Discharge + zoneData.totalCo2Import - zoneData.totalCo2Export;
}

```

madsnedergaard:
Hey @mrSeoD, any chance you could make the changes suggested by @tonypls so we can merge this PR? :)

madsnedergaard:
This was fixed in https://github.com/electricitymaps/electricitymaps-contrib/pull/5771

PR #5323, opened 2023-04-24
Add consumption data for MX

unitrium:
## Issue

The consumption data for Mexico is available as a live data set. So we should start collecting data as soon as possible so that we can use it for future estimation models.

## Description

Adds a fetch consumption function in the mexico parser.
Split Baja California in Baja California and Baja California Sur as those two regions are not interconnected.

### Preview

Testing to be added

- [x]  Adding testing

- [x]  Redo geometries


parsers/MX.py
```diff
@@ -228,6 +264,44 @@ def fetch_exchange(
...
+    logger: Logger = getLogger(__name__),
+) -> list:
+    """Gets the consumption data for a region using the live dashboard."""
+    # TODO the calls could be improved since we can get all the data in one call.
+    if session is None:
+        session = Session()
+    if target_datetime is not None:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+    response = session.get(MX_EXCHANGE_URL)
+    if response.status_code != 200:
```
VIKTORVAV99, line 281: should we use response.ok here instead? 
```suggestion
    if not response.ok:
```

parsers/MX.py
```diff
@@ -228,6 +264,44 @@ def fetch_exchange(
...
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
+    """Gets the consumption data for a region using the live dashboard."""
+    # TODO the calls could be improved since we can get all the data in one call.
+    if session is None:
+        session = Session()
+    if target_datetime is not None:
+        raise NotImplementedError("This parser is not yet able to parse past dates")
+    response = session.get(MX_EXCHANGE_URL)
```
VIKTORVAV99, line 280:
```suggestion
    response: Response = session.get(MX_EXCHANGE_URL)
```

VIKTORVAV99 (commented):
Some initial comments and suggestions.

Just a note as well, I can't seem to access the consumption page locally. So please do double check that the parser server can access the page (it should be able to) otherwise we will have to set up a proxy first.

unitrium:
I have updated the map, thanks!
Regarding the access to the page, it shouldn't be an issue since the page is the same from which we get the exchanges at the moment.

config/zones/MX-CE.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -96.22812130395499
     - 21.325344143417453
 timezone: null
```
VIKTORVAV99, line 6: Should we add the timezone here?
```suggestion
timezone: America/Tijuana
```

config/zones/MX-NE.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -96.63927161399994
     - 28.293549303557484
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

config/zones/MX-NO.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -99.30696277711047
     - 32.277751364000025
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

config/zones/MX-NW.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -104.88862626838755
     - 32.99936330200008
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

config/zones/MX-OC.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -97.5001883613038
     - 25.66554433840048
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

config/zones/MX-OR.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -89.87985327199993
     - 22.98054026911808
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

config/zones/MX-PN.yaml
```diff
@@ -4,3 +4,5 @@ bounding_box:
   - - -86.20059160099993
     - 22.123439846000053
 timezone: null
```
VIKTORVAV99, line 6:
```suggestion
timezone: America/Tijuana
```

VIKTORVAV99 (approved):
LGTM! 🎉

Just have some suggestions to add timezones but I think everything looks good code wise. Note: I did not run this locally as my access to the source is blocked.

PR #5326, opened 2023-04-24
Update datetime IN-WE parser 

mathilde-daugy:
## Issue

datetimes in the parser have a weird 7 minute delay added because of pytz

## Description

- remove datetime conversions with pytz and use arrow instead 
- remove functions related to production because production is now collected using `IN.py`

### Preview

NA

### Double check

### DIFF
--- a/parsers/IN_WE.py
+++ b/parsers/IN_WE.py
@@ -8,43 +8,14 @@ from typing import Optional
 
 import arrow
 import pandas as pd
-import pytz
 from requests import Response, Session
 
 from parsers.lib.config import refetch_frequency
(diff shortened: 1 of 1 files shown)


wobniarin:
Hi! I think the test is failing because the `tests/test_parser_interface` has a module we are no longer using ) (IN-WE.fetch_production). Shall we remove this import? 

`FAILED tests/test_parser_interface.py::ParserInterfaceTestcase::test_unused_files - ImportError: Module "electricitymap.contrib.parsers.IN_WE" does not define a "fetch_production" attribute`

wobniarin (approved):
LGTM! :)

PR #5329, opened 2023-04-25
Fix storage cannot be a None

unitrium:
## Issue

Returning a None as a default storage is causing issues. We should return an empty dict instead.


unitrium:
@FelixDQ  this is what is causing the issue in NI. I wasn't aware that None as a storage was not possible.

FelixDQ (approved):
Great, thanks! Feel free to revert the temporary fix I did yesterday after we have merged this into the monorepo.

PR #5330, opened 2023-04-25
Use new header for ESIOS parser

unitrium:
## Issue
We don't have any data for the ES->MA exchange since the 21st of February. Turns out they made an API change now the header is a bit different.

## Description

Update to new header format.


VIKTORVAV99 (approved):
LGTM! 🎉

tonypls (approved):
LGTM

PR #5331, opened 2023-04-25
prevent google translate on buttons

tonypls:
## Issue
https://github.com/facebook/react/issues/11538

## Description
When using the chrome base google translate while using the wind or solar buttons on mobile the whole app can crash

This PR prevents google translate operating on the buttons

Fixes this sentry error:
NotFoundError
Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.

copilot:poem


madsnedergaard (approved):
Woah, crazy edge case - nicely found! :D

PR #5232, opened 2023-03-23
Create parser for new data source for Puerto Rico

q--:
Implements the new data source for Puerto Rico.

Closes #3999

***Note:*** I'm not 100% sure about the fuel mapping; for example, I found out just before committing that one of the plants (see the comment in the source code) that I found elsewhere was running on fuel oil has actually been dual-fuel for some years, also accepting natural gas as a fuel. (So perhaps this should be changed to 'unknown'?) But I don't have time to work on this anymore today and I'm not sure when I'll work on this again, so if anyone else wants to look into this, please do.

***Note 2:*** `unknown` used to be renewable in the old parser. It no longer is in the new parser. Some configuration might have to be changed to reflect this.

PR #5334, opened 2023-04-26
Updated parser for Puerto Rico (US-PR)

MuffinCompiler:
## Issue

The parser for Puerto Rico has been down for a while since the data endpoints were not maintained. However, there is a new(?) data dashboard with a new data source which holds more information than previous sources.

## Description

Fetching production data for Puerto Rico. The dashboard is accessible [here](https://aeepr.com/es-pr/Paginas/NewGen/dashboard.aspx) and contains real time information about all power plants in Puerto Rico. The actual data source is a JavaScript file which gets polled by the dashboard. Therefore, we can just request the [JS file](https://aeepr.com/es-pr/generacion/Documents/DataJS/dataSource.js) once to parse the data. I think these endpoints are only accessible from the US or PR, so a proxy/VPN might be required.

Screenshot of the dashboard:

brave_95pq8eZNp3

Pull request contains:
- Updated parser for US-PR
- Basic unit tests
- Mock file to go along with the unit tests

Some points to discuss:
- There is also a price endpoint. However, I am not sure how to handle these vastly different prices of the power plants.
- We could rename the `US_PREPA.py` parser to `US_PR.py` to use a region identifier instead of the one of the electricity authority.
- Although the source contains information about current production of each power plant, it is not always clear if it is currently fueled by oil/gas/diesel. Some plants are capable of both combusting oil and gas, maybe even simultaneously. I tried to work out how different plants contribute, but this seems to be far too complex (and not important for us). Therefore, only coal and renewables are parsed from the sites directly (to distinguish types of renewables); and the remainder (oil+diesel, gas) is infered by the `dataByFuel` ratio (top right panel on dashboard). So, the returned production data should be accounted correctly.

### Double check

### DIFF
--- a/parsers/US_PREPA.py
+++ b/parsers/US_PREPA.py
@@ -34,15 +34,15 @@ Two more endpoints, not needed here. Contain status of power plants, and 24h his
 /es-pr/generacion/Documents/DataJS/dataUnitStatus.js"
 """
 
-VAILDATION_THRESHOLD_PERCENT = 5.0
+VAILDATION_THRESHOLD_PERCENT = 2.0
 
 # Source data identifiers -> Production types. See also comments at oil/gas validation
 SRC_POWER_TYPE_TO_PRODUCTION_TYPE = {
     # fossil
     "Turbina de Gas": "gas",
-    "COGEN": "",  # see get_production_type_from_site()
-    "Vapor": "oil",
-    "Ciclo Combinado": "gas",
+    "COGEN": "gas",  # see get_production_type_from_site()
+    "Vapor": "combined",  # oil & diesel & gas, hybrid
+    "Ciclo Combinado": "combined",  # oil & diesel & gas, hybrid
     # renewables
     "Hidroelectricas": "hydro",
     "Wind": "wind",
@@ -64,9 +64,6 @@ def get_production_type_from_site(json_site):
             return "gas"  # https://www.eia.gov/state/analysis.php?sid=RQ
         elif json_site["Desc"] == "AES":
             return "coal"
-    elif json_site["Type"] == "Vapor":
-        if json_site["Desc"] == "Costa Sur":
-            return "gas"  # https://www.eia.gov/state/analysis.php?sid=RQ
 
     if json_site["Type"] in SRC_POWER_TYPE_TO_PRODUCTION_TYPE:
         return SRC_POWER_TYPE_TO_PRODUCTION_TYPE[json_site["Type"]]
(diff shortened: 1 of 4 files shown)


MuffinCompiler:
> Before I do a review I would just like to let you know if there is another PR that adds a new parser for Puerto Rico open at #5232.
> 
> Now that is feature complete and just need some details to be ironed out.
> 
> I do plan on comparing the two and see which returns the most data and which seems to be more accurate before making any decisions but right now I'm leaning towards the parser in #5232 as it's already working (and this is not outside the us, and need to be updated to use the already existing proxy).

Thanks, shame on me, I totally missed that other pull request. 
However, the data source used in this pull request also contains the precise ratio of oil and gas as combustion fuel. A data issue that is quite complicated in PR (see code comments in both versions), since power plants are currently transforming from oil or diesel to gas and some are capable of combusting both fuel types.

parsers/US_PREPA.py
```diff
@@ -14,59 +14,227 @@
...
-    sourceData = dataSource[
-        "data"
-    ]  # The rest of the dataSource object contains unnecessary stuff like chart theme, label, axis names etc.
-    return sourceData
+VAILDATION_THRESHOLD_PERCENT = 2.0
+
+# Source data identifiers -> Production types. See also comments at oil/gas validation
+SRC_POWER_TYPE_TO_PRODUCTION_TYPE = {
+    # fossil
+    "Turbina de Gas": "gas",
```
q--, line 42: As I pointed out in my parser, in spite of the name, there are gas turbines that run on diesel; 'gas' doesn't mean 'natural gas' here

q--: But if you've found a data source that indicates all Puerto Rican gas turbines use natural gas, I'm interested to know

MuffinCompiler: Nope, I also can't provide a specific source for this specific information. I also found this the hard way first by assuming first they are based on gas, indeed. That's why I labeled them first as "gas" there. 

However, [later](https://github.com/MuffinCompiler/electricitymaps-contrib-pr/blob/4b27e41cbb6c96aec554975ea08903b84e58dc23/parsers/US_PREPA.py#L332-L359), the actual amounts of oil/gas are infered from the ratio source table in the data source, and not accumulated over each generation site. Therefore, the returned production values for 'gas' and 'oil' should be most accurate. Probably I should have set the type as "combined" instead of "gas" to express this.

MuffinCompiler:
Sorry, didn't know renaming a branch closes the pull request automatically... quickly restored

parsers/US_PREPA.py
MuffinCompiler: I just simplified the parser to infer coal, gas, and oil only from the % table. Renewables are parsed more specifically by type.

0100010001010011:
verify the new website https://genera-pr.com/data-generacion

pierresegonne:
any updates on this? Otherwise we might close it due to staleness

consideRatio:
The fetched .js data endpoint in this PR is no longer available it seems, I get:

```
The resource you are looking for has been removed, had its name changed, or is temporarily unavailable.
```

It seems that the https://aeepr.com/ website now only reports renewable electricity sources.

It seems like https://github.com/electricitymaps/electricitymaps-contrib/pull/5232#issuecomment-2498281610 has an actionable path forward, but is a stale PR as well but where anyone is welcome to continue the work.

PaulRoms:
I will close this Pr for staleness.
feel free to re-open a PR if necessary! 

Thank you again for your contribution!

PR #5335, opened 2023-04-27
Add consumption for CA BC

unitrium:
## Description

Start collecting consumption data for CA BC from BC hydro.


parsers/CA_BC.py
```diff
@@ -1,17 +1,63 @@
...
+    # Due to DST there might be some missing hours, we use shift_backward to fill them with the previous hour.
+    df["datetime"] = pd.to_datetime(
+        df["Date"] + " " + (df["HE"] - 1).astype("str"), format="%m/%d/%Y %H"
+    ).dt.tz_localize(TIMEZONE, nonexistent="shift_backward")
+    selected_times = df[df.datetime.dt.date == target_datetime.date()]
+    consumption_list = TotalConsumptionList(logger)
+    for row in selected_times.iterrows():
+        consumption_list.append(
+            zoneKey=zone_key,
+            datetime=row[1]["datetime"],
```
VIKTORVAV99, line 56: This need to have a native datetime and have a timezone, currently we are logging this error:
```
AssertionError: Datetimes must be returned as native datetime.datetime objects
```

VIKTORVAV99 (commented):
We need to change the datetimes to a native datetime.datetime object but besides that everything is working as expected as far as I can tell.

parsers/CA_BC.py
unitrium: Fixed this issue, indeed it was recorded as a timestamp instead of a datetime.

parsers/CA_BC.py
```diff
@@ -1,17 +1,63 @@
...
+@refetch_frequency(timedelta(days=1))
+def fetch_consumption(
+    zone_key: ZoneKey,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+):
+    r = session or Session()
+    if target_datetime is None:
+        target_datetime = datetime.now(tz=TIMEZONE)
```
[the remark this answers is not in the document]
unitrium: I think it would be best to return the latest month when fetching without a target datetime because at the moment they seem to publish the data for the last month.

parsers/CA_BC.py
```diff
@@ -1,17 +1,71 @@
...
+def fetch_consumption(
+    zone_key: ZoneKey,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+):
+    r = session or Session()
+    consumption_list = TotalConsumptionList(logger)
+    if target_datetime is None:
+        start_date = datetime.now(tz=TIMEZONE) - PUBLICATION_DELAY
```
VIKTORVAV99, line 36: Just to be sure we get overlapping data.
```suggestion
        start_date = datetime.now(tz=TIMEZONE) - PUBLICATION_DELAY - timedelta(days=1)
```

parsers/CA_BC.py
```diff
@@ -1,17 +1,71 @@
...
 import arrow
-from requests import Session
+import pandas as pd
+import pytz
+from requests import Response, Session
+
+from electricitymap.contrib.config import ZONES_CONFIG, ZoneKey
+from electricitymap.contrib.lib.models.event_lists import TotalConsumptionList
+from electricitymap.contrib.parsers.lib.config import refetch_frequency
+from electricitymap.contrib.parsers.lib.exceptions import ParserException
```
VIKTORVAV99, line 16:
```suggestion
from parsers.lib.config import refetch_frequency
from parsers.lib.exceptions import ParserException
```
For windows again, I think we should try and move these from parsers lib to contrib lib now that we have the other models and such as well in there.

VIKTORVAV99 (changes requested):
Final comments, looks good though.

VIKTORVAV99 (approved):

VIKTORVAV99:
Looks good, merge whenever you want!

PR #5336, opened 2023-04-27
Allow historical refetch of ESIOS

unitrium:
## Issue

We have a data gap for ES->MA from the 21st of february to the 25th of April. The parser was not able to fetch data for past datetimes while the data source supports it.

## Description

- Allow refetch of specific datatimes
- use the new exchange structure
- be more precise on timezones.


pierresegonne (approved):
goodonya

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5343, opened 2023-04-30
Fixing bugs in IN-MH parser where OCR was not able to read two fields from the image 

saurabhchatterjee23:
## Issue

Fixing bugs in IN-MH parser where `pytesseract` was not able to read two fields https://github.com/electricitymaps/electricitymaps-contrib/issues/4539

## Description
The IN-MH parser was failing for 3 different issues.

### Errors

I was able to reproduce 3 issues locally as shown in these logs [IN-MH logs](https://storage.googleapis.com/electricitymap-parser-logs/IN-MH.html)

#### Error 1
This error occurs when the coordinates of a field are misaligned and the OCR reads some images with line from the text box as a dash
```
could not convert string to float: '-'

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 172, in launch_parsers
    objs = res.get(timeout) or []

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 164, in <lambda>
    lambda: parser(

  File "/contrib/parsers/IN_MH.py", line 161, in fetch_production
    values[key] = max([float(text[count]), 0])          
```

When I was debugging the issue for this parser, I found out that some fields were read as '-'  from the image by `pytesseract` because it read the boundary of a text field as a dash. On digging further I also found that the code is supposed to read 23 distinct fields but it was able to read only 21 values.  For some reason, the coordinates of 3 fields were not correctly aligned which led to one of the fields being read as a dash and another as a dot.
##### Fix
Corrected the coordinates for SOLAR TTL, HYDRO, and PIONEER

#### Error 2
This error occurs when the URL returns an empty image for some reason which is read as a dot by the OCR
```
could not convert string to float: '.'

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 172, in launch_parsers
    objs = res.get(timeout) or []

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 164, in <lambda>
    lambda: parser(

  File "/contrib/parsers/IN_MH.py", line 161, in fetch_production
    values[key] = max([float(text[count]), 0])          
```
##### Fix
I have added verification to check if the parser got an image of a size greater than zero before reading text from the fields.

#### Error 3
This error occurs when the OCR is able to read less number of fields than the number of text fields it must have read,
because concatenating two numbers without space makes them read as one big number

```
list index out of range

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 172, in launch_parsers
    objs = res.get(timeout) or []

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 164, in <lambda>
    lambda: parser(

  File "/contrib/parsers/IN_MH.py", line 161, in fetch_production
    values[key] = max([float(text[count]), 0])        
```
When concatenating all the small images into one, for the PIONEER field the digits of the value got concatenated without any space, for example, image 123 joined 53 became 12353 without any space in between and was read as a single value instead of 2 separate values which lead to reading one less value and caused list index out of range

##### Fix
Changing the algorithm to not concatenate fields into one before converting the test to integer, instead each field separately to text and log an error for fields it is not able to read. Basically, make the best effort to read each field.

### Making the fix future-proof and fault tolerant to such misalignment of coordinates of the values

In the future, I am assuming that the coordinates of the boxes may go wrong, maybe because the website will change the layout of the image, that's an assumption.
This change in the algorithm will read fields as much as possible and skip those that it will not be able to read for some reason and log an error. It will not misinterpret any legit value, but instead, skip over fields `pytesseract` is not able to interpret

### Testing
- [X] Tested locally for the case where if the OCR is not able to read any field, it will not break the parser and successfully read all the other fields. It will leave an error log for the failed fields, but not fail open the parser
-  [X] The coordinates for the bonding box for SOLAR TTL and HYDRO are correct now, the OCR is able to read them properly, tested it by showing the image on my desktop for the field that the OCR was unable to read.

### Double check

### TODO

- [ ] Unit tests for IN-MH are missing, it is super important to have them because reading an image is complex and can go wrong in many ways, or when the website changes some format at their end


parsers/IN_MH.py
```diff
@@ -147,18 +148,28 @@ def fetch_production(
...
+            val = float(digit_text)
+        except ValueError:
+            # If the image cannot be converted to a valid number, log an error but do not break the parser
+            val = 0
+            logger.error(
+                "Error reading value for key %s, value read %s", key, digit_text
+            )
+        values[key] = val
+
+    logger.debug("values %s", values)
```
VIKTORVAV99, line 172: I'm not really sure why these changes are necessary?
Maybe you could explain why you made them?

saurabhchatterjee23: Added more detailed explanation above, let me know if it is not clear

parsers/IN_MH.py
```diff
@@ -145,20 +148,37 @@ def fetch_production(
...
+        digit_text = pytesseract.image_to_string(
+            imgs[index], lang="digits_comma", config="--psm 7"
+        )
+        val = 0
+        try:
+            val = float(digit_text)
+        except ValueError:
+            # If the image cannot be converted to a valid number, log an error but do not break the parser
+            val = 0
+            logger.error(
```
VIKTORVAV99, line 176: > Making the fix future-proof and fault tolerant to such misalignment of coordinates of the values
In the future, I am assuming that the coordinates of the boxes may go wrong, maybe because the website will change the layout of the image, that's an assumption.
This change in the algorithm will read fields as much as possible and skip those that it will not be able to read for some reason and log an error. It will not misinterpret any legit value, but instead, skip over fields pytesseract is not able to interpret.

While this would make the parser more fault tolerant it might also have a negative effect of returning partial data. Just making a note of it so others are aware.

VIKTORVAV99 (approved):
The changes in are pretty straight forward (after the added explanation) so I'll go ahead and approve and merge them.

PR #5346, opened 2023-05-01
Upgrade EIA parser to new data structure for exchanges

unitrium:
## Issue

The new data structure is now the favored method of formating data points.

## Description

Upgrades the EIA exchange parser to use the new structure and create an exchange unit test.

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -14,10 +14,6 @@ from typing import Any, Dict, List, Optional
 
 import arrow
 from dateutil import parser, tz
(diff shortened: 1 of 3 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

Do we want to proceed with upgrading other parsers as well?
Specifically the big multi zone parsers such as ENTSO-E and ES parsers?

unitrium:
@mathilde-daugy is working on updating the ENTSO-E parser, I think ES will be for a next round of updates :)

PR #5349, opened 2023-05-02
Fix IN parser + parser validation

mathilde-daugy:
## Issue

For 2021-08-16, the parser was not able to collect data from NPP because the format of the data source changed.
image
vs 
image

## Description

fetch_npp_production was updated so that the input df is not longer filtered by specific str
parser validation was implemented for fetch_production and fetch_consumption

### Preview
before: 
image

after: 
image

same problem is observed in IN-NO
image

### Double check

### DIFF
--- a/parsers/IN.py
+++ b/parsers/IN.py
@@ -21,7 +21,6 @@ from electricitymap.contrib.lib.models.event_lists import (
 )
 from electricitymap.contrib.lib.models.events import ProductionMix
 from parsers.lib.exceptions import ParserException
-from parsers.lib.validation import validate_consumption
 
 IN_TZ = "Asia/Kolkata"
 START_DATE_RENEWABLE_DATA = arrow.get("2020-12-17", tzinfo=IN_TZ).datetime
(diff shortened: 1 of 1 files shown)


parsers/IN.py
```diff
@@ -247,31 +254,23 @@ def fetch_npp_production(
...
-                production_in_zone.iloc[
-                    production_in_zone.index.get_indexer_for([mode])[0]
-                ].get("value")
-                / CONVERSION_GWH_MW,
-                3,
-            )
-        return production
+        df_npp = df_npp[["region", "production_mode", "value"]]
+
+        df_npp_filtered = df_npp.loc[~df_npp["production_mode"].isna()].copy()
```
pierresegonne, line 264: I am not sure to understand why this fixes the issue we've seen before

mathilde-daugy: image
When there is value in this col it corresponds to the total for the given mode for all the PPs in the sector in the state. This way we don't have to filter on "TYPE:" which may or may not be in the column text

pierresegonne (approved):

PR #5350, opened 2023-05-02
Add static method to merge production outputs

unitrium:
## Issue

Some production mix  parsers need to query several endpoints and merge the events together based on datetime.

## Description

Add a static method in ProductionBreakdown list to handle this case.

### DIFF
--- a/electricitymap/contrib/lib/models/event_lists.py
+++ b/electricitymap/contrib/lib/models/event_lists.py
@@ -3,6 +3,9 @@ from datetime import datetime
 from logging import Logger
 from typing import Any, Dict, List, Optional
 
+import numpy as np
+import pandas as pd
+
 from electricitymap.contrib.config import ZoneKey
 from electricitymap.contrib.lib.models.events import (
     Event,
@@ -38,6 +41,8 @@ class EventList(ABC):
 
 
 class ExchangeList(EventList):
+    events: List[Exchange]
+
     def append(
         self,
         zoneKey: ZoneKey,
@@ -54,6 +59,8 @@ class ExchangeList(EventList):
 
 
 class ProductionBreakdownList(EventList):
+    events: List[ProductionBreakdown]
+
     def append(
         self,
         zoneKey: ZoneKey,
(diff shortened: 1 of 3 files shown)


unitrium:
@VIKTORVAV99 I think this piece might also be useful for the AR parser we discussed a while ago.

VIKTORVAV99:
> @VIKTORVAV99 I think this piece might also be useful for the AR parser we discussed a while ago.

Yeah it should help things for sure, if I get the time I'll try and update the AR parser this week and test things out, we might finally be able to merge it.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -69,8 +76,50 @@ def append(
         if event:
             self.events.append(event)
 
+    @staticmethod
+    def merge_production_breakdowns(
+        ungrouped_production_breakdowns: List["ProductionBreakdownList"],
+        logger: Logger,
+        merge_zone_key: ZoneKey,
+        merge_source: str,
```
VIKTORVAV99, line 84: Instead of specifying the new source here maybe we should sort the existing soures and format them into the following format if they are diffrent and deduplicate them if not:
`source1, source2` I think that is already in use by some parsers and would make the experience consistant and would also allow us to change to a list later on when all the parsers use the new class in a easy manner. (We have talked about trying to make this a list or array for the frontend as well.)

Example from the SG parser:
```python
    return {
        "datetime": sg_data_to_datetime(data),
        "zoneKey": zone_key,
        "production": generation_by_type,
        "storage": {},  # there is no known electricity storage in Singapore
        "source": "emcsg.com, ema.gov.sg",
    }
```

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -69,8 +76,50 @@ def append(
         if event:
             self.events.append(event)
 
+    @staticmethod
+    def merge_production_breakdowns(
+        ungrouped_production_breakdowns: List["ProductionBreakdownList"],
+        logger: Logger,
+        merge_zone_key: ZoneKey,
```
VIKTORVAV99, line 83: The zone key should also always be the same between the imputs to prevent accidental merges between diffrent zones. Maybe we can just compare thes and set it in the method instead of defining it here?

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -69,8 +76,50 @@ def append(
...
+        df = df.groupby(level=0, dropna=False).sum(numeric_only=True, min_count=1)
+        result = ProductionBreakdownList(logger)
+        for row in df.iterrows():
+            production_mix = ProductionMix()
+            storage_mix = StorageMix()
+            for key, value in row[1].items():
+                if np.isnan(value):
+                    value = None
+                # The key is in the form of "production.<mode>" or "storage.<mode>"
+                prefix, mode = key.split(".")
```
VIKTORVAV99, line 105: key does not hava a defined type so the type checker is trhowing an error here that split does not exist on the type hashable.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -69,8 +76,50 @@ def append(
...
+                if prefix == "production":
+                    production_mix.set_value(mode, value)
+                elif prefix == "storage":
+                    storage_mix.set_value(mode, value)
+            result.append(
+                merge_zone_key,
+                row[0].to_pydatetime(),
```
VIKTORVAV99, line 112: Same here.

VIKTORVAV99 (commented):
Nice work!
I have some concerns about 2 things really, the source and the zone key.

I think it would also be nice to get a version of this for the production forecast (might not be needed yet) and one for the exchanges (can be used by at least 1 parser right away).

I don't think it's needed for the rest.

electricitymap/contrib/lib/models/event_lists.py
unitrium: Agree this might be better, as I saw some parsers do that like the India parser. Initially I thought, better to have the user in control of the final source to avoid any surprises but this might be better.

electricitymap/contrib/lib/models/event_lists.py
unitrium: Agree there should be at least some safe guards

unitrium:
Regarding forecasts, they are directly included because a forecast is the same structure. It just needs a little tweak to preserve it.
Agree that exchanges could also be useful, I will add it in a subsequent PR.

electricitymap/contrib/lib/models/event_lists.py
unitrium: I don't think there's an easy way out of this at the moment so I just added a type ignore.

mathilde-daugy (approved):

PR #5351, opened 2023-05-02
Update inconsistent lock file for UY

unitrium:
## Issue

Seems like the UY parser is still down. Looks like it has an inconsistency in the lock file which prevents it from installing opfcy.

## Description

Update the lock file to fix the issue.

Tested locally with a clean setup.


unitrium:
image
From the logs

VIKTORVAV99 (approved):

PR #5357, opened 2023-05-04
Fixing exchange and consumption parser for IN-WE

saurabhchatterjee23:
## Issue
Following parsers are failing and I can see exceptions in the logs
- Exchange
  - [IN-NO->IN-WE](https://storage.googleapis.com/electricitymap-parser-logs/IN-NO.html)
  - [IN-SO->IN-WE](https://storage.googleapis.com/electricitymap-parser-logs/IN-SO.html)
  - [IN-EA->IN-WE](https://storage.googleapis.com/electricitymap-parser-logs/IN-EA.html)
- Consumption
  - [IN-WE](https://storage.googleapis.com/electricitymap-parser-logs/IN-WE.html)

### Issue 1
```
'Region_Name'

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 172, in launch_parsers
    objs = res.get(timeout) or []

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value

  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker

  File "/projects/feeder-electricity/src/lib/fetch_data.py", line 164, in <lambda>
    lambda: parser(

  File "/contrib/parsers/lib/config.py", line 20, in wrapped_f
    result = f(*args, **kwargs)

  File "/contrib/parsers/IN_WE.py", line 192, in fetch_exchange
    exchanges = [

  File "/contrib/parsers/IN_WE.py", line 193, in <listcomp>
    format_exchanges_data(

  File "/contrib/parsers/IN_WE.py", line 106, in format_exchanges_data
    filtered_data["zone_key"] = filtered_data["Region_Name"].map(EXCHANGES_MAPPING)

  File "/usr/local/lib/python3.8/site-packages/pandas/core/frame.py", line 3807, in __getitem__
    indexer = self.columns.get_loc(key)

  File "/usr/local/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 3804, in get_loc
    raise KeyError(key) from err  
```

### Issue 2
```
invalid point: {'netFlow': 972.283, 'sortedZoneKeys': 'IN-SO->IN-WE', 'datetime': datetime.datetime(2023, 5, 4, 17, 0, tzinfo=tzfile('/usr/share/zoneinfo/Asia/Kolkata')), 'source': 'wrldc.in', 'schemaVersion': 3}, reason:Data from IN-SO->IN-WE can't be in the future, data was 2023-05-04T17:00:00+05:30, now is 2023-05-04T06:03:44.983187+00:00 
```
## Description
The source provides data for a given date from the beginning of the day (12 am IST) to the current timestamp
- It is possible for the source to miss data for some hours because of some issue with their system.
- The timestamp from the source is 12 hr format with missing 'AM/PM', which needs us to sanitize the data before we use it.
- The exchange and consumption parsers have a bug that tries to find data for an hour in the future or for an hour when there is no data from the source and fails with an exception. 

## Primary changes
- I rewrote the entire parser using SQL which makes the code much more readable.
  - I added a new package `pandasql` which lets me write SQL on JSON data, it is very neat :)
- The new parser fetches data from another API which provides the same data with timestamp in 24hr format which is more accurate to be consumed
- The new parser generates data points only for hours whose data is available, if doesn't fail if there is data missing for some hours which does happen from time to time for some reason from this data source
- The parser doesn't generate datapoint in future

- [ ] Write unit tests

### Double check

  - [X] `poetry run test_parser "IN-NO->IN-WE" exchange`
  - [X] `poetry run test_parser "IN-SO->IN-WE" exchange`
  - [X] `poetry run test_parser "IN-EO->IN-WE" exchange`
  - [X] `poetry run test_parser IN-WE consumption` 
- [X] poetry run test


parsers/IN_WE_new.py
```diff
@@ -0,0 +1,257 @@
...
+    "Region_Name": "WR-ER",
+    "Export_Ttc": 25000.0,
+    "Import_Ttc": 25000.0,
+    "Long_Term": 955.0,
+    "Short_Term": -125.0,
+    "Px_Import": 35.0,
+    "Px_Export": 0.0,
+    "Total": 322.0,
+    "Current_Loading": -1052.0,
+    "lastUpdate": "2023-04-05 12:24:09 PM"
```
VIKTORVAV99, line 41: Could you just give a quick explination to what the diffrent fields here are? Are any of them the "realtime" capacity?

parsers/IN_WE_new.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_exchange(
+    zone_key1: str,
+    zone_key2: str,
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 92:
```suggestion
) -> List[dict]
```

parsers/IN_WE_new.py
```diff
@@ -0,0 +1,257 @@
...
+CONSUMPTION_DATETIME_COLUMN_NAME = "current_datetime"
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_consumption(
+    zone_key: str = "IN-WE",
+    session: Optional[Session] = None,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
+) -> dict:
```
VIKTORVAV99, line 202:
```suggestion
) -> List[dict]:
```

validators/zone_specific_checks.py
```diff
@@ -34,8 +34,9 @@ def validate_production_has_fossil_fuel(events: pd.DataFrame) -> pd.Series:
         "production.oil",
         "production.gas",
     ]
-    fossil_fuel_cols_in_event = set.intersection(
-        set(fossil_fuel_cols), set(events.columns)
+
+    fossil_fuel_cols_in_event = list(
+        set.intersection(set(fossil_fuel_cols), set(events.columns))
     )
```
VIKTORVAV99, line 40: Why the change here?

poetry.lock
```diff
@@ -655,39 +729,37 @@ files = [
 
 [[package]]
 name = "pandas"
-version = "1.5.3"
+version = "2.0.1"
```
VIKTORVAV99, line 732: Note this PR will be blocked untill we can verify there are no breaking changes with pandas 2.0.0

VIKTORVAV99 (commented):
Some early feedback and it's looking really good so far.

Just a few things of note:
It looks like the data also has generation data (this is not the same as production) could you add a function for that as well? It should help make the estimations more accurate in the future.

Also this PR will be blocked untill we can verify there are no breaking changes in poetry 2.0. If you can change it back to 1.5.3 we can unblock this PR sooner.

Also if this new parser replaces IN_WE.py fully please archive the old parser.

parsers/IN_WE_new.py
saurabhchatterjee23: done

parsers/IN_WE_new.py
saurabhchatterjee23: done

poetry.lock
saurabhchatterjee23: Looks like the problem was with a newer version of `SQLAlchemy = 2.0` which was a dependency of `pandasql`. After adding a dependency on SQLAlchemy 1.4, the error is fixed without upgrading the version for `pandas`. We do not need to update the version of `pandas` anymore.

With 
- pandas 1.5.3
- SQLAlchemy - 2.0
- pandasql - 0.7.3 
I was getting the following exception

```
sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: 
```

which is fixed now with 
- pandas 1.5.3
- SQLAlchemy - 1.4
- pandasql - 0.7.3

saurabhchatterjee23:
> It looks like the data also has generation data (this is not the same as production) could you add a function for that as well? It should help make the estimations more accurate in the future.

What is the difference between generation data and production data? do you have any wiki I can read to understand it better?

Screenshot 2023-05-10 at 2 38 47 PM
Screenshot 2023-05-10 at 2 39 10 PM
Screenshot 2023-05-10 at 2 39 18 PM


Do you have any other parser doing it which I can read?
It is difficult to understand what each field represents in the JSON because there is no documentation about it, I am trying to find something so that I can add more comments

> Also this PR will be blocked until we can verify there are no breaking changes in poetry 2.0. If you can change it back to 1.4.4 we can unblock this PR sooner.

Done 

> Also if this new parser replaces IN_WE.py fully please archive the old parser.

Done

saurabhchatterjee23:
>Yes sorry I mean the generation forecast but perhaps that is not available?

generation forecast is not available from the source

>It seems like current generation data is available through if I am reading the images correctly?

That's correct, current generation data is available

saurabhchatterjee23:
>Both the exchange (undocumented) and the production parsers support including the current capacity values. In this case that would only be relevant for the exchanges. Take a look at https://github.com/electricitymaps/electricitymaps-contrib/pull/5203 for a proof of concept, there are some other limitations blocking that PR right now but this PR don't suffer from that.

I am working on writing `query_exchange_capacity_dayahead` method for this parser as import TTC and export TTC (TTC stands for Total Transfer Capability) can give us this information

saurabhchatterjee23:
I have taken care of the following feedback

1. Adding exchange capacity fields to fetch_exchange
2. Adding comments for fields in the JSON
3. Getting data for consumption from 'Demand' field instead of  'Act_Drawal' as per https://github.com/electricitymaps/electricitymaps-contrib/pull/5387

VIKTORVAV99 (commented):
I'm not sure you are aware but #5387 added parser validation for all the parser validation by using the new data type classes. Like this:
```
    exchange_list = ExchangeList(logger)
    for dt in get_date_range(target_datetime):
        net_flow = format_exchanges_data(
            zone_key1=zone_key1,
            zone_key2=zone_key2,
            data=data,
            target_datetime=dt,
        )
        exchange_list.append(
            zoneKey=ZoneKey(sortedZoneKeys),
            datetime=arrow.get(dt).replace(tzinfo="Asia/Kolkata").datetime,
            netFlow=net_flow,
            source="wrldc.in",
        )

    return exchange_list.to_list()
```
This PR currently reverts them and the plan is to gradually move over all parsers to this system (not just yet though) but the parsers that the team has updated should not be reverted, including this one.

saurabhchatterjee23:
> I'm not sure you are aware but #5387 added parser validation for all the parser validation by using the new data type classes. Like this:
> 
> ```
>     exchange_list = ExchangeList(logger)
>     for dt in get_date_range(target_datetime):
>         net_flow = format_exchanges_data(
>             zone_key1=zone_key1,
>             zone_key2=zone_key2,
>             data=data,
>             target_datetime=dt,
>         )
>         exchange_list.append(
>             zoneKey=ZoneKey(sortedZoneKeys),
>             datetime=arrow.get(dt).replace(tzinfo="Asia/Kolkata").datetime,
>             netFlow=net_flow,
>             source="wrldc.in",
>         )
> 
>     return exchange_list.to_list()
> ```
> This PR currently reverts them and the plan is to gradually move over all parsers to this system (not just yet though) but the parsers that the team has updated should not be reverted, including this one.

I have made similar changes to use ExchangeList and TotalConsumptionList instead, I was aware of these changes, but was unaware of the intent to move all parsers to use this new data-structure in the future. Not a big change for me


> I'm also not sure this longer needs a full rewrite of the old parser?
Yes we can do the same fixes in the current version.

Here is why I think we must move to the newer version
* The older version fetches data from API which has **wrong timestamps** - The timestamps are in 12 hour format but has missing AM/PM, the newer API that I am using gives timestamps in correct 24hour format
* The source has issues of **missing** data; there are scenarios when the data is unavailable for hours and even days. If the parser searches for data for hours for which the source is doesn't have data, it fails with that `Region_Name` error, we can fix that in the existing code but the rewrite is very clean in handling that
* The source does not provide future data, the older version would generate datapoints in future and theow error, I felt using something as simple as SQL could have never caused these in the firstplace, writing too much of code to process data leaves room for errors.
* The new version is much more readable and easy to maintain if we use simple SQL query to encapsulate the business logic
> Is there a way to migrate these changes to the existing parser instead?

Yes it can be done, but I will not recommend fixing existing code, after learning issues with the data source and I believe rewrite is clean, readable, easy to maintain and bugfree

saurabhchatterjee23:
There are too many commits in this CR, do you want me to squash them into one commit and create another PR?

VIKTORVAV99:
> Here is why I think we must move to the newer version
>
> The older version fetches data from API which has wrong timestamps - The timestamps are in 12 hour format but has missing AM/PM, the newer API that I am using gives timestamps in correct 24hour format
The source has issues of missing data; there are scenarios when the data is unavailable for hours and even days. If the parser searches for data for hours for which the source is doesn't have data, it fails with that Region_Name error, we can fix that in the existing code but the rewrite is very clean in handling that
The source does not provide future data, the older version would generate datapoints in future and throw error, I felt using something as simple as SQL could have never caused these bugs in the first place, writing too much of code to process data leaves room for errors.
The new version is much more readable and easy to maintain if we use simple SQL query to encapsulate the business logic

I was unaware of all these issues, makes sense then. I'm still not 100% sold on mixing in SQL if it's not needed but I understand why you used it. I'll do another review right away and then request a second review from the person working on the India zones at the moment.

> There are too many commits in this CR, do you want me to squash them into one commit and create another PR?

No need, we use squash merge so all commits will be squashed when it's merged. In fact we prefer multiple small commits in the PRs as it's easier to review.

parsers/IN_WE.py
```diff
@@ -175,33 +96,113 @@ def fetch_exchange(
...
-            data=data,
-            target_datetime=dt,
-        )
+
+    for index in result.index:
         exchange_list.append(
-            zoneKey=ZoneKey(sortedZoneKeys),
-            datetime=arrow.get(dt).replace(tzinfo="Asia/Kolkata").datetime,
-            netFlow=net_flow,
+            netFlow=result["exchange_value"][index],
```
VIKTORVAV99, line 126:
```suggestion
        exchange_list.append(
          # TODO: Add live capacity parsing once supported by validation.
            netFlow=result["exchange_value"][index],
```

VIKTORVAV99 (commented):
For me this looks good but I'll request an review from the relevant team members.

saurabhchatterjee23:
I have one more question @VIKTORVAV99
Is there any fixed time of day when the parsers run when they are annotated with `@refetch_frequency(timedelta(days=1))`? 
From the logs it doesn't appears so.

```
def get_date_range(dt: datetime):
    return pd.date_range(
        arrow.get(dt).floor("day").datetime,
        arrow.get(dt).ceil("day").floor("hour").datetime,
        freq="H",
    ).to_pydatetime()

```
This method is used to find hours for which the parser will fetch data from the source, it will always get data from **12 midnight to current datetime**, the parser will not find datapoints from the datasource in the future.

This parser will only fetch accurate and complete data for entire day if it is run in the last hour of the day, if the parser is run on the next day maybe delayed by let's say 15 minutes, it will try to fetch data for tomorrow and will not fetch data for the current day. For the parser to capture data in for entire day, given it is not decided at what hour of day it will be run, the parser must capture data for more than 24 hours irrespective of which date it is run for. 

Let's say we run the parser at 5pm, it should fetch data from yesterday 11am to today 5 pm for 1 day 6 hours so that there is a overlap in hours form the data collected yesterday. In the older version it will simply skip the data for yesterday's evening

We can even increase the frequency of these parsers to run every 12 hours and then collect data for last 24 hours

VIKTORVAV99:
> I have one more question @VIKTORVAV99
> Is there any fixed time of day when the parsers run when they are annotated with `@refetch_frequency(timedelta(days=1))`? 
> From the logs it doesn't appears so.
> 
> ```
> def get_date_range(dt: datetime):
>     return pd.date_range(
>         arrow.get(dt).floor("day").datetime,
>         arrow.get(dt).ceil("day").floor("hour").datetime,
>         freq="H",
>     ).to_pydatetime()
> 
> ```
> This method is used to find hours for which the parser will fetch data from the source, it will always get data from **12 midnight to current datetime**, the parser will not find datapoints from the datasource in the future.
> 
> This parser will only fetch accurate and complete data for entire day if it is run in the last hour of the day, if the parser is run on the next day maybe delayed by let's say 15 minutes, it will try to fetch data for tomorrow and will not fetch data for the current day. For the parser to capture data in for entire day, given it is not decided at what hour of day it will be run, the parser must capture data for more than 24 hours irrespective of which date it is run for. 
> 
> Let's say we run the parser at 5pm, it should fetch data from yesterday 11am to today 5 pm for 1 day 6 hours so that there is a overlap in hours form the data collected yesterday. In the older version it will simply skip the data for yesterday's evening
> 
> We can even increase the frequency of these parsers to run every 12 hours and then collect data for last 24 hours
> 

The parsers run every 15 minutes, the refetch frequency is just to indicate how much data is returned when we refetch historical data.

mathilde-daugy:
> Some early feedback and it's looking really good so far.
> 
> Just a few things of note: It looks like the data also has generation data (this is not the same as production) could you add a function for that as well? It should help make the estimations more accurate in the future.
> 
> Also this PR will be blocked untill we can verify there are no breaking changes in poetry 2.0. If you can change it back to 1.5.3 we can unblock this PR sooner.
> 
> Also if this new parser replaces IN_WE.py fully please archive the old parser.
 we do not want generation from this data source, it doesn't cover all of the production on the grid. We're using data from the Central Electricity Authority for production

mathilde-daugy:
@saurabhchatterjee23, thank you for all you hard work on this. We will take the time to properly review this PR before making any decision. 
Just a comment though, we are not really keen to use the `pandasql` library as it seems to be very new and we think that it could be unstable and cause trouble in the future. @unitrium if you can add anything on this topic . 

Thank you in advance for your patience :)

saurabhchatterjee23:
> @saurabhchatterjee23, thank you for all you hard work on this. We will take the time to properly review this PR before making any decision. Just a comment though, we are not really keen to use the `pandasql` library as it seems to be very new and we think that it could be unstable and cause trouble in the future. @unitrium if you can add anything on this topic .
> 
> Thank you in advance for your patience :)

Thanks for the appreciation. Take your time and review this parser's code. 

Just a FYI, pandasql is a new but it is very thin wrapper over SQLAlchemy which has been very stable and was launched in 2006.

PR #5362, opened 2023-05-04
JP `Unknown` emission factor for regions without discrete sources

### LINKED ISSUE #5361: [Data Issue]: JP carbon emission factor for zones without discrete power source is way off
### When did this happen?

"2023-05-05"

### What zones are affected?

Hokuriku, Chugoku, Shikoku, Okinawa, Tohoku, and Hokkaido

### What is the problem?

The zones mentioned above all have their carbon emission factor defaulted to 700gCO₂eq/kWh, even if many of these zones do contain nuclear and renewable energy sources.

It may be better to simply default these values for unknown sources in these regions to [2020 Japanese National Average](https://www.statista.com/statistics/1303959/japan-carbon-intensity-of-energy-production-per-kilowatt-hour/):
`
In 2020, the carbon intensity of energy production in Japan amounted to almost 0.22 kilogram of carbon dioxide per kilowatt hour. The figure decreased from about 0.24 kilogram of CO2 per kilowatt hour between 2012 to 2014.
`

Thus bringing down the number to 220gCO₂eq/kWh for Unknown data sources.

PPsyrius:
## Issue

Closes #5361.

## Description

This adds `unknown` emission source data for Japanese zones without discrete sources, which are:
- Hokuriku (`JP-HR`)
- Chugoku (`JP-CG`)
- Shikoku (`JP-SK`)
- Okinawa (`JP-ON`)
- Tohoku (`JP-TH`)
- Hokkaido (`JP-HKD`)

which now defaults to `462gCO₂eq/kWh` as per `Enerdata 2022`'s Japanese National Average data cited in [`Climate Transparency`'s Japan, Country Profile 2022](https://www.climate-transparency.org/wp-content/uploads/2022/10/CT2022-Japan-Web.pdf) instead of the default `700gCO₂eq/kWh` .

### Preview

```yml
emissionFactors:
  lifecycle:
    unknown:
      _comment: 'Source: Enerdata 2022 Japanese National Average'
      _url:
        - https://www.climate-transparency.org/wp-content/uploads/2022/10/CT2022-Japan-Web.pdf
      source: Enerdata 2022
      value: 462
```

### Double check

### DIFF
--- a/config/zones/JP-CG.yaml
+++ b/config/zones/JP-CG.yaml
@@ -14,9 +14,18 @@ capacity:
   solar: 55.9
   wind: 365
 disclaimer: The data provider (Chugoku Electric Power Co.,Inc.) provides live data only for solar and unknown. We are working on improving the data quality for this zone.
+emissionFactors:
+  lifecycle:
+    unknown:
+      _comment: 'Source: Enerdata 2022 Japanese National Average'
+      _url:
+        - https://www.climate-transparency.org/wp-content/uploads/2022/10/CT2022-Japan-Web.pdf
+      source: Enerdata 2022
+      value: 462
 contributors:
   - tmslaine
   - lorrieq
+  - PPsyrius
 parsers:
   consumptionForecast: JP.fetch_consumption_forecast
   price: JP.fetch_price
--- a/config/zones/JP-HKD.yaml
+++ b/config/zones/JP-HKD.yaml
@@ -20,7 +20,16 @@ contributors:
   - lorrieq
   - pierresegonne
   - wobniarin
+  - PPsyrius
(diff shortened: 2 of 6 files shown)


VIKTORVAV99:
I believe @wobniarin are working on the Japanese zones so I'll let her have the final say but it does look like an improvement even if it might not be as accurate as we would like (nuclear is still included). Maybe we can do some math to try and remove the nuclear share of this?

wobniarin:
Hi @PPsyrius! 

Thank you very much for this!  We have validated the data source and we will include this new value for all zones in Japan. We will approve it and make it ready to be seen on our website 🎉

wobniarin (approved):

PR #5364, opened 2023-05-05
Vietnam parser (continued)

MuffinCompiler:
## Issue

Continuation of #5055 as new pull request as suggested.

## Description

Created a consumption and price parser for Vietnam, Vietnam North, Vietnam Central, Vietnam South.
Copied over from #5055:

> Hey folks,
>
> I decided this weekend to look at the Vietnam data and made @alixunderplatz parser work as intended. I implemented the fetch_consumption and fetch_price for both current and historical data, for the three sub-zones (north, central, south), and Vietnam in total. Also created .yaml files accordingly, but without geo information.
>
>The data source is a bit... weird, especially regarding past data. If you request data for a given past day, you only get data for that day up to the current time. So you should poll past data late in the evening (Vietnam time) to get a complete set ;) On the upside, the source also has data for the upcoming 2 hours, probably a short-range consumption forecast, these future data points change every 10 minutes or so.
>
>There are a few more issues I encountered:
>
> - The daily graph on the data page does not match perfectly the archive of real time data by a difference of +/- 2% of total load. I guess there is some post-processing going on, the title of the daily chart has some word in it that translates to 'correlated load'. The real time data might also be more of a estimate.
> - The price in the daily chart is more significantly off. However, the real time data set is labeled as 'market price', while the daily chart prices translate to something like 'variable price'. The real time price also randomly plummets to 0, 1 or some other random low value. I set a simple threshold for some filtering, but this still gives weird values sometimes.
> 
> I also found a nice source of [capacities ](https://www.nldc.evn.vn/FullNewsg/100/Thong-tin-thi-truong-dien/default.aspx)of different power sources. This source seems also being updated from time to time. There is also a next-day forecast of some load, but not with a time, more like a maximum being expected. But this also gives us a nice estimate on the power sources! All in all, this should give us a parser similar to the Kuwait one.

Supported at the moment:
`poetry run test_parser [VN|VN-N|VN-C|VN-S] [consumption|price]`

### Double check

Closes: #5055
Closes: #1974


parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def filter_prices(per_zone_price):
+    """
+    Filter the prices for each zone is_price_valid lambda function. If they seem wrong, replace them with None.
+    """
+    for z, prices in per_zone_price.items():
+        per_zone_price[z] = [(p if is_price_valid(p) else None) for p in prices]
+
+    return per_zone_price
```
VIKTORVAV99, line 161: As mentioned before, I don't think this is something we should do.

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+        zone_prices = [
+            ps[i] for zone, ps in per_zone_price.items() if ps[i] is not None
+        ]
+
+        if len(zone_prices):
+            averaged_vn_prices.append(mean(zone_prices))
+        else:
+            averaged_vn_prices.append(None)
+
+    return averaged_vn_prices
```
VIKTORVAV99, line 182: Hmm, not sure I like that we are creating data here.

I think it would be best to leave it out and just add price data to the subzones.

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+    per_zone_price = {
+        "VN-N": fetched_data["data"][0],
+        "VN-C": fetched_data["data"][1],
+        "VN-S": fetched_data["data"][2]
+        # no total VN data
+    }
+
+    # filter out faulty values
+    per_zone_price = filter_prices(per_zone_price)
+    per_zone_price["VN"] = compute_national_average(per_zone_price)
```
VIKTORVAV99, line 213: Both of these can be removed based on the above comments.

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+
+    result_list = []
+
+    for i in range(len(per_zone_price[zone_key])):
+        # Convert to datetime. Data is for [0:30, 1:00, 1:30, ..., 23:30, 24:00]
+        # Or only up to latest available time
+        i_datetime = fetched_day.replace(
+            hour=int((i + 1) / 2), minute=(0 if i % 2 else 30)
+        ).floor("minute")
```
VIKTORVAV99, line 222: This is the same function as above in the consumption parser.

It can be extracted into its own function and reused in both.

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+
+        result_list.append(
+            {
+                "datetime": i_datetime.datetime,
+                "price": per_zone_price[zone_key][i],
+                "currency": "VND",
+                "zoneKey": zone_key,
+                "source": "nldc.evn.vn",
+            }
+        )
```
VIKTORVAV99, line 241: Just letting you know we are working on classes for this that has built in validation. They are not 100% done/ready yet so I'll check with the data team if they want us to add it to the parser.

VIKTORVAV99 (commented):
Overall this looks good works well.

I do think there are a few things that need to be addressed at first though.

config/zones/VN-C.yaml
```diff
@@ -0,0 +1,6 @@
+
+flag_file_name: vn.png
```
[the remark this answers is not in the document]
MuffinCompiler: Fixed in 5eb9af2

config/zones/VN.yaml
```diff
@@ -1,2 +1,32 @@
...
+  gas: 7398
+  hydro: 22345
+  oil: 1579
+  solar: 16567
+  unknown: 619
+  wind: 4667
+  
+sources:
+  EVN:
+    link: https://www.nldc.evn.vn/FullNewsg/100/Thong-tin-thi-truong-dien/default.aspx # capacities
```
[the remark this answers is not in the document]
MuffinCompiler: Moved to DataSources 65acfe1.

config/zones/VN.yaml
```diff
@@ -1,2 +1,32 @@
...
+  price: VN.fetch_price
+  consumption: VN.fetch_consumption
+
+bounding_box:
+  - - 135.84
+    - 33.82
+  - - 138.72
+    - 37.17
+  
+# emissionFactors:
```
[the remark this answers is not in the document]
MuffinCompiler: Done.

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+# Access via day, can also parse historical data
+
+tz = pytz.timezone("Asia/Ho_Chi_Minh")
+base_url_from_type = {
+    "consumption": "https://www.nldc.evn.vn/PhuTaiHandle.ashx?d=",  # + dd/mm/yyyy
+    "price": "https://www.nldc.evn.vn/GiaBienHandle.ashx?d=",  # + dd/mm/yyyy
+}
+# Prices are sometimes randomly 0₫, 1₫, or other small values.
+# We use here 1200 Đồng/kWh as threshold to accept price. Seems to be ~1700 in 2023.
+is_price_valid = lambda price: price > 1200
```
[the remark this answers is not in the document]
MuffinCompiler: I am not totally familiar with these prices, and how reliable they are. But I guess that the parser might be the wrong place to filter unreasonable data anyways. If you guys plan to install checks in the backend, that makes more sense. Removed here 09b29ec.

parsers/VN.py
MuffinCompiler: Done, as answered before. 09b29ec

parsers/VN.py
MuffinCompiler: Done. 09b29ec

parsers/VN.py
MuffinCompiler: 09b29ec

parsers/VN.py
MuffinCompiler: Good point, totally missed that. 573d208

parsers/VN.py
```diff
@@ -0,0 +1,255 @@
...
+
+    base_url = base_url_from_type[data_type]
+
+    today_dt_str = fetched_day.format("DD/MM/YYYY")
+
+    # Try fetch data for today.
+    today_response: Response = session.get(base_url + today_dt_str)
+
+    # Check if today's data is there.
+    if today_response.status_code == 200 and today_response.json() is not None:
```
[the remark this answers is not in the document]
MuffinCompiler: Fixed in 9b8e336.

parsers/VN.py
VIKTORVAV99: The new classes are not required yet so this can proceed without them.

VIKTORVAV99 (commented):
Okay so I think the changes in here look good now but our CI and build pipelines requires all zones to have a geometry, including subzones so we need to split up Vietnam into its subzones.

If you have the time and feel comfortable doing it please do so otherwise I can take a look at it this weekend, sadly won't have much time before then and the geometries can take a while.

MuffinCompiler:
> Okay so I think the changes in here look good now but our CI and build pipelines requires all zones to have a geometry, including subzones so we need to split up Vietnam into its subzones.
> 
> If you have the time and feel comfortable doing it please do so otherwise I can take a look at it this weekend, sadly won't have much time before then and the geometries can take a while.

It would be great if you could have a go at the geometries. Luckily, N/C/S are quite nicely separated by the individual [subregions of Vietnam](https://en.wikipedia.org/wiki/Northern,_Central_and_Southern_Vietnam).

VIKTORVAV99:
> > Okay so I think the changes in here look good now but our CI and build pipelines requires all zones to have a geometry, including subzones so we need to split up Vietnam into its subzones.
> > If you have the time and feel comfortable doing it please do so otherwise I can take a look at it this weekend, sadly won't have much time before then and the geometries can take a while.
> 
> It would be great if you could have a go at the geometries. Luckily, N/C/S are quite nicely separated by the individual [subregions of Vietnam](https://en.wikipedia.org/wiki/Northern,_Central_and_Southern_Vietnam).

I'll take a look at them this weekend!

shuuji3:
Hello, I think each region can be created using the bottom map on the website: https://www.nldc.evn.vn/FullNewsg/100/Thong-tin-thi-truong-dien/default.aspx

And it looks like this map is rendered from an SVG image, so maybe you can use it as a draft for the new geometry. You can fetch the SVG text by this CSS selector:

```js
document.querySelector('#highcharts-0 > svg > g.highcharts-series-group > g > g:first-child')
```

Screenshot 2023-05-13 at 13 17 50

I hope this would help the task. 🙂

VIKTORVAV99 (approved):
Thanks you @alixunderplatz and @MuffinCompiler for working on this parser!

I have added the geometries and updated the configs with the bounding boxes and added us as contributors for the zones.

There are some things left to do like creating translations and such for the zones but I think that is best to do in another PR, therefore I am approving this PR now so we can get it merged!

PR #5365, opened 2023-05-05
Resolves #5352: Added overflowY: auto style to ZoneHeader component

### LINKED ISSUE #5352: Zone name in panel is too wide and pushes gauges out
### Describe the bug
In some cases, especially for US zones in German (as the "USA" translation is long itself), the zone name is way too wide for the panel and therefore content is pushed out.

There's two bugs here: 

1. The gauges should not extend so far, and 
2. we should handle long zone names better so it does not overflow.

### To Reproduce

**Gauge missing:**

1. Go to https://app.electricitymaps.com/zone/US-MIDW-MISO on a mobile device
2. Change language to "Deutsch" 
3. See how the Renewable/regenerativ gauge is not showing any longer

**Overflow on mobile:**

1. Go to https://app.electricitymaps.com/zone/US-NW-NEVP on a mobile device
2. Change language to "Deutsch" 
3. See how the country label is overflowing

**Overflow on desktop:**

1. Go to https://app.electricitymaps.com/zone/US-MIDW-MISO on a desktop
2. Change language to "Deutsch" 
3. See how the country label is overflowing

### Expected behavior

1. The gauge should always show. Should be a somewhat easy CSS fix I'd imagine?
2. The zone name with Country label should not overflow. Harder to fix... A very simple quickfix could be to rename "Vereinigte Staaten" to "USA" in the German translation? 😅 

### Screenshots

Screenshot 2023-05-03 at 08 57 55

Screenshot 2023-05-03 at 09 01 51

Screenshot 2023-05-03 at 08 57 42

patrickwchoi:
## Description

To fix this issue, I just added a minor styling adjustment to the ZoneHeader Component: `style={{ overflowY: 'auto' }}`. I hope this is the change you're looking for. Also, this is my first OSS pull request so I'm sorry if I made any mistakes.

### Preview

image
image
As you can see, the piecharts and the USA header doesn't overflow anymore, even when the USA tag is lengthened. Not sure if you want the little scrolling bar at the bottom though.

### Double check

### DIFF
--- a/web/src/features/panels/zone/ZoneHeader.tsx
+++ b/web/src/features/panels/zone/ZoneHeader.tsx
@@ -61,7 +61,7 @@ export function ZoneHeader({ zoneId, data, isAggregated }: ZoneHeaderProps) {
   const isEstimated = estimationMethod !== undefined;
 
   return (
-    <div className="mt-1 grid w-full gap-y-5 sm:pr-4">
+    <div className="mt-1 grid w-full gap-y-5 sm:pr-4" style={{ overflowY: 'auto' }}>
       <ZoneHeaderTitle
         zoneId={zoneId}
         isEstimated={isEstimated}


VIKTORVAV99:
Hi @patrickwchoi and thanks for the PR!

I'm not sure this is what we had in mind in the issue. I think we should handle the overflow gracefully and try to decuple the gauges from the zone name instead of adding overflow scroll.

If you want to take a crack at solving that it would be great but don't feel forced to do so.

patrickwchoi:
> Hi @patrickwchoi and thanks for the PR!
> 
> I'm not sure this is what we had in mind in the issue. I think we should handle the overflow gracefully and try to decuple the gauges from the zone name instead of adding overflow scroll.
> 
> If you want to take a crack at solving that it would be great but don't feel forced to do so.

patrickwchoi:
I see, I thought it was too simple😅. Do you mind elaborating on how you want it to look like? I'm not sure what you mean by decuple. Do you want the gauges to stay present and only have overflow scroll on the country name?

VIKTORVAV99:
Sorry for not getting back to you before closing this PR, #5461 added the changes that resulted in the behavior we had in mind when creating the issue.

Unfortunately we can't proceed with both PRs so we had to close on of them. Even if I decided to go with the other PR we still value this one and hope you don't get discouraged from contributing in the future.

PR #5369, opened 2023-05-08
Use validation error in the event validation

unitrium:
## Issue

The consumption parser for the EIA raised errors due to some zones returning None for the consumption, most likely Genration only zones. This is not the expected behavior, the expected behavior was that the error would be catched and logged and an not added to the list.

## Description

Turns out the issue is that actually Pydentic raises a Validation error in the end with all the validation errors. Therefore rewrite the create static method to correctly catch those errors and log them.

This will fix the issue we see with consumption data for the US as zones reporting null will only log the error and not raise an error.


VIKTORVAV99 (approved):
Seems rather straight forward to me.

It's nice that you added some tests for it as well 👍🏼

PR #5371, opened 2023-05-08
Update zh translation

### LINKED ISSUE #5262: Wrong translation of "gas" in Chinese interface
## Description
When viewing electricity generation by energy source in the web app at app.electricitymaps.com (presumably also the other versions) with Chinese language setting, "Gas" is translated as "煤气". This is probably incorrect, and should instead be "天然气".

In this context "Gas" presumably means "natural gas" in the vast majority of cases. "煤气" means literally "coal gas", and refers specifically to gases made from coal. This may be used on a relatively modest scale for power generation at industrial plants with processes that generate incompletely combusted gases from coal, such as blast furnaces. Before natural gas came into widespread use, synthetic gas made from coal was one of the most common types of gas used for combustion, so maybe 煤气 popped up in some translation software for that reason. But these days, the vast majority of gas used for power generation is of course natural gas, which is "天然气" in Chinese.

qyearsley:
## Issue

Closes #5262

## Description

Updates the translation for gas (natural gas / fossil gas) from the colloquial 煤气 to the more specific 天然气. See for example that this is the term used in the wikipedia article https://zh.wikipedia.org/zh-tw/%E5%A4%A9%E7%84%B6%E6%B0%94 corresponding to the English "natural gas" article.

### Double check

- [n/a] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [n/a] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.


qyearsley:
Oops, I made a PR that included all of my changes for the other branch! Will close and try agian

PR #5374, opened 2023-05-09
Fix datetime not serializable error in error logging

unitrium:
## Issue

We get this error sometimes when parsing consumption:
`TypeError: Object of type datetime is not JSON serializable `

## Description

Make sure the datetime becomes a proper string when logging the error.


unitrium:
Closes https://linear.app/electricitymaps/issue/ELE-2075/datetime-is-not-json-serializable

mathilde-daugy (approved):

PR #5378, opened 2023-05-09
Fit Price data from ENTSOE parser with new data structure

unitrium:
## Description

Adapt the price data from ENTSO-E to the new Event data structure, clean up of the code and add testing.


parsers/ENTSOE.py
```diff
@@ -214,12 +216,13 @@
...
 # Some zone_keys are part of bidding zone domains for price data
-ENTSOE_PRICE_DOMAIN_OVERRIDE: Dict[str, str] = {
+ENTSOE_PRICE_DOMAIN_MAPPINGS: Dict[str, str] = {
     "AX": ENTSOE_DOMAIN_MAPPINGS["SE-SE3"],
     "DK-BHM": ENTSOE_DOMAIN_MAPPINGS["DK-DK2"],
     "DE": ENTSOE_DOMAIN_MAPPINGS["DE-LU"],
     "IE": ENTSOE_DOMAIN_MAPPINGS["IE(SEM)"],
     "LU": ENTSOE_DOMAIN_MAPPINGS["DE-LU"],
+    **ENTSOE_DOMAIN_MAPPINGS,
 }
```
VIKTORVAV99, line 226:
```suggestion
ENTSOE_PRICE_DOMAIN_MAPPINGS: Dict[str, str] = {
    **ENTSOE_DOMAIN_MAPPINGS, # Note: This has to be first so the domains are overwritten.
    "AX": ENTSOE_DOMAIN_MAPPINGS["SE-SE3"],
    "DK-BHM": ENTSOE_DOMAIN_MAPPINGS["DK-DK2"],
    "DE": ENTSOE_DOMAIN_MAPPINGS["DE-LU"],
    "IE": ENTSOE_DOMAIN_MAPPINGS["IE(SEM)"],
    "LU": ENTSOE_DOMAIN_MAPPINGS["DE-LU"],
}
```

parsers/test/mocks/ENTSOE/FR_prices.html
VIKTORVAV99, line 1: Can we change this to a XML file extensions since it is well XML?

Just for personal preference I suppose.

VIKTORVAV99 (changes requested):
There is one thing that needs to be addressed that has a suggested change and one thing I'd just like to change for personal preference. 

Otherwise it seems to function as expected.

unitrium:
Done for both.

VIKTORVAV99 (approved):
LGTM! 🎉

I fixed the last test as well.

PR #5381, opened 2023-05-09
New parser for Bangladesh

### LINKED ISSUE #5370: More up-to-date data source for Bangladesh
Hey all,

the Power Grid Company of Bangladesh has a new table for its production and import data. They seem to be updated only in batches every 5-6 hours, therefore afaik basically never fulfill the real time requirement of a <2h delay. The current parser parses Daily Reports from the day before via excel files (which match this data source). Is there interest in a parser for this data source, because it contains more recent data, but still not real-time?

https://erp.pgcb.gov.bd/web/generations/view_generations
https://erp.pgcb.gov.bd/web/generations/imported_power

MuffinCompiler:
## Issue

Closes #5370 

## Description

Rewrite the parser for Bangladesh to make use of the [new data source](https://erp.pgcb.gov.bd/web/generations/view_generations), supporting consumption, generation, and exchanges with Northeastern and Eastern India.
Data is being updated only in batches, therefore, most of the time, it does not fulfill the real-time criteria of <2h delay. (But it can from time to time!) Also historical data can be fetched. I added capacities to the config file from their last annual report.

Supported commands:
`poetry run test_parser BD generation`
`poetry run test_parser BD consumption`
`poetry run test_parser "BD->IN-NE" exchange`
`poetry run test_parser "BD->IN-EA" exchange`
All work with  `--target_datetime YYYYMMDD` as well.

Data also contains loadshed information, probably not important for us. The sum of consumption sources does not add up to the given total generation, I put the rest in 'unknown'.
Weird is also that the capacity for coal in the [annual report](https://bpdb.portal.gov.bd/sites/default/files/files/bpdb.portal.gov.bd/annual_reports/7b792f67_bf50_4b3d_9bef_8f9b568005c9/2022-11-29-05-22-0dea17e09d8a84e72a63312df6b5bdc6.pdf) (page 13) is lower than the current generation. Not sure what to make out of this.

### Double check


parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
+    for row in rows:
+        row_items = row.find_all("td")
+        row_items = [item.text.strip() for item in row_items]
+
+        row_dict = dict()
+
+        # date and time in [0]; [1] are DD-MM-YYYY; hh:mm:ss
+        row_dict["time"] = arrow.get(
+            row_items[0] + row_items[1], "DD-MM-YYYY" + "HH:mm:ss", tzinfo=tz
+        ).datetime
```
VIKTORVAV99, line 74: This currently fails when fetching really early data becasue the time is recorded as "HH:mm" not "HH:mm:ss" this should be updated to handle both senarions. I would try and parse the date seperatly and then the time seperatly and merge those. Preferable using the native datetime object but this is not a requirement.

parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
+    "Remarks",
+]
+
+
+def table_entry_to_float(entry: str):
+    """
+    Helper function to handle empty cells in table.
+    """
+    if entry == "":
+        return 0.0
```
VIKTORVAV99, line 46: If the entry is missing ("") then we should return None instead of 0. 0 should never be set manually (there are rare cases where negative production is reported and we set it to 0 but this is not the case here).

parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
 
-def excel_handler(shifted_target_datetime, logger: Logger) -> tuple:
-    """
-    Decides which url to request based on supplied arrow object.
-    Converts returned excel data into dataframe, format of data varies by date.
-    Returns a tuple containing (dataframe, bool).
-    """
+    # Find the table, there is only one, and verify its headers.
+    table = response_soup.find("table")
+    table_head = table.find("thead")
```
VIKTORVAV99, line 148: table can both be defined and undefined here so we should check that table is not none before accessing the find method. Otherwise this will have a type error.

parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
+    verify_table(table_head)
 
-    # NOTE file named 11-01-2019 actually covers 10-01-2019, pattern repeats!
-    # xls structure very different pre 11-01-18
-    # Before 03-07-2015 it's Daily_Report not Daily Report
-    # File name format adds space on 11-01-18
-    # Odd format around this time (XLS_SPAN)
-    # However file ending changes several days later! (╯°□°）╯︵ ┻━┻
+    # Table valid as we expect, parse the rows.
+    table_body = table.find("tbody")
```
VIKTORVAV99, line 152: Same here.

parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
+                f"Table headers mismatch with expected ones."
+                f"Expected: {TABLE_HEADERS}"
+                f"Parsed: {header_items}"
+            ),
+        )
 
 
-def exchange_processer(df, target_datetime, old_format=False) -> list:
+def query(
+    session: Optional[Session], target_datetime: Optional[datetime], logger: Logger
```
VIKTORVAV99, line 121:
```suggestion
    session: Session, target_datetime: Optional[datetime], logger: Logger
```
Session shoud be defined here to avoid type errors further down in the code.

parsers/BD.py
```diff
@@ -1,299 +1,295 @@
...
-    for index, row in df.iterrows():
-        exchange = row.to_dict()
-        flow = exchange.pop("HVDC", 0.0) + exchange.pop("Tripura", 0.0)
-        dt = timestamp_converter(index, target_datetime)
-        processed_data.append((-1 * flow, dt))
+    target_response: Response = session.get(target_url)
 
-    return processed_data
+    if not target_response.ok:
+        return_status = target_response.status_code
```
VIKTORVAV99, line 138: We can probably avoid this varable declaration and just include the target_response.status_code directly in the message instead.

VIKTORVAV99 (commented):
Thanks for the PR!

Overall this looks good and is working.
There is just a few small things that needs to be addressed. I think I got most of them but the type related changes might require further adjustment down the line.

parsers/BD.py
MuffinCompiler: Done.

parsers/BD.py
MuffinCompiler: Done, see bbd2c87.

parsers/BD.py
MuffinCompiler: Done, see bbd2c87.

parsers/BD.py
MuffinCompiler: Done at b688086. Also checking for None values when actually putting the data into the production/consumption/exchange dicts.

parsers/BD.py
MuffinCompiler: Wow well spotted :) This HH:mm format seems to be present just for a few hours in the first days the data is present. 40d0757

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for creating the new parser! It's very much appreciated!
I pushed some final changes required by our build and CI pipeline (and a small fix for a type error). I hope that's okay.

@mathilde-daugy I know you are still working on estimating the India zones and perhaps the exchange data from this parser might be of use. There are historical data available since 2015-04-19.

parsers/BD.py
```diff
@@ -1,299 +1,320 @@
...
-    return df, OLD_FORMAT
+        # And add sources if they are present in the table
+        known_sources_sum_mw = 0.0
+        for source_type in ["coal", "gas", "hydro", "oil", "solar"]:
+            if row[source_type] is not None:
+                # also accumulate the sources to infer 'unknown'
+                known_sources_sum_mw += row[source_type]
+                data["production"][source_type] = row[source_type]
+
+        # infer 'unknown'
```
q--, line 209: I just noticed that 'unknown' in the ElectricityMaps data is actually equal to the sum of both exchanges from India. I checked multiple datapoints in the graph.

Don't have time to create an issue right now, but perhaps someone else can.

VIKTORVAV99: I'll try and take a quick look at this on Monday and see if it's an issue in the parser.

PR #5385, opened 2023-05-11
Close #5319 AU-NT parser fix

### LINKED ISSUE #5319: AU-NT production parser down
## Description

This is an automatic error report generated for Australia Northern Territory (AU-NT).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser AU-NT production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/AU-NT.html">Explore the runtime logs</a>

celsomilne:
## Issue
Close #5319 

## Description
Fix error when running `poetry test_parser` by changing default behaviour when `target_datetime` is `None` in both `get_data` and `parse_consumption`

### Preview

N/A

### Double check

### DIFF
--- a/parsers/NTESMO.py
+++ b/parsers/NTESMO.py
@@ -112,7 +112,7 @@ def get_data(
     extraction_func: Callable[[bytes], pd.DataFrame],
     logger: Logger,
 ) -> pd.DataFrame:
-    
+
     # Check if the target datetime is None, and if it is, then initialize it to be today minus 2 days
     if target_datetime is None:
         target_datetime = datetime.now() - timedelta(days=2)
@@ -121,7 +121,7 @@ def get_data(
     assert target_datetime is not None, ParserException(
         "NTESMO.py", "Target datetime cannot be None."
     )
-    
+
     index = construct_year_index(target_datetime.year, session)
     try:
         data_file = get_historical_daily_data(
--- a/parsers/NTESMO.py
+++ b/parsers/NTESMO.py
@@ -112,9 +112,16 @@ def get_data(
     extraction_func: Callable[[bytes], pd.DataFrame],
     logger: Logger,
 ) -> pd.DataFrame:
+    
+    # Check if the target datetime is None, and if it is, then initialize it to be today minus 2 days
+    if target_datetime is None:
+        target_datetime = datetime.now() - timedelta(days=2)
+
+    # TODO : remove this assert statement, as it is handled above
     assert target_datetime is not None, ParserException(
         "NTESMO.py", "Target datetime cannot be None."
     )
+    
     index = construct_year_index(target_datetime.year, session)
     try:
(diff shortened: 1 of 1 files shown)


parsers/NTESMO.py
```diff
@@ -135,6 +146,11 @@ def parse_consumption(
     price: bool = False,
 ) -> List[dict]:
     data_points = []
+
```
unitrium, line 149: Since in this function the `target_datetime` cannot be None according to the type hints I don't think the datetime should be set here. It should rather be set at the fetch_production/consumption/price level.

parsers/NTESMO.py
```diff
@@ -68,6 +68,11 @@ class Generator(TypedDict):
     status_forcelist=[500, 502, 503, 504],
 )
 
+# Default behaviour for datetime
+def get_default_datetime_if_not_None(target_datetime: datetime) -> datetime:
+    if target_datetime is None:
+        return datetime.now(tz=AUSTRALIA_TZ) - timedelta(days=2)
```
unitrium, line 74:
```suggestion
        return datetime.now(tz=AUSTRALIA_TZ) - timedelta(days=DELAY)
```
There's a new delay variable to be used here.

parsers/NTESMO.py
```diff
@@ -112,9 +117,15 @@ def get_data(
     extraction_func: Callable[[bytes], pd.DataFrame],
     logger: Logger,
 ) -> pd.DataFrame:
+
+    # Get the default datetime if the target datetime is None
+    target_datetime = get_default_datetime_if_not_None(target_datetime)
```
unitrium, line 122: Same, the datetime cannot be None here according to the type hints.
So rather put this one level above in the fetch functions.

unitrium (changes requested):
Thanks for handling this, I would rather make a few changes before merging it.
Mainly as with other parsers we should set the type hint for the `target_datetime` in fetch functions as Optional[datetime] and inside those functions if it is None, have the default value.

parsers/NTESMO.py
celsomilne: Thanks for your comments! Instead of moving this into the fetch functions, would it be better to change the type hint to `target_datetime: Optional[datetime] = None`? I think this is a good fix, and similar to what appears in other parsers (e.g., [NL](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/NL.py) - see the `fetch_production` definition).

parsers/NTESMO.py
celsomilne: You're right - I will create a `DAY_DELAY` variable, and set it equal to 5, given the comment at the top of the file that "data is being published after 5 days"

celsomilne:
Closing as have submitted a second PR #5392 to respond to separate issue #5391

PR #5387, opened 2023-05-11
Update IN-WE consumption + parser validation

mathilde-daugy:
## Issue

Consumption data parser by `IN-WE.py` is way too low compared to production values available on NPP (parsed by `IN.py`)
The order of magnitude for total production is 50-60 GW whereas the consumption we get is around 25 GW.

Production collected for IN-WE:
image

Consumption collected for IN-WE:
image

## Description

After investigating the data source we realised that we weren't collecting the right data from the data source
image

The parser has also been updated to use the new parser validation structure.

### Preview

new consumption data for 2023-05-11:
image

old consumption data for 2023-05-11:
image

### Double check


pierresegonne (approved):
Nice one

pierresegonne (approved):

PR #5389, opened 2023-05-11
5352 Fix layout of gauges in ZoneHeader

montoyaaguirre:
## Issue

Partial fix to #5352 reported by @madsnedergaard 

## Description

Extracted the gauges container from the zone title container.

#### Before
Screenshot from 2023-05-11 21-11-23

#### Fix 
Gray country badge is hidden if it overflows
Screenshot from 2023-05-11 21-25-54

#### Alternative Fix
Gray country badge is visible if it overflows
Screenshot from 2023-05-11 21-29-54

### Double check


VIKTORVAV99:
Sorry for not getting back to this PR sooner, as you probably noticed this was already closed by merging #5461 as that fixes the issue and the resulting behavior is what we originally intended.

I would just like to say that we value this PR even if we decided to close it and proceed with another one, so please do not feel discouraged by that.

PR #5392, opened 2023-05-11
AU-NT default datetime fix, and readability improvements

### LINKED ISSUE #5391: AU-NT parser doesn't pass poetry tests
## Description

As referenced in PR #5385, there is an error when running `poetry test_parser AU-NT`, where `target_datetime` passed to `get_data` is `None`.

## Output of the error

```
  File "/Users/XXX/electricitymaps-contrib/parsers/NTESMO.py", line 115, in get_data
    assert target_datetime is not None, ParserException(
AssertionError: NTESMO.py Parser: Target datetime cannot be None.
```

# Suggested fix

As suggested in the review of the #5385, should change default behaviour + type hints to enable `None` for `target_datetime`.

celsomilne:
## Issue

Closes #5391

## Description

* As described in #5385, fixes default `target_datetime` value in `AU-NT`, enabling default behaviour when `target_datetime` is `None`, so that it passes poetry tests.
* Also did some cleanup of the code + made more readable with comments

### Preview

N/A

### Double check


parsers/NTESMO.py
```diff
@@ -108,16 +126,20 @@ def extract_demand_price_data(file: bytes) -> pd.DataFrame:
 
 def get_data(
     session: Session,
-    target_datetime: datetime,
     extraction_func: Callable[[bytes], pd.DataFrame],
-    logger: Logger,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
```
VIKTORVAV99, line 131: There should be no need for the default arguments here. The get data function should get these passed from the calling functions.

celsomilne: Resolves in d4d8fbf

celsomilne:
@VIKTORVAV99 have responded to all comments. Would be great to get your review when you have time!

parsers/NTESMO.py
```diff
@@ -68,28 +71,43 @@ class Generator(TypedDict):
     status_forcelist=[500, 502, 503, 504],
 )
 
+# Default behaviour for datetime
+def get_default_datetime_if_not_None(target_datetime: Optional[datetime]) -> datetime:
```
VIKTORVAV99, line 75:
```suggestion
def get_default_datetime_if_not_none(target_datetime: Optional[datetime]) -> datetime:
```

parsers/NTESMO.py
```diff
@@ -210,8 +256,15 @@ def fetch_consumption(
     target_datetime: datetime = arrow.now().shift(hours=-DELAY).to(AUSTRALIA_TZ),
```
VIKTORVAV99, line 256: I would use the new function you created here instead.
I.e. make this optional and then modify it before calling the get data function.

parsers/NTESMO.py
```diff
@@ -130,26 +152,32 @@ def get_data(
 
 def parse_consumption(
     raw_consumption: pd.DataFrame,
-    target_datetime: datetime,
-    logger: Logger,
+    target_datetime: Optional[datetime] = None,
+    logger: Logger = getLogger(__name__),
```
VIKTORVAV99, line 156: This function should always inherit the datetime and logger from the calling function so there is no need to set a default or adding a optional type here.

parsers/NTESMO.py
```diff
@@ -222,8 +275,15 @@ def fetch_price(
     target_datetime: datetime = arrow.now().shift(hours=-DELAY).to(AUSTRALIA_TZ),
```
VIKTORVAV99, line 275: Same here.

parsers/NTESMO.py
```diff
@@ -234,13 +294,21 @@ def fetch_production_mix(
...
```
VIKTORVAV99, line 294: And here.

VIKTORVAV99 (commented):
Some additional suggestions to reduce duplication and make sure the parser always behaves the same.  Let me know if there is any questions.

madsnedergaard:
Hey @celsomilne, would you mind taking a look at the comments here so we can get this PR merged? :)

madsnedergaard:

PR #5394, opened 2023-05-12
Production Breakdown correctly reports corrected negative values.

unitrium:
## Issue

When negative values were passed to a production breakdown, they were fixed by setting them to None. However when returning the dict those values were erased.

Furthermore, it would be great to be able to report values that have been corrected outside of the ProductionBreakdown.

## Description

- Fix None comming from corrected negative values being excluded from the resulting dictionary.

### DIFF
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -77,7 +77,8 @@ class ProductionMix(Mix):
         )
         if keep_corrected_negative_values:
             for corrected_negative_mode in self._corrected_negative_values:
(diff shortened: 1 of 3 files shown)


VIKTORVAV99 (approved):
As far as I can tell it looks good but it's starting to become a little complex and the file is getting longer and longer. Perhaps we should consider splitting up the events into their own files?

I also think we need to start documenting these better soon, the in code documentation is great but if we want contributors to use these I think we need to add a readme or perhaps a wiki entry (this can wait until v1 of these are finalized).

PR #5396, opened 2023-05-12
Fix exchange directions in AR parser

### LINKED ISSUE #5395: [Data Issue]: Exchanges in AR.py parser show mostly in wrong direction.
### When did this happen?

Issue is present apparently since the parser exists? Or the dashboard changed at some point.

### What zones are affected?

AR (and sub zones)

### What is the problem?

Exchange directions are wrong in the AR parser. They are fetched from the graphic on [this](https://cammesaweb.cammesa.com/) site and are based on arrow directions. However, most directions are the other way around in the emaps frontend.

See for example the `AR->PY` exchange. The parser (and emaps frontend) report the inverse value to the one in the source graphic. I think exchanges between AR zones are not processed yet, but they are mostly wrong, too. 
Preparing a pull request in a minute.

MuffinCompiler:
## Issue

Closes #5395.

## Description

Fixes the exchange direction by changing the expected arrow directions to match `zone_key1->zone_key2`.
National exchanges were not processed before I think due to missing yaml files, but international exchanges, e.g. `AR->PY` should now point in the right direction.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5368, opened 2023-05-07
Switches yaml package to gain 5x speed

madsnedergaard:
## Issue

Splitting up the config file in #5269  will be helpful in many cases, but we still often need the zones config data and that takes ~3.3s to load every single time!

## Description

To make this faster, I'm here switching from PyYAML to [ruamel.yaml](https://sourceforge.net/projects/ruamel-yaml/) in the config file. This yields a significant performance improvement from ~3.3s to ~0.6s.

I decided to leave PyYAML as we have more advanced usage in some scripts and I was not sure enough about replacing that at this point.

_I also first tried saving the combined zone configs to one file and loading that, but it was actually slower than loading every single file 😅_ 

### Preview

Before and after:

Screenshot 2023-05-07 at 21 12 33

Validated that output is the same:

Screenshot 2023-05-07 at 21 54 41

### Double check

- [x] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

PR #5398, opened 2023-05-13
Fix update capacity script

VIKTORVAV99:
## Issue

The update capacity script is currently not working.

## Description

Fixes the update capacity script and updates the Bulgaria capacities.

### Double check

- [x] I have tested my parser changes locally with `poetry run python scripts/ENTSOE_capacity_update.py  "zone_key"`


VIKTORVAV99:
Looks like the script nuked the emission factors, that's not great... 😅

I'll update the script to prevent this in the future.

VIKTORVAV99:
Update on this: there seems to be something in https://github.com/electricitymaps/electricitymaps-contrib/blob/579cf5065fb308fef1379a5b9c1911d4927344fb/electricitymap/contrib/config/__init__.py that mutates the ZONE_CONFIG (running print right after the ZONE_CONFIG declaration and at the end of the file returns different results). I don't think that should be happening and will be looking into a fix.

In the meantime I'll open a new PR to fix the capacity for BG.

VIKTORVAV99:
Quick fix to get this working as expected again but we should investigate why the ZONES_CONFIG was mutated in the first place and perhaps migrate other files that use it to use the reading function directly like I did in here.

madsnedergaard:
Would you mind sharing how you run the script? I'm getting a "No module named 'xmltodict'" when trying 😅 
We could also add a small README inside `/scripts` to explain how to run the different scripts.

scripts/ENTSOE_capacity_update.py
```diff
@@ -22,13 +22,17 @@
 import yaml
 from utils import ROOT_PATH, run_shell_command
 
-from electricitymap.contrib.config import CONFIG_DIR, ZONES_CONFIG
+from electricitymap.contrib.config import CONFIG_DIR
+from electricitymap.contrib.config.reading import read_zones_config
```
pierresegonne, line 26: Is that function new?

VIKTORVAV99: Yes and no, it was refractored out of the __init__.py in #5269 but it should be the same. It looks like something is not behaving as expected though but I don't know if that is a new behavior or if it just became evident now.

I'll try and look into this the coming weekend.

madsnedergaard (commented):
When I run the script it also changes indentation for all other fields (contributors, etc.), but might be easier to fix after #5368

madsnedergaard (approved):

VIKTORVAV99:
> Would you mind sharing how you run the script? I'm getting a "No module named 'xmltodict'" when trying 😅 We could also add a small README inside `/scripts` to explain how to run the different scripts.

I'll merge this as is now but I'll try and create some documentation for the scripts in there the coming days.

PR #5401, opened 2023-05-15
Resolving deprecation warnings when running tests.

saurabhchatterjee23:
## Issue

When running `poetry run test`, I was seeing some warnings for deprecation
```
tests/validators/test_zone_specific_checks.py::test_validate_production_has_fossil_fuel
  /home/runner/work/electricitymaps-contrib/electricitymaps-contrib/validators/zone_specific_checks.py:41: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.
    res = (events[fossil_fuel_cols_in_event] > 0).any(axis=1).astype(int)
```
```
parsers/test/test_OPENNEM.py::TestOPENNEM::test_process_solar_rooftop
  /home/runner/work/electricitymaps-contrib/electricitymaps-contrib/parsers/test/test_OPENNEM.py:12: FutureWarning: Argument `closed` is deprecated in favor of `inclusive`.
    idx = pd.date_range(
```
```
parsers/test/test_OPENNEM.py::TestOPENNEM::test_process_solar_rooftop
  /home/runner/work/electricitymaps-contrib/electricitymaps-contrib/parsers/test/test_OPENNEM.py:29: FutureWarning: Argument `closed` is deprecated in favor of `inclusive`.
    assert processed_df.index.equals(
```

## Description

Resolved the issues by upgrading the code to newer version to avoid these warnings,

### Double check

- [X] I have tested my parser changes locally with `poetry run test`

Closes: #5415


parsers/test/test_OPENNEM.py
```diff
@@ -13,7 +13,7 @@ def test_process_solar_rooftop(self):
             start="2021-01-01 00:00:00+00:00",
             end="2021-01-01 01:00:00+00:00",
             freq="5T",
-            closed="left",
+            inclusive="left",
```
VIKTORVAV99, line 16: Are the arguments for inclusive the same as closed?

VIKTORVAV99 (commented):
Thanks for opening this PR and helping to maintain our tests!

Just one question really then this can be merged!

parsers/test/test_OPENNEM.py
saurabhchatterjee23: Yes, this is what the documentation says

```
    closed : {None, 'left', 'right'}, optional
        Make the interval closed with respect to the given frequency to
        the 'left', 'right', or both sides (None, the default).

        .. deprecated:: 1.4.0
           Argument `closed` has been deprecated to standardize boundary inputs.
           Use `inclusive` instead, to set each bound as closed or open.
    inclusive : {"both", "neither", "left", "right"}, default "both"
        Include boundaries; Whether to set each bound as closed or open.
```

The unit test which uses this code continues to succeed and it generates the same results.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5402, opened 2023-05-15
Retain corrected negative production values when merging production outputs

unitrium:
## Issue

When merging different production outputs, the information on corrected negative values was being erased by the fact that we were creating a new object. The information needs to be propagated.

## Description

- To the dictionary form add a corrected negative modes output, that will retain some level of information on which modes were corrected.
- Rework the aggregation function so that we retain the information on corrected modes. Two cases can arise: either the mode value was set to 0 or None or either the mode is positive but one of the merged production values was a corrected value. In both cases we retain the information of which modes were corrected during the different steps.

### Preview

Adding unit tests to showcase how it should work and updating EIA.

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -117,20 +117,43 @@ def merge_production_breakdowns(
                 f"Trying to merge production outputs from multiple source types \
                 , got {len(source_types)}: {', '.join(source_types)}"
             )
-        df = df.groupby(level=0, dropna=False).sum(numeric_only=True, min_count=1)
+
+        def aggregate(value: pd.Series) -> Optional[float]:
```
VIKTORVAV99, line 121: I think this needs a union return type of `Union[Float, List, None]` list could probably be further typed if needed.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -117,20 +117,43 @@ def merge_production_breakdowns(
...
+                    # NaN values can arise from the aggregation. We want to keep them as None.
+                    # We do this at the last step to avoid changing the object type of the dataframe.
+                    if value is not None and np.isnan(value):
+                        value = None
+                    # The key is in the form of "production.<mode>" or "storage.<mode>"
+                    prefix, mode = key.split(".")  # type: ignore
+                    if prefix == "production":
+                        if mode in row[1]["correctedModes"]:
+                            # This is just to mark this mode as corrected, the value is not used.
+                            production_mix.set_value(mode, -1)
```
VIKTORVAV99, line 153: Would it be better to set this as a enum string or even enum number? Magic numbers should usually be avoided.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -117,20 +117,43 @@ def merge_production_breakdowns(
...
+                        production_mix.set_value(mode, value)
+                    elif prefix == "storage":
+                        storage_mix.set_value(mode, value)
```
VIKTORVAV99, line 156: Maybe add else statement after this to catch any unexpected prefixes and log them? Just for future reliablility?

electricitymap/contrib/lib/models/events.py
```diff
@@ -371,6 +371,9 @@ def to_dict(self) -> Dict[str, Any]:
             "storage": self.storage.dict(exclude_none=True) if self.storage else {},
             "source": self.source,
             "sourceType": self.sourceType,
+            "correctedModes": list()
+            if self.production is None
+            else list(self.production._corrected_negative_values),
```
VIKTORVAV99, line 376: Any advantage of using list() over []? Just for my own curiosity?

electricitymap/contrib/lib/tests/test_event_lists.py
```diff
@@ -312,6 +312,92 @@ def test_merge_production_list_predicted(self):
...
+        assert merged.events[0].production.wind is None
+        assert merged.events[0].production.coal == 30
+        assert merged.events[0].storage.hydro == 2
+        assert merged.events[0].production._corrected_negative_values == {"wind"}
+
+        assert merged.events[1].datetime == datetime(2023, 1, 3, tzinfo=timezone.utc)
+        assert merged.events[1].production.wind is None
+        assert merged.events[1].production.coal == 34
+        assert merged.events[1].storage.hydro == 2
+        assert merged.events[1].production._corrected_negative_values == {"wind"}
```
VIKTORVAV99, line 361: While not an issue per say all these declarations show type errors as production.* can be none and we don't explicitly do is not none first. Not sure if we need to "fix" this though but might be nice if we add typecheking to the CI later on.

VIKTORVAV99 (commented):
Not exactly sure how to test this but I think it look good for the most part. There are some type issues and some general suggestions.

So if the type errors are fixed this looks good to me but I would appriciate it if we waited for Mathildes review as well.

electricitymap/contrib/lib/tests/test_event_lists.py
unitrium: Added assertions to fix this.

electricitymap/contrib/lib/models/event_lists.py
unitrium: No in this case we just want to drop the column value as it doesn't match anything.

electricitymap/contrib/lib/models/event_lists.py
unitrium: I extended the set_value method to be able to flag modes. This will yield a better result I believe.

electricitymap/contrib/lib/models/events.py
unitrium: Actually this was to create a fully new object, but actually I read online that this is slightly less efficient so I changed it to [] instead, thanks for flagging this.

unitrium:
Sorry I actually pushed my commits to the wrong branch.

unitrium:
I cherry picked them to this branch, now it's all good.

unitrium:
Adding @pierresegonne and @FelixDQ for further review.

unitrium:
I'm adding some context here to describe a bit what are the different outcomes that this PR is trying to fix:
**Case 1 in merging with the same production type we get a positive outcome but the information that part of the value has been corrected is lost:**
Case1
Not that much of a big deal here, it's already our current setup. It's would just be nice to not loose that information.

unitrium:
Case 2: When merging with another mode the mode is getting set to None without the information that the mode has been corrected. Therefore when changing to the dict form the mode is removed and not present in the dict which could cause some problems down the line.
Case2

unitrium:
Case 3: The mode is corrected by a zero, in this case the resulting Event will have a None instead of a zero. In the flowtracing pipeline it's all the same but it creates a consistency problem because it is no longer highlighted that this is a corrected 0.
Case3

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -117,20 +117,46 @@ def merge_production_breakdowns(
...
-                if prefix == "production":
-                    production_mix.set_value(mode, value)
-                elif prefix == "storage":
-                    storage_mix.set_value(mode, value)
+                if str(key).startswith("production.") or str(key).startswith(
+                    "storage."
+                ):
+                    # NaN values can arise from the aggregation. We want to keep them as None.
+                    # We do this at the last step to avoid changing the object type of the dataframe.
+                    if value is not None and np.isnan(value):
```
pierresegonne, line 146: Can't we do that directly with https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html to the whole series at once?

pierresegonne (commented):
Have you thought about another way of doing this without having to rely on some sort of internal state?
For example, maybe we could do the negative production correction at the very end, in the to_dict function or something like this? This way we would not need to take so many considerations when doing the merging of the production events and it would simplify the logic

mathilde-daugy (approved):

electricitymap/contrib/lib/models/events.py
```diff
@@ -208,6 +253,59 @@ def to_dict(self) -> Dict[str, Any]:
...
+    """
+    An abstract class representing all types of electricity events that can be aggregated.
+    """
+
+    @staticmethod
+    def _unique_zone_key(df_view: pd.DataFrame) -> ZoneKey:
+        zone_key = df_view["zoneKey"].unique()
+        if len(zone_key) > 1:
+            raise ValueError(f"Cannot merge events from different zones: {zone_key}")
+        return zone_key[0]
```
VIKTORVAV99, line 266: Just as an edge case, we have a wrapping parser for the corsica and sardina exchange that merge DC and AC exchanges. Due to how the ENTSO-E parser was structured these needed to have different ZoneKeys.

electricitymap/contrib/lib/models/events.py
```diff
@@ -362,6 +456,45 @@ def create(
                 f"Error(s) creating production breakdown Event {datetime}: {e}"
             )
 
+    @staticmethod
+    def aggregate(events: List["ProductionBreakdown"]) -> "ProductionBreakdown":
```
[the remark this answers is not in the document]
tonypls: Ah nice, makes sense 😄

electricitymap/contrib/lib/models/events.py
```diff
@@ -208,6 +253,59 @@ def to_dict(self) -> Dict[str, Any]:
...
+        return source_type[0]
+
+    @staticmethod
+    def _unique_datetime(df_view: pd.DataFrame) -> datetime:
+        datetime = df_view.index.unique()
+        if len(datetime) > 1:
+            raise ValueError(
+                f"Cannot merge events from different datetimes: {datetime}"
+            )
+        return datetime[0]
```
VIKTORVAV99, line 289: We should probably normalize the time zones here first otherwise we might run into issues where there is diffrent timezone files but the value is the same.

electricitymap/contrib/lib/models/events.py
unitrium: Not really sure I see what you mean, could you point me to the place this is done ?
Anyways the exchange should have an exchange zone that is the combination of both zones ?

VIKTORVAV99: First we define overrides for them here in the ENTSOE parser:
```python
    "FR-COR-AC->IT-SAR": [
        ENTSOE_DOMAIN_MAPPINGS["IT-SACOAC"],
        ENTSOE_DOMAIN_MAPPINGS["IT-SAR"],
    ],
    "FR-COR-DC->IT-SAR": [
        ENTSOE_DOMAIN_MAPPINGS["IT-SACODC"],
        ENTSOE_DOMAIN_MAPPINGS["IT-SAR"],
    ],
```
Then in FR-COR_IT-SAR.py we do this:
```python
   # IT-SACOAC to IT-SAR
    AC_dataList = EXCHANGE_FUNCTION_MAP[type](
        zone_key1="FR-COR-AC",
        zone_key2="IT-SAR",
        session=session,
        logger=logger,
    )
    # IT-SACODC to IT-SAR
    DC_dataList = EXCHANGE_FUNCTION_MAP[type](
        zone_key1="FR-COR-DC",
        zone_key2="IT-SAR",
        session=session,
        logger=logger,
    )
    returnList: List[dict] = []

    # Compare the length and datetimes of the two data lists and
    # if they are the same combine the data from the two lists.
    if (
        len(AC_dataList) == len(DC_dataList)
        and AC_dataList[0]["datetime"] == DC_dataList[0]["datetime"]
        and AC_dataList[-1]["datetime"] == DC_dataList[-1]["datetime"]
    ):
        logger.info("Clean match! Merging data with zip")
        for AC, DC in zip(AC_dataList, DC_dataList):
            returnList.append(
                {
                    "datetime": AC["datetime"],
                    "netFlow": AC["netFlow"] + DC["netFlow"],
                    "sortedZoneKeys": "FR-COR->IT-SAR",
                    "source": AC["source"],
                }
            )
```
There is a fallback and a bit more code around it but that is the basics.

Would be nice if we could use the new merge function with this instead of my custom implementation.

unitrium: Ok, so I believe here the easiest would be to just add a correct mapping. I don't think it should be the merge function that should take care of this.

electricitymap/contrib/lib/models/events.py
unitrium: If the timezone is not the same this will result in an error anyways which is the intended behavior.
The unique part would see different objects.

PR #5403, opened 2023-05-15
Fit ENTSO-E production mix with new ProductionBreakdown structure

unitrium:
## Note:

To be reviewed and merged after #5402.

## Issue

The new ProductionBreakdown structure has been released, we should make sure our parsers conform to it.

## Description

- Update fetch production to use the new ProductionBreakdown structure
- Add testing of ENTSO-E fetch production.

### Double check


VIKTORVAV99:
Turning this into a draft until #5402 is merged

unitrium:
Ready for review now!

parsers/ENTSOE.py
```diff
@@ -1062,66 +1092,15 @@ def fetch_production(
...
-                if v < 0 and v > -50:
-                    # Set small negative values to 0
-                    logger.warning(
-                        "Setting small value of %s (%s) to 0." % (k, v),
-                        extra={"key": zone_key},
-                    )
-                    d["production"][k] = 0
-
+    parsed = parse_production(raw_production, logger, zone_key)
+    data = parsed.to_list()
```
VIKTORVAV99, line 1103: Looks like somewhere along the line we convert the datetime objects to timestamps, which don't work with the test_parser at least.

parsers/ENTSOE.py
```diff
@@ -736,44 +744,72 @@ def parse_scalar(
...
-                if is_production:
-                    productions[i][psr_type] += quantity
-                elif psr_type in ENTSOE_STORAGE_PARAMETERS:
-                    # Only include consumption if it's for storage. In other cases
-                    # it is power plant self-consumption which should be ignored.
-                    productions[i][psr_type] -= quantity
-            except ValueError:  # Not in list
-                datetimes.append(datetime)
-                productions.append(defaultdict(lambda: 0))
-                productions[-1][psr_type] = quantity if is_production else -1 * quantity
```
VIKTORVAV99, line 775: I do like the idea of moving up the validation and such here so perhaps the add_value() method should be implemented here if at all and return a production and storage mix.

VIKTORVAV99 (changes requested):
I'm not totally sold on creating a full production object for all the production modes and merging those instead of simply looping over them as before. It seems to add unneeded complexity even though it probably works very well.

We need to take care of the datetime issue before moving ahead either way.

parsers/ENTSOE.py
unitrium: Has been fixed with #5542

parsers/ENTSOE.py
unitrium: I'm actually not super sure about it. Since we are looping over all fuel timeseries, I would feel more safe using the merge production outputs to make sure that timestamps are aligned.
In the EIA, we saw that the add method wouldn't work because the timestamps are not all aligned.

VIKTORVAV99: Don't ENTSO-E TP normalize the times to the MTU resolution in UTC?
So they should all always be for the same hour or same 15 minutes depending on the country. But those should never be mixed.

unitrium: I hope they do it, but to be on the safe side, I rather avoid trusting the source to do transformations for us.
My reasoning here is that we iterate per fuel report, so it seems more efficient to match the data form when parsing it and merging it as a final step. Otherwise it would require adding some logic in the ENTSOE parser to find back the correct date to which we should add this mode. Since we have the logic for merging lists, I would rather use this.

VIKTORVAV99: Okay that makes sense. I'll take a deep look at the changes in here today.

VIKTORVAV99 (commented):
IT-SO currently fails as it is a aggregate of IT-CA and IT-SO (they split up the IT-SO zone in the source).

So I think we need to update the merge_production_outputs (in entsoe.py) to work with the new validation. It's throwing this issue: `Unknown zone: IT-CA (type=value_error)` Probably due to the fact there is no zone config for it?

Preferably we would split up the zone (I think we have enough data now) but we should do that in a separate PR.

unitrium:
Ah yeah, I didn't noticed this zone and function, I will update it.

unitrium:
I pushed additional changes, those remove the need for a `fetch_aggregated_production` function.

VIKTORVAV99:
> fetch_aggregated_production

I'll take a look right away!

VIKTORVAV99 (approved):
LGTM! 🎉

Everything is working as expected and all the tests are passing!

Let's merge this!

PR #5407, opened 2023-05-15
Automatically switch to correct "view mode" (zone/country)

madsnedergaard:
## Description

When clicking a zone in the ranking panel, the view will now switch to the correct mode depending on what type of zone you click. This also works for links, so going to `/zone/SE-SE1` will force the view mode to be "Zone".

### Preview

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/2f0c2c96-32ef-4f2c-8c0f-14a754596c2a

### Double check


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -1,21 +1,31 @@
...
+import { ZoneDataStatus, getHasSubZones, getZoneDataStatus } from './util';
 
 export default function ZoneDetails(): JSX.Element {
   const { zoneId } = useParams();
   const [timeAverage] = useAtom(timeAverageAtom);
   const [displayByEmissions] = useAtom(displayByEmissionsAtom);
+  const [viewMode, setViewMode] = useAtom(spatialAggregateAtom);
+  const isZoneView = viewMode === SpatialAggregate.ZONE;
+  const hasSubZones = getHasSubZones(zoneId);
+  const isSubZone = zoneId ? getCountryName(zoneId) !== '' : true;
```
tonypls, line 28: This is also an option 

```const isSubZone = zoneId.includes('-');```

File: web/src/features/panels/zone/ZoneDetails.tsx [hunk 2]
```diff
@@ -26,6 +36,19 @@ export default function ZoneDetails(): JSX.Element {
...
 
+  useEffect(() => {
+    if (hasSubZones === null) {
+      return;
+    }
+    // When first hitting the map (or opening a zone from the ranking panel),
+    // set the correct matching view mode (zone or country).
+    if (hasSubZones && isZoneView) {
+      setViewMode(SpatialAggregate.COUNTRY);
+    } else if (isSubZone && !isZoneView) {
```
tonypls, line 47: Maybe a little nitpicky but I find an if statement without the else a little easier to read

web/scripts/generateZonesConfig.ts
```diff
@@ -24,7 +24,9 @@ const mergeZones = (): ZonesConfig => {
   );
 
   const zoneFiles = fs.readdirSync(basePath);
-  const filesWithDirectory = zoneFiles.map((file) => `${basePath}/${file}`);
+  const filesWithDirectory = zoneFiles
+    .filter((file) => file.endsWith('.yaml'))
```
tonypls, line 28: safety first :)

tonypls (approved):
Not sure if it's an issue but when I first loaded the branch my local storage state was country view and the map state was zone view 
image

Also the fly to zone doesn't work on the spatial aggregate change from the ranking panel. 

Very nice renaming, optional improvement suggestions 🎉

web/scripts/generateZonesConfig.ts
madsnedergaard: I was very confused why it didn't work, but turns out the `.ds_store` file was throwing everything off 😬

madsnedergaard:
> Not sure if it's an issue but when I first loaded the branch my local storage state was country view and the map state was zone view 

Huh strange, dunno how/why and can't reproduce 🤔 

I have made a preview branch here, can you reproduce it? https://5a83f831.electricitymaps-app.pages.dev

> Also the fly to zone doesn't work on the spatial aggregate change from the ranking panel.

Good catch, fixed! :)

tonypls:
> > Not sure if it's an issue but when I first loaded the branch my local storage state was country view and the map state was zone view
> 
> Huh strange, dunno how/why and can't reproduce 🤔
> 
> I have made a preview branch here, can you reproduce it? [5a83f831.electricitymaps-app.pages.dev](https://5a83f831.electricitymaps-app.pages.dev)
> 
> > Also the fly to zone doesn't work on the spatial aggregate change from the ranking panel.
> 
> Good catch, fixed! :)

Must've just been a quirk when I was running locally from some old state. 

LGTM!

tonypls (approved):

madsnedergaard:
While testing again I realised my idea of using "geometries" as a dependency meant that switching beteen country/zone would run flyTo whatever zone you had selected.
So as an alternative, I have [in this commit](https://github.com/electricitymaps/electricitymaps-contrib/pull/5407/commits/c038f927d03b3052fbbbfaf3fc55b8974e80d03c) added new functionality to instead attempt to find the approximate area to center on: 
- For subzones (when in country-view), we center on the entire country
- For countries (when in zone-view), we center on the first subzone
- For the US, we handle it all a bit differently as the aggregated zone is so huge and the first subzone matching is Alaska 😅 

Here's a new preview for testing: https://c9d8b08f.electricitymaps-app.pages.dev/

tonypls:
I suspect this is related to this issue, it appears the app can now get in a state somewhere in-between zone and country mode. unsure how to replicate, experienced by me and Mathias after leaving app open. It could be due to the app updating in the background, let's monitor and see if it occurs again.

Pasted Graphic 6

PR #5408, opened 2023-05-16
Add merge Exchanges

unitrium:
## Issue

As mentioned on #5350, we also need a way of merging exchanges inside the new list datastructures as this is needed in the ENTSO-E parser at least. As it is looking at imports first and then merging with exports.

## Description

- Adds a static method to ExchangeList to merge outputs together.
- Add a parent class MergeableList to hold common code between ProductionBreakdownList and ExchangeList.
- Add testing of Exchange merges.

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -40,7 +40,49 @@ def to_list(self) -> List[Dict[str, Any]]:
         return [event.to_dict() for event in self.events]
 
 
-class ExchangeList(EventList):
+class MergeableList(EventList, ABC):
```
mathilde-daugy, line 43: The name is not very self explanatory.

unitrium: Any suggestions, nothing better comes to mind.
To me it describes best what this type of list is capable.

unitrium:
```suggestion
class MergeableEventList(EventList, ABC):
```
? But sounds like a repetion since Event is alreedy contained in the parent class.

VIKTORVAV99: My vote is for MergeableEventList. It creates some repetition but I think it's easier to understand. And if we add """comments""" IDEs should make it easy enough to understand?

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -57,8 +99,39 @@ def append(
...
+        ]
+
+        exchange_df = pd.concat(exchange_dfs)
+        exchange_df = exchange_df.rename(columns={"sortedZoneKeys": "zoneKey"})
+        zone_key, sources, source_type = ExchangeList.get_unique_zone_source(
+            exchange_df
+        )
+        exchange_df = exchange_df.groupby(level=0, dropna=False).sum()
+        for datetime, row in exchange_df.iterrows():
+            exchanges.append(zone_key, datetime, sources, row["netFlow"], source_type)
```
VIKTORVAV99, line 129: Looks like datetime is not getting the correct type here, I think we need to cast the type first. Right now it's a lable or hashable when it should be a datetime.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -87,14 +160,11 @@ def merge_production_breakdowns(
...
         There should be only one zone in the list of production breakdowns.
         """
-        if len(ungrouped_production_breakdowns) == 0:
-            return ProductionBreakdownList(logger)
-        if all(
-            len(breakdowns.events) == 0
-            for breakdowns in ungrouped_production_breakdowns
+        production_breakdowns = ProductionBreakdownList(logger)
+        if ProductionBreakdownList.is_completly_empty(
+            ungrouped_production_breakdowns, logger
```
VIKTORVAV99, line 165: Pylance is throwing the following error here:
```Argument of type "List[ProductionBreakdownList]" cannot be assigned to parameter "ungrouped_events" of type "List[MergeableList]" in function "is_completly_empty"
  "List[ProductionBreakdownList]" is incompatible with "List[MergeableList]"
    TypeVar "_T@list" is invariant
      "ProductionBreakdownList" is incompatible with "MergeableList"PylancereportGeneralTypeIssues
```
So I think we need to accept the input types for the is_completly_empty function to accept the ProductionBreakdownList and ExchangeList instead.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -57,8 +99,39 @@ def append(
...
+            for exchanges in ungrouped_exchanges
+            if len(exchanges.events) > 0
+        ]
+
+        exchange_df = pd.concat(exchange_dfs)
+        exchange_df = exchange_df.rename(columns={"sortedZoneKeys": "zoneKey"})
+        zone_key, sources, source_type = ExchangeList.get_unique_zone_source(
+            exchange_df
+        )
+        exchange_df = exchange_df.groupby(level=0, dropna=False).sum()
```
VIKTORVAV99, line 127: I assume this will work if there none/null events in on of the exchange lists, but should we not log this so we know if we are potentially missing data and should try and re-fetch it later?

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -40,7 +40,49 @@ def to_list(self) -> List[Dict[str, Any]]:
...
 
 
-class ExchangeList(EventList):
+class MergeableList(EventList, ABC):
+    """A wrapper around Events lists that can be merged."""
+
+    @classmethod
+    def is_completly_empty(
+        cls, ungrouped_events: List["MergeableList"], logger: Logger
+    ) -> "MergeableList":
```
VIKTORVAV99, line 49: The return signature here should be a bool.

electricitymap/contrib/lib/tests/test_event_lists.py
```diff
@@ -47,6 +47,27 @@ def test_append_to_list_logs_error(self):
...
+            zoneKey=ZoneKey("AT->DE"),
+            datetime=datetime(2023, 1, 1, tzinfo=timezone.utc),
+            netFlow=2,
+            source="trust.me",
+        )
+        exchanges = ExchangeList.merge_exchanges(
+            [exchange_list_1, exchange_list_2], logging.Logger("test")
+        )
+        assert len(exchanges) == 1
+        assert exchanges.events[0].netFlow == 3
```
VIKTORVAV99, line 69: I think I would like one more test to check the outcome if one of these are None/Null/N/A etc. Just to make sure we have the test and avoid unexpected changes if we ever decide to refractor things in the future.

VIKTORVAV99 (commented):
Bunch of small comments in regards to types plus some general questions and suggestions.

Will be nice to get this added!

electricitymap/contrib/lib/models/event_lists.py
unitrium: I have an issue with that approach as it means that the parent class is aware of what the child is doing which is not a good idea. I will see if I can fix it somehow.

electricitymap/contrib/lib/models/event_lists.py
unitrium: I think we need to ignore it anyways as we did in the merge production method.

unitrium: I don't think the cast is going to properly work.

electricitymap/contrib/lib/models/event_lists.py
unitrium: Apparently using Sequence is the best fix.
https://stackoverflow.com/questions/69891672/python-typing-how-to-handle-list-type-hinting-with-different-element-type-in-su

electricitymap/contrib/lib/tests/test_event_lists.py
unitrium: I added a test for NaN and negative values as well.

electricitymap/contrib/lib/models/event_lists.py
unitrium: This will be logged when the list you want to merge is created in the first place.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -40,7 +43,49 @@ def to_list(self) -> List[Dict[str, Any]]:
...
 
-class ExchangeList(EventList):
+class MergeableEventList(EventList, ABC):
+    """A wrapper around Events lists that can be merged."""
+
+    @classmethod
+    def is_completely_empty(
+        cls, ungrouped_events: Sequence["MergeableEventList"], logger: Logger
+    ) -> bool:
+        """Merge multiple lists of events into one."""
```
VIKTORVAV99, line 53: Can we update this to be more descriptive of just this function, maybe something like:

```suggestion
        """Checks if the lists to be merged has any data."""
```

What do you think?

VIKTORVAV99 (approved):
I added a suggestion to change the docstring but beyond that this looks good to me.

I also took the liberty to format the files and fix a small spelling error. I hope you don't mind.

electricitymap/contrib/lib/models/event_lists.py
unitrium: Yes I have added it now.

unitrium:
As discussed with @mathilde-daugy offline we decided to rename it to AggregatableEventList.
Will merge it after her review.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -40,7 +43,49 @@ def to_list(self) -> List[Dict[str, Any]]:
...
+    """An abstract class to supercharge event lists with aggregation capabilities."""
+
+    @classmethod
+    def is_completely_empty(
+        cls, ungrouped_events: Sequence["AggregatableEventList"], logger: Logger
+    ) -> bool:
+        """Checks if the lists to be merged has any data."""
+        if len(ungrouped_events) == 0:
+            return True
+        if all(len(exchanges.events) == 0 for exchanges in ungrouped_events):
```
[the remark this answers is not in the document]
unitrium: An oversight, this should be abstracted indeed.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -40,7 +43,49 @@ def to_list(self) -> List[Dict[str, Any]]:
...
+            logger.warning(f"All {cls.__name__} are empty.")
+            return True
+        return False
+
+    @classmethod
+    def get_unique_zone_source(
```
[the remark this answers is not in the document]
unitrium: Indeed, I changed it and split the classmethod into several chuncks.

pierresegonne (approved):
Looks better :)

Issue #4180, opened 2022-05-23
Create parser for Albania (source is available)


autipial:
Carbon intensity of the electricity produced in Albania is currently 461 gCO2eq/kWh. I suspect it could be wrong, since 100% of Albania's electricity come from renewable sources (mainly hydro + a small share of solar). 

The only thermal plant of the country ([TEC Vlora](http://www.kesh.al/en/asset/tec-vlora)) has never been functioning.

The installed capacity (listing only hydro though) is included in the daily data published by OST ([https://www.ost.al/te-dhena](https://www.ost.al/te-dhena)).

jarek:
I think Albania's value never got specified (I don't see it in `zoneOverrides` in https://github.com/electricitymap/electricitymap-contrib/blob/master/config/co2eq_parameters_all.json ) so the 461 g might be the default value for worldwide averages (it might be computed from the values in `fallbackZoneMixes -> defaults -> powerOriginRatios` in that file?).

If you or someone could link to recently monthly or yearly production data for Albania, we could add an override for this value.

In fact looks like it was previously added in https://github.com/electricitymap/electricitymap-contrib/pull/1264 but must have gotten lost somewhere along the way.

jarek:
https://www.iea.org/countries/albania says 2020 production was 5281 GWh hydro and 32 GWh solar which is pretty clear cut.

However, the issue of exchanges was raised in #1264. IEA 2019 data has yearly total of 11435 TJ = 3176 GWh imports and 2774 TJ = 770 GWh exports, with 5184 GWh own hydro generation. So imports can make up a significant part of electricity supply in Albania, at least in some years (for example in 2018 imports were 850 GWh and hydro production was 8552 GWh - so it is quite variable). That's a bit of a problem if Electricity Map doesn't know Albania's current generation, because it's hard to estimate carbon intensity of what's being exported: when Albania is exporting some power, is it exporting mostly Albanian hydro, or half hydro and half Montenegrin coal?

autipial:
Thanks @jarek.
Yearly production data can be found on the website of ERE (energy regulator of Albania). For 2021: [https://ere.gov.al/images/2022/04/Raporti%20Vjetor%202021%20perfundimtare.pdf](https://ere.gov.al/images/2022/04/Raporti%20Vjetor%202021%20perfundimtare.pdf).

- Total generation capacity for Albania as of 31/12/2021 is 2605 MW (p. 29).
[...]

PR #5416, opened 2023-05-18
Albania Consumption and ProductionPerUnit parser

VIKTORVAV99:
## Issue

Closes: #4180

## Description

This PR adds a consumption and production per unit parser for Albania in the hope that we can use this data to estimate/recreate the production breakdown in Albania. This is important as it often exports power to Montenegro, Kosovo and Greece and would significantly lower the carbon intensity of the first two zones during times of high exports.

Things to note: The data is significantly delayed by roughly 40 days.

### Double check


mathilde-daugy:
@VIKTORVAV99, thanks for your work on this PR. I'm not sure that we use production per unit anywhere so I think it would be more relevant to create a power plant mapping and then map each PP to a specific mode to get the power breakdown.

Let me know if this is a lot of extra work, I can help out next week :)

VIKTORVAV99:
> @VIKTORVAV99, thanks for your work on this PR. I'm not sure that we use production per unit anywhere so I think it would be more relevant to create a power plant mapping and then map each PP to a specific mode to get the power breakdown.
> 
> Let me know if this is a lot of extra work, I can help out next week :)

It is in ENTSO-E parser (which I based the structure on) but I don't know if it's actually used anywhere. The reason I didn't map it all to just hydro is because it's just the biggest plants and they don't represent all of the production as far as I can tell. The other production sources should also be almost 100% hydro with a small amount of solar though.

mathilde-daugy:
I have just checked and the only parsers that have production per unit are DK and SE. This data doesn't appear anywhere in the database though so we're not making any use of it. We only collect production per mode :)

VIKTORVAV99:
@mathilde-daugy I found another table in the data that has total generation and together with the named plants this means we can get the defined hydro production from the major plants and then calculate the unknown amount.
That would look like this for 16-03-2023:
| Hour | Sum named hydro power plants | Total   generation | Unknown |
|:----:|:----------------------------:|:------------------:|:-------:|
|   1  |              171             |         616        |   445   |
|   2  |              53              |         549        |   496   |
|   3  |               0              |         548        |   548   |
|   4  |               0              |         536        |   536   |
|   5  |               0              |         543        |   543   |
|   6  |               2              |         622        |   620   |
|   7  |              139             |         843        |   704   |
|   8  |              267             |        1155        |   888   |
|   9  |              353             |        1220        |   867   |
|  10  |              476             |        1228        |   752   |
|  11  |              384             |        1086        |   702   |
|  12  |              398             |         950        |   552   |
|  13  |              403             |         890        |   487   |
|  14  |              396             |         885        |   489   |
|  15  |              388             |         885        |   497   |
|  16  |              389             |         885        |   496   |
|  17  |              406             |        1036        |   630   |
|  18  |              482             |        1237        |   755   |
|  19  |              528             |        1371        |   843   |
|  20  |              528             |        1401        |   873   |
|  21  |              500             |        1385        |   885   |
|  22  |              419             |        1296        |   877   |
|  23  |              205             |        1039        |   834   |
|  24  |              113             |         794        |   681   |

Now according to IEA hydro production accounted for more than 99.9% of the generation in Albania so even the unknown generation should be hydro + a tiny bit solar.

So we have to decide on how we want to do this. Either we map all generation to hydro and call it a day, that would be the easiest option.

Or we map the known hydro to hydro and the rest to unknown and then calculate a new carbon emission factor for unknown that will be pretty much the same as hydro. But that way we don't make any assumptions about the production sources. Which may be better for accounting purposes later on, especially if the unknown share increases.

Feel free to ask anyone else on the team if they have any opinions about it as well.

mathilde-daugy:
I'll have a look at the data source and let you know. I think we can do the first option as i'm pretty sure that it will not make a big impact but i'll let you know later today

VIKTORVAV99:
Before implementing the changes we discussed in slack I took a look at the load data in ENTSO-E and the consumption data in the parser. These are vastly different so I think we need to investigate which one to use, the ENTSO-E values seems unrealistically low but we need to do a proper comparison.

I also checked the production per units in ENTSO-E, the units are included but all data is N/A so we can't use that, it would have made mapping the plants a lot easier and more robust.

Lastly I reached out to ENTSO-E about the status of getting the production information per source into the transparency platform and they should get back to me soon, however I think it might be best to hold of making any complicated parser changes until they do.

VIKTORVAV99:
Just to provide an update:

ENTSO-E got back to me and they have reached out to the TSO and will provide an update when they hear back from them.

VIKTORVAV99:
Closing this for now, we have discussed another solution for this internally.

PR #5420, opened 2023-05-20
Split expected emission factors into direct and lifecycle

qyearsley:
## Issue

Related to #3990

## Description

Splits the expected range values into direct and lifecycle.

Note that in the case of test failure, the output does have information whether the lifecycle or direct test case failed, although it may not be obvious from the message.

### Double check

- [] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.


tests/test_co2eq_parameters.py
```diff
@@ -254,8 +254,12 @@ class CO2eqParametersDirectAndLifecycleMixin:
...
-        """Apply the callback to each emission factor in the 'emissionFactors'
-        object.
+        """Apply the callback to each item in the 'emissionFactors' object.
+
+        The callback is called with the mode and factors for both the defaults
+        and the zone-specific overrides; each `factors` object is either a list
+        of factor dicts or a single factor dict, and each factor dict has the
+        key "value", and possibly other keys such as "source".
         """
         emission_factors = cls.parameters["emissionFactors"]
```
VIKTORVAV99, line 264: Pylance is reporting that parameters is unknown on the type `Type[CO2eqParametersDirectAndLifecycleMixin]` so we have to take care of that if we want to add any form of type checking in the future.

tests/test_co2eq_parameters.py
```diff
@@ -277,35 +299,12 @@ def test_emission_factor_value_ranges(self):
...
-            # Emissions are counted at discharge.
-            "battery charge": (0, 0),
-            "hydro charge": (0, 0),
-        }
-
         def check_range(mode, factor, zone):
             value = factor["value"]
-            low, high = ranges_by_mode[mode]
+            assert isinstance(value, (int, float))
+            low, high = self.ranges_by_mode[mode]
```
VIKTORVAV99, line 305: Same for ranges_by_mode as above.

tests/test_co2eq_parameters.py
```diff
@@ -384,6 +383,28 @@ class CO2eqParametersDirect(CO2eqParametersDirectAndLifecycleMixin, unittest.Tes
 
     parameters = CO2EQ_PARAMETERS_DIRECT
 
+    # Expected min and max values for emission factors, by mode.
+    ranges_by_mode = {
```
VIKTORVAV99, line 387:
```suggestion
    ranges_by_mode: Dict[str, Tuple[Union[int, float], Union[int, float]]] = {
```
You might have to update the imports from typing after adding this.

tests/test_co2eq_parameters.py
```diff
@@ -392,6 +413,28 @@ class CO2eqParametersLifecycle(
 
     parameters = CO2EQ_PARAMETERS_LIFECYCLE
 
+    # Expected min and max values for emission factors, by mode.
+    ranges_by_mode = {
```
VIKTORVAV99, line 417:
```suggestion
    ranges_by_mode: Dict[str, Tuple[Union[int, float], Union[int, float]]] = {
```

VIKTORVAV99 (commented):
The tests work as expected but we need to take a look at the types in the test itself, there are a lot of type errors in here.

I also noted that we still rely on the .ts file to get the possible modes. We should probably update this whole test to use the constants from the python parts instead so we decuple this from web.

The python constants are located at `electricitymap/contrib/config/constants`

tests/test_co2eq_parameters.py
qyearsley: Done; note that I discovered that type `float` sort of includes `int`; and there's also type numbers.Number which is already imported that includes float, int and other number types.

https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

tests/test_co2eq_parameters.py
qyearsley: Done

tests/test_co2eq_parameters.py
qyearsley: Done.

qyearsley: Note that many of the warnings were missing assert methods since the mixin class was not a subclass of TestCase.

One of the simplest ways to deal with this was to make it a subclass of TestCase; after doing that however it is found automatically by the test runner so to prevent unexpected behavior, in the latest revision I nested it inside of another class, like in https://stackoverflow.com/questions/1323455/python-unit-test-with-base-and-sub-class

qyearsley:
This doesn't yet resolve 100% of the issues flagged by pyright/pylance -- I propose resolving more of these issues in a separate PR.

PR #5421, opened 2023-05-20
[WIP] Remove nuclear from validation for Germany

mueslo:
## Issue

Ever since a few days ago Germany is only ever showing "estimated" data. This seems to be because ENTSO-E is only reporting Geothermal, Wind (offshore), and Brown coal / Lignite.  In any case, it should no longer be required for the nuclear key to be present for German output.

## Description

Remove 'nuclear' as a required ENTSO-E output for Germany

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -359,7 +359,6 @@ VALIDATIONS: Dict[str, Dict[str, Any]] = {
         "required": [
             "coal",
             "gas",
-            "nuclear",
             "wind",
             "biomass",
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Nuclear is still expected to be reported for Germany, even if it's 0, as it would be a indicator that there are data missing if it's missing.

We are working on improving the validation and will likely change how things work then so we still keep valid data and mark the other expected values as missing in some manner. But until that is fully shaped and implemented I don't think this change is necessary.

PR #5424, opened 2023-05-22
fix: IL parser

### LINKED ISSUE #4288: IL production parser down
## Description

This is an automatic error report generated for Israel (IL).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser IL production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/IL.html">Explore the runtime logs</a>

GuyNachshon:
## Issue

Closes #4288 

## Description
IL parser went down after the site redesign. I reached out to the person in charge of public information and was directed to [Noga](https://www.noga-iso.co.il/en/). this parser uses the API endpoint they use on their site.

**important to note**
currently, all of the consumption and production are labelled as `unknown` since there is no distribution map. *but* there is a consumption and production division between renewable energy and "regular energy" in MW [here](https://www.noga-iso.co.il/demandforecast). If you think this division is relevant, let me know, and I can add it.

### Double check

- [ X] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [X ] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

### DIFF
--- a/parsers/IL.py
+++ b/parsers/IL.py
@@ -1,8 +1,8 @@
 """
 Israel's electricity system load parser.
 
-Source: Israel Electric Corporation
-URL: https://www.iec.co.il/en/pages/default.aspx
+Source: Noga - The Israel Independent System Operator Ltd.
+URL: https://www.noga-iso.co.il/en/
 
 Shares of Electricity production in 2019:
     64.0% Gas
@@ -16,48 +16,24 @@ import re
 from datetime import datetime
 from logging import Logger, getLogger
 from typing import Optional
-
+import base64
 import arrow
-from bs4 import BeautifulSoup
 from requests import Session, get
 
-IEC_URL = "www.iec.co.il"
+NOGA_BASE_URL = "https://www.noga-iso.co.il/"
 IEC_PRODUCTION = (
     "https://www.iec.co.il/_layouts/iec/applicationpages/lackmanagment.aspx"
 )
-IEC_PRICE = "https://www.iec.co.il/homeclients/pages/tariffs.aspx"
+NOGA_ELECTRICITY_CONSUMPTION_API = f"{NOGA_BASE_URL}/Umbraco/Api/Documents/GetElectricalData"
+IEC_PRICE = "https://iecapi.iec.co.il//api/content/he-IL?pageRoute=content/tariffs/contentpages/businesselectricitytariff"
 TZ = "Asia/Jerusalem"
 
 
-def fetch_all() -> list:
-    """Fetch info from IEC dashboard."""
-    first = get(IEC_PRODUCTION)
-    first.cookies
-    second = get(IEC_PRODUCTION, cookies=first.cookies)
-    soup = BeautifulSoup(second.content, "lxml")
-
-    values: list = soup.find_all("span", class_="statusVal")
-    if len(values) == 0:
-        raise ValueError("Could not parse IEC dashboard")
-    del values[1]
-
-    cleaned_list = []
-    for value in values:
-        value = re.findall(r"\d+", value.text.replace(",", ""))
-        cleaned_list.append(value)
-
-    def flatten_list(_2d_list) -> list:
-        """Flatten the list."""
-        flat_list = []
-        for element in _2d_list:
-            if type(element) is list:
-                for item in element:
-                    flat_list.append(item)
-            else:
-                flat_list.append(element)
-        return flat_list
-
-    return flatten_list(cleaned_list)
+def fetch_all() -> dict:
+    """Fetch info from Noga API."""
+    with get(NOGA_ELECTRICITY_CONSUMPTION_API) as response:
+        data = response.json()
+    return data
 
 
 def fetch_price(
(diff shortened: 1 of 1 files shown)


parsers/IL.py
```diff
@@ -109,14 +106,23 @@ def fetch_production(
...
+
+    production = production.replace(',', '')
+    reserve = reserve.replace(',', '')
+
+    production = float(production) * 1000 + float(reserve) * 1000
 
     # all mapped to unknown as there is no available breakdown
     return {
         "zoneKey": zone_key,
         "datetime": arrow.now(TZ).datetime,
```
q--, line 123: NOGA has the timestamp on their web site (e.g.  Last update: 23/05/2023 19:04:43 ); it would be great if the parser could use that instead of the current timestamp

parsers/IL.py
```diff
@@ -66,24 +42,45 @@ def fetch_price(
...
+        raise ValueError("Could not parse IEC price")
+    general_message = components[0].get("text")
+    general_message = base64.b64decode(general_message).decode("utf-8")
+
+    date = extract_date_from_string(general_message)
+
+    table = components[1].get('table', [])
+    kw_prices = table[-1]
+    price_with_VAT = kw_prices[-1].get("value")
+    price_with_VAT = base64.b64decode(price_with_VAT).decode("utf-8")
```
VIKTORVAV99, line 63: We should check that kw_prices are not None or N/A before passing it to the decode function. 

I would also avoid defining the price_with_VAT and re-assigning it directly after, this can be inlined.

parsers/IL.py
```diff
@@ -66,24 +42,45 @@ def fetch_price(
...
 
     return {
         "zoneKey": zone_key,
         "currency": "NIS",
-        "datetime": extract_price_date(soup),
-        "price": float(price.p.text),
-        "source": IEC_URL,
+        "datetime": date,
+        "price": float(price_with_VAT),
+        "source": IEC_PRICE,
```
VIKTORVAV99, line 70: We don't want the full URL here, only the domain and subdomains.

parsers/IL.py
```diff
@@ -16,48 +16,24 @@
...
 from requests import Session, get
 
-IEC_URL = "www.iec.co.il"
+NOGA_BASE_URL = "https://www.noga-iso.co.il/"
 IEC_PRODUCTION = (
     "https://www.iec.co.il/_layouts/iec/applicationpages/lackmanagment.aspx"
 )
-IEC_PRICE = "https://www.iec.co.il/homeclients/pages/tariffs.aspx"
+NOGA_ELECTRICITY_CONSUMPTION_API = f"{NOGA_BASE_URL}/Umbraco/Api/Documents/GetElectricalData"
+IEC_PRICE = "https://iecapi.iec.co.il//api/content/he-IL?pageRoute=content/tariffs/contentpages/businesselectricitytariff"
```
VIKTORVAV99, line 28:
```suggestion
IEC_PRICE = "https://iecapi.iec.co.il/api/content/he-IL?pageRoute=content/tariffs/contentpages/businesselectricitytariff"
```
Should probably be just one slash here right?

parsers/IL.py
```diff
@@ -66,24 +42,45 @@ def fetch_price(
...
 
 
+def extract_date_from_string(general_message: str) -> datetime:
+    regexp = r"(\d{2}.\d{2}.\d{4})"
+    match = re.search(regexp, general_message)
+    if match:
+        date_str = match.group(1)
+        return datetime.strptime(date_str, "%d.%m.%Y")
+    else:
+        raise ValueError("Could not parse IEC price date")
```
VIKTORVAV99, line 81: Same here with the parser exception.

parsers/IL.py
```diff
@@ -109,14 +106,27 @@ def fetch_production(
         raise NotImplementedError("This parser is not yet able to parse past dates")
 
     data = fetch_all()
-    production = [float(item) for item in data]
+    production = data.get("Production")
+    reserve = data.get("CurrentReserve")
+
+    if not production or not reserve:
+        raise ValueError("Could not parse Noga production")
```
VIKTORVAV99, line 113: And here.

parsers/IL.py
```diff
@@ -109,14 +106,27 @@ def fetch_production(
...
+
+    date = data.get("PeakReserveDate")
+    time = data.get("PeakReserveTime")
+    date_time = arrow.get(f"{date} {time}", "DD/MM/YYYY HH:mm").datetime
+
+    production = production.replace(',', '')
+    reserve = reserve.replace(',', '')
+
+    production = float(production) * 1000 + float(reserve) * 1000
```
VIKTORVAV99, line 122:
```suggestion
    production = (float(production) + float(reserve)) * 1000
```

parsers/IL.py
```diff
@@ -109,14 +106,27 @@ def fetch_production(
...
-    production = [float(item) for item in data]
+    production = data.get("Production")
+    reserve = data.get("CurrentReserve")
+
+    if not production or not reserve:
+        raise ValueError("Could not parse Noga production")
+
+    date = data.get("PeakReserveDate")
+    time = data.get("PeakReserveTime")
+    date_time = arrow.get(f"{date} {time}", "DD/MM/YYYY HH:mm").datetime
```
VIKTORVAV99, line 117: We should make sure this generates valid datetimes before passing it to the production object.

parsers/IL.py
```diff
@@ -130,14 +140,19 @@ def fetch_consumption(
...
     data = fetch_all()
-    consumption = [float(item) for item in data]
-
-    # all mapped to unknown as there is no available breakdown
+    consumption = data.get("Production")
+    if not consumption:
+        raise ValueError("Could not parse Noga consumption")
+    date = data.get("PeakReserveDate")
+    time = data.get("PeakReserveTime")
+    date_time = arrow.get(f"{date} {time}", "DD/MM/YYYY HH:mm").datetime
```
VIKTORVAV99, line 148: Same here as above and it looks like this could be turned into a function used both above and here.

VIKTORVAV99 (commented):
Hi and thanks for the PR!

I left a lot of comment that mostly relates to making sure we get the expected values and log a message if we don't.

Take a extra look at the datetimes though, they need to be returned as datatime.datetime objects and the ones used for the production and consumption parser can be merged into one function.

GuyNachshon:
Cool, I'll get to it tomorrow. Regarding the structure and comments, I tried to leave as it was in the previous version, but I'll change it 😄

parsers/IL.py
GuyNachshon: Actually this is how they do the request for some reason 🤷🏽‍♀️
I'll check to see if it matters

GuyNachshon: Works both ways. changing for convinience

parsers/IL.py
```diff
@@ -66,24 +42,45 @@ def fetch_price(
...
         raise NotImplementedError("This parser is not yet able to parse past dates")
 
     with get(IEC_PRICE) as response:
-        soup = BeautifulSoup(response.content, "lxml")
+        data = response.json()
 
-    price = soup.find("td", class_="ms-rteTableEvenCol-6")
+    components = data.get("components")
+    if not components:
+        raise ValueError("Could not parse IEC price")
```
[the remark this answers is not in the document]
GuyNachshon: I'm not aware of a general parser error in Python, they only exist inside open-source packages that implement them. can you direct me towards what you mean?

parsers/IL.py
```diff
@@ -44,15 +60,19 @@ def fetch_price(
...
     if target_datetime is not None:
+        logger.warning(
+            "IL parser is not yet able to parse past dates, ignoring target_datetime"
+        )
         raise NotImplementedError("This parser is not yet able to parse past dates")
-
-    with get(IEC_PRICE) as response:
-        data = response.json()
+    if session is None:
+        session = Session()
```
VIKTORVAV99, line 68:
```suggestion
    session = session or Session()
```

parsers/IL.py
```diff
@@ -44,15 +60,19 @@ def fetch_price(
...
+    response = session.get(IEC_PRICE)
+    data = response.json()
```
VIKTORVAV99, line 70:
```suggestion
    response: Response = session.get(IEC_PRICE)
    if not response.ok:
        raise ParserException("IL.py", f"The returned response was not okay and returned the status: {response.status_code}", zone_key)
    data = response.json()
```

Here is an example on how to use ParserException as well.

VIKTORVAV99 (commented):
Adding some more comments but it seems all previous comments have not been addressed yet, will do a full review when all current comments have been addressed.

PR #5433, opened 2023-05-30
Fetch production for IN-SO

unitrium:
## Issue

We are kickstarting the development of IN-SO. Therefore we should start fetching production data for IN-SO.

## Description

Activates fetch production for IN-SO.

## Additional notes
We will hide the zone data on the map until the estimation model is ready to be used.

### Double check

### DIFF
--- a/config/zones/IN-SO.yaml
+++ b/config/zones/IN-SO.yaml
@@ -19,3 +19,4 @@ contributors:
 timezone: Asia/Kolkata
 parsers:
   consumption: IN.fetch_consumption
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Do we want to parse production data for IN-EA and IN-NE as well and disable them on the backend or should we hold off on that?

unitrium:
I would rather tackle one issue at a time and first focus on developing IN-SO, if everything goes well we can continue with the rest later :)

mathilde-daugy (approved):

PR #5436, opened 2023-06-01
Update FI.yaml

pnook:
Updated with latest wind capacity according to Fingrid ( https://www.fingrid.fi/en/electricity-market-information/wind-power-generation/ )

## Issue

Outdated wind power capacity

## Description

Update with latest capacity of wind power in Finland

### Preview

n/a

### Double check

### DIFF
--- a/config/zones/FI.yaml
+++ b/config/zones/FI.yaml
@@ -13,7 +13,7 @@ capacity:
   oil: 1051
   solar: 697
   unknown: 828
-  wind: 5575
+  wind: 5700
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
Thanks for updating the capacity!

I would just like to note that this source is not the same as we already use, but I double checked with the source we do use and it also has the updated value. But in the future if the source changes we also need to update the DATA_SOURCES.md file.

But since this match I'll approve and merge it right away.

PR #5439, opened 2023-06-02
Use UTC when labeling non-hourly time slider values

### LINKED ISSUE #5339: [Bug]: Index issue with yearly historical view
### When did this happen?

2023-04-28

### What zones are affected?

Germany

### What is the problem?

According to electricitymaps, German gCO₂eq/kWh did rise from 397 in 2019 to 447 in 2020. Mostly due to higher coal consumption. 

According to official data: https://www.destatis.de/EN/Press/2021/03/PE21_101_43312.html

Electricity from coal declined in 2020 compared to 2019 and was overtaken for the first time by windenergy.

What's the cause of the difference between both data sets?

sporadicus:
## Issue

Closes #5339

## Description

The time control interface contains 4 scopes of time durations for which the user may view averages: hourly (default), daily, monthly, and yearly. The user-selectable time values that appear in the slider for each of the above 4 time scopes are shown in the user's local timezone. However, when the chosen scope is daily, monthly, or yearly, the resulting user-selectable time values are incorrect for users whose timezone offset from UTC is a negative value (e.g. New York's UTC-05:00).

This fix causes the slider time values for the daily, monthly, and yearly scopes to use UTC for all users instead of their local timezone.

### Double check

- [x] I have run `npx prettier --check .` to ensure that my changes respect ElectricityMaps' contrib code formatting


VIKTORVAV99:
Sorry for taking some time on this, I was hoping we could add some tests as well but there don't seem to be a clear cut way to manipulate the timezone in vitest. If I can't find a way to do so by tomorrow I'll approve and merge this without tests.

PR #5441, opened 2023-06-04
Add wind data to BD (Bangladesh)

VIKTORVAV99:
## Issue

The source was recently updated with wind data and the table layout changed so the parser is currently failing.

## Description

Adds wind data and adjust the table parsing to adjust for the alteration in the source layout.

### Double check


unitrium (approved):
Thanks for fixing it ! Will deploy it tomorrow :)

VIKTORVAV99:
Can you run a backfill as well?

unitrium:
I ran a refetch, we will start a big flowtracing later with the changes from India as well.

VIKTORVAV99:
> I ran a refetch, we will start a big flowtracing later with the changes from India as well.

There should be data available from early 2015 if needed.

PR #5443, opened 2023-06-05
Small updates to geo files to reduce duplicate points and gaps

VIKTORVAV99:
## Description

To reduce duplicate points and gaps so the world generation works better and becomes smaller because it can reuse more lines/features.

There is still some cleanup and adjustments that need to be made but I thought it best to do these in small batches.

### Preview
Old:
image
New:
image
Old:
image
New:
image

### Double check


madsnedergaard (approved):
Nice! 🙌 
How did you go about doing this? :)

PR #5446, opened 2023-06-06
fix IN-EA_IN-SO exchange config

VIKTORVAV99:
## Issue

There was a extra parser field which cause the tests to fail.

## Description

Removes the extra field.


unitrium (approved):

unitrium:
Oh thanks! Don't know how that got in 😅

PR #5451, opened 2023-06-08
[WIP]: Upgrade ES parser to EventList structure

VIKTORVAV99:
## Issue

It still use the old no EventList structure.

## Description

Upgrades it to use the new EventList structure.

### Double check

### DIFF
--- a/parsers/ES.py
+++ b/parsers/ES.py
@@ -2,10 +2,12 @@
 
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
-from typing import Any, Callable, Dict, List, Literal, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 # The arrow library is used to handle datetimes
 from arrow import get
+from electricitymap.contrib.lib.models.event_lists import TotalConsumptionList
+from electricitymap.contrib.lib.types import ZoneKey
 
 # package "ree" is used to parse data from www.ree.es
 # maintained on github by @hectorespert at https://github.com/hectorespert/ree
@@ -37,7 +39,7 @@ from .lib.validation import validate
 SOURCE = "demanda.ree.es"
 
 # Literal list of valid zone keys for this parser
-ZONE_KEYS = Literal[
+SUPPORTED_ZONE_KEYS: List[ZoneKey] = [
     "ES",  # Spain
     "ES-CE",  # Ceuta
     "ES-CN-FVLZ",  # Fuerteventura/Lanzarote
@@ -57,7 +59,7 @@ ZONE_KEYS = Literal[
 # TODO: Update floors to be non zero.
 # Minimum valid zone demand. This is used to eliminate some cases
 # where generation for one or more modes is obviously missing.
-ZONE_FLOORS: Dict[ZONE_KEYS, float] = {
+ZONE_FLOORS: Dict[ZoneKey, float] = {
     "ES": 0,
     "ES-CE": 0,
     "ES-CN-FVLZ": 50,
@@ -74,7 +76,7 @@ ZONE_FLOORS: Dict[ZONE_KEYS, float] = {
     "ES-ML": 0,
 }
 
-ZONE_FUNCTION_MAP: Dict[ZONE_KEYS, Callable] = {
+ZONE_FUNCTION_MAP: Dict[ZoneKey, Callable] = {
     "ES": IberianPeninsula,
     "ES-CE": Ceuta,
     "ES-CN-FVLZ": LanzaroteFuerteventura,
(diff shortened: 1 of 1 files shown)


electricitymap/contrib/lib/models/events.py
VIKTORVAV99, line 1: The parser changes should be pretty straight forward even if they are relatively big.
Therefore I would appreciate if we focused on the changes in this file as these are the ones I am a little unsure of.

electricitymap/contrib/lib/models/events.py
```diff
@@ -119,6 +137,30 @@ def set_value(
             self._corrected_negative_values.add(mode)
         self.__setattr__(mode, value)
 
+    def add_value(
```
unitrium, line 140: I see a potential edge case here with negative values.

Imagine that we have the existing_value set to 2 and we get a new value with -1.
In this case we would call self.set_value with 2+(-1) = 1. Therefore we don't trigger the correct negative values function, which I think we should do. Also in this way we are mixing the self consumption from one mode with another which I don't think is super clear. But I acknowledge that how we always did it, so not sure it's that big of an issue.

unitrium: In my opinion it would make more sense to correct for each endpoint added individually before adding them together.

electricitymap/contrib/lib/models/events.py
```diff
@@ -24,6 +24,24 @@ def set_value(self, mode: str, value: Optional[float]) -> None:
         """
         self.__setattr__(mode, value)
 
+    def add_value(
```
unitrium, line 27: I'm wondering if it wouldn't be better to use the merge production outputs static method for this purpose. Though the PR for negative values is still on hold #5402.

parsers/ES.py
```diff
@@ -30,91 +27,113 @@
...
+
+def check_valid_parameters(
+    zone_key: ZoneKey,
+    session: Optional[Session],
+    target_datetime: Optional[datetime],
+    logger: Logger,
+    exchange: bool = False,
+):
+    """Raise an exception if the parameters are not valid for this parser."""
+    if exchange:
```
unitrium, line 91: Instead of using a boolean attribute, I think it would be better to use the supplied zone_key and check if it is in the exchange config.
This way we avoid making mistakes of saying this is an exchange while it is not.

parsers/ES.py
```diff
@@ -127,139 +146,110 @@ def fetch_island_data(
...
-        raise ParserException(
-            "ES.py",
-            f"No consumption data returned for zone: {zone_key}",
-            zone_key,
+    consumption = TotalConsumptionList(logger)
+    for event in island_data:
+        consumption.append(
+            zoneKey=zone_key,
+            datetime=datetime.utcfromtimestamp(event.timestamp).astimezone(
+                timezone.utc
```
unitrium, line 163: Double checking, is the island data in UTC as well or local time ?

VIKTORVAV99: As far as I know the whole demanda.ree.es platform use UTC timestamps (not dates) but I can double check this.

parsers/ES.py
```diff
@@ -30,91 +27,113 @@
...
+def check_valid_parameters(
+    zone_key: ZoneKey,
+    session: Optional[Session],
+    target_datetime: Optional[datetime],
+    logger: Logger,
+    exchange: bool = False,
+):
+    """Raise an exception if the parameters are not valid for this parser."""
+    if exchange:
+        print(zone_key)
```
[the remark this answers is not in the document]
VIKTORVAV99: It was left over from testing, I'll remove it.

parsers/ES.py
VIKTORVAV99: Good suggestion, I'll change it.

VIKTORVAV99:
This PR has now been updated so it only covers the add_value() function and update to the fetch_production function.

I'll make some changes that address the concerns that where raised but we might want to change the     _corrected_negative_values set to a dict that also contains how much it was corrected by, partially for accounting reasons and it would give us a better overview of what happens if the production is only partially corrected.

Another option is to create a nested dict that also contain the source mode and by how much it was corrected and if it was corrected to 0 or None.

I think these changes should be addressed in a separate PR though if we want to make them or similar changes.

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: Added a internal function to correct each endpoint, I think it might make sense to also record the amount for each mode that was corrected, mostly for accounting purposes in the future.

parsers/ES.py
```diff
@@ -170,69 +171,55 @@ def fetch_production(
...
-                }
-
-                ### Zone overrides
+    for event in island_data:
+        production = ProductionMix()
+        storage = StorageMix()
+        for mode in [
+            attribute
+            for attribute in getmembers(event)
+            if not attribute[0].startswith("__")
```
unitrium, line 185: I'm not a big fan of this. I adds some hidden logic and is not super future proof if we decide to change attributes in the future. Also I think part is not super readable.
Could you not just construct the mode directly through a mapping?

parsers/ES.py
```diff
@@ -170,69 +171,55 @@ def fetch_production(
...
                     )
+                    # Reset gas to 0, as it is now oil, normally this should be skipped or set to None but to conform to how REE does it, we set it to 0
+                    # We should aim to change this upstream in the future to make it uniform with other parsers going forward.
+                    production.gas = 0
 
-                # Hydro response is hydro storage
-                if zone_key == "ES-CN-HI":
-                    response_data["production"]["hydro"] = 0.0
-                    response_data["storage"]["hydro"] = -response.hydraulic
+                production.add_value(mode=PRODUCTION_MAPPING[mode[0]], value=mode[1])
```
unitrium, line 203: How about maybe adding some logic inside add_value so that if the production mode is None, it sets it to zero and then add the value ?
I think it would be a bit more safe ? And would avoid some of the code above.

electricitymap/contrib/lib/models/events.py
```diff
@@ -93,16 +111,27 @@ def __setattr__(
         value: Optional[float],
     ) -> None:
         """
-        Overriding the setattr method to check for negative values and set them to None.
-        This method also keeps track of the modes that have been corrected.
+        Overriding the setattr method to check that the name is a valid production mode.
         """
         if not name in PRODUCTION_MODES:
             raise ValueError(f"Unknown production mode: {name}")
-        if value is not None and value < 0:
```
unitrium, line 101: You shouldn't remove this part.
Otherwise if someone calls a production mix like this:

```production_mix.wind =-1```
It will not be handled as we want it.

File: electricitymap/contrib/lib/models/events.py (hunk 1 above)
Comment (unitrium) [reply electricitymap/contrib/lib/models/events.py:149]: Let's maybe not overcomplexify this as we are not using the amount yet. So let's just focus on one change and not take into account the amount.

File: electricitymap/contrib/lib/models/events.py (hunk 2 above)
Comment (unitrium) [reply electricitymap/contrib/lib/models/events.py:27]: I think yes you might be right, the add method might be better suited here.

**Review by unitrium — CHANGES_REQUESTED** [review]
I think this requires some changes. Especially since we are adding some code to handle production mix attributes in the fetch_production function. I don't think it's a good practice.
Let's keep it simple rather. I added some comments on how to make this easier. Essentially I think the add_value function should set to 0 if the mode doesn't exist yet and then add.

File: parsers/ES.py (hunk 6 above)
Comment (VIKTORVAV99) [reply parsers/ES.py:185]: Agreed, but due to the way the Response object in the Ree library is structured there it returns nothing to loop over.

I plan on creating a upstream PR to change this (and make it return None instead of 0 when there is no data).

**Comment by VIKTORVAV99** [comment]
@unitrium this should be ready to merge now with the changes for only the parser.

File: parsers/ES.py [hunk 9]
```diff
@@ -140,17 +139,15 @@ def fetch_consumption(
...
     ses = session or Session()
     island_data = fetch_island_data(zone_key, ses, target_datetime)
     consumption = TotalConsumptionList(logger)
     for event in island_data:
         consumption.append(
             zoneKey=zone_key,
-            datetime=datetime.utcfromtimestamp(event.timestamp).astimezone(
-                timezone.utc
-            ),
+            datetime=datetime.fromtimestamp(event.timestamp).astimezone(timezone.utc),
```
VIKTORVAV99, line 150: I noticed I made a error in the previous PR and corrected for UTC twice. This should fix that and I'll re-fetch the consumption and exchange data once this is merged.

unitrium (approved):
Looks good now!

PR #5455, opened 2023-06-09
Fix points in timezones ahead of UTC are discarded

unitrium:
## Issue

We saw for IN-EA, some exchanges being discarded because they were in the future. This comes from the fact that we are comparing the date with utc.

## Description

Making sure the check compares UTC timestamps + add unit test.

### Double check

### DIFF
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -178,14 +178,16 @@ class Event(BaseModel, ABC):
         return v
 
     @validator("datetime")
-    def _validate_datetime(cls, v, values: Dict[str, Any]):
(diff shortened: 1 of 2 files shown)


electricitymap/contrib/lib/models/events.py
```diff
@@ -178,14 +179,16 @@ def _validate_zone_key(cls, v):
         return v
 
     @validator("datetime")
-    def _validate_datetime(cls, v, values: Dict[str, Any]):
+    def _validate_datetime(cls, v: dt.datetime, values: Dict[str, Any]):
```
unitrium, line 182: Since there's an attribute datetime in the object, we need to rename the module a bit for type checker to understand it.

PR #5456, opened 2023-06-10
Restructure scheduled jobs

VIKTORVAV99:
## Issue

It looks like GitHub Security is picking up these as two diffrent jobs when they are run independently, this is not what we want.
image

## Description

Removes the cron schedules on the individual CI jobs and moves it to the "master" CI.
This will make us consume a little more resources but everything besides the CodeQL jobs and perhaps some tests should be cached at this point.

>**Note**
>I have removed the extra configuration from the security panel for now but it will be recreated if we don't merge this.

### Double check

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - master
   pull_request:
+  schedule:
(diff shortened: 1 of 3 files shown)


unitrium (approved):
Looks good to me!

PR #5457, opened 2023-06-11
add error checking on unknown zones + remove unused options params

tonypls:
## Issue
Felt inspired to do some minor performance improvements and prevent the crash when zone param is not valid

## Description
- This PR checks the zone details api response and throws an error if it doesn't contain any zone states (when it's not in our list of zones). 

- Removes the options param from useQuery as we weren't using it for anything.

- Explicitly defines props in a couple of places instead of passing and spreading

### Double check

### DIFF
--- a/web/src/api/getZone.ts
+++ b/web/src/api/getZone.ts
@@ -29,6 +29,9 @@ const getZone = async (
 
   if (response.ok) {
     const { data } = (await response.json()) as { data: ZoneDetails };
+    if (!data.zoneStates) {
+      throw new Error('No data returned from API');
+    }
     return data;
   }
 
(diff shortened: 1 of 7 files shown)


VIKTORVAV99 (approved):
As far as I can tell this looks good but there are some interesting results of using this locally 😅

In the future I think we should add a missing page as well, as I mentioned in the comments but I think that should be a seperate PR.

VIKTORVAV99:
> > As far as I can tell this looks good but there are some interesting results of using this locally 😅
> > 
> > In the future I think we should add a missing page as well, as I mentioned in the comments but I think that should be a seperate PR.
> 
> OOo, what are the interesting local results?

Well since we configured the mockserver to always return data no matter what zone key it has there can be some, well, interesting results. 😅

web/src/components/Flag.tsx
```diff
@@ -15,11 +15,7 @@ interface CountryFlagProps
   size?: number;
 }
 
-export function CountryFlag({
-  zoneId,
-  size = DEFAULT_FLAG_SIZE,
-  ...props
-}: CountryFlagProps) {
+export function CountryFlag({ zoneId, size = DEFAULT_FLAG_SIZE }: CountryFlagProps) {
```
madsnedergaard, line 18: This component should also accept a className to avoid breaking things :)

web/src/api/getZone.ts
```diff
@@ -29,6 +29,9 @@ const getZone = async (
 
   if (response.ok) {
     const { data } = (await response.json()) as { data: ZoneDetails };
+    if (!data.zoneStates) {
+      throw new Error('No data returned from API');
+    }
```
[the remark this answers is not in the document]
madsnedergaard: As an alternative we could adapt the "data missing" message component once the improvements to the app-backend have been merged - then we could more easily show something there, similar to the "We have no data for this zone" message :)

web/src/components/Flag.tsx
tonypls: Good spotting 🧐

tonypls:
Everyone happy for a merge?

VIKTORVAV99:
Thumbs up from me 👍🏻

But let's see what @madsnedergaard say 👀

madsnedergaard (approved):
Yes sorry, all good :)

PR #5459, opened 2023-06-12
Fix for issue #5083 (Graph selector offset when missing data)

### LINKED ISSUE #5083: Graph selector/marker is offset when a data point is missing
## Description
In yearly view when there is a datapoint missing the selector/marker gets offset causing a weird and unexpected behavior when hovering over the datapoints.
image

wrma6:
## Issue

Closes #5083 

## Description

This is my first contribution PR and I'm hoping to do many more in the future (as part of my degree), I'd love to hear any and all feedback!

Fixed the bug described in [issue 5083](https://github.com/electricitymaps/electricitymaps-contrib/issues/5083): "In yearly view when there is a datapoint missing the selector/marker gets offset causing a weird and unexpected behavior when hovering over the datapoints." It does seem that this issue should apply for all views, just most prominently viewed at the yearly timescale.

The issue with the current code is that the offset is calculated using the time difference between the first and second datetimes in the graph. This would only work if all datetimes have the same interval between them.

The fix for this is to compute the offset without the incorrect offset: for any given datetime, the interval is measured between said datetime and the next one. In the case where the datetime is the latest one, we use the value `endTime` that is used for the graph.

Speaking of `endTime`, there is a similar bug in the current code. The value is computed by examining the interval bewteen the last and second last datetimes (and then adding it to the last datetime). If there is a missing value in there, we will not get the proper time interval. The proper fix would be to define constants for the different time settings and then just add it to the last date time (ex. hour, day, year). I would be happy to add this to the PR as well, just not sure if you guys have conventions for hard-coded constants like these (or if these constants already exist somewhere)?

### Preview

image

image

### Double check

### DIFF
--- a/web/src/features/charts/elements/AreaGraph.tsx
+++ b/web/src/features/charts/elements/AreaGraph.tsx
@@ -263,6 +263,7 @@ function AreaGraph({
         timeScale={timeScale}
         valueScale={valueScale}
         datetimes={datetimes}
+        endTime={endTime}
         markerUpdateHandler={markerUpdateHandler}
         markerHideHandler={markerHideHandler}
         selectedLayerIndex={selectedLayerIndex}
--- a/web/src/features/charts/elements/GraphHoverline.tsx
+++ b/web/src/features/charts/elements/GraphHoverline.tsx
@@ -7,6 +7,7 @@ const GraphHoverLine = React.memo(
   ({
     layers,
     datetimes,
+    endTime,
     timeScale,
     valueScale,
     markerUpdateHandler,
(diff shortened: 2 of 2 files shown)


wrma6:
@VIKTORVAV99 I noticed that you were the author of the original issue, are you the right person to ask for review?

VIKTORVAV99:
> @VIKTORVAV99 I noticed that you were the author of the original issue, are you the right person to ask for review?

Yeah I'm in the team that is responsible for the frontend so that works great.

I'll put this on the todo list and hopefully I can do a full review tomorrow.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing it!

PR #5461, opened 2023-06-12
Issue 5352: Fixed layout overflow of ZoneHeaderTitle

### LINKED ISSUE #5352: Zone name in panel is too wide and pushes gauges out

luuu-xu:
## Issue

Closes #5352.
Closes #5389.
Closes #5365.

## Description

According to Issue #5352, using a language like German will cause the ZoneHeaderTitle to have a long width, it leads to overflowing of both/either the header and gauges. By setting **over-flow:hidden** and **truncate** on affecting components' CSS, the overflowing is fixed.

### Preview

Screenshot 2023-06-12 at 13 30 38
Screenshot 2023-06-12 at 13 31 00

### Double check

### DIFF
--- a/web/pnpm-lock.yaml
+++ b/web/pnpm-lock.yaml
@@ -1,8 +1,4 @@
-lockfileVersion: '6.1'
-
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
--- a/web/pnpm-lock.yaml
+++ b/web/pnpm-lock.yaml
@@ -4,6 +4,10 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
 dependencies:
   '@capacitor/app':
     specifier: ^4.1.1
--- a/web/pnpm-lock.yaml
+++ b/web/pnpm-lock.yaml
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
 
 dependencies:
   '@capacitor/app':
@@ -5098,7 +5102,7 @@ packages:
       '@storybook/csf-plugin': 7.0.0-alpha.47
       '@storybook/csf-tools': 7.0.0-alpha.47
       '@storybook/docs-tools': 7.0.0-alpha.47(react-dom@18.2.0)(react@18.2.0)(typescript@4.8.4)
-      '@storybook/mdx2-csf': 1.0.0-next.8
+      '@storybook/mdx2-csf': 1.1.0-next.1
       '@storybook/node-logger': 7.0.0-alpha.47
       '@storybook/postinstall': 7.0.0-alpha.47
       '@storybook/preview-web': 7.0.0-alpha.47(react-dom@18.2.0)(react@18.2.0)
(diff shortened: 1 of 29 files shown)


luuu-xu:
@VIKTORVAV99 Hi, can you please check this PR where simple CSS fixes now solve the gauges overflow issue? Thanks!

web/pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
```
VIKTORVAV99, line 5: Was this something you changed? I would prefer if we skipped this part from this PR.

VIKTORVAV99 (commented):
While I like this approach very much there are some elements that moved (that should not move)
Old version:
image
This PR:
image

This might not really matter in the long run but I would prefer if we didn't change this. (Mostly the country "chip/tag"

web/pnpm-lock.yaml
madsnedergaard: I think it's because of a newer version of pnpm - we should probably add something to the `engines` section to ensure we all use the same version :)
```json
{
    "engines": {
        "node": ">=10",
        "pnpm": ">=3"
    }
}
```

VIKTORVAV99: Ah yeah that might be it, could be smart to pin it to minor versions as well.

luuu-xu:
@VIKTORVAV99 Hi, sorry for not noticing the layout change, I accidentally changed the flex justify-content while solving the issue. I have made the correction and now it has the same layout as before and handles the gauges overflow too.

VIKTORVAV99 (approved):
LGTM! 🎉
Thanks for fixing it!

luuu-xu:
@VIKTORVAV99 Thank you for approving the changes! I see there are some checks failed causing the merge to be unsuccessful, is there anything I should do on my end?

VIKTORVAV99:
I don't think there is anything you have to do, we tried pinning the pnpm version and regenerating the lock files in another PR but there still seems to be issues. But I'll take care of those and then merge this.

web/pnpm-lock.yaml
VIKTORVAV99: Looks like https://github.com/pnpm/pnpm/releases/tag/v8.6.2 reverted the lockfile changes which is why we couldn't replicate the version bump locally (and in the other lockfiles).

So I'm just going to re-generate this lockfile and hopefully that should fix both the current issues and prevent future ones.

PR #5462, opened 2023-06-13
Upgrade ES parser consumption and exchange

VIKTORVAV99:
## Description

Split out from #5451

### Double check


parsers/ES.py
```diff
@@ -216,13 +225,6 @@ def fetch_production(
                     response_data["production"]["hydro"] = 0.0
                     response_data["storage"]["hydro"] = -response.hydraulic
 
-                response_data = validate(
-                    response_data,
-                    logger,
-                    floor=ZONE_FLOORS[zone_key],
-                    expected_range=expected_range,
-                )
-
```
VIKTORVAV99, line 225: This was not really used in this parser so to avoid adding back the entire mapping table (that was mostly 0) I removed it.

parsers/ES.py
```diff
@@ -327,10 +326,6 @@ def fetch_exchange(
     print(fetch_production("ES-CN-HI"))
 
     # Balearic Islands
-    print("fetch_consumption(ES-IB)")
```
unitrium, line 330: Why are those removed ?

unitrium: Let's keep them if anyone wants to test this code.

parsers/ES.py
```diff
@@ -236,30 +238,27 @@ def fetch_production(
...
 def fetch_exchange(
-    zone_key1: ZONE_KEYS,
-    zone_key2: ZONE_KEYS,
+    zone_key1: ZoneKey,
+    zone_key2: ZoneKey,
     session: Optional[Session] = None,
     target_datetime: Optional[datetime] = None,
     logger: Logger = getLogger(__name__),
 ) -> List[dict]:
+    sorted_zone_keys = "->".join(sorted([zone_key1, zone_key2]))
```
unitrium, line 247:
```suggestion
    sorted_zone_keys = ZoneKey("->".join(sorted([zone_key1, zone_key2])))
    check_valid_parameters(sorted_zone_keys, session, target_datetime, logger)
```
To propagate the type.

parsers/ES.py
```diff
@@ -30,91 +28,109 @@
...
+    """Raise an exception if the parameters are not valid for this parser."""
+    if "->" in zone_key:
+        if zone_key not in EXCHANGE_FUNCTION_MAP.keys():
+            zone_key1, zone_key2 = zone_key.split("->")
+            raise ParserException(
+                "ES.py",
+                f"This parser cannot parse data between {zone_key1} and {zone_key2}.",
+                zone_key,
+            )
     else:
```
unitrium, line 97: As explained above try not to use the else here.

unitrium (approved):
Some small code readability feedback but looks good to me! Thanks for taking the time to do this.

parsers/ES.py
VIKTORVAV99: We don't actually have a config for ES-IB so I removed the imported class for a small performance gain (miniscule at best but still).

So I removed these as well to not cause any confusion, this also brings it in like with how we do it for the other island group.

But I can add it back if you prefer.

unitrium: Ah indeed, I didn't saw that, agree better to remove it then.

parsers/ES.py
```diff
@@ -30,91 +28,109 @@
...
-        date = target_datetime.strftime("%Y-%m-%d")
+
+def check_valid_parameters(
+    zone_key: ZoneKey,
+    session: Optional[Session],
+    target_datetime: Optional[datetime],
+    logger: Logger,
+):
+    """Raise an exception if the parameters are not valid for this parser."""
+    if "->" in zone_key:
```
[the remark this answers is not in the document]
VIKTORVAV99: Hmm, while this might make readability easier won't it be a performance hit?

We would have to import EXCHANGES_CONFIG which in turn needs to read the files. It would also make us miss edge cases if the exchange don't exist at all due to a miss spelling or errors in parameter sorting/passing.

The current way should be very performant with just one logic check then branching after that with 1 logic check.

The your option would also have 2 logic checks but it would require additional imports and file reading as well?

I think we should prioritize performance in this parser as it's run often and even if I added the check it is improbable we will hit it.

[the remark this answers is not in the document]
VIKTORVAV99: Pushed a change that I think is the best of both worlds, let me know what you think.

VIKTORVAV99:
Merging this now, we can take a look at things in the other PR if needed.

PR #5463, opened 2023-06-13
Use india proxy for live demand

unitrium:
## Issue

Seems like the datasource has been geoblocked, therefore we need to use the proxy to be able to access it.

## Description

Use the India proxy for live demand data. Also sets the accept-content header to an empty string to let the proxy handle the decompression.

### Double check

### DIFF
--- a/parsers/IN.py
+++ b/parsers/IN.py
@@ -167,10 +167,10 @@ def fetch_consumption(
     for state in STATES_MAPPING[zone_key]:
         # By default the request headers are set to accept gzip.
         # If this header is set, the proxy will not decompress the content, therefore we set it to an empty string.
-        resp: Response = session.get(DEMAND_URL.format(proxy=INDIA_PROXY, state=state), headers={
-            "User-Agent": "Mozilla/5.0",
-            "Accept-Encoding": ""
-        })
+        resp: Response = session.get(
+            DEMAND_URL.format(proxy=INDIA_PROXY, state=state),
+            headers={"User-Agent": "Mozilla/5.0", "Accept-Encoding": ""},
+        )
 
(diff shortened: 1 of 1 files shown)


parsers/IN.py
```diff
@@ -165,8 +165,14 @@ def fetch_consumption(
 
     total_consumption = 0
     for state in STATES_MAPPING[zone_key]:
-        r: Response = session.get(DEMAND_URL.format(state=state))
-        soup = BeautifulSoup(r.content, "html.parser")
+        # By default the request headers are set to accept gzip.
+        # If this header is set, the proxy will not decompress the content, therefore we set it to an empty string.
```
VIKTORVAV99, line 169: TIL

This is probably why I had such problems working with the proxies in the past 😅

VIKTORVAV99 (approved):
Looks like it works, but I am still getting the warnings about data in the future, is this expected?

That is probably unrelated to this change though.

```2023-06-13 12:53:40,334 WARNING  test_parser                   
 Validation failed @ 2023-06-13 23:00:00+05:30: Data from IN-WE can't be in the future, data was 2023-06-13T23:00:00+05:30, now is 2023-06-13T10:53:40.334603+00:00```

mathilde-daugy (approved):

mathilde-daugy (approved):

parsers/IN.py
unitrium: Yes, that was actually a bit hard to debug if you don't have access to the proxy code 😅

unitrium: So I think there's an issue on the backend the update doesn't seem to have been propagated, I'll have a look

PR #5464, opened 2023-06-13
include time-zone from config file

wobniarin:
## Issue

We want to add a new feature to compute daytime and night time into the forecast system. To do that, we need to know the local time zone in order to get sunrise and sunset. 

## Description
This PR is linked to this one: https://github.com/electricitymaps/electricitymaps/pull/3621 

We need to include the TimeZone configuration in order to localize the sunrise and sunset times and compute the daytime feature.

### DIFF
--- a/electricitymap/contrib/config/types.py
+++ b/electricitymap/contrib/config/types.py
@@ -5,3 +5,4 @@ from typing import List, NewType, Tuple
 # representing a box with corners at 140.46°E, 39.64°S and 150.47°E, 33.48°S.
 Point = NewType("Point", Tuple[float, float])
 BoundingBox = NewType("BoundingBox", List[Point])
+TimeZone = NewType("TimeZone", str)
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -17,7 +17,8 @@ from electricitymap.contrib.config.zones import (
     zone_bounding_boxes,
     zone_parents,
 )
-from electricitymap.contrib.lib.types import ZoneKey
+from electricitymap.contrib.lib.types import ZoneKey, TimeZone
+
 
 CONFIG_DIR = Path(__file__).parent.parent.parent.parent.joinpath("config").resolve()
 
@@ -40,6 +41,12 @@ ZONE_BOUNDING_BOXES: Dict[ZoneKey, BoundingBox] = zone_bounding_boxes(ZONES_CONF
 
 # Make a mapping from subzone to the parent zone (full zone).
 ZONE_PARENT: Dict[ZoneKey, ZoneKey] = zone_parents(ZONES_CONFIG)
+
+# Prepare zone time zone
+ZONE_TIME_ZONE: Dict[ZoneKey, TimeZone] = {}
+for zone_id, zone_config in ZONES_CONFIG.items():
+    if "timezone" in zone_config:
+        ZONE_TIME_ZONE[zone_id] = zone_config["timezone"]
 
 # Zone neighbours are zones that are connected by exchanges.
 ZONE_NEIGHBOURS: Dict[ZoneKey, List[ZoneKey]] = generate_zone_neighbours(
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -10,14 +10,14 @@ from electricitymap.contrib.config.reading import (
     read_exchanges_config,
     read_zones_config,
 )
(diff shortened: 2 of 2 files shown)


wobniarin:
> Just to note, not all zone configs have timezones or valid timezone (Null), this should be addressed before merging this.
> 
> Let me know if you want some assistance with that and I'll try and get those done as soon as possible.

Thank you for the heads-up! I thought we had the timezone info for all zones. 

At the moment I am gonna leave it as a draft PR because I want to test how the model performs with this feature (tomorrow most likely) and then we can work on it together to make sure we have the timezone info available before merging this.

VIKTORVAV99:
> > Just to note, not all zone configs have timezones or valid timezone (Null), this should be addressed before merging this.
> > Let me know if you want some assistance with that and I'll try and get those done as soon as possible.
> 
> Thank you for the heads-up! I thought we had the timezone info for all zones.
> 
> At the moment I am gonna leave it as a draft PR because I want to test how the model performs with this feature (tomorrow most likely) and then we can work on it together to make sure we have the timezone info available before merging this.

Sounds good!

electricitymap/contrib/config/types.py
wobniarin, line 1: @VIKTORVAV99 Maybe you can help me here. How come we have two different files called types.py? 

One `electricitymap/contrib/config/types.py` and `electricitymap/contrib/lib/types.py`  which includes the NewType definition for ZoneKey. Do you know why we have this?

VIKTORVAV99: I think we wanted to separate the types that are used for the data in the config files from those we use internally so the first one only has the bounding_box, and point (and now timezone).

While the other has the ZoneKey type.

But honestly we could probably combine these into one type file in the future.

wobniarin: Thanks for the explanation :) I tag @mathilde-daugy for visibility, since we discussed this just 5 minutes ago :)

wobniarin:
> Hey, have you thought about using directly the `ZONES_CONFIG` object instead of creating a new one? From what I see, accessing the timezone from within the `ZONES_CONFIG` object is not complex.
> 
> I therefore would suggest using `ZONES_CONFIG` if the only thing we are doing is accessing that field.
> 
> The reason for this is that introducing new config objects increases the complexity of the configuration, which means in the long term more maintenance, more code to read etc etc.
> 
> Let me know if my explanation for this push-back is not clear :)

Hey, I don't clearly see what you mean. I am already using the `ZONES_CONFIG` object as we are doing for the bounding box and ZoneKey. 

Do you mean to include this part: 

```
# Prepare zone time zone
ZONE_TIME_ZONE: Dict[ZoneKey, TimeZone] = {}
for zone_id, zone_config in ZONES_CONFIG.items():
    if "timezone" in zone_config:
        ZONE_TIME_ZONE[zone_id] = zone_config["timezone"] 

``` 

directly inside the daytime_nighttime.py script? 

If we end up using the daytime_nighttime feature we will need to localize the timezone, so we will need this information anyway.  Isn't it a good opportunity to have the missing information for the zones?

pierresegonne:
> directly inside the daytime_nighttime.py script?

Yes that's what I meant. 

> If we end up using the daytime_nighttime feature we will need to localize the timezone, so we will need this information anyway. Isn't it a good opportunity to have the missing information for the zones?

Yes indeed, but what I am saying is that this information already exists. One just need to extract the timezone field from the `ZONES_CONFIG` object. I can jump on a call to explain if you don't see what I mean

wobniarin:
> > directly inside the daytime_nighttime.py script?
> 
> Yes that's what I meant.
> 
> > If we end up using the daytime_nighttime feature we will need to localize the timezone, so we will need this information anyway. Isn't it a good opportunity to have the missing information for the zones?
> 

> > Yes indeed, but what I am saying is that this information already exists. One just need to extract the timezone field from the `ZONES_CONFIG` object. I can jump on a call to explain if you don't see what I mean

Indeed this information exists, but not for all zones, regarding what Viktor said, isn't it? 

:) No worries Pierre, I think I got what you mean. I will check this out on my own, and then I will close this PR, or get back to you if I don't find my way. Tak!

pierresegonne:
> Indeed this information exists, but not for all zones, regarding what Viktor said, isn't it?

Indeed, but IMO adding an object that just contains a replica of what's already in `ZONES_CONFIG` won't solve this problem :) We should instead review the config files of the zones that don't have the timezone

madsnedergaard:
Should we close this PR now? :)

wobniarin:
Yes, let's close it.

PR #5467, opened 2023-06-15
Add pnpm to engines and pin versions

VIKTORVAV99:
## Description
Adds pnpm to the engines field and pins the version so the same major version to avoid breaking changes.
I also ran pnpm install in all the folders to update the lockfiles.

### Double check


madsnedergaard (approved):
LGTM! 🥳

tonypls (approved):

Issue #5426, opened 2023-05-23
[Data Issue]: new CZ (CEPS) parser is missing pumping demand for pumped hydro


q--:
### When did this happen?

Now that the new parser is live, I noticed that pumping data for pumped hydro is no longer available; IIRC it was available from the ENTSOE data source.

It looks like CEPS [does provide this data](https://www.ceps.cz/en/all-data#Load), so it should be possible to restore this functionality.

### What zones are affected?

Czechia/Czech Republic (CZ)

### What is the problem?

Now that the new parser is live, I noticed that pumping data for pumped hydro is no longer available; IIRC it was available from the ENTSOE data source.

It looks like CEPS [does provide this data](https://www.ceps.cz/en/all-data#Load), so it should be possible to restore this functionality.

VIKTORVAV99:
Ah I see, the new source only provide hydro storage production and not the pumping part.

As far as I can tell at first glance it should be an easy solution of just doing `pumping = load_including_pumping - load`

And then do `hydro_storage_production - pumping` later on.

It's a bit more to it in the actual implementation but that is the basics.

Frantisek12:
I've opened PR https://github.com/electricitymaps/electricitymaps-contrib/pull/5473 with the pumping demand implemented. Please go ahead and check it

oldrobotdev:
I'd like to take this one. Both earlier PRs (#5473, #7530) closed for inactivity rather than rejection, so the plan below follows the approach prescribed in the first comment here plus the review notes from #7530.

Checked today that the Load method still returns what we need, unauthenticated, on the same CepsData.asmx endpoint the parser already uses: value1 "Load including pumping [MW]" and value2 "Load [MW]" at 15-min resolution. On 2026-06-30 at 10:45 local the two were 9802.4 and 9365.6, so 436.8 MW of pumping; overnight the two series are equal.

Plan: fetch Load alongside Generation inside CZ.fetch_production with the same date window and QH/AVG/RT params so timestamps line up, compute pumping as value1 minus value2 per interval, and add it as positive hydro storage so net storage becomes pumping minus PsPP generation. Typed, using the existing ProductionBreakdownList/StorageMix conventions the file already uses. Tests get a second body-matched mock since Generation and Load POST to the same URL. PR to follow shortly.

PR #5473, opened 2023-06-16
#5426 new CZ (CEPS) parser is missing pumping demand for pumped hydro

Frantisek12:
## Issue

[](https://github.com/electricitymaps/electricitymaps-contrib/issues/5426)

Fixes: #5426 

## Description

fetch_consumption added and used to obtain pumping demand. fetch_consumption() not replaced in configuration for entsoe. Could be done if the format is correct. Please check

### Double check


parsers/CZ.py
```diff
@@ -195,6 +262,17 @@ def fetch_production(
...
 
+    if consumption is not None:
+        consumption_dict = {item["datetime"]: item for item in consumption}
+
+        for data_item in data_list:
+            date_time = data_item["datetime"]
+            if date_time in consumption_dict:
+                if float(consumption_dict[date_time]["pumping"]) > 0.0:
+                    data_item["storage"]["hydro"] = consumption_dict[date_time][
+                        "pumping"
```
Frantisek12, line 273: this weird formatting was done by `poetry run format`

parsers/CZ.py
```diff
@@ -68,6 +68,71 @@ def get_target_datetime(dt: datetime | None) -> datetime:
...
+
+    content = make_request(session, payload, zone_key).text
+    xml = BeautifulSoup(content, "xml")
+    mapper = get_mapper(xml)
+
+    data_tag = xml.find("data")
+    data_list: List[dict] = []
+
+    if data_tag is not None:
+        for values in data_tag:
```
VIKTORVAV99, line 104: This whole looping over xml elements is filled with type errors. Not sure we can merge it as is.

parsers/CZ.py
```diff
@@ -68,6 +68,71 @@ def get_target_datetime(dt: datetime | None) -> datetime:
...
+
+            try:
+                consumption_with_pumping = float(values["value1"])
+                consumption = float(values["value2"])
+                date_time = values["date"]
```
VIKTORVAV99, line 109: Causing several issues here.

parsers/CZ.py
```diff
@@ -68,6 +68,71 @@ def get_target_datetime(dt: datetime | None) -> datetime:
     return dt
 
 
+def fetch_consumption(
```
VIKTORVAV99, line 71: This should not be named fetch_consumption if it don't return the expected consumption element.

Should be renamed to get_consumption

VIKTORVAV99 (changes requested):
Sorry for taking so long with this, that's on me, but there are a lot of type errors and undocumented logic at work here. I don't think we can merge it as is.

madsnedergaard:

PR #5474, opened 2023-06-18
Fix: production history shows mt instead of kg

### LINKED ISSUE #4317: Tooltip of production history shows mt instead of kg for unit of emissions
See this screenshot:
image

Similar to #793.

wrma6:
## Issue

Closes #4317 

## Description

This should be the fix for [this open issue](https://github.com/electricitymaps/electricitymaps-contrib/issues/4317), just one simple change in the end. The issue seems to be that the value used for comparison in `formatCo2` is the parameter that uses units `gCO₂ / h` instead of the final calculated value (that is the number formatted) that uses units `tCO₂ / min`.

### Preview

image

### Double check

### DIFF
--- a/web/src/utils/formatting.ts
+++ b/web/src/utils/formatting.ts
@@ -27,7 +27,7 @@ export const formatCo2 = function (d: number, numberDigits: number = DEFAULT_NUM
     return d;
   }
 
-  return d >= 1
+  return value >= 1
     ? `${d3.format(`.${numberDigits}s`)(value)}t ${translate('ofCO2eqPerMinute')}` // a ton or more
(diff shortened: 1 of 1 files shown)


wrma6:
@jarek I saw that you responded to the original issue, are you the right person to ask for review?

jarek:
Thanks for making the PR! As far I know, @VIKTORVAV99 should be best for reviewing front-end PRs.

VIKTORVAV99:
Yep that would be me, I'll take a look at this as soon as possible.

Thanks for the mention @jarek.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing it @wrma6!

PR #5478, opened 2023-06-19
Use Merit India to get consumption data

unitrium:
## Issue

Seems like the merit india dashboard is less overloaded then the other dashboard. We would get less timeout issues. Also needs the proxy as it is geoblocked.

## Description

Switch to Merit India for the consumption data.

### Double check


VIKTORVAV99:
I'll do a review after lunch 👍🏼

mathilde-daugy (approved):

PR #5480, opened 2023-06-19
Changes to engines requirements to greater than X instead of strict major

madsnedergaard:
## Description

In #5467 versions were added, but I think pinning to a specific major is a bit too strict especially for Node :)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -4,8 +4,8 @@
   "description": "A real-time visualisation of the CO2 emissions of electricity consumption",
   "license": "AGPL-3.0-or-later",
   "engines": {
-    "node": "^18",
-    "pnpm": "^8"
+    "node": ">=18",
+    "pnpm": ">=8"
   },
   "type": "module",
   "repository": {


VIKTORVAV99 (commented):
I guess this is a tradeoff between restricting it to the version we have tested vs what is installed locally.

Bumping node is fine but I'm not sure we should allow pnpm above 8 as the CI strictly made to run at 8 and there has been major breaking changes before.

web/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "A real-time visualisation of the CO2 emissions of electricity consumption",
   "license": "AGPL-3.0-or-later",
   "engines": {
-    "node": "^18",
-    "pnpm": "^8"
+    "node": ">=18",
+    "pnpm": ">=8"
```
madsnedergaard, line 8:
```suggestion
    "pnpm": "^8"
```

madsnedergaard:
Yeah you're right on pnpm, I've reverted that part :)

VIKTORVAV99 (approved):
LGTM! 🎉

Do you want to change the mockserver and mobile configs as well before merging?

madsnedergaard:
> LGTM! 🎉
> 
> Do you want to change the mockserver and mobile configs as well before merging?

Thanks, missed that! :)

PR #5481, opened 2023-06-19
include capacity changes for JP-HKD

wobniarin:
## Issue

We have different historical capacity values for JP-HKD. The parser does not retrieve installed capacity, but we can include manually from the data sources. The capacity is usually updated at the end of the fiscal year (in Japan, around March) 

## Description

We include a function that returns the capacity value for JP-HKD based on the year that is queried. The capacity value is included into the json blob returned. Right now we only include yearly values for wind. 

### Preview

image

### Double check

### DIFF
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -76,9 +76,7 @@ def fetch_production(
     datalist = []
 
     for i in df.index:
-        capacity = get_wind_capacity(
-                    df.loc[i, "datetime"].to_pydatetime(), zone_key
-                )
+        capacity = get_wind_capacity(df.loc[i, "datetime"].to_pydatetime(), zone_key)
         data = {
             "zoneKey": zone_key,
             "datetime": df.loc[i, "datetime"].to_pydatetime(),
@@ -94,9 +92,7 @@ def fetch_production(
                 "geothermal": None,
                 "unknown": df.loc[i, "unknown"],
             },
-            "capacity": {
-                "wind":capacity if capacity is not None else {}
-                            },
+            "capacity": {"wind": capacity if capacity is not None else {}},
             "source": "occtonet.or.jp, {}".format(sources[zone_key]),
         }
         datalist.append(data)
--- a/parsers/JP.py
+++ b/parsers/JP.py
@@ -9,8 +9,8 @@ import arrow
 import pandas as pd
 from requests import Session
 
-from parsers import occtonet
 from electricitymap.contrib.config import ZONES_CONFIG
+from parsers import occtonet
 from parsers.lib.config import refetch_frequency
 
 # Abbreviations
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (commented):
Not sure if I was meant to review this already as it's still a draft but it works well if you remove the breakpoint 😉

This could perhaps benefit from being turned into a dictionary mapping but it would only make really make a difference if we add more values.

wobniarin:
> Not sure if I was meant to review this already as it's still a draft but it works well if you remove the breakpoint 😉
> 
> This could perhaps benefit from being turned into a dictionary mapping but it would only make really make a difference if we add more values.

> Not sure if I was meant to review this already as it's still a draft but it works well if you remove the breakpoint 😉
> 
> This could perhaps benefit from being turned into a dictionary mapping but it would only make really make a difference if we add more values.



Hey, no. It is still a WIP and as you have seen I forgot a breakpoint. I will mark it as ready for review tomorrow when I have run formatting and removed the missing breakpoints there. Sorry for that, and thanks for the early feedback :)

parsers/JP.py
```diff
@@ -71,13 +86,17 @@ def fetch_production(
                 "geothermal": None,
                 "unknown": df.loc[i, "unknown"],
             },
+            "capacity":
+            {"wind": get_wind_capacity(df.loc[i, "datetime"].to_pydatetime(), zone_key) if zone_key in ["JP-HKD"] else None},
             "source": "occtonet.or.jp, {}".format(sources[zone_key]),
         }
         datalist.append(data)
-
+    breakpoint()
```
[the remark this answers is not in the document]
wobniarin: yes! thanks for pointing it out :)

parsers/JP.py
```diff
@@ -39,6 +40,22 @@
...
+    capacity = ZONE_CONFIG["capacity"]["wind"]
+    if zone_key == "JP-HKD":
+        if datetime.year <= 2019:
+            capacity = 480
+        elif datetime.year == 2020:
+            capacity = 520
+        elif datetime.year >= 2021:
+            capacity = 577
+    else:
+        raise NotImplementedError("This parser can only fetch wind capacity for JP-HKD")
```
VIKTORVAV99, line 55: Won't this assertation make the else statment redundant as it will throw before it is ever reached?

You could also use a ParserException instead to log the zone_key but I don't think that is required.

VIKTORVAV99 (commented):
Just one small thing, then I think we can merge it.

parsers/JP.py
wobniarin: Actually, you are right. 

Now I am testing it for `JP-TH` since it uses the same parser and now we get this as an output. `capacity:None` 

image

Do you think that can mess down the road with capacities? Should I only include the dict 

``` 
{'capacity': {'wind': None},
``` 

for zones that are not JP-HKD?

VIKTORVAV99: I think it would be better to do something like:
`"capacity": capacities if capacities is not None else {}`

That way we don't contaminate the database with wrong capacity values which I think would affect the app.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5491, opened 2023-06-20
Updates script to remove zone

madsnedergaard:
## Description

I'm about to clean up the older India zones, but wanted better tooling in place first :)

Improves the "remove zone" script by:
- Cleans up the script by splitting into functions and improving logging
- Adding new functionality for moving parsers and finding files using the zonekey
- Ensures touched files are formatted correctly


scripts/remove_zone.py
```diff
@@ -37,73 +39,118 @@ def _is_zone_in_exchange(exchange_config_path: str) -> bool:
...
 
-    run_shell_command(f"npx prettier --write {geo_json_path}", cwd=ROOT_PATH)
+    run_shell_command(
+        f"npx prettier --config {prettier_config_path} --write {geo_json_path}",
+        cwd=ROOT_PATH,
+    )
+    run_shell_command(
+        f"pnpm generate-world",
+        cwd=ROOT_PATH / "web",
+    )
```
VIKTORVAV99, line 88: This will fail if the zone we are removing is a subzone since we are not updating the parent config file.

VIKTORVAV99 (changes requested):
Some early feedback.

Will do another review later on.

scripts/remove_zone.py
```diff
@@ -37,73 +39,118 @@ def _is_zone_in_exchange(exchange_config_path: str) -> bool:
...
             pass
 
-    # TODO: we could check if part of subZoneNames
-    # TODO: we could detect parsers that are only used by this zone
-    #       so we can clean those files up.
 
+def remove_translations(zone_key: ZoneKey):
     for locale_file in LOCALE_FILE_PATHS:
-        with JsonFilePatcher(locale_file, indent=4) as f:
+        with JsonFilePatcher(locale_file, indent=2) as f:
```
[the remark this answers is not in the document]
madsnedergaard: Strange, which zone did you try to remove? It might be because it's trying to read a file that is not JSON (like .DS_Store) 🤔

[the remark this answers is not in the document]
VIKTORVAV99: SE-SE1 😅

It was the first that came to mind...

scripts/remove_zone.py
madsnedergaard: Hmm, which part would fail?
And yes, we are not doing anything about parent zone configs - but the "file searcher" should point out that the zone-key is used in the config and it's asking to manually clean that :)

VIKTORVAV99: The generate-world script tries to look up all the zones defined as subzones in the parent config to find the "borders", since we are removing the zone from the geojson but not from the subzone field in the parent the script won't be able to find it and will crash.

I think that the crash happens before the "file searcher" so it would go unnoticed.

madsnedergaard: Ah, now I get it! Fixed :)

scripts/remove_zone.py
```diff
@@ -37,73 +39,118 @@ def _is_zone_in_exchange(exchange_config_path: str) -> bool:
...
+                        )
+    if not results:
+        print(f'🧹 Found no additional files mentioning "{text}" in contrib repository.')
+        return
+
+    print(
+        f"❗️ Found {len(results)} files mentioning {text}, please manually clean these files:"
+    )
+    for result in results:
+        print(f"  - {result}")
```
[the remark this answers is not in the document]
madsnedergaard: I think so. I haven't set up my Windows desktop after moving, so I cannot really test or fix it right now. How about adding a warning that this script only works on MacOS as a first step before merging?

VIKTORVAV99:
@madsnedergaard this looks good now and works (kind of). It needs a few tweaks to account for other updates (prettier needs to be pinned to v2, some updated path strings and such) would you mind if I pushed those directly, after resolving the merge conflicts?

I was also planning on adding a guard to only let it run on Posix systems for now if that is okay.

madsnedergaard:
> @madsnedergaard this looks good now and works (kind of). It needs a few tweaks to account for other updates (prettier needs to be pinned to v2, some updated path strings and such) would you mind if I pushed those directly, after resolving the merge conflicts?
> 
> I was also planning on adding a guard to only let it run on Posix systems for now if that is okay.

That would be great, please do 😍

VIKTORVAV99 (approved):
LGTM! 🎉

Feel free to merge anytime but I would suggest double checking so I didn't mess up the merge resolution.

madsnedergaard:
> LGTM! 🎉
> 
> Feel free to merge anytime but I would suggest double checking so I didn't mess up the merge resolution.

Thanks for adding the changes! 🙏

PR #5497, opened 2023-06-21
Deduplicate contributor names in zones.json

VIKTORVAV99:
## Issue
We currently store contributor names across multiple zones which means that if one contributor has contributed to many zones the size of the zones.json file grows exponentially in file size even if the total contributor count across all zones just increased by 1. 

## Description
This PR stores the contributors in a deduplicated array at the top level and replace the zone specific contributor names with a index relating to their position in the deduplicated array. This results in significant size savings when a contributor has contributed to more than one zone with a minimal additional cost (a single int value and a comma, right now that would be 3 chars) if they have just contributed to one zone.

### Additional improvements that can be made:

- We can get a few more bytes out of this if we sort the contributor names by the amount of zone configs they are added in but I did not think that was worth it as a initial step as it would only be a few bytes at best at the current time.

- We could also just include them in the deduplicated array if they have contributed to more than 1 zone  and otherwise keep the name in the zone specific config. But this adds even more complexity and would not result in any major size reductions at this point. 

>**Note**
> This PR is a proof of concept and needs some renaming, cleanup and a thorough review before merging but it's technically functional.

### Double check


VIKTORVAV99:
This reduce the size as much as I had initially intended but it does reduce it by about 8 kb before compression. However the size reduction will be greater as contributors contribute to more zones.

I'm sure there are a few things that could be improved in here and this is nothing urgent so if you have some extra time feel free to do thorough review.

web/src/features/panels/zone/util.ts
```diff
@@ -32,7 +33,7 @@ export enum ZoneDataStatus {
   UNKNOWN = 'unknown',
 }
 
-const zonesConfig: Record<string, zoneConfigItem | undefined> = zonesConfigJSON;
+const zonesConfig = zonesConfigJSON as unknown as CombinedZonesConfig;
```
madsnedergaard, line 36: Let's move this one to the top as it's used inside `getHasSubZones` above

web/geo/generateWorld.ts
```diff
@@ -29,7 +29,7 @@ const EXCHANGE_OUT_PATH = path.resolve(
 
 const fc: WorldFeatureCollection = getJSON(config.WORLD_PATH);
 const zoneConfig = mergeZones();
-const aggregates = generateAggregates(fc, zoneConfig);
+const aggregates = generateAggregates(fc, zoneConfig.zonesConfig);
```
madsnedergaard, line 32: In all the places where this is changed, could we rename it from `zoneConfig.zonesConfig` to something less confusing? 😄 
Maybe `zones.config` or just `zones.zonesConfig`?

madsnedergaard (commented):
Added a few comments, otherwise I think it makes sense - if you want to add a quick win on top of this, we could exclude timezone from the generated file as I don't think we're using it in the app? :)

web/geo/generateWorld.ts
VIKTORVAV99: I was thinking about this after I created the PR. I think maybe config.zones make sense then we also have config.contributors.

In the future we could also add the exchanges to the same object but I haven't explored that.

VIKTORVAV99:
> Added a few comments, otherwise I think it makes sense - if you want to add a quick win on top of this, we could exclude timezone from the generated file as I don't think we're using it in the app? :)

I thought you where going to push the changes we played around with 😅
(include list instead of exclude list).

But I can do that in this PR if you wish (or you can push your changes here).

madsnedergaard:
> > Added a few comments, otherwise I think it makes sense - if you want to add a quick win on top of this, we could exclude timezone from the generated file as I don't think we're using it in the app? :)
> 
> I thought you where going to push the changes we played around with 😅 (include list instead of exclude list).

Oh yeah, forgot all about that! I have pushed a new commit with the changes here :)

web/geo/generateWorld.ts
madsnedergaard: Oh, guess we could rename the other `config` constant otherwise?

VIKTORVAV99:
I think I finally got everything ready in here but I noticed that there are some things in the backend that also use the mergeZones function (that I renamed to getConfig) that need to be updated as well.

I'm not really sure how to go about this as I don't want to merge this before that is updated but I can't update it without merging this 😅

madsnedergaard:
Haha yeah, that is definitely a problem 😅 
I think the best option is to make a branch on the backend where contrib points at this specific branch, and then when it's reviewed and ready to merge, we merge this one first and then the backend branch (including a submodule bump). Happy to help on the release part when we get to that :)

VIKTORVAV99:
> Haha yeah, that is definitely a problem 😅 I think the best option is to make a branch on the backend where contrib points at this specific branch, and then when it's reviewed and ready to merge, we merge this one first and then the backend branch (including a submodule bump). Happy to help on the release part when we get to that :)

That sounds like good plan, I'll wait until this is reviewed and then do the backend changes.

I'll definitely take you up on that help with the release when we are at that point.

madsnedergaard (approved):
LGTM

q--:
👍

 You can disregard my comment, then.

VIKTORVAV99:
@madsnedergaard can we just merge this? I looked at the backend again and it looks like we copied all the files over instead of importing from contrib when it was migrated to TS. So the backend should be unaffected?

madsnedergaard (approved):
Let's do it :)

PR #5501, opened 2023-06-22
useState instead of jotai for is map moving + nuke tooltip when moving

tonypls:
## Issue
Some android devices struggle with the map providing a laggy performance

## Description
This PR removes some of the render control logic from jotai global state and controls them just with local state and passing props. This is experimental to test if this provides a smoother experience.


madsnedergaard:
Interesting idea - would be cool if we could measure the performance before and after this to get some numbers. Maybe a hackday project for tomorrow? 😄

VIKTORVAV99:
Also worth nothing is that Jotai released v2 some time ago that changes a few things, perhaps it's more performant than the v1 version in our case?

https://jotai.org/docs/guides/migrating-to-v2-api

tonypls:
> Interesting idea - would be cool if we could measure the performance before and after this to get some numbers. Maybe a hackday project for tomorrow? 😄

As a quick test I'll try get hold of Ciril's phone which he's had some performance issues on in the past.

tonypls:
Gave this branch a test and compared it to production on Ciril's phone and it was quite a bit smoother. 

I think the main improvement is the hiding of the exchange layer as the touch event begins over when the map starts moving.

Would love to dig in the details of  the performance improvement numbers but that might not be till next hackday

tonypls:
Closing this in favour of #5565. 

I still think there is value in using vanilla react useState over a Jotai atom for performance but let's discuss the value of performance vs complexity and tackle that in another PR if we want.

PR #5508, opened 2023-06-24
Changes data caching and refetching logic

### LINKED ISSUE #5064: Make background data refresh work again
It seems like the functionality isn't working any more. 
We should get it back, so that the app will call `/state` and recolor the map on an interval. This is especially useful for anyone using the app as a dashboard.

madsnedergaard:
## Issue

Closes #5064

### Why

This PR fixes several problems with the way we're fetching data from the backend:
1. Every 5 minutes we are refetching everything, which is wasting a lot of MBs if leaving the app open
2. Due to our caching, data might sometimes come out of sync between state and details (e.g. the map tooltip and the left panel)
3. We might not show the latest available data (if visiting the app just before the new hour, caching means the new hour is not shown yet)

### How 

In short, I have made some changes here based on the fact that we don't change data within the same hour, so we only really need to refresh data when crossing into the next hour (e.g. visiting at 3:58pm and staying for 5 minutes).

This was achieved by:
- changing queries to stay cached and fresh forever (it's only in memory though)
- added a new function that every minutes check if we have entered a new hour...
- and if yes, we invalidate the hourly state and the hourly zone details

Things that I have ensured work nicely:
- [x] Map changes colors
- [x] Map tooltip data refreshes
- [x] If a zone is selected and shown in left panel, data refreshes
- [x] Clicking another zone makes it refetch data for that zone (even though it was previously cached)
- [x] It does not download detailed data for all zones previously visited (which would be a waste if user does not want to look at that zone again)
- [x] Weather layers work as before

### Preview

_I changed it to refetch every minute and added some logging to visualise the effect:_

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/75e31ee9-8e03-46d0-9714-0fcdce03aa1a

### Double check


web/src/utils/refetching.ts
```diff
@@ -0,0 +1,33 @@
...
+      // Invalidate hourly zone queries - this matches all zone queries
+      queryClient.invalidateQueries({
+        queryKey: [QUERY_KEYS.ZONE, { aggregate: TimeAverages.HOURLY }],
+      });
+
+      // Reset the start hour
+      startHour = currentHour;
+    }
+  }, ONE_MINUTE);
+}
```
VIKTORVAV99, line 33:
```suggestion
export function refetchDataOnHourChange(queryClient: QueryClient) {
  let startHour = new Date().getUTCHours();
  setInterval(() => {
    const currentHour = new Date().getUTCHours();
    if (startHour !== currentHour) {
      console.info(`Refetching data for new hour: ${currentHour}`);

      // Invalidate hourly state query
      queryClient.invalidateQueries({
        queryKey: [QUERY_KEYS.STATE, { aggregate: TimeAverages.HOURLY }],
      });

      // Invalidate hourly zone queries - this matches all zone queries
      queryClient.invalidateQueries({
        queryKey: [QUERY_KEYS.ZONE, { aggregate: TimeAverages.HOURLY }],
      });

      // Reset the start hour
      startHour = currentHour;
    }
  }, ONE_MINUTE);
}
```

This should probably use hours instead of minutes right?
(UTC hours prevents timezone weirdness from interfering with this)

VIKTORVAV99: I'm thinking it might be better to set this as a timeout with the following value instead though to avoid checking it all the time?
```js
  const currentDate = new Date();
  const refetchIn =
    (60 * 60 * 1000) -
    (currentDate.getUTCMinutes() * 60 * 1000 +
      currentDate.getUTCSeconds() * 1000 +
      currentDate.getUTCMilliseconds());
  setTimeout(() => console.log('Refetch Now'), refetchIn);
```
This is just a late night thought though, I don't really know what is best or more performant.

web/src/main.tsx
```diff
@@ -53,11 +53,16 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       retry: MAX_RETRIES,
-      refetchInterval: REFETCH_INTERVAL_FIVE_MINUTES,
+      refetchOnWindowFocus: false,
+      // by default data is cached and valid forever, as we handle invalidation ourselves
+      cacheTime: Number.POSITIVE_INFINITY,
```
tonypls, line 58: Is one hour or 24 maybe more sensible?

web/src/api/getAppVersion.ts
```diff
@@ -13,7 +13,6 @@ async function getVersion(): Promise<{ version: string }> {
 
 export const useGetAppVersion = () =>
   useQuery<{ version: string }>([], async () => getVersion(), {
-    refetchInterval: REFETCH_INTERVAL_ONE_HOUR,
-    refetchOnWindowFocus: false,
```
tonypls, line 17: The getVersion function does pop up a lot in sentry. I think because it's trying to access local files. With refetch on window focus we may get more of these errors

web/src/utils/refetching.ts
tonypls: Alternatively we could use the refetchIn variable that Viktor has suggested here as the refetchInterval.

```
const getTimeToNextHour = () => {
  const currentDate = new Date();
  const refetchIn =
    60 * 60 * 1000 -
    (currentDate.getUTCMinutes() * 60 * 1000 +
      currentDate.getUTCSeconds() * 1000 +
      currentDate.getUTCMilliseconds());

  return refetchIn;
};

const useGetState = (): UseQueryResult<GridState> => {
  const timeToNextHour = getTimeToNextHour();
  const [refetchInterval, setRefetchInterval] = useState(timeToNextHour);
  return useQuery<GridState>(
    [QUERY_KEYS.STATE, timeAverage],
    async () => getState(timeAverage),
    {
      refetchInterval,
      staleTime: ONE_HOUR,
      refetchOnWindowFocus: true,
      onSuccess: () => setRefetchInterval(getTimeToNextHour()),
    }
  );
};

export default useGetState;
```

tonypls (commented):
Nice idea, couple of comments

web/src/api/getAppVersion.ts
VIKTORVAV99: could we avoid this by bundling the app version in the config file instead when we generate that locally?

web/src/utils/refetching.ts
madsnedergaard: Whoops yes, just changed it to minutes to make the preview video 😅 

I thought about using refetchInterval too, but if I visit the app at 4:55pm the refetchInterval would be 5 minutes,  and after the hour change we don't want it to refetch that often. So I don't think calculating the "time until next hour" just once is going to work :)

web/src/api/getWeatherData.ts
```diff
@@ -111,8 +111,8 @@ const useGetWeather = (
     [type],
     async () => await getWeatherData(type),
     {
-      staleTime: REFETCH_INTERVAL_FIVE_MINUTES,
-      refetchOnWindowFocus: false,
+      staleTime: FIVE_MINUTES,
+      cacheTime: FIVE_MINUTES,
```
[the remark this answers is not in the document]
madsnedergaard: Yes, but I overwrite the default in `main.tsx` and therefore set it to 5 min here 😉

web/src/utils/refetching.ts
tonypls: That's why I added this line :D

```onSuccess: () => setRefetchInterval(getTimeToNextHour()),```

madsnedergaard: Ah sorry, didn't see the code-block was scrollable 😂 Will test it!

web/src/api/getAppVersion.ts
VIKTORVAV99: Isn't that better suited for a service worker? I belive my PWA PR adds this automatically.

web/src/utils/refetching.ts
tonypls: This function has been deprecated though, I liked that this wouldn't reinvent the wheel but maybe it's best to ignore this approach.

madsnedergaard: Where do you see that `onSuccess` is deprecated?

tonypls: https://tanstack.com/query/latest/docs/react/reference/useQuery
image

madsnedergaard: Ah thanks, turns out they even did a blog post about it: https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

As an alternative, we could do something like this using a useEffect hook, but it also gets quite messy and I'm unsure it if will cause bugs since we use the getZone hook in multiple places 🤔 



Screenshot 2023-06-26 at 11 20 10

Screenshot 2023-06-26 at 11 21 41

web/src/api/getAppVersion.ts
madsnedergaard: Yes indeed, let's discuss that in the PWA PR :)

web/src/utils/refetching.ts
tonypls: I messaged Dominik on twitter about our use case and the deprecated function, this was his reply:
image

tonypls (approved):
Calling a function every minute doesn't feel like the most eloquent solution but it's a damn sight better than what we currently do. 

I'm happy to approve this and if we want to do something different in the future we can.

VIKTORVAV99 (approved):
I agree with Tony, we should try and find a way to avoid running this every minute but let's merge the current improvements and then we can try and figure out a better way to do it in the future.

madsnedergaard:
Haha thanks both, and I agree! It's not elegant and smart, but it works and it has no performance impact - and it's very easy to reason about :)

PR #5509, opened 2023-06-25
Add geothermal and biomass to Taiwan

VIKTORVAV99:
## Issue
Taiwan has geothemal and biomass production we are currently not parsing.

## Description
The values for geothermal and biomass are not listed as their own "fueltypes" but go under Other Renewable Energy, fortunately for us they have Geothermal and Biofuel in their name.

This PR changes things so we look at the name for the plants that has the fueltype Other Renewable Energy and then manually assign their fueltype based from their name.

>**Note**
>It is currently not possible to update the fetch production in the tw.py parser to the new parser classes as they don't support capacity parsing yet.

Closes: #5107 

### Double check


unitrium (approved):
Looks good!

PR #5518, opened 2023-06-26
update india zone name

tonypls:
## Issue
The data portal scripts depend on the zoneName of a parent country matching the countryName of a zone. 

## Description
This PR renames "Mainland India" to "India" to match the country name fields

### Double check

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -776,7 +776,7 @@
       "zoneName": "Isle of Man"
     },
     "IN": {
-      "zoneName": "Mainland India"
+      "zoneName": "India"
     },
     "IN-EA": {
       "countryName": "India",


VIKTORVAV99 (commented):
I knew this would cause problems some day.
As it stands we still have one India zone that is not aggregated because we don't have a parser for it, [Andaman and Nicobar Islands](https://app.electricitymaps.com/zone/IN-AN), and by renaming it to just India we would be miss representing it in the app.

How does this work for the US? Those zones don't match either?

tonypls:
> I knew this would cause problems some day. As it stands we still have one India zone that is not aggregated because we don't have a parser for it, [Andaman and Nicobar Islands](https://app.electricitymaps.com/zone/IN-AN), and by renaming it to just India we would be miss representing it in the app.
> 
> How does this work for the US? Those zones don't match either?

We can do this on the backend

```
        if zone_name == "Contiguous United States":
            zone_name = "USA"
            country_name = "United States of America"
```

tonypls:
Closed in favour of ignoring India in the export scripts

PR #5520, opened 2023-06-26
Remove flag_file_name from all zone configs

VIKTORVAV99:
## Issue

The flag_file_name is no loger used but still included in almost all zone configs.

## Description

This removes the flag_file_name from all the zone configs and updates any typings to account for this.

### Double check


scripts/remove_flag_file_name.py
```diff
@@ -0,0 +1,23 @@
+import yaml
```
madsnedergaard, line 1: I assume this made things much easier when doing it :D
Not sure we really need to keep the script though?

VIKTORVAV99: Considering there was 250+ files to look though it sure did 😅

But you are right, we don't need to keep the script. I'll push a commit that removes it.

VIKTORVAV99:
Hmm Cypress e2e is failing because of some unknown script error?
This should not have affected that in any way 😅

VIKTORVAV99:
Looks like cypress is failing on master as well so it's not related to this PR.

madsnedergaard (approved):

PR #5523, opened 2023-06-27
Fix for Mongolia parser

### LINKED ISSUE #5470: MN production parser down
## Description

This is an automatic error report generated for Mongolia (MN).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser MN production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/MN.html">Explore the runtime logs</a>

MuffinCompiler:
## Issue

Closes #5470

## Description

Endpoint for Mongolia data changed from table (parsed) from HTML, to JSON. Available data fields are still exactly the same as in previous data source (wind, solar, exchange, unknown which infered from total consumption). Changed general structure a bit to remove duplicated code in consumption and production methods.

Also changed a single import in the `IN-EA` parser. The present absolute import path yields runtime errors for me. All other parsers use the import I changed it to.

### Double check

- [x] I have tested my parser changes locally with `poetry run test_parser MN [production|consumption]`

### DIFF
--- a/parsers/IN_EA.py
+++ b/parsers/IN_EA.py
@@ -7,7 +7,7 @@ from requests import Session
 
 from electricitymap.contrib.config import ZoneKey
 from electricitymap.contrib.lib.models.event_lists import ExchangeList
-from electricitymap.contrib.parsers.lib.config import refetch_frequency
+from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
 IN_WE_PROXY = "https://in-proxy-jfnx5klx2a-el.a.run.app"
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import json
 from datetime import datetime
 from logging import Logger, getLogger
 from typing import Any, Dict, Optional
@@ -74,7 +73,7 @@ def query(session: Session) -> Dict[str, Any]:
         )
 
     # Read as JSON
-    response_json = json.loads(target_response.text)
+    response_json = target_response.json()
     query_result = parse_json(response_json)
 
     return query_result
@@ -98,7 +97,7 @@ def fetch_production(
         - query_data["importMW"]
         - query_data["solarMW"]
         - query_data["windMW"],
-        2,
+        13,
     )
 
     dataset_production = {
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -3,7 +3,7 @@
 import json
 from datetime import datetime
 from logging import Logger, getLogger
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Optional
 
 import arrow
 from pytz import timezone
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -8,6 +8,12 @@ import arrow
 from pytz import timezone
 from requests import Response, Session
 
+from electricitymap.contrib.config import ZoneKey
+from electricitymap.contrib.lib.models.event_lists import (
+    ProductionBreakdownList,
+    TotalConsumptionList,
+)
+from electricitymap.contrib.lib.models.events import ProductionMix
 from parsers.lib.exceptions import ParserException
 
 NDC_GENERATION = "https://disnews.energy.mn/test/convert.php"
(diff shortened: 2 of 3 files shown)


parsers/IN_EA.py
```diff
@@ -7,7 +7,7 @@
 
 from electricitymap.contrib.config import ZoneKey
 from electricitymap.contrib.lib.models.event_lists import ExchangeList
-from electricitymap.contrib.parsers.lib.config import refetch_frequency
```
VIKTORVAV99, line 10: I assume you are on windows?
I had the same issues so this change is perfectly fine.

parsers/MN.py
```diff
@@ -1,16 +1,83 @@
...
+
+    if not target_response.ok:
+        raise ParserException(
+            parser="MN.py",
+            message=f"Data request did not succeed: {target_response.status_code}",
+        )
+
+    # Read as JSON
+    response_json = json.loads(target_response.text)
+    query_result = parse_json(response_json)
```
VIKTORVAV99, line 78: If it is actually json you can save a bit of trouble and simplify this a bit by calling response.json() instead of response.text

parsers/MN.py
```diff
@@ -21,34 +88,27 @@ def fetch_production(
...
+    query_data = query(session)
 
+    # Calculated 'unknown' production from available data (consumption, import, solar, wind).
+    # 'unknown' consists of 92.8% coal, 5.8% oil and 1.4% hydro as per 2020; sources: IEA and IRENA statistics.
+    query_data["unknownMW"] = round(
+        query_data["consumptionMW"]
+        - query_data["importMW"]
+        - query_data["solarMW"]
+        - query_data["windMW"],
+        2,
```
VIKTORVAV99, line 101:
```suggestion
        13,
```

parsers/MN.py
```diff
@@ -64,24 +124,13 @@ def fetch_consumption(
...
 
-    time_combined = datetime.combine(date_pretty, time_pretty)
+    query_data = query(session)
 
     data = {
         "zoneKey": zone_key,
-        "datetime": time_combined.replace(tzinfo=timezone(TZ)),
-        "consumption": consumption_MW,
+        "datetime": query_data["time"],
+        "consumption": query_data["consumptionMW"],
```
VIKTORVAV99, line 133: I don't know if you are aware of this but we have added parser classes with built in validation, it would be great if those could be implemented.

You can find the documentation here: https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser

If you don't feel comfortable doing it I'll do it when this PR is otherwise ready to be merged.

VIKTORVAV99 (commented):
Some general suggestions and improvements to simplify things.

PS: Nice to see we are using a JSON endpoint now, they are so much easier to work with!

parsers/MN.py
MuffinCompiler: Done!

parsers/MN.py
MuffinCompiler: d14dfca
I wasn't aware of that, but great to see a change to better Parser return objects!

Currently, even though this parser only returns one breakdown, I think the interface requires the `ProductionBreakdown` to be packed in a `ProductionBreakdownList`? Because the documentation says "In cases where the parser function collects data for more than one datetime, the data should be stored in the following [list] objects...", but returning only a single `ProductionBreakdown` gives an error.

Also since there is a ProductionBreakdown and a TotalProduction, I opted to use the breakdown, even though we only have solar/wind/unknown here. 

Feel free to change things if the new objects should be used differently :)

VIKTORVAV99: Thanks for updating it!

So far things look good from a quick glance but I'll do a full review tomorrow morning.

As for the ProductionBreakdownList, this is required even if there is just a single data point so we can simplify things in the future.

As you noticed there is a to_list() function, in the future the goal is to remove the need for this and return the production breakdown list as is to enable us to further enhance validation and unify the parsers.

This also lays the groundwork so we can return multiple data types (consumption and production for example) in the same parser run making everything a bit more efficient (though this is not yet fully shaped).

VIKTORVAV99 (approved):
LGTM! 🎉

I changed the date/time parsing a bit so we don't need arrow but this works great!

PR #5525, opened 2023-06-28
Add simplified chinese translations

helicopter-z:
## Description

Updated simplified Chinese translations to be up-to-date with the English strings. 

### Double check


VIKTORVAV99 (approved):
Double checked a few of these with a translator and as far as I can tell they all look good!

Thanks for updating these translations!

PR #5529, opened 2023-06-28
Update CH with 2022 data

danijelst:
Update historical data based on the mentioned source and add the record for 2022

## Description

Update data to the latest value provided by uvek-gis.admin.ch

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for updating it!

VIKTORVAV99:
Looks like we need to update the test before merging this.

Could you do it? Otherwise I'll do so tomorrow morning and then merge it.

PR #5530, opened 2023-06-29
Add zone for Lord Howe Island

### LINKED ISSUE #3286: Add a new zone : Lord Howe Island
### description

Lord Howe Island is a small electricity zone and electricity-independent from Australia 🇦🇺 .

#### geographical
https://en.wikipedia.org/wiki/Lord_Howe_Island

### data

public dashboard: http://photonscada.com/data/perspective/client/LHI => live data via WebSocket

### capacity source
https://www.lhib.nsw.gov.au/infrastructure/renewable-energy

wrma6:
## Issue

Closes #3286 

## Description

Add a new zone for Lord Howe Island, which includes:
- Defining the required configs for a new zone, I called the zone `AU-LH` (not sure if there is formal procedure for names)
- Defining the geography using QGIS
- Writing the Python parser

** CURRENT ISSUE **: the site used as the data source is [this](http://photonscada.com/data/perspective/client/LHI). I noticed that parsers would usually use BeautifulSoup in this case, but it doesn't actually work for this site. I think it's because the websit e loads dynamically through Javascript, which the standard Python `requests` module can't handle. Instead, I've used the `requests_html` library which runs Chromium in headless mode to load the page properly. However, this means that my parser doesn't follow the defined parser prototype (`session` is of different type, from a different module). I would love to get some guidance on how to proceed from here, or maybe there is another pre-existing way to deal with cases like these?

### Preview
Screenshot 2023-06-23 225537

image

### Double check


VIKTORVAV99:
That would be me in most cases 😅

I'll try and take a look at it today.

parsers/AU-LH.py
```diff
@@ -0,0 +1,200 @@
...
+
+    production_list = ProductionBreakdownList(logger=logger)
+    production_list.append(
+        zoneKey=zone_key,
+        datetime=datetime.now(timezone.utc),
+        production=ProductionMix(
+            gas=data["diesel"],
+            solar=data["solar"],
+        ),
+        storage=StorageMix(battery=data["battery"]),
```
VIKTORVAV99, line 122:
```suggestion
        storage=StorageMix(battery=-data["battery"]),
```
Storage values should be inverted as we count positive as storage charge and negative as storage discharge.

parsers/AU-LH.py
```diff
@@ -0,0 +1,200 @@
...
+        source=DATA_URL,
```
VIKTORVAV99, line 123: This should only be the domain. Not the full URL. In this case that would be "photonscada.com".

parsers/AU-LH.py
```diff
@@ -0,0 +1,200 @@
...
+        session = HTMLSession()
+
+    data = fetch_data(zone_key, session)
+
+    consumption_list = TotalConsumptionList(logger=logger)
+    consumption_list.append(
+        zoneKey=zone_key,
+        datetime=datetime.now(timezone.utc),
+        consumption=data["consumption"],
+        source=DATA_URL,
```
VIKTORVAV99, line 154: Same here.

parsers/AU-LH.py
```diff
@@ -0,0 +1,200 @@
...
+
+    production_list = TotalProductionList(logger=logger)
+
+    production_list.append(
+        zoneKey=zone_key,
+        datetime=datetime.now(timezone.utc),
+        value=data["solar"] + data["diesel"],
+        source=DATA_URL,
+    )
+
```
VIKTORVAV99, line 188: This is not really what this data type is supposed to return. It is ususally used when the total generation/production is reported seperatly from the individual production modes.

It is also missing battery discharge.

parsers/AU-LH.py
```diff
@@ -0,0 +1,200 @@
...
+    # determine the sign of the battery value based on the other values
+    if value_dict["diesel"] + value_dict["solar"] < value_dict["consumption"]:
+        value_dict["battery"] *= -1
+
+    return value_dict
+
+
+def fetch_production(
+    zone_key: ZoneKey,
+    session: HTMLSession = None,
```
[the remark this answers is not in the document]
wrma6: Yeah, I mentioned this in the PR description... is it OK to just keep it in the function signature and then open my own HTMLSession instead?

parsers/AU-LH.py
wrma6: Ok, I'll remove it

parsers/AU-LH.py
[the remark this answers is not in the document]
VIKTORVAV99: The expected type should still be a "normal" session even if you don't use it.

Then you can create your own session inside the function.

parsers/AU-LH.py
wrma6: I think this is properly considered in the code (line 101 in the new push to the branch), I invert the number if needed in that conditional statement. The battery number can be charge or discharge depending on the time of day.

VIKTORVAV99:
> I added `requests_html` to `pyproject.toml`, and then ran `poetry lock` + `poetry install -E parsers`. Is this the correct way to add dependencies? I noticed that `poetry.lock` changed a lot, even for modules that aren't related to requests_html.

Usually `poetry add "dependency" -E parsers` is the way to go. But it's normal for poetry to bump packages that are in range when running poetry lock.

wrma6:
Using `poetry add "dependency" -E parsers` seems to do some weird things to the file for me, I get `requests-html = {version = "^0.10.0", extras = ["parsers"]}` instead of something like `odfpy = {version="^1.4.1", optional=true}` that the other packages have.

Let me know if the current setup doesn't do it, I'll try and figure out another way if that's the case.

VIKTORVAV99:
> Using `poetry add "dependency" -E parsers` seems to do some weird things to the file for me, I get `requests-html = {version = "^0.10.0", extras = ["parsers"]}` instead of something like `odfpy = {version="^1.4.1", optional=true}` that the other packages have.
> 
> Let me know if the current setup doesn't do it, I'll try and figure out another way if that's the case.

That is normal, poetry has updated their syntax but it should not cause any problems. I'm taking another look at this later today.

VIKTORVAV99 (changes requested):
Been taking another look at this and I'm very hesitant to using request_html.

For one it downloads an entire chromium version on the first run, something we should not be doing as it introduces external security risks (outside of source control).

Secondly it seems like the package is no longer maintained, no updates have been made to the source repository in years, which means we can't trust it to support newer versions of python when we want to upgrade. (It also looks like it's using a old chromium version).

So I don't think we can or should merge this as is. If you wish we could merge the geometries and config files though, just not the parser and dependency changes.

wrma6:
Thanks for the review @VIKTORVAV99 , I can definitely understand you concerns especially considering Lord Howe is not the biggest region out there. What would you prefer, only to merge the geometries or we could just close this PR until a better solution maybe reveals itself in the future?

VIKTORVAV99:
We can definitely merge the geometries and config and just skip the parser for now.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5532, opened 2023-06-29
Fix and upgrade ZA.py to get South Africa data again.

VIKTORVAV99:
## Issue
The South Africa parser is down.

## Description
This PR fixes and cleans up the parser. It was originally intended to use our new parser classes but after talking to Robin we decided to hold off on that until we have a better way to merge production outputs for the same production modes.

closes: #5143

### Double check


electricitymap/contrib/config/model.py
```diff
@@ -69,7 +69,7 @@ def get_function(self, type: str) -> Optional[Callable]:
         """
         function_str = getattr(self, type)
         if function_str:
-            return import_string(f"electricitymap.contrib.parsers.{function_str}")
```
VIKTORVAV99, line 72: This was just to get the tests running locally, let me know if I should revert it.

parsers/ZA.py
```diff
@@ -18,118 +55,77 @@ def get_url() -> str:
...
-                    data["production"][column_mapping[j]] + float(prod_data_value),
-                    13,
-                )
-
-        all_data.append(data)
-
-    return all_data
+    if not res.ok:
+        raise ParserException(
+            "ZA.py",
```
madsnedergaard, line 75: Not related to this PR, but couldn't we infer the filename in a smarter way? I also see that we in some places use `zone_key` instead of the filename - so maybe a general approach could help clean that up too :)

madsnedergaard (approved):
Looks good to me! 
Sad that data is quite delayed, is it just temporarily?

`max returned datetime: 2023-06-26T21:00:00+00:00 UTC  :( >2h from now !!!`

VIKTORVAV99:
I think that's the state of things sadly but it will help guide our estimations and fill the historical data at least.

parsers/ZA.py
VIKTORVAV99: For the parser exception we log the parser, message and zone_key. But yeah hopefully we can make this more "automatic". I think Robin has improving logging on some roadmap somewhere 😉

PR #5538, opened 2023-06-30
Fetches latest hour separately before fetching all 24 hours

madsnedergaard:
## Issue

Waiting to load all 24 hours of data before showing anything on the map is not a great user experience, especially on slow internet.

## Description

_This work is the result of @VIKTORVAV99 and I peer-coding in a hotel lobby at midnight 🦉🚀_ 

Enables the app to fetch the latest hour of data at first, and then afterwards fetch the rest. 
This means the app loads and displays the map way faster, so users don't have to stare at the loading overlay until everything is there.

### Preview

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/45ebc1df-defa-4dd2-9966-2b3e0438bbb0

### Double check


web/src/api/getState.ts
```diff
@@ -24,9 +24,26 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
+
+  // Then fetch the rest of the data
+  const all_data = useQuery<GridState>(
+    [QUERY_KEYS.STATE, { aggregate: timeAverage }],
+    async () => getState(timeAverage),
+    {
+      // The query should not execute until the hour_zero query is done
+      enabled: Boolean(hour_zero.isLoading === false && hour_zero.data),
+    }
   );
```
VIKTORVAV99, line 45: Something became evident in the preview deployment that might result in us having to change this.

Due to the fact that the backend PR hasn't been merged this always return a 400 status code.

While this would naturally resolve itself when we have merged said PR we should probably make it more robust in case the first query fails and still fetch the all data one since it includes the same data.

Another thing that came to mind is that we only need to fetch the first hour or hour zero when the user is first opening the app, not every time we fetch the hourly state (why didn't we think of this before? 😅).

I realize this adds complexity but I think it is worth it to make the app more robust and efficient.

madsnedergaard: Good point on the 400, I'll attempt to make it proceed despite failing on the first one.

Since we now only fetch hourly once per hour, I don't think the additional request is a problem really :)

VIKTORVAV99 (approved):
Yay! It works great!

I think we should wait to merge this until we have decided on the endpoint name and merged the backend PR though.

madsnedergaard:
> This is great, I found a significant performance increase if we add `useGetState();` in app.tsx, it's not a blocking call and gets the ball rolling on our biggest request first.

Great point, it definitely improves things! :)

Before: 
Screenshot 2023-08-08 at 22 51 55

After: 
Screenshot 2023-08-08 at 22 52 05

VIKTORVAV99:
Eeh, changing the loading order did something unexpected...
Now all the data will be fetched before the map is done loading so we will always have the full hourly state when the app is finished loading. Making the first hour loading somewhat redundant...

If we can start preloading the world file earlier though that might shake things up again. (or just make it smaller)

madsnedergaard:
> Eeh, changing the loading order did something unexpected... Now all the data will be fetched before the map is done loading so we will always have the full hourly state when the app is finished loading. Making the first hour loading somewhat redundant...

Still think it's worth it for slower networks - as far as I am aware, the downloads should happen something like this:

```
/last_hour     | <--->
/hourly        |      <----->
map            | <----->
```

madsnedergaard:
The "full" 24 hour query is only enabled after the first_hour query is successful :)

Screenshot 2023-08-09 at 20 31 28

VIKTORVAV99:
> The "full" 24 hour query is only enabled after the first_hour query is successful :)
> 
> Screenshot 2023-08-09 at 20 31 28
> 

Oh right.
But the biggest benefit here should be to mobile users either way. And if they are using the app the world file will come bundled.

VIKTORVAV99:
> > The "full" 24 hour query is only enabled after the first_hour query is successful :)
> > 
> > Screenshot 2023-08-09 at 20 31 28
> > 
> 
> Oh right.
> But the biggest benefit here should be to mobile users either way. And if they are using the app the world file will come bundled. So this should give them some nice speed gains.

PR #5539, opened 2023-06-30
add_value method

VIKTORVAV99:
## Description

Adds a add value method to loop over a list that should be mapped to the same mode and sum them for instances where multiple source type maps to the same electricity maps type.

### Double check


electricitymap/contrib/lib/models/events.py
```diff
@@ -119,11 +154,31 @@ def set_value(
         If correct_negative_with_zero is set to True, negative values will be set to 0 instead of None.
         This method keeps track of values that have been corrected.
         """
-        if correct_negative_with_zero and value is not None and value < 0:
-            value = 0
-            self._corrected_negative_values.add(mode)
+        value = self._correct_negative_value(mode, value, correct_negative_with_zero)
         self.__setattr__(mode, value)
 
+    def add_value(
```
unitrium, line 160: Since we are doing OOP, how about bringing some part inside the base Mix class?
I think the part about adding the value can be shared.

electricitymap/contrib/lib/models/events.py
```diff
@@ -108,6 +128,21 @@ def __setattr__(
...
+    def _correct_negative_value(
+        self, mode: str, value: Optional[float], correct_negative_with_zero: bool
+    ) -> Union[float, None]:
+        """
+        Corrects a negative value by setting it to None or 0.
+        This method also keeps track of the modes that have been corrected.
+        """
+        if value is not None and value < 0:
+            self._corrected_negative_values.add(mode)
+            if correct_negative_with_zero == True:
```
unitrium, line 140:
```suggestion
            value = 0 if correct_negative_with_zero else None
```

unitrium: Might be easier one line

unitrium (commented):
Small suggestions, I think we can improve a bit readability :)

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: Sure thing, I thought we might want to add something else here in the future (like record the value) but it would be trivial to change back.

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: They already are in the base mix class?
This just overides that so it also corrects for negative values which we don't want to do for storage but I guess we could move it all up and use a flag instead?

electricitymap/contrib/lib/models/events.py
```diff
@@ -119,11 +154,31 @@ def set_value(
...
+        mode: str,
+        value: Optional[float],
+        correct_negative_with_zero: bool = False,
+    ) -> None:
+        """Adds the provided value to the existing value of the provided mode. Negative values are set to None by default.
+        If correct_negative_with_zero is set to True, negative values will be set to 0 instead of None.
+        This method keeps track of modes that have been corrected.
+        """
+        value = self._correct_negative_value(mode, value, correct_negative_with_zero)
+        if value is None:
```
unitrium, line 167:
```suggestion
        super().add_value(mode, value)
```
In that case I suggest doing this and remove the code below :)

electricitymap/contrib/lib/models/events.py
unitrium: Indeed, I didn't expended the code so I didn't see it was already in the Mix.
Then you can directly go for my suggestion below

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: Oh I have not really used that approach before, seems useful!

I'll test it!

VIKTORVAV99:
I realize this PR don't have any tests, do you want me to add some specific ones that just test the add value method before we merge this (after we are otherwise happy with the implementation).

unitrium:
Actually yes, that was the last thing I had in mind, a small unit test to demonstrate that it works and make sure we don't do regressions in the future.

VIKTORVAV99:
> Actually yes, that was the last thing I had in mind, a small unit test to demonstrate that it works and make sure we don't do regressions in the future.

I'll add 3 ones for the ProductionMix and 2 for StorageMix, that should cover all scenarios I think. I'll get those done this weekend.

pyproject.toml
```diff
@@ -105,7 +105,8 @@ load-plugins = "pylint_pydantic"
 [tool.pytest.ini_options]
 testpaths = [
     "tests",
-    "parsers/test"
+    "parsers/test",
+    "electricitymap/contrib/lib/tests",
```
VIKTORVAV99, line 109: Turns out we weren't running any of the new tests 😅

unitrium (approved):
Great thanks for implementing this!

VIKTORVAV99:
> Great thanks for implementing this!

Whooo! 🥳

PR #5540, opened 2023-06-30
Allow price data to be in the future

unitrium:
## Issue

Price data is published for the day ahead, so they need to be able to be in the future somehow.

## Description

Those are not forecasts, it's a measured event since the market cleared. So it doesn't make sense to have it as a forecasted event.

For now, we can just remove the check on the future for price data. Later on when we focus on whole sale price data, we can decide if a bigger gap makes sense.

### Double check


electricitymap/contrib/lib/models/events.py
```diff
@@ -573,11 +573,20 @@ class Price(Event):
...
         return v
 
+    @validator("datetime")
+    def _validate_datetime(cls, v: dt.datetime) -> datetime:
+        """Prices are given for the day ahead, so we should allow them to be in the future."""
+        if v.tzinfo is None:
+            raise ValueError(f"Missing timezone: {v}")
+        if v < LOWER_DATETIME_BOUND:
+            raise ValueError(f"Date is before 2000, this is not plausible: {v}")
+        return v
```
VIKTORVAV99, line 588: This goes for the other events as well but should we add a check to make sure the datetime is actually a datetime here, before it's checked by the other validation?
```suggestion
    @validator("datetime")
    def _validate_datetime(cls, v: dt.datetime) -> datetime:
        """Prices are given for the day ahead, so we should allow them to be in the future."""
        if not isinstance(v, datetime):
            raise TypeError(f"Date is not a native datetime object: {v}")
        if v.tzinfo is None:
            raise ValueError(f"Missing timezone: {v}")
        if v < LOWER_DATETIME_BOUND:
            raise ValueError(f"Date is before 2000, this is not plausible: {v}")
        return v
```

unitrium: As far as I'm aware pydentic enforces the type before running the validation

VIKTORVAV99: Just did a quick check and it seems to allow more than just datetimes:
`invalid type; expected datetime, string, bytes, int or float (type=type_error)`
But then again it also seems to convert non datetime types like timestamps to a datetime so it might not be a huge concern.

unitrium: Yes so actually it allows a lot of different things that are accepted but they all will get converted to datetimes.
https://docs.pydantic.dev/dev-v2/usage/types/datetime/
I would suggest for now not changing this, I will open another PR to make improvements there.

VIKTORVAV99: Sounds good, I'll approve this so we can merge it.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5542, opened 2023-06-30
Vonvert pandas timestamp to datetime before merging

unitrium:
## Issue

Seems like datetimes were converted to timestamps during the merge operation. We should stick to python datetimes.

## Description

Ensure that we get a datetime after merging events together.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5548, opened 2023-07-01
Add Réunion and update FR_O.py to parser classes

VIKTORVAV99:
## Description

This PR adds the Reunion streaming dataset and updates the parser to use the new parser classes.

### Double check


parsers/FR_O.py
```diff
@@ -157,92 +169,89 @@ def fetch_data(
...
+
+        production_breakdown_list.append(
+            zoneKey=zone_key,
+            datetime=datetime.fromisoformat(production_object[f"{date_string}"]),
+            production=production,
+            storage=storage,
+            source="edf.fr",
+            sourceType=EventSourceType.estimated
+            if production_object["statut"] == "Estimé"
+            else EventSourceType.measured,
```
VIKTORVAV99, line 229: I don't know if this is actually supported for external estimations so if it's not I can revert it (or add a enum for estimated_external).

unitrium: It's exactly for this kind of parser that I had this in mind so it's supported :)
The only thing is that there's no action on it in the backend for now.

parsers/FR_O.py
```diff
@@ -157,92 +169,89 @@ def fetch_data(
...
-                f"Rate limit exceeded. Please try again later after: {data.reset_time}",
+                f"Rate limit exceeded. Please try again later after: {data.get('reset_time')}",
             )
         elif data.get("error_code") == "ODSQLError":
             raise ParserException(
                 "FR_O.py",
                 "Query malformed. Please check the parameters. If this was previously working there has likely been a change in the API.",
             )
-    return data, DATE_STRING_MAPPING[zone_key]
+    if isinstance(data, list):
```
unitrium, line 179: One suggestion I would make here is to invert the condition so that you raise first if there an issue, otherwise you continue your flow path.
It limits the depth of the function and improves readability in my opinion :)
```suggestion
    if not isinstance(data, list):
```

parsers/FR_O.py
```diff
@@ -157,92 +169,89 @@ def fetch_data(
...
+                logger.warning(
+                    f"Unknown mode_key: '{mode_key}' encountered for {zone_key}."
+                )
+
+        production_breakdown_list.append(
+            zoneKey=zone_key,
+            datetime=datetime.fromisoformat(production_object[f"{date_string}"]),
+            production=production,
+            storage=storage,
+            source="edf.fr",
```
unitrium, line 226: I would suggest having the source as a constant at the top of the file

parsers/FR_O.py
VIKTORVAV99: Yay! Last time I added it the whole parser broke... 😅

unitrium (approved):
Looks good to me!
Small code suggestions. If you have time, I would suggest adding some tests but you can already go ahead

VIKTORVAV99:
I'll add the suggestions and get some tests for the realtime data sources.

PR #5550, opened 2023-07-03
Add validation for RS hydro in ENTSOE.py

VIKTORVAV99:
## Issue

The Serbia production values for hydro have been over 6 times the installed capacity for several hours the last few days but are have been corrected in the ENTSO-E transparency platform. 

## Description

This PR adds validation for RS hydro production so it can never go over double it's installed capacity. Which means it should exclude all the erroneous values while still allowing it to increase without us having to adjust it.

### Double check


madsnedergaard (approved):
Nice, I was just about to do the same! 🥳

PR #5551, opened 2023-07-03
Fix negative solar gets discarded for BR

unitrium:
## Issue

We noticed several datapoints getting discarded for BR zones because they have negative solar. Those negative values are self consumption and should not be discarded but set to 0.

## Description

Updates the code for BR production the the new Event structure and allow negative solar to be corrected to zero.
Update the tests accordingly.

### Double check


parsers/BR.py
```diff
@@ -55,62 +63,51 @@ def get_data(session: Optional[Session]):
...
-        GENERATION_MAPPING.get(name, "unknown"): val
-        for name, val in totals.items()
-        if name not in entries_to_remove
-    }
+    del totals["total"]
+    production = ProductionMix()
+    for mode, value in totals.items():
+        mode_name = GENERATION_MAPPING.get(mode, "unknown")
+        production.add_value(
+            mode_name, value, mode_name in CORRECTED_NEGATIVE_PRODUCTION
```
VIKTORVAV99, line 84: Why do we want to single out solar and not correct all modes with negative values?

parsers/BR.py
```diff
@@ -55,62 +63,51 @@ def get_data(session: Optional[Session]):
...
-        + totals.get("itaipu50HzBrasil", 0.0)
-        + totals.get("itaipu60Hz", 0.0)
-    )
-    entries_to_remove = {"hidraulica", "itaipu50HzBrasil", "itaipu60Hz", "total"}
-    mapped_totals = {
-        GENERATION_MAPPING.get(name, "unknown"): val
-        for name, val in totals.items()
-        if name not in entries_to_remove
-    }
+    del totals["total"]
```
VIKTORVAV99, line 79: If we are getting a total value should we not also add a TotalProductionList (total generation parser)?

I can do that in a seperate PR if you want to get this merged first.

parsers/BR.py
```diff
@@ -55,62 +63,51 @@ def get_data(session: Optional[Session]):
...
 
 
-def production_processor(json_data, zone_key: str) -> tuple:
+def production_processor(
+    json_data: dict, zone_key: str
+) -> Tuple[datetime, ProductionMix]:
     """Extracts data timestamp and sums regional data into totals by key."""
 
-    dt = arrow.get(json_data["Data"])
+    dt = arrow.get(json_data["Data"]).datetime
```
VIKTORVAV99, line 71: This looks to be a propper ISO date string, we can probably replace arrow with datetime.fromisoformat() here.

VIKTORVAV99 (approved):
Some general comments and suggestions for further improvments but things look good.

parsers/BR.py
unitrium: So far I do not have information on whether other modes are reporting negative consumption and if this is self consumption.
Since the issue only arised with solar so far, I would limit it to solar. Other modes will be set to None.

parsers/BR.py
unitrium: True that would be even better!

parsers/BR.py
VIKTORVAV99: What about consumption? I think "cargaVerificada" is load and if we subtract the exports from that we should get the consumption data.

PR #5553, opened 2023-07-04
Use refetch token if specified datetime in ENTSOE parser

unitrium:
## Issue

We are rolling out an automatic data refetch. For ENTSOE we risk hitting API rate limitations. Therefore when we are refetching data, we should be using a specific token.

## Description

When the target datetime is specified, we should use the refetch token as when the target datetime is specified it means this is a refetch.

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -483,7 +483,6 @@ def query_ENTSOE(
             message="target_datetime has to be a datetime in query_entsoe",
         )
 
-
     # make sure we have an arrow object
     params["periodStart"] = (target_datetime + timedelta(hours=span[0])).strftime(
         "%Y%m%d%H00"  # YYYYMMDDHH00
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -472,24 +472,28 @@ def query_ENTSOE(
     Raises an exception if no API token is found.
     Returns a request object.
     """
+    env_var = "ENTSOE_REFETCH_TOKEN"
     if target_datetime is None:
         target_datetime = datetime.utcnow()
-    if isinstance(target_datetime, datetime):
-        # make sure we have an arrow object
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (commented):
I'm a bit unsure if this is needed.

I don't think our rate limit issues are due to the API token but the IP address rate limiting. Not to mention we already have the ability to use multiple API tokens in the parser and are randomizing which is in use each run.

By introducing another ENV variable all local debugging would need to set multiple ENV variables (even if it's the same token).

I would rather see us focusing on using a proxy or list of proxies to spread out the API calls across IP addresses rather than tokens.

unitrium:
We have introduced a new IP, that I will use for the refetches in a subsequent PR. To avoid any flagging we are going to have a new token being used with this new IP address.

VIKTORVAV99:
Ah okay, I suppose that makes sense.
But then I'd like to add a fallback for local debugging, something like this:
```python
if get_token(ENTSOE_REFETCH_TOKEN) is None: # Might need changing, have not tested locally.
  env_var = ENTSOE_TOKEN
```

unitrium:
I'm not sure we can easily introduce this without making big changes because we don't have any debug parameter somewhere.
So I think the easiest is just to require two env variables, which doesn't seem like a big change to me ?

VIKTORVAV99:
> I'm not sure we can easily introduce this without making big changes because we don't have any debug parameter somewhere. So I think the easiest is just to require two env variables, which doesn't seem like a big change to me ?

It's just that it's a breaking change we don't really document anywhere. But I suppose it only affect people running the parser locally so it might be enough to just update the wiki?

unitrium:
Yes, agree updating the wiki should be enough.

VIKTORVAV99:
I'll update the wiki as soon as this is merged. 🙂

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5558, opened 2023-07-05
Improve pydantic validation

unitrium:
## Issue

A newer version of pydantic is available with some nice new validations.

## Description

- upgrade to pydantic v2.
- use AwareDatetime for datetimes in the Event class

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Does this also take care of the datetime issues we discussed in the other PR?

unitrium:
Not yet, I'm still a bit hesitant on introducing a strict type validation that goes beyond pydantic since they anyways convert it to a datetime in the end.

PR #5559, opened 2023-07-05
Updates missing bounding boxes (and improves the script)

madsnedergaard:
## Description

To improve our estimations, zones should have bounding boxes!
This PR updates the script and adds the missing ones.

The only zones without a bounding box is a small US zone that is deprecated and Russia (as it cross the 180th Meridian line and therefore causes problems).

### Preview

This is less useful for very large aggregated zones like India, but thought it would be nice to just do it for all zones while at it :)

Screenshot 2023-07-05 at 15 30 14


web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -72,18 +86,44 @@ for (const zone of zonesGeo.features) {
   ];
 }
 
+if (isAggregate) {
+  const averageBoundingBox: number[][] = Object.values(boundingBoxes).reduce(
```
VIKTORVAV99, line 90: Huh I thouth ESLint would complain about the usage of reduce.

web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -1,36 +1,50 @@
...
+if (zoneFeatures.length <= 0) {
+  console.info(`Zone ${zoneKey} does not exist in geojson, using subzones instead`);
+  isAggregate = true;
+  zoneFeatures = zonesGeo.features.filter((d) => {
+    return d.properties.countryKey === zoneKey;
+  });
+}
+zonesGeo.features = zoneFeatures;
 
 let allCoords = [];
```
VIKTORVAV99, line 46: Can't add a full suggestion as you didn't change the lines below but I think we should do something like this:
```ts
let allCoords: Position[] = [];
let boundingBoxes: { [key: string]: number[][] } = {};

for (const zone of zonesGeo.features) {
  allCoords = [];
  const geometryType = zone.geometry.type;
  for (const coords1 of zone.geometry.coordinates) {
    for (const coord of coords1[0]) {
      allCoords.push(coord as Position);
    }
  }
```
That way we get rid of a type error.

VIKTORVAV99 (approved):
Looks good!

Just one suggestion but I don't think you caused the error. 🙂

web/geo/generateZoneBoundingBoxes.ts
madsnedergaard: Nice, I was a bit annoyed by the type error, so thanks for the suggestion! I had to do some additional changes on top of your suggestion here to avoid additional errors :)

web/geo/generateZoneBoundingBoxes.ts
madsnedergaard: Replaced with a simple loop :)

VIKTORVAV99: Do you want to do the same for the arrows? 😉
image

config/zones/FO-MI.yaml
```diff
@@ -72,14 +78,16 @@ parsers:
...
+      https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
   UNECE 2022:
-    link: https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf
-  ? UNECE 2022, WindEurope "Wind energy in Europe, 2021 Statistics and the outlook
-    for 2022-2026" Wind Europe Proceedings (2021)
-  : link:
+    link: >-
+      https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf
+  UNECE 2022, WindEurope "Wind energy in Europe, 2021 Statistics and the outlook for 2022-2026" Wind Europe Proceedings (2021):
+    link: >-
```
VIKTORVAV99, line 90: Huh it actually cleaned these up? I'm a bit surprised but it's nice!

VIKTORVAV99 (approved):
LGTM! 🎉

(In fact a little too good, this apparently cleans up all out YAML files! 🥳)

web/geo/generateZoneBoundingBoxes.ts
madsnedergaard: Eeeh, let's save that for another day 😅

PR #5563, opened 2023-07-06
Fix ES.py to not update global constant

VIKTORVAV99:
## Issue

Somehow it seems that the global constant was perserved between different parser runs.
This caused subsequent runs to have the wrong global constant.

## Description

This PR prevents the updating of the global constant by reassigning the needed values locally instead and then overriding the local value when needed.

### Double check


parsers/ES.py
```diff
@@ -177,17 +177,17 @@ def fetch_production(
                 # NOTE the LNG terminals are not built yet, so power generated by "gas" or "combined" in ES-CN domain is actually using oil.
                 # Recheck this every 6 months and move to gas key if there has been a change.
                 # Last checked: 2022-06-27
+                MAPPING = PRODUCTION_MAPPING
                 if zone_key.split("-")[1] == "CN":
```
VIKTORVAV99, line 181: I know we talked about using an explicit mapping but since we also assign gas to 0 to be backwards compatible (until I/we update the ree library) we still need the if statement.

parsers/ES.py
```diff
@@ -177,17 +177,17 @@ def fetch_production(
...
-                        {
-                            "gas": "oil",
-                            "combined": "oil",
-                        }
-                    )
+                    MAPPING = {
+                        **PRODUCTION_MAPPING,
+                        "gas": "oil",
+                        "combined": "oil",
+                    }
```
VIKTORVAV99, line 186: I guess update would still have worked but this seemed more explicit.

parsers/ES.py
```diff
@@ -177,17 +177,17 @@ def fetch_production(
                 # NOTE the LNG terminals are not built yet, so power generated by "gas" or "combined" in ES-CN domain is actually using oil.
                 # Recheck this every 6 months and move to gas key if there has been a change.
                 # Last checked: 2022-06-27
+                MAPPING = PRODUCTION_MAPPING
```
unitrium, line 180:
```suggestion
                fuel_mapping = PRODUCTION_MAPPING
```
Since the value might change it's not a constant, so lower case keys should be used.

parsers/ES.py
unitrium: Ok then let's not over complexify it.

parsers/ES.py
```diff
@@ -177,17 +177,17 @@ def fetch_production(
...
                 if zone_key.split("-")[1] == "CN":
-                    PRODUCTION_MAPPING.update(
-                        {
-                            "gas": "oil",
-                            "combined": "oil",
-                        }
-                    )
+                    MAPPING = {
```
unitrium, line 182:
```suggestion
                    fuel_mapping = {
```

parsers/ES.py
```diff
@@ -177,17 +177,17 @@ def fetch_production(
...
+                        **PRODUCTION_MAPPING,
+                        "gas": "oil",
+                        "combined": "oil",
+                    }
                     # Reset gas to 0, as it is now oil, normally this should be skipped or set to None but to conform to how REE does it, we set it to 0
                     # We should aim to change this upstream in the future to make it uniform with other parsers going forward.
                     production.gas = 0
-                production.add_value(PRODUCTION_MAPPING[mode], value, True)
+                production.add_value(MAPPING[mode], value, True)
```
unitrium, line 190:
```suggestion
                production.add_value(fuel_mapping[mode], value, True)
```

unitrium (approved):
Approving, just a small change about the variable.

PR #5564, opened 2023-07-06
update to jotai v2

tonypls:
## Issue
Update Jotai to V2

## Description
This PR updates Jotai to version 2.2.2. Migration guide [here](https://jotai.org/docs/guides/migrating-to-v2-api)

We don't use any of the deprecated functions so this was an easy upgrade 😄 

### Double check


VIKTORVAV99:
It looks like this function use "something" that is no loger supported (but not covered in the migration guide).
https://github.com/electricitymaps/electricitymaps-contrib/blob/fc0c0ba523ab1746ad03dc3668db6805032174dd/web/src/utils/state/atomWithCustomStorage.ts#L113-L117
But I can't actually find an instance where we use that function?

VIKTORVAV99:
I noticed that the onboarding is now flashing when reloading:

With Jotai v2:
https://github.com/electricitymaps/electricitymaps-contrib/assets/30777521/d1bdf75c-25e4-46c4-a63e-da70ec63f292

With Jotai v1:

https://github.com/electricitymaps/electricitymaps-contrib/assets/30777521/eaa74fe9-ae16-4040-aaef-f5cf75c940e2

tonypls:
> It looks like this function use "something" that is no loger supported (but not covered in the migration guide).
> 
> https://github.com/electricitymaps/electricitymaps-contrib/blob/fc0c0ba523ab1746ad03dc3668db6805032174dd/web/src/utils/state/atomWithCustomStorage.ts#L113-L117
> 
> 
> But I can't actually find an instance where we use that function?

This file currently serves as documentation for our ambitions to implement atoms with storage at some point. Maybe we should trim it ✂️

VIKTORVAV99:
> > It looks like this function use "something" that is no loger supported (but not covered in the migration guide).
> > 
> > https://github.com/electricitymaps/electricitymaps-contrib/blob/fc0c0ba523ab1746ad03dc3668db6805032174dd/web/src/utils/state/atomWithCustomStorage.ts#L113-L117
> > 
> > 
> > But I can't actually find an instance where we use that function?
> 
> This file currently serves as documentation for our ambitions to implement atoms with storage at some point. Maybe we should trim it ✂️ 

Can't we just comment the whole thing out? 😅

tonypls:
> I noticed that the onboarding is now flashing when reloading:
> 
> With Jotai v2: [electricitymaps/electricitymaps-contrib/assets/30777521/d1bdf75c-25e4-46c4-a63e-da70ec63f292](https://github.com/electricitymaps/electricitymaps-contrib/assets/30777521/d1bdf75c-25e4-46c4-a63e-da70ec63f292)
> 
> With Jotai v1:
> 
>  Inspelning.2023-07-06.175124.mp4

Good spotting.

"atomWithStorage util's delayInit is removed as being default. Also it will always render initialValue on first render, and the stored value, if any, on subsequent renders. The new behavior differs from v1. See https://github.com/pmndrs/jotai/discussions/1737 for more information."

madsnedergaard (approved):
LGTM! :)

PR #5565, opened 2023-07-06
`Mobile performance`:Hide exchange arrows on touch start

tonypls:
## Issue
Slow performance on older android devices

## Description
This PR hides the exchange arrows when the onTouchStart event handler is triggered. 

As discussed in PR #5501 the main performance benefit was seen from hiding the arrows, this implementation should keep the same user experience on desktop.

### Double check


madsnedergaard (approved):
Thanks! I still think the changes you made (like moving from atom to a prop for exchange arrow component) might be worth it, will keep the PR in mind (and written down) when we later look into performance more dramatically :)

PR #5577, opened 2023-07-08
Add zone for Southeast Alaksa Power Agency (Alaska)

### LINKED ISSUE #5552: Data for Wrangell, Petersburg and Ketchikan (Alaska, USA)
The Southeast Alaska Power Agency services Ketchikan, Wrangell and Petersburg in Alaska.

They have a page with SCADA data available: https://seapahydro.org/scada

The abbreviations seem pretty intuitive: TYL =TYEE LAKE hydropower plant, SWL = SWAN LAKE  hydropower plant; WRG = WRANGELL, PTG = PETERSBURG, KTN = KETCHIKAN

The page has hydro and diesel generation. As the latter is currently 0 MW, I think it's only used as a backup.

SEAPA SCADA

wrma6:
## Issue

Closes #5552 

## Description

- Re-define Alaska geometry to desginate zone for SEAPA (Ketchikan, Wrangell and Petersburg)
- Add parser for production data from https://seapahydro.org/scada

### Preview

image

### Double check


parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,66 @@
+from datetime import datetime, timezone
+from logging import Logger, getLogger
+from typing import List, Optional, Union
+
+from requests import Response, Session
+from requests_html import HTMLSession
```
VIKTORVAV99, line 6: Why are we using request_html here when we can get the data by calling the API directly with requests?
https://seapahydro.org/api/scada/index

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,66 @@
...
+from parsers.lib.exceptions import ParserException
+
+PARSER_NAME = "US_AK_SAP.py"
+
+SOURCE = "seapahydro.org"
+DATA_URL = "https://seapahydro.org/scada"
+
+
+def get_value_by_id(res: Response, id) -> float:
+    return float(res.html.find(f"#{id}", first=True).text)
```
VIKTORVAV99, line 20: In combination with the above calling the API directly means we would get a json object that we can convert to a dict directly and avoid using HTML parsing and scraping.

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    production_list = ProductionBreakdownList(logger=logger)
+    production_list.append(
+        zoneKey=zone_key,
+        datetime=datetime.now(timezone.utc),
+        production=ProductionMix(
+            gas=data["ktn_diesel"],
+            hydro=data["seapa_total"] + data["ktn_hydro"],
+        ),
```
VIKTORVAV99, line 54: if you need to add multiple source values to the same production mode you should be using the add_value() function of the ProductionMix class as it will run some basic validation on each source data instead of the combined source data.

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,66 @@
...
+
+    session.close()
+
+    return production_list.to_list()
+
+
+if __name__ == "__main__":
+    """Main method, never used by the Electricity Maps backend, but handy for testing."""
+
+    print(fetch_production(ZoneKey("AU-LH")))
```
VIKTORVAV99, line 66: This ZoneKey don't match.

VIKTORVAV99 (commented):
Some comments on the parser parts.
I thought it best to limit my review to the parser and when that is okay/approved move on to the other parts of the PR.

parsers/US_AK_SAP.py
wrma6: Ah yeah you're definitely right about that one, I should've checked for that. I'm a bit inexperienced at this kind of stuff, sorry.

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,64 @@
...
+        "seapa_total": get_value(data, "ss_mw"),  # 2 hydro plants owned by SEAPA
+        "ktn_hydro": get_value(
+            data, "ktn_hydro_mw"
+        ),  # hydro from pre-existing Ketchikan plants
+        "ktn_diesel": get_value(data, "ktn_diesel_mw"),  # backup Ketchikan diesel
+    }
+
+    production_mix = ProductionMix(gas=data["ktn_diesel"])
+    production_mix.add_value("hydro", data["seapa_total"], True)
+    production_mix.add_value("hydro", data["ktn_hydro"], True)
```
VIKTORVAV99, line 46:
```suggestion
    production_mix = ProductionMix()
    production_mix.add_value("gas", gas=data["ktn_diesel"], True)
    production_mix.add_value("hydro", data["seapa_total"], True)
    production_mix.add_value("hydro", data["ktn_hydro"], True)
```
This could also be done it a loop but since it's only 3 elements we can keep it like this for now.

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,64 @@
...
+
+    production_list = ProductionBreakdownList(logger=logger)
+    production_list.append(
+        zoneKey=zone_key,
+        datetime=datetime.now(timezone.utc),
+        production=production_mix,
+        source=SOURCE,
+    )
+
+    session.close()
```
VIKTORVAV99, line 56: This is not needed.
```suggestion
```

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,64 @@
...
+DATA_URL = "https://seapahydro.org/api/scada/index"
+
+
+def get_value(data: dict, key: str) -> float:
+    return float(data[key]["text"])
+
+
+def fetch_production(
+    zone_key: ZoneKey,
+    session: Session = Session(),
```
VIKTORVAV99, line 24: It has previusly been pointed out to us/me that having a class constructor as the default argument can cause problems (but shouldn't).
I suggesst setting this to:
```suggestion
    session: Optional[Session] = None,
```
And then do
`session = session or Session()` inside the function call.

VIKTORVAV99 (commented):
Just some small suggestions otherwise this seems to work well.

teejas:
hey, i've been looking into this PR + related issue because i was interested in contributing a new parser of my own and was viewing this for reference. i have a couple questions about the data being scraped from SEAPA and how it's being interpreted.

so based on the code as it is: it looks like `ktn_diesel_mw`, `ktn_hydro_mw`, and `ss_mw` (key names taken from the [SEAPA API response](https://seapahydro.org/api/scada/index)) are all being read as power produced by SEAPA. therefore they are all considered in a `fetch_production()` function.

according to the [SEAPA website](https://seapahydro.org/about): "The Southeast Alaska Power Agency (SEAPA) is a regional Joint Action Agency of the State of Alaska that owns the Swan Lake and Tyee Lake hydroelectric facilities, 14 miles of submarine cables, and 175 miles of overhead transmission lines serving Ketchikan, Wrangell, and Petersburg, Alaska." Meaning power is only produced at the Swan and Tyee Lake facilities, and it is transmitted to and consumed by the Ketchikan, Wrangell, and Petersburg cities.

therefore, i believe `ss_mw` is the **only** actual measure of power produced by SEAPA, it is the sum total of power produced by Swan and Tyee Lake hydroelectric facilities (`swl_mw` and `tyl_mw` can also be used to get the individual measures of each respective facility). Further, `wrg_mw`, `ptg_mw`, and `ktn_mw` are likely measures of how much power is being consumed or transmitted to the Alaskan cities of Wrangell, Petersburg, and Ketchikan respectively. I believe (but am less sure) that `ktn_hydro_mw` and `ktn_diesel_mw` is a further breakdown of `ktn_mw` to how much power from each specific source (hydro or gas) is being used?

in summary: `ss_mw` is the only actual measure of power being produced and should go in the `fetch_production()` function. `ktn_mw`, `wrg_mw`, and `ptg_mw` should be considered in a `fetch_consumption()` function.

p.s. excuse my long and wordy explanation, i wanted to be as clear as possible

teejas:
@wrma6 im making a PR to your PR: https://github.com/wrma6/electricitymaps-contrib/pull/1/

this splits it up into both fetch_production and fetch_consumption and references the data appropriately. im also looking into other data sources to supplement what is only hydroelectric and some consumption data (i.e. EIA and AK Energy Authority).

wrma6:
@teejas thanks for your comments on my PR!

What you said for SEAPA's power generation is correct (from what I understand), but I don't think that Ketchikan takes 100% of its power from SEAPA. I think that's what the `ktn_diesel_mw` and `ktn_hydro_mw` values are for. The diesel is just a backup so it's usually 0, but as shown in the dashboard there's multiple hydro plants, labelled as KTN, WHITMAN, BEAVER FALLS and UPPER SILVIS Lakes. If you add these values up, they equal the value reported by `ktn_hydro_mw` so I think that is meant to be counted in production as well. I also referred to this source, https://www.ktn-ak.us/hydro-electric, which talks about both the hydro and diesel generation in Ketchikan (and to me it seems consistent with the above observations).

Also, thanks for going through the effort of making a PR on top of mine. I did try to take a look, but there's a lot of files changed in there. Did you mean to add those changes in?

teejas:
> @teejas thanks for your comments on my PR!
> 
> What you said for SEAPA's power generation is correct (from what I understand), but I don't think that Ketchikan takes 100% of its power from SEAPA. I think that's what the `ktn_diesel_mw` and `ktn_hydro_mw` values are for. The diesel is just a backup so it's usually 0, but as shown in the dashboard there's multiple hydro plants, labelled as KTN, WHITMAN, BEAVER FALLS and UPPER SILVIS Lakes. If you add these values up, they equal the value reported by `ktn_hydro_mw` so I think that is meant to be counted in production as well. I also referred to this source, https://www.ktn-ak.us/hydro-electric, which talks about both the hydro and diesel generation in Ketchikan (and to me it seems consistent with the above observations).
> 
> Also, thanks for going through the effort of making a PR on top of mine. I did try to take a look, but there's a lot of files changed in there. Did you mean to add those changes in?

No those changes were just from merging main → your branch i believe. the only changes i made were to the file in `parsers/`.

wrma6:
No, I think those values aren't directly related to each other. `ktn_mw` is electricity from SEAPA itself while the other two values are not.

q--:
So I did some looking around, and [found on the website of Ketchikan](https://www.ketchikan.gov/hydro-electric) that Ketchikan Public Utilities owns the Ketchikan Lakes, Silvis Lakes, Beaver Falls, and (probably) Whitman Lake generation facilities, while SEAPA owns the Swan Lake  and Tyee Lake hydropower plants.  So "KTN HYDRO"  is probably generation from Ketchikan-owned hydro power plants, and "SEAPA SYSTEM TOTAL" doesn't include Ketchikan-owned generation (and seems to add up to the total generated by the SEAPA-owned "SWAN LAKE" and "TYEE LAKE" power plants

parsers/US_AK_SAP.py
```diff
@@ -0,0 +1,64 @@
...
+
+    data = {
+        "seapa_total": get_value(data, "ss_mw"),  # 2 hydro plants owned by SEAPA
+        "ktn_hydro": get_value(
+            data, "ktn_hydro_mw"
+        ),  # hydro from pre-existing Ketchikan plants
+        "ktn_diesel": get_value(data, "ktn_diesel_mw"),  # backup Ketchikan diesel
+    }
+
+    production_mix = ProductionMix(gas=data["ktn_diesel"])
```
q--, line 44: Shouldn't this be `oil`, not `gas`?

parsers/US_AK_SAP.py
q--, line 1: With the new parser naming conventions, this probably needs to be renamed to `SEAPA.py`.

web/public/locales/en.json
```diff
@@ -1508,6 +1508,10 @@
       "countryName": "USA",
       "zoneName": "Alaska"
     },
+    "US-AK-SAP": {
```
q--, line 1511: Not exactly sure what the policy for zone names is... It might make sense to use `SEAPA` instead of `SAP`, as that's the name of the power agency. @VIKTORVAV99, what do you think?

VIKTORVAV99: SEAPA would be slightly easier to remember and work with.

wrma6:
Thanks for the review guys, I've made the changes you both suggested. I see there are currently some lint errors for the file, but my local pylint complains when I try to satisfy the new conventions. What's the solution for this (or someone else can just push the lint changes)?

VIKTORVAV99:
> Thanks for the review guys, I've made the changes you both suggested. I see there are currently some lint errors for the file, but my local pylint complains when I try to satisfy the new conventions. What's the solution for this (or someone else can just push the lint changes)?

Make sure you have installed the latest dependencies with `poetry install -E parsers`, we switched from pylint, black and isort to ruff since they were so incredibly slow by comparison. Also make sure you are using python 3.10 or above to make sure all the new syntax works.

After that has been done you should run `poetry run format`, that will run the automatic fixes and formatting. Then you need to fix the rest manually, I'd recommend running `poetry run format` again after that though to ensure the no formatting mistakes slipped by (the code will be a lot shorter for the affected lines for example).

q-- (approved):
Looks fine to me as far as I can tell without actually running the code. Cypress e2e test failure seems unrelated to this PR

PR #5578, opened 2023-07-09
Add retries to CO parser for most recent data

### LINKED ISSUE #5503: CO production parser down
## Description

This is an automatic error report generated for Colombia (CO).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser CO production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/CO.html">Explore the runtime logs</a>

wrma6:
## Issue

Closes #5503 

## Description

Add retries (each retry moves the datetime back by one day) to the CO parser, since data freshness seems unreliable.

### Double check


VIKTORVAV99 (commented):
Looks good for the most part, just have some small comments to make this easier to maintain in the future.

wrma6:
Followed your suggestion, code should be simplified now.

VIKTORVAV99:
I think this is fine at least for now. I in the future I would like to rewrite the parser a little so it don't combine values for the modes in the mapping before assigning it to the electricity maps mode and use the add_value function instead. That way we would be getting a more standardized parser and more granular correction of negative consumption. At this point we could introduce a fetch data function.

The changes you have here works well so no real comments on those but there are some unbound type errors with the new if statements that I would like to see fixed before merging this. Let me know if you need any help with it and I can push the needed changes directly.

wrma6:
Hmm, I don't see the unbound type errors... is it coming from poetry run lint?

VIKTORVAV99:
It's coming from the pylance python extension in VS Code. We have been waiting to do me extensive type checking in CI but unfortunately many of the old parsers have type errors that needs to be resolved first or it would never pass.

wrma6:
Ok, I tried to fix the warnings from pylance, let me know if that worked

VIKTORVAV99 (approved):
I added the last changes needed for the typing to be accurate but nice work!

Everything else looks good so this is ready to be merged!
Thanks for all the help!

PR #5585, opened 2023-07-10
Add LegacyHttpAdapter for TR parser SSL certificate issue

### LINKED ISSUE #5482: GB-ORK production parser down
## Description

This is an automatic error report generated for Great Britain Orkney Islands (GB-ORK).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser GB-ORK production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/GB-ORK.html">Explore the runtime logs</a>

### LINKED ISSUE #5484: TR production parser down
## Description

This is an automatic error report generated for Turkey (TR).

Issues:
- No recent data found for `consumptionForecast` parser
- No recent data found for `generationForecast` parser
- No recent data found for `productionPerModeForecast` parser

## Suggestions
- Try running the parser locally using the command `poetry run test_parser TR production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/TR.html">Explore the runtime logs</a>

wrma6:
## Issue

Closes #5484
Closes #5482

## Description

Solve the SSL issue that is breaking the TR parser - use older SSL version instead of the current one used by the Google Cloud machines.

### Double check


parsers/TR.py
```diff
@@ -37,13 +39,30 @@
...
+            num_pools=connections, maxsize=maxsize, block=block, ssl_context=ctx
+        )
+
+
+# Use a LegacyHttpAdapter to avoid "unsafe legacy renegotiation disabled" error (issue #5484)
+# Original code source: https://stackoverflow.com/questions/71603314/ssl-error-unsafe-legacy-renegotiation-disabled
+def create_legacy_session():
+    session = Session()
+    session.mount("https://", LegacyHttpAdapter())
+    return session
```
VIKTORVAV99, line 56:
```suggestion
def get_legacy_session(session: Session):
    return session.mount("https://", LegacyHttpAdapter())
```

parsers/TR.py
```diff
@@ -37,13 +39,30 @@
...
+
+
 def fetch_data(session: Session, target_datetime: datetime, kind: str) -> dict:
     url = "/".join((EPIAS_MAIN_URL, KINDS_MAPPING[kind]["url"]))
     params = {
         "startDate": (target_datetime).strftime("%Y-%m-%d"),
         "endDate": (target_datetime + timedelta(days=1)).strftime("%Y-%m-%d"),
     }
-    r: Response = session.get(url=url, params=params)
+    r: Response = create_legacy_session().get(url=url, params=params)
```
VIKTORVAV99, line 65:
```suggestion
    r: Response = get_legacy_session(session).get(url=url, params=params)
```

VIKTORVAV99 (commented):
I don't think we should be creaing a new session all the time.

So let's re-use the session we already create and just mount the adapter on it.

wrma6:
With your suggested change, we would be mounting the LegacyHttpAdapter to the session for all subsequent uses of it then, right? To me that seems like something we don't want since we should use the standard adapter (which is technically the more "secure" one) for the other sites if we can.

VIKTORVAV99:
> With your suggested change, we would be mounting the LegacyHttpAdapter to the session for all subsequent uses of it then, right? To me that seems like something we don't want since we should use the standard adapter (which is technically the more "secure" one) for the other sites if we can.

While that shouldn't happen I do realize there is a possibility of that. But right now we are creating a Session when the fetch_production function is first called, then yet another one when calling the new function.

So if you are still concerned with the above change I suggest we set the Optional session in the fetch functions to none instead. That way we are only creating it once either way.

wrma6:
No concerns from my side about your approach, I was in fact unaware that the default is for the parser to be called without a pre-existing Session parameter. I took your suggestion and modified it a little bit, let me know what you think.

VIKTORVAV99 (approved):
LGTM! 🎉

I pushed some changes that are unrelated to your PR on it's own but that has raised som potential concerns in another PR a while back.

However since what we are changing could be fairly important I would like another set of eyes on it before merging.

parsers/TR.py
```diff
@@ -37,12 +39,28 @@
...
+        self.poolmanager = urllib3.poolmanager.PoolManager(
+            num_pools=connections, maxsize=maxsize, block=block, ssl_context=ctx
+        )
+
+
+# Use a LegacyHttpAdapter to avoid "unsafe legacy renegotiation disabled" error (issue #5484)
+# Original code source: https://stackoverflow.com/questions/71603314/ssl-error-unsafe-legacy-renegotiation-disabled
+def mount_legacy_adapter(session: Session):
+    session.mount("https://", LegacyHttpAdapter())
+
```
VIKTORVAV99, line 55: @unitrium I noticed #5482 is suffering the same problem, do you think it would make sense to bring this into a lib instead?

unitrium: Yes, I think it would make sense to reuse this.

wrma6:
Hey, I added the changes as (I think) you wanted, let me know what you think.

parsers/lib/session.py
```diff
@@ -0,0 +1,19 @@
...
+        ctx.options |= 0x4  # OP_LEGACY_SERVER_CONNECT
+        self.poolmanager = urllib3.poolmanager.PoolManager(
+            num_pools=connections, maxsize=maxsize, block=block, ssl_context=ctx
+        )
+
+
+# Use a LegacyHttpAdapter to avoid "unsafe legacy renegotiation disabled" error
+# Original code source: https://stackoverflow.com/questions/71603314/ssl-error-unsafe-legacy-renegotiation-disabled
+def mount_legacy_adapter(session: Session):
+    session.mount("https://", LegacyHttpAdapter())
```
VIKTORVAV99, line 19: This works great for the current parser but to make it a little more robust I suggest we change it to something like this instead:
```suggestion
def get_legacy_adapter(session: Session, logger: Logger):
    local_session = deepcopy(session)
    if "https://" in local_session.adapters or "http://" in local_session.adapters:
        logger.warning("Session already has adapters, they will be overridden.")
    local_session.mount("https://", LegacyHttpAdapter())
    local_session.mount("http://", LegacyHttpAdapter())
    return local_session
```
This adopts a functional style for the function to avoid poluting the original session object with the changes.

I double checked in the backend and a new session is only created per *parser* not per parser function. For most parsers this won't be an issue as all the functions call the same domain. But a few parsers have diffrent domains for say production and consumption. The above changes make sure we only change it in the parser function that is called.

VIKTORVAV99 (changes requested):
I added a suggestion that will make this more reusable and a little more safe to use.

If you agree with the changes feel free to commit them directly otherwise let's discuss it a bit.

parsers/lib/session.py
VIKTORVAV99: Another option to the above would be to create a new session in this function instead, like this:
```suggestion
def get_session_with_legacy_adapter():
	    session = Session();
	    session.mount("https://", LegacyHttpAdapter())
	    session.mount("http://", LegacyHttpAdapter())
	    return session
```
It would simplify things a bit.

wrma6:
I think in the case that we're looking to deep copy the session, we might as well just create a new session instead. Let me know if you would prefer the other option and I can change it.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for finding a relatively simple fix for the issue!

PR #5590, opened 2023-07-12
Fix Alaska Bounding Box

VIKTORVAV99:
## Issue
Bo Dong alerted us on slack that the Alaska bounding box was incorrect. This is due to the fact it crosses the 180th meridian and our current script can't handle this.

## Description
I have manually updated the bounding box and added a override for it in the bounding box generation script.

### Double check


madsnedergaard (approved):
Oh cool, thanks for fixing it!

Issue #5589, opened 2023-07-12
pumped hydro consumption shown as generation in some parts of Europe


KabelWlan:
## Description
Generation is displayed correctly but it seems like consumption is shown as generation as well. I've seen this happen for Austria, Belgium, Germany, Italy, Lithuania, Portugal, Slovakia and Spain (all countries with available pumped hydro consumption on EntsoE).

Here's an example:
![image](https://github.com/electricitymaps/electricitymaps-contrib/assets/49598085/b108107c-00cb-49d0-9d7f-7aeecc2e284a)

VIKTORVAV99:
I'll look into this as soon as possible.

PR #5591, opened 2023-07-12
Fix ENTSOE.py storage consumption

VIKTORVAV99:
## Issue
Fixes: #5589 

## Description

Adds back the is_production check and inverts the value if it's not true to fix the consumption values.

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -810,9 +810,12 @@ def parse_production(
         for entry in timeseries.find_all("point"):
             quantity = float(entry.find_all("quantity")[0].contents[0])
             position = int(entry.find_all("position")[0].contents[0])
+            is_production = (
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -810,9 +810,12 @@ def parse_production(
         for entry in timeseries.find_all("point"):
             quantity = float(entry.find_all("quantity")[0].contents[0])
             position = int(entry.find_all("position")[0].contents[0])
+            is_production = (
```
unitrium, line 813: Could you add a comment describing what this is for here ?

unitrium (approved):
Ah ok that's what this was for!
Could you just add a comment on the is_production?

parsers/ENTSOE.py
VIKTORVAV99: Sure thing!

PR #5594, opened 2023-07-12
Update PT.yaml

Musketeiro:
Updated solar capacity per the official figures from REN for June 2023.


VIKTORVAV99 (approved):
Thanks for keeping these updated! 🎉

PR #1328, opened 2018-04-12
More bounding boxes

maxbellec:
`getBoundingBoxes.json` generates a bounding box with the min and max lat and lon of every zone, plus a 0.5 padding on each side.

Comparison with current data shows it's very similar
```
NO
-9.62 57.49 31.58 71.68 
4.26 57.73 20.17 70.87 

BO
-70.17 -23.40 -56.97 -9.18 
-69.59 -22.87 -57.50 -9.76 

MN
87.24 41.09 120.41 52.63 
87.75 41.60 119.77 52.05 

OM
51.48 16.14 60.34 26.89 
52.00 16.65 59.81 26.40 

JP
122.44 23.71 154.49 46.02 
129.41 31.03 145.54 45.55 

KE
33.39 -5.18 42.39 5.53 
33.89 -4.68 41.86 5.51 

ZM
21.48 -18.57 34.17 -7.69 
21.89 -17.96 33.49 -8.24 

PL
13.62 48.49 24.64 55.34 
14.07 49.03 24.03 54.85 

BI
28.49 -4.96 31.33 -1.80 
29.02 -4.50 30.75 -2.35 

MA
-17.51 20.92 -0.53 36.43 
-17.02 21.42 -1.12 35.76 

TF
39.23 -50.22 78.09 -11.05 
68.72 -49.77 70.56 -48.62 

MM
91.67 9.29 101.67 29.04 
92.30 9.93 101.18 28.34 

HT
-74.99 17.53 -71.14 20.59 
-74.46 18.03 -71.62 19.92 

AT
9.02 45.88 17.65 49.51 
9.48 46.43 16.98 49.04 

GN
-15.58 6.69 -7.16 13.17 
-15.13 7.31 -7.83 12.59 

DZ
-9.18 18.48 12.47 37.59 
-8.68 19.06 12.00 37.12 

CR
-87.62 5.02 -82.06 11.71 
-85.94 8.23 -82.55 11.22 

MZ
29.71 -27.36 41.35 -9.97 
30.18 -26.74 40.78 -10.32 

SK
16.34 47.25 23.04 50.10 
16.88 47.76 22.56 49.57 

CH
5.45 45.32 10.97 48.30 
5.96 45.82 10.49 47.81 

IL
33.75 28.99 36.39 33.91 
34.27 29.50 35.84 33.28 

IS
-25.04 62.90 -13.00 67.06 
-24.33 63.50 -13.61 66.53 

HN
-89.86 12.48 -82.63 17.92 
[...]

PR #5596, opened 2023-07-12
[WIP] Adds many more missing bounding boxes

madsnedergaard:
## Issue

Follow up to #1328 

## Description

Turns out there were a lot of zones that also didn't have bounding boxes 😅 
Guess it's less important when we don't have data for the zones, but we might as well just add the bounding boxes now so that they are ready for when we have parsers!

I also fixed a few odd ones and made the script able to detect missing boxes.

**Impact:** Since we exclude bounding boxes in frontend, it doesn't change anything. It might make the need to switch yaml parser library more urgent though :)

### Preview

use https://geojson.io and open this geojson file with all the lines:

https://www.dropbox.com/s/arfylzh165nxje0/lines.geojson?dl=0


web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -6,117 +6,139 @@ import { saveZoneYaml } from './files.js';
...
 
-if (!zoneConfig) {
-  console.error(`ERROR: Zone ${zoneKey} does not exist in configuration`);
-  process.exit(1);
-}
+  // do not modifiy current bounding boxes
+  if (zoneConfig.bounding_box) {
+    console.error(`Ignoring ${zoneKey} as it already have a bounding box`);
+    return;
+  }
```
VIKTORVAV99, line 30: Why do we want to ignore existing zones? If they have changed the bouding box should also be updated (like with the texas change).

web/geo/generateZoneBoundingBoxes.ts
```diff
@@ -6,117 +6,139 @@ import { saveZoneYaml } from './files.js';
...
-  boundingBoxes[zone.properties.zoneName] = [
-    [minLon - 0.5, minLat - 0.5],
-    [maxLon + 0.5, maxLat + 0.5],
-  ];
-}
+  if (isAggregate) {
+    const averageBoundingBox: number[][] = [
+      [200, 200],
+      [-200, -200],
+    ];
```
VIKTORVAV99, line 89: Do we need a average bouding box or could we use a initial state of Number.NaN istead?
I think it would make more sense to start from zero so to speak.

Maybe call it temp bounding box insteade of average as well.

VIKTORVAV99 (commented):
For the most part this looks good, I have some general suggestions and thoughts though.

PS: Why is US-NW-AVRN not in the world file?

madsnedergaard:
Closing this, maybe some day in the future ¯\\\_(ツ)\_/¯

PR #5623, opened 2023-07-20
generate arrows working

tonypls:
## Issue
Our generate-arrows.mjs script no longer worked.
## Description
This PR updates the files required for our generate arrows script that creates animated exchange arrows over a range of co2 intensities and colors.

This PR was motivated by investigating how our change in CO2 intensity range in #4789 effects the difference between the map zone colors and the exchange colors. This is because we use discrete range steps for each arrow color (0, 80, 160...)

Updating the range (from 0 - 800 to 0 - 1500) of our exchange gifs actually makes the problem worse. I suggest we leave the current arrow color range and perhaps in the future do something smart with SVGs that matches the color exactly.

I've included what's required to run generate arrows below just incase.

### Double check


madsnedergaard:
Oh, did we accidentally remove the highlights at some point? And annoying that it wasn't any better :)

Seems like there's duplicate assets in this PR in both public/images and in scripts?

madsnedergaard (approved):
LGTM! :)

PR #5640, opened 2023-07-26
Add zone disclamers for production transfers in the US

unitrium:
## Issue

In the US we have some generation-only zones that have been integrated with other zones directly. We should therefore add a disclamer and explicit our methodology.

## Description

Add a disclaimer to the zones that have been integrated and point towards the wiki page.
For now disclaimers don't support links.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5642, opened 2023-07-26
Incorporate US-FLA-NSB into US-FLA-FPL and US-SE-AEC into US-SE-SOCO

pierresegonne:
## Issue

Some of the deprecated US zones have been incorported into other balancing authorities. 

The EIA still returns historical data for those zones, so I suggest we merge the historical data for those zones into the zones that they merged with.

## Description

Merges the deprecated zones with the following strategy:

- `US-SW-GRMA`. A former generation only BA, with only one gas plant. Merged with `US-SW-AZPS`.
There is no historical production data, but historical data for the `US-SW-AZPS->US-SW-GRMA` exchange. I'm proposing not to do anything about this. We could change the parser so that the historical production for `AZPS` adds the exchanged power as gas power, but that would require non-trivial changes to the parser, and has some not so certain implications for the data (does the EIA already account for that production for AZPS??)

- `US-SE-AEC`. Got merged with `US-SE-SOCO`. Transferred the production, cancelled the internal exchange `US-SE-AEC->US-SE-SOCO` and transferred the exchanged to/from AEC to SOCO.
- `US-FLA-NSB`. Got merged with `US-FLA-FPL`. Transferred the production, cancelled the internal exchange `US-FLA-FPL->US-FLA-NSB` and transferred the exchanged to/from NSB to FPL.

I also moved the zones config to the retired_zones folder instead of simply deleting them altogether. Not sure it's the intended purpose.

### Double check


unitrium:
Hum I'm a bit wondering about historical exchanges if we do this. Wouldn't this cause those production being counted twice, once through the exchanges and once through the integration?

VIKTORVAV99:
> Hum I'm a bit wondering about historical exchanges if we do this. Wouldn't this cause those production being counted twice, once through the exchanges and once through the integration?

Agreed, I think we need to remove the exchange between the parent zone and the child zone and then re-map and aggregate all the other exchanges for the zones so we get an accurate accounting of the exchange and production data.

pierresegonne:
The exchanges between the parent/children zones have already been commented out a while ago.

I'll check whether we have exchanges with external zones

pierresegonne:
FYI I'm capturing all my learnings on the internal issue. Will transfer the consolidated knowledge here once I'm confident about a plan of action

pierresegonne:
I've updated the description and the PR is ready to be reviewed again. Thanks for the initial critical assesment :)

VIKTORVAV99 (approved):
Looks good to me!

Thanks for implementing the suggested changes.
PS: Nice that you added back the configs in the retired_zones folder.

pierresegonne:
If you have any objections / doubts @mathilde-daugy @unitrium let me know sooner rather than later, as I'll start reprocessing the data soon after contrib has been updated with this change

PR #5646, opened 2023-07-27
Fix DK parser

mathilde-daugy:
## Issue

in the request, the limit is set to 144 which means that we are only collecting data for half of the day 
image

## Description

`limit` in `params` in `fetch_data` is increased to 500 to make sure that we collect 288 datapoints as we should

### Preview

image

### Double check

### DIFF
--- a/parsers/DK.py
+++ b/parsers/DK.py
@@ -36,14 +36,14 @@ def fetch_data(
         target_datetime = target_datetime.replace(tzinfo=None)
 
     params = {
-        "limit": 144,
+        "limit": 500,
         "filter": '{"PriceArea":"DK1"}'
         if price_area == "DK1"
         else '{"PriceArea":"DK2"}',
-        "start": (target_datetime - timedelta(days=1)).isoformat(timespec="minutes")
+        "start": (target_datetime - timedelta(days=1)).strftime("%Y-%m-%d")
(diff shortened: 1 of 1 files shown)


parsers/DK.py
```diff
@@ -100,34 +103,26 @@ def fetch_exchange(
...
-    else:
-        return return_list
+        return all_exchange_data.to_list()
 
 
 if __name__ == "__main__":
-    print("fetch_exchange(DK-DK2, SE-SE4) ->")
-    print(fetch_exchange("DK-DK2", "SE-SE4"))
+    print("fetch_exchange(DK-DK1, DE) ->")
+    print(fetch_exchange("DK-DK1", "DE", target_datetime=datetime(2023, 3, 27)))
```
unitrium, line 128: I would advise not to include those changes.

parsers/DK.py
```diff
@@ -60,7 +60,10 @@ def fetch_data(
         else:
             return data
     else:
-        raise ParserException(parser="DK.py", message="No data was returned")
+        raise ParserException(
```
unitrium, line 63: Please add the zone_key option as well.

unitrium (approved):
Looks good, please just fix the two comments.

Issue #5627, opened 2023-07-23
Add display for aggregated net cross-border electricity exchange


glaubitz:
Currently, Electricity Maps will display electricity imports and exports for a selected country brown down by county.

It would be nice to have an additional bar shown for each country which will show the net cross-border electricity exchange, so that it's immediately visible how much of the electricity available within a country comes from imports (positive flow) or how much is exported (negative flow).

Issue #4430, opened 2022-08-08
Create time series graph of net exchanges to give insight into net autarky


TheNewCivilian:
**Is your feature request related to a problem?**
To monitor and tackle the rise of a possible energy crisis, per country energy autarky or clear contract regulated energy dependence needs to be preserved.

**Describe the solution you'd like**
To get a clearer view on the energy autarky of a country a time series representation of total summed import / export volumes could be very helpful.

![Frame 49](https://user-images.githubusercontent.com/24456478/183401285-d776148f-2f0a-4c66-8470-99d80cdd356a.png)

**Describe alternatives you've considered**
A new map view setting, showing all live dependencies following a positive (exporting), neutral, negative (importing) color scheme. e.g. In case of an global energy shortage we could see more and more countries going red.

VIKTORVAV99:
I'm not sure this would accurately represent energy dependance/independence as there is market conditions that causes a country to import/export even when it's not strictly needed from a grid security standpoint. This would also be a issue with countries that have internal bidding zones where there is a internal market where one bidding zone is constantly exporting and others are constantly importing. This is only natural and happens in every country but internal bidding zones make it much more apparent.

However it would be nice with a graph, similar to the price graph, that showed the net import/exports without any mention of dependance/independence.

Kongkille:
We are also intending to work on a feature to aggregate regions into countries so it's easier to compare countries to each other. In that case, I also think this feature becomes even more relevant. 

If you agree that this also addresses your original issue @TheNewCivilian , perhaps we rename this issue to be about creating a time series graph of net exchanges?

PR #5647, opened 2023-07-27
Net exchange graph

VIKTORVAV99:
## Description

This PR add a net exchange graph to visualize the net flows for a zone and includes support for both power (MW) and emissions (gco2/min).

Closes: #4430
Closes: #5627

### Preview:
image

TODO: 

- [x] Make it scale better.
- [x] Create a NetExchange tooltip
- [x] Fix Net Exchange tooltip sizing.
- [x] #5700 (Also cherry picked to this branch).
- [x] Hide graph when in production mode.
- [x] #5709 (Also cherry picked to this branch).
- [x] Investigate why the graphs don't show the correct scaling for a few zones.
- [x] Fix scaling in when there is only export.
- [x] Add carbon emission exchange values (when in emission mode).

### Double check


FelixDQ:
This is really cool @VIKTORVAV99. For inspiration, we built an app for ENTSO-E a long time ago showing the net exchange for EU countries ([ios link](https://apps.apple.com/us/app/entso-e-transparency-platform/id1481574577) & [android link](https://play.google.com/store/apps/details?id=eu.entsoe.tpapp&hl=en&gl=US&pli=1)). It was based on an old forked version of Electricity Maps.

image
Blue countries are net exporting, and yellow countries are net importing.

VIKTORVAV99:
> This is really cool @VIKTORVAV99. For inspiration, we built an app for ENTSO-E a long time ago showing the net exchange for EU countries ([ios link](https://apps.apple.com/us/app/entso-e-transparency-platform/id1481574577) & [android link](https://play.google.com/store/apps/details?id=eu.entsoe.tpapp&hl=en&gl=US&pli=1)). It was based on an old forked version of Electricity Maps.
> 
> image Blue countries are net exporting, and yellow countries are net importing.

Oh that's pretty cool! I was not aware of that.

For this issue/PR in particular I think I (or we) will only focus on the net exchange graph but I think it would be pretty cool to add a view like this to the map. @madsnedergaard and @tonypls what do you think about this?

madsnedergaard:
So cool to come back from vacation and see this PR! 😍 
I think the graph is a really cool first (but very important!) step in making exchanges more understandable to people. A view would be cool, but I also think we could experiment with changing the arrows so that they communicate the same thing more clearly :)

web/src/features/charts/hooks/useNetExchangeChartData.ts
```diff
@@ -0,0 +1,93 @@
...
+    valueAxisLabel,
+    layerStroke: undefined,
+  };
+
+  return { data: result, isLoading, isError };
+}
+
+interface ValuesInfo {
+  valueAxisLabel: string; // For example, GW or tCO₂eq/min
+  valueFactor: number; // TODO: why is this required
```
VIKTORVAV99, line 72: This is needed so the y axis scales properly.

web/public/locales/en.json
```diff
@@ -146,7 +147,10 @@
     "selectLanguage": "Select language",
     "lowCarbDescription": "Includes renewables and nuclear",
     "dataIsDelayed": "Data is delayed",
-    "price": "Price"
+    "price": "Price",
+    "netExchange": "Net exchange",
+    "importing": "Importing",
+    "exporting": "Exporting"
```
VIKTORVAV99, line 153: Could probably use more descriptive strings here to avoid confusion. Any suggestion?

VIKTORVAV99 (commented):
Lots of stuff has changed in here but I have tried to split it up the global changes in separate PRs so all changes should be tied together and reviewed as one. 

Don't hold back on the feedback!

silimotion:
I have a suggestion that can improve the usefulness of the graph. We could add how much percentage of the energy supply do the imports/exports represent, similarly to how each connection is represented in the origin of electricity graph. E.g "Exporting 413 MW, representing 11.2% of electricity available in #country". This could also be done in the emissions tab: "representing 11.2% of emissions".

On another note, I think it would be more aesthetically pleasant if the numbers were in bold, as in the price graph

VIKTORVAV99:
> I have a suggestion that can improve the usefulness of the graph. We could add how much percentage of the energy supply do the imports/exports represent, similarly to how each connection is represented in the origin of electricity graph. E.g "Exporting 413 MW, representing 11.2% of electricity available in #country". This could also be done in the emissions tab: "representing 11.2% of emissions".
> 
> On another note, I think it would be more aesthetically pleasant if the numbers were in bold, as in the price graph

That's a good suggestion but I think it would be better to do that in a follow up PR so this doesn't grow even bigger.

I'll see about making the value in the tooltip bold though.

web/src/features/charts/hooks/useNetExchangeChartData.ts
```diff
@@ -0,0 +1,93 @@
...
+    markerFill,
+    valueAxisLabel,
+    layerStroke: undefined,
+  };
+
+  return { data: result, isLoading, isError };
+}
+
+interface ValuesInfo {
+  valueAxisLabel: string; // For example, GW or tCO₂eq/min
```
VIKTORVAV99, line 71: In a follow up PR I will introduce strict typing here and add enums for the power values. (And implement those across the board).

web/public/locales/en.json
madsnedergaard: I think this is okay :)

web/src/features/charts/hooks/useNetExchangeChartData.ts
madsnedergaard: Let's remove the TODO comment then :D

web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
```diff
@@ -25,7 +25,7 @@ export default function AreaGraphToolTipHeader(props: AreaGraphToolTipHeaderProp
             }}
             className="rounded-sm font-bold"
           ></div>
-          <p className="text-base">{title}</p>
+          <p className="pr-4 text-base">{title}</p>
```
madsnedergaard, line 28: It's a tricky balance here, cause this might break things a bit for a few languages. Maybe just do `pr-2` instead?
Screenshot 2023-08-16 at 21 32 17

web/src/utils/helpers.ts
```diff
@@ -108,3 +108,33 @@ export function getRenewableRatio(
...
   return (isConsumption ? renewableRatio : renewableRatioProduction) ?? Number.NaN;
 }
+
+/**
+ * Function to round a number to a specific amount of decimals.
+ * @param {number} number - The number to round.
+ * @param {number} decimals - Defaults to 2 decimals.
+ * @returns {number} Rounded number.
+ */
+export const round = (number: number, decimals = 2): number => {
```
madsnedergaard, line 118: Hmm, not sure why we need this function compared to `.toFixed()`? We also have the same in `geo/utilities.ts`, dunno if we should try merging those or just leave it like this

madsnedergaard (commented):
This is so cool @VIKTORVAV99, thanks for implementing it! 😍 
A few very minor comments

web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
VIKTORVAV99: I'll see if I can find something that works. But in general I think we should scale the tooltip (on desktop at least) so the title and datetime always fit.

web/src/utils/helpers.ts
VIKTORVAV99: Technically to fixed don't round things but cuts them off.

So 5.967 would be 5.96 instead of 5.7 with toFixed (that is if I remember correctly).

But I'm hoping we move the netExchage calculation to the app-backend (netCO2exchange is already in there) and maybe rework how the formatting work so we return the correct number instead of the division factor.

But that's something for future improvements.

web/src/features/charts/hooks/useNetExchangeChartData.ts
VIKTORVAV99: This was more to get everyone on the same page since these TODO comments are in more graphs.

Plan was to remove them once reviews where in progress and everyone was on the same page.

web/src/utils/helpers.ts
VIKTORVAV99: Here is another difference between the two:
```

> const round = (number, decimals = 2) => {  return (    (Math.round((Math.abs(number) + Number.EPSILON) * 10 ** decimals) / 10 ** decimals) *    Math.sign(number)  );};
undefined
> round(0.0001)
0
> 0.0001.toFixed(2)
'0.00'
>
```
So I do like "my" round function better as it gives:

1. Number output
2. The expected amount of digits.

web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
madsnedergaard: Makes sense if possible :)

web/src/utils/helpers.ts
madsnedergaard: Good point, thanks for the explanation - let's stick with yours:)

madsnedergaard (approved):
🥳🎉

tonypls:
Cool to see the relationship between renewables and imports like this:
image

VIKTORVAV99:
> Cool to see the relationship between renewables and imports like this:
> image
> 

Yeah, it's pretty interesting! Looking forward to getting it out there and seeing what users say.

PR #5677, opened 2023-08-01
Dependency maintenance

VIKTORVAV99:
## Description

Some general dependency maintenance and type updates to correct usage after todays dependency updates.

### Double check


web/src/components/Flag.tsx
```diff
@@ -7,10 +7,7 @@ function getCountryName(zoneId: string) {
   return flagName.toUpperCase();
 }
 
-type HTMLSVGElement = HTMLElement & SVGElement;
-interface CountryFlagProps
-  extends React.HTMLAttributes<HTMLSVGElement>,
-    React.SVGAttributes<HTMLSVGElement> {
```
VIKTORVAV99, line 13: This caused:
```
src/components/Flag.tsx:11:11 - error TS2320: Interface 'CountryFlagProps' cannot simultaneously extend types 'HTMLAttributes<HTMLSVGElement>' and 'SVGAttributes<HTMLSVGElement>'.
  Named property 'children' of types 'HTMLAttributes<HTMLSVGElement>' and 'SVGAttributes<HTMLSVGElement>' are not identical.

11 interface CountryFlagProps
```

web/src/translation/translation.ts
```diff
@@ -1,11 +1,10 @@
-import { TFunction } from 'i18next';
 import { useTranslation as useTranslationHook } from 'react-i18next';
 import { vsprintf } from 'sprintf-js';
 import i18next from './i18n';
 
 // Todo: We should get rid of vsprintf and use i18next interpolation instead
 const translateWithTranslator = (
-  translator: TFunction,
```
VIKTORVAV99, line 8: This caused:
```
src/translation/translation.ts:1:10 - error TS2614: Module '"i18next"' has no exported member 'TFunction'. Did you mean to use 'import TFunction from "i18next"' instead?

1 import { TFunction } from 'i18next';
```
We could also import t directly from i18next if you prefer, the type should be the same.

VIKTORVAV99 (commented):
Some of the dependency updates I did earlier today "fixed" their own typing which in turn broke ours... Most likely cause we where using them wrong.

web/src/translation/translation.ts
tonypls: Might be a tinsy bit nicer to import the type, but up to you :)

tonypls (approved):
LGTM

web/src/translation/translation.ts
VIKTORVAV99: t has a type defined internally as:
```ts
t: TFunction<[DefaultNamespace, ...Exclude<FlatNamespace, DefaultNamespace>[]]>;
```
But TFunction is not exported so this way felt a little simpler without having to do any workaround or importing directly from their internal type files.

PR #5691, opened 2023-08-02
Return ProductionMix fields when set to None

unitrium:
## Issue

By design in the ProductionBreakdown, the storage and production mix were excluding all modes whose values were None.

This was done to avoid creating data that doesn't exist. Indeed, returning None for all modes even when the datasource didn't give any information about this mode was thought to be problematic because we would be creating data that doesn't exist.

However, some zones like US-SW-PNM started returning Nones for some production modes instead of zeros randomly. Therefore this interfers with the exclusion, as modes set to None are excluded.

Ideally we want to exclude modes that have not been set, whose default value is therefore None but not modes that have been intentionally set to None.

## Description

- Update the method `to_dict` from ProductionBreakdown to exclude only unset modes.
- Add testing to make sure the modes are correctly returned through all possible ways of creating them (object instanciation, `add_value`, `set_value`)
- Update the base`add_value` method from Mix so that all paths go through `__setattr__` as it will mark the mode as set. Otherwise calling `add_value` with a None would not set the mode properly.

### Double check


VIKTORVAV99:
> @VIKTORVAV99, I'm also thinking that actually the `set_value` method is rendered a bit useless now that we have the `add_value` method. I will set up a PR to remove it and simplify the code.

I was thinking it could still be used for creating overrides but I suppose you could assign the value directly to the object property if that is the case.

unitrium:
> I was thinking it could still be used for creating overrides but I suppose you could assign the value directly to the object property if that is the case.

I'm not sure overrides are something we would like to encourage in the future, ideally everything should be supported by our interface. So to keep things simple the add_value method should be sufficient.

VIKTORVAV99 (approved):
LGTM! 🎉

electricitymap/contrib/lib/models/events.py
```diff
@@ -546,11 +545,11 @@ def to_dict(self) -> Dict[str, Any]:
             "datetime": self.datetime,
             "zoneKey": self.zoneKey,
             "production": self.production.dict(
-                exclude_none=True, keep_corrected_negative_values=True
+                exclude_unset=True, keep_corrected_negative_values=True
```
pierresegonne, line 548: Nice that pydantic allows this

PR #5693, opened 2023-08-02
Fix ES->MA connection

### LINKED ISSUE #5645: Wrong data in Spain<->Morocco connection
## Description
Just as I suspected, the data for the Spain<->Morocco connection is wrong, as can be seen on the [ESIOS webpage](https://www.esios.ree.es/en/analysis/10209) (ignore the lack of data during last night, that's not linked to this).
It is related to the fact that the parser returns MWh of fifteen minutes intervals, while the backend expects MW, so averages instead of adding.
Apart from changing something in the backend, the easiest solution that I found is forcing the API to return hour intervals. We would have to add a small delay also, to prevent returning incomplete data. (e.g. server only adds 11:00 and 11:15 if the provided date is 11:24)
We could alternatively multiply the 15 minutes interval data by 4 to get the MW. This would increase the granularity but would imply adding specific cases for this exchange, the parser would break if the granularity ever changes, etc. Whereas by forcing the granularity to 1 hour we make sure the granularity is always correct.
What is your opinion?

silimotion:
## Issue

Closes #5645 

## Description

I added a new mapping to take into account the granularity of the data, as discussed in #5645

### Double check

### DIFF
--- a/config/zones/TW.yaml
+++ b/config/zones/TW.yaml
@@ -4,17 +4,17 @@ bounding_box:
   - - 121.951243931
     - 25.2954588893
 capacity:
-  biomass: 38.2
-  coal: 14681
-  gas: 18365
-  geothermal: 6.4
-  hydro: 2094
-  hydro storage: 2609
-  nuclear: 2865
-  oil: 1586
-  solar: 7700
+  biomass: 92
+  coal: 14697
+  gas: 18380
+  geothermal: 5
+  hydro: 2098
+  hydro storage: 2602
+  nuclear: 2887
+  oil: 1593
+  solar: 9724
   unknown: 632
-  wind: 1062
+  wind: 1581
 contributors:
   - corradio
   - MariusKroon
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -17,14 +17,6 @@ LOWER_DATETIME_BOUND = datetime(2000, 1, 1, tzinfo=timezone.utc)
 
 
 class Mix(BaseModel, ABC):
-    def set_value(self, mode: str, value: Optional[float]) -> None:
-        """
-        Sets the value of a production mode.
-        This can be used if the Production has been initialized empty
-        and is being filled in a loop.
-        """
-        self.__setattr__(mode, value)
-
     def add_value(
         self,
         mode: str,
@@ -73,7 +65,7 @@ class ProductionMix(Mix):
         Overriding the constructor to check for negative values and set them to None.
         This method also keeps track of the modes that have been corrected.
         Note: This method does NOT allow to set negative values to zero for self consumption.
-        As we want self consumption to be set to zero, on a fine grained level with the set_value method.
+        As we want self consumption to be set to zero, on a fine grained level with the `add_value` method.
         """
         super().__init__(**data)
         for attr, value in data.items():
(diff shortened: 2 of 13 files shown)


parsers/ESIOS.py
```diff
@@ -55,7 +61,10 @@ def fetch_exchange(
     }
 
     zone_key = ZoneKey("->".join(sorted([zone_key1, zone_key2])))
-    if zone_key not in EXCHANGE_ID_MAP.keys():
+    if (
+        zone_key not in EXCHANGE_ID_MAP.keys()
+        or zone_key not in EXCHANGE_GRANULARITY_FACTOR_MAP.keys()
```
unitrium, line 66:
```suggestion
        or zone_key not in EXCHANGE_GRANULARITY_FACTOR_MAP
```
No need for the `keys` method, you can directly make sure it's here.

parsers/ESIOS.py
```diff
@@ -55,7 +61,10 @@ def fetch_exchange(
     }
 
     zone_key = ZoneKey("->".join(sorted([zone_key1, zone_key2])))
-    if zone_key not in EXCHANGE_ID_MAP.keys():
+    if (
+        zone_key not in EXCHANGE_ID_MAP.keys()
```
unitrium, line 65:
```suggestion
        zone_key not in EXCHANGE_ID_MAP
```

parsers/ESIOS.py
```diff
@@ -24,6 +24,12 @@
     "ES->MA": "10209",
 }
 
+# Map each exchange to the needed factor to adjust from MWh to MW. Depends on the time granularity of the API for each request
+EXCHANGE_GRANULARITY_FACTOR_MAP = {
+    "AD->ES": 1,
+    "ES->MA": 4,
```
unitrium, line 30: Could you just add a comment that it's a 15min interval, so that when we look back at the code we understand this factor.

unitrium (commented):
Thanks, could you just add a bit more comments so that we can better understand the mapping in the future? I will let @VIKTORVAV99 do the final review

parsers/ESIOS.py
```diff
@@ -24,6 +24,13 @@
     "ES->MA": "10209",
 }
 
+# Map each exchange to the needed factor to adjust from MWh to MW. Depends on the time granularity of the API for each request
+# E.g ES->MA is 4 because the API returns 15 minutes intervals data (15 min = 1/4 of an hour; P=E/t).
+EXCHANGE_GRANULARITY_FACTOR_MAP = {
```
VIKTORVAV99, line 29:
```suggestion
EXCHANGE_MULTIPLICATION_FACTOR_MAP = {
```

parsers/ESIOS.py
```diff
@@ -55,7 +62,10 @@ def fetch_exchange(
     }
 
     zone_key = ZoneKey("->".join(sorted([zone_key1, zone_key2])))
-    if zone_key not in EXCHANGE_ID_MAP.keys():
+    if (
+        zone_key not in EXCHANGE_ID_MAP
+        or zone_key not in EXCHANGE_GRANULARITY_FACTOR_MAP
```
VIKTORVAV99, line 67:
```suggestion
        or zone_key not in EXCHANGE_MULTIPLICATION_FACTOR_MAP
```

parsers/ESIOS.py
```diff
@@ -82,6 +92,8 @@ def fetch_exchange(
             -value["value"] if zone_key.partition("->")[0] == "ES" else value["value"]
         )
 
+        net_flow *= EXCHANGE_GRANULARITY_FACTOR_MAP[zone_key]
```
VIKTORVAV99, line 95:
```suggestion
        net_flow *= EXCHANGE_MULTIPLICATION_FACTOR_MAP[zone_key]
```

VIKTORVAV99 (commented):
Overall this looks good, but I would rename the mapping to better represent the value it contains, i.e. the multiplication factor, not the granularity itself.

VIKTORVAV99 (approved):
LGTM! 🎉
Thanks for implementing the fix.

PR #5694, opened 2023-08-03
Remove set_value from the Mix class in favour of add_value

unitrium:
## Issue

`add_value` provides essentially the same functionality as `set_value`. To simplify the code around ProductionBreakdowns, removing the `set_value` method seems like the best way forward.

## Description

Replacing all set_value by add_value and removing set_value code.
Update the merge production code to reflect those changes.
Add a test to make sure we don't return unecessary fields.

### Double check


VIKTORVAV99:
I just took a quick glance now and noticed some of the parsers that used set_value() manually sum values from other modes first (the CO parser specifically).

Do we want to change that? And if we do do we want to do it in the same PR?

electricitymap/contrib/lib/models/events.py
```diff
@@ -177,19 +155,16 @@ def corrected_negative_modes(self) -> AbstractSet[str]:
...
+            ):
                 value = getattr(production_mix, mode)
-                if value is None:
-                    continue
-                if getattr(merged_production_mix, mode) is None:
-                    merged_production_mix.set_value(mode, 0)
-                merged_production_mix.set_value(
-                    mode, getattr(merged_production_mix, mode) + value
-                )
+                merged_production_mix.add_value(mode, value)
```
VIKTORVAV99, line 167: This seems so much simpler. Nice implementation.

electricitymap/contrib/lib/models/events.py
```diff
@@ -220,17 +195,14 @@ def merge(cls, storage_mixes: List["StorageMix"]) -> "StorageMix":
...
                 value = getattr(storage_mix_to_merge, mode)
-                if value is not None:
-                    current_value = getattr(storage_mix, mode)
-                    if current_value is None:
-                        storage_mix.set_value(mode, value)
-                    else:
-                        storage_mix.set_value(mode, current_value + value)
-        return storage_mix
+                merged_storage_mix.add_value(mode, value)
+        return merged_storage_mix
```
VIKTORVAV99, line 205: Unless I am mistaken this is now the same (functionality wise) as the production merge method and both can be moved to the top level Mix class?

(or at least this can, since we are overriding the add_value() in the production mix to make sure it's not negative)

VIKTORVAV99 (commented):
Just one small thing.

electricitymap/contrib/lib/models/events.py
unitrium: I don't think it would because there's a slight difference in the production mixes where we copy the corrected modes set over to the new merged mix.
I would maybe keep as it is in this PR and give it another thought in another PR to keep this one small and focused on one change.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5698, opened 2023-08-04
Update GB emission factors

VIKTORVAV99:
## Issue

Some of the GB emission factors where not using the values they should.

## Description

This changes the affected emissions factors to use the EU28 and European values for the relevant emission factors.

### Double check


tonypls:
For reference [here's the source](https://unece.org/sites/default/files/2022-04/LCA_3_FINAL%20March%202022.pdf)

madsnedergaard (approved):
LGTM :)

PR #5699, opened 2023-08-04
Added aria labels to icon only buttons

mikkelsonnenschein:
## Issue
Issue described here: https://github.com/electricitymaps/electricitymaps-contrib/issues/5695

## Description
- Added a new `ariaLabel` prop to the Map button component that can be passed strings from the localisation files
- Added an aria section in localisation file and a sub section with labels. The idea here is that the aria-section in the localisation file could contain multple aria types like [descriptions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) or [placeholders](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-placeholder) etc.

## Preview

Screenshot 2023-08-04 at 14 20 15. 

Accessibility score: 78 -> 93. 

Screenshot 2023-08-04 at 14 20 56. 

Aria labels applied to 5 right side map buttons + side panel show/hide -button

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -148,6 +148,17 @@
     "dataIsDelayed": "Data is delayed",
     "price": "Price"
   },
+  "aria": {
+    "label": {
+      "selectLanguage": "Select language",
+      "colorBlindMode": "Color blind mode",
+      "windLayer": "Wind layer",
+      "solarLayer": "Solar layer",
+      "changeTheme": "Change theme",
+      "hideSidePanel": "Hide side panel",
+      "showSidePanel": "Show side panel"
+    }
+  },
   "aggregateButtons": {
     "country": "country",
     "zone": "zone"
--- a/web/src/features/map-controls/LanguageSelector.tsx
+++ b/web/src/features/map-controls/LanguageSelector.tsx
@@ -31,6 +31,7 @@ export function LanguageSelector({ isMobile }: { isMobile?: boolean }) {
           <MapButton
             icon={<HiLanguage size={20} style={{ strokeWidth: '0.5' }} />}
             tooltipText={__('tooltips.selectLanguage')}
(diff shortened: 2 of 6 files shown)


VIKTORVAV99:
Nice work!
I don't see anything that needs changing by looking at the code but I'll do some testing before approving and merging as well. On Monday at the latest.

I'd also like to add Swedish translations directly in this PR if you don't mind?

mikkelsonnenschein:
Sounds good! And ofc please do.

web/public/locales/en.json
```diff
@@ -148,6 +148,17 @@
     "dataIsDelayed": "Data is delayed",
     "price": "Price"
   },
+  "aria": {
+    "label": {
+      "selectLanguage": "Select language",
+      "colorBlindMode": "Color blind mode",
+      "windLayer": "Wind layer",
+      "solarLayer": "Solar layer",
```
VIKTORVAV99, line 156:
```suggestion
      "colorBlindMode": "Toggle color blind mode",
      "windLayer": "Toggle wind layer",
      "solarLayer": "Toggle solar layer",
```
Either I think we should add toggle to these translations (like the suggestion) or we should add a Turn on/Turn off or Show/Hide version for these like the side panel.

Let me know what you think and once we have decided I'll add the Swedish translations and merge this.

mikkelsonnenschein: Hi, to give some context the reason why i chose to leave it out was that Voice Over on Mac includes it natively, both the type (toggle) aswell as the state (selected). 

Toggle - Aria label

VIKTORVAV99: I was not aware of that!
Let's leave it as is then!

I'll add the Swedish translations later today.

VIKTORVAV99 (approved):
Looks good to me! 🎉

Thanks for implementing this!

PR #5700, opened 2023-08-05
Fix negative scaling in graphs

VIKTORVAV99:
## Issue

Our graphs currently don't take negative values into account at all when scaling.

## Description

This PR ensures that negative values are taken into account, it should also have a performance improvement in the affected function.

This fixes our price graphs and is needed for #5647.

#### Before:
image
#### After:
image

### Double check

### DIFF
--- a/web/src/features/charts/elements/AreaGraph.tsx
+++ b/web/src/features/charts/elements/AreaGraph.tsx
@@ -23,13 +23,24 @@ const Y_AXIS_WIDTH = 40;
 const Y_AXIS_PADDING = 4;
 
 const getTotalValues = (layers: any) => {
-  const values = layers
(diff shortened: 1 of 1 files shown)


web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -23,13 +23,24 @@ const Y_AXIS_WIDTH = 40;
 const Y_AXIS_PADDING = 4;
 
 const getTotalValues = (layers: any) => {
```
VIKTORVAV99, line 25: We should try and create a interface for layer (or layers) in a separate PR as we use it in several places.

madsnedergaard (approved):
Looks good to me! 🥳

PR #5702, opened 2023-08-06
Fix CY solar and biomass

silimotion:
## Issue
Recently the solar and biomass data from Cyprus stopped being recorded. This fixes it.

Screenshot from 2023-08-06 18-31-24

### Double check


VIKTORVAV99 (approved):
I was not aware we where parsing solar and biomass like this for CY.

But let's get the parser working again before we look into alternative sources that are more clearly defined.

PR #5710, opened 2023-08-07
Fixes breakdown tooltips on Safari

### LINKED ISSUE #5689: Not matching values, error in country name.
Bildschirm­foto 2023-08-01 um 20 50 17
As shown, the country name is not eligible and the indicated imported power is not matching the percent value.
Regards
Michael

madsnedergaard:
## Issue

Closes #5689

## Description

Not 100% sure what's going on, but having the root element contain the positioning style caused the weird bug where previous content "stayed" and overlapped new tooltip content (see video below).

I managed to get things working again in both Safari and Chrome (and on desktop + mobile) by creating a new div that contains the tooltip just below the root portal element.

### Preview

**Before:**

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/5cf434fe-5c6d-4dfc-9c12-d14db82e9a62

**After:**

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/7f99c41d-7b17-483d-8bca-bf0fddcf5957

### Double check


VIKTORVAV99 (approved):
Safari really is the new IE isn't it?

Hopefully this takes care of the issue permanently.

madsnedergaard:
Haha yeah, but I actually just realised that this is also how we do it in the AreaGraphTooltip component :)

VIKTORVAV99:
> Haha yeah, but I actually just realised that this is also how we do it in the AreaGraphTooltip component :)

Perhaps we should try and unify those a bit at some point?
(I have noticed we duplicate a lot between the area graphs (and tooltips) so if you run out of issues on Linear you can always tackle these 😉)

tonypls (approved):
LGTM

PR #5711, opened 2023-08-09
Add wind layer test data and particle scaling based on zoom

### LINKED ISSUE #1157: Wind stream lines density unbalanced with map scale

wrma6:
## Issue

Closes #1157 

## Description

There are two parts of this PR. The first is to add wind layer test data so that the wind layer can be seen locally. This is done by intercepting the calls for the wind files. The code does request different times and throws certain errors when assumptions don't hold, but I found that everything works if we just set the time in the mock data to the requested time from the code. This means we only need one file of mockdata to handle all the calls.

The second part is to add scaling for particle count and velocity based on zoom. I tried a bunch of methods and wanted to keep things simple, so I ended up just using the square root of the "zoom value" reported by the map object accessible to the wind layer. I also adjusted the constants so that particle count and velocity is mostly unchanged when the map is mostly zoomed out (so the changes mostly affect when the user zooms in a lot). See below for comparisons.

### Preview
Proposed screenshots
image

image

Current screenshots
image

image

### Double check


VIKTORVAV99:
So the changes in here look good as far as I can tell but as you may have noticed this particular file (wind.ts) is a bit messy...

So I would appreciate if we could take care of the GitHub Code scanning alerts (just the ones that relate to your changes). So we can start getting this file cleaned up a bit or at least not introduce more messy code. 😅

Does that sound okay? I'll happily help if needed 🙂

wrma6:
Hopefully should be good now? My linting/formatting setup has always been a bit weird so far.

Just noticed all of the ignores at the top of windy.ts, maybe now that I've taken all the time understand what it's doing I could re-write it? If no one's already working on it I guess

VIKTORVAV99:
If you want to take a crack at it feel free to do so. But I suggest doing that in a seperate PR.

These files are very old and haven't really been maintained for some reason thought so it might not be as straight forward as it seems.

I'm on my phone now but I'll double check the latest changes tomorrow and merge it if all is good (which I think it is).

PS: GitHub code scanning don't care about the disabled eslint rules (in the file) but will only fail on new errors. (changed lines also count as new).

VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard:
Thank you so much for jumping into this particular messy part of the codebase and fixing a looooong standing and annoying bug 😍 

To add some context, the Windy code was forked from some deprecated code and tweaked to make it work for our use case, but we've kinda been waiting and hoping something nice third-party would appear, so we could avoid having to maintain this code :)

wrma6:
I always had my eyes on this issue since to me since I felt like it was a very visible bug. Like a lot of other people though the code did turn me away for a while especially since it did also involve figuring out how to enable the layers locally on dev as well 😄.

Now that have figured out how it works I'd also like to help with re-writing the code to clean it all up. Viktor has warned me this might prove to be a difficult challenge and I'll be switching terms at school from study to work so I'm unsure about my time to work on this in the near future. But I'll do my best to get it done eventually 👍

PR #5715, opened 2023-08-10
Add solar mock data and fix solar layer moving issue

### LINKED ISSUE #3217: Solar layer is moving when zooming
## Description

As mentioned by @veqtrus in #3215, it seems like the solar layer is being offset incorrectly when zooming on the map.

### Examples

See the cloud of Queensland in this example when zooming:

https://user-images.githubusercontent.com/3296643/125889923-b578d356-dfce-41d2-b494-a6aa86567e5a.mp4

And when moving around:

https://user-images.githubusercontent.com/3296643/125890099-9a8aed60-740f-4204-9db3-ab07fe701b65.mp4

wrma6:
## Issue

Closes #3217

## Description

I added solar mock data following much of the same process as in https://github.com/electricitymaps/electricitymaps-contrib/pull/5711. Assuming that one gets merged first, I will change this PR and merge the interception of mockserver calls into one.

As for the issue of the solar layer moving, I'm pretty sure it's simply because the code was using a (simple) custom "unproject" function instead of the one that comes with the map (which I believe takes into account the mercator projection). This should be why the latitudes (ie. vertical zooming) were messed up.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard:
Thanks for fixing this too! 💪 
Did you notice any performance changes here, as mentioned in the comment?
```
 // simpler but also slower as unproject method seems more
// complex than just these basic arithmetic operations.
```

wrma6:
Not visibly, but I didn't do any sort of formal performance testing to really see the difference. I think it should be OK however, as the wind layer does currently use the unproject method (I believe more frequently as well).

VIKTORVAV99:
I also did not see any noticeable performance issues when tested locally so it should be fine.

If it presents itself down the line we might have to re-evalue how this is done though.

PR #5719, opened 2023-08-11
Update Belgian wind & solar capacities

Fred-3000:
Same sources as before.

## Issue

None

## Description

Update Belgian wind & solar capacities


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for keeping these capacities updated.

Issue #2883, opened 2020-12-29
More & better data for Bolivia (historical with better breakdown by type + forecast with good breakdown by type)


q--:
While investigating  #2882, I came across some better data:

* Historical generation by power plant up to yesterday.
 * [List of reports](https://www.cndc.bo/estadisticas/index.php), select "Despacho de carga realizado" in the dropdown menu
  * [URL to an actual report](https://www.cndc.bo/php/dload.php?f=dcdr_011220.htm&d=boldia). As you can see, the URL has a date in DDMMYY format that can be altered to get data for a different day. The list of reports only has links to the last month, but by changing the date in this URL you can go back a lot further.
  * Although the the report has no data for the current day (we'll need #2882 for that), it actually does distinguish between wind and solar, partially resolving #2666

TODO type the rest, saving so I don't lose stuff

IV1T3:
The archive above mentioned by @q-- goes actually as far back as February 1st, 2006 with hourly data and split by type: 

https://www.cndc.bo/media/archivos/boletindiario/dcdr_010206.htm

Nil-Cipher:
Is anyone still doing this?

q--:
> Is anyone still doing this?

Not I, at least

PR #5722, opened 2023-08-13
Made function to parse Bolivia website for hourly production data

Nil-Cipher:
## Issue

 #2883

## Description

Add functionality to parse historical production data for Bolivia.

### Double check

### DIFF
--- a/parsers/BO-enhanced.py
+++ b/parsers/BO-enhanced.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+
+# The arrow library is used to handle datetimes consistently with other parsers
+from datetime import datetime
+from logging import Logger, getLogger
+from typing import List, NamedTuple, Optional
+
+import arrow
+
+# BeautifulSoup is used to parse HTML to get information
+from bs4 import BeautifulSoup
+from requests import Session
+
+tz_bo = "America/La_Paz"
+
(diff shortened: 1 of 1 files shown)


q--:
By the way, one of the reasons I opened https://github.com/electricitymaps/electricitymaps-contrib/issues/2883 is IIRC that the current source for Bolivia originally didn't separate wind from solar and lumped biomass with thermal. I'm not sure if there's still some biomass included with thermal (biogas perhaps? Or I'm making it up), but with the more extensive data now also offered by the current source, the main advantage would be the ability to distinguish between the various types of thermal generation. The I

parsers/BO-enhanced.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+
+# The arrow library is used to handle datetimes consistently with other parsers
+from datetime import datetime
+from logging import Logger, getLogger
+from typing import List, NamedTuple, Optional
+
+import arrow
```
VIKTORVAV99, line 8: Just providing some early feedback here but we have decided internally to move away from arrow and only use datetime when possible.

madsnedergaard:
@Nil-Cipher, do you remember what's missing from making this ready for review? We'd like to either get it done and merged, or close the PR for now :)

PR #5729, opened 2023-08-14
Updated capacities for Hungary

RolandSzep:
## Description

Updated capacities for Hungary. The actual data is from this [pdf](https://www.mavir.hu/documents/10258/246425808/BT_2015-20230701_ig_BR+NT_HU_v1.pdf).
I have also updated DATA_SOURCES.md with a link to the subpage where the latest pdf can always be found.
Biomass capacity should still be obtained from ENTSO-E, because in the pdf provided by MAVIR biomass is combined with other sources.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5731, opened 2023-08-14
Day/night indicators on the timeslider ☀️🌙

madsnedergaard:
## Issue

Day-night indicators [were added](https://github.com/electricitymaps/electricitymaps-contrib/pull/2871), but then removed with the historical view feature.

## Description

This PR implements an indication of day/night on the TimeSlider, to make it easier to understand local time for zones when exploring.

There's a lot of code here, but happy to talk anyone through it :)

### Preview

See preview: TBD


web/src/components/TimeSlider.tsx
```diff
@@ -17,18 +85,61 @@ function TimeSlider({ onChange, numberOfEntries, selectedIndex }: TimeSliderProp
...
+    <TimeSliderBasic {...props} trackBackground={trackBackground} thumbIcon={thumbIcon} />
+  );
+}
+
+function TimeSlider(props: TimeSliderProps) {
+  const zoneId = getZoneFromPath();
+  const [timeAverage] = useAtom(timeAverageAtom);
+
+  const showNightTime =
+    zoneId && timeAverage === TimeAverages.HOURLY && props.numberOfEntries === 24;
```
VIKTORVAV99, line 136: I believe this should be 25?
As we show from 09:00 to 09:00.

Is the props length even needed though?

web/src/hooks/nightTimes.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  // get latitude and longitude of zone
+  const feature = geometries.features.find(
+    (feature) => feature.properties.zoneId === zoneId
+  );
+  const coord = feature?.properties.center;
+  if (!coord) {
+    console.warn('no center coordinates available');
+    return undefined;
+  }
```
VIKTORVAV99, line 33: In what scenario can this happen? All zones should have centers calculated and they should always be available when the feature is available.

web/src/components/TimeSlider.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+  'linear-gradient(90deg,rgb(243 244 246) 0%, rgb(209 213 219 ) 0%, rgb(209 213 219 ) 12.5%, rgb(243 244 246) 12.5%,rgb(243 244 246) 62.5%, rgb(209 213 219 ) 62.5%, rgb(209 213 219 ) 100%, rgb(243 244 246) 100%)';
+
+const meta: Meta<typeof TimeSliderBasic> = {
+  title: 'basics/TimeSlider',
+  component: TimeSliderBasic,
+};
+
+export default meta;
+
+function SliderWithControls(arguments_: TimeSliderProps & { nightTimeSets?: any }) {
```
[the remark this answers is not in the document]
VIKTORVAV99: I'm with Eslint on this one, let's not add more any types. (It's hard enough to remove the ones we have.)

VIKTORVAV99 (commented):
Some initial comments, will take a another look and test it tomorrow.

VIKTORVAV99:
Looks like the tests are failing cause the values are offset by +2 (possibly due to timezones?).

web/src/components/TimeSlider.tsx
madsnedergaard: Apparently not, logging `props.numberOfEntries` returns 24 when in hourly :)

And yes, if we remove it there's an annoying flickering going when switching between time aggregates due to the way things are rendered:

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/b8185625-a245-4747-ab20-0eb8e7e75288

VIKTORVAV99: Oh that was a nasty bug. Let's keep it in there!

web/src/hooks/nightTimes.ts
madsnedergaard: It's only due to using the `.find()` lookup which can return `undefined` if element is not found. Happy to redo that part if you have a better solution in mind :)

web/src/components/TimeSlider.stories.tsx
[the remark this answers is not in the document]
madsnedergaard: You guys are teaming up, huh? 😅
But fair enough, fixed now!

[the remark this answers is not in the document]
VIKTORVAV99: I spent way to long removing a any type yesterday because as soon as I fixed the error in one place it just cascaded to another place...

So yes!

web/src/hooks/nightTimes.ts
VIKTORVAV99: Hmm, yeah that is a problem. All the solutions that I can think of involve bigger changes.

Either we shift to using the bounding box from the zone config (which we can access directly) and use that to get the center instead.
Or we change how the config files in general are computed in a future version where we can use the center calculated by topojson but insert it into the config instead (at build).

But that seems like a lot of work for just this... (could help us optimize the loading though by shrinking the world file and moving it over to the config instead))

web/public/images/slider-thumb-night.svg
```diff
@@ -0,0 +1 @@
+<svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 24 24" aria-hidden="true" height="14px" width="14px" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" d="M9.528 1.718a.75.75 0 01.162.819A8.97 8.97 0 009 6a9 9 0 009 9 8.97 8.97 0 003.463-.69.75.75 0 01.981.98 10.503 10.503 0 01-9.694 6.46c-5.799 0-10.5-4.701-10.5-10.5 0-4.368 2.667-8.112 6.46-9.694a.75.75 0 01.818.162z" clip-rule="evenodd"></path></svg>
```
VIKTORVAV99, line 1:
```suggestion
<svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 24 24" aria-hidden="true" height="1em" width="1em" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" d="M9.528 1.718a.75.75 0 01.162.819A8.97 8.97 0 009 6a9 9 0 009 9 8.97 8.97 0 003.463-.69.75.75 0 01.981.98 10.503 10.503 0 01-9.694 6.46c-5.799 0-10.5-4.701-10.5-10.5 0-4.368 2.667-8.112 6.46-9.694a.75.75 0 01.818.162z" clip-rule="evenodd"></path></svg>
```
I'm not sure these even need a high or width but let's use the same one for both for consistency.

madsnedergaard: It's actually on purpose, to make the icons look more like they are the same size 😄 

Frame 1 (7)

web/src/components/TimeSlider.stories.tsx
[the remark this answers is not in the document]
madsnedergaard: Haha yeah, in general I agree (while in Stories I don't really care, but might as well just type it up :))

VIKTORVAV99:
Also if you press a gray zone (with no data) the whole app freezes...

And less critical if there is a datapoint missing (at least in the end) the new icons disappear but it still works.

madsnedergaard:
> I noticed that when I switch between countries it goes back to the <> arrow real quick before it updates. Would it be possible to avoid that without adding too much code? (or use something else like a dot instead so it is less noticeable)

Yeah, it's a bit annoying - but I can't really see a good way of fixing it. It happens due to this, as we don't have any data for the newly selected zone until it's done fetching:

Screenshot 2023-08-15 at 22 26 30

madsnedergaard:
Well actually, we can use an effect to keep the old one around until the next one is available. It works like this, what do you think?

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/a5001868-5cc5-4aaa-a8b0-3efb78082472

VIKTORVAV99:
> Well actually, we can use an effect to keep the old one around until the next one is available. It works like this, what do you think?
> 
>  Kapture.2023-08-15.at.22.34.14.mp4

I think that works well!

Most of the time the icon should be the same for neighboring zones at least.

madsnedergaard:
> Also if you press a gray zone (with no data) the whole app freezes...

Whoops, good catch - fixed 😅 

> Did you use a specific zone for testing this?

And less critical if there is a datapoint missing (at least in the end) the new icons disappear but it still works.

VIKTORVAV99:
> > Also if you press a gray zone (with no data) the whole app freezes...
> 
> Whoops, good catch - fixed 😅
> 
> > Did you use a specific zone for testing this?
> 
> And less critical if there is a datapoint missing (at least in the end) the new icons disappear but it still works.

I used: https://app.electricitymaps.com/zone/CR

It has some values missing now but hopefully that will resolve it self.

web/public/images/slider-thumb-night.svg
madsnedergaard: Haha, me too - which is why we're using the one that is 14px instead of 1em (16px) 😂

VIKTORVAV99: Oh!

web/src/components/TimeSlider.stories.tsx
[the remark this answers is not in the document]
VIKTORVAV99: It makes the numbers in the security tab go down (or at least not up) 😅

madsnedergaard:
> > > Also if you press a gray zone (with no data) the whole app freezes...
> > 
> > 
> > Whoops, good catch - fixed 😅
> > > Did you use a specific zone for testing this?
> > 
> > 
> > And less critical if there is a datapoint missing (at least in the end) the new icons disappear but it still works.
> 
> I used: [app.electricitymaps.com/zone/CR](https://app.electricitymaps.com/zone/CR)
> 
> It has some values missing now but hopefully that will resolve it self.

Hmm, I made it work for Costa Rica, but if there are more missing hours it goes completely off 🤔 

Honduras right now:
Screenshot 2023-08-15 at 23 01 11

vs how it should be:
Screenshot 2023-08-15 at 23 01 16

The problem is that I am using a `closestIndexTo(nightStart, datetimes)` util which is really helpful, but unfortunately also means things go crazy when there's only e.g. 5 entries sent. One solution could be to just ignore the datetimes completely and use the timestamp from `/state` and go 25 hours back on that, but I don't know if that would have other consequences 🤔

VIKTORVAV99:
> > > > Also if you press a gray zone (with no data) the whole app freezes...
> > > 
> > > 
> > > Whoops, good catch - fixed 😅
> > > > Did you use a specific zone for testing this?
> > > 
> > > 
> > > And less critical if there is a datapoint missing (at least in the end) the new icons disappear but it still works.
> > 
> > 
> > I used: [app.electricitymaps.com/zone/CR](https://app.electricitymaps.com/zone/CR)
> > It has some values missing now but hopefully that will resolve it self.
> 
> Hmm, I made it work for Costa Rica, but if there are more missing hours it goes completely off 🤔
> 
> Honduras right now: Screenshot 2023-08-15 at 23 01 11
> 
> vs how it should be: Screenshot 2023-08-15 at 23 01 16
> 
> The problem is that I am using a `closestIndexTo(nightStart, datetimes)` util which is really helpful, but unfortunately also means things go crazy when there's only e.g. 5 entries sent. One solution could be to just ignore the datetimes completely and use the timestamp from `/state` and go 25 hours back on that, but I don't know if that would have other consequences 🤔

Hmm, another option would be to fill all the datetimes (even if they are empty) so it's always 25. That could in theory fix some other problems with the graphs depending on how it's done.

Or we just enable estimations for these 😆 

But perhaps we can skip that for now? I would like to get the PR merged before it becomes too large.

madsnedergaard:
Yes, let's get this merged and then we can revisit the missing hours later :)

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5732, opened 2023-08-15
TH `Unknown` emission factor 1998-2022

PPsyrius:
## Description

Updating Thailand's `Unknown` emission factor to use [Energy Policy and Planning Office (EPPO)](https://www.eppo.go.th/index.php/en/en-energystatistics/co2-statistic)'s figure for 1998-2022, with [Emission Assumptions](https://www.eppo.go.th/images/Energy-Statistics/energyinformation/Energy_Statistics/Emission/EmissionAssumptions_TH.pdf) for CO<sub>2</sub>, N<sub>2</sub>O, and CH<sub>4</sub> based on IPCC 2006 while NO<sub>x</sub> and CO are from earlier 1996 publishing.

Here's the [backup Powerpoint file](https://github.com/electricitymaps/electricitymaps-contrib/files/12341077/T09_01_15.pptx) should their website become unavailable somehow.

### Preview

image

### Double check


config/zones/TH.yaml
```diff
@@ -19,18 +19,108 @@ capacity:
 contributors:
   - PPsyrius
 emissionFactors:
-  direct:
-    unknown:
-      _url: https://docs.google.com/spreadsheets/d/16j5BVc4L34BDqVE0GeDBv0P3zQupT6zU77Kj5UK7GjE/edit?usp=sharing
-      source: IEA 2020; assumes 9.1% biomass, 19.3% coal, 64.7% gas, 2.5% hydro, 2.6%
-        solar, 1.7% wind, 0.1% other
-      value: 386.645
```
VIKTORVAV99, line 27: I don't think these should necessarily be removed. Is there any  particular reason for it?

config/zones/TH.yaml
```diff
@@ -19,18 +19,108 @@ capacity:
...
   lifecycle:
     unknown:
-      _url: https://docs.google.com/spreadsheets/d/16j5BVc4L34BDqVE0GeDBv0P3zQupT6zU77Kj5UK7GjE/edit?usp=sharing
-      source: IEA 2020; assumes 9.1% biomass, 19.3% coal, 64.7% gas, 2.5% hydro, 2.6%
-        solar, 1.7% wind, 0.1% other
-      value: 509.277
+      - _comment: Gross Energy Generation of EGAT and Net Energy Generation of IPP, SPP, and VSPP
+        datetime: '1998-01-01'
+        source: EPPO 1998
+        value: 636
```
VIKTORVAV99, line 27: My only worry is that the emission factors used in the calculations are may be different than those we use as they are older (from 2009 vs 2013 for the global ones)

config/zones/TH.yaml
```diff
@@ -19,18 +19,108 @@ capacity:
...
+        source: EPPO 2020
+        value: 442
+      - _comment: Gross Energy Generation of EGAT and Net Energy Generation of IPP, SPP, and VSPP
+        datetime: '2021-01-01'
+        source: EPPO 2021
+        value: 433
+      - _comment: Gross Energy Generation of EGAT and Net Energy Generation of IPP, SPP, and VSPP
+        datetime: '2022-01-01'
+        source: EPPO 2022
+        value: 409
```
VIKTORVAV99, line 123: I would just add the source as EPPO or EPPO 2023 since it was released this year and then add a top level sources object like this for example:
```yaml
sources:
  EU-ETS, ENTSO-E 2021:
    link: https://drive.google.com/file/d/15UeUHyhcjVw8fGLc0Zx_HELSx20cw_-e/view?usp=share_link
```

VIKTORVAV99 (commented):
Overall this change looks good for the most part. I just have some questions about the emission factors used in the data source that could potentially skew things a little.

config/zones/TH.yaml
PPsyrius: I assume it should be removed as it was from a different calculation, but yeah - will put them back in.

config/zones/TH.yaml
PPsyrius: I've found [CEIC's data](https://www.ceicdata.com/en/thailand/carbon-dioxide-emissions-statistics/carbon-dioxide-emission-per-electricity-generation) to be quite close to EPPO's figure, but they're likely using different calculation methods even if they shared the same source i.e. CEIC's `407 gCO₂eq/kWh` vs EPPO's own `409 gCO₂eq/kWh` for 2022.

While we can try backtracking how EPPO reached the figure above ourselves, [their source simply group all renewable sources except hydro as "renewables" and dump all of Laos's Hydro and Coal as "import"](https://view.officeapps.live.com/op/view.aspx?src=https%3A%2F%2Fwww.eppo.go.th%2Fepposite%2Fimages%2FEnergy-Statistics%2Fenergyinformation%2FEnergy_Statistics%2FElectricity%2FT05_02_04-2.xls&wdOrigin=BROWSELINK), making it a bit daunting to replicate.

[Here's my own calculation](https://docs.google.com/spreadsheets/d/1s6d6qQnE__-pque1urlXY5iseRVjhiBZr-pATPwInPw/edit?usp=sharing) using the contract obligation capacity instead of total capacity for IPP/SPP/VSPP from ERCSPP database, arriving at `406.267 gCO₂eq/kWh`. Alternatively, another backtracking attempt tries to breakdown "renewables" and "import" columns from EPPO's 2022 total electricity generation by source arriving at a whopping `447.488 gCO₂eq/kWh` instead. (see Sheet 2)

VIKTORVAV99: If the differences are that small there should be no issues. We will always be off by some degree either way due to the fact it is all unknown.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for addressing the feedback.

PR #5734, opened 2023-08-15
Fix zone switching from URL

tonypls:
## Issue
The zone / country switching was not working when going from a country URL to a zone URL

## Description
This PR updates the logic in our zone details useEffect to only use isSubZone which is taken from the params, the issue was that the spatialAggregateAtom was not updated on first render so the switch would not happen when navigating from a URL. If this atom was in the useEffect dependency array it would fix the URL issue but conflicts with the way we handle switching in the toggle and you would cause a double toggle every time it was clicked.

I've also added the geometries as a dependency to the map zone switching / fly to zone useEffect, this means the map now uses our ```getApproximateFeature``` function and will show an approximate zone when the selected zone is not present on the map.
 
### Things to test
URL zone switching:
- zone -> country
- country -> zone
Ranking panel zone switching:
- zone -> country
- country -> zone
Map zone switching; toggle the spatial aggregate toggle from various states

### Double check

### DIFF
--- a/web/src/features/panels/zone/ZoneDetails.tsx
+++ b/web/src/features/panels/zone/ZoneDetails.tsx
@@ -31,7 +31,7 @@ export default function ZoneDetails(): JSX.Element {
   const hasSubZones = getHasSubZones(zoneId);
   const isSubZone = zoneId ? zoneId.includes('-') : true;
   const { data, isError, isLoading } = useGetZone();
-
+  console.log('isSubZone:', isSubZone);
   // TODO: App-backend should not return an empty array as "data" if the zone does not
   // exist.
   if (Array.isArray(data)) {
@@ -44,10 +44,10 @@ export default function ZoneDetails(): JSX.Element {
     }
     // When first hitting the map (or opening a zone from the ranking panel),
     // set the correct matching view mode (zone or country).
-    if (!isSubZone) {
+    if (hasSubZones && !isSubZone) {
       setViewMode(SpatialAggregate.COUNTRY);
     }
-    if (isSubZone) {
+    if (!hasSubZones && isSubZone) {
       setViewMode(SpatialAggregate.ZONE);
     }
   }, []);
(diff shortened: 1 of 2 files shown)


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -45,10 +44,10 @@ export default function ZoneDetails(): JSX.Element {
     }
     // When first hitting the map (or opening a zone from the ranking panel),
     // set the correct matching view mode (zone or country).
-    if (hasSubZones && isZoneView) {
+    if (!isSubZone) {
```
VIKTORVAV99, line 47: I don't think this will work. Now it will always switch to country mode when clicking on a zone that don't have any subzones.

For example if you are in "zone mode" and click on Germany it will switch to "country mode".

We want to preserve the mode if the zone don't have any subzones at all and only switch if there are "two levels" to a zone.

VIKTORVAV99 (changes requested):

web/src/features/panels/zone/ZoneDetails.tsx
tonypls: Okay, I've reintroduced the hasSubZone condition. Do you think this will cover all cases?

VIKTORVAV99: Seems to work great!

If you remove that leftover console.log() we can merge this.

VIKTORVAV99 (approved):
LGTM! 🎉

tonypls:
I assume this previously worked btw and is potentially something that changed when we upgraded Jotai. Adding some routing tests for these cases could be useful

VIKTORVAV99:
> I assume this previously worked btw and is potentially something that changed when we upgraded Jotai. Adding some routing tests for these cases could be useful

Agreed. I would create a issue on linear and merge this first so it is in place for the data platform links as soon as possible though.

PR #5735, opened 2023-08-15
updated capacities for JP-KY

wobniarin:
## Issue
When implementing the estimation method for JP-KY, we have found that the capacities are outdated according to the last integrated report and the data we fetch from the parser. 

## Description
We update solar and wind capacities to match the latest Integrated report from the TSO Kyuden (page 33) [Link](https://www.kyuden.co.jp/library/pdf/en/ir/integratedreport/2022/en_integratedreport_2022.pdf)

### Preview
N/A

### DIFF
--- a/config/zones/JP-KY.yaml
+++ b/config/zones/JP-KY.yaml
@@ -4,12 +4,13 @@ bounding_box:
   - - 131.89
     - 35.11
 capacity:
-  geothermal: 210.2
+  geothermal: 553
   hydro: 1247
   hydro storage: 2350
   nuclear: 5258
   unknown: 9976.94
-  wind: 3.25
+  wind: 630
+  solar: 10910
 contributors:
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

config/zones/JP-KY.yaml
```diff
@@ -4,12 +4,12 @@ bounding_box:
...
 capacity:
-  geothermal: 210.2
+  geothermal: 553
   hydro: 1247
   hydro storage: 2350
   nuclear: 5258
-  unknown: 9976.94
-  wind: 3.25
+  wind: 630
+  solar: 10910
```
pierresegonne, line 12: That's a lot! Go Japannnnn

pierresegonne (approved):

PR #5737, opened 2023-08-15
Remove unused `countryName` in world files

VIKTORVAV99:
## Issue
The world.json file is very big.

## Description
Removes all mentions of the unused `countryName` (in relation to geographies) and shaves of 10kb from the file size.

### Double check


madsnedergaard:
Hmm, I'm wondering if we should exclude the field during the `generateWorld` script instead of removing it completely form the `world.geojson` file? Might be nice to keep it there for searching and looking up the name of a zone :)

VIKTORVAV99:
> Hmm, I'm wondering if we should exclude the field from the generated `world.json` instead of removing it completely form the `world.geojson` file? Might be nice to keep it there for searching and looking up the name of a zone :)

I did consider that, but it adds some overhead to the build process and my thinking was since the zone key and country key are still in there it's not that critical.

madsnedergaard:
> > Hmm, I'm wondering if we should exclude the field from the generated `world.json` instead of removing it completely form the `world.geojson` file? Might be nice to keep it there for searching and looking up the name of a zone :)
> 
> I did consider that, but it adds some overhead to the build process and my thinking was since the zone key and country key are still in there it's not that critical.

Hmm, not sure I quite agree - we're already looping through each feature when calculating the center, so it would be very simple to add a `delete geo.properties.countryName` (or make it all immutable) here:

https://github.com/electricitymaps/electricitymaps-contrib/blob/be179e3fc4b5cd96e3f4401234a515939c336e76/web/geo/generateTopojson.ts#L66-L71

I personally find it useful to keep the countryName to avoid having to look up zone names, but maybe I'm also not the average contributor here 😅

VIKTORVAV99:
But how often do you look at the names in the geo files? (I've never done it) All the names are in the translation files as well.

But I can add it back to the geo files and remove it in the loop but it's resource intensive and would increase the build time even if it's marginal.

VIKTORVAV99:
> I think you're overthinking it here, I just timed the difference with a very sloppy implementation:
> 
> ```
> const { zoneName, ...props } = geo.properties;
> newObjects[zoneName] = {...geo, properties: props};
> ```
> 
> and it's really non-existing, maybe up to 10ms at best - I don't think that's hurting the developer experience 😄
> 
> With that said it's also not something I feel strongly about, so I'm fine with merging it if you think so :)

Hehe perhaps, I'll change it to just remove in the build step.

PR #5740, opened 2023-08-16
Upgrade production and exchanges for CA-NB to new structure

unitrium:
## Issue

Part of the data structure upgrade

## Description

Improve code readbility and upgrade to new datastructure.

### Double check

### DIFF
--- a/parsers/CA_NB.py
+++ b/parsers/CA_NB.py
@@ -1,17 +1,31 @@
 #!/usr/bin/env python3
 
 # The arrow library is used to handle datetimes consistently with other parsers
-from datetime import datetime
+from datetime import datetime, timezone
 from logging import Logger, getLogger
-from typing import Optional
-
-import arrow
+from typing import Any, Dict, List, Optional
 
 # BeautifulSoup is used to parse HTML to get information
 from bs4 import BeautifulSoup
 from requests import Session
 
-timezone = "Canada/Atlantic"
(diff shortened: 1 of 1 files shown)


parsers/CA_NB.py
```diff
@@ -74,25 +88,29 @@ def fetch_production(
...
+    production.append(
+        zoneKey=zone_key,
+        # Using the current utc time because the page returns the current time.
+        datetime=datetime.now(tz=timezone.utc).replace(
+            minute=0, second=0, microsecond=0
+        ),
+        source=SOURCE,
+        production=ProductionMix(
+            unknown=generated,
+        ),
```
VIKTORVAV99, line 101: not part of this PR but would it make sense to switch these to use totalgeneration instead at some point instead of adding it as unknown to the production mix? (Would probably require other changes.)

VIKTORVAV99 (approved):

parsers/CA_NB.py
unitrium: Yeah I was also thinking that. The problem is that currently the load data is not taken into account and therefore not shown on the map. But I'm thinking of creating an estimation model that would take the load and put it into unknown, seems like the best thing to do for me.

VIKTORVAV99: Yeah, that way we don't pollute the database either.

PR #5742, opened 2023-08-16
Fix area graph color for storage

VIKTORVAV99:
## Issue

The color in the area graphs for hydro storage and battery storage turned black due to a space in the name.

## Description
So we did some checking further down the line that I was not aware of. This PR fixes that check to work with the new structure.

#### Before:
image
#### After:
image

### Double check

### DIFF
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -4,6 +4,8 @@ import { area, curveStepAfter } from 'd3-shape';
 import React from 'react';
 import { detectHoveredDatapointIndex, getNextDatetime, noop } from '../graphUtils';
 import { AreaGraphElement } from '../types';
+import { modeColor } from 'utils/constants';
+import { ElectricityModeType } from 'types';
 
 interface AreaGraphLayersProps {
   layers: any[];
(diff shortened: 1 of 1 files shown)


web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -66,7 +68,7 @@ function AreaGraphLayers({
   return (
     <g>
       {layers.map((layer, ind) => {
-        const isGradient = typeof layer.fill === 'function';
```
VIKTORVAV99, line 69: This did no longer work since we are always returning a function.

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -103,7 +105,7 @@ function AreaGraphLayers({
               className={layers.length > 1 ? 'sm:hover:opacity-75' : ''}
               style={{ cursor: 'pointer' }}
               stroke={layer.stroke}
-              fill={isGradient ? `url(#${gradientId})` : layer.fill}
+              fill={isGradient ? `url(#${gradientId})` : layer.fill(layer.key)}
```
VIKTORVAV99, line 108: Which in turn broke this as the URL did not support gradientId with a space in it.

URL encoding also works to fix this issue but it creates gradients for the static colors and inflating the dom. Something we want to avoid.

madsnedergaard (approved):
Good catch, thanks for fixing it!

PR #5744, opened 2023-08-16
Make IN consumption parser use a ThreadPoolExecutor to run faster

pierresegonne:
## Issue

https://storage.googleapis.com/electricitymap-parser-logs/IN-NO.html shows that the consumption parser times out quite frequently

## Description

Use multithreading to parallelise the different http requests required to compute total consumption

### Preview

Screenshot 2023-08-16 at 16 41 00

It took 500+ seconds before

### Double check


unitrium (commented):
Hum that would be very nice, I'm just afraid of being flagged and banned if we starting having multiple threads.

pierresegonne:
> Hum that would be very nice, I'm just afraid of being flagged and banned if we starting having multiple threads.

Why would that happen? We're still making the same number of requests

mathilde-daugy (approved):

mathilde-daugy:
I don't know what to make of @unitrium's comment but to me it looks good :)

pierresegonne:
Screenshot 2023-08-17 at 10 43 56

Took the precaution to try running it within our feeder and it works fine (was wondering how doing multithreading within a multiprocessing setup would interface)

pierresegonne:
@FelixDQ  and @VIKTORVAV99  I tagged you, in case you have previous context on doing something similar on other parsers. 
I'll move ahead shortly with the change if there are no further objections

VIKTORVAV99:
> @FelixDQ and @VIKTORVAV99 I tagged you, in case you have previous context on doing something similar on other parsers. I'll move ahead shortly with the change if there are no further objections

I don't think we have used multithreading in the parsers before but it makes sense to do so when they require multiple endpoints and there is probably a few parsers that could benefit from it. I'm just curious though. How many cores does the parser actually have access to when running in the production environment?

pierresegonne:
> How many cores does the parser actually have access to when running in the production environment?

I'm not sure, the feeder can scale automatically to multiple instances, but I don't know how many cores are provisioned on each

pierresegonne:
Let's try and see :)

PR #5745, opened 2023-08-16
Upgrade CA-QC to new parser datastructure

unitrium:
## Issue

Part of the datastructure upgrade.

## Description

Make consumption and production data use the new parser datastructure for CA-QC.
Adding snapshot tests to easily test the output of those commands.
Update the `to_dict` method so that `EventType` returns a plain string which is easier for testing.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5746, opened 2023-08-16
Fix cy capacity parser

### LINKED ISSUE #5743: Broken CY capacity parser
## Description
Cyprus capacity is broken, caused by a change in the website. This change happened before the one that needed my previous PR #5702.
I can fix this, but I need some guidance as to what does the backend expect. Should it return the capacity when asked for any date, or only for the current time? Also, why is the capacity and production data returned together, instead of separated? In this case being returned in different calls would be more reasonable, in my opinion. Does the backend support this?

silimotion:
## Issue

Closes #5743 

## Description

This solves the issue by updating the name of the html table that should be parsed
I also added some error handling to only run the code when using the realtime source

### Double check


VIKTORVAV99 (changes requested):
Does this work locally for you?
Capacity seems to be an empty object no matter if it's run with a target datetime or not.

VIKTORVAV99:
🤔 It should not be doing that.

VIKTORVAV99:
The test parser returns None as expected.
So it's something in the parser itself that is changing it. I suspect it is how it's set up with the class based structure.
If I log the target_datetime before any of the class stuff I get this:
```
❯ poetry run test_parser CY
Target datetime: None
```

silimotion:
Oh, I see it now. It seems it sets the current time if there is no date in line 122. I can edit it later. As there is always a target_datetime some code was never run.

silimotion:
It should work now. I missed the line where the target_datetime was set because I though I was editing the entrypoint because both functions have the same "fetch_production" name. Can we rename the function inside the Class CyprusParser to something like parse_production? This would stop the name collision. Or just altogether refactor the module to remove the CypusParser class? I just don't see the point in it.

parsers/CY.py
```diff
@@ -115,10 +119,8 @@ def fetch_production(
     """Requests the last known production mix (in MW) of a given country."""
     assert zone_key == "CY"
 
-    target_datetime = target_datetime or datetime.utcnow()
-
     parser = CyprusParser(session or Session(), logger)
-    if isinstance(target_datetime, datetime):
+    if isinstance(target_datetime, Optional[datetime]):
         return parser.fetch_production(target_datetime)
```
VIKTORVAV99, line 124: This won't work. The isinstance method can not be used with a optional type (at least not in python 3.8). But I don't think you need it at all with the other changes you have made.
```suggestion
    return parser.fetch_production(target_datetime)
```

VIKTORVAV99 (changes requested):

VIKTORVAV99 (approved):
Thanks for updating the parser!

PR #5747, opened 2023-08-17
Archive oahu parser

### LINKED ISSUE #5660: [Data Issue]: Wrong data for USA – Oahu
### When did this happen?

2023-07-31

### What zones are affected?

USA-Oahu

### What is the problem?

Hawaii's last coal plant closed on September 1, 2022, (properly marked as having 0 installed capacity) yet the data for Oahu still shows about 16% of its electricity (124 MW) coming from coal. This data is marked as being an estimation, but that's very incorrect, and no data would be better than such outright wrong and outdated data. The amount of solar installed capacity is also far too low, 241 MW vs. more than 1000 MW in reality so the generation should also be much higher. Please update Oahu or just remove it altogether. 
image

nickwohnhas:
## Issue
The data source for Oahu no longer displays any information: https://www.islandpulse.org/.  

## Description
Per the discussion [here](https://github.com/electricitymaps/electricitymaps-contrib/issues/5660) it may be preferable to archive the data parser altogether because the map is currently showing incorrect data for Oahu, HI.  I made an attempt to remove the relevant code, but I'm very new to this project so if I missed something or this isn't needed please let me know!  I did see that the mockserver has some JSON data for Oahu and I can remove that if necessary.  

### Preview

Screen Shot 2023-08-16 at 10 06 55 PM

### Double check


parsers/US_HI.py
VIKTORVAV99, line 1: It would be better to move this to `parsers/archived` instead of fully removing it.

VIKTORVAV99 (changes requested):
Thanks for opening a PR!
There are a few things that will have to change in here as it's currently removing a bit too much.

It also won't close the issue as Oahu is also estimated. But I'll open a PR for that in the backend.

parsers/US_HI.py
nickwohnhas: Hey @VIKTORVAV99 , thanks so much for the review! I went ahead and restored all of the deleted files, and moved the `US_HI.py` parser into `parsers/archived`.

VIKTORVAV99: Thanks, but in you restored a tiny bit too much. We still need to remove the parsers object from the config file otherwise the backend will try and call the parser.

VIKTORVAV99:
Just to provide an update, the backend PR is still open but I expect we can merge that soon. As soon that is merged I will merge this.

nickwohnhas:
> Just to provide an update, the backend PR is still open but I expect we can merge that soon. As soon that is merged I will merge this. (Assuming the parser objects have been removed from the zone config).

Awesome!  I'm working on removing the parser from the zone config now.  Sorry for the delay

parsers/US_HI.py
nickwohnhas: I've removed the parser object from the config file.  If I've misunderstood please let me know, but I think that was the last outstanding item on this PR.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5749, opened 2023-08-17
update capacities for JP-CG

wobniarin:
## Issue

The capacity values we have for JP-CG are outdated. 

## Description

We update them with more recent values. 

### Preview

N/A

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5750, opened 2023-08-18
[Hackday] - Download CSV button in App

tonypls:
## Issue
**Hackday Project**

Download annual CSV data in the app

## Description
This PR adds a download modal to the zone details panel. 

This allows users to select and download hourly, daily, monthly or yearly CSVs for 2021 and 2022. 

### Double check

### DIFF
--- a/web/src/components/Modal.tsx
+++ b/web/src/components/Modal.tsx
@@ -27,7 +27,7 @@ export default function Modal({ isOpen, setIsOpen, title, children }: ModalProps
           >
             <HiXMark size="18" />
           </Dialog.Close>
-          <div className="px-2 py-3 sm:p-[25px_55px]">{children}</div>
+          <div className="px-2 py-3 sm:p-[25px_32px]">{children}</div>
         </Dialog.Content>
       </Dialog.Portal>
     </Dialog.Root>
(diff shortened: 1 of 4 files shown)


VIKTORVAV99:
I know this is draft but I got a little exited so went to check it out.

Generally I think it works well and makes sense IMO but there are some polish and final touches needed, but the most important one seems to be if a user tries to download data for a zone we don't have in the data portal, for example: [data.electricitymaps.com/RE_2021_yearly.csv](https://data.electricitymaps.com/RE_2021_yearly.csv), then things more or less crash (Not really but users ger re-directed to the error page).

So we need to figure out how to keep them in sync and only show the download link for the data portal files (or add all zones to the data portal).

tonypls:
> I know this is draft but I got a little excited so went to check it out.
> 
> Generally I think it works well and makes sense IMO but there are some polish and final touches needed, but the most important one seems to be if a user tries to download data for a zone we don't have in the data portal, for example: [data.electricitymaps.com/RE_2021_yearly.csv](https://data.electricitymaps.com/RE_2021_yearly.csv), then things more or less crash (Not really but users ger re-directed to the error page).
> 
> So we need to figure out how to keep them in sync and only show the download link for the data portal files (or add all zones to the data portal).

Yeah, I think it would make sense to do this on the app-backend as the logic is already in python. Also I think we'd need some product input to discuss if we to collect emails before downloading etc. Will keep the discussion moving when we're next priortizing the roadmap.

PR #5760, opened 2023-08-20
Add missing German translations

sebastianlay:
## Issue

Some strings are missing in the German translation. These were just the ones that jumped out at me as a first time visitor to the website.

There is an additional text that does not seem to support translating in https://github.com/electricitymaps/electricitymaps-contrib/blob/a58470c9830bda73df5ecbb6a30be42cd0ed205a/web/src/features/charts/BreakdownChart.tsx#L42
I might open another PR to change that.

## Description

This change should fix the obvious missing translations.

### Double check

- ~[ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"`~ (not applicable)
- ~[ ] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.~ (not applicable)


VIKTORVAV99 (approved):
Thanks for adding these!

VIKTORVAV99:
>There is an additional text that does not seem to support translating in https://github.com/electricitymaps/electricitymaps-contrib/blob/a58470c9830bda73df5ecbb6a30be42cd0ed205a/web/src/features/charts/BreakdownChart.tsx#L42
I might open another PR to change that.

The goal is that this wont be needed at all eventually but if you want to open a PR for it in the meantime feel free to do so.

PR #5766, opened 2023-08-21
Fix SG parser

madsnedergaard:
## Issue

#5757 

## Description

The image was moved to a new URL. 

Tried testing locally, but it's stalling on tesseract - but at least it's not throwing the error like it was before I updated the url :)

### Double check


tonypls (approved):
LGTM

image

Issue #5763, opened 2023-08-21
Update hungary capacities


FelixDQ:
## Description

From the feedback form:
> According to MAVIR (Hungarian Electricity Industry Transmission System Controller) data, Hungary has the following capacities on 01.08.2023:  Nuclear: 2026,6 MW Coal, Lignite, Fired Together: 926,3 MW Hydrocarbon Gas: 3219,5 MW Hydrocarbon Oil,Diesel: 424,8 MW Other conventional (Battery): 21,3 MW Solar (industrial-sized (over 50 kW) PV): 3122,1 MW Wind: 323,3 MW Biomass, Biogas: 343,9 MW Hydro: 61,2 MW Wasted: 49,4 MW Geothermal: 2,7 MW  source: https://www.mavir.hu/documents/10258/246465111/BT_2015-20230801_ig_BR+NT_HU_v1.pdf/406e9cf1-18e2-c6c6-850c-f4f88765aada?t=1692102013071

PR #5767, opened 2023-08-22
Updated Hungarian capacities

RolandSzep:
## Description

Updated Hungarian solar and battery storage capacities for the month of August. 

### Double check

### DIFF
--- a/web/src/features/map-controls/MapControls.tsx
+++ b/web/src/features/map-controls/MapControls.tsx
@@ -1,6 +1,7 @@
 import { Button } from 'components/Button';
 import { isInfoModalOpenAtom, isSettingsModalOpenAtom } from 'features/modals/modalAtoms';
 import { useAtom, useSetAtom } from 'jotai';
+import { useTransition } from 'react';
 import { FiWind } from 'react-icons/fi';
 import { HiOutlineEyeOff, HiOutlineSun } from 'react-icons/hi';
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
Thanks for the PR!

Before merging this could you take a look and see if we can update the capacities with the data in this issue https://github.com/electricitymaps/electricitymaps-contrib/issues/5763?

RolandSzep:
> Thanks for the PR!
> 
> Before merging this could you take a look and see if we can update the capacities with the data in this issue #5763?

Sure, but these are gross capacities, and we are currently using net capacities (I checked and that’s what is reported towards ENTSO-E also). Net capacities are not reported in the latest pdf published by MAVIR for some reason (they were in the previous pdf). To be honest I don’t what the difference is between net and gross capacities, so I will do as you say. 😅

VIKTORVAV99:
> > Thanks for the PR!
> > Before merging this could you take a look and see if we can update the capacities with the data in this issue #5763?
> 
> Sure, but these are gross capacities, and we are currently using net capacities (I checked and that’s what is reported towards ENTSO-E also). Net capacities are not reported in the latest pdf published by MAVIR for some reason (they were in the previous pdf). To be honest I don’t know what the difference is between net and gross capacities, so I will do as you say. 😅

Not sure what the exact difference is either but gross could be installed capacity and net installed capacity - unavailable due to maintenance and such. If that is the case it's gross we want.

PR #5771, opened 2023-08-22
Make total values in tooltip respect selected mix-mode

madsnedergaard:
## Issue

Closes #5701

Closes #5140, #5759

## Description

This makes the calculation in breakdown tooltips take the selected mode (production/consumption) instead of only using production.

### Preview

Screenshot 2023-08-22 at 22 33 17

Screenshot 2023-08-22 at 22 33 09

### Double check

- [x] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

Issue #4428, opened 2022-08-07
Show energy instead of power in aggregated view table


corradio:
The aggregated view currently shows GW (power). It represents the average power of a zone during the year.
I think it would be more intuitive to show GWh or TWh (energy). It would represent the total energy of a zone.
Also, this is a value which is usually communicated about by experts and by journalists. So far, I've never seen the power value being used in the media.

Mathematically, we could simply take the power value and multiply by the number of hours (it would therefore be robust to missing values).

<img width="420" alt="image" src="https://user-images.githubusercontent.com/1655848/183297361-0f692922-8fc2-4128-8479-b36003a347cd.png">

VIKTORVAV99:
I kind of suggested the same thing in https://github.com/electricitymap/electricitymap-contrib/issues/3938#issuecomment-1171088990 but it probably a good idea for it to have it's own issue.

This should also be changed on the backend so it returns the total energy instead of average power so we can avoid doing these calculations in the frontend.

As for the calculations themselves I think it would be a good idea to include a disclaimer if there is a period with missing information so people don't take it as the absolute truth when we know there is data missing even if it might not change anything. (similar to how the estimated data label work).

Kongkille:
Yes this would be a nice follow up. I am guessing we'd want to do the same for the 'carbon emissions' tab? To show total CO2?

<img width="438" alt="image" src="https://user-images.githubusercontent.com/45588799/183846378-8cc461b9-0b40-4329-89a9-41c6dde8ce04.png">

VIKTORVAV99:
Yeah that seems only logical. Perhaps we could update the current ones to show a value per hour as well since all other values are per hour?

Kongkille:
> Yeah that seems only logical. Perhaps we could update the current ones to show a value per hour as well since all other values are per hour?

Yes it's a bit curious that it's t/min. I am not sure if there's any specific reasoning for it

VIKTORVAV99:
@Kongkille @madsnedergaard @tonypls is this something we could include in the rewrite?
Would be nice to get it added as it would be more clear for the historical views.

madsnedergaard:
While it would be really nice to do, we're already fighting hard to not increase scope of the rewrite project - so let's fix it afterwards :)

VIKTORVAV99:
> While it would be really nice to do, we're already fighting hard to not increase scope of the rewrite project - so let's fix it afterwards :)

[...]

PR #5772, opened 2023-08-23
Support larger emission numbers

madsnedergaard:
## Description

Changes the app to handle larger values for emissions as part of addressing #4428

~~**BLOCKED BY ** #5771 as my local git history is a mess, so I'll push some more stuff here after that PR is merged~~


web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -76,6 +76,8 @@ const getLayers = (
   }));
 };
 
+const defaultFormatTickHandler = (t: number) => t.toString();
+
```
VIKTORVAV99, line 80: Cant we simply use String() instead?

If I remember correctly it should do the same.

web/src/features/charts/hooks/useBreakdownChartData.ts
```diff
@@ -87,8 +87,9 @@ export default function useBreakdownChartData() {
         // in GW or MW
         entry.layerData[key] = Math.max(0, exchangeValue / valueFactor);
         if (displayByEmissions) {
-          // in tCO₂eq/min
-          entry.layerData[key] *= (value.exchangeCo2Intensities || {})[key] / 1e3 / 60;
+          // in gCO₂eq/hour
+          entry.layerData[key] =
+            (value.exchangeCo2Intensities || {})[key] * Math.max(0, exchangeValue);
```
VIKTORVAV99, line 92:
```suggestion
            value.exchangeCo2Intensities?.[key] * Math.max(0, exchangeValue);
```

web/src/features/charts/hooks/useBreakdownChartData.ts
```diff
@@ -130,13 +131,11 @@ function getStorageValue(
...
+  const invertedValue = -1 * Math.min(0, storageValue);
 
-  if (displayByEmissions) {
-    scaledValue *= value.dischargeCo2Intensities[storageKey] / 1e3 / 60;
-  }
-
-  return scaledValue;
+  return displayByEmissions
+    ? invertedValue * value.dischargeCo2Intensities[storageKey] * valueFactor
+    : invertedValue / valueFactor;
```
VIKTORVAV99, line 138: With all this multiplication or division there is a risk we introduce floating point errors.

Are we sure these are rounded upstream?

web/src/features/charts/BreakdownChart.tsx
```diff
@@ -29,6 +31,11 @@ function BreakdownChart({
 
   const { chartData, valueAxisLabel, layerFill, layerKeys } = data;
 
+  // Find highest daily emissions to show correct unit on chart
+  const maxEmissions = max(chartData.map((day) => sum(Object.values(day.layerData))));
```
VIKTORVAV99, line 35: I think this should use absolute values (all scaling should) so we don't risk exports being wrong.

VIKTORVAV99 (commented):
Just some quick thoughts.

web/src/features/charts/hooks/useBreakdownChartData.ts
madsnedergaard: It's not ideal, but we'll attempt to move most of these calculation to app-backend in follow-up work :)

web/src/features/charts/BreakdownChart.tsx
madsnedergaard: Not sure I understand what you mean here.
This is used to make sure we show "0t" (or whatever unit the highest value on the chart is) instead of "0g" at the beginning of the y-axis.

web/src/features/charts/elements/AreaGraph.tsx
madsnedergaard: In this case, where we don't accept any null-ish inputs it's probably fine to use String() :)
Updated now.

web/src/features/charts/BreakdownChart.tsx
VIKTORVAV99: Albeit an edge case if say a zone has 3 modes that are each producing 500kg in emissions and then a export/storage of 1t (1000kg) then the unit should probably be 0t and not 0kg.

Might not matter much since we aren't showing negative values (yet) but if nothing has changes the plan is still to show storage charge and exports in the breakdown graph right?

madsnedergaard: Hmmmm, wouldn't it be weird if this graph included exports/storage? At last currently it's showing the "origin of emissions", so not accounting for what's exported. How would you show both in a nice and not confusing way? 🤔 

Screenshot 2023-08-30 at 07 58 47

Anyway, sounds like a future discussion and I suggest we wait with optimizing the units for this until then :)

VIKTORVAV99: We have discussed it a while ago related to this issue: https://github.com/electricitymaps/electricitymaps-contrib/issues/5259

But perhaps we need to rethink the implementation?

madsnedergaard: Oh sorry, I forgot all about that! Let's keep it as a separate discussion, I'd like to take a step back and try making a few examples (in Figma, Paint, or whatever) to discuss usability if we change and what makes sense for users :)

web/cypress/e2e/countrypanel.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('Country Panel', () => {
 
     cy.contains('Carbon emissions').should('have.attr', 'aria-checked', 'false');
     cy.contains('Carbon emissions').click().should('have.attr', 'aria-checked', 'true');
-    cy.contains('0 t/min');
+    cy.contains('0t/min');
```
tonypls, line 24: Random fact, a tonne is also known as a Megagram (Mg)

web/src/utils/formatting.ts
```diff
@@ -19,18 +18,37 @@ export const formatPower = function (
...
+  const checkAgainst = valueToMatch ? valueToMatch / 1e6 : value;
+
+  // grams and kilograms
+  if (Math.round(checkAgainst) < 1) {
+    return `${d3.format(`,.0~s`)(value * 1e6)}g`;
+  }
+
+  // tons
+  if (Math.round(checkAgainst) < 1e5) {
+    const decimals = value < 1 ? 1 : 0;
```
tonypls, line 41: Would 2dp across the board be nice give users a bit more accuracy?

tonypls (approved):
Few small comments:

I get many of these errors in my console when hovering emissions charts:
image


Should we apply the formatting to the Net Exchange chart too? If we don't use the same formatting I think we should atleast make the unit layout consistent.

This:
image


More similar to this:
image

tonypls:
I also think we should be consistent across all units and use a space between values and units. Aparrently it's more sciencey to have a space [sciencey article on units](https://physics.nist.gov/cuu/Units/checklist.html). I can do a sweet in a separate PR if you want to keep this one from growing

web/src/utils/formatting.ts
```diff
@@ -19,18 +18,37 @@ export const formatPower = function (
   return power;
 };
 
-export const formatCo2 = function (d: number, numberDigits: number = DEFAULT_NUM_DIGITS) {
-  let value = d;
+export const formatCo2 = function (gramPerHour: number, valueToMatch?: number) {
```
VIKTORVAV99, line 21: This feels a little complext in relation to the scale power function we have below. Do we really need all the d3 format stuff instead of just rounding dividing, rounding and returning the unit ourselves. We already do return the units manually.

VIKTORVAV99 (approved):
Approving as there is nothing I consider blocking. But there are some things we can consider changing in a follow up PR.

And I agree with Tony on the space between units (we also use that in the current version of the app).

web/cypress/e2e/countrypanel.spec.ts
VIKTORVAV99: > I think t for tonne is good here, Megagram is probably only known by a niche group.

I'm just worried it might get confused with the long ton (1016 kg) or the short ton (907 kg)

web/src/utils/formatting.ts
```diff
@@ -19,18 +18,37 @@ export const formatPower = function (
...
+  // Hundred thousands of tons
+  if (Math.round(checkAgainst) < 1e6) {
+    return `${d3.format(`,.1~f`)(value / 1e6)}Mt`;
   }
 
-  return value >= 1
-    ? `${d3.format(`.${numberDigits}s`)(value)}t ${translate('ofCO2eqPerMinute')}` // a ton or more
-    : `${d3.format(`.${numberDigits}s`)(value * 1e6)}g ${translate('ofCO2eqPerMinute')}`;
+  // megatons or above
+  return `${d3.format(`,.2~s`)(value)}t`;
```
madsnedergaard, line 51: We can fix the missing space, but it requires using regex to split the result and then put a space in before joining it together again. This is because d3-format returns `4M` when using the SI-prefix

web/src/utils/formatting.ts
madsnedergaard, line 41:
```suggestion
    const decimals = value < 1 ? 2 : 1;
```

madsnedergaard (commented):
> I get many of these errors in my console when hovering emissions charts

No longer seems to appear :)

> tooltip styling
Fixed in another PR

PR #5773, opened 2023-08-24
add EU_ZONES to config

mathilde-daugy:
## Issue

When developing a data validation page on Dora we realised that the EU_ZONES list and config could be added to the general config as it can be used by other projects 
image

## Description

Add list of EU zones
Create EU_ZONES_CONFIG

### Preview

Na

### Double check


electricitymap/contrib/config/__init__.py
```diff
@@ -24,6 +24,37 @@
...
+EU_ZONES = [
+    "AT",
+    "BE",
+    "BG",
+    "CY",
+    "CZ",
+    "DE",
+    "DK",
+    "EE",
+    "ES",
```
pierresegonne, line 36: What about the smaller zones, i.e `FR-COR` etc?

pierresegonne: Do we want to include them here?

mathilde-daugy: good question, i think in the future if we want to aggregate data we will refer to continental EU and it might be misleading to add the smaller zones. 
We can make 2 lists though

FelixDQ (commented):
Could we consider moving this configuration to the zones yaml files? Ideally, we want those to contain all configurations related to zones

electricitymap/contrib/config/__init__.py
VIKTORVAV99: Do the data source we will use to compare it have data for EU28 (before GB exited) if so we might want to account for that as well?

mathilde-daugy: https://iea.blob.core.windows.net/assets/c5b1e844-3f51-4d56-b395-87d7db87cadd/Monthlyelectricitystatistics_Documentation_2023.pdf

pierresegonne:
> Could we consider moving this configuration to the zones yaml files?

This could be a good idea indeed, but might need further thinking on what we want to include in those files, so for me it's ok to start having it as proposed

madsnedergaard:
> Could we consider moving this configuration to the zones yaml files? Ideally, we want those to contain all configurations related to zones

I think it's a good idea, but probably requires a bit more shaping in order to make a data structure/naming that allows any zone to be part of a larger region (EU, the Americas, etc.) - so I'm fine with using this list as first :)

mathilde-daugy:
I will merge this for now and we can take the time to improve the config files later :)

madsnedergaard:
> > > Could we consider moving this configuration to the zones yaml files? Ideally, we want those to contain all configurations related to zones
> > 
> > 
> > I think it's a good idea, but probably requires a bit more shaping in order to make a data structure/naming that allows any zone to be part of a larger region (EU, the Americas, etc.) - so I'm fine with using this list as first :)
> 
> For larger areas I think we should do synchronous grids instead on administrative areas (or both) as they have a larger interconnectivity and influence each other on a broader scale. (Also something Samuel Perkin suggested).
> 
> These would not necessarily be the same as the administrative zones (and would not even follow country borders in some cases) so I think it would require further discussion and shaping on our part.

Yeah, that makes great sense! There's probably going to be some cases where we need to differ a bit from the actual synchronous grids though, where Denmark as an example appears to be split across two different grids 😅

PR #5777, opened 2023-08-25
TotalConsumption events : add validation to filter 0 values

mathilde-daugy:
## Issue
Parsed consumption can be 0 MW 
image

## Description

add parser exception to avoid getting 0s 

### Preview

NA

### Double check

### DIFF
--- a/parsers/IN.py
+++ b/parsers/IN.py
@@ -279,8 +279,7 @@ def fetch_consumption_from_meritindia(
 
         for future in concurrent.futures.as_completed(futures):
             total_consumption += future.result()
-    if total_consumption == 0:
-        raise ParserException(parser="IN.py", message=f"{target_datetime}: No valid consumption data found", zone_key=zone_key)
+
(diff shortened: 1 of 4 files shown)


parsers/IN.py
```diff
@@ -279,7 +279,8 @@ def fetch_state_consumption(session, state):
 
         for future in concurrent.futures.as_completed(futures):
             total_consumption += future.result()
-
+    if total_consumption == 0:
+        raise ParserException(parser="IN.py", message=f"{target_datetime}: No valid consumption data found", zone_key=zone_key)
```
VIKTORVAV99, line 283: Instead of doing it in the parser it might make sense to do it in the TotalConsumptionList class so it's used for other zones as well. Parsed consumption should never be 0.

pierresegonne: Good point, let's do that

pierresegonne (approved):

PR #5778, opened 2023-08-25
Prevent overflow on tooltip datetimes and electricity / emmissions toggle

tonypls:
## Issue
We have some display edge cases

## Description
This PR fixes toggle and tooltip overflow issues

### Preview
Old:
image
image

New:
image

image

### Double check

### DIFF
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -41,14 +41,12 @@ export default function ToggleButton({
   return (
     <div
       className={twMerge(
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
While it's definitely a big improvement there is still come cases where it looks a bit odd.
Like this one:
image
Could we prevent the big left margin without changing the whole structure so it's a bit more centered at least?

madsnedergaard (approved):
I vote for merging this now, and then improving it further later 🥳 
(I also think it requires more work if we want to support an even wider panel dynamically, and we'll most likely be revisiting the whole left-panel again in a somewhat near future)

PR #5789, opened 2023-08-29
Upgrade MX parser

unitrium:
## Issue

Upgrading the remaining functions of the MX parser to the Event datastructure.

## Description
- Upgrade exchanges to ExchangesList
- Upgrade Production to ProductionBreakdownList.

### Double check


VIKTORVAV99 (approved):
Nice work getting rid of Arrow as well. Should make it easy to switch to the built in timezone package in the newer version of python when we upgrade.

PR #5798, opened 2023-08-31
Fix EIA parser reports 0 hydro production as 0 hydro storage

unitrium:
## Issue

We noticed that US-NW-NWMT went down on the map. The reason for this is that estimations are failing because of an outlier in the data. The EIA reported 0 hydro production at 2023-08-23 15:00:00.000000 +00:00 which got interpreted by our parser as a 0 hydro storage. In itself its not wrong but this extra transformation is causing troubles down the line so we should stick to the normal data as much as possible and therefore report it as 0 production.

## Description

- Update the `create_production_storage` so that 0 hydro production is reported as production instead of storage.
- simplify the function logic by inverting the if clause.

### Double check

### DIFF
--- a/parsers/EIA.py
+++ b/parsers/EIA.py
@@ -447,7 +447,7 @@ def create_production_storage(
     production_value = production_point["value"]
     production_mix = ProductionMix()
     storage_mix = StorageMix()
-    if production_value > 0:
+    if production_value >= 0:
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5799, opened 2023-08-31
Support higher energy values and change labels to watt hour

madsnedergaard:
## Issue

Part of #4428

## Description

Changes power formatting to handle higher values, and also changes from power (KW) to energy (KWh).

### Double check


web/src/utils/formatting.ts
```diff
@@ -12,12 +12,21 @@ export const formatPower = function (
...
     return d;
   }
-  const power = `${d3.format(`.${numberDigits}s`)(d * 1e6)}W` //Add a space between the number and the unit
+  const power = `${d3.format(`.${numberDigits}s`)(d * 1e6)}Wh` //Add a space between the number and the unit
     .replace(/([A-Za-z])/, ' $1')
     .trim();
   return power;
 };
 
+export const formatPowerWithSameUnit = function (
```
tonypls, line 21: Doesn't look like this one is being used

tonypls (approved):
LGTM apart from the unused function

web/src/utils/formatting.ts
```diff
@@ -64,21 +73,35 @@ const scalePower = function (maxPower: number | undefined) {
...
 
   if (value < 1e3) {
     return {
-      unit: PowerUnits.MEGAWATTS,
+      unit: EnergyUnits.MEGAWATT_HOURS,
       formattingFactor: 1,
     };
   }
 
+  if (value < 1e6) {
```
VIKTORVAV99, line 88: does it make sense to turn this into a switch case now that the if satements have grown?
I feel like they are easier to read over repeated if statements.

web/src/utils/formatting.ts
```diff
@@ -64,21 +73,35 @@ const scalePower = function (maxPower: number | undefined) {
 
   if (value < 1) {
```
VIKTORVAV99, line 74: Do we always want to use EnergyUnits or do we want to use PowerUnits for hourly and energy for the rest. For hourly they have the same value and meaning.

VIKTORVAV99 (approved):
No functional problems so approving. But I left some comments to think about.

web/src/utils/formatting.ts
madsnedergaard: I think it makes most sense to stick with energy for consistency :)

web/src/utils/formatting.ts
madsnedergaard: Good point, I've rewritten it to be cleaner by using a mapping :)

PR #5813, opened 2023-09-01
Feature flagging 🎏 

madsnedergaard:
## Description

This calls a new endpoint on app-backend to fetch a list of enabled feature flags. It also exposes a small widget for toggling feature flags in the app when using a `?ff` url param. Name of this could probably be improved a lot, but I wanted to keep it short and easy... Open to suggestions though :)

I initially started with https://github.com/sergiodxa/flagged, but found that we wanted to customise it all to use TanStack Query and it didn't make sense to continue using a very small library for this. We could copy [their render prop](https://github.com/sergiodxa/flagged) implementation later on if we want to use it like this.

### Preview

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/9b620f03-ac05-4554-b734-ce47f87c0357

**Example usage:**

Screenshot 2023-09-01 at 16 41 42


web/src/features/feature-flags/api.ts
```diff
@@ -0,0 +1,40 @@
...
+    useQuery<FeatureFlags>([QUERY_KEYS.FEATURE_FLAGS], async () => getFeatureFlags(), {
+      suspense: true,
+    }).data || {}
+  );
+}
+
+export function useFeatureFlag(name: string): boolean {
+  const features = useFeatureFlags();
+
+  return features?.[name] || false;
```
VIKTORVAV99, line 39: Any reason why we don't return the key directly?
```suggestion
  return useFeatureFlags()?.[name] || false;
```

web/src/features/feature-flags/api.ts
```diff
@@ -0,0 +1,40 @@
...
+  const requestOptions: RequestInit = {
+    method: 'GET',
+    headers: await getHeaders(path),
+  };
+
+  try {
+    const response = await fetch(`${getBasePath()}${path}`, requestOptions);
+    if (response.ok) {
+      const data = await response.json();
+      return data;
```
VIKTORVAV99, line 16: Same here.
```suggestion
      return await response.json();
```

web/src/features/feature-flags/api.ts
```diff
@@ -0,0 +1,40 @@
+import { useQuery } from '@tanstack/react-query';
+import { QUERY_KEYS, getBasePath, getHeaders } from 'api/helpers';
+import { FeatureFlags } from './types';
+
+export async function getFeatureFlags(): Promise<FeatureFlags> {
+  const path = '/feature-flags';
```
VIKTORVAV99, line 6: If this ever changes in the future we only have to change it once.
```suggestion
  const path = `/${QUERY_KEYS.FEATURE_FLAGS}`;
```
Don't forget to add the import.

web/src/features/feature-flags/api.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (response.ok) {
+      const data = await response.json();
+      return data;
+    }
+
+    throw new Error(await response.text());
+  } catch (error) {
+    // If the request fails, we will return an empty object instead of throwing an error
+    // as the app might still be functional without feature flags
+    console.log(error);
```
VIKTORVAV99, line 23:
```suggestion
    console.error(error);
```

VIKTORVAV99 (commented):
Some additional comments. Nothing blocked so will approve when resolved either way.

web/src/features/feature-flags/api.ts
madsnedergaard: No specific reason, I guess I personally find my version a bit more readable and easier to comprehend though :)

web/src/features/feature-flags/api.ts
madsnedergaard: I personally think it's easier to read and understand the more explicit approach (as "return await" might require a bit of extra mental processing), but I'm also fine with changing it :)

web/src/features/feature-flags/api.ts
VIKTORVAV99: I think there is a small performance advantage to returning it directly but since this is a function that won't run that often I don't think it matters much.

Go with what you prefer on both of them 🙂

VIKTORVAV99 (approved):
LGTM! 🎉

web/src/features/feature-flags/api.ts
madsnedergaard: It most likely does, but even if it doesn't I think the performance cost of assigning a constant would be extremely small 😄

PR #5814, opened 2023-09-01
Prefetch data with TanStack Query

madsnedergaard:
## Description

During work on #5813 I found that we can utilise a `prefetchQuery` to load data initially, instead of leaving unused hooks around to force early downloads. This implements that for the new feature-flag query and for "last_hour" state query.

I also switched to using Suspense for data-fetching as well, so that we can simplify the logic of when to show a loader and when not :)


web/src/App.tsx
```diff
@@ -31,13 +30,18 @@ if (isProduction) {
   });
 }
 
+function LoadingScreen() {
+  return (
+    <main className="fixed flex h-screen w-screen flex-col pt-[56px]">
```
madsnedergaard, line 35: I decided to leave the header out and just keep the same padding to top - I was maybe afraid that people would click the outbound links in the header while waiting (if on a slow connection), not knowing that this would take them away and require restarting any data fetching.

VIKTORVAV99: Smart move.

VIKTORVAV99 (commented):
As far as I can tell it works but there is a type error which blocks CI. Could you look into that?

madsnedergaard:
> As far as I can tell it works but there is a type error which blocks CI. Could you look into that?

Ah, fixed :)

madsnedergaard:
Ugh, this means that the app now shows the error screen if fetching state fails. I'll spend some more time on this later to see if I can make it put a nice alert on top of the map instead :)

madsnedergaard:
> Ugh, this means that the app now shows the error screen if fetching state fails. I'll spend some more time on this later to see if I can make it put a nice alert on top of the map instead :)

I actually couldn't and it required more work, so I think we should drop the Suspense work and just have prefetching. Will pick it up again at some point soon

PR #5815, opened 2023-09-01
Render solar layer for the whole planet

Ugzuzg:
## Issue

Currently, the solar layer doesn't handle map interaction very well.

[Screencast from 2023-09-01 17-22-07.webm](https://github.com/electricitymaps/electricitymaps-contrib/assets/1078004/071564c3-e845-47e3-9af4-6e525073688a)

It also shifts down somehow when the map wraps around:
Screenshot from 2023-09-01 15-38-56

Due to the fact that the layer only renders what's in view and then applies blur on that boundary, solar patterns change in close-ups while navigating and zooming the map, but the underlying data is still the same.

[Screencast from 2023-09-02 10-40-51.webm](https://github.com/electricitymaps/electricitymaps-contrib/assets/1078004/d3f71974-1c46-49d7-8bc5-ab00c6e75845)

Average execution time per render: 640ms. Rendered on every map interaction.

## Description

Render the solar layer for the whole planet to reduce the choppiness and add it as a mapbox layer.

Average execution time per render: 200ms. Rendered only when solarData changes.

### Preview

[2023-09-01.webm](https://github.com/electricitymaps/electricitymaps-contrib/assets/1078004/43440780-aff7-443c-9acd-607a31ce6c8a)

### Double check


VIKTORVAV99 (approved):
I have been testing this a few days now and it seems to work really well.

My only concern is how it affects lower powered devices but I think we should give it a try.

PR #5817, opened 2023-09-01
some windy fixes

VIKTORVAV99:
## Issue

This file is a mess 😅

## Description

Extracts some of the inner functions to the outer scope so they can be resued by the browsers.

Fixes var declarations and replace them with const and let.

Fixes some type declarations.

### Double check


VIKTORVAV99:
I know the CI is failing on this but it can be ignored for now as it is just me shifting code around which makes the code scanning think it's new code when it's not.

madsnedergaard (approved):
Well, it seems to work on deployment and you added some tests that are passing, so I guess that's good enough - because reviewing this code is also close to impossible 😅

VIKTORVAV99:
> Well, it seems to work on deployment and you added some tests that are passing, so I guess that's good enough - because reviewing this code is also close to impossible 😅 

Which is just one of the many reasons we need to clean these up... 😅

I tested it pretty through locally and didn't find any issues but it's not the easiest code to test...

PR #5818, opened 2023-09-02
Update FI.yaml

pnook:
Latest update as per Fingrid


VIKTORVAV99 (approved):

PR #5824, opened 2023-09-04
Complete rewrite of ES.py parser

### LINKED ISSUE #5595: Missing oil generation in Ibiza and Menorca
## Description
It seems the values for diesel (oil) generation are not being recorded:

Screenshot_20230712_164941
Screenshot_20230712_165018

This happens also in Menorca.

Running the parser locally shows oil production as 0 MW:
```bash
{'correctedModes': [],
  'datetime': datetime.datetime(2023, 7, 12, 17, 0, tzinfo=datetime.timezone.utc),
  'production': {'biomass': 0.0,
                 'coal': 0.0,
                 'gas': 82.4,
                 'hydro': 0.0,
                 'nuclear': 0.0,
                 'oil': 0.0,
                 'solar': 1.5,
                 'unknown': 0.0,
                 'wind': 0.0},
  'source': 'demanda.ree.es',
  'storage': {},
  'zoneKey': 'ES-IB-IZ'}
```

silimotion:
## Issue
Closes #5595 

## Description

Complete rewrite of ES.py parser to stop using ree dependency and fix bugs that caused improper reporting of data in balearic islands.

I've checked all possible requests, and everything seems to work the same as before except for:
- Instead of returning 0 when no information is available (e.g. nuclear in menorca), it returns None in that fuel categories.
- Hydro is set by default to hydro storage, as in the data it is hydro + hydro storage. This doesn't change any production values as there aren't islands with hydro power plants except El Hierro. There is a hydro storage being built in another canary island so when it comes online this should keep working.

### Preview
Screenshot_20230904_122057

### Double check


silimotion:
It seems tests will have to be rewritten as they use ree package also. It should be done in this same PR right?

VIKTORVAV99:
> It seems tests will have to be rewritten as they use ree package also. It should be done in this same PR right?

Yes, otherwise it won't pass CI.

parsers/ES.py
```diff
@@ -1,27 +1,87 @@
...
+    "Gomera": "Canarias",
+    "LaPalma": "Canarias",
+    "Tenerife": "Canarias",
+    "LanzaroteFuerteventura": "Canarias",
+    "ElHierro": "Canarias",
+}
+TIMEZONES_MAPPING = {
+    "IberianPeninsula": "Europe/Madrid",
+    "Ceuta": "Europe/Madrid",
+    "Melilla": "Europe/Madrid",
```
VIKTORVAV99, line 73:
```suggestion
    "Ceuta": "Africa/Ceuta",
    "Melilla": "Africa/Ceuta",
```

parsers/ES.py
```diff
@@ -1,27 +1,87 @@
...
+    "Menorca": "Baleares",
+    "Ibiza": "Baleares",
+    "Formentera": "Baleares",
+    "GranCanaria": "Canarias",
+    "Gomera": "Canarias",
+    "LaPalma": "Canarias",
+    "Tenerife": "Canarias",
+    "LanzaroteFuerteventura": "Canarias",
+    "ElHierro": "Canarias",
+}
```
VIKTORVAV99, line 69: These mappings should be using the zone keys instead of the names. This also includes the timezones.

parsers/ES.py
```diff
@@ -41,41 +101,26 @@
...
+    ZoneKey("ES-CN-HI"): "ElHierro",
+    ZoneKey("ES-CN-IG"): "Gomera",
+    ZoneKey("ES-CN-LP"): "LaPalma",
+    ZoneKey("ES-CN-TE"): "Tenerife",
+    ZoneKey("ES-IB-FO"): "Formentera",
+    ZoneKey("ES-IB-IZ"): "Ibiza",
+    ZoneKey("ES-IB-MA"): "Mallorca",
+    ZoneKey("ES-IB-ME"): "Menorca",
+    ZoneKey("ES-ML"): "Melilla",
 }
```
VIKTORVAV99, line 117: This is no longer needed since we aren't importing the functions from REE.

parsers/ES.py
```diff
@@ -41,41 +101,26 @@
...
-    "nuclear": "nuclear",
-    "other": "unknown",
-    "solar": "solar",
-    "vapor": "oil",
-    "waste": "biomass",
-    "wind": "wind",
+    "ES->ES-IB-MA": ZoneKey("ES"),
+    "ES-IB-IZ->ES-IB-MA": ZoneKey("ES-IB-MA"),
+    "ES-IB-FO->ES-IB-IZ": ZoneKey("ES-IB-FO"),
+    "ES-IB-MA->ES-IB-ME": ZoneKey("ES-IB-MA"),
```
VIKTORVAV99, line 123: Same with this.

parsers/ES.py
```diff
@@ -1,27 +1,87 @@
...
-    Mallorca,
-    Melilla,
-    Menorca,
-    Response,
-    Tenerife,
-)
+from arrow import get, utcnow
+
+FUEL_MAPPING = {
+    "dem": "demand",
```
VIKTORVAV99, line 11: Should not be included in here as demand is not a fuel.

I would create a new list with ignored values that includes dem and other values we want to ignore in the future.

parsers/ES.py
```diff
@@ -113,17 +158,50 @@ def check_valid_parameters(
...
+            "ES.py",
+            f"Failed fetching data for {zone_key}",
+            zone_key,
+        )
+    json = loads(res.text.replace("null(", "", 1).replace(r");", "", 1))
+    data = json["valoresHorariosGeneracion"]
+    responses = []
+    for value in data:
+        ts = value.pop("ts")
+        arrow = get(ts + " " + timezone, "YYYY-MM-DD HH:mm ZZZ")
```
VIKTORVAV99, line 188: This should be using a f string and if possible datetime.datetime instead of arrow.

parsers/ES.py
```diff
@@ -113,17 +158,50 @@ def check_valid_parameters(
...
+        response = {}
+        response["timestamp"] = str(arrow)
```
VIKTORVAV99, line 190: Why are you turning this into a string here? We can return it directly as a datetime object later on.

parsers/ES.py
```diff
@@ -113,17 +158,50 @@ def check_valid_parameters(
...
+            if key in fuel_mapping:
+                if fuel_mapping[key] in response.keys():
+                    response[fuel_mapping[key]] += value[key]
+                else:
+                    response[fuel_mapping[key]] = value[key]
+            elif key in LINK_MAPPING:
+                if LINK_MAPPING[key] in response.keys():
+                    response[LINK_MAPPING[key]] += value[key]
+                else:
+                    response[LINK_MAPPING[key]] = value[key]
```
VIKTORVAV99, line 201: This should not be aggregated here. We should aggregate it later on using the parser classes, they contain built in validation to ensure data quality which will be degraded if we aggregate it here first.

parsers/ES.py
```diff
@@ -163,40 +243,46 @@ def fetch_production(
...
+            "biomass",
+            "coal",
+            "gas",
+            "geothermal",
+            "nuclear",
+            "oil",
+            "solar",
+            "wind",
+            "unknown",
+        ]:
```
VIKTORVAV99, line 275: It should loop over all modes here, not just the ones that has a defined mode so we can log if something is missing.

parsers/ES.py
```diff
@@ -163,40 +243,46 @@ def fetch_production(
...
-                    production.hydro = 0
-
+        if "hydro" in event.keys():
+            storage.hydro = -event["hydro"]
         productionEventList.append(
             zoneKey=zone_key,
-            datetime=datetime.fromtimestamp(event.timestamp).astimezone(timezone.utc),
+            datetime=datetime.fromisoformat(event["timestamp"]).astimezone(
+                timezone.utc
+            ),
```
VIKTORVAV99, line 285: As noted, with the above changes that i suggested datetime.fromisoformat is not needed.

VIKTORVAV99 (changes requested):

parsers/ES.py
```diff
@@ -37,45 +21,107 @@
...
+    "trn": "production.unknown",
+    "otrRen": "production.unknown",
+    "cogen": "production.gas",
+    "cb": "exchange.pe_ma",
+    "icb": "exchange.pe_ma",
+    "emm": "exchange.ma_me",
+    "emi": "exchange.ma_ib",
+    "eif": "exchange.ib_fo",
+    "inter": "exchange.int",
+}
```
VIKTORVAV99, line 68: Why prefix these are require us to split the strings later on instead of having multiple lists which is both faster and easier to maintain?

parsers/ES.py
```diff
@@ -37,45 +21,107 @@
...
+    ZoneKey("ES-CN-HI"): "Atlantic/Canary",
+    ZoneKey("ES-CN-IG"): "Atlantic/Canary",
+    ZoneKey("ES-CN-LP"): "Atlantic/Canary",
+    ZoneKey("ES-CN-TE"): "Atlantic/Canary",
+    ZoneKey("ES-IB-FO"): "Europe/Madrid",
+    ZoneKey("ES-IB-IZ"): "Europe/Madrid",
+    ZoneKey("ES-IB-MA"): "Europe/Madrid",
+    ZoneKey("ES-IB-ME"): "Europe/Madrid",
+    ZoneKey("ES-ML"): "Africa/Ceuta",
 }
```
VIKTORVAV99, line 116: These can probably be combined, either as a dict with the zoneKey as the key and appropriate subkeys for each of the values or as a named tuple.

parsers/ES.py
```diff
@@ -37,45 +21,107 @@
...
-    "hydraulic": "hydro",
-    "nuclear": "nuclear",
-    "other": "unknown",
-    "solar": "solar",
-    "vapor": "oil",
-    "waste": "biomass",
-    "wind": "wind",
+SOURCE = "demanda.ree.es"
+
+EXCHANGE_FUNCTION_MAP: Dict[str, Callable] = {
```
VIKTORVAV99, line 120: These types have to be updated and so should the naming, it's no longer a function map and they are no longer callable. It should probably be a map of two ZoneKeys.

parsers/ES.py
```diff
@@ -113,17 +159,61 @@ def check_valid_parameters(
...
+                elif data_mapping[mode] == "demand":
+                    response["demand"] = value[mode]
+            except KeyError:
+                raise ParserException("ES.py", f"Found unknown key:{mode}", mode)
+        response["production"] = production
+        response["storage"] = storage
+        response["exchange"] = exchange
+        responses.append(response)
+    if responses:
+        return responses
```
VIKTORVAV99, line 216: This is very complicated comnpared to the old version. Espeically with all the string matching and replacing going on.

parsers/ES.py
```diff
@@ -163,42 +253,29 @@ def fetch_production(
...
+
+    if zone_key == "ES-IB-ME" or zone_key == "ES-IB-FO":
+        data_mapping["gas"] = "production.oil"
+    if zone_key == "ES-IB-IZ":
+        data_mapping["die"] = "prosuction.gas"
+
+    island_data = fetch_island_data(zone_key, ses, target_datetime, data_mapping)
     productionEventList = ProductionBreakdownList(logger)
-
     for event in island_data:
```
VIKTORVAV99, line 273: It would be better if production logic was handled here, and exhange in exchange, etc.

parsers/ES.py
```diff
@@ -113,17 +159,61 @@ def check_valid_parameters(
...
+    responses = []
+    for value in data:
+        ts = value.pop("ts")
+        arrow = get(f"{ts} {tz}", "YYYY-MM-DD HH:mm ZZZ")
+        response = {}
+        response["timestamp"] = arrow
+        production = ProductionMix()
+        storage = StorageMix()
+        exchange = {}
+        for mode in value:
```
VIKTORVAV99, line 195: If you still want to have a preprocess function I suggest splitting it out from the fetch data function as they do two very seperate things.

VIKTORVAV99 (changes requested):
There are some things in here that are way more complex than they need to be which will hurt maintainability long term. These need to be adressed before we can proceede.

parsers/ES.py
VIKTORVAV99: That parser is under development and should probably not be used as an example until it is done. It will also likely be refactored now that we are adding storage to the parser. A better example would be ENTSOE or EIA as they are fully feature complete and handle multiple zones.

Adding so much reliance on string values will make it harder to maintain and it will be easier to introduce errors.

silimotion: Ok, I will not base my code on that parser. However, I don't see how could we make this code work and raise a warning when new unknown keys are passed. I understood from a previous message of yours that this was necessary, but after re-reading it I have noticed I may have misunderstood. I think that it is a nice addition though. Could you guide me in how would you like the parser to be structured like? Should it pass a dict with the original keys (e.g. dem, die, nuc, etc...) to all the fetch_x functions, and the conversion to Electricity Maps data be done there? It would lose the ability to check that all keys have been used.

silimotion: I now see that this current implementation is a bit fragile. I can implement this other code, if you don't have a better idea.

VIKTORVAV99: If we split them into 3 lists, production, exchange, and ignored it would be easy. Either we check everything in a preprocess and make sure the key are in one of the lists. Or we check it in each of the fetch function. Both have their pros and conns, doing everything in a preprocess step will make the code slightly more complex but less code duplication.

In the second approach we do it in each parser and duplicate (some of) the code but it makes it less complex and easier to understand in my opinion.

For the preprocess approach I would do something like this:

```python
  for key of keys:
    if key in production:
      # handle production
    elif key in exchange:
      # handle exchanges
    elif key == "dem"
      # handle demand
    else:
      logger.warning("Log warning about key"
```

Here is how I would do it for production with the second approach.

```python
  for key of keys:
    if key in production_map:
      # Handle production here
    elif key not in exchange_map and key not in ignored_values: # We don't need to check for dem explicitly here as it would be in ignored values, in the preprocess step we have to explicitly check it instead.
      logger.warning("Log warning about key")
```

Both avoid the string prefix by putting them in different lists though.

parsers/ES.py
silimotion: I think joining all in one tuple would make it more complicated and less self-explanatory, even though it would be less repetitive.

silimotion:
I have a few questions / things to note before finishing this PR:
- We could update the Canary Islands LNG check, even though it is not related to the PR
- We need to remove ree dependency from the repository dependencies, I waited till the end because I wasn't sure if it would be easier for you to test the results with it or something
- I set "cogen" (Cogeneration) to gas because I found a REE document that explains the weights of each generation type:
Screenshot_20230909_222314
So ~90% is gas and ~10% is oil. Maybe it would be better to set it to unknown, or add 0.9 times the value to gas and 0.1 to oil. This document also contains data about the CO2 intensity of the different production types but I think that should be discussed later.

VIKTORVAV99:
> I have a few questions / things to note before finishing this PR:
> 
> * We could update the Canary Islands LNG check, even though it is not related to the PR
> * We need to remove ree dependency from the repository dependencies, I waited till the end because I wasn't sure if it would be easier for you to test the results with it or something
> * I set "cogen" (Cogeneration) to gas because I found a REE document that explains the weights of each generation type:
>   Screenshot_20230909_222314
>   So ~90% is gas and ~10% is oil. Maybe it would be better to set it to unknown, or add 0.9 times the value to gas and 0.1 to oil. This document also contains data about the CO2 intensity of the different production types but I think that should be discussed later.

REE should be removed completely in this PR as there is no need for it anymore.

We should under no circumstances add a static split in the parser. If needed we should adjust the unknown emission factor.

parsers/ES.py
```diff
@@ -112,18 +157,44 @@ def check_valid_parameters(
...
+    target_datetime: Optional[datetime],
+):
+    """Fetch data for the given zone key."""
+    tz = TIMEZONES_MAPPING[zone_key]
+    data = fetch_data_ree(zone_key, session, target_datetime, tz)
+    responses = []
+    for value in data:
+        ts = value.pop("ts")
+        date = datetime.fromisoformat(ts).replace(tzinfo=ZoneInfo(tz))
+        value["timestamp"] = date
```
VIKTORVAV99, line 194: This should not be called timestamp if it is in face a datetime object.

parsers/ES.py
```diff
@@ -112,18 +157,44 @@ def check_valid_parameters(
...
+def fetch_data_ree(
+    zone_key: ZoneKey, session: Session, target_datetime: Optional[datetime], tz: str
+) -> Dict:
     if target_datetime is None:
-        date = target_datetime
+        date = datetime.now(tz=ZoneInfo(tz)).strftime("%Y-%m-%d")
     else:
         date = target_datetime.strftime("%Y-%m-%d")
+    system = ZONE_MAPPING[zone_key]
+    zone = API_CODE_MAPPING[zone_key]
```
VIKTORVAV99, line 168: As specified before I think it would be better to combine these lists into one.

Something like:
```python
{
  ZoneKey("ES"): {
    "API_CODE": "code",
    "SYSTEM": "system",
    "TZ": "timezone"
  },
  # ...
}
```
It would remove a lot of duplicate code and have a nice and clear object grouping for each zone.

parsers/ES.py
```diff
@@ -163,40 +234,41 @@ def fetch_production(
...
-                    # Reset gas to 0, as it is now oil, normally this should be skipped or set to None but to conform to how REE does it, we set it to 0
-                    # We should aim to change this upstream in the future to make it uniform with other parsers going forward.
-                    production.gas = 0
-                production.add_value(fuel_mapping[mode], value, True)
-                # Zone specific override for El Hierro
-                # Hydro response is hydro storage, not hydro production
-                if zone_key == "ES-CN-HI" and mode == "hydraulic":
-                    storage.hydro = -value
-                    production.hydro = 0
+        timestamp = event.pop("timestamp").astimezone(timezone.utc)
```
VIKTORVAV99, line 260: Why are we poping the timestamp (that is actually a datetime) instead of just assigning it later on?

And as we are adding a timezone in the preprocess step we don't need to do astimezone here.

parsers/ES.py
```diff
@@ -1,27 +1,10 @@
...
-    Ibiza,
-    LanzaroteFuerteventura,
-    LaPalma,
-    Mallorca,
-    Melilla,
-    Menorca,
-    Response,
-    Tenerife,
-)
+from typing import Dict, Optional
```
VIKTORVAV99, line 6: Neither of these type imports are needed in python 3.10

We can do `dict` instead of `Dict` and `type | None` instead of `Optional[type]`.

VIKTORVAV99 (commented):
Some last notes, some are related to the migration to Python 3.10 that happened during this PR was opened.

parsers/ES.py
```diff
@@ -112,18 +176,44 @@ def check_valid_parameters(
         )
 
 
-def fetch_island_data(
-    zone_key: ZoneKey, session: Session, target_datetime: datetime | None
-) -> list[Response]:
-    """Fetch data for the given zone key."""
+def fetch_data_ree(
```
VIKTORVAV99, line 179:
```suggestion
def get_ree_data(
```

parsers/ES.py
```diff
@@ -147,8 +237,8 @@ def fetch_consumption(
     for event in island_data:
         consumption.append(
             zoneKey=zone_key,
-            datetime=datetime.fromtimestamp(event.timestamp).astimezone(timezone.utc),
-            consumption=event.demand,
+            datetime=event["datetime"].astimezone(timezone.utc),
```
VIKTORVAV99, line 240:
```suggestion
            datetime=event["datetime"]
```

parsers/ES.py
```diff
@@ -215,35 +309,36 @@ def fetch_exchange(
...
         else:
-            net_flow = response.link["pe_ma"]
+            net_flow = exchanges["pe_ma"]
 
         exchangeList.append(
             zoneKey=sorted_zone_keys,
-            datetime=datetime.fromtimestamp(response.timestamp).astimezone(
-                timezone.utc
-            ),
+            datetime=event["datetime"].astimezone(timezone.utc),
```
VIKTORVAV99, line 341:
```suggestion
            datetime=event["datetime"],
```

parsers/ES.py
```diff
@@ -215,35 +309,36 @@ def fetch_exchange(
...
     )
 
     exchangeList = ExchangeList(logger)
-    for response in responses:
+    for event in responses:
+        exchanges = {}
+        for key in event:
+            if key in EXCHANGE_PARSE_MAPPING.keys():
+                exchanges[EXCHANGE_PARSE_MAPPING[key]] = event[key]
+            elif key not in PRODUCTION_PARSE_MAPPING and key != "dem" and key != "hid":
```
VIKTORVAV99, line 326:
```suggestion
            elif key not in PRODUCTION_PARSE_MAPPING and key != "dem" and key != "hid" and key != "datetime":
```

I would suggest declaring a list for this somehwere as it is reused in several places, or move all the warnings into the preprocess step by checking all the keys there (dem, production and exchange).

parsers/ES.py
```diff
@@ -112,18 +176,44 @@ def check_valid_parameters(
...
+    target_datetime: datetime | None,
+):
+    """Fetch data for the given zone key."""
+    tz = ZONE_MAPPING[zone_key]["TZ"]
+    data = fetch_data_ree(zone_key, session, target_datetime, tz)
+    responses = []
+    for value in data:
+        ts = value.pop("ts")
+        date = datetime.fromisoformat(ts).replace(tzinfo=ZoneInfo(tz))
+        value["datetime"] = date
```
VIKTORVAV99, line 213:
```suggestion
        value["datetime"] = datetime.fromisoformat(ts).replace(tzinfo=ZoneInfo(tz))
```

parsers/ES.py
```diff
@@ -112,18 +176,44 @@ def check_valid_parameters(
...
+    zone_key: ZoneKey,
+    session: Session,
+    target_datetime: datetime | None,
+):
+    """Fetch data for the given zone key."""
+    tz = ZONE_MAPPING[zone_key]["TZ"]
+    data = fetch_data_ree(zone_key, session, target_datetime, tz)
+    responses = []
+    for value in data:
+        ts = value.pop("ts")
```
VIKTORVAV99, line 211: Why are we poping ts instead of using it directly?

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for bearing with me through all the reviews!

PR #5821, opened 2023-09-03
Update dark-mode styling of map controls

madsnedergaard:
## Description

Make the map control buttons and selectors look nicer and more similar to the toggles.
Also updated styling on a bunch of tooltips that were missed earlier.

### Preview

Screenshot 2023-09-04 at 00 18 59

PR #5825, opened 2023-09-04
Improve map-control buttons look and feel

madsnedergaard:
## Issue

Follow-up from #5821

## Description

Makes the buttons in light mode nice and crisp too, as well as aligning hover styling on both themes.

### DIFF
--- a/web/src/features/map-controls/MapButton.tsx
+++ b/web/src/features/map-controls/MapButton.tsx
@@ -25,7 +25,7 @@ export default function MapButton({
     <TooltipWrapper tooltipContent={tooltipText}>
       <Component
         onClick={onClick}
-        className={`pointer-events-auto flex h-8 w-8 items-center justify-center rounded bg-white/90 text-left shadow-lg backdrop-blur-sm transition hover:bg-gray-100 dark:bg-gray-800/80 dark:hover:bg-gray-900/90 ${className}`}
+        className={`pointer-events-auto flex h-8 w-8 items-center justify-center rounded bg-white/80 text-left shadow-lg backdrop-blur-sm transition  hover:bg-white dark:bg-gray-800/80 dark:hover:bg-gray-900/90 ${className}`}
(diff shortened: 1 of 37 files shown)


web/src/features/map-controls/MapButton.tsx
```diff
@@ -25,7 +25,7 @@ export default function MapButton({
     <TooltipWrapper tooltipContent={tooltipText}>
       <Component
         onClick={onClick}
-        className={`pointer-events-auto flex h-8 w-8 items-center justify-center rounded bg-white/90 text-left shadow-lg backdrop-blur-sm transition hover:bg-gray-100 dark:bg-gray-800/80 dark:hover:bg-gray-900/90 ${className}`}
+        className={`pointer-events-auto flex h-8 w-8 items-center justify-center rounded bg-white/80 text-left shadow-lg backdrop-blur-sm transition  hover:bg-white dark:bg-gray-800/80 dark:hover:bg-gray-900/90 ${className}`}
```
VIKTORVAV99, line 28:
```suggestion
        className={`pointer-events-auto flex h-8 w-8 items-center justify-center rounded bg-white/80 text-left shadow-lg backdrop-blur-sm transition hover:bg-white dark:bg-gray-800/80 dark:hover:bg-gray-900/90 ${className}`}
```

web/src/index.css
VIKTORVAV99, line 1: Should we change the file extension to postcss? I think it makes sense now that we are no longer using plain css as some language servers (vs code extensions) might behave differently.

VIKTORVAV99 (approved):
Approving!

The comments are nothing of note really, just a removal of a space and a thought that don't really relate to the changes themselves.

web/src/index.css
madsnedergaard: Hmm, not sure about that - think it will cause more problems by editors and people not knowing which language a `.postcss` file is :)

PR #5826, opened 2023-09-05
Select api version based on feature flag

tonypls:
## Issue
Draft PR for inspiration if we want to do something like this later

## Description
This PR adds allows the zone api to select between v6 and v7 based on the total-energy feature flag

### Double check


madsnedergaard:
Ah nice! I tried something similar at first, but it introduced some issues that I now cannot remember 😅 
Maybe it was something about the order - we need to make sure feature-flags are fetched before state begins, so it knows which version to use. But I guess we could change the `enabled` param depending on this if we want to

tonypls:
Closing as we don't use this pattern for now :)

PR #5831, opened 2023-09-05
Update emission naming values

tonypls:
## Issue
We previously considered emission to be grams of CO2 / minute, we're moving to total emissions for each datetime and granularity

## Description
This PR updates units in the app related to emissions to no longer per minutes

The charts made sense with tonne as the base unit in the axis labels. It would be worth checking this still makes sense when values are very low. 

### Double check

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -208,6 +208,7 @@
   "electricityImportedFrom": "<b>%1$s %%</b> of electricity available in %4$s <b>%2$s</b> is imported from %5$s <b>%3$s</b>",
   "emissionsImportedFrom": "<b>%1$s %%</b> of emissions from electricity available in %4$s <b>%2$s</b> are imported from %5$s <b>%3$s</b>",
   "ofCO2eqPerMinute": "of CO₂eq per minute",
+  "ofCO2eq": "of CO₂eq",
   "theMap": {
     "groupName": "The map",
     "mapColors-question": "What do the colors on the map mean?",
--- a/web/src/features/charts/EmissionChart.tsx
+++ b/web/src/features/charts/EmissionChart.tsx
@@ -32,7 +32,7 @@ function EmissionChart({ timeAverage, datetimes }: EmissionChartProps) {
         data={chartData}
         layerKeys={layerKeys}
         layerFill={layerFill}
-        valueAxisLabel="CO₂eq / min"
+        valueAxisLabel="tCO₂eq"
         markerUpdateHandler={noop}
         markerHideHandler={noop}
         datetimes={datetimes}
--- a/web/src/features/charts/hooks/useBreakdownChartData.ts
+++ b/web/src/features/charts/hooks/useBreakdownChartData.ts
@@ -161,7 +161,7 @@ function getGenerationValue(
 }
 
 interface ValuesInfo {
-  valueAxisLabel: string; // For example, GW or tCO₂eq/min
+  valueAxisLabel: string; // For example, GW or tCO₂eq
   valueFactor: number;
 }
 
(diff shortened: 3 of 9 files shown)


VIKTORVAV99 (commented):
Can we add a enum for these? Seems like we are using CO2 with both the normal and small 2 which doesn't really make sense when it relates to the same unit.

tonypls:
> Can we add a enum for these? Seems like we are using CO2 with both the normal and small 2 which doesn't really make sense when it relates to the same unit.

I think all the big 2's are pointing to translations :)

VIKTORVAV99:
> > Can we add a enum for these? Seems like we are using CO2 with both the normal and small 2 which doesn't really make sense when it relates to the same unit.
> 
> I think all the big 2's are pointing to translations :)

Oh yeah that seems to be the case, would you mind deleting all the old translation strings then (I assume these translations can't be reused?)?

tonypls:
Will remove the old translations and make the test pass 👍

web/public/locales/en.json
```diff
@@ -208,6 +208,7 @@
   "electricityImportedFrom": "<b>%1$s %%</b> of electricity available in %4$s <b>%2$s</b> is imported from %5$s <b>%3$s</b>",
   "emissionsImportedFrom": "<b>%1$s %%</b> of emissions from electricity available in %4$s <b>%2$s</b> are imported from %5$s <b>%3$s</b>",
   "ofCO2eqPerMinute": "of CO₂eq per minute",
```
madsnedergaard, line 210: Should we delete all of these?

web/src/features/charts/EmissionChart.tsx
```diff
@@ -32,7 +32,7 @@ function EmissionChart({ timeAverage, datetimes }: EmissionChartProps) {
         data={chartData}
         layerKeys={layerKeys}
         layerFill={layerFill}
-        valueAxisLabel="CO₂eq / min"
+        valueAxisLabel="tCO₂eq"
```
madsnedergaard, line 35: I think this should be without `t` prefix, as the unit (t/kg/Pt, etc.) is now shown on the axis tick labels instead.
```suggestion
        valueAxisLabel="CO₂eq"
```

web/src/features/charts/hooks/useBreakdownChartData.ts
```diff
@@ -174,7 +174,7 @@ function getValuesInfo(
   );
 
   const format = scalePower(maxTotalValue);
-  const valueAxisLabel = displayByEmissions ? 'CO₂eq / min' : format.unit;
+  const valueAxisLabel = displayByEmissions ? 'tCO₂eq' : format.unit;
```
madsnedergaard, line 177: Same comment here regarding excluding `t`

web/src/features/charts/hooks/useNetExchangeChartData.ts
```diff
@@ -82,7 +82,7 @@ function getValuesInfo(
 
   const { unit, formattingFactor } = displayByEmissions
     ? {
-        unit: 'tCO₂eq/min',
+        unit: 'tCO₂eq',
```
madsnedergaard, line 85: Same same

madsnedergaard (commented):
Nice, a few small comments

web/public/locales/en.json
tonypls: Yes! I've removed all of them / translated some to the equivalent of "of CO2eq" in the respective language. @VIKTORVAV99 Could you confirm the Swedish is correct 🇸🇪 ?

madsnedergaard (approved):
LGTM 🥳

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5835, opened 2023-09-06
PH-updated parser to handle battery storage

mathilde-daugy:
## Issue

Power plant mapping was updated with battery storage facilities but the current parser was not built to capture that

## Description

update the parser to collect battery storage
remove the pivot table part because it is not very efficient method

### Preview

image

### Double check

### DIFF
--- a/parsers/PH.py
+++ b/parsers/PH.py
@@ -7,15 +7,13 @@ import pandas as pd
 import requests
 from requests import Response, Session
 
+from electricitymap.contrib.config.constants import PRODUCTION_MODES
 from electricitymap.contrib.lib.models.event_lists import (
+    EventSourceType,
     ExchangeList,
     ProductionBreakdownList,
 )
-from electricitymap.contrib.lib.models.events import (
-    EventSourceType,
-    ProductionMix,
-    StorageMix,
-)
+from electricitymap.contrib.lib.models.events import ProductionMix, StorageMix
 from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
@@ -453,6 +451,10 @@ EXCHANGE_KEY_MAPPING = {
 }
 KIND_TO_URL = {"production": DIPC_URL, "exchange": RTDHS_URL}
 KIND_TO_POST_ID = {"production": "5754", "exchange": "5770"}
+PRODUCTION_MIX_TO_MODE = {
+    "production": PRODUCTION_MODES,
+    "storage": ["hydro_storage", "battery"],
+}
 
 
 class MarketReportsItem(NamedTuple):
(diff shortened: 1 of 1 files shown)


parsers/PH.py
```diff
@@ -681,25 +682,39 @@ def fetch_production(
...
+    for dt in df.index.unique():
+        df_dt = df.loc[dt]
         production_mix = ProductionMix()
-        for mode in [m for m in row.index if "production." in m]:
-            production_mix.add_value(mode.replace("production.", ""), row[mode])
         storage_mix = StorageMix()
-        if "storage.hydro" in row.index:
-            storage_mix.add_value("hydro", row["storage.hydro"])
+        for idx, row in df_dt.iterrows():
+            if row["production_mix"] == "production":
```
VIKTORVAV99, line 695: Wouldn't `if row["mode"] in PRODUCTION_MODES` be easier and we could skip the whole "production_mix" column? (same for storage but with a storage list)

VIKTORVAV99 (commented):
The battery storage values seems to only be positive (storage) but no negative values (discharge). Is storage and discharge reported on units or is there something we are missing in here?

Also one suggestion, but things seems to be working as expected so feel free to ignore.

parsers/PH.py
```diff
@@ -611,15 +616,11 @@ def aggregate_per_datetime_mode(df: pd.DataFrame) -> pd.DataFrame:
...
+def add_production_mix_column(df: pd.DataFrame) -> pd.DataFrame:
     df = df.copy()
-    df = df.pivot(index="datetime", columns="mode")
-    # Flatten columns and make them "production.{mode}"
-    df.columns = df.columns.to_series().str.join(".")
-    # Handle storage
-    if "production.hydro_storage" in df.columns:
-        df = df.rename(columns={"production.hydro_storage": "storage.hydro"})
-        df["storage.hydro"] *= -1
+    df["production_mix"] = None
```
pierresegonne, line 621: I don't think the name of the column is self-descriptive.

If I understand correctly, we end up with:

```
idx, mode, value, production_mix 
dt, coal, 1239, "production"
```

If that's the case, then I would second Viktor's comment: do we even need that, and can't we directly use the mode column for that info?

pierresegonne (changes requested):

pierresegonne:
> The battery storage values seems to only be positive (storage) but no negative values (discharge). Is storage and discharge reported on units or is there something we are missing in here?

I think it's because we filter items with positive contributions to the grid. We could make an exception for resources that are matched to a storage system

VIKTORVAV99:
> > The battery storage values seems to only be positive (storage) but no negative values (discharge). Is storage and discharge reported on units or is there something we are missing in here?
> 
> I think it's because we filter items with positive contributions to the grid. We could make an exception for resources that are matched to a storage system

Yeah I think we have to do that, otherwise we would loose data that we might not be able to refetch on a later date which might influence the carbon intensity if we re-work how storage emission factors are attributed (say only during charging/storage instead of a yearly average as an example).

parsers/PH.py
```diff
@@ -681,25 +682,39 @@ def fetch_production(
...
-            storage_mix.add_value("hydro", row["storage.hydro"])
+        for idx, row in df_dt.iterrows():
+            if row["production_mix"] == "production":
+                production_mix.add_value(row["mode"], row["production"])
+            elif row["production_mix"] == "storage":
+                if row["mode"] == "hydro_storage":
+                    mode = "hydro"
+                else:
+                    mode = row["mode"]
+                storage_mix.add_value(mode, row["production"])
```
VIKTORVAV99, line 702:
```suggestion
                storage_mix.add_value(mode, -row["production"])
```
This gets us every time...
can we add something in the long term planning to fix this?

VIKTORVAV99:
> > The battery storage values seems to only be positive (storage) but no negative values (discharge). Is storage and discharge reported on units or is there something we are missing in here?
> 
> I think it's because we filter items with positive contributions to the grid. We could make an exception for resources that are matched to a storage system

Do we even need filter this? The new parser classes would correct these to 0 when adding them to production, or would that include a lot more items that are consumption only?

pierresegonne:
> Do we even need filter this? The new parser classes would correct these to 0 when adding them to production, or would that include a lot more items that are consumption only?

I would rather not rely on an downstream process to correct these values when we know that they don't even represent generation units. The data contains data both for consumer nodes and generators :)

PR #5836, opened 2023-09-06
Round values in add_value() method

VIKTORVAV99:
## Issue
Due to how python works we accidentally introduce floating point errors when adding values, this is a random behaviour.

## Description

This PR introduces rounding to the add value function to eliminate these floating point errors. 6 digits was chosen as the number of decimals as it gives us a precision of 1 watt, it is hard to imagine our source values would be more accurate than this considering the scales we are looking at but this can be increased if deemed necessary.

### Double check


mathilde-daugy:
i don't see any issue with this, thanks Viktor. I'll approve but please resolve the tests and prettier issue :)

mathilde-daugy (approved):

VIKTORVAV99:
I'll update the tests and double check all formatting then merge this.

PR #5838, opened 2023-09-07
Upgrade AR production parser to the new Event datastructure

unitrium:
## Issue

As part of the upgrade to the new Event datastructure.
This update required a bit of extra work as the production is split between conventional and renewable. When merging together those two productions sets we need to make sure they have matching timestamps. Otherwise we will create parser events with only conventional or renewable. At the moment our infrastructure would consider that the missing modes are set to 0 and this would therefore create issues with the calculations of the average hourly production.

## Description

- Upgrade the AR `fetch_production` function to use the new ProductionBreakdownList.
- Add testing
- Add an option in `merge_production_breakdowns` to only have merge on matching timestamps.

### Preview

Adding some unit tests to showcase this.

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -227,6 +230,8 @@ def merge_production_breakdowns(
 
         df = df.drop(columns=["source", "sourceType", "zoneKey"])
         df = df.groupby(level=0, dropna=False)["data"].apply(list)
+        if matching_timestamps_only:
```
unitrium, line 233: Here we make sure that the grouped events for each row match the number of list we want to merge.
This eliminates rows were a datetime is missing in the other list.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -204,18 +204,21 @@ def append(
...
         ungrouped_production_breakdowns: List["ProductionBreakdownList"],
         logger: Logger,
+        matching_timestamps_only: bool = False,
     ) -> "ProductionBreakdownList":
         """
         Given multiple parser outputs, sum the production and storage
         of corresponding datetimes to create a unique production breakdown list.
         Sources will be aggregated in a comma-separated string. Ex: "entsoe, eia".
         There should be only one zone in the list of production breakdowns.
+        Matching timestamps only will only keep the timestamps where all the production breakdowns have data.
```
VIKTORVAV99, line 214: Nice, this means we might be able to move forward with splitting up AR after this.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -227,6 +230,8 @@ def merge_production_breakdowns(
 
         df = df.drop(columns=["source", "sourceType", "zoneKey"])
         df = df.groupby(level=0, dropna=False)["data"].apply(list)
+        if matching_timestamps_only:
+            df = df[df.apply(lambda x: len(x) == len_ungrouped_production_breakdowns)]
```
VIKTORVAV99, line 234: Should we also try and log the datapoints that where discarded or just the amount? Could help with debugging later on?

VIKTORVAV99 (approved):
No issues as far as I can tell it's very nice to get rid of much of the duplicate functionality in the parser.

Looks like my previous PR messed a little with your tests but I'm approving anyway.

Nice work!

unitrium:
@FelixDQ here's another example where it shows that having the productions modes being independent would make sense.
Basically here it can happen that the timestamp from the conventional production doesn't match the timestamp of the renewable production. We manually filter out events where they don't match to avoid creating disjoint events because otherwise that would mess up the average calculations in the flowtracing pipeline as it interprets a missing mode as a 0.

If we had those events separate it would not make a difference to have disjoint events :)

electricitymap/contrib/lib/models/event_lists.py
unitrium: Yes I think this would simplify things quite a lot!

electricitymap/contrib/lib/models/event_lists.py
unitrium: Yes I think that could be a useful information, I will add it

PR #5839, opened 2023-09-07
take into account empty breakdowns

unitrium:
## Issue

I actually noticed that in the case of empty event lists all the datapoints would be discarded. In the case of empty list we should not discard everything in the other.

## Description

Merging on matching timestamps takes into account fully empty event lists.

### Double check

### DIFF
--- a/electricitymap/contrib/lib/models/event_lists.py
+++ b/electricitymap/contrib/lib/models/event_lists.py
@@ -218,13 +218,14 @@ class ProductionBreakdownList(AggregatableEventList):
             ungrouped_production_breakdowns, logger
         ):
             return production_breakdowns
-        len_ungrouped_production_breakdowns = len(ungrouped_production_breakdowns)
+        filtered_ungrouped_breakdowns = [
+            ungrouped_breakdowns.dataframe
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (commented):
Wouldn't this risk overwriting valid data with invalid data if say the longer on two lists still have data and the shorter of the lists miss data at one point.

Say the long list has data for hour 1 and 2 normally and the short one has for hour 2.

Normally it would only return hour 2 then.

But if the short list was empty we would return hour 1 and 2 with this change witch would overwrite hour 1 with potentially erroneous data.

Or am I missing something?

VIKTORVAV99:
> No indeed that would the intended behavior. I have seen that in the US, since we are getting each mode separately we are getting empty lists so it would be to avoid having a missing mode erase everything.

Wouldn't it be better to do a hard exit if that is the case or return an empty list and let the estimations handle it instead of only having partial data that could potentially overwrite complete past data?

unitrium:
Hum indeed, you're probably right, since the behaviour has to be explicitly enabled, it might be better to not do this.

PR #5840, opened 2023-09-07
[total energy] Adds new introduction modal

madsnedergaard:
## Description

Adds a new modal that will explain the change. It should only show until November, assuming that any user who haven't seen it after that time is most likely a new user.

I also made sure it only pops up after onboarding has been seen, to avoid confusing new users :)

### Preview

Screenshot 2023-09-07 at 14 16 21


tonypls:
Maybe not enough contrast here on darkmode:
image

web/src/features/modals/TotalEnergyIntroModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+export default function TotalEnergyIntroModal() {
+  const { __ } = useTranslation();
+  const [hasTotalEnergyIntroBeenSeen, setHasTotalEnergyIntroBeenSeen] = useAtom(
+    hasTotalEnergyIntroBeenSeenAtom
+  );
+  const [hasOnboardingBeenSeen] = useAtom(hasOnboardingBeenSeenAtom);
+  const [searchParameters] = useSearchParams();
+  const skipOnboarding = searchParameters.get('skip-onboarding') === 'true';
+  // Stop showing this modal roughly a month after the feature is released
+  const isExpired = new Date() > new Date('2023-11-01');
```
VIKTORVAV99, line 66: I suggest we make this longer on the apps at least (but only show it once).

One(ish) month is not that long and even if we create a release now it will be less with publishing delays, distribution delays and then for however long it takes for people to update (or the device to auto update).

madsnedergaard: Sure, we can always remove it again later if we get feedback about it :)

PR #5854, opened 2023-09-11
Update capacity to use power and production to use energy

tonypls:
## Issue
PWh isn't a common measure and capacity is measured by power

## Description
This PR updates our formatting to show a maximum of TW / TWh and will use power values for capacity

### Preview
image
### Double check

### DIFF
--- a/web/src/utils/formatting.ts
+++ b/web/src/utils/formatting.ts
@@ -29,26 +29,24 @@ export const formatCo2 = function (grams: number, valueToMatch?: number): string
     return '?';
   }
 
-  // Assume gCO₂ input
-  const value = grams;
-
   // Ensure both numbers are at the same scale
-  const checkAgainst = valueToMatch ?? value;
+  const checkAgainst = valueToMatch ?? grams;
 
+  //Values less than 1Mt
   if (Math.round(checkAgainst) < 1e9) {
-    let decimals = value < 1 ? 2 : 1;
+    let decimals = grams < 1 ? 2 : 1;
     // Remove decimals for large values
-    if (value > 1_000_000) {
+    if (grams > 1_000_000) {
       decimals = 2;
     }
     if (checkAgainst < 1e6) {
-      return addSpaceBetweenNumberAndUnit(`${d3.format(`,.${decimals}~s`)(value)}g`);
+      return addSpaceBetweenNumberAndUnit(`${d3.format(`,.${decimals}~s`)(grams)}g`);
     }
 
-    return addSpaceBetweenNumberAndUnit(`${d3.format(`,.${decimals}~r`)(value / 1e6)}t`);
+    return addSpaceBetweenNumberAndUnit(`${d3.format(`,.${decimals}~r`)(grams / 1e6)}t`);
   }
-  // megato(ns or above as a default
-  return addSpaceBetweenNumberAndUnit(`${d3.format(',.2~s')(value / 1e6)}t`);
+  // tonnes or above with significant figures as a default
+  return addSpaceBetweenNumberAndUnit(`${d3.format(',.2~s')(grams / 1e6)}t`);
 };
 
 const scalePower = function (maxPower: number | undefined) {
--- a/web/src/utils/formatting.ts
+++ b/web/src/utils/formatting.ts
@@ -23,40 +23,32 @@ export const formatEnergy = function (
   return addSpaceBetweenNumberAndUnit(power);
 };
 
-export const formatCo2 = function (grams: number, valueToMatch?: number) {
-  if (grams == undefined || Number.isNaN(grams)) {
+export const formatCo2 = function (grams: number, valueToMatch?: number): string {
+  // Validate input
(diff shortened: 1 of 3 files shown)


VIKTORVAV99:
Pettawatt is pretty common when it comes to lasers 😅

I'm not sure it would be more helpful with large TW numbers but then again I'm not exactly our normal user in this scenario...

tonypls:
> Pettawatt is pretty common when it comes to lasers 😅
> 
> I'm not sure it would be more helpful with large TW numbers but then again I'm not exactly our normal user in this scenario...
> 
> It's capacity over time (not hourly) also measured in kWh , MWh etc to account for required maintenance and such (we don't take this into account right now but still).

I think because it's only really US and most the time other sources use TW it's the best we can use here for now. 
Here's an example of our world in data using high numbers to stick with TWh
image


The capacity was feedback from the team that now more about energy than me 😅 A good point too, I think we can add energy capacity in the future when we need it 👍

VIKTORVAV99:
> > Pettawatt is pretty common when it comes to lasers 😅
> > 
> > I'm not sure it would be more helpful with large TW numbers but then again I'm not exactly our normal user in this scenario...
> > 
> > It's capacity over time (not hourly) also measured in kWh , MWh etc to account for required maintenance and such (we don't take this into account right now but still).
> 
> I think because it's only really US and most the time other sources use TW it's the best we can use here for now. 
> Here's an example of our world in data using high numbers to stick with TWh
> image
> 
> 
> The capacity was feedback from the team that know more about energy than me 😅 A good point too, I think we can add energy capacity in the future when we need it 👍 

I'm expecting that pettawatt will be more common once we do larger aggregates (like the whole EU or similar) but I suppose we can still show it in TW/TWh but we might have to adapt some things to handle the bigger numbers as the current UI has not been designed for them.

Capacity is only really a concern once we aggregate daily, monthly and yearly capacity which probably come after capacity parsing so a few steps to do first...

web/src/utils/formatting.ts
```diff
@@ -11,18 +11,29 @@ function addSpaceBetweenNumberAndUnit(inputString: string) {
 
 export const formatEnergy = function (
```
VIKTORVAV99, line 12: perhaps we should rename this now that it don't format just energy?

web/src/utils/formatting.ts
```diff
@@ -11,18 +11,29 @@ function addSpaceBetweenNumberAndUnit(inputString: string) {
...
+  isCapacity = false
 ) {
   // Assume MW input
   if (d == undefined || Number.isNaN(d)) {
     return d;
   }
   const significantFigures = d.toString().length > 1 ? numberDigits : 1;
-
-  const power = `${d3.format(`.${significantFigures}s`)(d * 1e6)}Wh`;
+  const unit = isCapacity ? 'W' : 'Wh';
```
VIKTORVAV99, line 22: Since we are doing this we might as well do it for the hourly production values as well so they match the capacity values?

web/src/utils/formatting.ts
```diff
@@ -11,18 +11,29 @@ function addSpaceBetweenNumberAndUnit(inputString: string) {
 
 export const formatEnergy = function (
   d: number,
-  numberDigits: number = DEFAULT_NUM_DIGITS
+  numberDigits: number = DEFAULT_NUM_DIGITS,
+  isCapacity = false
```
VIKTORVAV99, line 15: Maybe change this to hourly?

VIKTORVAV99 (commented):
Just some notes.

web/src/utils/formatting.ts
```diff
@@ -11,18 +11,29 @@ function addSpaceBetweenNumberAndUnit(inputString: string) {
...
   return addSpaceBetweenNumberAndUnit(power);
 };
 
+export const formatPower = function (
+  d: number,
+  numberDigits: number = DEFAULT_NUM_DIGITS
+) {
+  return formatEnergy(d, numberDigits, true);
+};
+
```
madsnedergaard, line 36: Just had a thought: how about flipping the two functions around, so `formatPower` is the default that does the d3 logic, and then `formatEnergy` just adds "h" to the output of the `formatPower`?

```js
export const formatPower = function (
  d: number,
  numberDigits: number = DEFAULT_NUM_DIGITS
) {
  return `${formatEnergy(d, numberDigits)h`;
};
```

web/src/utils/formatting.ts
tonypls: I think the idea is that capacity is an instantaneous measure and our hourly values are hourly so the mismatch is intentional. It could be confusing that when talking production values we provide different base units depending on time aggregation

VIKTORVAV99: Agreed but MW and MWh has the same value since we are showing things in hourly resolution anyhow.

tonypls: Seems reasonable 👍

madsnedergaard (approved):
Beautiful, thanks! 💜

PR #5855, opened 2023-09-11
fix IL production parser and add total production

VIKTORVAV99:
## Issue

[ELE-2774](https://linear.app/electricitymaps/issue/ELE-2774/priority-zones-down-il-za-in)

## Description

Get the production parser for IL up and running again. Also adds a totalProduction function but that is so far unused (I and Robin have talked about switching to that from unknown when we only have the total production).

### Double check

### DIFF
--- a/parsers/IL.py
+++ b/parsers/IL.py
@@ -19,7 +19,16 @@ from typing import Optional
 
 import arrow
 from bs4 import BeautifulSoup
-from requests import Session, get
+from requests import Response, Session
+
+from electricitymap.contrib.lib.models.event_lists import (
+    ProductionBreakdownList,
+    TotalProductionList,
+)
+from electricitymap.contrib.lib.models.events import ProductionMix
+from electricitymap.contrib.lib.types import ZoneKey
+
+URL = "https://www.noga-iso.co.il/Umbraco/Api/Documents/GetElectricalData"
 
 IEC_URL = "www.iec.co.il"
 IEC_PRODUCTION = (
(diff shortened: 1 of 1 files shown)


madsnedergaard (approved):
LGTM 🥳

q--:
The pull request https://github.com/electricitymaps/electricitymaps-contrib/pull/5424 with another fix for this is still open

q--:
> Also adds a totalProduction function but that is so far unused (I and Robin have talked about switching to that from unknown when we only have the total production).

Have you considered adding support for dividing "unknown" production between "regular unknown" and "renewable unknown", or having multiple types of "unknown" that we know consist of certain production types?

We've previously encountered this with Bolivia before they split out wind and solar (before that, they just had "thermal", "hydro" and "renewable" IIRC).

Now with Israel there is again some useful data we aren't retrieving.

Back with Bolivia the concern was about the undesirability of adding another category to the production panel, but if the parser output is being used by an estimation model that should no longer be a concern.

VIKTORVAV99:
> > Also adds a totalProduction function but that is so far unused (I and Robin have talked about switching to that from unknown when we only have the total production).
> 
> Have you considered adding support for dividing "unknown" production between "regular unknown" and "renewable unknown", or having multiple types of "unknown" that we know consist of certain production types?
> 
> (We've previously encountered data with "unknown" and "renewable unknown"  with Bolivia, before they split out wind and solar; before that, they just had "thermal", "hydro" and "renewable" IIRC.)
> 
> Now with Israel there is again some useful data we aren't retrieving.
> 
> Back then with Bolivia the concern was about the undesirability of adding another category to the production panel, but if the parser output is being used by an estimation model that should no longer be a concern.

While I'd like to see more granular source data it's not as straight forward as it seems. I don't know all the hurdles but I know there are many to adding a new "mode" everything from data validation to database tables would have to be updated. But I agree it could be useful, especially for the internal stuff which is why I suggested this a while back https://github.com/electricitymaps/electricitymaps-contrib/discussions/4219#discussioncomment-3110126

VIKTORVAV99:
> The pull request https://github.com/electricitymaps/electricitymaps-contrib/pull/5424 with another fix for this is still open

The plan was to get the production fix merged quickly and then get the other fix merged. Hopefully I'll have some time to clean up that other PR later today.

PR #5859, opened 2023-09-12
Refactor parser names

unitrium:
## Issue

As discussed with the team, we would like to refactor parser names. The reasoning is that we used to use the zone key name for some parsers while we were using the source name for others.
Some of those parsers would be used by different zones so naming the parser with the zone key wasn't the best choice.
Ideally we would like 1 parser = 1 source, which can then be used by multiple zones.

## Description

- Rename MX parser to CENACE.
- Rename AR parser to CAMMESA.

### Double check


VIKTORVAV99 (approved):
There was some leftover references in some logging messages but I pushed the changes so now this LGTM.

PR #5860, opened 2023-09-12
Use power units for hourly and energy units for aggregated views

tonypls:
## Issue
Capacity is expressed in power (W) and power production is expressed in energy (Wh), they are both shown in the hourly view

## Description
This PR displayed only power values in the hourly view and energy values in the time aggregated views

### Double check


web/src/utils/formatting.ts
```diff
@@ -72,13 +73,21 @@ const scalePower = function (maxPower: number | undefined) {
...
+        [1, PowerUnits.MEGAWATTS],
+        [1e-3, PowerUnits.KILOWATTS],
+      ]
+    : [
+        [1e9, EnergyUnits.PETAWATT_HOURS],
+        [1e6, EnergyUnits.TERAWATT_HOURS],
+        [1e3, EnergyUnits.GIGAWATT_HOURS],
+        [1, EnergyUnits.MEGAWATT_HOURS],
+        [1e-3, EnergyUnits.KILOWATT_HOURS],
+      ];
```
madsnedergaard, line 90: In future it would be nice to do something like this, but don't think it's worth it now :)
```suggestion
const unit = isPower ? Units.Power : Units.Energy;
const thresholds: [number, EnergyUnits | PowerUnits][] = [
      [1e9, unit.PETA],
      [1e6, unit.TERA],
      [1e3, unit.GIGA],
      [1, unit.MEGA],
      [1e-3, unit.KILO],
]
```

madsnedergaard (approved):
LGTM

PR #5861, opened 2023-09-12
Automatically sort imports 

tonypls:
## Issue
Manually sorted imports 📚 

We've started doing more import sorting in PRs, this is great but it can add extra noise for reviewers so I think it's nice to agree to a sorting pattern and apply it across the repo in one fell swoop.

## Description
This PR adds some sorting config to the eslintrc.

Sorting order is up for discussion so feel free to provide suggestions.

This is the most logic sorting pattern I could find online (it's not perfect, see node imports and non-relative internal imports and bundled with external imports. But consistent is better than perfect 🙏 )

### Double check

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -20,6 +20,25 @@
     "prettier"
   ],
   "rules": {
+    "import/order": [
+      "error",
+      {
+        "groups": ["external", "internal"],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          }
+        ],
+        "pathGroupsExcludedImportTypes": ["react"],
+        "newlines-between": "always",
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true
+        }
+      }
+    ],
     // TODO: TEMPORARILY DISABLED DURING INITIAL DEVELOPMENT
     "@typescript-eslint/no-empty-interface": "off",
     "@typescript-eslint/no-empty-function": "off",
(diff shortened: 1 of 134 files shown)


web/.eslintrc.json
```diff
@@ -20,6 +20,25 @@
...
+      "error",
+      {
+        "groups": ["external", "internal"],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          }
+        ],
```
VIKTORVAV99, line 33: Should we add a group for "node*" imports as well? Would be nice to group all the internal stuff seperate from our code imports.

madsnedergaard:
Love the idea! :)
Did you look into using https://github.com/lydell/eslint-plugin-simple-import-sort instead?
Here's a list of how they differ: https://github.com/lydell/eslint-plugin-simple-import-sort#how-is-this-rule-different-from-importorder

web/.eslintrc.json
```diff
@@ -20,6 +20,25 @@
     "prettier"
   ],
   "rules": {
+    "import/order": [
+      "error",
+      {
+        "groups": ["external", "internal"],
```
madsnedergaard, line 26: If we stick with `import/order` (see other comment), I think we should use the full grouping they suggest:
`["builtin", "external", "internal", "parent", "sibling", "index", "object", "type"]`

tonypls:
Closing to re-do with https://github.com/lydell/eslint-plugin-simple-import-sort

PR #5862, opened 2023-09-12
Shows a message if we dont have enough data for the charts

madsnedergaard:
## Issue

Screenshot 2023-09-12 at 14 50 01

Screenshot 2023-09-12 at 14 50 07

See https://staging.electricitymaps.com/zone/TH on yearly, and switching between time scales makes the app crash.

## Description

Shows a message instead of crashing the app if there is not enough data to show charts :)

### Preview

Screenshot 2023-09-12 at 14 44 56


tonypls (approved):
Nice 🙌  The changes to show missing values in the charts will also add more safety to this

web/src/features/charts/BreakdownChart.tsx
```diff
@@ -38,6 +39,17 @@ function BreakdownChart({
 
   const titleDisplayMode = displayByEmissions ? 'emissions' : 'electricity';
   const titleMixMode = mixMode === Mode.CONSUMPTION ? 'origin' : 'production';
+
+  const hasEnoughDataToDisplay = datetimes?.length > 2;
```
VIKTORVAV99, line 43: Should this not be:
```suggestion
  const hasEnoughDataToDisplay = datetimes?.length >= 2;
```
Or do we only want to show it when we have more than two datapoints?

web/src/features/charts/NotEnoughDataMessage.tsx
```diff
@@ -0,0 +1,12 @@
+import { ChartTitle } from './ChartTitle';
+
+export function NotEnoughDataMessage({ translationKey }: { translationKey: string }) {
+  return (
+    <div className="w-full">
+      <ChartTitle translationKey={translationKey} hasLink={false} />
+      <div className="my-2 rounded bg-gray-200 py-4 text-center text-sm dark:bg-gray-800">
+        <p>Not enough data available to display chart</p>
```
VIKTORVAV99, line 8: It should be possible to translate this string.

VIKTORVAV99 (changes requested):
Just one thing about translating one of the string that needs to be addressed

web/src/features/charts/BreakdownChart.tsx
madsnedergaard: We currently expect these things to be available, so we kinda do need more than two:
```js
  const startTime = datetimes?.at(0);
  const lastTime = datetimes?.at(-1);
  const interval = datetimes?.at(-2);
```

I briefly looked into making it work with just one or two datapoints, but that was a lot more work that I didn't want to get started on now 😅

web/src/features/charts/NotEnoughDataMessage.tsx
madsnedergaard: You are absolutely right, I have made it possible and added translations for the most common languages :)

VIKTORVAV99 (approved):
Thanks for fixing the translations!
(and sneaking in some extra Danish ones... 😁)

PR #5864, opened 2023-09-12
add a value factor for axis ticks of emissions breakdown chart

tonypls:
## Issue
CO2 axis scales don't match
image

## Description
This PR scales the values of the emission breakdown chart by an order of magnitude.

The CO2 values are calculated using CO2intensity (g/kWh) * energy of each mode (MW), energy is provided with a base unit of MW so we need to multiply by 1000 to get the correct value

### Preview
image

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #5868, opened 2023-09-13
Refactor KR parser (bis)

### LINKED ISSUE #5845: South Korea data doesn't seem right
## Description
I've noticed that South Korean data does not match the source, nor the results of running the parser locally. There aren't errors in the logs neither. The app shows that some data is being estimated:
Screenshot_20230910_213433
Screenshot_20230910_213642
Screenshot_20230910_213851
It is especially notable in the fact that nuclear should be a stable line but it is load following in the app. The parser seems to be getting the data correctly (~20GW nuclear). It seems the problem is in the backend, possibly in the estimation model. It seems that the estimation is needed to transform "renewables" in the graph to the different production types of the app, but it is not functioning correctly as the graphs don't match.

pierresegonne:
## Issue

We want to integrate parser data for `KR` (instead of current estimates).
Is a prerequisite to close #5845 

## Description

To understand better the parser and how robust it is I went ahead to refactor it. 
@unitrium did the same simultaneously 😅. I took some changes from his PR #5867, most notably the tests

### Preview

Tested with:
- `poetry run python test_parser.py KR production`
- `poetry run python test_parser.py KR production --target_datetime '2022-04-09 00:00:00+00:00'`
- `poetry run python test_parser.py KR consumption`
- `poetry run python test_parser.py KR price`
- `poetry run python test_parser.py KR price --target_datetime '2023-09-10 00:00:00+00:00'`
- `poetry run python test_parser.py KR price --target_datetime '2023-04-09 00:00:00+00:00'` -> fails as expected

### Double check


pierresegonne:
For now `poetry run python test_parser.py KR production --target_datetime '2022-04-09 00:00:00+00:00'` fails with:

```requests.exceptions.SSLError: HTTPSConnectionPool(host='new.kpx.or.kr', port=443): Max retries exceeded with url: /powerSource.es?mid=a10606030000&device=chart (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1131)')))```

Currently looking into it

File: parsers/KPX.py [hunk 1]
```diff
@@ -164,29 +123,62 @@ def fetch_price(
...
+                source=KR_SOURCE,
+                price=price_value,
+                currency=KR_CURRENCY,
+            )
+
+    return price_list.to_list()
 
-            all_data.append(data)
 
+def _parse_long_term_prod_data(html: str) -> Dict[datetime, ProductionMix]:
```
unitrium, line 137:
```suggestion
def _parse_long_term_prod_data(html: str, zone_key: ZoneKey) -> ProductionBreakdownList:
```
You could directly create the list here instead of creating an adhoc dict with timestamps as indexes.

parsers/KPX.py
```diff
@@ -196,115 +188,132 @@ def get_long_term_prod_data(
...
-    table_rows = soup.find_all("tr")[1:]
-
-    for row in table_rows:
-
-        sanitized_date = [value[:-1] for value in row.find_all("td")[0].text.split(" ")]
-        curr_prod_datetime_string = (
-            "-".join(sanitized_date[:3]) + "T" + ":".join(sanitized_date[3:]) + ":00"
+    production_list = ProductionBreakdownList(logger)
+    events = _parse_long_term_prod_data(res.text)
+    for dt, event in events.items():
```
unitrium, line 198: As explained above here you could directly return the ProductionBreakdownList.

parsers/KPX.py
```diff
@@ -196,115 +188,132 @@ def get_long_term_prod_data(
...
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=dt,
+            source=KR_SOURCE,
+            production=event,
         )
-        arw_datetime = arrow.get(
-            curr_prod_datetime_string, "YYYY-MM-DDTHH:mm:ss", tzinfo=TIMEZONE
-        ).datetime
+    return production_list.to_list()
```
unitrium, line 205:
```suggestion
    return production_list
```
Usually we keep the `.to_list()` for the very last step in the `fetch_production` function so that when we get at a later stage and those functions can directly return the ProductionBreakdownList objects we have one single place to remove it from.

parsers/KPX.py
```diff
@@ -196,115 +188,132 @@ def get_long_term_prod_data(
...
+        )
+        production_mix.add_value(
+            "nuclear",
+            round(float(item["nuclearPower"]), 5),
+        )
+        production_mix.add_value("oil", round(float(item["oil"]), 5))
+        production_mix.add_value("unknown", round(float(item["newRenewable"]), 5))
+        storage_mix = StorageMix()
+        storage_mix.add_value("hydro", -round(float(item["raisingWater"]), 5))
+        timed_data[date] = {"production": production_mix, "storage": storage_mix}
```
unitrium, line 255: Same here, you can directly return the list, as in #5867

parsers/KPX.py
```diff
@@ -196,115 +188,132 @@ def get_long_term_prod_data(
...
-            data["production"]["gas"] = chart_data_values["gas"]
-            data["production"]["nuclear"] = chart_data_values["nuclear"]
-            data["production"]["oil"] = chart_data_values["oil"]
-            data["production"]["hydro"] = chart_data_values["hydro"]
-            data["production"]["unknown"] = chart_data_values["renewable"]
+    if target_datetime is None:
+        target_datetime = _now.datetime
 
-            all_data.append(data)
+    if target_datetime.date() == _now.date():
```
unitrium, line 300: I think it's a bit of a doggy comparison. If I understand correctly the thinking, if we don't pass a target datetime, we should use the real time source right ?
If that's the case you can eliminate this if by having  
```
if target_datetime is None:
  return get_real_time
 return get_granular_real_time_prod_data(
            zone_key=zone_key, session=session, logger=logger
        )

File: parsers/test/test_KPX.py [hunk 6]
```diff
@@ -0,0 +1,62 @@
...
+                    "datetime": element["datetime"].isoformat(),
+                    "production": element["production"],
+                    "storage": element["storage"],
+                    "source": element["source"],
+                    "zoneKey": element["zoneKey"],
+                    "sourceType": element["sourceType"].value,
+                }
+                for element in production
+            ]
+        )
```
unitrium, line 62: I would suggest to add a test for when it's not using realtime data :) But feel free to ignore if it's too much extra work.

unitrium (commented):
A few suggestions, let me know once I can review again :)

parsers/KPX.py
pierresegonne: I did this on purpose to avoid having to propagate the zone_key and logger arguments where it's not really needed, but I'm open for discussion here

parsers/KPX.py
pierresegonne: Same

parsers/KPX.py
pierresegonne: Good point

parsers/KPX.py
unitrium: I don't really see it as an issue to propagate the zone key and the logger, correctly encapsulating the function's behavior seems like the best way to do it.
Basically by using a custom wraping (datetime index dict of Mixes), the function making use of the parsed results needs some inside knowledge about how things are done in the parsing function. Mainly that the dict is indexed by datetimes and those refer to the time of the production/storage. It's generally not the best practice because ideally you should directly have all you need from the function's signature.

unitrium: Besides, in this case you end up looping twice over the same data, not a big deal because the dataset is small but we could save ourselves one loop.

parsers/KPX.py
pierresegonne: Good catch - will simplify this

pierresegonne:
For the SSL problem - this should not be a problem in production https://console.cloud.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22feeder-electricity%22%0Aresource.labels.location%20%3D%20%22europe-west1%22%0AjsonPayload.zone_key%20%3D%20%22KR%22%0A--Hide%20similar%20entries%0A-%2528jsonPayload.message%3D~%22This%20parser%20is%20not%20able%20to%20parse%20dates%20before%20%2528%5Cd%7B4%7D-%25280%5B1-9%5D%7C1%5B0-2%5D%2529-%25280%5B1-9%5D%7C%5B12%5D%5B0-9%5D%7C3%5B01%5D%2529%2529.%22%2529%0A--End%20of%20hide%20similar%20entries;cursorTimestamp=2023-09-13T12:52:20.210181Z;duration=PT1H?project=tmrow-152415

pierresegonne:
@unitrium All comments have been adressed + added a snapshot for historical

pierresegonne:
There is a discrepancy between historical and real-time data though. 
Historical does not seem to show `oil`, `hydro`, `hydro storage` nor `unknown`

parsers/KPX.py
```diff
@@ -232,79 +182,109 @@ def get_long_term_prod_data(
...
-    res_0 = r0.get(REAL_TIME_URL, verify=False)
-    chart_data = extract_chart_data(res_0.text)
+    for item in json_obj:
+        if item["regDate"] == "0":
+            break
 
-    return chart_data
+        dt = TIMEZONE.localize(datetime.strptime(item["regDate"], "%Y-%m-%d %H:%M"))
+
+        production_mix = ProductionMix()
```
unitrium, line 225:
```suggestion
        production_mix = ProductionMix(
           coal=float(item["coal"]) + float(item["localCoal"]),
           gas=float(item["gas"]),
           hydro=float(item["waterPower"]),
           nuclear=float(item["nuclearPower"]),
           oil=float(item["oil"]),
           unknown=float(item["newRenewable"])
        )
```
You can also initialize the values like this. I think it's a bit more compact but up to you :)
Also @VIKTORVAV99  recently added a default rounding so that all values across all parsers are rounded by the same method, so no need for this round anymore.

parsers/KPX.py
```diff
@@ -232,79 +182,109 @@ def get_long_term_prod_data(
...
+            "hydro",
+            round(float(item["waterPower"]), 5),
+        )
+        production_mix.add_value(
+            "nuclear",
+            round(float(item["nuclearPower"]), 5),
+        )
+        production_mix.add_value("oil", round(float(item["oil"]), 5))
+        production_mix.add_value("unknown", round(float(item["newRenewable"]), 5))
+        storage_mix = StorageMix()
```
unitrium, line 244:
```suggestion
        storage_mix = StorageMix(
         hydro = -float(item["raisingWater"]
        )
```

unitrium (approved):
Looks good! A final suggestion but feel free to move forward.

silimotion:
Hi, I've noticed that the current implementation lacks the solar data. Can it be added in this same PR?

pierresegonne:
> Hi, I've noticed that the current implementation lacks the solar data. Can it be added in this same PR?

I'll see what I can do :)

parsers/KPX.py
```diff
@@ -165,27 +165,15 @@ def parse_chart_prod_data(
...
-        production_mix.add_value(
-            "coal",
-            round(float(item["coal"]) + float(item["localCoal"]), 5),
+        production_mix = ProductionMix(
+            coal=float(item["coal"]) + float(item["localCoal"]),
+            gas=float(item["gas"]),
+            hydro=float(item["waterPower"]),
+            nuclear=float(item["nuclearPower"]),
+            oil=float(item["oil"]),
+            unknown=float(item["newRenewable"]),
```
VIKTORVAV99, line 174: Sorry this is a bit late but we should avoid assigning values like this directly as they don't get the same validation checks then. (Right now it will not correct for negative values if they are present or correct floating point errors.)

pierresegonne: Ah shit only saw your comment now - do you have time to make a quick PR to fix that? I'm planning on releasing a new version in the morning so we could have that in there

VIKTORVAV99: > Ah shit only saw your comment now - do you have time to make a quick PR to fix that? I'm planning on releasing a new version in the morning so we could have that in there

I actually just double checked and the negative values will be corrected. But some of the other validation we do will not. So feel free to merge this and I'll open a new PR during the day.

Issue #4338, opened 2022-07-11
Update Python union and optional types to use `|` syntax


VIKTORVAV99:
The current type syntax is not very user friendly, we should update all the types to use `type1 | type2` instead of `Union[type1, type2] ` and `type | null` instead of `Optional[type]` as soon as possible.

Another benefit of this is that we get less imports as the new type deceleration is native and don't rely on importing Union or Optional from typing. 

- [x] #4337
- [ ] Update types

shuuji3:
Have you tried a helpful tool called `pyupgrade`? It can automatically rewrite `Union` and `Option` to new type notation with `--py310-plus` flag: https://github.com/asottile/pyupgrade/#pep-604-typing-rewrites.

Also, I can perform other refactorings like the conversion from `str.format()` to f-string. It looks like it always works perfectly but it can reduce manual work and error for refactoring.

VIKTORVAV99:
> Have you tried a helpful tool called `pyupgrade`? It can automatically rewrite `Union` and `Option` to new type notation with `--py310-plus` flag: https://github.com/asottile/pyupgrade/#pep-604-typing-rewrites.
> 
> Also, I can perform other refactorings like the conversion from `str.format()` to f-string. It looks like it always works perfectly but it can reduce manual work and error for refactoring.

The problem isn't really the refactoring itself but I'm waiting for #4335 and #4337 before I do anything.

However that tool looks interesting so I'll keep it in mind when this is ready to be fixed.

PR #5882, opened 2023-09-13
automatic migration to python 3.10 syntax

VIKTORVAV99:
## Issue

We are using old syntax.
Closes: #4338

## Description

Use pyupgrade to automatically migrate to new python syntax supported in python 3.10.

### Double check


VIKTORVAV99:
Sorry for the massive change in here, the only thing I am really concerned with is this commit: https://github.com/electricitymaps/electricitymaps-contrib/pull/5882/commits/b535f59691d0da80c28df22590806b2a0996451d

unitrium:
Will review it this morning:)

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -1,7 +1,8 @@
 from abc import ABC, abstractmethod
+from collections.abc import Sequence
```
unitrium, line 2: I guess they are supporting both from `collections` and from `typing` for a few extra versions right ?

unitrium (approved):
OMG that looks so much better!
Thanks for running this, I guess I will do it as well on the monorepo later

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: Should be, but now that we can use the classes directly this (dict instead of Dict) this should be the way forward.

VIKTORVAV99:
> OMG that looks so much better! Thanks for running this, I guess I will do it as well on the monorepo later

Only thing that was a pain in the ass was removing all the unused imports...

PR #5888, opened 2023-09-14
Refactor DK-BHM parser

unitrium:
## Issue

Use the new event datastructure to refactor the new DK-BHM parser.

### Double check


parsers/BORNHOLM_POWERLAB.py
```diff
@@ -0,0 +1,86 @@
...
+    production = ProductionBreakdownList(logger)
+    production.append(
+        zoneKey=zone_key,
+        datetime=datetime.fromtimestamp(response["latest"], tz=TIMEZONE),
+        source=SOURCE,
+        production=ProductionMix(
+            biomass=response["sub"]["factory"],
+            solar=response["sub"]["solar_cells"],
+            wind=response["sub"]["wind_turbines"],
+        ),
```
VIKTORVAV99, line 49: granted these are only a single value and there are no math involved but I think we should still consider using add_value() for them.

parsers/BORNHOLM_POWERLAB.py
```diff
@@ -0,0 +1,86 @@
...
+    response = session.get(LATEST_DATA_URL).json()
+    exchange = ExchangeList(logger)
+    sorted_zone_keys = sorted([zone_key1, zone_key2])
+    flow = 1 if zone_key1 == sorted_zone_keys[0] else -1
+    exchange.append(
+        zoneKey=ZoneKey("->".join(sorted_zone_keys)),
+        datetime=datetime.fromtimestamp(response["latest"], tz=TIMEZONE),
+        source=SOURCE,
+        netFlow=flow * response["sub"]["seacable"],
+    )
```
VIKTORVAV99, line 75:
```suggestion
    direction = 1 if zone_key1 == sorted_zone_keys[0] else -1
    exchange.append(
        zoneKey=ZoneKey("->".join(sorted_zone_keys)),
        datetime=datetime.fromtimestamp(response["latest"], tz=TIMEZONE),
        source=SOURCE,
        netFlow=direction * response["sub"]["seacable"],
    )
```

parsers/BORNHOLM_POWERLAB.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+# The arrow library is used to handle datetimes
+from datetime import datetime
+from logging import Logger, getLogger
+from typing import List, Optional
+
+from pytz import timezone
```
VIKTORVAV99, line 7: We can use `zoneinfo` instead of pytz now (new python built in since python 3.9), it should have the same API/function as timezone if I remember correctly.

This is just a bonus.

VIKTORVAV99 (commented):
Some small things.

parsers/BORNHOLM_POWERLAB.py
```diff
@@ -0,0 +1,86 @@
...
+
+PRODUCTION_MAPPING = {
+    "wind": "wind_turbines",
+    "biomass": "factory",
+    "solar": "solar_cells",
+}
+LATEST_DATA_URL = "http://bornholm.powerlab.dk/visualizer/latestdata"
+
+SOURCE = "bornholm.powerlab.dk"
+TIMEZONE = timezone("Europe/Copenhagen")
```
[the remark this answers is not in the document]
unitrium: You're right it's UTC

parsers/BORNHOLM_POWERLAB.py
unitrium: In this case I find it easier to have the modes in the constructor. No real reason behind this but it's more for convenience, since both are supported I don't think we should favor one over the other.

unitrium:
Got time to get back to it, ready to review again :)

VIKTORVAV99 (approved):
LGTM!

Issue #5883, opened 2023-09-13
Wrong values for the tick in the breakdown graph for total energy


VIKTORVAV99:
**Describe the bug**

**To Reproduce**
1. Go to Italy in country mode
2. Click on yearly
3. Look at the breakdown graph ticks.
4. See two ticks with 100 TWh

**Expected behavior**
The second tick should be 120 TWh

**Screenshots**
<img width="469" alt="image" src="https://github.com/electricitymaps/electricitymaps-contrib/assets/30777521/7c7f5c50-41a7-48a3-a25b-3e10a43e3be5">

**Desktop (if relevant):**
 - OS: MacOS 13.5.2
 - Browser: Edge
 - Version: 116.0.1938.81

**Additional context**
Something about the formatting is off. It seems to happen every time there is 7 ticks in the axis.

VIKTORVAV99:
cc @madsnedergaard @tonypls

PR #5889, opened 2023-09-14
format axis to 2sf and reduce font size when more than 4 axis ticks

tonypls:
## Issue
Fixes: #5883 

## Description
This PR updates the bar breakdown chart axis formatting to display 2 significant figures. 

I've also update the font size of the axis to be slightly reduced in the rare case we show 8 axis ticks instead of 4. This prevents the axis labels overlapping

### Preview
Before:
image

Now:
image

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard (approved):
Makes sense, thanks for fixing it. I guess we could also set a limit of ticks in the future, to not allow more than 6 ticks :)

tonypls:
> Makes sense, thanks for fixing it. I guess we could also set a limit of ticks in the future, to not allow more than 6 ticks :)

I tried but unfortunately I couldn't find an easy way to do this within d3.

PR #5890, opened 2023-09-14
Ensure modal is always closeable

madsnedergaard:
## Issue

On small screens (iPhone SE) the new total intro modal cannot be closed.

## Description

This ensures that the close button is always visible (by setting a max-height on modals), and hides the image on small screens. Also scales up the size of it depending on screen size.

Note: I tried getting a scrollbar in place when content is higher than screen, but couldn't get it working in a nice way - so rolling with this first step for now.

### Preview

Now it should be closeable even on a watch!

Screenshot 2023-09-14 at 10 11 38

### Double check

### DIFF
--- a/web/src/components/Modal.tsx
+++ b/web/src/components/Modal.tsx
@@ -26,7 +26,7 @@ export default function Modal({
           // Avoid close button being auto-focused initially, as pressing space will otherwise close the modal
           onOpenAutoFocus={(event: Event) => event.preventDefault()}
           data-test-id={testId}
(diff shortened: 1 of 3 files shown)


VIKTORVAV99 (approved):
Approving!

PS: I just noticed the two bottom corners are not rounded 🙃

madsnedergaard:
They looked rounded to me? 😅  (screenshot from production)
Screenshot 2023-09-14 at 10 25 32

PR #5891, opened 2023-09-14
Rename PH parser to IEMOP

pierresegonne:
## Issue

`PH` is the country name not the zone name

## Description

Adopt new convention which use source names for parser

### Double check


unitrium (approved):
Thanks for taking up on the trend!

VIKTORVAV99 (approved):
🥳

PR #5894, opened 2023-09-14
remove include and add public to exclude

VIKTORVAV99:
## Issue

It was causing errors in the folders that was not part of the include, like scripts.

## Description

Removes the includes to run on all files except those in exclude, exclude public.

### Double check

### DIFF
--- a/web/cypress/tsconfig.json
+++ b/web/cypress/tsconfig.json
@@ -11,6 +11,5 @@
     "strict": true,
     "target": "ESNext",
     "types": ["cypress", "@testing-library/cypress"]
-  },
-  "include": ["."]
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
Well shit that was a lot of additional type errors...

I don't have time to look at that right now but will get it sorted as soon as possible.

VIKTORVAV99:
@madsnedergaard do vscode show errors for you for the node methods in `/electricitymaps-contrib/web/scripts/generateZonesConfig.ts`?

It does for me but I can't figure out what the issue is since it pass tsc when run, so thinking it might just be a issue with the extension?

PR #5897, opened 2023-09-15
Make monthly include 1 more month

### LINKED ISSUE #5865: 13 Months tab instead of 12 Months
Hi All,

May I suggest an update to 13 Months tab instead of 12 Months. This way it is easier to compare the same month from the previous year. I noticed the 24 hours tab is in fact 25 hours. In any case my proposal is not associated with the  tab name, but rather with the display of the information. I find it useful to be able to compare the same hour from the previous day, same date from previous month and same month from previous year. Hope this makes sense. 

Regards,
Aleks

madsnedergaard:
## Issue

Closes #5865

## Description

This makes it easier to compare last month with the same month last year.
As an example, I would today be able to see August 2023 and August 2022.

I also fixed a small visual bug where the tick text was slightly off.

### Preview

Tick positioning

### Double check


VIKTORVAV99 (commented):
No issues with the code as far as I am concerned but cypress expects a call to the v6 state. We might want to update that 😅

VIKTORVAV99 (approved):
LGTM! 🎉

PR #5899, opened 2023-09-15
Fix NP config

pierresegonne:
## Issue

The Nepalese bounding box is inverted, timezone is not correct as well.

## Description

Well, fix that

### Double check


VIKTORVAV99 (approved):
Very strange, thanks for fixing it!

PR #5902, opened 2023-09-15
Don't fly to zone when geometry changes and uppercase zones

tonypls:
## Issue
Currently when a zone is selected and we toggle between the country and zone view it can trigger the fly to zone action causing some unexpected behaviour.

## Description
This PR removes the geometries from the useEffect dependency array, this was previously used when the ranking panel included zones from both the country and zone view but is not longer required.

This PR also redirects the app when zoneKeys are not capitalized. Previously this just showed an empty zone. 

### Double check


madsnedergaard (approved):
Nice fixes! :)

VIKTORVAV99 (approved):
Very nice with the upper case fix!

PR #5915, opened 2023-09-18
bump prettier and plugin

VIKTORVAV99:
### Double check


VIKTORVAV99:
The current (hacky) implementation we are using is no longer supported. Closing this for now.

PR #5916, opened 2023-09-18
Remove KW `fetch_production`

unitrium:
## Issue

The parser for KW is creating parser data making assumptions based on the consumption to create production data. We want to avoid this. Parser should just collect the data available from datasources and format it to electricityMaps format.

The conversion from consumption to production in selected zones will be handled by one of our estimation models with a clear disclaimer that the data is estimated and not measured.

## Description

- remove `fetch_production` from KW parser.

Initially I was intending on removing it from TH as well as we are facing similar issues but the change is a bit more complex as it is generation data and not consumption data. Will do a follow up PR later.

### Double check

### DIFF
--- a/config/zones/KW.yaml
+++ b/config/zones/KW.yaml
@@ -80,5 +80,4 @@ fallbackZoneMixes:
         wind: 0.0
 parsers:
   consumption: KW.fetch_consumption
-  production: KW.fetch_production
 timezone: Asia/Kuwait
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (approved):
LGTM!

Is the new estimation method already up and running or should we wait to merge this?

unitrium:
I hope to get it up and running later today

madsnedergaard (approved):
LGTM :)

PR #5925, opened 2023-09-21
Update Mexico's emission factors

unitrium:
## Issue

Since we are now using an estimation for mexico based on it's consumption we should adjust the emission factors of the region for unknown. For now we are going to use the annual breakdown from the IEA from 2020 until a more recent and better source can be used.

## Description

Update the emission factor of all Mexico zones based on the annual breakdown.

### Double check

### DIFF
--- a/config/zones/MX-BC.yaml
+++ b/config/zones/MX-BC.yaml
@@ -11,6 +11,17 @@ fallbackZoneMixes:
       gas: 0.923
       wind: 0.077
   source: "Electricity Maps & Dra. Gabriela Mu\xF1oz Mel\xE9ndez"
+emissionFactors:
+  direct:
+    unknown:
+      - datetime: '2017-01-01'
+        source: "Electricity Maps & Dra. Gabriela Mu\xF1oz Mel\xE9ndez"
+        value: 453
+  lifecycle:
+    unknown:
+      - datetime: '2017-01-01'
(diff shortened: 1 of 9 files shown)


config/zones/MX-BC.yaml
```diff
@@ -11,6 +11,17 @@ fallbackZoneMixes:
       gas: 0.923
       wind: 0.077
   source: "Electricity Maps & Dra. Gabriela Mu\xF1oz Mel\xE9ndez"
+emissionFactors:
+  direct:
+    unknown:
+      - datetime: '2017-01-01'
+        source: "Electricity Maps & Dra. Gabriela Mu\xF1oz Mel\xE9ndez"
```
VIKTORVAV99, line 18: I think we can get rid of the unicode encoding here, make it looks a little better when looking at the file.

VIKTORVAV99 (approved):
I ran prettier so this looks good now I think. Just one tiny suggestion about some text formatting.

config/zones/MX-BC.yaml
unitrium: Not entierly sure if our whole pipeline will support it, so to be sure I would rather keep it as it is :)

VIKTORVAV99: Yaml it self supports it but that is a valid point.

let's keep it

unitrium: Though I hope it supports it but I don't want to spend time fixing it if it doesn't right now 😅

PR #5930, opened 2023-09-22
Fix KeyError and refactor TW parser

pierresegonne:
## Issue

`TW` was not getting any data recently. The reason is that one production mode stopped being reported in the source: https://storage.googleapis.com/electricitymap-parser-logs/TW.html

## Description

* Makes parser robust for such issues
* Refactor
* Test and tralala

### Preview

Screenshot 2023-09-22 at 16 12 54

### Double check


parsers/TAIPOWER.py
```diff
@@ -1,43 +1,45 @@
...
     target_datetime: datetime | None = None,
     logger: Logger = getLogger(__name__),
 ) -> dict:
     if target_datetime:
         raise NotImplementedError("This parser is not yet able to parse past dates")
 
-    url = "http://www.taipower.com.tw/d006/loadGraph/loadGraph/data/genary.txt"
-    s = session or Session()
-    response = s.get(url)
+    s = session
```
unitrium, line 29:
```suggestion
```
You can use directly the session object ?

parsers/TAIPOWER.py
```diff
@@ -1,43 +1,45 @@
...
+    response = s.get(PRODUCTION_URL)
```
unitrium, line 30:
```suggestion
    response = session.get(PRODUCTION_URL)
```

unitrium (approved):
Thanks for handling this, I will have a look on how to support capacity natively during slack week.

PR #5934, opened 2023-09-22
upgrade LK parser and rename it

VIKTORVAV99:
## Issue

The parser didn't use the new parser classes and where missing tests.

## Description

- Upgraded the parser to use parser classes
- Removed arrow
- Renamed parser
- Added tests

### Double check


VIKTORVAV99:
not really sure why the test is failing here, it's passing for me locally. I would appreciate if you could take a look.

parsers/CEB.py
```diff
@@ -40,14 +40,16 @@ def fetch_production(
     logger: Logger = getLogger(__name__),
 ):
     """Requests the previous day's production mix (in MW) for Sri Lanka, per quarter-hour"""
-    if target_datetime is not None:
-        raise NotImplementedError(
-            "It is not possible to fetch historical data for this datasource, it only has data for the previous day"
-        )
+    if target_datetime is None:
+        target_datetime = datetime.now(tz=timezone.utc)
```
q--, line 44: Did you take into account that the [data source](//cebcare.ceb.lk/gensum/details) only has data up to the previous day (inclusive)?

VIKTORVAV99: Yep, when I tested it returns the most up to date data when it's the current date even if the current date doesn't exist yet.

We can do a time delta of - 1 day if you prefer that though.

q--: No, if it works it's fine.

parsers/CEB.py
```diff
@@ -0,0 +1,97 @@
...
+    zone_key: ZoneKey = ZoneKey("LK"),
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+):
+    """Requests the previous day's production mix (in MW) for Sri Lanka, per quarter-hour"""
+    if target_datetime is None:
+        target_datetime = datetime.now(tz=timezone.utc)
+
+    ses = session or Session()
```
unitrium, line 46:
```suggestion
    if session is None:
        session = Session()
```
I would recommend avoiding renaming to shorter variable names. It often can result in confusions. The more you can read your code without thinking about what this variable is, the better!

unitrium (approved):
Looks good thanks for handling this!

q--:
Could the test error have anything to do with time zone settings? A difference of two hours is exactly the difference between Central European Summer Time, which I presume your laptop is set to, and UTC.

VIKTORVAV99:
> Could the test error have anything to do with time zone settings? A difference of two hours is exactly the difference between Central European Summer Time, which I presume your PC is set to, and UTC.

Could be it but I thought I made sure all date times where time zone aware and only used UTC. I'll have to double check though.

PR #5936, opened 2023-09-22
fix MX-BC->US-CAL-CISO exchange

VIKTORVAV99:
## Issue

The MX-BC->US-CAL-CISO exchange where not working.

## Description

For some reason this particular exchange was using the US_CA.py parser but the actual source was the same as the CENACE.py parser so I added the exchange to the CENACE.py parser and removed the relevant code from the US_CA.py parser. 

I also took the opportunity to switch to ZoneInfo instead of pytz in the CENACE.py parser.

### Double check


VIKTORVAV99:
@madsnedergaard could we look into adding automatic issues for downed exchanges as well?
Should be considered almost as critical as production data IMO as we do flowtracing and it won't work without it.

unitrium (approved):
Thanks for taking care of this!

madsnedergaard:
> @madsnedergaard could we look into adding automatic issues for downed exchanges as well? Should be considered almost as critical as production data IMO as we do flowtracing and it won't work without it.

It is technically harder, but I will add it to the roadmap :)

PR #5937, opened 2023-09-23
fix KPX parser

VIKTORVAV99:
## Issue

We where assigning the production and hydro values directly, we should not do that as it can introduce floating point errors when one Emaps mode has multiple source modes.

## Description

This PR changes the KPX parser to use mappings and looping over the source modes and then assigning the values with add_value() instead which has been designed to prevent the issue mentioned above.

### Double check


VIKTORVAV99:
@mathilde-daugy her you can see a pretty good example of the floating point errors we introduce in the data when not using the new parser classes.

pierresegonne (approved):
🙌🏻

PR #5940, opened 2023-09-23
Fix MN parser

### LINKED ISSUE #5905: MN parser down
## Description
Mongolia parser is down, but the bot hasn't open an issue. I am opening this one to alert the maintainers.
It seems this time the parser is down because of a change in the json that was being parsed:
> raise ParserException(
parsers.lib.exceptions.ParserException: MN.py Parser: Fetched keys from source dict_keys(['date', 'syssum', 'tpp', 'sumnar', 'sums', 'energyimport', 't']) do not match expected keys dict_values(['date', 'syssum', 'sumnar', 'sums', 'energyimport', 't']).

A new key called tpp (thermal power plants?) has being added. The value of this new key doesn't match the previously calculated unknown production (so tpp plus other keys don't add up to consumption). What should be done to fix this? It seems an unknown source is being added.

By the way, a bit off-topic, but I have noticed that the Mongolia parser outputs global exchange data. We currently get the exchange with Russia from its parser, so we could calculate the exchange with China by substracting the other exchange. Is this possible?

silimotion:
## Issue

Closes #5905

## Description

This PR fixes the Mongolia parser by ignoring the new "tpp" value reported, as reasoned in the linked issue.

I also changed pytz to zoneinfo for the timezone, as the previous pytz use was incorrect (+7:08 timezone instead of +8:00), and pytz is in maintenance mode.

### Double check


VIKTORVAV99:
Took another look at this and it seems like we should no longer ignore the TPP value, 

Syssum (consumption) minus TPP, and the other values (including imports) result in a leftover of 10 mw which seems reasonable as unknown now that TPP is reported separately.

So TTP and remainder should both be reported as unknown (small difference but it gives us more flexibility in the future). There is also a bunch of new keys added to the JSON response that need to be handled before this can be merged.

VIKTORVAV99 (changes requested):
Took another look at this and it seems like we should no longer ignore the TPP value,

Syssum (consumption) minus TPP, and the other values (including imports) result in a leftover of 10 mw which seems reasonable as unknown now that TPP is reported separately.

So TTP and remainder should both be reported as unknown (small difference but it gives us more flexibility in the future). There is also a bunch of new keys added to the JSON response that need to be handled before this can be merged.

silimotion:
> Syssum (consumption) minus TPP, and the other values (including imports) result in a leftover of 10 mw which seems reasonable as unknown now that TPP is reported separately.

Sometimes this leftover is negative, as in the date I pointed out in a comment in the issue. I suppose it is due to batteries because there is an image of them in the website ([https://ndc.energy.mn/](https://ndc.energy.mn/)). I would suggest waiting until they add them to the json response to see if the leftover is negative then.

> So TTP and remainder should both be reported as unknown (small difference but it gives us more flexibility in the future). 

Do you mean adding both? Wouldn't that be the same as what is returned now? As I said, as the leftover may be negative I don't know if it would really carry any meaning.

> There is also a bunch of new keys added to the JSON response that need to be handled before this can be merged.

If you don't mind I will change the parser so that it raises an alert instead of an error, as it seems they are adding keys frequently.

VIKTORVAV99:
> > Syssum (consumption) minus TPP, and the other values (including imports) result in a leftover of 10 mw which seems reasonable as unknown now that TPP is reported separately.
> 
> Sometimes this leftover is negative, as in the date I pointed out in a comment in the issue. I suppose it is due to batteries because there is an image of them in the website (https://ndc.energy.mn/). I would suggest waiting until they add them to the json response to see if the leftover is negative then.
> 
> > So TTP and remainder should both be reported as unknown (small difference but it gives us more flexibility in the future).
> 
> Do you mean adding both? Wouldn't that be the same as what is returned now? As I said, as the leftover may be negative I don't know if it would really carry any meaning.

I know the value won't change but if they change it later on it will be easier to make changes as TPP is already split out (instead of ignored).

> 
> > There is also a bunch of new keys added to the JSON response that need to be handled before this can be merged.
> 
> If you don't mind I will change the parser so that it raises an alert instead of an error, as it seems they are adding keys frequently.

We have been thinking about doing the same for other parsers so feel free to go ahead.

parsers/MN.py
```diff
@@ -38,9 +43,8 @@ def parse_json(web_json: dict) -> dict[str, Any]:
 
     # Validate first if keys in fetched dict match expected keys
     if set(JSON_QUERY_TO_SRC.values()) != set(web_json.keys()):
-        raise ParserException(
-            parser="MN.py",
-            message=f"Fetched keys from source {web_json.keys()} do not match expected keys {JSON_QUERY_TO_SRC.values()}.",
+        logger.warning(
+            msg=f"Fetched keys from source {web_json.keys()} do not match expected keys {JSON_QUERY_TO_SRC.values()}.",
         )
```
VIKTORVAV99, line 48:
```suggestion
        logger.error(
            msg=f"Fetched keys from source {web_json.keys()} do not match expected keys {JSON_QUERY_TO_SRC.values()}.", extra={zone_key: zone_key, parser: "MN.py"}
        )
```

parsers/MN.py
```diff
@@ -93,23 +97,25 @@ def fetch_production(
...
-        solar=query_data["solarMW"],
-        wind=query_data["windMW"],
-        unknown=query_data["unknownMW"],
-    )
+    prod_mix = ProductionMix()
+    prod_mix.add_value("solar", query_data["solarMW"])
+    prod_mix.add_value("wind", query_data["windMW"])
+    prod_mix.add_value(
+        "unknown", query_data["leftoverMW"] + query_data["thermal"]
+    )  # Thermal is currently unknown
```
VIKTORVAV99, line 118: There shuld be no math operations for adding the values here, prefarable we should just loop over all the values so it is easy to add new keys/modes in the future.

This does work though so it's not a blocker.

VIKTORVAV99 (changes requested):
We are missing the logger argument for the query function call in the fetch consumption function.

VIKTORVAV99 (changes requested):
We are missing the logger argument for the query function call in the fetch consumption function.

VIKTORVAV99 (changes requested):
We are missing the logger argument for the query function call in the fetch consumption function.

Otherwise the rest is just suggestions.

parsers/MN.py
silimotion: I think I've understood what you meant here, but please point out if not

silimotion:
The endpoint was down for several weeks. Now the data is available on an slightly different URL, and it not longer contains the new keys.

I have updated the PR to use the new endpoint.

VIKTORVAV99 (approved):
LGTM!

Thanks for fixing it (and fixing it again when the source changed)!

silimotion:
@VIKTORVAV99 It seems like the second last merge undid the change to the endpoint, so the parser isn't working.

Can we fix it from here or is another PR needed?

VIKTORVAV99:
> @VIKTORVAV99 It seems like the second last merge undid the change to the endpoint, so the parser isn't working.
> 
> Can we fix it from here or is another PR needed?

Oh that not great... Merge conflicts really have a way of messing things up 😅

Since this is already merged even if you used the same branch it would need a new PR.

The good news is that we can merge it right away 🙂

PR #5947, opened 2023-09-25
Update FI.yaml

pnook:
Latest wind capacity

### Double check


VIKTORVAV99 (approved):

PR #5952, opened 2023-09-26
upgrade ENTSO-E exchange to parser classes

VIKTORVAV99:
## Issue

The ENTSO-E parser still used the old parser structure.

## Description

Upgrades and simplifies the exchange parsing in the ENTSO-E parser. (Bonus: removal of pytz).

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -932,10 +932,8 @@ def parse_production_per_units(xml_text: str) -> Any | None:
 def parse_exchange(
     xml_text: str,
     is_import: bool,
-    zoneKey1: ZoneKey,
-    zoneKey2: ZoneKey,
+    exchangeKey: ZoneKey,
     logger: Logger,
-    existingExchangeList: ExchangeList | None = None,
     forecasted: bool = False,
 ) -> ExchangeList:
     if not xml_text:
@@ -944,10 +942,10 @@ def parse_exchange(
             message="No forcasted exchange data found"
             if forecasted
             else "No exchange data found",
-            zone_key="->".join(sorted([zoneKey1, zoneKey2])),
+            zone_key=exchangeKey,
         )
 
-    newExchangeList = ExchangeList(logger)
+    exchangeList = ExchangeList(logger)
 
     soup = BeautifulSoup(xml_text, "html.parser")
     # Get all points
@@ -971,8 +969,8 @@ def parse_exchange(
             position = int(entry.find_all("position")[0].contents[0])
             datetime = datetime_from_position(datetime_start, position, resolution)
             # Find out whether or not we should update the net production
-            newExchangeList.append(
-                zoneKey=ZoneKey("->".join(sorted([zoneKey1, zoneKey2]))),
+            exchangeList.append(
+                zoneKey=exchangeKey,
                 datetime=datetime,
                 source=SOURCE,
                 netFlow=quantity,
@@ -981,12 +979,7 @@ def parse_exchange(
                 else EventSourceType.measured,
             )
 
-    if existingExchangeList is not None:
-        return ExchangeList.merge_exchanges(
-            ungrouped_exchanges=[existingExchangeList, newExchangeList], logger=logger
-        )
-    else:
-        return newExchangeList
+    return exchangeList
 
 
 def parse_prices(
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -1000,6 +1018,67 @@ def parse_prices(
...
+                    xml_text=raw_exchange,
+                    is_import=True,
+                    zoneKey1=zone_key1,
+                    zoneKey2=zone_key2,
+                    logger=logger,
+                    existingExchangeList=parsed,
+                    forecasted=forecasted,
+                )
+        return parsed.to_list()
+    else:
```
unitrium, line 1073: Lots of nested ifs, I would recommend getting at least rid of one level by inverting this.
You could do:
```
if raw_exchange is None:
    raise
do stuff
```

parsers/ENTSOE.py
```diff
@@ -1000,6 +1018,67 @@ def parse_prices(
...
+    if raw_exchange is not None:
+        parsed = parse_exchange(
+            raw_exchange,
+            is_import=False,
+            zoneKey1=zone_key1,
+            zoneKey2=zone_key2,
+            logger=logger,
+            forecasted=forecasted,
+        )
+        if parsed:
```
unitrium, line 1055: Are you sure about this ?
Shouldn't we try to get exports and imports independently of whether there's an import or export ?

parsers/ENTSOE.py
```diff
@@ -1000,6 +1018,67 @@ def parse_prices(
...
+            if raw_exchange is not None:
+                parsed = parse_exchange(
+                    xml_text=raw_exchange,
+                    is_import=True,
+                    zoneKey1=zone_key1,
+                    zoneKey2=zone_key2,
+                    logger=logger,
+                    existingExchangeList=parsed,
+                    forecasted=forecasted,
+                )
```
unitrium, line 1071: Shouldn't we merge the import and the exports ?

parsers/ENTSOE.py
```diff
@@ -1000,6 +1018,67 @@ def parse_prices(
...
+    key = ZoneKey("->".join(sorted_zone_keys))
+    if key in ENTSOE_EXCHANGE_DOMAIN_OVERRIDE:
+        domain1, domain2 = ENTSOE_EXCHANGE_DOMAIN_OVERRIDE[key]
+    else:
+        domain1 = ENTSOE_DOMAIN_MAPPINGS[zone_key1]
+        domain2 = ENTSOE_DOMAIN_MAPPINGS[zone_key2]
+    # Grab exchange
+    # Export
+    raw_exchange = (
+        query_exchange_forecast(domain1, domain2, session, target_datetime)
```
unitrium, line 1042: Since the functions have the same parameters you could have some like this:
```
query_function =  query_exchange
if forecasted:
   query_function = query_exchange_forecast
...

query_function(domain1, domain2, session, target_datetime)
...
query_function(domain2, domain1, session, target_datetime)
```
This way you don't have to redo the if else.

unitrium (commented):
A few extra questions before getting there.
Since this is an important parser I would recommend creating tests for this. If you don't have time I can quickly setup a test case for this :)

parsers/ENTSOE.py
VIKTORVAV99: As far as I am aware they are always matched and since we parse them alphabetically the order will always be the same. But I can look into making this more robust.

parsers/ENTSOE.py
VIKTORVAV99: They are merged in parse_exchange function if the existingExchangeList is not empty/None. I'll try and clear this up a bit though.

VIKTORVAV99:
> A few extra questions before getting there. Since this is an important parser I would recommend creating tests for this. If you don't have time I can quickly setup a test case for this :)

I played around with testing it a little and I'm not sure it's as easy to test as the other endpoints in the ENTSOE parser as it calls the same endpoint twice with different params to get the import/exports.

So if you want to do it feel free to do so.

However when I played around with the XML files I found out we can get the direction automatically instead of setting it manually (first call being exports and the second hardcoded as imports). But it will increase the complexity a bit. Do you think that is something we should implement?

VIKTORVAV99:
Figured out a good way to add tests but I'm going to close this PR and create a separate one for the tests and then after those are merged re-create this to avoid merge conflict messiness.

PR #5967, opened 2023-10-01
fix wrong term in zh-TW

hms5232:
## Issue

None.

## Description

Fix "server" translation: in Taiwan is "伺服器", "服務器" is used in China (zh-CN).


tonypls (approved):
LGTM 🎉  Thanks for the contribution!

PR #5979, opened 2023-10-03
Remove US generation only zones configs

pierresegonne:
## Issue

We still have config files for zones that have been merged into another for the US

## Description

Remove these deprecated config files

### Double check


VIKTORVAV99 (approved):
LGTM 🎉

Issue #3206, opened 2021-07-05
update CA-NS parser


jarek:
We're logging error messages like "discarding datapoint at 2021-07-04 18:00:00+00:00 due to coal percentage out of bounds: 0.24559999999999998"

Our current parser rejects points that have less than 25% coal in the fuel mix, in an effort to duplicate validation functionality of the source website (added in #1193). However the source website appears to have changed and AFAICT there is no longer validation there (`ChartTodaysPower` function in https://www.nspower.ca/bundles/mainjs?v=6UoQDTSr7sd0yMo3fE3LJgJn4iAuAmXV0JumYpeHxnc1). We should at very least relax the validation bounds a little bit, less than 25% coal is now apparently not uncommon. We should also check the known-max-capacities in the validation section of our parser to make sure they're still accurate.

Also, the current/new source website uses a different API endpoint - we use https://www.nspower.ca/library/CurrentLoad/CurrentMix.json, the website now uses https://www.nspower.ca/api/todayspower/powermix. However the new one appears to be inferior as it doesn't have short-term-historical data as the old one does, and as long as the old URL still works it's probably better to keep using it?

We would also want to update source page in any documentation ("real time data sources" list) to https://www.nspower.ca/cleanandgreen/todays-energy-stats

skovhus:
Thanks for creating this issue @jarek

jarek:
The source page https://www.nspower.ca/cleanandgreen/todays-energy-stats now returns an empty 404 response, and there's a leftover "Today's Energy Stats" box on pages like https://www.nspower.ca/cleanandgreen/renewable-energy-sources/tidal-power but it doesn't link to anything unlike the boxes next to it :/ Has too much transparency became inconvenient?

The source JSON links linked above still work. But they log too much hydro: currently (Sun 16 Jan 2022, 9pm NS time) they log 1892.58 MW load x 25.34% hydro supply = 479.58 MW (and the hour before it was 554 MW) whereas I'm not finding any sources for new hydro plants that would bring hydro capacity above the previously published 400-420 MW. The error logs at https://storage.googleapis.com/electricitymap-parser-logs/CA-NS.html recently log hydro being too high as the reason for rejecting datapoints.

[...]

PR #5985, opened 2023-10-04
Fix CA-NS parser

silimotion:
## Description
Nova Scotia data is missing for some hours:

Screenshot_20231004_220430

This is caused by the validation done inside the parser, which was written 6 years ago. It seems the production has changed a bit since, which makes the checks fail.
 I've updated the checks, but I think we could consider removing the validation altogether to avoid a similar situation in the future. The opinion of @jarek could be useful, as he made the validation part. Especially since the link in the documentation throws a 404.

Looking around I've found #3206, an interesting discussion regarding this, but that didn't spawn a PR. I think this could be merged to stop discarding data automatically, and implement whatever solution emerges from other discussions later on.

I've used the opportunity to also remove the arrow dependency from this parser

### Double check

### DIFF
--- a/parsers/CA_NS.py
+++ b/parsers/CA_NS.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python3
 
-# The arrow library is used to handle datetimes
-from datetime import datetime
+# The datetime library is used to handle datetimes
+from datetime import datetime, timezone
 from logging import Logger, getLogger
 
-import arrow
 from requests import Session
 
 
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):

VIKTORVAV99 (approved):
LGTM! 🎉

I do agree that we should look over this validation and maybe even see if there is a new data source availabe when we update this parser to use parser classes.

PR #5997, opened 2023-10-09
Bar chart tooltips are not rendered from the correct hover location

tonypls:
## Issue
Currently we anchor the chart tick and render tooltips from a mouse hover that is 0.5 width to the left of the bar being hovered
## Description
This PR updates our detect hover point function to detect the hover event 0.5 * bar width to the right.

### Double check


VIKTORVAV99 (approved):
I can't really tell the difference but this looks good to me code wise and it don't break anything so let's merge it.

PR #5987, opened 2023-10-06
Revive: updated lifecycle estimates for CA-ON

jayaddison:
## Issue

Closes #4886.

## Description

Following on from and building on work from existing pull request #4926, this changeset adds lifecycle emission factors for the three production sources (gas, hydro, wind) in Ontario, Canada presented in the paper ["Life cycle greenhouse gas emissions of electricity generation in the province of Ontario, Canada" by E. Mallia and G. Lewis](https://link.springer.com/article/10.1007/s11367-012-0501-0).

As mentioned in the issue report, this paper has previously been discussed [here](https://github.com/electricitymaps/electricitymaps-contrib/discussions/4869) and [here](https://github.com/electricitymaps/electricitymaps-contrib/discussions/4791#discussioncomment-4295383).

Most of the pull request was in good shape; getting the source references to match is the main change and was verified locally using the following small Python script:

```python
import yaml

with open('config/zones/CA-ON.yaml') as f:
    content = f.read()

c = yaml.safe_load(content)
ref1 = list(c['sources'].keys())[0]
ref2 = c['emissionFactors']['lifecycle']['gas']['source']
assert ref1 == ref2, "verification failed"
```

### Preview

N/A

### Double check

- [ ] ~~I have tested my parser changes locally with `poetry run test_parser "zone_key"`~~
- [ ] ~~I have run `pnpx prettier --write .` and `poetry run format` to format my changes.~~

PR #6001, opened 2023-10-11
Add validation test for emissions factor references provided in zone configs

jayaddison:
## Description

While developing #5987 it was nontrivial for me to check whether the `source` field defined for an emission factor matched an existing top-level `sources` entry in the zone.

This pull request provides a test case to perform that validation automatically, and also caters for a number of existing situations where any of the three of {averaged Electricity Maps data, or implicit well-known source references, or assumptions} are made about the emissions factor estimates.

### Double check

- [ ] ~~I have tested my parser changes locally with `poetry run test_parser "zone_key"`~~
- [x] I have run ~~`pnpx prettier --write .` and~~ `poetry run format` to format my changes.


tests/test_zones_json.py
```diff
@@ -7,6 +7,24 @@
 
 
 class ZonesJsonTestcase(unittest.TestCase):
+    # Add well-known sources that don't require config-based references here
+    GLOBAL_SOURCE_REFERENCES = {
```
jayaddison, line 11: I think it'd be preferable to create a separate YAML reference sources file that contains these and any relevant metadata -- extracting a constant set here to hold them is something of a migration phase to figure out how many items exist and to spot patterns in them.

tests/test_config_model.py
```diff
@@ -11,6 +11,45 @@ def test_pydantic_model(self):
...
+        "EU-ETS, ENTSO-E 2021",
+        "Guatemala AMM 2021-2022",
+        "IEA 2019",
+        "IEA 2020",
+        'Oberschelp, Christopher, et al. "Global emission hotspots of coal power generation."',
+        "Tidal (IPCC 2014)",
+        "https://www.iea.org/data-and-statistics/charts/electricity-generation-mix-in-mexico-1-jan-30-sep-2019-and-2020",
+    }
+
+    def test_zone_sources(self):
```
jayaddison, line 32: @pierresegonne following on from [my question yesterday](https://github.com/electricitymaps/electricitymaps-contrib/pull/4435#discussion_r1354593945): this test case is the reason I was wondering about allowing the `emissionFactors` key to remain on the in-memory zone config representations.

These changes are work-in-progress; defining a set of well-known `sources` within the code doesn't seem ideal, so I'm planning some follow-up these changes with a migration of `GLOBAL_SOURCE_REFERENCES` into another, new config YAML file (`config/sources.yaml`, or similar).

Also note: given your mention of memory usage as a concern: the changes here will require more memory not only for the loaded-YAML object representation of `emissionFactors`, but then another bunch of memory allocations for the `pydantic` field to hold them on each zone instance.

jayaddison: Ah, hold on.  After re-reading your response, I double-checked and -- with the updated understanding from developing these changes -- I think I understand that these stats are already available in the `electricitymap.contrib.config.CO2EQ_PARAMETERS` object at runtime.  In particular, using a Python interactive shell appears to show the data under `CO2EQ_PARAMETERS['emissionFactors']['zoneOverrides']`.

That would seem to make sense, and could mean that I don't need the `ZoneEmissionFactors` model that I'd added here...

...and then the `del` statement can be restored too.

config/zones/UA.yaml
```diff
@@ -90,7 +90,7 @@ emissionFactors:
       _comment: estimated yearly share of solar & wind in this mixed renewable category
         based on installed capacity in 2018
       _url: https://en.wikipedia.org/wiki/Renewable_energy_in_Ukraine
-      source: renewable mix; assumes 50% solar, 50% wind
+      source: assumes 50% solar, 50% wind renewable mix
```
jayaddison, line 93: According to figures from the [latest currently-available iteration of the same Wikipedia source page](https://en.wikipedia.org/w/index.php?title=Renewable_energy_in_Ukraine&oldid=1153471433#History), in Y2021 the ratio of `solar:wind` capacity was `6227:1673`.

However: those figures predate the conflict in Ukraine that began in Y2022 and so the numbers may not correspond to the situation more recently.  Note also that data for Ukraine is currently unavailable on Electricity Maps (the parser has been down for a while; I'm not cross-linking the relevant issue thread here because it's not relevant to most of this pull request).

jayaddison: Also: installed capacity ratio doesn't necessarily indicate actual output capacity.  So even though the capacity ratio may have changed, the ratio itself may not be suitable to map 1-1 to the percentages assumed here.

config/zones/IN-KA.yaml
```diff
@@ -83,10 +83,8 @@ emissionFactors:
         source: Electricity Maps, 2021 average
         value: 459.2042079468457
     unknown:
-      _comment: 'Source: Souther Regional Load Dispatch Center Allocation Limits'
-      _url:
-        - http://srldc.org/2018-19srallocation.aspx
-      source: calculated 20% nuclear, 80% thermal (coal)
+      _comment: calculated 20% nuclear, 80% thermal (coal)
+      source: Souther Regional Load Dispatch Center Allocation Limits 2018-2019; SRLDC 2018-2019 derived calculation for Karnataka
```
jayaddison, line 87: @zuron7 this is data you collected a few years ago now, but I wondered whether you know where there might be updated data for Karnataka from SRLDC (or elsewhere)?  It looks like their website has moved from `srldc.org` to `srldc.in`, and I'm having difficulty accessing it.

jayaddison:
I'm going to open this as 'ready for review'.  Initially my goal here was solely to write a test case to make it easier to detect source-related problems - along the way the work has branched off into a few other source related cleanups and questions, but I think I should pause before going any further into the details on those.

VIKTORVAV99:
Since as you mentioned this has gotten a little bigger than initially intended I won't have time to review it today but at first glance everything seems pretty straightforward. So hopefully I can test and review it this weekend.

jayaddison:
Thanks @VIKTORVAV99 - it can (dare I say should?) wait until next week, though.

tests/test_config_model.py
```diff
@@ -11,6 +11,40 @@ def test_pydantic_model(self):
...
+                        )
+                        for source in estimate.source.split(";"):
+                            source = source.strip()
+                            if source.startswith("assumes"):
+                                continue
+                            if source.startswith("Electricity Maps"):
+                                continue
+                            if source in self.GLOBAL_SOURCE_REFERENCES:
+                                continue
+                            self.assertIn(source, CONFIG_MODEL.zones[zone_key].sources)
```
VIKTORVAV99, line 46: This line here has a type error and prints the following using pylance:
> Argument of type "dict[str, Source] | None" cannot be assigned to parameter "container" of type "Iterable[Any] | Container[Any]" in function "assertIn"

tests/test_config_model.py
```diff
@@ -11,6 +11,40 @@ def test_pydantic_model(self):
...
+        "EIA 2020/BEIS 2021",
+        "EU-ETS 2021",
+        "EU-ETS, ENTSO-E 2021",
+        "IEA 2019",
+        "IEA 2020",
+        'Oberschelp, Christopher, et al. "Global emission hotspots of coal power generation."',
+    }
+
+    def test_zone_sources(self):
+        for measurement_basis, model in CO2EQ_CONFIG_MODEL:
```
VIKTORVAV99, line 27: Since measurement_basis is not used I would underscore it.

VIKTORVAV99 (commented):
Works really well, most of the changes are straight forward and nothing special to note.

Just some small comments but nothing that should affect functionality.

tests/test_config_model.py
jayaddison: Ah, my mistake committing a partial change-in-progress there: I was considering adding that variable to the [error message here](https://github.com/jayaddison/electricitymaps-contrib/blob/42e27fe69891c3d39c2e815d31c1fbd41198ed14/tests/test_config_model.py#L36) to help people figure out where the source entry is missing.  What do you think?

tests/test_config_model.py
jayaddison: Ok; and it seems that this is due to the fact that the field is optional (an `x in foo` check is invalid if `foo` is `None`).  But it _is_ correct for it to be an optional field.  And `pyright` isn't convinced-enough by adding a `assertIsNotNone` on the previous line to eliminate the `| None` possibility.

Thinking about how to resolve this.. (adding `or {}` to the end of this line would be a quick although not-ideal way).

jayaddison: > And pyright isn't convinced-enough by adding a assertIsNotNone on the previous line to eliminate the | None possibility.

See microsoft/pyright#2007 for the (reasonable) rationale behind that.  However, the workaround suggested (add an `assert ... is not None` - attempted locally as `assert CONFIG_MODEL.zones[zone_key].sources is not None`) does not seem to be working for me.  I'll try to figure out whether that's a `pyright` bug.

jayaddison: Yep, I think there could be a `pyright` bug here.  `assert .. is not None` doesn't seem to preserve the asserted non-empty condition in the case where attribute access is used here:

```py
                            zone_sources = CONFIG_MODEL.zones[zone_key].sources
                            assert CONFIG_MODEL.zones[zone_key].sources is not None
                            assert zone_sources is not None
                            self.assertIn(source, CONFIG_MODEL.zones[zone_key].sources)  # pyright complains here
                            self.assertIn(source, zone_sources)  # pyright does not complain here
```

jayaddison: Hmm.  Maybe that's as-expected - possibly because the `pydantic.BaseModel` that `Zone` derives from implements `__getattr__`, so the results of attribute access could change from call to call.

jayaddison: Ok, this should now be resolved (96974de4253351bdfdd14af1f8d68c9a9c559316 - hold the container's value in a local variable and use an `assert ... is not None` to help `pyright` narrow the typechecking for it).

tests/test_config_model.py
jayaddison: Hopefully resolved; I think that making the error messages from this unit test as unambiguous as possible would make sense, so I've standardized the format of them a bit, and included the full path to the `source` key when it's missing (making use of the `measurement_basis` variable).

VIKTORVAV99 (approved):
LGTM! 🎉

jayaddison:
Thank you @VIKTORVAV99!  Taking a look at resolving the merge conflicts now.

jayaddison:
I've resolved the conflicts locally, although tests fail (as intended, I think) due to a recently-added `CH` zone source without a top-level definition in the zone config.

I'll wait to hear back from the [related `CH` zone thread](https://github.com/electricitymaps/electricitymaps-contrib/pull/6032#discussion_r1369259824) to confirm whether there's a preferred way to attribute that updated source.

pierresegonne:
I think I understand what you had in mind for `CH`. Yes you can definitely create:

```
sources:
  - SFOE, Electricity Maps:
    link: ....
```

And then only reference `SFOE, ELectricity Maps` in the individual entries for emission factors

jayaddison:
Thanks @pierresegonne - yep, you understood my question :) (I also read your [earlier comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/6032#discussion_r1369698894) before this).

jayaddison:
Note: if this test causes too much friction in future when editing zone configs, please feel free to skip/remove them, I won't be offended.

On the other hand, if they turn out to be particularly useful, I'd be willing to try refactoring the logic into a `pydantic` field validator (like [these I see in the `events` module](https://github.com/electricitymaps/electricitymaps-contrib/blob/e53c5a41c2dfc72effd8afc50e8e356cf8049b12/electricitymap/contrib/lib/models/events.py#L235)) but I think it's worth figuring out whether the test is valuable, first.

VIKTORVAV99:
> Note: if this test causes too much friction in future when editing zone configs, please feel free to skip/remove them, I won't be offended.
> 
> On the other hand, if they turn out to be particularly useful, I'd be willing to try refactoring the logic into a `pydantic` field validator (like [these I see in the `events` module](https://github.com/electricitymaps/electricitymaps-contrib/blob/e53c5a41c2dfc72effd8afc50e8e356cf8049b12/electricitymap/contrib/lib/models/events.py#L235)) but I think it's worth figuring out whether the test is valuable, first.

I just merged it so let's let it run a few weeks maybe (we don't get a ton of updates to the config files). But at least in my opinion the more work that can be done by CI (and is reliable) the more time we can spend doing other things like more in depth reviews or developing new features for example.

Thanks once again!

Issue #6011, opened 2023-10-15
Upgrade all parsers with data-validating classes


kruschk:
I noticed on the [Building a new parser](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser#parser-validation) wiki page that several classes have been introduced to ensure parsed data is structured correctly; however, only about a third of all the parsers have been updated to use them. As promised in [my comment on #5986](https://github.com/electricitymaps/electricitymaps-contrib/issues/5986#issuecomment-1762369470), I've compiled the following list to keep track of which parsers have been upgraded and which have not:

- [x] [AW](../blob/master/parsers/AW.py)
- [ ] [AX](../blob/master/parsers/AX.py)
- [ ] [BD](../blob/master/parsers/BD.py)
- [ ] [BG](../blob/master/parsers/BG.py)
- [ ] [BO](../blob/master/parsers/BO.py)
- [x] [CAMMESA](../blob/master/parsers/CAMMESA.py)
- [x] [CA\_AB](../blob/master/parsers/CA\_AB.py)
- [x] [CA\_BC](../blob/master/parsers/CA\_BC.py)
- [x] [CA\_NB](../blob/master/parsers/CA\_NB.py)
- [ ] [CA\_NS](../blob/master/parsers/CA\_NS.py)
- [ ] [CA\_ON](../blob/master/parsers/CA\_ON.py)
- [ ] [CA\_PE](../blob/master/parsers/CA\_PE.py)
- [x] [CA\_QC](../blob/master/parsers/CA\_QC.py)
- [ ] [CA\_SK](../blob/master/parsers/CA\_SK.py)
- [ ] [CA\_YT](../blob/master/parsers/CA\_YT.py)
- [x] [CENACE](../blob/master/parsers/CENACE.py)
- [ ] [CH](../blob/master/parsers/CH.py)
- [ ] [CL](../blob/master/parsers/CL.py)
- [x] [CO](../blob/master/parsers/CO.py)
- [ ] [CR](../blob/master/parsers/CR.py)
- [ ] [CY](../blob/master/parsers/CY.py)
- [ ] [CZ](../blob/master/parsers/CZ.py)
- [x] [DK](../blob/master/parsers/DK.py)
- [ ] [DK\_BHM](../blob/master/parsers/DK\_BHM.py)
- [ ] [DO](../blob/master/parsers/DO.py)
- [x] [EIA](../blob/master/parsers/EIA.py)
[...]

Issue #5986, opened 2023-10-06
CA-SK production parser down


electricitymapsbot:
## Description

This is an automatic error report generated for Canada Saskatchewan (CA-SK).

Issues:
- No recent data found for `production` parser
- No recent data found for `consumption` parser

## Suggestions
- Try running the parser locally using the command `poetry run test_parser CA-SK production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/CA-SK.html">Explore the runtime logs</a>

You can see an overview of all parser issues [here](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Parser-issues).

kruschk:
Hi! I contributed a few times last year, and Hacktober has inspired me to get back to it (I'm a sucker for gamification :sweat_smile: ). I hope I'm still welcome to lend a hand!

I dug into this issue a bit, and it seems like the [consumption URL](https://www.saskpower.com/ignitionapi/Content/getnetload) has started returning `"Data Unavailable"` instead of the numeric value that (I believe) it used to. I figured this was likely just a transient thing, but then I noticed the JSON returned by the [production URL](https://www.saskpower.com/ignitionapi/PowerUseDashboard/GetPowerUseDashboardData) contains a `systemDemand` key that may not have been present before, so perhaps this was an API change. If that's the case, could we use that value instead? For reference, here's the relevant [SaskPower dashboard](https://www.saskpower.com/Our-Power-Future/Our-Electricity/Electrical-System/Where-Your-Power-Comes-From).

electricitymapsbot:
Closing this as the electricity data is flowing again.

kruschk:
I'm not sure why the bot closed this; the issue persists. I've noticed one other thing: the source is still reporting data from October 8, so something is likely wrong on SaskPower's end. Should I try contacting them to see if they're aware of the issue?

VIKTORVAV99:
If the bot closed the issue we got data again (it will open a new issue if it stops working again).

If you want to reach out feel free to do so but it's nothing that's expected or required.

kruschk:
I just thought it seemed odd because errors are still appearing in the runtime logs.

My last comment was wrong: the source is still updating because it now claims the data is "as of October 9, 2023", so I suppose it's just lagging by a few days. That said, the application shows data for October 10, so I'm not sure what's going on there. Looking at the parser logic, I don't see why this would be the case.

[...]

PR #6013, opened 2023-10-15
refactor: upgrade CA-SK with event classes

kruschk:
# Issue

#5986, #6011

# Description

Upgrade the parser to use the `ProductionBreakdown`, `ProductionMix`, `TotalConsumption`, and `ZoneKey` classes. This should yield no functional change. Note that this does not resolve the current issue with the `fetch_consumption` function. Additionally:

- Conform to Black's 88-column default line length limit.
- Factor out duplicate code.

# Double check

### DIFF
--- a/parsers/CA_SK.py
+++ b/parsers/CA_SK.py
@@ -1,8 +1,8 @@
-import typing
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
+from typing import Any
+from zoneinfo import ZoneInfo
 
-from pytz import timezone
 from requests import Session
 
 from electricitymap.contrib.config import ZoneKey
@@ -10,7 +10,7 @@ from electricitymap.contrib.lib.models.event_lists import ProductionBreakdownLis
 from electricitymap.contrib.lib.models.events import ProductionMix, TotalConsumption
 from parsers.lib.exceptions import ParserException
 
-TIMEZONE = timezone("America/Regina")
+TIMEZONE = ZoneInfo("America/Regina")
 PRODUCTION_URL = (
     "https://www.saskpower.com/ignitionapi/PowerUseDashboard"
     "/GetPowerUseDashboardData"
@@ -37,7 +37,7 @@ def _request(
     target_datetime: datetime | None,
     url: str,
     zone_key: ZoneKey,
-) -> typing.Any:
+) -> dict | str:
     # The source does not offer historical data, so bail out if it's requested.
     if target_datetime:
         raise ParserException("CA_SK.py", "Unable to fetch historical data", zone_key)
@@ -62,19 +62,19 @@ def fetch_production(
     session: Session | None = None,
     target_datetime: datetime | None = None,
     logger: Logger = getLogger(__name__),
-) -> list[dict[str, typing.Any]]:
+) -> list[dict[str, Any]]:
     payload = _request(session, target_datetime, PRODUCTION_URL, zone_key)
     # Date is in the format "Jan 01, 2020"
     date = datetime.strptime(payload["SupplyDataText"], "%b %d, %Y")
-    production_mix = ProductionMix(
-        **{
-            PRODUCTION_MAPPING[value["type"]]: value["totalGenerationForType"]
-            for value in payload["PowerCacheData"]["generationByType"]
-        }
-    )
+    production_mix = ProductionMix()
+    for generation_by_type in payload["PowerCacheData"]["generationByType"]:
+        production_mix.add_value(
+            PRODUCTION_MAPPING[generation_by_type["type"]],
+            generation_by_type["totalGenerationForType"],
+        )
     production_breakdown_list = ProductionBreakdownList(logger)
-    # Convert the daily average returned by the API into hourly values. This is
-    # a bit of a hack, but it's required because the back-end requires hourly
+    # Copy the daily average returned by the API into hourly values. This is a
+    # bit of a hack, but it's required because the back-end requires hourly
     # datapoints while the API only provides daily averages.
     for hour in range(24):
         production_breakdown_list.append(
@@ -91,7 +91,7 @@ def fetch_consumption(
     session: Session | None = None,
     target_datetime: datetime | None = None,
     logger: Logger = getLogger(__name__),
-) -> dict[str, typing.Any]:
+) -> dict[str, Any]:
     payload = _request(session, target_datetime, CONSUMPTION_URL, zone_key)
     # The source refreshes every 5 minutes, so we assume the current data is
     # from 5 minutes before the most recent multiple of 5 minutes.
--- a/parsers/CA_SK.py
+++ b/parsers/CA_SK.py
@@ -1,142 +1,108 @@
+import typing
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
 
 from pytz import timezone
-from requests import Response, Session
+from requests import Session
 
+from electricitymap.contrib.config import ZoneKey
+from electricitymap.contrib.lib.models.event_lists import ProductionBreakdownList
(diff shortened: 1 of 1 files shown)


parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
+import typing
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
 
 from pytz import timezone
```
VIKTORVAV99, line 5: Since we updated python we can use the zoneinfo built-in instead, which means less external dependencies (which makes my job easier 😅)

It will also fix some odd extra minutes we have been observing with pytz for some timezones.

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
...
+)
 
 
-def validate_no_datetime(target_datetime: datetime | None, zone_key) -> None:
+def _request(
+    session: Session | None,
+    target_datetime: datetime | None,
+    url: str,
+    zone_key: ZoneKey,
+) -> typing.Any:
```
VIKTORVAV99, line 40: Since we are returning json the type should be `list | dict | None`, which makes the type checking a little bit better.

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
...
-                "production": production_data,
-                "source": "saskpower.com",
-            }
+    date = datetime.strptime(payload["SupplyDataText"], "%b %d, %Y")
+    production_mix = ProductionMix(
+        **{
+            PRODUCTION_MAPPING[value["type"]]: value["totalGenerationForType"]
+            for value in payload["PowerCacheData"]["generationByType"]
+        }
+    )
```
VIKTORVAV99, line 74: This is great and simple but due to how the ProductionMix class works it would be better to use the add_value() function of the class and use that inside the loop.

This makes sure all the data sanitation and validation runs on each value and makes it more in line with how other parsers are working / will work.

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
...
-            "source": "saskpower.com",
-        }
-    ]
+    return TotalConsumption.create(
+        consumption=float(payload),
+        datetime=assumed_datetime,
+        logger=logger,
+        source="saskpower.com",
+        zoneKey=zone_key,
+    ).to_dict()
```
VIKTORVAV99, line 108: This should use the append function instead of the create function. The create function is an internal function that don't return anyting (it assings things internally).

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
+import typing
```
VIKTORVAV99, line 1: Since we only use Any we should consider importing that directly.

VIKTORVAV99 (commented):
Great start!

Thanks for doing this, a few things to consider before we can merge this though.

parsers/CA_SK.py
kruschk: I didn't want to restrict the return types because a JSON payload could feasibly be a `bool`, `float`, `int`, or `str` as well. In fact, while the production URL returns a `dict`, judging from the existing parser, the consumption source seems to return a `str` (it's hard to be sure while it's down). How about this?

```suggestion
) -> dict | str:
```

parsers/CA_SK.py
kruschk: I'm not sure how that would look; when I tried it raised an `AttributeError: 'TotalConsumption' object has no attribute 'append'`. Could you give me an example of what you intend? What I have seems to work fine if I run the consumption parser with some hard coded payload, e.g., `payload = "37"`.

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
...
+            f"Request to {url} failed. Response Code: {response.status_code}\n"
+            f"Error:\n{response.text}",
             zone_key,
         )
+    return response.json()
 
 
 def fetch_production(
-    zone_key: str = "CA-SK",
+    zone_key: ZoneKey = ZoneKey("CA-SK"),
```
kruschk, line 61: The [example parser](../../blob/master/parsers/examples/example_parser.py) uses `ZoneKey` for the type hint, but the wiki documentation still uses `str`. Should the wiki be updated?

parsers/CA_SK.py
```diff
@@ -1,142 +1,108 @@
...
     session: Session | None = None,
```
kruschk, line 62: The examples and wiki all seem to agree that a `Session` instance should be provided as the default (as shown in the suggested change), but that can cause unexpected behaviour with mutable objects because the object will only be instantiated once (when the function is defined, [per this Stack Overflow post](https://stackoverflow.com/questions/1132941)), so I'm wondering whether it's the best approach. Has this issue already been considered?

```suggestion
    session: Session = Session(),
```

parsers/CA_SK.py
VIKTORVAV99: The thing is that the new parser classes where designed with the assumption that all data should be lists or multiple data points (this works even for single datapoints) so it should looks something like:
```python
totalConsumption = TotalConsumption(logger)
totalConsumption.append(/* ... */)
return totalConsumption.to_list()
```

Once all parsers have been upgraded we plan on deprecating a dict as the return signature.

parsers/CA_SK.py
```diff
@@ -1,142 +1,107 @@
...
-    if not response.ok:
-        raise ParserException(
-            "CA_SK.py",
-            f"Failed to fetch production data. Response Code: {response.status_code}\nError:\n{response.text}",
-            zone_key,
-        )
-
-    raw_data = response.json()
+) -> list[dict[str, Any]]:
+    payload = _request(session, target_datetime, PRODUCTION_URL, zone_key)
```
VIKTORVAV99, line 65:
```suggestion
    payload = _request(session, target_datetime, PRODUCTION_URL, zone_key)
    if isinstance(payload, str):
        raise ParserException("CA_SK.py", "Expected dict but returned payload was str", zone_key)
```

Since str was added as a return type we need to check that this is a dict for type narrowing before we can access it without type errors.

VIKTORVAV99 (commented):
Some finishing touches are needed but beyond these it looks good!

parsers/CA_SK.py
```diff
@@ -1,142 +1,107 @@
...
             zone_key,
         )
+    return response.json()
 
 
 def fetch_production(
-    zone_key: str = "CA-SK",
-    session: Session | None = None,
+    zone_key: ZoneKey = ZoneKey("CA-SK"),
+    session: Session = Session(),
```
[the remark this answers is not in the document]
kruschk: Yes, I brought up that problem in an [earlier comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/6013#discussion_r1361358339) :wink:. I just changed it to agree with the documentation/examples in case it really was the desired behaviour; since it's not, I'll revert that commit. I guess the documentation should be updated accordingly. right?

parsers/CA_SK.py
kruschk: By that logic, wouldn't we also want to check for a `dict` in the other parser function? I don't think there's any real need for a runtime type check, because we already know that the consumption URL always responds with a `str` and the production URL always returns a `dict`. The type hint alludes to this fact, but doesn't make it explicit.

parsers/CA_SK.py
kruschk: I can't get your snippet to work unless I replace `TotalConsumption` with `TotalConsumptionList`, so I went with that. Let me know if that's not what you meant!

parsers/CA_SK.py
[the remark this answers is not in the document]
VIKTORVAV99: It should 😅.

VIKTORVAV99 (approved):

VIKTORVAV99 (approved):
LGTM! 🎉

PR #6015, opened 2023-10-15
Fix total consumption tooltip numbers

madsnedergaard:
## Issue

Closes https://github.com/electricitymaps/electricitymaps-contrib/issues/5973

## Description

During recent work on #4428, we accidentally changed the output used in the breakdown tooltip.

Screenshot 2023-10-15 at 22 58 54

The value highlighted in screenshot was intended to be the total amount of electricity/emissions **available** for the zone. This means production + discharge + imports.

### Preview

Screenshot 2023-10-15 at 23 00 30


web/src/features/charts/graphUtils.test.ts
```diff
@@ -93,11 +93,11 @@ describe('getElectricityProductionValue', () => {
   });
 });
 
-describe('getTotalEmissionsOrElectricity', () => {
+describe('getTotalEmissionsAvailableOrElectricity', () => {
```
VIKTORVAV99, line 96:
```suggestion
describe('getTotalEmissionsAvailableOrElectricityAvailable', () => {
```
Small nitpick

web/src/features/charts/graphUtils.ts
```diff
@@ -74,34 +74,34 @@ export const getGenerationTypeKey = (name: string): GenerationType | undefined =
...
-export function getTotalEmissions(zoneData: ZoneDetail, mixMode: Mode) {
-  const isConsumption = mixMode === Mode.CONSUMPTION;
-  if (isConsumption) {
-    if (zoneData.totalCo2Consumption === null) {
-      return Number.NaN;
-    }
-    return zoneData.totalCo2Consumption;
-  }
+/** Returns the total emissions that is available in the zone (e.g. production + discharge + imports) */
+export function getTotalEmissionsAvailable(zoneData: ZoneDetail, mixMode: Mode) {
```
VIKTORVAV99, line 91: While this function is pretty simple and easy to understand it is almost identical to the one above. Do you think it makes sense to try and combine them?

web/src/features/charts/tooltipCalculations.test.ts
```diff
@@ -112,27 +112,29 @@ const zoneDetailsData = {
...
-      production: 41_161,
-      zoneKey: 'FR',
-      storage: undefined,
-      isExport: false,
-      emissions: 211_155_930,
-      usage: 41_161,
-    };
-    expect(data).toEqual(expectedData);
+
+    expect(actual).toMatchInlineSnapshot(`
```
VIKTORVAV99, line 122: Cool, I didn't know about inline snapshot.

VIKTORVAV99 (approved):
Works great!

Nothing blocking as far as I am concerned, just a 2 small things to look over first but they are not functional changes.

PS: I'll kind of miss the "wrong way" of doing it, but I get that it was confusing for most people.

web/src/features/charts/graphUtils.ts
madsnedergaard: Haha, it used to be one, but we decided to split it as it was simply too confusing when using it to have one function for both use cases :D Also switching between the two "types" with an argument became harder to understand and reason about.

IntermittentNRG:
I'm confused. What is discharge and available electricity?

Is 100% still == consumption?

VIKTORVAV99:
> I'm confused. What is discharge and available electricity?
> 
> Is 100% still == consumption?

Discharge is storage and available energy is production + discharge + imports.

madsnedergaard:
Sorry that I didn't make things clear :)

The number here should represent the combined amount of electricity that is available within this zone. This means all the electricity produced by the zone, all that is imported and anything potentially being used from storage (batteries or pumped hydro). 

This might all be consumed by the zone, but not necessarily - [North Central Sweden](https://app.electricitymaps.com/zone/SE-SE2) is a great example as more electricity is produced than the zone consumes it self.

So the number should in the tooltip was never intended to be the total amount of electricity **consumed** by the zone (what's left after exports), but we accidentally made it so during recent changes.

Hope that makes sense, as Viktor points out there's value in both numbers, but for now we think sticking to the original intent is the least confusing path for the majority of users :)

PR #6029, opened 2023-10-19
Northern Ireland (GB-NIR) Coal Capacity Reduced

brandongalbraith:
## Issue

https://github.com/electricitymaps/electricitymaps-contrib/issues/6028 https://github.com/electricitymaps/electricitymaps-contrib/issues/6007

## Description

Kilroot power station, the last coal power plant in Northern Ireland, [has closed as of 30 September](https://www.soni.ltd.uk/how-the-grid-works/system-information/) per referenced issues. Data source in DATA_SOURCES.md reflects this capacity update.

### Double check


VIKTORVAV99:
No I don't think so, I just have had a few other things on my plate the last few days. I'm on my phone now but I'll get this merged later today or tomorrow at the latest.

Sorry about that.

VIKTORVAV99:
No need for apologies!
I haven't done a full review yet; I basically just need to check that the value match the source we have in the sources file.

VIKTORVAV99 (commented):
Okay so it looks like we are just using a PDF from SONI for the capacity data. This still lists the coal plants so either we need to find an updated version or just add a specific entry for coal pointing to the BBC article.

The last would be easiest but if we can find an updated PDF we could update all the capacities.

VIKTORVAV99:
I think we would be fine with an article about the coal plant closing for a reputable source. Then we can merge this and update the sources the next time they do a regular capacity update.

That way we can get in the updated value sooner rather than later.

How does that sound?

VIKTORVAV99:
> Ready for review when time permits 👀 👍 

Will try and do so tomorrow,otherwise it will be done on Monday!

VIKTORVAV99 (approved):
Nice!

I think we should be able to trust the BBC 😅

PR #6030, opened 2023-10-19
wfix solar layer crash from bookmark launch

tonypls:
## Issue
Opening the app from a bookmark in chrome and firefox can cause crashes

## Description
See the error below from the `src/features/weather-layers/solar/SolarLayer`
```

Error: Style is not done loading.

React ErrorBoundary Error: Style is not done loading.
```
I think we could improve this further by not loading the layer until it's ready to be loaded but this solves the issue for users effected by this now.

### Double check


madsnedergaard (approved):
LGTM, thanks for fixing it!

PR #6032, opened 2023-10-19
Update unknown emission factor in CH

pierresegonne:
## Issue

The Swiss Federal Office of Energy (SFOE) reached out to us to encourage us to re-examine our unknown emission factor for CH. 

## Description

* Apply computations detailed in https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E/edit#gid=1983757858

The gist of the problem is that there is a large mismatch between total production as reported by the SFOE and what we get from ENTSO-E (Swissgrid). 
This means that we need to take assumptions, as detailed in the sheet, to determine the makeup of the unknown production. 

The vision we will have will be different from what SFOE says as long as there is that gap between SFOE and Swissgrid. It will result in CH having dirtier production than what actually happens according to SFOE (missing ~10TWh of hydro!). But I don't think we should at this point correct the Swissgrid data.


pierresegonne:
FYI @jayaddison

pierresegonne:
Is a refresher for https://github.com/electricitymaps/electricitymaps-contrib/pull/4037#issuecomment-1761727603

pierresegonne:
@FelixDQ do you remember if `isLowCarbon`, `isRenewable` accepts different datetimes?

VIKTORVAV99:
> @FelixDQ do you remember if `isLowCarbon`, `isRenewable` accepts different datetimes? (don't have time to check before logging off for the day)

I believe they do as SE (aggregated zone) have specific values for each year for both isLowCarbon and isRenewable.

config/zones/CH.yaml
```diff
@@ -73,11 +73,30 @@ emissionFactors:
         source: Electricity Maps, 2021 average
         value: 113.07156762843732
     unknown:
-      _comment:
-        'Source ENTSOE, Swiss government and Electricity Maps: https://github.com/electricitymaps/electricitymaps-contrib/pull/2898
-        and https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E/'
-      source: 52% hydro storage, 48% unknown thermal
-      value: 288
+      - datetime: '2017-01-01'
+        _comment: 'Assumed make up of the unknown production is detailed here: https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E'
```
VIKTORVAV99, line 77: Are we sure the unknown renewables are hydro? They currently have a 24 gCO2eq/kWh.

config/zones/CH.yaml
```diff
@@ -73,11 +73,30 @@ emissionFactors:
         source: Electricity Maps, 2021 average
         value: 113.07156762843732
     unknown:
-      _comment:
-        'Source ENTSOE, Swiss government and Electricity Maps: https://github.com/electricitymaps/electricitymaps-contrib/pull/2898
-        and https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E/'
-      source: 52% hydro storage, 48% unknown thermal
-      value: 288
+      - datetime: '2017-01-01'
```
VIKTORVAV99, line 76: In the API we have data going back as far as 2015, do we have the values needed to calculate the unknown CO2eq value for 2015 and 2016 as well?

VIKTORVAV99 (commented):
Just some thought before approving.

FelixDQ:
> @FelixDQ do you remember if isLowCarbon, isRenewable accepts different datetimes? (don't have time to check before logging off for the day)

Yup, we support datetimes for all data that goes into the object that used to be the co2eqparameters.json file.

config/zones/CH.yaml
FelixDQ: In that case it defaults to the oldest date. Definition for how we parse them:
```
`n` dates are sorted in ascending order (d1, d2, ..., dn)
d1 is valid from from (epoch to d2)
d2 is valid from (d2 to d3)
dn is valid from (dn to end_of_time)
```

config/zones/CH.yaml
pierresegonne: Well no .. but that's my assumption. It's unclear what it is, we just know it's "unknown renewable"

VIKTORVAV99: I think it's fine for now as we haven't determined the exact composition. But it's likely it's not 100% hydro.

VIKTORVAV99 (approved):
LGTM! 🎉

mathilde-daugy (approved):

config/zones/CH.yaml
```diff
@@ -73,11 +73,30 @@ emissionFactors:
...
+        _comment: 'Assumed make up of the unknown production is detailed here: https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E'
+        source: SFOE, Electricity Maps
```
jayaddison, line 78: To help figure out merge conflict resolution for #6001: I _think_ this indicates that the linked spreadsheet combines data and/or discussion from SFOE with Electricity Maps' own calculations.

Also as context: the spreadsheet is an [existing `source` entry for `hydro`](https://github.com/electricitymaps/electricitymaps-contrib/blob/e53c5a41c2dfc72effd8afc50e8e356cf8049b12/config/zones/CH.yaml#L301) -- seeming to reaffirm that it's derived calculations, now with input from SFOE in addition.

Based on that, to resolve #6001 I think I'd lean towards adding a single top-level `source` entry to the zone YAML file, named `SFOE, Electricity Maps`.  I did notice that a few other zones use semi-colons to indicate that multiple sources were referenced to gather metrics.. this case seems potentially different, in that it's one document that has been produced by two groups in collaboration.

Please let me know if that's roughly on-track!

pierresegonne: > I think this indicates that the linked spreadsheet combines data and/or discussion from SFOE with Electricity Maps' own calculations.

Correct

> Also as context: the spreadsheet is an [existing source entry for hydro](https://github.com/electricitymaps/electricitymaps-contrib/blob/e53c5a41c2dfc72effd8afc50e8e356cf8049b12/config/zones/CH.yaml#L301) -- seeming to reaffirm that it's derived calculations, now with input from SFOE in addition.

Yes, this will be updated in the coming days as data is reprocessed with the new unknown emission factor. We will be recomputing the emission factor for hydro discharge as well. 

>  I did notice that a few other zones use semi-colons to indicate that multiple sources were referenced to gather metrics..

There is no single convention here, this is only used as information for users to get a sense of where the data is coming from. 

> I think I'd lean towards adding a single top-level source entry to the zone YAML

Not sure exactly what you have in mind, but be aware that we might have multiple sources for one zone (ex: one source for each emission factor)

PR #6033, opened 2023-10-20
Add US State borders to the map 🇺🇸 

tonypls:
## Description
State borders on the map 🇺🇸 

This PR adds US state borders to the map.

A new state borders geojson file contains a LineString geometry for the state borders and Point geometries for each of the state labels. 

Todo:

- [x] Create base state layer
- [x] Convert / compress to topoJSON (maybe)
- [x] Light mode styling of state borders
- [x] Dark mode styling of state borders
- [ ] Hovered zone border above state borders
- [x] State naming

### Preview

image

### Double check


VIKTORVAV99:
Tony have we considered putting this behind a button/toggle like the solar and wind layers? 

While I think this is an important feature to have I don't think it is needed for everyone and some might not want it at all.

jarek:
Nice. With the switch, we could do India, Brazil, Sweden, etc afterwards.

VIKTORVAV99:
Not sure Sweden for examinerad is needed, we don't really have states (in the same sense) and everyone is aware of the bidding zone they are in.

tonypls:
Cloudflare deployments are back up! Deployment preview here --------------------------------------------------^^^

web/src/features/map/Map.tsx
```diff
@@ -84,14 +89,19 @@ export default function MapPage(): ReactElement {
         'fill-color': '#FFFFFF',
         'fill-opacity': ['case', ['boolean', ['feature-state', 'hover'], false], 0.3, 0],
       } as mapboxgl.FillPaint,
+      statesBorder: {
+        'line-color': theme.stateBorderColor,
+        'line-width': 1.4,
+        'line-opacity': 0.9,
+        'line-dasharray': [1, 1],
+      } as mapboxgl.LinePaint,
```
VIKTORVAV99, line 97: Not sure if this comment is in the right place but would it be possible to change the border color between light and dark mode?

web/src/features/map/Map.tsx
```diff
@@ -349,11 +364,64 @@ export default function MapPage(): ReactElement {
...
+            'text-transform': 'uppercase',
+
+            'text-font': ['Poppins SemiBold'],
+          }}
+          paint={{
+            'text-color': 'white',
+            'text-halo-color': '#111827',
+            'text-halo-width': 0.5,
+            'text-halo-blur': 0.25,
+            'text-opacity': 0.9,
```
VIKTORVAV99, line 398: And same here with the names. Should they change depending on theme?

VIKTORVAV99 (approved):
Looking really good! Nice work Tony!

I have some thoughts but they should be considered non blocking and can be addressed in a follow up PR.

There is also some small miss alignments in the borders if you zoom in a lot but we should just address that later on, it's possible we need to tweak the US zone borders a little as well.

Lastly about the toggle for turning this on and off, it might not be needed anymore (this turned out better than I thought). But we should keep it in mind if we reorganize settings and such.

web/public/fonts/Poppins SemiBold/0-255.pbf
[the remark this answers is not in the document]
tonypls: Yes, it's just for the map annoyingly the map needs this strange .pbf format. I think it's quite nice to co-locate fonty related stuff but the file naming suggestion is a good point! I'll update that now :)

web/src/features/map/Map.tsx
tonypls: I think the names might be best to stay the same as they're above quite saturated dynamic colours from the zone colour whether you're in dark or light mode.

madsnedergaard (approved):
LGTM!

PR #6034, opened 2023-10-20
fix(parser): Add expected range for gas to Peru

FelixDQ:
The Peru parser has been outputting incomplete data:
image

The bounds were based on the the max/min values from the database:
Highest values:
image

Lowest values:
image


mathilde-daugy (approved):

mathilde-daugy:
Let's monitor in the future to ensure that we don't start discarding all data points in gas produciton increases or something

PR #6042, opened 2023-10-21
Upgrade IE parser to ParserFunctions

VIKTORVAV99:
## Issue

The IE parser was not using ParserFunctions

## Description

This PR upgrades the existing parser to use parser functions as well as adding a parser function for total generation.

### Double check


config/zones/GB-NIR.yaml
```diff
@@ -229,11 +229,12 @@ fallbackZoneMixes:
...
-  consumption: IE.fetch_consumption
-  consumptionForecast: IE.fetch_consumption_forecast
+  consumption: SMARTGRIDDASHBOARD.fetch_consumption
+  consumptionForecast: SMARTGRIDDASHBOARD.fetch_consumption_forecast
   price: ENTSOE.fetch_price
-  production: IE.fetch_production
-  productionPerModeForecast: IE.fetch_wind_forecasts
+  production: SMARTGRIDDASHBOARD.fetch_production
+  productionPerModeForecast: SMARTGRIDDASHBOARD.fetch_wind_forecasts
+  generationForecast: SMARTGRIDDASHBOARD.fetch_generation
```
VIKTORVAV99, line 237: Apparently we don't have a `generation` type for the parsers functions? Only `generationForecast`?

parsers/SMARTGRIDDASHBOARD.py
VIKTORVAV99, line 1: If it's easier to review I can rename this back to the original name and then just switch back before merging it.

VIKTORVAV99:
@florianscheidl adding you as a reviewer for this as I saw you where working on [ELE-3449](https://linear.app/electricitymaps/issue/ELE-3449/gb-nir), this PR will give us some additional data that could be used to guide the estimation models.

parsers/SMARTGRIDDASHBOARD.py
florianscheidl: Sounds like a good idea :)

VIKTORVAV99: We did not have any tests before 😅

config/zones/IE.yaml
florianscheidl, line 1: Can you remind me why we have fetch_generation and fetch_production? 
As far as I understand, fetch_production fetches demand (similar to fetch_consumption which fetches demand_forecast or demand based on availability). Initially I would have expected fetch_production to fetch generation but then fetch_generation would be redundant. Thank you for clarifying :)

parsers/IE.py
florianscheidl, line 1: As discussed in the previous comment, I wanted to clarify the role of fetch_production vs fetch_generation.

florianscheidl (commented):
Overall looks very nice to me :) 
I left some comments asking for clarification regarding fetch_generation and fetch_production, which confused me a bit.

config/zones/IE.yaml
VIKTORVAV99: Fetch generation only fetches the total generation, this is a singel number of the total amount of MW produced, fetch production breaks down this per mode instead (not super good at that in this parser).

VIKTORVAV99: I can rename it to fetchTotalGeneration to make it more clear.

florianscheidl: I see. I think that would be easier to understand. 
Getting back to fetch_production: why is it fetching demand data rather than generation data?

florianscheidl: Just noticed/remembered that we calculate the production based on demand instead of the available generation. Then this makes sense :)
LGTM then 🥳 

For context: when researching the issue ([ELE-3449](https://linear.app/electricitymaps/issue/ELE-3449/gb-nir)), we discovered a gap between the production value (as we currently calculate it) p0 = demand - imports - wind generation - unknown vs the reported production value p1 (logged as generation in smartgrid). This applies both for NIR and ROI. Moreover, the difference d_ROI = p0_ROI-p1_ROI was roughly equal to -d_NIR = p0_NIR-p1_NIR, so we assumed this had to do with exchanges within the island of Ireland. Right now we are in contact with EirGrid to resolve this question. So once we hear back, we might propose some changes to the parsers.

florianscheidl (approved):

config/zones/IE.yaml
VIKTORVAV99: That makes sense and I suspected as much when I saw the issue and I was hoping that using the generation data as well could help improve things.

Feel free to let me know when we hear back from them, updating the parser after these changes and added tests should be trivial.

VIKTORVAV99:
I'll rename things again and then merge this 🙂

PR #6050, opened 2023-10-25
refactor: upgrade CA-NS with event classes

kruschk:
# Issue

Refs: #6011, #6050

# Description

Upgrade the parser to use the `ExchangeList`, `ProductionBreakdownList`, `ProductionMix`, and `ZoneKey` classes. This should yield no functional change. Additionally:

- Define global constants to replace a number of local variables and literals.
- Try to use more consistent names throughout.

# Double check


parsers/CA_NS.py
```diff
@@ -1,213 +1,210 @@
...
+    "imports": (0, 0.50),
+}
+ZONE_KEY = ZoneKey("CA-NS")
+
+
+def _get_ns_info(
+    session: Session, logger: Logger
+) -> (ExchangeList, ProductionBreakdownList):
+    base_loads = session.get(LOAD_URL).json()  # Base loads in MW
+    mixes_percent = session.get(MIX_URL).json()  # Electricity breakdowns in %
```
kruschk, line 53: I've observed some interesting characteristics in the pair of JSON arrays returned by these endpoints:

1. The arrays $a$ and $b$ appear to be parallel in the sense that element $a_i$ corresponds with element $b_i$ for all $i$. I've convinced myself that this is true during testing by inspecting the timestamps. We can use this fact to our advantage by `zip`ping the two arrays together and iterating over the resulting `tuple`s instead of the current approach of iterating over one and then performing a linear search to find the corresponding entry in the other. Unfortunately, a race condition is lurking here: we could get unlucky and make one request immediately before a new datapoint comes in and the other immediately afterward, resulting in desynchronised arrays. We could address this problem by verifying that all timestamps agree between the two arrays; if they don't, I imagine we could simply raise an exception and bail out, and we'll get that datapoint on the next go round. Alternatively (additionally?), we could retry the requests a few times (likely after sleeping for a second or so) until the returned arrays do agree.
3. Each array is of length 25, not 24. The first (oldest, according to the timestamp) element of the base load array always indicates $0 \ \mathrm{MW}$, so I think we should just ignore this entry in both arrays.

kruschk: I went ahead and implemented this in a second commit!

parsers/CA_NS.py
```diff
@@ -1,213 +1,198 @@
...
+    "hydro": (0, 0.60),
+    "wind": (0, 0.55),
+    "imports": (0, 0.50),
+}
+ZONE_KEY = ZoneKey("CA-NS")
+
+
+def _get_ns_info(
+    session: Session, logger: Logger
+) -> (ExchangeList, ProductionBreakdownList):
```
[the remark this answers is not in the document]
kruschk: Whoops! :sweat_smile:

parsers/CA_NS.py
```diff
@@ -1,213 +1,198 @@
...
+    base_loads = session.get(LOAD_URL).json()  # Base loads in MW
+    mixes_percent = session.get(MIX_URL).json()  # Electricity breakdowns in %
+    if any(
+        base_load["datetime"] != mix_percent["datetime"]
+        for base_load, mix_percent in zip(base_loads, mixes_percent)
+    ):
+        raise ParserException(PARSER, "source data is out of sync", ZONE_KEY)
```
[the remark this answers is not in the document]
kruschk: Did you see [this comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/6050#discussion_r1371097592)? As far as I can tell, the two arrays should always be synchronised; if they're not, it's probably because of a race condition, and I wouldn't trust the data in them at all. I think it would be better to retry the requests rather than falling back to a search.

parsers/CA_NS.py
```diff
@@ -1,213 +1,198 @@
...
-            "wind": mix["Wind"] / 100.0,
-            "imports": mix["Imports"] / 100.0,
+        mix_fraction = {
+            "coal": mix_percent["Solid Fuel"] / 100.0,
+            "gas": (
+                mix_percent["HFO/Natural Gas"]
+                + mix_percent["CT's"]
+                + mix_percent["LM 6000's"]
+            )
+            / 100.0,
```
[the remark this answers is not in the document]
kruschk: I'm not sure what you mean. How would you split these? Which operations are unnecessary? This object maps the source's production modes (plus imports) to each of Electricity Maps' production modes (plus imports), but it only does that once for each entry in the source array.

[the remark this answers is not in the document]
VIKTORVAV99: add_value() supports adding values for the same mode multiple times. So instead of doing additions or summing things in the parsers we should just be adding the source values directly.

This ensures that the parsers are kept as simple as possible and that all the source values are correlated on an individual basis if needed.

parsers/CA_NS.py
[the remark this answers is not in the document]
VIKTORVAV99: I did but assuming the date times that are returned are always accurate that would not be a problem and it would be better for our internal system if it only raised errors on actual failures.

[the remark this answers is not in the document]
kruschk: I've never seen the two arrays disagree in practice, so I'm not sure the added complexity of a fallback search would be worthwhile. I think using retries could achieve the same benefit in a simpler way. I'll add a commit to show what I mean; maybe you'll like it!

parsers/CA_NS.py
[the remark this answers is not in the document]
kruschk: Ahh, I think I see what you're saying. I'll try to use that instead, but it might be a bit tricky to integrate it properly because of the custom validation logic in this parser.

parsers/CA_NS.py
```diff
@@ -1,213 +1,178 @@
...
+    # arrays agree or the retry limit is reached.
+    for _ in range(3):
+        base_loads = session.get(LOAD_URL).json()  # Base loads in MW
+        mixes = session.get(MIX_URL).json()  # Electricity mix breakdowns in %
+        if all(
+            base_load["datetime"] == mix["datetime"]
+            for base_load, mix in zip(base_loads, mixes)
+        ):
+            break
+        time.sleep(2)
```
VIKTORVAV99, line 37: While this is a smart move on it's own I'm not sure it works well with the backend as it already has it's own retry logic.

I have experienced this in another parser where it is the same 99.99% but when we refetched realy data one of the lists had a few hours more data which caused it to fail. By adding a fallback loop we avoid scenarios like this.

parsers/CA_NS.py
```diff
@@ -1,213 +1,182 @@
...
+        # Sanity checks: verify that reported production doesn't exceed listed
+        # capacity by a lot. In particular, we've seen error cases where hydro
+        # production ends up calculated as 900 MW which greatly exceeds known
+        # capacity of 418 MW.
+        if (
+            100 < production_mix.biomass
+            or 1300 < production_mix.coal
+            or 700 < production_mix.gas
+            or 500 < production_mix.hydro
+            or 700 < production_mix.wind
```
VIKTORVAV99, line 112: Since these can potentially be none this is throwing a bunch of type errors. We shoud use getters here instead with a fallback value or check that all are valid first.

Something like all() not None could work?

parsers/CA_NS.py
```diff
@@ -1,213 +1,182 @@
...
+            or 60 < mix["Hydro"]
+            or 50 < mix["Imports"]
+            # The validation JS reports an error when Solid Fuel (coal) is over
+            # 85%, but as far as I can tell, that can actually be a valid
+            # result, I've seen it a few times. Use 98% instead.
+            or 98 < mix["Solid Fuel"]
+            or 55 < mix["Wind"]
+            # Gas
+            or 50 < mix["HFO/Natural Gas"] + mix["CT's"] + mix["LM 6000's"]
+        ):
```
jarek, line 63: I would suggest removing these filters (I was the person who added them in the first place). They weren't accurate during last summer (datapoints were being ignored due to these filters), and that Javascript file no longer exists.

(actually this data doesn't seem to be shown anywhere on NS Power website anymore, but let's hope they keep the API endpoint up :hand_over_mouth: )

parsers/CA_NS.py
```diff
@@ -1,213 +1,182 @@
...
-            }
+        # In this source, imports are positive. In the expected result for
+        # CA-NB->CA-NS, "net" represents a flow from NB to NS, i.e., an import
+        # to NS, so the value can be used directly. Note that this API only
+        # specifies imports; when NS is exporting energy, the API returns 0.
+        exchanges.append(
+            datetime=timestamp,
+            netFlow=load * mix["Imports"] / 100,
+            source=SOURCE,
+            zoneKey=ZoneKey("CA-NB->CA-NS"),
```
jarek, line 92: Not strictly related to this refactor, but just to mention. The assumption that all imports are from NB is now problematic because Nova Scotia's interconnection to Newfoundland is now live. But I've never seen live data on its operation, the closest thing was that it amounted to 14% of NS supply in 2023H1, per https://www.nspower.ca/cleanandgreen/clean-energy "Our Energy Stats" chart

parsers/CA_NS.py
kruschk: Thanks for chiming in! Do "these filters" include the capacity filters (starting on line 103) as well, or only the percentage filters shown here?

parsers/CA_NS.py
kruschk: Hmm, that's a good catch. Do you know which linter emits that warning? I don't see it when I run `poetry run check`.

parsers/CA_NS.py
jarek: The percentage filters originally came in from that JS file, and the capacity filters were our own, I think. The capacity filters seem alright still, except I'm not sure if this endpoint reports Maritime Link import as part of the general hydro, in which case hydro could exceed the capacity. But we can leave it for now and check logs.

parsers/CA_NS.py
q--: For reference related to your comment: the daily generation reports for Newfoundland *might* be usable as a source for the interconnector data, see
https://github.com/electricitymaps/electricitymaps-contrib/issues/2541#issuecomment-1879357645

q--: Hey, look: I found a page the NSPower website that has the information you need (Maritime Link interconnector) and more, updated roughly every 5 minutes, with more granular data than the current source: [it's called Daily Report, but under the heading "Current System Conditions" you'll find "10 minute averages" of e.g. load, wind, imports and exports through various interconnectors that are at most five minutes old](https://www.nspower.ca/oasis/system-reports-messages/daily-report).

q--: So I dug deeper into this, and it turns out we aren't actually using this exchange data since [we get the data on the NS-NB interconnection from New Brunswick Power](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/config/exchanges/CA-NB_CA-NS.yaml), not from NSPower. So it might be best to just delete the exchanges related part.

kruschk: Hey @VIKTORVAV99, any thoughts on this? Should we simply delete the exchange parser if it's not used?

q--: To clarify, my idea was to delete the exchange parser now so we are more certain we don't accidentally fetch incorrect data, and then re-add an exchange parser based on the new source from #6317 in a separate pull request.

But doing both in this PR is also acceptable, if you've got the time for it.

kruschk: Oh, gotcha! It's not entirely clear to me what needs to be done to implement the new exchange parser (there are a lot of links to follow and I'm a bit lost), so I'll just delete it here and we can address it separately. Would those changes have any impact on the production parser?

q--: The exchange stuff shouldn't. The "[Current System Conditions](https://www.nspower.ca/oasis/system-reports-messages/daily-report)" also has wind generation, but as the page has data updated roughly every 5 minutes I think it'll be hard to integrate with the hourly data provided by the current source &ndash; and probably not that useful either, if the current source stays online.

As for implementing the exchange parser, the short summary is: from https://www.nspower.ca/oasis/system-reports-messages/daily-report, under "**Current System Conditions**", grab:
1. The timestamp, e.g. "Last Updated: 10-Jan-24 20:38:10";
2. "NS Export" for (most likely) the New Brunswick interconnector (optional, already provided by the New Brunswick parser);
3. "Maritime Link Import" for the Newfoundland interconnector.

The rest of the stuff in the issue is useful for exchange configuration and historical data.

kruschk: Awesome, thanks for clarifying! That doesn't sound as arduous as I thought; maybe I can roll it into this PR. I'll try to carve out some time to work on it this weekend.

kruschk: Alright, that wasn't so bad! A few notes:

- I created the `CA-NL-NF->CA-NS` exchange, but this is my first time doing so, so I very well could have done something wrong. In particular, I haven't tested how it will look because I don't have the front-end set up on my machine (I used Google Maps for the coordinates and a protractor for the rotation :grin:). Further, I wasn't sure whether to use `CA-NL-LB` or `CA-NL-NF` for the other zone (I chose the latter based on the physical location of the link).
- I also implemented the `CA-NB->CA-NS` exchange, but the back-end is still configured to use the NB parser for this pair.

jarek:
Can this five-month-old PR be reviewed and merged?

Current JSON endpoints aren't sending current data (https://github.com/electricitymaps/electricitymaps-contrib/issues/5587), if this persists, we would probably want to switch to the OASIS report being implemented in this PR, so having it merged would be a useful base.

jarek:
Can this five-month-old PR be reviewed and merged?

Current JSON endpoints aren't sending current data (https://github.com/electricitymaps/electricitymaps-contrib/issues/5587), if this persists, we would probably want to switch to the OASIS report being implemented in this PR, so having it merged would be a useful base.

parsers/CA_NS.py
```diff
@@ -1,212 +1,166 @@
...
+    timestamp = datetime.strptime(
+        soup.find(string="Current System Conditions").find_next("td").em.i.string,
+        "%d-%b-%y %H:%M:%S",
+    ).replace(tzinfo=ZoneInfo("America/Halifax"))
+
+    # Choose the appropriate exchange figure for the requested zone pair.
+    exchange = (
+        -float(soup.find(string="NS Export ").find_next("td").string)
+        if sorted_zone_keys == ZoneKey("CA-NB->CA-NS")
+        else float(soup.find(string="Maritime Link Import ").find_next("td").string)
```
VIKTORVAV99, line 144: We need to ensure all these soup.find() are not none before we pass them to functions that can't handle None values for type safety.

Then I think we can merge this.

kruschk: I'm not sure how we should handle this, because it essentially boils down to how the HTML is structured. We could catch errors and set `exchange = None`, for example, but if Nova Scotia Power update the HTML, we might wind up ignoring errors that actually require a parser fix. I think it's reasonable to assume the HTML has a particular structure, and if that changes, Python will let us know by raising an exception.

VIKTORVAV99: What we should do is add clear error handling and throw exceptions with clear messages if that is the case. Not rely on the whole thing crashing due to the wrong type being used.

If we add type checking to the CI, which we want to do at some point, we'd have to change this anyway.

parsers/CA_NS.py
```diff
@@ -132,17 +132,25 @@ def fetch_exchange(
...
-    )
+    try:
+        exchange = (
+            -float(soup.find(string="NS Export ").find_next("td").string)
+            if sorted_zone_keys == ZoneKey("CA-NB->CA-NS")
+            else float(soup.find(string="Maritime Link Import ").find_next("td").string)
+        )
+    except (AttributeError, TypeError):
+        logger.warning("unable to extract exchange data (error: {error})")
+        exchange = None
```
q--, line 153: I actually preferred the 'parser crashing' situation. Then you could easily see in the error log on which line the error occurred, and therefore which code needed updating without having to run and debug the parser first. (Now it can be any of the two, if I'm not mistaken.)

Perhaps this 'raise error from e' pattern I saw in another parser would work? If I'm not mistaken it might preserve the original stack trace and tell you on which line the error occurred.

https://github.com/electricitymaps/electricitymaps-contrib/blob/76b174e650103f6f9bfbf5f03810aa846022aaa7/parsers/ENTSOE.py#L1020-L1027

VIKTORVAV99: Agreed, and yes it will perserve the original stack trace which helps with debugging 🙂

kruschk: Whoops, I had botched that anyway because I hadn't bound `error` (e.g., `except ... as error:`). I like this `raise ... from error` pattern; thanks for suggesting it!

mathilde-daugy:
Hi, we will merge a quick fix today to make sure that the data stops being overwritten with 0s. This PR looks great and should be merged after as the quality of the refactoring is really good. :)

parsers/CA_NS.py
```diff
@@ -1,212 +1,179 @@
...
+        if timestamp in loads:
+            load = loads[timestamp]
         else:
-            # if not found, assume 1244 MW, based on average yearly electricity available for use
-            # in 2014 and 2015 (Statistics Canada table Table 127-0008 for Nova Scotia)
+            # If a base load corresponding with this timestamp is not found,
+            # assume 1244 MW based on the average yearly electricity available
+            # for use in 2014 and 2015 (Statistics Canada table 127-0008 for
+            # Nova Scotia).
             load = 1244
```
mathilde-daugy, line 72: We should remove this here because we are generating synthetic data and we don't keep track of it. It is better to have a None that will later be estimated

jarek: for the record, this was first added in https://github.com/electricitymaps/electricitymaps-contrib/pull/503 back in 2017

mathilde-daugy: yes, i know. We are trying to improve the overall quality of our data and this means that we want to minimize creating data points at the parser level and we want to ingest as little synthetic data as possible. If these points are missing, they will be estimated by our pipelines which will detect the gap in the time series

kruschk: When the base load is `None`, every mode in the `ProductionMix` object should be set to `None` because each one is derived from the base load. In that case, can we just skip the offending data point / event entirely?

q--: > yes, i know. We are trying to improve the overall quality of our data and this means that we want to minimize creating data points at the parser level and we want to ingest as little synthetic data as possible. If these points are missing, they will be estimated by our pipelines which will detect the gap in the time series

FYI, if you want to correct existing data, NS power has [historical data on load from 2012 on](https://www.nspower.ca/oasis/monthly-reports/hourly-total-net-nova-scotia-load). So it should be possible to correct all existing datapoints if you want to.

parsers/CA_NS.py
```diff
@@ -1,212 +1,175 @@
...
+        for load in session.get(LOAD_URL).json()[1:]
+    }
+    mixes = session.get(MIX_URL).json()[1:]  # Electricity mix breakdowns in %
+
+    production_breakdowns = ProductionBreakdownList(logger)
+    for mix in mixes:
+        timestamp = _parse_timestamp(mix["datetime"])
+        if timestamp not in loads:
+            logger.warning(
+                "unable to find base load for %s", timestamp, extra={"key": ZONE_KEY}
```
VIKTORVAV99, line 67:
```suggestion
                f"unable to find base load for {timestamp}", extra={"zone_key": ZONE_KEY}
```

parsers/CA_NS.py
```diff
@@ -1,212 +1,175 @@
...
+            )
+            continue
+        load = loads[timestamp]
+        if load <= 0:
+            logger.warning("invalid base load of %s MW", load, extra={"key": ZONE_KEY})
```
VIKTORVAV99, line 72:
```suggestion
            logger.warning(f"invalid base load of {load} MW", extra={"zone_key": ZONE_KEY})
```

parsers/CA_NS.py
```diff
@@ -1,212 +1,175 @@
...
+            or 700 < (production_mix.wind or 0)
+        ):
+            logger.warning(
+                "discarding datapoint at %s because some mode's production is "
+                "infeasible: %s",
+                timestamp,
+                production_mix,
+                extra={"key": ZONE_KEY},
+            )
+            continue
```
VIKTORVAV99, line 102: @mathilde-daugy do we still want to do this at a parser level or should we do it with the new validation layer in the backend?

VIKTORVAV99 (approved):
Looks good to me!

Thanks for bearing with me on this PR and sorry for some reviews taking longer than they should.

I'm approving this now but will let @mathilde-daugy have the final say if we should remove the in parser validation for this parser.

PS: I left some suggestions that have some tiny performance improvements as it avoid a reverse walk in the string for the value replacement.

parsers/CA_NS.py
kruschk: I typically prefer f-strings too, but I switched these `logger.warning` calls to %-formatting on the recommendation of [pylint W1203](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html).

kruschk: Sounds good! I'll apply these changes once we've received Mathilde's feedback on your other comment.

parsers/CA_NS.py
mathilde-daugy: the validation layer is not yet ready so we can keep that dor now and update when relevant

parsers/CA_NS.py
mathilde-daugy: that's great to know, this could be useful to fill in the missing data!

@kruschk when all modes are None the parser validation layer will filter out the points that have completely empty (all `None`) so there is nothing to be done in the parser actually

kruschk:
All good to merge?

PR #6053, opened 2023-10-26
[WIP] Light mode styling

tonypls:
## Description
This PR adds styling to the state borders TODO; Peer coding with @rasmusjeppsen to finalize UX and styling

### Double check

* Update unknown emission factor in CH

* Update isLowCarbon isRenewable

* Update with unknown renewable being biomass…ates' into tvs/USA-state-borders-base-branch…-state-borders…ates' into tvs/USA-state-borders-base-branch…-state-borders


madsnedergaard:
Very nice! 
I'd love to hear more thoughts on the styling you chose and why, maybe we should sit together and talk it through? :)
Personally I think it would be nice to make it the borders and labels more subtle, so the actual balancing authorities shine more and clearly stand out as the most important lines. But I might also be missing context and your thoughts on this!

A tried tweaking it a bit to better illustrate what I mean, hope it doesn't come as too extreme or too similar to Google Maps styling 😅 

## Zoomed out

### Before

Screenshot 2023-10-31 at 12 45 45

### After

Screenshot 2023-10-31 at 12 46 14


## Zoomed in

### Before

Screenshot 2023-10-31 at 12 46 56

### After

Screenshot 2023-10-31 at 12 46 47

madsnedergaard:
Another thing I just noticed is the duplicate labels, which seems quite strange 😄 

Screenshot 2023-10-31 at 12 57 15

Found this solution to fixing it, but it might require quite some work: https://github.com/mapbox/mapbox-gl-js/issues/7846

tonypls:
> Very nice! I'd love to hear more thoughts on the styling you chose and why, maybe we should sit together and talk it through? :) Personally I think it would be nice to make it the borders and labels more subtle, so the actual balancing authorities shine more and clearly stand out as the most important lines. But I might also be missing context and your thoughts on this!
> 
> A tried tweaking it a bit to better illustrate what I mean, hope it doesn't come as too extreme or too similar to Google Maps styling 😅
> 


I did experiment with the dashed line styling and I think it looks good but one problem is the state bounding boxes overlap where states meet, when dashed style is applied here the dash will appear inconsistent and look at bit off. 

This can be fixed by making the features "lines" instead of "multipolygons" but this then means we'd have to place the labels at a chosen lat lon instead of relying on the built in polygon center that is automatically computed by mapbox to place the label.

There's a tradeoff to be made and here a few options:

- Dash the lines and deal with overlapping dashing which doesn't look perfect
- Dash the lines and spend some extra time converting polygons to lines + placing labels (the current placement is dynamic based on zoom, can cause duplicate labels)
- Use the solid line and adjust transparency + line width

Yeah I think it would be good to sit down and compare what's better

tonypls:
@madsnedergaard I've merged the dashed line branch into this one and included the custom font file :)

web/src/features/map/Map.tsx
```diff
@@ -102,11 +112,15 @@ export default function MapPage(): ReactElement {
     }
 
     // An issue where the map has not loaded source yet causing map errors
-    const isSourceLoaded = map.getSource('zones-clickable') != undefined;
+    const isSourceLoaded =
+      map.getSource('zones-clickable') != undefined &&
+      map.isSourceLoaded('zones-clickable') &&
+      map.isSourceLoaded('states') &&
+      map.getSource('states') != undefined;
```
madsnedergaard, line 119: Should the order of 'states' here be opposite? E.g. first check that it is not undefined, then check if it is loaded

web/src/features/map/Map.tsx
```diff
@@ -349,11 +363,64 @@ export default function MapPage(): ReactElement {
...
+            'text-halo-color': '#111827',
+            'text-halo-width': 0.5,
+            'text-halo-blur': 0.25,
+            'text-opacity': 0.9,
+          }}
+          maxzoom={4.5}
+          minzoom={3}
+        />
+      </Source>
       <CustomLayer>
```
madsnedergaard, line 424: We should start breaking this file down into smaller modules soon 😅 
(but doesn't have to be now, just before we do additional work next time)

madsnedergaard (commented):
One minor comment, otherwise it looks good to me

web/src/features/map/Map.tsx
tonypls: Yeah for sure! I have a slack week task I've been meaning to get around to.

madsnedergaard (approved):

madsnedergaard:
Deploy preview is failing because [Cloudflare is down](https://www.cloudflarestatus.com/), just ignore this

PR #6054, opened 2023-10-26
Updates Poetry version

madsnedergaard:
## Description

As part of updating our CI setup, we need to bump `poetry` version and therefore also `requests` version.


VIKTORVAV99 (approved):
Looks like I already have been using this locally without issues and dependabot has already updated the lock file version so this should be good to go!

PR #6056, opened 2023-10-26
Fix Vietnam parser

### LINKED ISSUE #5775: Broken Vietnam parser
## Description
Vietnam parser (consumption and price) crashes when run. This happens because the currently used URLs redirect to the 404 page. Data seems to have moved to other URLs inside the same page. After investigating I've found:
- [https://www.nldc.evn.vn/api/services/app/Pages/GetChartPhuTaiVM](https://www.nldc.evn.vn/api/services/app/Pages/GetChartPhuTaiVM)
  - Contains consumption data for all subregions and whole country, but only for the present day
- [https://www.nldc.evn.vn/api/services/app/Pages/GetChartGiaBienVM](https://www.nldc.evn.vn/api/services/app/Pages/GetChartGiaBienVM)
  - Similar to previous but for price
- [https://www.nldc.evn.vn/api/services/app/Dashboard/GetBieuDoTuongQuanPT?day=24%2F08%2F2023](https://www.nldc.evn.vn/api/services/app/Dashboard/GetBieuDoTuongQuanPT?day=24%2F08%2F2023)
  - Contains consumption and price data for the whole country only, but allows historical
- [https://www.nldc.evn.vn/ThiTruongDien](https://www.nldc.evn.vn/ThiTruongDien)
  - This page contains a table with the last day production data. I link it because maybe the estimation model could use it to predict production.

The current parser seems to be designed to get historical regional data, so I ask for guidance as to what to implement in this case to fix the parser (only daily regional data, historical country data, both (don't now if backend supports this), other, etc)

silimotion:
## Issue

Closes #5775

## Description
This PR fixes the consumption and price parser for Vietnam using the endpoints explained in the issue. Unfortunately, the parser does not support the subzones. This is because the way the data is published in the live endpoint, data for 00:00 local time is never available. We can work around this using the historical data, but it is not available for subzones.

Please note that there is daily production data which could complement this consumption data also available.

### Preview

```
{'consumption': 28967.8,
  'datetime': datetime.datetime(2023, 10, 27, 5, 30, tzinfo=zoneinfo.ZoneInfo(key='Asia/Ho_Chi_Minh')),
  'source': 'nldc.evn.vn',
  'zoneKey': 'VN'}
```

### Double check


parsers/VN.py
```diff
@@ -1,216 +1,163 @@
...
+            price=data[f"giaBien{ZONE_CODE[ZoneKey('VN')]}"],
+            currency="VND",
+            zoneKey=ZoneKey("VN"),
+            source="nldc.evn.vn",
         )
+    return result_list.to_list()
+
+
+def fetch_historical_data(
+    target_datetime: datetime,
```
VIKTORVAV99, line 145: Since target_datetime can be None this need to either be updated to allow None or it needs to be handled earlier so target_datetime is never None.

parsers/VN.py
```diff
@@ -1,216 +1,163 @@
...
+    session: Session = Session(),
+):
+    res = session.get(
+        f"{HISTORICAL_DATA}?day={(target_datetime - timedelta(minutes=30)).strftime('%d/%m/%Y')}"
+    )
+    if not res.ok:
+        return ParserException
```
VIKTORVAV99, line 152: It should not return a ParserException, if it is not okay it should either `raise` a parser exception or log an error depending on how critical the error is to the parser.

parsers/VN.py
```diff
@@ -1,216 +1,163 @@
...
-    if len(result_list) == 0:
-        raise ParserException(
-            parser="VN.py", message=f"No valid prices for requested day found."
-        )
 
-    return result_list
+def fetch_live_data(data_type=str, session: Session = Session()):
+    res = session.get(LIVE_DATA[data_type])
+    if not res.ok:
+        return ParserException
```
VIKTORVAV99, line 159: Same here as the other comment.

VIKTORVAV99 (changes requested):

parsers/VN.py
```diff
@@ -1,216 +1,167 @@
...
+def fetch_consumption(
+    zone_key: ZoneKey,
+    session: Session = Session(),
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+):
+    if target_datetime is None:
+        live_consumption = fetch_live_consumption(session, logger)
+        # Live data is empty at 00:00, resort to historical from previous day (which includes 00:00)
+        if len(live_consumption) > 0:
```
VIKTORVAV99, line 40: Since this supports the subzones as well we should collect the data going forward even if it don't have historical data.

The same goes for price.

parsers/VN.py
```diff
@@ -1,216 +1,167 @@
...
+    if not res.ok:
         raise ParserException(
-            parser="VN.py", message=f"No valid prices for requested day found."
+            parser="VN.py", message=f"Request failed: {res.status_code}"
         )
+    return json.loads(res.text)["result"]["data"]["phuTai"]
+
 
-    return result_list
+def fetch_live_data(data_type=str, session: Session = Session()):
```
VIKTORVAV99, line 158:
```suggestion
def fetch_live_data(data_type:str, session: Session):
```

The session should probably be set further up?

parsers/VN.py
```diff
@@ -1,216 +1,167 @@
...
-def fetch_target_data(data_type: str, target_datetime, session: Session):
-    """
-    Fetch the day containing the given target_datetime.
-    """
 
-    # Convert timezone, and get the day
-    target_datetime_vn = arrow.get(target_datetime).to(tz)
+def fetch_consumption(
+    zone_key: ZoneKey,
+    session: Session = Session(),
```
VIKTORVAV99, line 33:
```suggestion
    session: Session | None,
```
Due to python weirdness we should not set defualt object like this.

it should be set just below as:
`session = session or Session()`

That goes for all the top level functions, then we should just pass on the session.

parsers/VN.py
```diff
@@ -1,216 +1,167 @@
...
+        live_consumption = fetch_live_consumption(session, logger)
+        # Live data is empty at 00:00, resort to historical from previous day (which includes 00:00)
+        if len(live_consumption) > 0:
+            return live_consumption
+        else:
+            return fetch_historical_consumption(
+                datetime.now().astimezone(tz) - timedelta(days=1), session, logger
+            )
+    else:
+        return fetch_historical_consumption(target_datetime, session, logger)
```
VIKTORVAV99, line 47: Why is there a astimezone on the first statement but not the seccond?

if you just want to set the timezone you can do that directly with datetime.now(tz=tz)

parsers/VN.py
```diff
@@ -1,216 +1,167 @@
...
+        live_consumption = fetch_live_price(session, logger)
+        # Live data is empty at 00:00, resort to historical from previous day (which includes 00:00)
+        if len(live_consumption) > 0:
+            return live_consumption
+        else:
+            return fetch_historical_price(
+                datetime.now().astimezone(tz) - timedelta(days=1), session, logger
+            )
+    else:
+        return fetch_historical_price(target_datetime, session, logger)
```
VIKTORVAV99, line 66: Same here.

parsers/VN.py
silimotion: Ok, but mind that there won't be data for 00:00 to 00:30 for the subzones as that data is never published in the live data endpoint

VIKTORVAV99: That should be acceptable either, but are we sure they follow a forward looking timestamp and not a backwards looking one?

Basically are we sure that 00:00 is for 00:00 - 00:29:59 and that it's not 00:30 that is for 00:00:01 – 00:30?

silimotion: Good thing you pointed that out. I assumed that the data was forward-looking but given that the historical endpoint returns data timestamps from 00:30 to 00:00 I now believe it is backwards-looking.

Does this mean that all data returned should be substracted 30 minutes to match internal Electricity Maps data format?

VIKTORVAV99: It does, a simple -timedelta(minutes=30) should do the trick. Just make sure to add a comment about it.

VIKTORVAV99 (approved):
Thanks for fixing this!

And sorry for it taking so long to get this merged!

VIKTORVAV99:
Yeah I noticed that too, haven't really had the time to look into it but I'll try and do so before the end of the week.

VIKTORVAV99:
Found the issue, and it's pretty simple to fix, just a case of a missing config in the backend. But I think I'll wait a bit with that until we add local overrides for the unknown emission factor. Should make the data a bit more accurate. Hopefully we can get that done soon but the whole team is quite busy this week.

PR #6058, opened 2023-10-27
Capacity parser for EMBER

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
```

## Description

This PR includes a capacity parser for all EMBER zones

### Preview
image

### Double check


pyproject.toml
```diff
@@ -37,6 +37,7 @@ openpyxl = {version="^3.1.2", optional=true}
 pydataxm = {version="^0.3.2", optional=true}
 ruamel-yaml = "^0.17.24"
 odfpy = {version = "^1.4.1", optional = true}
+pycountry = "^22.3.5"
```
VIKTORVAV99, line 40: This should probably be added as an optional dependency in the parser group to match the others? (or even a new capacity parser group)

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+from datetime import datetime
+
+import pandas as pd
+import pycountry
+from bs4 import BeautifulSoup
+from requests import Response, Session
+
+from electricitymap.contrib.config import ZoneKey
+
+""" Collects capacity data from the yearly electricity data from Ember"""
```
wobniarin, line 11: Would it make sense to include the link to the readme file or documentation you are writing for the capacity updates?

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+    df_capacity = df_capacity.rename(
+        columns={
+            "country_code_iso2": "zone_key",
+            "Year": "datetime",
+            "Variable": "mode",
+            "Value": "value",
+        }
+    )
+    df_capacity["datetime"] = df_capacity["datetime"].apply(lambda x: datetime(x, 1, 1))
+    df_capacity["value"] = df_capacity["value"] * 1000
```
wobniarin, line 127: I would specify that we do this to get MW

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+def fetch_production_capacity(target_datetime: datetime, zone_key: ZoneKey) -> dict:
+    all_capacity = fetch_production_capacity_for_all_zones(target_datetime)
+    if zone_key in all_capacity:
+        zone_capacity = all_capacity[zone_key]
+        print(
+            f"Fetched capacity for {zone_key} in {target_datetime.year}: \n {zone_capacity}"
+        )
+        return zone_capacity
+    else:
+        raise ValueError(f"No capacity data for {zone_key} in {target_datetime.year}")
```
wobniarin, line 172: Same as with IRENA, we might need to standardise the way we handle exceptions to all the capacity parsers. Do we want to raise an Exception or we want to log it as a warning?

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+def get_capacity_dict_from_df(df_capacity: pd.DataFrame) -> dict:
+    all_capacity = {}
+    for zone in df_capacity.index.unique():
+        df_zone = df_capacity.loc[zone]
+        zone_capacity = {}
+        for i, data in df_zone.iterrows():
+            mode_capacity = {}
+            mode_capacity["datetime"] = data["datetime"].strftime("%Y-%m-%d")
+            mode_capacity["value"] = round(float(data["value"]), 0)
+            mode_capacity["source"] = "Ember, Yearly electricity data"
```
unitrium, line 149: As with other parsers, I think it's better to have this as a top level constant rather than hidden in the code.

unitrium, line 149:
```suggestion
            mode_capacity["source"] = SOURCE
```

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+    "Coal": "coal",
+    "Gas": "gas",
+    "Hydro": "hydro",
+    "Nuclear": "nuclear",
+    "Other Fossil": "unknown",  # mostly oil it seems
+    "Other Renewables": "unknown",
+    "Solar": "solar",
+    "Wind": "wind",
+}
+
```
unitrium, line 25:
```suggestion
SOURCE = "https://ember-climate.org/, Ember, Yearly electricity data"
```

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+    "KW": {"Other Fossil": "oil"},
+    "MN": {"Other Fossil": "coal"},
+    "SG": {"Other Fossil": "coal"},
+    "SV": {"Other Fossil": "oil"},
+    "TR": {"Other Fossil": "oil", "Other Renewables": "geothermal"},
+    "TW": {"Other Fossil": "oil"},
+    "ZA": {"Other Fossil": "oil"},
+}
+
+EMBER_ZONES = [
```
unitrium, line 40: Didn't we also talked about making this part of the yaml config ?

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+    "ZA",
+]
+
+
+def map_variable_to_mode(row: pd.Series) -> pd.DataFrame:
+    zone = row["zone_key"]
+    variable = row["mode"]
+    if zone in SPECIFIC_MODE_MAPPING:
+        if variable in SPECIFIC_MODE_MAPPING[zone]:
+            row["mode"] = SPECIFIC_MODE_MAPPING[zone][variable]
```
unitrium, line 80: Usually it's not a good practice to update mutable objects like this especially if you are using it as a variable.
It can create unexpected side effects and also for the viewer it's hard to follow what is happening.

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+    "SV",
+    "TH",
+    "TR",
+    "TW",
+    "UY",
+    "ZA",
+]
+
+
+def map_variable_to_mode(row: pd.Series) -> pd.DataFrame:
```
unitrium, line 75: Are you sure this is a Series ? Looks more like a dataframe ?

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+        else:
+            row["mode"] = EMBER_VARIABLE_TO_MODE[variable]
+    else:
+        row["mode"] = EMBER_VARIABLE_TO_MODE[variable]
+    return row
+
+
+def get_data_from_url(session: Session) -> pd.DataFrame:
+    base_url = "https://ember-climate.org"
+    yearly_catalogue_url = base_url + "/data-catalogue/yearly-electricity-data/"
```
unitrium, line 90: How about merging those two together and maybe putting them at the top level for better readability ?

unitrium: Ah ok maybe not, since you are using the base url in different places, I would maybe still put the base url as a constant at top level.

electricitymap/capacity_parsers/EMBER.py
unitrium: Or maybe the return is wrong ?

electricitymap/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,176 @@
...
+            "country_code_iso2": "zone_key",
+            "Year": "datetime",
+            "Variable": "mode",
+            "Value": "value",
+        }
+    )
+    df_capacity["datetime"] = df_capacity["datetime"].apply(lambda x: datetime(x, 1, 1))
+    df_capacity["value"] = df_capacity["value"] * 1000  # convert from GW to MW
+
+    df_capacity = df_capacity.apply(map_variable_to_mode, axis=1)
```
unitrium, line 129: I think a more lean way to do this would be to create a new column named mapped mode which is created by applying on each row your operation and returning the mapped mode.

unitrium (commented):
A few comments, feel free to request additional help or reviews :)

electricitymap/capacity_parsers/EMBER.py
mathilde-daugy: yes we did but i've not updated any config yet, this is just a way of keeping track of which zones are updated using this parser

electricitymap/contrib/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,182 @@
...
+    "TH",
+    "TR",
+    "TW",
+    "UY",
+    "ZA",
+]
+
+
+def map_variable_to_mode(data: pd.Series) -> str:
+    data = data.copy()
```
unitrium, line 77: I don't think you need to do a copy since you're not modifying the data.

electricitymap/contrib/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,182 @@
...
+    zone = data["zone_key"]
+    variable = data["variable"]
+    if zone in SPECIFIC_MODE_MAPPING:
+        if variable in SPECIFIC_MODE_MAPPING[zone]:
+            mode = SPECIFIC_MODE_MAPPING[zone][variable]
```
unitrium, line 82:
```suggestion
            return SPECIFIC_MODE_MAPPING[zone][variable]
```
One final code style suggestion would be to reduce the amount of else, by directly returning :)

electricitymap/contrib/capacity_parsers/EMBER.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
```
unitrium, line 83: The last two else are the same no ? So we could combine them ?

PR #6059, opened 2023-10-27
Create EIA capacity parser

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
```

## Description

This PR includes a capacity parser for all EIA zones

### Preview
image

### Double check


electricitymap/contrib/capacity_parsers/EIA.py
```diff
@@ -0,0 +1,100 @@
...
+from parsers.EIA import REGIONS
+from parsers.lib.utils import get_token
+
+logger = getLogger(__name__)
+
+CAPACITY_URL = "https://api.eia.gov/v2/electricity/operating-generator-capacity/data/?frequency=monthly&data[0]=nameplate-capacity-mw&facets[balancing_authority_code][]={}"
+API_KEY = get_token("EIA_KEY")
+
+US_ZONES = {key: value for key, value in REGIONS.items() if key.startswith("US-")}
+TECHNOLOGY_TO_MODE = {
```
pierresegonne, line 19: Can't we have that from the normal EIA parser?

electricitymap/contrib/capacity_parsers/EIA.py
```diff
@@ -0,0 +1,100 @@
...
+from electricitymap.contrib.config import ZoneKey
+from parsers.EIA import REGIONS
+from parsers.lib.utils import get_token
+
+logger = getLogger(__name__)
+
+CAPACITY_URL = "https://api.eia.gov/v2/electricity/operating-generator-capacity/data/?frequency=monthly&data[0]=nameplate-capacity-mw&facets[balancing_authority_code][]={}"
+API_KEY = get_token("EIA_KEY")
+
+US_ZONES = {key: value for key, value in REGIONS.items() if key.startswith("US-")}
```
pierresegonne, line 18: Does the EIA report capacity for all US zones?

electricitymap/contrib/capacity_parsers/EIA.py
```diff
@@ -0,0 +1,100 @@
...
+    zone_key: ZoneKey,
+    target_datetime: datetime,
+) -> Union[Dict, None]:
+    url_prefix = CAPACITY_URL.format(REGIONS[zone_key])
+    start_date = target_datetime.strftime("%Y-%m-01")
+    end_date = target_datetime.replace(
+        day=calendar.monthrange(target_datetime.year, target_datetime.month)[1]
+    ).strftime("%Y-%m-%d")
+    url = f"{url_prefix}&api_key={API_KEY}&start={start_date}&end={end_date}&sort[0][column]=period&sort[0][direction]=desc&offset=0&length=5000"
+    r: Response = Session().get(url)
```
pierresegonne, line 79: Same comment about the session as for OPENNEM

pierresegonne (approved):

electricitymap/contrib/capacity_parsers/EIA.py
mathilde-daugy: image
It is not the same mapping as it is already aggregated for the hourly generation route. 
I did the mapping using the technology rather than the code because i think it is more comprehensive and easier to spot if i made a mistake

electricitymap/contrib/capacity_parsers/EIA.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        return capacity_dict
+    else:
+        logger.warning(
+            f"Failed to fetch capacity data for {zone_key} at {target_datetime.strftime('%Y-%m')}"
+        )
+
+
+def fetch_production_capacity_for_all_zones(
+    target_datetime: datetime, session: Session = None
```
unitrium, line 97:
```suggestion
    target_datetime: datetime, session: Session | None = None
```
You might need to fix the typing here.

electricitymap/contrib/capacity_parsers/EIA.py
```diff
@@ -0,0 +1,105 @@
...
+    df_aggregated = df.groupby(["mode"])[["nameplate-capacity-mw"]].sum().reset_index()
+    capacity_dict = {}
+    for mode in df_aggregated["mode"].unique():
+        mode_dict = {}
+        mode_dict["value"] = float(
+            df_aggregated.loc[df_aggregated["mode"] == mode][
+                "nameplate-capacity-mw"
+            ].sum()
+        )
+        mode_dict["source"] = "EIA"
```
unitrium, line 63:
```suggestion
        mode_dict["source"] = SOURCE
```
I would recommend following the same structure as with other parsers.

unitrium (approved):
A few code style suggestions but looks great!

PR #6062, opened 2023-10-28
refactor: upgrade CA-YT with event classes

kruschk:
# Issue

Refs: #6011, #6062

# Description

Upgrade the parser to use the ProductionBreakdownList, ProductionMix, StorageMix, and ZoneKey classes. This should yield no functional change. Additionally:

- Define global constants to replace magic values.
- Eliminate a few unnecessary variables.

# Double check


parsers/CA_YT.py
```diff
@@ -1,128 +1,115 @@
...
+        datetime=datetime.strptime(f"{date} {time}", "%A, %B %d, %Y %I:%M %p").replace(
+            tzinfo=TIMEZONE
+        ),
+        production=ProductionMix(
+            coal=0,
+            geothermal=0,
+            hydro=_parse_mw(soup.find("div", class_="load_hydro").div.text),
+            nuclear=0,
+            unknown=_parse_mw(thermal.text) if thermal else 0,
+        ),
```
VIKTORVAV99, line 101: Sam here as the other parser, even if this works I would prefer to user the add_value method so all parsers use the same method going forward.

parsers/CA_YT.py
```diff
@@ -1,128 +1,115 @@
...
-    }
-
-    if hydro_capacity:
-        data.update({"capacity": {"hydro": hydro_capacity}})
-
-    return data
+    if zone_key != ZONE_KEY:
+        raise ParserException("CA_YT.py", "Cannot parse zone '{zone_key}'", zone_key)
+    if target_datetime:
+        raise ParserException("CA_YT.py", "Unable to fetch historical data", zone_key)
```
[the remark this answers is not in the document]
VIKTORVAV99: Both are fine and I don't think we have a specific convention for this.

parsers/CA_YT.py
```diff
@@ -1,128 +1,119 @@
...
-
-    def parse_mw(text):
+    def _parse_mw(text):
+        """
+        Extract the power value from the source's HTML text content. The text
+        is formatted as, e.g., "37.69 MW - hydro".
+        """
         try:
-            return float(text[: text.index("MW")])
+            return float(text[: text.index(" MW")])
```
VIKTORVAV99, line 69: Was the additional space here intentional?

parsers/CA_YT.py
```diff
@@ -1,128 +1,119 @@
...
+    production_breakdowns = ProductionBreakdownList(logger=logger)
+    production_breakdowns.append(
+        datetime=datetime.strptime(f"{date} {time}", "%A, %B %d, %Y %I:%M %p").replace(
+            tzinfo=TIMEZONE
+        ),
+        production=production_mix,
+        source=SOURCE,
+        storage=StorageMix(
+            hydro=_parse_mw(hydro_capacity.div.text) if hydro_capacity else None
+        ),
```
VIKTORVAV99, line 108: Why is this added to storage mix?

This was originally added to the capacity object (this is not supported by the parser classes though)

parsers/CA_YT.py
```diff
@@ -1,128 +1,119 @@
...
 from bs4 import BeautifulSoup
 from requests import Session
 
-timezone = "America/Whitehorse"
+from electricitymap.contrib.lib.models.event_lists import ProductionBreakdownList
+from electricitymap.contrib.lib.models.events import ProductionMix, StorageMix
+from electricitymap.contrib.lib.types import ZoneKey
+from parsers.lib.exceptions import ParserException
+
+SOURCE = "www.yukonenergy.ca"
```
VIKTORVAV99, line 15: While this contained www before I think it should just be the following:

```suggestion
SOURCE = "yukonenergy.ca"
```

VIKTORVAV99 (commented):
A few things to check on before this can proceede

VIKTORVAV99 (approved):
LGTM!

Thanks for doing this!

I took the opportunity to rename the parser to match what we have discussed internally, I hope you don't mind.

parsers/CA_YT.py
kruschk: I did this intentionally so the string passed to `float` wouldn't always contain an extra space (which just gets ignored), but it's not necessary!

parsers/CA_YT.py
kruschk: Good question... I think I thought they were the same for some reason. I guess this should be removed! :grimacing:

kruschk:
I don't mind at all! Thanks for accepting this!

parsers/CA_YT.py
kruschk: I see you've fixed this in another PR, thanks!

PR #6064, opened 2023-10-29
refactor: upgrade CA-PE with event classes

kruschk:
# Issue

Refs: #6011

# Description

Upgrade the parser to use the ExchangeList, ProductionBreakdownList, ProductionMix, and ZoneKey classes. This should yield no functional change. Additionally:

- Define global constants to replace local variables and literals.
- Simplify the logic that retrieves event data from the source by transforming the response payload into a more convenient data structure.

# Double check


parsers/CA_PE.py
```diff
@@ -1,160 +1,145 @@
...
-        },
-        "storage": {},
-        "source": "princeedwardisland.ca",
-    }
-
-    return data
+            # used as peakers and back-up.
+            oil=event["fossil"],
+            wind=event["wind"],
+        ),
```
VIKTORVAV99, line 85: Could we loop over these instead? I would prefer if we used the add_value() method to keep it in line with other parsers and make sure any current and future validation is run on it.

VIKTORVAV99:
Just would like to mention that I would prefer if force-push was not used, it makes it very hard to compare changes and impossible to compare changes between commits.

Don't worry about creating a lot of small commits either, they are preferable as we squash merge all PRs anyway.

Just something to keep in mind 🙂

kruschk:
> Just would like to mention that I would prefer if force-push was not used, it makes it very hard to compare changes and impossible to compare changes between commits.

Oh, sure! Sorry about that. I was just rebasing onto `master`; I didn't make any changes to the first commit. If it helps at all, you can select the range of commits to view on the [Files Changed](https://github.com/electricitymaps/electricitymaps-contrib/pull/6064/files) tab:

image

Anyway, I'll avoid changing commit hashes going forward! :slightly_smiling_face:

> Don't worry about creating a lot of small commits either, they are preferable as we squash merge all PRs anyway.

Sounds good! If there are trivial commit messages in the PR (e.g., fixup foo, integrate review comments, etc.), do you typically remove those during the squash?

VIKTORVAV99:
> > Just would like to mention that I would prefer if force-push was not used, it makes it very hard to compare changes and impossible to compare changes between commits.
> 
> 
> 
> Oh, sure! Sorry about that. I was just rebasing onto `master`; I didn't make any changes to the first commit. If it helps at all, you can select the range of commits to view on the [Files Changed](https://github.com/electricitymaps/electricitymaps-contrib/pull/6064/files) tab:
> 
> 
> 
> image
> 
> 
> 
> Anyway, I'll avoid changing commit hashes going forward! :slightly_smiling_face:
> 
> 
> 
Yeah there are ways it's just that it kind of breaks some of the built in change detection around PR comments and don't work as well in the VS code extension.

> > Don't worry about creating a lot of small commits either, they are preferable as we squash merge all PRs anyway.
> 
> 
> 
> Sounds good! If there are trivial commit messages in the PR (e.g., fixup foo, integrate review comments, etc.), do you typically remove those during the squash?

I think we usually leave the commit messages as is but if there are multiple fix commits in a row for example I have combined them or removed them entirely on occasion.

parsers/CA_PE.py
```diff
@@ -1,160 +1,147 @@
...
-    # "Wind Power Exported Off-Island is that portion of wind generation that is supplying
-    # contracts elsewhere. The actual electricity from this portion of wind generation
-    # may stay within PEI but is satisfying a contractual arrangement in another jurisdiction."
-    # We are ignoring these paper exports, as they are an accounting/legal detail
-    # that doesn't actually reflect what happens on the wires.
-    # (New Brunswick being the only interconnection with PEI, "exporting" wind power to NB
-    # then "importing" a balance of NB electricity likely doesn't actually happen.)
-    imported_from_nb = (
-        pei_info["pei_load"] - pei_info["pei_fossil_gen"] - pei_info["pei_wind_gen"]
+        raise ParserException("CA_PE.py", "Unable to fetch historical data", ZONE_KEY)
```
VIKTORVAV99, line 103: The zone_key here should be the exchange key.

parsers/CA_PE.py
```diff
@@ -1,160 +1,147 @@
...
+        headers={"Content-Type": "application/json"},
+    ).json()["data"]
+    # Collect the production modes and power values into a lookup table.
+    modes = dict(
+        (
+            object_["data"]["header"].split(": ", maxsplit=1)
+            for object_ in event
+            if object_["type"] == "GaugeChart"
+        )
     )
```
VIKTORVAV99, line 44: It would be good if this used dict comprehension instead.

VIKTORVAV99 (commented):
Just a suggestion, otherwise this looks good.

parsers/CA_PE.py
kruschk: If I recall correctly, using a dictionary comprehension would have required duplicate code to separate the key from the value, so I did this instead.

parsers/CA_PE.py
kruschk: Uh oh, it looks like this didn't get corrected in time. Do we need another PR?

VIKTORVAV99: Ah I missed my own comment! 😅

It's probably fine but if you wanted you could add a consumption parser that takes the load data and returns that (and slip in this change).

I added it to my todo list but if you want to do it feel free to do so!

kruschk: I see you've fixed this one too, thanks again!

PR #6070, opened 2023-10-30
Capacity parser - IRENA

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
```

## Description

This PR includes a capacity parser for all IRENA zones. This data source will be used if the data from EMBER is not available or insufficient.

### Preview
from:  
image
to: 
image

### Double check


wobniarin:
It seems like linting is failing because `pycountry` is not imported, we should add it in the venv :)

electricitymap/contrib/capacity_parsers/IRENA.py
```diff
@@ -0,0 +1,146 @@
...
+    12: "geothermal",
+    13: "coal",
+    14: "oil",
+    15: "gas",
+    16: "unknown",
+    17: "nuclear",
+    18: "unknown",
+}
+
+SPECIFIC_MODE_MAPPING = {"IS": {16: "oil"}}  # unknown -> oil
```
wobniarin, line 35: Is it because IRENA returns this as unknown when we know it is oil? Maybe we could write a longer comment to explain that

electricitymap/contrib/capacity_parsers/IRENA.py
```diff
@@ -0,0 +1,146 @@
...
+                            "datetime": datetime_value,
+                            "value": value,
+                        }
+                    },
+                }
+    return capacity_dict
+
+
+def fetch_production_capacity(target_datetime: datetime, zone_key: ZoneKey) -> dict:
+    all_capacity = get_capacity_data_for_all_zones(target_datetime)
```
wobniarin, line 125: Minor suggestion: In the beginning I didn't understand why we were getting capacity data for all zones instead of just querying the zone we wanted. Now I get it. Would it make sense to include a comment that explains why we get data for all zones (the .px data has data for all zones in one file) and then we select what we want?

electricitymap/contrib/capacity_parsers/IRENA.py
```diff
@@ -0,0 +1,146 @@
...
+    zone_capacity = all_capacity[zone_key]
+
+    if zone_capacity:
+        print(
+            f"Fetched capacity for {zone_key} in {target_datetime.year}: \n{zone_capacity}"
+        )
+        return zone_capacity
+    else:
+        raise ValueError(f"No capacity data for {zone_key} in {target_datetime.year}")
```
wobniarin, line 134: Do we want the fetch capacity to crash if there is no capacity data? In the REE parser we used `logger.warning` to log it but continue with the process. Would it make sense to use the same approach across all capacity parsers?

electricitymap/contrib/capacity_parsers/IRENA.py
mathilde-daugy: there are two ways of doing it actually, we could pass the zone param in the web query and fetch zone by zone but i think this is slightly more efficitent as we only do one query and then filter it

mathilde-daugy:
> It seems like linting is failing because `pycountry` is not imported, we should add it in the venv :)

this works now, it was fixed when the ember parser got merged

electricitymap/contrib/capacity_parsers/IRENA.py
wobniarin: makes sense

wobniarin (approved):
looks good! Thanks for implementing all the feedback 🚀

PR #6071, opened 2023-10-30
Capacity parser - REE

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ree.es"
    └── value: 5233
```

## Description

This PR includes a capacity parser for all REE zones (includes ES and all ES territories)

### Preview
image

### Double check


electricitymap/contrib/capacity_parsers/REE.py
```diff
@@ -0,0 +1,97 @@
...
+    ree_capacity = {}
+    for zone in ZONE_KEY_TO_GEO_LIMIT:
+        zone_capacity = fetch_production_capacity(zone, target_datetime)
+        ree_capacity[zone] = zone_capacity
+    print(f"Fetched capacity for REE zones on {target_datetime.date()}")
+    return ree_capacity
+
+
+if __name__ == "__main__":
+    fetch_production_capacity("ES", datetime(2023, 1, 1))
```
wobniarin, line 97: Is this only for testing purposes? Maybe we don't need it in the parser itself?

electricitymap/contrib/capacity_parsers/REE.py
```diff
@@ -0,0 +1,97 @@
...
+                if mode in capacity:
+                    capacity[mode]["value"] += value
+                else:
+                    mode_capacity = {
+                        "datetime": target_datetime.strftime("%Y-%m-%d"),
+                        "value": value,
+                        "source": "ree.es",
+                    }
+                    capacity[mode] = mode_capacity
+        print(
```
wobniarin, line 75: Just as a suggestion, maybe it would make sense to include this as a `logger.info` instead of printing it? Maybe we just want to log errors and warnings 🤔

electricitymap/contrib/capacity_parsers/REE.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime
+from logging import getLogger
+
+from requests import Response, Session
+
+from electricitymap.contrib.config import ZoneKey
+
+"""Disclaimer: Capacity for the Spanish isles is only avilable per archipelago."""
```
wobniarin, line 8: We are just going to consider that all isles in one archipelago have the same capacity, isn't it?

wobniarin (approved):
LGTM 🚀 just some minor comments and questions

electricitymap/contrib/capacity_parsers/REE.py
```diff
@@ -0,0 +1,97 @@
...
+    "ES-CN-TE": "canarias",
+    "ES-CE": "ceuta",
+    "ES-ML": "melilla",
+}
+
+
+def fetch_production_capacity(zone_key: ZoneKey, target_datetime: datetime):
+    geo_limit = ZONE_KEY_TO_GEO_LIMIT[zone_key]
+    geo_ids = GEO_LIMIT_TO_GEO_IDS[geo_limit]
+    url = f"https://apidatos.ree.es/es/datos/generacion/potencia-instalada?start_date={target_datetime.strftime('%Y-01-01T00:00')}&end_date={target_datetime.strftime('%Y-12-31T23:59')}&time_trunc=year&geo_trunc=electric_system&geo_limit={geo_limit}&geo_ids={geo_ids}&tecno_select=all"
```
wobniarin, line 57: would it make sense to split this url as a base_url and all the different headers? so it is a bit more readable?

silimotion:
I have read the code and it seems to me that the data returned for the spanish islands would be incorrect. I believe that the parser will return the same data for all balearic islands and canary islands, as it uses the same url for all islands in each group. Specifically, it will return the sum of all capacity in the islands in each group.

Hope this helps!

wobniarin:
> I have read the code and it seems to me that the data returned for the spanish islands would be incorrect. I believe that the parser will return the same data for all balearic islands and canary islands, as it uses the same url for all islands in each group. Specifically, it will return the sum of all capacity in the islands in each group.
> 
> Hope this helps!

Yep, indeed. I think we might need to split the data from each archipelago to each island, since the parser itself does not return capacity data for each island.

electricitymap/contrib/capacity_parsers/REE.py
VIKTORVAV99: I would split this into the url and params and then use the params object in the request get function, much cleaner to look at since it's just dict.

electricitymap/contrib/capacity_parsers/REE.py
```diff
@@ -0,0 +1,108 @@
...
+    # "ES-CN-TE": "canarias",
+    "ES-CE": "ceuta",
+    "ES-ML": "melilla",
+}
+
+
+def fetch_production_capacity(zone_key: ZoneKey, target_datetime: datetime):
+    geo_limit = ZONE_KEY_TO_GEO_LIMIT[zone_key]
+    geo_ids = GEO_LIMIT_TO_GEO_IDS[geo_limit]
+    url = "https://apidatos.ree.es/es/datos/generacion/potencia-instalada?"
```
VIKTORVAV99, line 58:
```suggestion
    url = "https://apidatos.ree.es/es/datos/generacion/potencia-instalada"
```

wobniarin (approved):
🚀 nice job! LGTM!

PR #6073, opened 2023-10-31
Capacity parser - OPENNEM

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
```

## Description

This PR includes a capacity parser for all OPENNEM zones. Solar capacity is available for the northern territory but should be dealt with separately as conventional thermal capacity is not available here. 

### Preview
image

### Double check


electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,135 @@
...
+                "source": SOURCE,
+            }
+        capacity[zone] = zone_capacity
+    return capacity
+
+
+def fetch_production_capacity(zone_key: ZoneKey, target_datetime: datetime):
+    capacity = fetch_production_capacity_for_all_zones(target_datetime)[zone_key]
+    if capacity:
+        print(f"Updated capacity for {zone_key} in {target_datetime}: \n{capacity}")
```
pierresegonne, line 111: let's use a logger instead

electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,135 @@
...
+    "bioenergy_biogas": "biomass",
+    "gas_ccgt": "gas",
+    "gas_wcmg": "gas",
+    "gas_recip": "gas",
+    "bioenergy_biomass": "biomass",
+}
+
+
+def get_opennem_capacity_data() -> dict:
+    url = "https://api.opennem.org.au/facility/"
```
pierresegonne, line 37: I would extract that to a const

electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,135 @@
...
+    r: Response = Session().get(url)
```
pierresegonne, line 38: Should we do like for other parsers are allow passing the session as an arg? 
If we only intend to run things manually, I guess it would be fine to leave it as is. 
I'm asking a question in #engineering around the reason for passing sessions as args in the production parsers.

pierresegonne: https://electricitymaps.slack.com/archives/C01PWQYKX2R/p1698752273827549

electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,135 @@
...
+
+    capacity_df["datetime"] = capacity_df["datetime"].apply(
+        lambda x: pd.to_datetime(x).replace(
+            month=1, day=1, hour=0, minute=0, second=0, microsecond=0, tzinfo=None
+        )
+    )
+    return capacity_df
+
+
+def filter_capacity_data_by_datetime(
```
pierresegonne, line 64: +1 for testing this

pierresegonne (commented):
Looking pretty good, I'm waiting to know more about the session to approve / request changes

electricitymap/contrib/capacity_parsers/OPENNEM.py
mathilde-daugy: In the way i had planned this, we should run this manually once or twice a year so this is why i did it like that. If you think it's better the other way we can switch but it may be overkill?

pierresegonne: Thanks for updating, let's follow the same logic as in the other parsers, i.e providing at the higher level (fetch_production_capacity) session as an optional argument with `Session()` as the default value

VIKTORVAV99: I think aligning these with the other parses as much as possible makes sense, if the data sources start providing monthly updates we should probably consider automating these and run them on a schedule (not as often as the production parsers though but maybe once per week).

electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        df = df.loc[df["datetime"] <= target_datetime]
+    return df
+
+
+def fetch_production_capacity_for_all_zones(
+    target_datetime: datetime, session: Session | None = None
+):
+    if session is None:
+        session = Session()
```
pierresegonne, line 89:
```suggestion
session = session or Session()
```

electricitymap/contrib/capacity_parsers/OPENNEM.py
```diff
@@ -0,0 +1,153 @@
...
+            zone_capacity[data["mode"]] = {
+                "datetime": target_datetime.strftime("%Y-%m-%d"),
+                "value": round(data["value"], 2),
+                "source": SOURCE,
+            }
+        capacity[zone] = zone_capacity
+    return capacity
+
+
+def fetch_production_capacity(
```
pierresegonne, line 114: Same as for ONS, shouldn't we have session as an optional argument here as well?

pierresegonne (approved):
Well done, minor things so I'm approving so you're not blocked

PR #6092, opened 2023-11-02
Capacity parser - GB

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
```

## Description

This PR includes a capacity parser for GB

### Preview

image

### Double check

### DIFF
--- a/electricitymap/contrib/capacity_parsers/GB.py
+++ b/electricitymap/contrib/capacity_parsers/GB.py
@@ -1,0 +1,50 @@
+from datetime import datetime
+
+from bs4 import BeautifulSoup
+from requests import Response, Session
+
+from electricitymap.contrib.config import ZoneKey
+
+MODE_MAPPING = {
+    '"Wind Onshore"': "wind",
+    '"Wind Offshore"': "wind",
+    '"Solar"': "solar",
+    '"Other renewable"': "unknown",
+    '"Other"': "unknown",
+    '"Nuclear"': "nuclear",
+    '"Hydro Run-of-river and poundage"': "hydro",
+    '"Fossil Hard coal"': "coal",
+    '"Fossil Gas"': "gas",
+    '"Biomass"': "biomass",
+    '"Hydro Pumped Storage"': "hydro storage",
+}
+
+
+def fetch_production_capacity(zone_key:ZoneKey, target_datetime: datetime) -> dict:
+    url = f"https://www.bmreports.com/bmrs/?q=ajax/year/B1410/{target_datetime.year}/"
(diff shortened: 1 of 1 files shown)


electricitymap/contrib/capacity_parsers/GB.py
```diff
@@ -0,0 +1,58 @@
...
+SOURCE = "bmreports.com"
+
+
+def fetch_production_capacity(
+    zone_key: ZoneKey, target_datetime: datetime, session: Session
+) -> Union[Dict, None]:
+    url = f"https://www.bmreports.com/bmrs/?q=ajax/year/B1410/{target_datetime.year}/"
+    r: Response = session.get(url)
+
+    if r.status_code == 200:
```
unitrium, line 34: I would recommend inverting this.
```suggestion
    if not r.ok:
       raise
```

unitrium: The reason is that you will nest the rest of the code one level less and it is then more readable.

electricitymap/contrib/capacity_parsers/GB.py
```diff
@@ -0,0 +1,58 @@
...
+                    "datetime": target_datetime.strftime("%Y-%m-%d"),
+                    "value": int(item.find("quantity").string),
+                    "source": SOURCE,
+                }
+        logger.info(
+            f"Fetched capacity for {zone_key} on {target_datetime.date()}: \n{capacity}"
+        )
+        return capacity
+    else:
+        logger.error(f"GB: No capacity data available for year {target_datetime.year}")
```
unitrium, line 54: Maybe just raise ?

electricitymap/contrib/capacity_parsers/GB.py
```diff
@@ -0,0 +1,58 @@
...
+    '"Hydro Pumped Storage"': "hydro storage",
+}
+
+SOURCE = "bmreports.com"
+
+
+def fetch_production_capacity(
+    zone_key: ZoneKey, target_datetime: datetime, session: Session
+) -> Union[Dict, None]:
+    url = f"https://www.bmreports.com/bmrs/?q=ajax/year/B1410/{target_datetime.year}/"
```
unitrium, line 31: I would recommend putting the base url at the top.

unitrium: As a constant.

unitrium (approved):
Small improvements suggestions

PR #6093, opened 2023-11-02
Capacity parser - CL

mathilde-daugy:
## Issue

capacity
```
capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: SOURCE
    └── value: 5233
```

## Description

This PR includes a capacity parser for **CL**
**NB**: Mapping conventional thermal as unknown as the production parser data is aggregated

### Preview

image
### Double check


electricitymap/contrib/capacity_parsers/CL.py
```diff
@@ -0,0 +1,70 @@
...
+from logging import getLogger
+from typing import Dict, Union
+
+import pandas as pd
+from bs4 import BeautifulSoup
+from requests import Response, Session
+
+from electricitymap.contrib.config import ZoneKey
+
+logger = getLogger(__name__)
```
madsnedergaard, line 11: Should we set log level here? I don't see any logger output otherwise, but might be because I'm running the script in the wrong way (`poetry run python CL.py`) 😅 
```suggestion
logger = getLogger(__name__)
basicConfig(level=INFO)
```
(above suggestion requires `from logging import basicConfig, getLogger, INFO`)

electricitymap/contrib/capacity_parsers/CL.py
```diff
@@ -0,0 +1,70 @@
+from datetime import datetime
```
madsnedergaard, line 1: Should this file be renamed to only cover the specific subzone (`CL-SEN`)? Otherwise I think we should make it clear that this data is not valid for the other subzones :)

madsnedergaard (approved):
A few nitpicks, but otherwise LGTM!

And then some high-level comments that does not relate to this specific parser, but to the capacity parsers in general. If you don't mind, I'd be happy to help create some PRs to implement these changes :) 

1. Have you thought of getting more contributors to help build these parsers? Could be a great way to expand coverage and get more people involved :)
2. Don't know if this was already discussed, but I think it would make sense to have the capacity_parsers folder on top level next to /parsers, and then create a symlink like we did with parsers folder.
3. We should add a README in the folder describing what these capacity-parsers are and how to run them - maybe even update `test_parser` to support a "capacity" type similar to "production"?

Approving nonetheless as none of this is really blocking :)

PR #6094, opened 2023-11-02
EIA capacity parser - update fetch_production_capacity_for_all_zones

mathilde-daugy:
## Description

fixed return type 
update for loop to only add zones with capacity dict and not those with None


wobniarin (approved):
Maybe it would be worth adding a comment that says that the loop is to ensure that you don't append empty dict to a zone.

It looks good to me so I approve!

PR #6095, opened 2023-11-03
Moving IN-WE to a new API which has 24 hour datetime data unlike 12 hour

saurabhchatterjee23:
## Description
The older API endpoints provided data in 12 hour format without AM and PM, because of which there was quite a lot of logic in code to parse that data.

The new APIs from the same data source provide data in 24 hour format which is more correct and simple to process.

Also made changes to methods to use Pandas groupby on datetime hour which simplified the code even further and fixed some bugs which would arise in the earlier version for missing datapoints from the source

### Double check

- [X]  `poetry run test_parser IN-WE consumption`
- [X]  `poetry run test_parser "IN-NO->IN-WE" exchange`

### DIFF
--- a/parsers/IN_WE.py
+++ b/parsers/IN_WE.py
@@ -18,153 +18,144 @@ from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
+"""
+This module is used to read the exchange values on a hourly basis for a day from India West to North, East, South. 
+This module also calculates the total electricity consumption by India West.
+The data is fetched from https://www.wrldc.in using APIs which provide data in JSON format
+India West constitutes of Gujrat, Madhya Pradesh, Chattishgarh, Maharashtra, Goa, DD, DNH
+
+These APIs return the data in 24 hour format from midnight 12:00am to current time at which the parser is run  
+"""
+
+
 IN_WE_PROXY = "https://in-proxy-jfnx5klx2a-el.a.run.app"
-EXCHANGE_URL = f"{IN_WE_PROXY}/InterRegionalLinks_Data.aspx/Get_InterRegionalLinks_Region_Wise?host=https://www.wrldc.in"
-CONSUMPTION_URL = f"{IN_WE_PROXY}/OnlinestateTest1.aspx/GetRealTimeData_state_Wise?host=https://www.wrldc.in"
+IN_TZ = "Asia/Kolkata"
+
+""" 
+Exchange data parser
+source: https://www.wrldc.in/content/166_1_FlowsonInterRegionalLinks.aspx
+API : https://www.wrldc.in/InterRegionalLinks_Data.aspx/Get_InterRegionalLinks_Data
+"""
+EXCHANGE_URL = f"{IN_WE_PROXY}/InterRegionalLinks_Data.aspx/Get_InterRegionalLinks_Data?host=https://www.wrldc.in"
+
+"""
+Consumption data parser
+source: https://www.wrldc.in/content/164_1_StateScheduleVsActual.aspx
+API: https://www.wrldc.in/OnlinestateTest1.aspx/GetRealTimeData
+"""
+CONSUMPTION_URL = (
+    f"{IN_WE_PROXY}/OnlinestateTest1.aspx/GetRealTimeData?host=https://www.wrldc.in"
+)
 
 EXCHANGES_MAPPING = {
-    "WR-SR": "IN-SO->IN-WE",
-    "WR-ER": "IN-EA->IN-WE",
-    "WR-NR": "IN-NO->IN-WE",
-}
-
-KIND_MAPPING = {
-    "exchange": {"url": EXCHANGE_URL, "datetime_column": "lastUpdate"},
-    "consumption": {"url": CONSUMPTION_URL, "datetime_column": "current_datetime"},
+    "IN-SO->IN-WE": "WR-SR",
+    "IN-EA->IN-WE": "WR-ER",
+    "IN-NO->IN-WE": "WR-NR",
 }
 
 
-def get_date_range(dt: datetime):
-    return pd.date_range(
-        arrow.get(dt).floor("day").datetime,
-        arrow.get(dt).ceil("day").floor("hour").datetime,
-        freq="H",
-    ).to_pydatetime()
-
-
-def fetch_data(
-    kind: str | None = None,
-    session: Session | None = None,
-    target_datetime: datetime | None = None,
-) -> dict:
-    """- get production data from wrldc.in
-    - filter all rows with same hour as target_datetime"""
+def fetch_data(url: str, session: Session, target_datetime: datetime) -> dict:
+    """
+    get data from wrldc.in
+    """
     assert target_datetime is not None
-    assert kind is not None
 
     r = session or Session()
-    payload = {"date": target_datetime.strftime("%Y-%m-%d")}
+    payload = {"date": target_datetime.strftime("%Y-%m-%d"), "Flag": 24}
 
-    resp: Response = r.post(url=KIND_MAPPING[kind]["url"], json=payload)
+    resp: Response = r.post(url=url, json=payload)
 
     try:
         data = json.loads(resp.json().get("d", {}))
     except json.decoder.JSONDecodeError:
         raise ParserException(
             parser="IN_WE.py",
-            message=f"{target_datetime}: {kind} data is not available",
+            message=f"{target_datetime}: {url} data is not available",
         )
-
-    datetime_col = KIND_MAPPING[kind]["datetime_column"]
-    for item in data:
-        item[datetime_col] = datetime.strptime(item[datetime_col], "%Y-%d-%m %H:%M:%S")
-        dt = arrow.get(item[datetime_col])
-        if dt.second >= 30:
-            item[datetime_col] = dt.shift(minutes=1).floor("minute").datetime
-        else:
-            item[datetime_col] = dt.floor("minute").datetime
     return data
 
 
-def filter_raw_data(
-    kind: str,
-    data: dict,
-    target_datetime: datetime,
-) -> pd.DataFrame:
+def process_exchanges_data(
+    data: dict, zone_key1: str, zone_key2: str, logger: Logger
+) -> ExchangeList:
     """
-    Filters out correct datetimes (source data is 12 hour format)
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (changes requested):
Hi and thanks for the PR!

This is very hard to review as a lot of things have changed and/or been moved around. So if possible I would suggest splitting up this PR in several PRs that can be evaluated on their own.

For example this seems to remove the use of arrow, that should probably be a PR on it's own, and so would moving to the 24 hour API etc.

Otherwise it will take a long time to review this PR and introduces additional risk that we missed something that alters it's behaviour or output in a way we didn't expect.

saurabhchatterjee23:
>This is very hard to review as a lot of things have changed and/or been moved around. So if possible I would suggest splitting up this PR in several PRs that can be evaluated on their own.

I can understand that it is difficult to review all the changes, it becomes even more difficult to review if someone is trying to understand the changes by going through the diff. I feel it is much easier to understand the changes if it is clear how the code was working earlier and how it can be simplified if the data is available in 24-hour format. I will add details in the old and the newer version of the parser in the PR explaining what exactly was the logic before my change and how it changed in the newer version. I am confident that it will be easy to review after that explanation.



>For example this seems to remove the use of arrow, which should probably be a PR on its own, and so would moving to the 24 hour API etc.

I have not used `arrow` extensively before, I will read about how it is different to use arrow to convert pandas datetime instead of `to_pydatetime()`



>Otherwise it will take a long time to review this PR and introduces additional risk that we missed something that alters it's behaviour or output in a way we didn't expect.

I agree, without unit tests, it will be really difficult to review how the output will change unexpectedly in the newer version

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
-CONSUMPTION_URL = f"{IN_WE_PROXY}/OnlinestateTest1.aspx/GetRealTimeData_state_Wise?host=https://www.wrldc.in"
+"""
+This module is used to read the exchange values on a hourly basis for a day from India West to North, East, South. 
+This module also calculates the total electricity consumption by India West.
+The data is fetched from https://www.wrldc.in using APIs which provide data in JSON format
+India West constitutes of Gujrat, Madhya Pradesh, Chattishgarh, Maharashtra, Goa, DD, DNH
 
-EXCHANGES_MAPPING = {
-    "WR-SR": "IN-SO->IN-WE",
-    "WR-ER": "IN-EA->IN-WE",
```
saurabhchatterjee23, line 27: I reversed this mapping because we simply need to get records in data which match to WR-SR or WR-ER or WR-NR

- In the older version we were joining the data with this mapping to add a new column which represents sortedZoneKey
+ In the newer version instead of performing a join, I am simply filtering by region name like WR-SR, WR-ER, WR-NR

It is the same filtering logic, just easier to read

VIKTORVAV99:
> > This is very hard to review as a lot of things have changed and/or been moved around. So if possible I would suggest splitting up this PR in several PRs that can be evaluated on their own.
> 
> I can understand that it is difficult to review all the changes, it becomes even more difficult to review if someone is trying to understand the changes by going through the diff. I feel it is much easier to understand the changes if it is clear how the code was working earlier and how it can be simplified if the data is available in 24-hour format. I will add details in the old and the newer version of the parser in the PR explaining what exactly was the logic before my change and how it changed in the newer version. I am confident that it will be easy to review after that explanation.

Which is why I am suggesting splitting up this PR, it't becomes much easier to review if there is "one" change per PR and will allow us to either implement the changes or request changes a lot quicker.
> 
> > For example this seems to remove the use of arrow, which should probably be a PR on its own, and so would moving to the 24 hour API etc.
> 
> I have not used `arrow` extensively before, I will read about how it is different to use arrow to convert pandas datetime instead of `to_pydatetime()`
> 
Note I don't want to use arrow at all really, but change things with dates and times have a tendency to introduce bugs. Which is why I would prefer it as a separate PR.

> > Otherwise it will take a long time to review this PR and introduces additional risk that we missed something that alters it's behaviour or output in a way we didn't expect.
> 
> I agree, without unit tests, it will be really difficult to review how the output will change unexpectedly in the newer version

Even unit tests might not cover all scenarios, they are good for regression testing but big changes like in here can easily include things that are not covered in the tests right now.

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
+    get data from wrldc.in
+    """
     assert target_datetime is not None
-    assert kind is not None
 
     r = session or Session()
-    payload = {"date": target_datetime.strftime("%Y-%m-%d")}
+    payload = {"date": target_datetime.strftime("%Y-%m-%d"), "Flag": 24}
 
-    resp: Response = r.post(url=KIND_MAPPING[kind]["url"], json=payload)
```
saurabhchatterjee23, line 58: In the new code I am simply passing the URL as a pram instead of a type and then reading it from he KIND_MAPPING

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
         data = json.loads(resp.json().get("d", {}))
     except json.decoder.JSONDecodeError:
         raise ParserException(
             parser="IN_WE.py",
-            message=f"{target_datetime}: {kind} data is not available",
+            message=f"{target_datetime}: {url} data is not available",
         )
-
-    datetime_col = KIND_MAPPING[kind]["datetime_column"]
-    for item in data:
```
saurabhchatterjee23, line 69: This logic rounds the datetime field to nearest minute, which is not needed in the newer version where we aggregate by datetime hour

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
-
-    dt_12_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %I:%M")).datetime
-    datetime_col = KIND_MAPPING[kind]["datetime_column"]
-    filtered_data = pd.DataFrame(
-        [item for item in data if item[datetime_col].hour == dt_12_hour.hour]
-    )
-    return filtered_data
-
 
-def format_exchanges_data(
```
saurabhchatterjee23, line 98: This method `format_exchanges_data` was called for every hour generated from `get_date_range(dt: datetime):` and would give net_flow for that hour. In the newer version this is not needed by simply grouping on datetime field

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
+    - filter out the records that correspond to the zone_key
+    - convert the datetime field to contain only date and hour(remove minute and secods)
+    - set timezone to IST for datetime field
+    - average all data points grouping by target_datetime hour
+    - generate the Exchange list from the Current_Loading column
     """
-    assert len(data) > 0
-    assert kind != ""
-
-    dt_12_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %I:%M")).datetime
```
saurabhchatterjee23, line 90: `filter_raw_data` method was called for every hour and it would filter data for that target_datetime hour.

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
-    datetime_col = KIND_MAPPING[kind]["datetime_column"]
-    filtered_data = pd.DataFrame(
-        [item for item in data if item[datetime_col].hour == dt_12_hour.hour]
```
saurabhchatterjee23, line 93: There is definitely a bug here. In the older version given the older data had datetime field in 12 hour format with missing AM PM value, this will filter aggregate the data for morning and evening in the same bucket. For example, if the data has records for 2023-11-08 11:00 **AM** and 2023-11-08 11:00 **PM** it will all be considered to be in the same hour

_The older API provided data in a format which can never be correctly interpreted, improving the correctness is the primary motivation I am pushing for this change_

Because of this bug, let's say the parser is run at 2023-11-08 18:23, the source will only provide data till 18:23; however the older version of the parser would generate data in the future as well till 23:00 hour because it will mistakenly copy the data from 11 hour to 23 hour.

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
+    df = df[df["Region_Name"] == EXCHANGES_MAPPING[zone_key]]
+    df[datetime_column] = (
+        pd.to_datetime(df[datetime_column], format="%Y-%m-%d %H:%M:%S")
+        .dt.tz_localize(IN_TZ)
+        .dt.floor("h")
+    )
+    df = df.groupby(["Region_Name", datetime_column]).mean().round(3)
+    df[value_column] = -df[value_column]
 
-    if target_datetime.hour >= 12:
```
saurabhchatterjee23, line 115: This logic was to compensate for 12 hr dataformat, we do not need it in the newer version

parsers/IN_WE.py
```diff
@@ -18,153 +18,144 @@
...
-    df_exchanges = filtered_data.loc[filtered_data["zone_key"] == sortedZoneKeys]
+    df = pd.DataFrame(data, columns=["Region_Name", datetime_column, value_column])
+    df = df[df["Region_Name"] == EXCHANGES_MAPPING[zone_key]]
+    df[datetime_column] = (
+        pd.to_datetime(df[datetime_column], format="%Y-%m-%d %H:%M:%S")
+        .dt.tz_localize(IN_TZ)
+        .dt.floor("h")
+    )
+    df = df.groupby(["Region_Name", datetime_column]).mean().round(3)
+    df[value_column] = -df[value_column]
```
saurabhchatterjee23, line 104: This is the core logic, which trims the datetime to hour and generates data for every hour and then takes the mean within every hour. The logic is the same as the older version, the difference here is that I am not iterating and querying the dataframes for every hour in a separate query unlike wat we did in the older version

saurabhchatterjee23:
>Which is why I am suggesting splitting up this PR, it't becomes much easier to review if there is "one" change per PR and will allow us to either implement the changes or request changes a lot quicker.

I feel this is PR for one change to move to new API with 24 hour dateformat, just moving to the new API makes a lot of old code obsolete; I am finding it difficult to understand how to slit it into multiple PRs

I will put more thought into it

saurabhchatterjee23:
This is a smaller PR with bare minimum changes to parse 24-hour data using the same logic as in the older parser https://github.com/electricitymaps/electricitymaps-contrib/pull/6115
Let me know if you find it difficult to understand

madsnedergaard:
Closing this for now, it was a nice contribution and we appreciate your work (🙏), but given the age of this PR and the somewhat limited value this brings, we're focusing efforts on bigger projects like expanding coverage and improving data quality :)

PR #6097, opened 2023-11-03
Removes missing import

madsnedergaard:
## Issue

Follow-up to #6087, deployment breaks because of this missing import


madsnedergaard:
cc @FelixDQ. We should find a way to run Earthly with contrib during GitHub CI checks :)

Self-merging again to get deployments working.

PR #6099, opened 2023-11-03
Changes to use Node v20 (and not pin to 20.6)

madsnedergaard:
## Description

Turns out Node v20.6 has problems, so pinning more losely to just v20.

See nice explanation of why here: https://github.com/electricitymaps/electricitymaps-contrib/issues/5984#issuecomment-1792467142

### DIFF
--- a/web/Earthfile
+++ b/web/Earthfile
@@ -1,5 +1,5 @@
-VERSION 0.7
-FROM node:20.6
+VERSION --use-visited-upfront-hash-collection 0.7
+FROM node:20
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):

VIKTORVAV99:
Approved so we can get this deployed but was it not possible to use the `.node-version` file so these always stay in sync with local and the contrib CI?

madsnedergaard:
I don't see an easy way of getting the number from .node-version into Earthfile unfortunately :)

PR #6101, opened 2023-11-04
Update no-NB.json

haraldgroven:
Added new missing translations to locale `no-NB` in order to bring Norwegian UI back to latest version. 

Notes
The tool `translation-helper` skipped and didn't write several strings, eg. Sweden's regions. 
I haven't checked if URLs in locale point to correct (current) URLs.


web/public/locales/no-NB.json
```diff
@@ -705,7 +752,7 @@
       "zoneName": "Guyana"
     },
     "HK": {
-      "zoneName": "Hong Kong"
+      "zoneName": "Hong-Kong"
```
VIKTORVAV99, line 755:
```suggestion
      "zoneName": "Hongkong"
```

At least according to Wikipedia 😅

VIKTORVAV99 (commented):
Just one thing to look at as far as I can tell.

web/public/locales/no-NB.json
madsnedergaard: Seems like this was not changed, did you perhaps forget to push it or was the dashed version actually correct? :)

madsnedergaard (approved):
LGTM! :)

PR #6103, opened 2023-11-06
Fixes missing script

madsnedergaard:
## Issue

Internal build pipeline is blocked.

## Description

Adds the missing script used. 
FYI, I will work on running Earthly tests and build on PR builds here so we don't just discover these things when trying to deploy it all 😅


madsnedergaard:
Self-merging as this is blocking our CI.

PR #6105, opened 2023-11-06
Run Earthly on Actions

madsnedergaard:
## Description

We currently only get to test that [Earthly](https://earthly.dev/) builds and tests are working once we update the contrib repository in our backend repo. 

This is not great, as we might merge breaking changes to `master` here and have to create and merge new changes before we can test again. See example below of me desperately trying to fix our blocked CI/CD pipeline, because we never tested that the app could actually be built:

Screenshot 2023-11-06 at 16 24 57


VIKTORVAV99:
Hehe I was going to do the same thing 😅

madsnedergaard:
> Hehe I was going to do the same thing 😅

Haha nice! ❤️ Got it working now, will also be restructuring things a bit afterwards if possible :)

.github/workflows/CI.yml
```diff
@@ -93,10 +99,8 @@ jobs:
     name: Deploy preview
     needs: [eslint, typecheck, setup, validate_generate_files, unit_test]
     if: >-
-      (needs.setup.outputs.frontend != 'false' ||
-      needs.setup.outputs.workflows != 'false') &&
-      github.event.pull_request.head.repo.full_name == github.repository &&
-      github.ref_name != 'master'
```
VIKTORVAV99, line 99: This was added so the jobs wouldn't fail because the forks don't have access to our secrets. Would this new job not face the same issues?

madsnedergaard: Oh yes, this was just a merge conflict gone wrong - thanks! :D

madsnedergaard:
It works now, it was failing on formatting due to missing .prettierignore 🤦

VIKTORVAV99 (approved):
Let's see if it works on master!

PR #6109, opened 2023-11-07
Paint map when zones source not fully loaded

tonypls:
## Issue
The aggregate toggle was not colouring previously un-rendered zones. 

## Description
This PR removes the check for the zone source layer to be loaded in the useEffect that colours the zones. 

This fixes the bug but it can be improved a lot to reduce the # of renders and complexity of this useEffect. I'll update this in the map.tsx refactor I'm currently working on. 

### Double check

### DIFF
--- a/web/src/features/map/Map.tsx
+++ b/web/src/features/map/Map.tsx
@@ -114,7 +114,6 @@ export default function MapPage(): ReactElement {
     // An issue where the map has not loaded source yet causing map errors
     const isSourceLoaded =
       map.getSource('zones-clickable') !== undefined &&
-      map.isSourceLoaded('zones-clickable') &&
       map.getSource('states') !== undefined &&
       map.isSourceLoaded('states');
 


VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard (commented):
Hmm, didn't it work with this before? 🤔 
And wouldn't we hit risk the crashes again, as mentioned in the comment?
`    // An issue where the map has not loaded source yet causing map errors`

tonypls:
> Hmm, didn't it work with this before? 🤔 And wouldn't we hit risk the crashes again, as mentioned in the comment? ` // An issue where the map has not loaded source yet causing map errors`

The check to see if it's undefined was there but the source being loaded check was added by me in the state borders PR here: https://github.com/electricitymaps/electricitymaps-contrib/pull/6033/files#diff-c34d49e3311360a9ac86be9d6a213957e6a655f1011de1c31aed2ee96fb07207R117

madsnedergaard:
Ah great, missed that - thanks!

PR #6113, opened 2023-11-08
Add support for None values in netFlow, value, and consumption fields

VIKTORVAV99:
This pull request adds support for None values in the netFlow, value, and consumption fields of the Exchange, TotalProduction, and TotalConsumption classes. This allows for more flexibility in data input and better handling of missing data. The _none_safe_round function has also been added to safely round values that may be None.

This is more of a personal preference rather than hard requirement so I'm open to discussion about this. However this unifies how we handle values passed in append (both StorageMix and ProductionMix can already be None) and moves more logic to the classes rather than requirering it in the parsers.

If we want to go forward with it I'll create and update the tests for it.

### DIFF
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -16,6 +16,15 @@ from electricitymap.contrib.lib.types import ZoneKey
 LOWER_DATETIME_BOUND = datetime(2000, 1, 1, tzinfo=timezone.utc)
 
 
+def _none_safe_round(value: float | None, precision: int = 6) -> float | None:
+    """
+    Rounds a value to the provided precision.
+    If the value is None, it is returned as is.
+    The default precision is 6 decimal places, which gives us a precision of 1 W.
+    """
+    return None if value is None else round(value, precision)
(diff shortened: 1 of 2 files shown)


electricitymap/contrib/lib/models/events.py
```diff
@@ -382,10 +400,12 @@ def to_dict(self) -> dict[str, Any]:
 class TotalProduction(Event):
```
VIKTORVAV99, line 400: We should also consider renaming this to TotalGeneration IMO so it's less likely to be confused with the production breakdown.

mathilde-daugy:
This makes good sense to me. I think if @unitrium has no objection we can merge this np :)

mathilde-daugy (approved):

VIKTORVAV99:
Thanks, if @unitrium don't have any objections I'll create some new tests for this and update tests if needed.

(as well as fixing the edge case were some parsers returned strings that was somehow converted automatically before (unwanted behaviour))

unitrium (approved):
No problems for me!

PR #6114, opened 2023-11-08
Refactor map.tsx

tonypls:
## Issue
Map.tsx is a becoming a bit hard to read and not as extensible as we'd like.

## Description
This PR intends to make Map.tsx more readable and extensible. 

The main changes are:
- Lift the source loaded check and map reference to the top level of the file out of the useEffects. 
- Make the map colouring useEffect dependency array only contain variables that can change the map so this useEffect doesn't rely on the second order effect to the map reference
- Use the initialViewState to increase performance over the flyToZone useEffect
- Extract the states layer to it's own function

Would appreciate feedback 🙏 

### Double check


web/src/features/map/Map.tsx
```diff
@@ -371,64 +361,14 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
...
-            'text-color': 'white',
-            'text-halo-color': '#111827',
-            'text-halo-width': 0.5,
-            'text-halo-blur': 0.25,
-            'text-opacity': 0.9,
-          }}
-          maxzoom={4.5}
-          minzoom={3}
-        />
+        <StatesLayer mapStyles={mapStyles} />
```
madsnedergaard, line 371: Makes good sense to extract this to it's own component/file. How about also splitting "mapStyles" and moving the relevant styling to the new file as well?
And could we move `<Source />` too, so everything related to states is co-living in the same place? :)

web/src/features/map/map-layers/StatesLayer.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+          'text-font': ['poppins-semibold'],
+        }}
+        paint={{
+          'text-color': 'white',
+          'text-halo-color': '#111827',
+          'text-halo-width': 0.5,
+          'text-halo-blur': 0.25,
+          'text-opacity': 0.9,
+        }}
```
madsnedergaard, line 34: You could also consider extracting these (layout and styling) to a constant, to avoid this inline-style. That would also allow reusing the text layout configuration something like this:
```
layout={{
  ...textLayout,
  'text-field': ['get', 'stateId'],
}}
```

web/src/features/map/Map.tsx
```diff
@@ -104,25 +109,25 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
     [theme]
   );
 
-  const { isLoading, isSuccess, isError, data } = useGetState();
+  const { isLoading, isError, data } = useGetState();
   const mapReference = useRef<MapRef>(null);
   const { worldGeometries, statesGeometries } = useGetGeometries();
+  const map = mapReference.current?.getMap();
+  map?.touchZoomRotate.disableRotation();
+  map?.touchPitch.disable();
```
madsnedergaard, line 117: Doesn't this part need to be within a useEffect so that it runs when mapReference changed? Is it always available immediately upon rendering Map?

web/src/features/map/Map.tsx
```diff
@@ -371,64 +361,14 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
...
-      >
-        <Layer id="zones-clickable-layer" type="fill" paint={styles.zonesClickable} />
-        <Layer id="zones-hoverable-layer" type="fill" paint={styles.zonesHover} />
-        <Layer id="zones-border" type="line" paint={styles.zonesBorder} />
+      <Layer id="ocean" type="background" paint={mapStyles.ocean} />
+      <Source id={ZONE_SOURCE} promoteId={'zoneId'} type="geojson" data={worldGeometries}>
+        <Layer id="zones-clickable-layer" type="fill" paint={mapStyles.zonesClickable} />
+        <Layer id="zones-hoverable-layer" type="fill" paint={mapStyles.zonesHover} />
+        <Layer id="zones-border" type="line" paint={mapStyles.zonesBorder} />
       </Source>
```
madsnedergaard, line 369: If you want to go further, you could attempt to extract the zone-related layers to it's own component - but it might not be that easy 😅

web/src/features/map/Map.tsx
```diff
@@ -340,14 +326,18 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
     }
   };
 
+  const initialViewState =
```
madsnedergaard, line 329: This would only work after the call to backend has succeeded AFAIK (why it was previously using an effect depending on `isSuccess`). Will we in a real-world scenario actually have backend data before the map is loaded? 🤔

madsnedergaard (commented):
Nice, thanks for taking this one! I've left some comments and ideas, happy to discuss further :)

web/src/features/map/map-layers/StatesLayer.tsx
tonypls: Started doing this one but felt very styled-component-y. I think given this file is quite small inline works better to match our tailwind style of styling

web/src/features/map/Map.tsx
tonypls: Good point ✅

web/src/features/map/Map.tsx
tonypls: I think the map reference should stay the same but I've moved this to inside a useEffect so it's not called every re-render

web/src/features/map/map-layers/StatesLayer.tsx
tonypls: Actually I've gone back to your suggestion now I've moved the zone layer out and this styles to the top pattern works for all the layers 😅

web/src/features/map/Map.tsx
tonypls: Have moved this out too, what do ya think?

web/src/features/map/map-layers/StatesLayer.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Layer
+        id="state-labels-name"
+        type="symbol"
+        source="states"
+        layout={{
+          'text-field': ['get', 'stateName'],
+          ...stateLabelLayour,
+        }}
+        paint={stateLabelPaint}
+        minzoom={4.5}
```
VIKTORVAV99, line 51: Is there any risk the names and IDs are shown at the same time when the zoom level is exactly 4.5?

web/src/features/map/Map.tsx
VIKTORVAV99, line 1: I can't comment on the specific lines but the useMemo and Source imports are no longer used.

web/src/features/map/Map.tsx
```diff
@@ -371,65 +329,13 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
...
-          }}
-          maxzoom={4.5}
-          minzoom={3}
-        />
-      </Source>
+      <Layer
+        id="ocean"
+        type="background"
+        paint={{ 'background-color': theme.oceanColor }}
+      />
```
VIKTORVAV99, line 336: I would consider moving this out as well even if it don't really make a difference code wise. But just so we put all layers as their own files in map-layers.

VIKTORVAV99 (commented):
Not seeing anything that would be blocking, just a few suggestions/nitpicks.

web/src/features/map/map-layers/StatesLayer.tsx
tonypls: As far as I can tell the implementation doesn't have any overlap

madsnedergaard (approved):
Forgot to submit review -.-'

PR #6115, opened 2023-11-09
Change the IN_WE parser to consume data from a new API from the same data source

saurabhchatterjee23:
## Issue
This PR breaks down the changes in https://github.com/electricitymaps/electricitymaps-contrib/pull/6095 into smaller incremental PRs

## Description

This PR is to change the IN_WE parser to consume data from a new API from the same data source which provides data in 24-hour format using the same algorithm used in the older version

### Double check

- [X] poetry run test_parser IN-WE consumption
- [X] poetry run test_parser "IN-NO->IN-WE" exchange


saurabhchatterjee23:
I ran the older and newer version of parser at 12pm IST today.
I compared the difference between the output of the older and the newer version
https://jsoncompare.com/#!/diff/id=a6fbddbdc41dc41ab35bee2e9548cb2b&fullscreen/

The data for the first 12 hours of the day matches precisely.
The new parser doesn't generate data beyond 12 pm because there is no data available from the source, the older version did generate data in the future because of a bug in the code

parsers/IN_WE.py
```diff
@@ -108,6 +110,8 @@ def format_exchanges_data(
     filtered_data = filter_raw_data(
         kind="exchange", data=data, target_datetime=target_datetime
     )
+    if filtered_data.empty:
```
mathilde-daugy, line 113: We do not want to return 0 here and rather a `ParserException`. Returning a 0 here would propagate a "fake" value throughout our entire pipelines. We would rather have a missing datapoint that can later be estimated than a wrong value

parsers/IN_WE.py
```diff
@@ -145,16 +149,12 @@ def format_consumption_data(
         data=data,
         target_datetime=target_datetime,
     )
+    if filtered_data.empty:
```
mathilde-daugy, line 152: Same comment here

mathilde-daugy (changes requested):

mathilde-daugy:
Hi @saurabhchatterjee23, thank you for breaking up your PR in smaller chunks. It is much easier to review and to see how the changes impact the parser. 

I have made a comment about not return 0 when data is missing but other than that, this looks great!

parsers/IN_WE.py
VIKTORVAV99: It wouldn't even propagate as the new ParserClasses will reject them (showing their benefit already!) but we should avoid it anyhow as it is now producing errors instead.

parsers/IN_WE.py
```diff
@@ -67,7 +69,7 @@ def fetch_data(
 
     datetime_col = KIND_MAPPING[kind]["datetime_column"]
     for item in data:
-        item[datetime_col] = datetime.strptime(item[datetime_col], "%Y-%d-%m %H:%M:%S")
+        item[datetime_col] = datetime.strptime(item[datetime_col], "%Y-%m-%d %H:%M:%S")
         dt = arrow.get(item[datetime_col])
```
VIKTORVAV99, line 73: Since we are changing things in here anyway I would prefer if we could get rid of this usage of arrow here. Python datetimes should be able to do the same thing without much hassle.

parsers/IN_WE.py
```diff
@@ -87,10 +89,10 @@ def filter_raw_data(
     assert len(data) > 0
     assert kind != ""
 
-    dt_12_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %I:%M")).datetime
+    dt_24_hour = arrow.get(target_datetime.strftime("%Y-%m-%d %H:%M")).datetime
```
VIKTORVAV99, line 92: here as well, as well as updating the comment above about the 12 hour data format.

VIKTORVAV99 (changes requested):
I agree with @mathilde-daugy comments but I have some additional notes as well.

parsers/IN_WE.py
saurabhchatterjee23: I will do that in another PR

parsers/IN_WE.py
saurabhchatterjee23: Done

parsers/IN_WE.py
```diff
@@ -67,7 +69,7 @@ def fetch_data(
 
     datetime_col = KIND_MAPPING[kind]["datetime_column"]
     for item in data:
-        item[datetime_col] = datetime.strptime(item[datetime_col], "%Y-%d-%m %H:%M:%S")
+        item[datetime_col] = datetime.strptime(item[datetime_col], "%Y-%m-%d %H:%M:%S")
```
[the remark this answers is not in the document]
saurabhchatterjee23: Yes need this to convert String datetime in 24h to python datetime

parsers/IN_WE.py
mathilde-daugy: fair point but i would rather not create synthetic 0s even at the parser level. It is not good practice :)

saurabhchatterjee23: I cannot throw ParserException here because the exception will break the loop from where this method is called.
I did two changes to handle this properly. 

1. I have updated the get_date_range to generate time range from midnight to **hour of target_datetime**. It will avoid generating dates in the future
2. I am aware of edge cases when the source API has data missing for few hours for some reason, for those cases  the datapoints will not be added to the output

parsers/IN_WE.py
saurabhchatterjee23: Done

parsers/IN_WE.py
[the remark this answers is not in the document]
VIKTORVAV99: please do use date time.fromisoformat() instead of pandas. We strive to use native date time handling as much as possible for performance and maintainability.

parsers/IN_WE.py
VIKTORVAV99: If there really are data for future dates that should be taken a look at cause then it might mean this data is actually forecasted and then all data should be returned instead (using the consumptionForecast function) or if they are split between measured and forecasted we should do the same.

saurabhchatterjee23: From this datasource I can confirm there is no future data available, however I agree, that instead of our code deciding which all hours to generate data for from the datasource; our code must simply rely on the data and generate datapoint for hours for which data is available.

In the future, PRs, I will make changes to use **group by on hour** to aggregate data by hour, it will simply generate datapoints for hours for which any data, and skip hours for which data is not there

saurabhchatterjee23:
I will wait for these changes to be pulled in before creating the next PR, does that sound reasonable ?

VIKTORVAV99 (commented):
A few more comments that needs to be addressed first.

parsers/IN_WE.py
```diff
@@ -176,7 +173,7 @@ def fetch_exchange(
     logger: Logger = getLogger(__name__),
 ) -> list[dict[str, Any]]:
     if target_datetime is None:
-        target_datetime = arrow.utcnow().datetime
+        target_datetime = arrow.now(TZ).datetime
```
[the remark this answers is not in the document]
saurabhchatterjee23: done

saurabhchatterjee23:
> A few more comments that needs to be addressed before we can merge this.

Done

saurabhchatterjee23:
@VIKTORVAV99 Do you have any update on this ?

VIKTORVAV99:
> @VIKTORVAV99 Do you have any update on this ?

I did not see that you had made changes until you pinged me. A good practice would be to re-request a review after changes have been made (request for reviews have different notification settings and workflow integrations).

But I'll take a look at this later today if I have the time.

saurabhchatterjee23:
> > @VIKTORVAV99 Do you have any update on this ?
> 
> I did not see that you had made changes until you pinged me. A good practice would be to re-request a review after changes have been made (request for reviews have different notification settings and workflow integrations).
> 
> But I'll take a look at this later today if I have the time.

Thanks for letting me know, I am a bit new to GitHub reviews.

parsers/IN_WE.py
```diff
@@ -210,7 +209,7 @@ def fetch_consumption(
     logger: Logger = getLogger(__name__),
 ) -> list[dict[str, Any]]:
     if target_datetime is None:
-        target_datetime = arrow.utcnow().datetime
+        target_datetime = datetime.now(TZ)
```
VIKTORVAV99, line 212: Changing the timezone of this from UTC also changes the returned datetimes, I'm not sure what is correct but let's put it back to UTC for now (and in the fetch exchange function).

@mathilde-daugy do you know if this has to be in UTC or not?

mathilde-daugy: the datetimes should be localized either way, as long as it is consistent across the script then there shouldn't be an error

VIKTORVAV99: Okay, I'll create a suggestion to fix this then :)

parsers/IN_WE.py
```diff
@@ -53,7 +58,7 @@ def fetch_data(
     assert kind is not None
 
     r = session or Session()
-    payload = {"date": target_datetime.strftime("%Y-%m-%d")}
+    payload = {"date": target_datetime.strftime("%Y-%m-%d"), "Flag": 24}
```
VIKTORVAV99, line 61:
```suggestion
    payload = {
        "date": target_datetime.astimezone(tz=TZ).strftime("%Y-%m-%d"),
        "Flag": 24,
    }
```
To make sure all requests are localized.

madsnedergaard:
Closing this for now, it was a nice contribution and we appreciate your work (🙏), but given the age of this PR and the somewhat limited value this brings, we're focusing efforts on bigger projects like expanding coverage and improving data quality :)

PR #6130, opened 2023-11-10
Optimize zone configuration loading and contributors upsert

VIKTORVAV99:
Refactor generateZonesConfig.ts to optimize zone configuration loading, simplify code, and upsert subzone contributors. This PR improves the performance of the getConfig function and ensures that contributors are properly added to parrent zones.


web/scripts/generateZonesConfig.ts
```diff
@@ -40,29 +40,45 @@ const getConfig = (): CombinedZonesConfig => {
...
+    const zoneContributors = new Set<string>();
+
+    for (const subZoneName of config.subZoneNames ?? []) {
+      const subZonePath = `${basePath}/${subZoneName}.yaml`;
+      const subZoneConfig = yaml.load(fs.readFileSync(subZonePath, 'utf8')) as ZoneConfig;
+      if (subZoneConfig.contributors) {
+        for (const contributor of subZoneConfig.contributors) {
+          contributors.add(contributor);
+          zoneContributors.add(contributor);
+        }
```
VIKTORVAV99, line 57: We should consider making this async in the future as for example the US has a lot of sub zones that could be processed in async. However this would make it none deterministic and more complicated so I opted against it.

VIKTORVAV99: I had an idea that would solve and speed this up, will mark as draft and refractor.

madsnedergaard (approved):
Code makes sense, added a small unit test to verify the changes :)

VIKTORVAV99:
Not really sure why the CI is failing on this?
Is there any reason why we have a special tsconfig.node.js instead of using the normal tsconfig.js for all of the codebase?

cc @madsnedergaard

madsnedergaard:
> Not really sure why the CI is failing on this? Is there any reason why we have a special tsconfig.node.js instead of using the normal tsconfig.js for all of the codebase?
> 
> cc @madsnedergaard

Can't remember by now why we setup things like this 🤔 But hope we can soon trim things down by migrating to `bun` 😎

VIKTORVAV99:
> > Not really sure why the CI is failing on this? Is there any reason why we have a special tsconfig.node.js instead of using the normal tsconfig.js for all of the codebase?
> > cc @madsnedergaard
> 
> Can't remember by now why we setup things like this 🤔 But hope we can soon trim things down by migrating to `bun` 😎

That would be nice! The performance increases would just be a bonus at this point 😅

PR #6132, opened 2023-11-11
Update no-NB.json completed translation

haraldgroven:
This PR contains a 100% translation of strings in `no-NB.json` in sync with these in master file `en.json`. 

Translated using desktop app, since the translation tool `translation-helper` was unable to save file with all missing strings. The desktop app may have altered whitespace and ordering in json file, so I hope reviewers may run appropriate formatting tools to adhere to whatever style used in project.


web/public/locales/no-NB.json
```diff
@@ -1352,6 +1443,22 @@
...
+      "zoneName": "Norra Mellansverige (Sundsvall)"
+    },
+    "SE-SE3": {
+      "countryName": "Sverige",
+      "zoneName": "Södra Mellansverige (Stockholm)"
+    },
+    "SE-SE4": {
+      "countryName": "Sverige",
+      "zoneName": "Södra Sverige (Malmö)"
+    },
```
VIKTORVAV99, line 1461:
```suggestion
    "SE-SE1": {
      "countryName": "Sverige",
      "zoneName": "Norra Sverige"
    },
    "SE-SE2": {
      "countryName": "Sverige",
      "zoneName": "Norra Mellansverige"
    },
    "SE-SE3": {
      "countryName": "Sverige",
      "zoneName": "Södra Mellansverige"
    },
    "SE-SE4": {
      "countryName": "Sverige",
      "zoneName": "Södra Sverige"
    },
```
Since we don't use these "aliases" for the Swedish bidding zones in either the English or Swedish translation I would skip them here as well to avoid them getting truncated in the app.

web/public/locales/no-NB.json
```diff
@@ -1,270 +1,301 @@
...
-      "header": "Bytt mellom visning av utslipp og elektrisitet",
-      "text": "Bytt mellom to alternative visninger som viser en el-markedssone informasjon om enten elektrisitet eller utslipp. Verdiene som vises representerer totalen for den valgte perioden, og informasjonen er sammenlignbar med tallene for produksjon eller forbruk."
-    },
-    "view6": {
-      "header": "Utforsk historiske data",
-      "text": "Nede til venstre, kan du se både sanntids og historiske utlipps- og elektrisitetdata for verden. Du kan endre på de aggregerte dataene ved bruke bryterne for å endre på tidslinja slik at du kan se utvikling over tid."
-    }
+  "aggregateButtons": {
+    "country": "land",
+    "zone": "prisområde"
```
VIKTORVAV99, line 17: We should not use price area as that may not always be the case. This can be but are not limited to administrative areas, island boundaries and bidding zones.

So we need to make this wording more general.

web/public/locales/no-NB.json
```diff
@@ -1,270 +1,301 @@
...
-    "by-source": {
-      "emissions": "Karbonutslipp etter kilde",
-      "electricity-production": "Elektrisitetsproduksjon etter kilde",
-      "electricity-consumption": "Elektrisitetsforbruk etter kilde",
-      "total-emissions": "Totale karbonutslipp etter kilde",
-      "total-electricity-production": "Total elektrisitetsproduksjon etter kilde",
-      "total-electricity-consumption": "Total elektrisitetsforbruk etter kilde"
+  "aria": {
+    "label": {
+      "changeTheme": "Endre utseende",
```
VIKTORVAV99, line 21: Should this not be more like "Endre tema" than "Endre utseende"? As we just allow them to change theme and not individual looks?

web/public/locales/no-NB.json
```diff
@@ -1,270 +1,301 @@
...
+      "colorBlindMode": "Fargeblind-tilpasset",
+      "hideSidePanel": "Skjul sidepanel",
+      "selectLanguage": "Velg språk",
+      "showSidePanel": "Vis sidepanel",
+      "solarLayer": "Sol-lag",
+      "windLayer": "Vind-lag"
     }
   },
+  "battery storage": "Batterilagring",
+  "biomass": "Biomasse",
```
VIKTORVAV99, line 31: These should be lower case to match the English ones. (As well as the other production modes.)

VIKTORVAV99 (commented):
Thanks for the PR!

A few notes, I might have commented on things that was not directly changed by you though.

web/public/locales/no-NB.json
haraldgroven: The sub-country regions are indeed "price areas/bidding zones" in my country. 
May be generalized to "**region**" for countries where these are not "price areas/bidding zones".

VIKTORVAV99: Zone should be translated to Sone though right?

I think that would be better than region.

haraldgroven: **Region** is understood as sub-country division, and perhaps clearer. 
**Sone** have a less clear meaning.

VIKTORVAV99: > **Region** is understood as sub-country division, and perhaps clearer. **Sone** have a less clear meaning.

It does not need to be a subarea of a country though, all existing zones are but we expect that in the coming years there will be multi country hubs for electricity production in the north and Baltic Sea for example. There could also be a cross border area (Ireland could be such a option for example).

haraldgroven: Ok, maybe "**Sone**" is lowest common denominator for this geographical division.

It would have been great if it had been possible to standardise this sub-division to "prize zone" / "bidding zone" with same meaning across countries.

VIKTORVAV99: Yes, but currently it is not standardised across the world, and even if it was at the current state of our data sources it would not be possible to use it everywhere anyhow.

VIKTORVAV99:
Did you forget to push the discussed changes? I see the discussions have been marked as resolved but there are no code changes.

VIKTORVAV99 (approved):
LGTM!

Thanks for translating the new strings!

PR #6134, opened 2023-11-11
Change pytz to timezone.utc and ZoneInfo

VIKTORVAV99:
This PR updates all python code to use timezone.utc and ZoneInfo instead of pytz for better compatibility, consistency and speed.

Arrow is next 😉

### DIFF
--- a/parsers/ELEXON.py
+++ b/parsers/ELEXON.py
@@ -11,13 +11,12 @@ https://bscdocs.elexon.co.uk/guidance-notes/bmrs-api-and-data-push-user-guide
 """
 
 import re
-from datetime import date, datetime, time, timedelta
+from datetime import date, datetime, time, timedelta, timezone
 from io import StringIO
 from logging import Logger, getLogger
 
 import arrow
 import pandas as pd
-import pytz
 from requests import Session
 
 from electricitymap.contrib.config.constants import PRODUCTION_MODES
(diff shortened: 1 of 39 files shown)


parsers/ELEXON.py
```diff
@@ -11,13 +11,12 @@
 """
 
 import re
-from datetime import date, datetime, time, timedelta
+from datetime import date, datetime, time, timedelta, timezone
 from io import StringIO
```
wobniarin, line 15: More a curiosity/learning for me than a change request: Why are we using `timezone` here? Shouldn't it make sense to import ZoneInfo as we do for the other zones and get `TIMEZONE = ZoneInfo("Europe/London")` from there?

wobniarin (commented):
very straightforward! I only have a question regarding when to use `ZoneInfo` and when to use `timezone.utc`. I will approve afterwards :)

parsers/ELEXON.py
VIKTORVAV99: timezone from datetime have a singleton with a static definition for utc, this is the recommended approach by the python documentation.

VIKTORVAV99:
> very straightforward! I only have a question regarding when to use `ZoneInfo` and when to use `timezone.utc`. I will approve afterwards :)  

Basically we should use timezone.utc whenever it's UTC (only) and ZoneInfo otherwise. (As per the python documentation)

wobniarin:
> > very straightforward! I only have a question regarding when to use `ZoneInfo` and when to use `timezone.utc`. I will approve afterwards :)
> 
> Basically we should use timezone.utc whenever it's UTC (only) and ZoneInfo otherwise. (As per the python documentation)

Thanks a lot for the explanation! Are we planning on documenting it somewhere so that contributors can learn these best practices? I think we should also share them with the team!

wobniarin (approved):

VIKTORVAV99:
> > > very straightforward! I only have a question regarding when to use `ZoneInfo` and when to use `timezone.utc`. I will approve afterwards :)
> > 
> > 
> > Basically we should use timezone.utc whenever it's UTC (only) and ZoneInfo otherwise. (As per the python documentation)
> 
> Thanks a lot for the explanation! Are we planning on documenting it somewhere so that contributors can learn these best practices? I think we should also share them with the team!

I'll make sure our wiki pages are updated and I can write together a small email and send that out to engineering tomorrow.

PR #6139, opened 2023-11-14
[Performance]: bump black and flake8

VIKTORVAV99:
## Issue

CI is slower than we would like.

## Description

Both flake8 and black is used in the CI and they have both received performance updates, we should upgrade to take advantage of that.


VIKTORVAV99:
Closing this in favour of just adding ruff instead. Which is much much faster.

PR #6140, opened 2023-11-14
Make UY parser throw clearer error

pierresegonne:
## Issue

The `UY` data source is out, but the logs are quite cryptic about the error https://storage.googleapis.com/electricitymap-parser-logs/UY.html 

## Description

* Raise `ParserException` when the data to fetch the url can't be accessed. 

### Double check

### DIFF
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -110,7 +110,7 @@ def fetch_production(
 ) -> list:
     """collects production data from ADME and format all data points for target_datetime"""
     if target_datetime is None:
-        target_datetime = datetime.now().replace(tzinfo=UY_TZ)
+        target_datetime = datetime.now(tz=UY_TZ)
     session = session or Session()
 
     data = fetch_data(
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -56,11 +56,16 @@ def get_adme_url(target_datetime: datetime, session: Session) -> str:
         if tag.button is not None:
             if tag.button.string == "Archivo Scada Detalle 10minutal":
                 data_url = "https://pronos.adme.com.uy" + tag.get("href")
+
+    if not data_url:
+        raise ParserException(
+            parser="UY.py",
+            message="Impossible to fetch data url from ADME",
+        )
     return data_url
 
 
 def fetch_data(
-    zone_key: str,
     session: Session,
     target_datetime: datetime,
     sheet_name: str,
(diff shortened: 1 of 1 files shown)


parsers/UY.py
```diff
@@ -149,11 +153,10 @@ def fetch_consumption(
 ) -> list:
     """collects consumption data from ADME and format all data points for target_datetime"""
     if target_datetime is None:
-        target_datetime = arrow.utcnow().replace(tzinfo=UY_TZ).datetime
+        target_datetime = datetime.now().replace(tzinfo=UY_TZ)
```
pierresegonne, line 156: Tried to get rid of arrow, but could not replace all of it, as I don't know what's the format of the string returned by the source where `arrow.get` is used

parsers/UY.py
```diff
@@ -105,11 +110,10 @@ def fetch_production(
 ) -> list:
     """collects production data from ADME and format all data points for target_datetime"""
     if target_datetime is None:
-        target_datetime = arrow.utcnow().replace(tzinfo=UY_TZ).datetime
+        target_datetime = datetime.now().replace(tzinfo=UY_TZ)
```
VIKTORVAV99, line 113:
```suggestion
        target_datetime = datetime.now(tz=UY_TZ)
```

parsers/UY.py
VIKTORVAV99, line 156:
```suggestion
        target_datetime = datetime.now(tz=UY_TZ)
```

parsers/UY.py
```diff
@@ -184,11 +187,10 @@ def fetch_exchange(
 ) -> list:
     """collects exchanges data from ADME and format all data points for target_datetime"""
     if target_datetime is None:
-        target_datetime = arrow.utcnow().replace(tzinfo=UY_TZ).datetime
+        target_datetime = datetime.now().replace(tzinfo=UY_TZ)
```
VIKTORVAV99, line 190:
```suggestion
        target_datetime = datetime.now(tz=UY_TZ)
```

VIKTORVAV99 (approved):
Just a suggestion around getting the datetime but nothing blocking.

You might also want raise an error on the initial URL with `if not r.ok`

parsers/UY.py
VIKTORVAV99: I'll be taking a look at our usage of arrow and removing it the next days so hopefully I can remove the last bits then.

pierresegonne: Nice! But FYI here we won't be able to get rid of it until the source comes back online

PR #6144, opened 2023-11-15
Remove arrow part 1

VIKTORVAV99:
This PR removes arrow in some easy and low hanging places, in some parsers it don't fully remove arrow but this will be addressed in later PRs.


unitrium (approved):
Looks good! Thanks for handling this

PR #6147, opened 2023-11-16
Capacity update - add productionCapacity to parsers

mathilde-daugy:
## Description

The goal of this PR is to add the capacity parser to the `parsers` config for the relevant zones. This is a necessary step in the [capacity update process](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Capacity-update-process)

The key `productionCapacity` is added to the `parsers` config for each zone that is updated automatically. The list of these zones can be found on the wiki [here](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Capacity-update-process#automatic_update)

The next step will be to create a poetry script that reads directly from the `parsers` config using `PARSER_KEY_TO_DICT["productionCapacity"]`

**NB** some config files have been reformatted when running `prettier`
**NB 2** this update has no impact on feeder-electricity as the `fetch_all` function only runs on the kinds : `["production",  "productionPerModeForecast", ]` -> they are hardcoded

### Preview

image

### Double check


VIKTORVAV99 (changes requested):
We need to update this part of our validation code before this can pass CI and be merged. Otherwise all python tests will fail.

https://github.com/electricitymaps/electricitymaps-contrib/blob/a0689e0156fbef49b8cd72c2bd1a9cf2ec2c1a02/electricitymap/contrib/config/model.py#L75-L82

mathilde-daugy:
@VIKTORVAV99 I've implemented the necessary changes, this is good for review now :)

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for taking the time and doing the other changes so quickly!

mathilde-daugy:
> LGTM! 🎉
> 
> Thanks for taking the time and doing the other changes so quickly!

Thank you for your reactivity!

PR #6149, opened 2023-11-16
Capacity parsers - update return type again

mathilde-daugy:
## Issue
The return types allowed are as follows: 
image

## Description

In order to comply with the allowed return types, the capacity parser `fetch_production_capacity` need to be update from `dict | None` to `dict[str, Any] | None`

### Preview

NA

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉
_again_

Issue #6136, opened 2023-11-12
[Bug] [Firefox] Zone consumption data greyed-out on map after initial page-load


jayaddison:
**Describe the bug**
When the `electricitymap` app page is opened, initially the zone consumption data appears to be missing -- all of the zone polygons are displayed in grey.  Cross-border exports do appear as expected, however.

Workaround: toggle the time from the default of 'live' to an hour ago and then back again -- or toggle the view mode from the default of 'consumption' to 'production' and back again.

**To Reproduce**
1. https://app.electricitymaps.com/map
2. Observe that the country zone polygons are coloured grey initially.

**Expected behavior**

**Screenshots**
![image](https://github.com/electricitymaps/electricitymaps-contrib/assets/55152140/32bdfce8-8251-4561-9728-b88eb8110f4c)

**Desktop:**
 - OS: Debian Trixie (13) GNU/Linux
 - Browser: Firefox
 - Version: 115.4.0

**Additional context**
This is a problem that I've only noticed within the past week -- making me think that it could be a regression.

jarek:
This also happens to me on Firefox. A reload usually fixes it. I wonder if this is some sort of race condition?

Here is developer console of a page load when it was greyed out. I have several privacy tools installed, I don't know if any of them are implicated, but the fact that it will usually load when I reload makes me suspect that it's not _directly_ causing it.

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://o192958.ingest.sentry.io/api/4504366922989568/envelope/?sentry_key=bbe4fb6e5b3c4b96a1df95145a91e744&sentry_version=7&sentry_client=sentry.javascript.react%2F7.77.0. (Reason: CORS request did not succeed). Status code: (null).

Loading failed for the <script> with source “https://plausible.io/js/plausible.local.js”. [map:117:105](https://app.electricitymaps.com/map)
Loading failed for the <script> with source “https://cdn.mouseflow.com/projects/8964f228-0fea-41ca-a8a6-dc23a38dc526.js”. [map:1:1](https://app.electricitymaps.com/map)
· · ·
· ϟ · Welcome to Electricity Maps!
· · ·
🌱 Ready to work on fixing climate change full-time?
https://electricitymaps.com/jobs
🐙 Got comments or want to contribute?
https://github.com/electricityMaps/electricitymaps-contrib
[...]

PR #6157, opened 2023-11-20
bug(web): Remove check for loaded source (again)

tonypls:
## Issue
This bug is back: #6109 reintroduced here: #6114 🤦 

## Description
Remove the check that the is source loaded before painting the map, this means we also paint the no visible zones in the background so the toggle between zone and country works as expected.

### Double check


VIKTORVAV99:
Is this what is causing #6136 perhaps?

VIKTORVAV99 (approved):

tonypls:
> Is this what is causing #6136 perhaps?

Unfortunately not, I can replicate that one locally but it's a bit hard to fix without digging deep into it.

PR #6158, opened 2023-11-20
Fix warning in the FR parser for depreciated pandas default

unitrium:
## Issue

.mean() is yielding a warning now because the default numeric_only value is depreciated.
Nothing urgent but it's a quick fix and it's closing an alert on our systems.

## Description

Set the default `numeric_only` to True and add a snapshot test to make sure nothing breaks.


VIKTORVAV99 (approved):
Simple fix!

And very nice with the inclusion of additional tests!

madsnedergaard:
This PR caused two failing builds, would you mind fixing them in a follow-up PR? :)

Screenshot 2023-11-20 at 15 36 27

We currently don't require all checks to be valid before approval, as the dynamic nature of the current setup makes this harder - so it's quite easy to miss them :)

PR #6169, opened 2023-11-22
[Performance]: Ruff

VIKTORVAV99:
## Description

This PR will replace black, flake8 and isort with ruff for a massive performance increase.

> [!Note]
> PR turned out to be bigger than expected, let me know if I should split things up 🙂


madsnedergaard:
Yaaay, looking forward to this 😍

VIKTORVAV99:
> Yaaay, looking forward to this 😍

It's turning out to be a bit bigger PR than I initially expected so let me know if you want me to split it up a bit 🙂

scripts/tooling.py
```diff
@@ -20,26 +20,27 @@ def _run(cmd: str):
...
+    _run("ruff format .")
 
 
-def lint():
-    _run(
-        "flake8 electricitymap tests parsers --count --select=F401,E901,E999,F821,F822,F823 --show-source --statistics"
-    )
-    _run("black --check .")
-    _run("isort -c .")
+def pylint():
```
VIKTORVAV99, line 27: This is crazy slow, do we actually need it or can we remove it?

I looked at the pylint rules added to Ruff so far but they don't seem to be available. Check here: https://docs.astral.sh/ruff/rules/#pylint-pl

madsnedergaard: cc @FelixDQ and @unitrium, think we can remove it now that we use ruff?

madsnedergaard (commented):
Looks good to me, leaving final review to @unitrium :)

scripts/tooling.py
FelixDQ: Yes, we have removed pylint on the internal repo in favor of ruff. The rules don't overlap 100%, but we believe that the rules from ruff are sufficient. So feel free to remove pylint from contrib as well.

VIKTORVAV99: Thanks! Will do so right away!

madsnedergaard (approved):
Let's just get this merged, it would be so nice to get rid of isort and friends!

PR #6170, opened 2023-11-23
Add better error handling in ENTSOE.

unitrium:
## Issue

Our logs are not really helpful right now to debug ENTSO-E errors in Google Cloud.
We don't get the error reason for the query and we don't get the status code.
image

Furthermore, since the error gets raised in the `query` function we don't get the zone key so it's impossible to know which zones are affected by this outage.

## Description

- Rewrite the Exception message in `query_ENTSO-E` to log the status code and the error reason instead of the whole html text.
- Add try catches around the query functions to catch parser exceptions and raise them with additional payload like the `zone_key`.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -531,12 +531,8 @@ def query_ENTSOE(
             error_text = soup.find_all("text")[0].prettify()
             if "No matching data found" in error_text:
                 exception_message = "No matching data found"
-            else:
-                exception_message = (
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
Nice! Wasn't really aware of the from clause before but looked it up and it seems very useful. I'll use it in the ruff PR as well to fix some longing errors there!

madsnedergaard (approved):
Nice, thanks 🙏

PR #6171, opened 2023-11-23
Capacity update - create poetry command 

mathilde-daugy:
## Issue

The last step of the capacity update process is to enable the contributors to update configurations using a poetry command. 
image

## Description

In the file `capacity_update.py`:

 `capacity_update` that take the parameter
-  zone: str -  to update the capacity of a specific zone (eg FR)
- source: str -  to update the capacities of all zones in a data source (eg ENTSO-e)
- target_datetime:  str - date for which we want the capacity
- update_aggregate: bool - if the selected zone is part of a country, then this enables the user to also update the aggregated capacity

In the file `update_capacity_configuration.py`

Function to update the capacity config for one zone or an aggregated zone. In  both cases, we need to treat different cases:
**1. the capacity configuration is a float or and int.** It will be updated to a dict in the following format:

capacity
├── wind
    ├── datetime: "2023-01-01"
    ├── source: "ENTSOE"
    └── value: 5233
**2. the capacity configuration is a dict.** It will be updated to a list of dicts in the following format
`[{"datetime":"2023-01-01","source":"abc","value":3},{"datetime":"2022-01-01","source":"abc","value":4}]`

**3. the capacity configuration is a list.** The new capacity will be appended to the list if the datetime is not already in the list otherwise it will replace the existing dict.

**In the case of an aggregated zone**

- if not all the formats are the same then we do not update the capacity
- if the format is a dict, the capacities are updated only if the datetime keys are the same for all subzones
- if the format is a list, the subzone capacities are extracted and summed if they have the same datetime, otherwise they are not added

### Preview

image

### Double check


capacity_update.py
```diff
@@ -0,0 +1,109 @@
...
+    zone: a two letter zone from the map or a zone group (EIA, ENTSOE, EMBER, IRENA)
+    target_datetime: ISO 8601 string, such as 2018-05-30 15:00
+    \n
+    Examples
+    -------
+    >>> poetry run capacity_update --zone FR --target_datetime "2022-01-01"
+    >>> poetry run capacity_update --source ENTSOE --target_datetime "2022-01-01"
+    """
+    assert zone is not None or source is not None
+    assert not (zone is None and source is None)
```
unitrium, line 63: Try using a XOR to combine the two above into one.

capacity_update.py
```diff
@@ -0,0 +1,109 @@
...
+
+    session = Session()
+    parsed_target_datetime = None
+    if target_datetime is not None:
+        parsed_target_datetime = datetime.fromisoformat(target_datetime)
+    else:
+        raise ValueError("target_datetime must be specified")
+
+    if source is not None:
+        if source not in CAPACITY_PARSER_SOURCE_TO_ZONES:
```
unitrium, line 73: Extract to its own function

capacity_update.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+        for zone in source_capacity:
+            if not source_capacity[zone]:
+                print(f"No capacity data for {zone} in {target_datetime}")
+            else:
+                update_zone_capacity_config(zone, source_capacity[zone])
+
+    elif zone is not None:
+        if zone not in CAPACITY_PARSERS:
```
unitrium, line 92: Same as above

scripts/update_capacity_configuration.py
```diff
@@ -0,0 +1,220 @@
...
+def update_zone_capacity_config(zone_key: ZoneKey, data: dict) -> None:
+    """Update the capacity config for a zone"""
+    if zone_key not in ZONES_CONFIG:
+        raise ValueError(f"Zone {zone_key} does not exist in the zones config")
+
+    _new_zone_config = deepcopy(ZONES_CONFIG[zone_key])
+    if "capacity" in _new_zone_config:
+        capacity = _new_zone_config["capacity"]
+
+        if all(isinstance(capacity[m], float | int) for m in capacity.keys()):
```
unitrium, line 31: Add comment to specify that is the old format and add a TODO to remove once everything is to the new format.

scripts/update_capacity_configuration.py
```diff
@@ -0,0 +1,220 @@
...
+
+    _new_zone_config["capacity"] = capacity
+
+    # sort keys
+    _new_zone_config["capacity"] = sort_config_keys(_new_zone_config["capacity"])
+
+    write_zone_config(zone_key, _new_zone_config)
+
+
+def get_zone_capacity_config(
```
unitrium, line 46: There seems to be a mismatch between the function name and the description, the name is get, while the description says that we update.

scripts/update_capacity_configuration.py
```diff
@@ -0,0 +1,220 @@
...
+
+    for mode in CAPACITY_MODES:
+        mode_capacity_configs = [
+            capacity_config[mode] for capacity_config in all_capacity_configs
+        ]
+
+        if not check_capacity_config_type(
+            mode_capacity_configs, type(mode_capacity_configs[0])
+        ):
+            logger.warning(
```
unitrium, line 125: I think it would be better to raise. So that the person first fixes all child before.

scripts/update_capacity_configuration.py
```diff
@@ -0,0 +1,220 @@
...
+        updated_capacity["source"] = ", ".join(
+            [elem for elem in sources if elem is not None]
+        )
+        return sort_config_keys(updated_capacity)
+
+
+def compute_aggregated_value(capacity_config: list[dict[str, Any]]) -> float | None:
+    """Compute the aggregated capacity value as the sum of all subzone capacities"""
+    if all(capacity_config["value"] is None for capacity_config in capacity_config):
+        aggregated_value = None
```
unitrium, line 167:
```suggestion
        return None
```
This way you can reduce the imbrications by one level.

unitrium (approved):
Great job!

PR #6173, opened 2023-11-24
Increase significant digits

### LINKED ISSUE #5921: Reduced data precision
## Description
I've noticed that the data available in the map has only two digits of precision, instead of the usual three. I feel this is quite irritating as it makes visualizing tiny differences more difficult, especially around numbers near the boundary, such as in this 
example:
Screenshot_20230920_014826
Screenshot_20230920_014845

We can see that they are different values because the percentage of capacity they represent is different.
Was this change on purpose? If so, why was this change implemented? In the case I've posted the difference between possible values is of up to 8% (13.5/12.5), quite significant.

It is possible this was introduced with the change from power to energy units.

Hope all this makes sense, and this issue hasn't been raised before.

silimotion:
## Issue

Closes #5921

## Description

This PR increases the default number of digits used by the tooltips from 2 to 3. I believe this is better as it shows more detailed data, but it's open to discussion.

### Preview

Before:
Screenshot_20231124_115544

After:
Screenshot_20231124_012718

### Double check


VIKTORVAV99 (dismissed):
This makes sense to me but I think it would be good to have a second pair of eyes take look at this as well.

So @madsnedergaard, @tonypls and @Alportan what do you guys think?

madsnedergaard (commented):
I overall think it makes sense, except for a single use case - the x-axis label on the breakdown chart:

Screenshot 2023-11-27 at 10 51 43

As an alternative, we can make the number optionally adjustable and then just change it for that specific place? :)

Alportan:
It does make sense to increase the accuracy of our data by adding decimals. 🙌

**💡 Idea for the future:** Agree on the number of decimals we display for different data points to ensure visual consistency.

VIKTORVAV99 (changes requested):
Looks like we all agree this would be a nice change.

However if you could update the chart to use 2 digits only that would be great!

silimotion:
I have set the BreakdownChart to 2 digits now because I've seen some zones in the map with labels such as "15 GW" that could be incorrectly rounded. Are there any checks in the code so that the calculated values in the labels only need 2 digits and never more?

VIKTORVAV99 (approved):
LGTM!

Thanks for making the last changes!

PR #6174, opened 2023-11-24
Update dependabot configuration to include ruff updates

VIKTORVAV99:
Ruff seems to be releaseing weekly updates. In order to ensure we get the benefits of that we should enable dependabot to update ruff on a daily shedule.


madsnedergaard (approved):
Makes sense :)

VIKTORVAV99:
Thanks!

Will merge this after the ruff PR 🙂

PR #6175, opened 2023-11-24
Update BE solar capacity

Fred-3000:
An update seemed fit as 800 MW has come online the past 3 months, or in other words: a 10% increase! 🥳
[Same source. ](https://www.elia.be/en/grid-data/power-generation/solar-pv-power-generation-data)


VIKTORVAV99 (approved):
Thanks for updating these!

PR #6176, opened 2023-11-25
Change Python caching strategy

VIKTORVAV99:
This PR removes the built in caching in favour of our custom cache.

Before this we accidentally created duplicate caches (with the same data) as well as spending extra time restoring and saving both caches.

It won't cut down CI times that much (around 4 sec) but right now we are just wasting resources.

It also fixes the version printing for the poetry version to ensure the correct version is used for the venv.


madsnedergaard (approved):
Nice, looks good to me! :)

Issue #6135, opened 2023-11-12
Remove usage of arrow


VIKTORVAV99:
## Description
The built in datetime class is faster and less error prone than arrow so we should migrate to fully using datetime instead of arrow in our parsers.

This probably needs to be done in steps so it's easier to review as it can have a big impact on how parsers are structured.

### Parsers missing:
- [x] [parsers/EIA.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/EIA.py)
- [x] [parsers/ELEXON.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/ELEXON.py)
- [x] [parsers/RU.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/RU.py)
- [x] [parsers/occtonet.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/occtonet.py)
- [x] [parsers/statnett.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/statnett.py)
- [x] [parsers/GSO.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/GSO.py)
- [x] [parsers/IL.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IL.py) - #7348
- [x] [parsers/IN.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IN.py)
- [x] [parsers/IN_PB.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IN_PB.py)
- [x] [parsers/IN_UT.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IN_UT.py)
- [x] [parsers/IN_WE.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IN_WE.py)
- [x] [parsers/IQ.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/IQ.py)
- [x] [parsers/JP.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/JP.py)
- [x] [parsers/NG.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/NG.py)
- [x] [parsers/CO.py](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/CO.py)
[...]

PR #6179, opened 2023-11-28
refactor: removed arrow from CA-AB parser

Barissa-Imran:
## Issue

Ref: #6135 

## Description

Upgrade the parser to use `datetime`. Parser functionality has not changed.

- add `24` hour handling (midnight next day)

### Double check


parsers/CA_AB.py
```diff
@@ -79,9 +79,21 @@ def fetch_price(
...
+                date_time = datetime.strptime(row[0], "%m/%d/%Y %H").replace(
+                    tzinfo=TIMEZONE
+                )
+            except ValueError:
+                # Handle special case where hour is `24`
+                row[0] = row[0].replace(" 24", " 00")
+                date_time = datetime.strptime(row[0], "%m/%d/%Y %H").replace(
+                    tzinfo=TIMEZONE
+                )
+                date_time += timedelta(days=1)
```
VIKTORVAV99, line 92: I think it would be easier to split the date string into two parts, the date and the time.
Once we have the time seperate we can parse it as an int and do:
```python
datetime.strptime(f"{date} {hour - 1}", "%m/%d/%Y %-H")
```

This would remove the need for the seccond special operation and would give us an accurate index for all hours. If hour 24 is the next day we can simply shift back the hours by 1 using `+ timedelta(hours=1)` as that is fully datetime aware.

Barissa-Imran: this implementation is better actually see latest commit

parsers/CA_AB.py
```diff
@@ -79,9 +79,11 @@ def fetch_price(
     prices = PriceList(logger)
     for row in csv.reader(response.text.split("\r\n\r\n")[2].splitlines()[1:]):
         if row[1] != "-":
+            date, hour = row[0].split()
             prices.append(
                 zoneKey=zone_key,
-                datetime=arrow.get(row[0], "MM/DD/YYYY HH", tzinfo=TIMEZONE).datetime,
+                datetime=datetime.strptime(f"{date} {int(hour) - 1}", "%m/%d/%Y %H")
+                + timedelta(hours=1),
```
VIKTORVAV99, line 86:
```suggestion
                datetime=datetime.strptime(f"{date} {int(hour) - 1}", "%m/%d/%Y %H").replace(tzinfo=TIMEZONE),
                + timedelta(hours=1),
```

We need to attach a the timezone as well.

Also are you sure that the value of hour 24 are for 00 the next day and not simply hour 23 but 1 indexed? Otherwise we don't need the timedelta.

madsnedergaard:
Closing this as it has been fixed in https://github.com/electricitymaps/electricitymaps-contrib/pull/6559

PR #6182, opened 2023-11-28
Add support for fetching historical data for all French overseas regions

VIKTORVAV99:
## Issue

Currently our historical data parsing doesn't work for all zones.

## Description
While looking for something else entirely (capacity data) I found a nice dataset on the main EDF dataportal that has historical data for all the French overseas zones!

### Double check


parsers/FR_O.py
```diff
@@ -185,21 +182,25 @@ def fetch_data(
...
         raise ParserException(
             "FR_O.py",
-            f"Unexpected data format for {zone_key} for {target_datetime.strftime('%Y')}"
+            f"Unexpected data format for {zone_key} for {target_datetime}"
             if target_datetime
             else f"Unexpected data format for {zone_key}.",
             zone_key,
         )
-    return data, DATE_STRING_MAPPING[zone_key]
+    source = url.split("//")[1].split("/")[0]
```
VIKTORVAV99, line 190: Do we want to just set the source as EDF for all of these datatypes and regions?
All our data comes from EDF subdomains.

VIKTORVAV99:
@mathilde-daugy if you have the time could you give this a quick review so we can fetch the data before we re-run the flow tracing?

parsers/FR_O.py
mathilde-daugy: i think edf is good enough

mathilde-daugy: today we use opendata.edf.fr, i think we should keep this source

mathilde-daugy (commented):
Looks good to me :)

madsnedergaard (approved):

parsers/FR_O.py
VIKTORVAV99: We use different ones of all the regions, FR-COR for example use opendata-corse.edf.fr

So I would prefer using edf.fr as the source as it covers them all and is the organisation in it self.

mathilde-daugy: i would use opendata.edf.fr to signify where we get the data too. Edf is a massive company and it can be hard to get back to the data source

VIKTORVAV99: True, let's do `opendata.edf.fr` then 🙂

PR #6184, opened 2023-11-28
[IN-WE] Handle errors better during expected downtime

madsnedergaard:
## Description

- Handle all JSON parsing errors the same 
- Raise different error during the expected outage period over the weekend

### Double check


mathilde-daugy (approved):

VIKTORVAV99:
Don't you dare merge this without formatting first! 😅

consideRatio:
Hey @madsnedergaard! I'm working working #6135 for the removal of arrow for the IN-WE parser, and picked up on a possible bug where `datetime.now()` is used instead of `datetime.now(<api servers timezone>)` when deciding on if an outage is expected or not.

Goal outcomes:
1. _A reference comment about this outage clarifying some details_
   - Do we have external docs about this to link to, or is it an observation we have made without external docs?
   - Do we know it's the API, or the proxy we use to reach it?
   - A clarification: is data for these days not available by the API server, or is the data not served by the API server these days?
2. _A verification/fix of using `datetime.now()` instead of `datetime.now(ZoneInfo("Asia/Kolkata"))`_

madsnedergaard:
> Hey @madsnedergaard! I'm working working #6135 for the removal of arrow for the IN-WE parser, and picked up on a possible bug where `datetime.now()` is used instead of `datetime.now(<api servers timezone>)` when deciding on if an outage is expected or not.
> 
> Do you remember something about this expected downtime?
> 
> Goal outcomes:
> 
> 1. _A reference comment about this outage clarifying some details_
>    
>    * A clarification: is data for these days not available by the API server, or is the data not served by the API server these days?
>    * Do we have external docs about this to link to, or is it an observation we have made without external docs?
>    * Do we know it's the API, or the proxy we use to reach it?
> 2. _A verification/fix of using `datetime.now()` instead of `datetime.now(ZoneInfo("Asia/Kolkata"))`_

Hey @consideRatio, thanks for asking.

I think you're right that it is indeed a bug - I can't remember if it has actually caused any problems, but I think the change you made in #7627 looks reasonable, so let's try it :)

As far as I remember, the data is manually uploaded hence why it's not available until after the weekend 😅 It's not documented anywhere, but just something we discovered while developing the parser.

PR #6186, opened 2023-11-28
[Capacity parsers] - Update parser for DE

mathilde-daugy:
## Issue

ENTSOE capacity data doesn't include battery storage. The zones affected by this are DE and FR. 

## Description

For DE we can get installed capacity from [energy-charts](https://www.energy-charts.info/charts/installed_power/chart.htm?l=en&c=DE&legendItems=1110111111111&year=2022)

The data is consistent with ENTSO-e and it includes battery storage

### Preview

image

### Double check


electricitymap/contrib/capacity_parsers/DE.py
```diff
@@ -0,0 +1,93 @@
...
+    try:
+        return float(value) * CONVERT_GW_TO_MW
+    except:
+        return None
+
+
+def fetch_production_capacity(
+    zone_key: ZoneKey, target_datetime: datetime, session: Session
+) -> dict[str, Any] | None:
+    r: Response = session.get(REQUEST_URL)
```
VIKTORVAV99, line 44: If this is run via the test parser it will crash from the lack of a logger object and that session is none.

(In case someone wants to run it without updating the config file.)

VIKTORVAV99 (commented):
It works if it's run with the update_capacity script but running it with the test parser fails (not sure if it's even ment to be run via the test parser).

There is also a small type error in the convert to mw function but I left a suggestion.

electricitymap/contrib/capacity_parsers/DE.py
VIKTORVAV99: Ah okay!

That makes sense I think we can just add an early return in the test_parser that says it doesn't support capacity.

But we can do that in another PR when we have the time.

VIKTORVAV99 (approved):

electricitymap/contrib/capacity_parsers/DE.py
mathilde-daugy: i will make a note of that ;)

PR #6192, opened 2023-11-29
refactor: remove arrow from CA-QC

Barissa-Imran:
## Issue
#6135 

## Description
Remove arrow module from parser. maintained its functionality
- replace `arrow` with `datetime`

### Double check


parsers/CA_QC.py
```diff
@@ -37,7 +35,9 @@ def fetch_production(
     now = datetime.now(tz=TIMEZONE)
     for elem in data:
         values = elem["valeurs"]
-        timestamp = arrow.get(elem["date"], tzinfo=TIMEZONE).datetime
+        timestamp = datetime.strptime(elem["date"], "%Y-%m-%dT%H:%M:%S").replace(
+            tzinfo=TIMEZONE
+        )
         # The datasource returns future timestamps or recent with a 0.0 value, so we ignore them.
```
VIKTORVAV99, line 41:
```suggestion
        if isinstance(elem["date"], str):
            timestamp = datetime.fromisoformat(elem["date"]).replace(tzinfo=TIMEZONE)
```
We should make sure the date field is a string before we use it and we should stick with fromisoformat whenever possible.

parsers/CA_QC.py
```diff
@@ -71,7 +71,9 @@ def fetch_consumption(
         if "demandeTotal" in elem["valeurs"]:
             consumption.append(
                 zoneKey=zone_key,
-                datetime=arrow.get(elem["date"], tzinfo=TIMEZONE).datetime,
+                datetime=datetime.strptime(elem["date"], "%Y-%m-%dT%H:%M:%S").replace(
+                    tzinfo=TIMEZONE
+                ),
```
VIKTORVAV99, line 76: We should use isoformat here as well.

VIKTORVAV99 (commented):
Two small things otherwise this looks good!

PR #6205, opened 2023-11-30
Enable ruff rule DTZ003 and DZT004

VIKTORVAV99:
This PR enables the ruff rule DZT003 which will dissallow usage of `datetime.utcnow()` and rule DZT004 which will dissallow the usage of `datetime.utcfromtimestamp()`.

Both of these functions are deprecated and will be removed in future python versions and have direct replacements.

### DIFF
--- a/parsers/ESKOM.py
+++ b/parsers/ESKOM.py
@@ -54,7 +54,7 @@ PRODUCTION_IDS = [0, 6, 8, 9, 10, 11, 16, 17, 18, 19]
 def get_url() -> str:
     """Returns the formatted URL"""
     date = datetime.now(timezone.utc)
-    return f"https://www.eskom.co.za/dataportal/wp-content/uploads/{date.strftime('%Y/%m')}/{date.strftime('%m')}/Station_Build_Up.csv"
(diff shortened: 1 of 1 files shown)


parsers/ESKOM.py
```diff
@@ -53,8 +53,8 @@
 
 def get_url() -> str:
     """Returns the formatted URL"""
-    date = datetime.utcnow()
-    return f"https://www.eskom.co.za/dataportal/wp-content/uploads/{date.strftime('%Y')}/{date.strftime('%m')}/Station_Build_Up.csv"
+    date = datetime.now(timezone.utc)
+    return f"https://www.eskom.co.za/dataportal/wp-content/uploads/{date.strftime('%Y/%m')}/{date.strftime('%m')}/Station_Build_Up.csv"
```
madsnedergaard, line 57: Why the URL change here? :)

VIKTORVAV99: Oh that was just me trying things out, I thought I reverted that...

This does work though but I'll change it back to keep the PR as clean as possible.

PR #6219, opened 2023-12-04
[CA-BC] Updates unknown emission factor for zone

### LINKED ISSUE #6129: [Data Issue]: Estimate for British Columbia (CA-BC) assumes generation thermal but it's mostly hydro
### When did this happen?

Recently but not sure of the exact date. I checked a couple of days ago and British Columbia has always been a "grey" province with no information other than transmission to Alberta and Washington state.

### What zones are affected?

CA-BC

### What is the problem?

The map https://app.electricitymaps.com/zone/CA-BC shows British Columbia as having a carbon intensity now of:
> 700 gCO₂eq/kWh (Source: assumes thermal (coal, gas, oil or biomass))

British Columbia produces most of it's electricity from hydro. It would be better to use [this data from the federal government of Canada](https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-british-columbia.html) to estimate the unknown CO₂eq emissions instead of assuming it is thermal.

madsnedergaard:
## Issue

Closes #6129

## Description

Haven't done this in a while, so can't 100% remember who we do things 😅 

The most recent numbers I could find are from 2019/2020, so it's a bit outdated, but still better than the existing thing.

**Source:** https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-british-columbia.html#s1

**Calculation:**

Screenshot 2023-12-04 at 09 25 16


VIKTORVAV99 (changes requested):
We also need to add the direct values and RE ratio and Low Carbon Ratio or things will look weird 😅

I have a bunch of time on the train later today if you want me to add them instead 🙂

madsnedergaard:
> We also need to add the direct values and RE ratio and Low Carbon Ratio or things will look weird 😅
> 
> I have a bunch of time on the train later today if you want me to add them instead 🙂

Oh yeah okay, it also seemed too easy 😄 
That would be great, then I'll afterwards attempt to document things better (like what does `powerOriginRatios` even do) in the wiki :)

VIKTORVAV99:
> > We also need to add the direct values and RE ratio and Low Carbon Ratio or things will look weird 😅
> > I have a bunch of time on the train later today if you want me to add them instead 🙂
> 
> Oh yeah okay, it also seemed too easy 😄 That would be great, then I'll afterwards attempt to document things better (like what does `powerOriginRatios` even do) in the wiki :)

Honestly I'm not sure they are used anymore. But I think they are static fallbacks for when the live data is missing, but those use estimates now so they do, they also haven't been updated for a while. 😅

I think it would be best to have a talk with @pierresegonne about what we use these for and if we even need them anymore?

pierresegonne:
You're right about this, the powerOriginRatios are fallbacks for when data is missing. As long as we're not estimating exchanges across the board we should keep them :)

VIKTORVAV99 (approved):

PR #6226, opened 2023-12-06
Add VN unknown emission factor

silimotion:
## Description

This adds the unknown emission factor for Vietnam, prerequisite to it showing on the map. I've used EMBER data and the default emission factors from the wiki.

### Double check

### DIFF
--- a/config/zones/VN-C.yaml
+++ b/config/zones/VN-C.yaml
@@ -13,4 +13,41 @@ parsers:
   price: VN.fetch_price
   consumption: VN.fetch_consumption
 
+emissionFactors:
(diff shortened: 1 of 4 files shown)


VIKTORVAV99 (approved):
Thanks for updating the files!

silimotion:
@VIKTORVAV99 
Is anything left?

VIKTORVAV99:
Not sure what you mean?

VIKTORVAV99:
If you mean to enable the estimations not really, we just need to add it to the backend config. But we are doing some db changes so not really changing those configs until that is done.

Issue #6247, opened 2023-12-13
The labeler action is broken


pierresegonne:
## Description

The labeler action seems broken - it fails on all new PRs, whether they are valid or not.

Fails wiht: `Error: TypeError: Input does not meet YAML 1.2 "Core Schema" specification: sync-labels`

VIKTORVAV99:
Yeah I noticed that, was a bit quick with a dependency update that passed all the check but I'll have a fix ready today.

PR #6248, opened 2023-12-13
migrate labeler config to new schema

VIKTORVAV99:
## Issue

Fixes: #6247 

## Description

Migrates the labeler config to use the new schema syntax.

### Preview

It won't be working until this is merged to master. So no way to preview it in here.

### Double check

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,45 +1,66 @@
 infrastructure:
-  - .github/**/*
-  - docker-compose.yml
-  - web/Dockerfile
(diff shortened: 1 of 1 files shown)


pierresegonne:
If we can't see that it works, how have you tested it? 😅

VIKTORVAV99:
> If we can't see that it works, how have you tested it? 😅

This is one of those situations where we will be testing on master 😅 🙈

pierresegonne (approved):
YOLO

PR #6249, opened 2023-12-13
[Labeler]: Switch sync labels to a bool

VIKTORVAV99:
The first pr fixed the schema but a new error poped up.

We used a empty string for the config of the sync labels option which is no longer supported, this PR sets it to a bool instead.


tonypls (approved):
LGTM

PR #6252, opened 2023-12-14
Switch RUFF to use SARIF output format for code analytics

VIKTORVAV99:
## Description
Switching to SARIF format should give us more contextual and better code annotations in GitHub. This format is also cross platform so it will be easier to integrate with other systems if we ever decide to do that.

SARIF as an output format was added to RUFF in version 0.1.8 which is why it wasn't used from the start.

### DIFF
--- a/.github/workflows/python_formatting.yml
+++ b/.github/workflows/python_formatting.yml
@@ -36,6 +36,13 @@ jobs:
       - name: Check formatting
         run: |
           poetry run ruff format . --check
+        continue-on-error: true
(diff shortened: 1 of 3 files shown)


.github/workflows/python_formatting.yml
```diff
@@ -36,6 +36,13 @@ jobs:
       - name: Check formatting
         run: |
           poetry run ruff format . --check
```
hussam789, line 38:
```suggestion

PR #6256, opened 2023-12-16
Fix initial map state uncolored zones issue

### LINKED ISSUE #6240: Country view broken unless switching to zone and back
**Describe the bug**
Since a couple of weeks, when I open Electricitymap the Country-level view map is gray without co2-intensities showing. When I switch to zone-view and back to country-view it works again.

This is pretty annoying, especially on mobile because the switching takes more clicks there.
It happens to me 

**To Reproduce**
1. Go to Electricitymap
2. The map is gray
3. Switch to zone view, the map shows colors again
4. Switch to country view, the map still shows colors

**Expected behavior**
I expect the map to show colors in country view from the beginning

**Screenshots**
This is the default view
grafik

**Desktop (if relevant):**
 - OS: Win10
 - Browser Firefox
 - Version 120
 - ublock origin enabled, but the error happens without as well
 - clearing cache did not help

**Smartphone (if relevant):**
 - OS: Android 13
 - Browser Firefox
 - Version 114
 -

Ugzuzg:
## Issue

Closes https://github.com/electricitymaps/electricitymaps-contrib/issues/6136
Closes #6240

## Description

The main issue is that isSourceLoaded inside the effect that renders the zones is not reactive and the change of the source state doesn't trigger a re-render of the component. Subscribing to that and changing the state fixes the issue.

Adding map to a dependency list of the effects makes sure that the map reference doesn't get stale in the closure and fixes the hot reload issue on my machine.

### Double check


tonypls:
Thanks for looking into this! 

I'm still getting the issue on firefox, now it's happening on every refresh on firefox for me.


image

Ugzuzg:
I think, it should be fixed now.

larkinox:
Just to confirm that I'm also still seeing this issue. Firefox 120.0.1 (64-bit), fresh visit to the map in a private browsing window. Happy to troubleshoot further if useful.

image

tonypls (approved):
LGTM, thanks for fixing!

tonypls:
Failing test is unrelated btw!

jayaddison:
Thank you @Ugzuzg!

PR #6262, opened 2023-12-19
refactor(parser structure): Refactor IS parser

unitrium:
## Description

Refactoring the IS parser to use the new datastructure.

### Preview
Testing is currently pending because the website seems to be down so I can't get a test sample.

### Double check


VIKTORVAV99 (approved):
I think the endpoint is configured to only accept our production server IP and IPs in Iceland. So might be hard to get a a sample.

The changes look good though!

unitrium:
Ah ok that might explain some stuff! I will try to get a sample through a proxy

PR #6263, opened 2023-12-19
feat(app): Use the v8 backend

VIKTORVAV99:
## Issue

Closes: AVO-6

### Double check


VIKTORVAV99:
I don't plan on merging this anytime soon as I want all of the other contrib changes deployed before we merge and deploy this to take your time.

There is a lot of changes in here so don't hesitate to adress anything you feel is wrong or strange.

VIKTORVAV99:
Added JSDoc comments to the new types so there is a explanation of the value like this:
image

web/src/features/exchanges/ExchangeTooltip.tsx
```diff
@@ -14,11 +14,16 @@ interface ExchangeTooltipProperties {
 
 export default function ExchangeTooltip(
   properties: ExchangeTooltipProperties
-): ReactElement {
-  const { key, netFlow, co2intensity } = properties.exchangeData;
+): ReactElement | null {
+  const { key, f, ci } = properties.exchangeData;
+  if (!f || !ci) {
+    return null;
+  }
```
VIKTORVAV99, line 21: This will never be triggered as the ExchangeArrow function where this is called from has already exited early. I'll do a bigger revamp, code cleanup and deduplication of these tooltips in a different PR.

madsnedergaard:
Let me know when this is ready for review :)

VIKTORVAV99:
> Let me know when this is ready for review :)

The frontend is good to go but I think I found a bug in the backend with the estimation filtering I need to figure out first before we deploy this to all users.

But if you want to get a head start on the review feel free to do so!

VIKTORVAV99:
I'll take care of the conflict but otherwise this PR is now good to go!

web/src/features/exchanges/ExchangeTooltip.tsx
```diff
@@ -17,10 +17,11 @@ export default function ExchangeTooltip({
   exchangeData,
   isMobile,
 }: ExchangeTooltipProperties): ReactElement {
-  const { key, netFlow, co2intensity } = exchangeData;
+  const { key, f, ci } = exchangeData;
```
madsnedergaard, line 20: I'm sorry for pointing this out quite late, but I think we should try to do the remapping as early as possible instead of using the shortened version all the way down.

In this exchange we could make the remapping in `useExchangeArrowsData` so it's passed down :)

VIKTORVAV99: Will take care of it later today 🙂

madsnedergaard: Thanks, I know it's quite late to bring this up 😅

web/src/features/panels/ranking-panel/getRankingPanelData.ts
```diff
@@ -33,15 +33,21 @@ export const getRankedState = (
   if (!data) {
     return [];
   }
-  const zonesData = data.data;
-  const keys = Object.keys(zonesData.zones) as Array<keyof GridState>;
+  const gird = data.data.datetimes[datetimeIndex];
```
madsnedergaard, line 36: Is this supposed to be `grid`?

VIKTORVAV99: I think so, as in grid state.

But I'll double check 🙂

web/src/features/map/MapTooltip.tsx
```diff
@@ -92,26 +68,24 @@ export function TooltipInner({
...
   hasOutage,
-  estimationMethod,
-  estimatedPercentage,
+  estimated,
 }: {
-  hasOutage: boolean | undefined;
-  estimationMethod: string | undefined;
-  estimatedPercentage: number | undefined;
+  hasOutage?: boolean | null;
+  estimated?: number | boolean | null;
```
madsnedergaard, line 74: I think this idea of a property that can be both a number and a boolean is going to cause problems down the road, but we can reconsider this at a later point :)

web/src/features/panels/ranking-panel/getRankingPanelData.ts
```diff
@@ -33,15 +33,19 @@ export const getRankedState = (
   if (!data) {
     return [];
   }
-  const zonesData = data.data;
-  const keys = Object.keys(zonesData.zones) as Array<keyof GridState>;
+  const girdState = data.data.datetimes[datetimeIndex];
```
madsnedergaard, line 36: `gird` should probably be grid here too, right?

web/src/features/panels/zone/ZoneHeaderGauges.tsx
```diff
@@ -36,19 +36,16 @@ export function ZoneHeaderGauges({ data }: { data?: ZoneDetails }) {
...
+    { c: { rr: renewableRatio }, p: { rr: renewableRatioProduction } },
+    isConsumption
   );
   const fossilFuelPercentage = getFossilFuelRatio(
-    isConsumption,
-    fossilFuelRatio,
-    fossilFuelRatioProduction
+    { c: { fr: fossilFuelRatio }, p: { fr: fossilFuelRatioProduction } },
+    isConsumption
   );
```
madsnedergaard, line 49: couldn't we just pass the whole "selectedData" variable to the util functions here instead of reconstructing the object?

madsnedergaard (commented):
Thanks for making the changes, it's definitely much easier to understand and use now 😄 
A few additional comments

web/src/features/map/MapTooltip.tsx
VIKTORVAV99: Possibly but considering what we learned about the estimations recently I think that maybe it should just be a number for all time aggregations.

web/src/features/panels/ranking-panel/getRankingPanelData.ts
VIKTORVAV99: That's what I get for trying to spell in the morning 😅

I'll take care of it...

web/src/features/panels/zone/ZoneHeaderGauges.tsx
VIKTORVAV99: Possibly, and might be a better solution. I'll take a look!

VIKTORVAV99: So this is because this is using the zone data and not the state data, when the state data is used it's passed like this:
```js
  const intensity = getCarbonIntensity(zoneData, isConsumption);
```

So if we want to use the same function for both one of them need reconstruct the object in one place or the other and I think it's better the zone details should do it  than the state related things as they are used more and need to be faster.

I'm considering changing the usage of the carbon intensity from the details file to the state files though as those are already loaded. That means they will be used the same way. But that is a separate PR for sure.

VIKTORVAV99: It would be possible to make the function handle both kinds of data and have internal logic on which to use but unsure how beneficial that would be over constructing the object. So I'll leave that up to you to decide then I'll implement it if we decide to do so.

madsnedergaard: Ah okay, I see - let's keep it like this then :)

madsnedergaard:
I just hit the same "state and details not in sync" problem that we've had before, where visiting at 21:55 meant the numbers were out of sync due to caching. This became more clear while testing this PR as no users are using `v8` yet. I suggest we pick up this problem soon-ish after the app-backend caching project is done :)

Screenshot 2024-04-11 at 22 00 03
Screenshot 2024-04-11 at 21 59 49

Don't think it's blocking for this PR though, just thought I'd share it since we talked about it earlier today.

madsnedergaard (approved):

VIKTORVAV99:
Will create a new release tomorrow morning which includes this and monitor it during the day for any issues.

VIKTORVAV99:
Thanks for the reviews!

tonypls:
I know I'm late to the party here but it feels like the right place to discuss.

Would it be beneficial to convert all shortened keys into human-readable ones at the API query level? Revisiting the codebase and having to refer to an interface to decode a variable adds another layer of complexity.

Consider the difference in cognitive load between `const { e, o } = zoneData;` and `const { isEstimated, hasOutage } = zoneData;`. Making variable names human-readable from the API layer down could enhance the developer experience by improving readability. Although there might be a slight increase in bundle size, variable names are generally minified, so the impact should be negligible. 

What are your thoughts on prioritizing readability in this context?

VIKTORVAV99:
> I know I'm late to the party here but it feels like the right place to discuss.
> 
> 
> 
> Would it be beneficial to convert all shortened keys into human-readable ones at the API query level? Revisiting the codebase and having to refer to an interface to decode a variable adds another layer of complexity.
> 
> 
> 
> Consider the difference in cognitive load between `const { e, o } = zoneData;` and `const { isEstimated, hasOutage } = zoneData;`. Making variable names human-readable from the API layer down could enhance the developer experience by improving readability. Although there might be a slight increase in bundle size, variable names are generally minified, so the impact should be negligible. 
> 
> 
> 
> What are your thoughts on prioritizing readability in this context?
> 
> 

I would be happy to improve the Developer Experience of this, especially if we end up changing the details data as well as that's even more complex. But I'm not sure how much of a priority it should be?

Bundle size issue of that in particular would be almost non existent so I don't think that should be a concern. A bigger concern is the runtime cost of converting all the keys IMO.

I can take a look at it during the slack week and then we can make a more informed decision on what we want to do if you wish?

tonypls:
> > I know I'm late to the party here but it feels like the right place to discuss.
> > Would it be beneficial to convert all shortened keys into human-readable ones at the API query level? Revisiting the codebase and having to refer to an interface to decode a variable adds another layer of complexity.
> > Consider the difference in cognitive load between `const { e, o } = zoneData;` and `const { isEstimated, hasOutage } = zoneData;`. Making variable names human-readable from the API layer down could enhance the developer experience by improving readability. Although there might be a slight increase in bundle size, variable names are generally minified, so the impact should be negligible.
> > What are your thoughts on prioritizing readability in this context?
> 
> I would be happy to improve the Developer Experience of this, especially if we end up changing the details data as well as that's even more complex. But I'm not sure how much of a priority it should be?
> 
> Bundle size issue of that in particular would be almost non existent so I don't think that should be a concern. A bigger concern is the runtime cost of converting all the keys IMO.
> 
> I can take a look at it during the slack week and then we can make a more informed decision on what we want to do if you wish?

I do find it slows me down significantly when navigating the code base, as an example I often highlight and jump to variable names but in this case it's no longer a possibility: 
image

Ideally, we should prioritize enhancements to developer experience that restore the usability we previously enjoyed. I have created a ticket to address this [here](https://linear.app/electricitymaps/issue/AVO-190/map-v8-api-to-human-readable-variable-names) happy to pick it up myself during slack week too if you're busy.

VIKTORVAV99:
> > > I know I'm late to the party here but it feels like the right place to discuss.
> > > Would it be beneficial to convert all shortened keys into human-readable ones at the API query level? Revisiting the codebase and having to refer to an interface to decode a variable adds another layer of complexity.
> > > Consider the difference in cognitive load between `const { e, o } = zoneData;` and `const { isEstimated, hasOutage } = zoneData;`. Making variable names human-readable from the API layer down could enhance the developer experience by improving readability. Although there might be a slight increase in bundle size, variable names are generally minified, so the impact should be negligible.
> > > What are your thoughts on prioritizing readability in this context?
> > 
> > 
> > I would be happy to improve the Developer Experience of this, especially if we end up changing the details data as well as that's even more complex. But I'm not sure how much of a priority it should be?
> > Bundle size issue of that in particular would be almost non existent so I don't think that should be a concern. A bigger concern is the runtime cost of converting all the keys IMO.
> > I can take a look at it during the slack week and then we can make a more informed decision on what we want to do if you wish?
> 
> I do find it slows me down significantly when navigating the code base, as an example I often highlight and jump to variable names but in this case it's no longer a possibility: image
> 
> Ideally, we should prioritize enhancements to developer experience that restore the usability we previously enjoyed. I have created a ticket to address this [here](https://linear.app/electricitymaps/issue/AVO-190/map-v8-api-to-human-readable-variable-names) happy to pick it up myself during slack week too if you're busy.

Makes sense. Although if you limited that to match exactly you'd find it pretty fast 😉

Personally I usually use the reference navigator which shows you exactly where an object is used:
image

madsnedergaard:
I missed that we didn't update it everywhere, we should indeed have the full, readable names wherever possible IMO :)

PR #6264, opened 2023-12-19
Update v8 mockdata to include last_hour

VIKTORVAV99:
## Issue

It appears that the last_hour mock data was not included before.

## Description

Updates the v8 mock data to include last_hour so it can be used in the tests for #6263


tonypls (approved):
LGTM 🚀

PR #6265, opened 2023-12-20
refactor(parser validation): Refactor BD part 1

unitrium:
## Description

I'm starting a refactor of this parser. To avoid a big messy PR I'm splitting it into smaller PRs.
This PR just renames the parser and create a baseline test which would make sure that subsequent PRs don't break anything.

### Double check


VIKTORVAV99 (approved):
LGTM!

I'll mark the flagged code as used in tests later today 🙂

unitrium:
I have dismissed the warnings but I actually agree that I could sanitize the html input a bit, people will execute this test on their computers.

PR #6267, opened 2023-12-20
refactor(parser-validation): refactor BD part 2

unitrium:
To be reviewed after the other PR to fix names has been merged.
## Description

- Remove arrow.
- Fix an issue regarding the date. The source is reporting 24:00 as the first hour of the day instead of 00:00. Used with arrow's get it caused confusion because we mapped it to the day before (reflected in the snapshot changes). Now we correctly map it as the hour of the next day.
- Convert to using TotalConsumption.

### Double check

### DIFF
--- a/parsers/ERP_PGCB.py
+++ b/parsers/ERP_PGCB.py
@@ -5,10 +5,13 @@ from logging import Logger, getLogger
 from typing import Any
 from zoneinfo import ZoneInfo
 
-import arrow
 from bs4 import BeautifulSoup, Tag
 from requests import Response, Session
 
+from electricitymap.contrib.lib.models.event_lists import (
+    TotalConsumptionList,
+)
+from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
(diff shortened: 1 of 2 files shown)


unitrium:
Can be reviewed now, the other part has been merged

parsers/ERP_PGCB.py
```diff
@@ -166,7 +164,7 @@ def query(
             parser="BD.py",
             message=("Could not find table header in returned HTML."),
         )
-    verify_table(table_head)
+    verify_table_header(table_head)
```
VIKTORVAV99, line 167: This does cause some type errors as the table head can be of Tag | NavigatableString | Int but I don't think it will cause any runtime issues as it is right now.

parsers/ERP_PGCB.py
```diff
@@ -68,47 +71,42 @@ def parse_table_body(table_body: Tag):
...
-        row_dict["bd_import_tripura"] = table_entry_to_float(row_items[12])
-        row_dict["remarks"] = row_items[13]
-
-        row_data.append(row_dict)
+            parsed_time = datetime.strptime(row_items[1], "%H:%M").time()
+
+        row_data.append(
+            {
+                "time": datetime.combine(parsed_day, parsed_time, tzinfo=TIMEZONE),
+                "total_generation": table_entry_to_float(row_items[2]),  # MW
```
VIKTORVAV99, line 91: Shall we also add a parser function for this?
I know it's not really used by us right now but would be nice to start collecting these to use in validation rules in the future?

VIKTORVAV99 (approved):
Everything works as expected.
There is a small type error though but it should not impact runtime functionality, would be good to double check this in the follow up PR for production parsing though.

PR #6271, opened 2023-12-21
refactor(parser-valiation): refactor BO parser

unitrium:
## Description

Refactor the BO parser to use the new datastructure. Add test coverage and minor improvements.

### Double check


parsers/CNDC.py
```diff
@@ -0,0 +1,152 @@
...
+def parse_generation_forecast(
+    zone_key: ZoneKey, date: datetime, raw_data: list[dict], logger: Logger
+) -> TotalProductionList:
+    result = TotalProductionList(logger)
+    assert date.tzinfo == tz_bo
+    for hour_row in raw_data:
+        [hour, forecast, total, thermo, hydro, wind, solar, bagasse] = hour_row
+
+        # "hour" is one-indexed
+        timestamp = get_timestamp(query_date=date, hour=hour - 1)
```
VIKTORVAV99, line 73: Not sure if you changed this part but I would prefer if it was called something like get_datetime instead as it actually returns a datetime.

parsers/CNDC.py
```diff
@@ -0,0 +1,152 @@
...
+            datetime=timestamp,
+            production=ProductionMix(
+                hydro=hydro,
+                solar=solar,
+                wind=wind,
+                biomass=bagasse,
+                # NOTE: thermo includes gas + oil mixed, so we set these as unknown for now
+                # The modes here should match the ones we extract in the production payload
+                unknown=total - (hydro + solar + wind + bagasse),
+            ),
```
VIKTORVAV99, line 112: Why are we not assigning thermo directly to unknown?
Or are there additional missing production?

parsers/CNDC.py
```diff
@@ -0,0 +1,152 @@
...
+        timestamp = get_timestamp(query_date=date, hour=hour - 1)
+        modes_extracted = [hydro, solar, wind, bagasse]
+
+        if total is None or None in modes_extracted:
+            continue
+
+        result.append(
+            zoneKey=zone_key,
+            datetime=timestamp,
+            production=ProductionMix(
```
VIKTORVAV99, line 104: By not using the add value function we introduce floating point errors like this: `'unknown': 777.4199999999998,`
We could possibly change how the ProductionMix class works to avoid this or ose the add_value function everywhere.Either or I think we should avoid introducing these.

VIKTORVAV99 (commented):
Everything works as expected but I have a few comments before approving.

parsers/CNDC.py
unitrium: I think there might be additional unknowns we want to account for. This was already in the original parser.

parsers/CNDC.py
unitrium: I think I would rather update the __set__attr method of the Mix class so that it's coherent with the add method. I will do a follow up PR to address this.

VIKTORVAV99 (approved):

PR #6275, opened 2023-12-22
refactor: remove arrow from SV parser

yujia21:
## Issue

[6135](https://github.com/electricitymaps/electricitymaps-contrib/issues/6135)

## Description

Removes arrow module from parser, replacing with `datetime`.

### Double check


parsers/SV.py
```diff
@@ -133,9 +135,9 @@ def data_processer(data) -> list:
     joined_data = sorted(d.values(), key=itemgetter("datetime"))
 
     def get_datetime(hour):
-        at = arrow.now("UTC-6").floor("hour")
-        dt = (at.replace(hour=int(hour), minute=0, second=0)).datetime
-        return dt
+        return datetime.now(TIMEZONE).replace(
+            hour=int(hour), minute=0, second=0, microsecond=0
+        )
```
VIKTORVAV99, line 140: The previous version had a static offset of -6 hours from UTC, are you sure the new timezone will match this year round? Otherwise it would be better to add a static offset with datetime.timezone() instead.

yujia21: Good point, just checked! El Salvador no longer uses daylight savings (https://www.timeanddate.com/time/change/el-salvador)

VIKTORVAV99 (approved):
LGTM! 🥳

Thanks for doing this!

PR #6280, opened 2023-12-22
refactor: remove arrow from TR parser

yujia21:
## Issue

[6135](https://github.com/electricitymaps/electricitymaps-contrib/issues/6135)

## Description

Technically the `.replace(tzinfo=TR_TZ)` isn't necessary as the string to be converted contains "+03:00", but to keep the exact same tzinfo as master branch I included the replace.

### Double check


parsers/TR.py
```diff
@@ -37,6 +36,16 @@
...
+def _str_to_datetime(date_string: str) -> datetime:
+    """
+    Converts string received into datetime format.
+    String received is almost in isoformat, just missing : in the timezone
+    """
+    return datetime.fromisoformat(date_string[:-2] + ":" + date_string[-2:]).replace(
+        tzinfo=TR_TZ
+    )
+
+
```
VIKTORVAV99, line 48: I would wrap this in a try except cause so we can get some clear error logging if this fails (in case they "fix" the format). I would also remove the tzinfo replacement for the same reason. Since they include it in the data we should use it (in case it changes in the future).

VIKTORVAV99 (changes requested):
Looks good, just one comment to increase the robustness a bit since the data source don't return a true ISO format datetime.

VIKTORVAV99 (approved):
LGTM! Thanks for doing this!

PR #6283, opened 2023-12-22
Add CN-MN exchange and parser

silimotion:
## Description

This PR adds the CN->MN exchange. To calculate the exchange, I used the total imports data from the MN source, and the MN->RU-2 from the Russian source. This is based on the assumption that Mongolia only trades with this two countries. I think this is a reasonable assumption, given geography and Mongolian grid.
### Preview

Screenshot_20231222_130854

``` bash
Parser result:
[{'datetime': datetime.datetime(2023, 12, 22, 19, 0, tzinfo=zoneinfo.ZoneInfo(key='Asia/Ulaanbaatar')),
  'netFlow': 619.82134,
  'sortedZoneKeys': 'CN->MN',
  'source': 'https://ndc.energy.mn/',
  'sourceType': <EventSourceType.measured: 'measured'>}]
---------------------
took 4.39s
min returned datetime: 2023-12-22 11:00:00+00:00 UTC
max returned datetime: 2023-12-22 11:00:00+00:00 UTC  -- OK, <2h from now :) (now=2023-12-22T12:10:49+00:00 UTC)
```

### Double check


parsers/MN.py
```diff
@@ -143,6 +145,46 @@ def fetch_consumption(
...
+            data["datetime"] == query_data["time"]
+            and data["sortedZoneKeys"] == "MN->RU-2"
+            and "netFlow" in data.keys()
+        ):
+            exchange_list.append(
+                datetime=query_data["time"],
+                zoneKey=sorted_zone_keys,
+                # We calculate the flow with the total imports and the MN->RU-2 exchange, as Mongolia has only two connections
+                netFlow=query_data["importMW"] + data["netFlow"],
+                source="https://ndc.energy.mn/",
```
VIKTORVAV99, line 182: Not sure if you know this but we have created generic functions to do this and ensure everything match. You can asses it with
```python
exchange_list.merge_exchanges()
```
The only issue is that it only accepts a list of `ExchangeList` object to merge.

VIKTORVAV99 (commented):
We have already created some of this functionality to ensure it's consistent across parsers but it don't work until the RU.py parser has been upgraded. So I would like to hold of with this until that is upgraded.

(Some of the background for that is that we are doing a code freeze over the holidays anyhow so while we will be merging things in contrib they won't be deployed until early jan so there is no hurry to get this merged.)

parsers/MN.py
silimotion: I am not sure if I have understood correctly, but isn't this function meant to merge ExchangeLists of the same sortedzones? In this case I am substracting the russian connection to the total.

VIKTORVAV99:
I could not get this to work when I tested it as no matching data could be found, I'll try again tomorrow.

parsers/MN.py
VIKTORVAV99, line 182:
```suggestion
                netFlow=query_data["importMW"] - data["netFlow"],
```
The current implementation seems to be inversed.
Please double check this yourself.

It would also be good if we could add a test or two for this so we can validate the expected behavior.

silimotion: > seems to be inversed

I think it isn't.
Let's see:
CN->MN + RU->MN = Total Imports
CN->MN = Total Imports + MN->RU

I'll work on those tests. It would be optimal to be able to test the parser without requiring the sources to have matching data.

VIKTORVAV99: the easiest way would probably be to add snapshot tests using mock data. Like I did in #6285.

That way it don't rely on an online source, changing data or creating manual tests are are both logic intensive and hard to maintain long term.

silimotion: I was thinking of doing something similar to what I did for the REE parser but I will take a look at these snapshot tests.

VIKTORVAV99:
@florianscheidl I just come to think about your project in relation to this. Would it be useful for you to know the total imports/exports in some form?

If so perhaps we can begin shaping the requirements for it as we don't store that at the moment? (MN is not the only zone that has this data in some form.)

florianscheidl:
Hey, @VIKTORVAV99, that would be interesting to pursue! Let's chat about it some time soon.

PaulRoms:

PR #6285, opened 2023-12-22
Add snapshot test for ENTSO-E exchange functions

VIKTORVAV99:
## Issue

Low testing cover for the ENTSO-E parser.

## Description

This PR adds a snapshot test for the fetch_exchange and the fetch_exchange_forecast function in the ENTSO-E parser.

### Double check

- [x] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [x] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

PR #6293, opened 2023-12-27
Upgrade ENTSOE exchange parsing to ExchangeList

VIKTORVAV99:
## Issue

The ENTSO-E parser has not been updated to ExchangeLists yet and contain a lot of duplicate code.

## Description

This PR updates the exchange functions to use ExchangeLists and unifies the parsing to between the two functions so that it's easier to maintain in the future.

> [!IMPORTANT]
> This PR targets the branch of #6285, which should be merged first!

### Double check


parsers/test/snapshots/snap_test_ENTSOE.py
VIKTORVAV99, line 1: All the date times in here are inverted so it was not as easy to review as I hoped but they should match.

I guess we could change the ordering of the output for the tests in #6285 so we get a better diff.

VIKTORVAV99:
Marking this as blocked as it breaks the `FR-COR_IT-SAR.py` wrapper that aggregate two exchanges. Will add a PR to do this directly in the ENTSOE parser instead that targets this branch.

Once both have been reviews the second PR can be merged into this one and then both be merged to master without any breaking changes.

VIKTORVAV99:
@unitrium would you prefer if I merged #6294 into this and have one big PR or would it be easier to review as it is now?

unitrium (approved):
Great job regrouping all under the ENTSO-E parser!

PR #6294, opened 2023-12-28
Modify raw exchange function to handle aggregate exchanges in ENTSOE

VIKTORVAV99:
## Issue

We currently have a wrapper that merges the AC and DC exchanges between FR-COR and IT-SAR, this will break when we migrate to ExchangeLists as they will validate the ZoneKeys for the two exchanges, and will fail as they don't exist in the exchange config files.

## Description
This PR creates a way to mark exchanges as aggregates directly in the ENTSOE parser and modifies the function created in #6293 to loop over domain pairs instead of just accepting one pair to fetch. This is done in a way that don't alter the existing behaviour.

> [!IMPORTANT]
> This PR targets the #6293 PR to avoid a massive diff, but should be merged into #6293 after that and this has been reviewed in order to make sure that we don't break the existing way we fetch this exchange.

### Double check


VIKTORVAV99:
This works as is but I will add more snapshot tests tomorrow in order to make sure we have some regression testing in place for future code changes.

parsers/ENTSOE.py
```diff
@@ -231,6 +223,14 @@
     "RU-1->UA": [ENTSOE_DOMAIN_MAPPINGS["RU"], ENTSOE_DOMAIN_MAPPINGS["UA-IPS"]],
     "SK->UA": [ENTSOE_DOMAIN_MAPPINGS["SK"], ENTSOE_DOMAIN_MAPPINGS["UA-IPS"]],
 }
+
+EXCHANGE_AGGREGATES: dict[str, list[list]] = {
+    "FR-COR->IT-SAR": [
+        [ENTSOE_DOMAIN_MAPPINGS["IT-SACOAC"], ENTSOE_DOMAIN_MAPPINGS["IT-SAR"]],
+        [ENTSOE_DOMAIN_MAPPINGS["IT-SACODC"], ENTSOE_DOMAIN_MAPPINGS["IT-SAR"]],
+    ],
```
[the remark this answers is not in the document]
VIKTORVAV99: I agree with this but I decided it was best to match the other list formats for now.
I will create a separate PR that focuses on just cleanup of this and better typing after these ones are merged.

parsers/ENTSOE.py
```diff
@@ -1208,61 +1208,69 @@ def get_raw_exchange(
...
+        domain1, domain2 = domain_pair if is_import else domain_pair[::-1]
+        try:
+            raw_exchange = query_function(domain1, domain2, session, target_datetime)
+        except Exception as e:
+            raise ParserException(
+                parser="ENTSOE.py",
+                message=f"Failed to query {'import' if is_import else 'export'} for {domain1} -> {domain2}",
+                zone_key=sorted_zone_keys,
+            ) from e
+        if raw_exchange is not None:
```
unitrium, line 1247: A personal preference of mine would be to just raise if it's None and carry on with the rest.

unitrium (approved):
Looks good to me!

parsers/ENTSOE.py
VIKTORVAV99: I'll change that when I add the new snapshot tests.

PR #6296, opened 2023-12-29
fix(Taipower parser down): update number of expected columns in the TW parser

unitrium:
## Issue

The Taipower parser is currently down because the source returns an additional column now. Seems that it is an empty column but it's preventing the parser from parsing the data.
image

## Description

- Update the parser to handle this extra column. We don't use this column.
- Update the mock data with the updated source data.
- Update the snapshot.

### Double check


VIKTORVAV99 (approved):
LGTM but I don't think Mads wanted to do any deployments until the code freeze is over. Just so you know.

PR #6299, opened 2023-12-30
Upgrade ENTSOE Consumption and Generation to new parser functions

VIKTORVAV99:
## Issue

The consumption and generation parsers don't currently use the new parser functions.

## Description

Modifies the code to use the new parser functions for consumption and generation. As well as snapshot tests for them all of them.

### Double check


parsers/ENTSOE.py
```diff
@@ -825,50 +827,6 @@ def parse_production(
...
-
-        for entry in timeseries.find_all("point"):
-            quantity = float(entry.find_all("quantity")[0].contents[0])
-            if quantity == 0:
-                continue
-            position = int(entry.find_all("position")[0].contents[0])
-            datetime = datetime_from_position(datetime_start, position, resolution)
-            res[datetime] = res[datetime] + quantity if datetime in res else quantity
-
-    return res
```
VIKTORVAV99, line 869: This don't work at all for realtime consumption as far as I can tell and all zones I have checked returned nothing and it is the number 1 cause of warning logs for the ENTSOE zones.

If it's actually required I can add it back but the current implementation made it so we only return 1 datapoint for the consumption function for realtime data.

unitrium (approved):
Sorry for overlooking this. Looks good to me!

PR #6245, opened 2023-12-13
[REF] EU emission factors 2022

mathilde-daugy:
Issue
Update US emission factors with EU-ETS 2022 values. The process is the same as last year, see [wiki](https://github.com/electricitymaps/electricitymaps-contrib/wiki/EU-emission-factors) for documentation

Computation notebook: [EU_emission_factors.ipynb](https://colab.research.google.com/drive/1buA-yHQCF711F53IX7_GAxYMJ9p4qJoT?usp=sharing)

Description
Updated yaml files for all EU zones that have REF

Preview
NA

Double check
 I have tested my parser changes locally with poetry run test_parser "zone_key"
 I have run pnpx prettier --write . and poetry run format to format my changes.

PR #6308, opened 2024-01-02
[REF] update EU emission factor and remove outliers

mathilde-daugy:
## Issue

see comment on PR #6245, gas factor was too high for DE. 
I've generated the emission factors and used a lower upper bound for gas.
The default coal factor is also updated (also because of a modification of the validation limits)


VIKTORVAV99 (approved):
LGTM!

I am curious though, did we incorporate any of w-flo's work in the repo that was linked? I think it included a few more power plant mappings that would give us more accurate data.

mathilde-daugy:
> LGTM!
> 
> I am curious though, did we incorporate any of w-flo's work in the repo that was linked? I think it included a few more power plant mappings that would give us more accurate data.

Unfortunately this is not something i have had the time to do, he has done a lot of work and i need to prioritise this but haven't yet done so :(

cgicgi:
thanks a lot for this quick reaction!

PR #6319, opened 2024-01-06
pre-commit-config

VIKTORVAV99:
## Description

Adds a pre-commit-config so people can install pre-commit hooks locally to reduce CI runs and code churn.

This is done by installing [`pre-commit`](https://pre-commit.com) with for example `brew` and then running `pre-commit install`.

For now this just includes ruff.

### Double check


VIKTORVAV99:
Not sure if this was what you hand in mind @madsnedergaard but this seems to be the cleanest solution for ruff at least.

madsnedergaard (approved):
Nice :)

PR #6325, opened 2024-01-08
Total FI-SE exchange capacity for country view

msaynevirta:
## Issue

FI-SE exchange capacity is missing in the country view.

## Description

Combined capacity of FI-SE1 and FI-SE3 has been added to the FI-SE config file.

### Double check


VIKTORVAV99 (approved):
Thanks for adding these!

PR #6331, opened 2024-01-10
Update EFs, capacity, lowCarbon, isRenewable

florianscheidl:
## Issue

See [ELE-3561](https://linear.app/electricitymaps/issue/ELE-3561/updated-ch-emission-factors)

Emission factors, capacity, lowCarbon, isRenewable values were out of date for Switzerland

## Description

Pulled new data, (see SFOE, Electricity Maps [source](https://docs.google.com/spreadsheets/d/1FLHQ6e9Es08BIqX654BM3SEb_fuAk4k4O-6cmRXyw_E)), which updated values from previous years.

### Double check

### DIFF
--- a/config/zones/CH.yaml
+++ b/config/zones/CH.yaml
@@ -12,7 +12,7 @@ capacity:
   hydro storage: 4605
   nuclear: 3014.6
   oil: 0.4
-  solar: 3673.3
+  solar: 4736.7
   wind: 88.4
 contributors:
   - corradio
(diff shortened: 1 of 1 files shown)


config/zones/CH.yaml
```diff
@@ -342,22 +342,22 @@ isRenewable:
   unknown:
     - datetime: '2017-01-01'
       source: SFOE, Electricity Maps
-      value: 0.53
+      value: 0.72
```
VIKTORVAV99, line 345: This is quite a large change, do you know why that might be?

VIKTORVAV99 (approved):
PR looks good but I have a question about the values themselves as it looks to be some really big changes. Do we know why?

config/zones/CH.yaml
florianscheidl: Yes, in the source file, I think we treated the total unknown **thermal** power production (which includes non-renewable as well as renewable = biomass sources) as non-renewable energy source. Since biomass makes up a significant share of this and has a much lower emission factor, this led to the decrease in emission factors (and renewable and low carbon shares). Happy to jump on a brief call if you'd like to review these findings 😃

VIKTORVAV99: Ah then it makes perfect sense!
No need for a call as this is not the first time this has happened 😅.

PR #6334, opened 2024-01-11
Enable ruff for the test files

VIKTORVAV99:
## Issue

We where not running ruff on the test files.

## Description

This PR enables that and fixes some of the issues in the files and ignores the false positives due to star imports.

### Double check


unitrium (approved):
Seems good to me

Issue #6339, opened 2024-01-12
Battery capacity in Belgium


kroeckx:
On https://www.elia.be/en/publications/reports, it says Belgium has 50 MW / 100 MWh of battery installation (on 2023-10-01). Not sure how to add it to the config.

PR #6340, opened 2024-01-12
Add Belgian battery capacity

kroeckx:
Fixes: #6339

### Double check

I have no idea if "battery: " is correct syntax. I suspect it's not, since I can't find any examples with that and that there is currently just no way to specify it's the power output, only the the storage capacity.


config/zones/BE.yaml
```diff
@@ -15,6 +15,8 @@ capacity:
   oil: 455
   solar: 8595
   wind: 5315
+  battery: 50
+  battery storage: 100
```
VIKTORVAV99, line 19:
```suggestion
  battery storage: 50
```
We don't support the storage values and only the discharge value should be added, at least for now.

We also need to add the source to `DATA_SOURCES.md` before we can merge this.

kroeckx:
I've pushed a new version.

DATA_SOURCES.md
```diff
@@ -198,6 +198,7 @@ For many European countries, data is available from [ENTSO-E](https://transparen
   - Solar: [ELIA](https://www.elia.be/en/grid-data/power-generation/solar-pv-power-generation-data)
   - Hydro: [ELIA](https://www.elia.be/-/media/project/elia/shared/documents/elia-group/publications-pdfs/20200414_elia_annual-report-sustainability_en.pdf)
   - Wind: [ELIA](https://www.elia.be/en/grid-data/power-generation/wind-power-generation)
+  - Battery: [ELIA](https://www.elia.be/en/publications/reports)
```
VIKTORVAV99, line 201: It would be great if we could specify which report it was here. They seem to have quite a few 😅

kroeckx:
It's on that page itself.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for adding it!

PR #6356, opened 2024-01-16
Update PT solar capacity.

Musketeiro:
Updated the solar capacity for PT. Figures are taken from REN which is already on the DATA_SOURCES.md file.

### DIFF
--- a/config/zones/PT.yaml
+++ b/config/zones/PT.yaml
@@ -12,7 +12,7 @@ capacity:
   hydro storage: 3585
   nuclear: 0
   oil: 0
-  solar: 2141
+  solar: 2611
   unknown: 28
   wind: 5374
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
Thanks for updating these!

VIKTORVAV99:
We are now getting this directly from ENTSO-E to streamline the process and ensure we have versioned data. I do see that the values have been updated though so I'll update these to the latest available in ENTSO-E. It's possible not all installed capacity is reported as either production or capacity but that's on the TSOs side that report data to ENTSO-E and nothing we have control over. We need these to be kept in line for better outlier detection and estimations.

Musketeiro:
Alright. Thank you for the reply.

PR #6361, opened 2024-01-18
fix(parser): Fix NTESMO parser's datetime issue

unitrium:
## Issue

When running `test-parser AU-NT` we get an error because the datetime is None. This is due because we pass a None in test-parser and this parser doesn't support it.

## Description

Replacing the way None are interpreted in the AU-NT parser and removing arrow in the same instannce.


VIKTORVAV99 (approved):
LGTM!

Nice to get rid of arrow as well!

PR #6363, opened 2024-01-19
QA capacity update

florianscheidl:
## Issue

The capacity data for Qatar lacked gas and the values and the values for biomass and solar are outdated.

## Description

Updated capacity values from 2021, IRENA data (see https://pxweb.irena.org/pxweb/en/IRENASTAT/IRENASTAT__Power%20Capacity%20and%20Generation/ELECCAP_2023_cycle2.px/).


config/zones/QA.yaml
```diff
@@ -3,9 +3,10 @@ bounding_box:
     - 24.059871521000062
   - - 52.11654707100007
     - 26.660101630000042
-capacity:
-  biomass: 38
-  solar: 5
+capacity: # https://docs.google.com/spreadsheets/d/1ownS3qse6zDG4gNnKEZDroIKLNiqZcULqp34LJUxrL0/edit?usp=sharing
+  biomass: 19
+  gas: 10576
+  solar: 5.1
```
VIKTORVAV99, line 9: Do we want to version the capacity to only be for 2021 now that we have that capability or does that not matter in this case?

florianscheidl: That might make sense :)
What syntax do we use to version the capacities?

VIKTORVAV99: Here is a example:
```yaml
apacity:
  battery storage:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 3166.3
  biomass:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 170.1
  coal:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 14329.7
  gas:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 65837.7
  geothermal:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 0.0
  hydro:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 558.0
  hydro storage:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 0.0
  nuclear:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 5138.6
  oil:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 786.5
  solar:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 13478.1
  unknown:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 229.7
  wind:
    datetime: '2023-10-01'
    source: EIA.gov
    value: 37042.2
 ```
 And since each value is valid from its datetime to the next datetime added we only need to add one value for solar from 2016-01-01. And biomass and gas from 2021-01-01

florianscheidl: Ah, perfect, will do that 😃

PR #6369, opened 2024-01-21
refactor: upgrade CA-ON with event classes

kruschk:
# Issue

Refs: #6011

# Description

- Upgrade the parser to use appropriate event classes. This should yield no functional change.
- Replace arrow with equivalent functionality from datetime.
- Tidy up and refactor the existing code.
- Transform the source data manually rather than relying on pandas.

# Double check


parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
 )
-EXCHANGES_URL = "http://reports.ieso.ca/public/IntertieScheduleFlow/PUB_IntertieScheduleFlow_{YYYYMMDD}.xml"
-
-XML_NS_TEXT = "{http://www.theIMO.com/schema}"
+PRODUCTION_URL = "http://reports.ieso.ca/public/GenOutputCapability/PUB_GenOutputCapability_{YYYYMMDD}.xml"
+SOURCE = "ieso.ca"
+# IESO says "Eastern Standard Time is used year round." This means daylight
+# savings is not used (that is called "Eastern Daylight Time"), and we need to
+# use UTC-5 rather than 'Canada/Eastern'.
+TIMEZONE = datetime.timezone(datetime.timedelta(hours=-5), name="UTC-5")
```
VIKTORVAV99, line 77: I would prefer it if we imported datetime.datetime, timezone, and timedelta seperatly to avoid cluttering up lines like this.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
     logger: Logger,
-    url_template,
-):
-    dt = (
-        arrow.get(target_datetime)
-        .to(tz_obj)
-        .replace(hour=0, minute=0, second=0, microsecond=0)
+    session: Session | None,
+    target_datetime: datetime.datetime | None,
+    template: str,
```
VIKTORVAV99, line 85: I would rename this to make it more clear, perhaps even split out the URL formatting from the fetch function.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
-    generators = (
-        xml.find(XML_NS_TEXT + "IMODocBody")
-        .find(XML_NS_TEXT + "Generators")
-        .findall(XML_NS_TEXT + "Generator")
-    )
+    # Collect the source data into a dictionary keying ProductionMix objects by
+    # the time of day at which they occurred.
+    mixes = {}
+    for generator in xml.iter(NAMESPACE + "Generator"):
+        mode = MODES[generator.findtext(NAMESPACE + "FuelType")]
```
VIKTORVAV99, line 130: To avoid type errors this need to be checked so it's not None first.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+            ).add_value(
+                mode,
+                # Sometimes the XML data has no EnergyMW tag for a given plant
+                # at a given hour. The report XSL formats this as "N/A"; we
+                # return 0.
+                # TODO: Is XSL supposed to be XML? I've never seen an N/A in
+                # the response, and if one appeared we'd have to test for
+                # generation == "N/A", not generation is None. If the EnergyMW
+                # element is just absent, then the generation is None test
+                # makes sense.
```
VIKTORVAV99, line 144: Not sure what you mean here?
All the files returned seem to be XML files.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+                # TODO: Should we set this to None instead of 0?
+                0 if generation is None else float(generation),
```
VIKTORVAV99, line 146: If it's missing it should be None. The .add_value() can accept None values directly.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+    # Collect the source data into a dictionary keying ProductionMix objects by
+    # the time of day at which they occurred.
+    mixes = {}
+    for generator in xml.iter(NAMESPACE + "Generator"):
+        mode = MODES[generator.findtext(NAMESPACE + "FuelType")]
+        for output in generator.iter(NAMESPACE + "Output"):
+            generation = output.findtext(NAMESPACE + "EnergyMW")
+            mixes.setdefault(
+                datetime.time(hour=_parse_hour(output)), ProductionMix()
+            ).add_value(
```
VIKTORVAV99, line 135: Not sure I like this, it hides the types and makes it harder to maintain, debug and contribute long term.

I would probably try and zip the lists instead and then loop over the zipped list to add the values.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
-    # sort it for clean-ness and easier debugging.
-    data = sorted(data, key=lambda dp: dp["datetime"])
-
-    return data
+    return production_breakdowns.to_list()
 
 
 def fetch_price(
-    zone_key: str = "CA-ON",
+    zone_key: str = ZoneKey("CA-ON"),
```
VIKTORVAV99, line 162:
```suggestion
    zone_key: ZoneKey = ZoneKey("CA-ON"),
```

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+    # 5-minute price, but its archives only go back roughly 4 days (see "5
+    # Minute Market Clearing Price" at http://www.ieso.ca/power-data ).
+    prices = PriceList(logger)
+    for hoep in xml.iter(NAMESPACE + "HOEP"):
+        prices.append(
+            currency="CAD",
+            datetime=datetime.datetime.combine(
+                date, datetime.time(hour=_parse_hour(hoep)), tzinfo=TIMEZONE
+            ),
+            price=float(hoep.findtext(NAMESPACE + "Price")),
```
VIKTORVAV99, line 187: While the price argument handles None values the float does not.

I would assign the value to a variable and then check so that variable is not None.
Something like:
```python
price_string = hoep.findtext(NAMESPACE + "Price")
price = float(price_string) if price_string else None
```

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+            continue
+        for actual in intertie.iter(NAMESPACE + "Actual"):
+            # The IESO day starts with Hour 1, Interval 1. At 11:37, we might
+            # have data for up to Hour 11, Interval 12. This means it is
+            # necessary to subtract 1 from the values (otherwise we would have
+            # data for 12:00 already at 11:37).
+            time = datetime.time(
+                hour=_parse_hour(actual),
+                minute=5 * (int(actual.findtext(NAMESPACE + "Interval")) - 1),
             )
```
VIKTORVAV99, line 233: Same here, find tixt can return None which int can't handle.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
+            # In the source data, flows out of Ontario (i.e., exports) are
+            # positive. For us, positive flow follows the direction of the
+            # arrow in sorted_zone_keys, so change the sign of the flow if
+            # necessary.
+            flow = float(actual.findtext(NAMESPACE + "Flow"))
```
VIKTORVAV99, line 238: And here.

parsers/CA_ON.py
```diff
@@ -64,331 +55,265 @@
...
     session: Session | None = None,
-    target_datetime: datetime | None = None,
+    target_datetime: datetime.datetime | None = None,
     logger: Logger = getLogger(__name__),
-) -> list:
+) -> list[dict[str, Any]]:
     """Requests the last known power exchange (in MW) between two regions."""
 
-    sorted_zone_keys = "->".join(sorted([zone_key1, zone_key2]))
+    sorted_zone_keys = "->".join(sorted((zone_key1, zone_key2)))
```
VIKTORVAV99, line 204:
```suggestion
    sorted_zone_keys = ZoneKey("->".join(sorted((zone_key1, zone_key2))))
```

VIKTORVAV99 (commented):
Really good start!

There is some type errors and other small things that needs to be adressed before we can merge this but it seems to be working!

I would also love it if you could add some snapshot tests to make sure we are not causing a regression. (These should probably be added in a separate PR that is merged first).

parsers/CA_ON.py
kruschk: I was referring to the original comment, above, but I figured it out! It's saying that the expected `EnergyMW` element will occasionally be missing from the XML, and the table will show an "N/A" entry in its place when viewed in the browser. From the parsers' perspective, the `generation is None` test is correct.

parsers/CA_ON.py
kruschk: It's essentially just simulating a [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#defaultdict-objects); I could use one of those instead, if you prefer.

I'm not sure which lists you're referring to for zipping... :thinking:

VIKTORVAV99: Took a quick crack at what I meant and probably wouldn't work on it's own without grouping the values by the hour later on.

I really just want to do this in a way that perserves the type hinting for both the `ProductionMix()` and `.add_value()`, we plan on introducing strong typing to the CI at some point I really don't want to introduce potential erros when now that we are re-doing the parsers anyway.

kruschk: Ahh, I finally got `mypy` to work! I had to delete the `__init__.py` file in the root of the project; otherwise, `mypy` treats the root directory as a Python package and errors out. I added some type hints (to satisfy `mypy`) and switched over to `collections.defaultdict`; perhaps that makes the intent clearer?

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for upgrading it!

kruschk:
Thanks!

I agree that some automated tests would be a great help. They should be easy enough to implement for this parser, but I'd need to familiarise myself with the framework first. That said, I performed some spot checks between the previous implementation and this one, so I'm reasonably confident that they exhibit the same behaviour.

Issue #5121, opened 2023-02-23
Fix ESLint error in weather layers


VIKTORVAV99:
Currently ESLint is disabled for most of the weather layer files, we should strive to get ESLint enabled for these and fix the underlying errors instead.

qyearsley:
The files there that have eslint disabled

VIKTORVAV99:
> The files there that have eslint disabled (e.g. https://github.com/electricitymaps/electricitymaps-contrib/blob/master/web/src/features/weather-layers/wind-layer/windy.ts) have comments like:
> 
> ```
> // TODO: Rewrite completely
> // This file was taken from https://github.com/esri/wind-js, and modified
> ```
> 
> So, does this mean we want to refactor and keep the logic and behavior the same, but improve readability and style? Asking because I didn't want to potentially waste time trying to improve what's there if someone's gonna rewrite completely.

Hi @qyearsley!

Yeah we want to rewrite these but they seems to be really fragile so nobody has really taken the steps to do it yet. And yes the inputs and outputs need to stay the same.

If you want to take a crack at it I think the rest of the team will be very welcome to it as well but keep in mind this might me a lot more difficult than it initially seems.

So if you wish I'll happily assign you the issue if you want to work on it.

CC @madsnedergaard and @tonypls for visability.

qyearsley:
Makes sense, thanks.

It sounds like verifying that nothing is broken after each incremental change (even if it appears like it shouldn't affect behavior) would be a good idea; there are not currently unit tests for wind-layer/windy.ts or wind-layer/scales.ts; what's the most useful thing I can do if I try to make a change and want to verify that it should still work? `pnpm run mockserver` or something else?

VIKTORVAV99:
> Makes sense, thanks.
> 
> It sounds like verifying that nothing is broken after each incremental change (even if it appears like it shouldn't affect behavior) would be a good idea; there are not currently unit tests for wind-layer/windy.ts or wind-layer/scales.ts; what's the most useful thing I can do if I try to make a change and want to verify that it should still work? `pnpm run mockserver` or something else?

Honestly I'm not sure, we don't have any mock data for the solar or wind layers and contributors don't have access to the API for local development. 

I think it maybe it would be best to create unit tests first and then refractor the code. These tests should just check input and output as the internal code would change. I can probably add a snapshot of the weather data if you wish to proceed with this.

qyearsley:
[...]

PR #6374, opened 2024-01-22
Wind layer fixes

wrma6:
## Issue

Partial fix (wind layer side) for #5121

## Description
- Add proper typing to everything in `windy.ts`
- Refactor logic into using mostly classes rather than nested callbacks
- Cleanup a lot of uncessary logic and confusing variable names

Probably still could be improved more but wind layer code should finally be somewhat understandable.

### Double check


VIKTORVAV99:
This is great! Will try and dig into this PR next week but since this is a complex piece of code and there is a lot of changes that could take a while.

web/src/features/weather-layers/wind-layer/calc.ts
```diff
@@ -0,0 +1,91 @@
...
+  const rx = 1 - x;
+  const ry = 1 - y;
+  const a = rx * ry,
+    b = x * ry,
+    c = rx * y,
+    d = x * y;
+  const u = g00[0] * a + g10[0] * b + g01[0] * c + g11[0] * d;
+  const v = g00[1] * a + g10[1] * b + g01[1] * c + g11[1] * d;
+  return [u, v, Math.hypot(u, v)];
+};
```
VIKTORVAV99, line 26: It would be great if we could add some unit tests to these, there are a bunch of math going on and while I think they look correct it would be nice to have a test in place for it.

web/src/features/weather-layers/wind-layer/calc.ts
```diff
@@ -0,0 +1,91 @@
...
+  const projected = map.project([lon, lat]);
+  return [projected.x, projected.y];
+};
+
+/**
+ * Calculate distortion of the wind vector caused by the shape of the projection at point (x, y). The wind
+ * vector is modified in place and returned by this function.
+ * @returns {WindVector}
+ */
+export const distort = (
```
VIKTORVAV99, line 42: That goes for this function as well.

web/src/features/weather-layers/wind-layer/calc.ts
```diff
@@ -0,0 +1,91 @@
...
+  // Scale distortion vectors by u and v, then add.
+  wind[0] = d[0] * u + d[2] * v;
+  wind[1] = d[1] * u + d[3] * v;
+  return wind;
+};
+
+/**
+ * @returns {Bounds} puts boundary info into one object
+ */
+export const buildBounds = (bounds: number[][], width: number, height: number) => {
```
VIKTORVAV99, line 84: and this

VIKTORVAV99 (commented):
Overall as far as I can tell this looks good, but I would really love some new unit tests for all the new functions just to double check they are working as expected and to make sure they don't regress in the future.

Nice work!

wrma6:
@VIKTORVAV99 Added the tests you asked (one of the methods already had tests so I just copied), let me know what you think

wrma6:
Hey @VIKTORVAV99 sorry to ping for this, but is it possible for us to get this PR merged within the next 2 weeks or so? The reason for this is that I have to write a report about the work that I did so it would be cool to have everything in order before having to do that.

VIKTORVAV99:
> Hey @VIKTORVAV99 sorry to ping for this, but is it possible for us to get this PR merged within the next 2 weeks or so? The reason for this is that I have to write a report about the work that I did so it would be cool to have everything in order before having to do that.

I'll see what I can do!

VIKTORVAV99 (approved):
LGTM! 🥳

Thanks for doing this, it was very much needed!

Safari is still crap, but it was so before as well so no regression, and it seems like Chromium based browsers works as intended at least so I'm going to merge this 🙂

Nice work!

wrma6:
Thank you very much for the quick review time!

PR #6379, opened 2024-01-22
Support unknown/generic estimation models in text

madsnedergaard:
## Description

Adds support for a shared/generic description, in case a new model is added before we have time to update translations on the app.

### Preview

Screenshot 2024-01-22 at 14 26 03


madsnedergaard:
cc @silkeholmebonnen, I added a fallback for unknown estimation models and updated text text with translations done by the team :)

VIKTORVAV99 (approved):

PR #6380, opened 2024-01-22
New and improved data labelling! 🏷️ 

madsnedergaard:
## Issue

The current "estimated" badge is not very clear and figuring out what it actually means requires digging deep into the GitHub Wiki.

## Description

This makes it much easier for understand **why** some data is estimated and hopefully also means more people realise that data might not be completely accurate due to missing data from the source.

**Kudos to @silkeholmebonnen for doing all the code here and @Alportan for the design 🙌**
Due to ongoing discussions around this topic these days, we have decided to release this change earlier instead of polishing it :)

### Preview
Screenshot 2024-01-22 at 14 32 58
Screenshot 2024-01-22 at 14 33 05
Screenshot 2024-01-22 at 14 33 13

Screenshot 2024-01-22 at 14 34 07

**Note:** We will expand upon this in the coming days with more translations, improved text for more estimation models and more.


web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+  cardType: string;
+  estimationMethod: string | undefined;
+  outageMessage: ZoneDetails['zoneMessage'];
+}) {
+  if (cardType == 'outage') {
+    return <OutageCard outageMessage={outageMessage} />;
+  } else if (cardType == 'aggregated') {
+    return <AggregatedCard />;
+  } else if (cardType == 'estimated') {
+    return <EstimatedCard estimationMethod={estimationMethod} />;
```
VIKTORVAV99, line 21: This should probably be a switch case, but let's skip that in this PR.

VIKTORVAV99 (approved):
LGTM!

web/src/features/panels/zone/EstimationCard.tsx
madsnedergaard: Yes and yes :D

PR #6383, opened 2024-01-22
Add estimation drift message for filtered threshold zone

VIKTORVAV99:
Feel free to rework the wording how you see fit. 🙂

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -122,6 +122,11 @@
       "pill": "Not realtime",
       "body": "The data for this zone is not published in realtime and only provides the total production. Both hourly production and production sources are estimated based on historical data for each production source."
     },
+    "threshold_filtered": {
+      "title": "Estimation drift",
+      "pill": "Hidden",
+      "body": "The estimations for this zone has begun to drift too far from the actual values. The data is therefore not shown on the map."
+    },
     "outage": {
       "title": "Ongoing issues",
       "pill": "Unavailable"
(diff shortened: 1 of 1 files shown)


web/public/locales/en.json
```diff
@@ -122,6 +122,11 @@
       "pill": "Not realtime",
       "body": "The data for this zone is not published in realtime and only provides the total production. Both hourly production and production sources are estimated based on historical data for each production source."
     },
+    "threshold_filtered": {
+      "title": "Estimation drift",
+      "pill": "Hidden",
+      "body": "The estimations for this zone has begun to drift too far from the actual values. The data is therefore not shown on the map."
```
q--, line 128: 'the estimations (...) have' or 'the estimation (...) has', if I'm not mistaken

VIKTORVAV99: Nice catch! Thanks!
I'll update this tomorrow.

madsnedergaard, line 128: I think it's worth focusing more on the end-user experience, as they technically won't know that we have disabled estimations (or even had it in the first place). They just see a grey map, so I think we should focus on explaining the situation instead:

```suggestion
      "title": "Missing data",
      "pill": "Data outage",
      "body": "The data provider for this zone has been missing data for a prolonger period. We are monitoring the issue and will update the app as soon as data becomes available again."
```

VIKTORVAV99: Maybe something like this would work better in that case:
```
"The data provider for this zone has been missing data for a prolonged period. We are monitoring the issue and the zone will update as soon as data becomes available again."
```
Less emphasis on us updating the app and more emphasis on that it affects just that zone?

madsnedergaard: Makes sense, let's go with that!

madsnedergaard (approved):

PR #6384, opened 2024-01-22
Update Singapore solar capacity

q--:
Source: https://www.ema.gov.sg/consumer-information/solar/solar-irradiance-map

This is also the page used to fetch the solar data by the production parser.

---
I also updated the capacity data source, as the old URL 404's.


VIKTORVAV99 (approved):
Thanks for updating these!

PR #6386, opened 2024-01-23
update(capacity) - DE 2022 & 2023

mathilde-daugy:
Capacity update for DE for 2022 and 2023. Date for nuclear is the date of the shutdown of the last PP


VIKTORVAV99 (approved):

unitrium:
Hey,
I just noticed when running update contrib that the geothermal capcity disapeared. Is it intentional ?

mathilde-daugy:
forgot to check that actually. It looks like we getting 21 MW from ENTSOe continuously so i will update

PR #6387, opened 2024-01-23
Set auto_improve to false

VIKTORVAV99:
## Issue

We are getting a lot of suggestions that are not really valid as it don't seem to consider the whole codebase or how we do things internally.

## Description

Changes the auto improve to `false` and keeps the auto review parts.


madsnedergaard:
Oh, I see you both changed it, already merged #6389 :)

PR #6388, opened 2024-01-23
improvement(parsers): Refactor SV parser

unitrium:
## Issue
Refactoring the SV parser to make it compatible with the new datastructure. Also the parser was quite obscure as to how it parsed the data, so took some time to clarify all of it. It looks like we could also get the exchange data which would be addressed in a follow up PR.
Also the thermal energy was not used previously.

## Description

Refactor the parser with a more readable parsing and mapping of the different modes and dates.
Map thermal to unknown.
Add testing.

### Double check


VIKTORVAV99:
I'll give this a review later today 🙂

unitrium:
Not sure why I get all those changes. Will investigate

parsers/ESTADISTICO_UT.py
```diff
@@ -0,0 +1,194 @@
...
+from requests import Response, Session
+
+from electricitymap.contrib.lib.models.event_lists import (
+    ProductionBreakdownList,
+    ProductionMix,
+)
+from electricitymap.contrib.lib.types import ZoneKey
+
+# This parser gets hourly electricity generation data from ut.com.sv for El Salvador.
+# El Salvador does have wind generation but there is no data available.
```
q--, line 19: This line is obsolete

parsers/ESTADISTICO_UT.py
```diff
@@ -0,0 +1,193 @@
...
+                hour=int(hour),
+                tzinfo=TIMEZONE,
+            )
+            double_json = double_json.replace(
+                f"new Date({year},{month},{day},{hour},{minute},{second},{ms})",
+                f'\\"{dt.isoformat()}\\"',
+            )
+    data = json.loads(double_json)
+    jsresult = data["result"]
+    clean_json = json.loads(jsresult[1:-1])
```
q--, line 110: By the way, if you keep the old data source there's also a CSV/XLS/XLSX export that might be slightly easier to work with: 
'Export to Excel' menu on the Generación por tipo de tecnología table

parsers/test/test_ESTADISTICO_UT.py
```diff
@@ -0,0 +1,49 @@
...
+        self.adapter = Adapter()
+        self.session.mount("https://", self.adapter)
+        self.session.mount("http://", self.adapter)
+        self.adapter.register_uri(
+            GET,
+            DAILY_OPERATION_URL,
+            text=resources.files("parsers.test.mocks.ESTADISTICO_UT")
+            .joinpath("production.html")
+            .read_text(),
+        )
```
VIKTORVAV99, line 23: I have been using Path from pathlib instead of resources, I like that it has a shorter syntax.

Not that it really matters but it would probably be nice if we decided on which to use for consistency across the test files.

parsers/test/test_ESTADISTICO_UT.py
```diff
@@ -0,0 +1,49 @@
...
+            [
+                {
+                    "datetime": element["datetime"].isoformat(),
+                    "production": element["production"],
+                    "storage": element["storage"],
+                    "source": element["source"],
+                    "zoneKey": element["zoneKey"],
+                }
+                for element in production
+            ]
```
VIKTORVAV99, line 48:
```suggestion
            [
                {
                    "datetime": element["datetime"].isoformat(),
                    "production": element["production"],
                    "storage": element["storage"],
                    "source": element["source"],
                    "zoneKey": element["zoneKey"],
                    "sourceType": element["sourceType"].value,
                }
                for element in production
            ]
```

VIKTORVAV99 (commented):
Just some notes, but it looks like things are working as expected.
Nice work!

parsers/ESTADISTICO_UT.py
```diff
@@ -0,0 +1,193 @@
...
+def data_processer(
+    zone_key: ZoneKey, data: list[dict], logger: Logger
+) -> ProductionBreakdownList:
+    """
+    Takes data in the form of a list of lists.
+    Converts each list to a dictionary.
+    Joins dictionaries based on shared datetime key.
+    Maps generation to type.
+    """
+    per_mode_production: dict[str, ProductionBreakdownList] = {}
```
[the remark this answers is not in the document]
unitrium: The modes are returned each with their own timestamp in an unsorted way so it's a bit hard to add them correctly, I feel this is the easiest, I don't think we can solve this with a change in the model.

parsers/test/test_ESTADISTICO_UT.py
unitrium: Yeah I followed the same structure as most tests. Since this is rather a preference than an implementation design I would leave it to the dev to decide.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #6393, opened 2024-01-24
add capacity-source-to-zones config

mathilde-daugy:
## Issue

The list of zones is currently hardcoded in the capacity parsers which makes it hard to maintain. If the list in not up to date, we can update data for the wrong zones. The source should be extracted from the configuration files. 

## Description
- Created `CAPACITY_PARSER_SOURCE_TO_ZONES` mapping that returns the source as key and a list of all zones using the source as values 
- **CAVEAT** i added this mapping in the `__init__`  of `capacity_parsers` because i'm unsure of where is the best place for it. Open to suggestions though :)
- updated EIA, ENTSOE, IRENA and EMBER capacity parsers to use this mapping


electricitymap/contrib/capacity_parsers/__init__.py
VIKTORVAV99, line 1: Perhaps it would be better to create a set of utilities instead of adding it here.

I suggest a subfolder to capacity parsers and then adding this there instead and I would rework it so it's a function instead of a variable so it's run only when called.

electricitymap/contrib/config/capacity.py
```diff
@@ -1,4 +1,25 @@
...
+CAPACITY_PARSER_SOURCE_TO_ZONES = {}
+for zone_id, zone_config in ZONES_CONFIG.items():
+    if zone_config.get("parsers", {}).get("productionCapacity") is None:
+        continue
+    source = zone_config.get("parsers", {}).get("productionCapacity").split(".")[0]
+    if source not in CAPACITY_PARSER_SOURCE_TO_ZONES:
+        CAPACITY_PARSER_SOURCE_TO_ZONES[source] = []
+    CAPACITY_PARSER_SOURCE_TO_ZONES[source].append(zone_id)
+
+ZONE_TO_CAPACITY_PARSER_SOURCE = {}
```
unitrium, line 19: I think it would be better performance wise to do it with list comprehension.
```suggestion
ZONE_TO_CAPACITY_PARSER_SOURCE = {
    zone: source
    for source, zones in ZONES_CONFIG.items()
    for zone in zones
}
```

unitrium (approved):
Small suggestion

electricitymap/contrib/config/capacity.py
```diff
@@ -1,4 +1,27 @@
...
+        continue
+    source = zone_config.get("parsers", {}).get("productionCapacity").split(".")[0]
+    if source not in CAPACITY_PARSER_SOURCE_TO_ZONES:
+        CAPACITY_PARSER_SOURCE_TO_ZONES[source] = []
+    CAPACITY_PARSER_SOURCE_TO_ZONES[source].append(zone_id)
+
+ZONE_TO_CAPACITY_PARSER_SOURCE = {
+    zone: source for source, zones in ZONES_CONFIG.items() for zone in zones
+}
+for source, zones in CAPACITY_PARSER_SOURCE_TO_ZONES.items():
```
unitrium, line 22:
```suggestion
```

electricitymap/contrib/config/capacity.py
```diff
@@ -1,4 +1,27 @@
...
+    for zone in zones:
```
unitrium, line 23:
```suggestion
```

electricitymap/contrib/config/capacity.py
```diff
@@ -1,4 +1,27 @@
...
+        ZONE_TO_CAPACITY_PARSER_SOURCE[zone] = source
```
unitrium, line 24:
```suggestion
```

unitrium:
Before merging, remember to remove the duplicated code

PR #6402, opened 2024-01-25
Update package dependencies and testing

VIKTORVAV99:
Updates vite + vite assosicated packages to new major versions.

Ran pnpm upgrade in web and mockserver to upgrade all patch versions.

Upgrades nodemon in mockserver to a new major version.


tonypls (commented):
Looks good! Not sure why the tests fail, ping me if there's further changes to review to get them passing 🙏

VIKTORVAV99:
> Looks good! Not sure why the tests fail, ping me if there's further changes to review to get them passing 🙏

Looks like react-map-gl was bumped to a new minor version instead of a patch which had new type definitions (or added type definitions). I downgraded it to just a patch upgrade instead.

VIKTORVAV99:
Something extremely weird is going on. All the text is gone from the component tests 😅

VIKTORVAV99:
Will pause this PR a bit since we have a few higher priority things to do.

VIKTORVAV99:
There was a typing conflict between the cypress component tests and our unit tests so I removed the jest types (vitest support chai too). However this caused some type errors in the unit tests instead which I then shifted to chai style assertions to resolve.

I hope this make sense and I would be happy to split out the test parts in another PR if need be.

web/src/components/TimeSlider.test.ts
```diff
@@ -2,8 +2,8 @@ import { COLORS, getThumbIcon, getTrackBackground } from './TimeSlider';
 
 describe('getTrackBackground', () => {
   it('returns the day color when no sets are provided', () => {
-    expect(getTrackBackground(false, 10)).toEqual(COLORS.light.day);
-    expect(getTrackBackground(true, 10)).toEqual(COLORS.dark.day);
+    expect(getTrackBackground(false, 10)).to.eq(COLORS.light.day);
+    expect(getTrackBackground(true, 10)).to.eq(COLORS.dark.day);
```
VIKTORVAV99, line 6: An alternative to changing all of these would be to import all the functions from vitest directly in all the tests, then both syntaxes would be valid as far as `tsc` is concerned.

web/src/utils/styling.test.ts
```diff
@@ -9,10 +9,10 @@ describe('useMediaQuery', () => {
   it('renders', () => {
     window.resizeTo(BELOW_MIN_WIDTH, 0);
     const { result } = renderHook(() => useMediaQuery(`(min-width: ${MIN_WITDH}px)`));
-    expect(result.current).toBeFalsy();
+    expect(result.current).to.be.false;
 
     act(() => window.resizeTo(MIN_WITDH, 0));
 
-    expect(result.current).toBeTruthy();
+    expect(result.current).to.be.true;
```
tonypls, line 16: Nice, I kinda like the new syntax

web/src/features/charts/tooltipCalculations.test.ts
```diff
@@ -119,22 +119,20 @@ describe('getProductionTooltipData', () => {
...
-        "isExport": false,
-        "production": 41161,
-        "storage": undefined,
-        "totalElectricity": 84545.75,
-        "totalEmissions": 9590208033016.975,
-        "usage": 41161,
-        "zoneKey": "FR",
-      }
-    `);
+    expect(actual).to.deep.equal({
```
tonypls, line 122: Does this mean we'll lose the ability to update snapshots?

VIKTORVAV99: Oh didn't think about that, vitest also has snapshots so I'll just add it as an import and revert to the snapshots.

madsnedergaard (approved):
LGTM

PR #6405, opened 2024-01-28
Fix ENTSO-E query targets, refetch_frequency, and request data -72 hours

VIKTORVAV99:
## Issue

There seems to be a misalignment to what the parser fetches and what it said it would fetch in the docstrings. The time delta for some also seems wrong as this should only be the lookback time?

## Description
This PR does 3 things:

1. Changes the default lookback hours to -72, this reduces the amount of queries we need to do when refetching larger time periods by 1/3.
2. Changes the requested date times to what is actually said in the docstrings.
3. Changes the refetch_frequency to be the lookback from the target datetime (only the amount of hours we fetch before the target datetime).

### Double check


unitrium (approved):
Looks good to me!

PR #6435, opened 2024-02-01
fix(exchange tooltip): Add optional chaining to prevent crash if there is no data for exchanges

VIKTORVAV99:
## Issue

In the exchange tooltip we used direct property access to get the data, this works great if there is data, however if there is not this causes a hard crash.

## Description

This PR adds optional chaining so things fail gracefully and don't crash if there is no data for an exchange. This matches the behaviour we use in other places to access the same data.

### Double check

### DIFF
--- a/web/src/features/charts/tooltipCalculations.ts
+++ b/web/src/features/charts/tooltipCalculations.ts
@@ -90,10 +90,9 @@ export function getExchangeTooltipData(
 ) {
   const { zoneKey, exchangeCo2Intensities, exchangeCapacities } = zoneDetail;
 
-  const co2Intensity = exchangeCo2Intensities[exchangeKey];
-
-  const exchangeCapacityRange = (exchangeCapacities || {})[exchangeKey];
-  const exchange = (zoneDetail.exchange || {})[exchangeKey];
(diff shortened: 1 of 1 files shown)


web/src/features/charts/tooltipCalculations.ts
```diff
@@ -90,10 +90,9 @@ export function getExchangeTooltipData(
 ) {
   const { zoneKey, exchangeCo2Intensities, exchangeCapacities } = zoneDetail;
 
-  const co2Intensity = exchangeCo2Intensities[exchangeKey];
-
-  const exchangeCapacityRange = (exchangeCapacities || {})[exchangeKey];
-  const exchange = (zoneDetail.exchange || {})[exchangeKey];
+  const co2Intensity = exchangeCo2Intensities?.[exchangeKey];
+  const exchangeCapacityRange = exchangeCapacities?.[exchangeKey];
+  const exchange = zoneDetail?.exchange?.[exchangeKey];
```
VIKTORVAV99, line 95: It's possible the first optional chain is not needed on this row but I thought it safer to include it than leave it out.

tonypls (approved):
LGTM nice!!!

PR #6438, opened 2024-02-02
chore(ruff): Upgrade to ruff 0.2.0

VIKTORVAV99:
## Description

Ruff has released a new "major" version, we should adopt it to add more auto fixes to rules as well as to benefit from upstream bug fixes.

### Preview

There was no changes with our current ruleset.

### Double check


VIKTORVAV99:
There has already been a new release of ruff, closing until we are ready to update on both contrib and the backend.

PR #6440, opened 2024-02-02
fix(CA-NS) - data gets overwritten by 0

mathilde-daugy:
## Issue

Data in the DB would get overwritten by 0 MW a day later because the first data point in the response was 0MW
image

Also it seems like the data is delayed by over a week but we haven't been able to confirm if this is a bug or not (@lav-julien)

## Description

Load data points at 0MW are filtered out as total demand should never be 0 MW
- added parser validation

### Double check

### DIFF
--- a/parsers/CA_NS.py
+++ b/parsers/CA_NS.py
@@ -14,7 +14,9 @@ from electricitymap.contrib.lib.models.event_lists import (
 from electricitymap.contrib.lib.models.events import ProductionMix
 from electricitymap.contrib.lib.types import ZoneKey
 
-SOURCE ="nspower.ca"
+SOURCE = "nspower.ca"
+
+
 def _get_ns_info(requests_obj, logger: Logger):
     zone_key = ZoneKey("CA-NS")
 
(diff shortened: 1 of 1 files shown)


lav-julien:
For the week delay I think it's just because the datasource is down now, before (2 weeks ago) we were receiving data close to real-time!

parsers/CA_NS.py
```diff
@@ -97,7 +113,7 @@ def _get_ns_info(requests_obj, logger: Logger):
             )
 
         electricity_mix = {
-            gen_type: percent_value * load
+            gen_type: round(percent_value * load, 3)
```
VIKTORVAV99, line 116: Not sure round is needed, the production_mix.add_value() will round any number to 6 decimals internally.

VIKTORVAV99 (approved):
Seems to work as expected, nice work!

We already have a PR open for CA-NS though but I think this is pretty simple and they can work together.

Would you mind adding a test or snapshot test so we don't regress in the future.

mathilde-daugy:
@VIKTORVAV99 will merge this now and we should go ahead with the other PR that does good refactoring

VIKTORVAV99:
> @VIKTORVAV99 will merge this now and we should go ahead with the other PR that does good refactoring

Sounds good, I'll probably create a separate PR to add a snapshot test to make sure the other PR don't break stuff.

PR #6443, opened 2024-02-02
chore(react-i18next): change usage of custom useTranslation to react-i18next

VIKTORVAV99:
## Description
Updates out translation code to make use of i18n interpolation instead of sprintf-js. This should help with performance, developer experience and bundle size.

### Todo:

- [x] Remove usage of custom useTranslation hook
- [x] Fix type errors
- [x] Fix tests
- [x] Update translation files
- [x] Remove old custom code

### Double check


web/cypress.config.ts
```diff
@@ -15,5 +15,17 @@ export default defineConfig({
...
+        if (browser.family === 'chromium' && browser.name !== 'electron') {
+          launchOptions.preferences.default.intl = {
+            accept_languages: 'en-US,en',
+            selected_languages: 'en-US,en',
+          };
+
+          return launchOptions;
+        }
+      });
+    },
```
VIKTORVAV99, line 29: This is just so the local setup uses the same as the CI everywhere. Was quite annoying when it picked Swedish instead of English all the time.

web/src/features/map-controls/MapOptionSelector.tsx
```diff
@@ -15,7 +14,6 @@ export default function MapOptionSelector({
   children,
   isMobile,
 }: MapOptionSelectorProps) {
-  const { __ } = useTranslation();
```
VIKTORVAV99, line 18: since we consider any variable that begins with `_` to be a placeholder/allowed unused we didn't catch these unused hooks before. 

There are a few more that I removed as well.

web/src/App.tsx
```diff
@@ -79,10 +78,10 @@ export default function App(): ReactElement {
             <Sentry.ErrorBoundary fallback={ErrorComponent} showDialog>
               {isSuccess && isNewVersionAvailable && (
                 <Toast
-                  title={__('misc.newversion')}
+                  title={i18n.t('misc.newversion')}
```
madsnedergaard, line 81: Why not use the `useTranslation` hook here?

web/src/features/charts/CarbonChart.tsx
```diff
@@ -23,7 +25,7 @@ function CarbonChart({ datetimes, timeAverage }: CarbonChartProps) {
 
   const hasEnoughDataToDisplay = datetimes?.length > 2;
 
-  const badgeText = getBadgeText(chartData);
+  const badgeText = getBadgeText(chartData, t);
```
madsnedergaard, line 28: Nice catch :)

web/src/features/charts/graphUtils.ts
```diff
@@ -160,22 +160,20 @@ export function getElectricityProductionValue({
   return generationTypeStorage === 0 ? 0 : -generationTypeStorage;
 }
 
-export function getBadgeText(chartData: AreaGraphElement[]) {
-  const { __ } = useTranslation();
-
+export function getBadgeText(chartData: AreaGraphElement[], t: TFunction) {
   const allEstimated = chartData.every(
```
madsnedergaard, line 164: Nice with this change so we don't need the hook in here :)

web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
```diff
@@ -42,10 +42,10 @@ export default function AreaGraphToolTipHeader(props: AreaGraphToolTipHeaderProp
             <Badge
               pillText={
                 estimatedPercentage
-                  ? i18n.t('estimation-card.aggregated_estimated.pill', {
+                  ? t('estimation-card.aggregated_estimated.pill', {
                       percentage: estimatedPercentage,
                     })
-                  : __('estimation-badge.fully-estimated')
+                  : t('estimation-badge.fully-estimated')
               }
```
madsnedergaard, line 49: It's weird how we were mixing `i18n.t` and `__`, so it's great to get it all aligned 😅

web/src/features/charts/tooltips/BreakdownChartTooltip.tsx
```diff
@@ -134,31 +135,31 @@ export function BreakdownChartTooltipContent({
...
+          <CountryFlag className="shadow-3xl" zoneId={selectedLayerKey} />
+        ),
+      }) // Eg: "7 % of electricity in Denmark is imported from Germany"
+    : t(originTranslateKey, {
+        percentageUsage: percentageUsage.toString(),
+        zoneName: getZoneName(zoneKey),
+        selectedLayerKey: t(selectedLayerKey),
+        zoneFlag: renderToString(<CountryFlag className="shadow-3xl" zoneId={zoneKey} />),
+      }); // Eg: "20 % of electricity in Denmark comes from biomass"
   const title = isExchange
```
madsnedergaard, line 160: Omg yes, I have many times struggled with trying to reason about what the hell this exact translation would output - the named variables makes it so much easier 😍

madsnedergaard (approved):
My fingers hurt from scrolling, but it all looks great and on the preview app everything looks as expected! Thanks for taking the time to do this one, it's cleaning up a lot of tech debt 🙏 💚

web/src/App.tsx
VIKTORVAV99: Good question 😅

I'll change it...

web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
VIKTORVAV99: That was me like last week, got so annoyed by sprintf I ended up removing it everywhere 😅

web/src/features/charts/tooltips/BreakdownChartTooltip.tsx
VIKTORVAV99: And now you know why this PR exists 😅

web/src/App.tsx
VIKTORVAV99: I remember why I did it like this now...

Otherwise the i18n import becomes unused and well vite removes the whole thing 😅

It should still be the same instance though so should not impact any functionality.

VIKTORVAV99:
@madsnedergaard should we wait for a review from Tony or merge this now? I would like to merge this as soon as possible since it changes 88 files and merge conflicts are only a matter of time...

web/src/App.tsx
madsnedergaard: Ah okay, fair :)

PR #6447, opened 2024-02-04
feat(code splitting): do more granular code splitting to improve first load

VIKTORVAV99:
## Description

This PR mainly does 2 things:

- Splits up the code using react.lazy() to create more modular modules.
- Adds new suspense boundaries so these modules (or set of modules) can be shown as soon as they are ready.

### Background
The reason for doing this is simply to both improve the performance of the app as well as the perceived performance of the app.

For example by putting the header in it's own suspense boundary we can load it way before the other parts of the app, increasing the perceived performance and first load experience.

While code splitting the left panel into separate modules allows it to be shown faster as it only needs to load either the ranking panel or country panel but never both.

### Things of note
The header, loading layer, onboarding, and main app now load independently. This makes the app feel way faster but it has a gotcha where the onboarding can load over the loading layer before the map is ready. I don't think this is necessary a bad thing, especially since it should only happen once, but something to keep in mind and look a bit extra at.

### Double check


web/src/App.tsx
```diff
@@ -22,6 +18,10 @@ import trackEvent from 'utils/analytics';
 
 const MapWrapper = lazy(async () => import('features/map/MapWrapper'));
 const LeftPanel = lazy(async () => import('features/panels/LeftPanel'));
+const LegendContainer = lazy(async () => import('components/legend/LegendContainer'));
+const FAQModal = lazy(() => import('features/modals/FAQModal'));
+const InfoModal = lazy(() => import('features/modals/InfoModal'));
+const SettingsModal = lazy(() => import('features/modals/SettingsModal'));
```
q--, line 24: Only one of these is `async`. Intentional?

VIKTORVAV99: Good catch, it is not!

VIKTORVAV99:
> I think this is generally really great! I think the "onboarding showing before map is loaded" is actually fine, then you can start reading while waiting :)
> 
> I do however have a slight problem with the panels - if you are on a slow connection, it is kinda weird that you'll first see a map and then when clicking a country you only see a grey screen. I suggest we either load the zone details panel earlier or add a loading indicator to let people know something is going on. Here's a video showing it:
> 
>  Kapture.2024-02-07.at.15.51.55.mp4

Oh that's pretty bad, never tested it in mobile view. (Probably should have...)

I'll see what I can do but a loading indicator is probably best until we have fallback mock containers for graphs.

web/src/App.tsx
```diff
@@ -83,16 +85,26 @@ export default function App(): ReactElement {
...
+                <OnboardingModal />
+              </Suspense>
+              <Suspense>
+                <FAQModal />
+                <InfoModal />
+                <SettingsModal />
+              </Suspense>
+              <Suspense>
+                <LeftPanel />
+                <MapWrapper />
```
q--, line 101: On mobile, if people arrive via a direct link to a zone/country, only the "left panel" is shown. It might make sense not to wait on the map in that case.

VIKTORVAV99: I'll take a look at that tomorrow to see how it affects non-mobile users. Otherwise that makes a lot of sense.

madsnedergaard (approved):
Great! 👏

PR #6448, opened 2024-02-04
feat(early hints): Add early hint headers for Cloudflare

VIKTORVAV99:
## Description

Apparently this is already active:
```
➜  web git:(vik/early_hints) curl -X HEAD -I https://app.electricitymaps.com
HTTP/2 103 
link: <https://app-backend.electricitymap.org/>; rel=preconnect, <https://o192958.ingest.sentry.io>; rel=preconnect, <https://plausible.io>; rel=preconnect

HTTP/2 200 
date: Sun, 04 Feb 2024 13:27:26 GMT
content-type: text/html; charset=utf-8
access-control-allow-origin: *
cache-control: public, max-age=0, must-revalidate
link: <https://app-backend.electricitymap.org/>; rel="preconnect", <https://o192958.ingest.sentry.io>; rel="preconnect", <https://plausible.io>; rel="preconnect"
referrer-policy: strict-origin-when-cross-origin
x-content-type-options: nosniff
report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=3ucCajQHrz%2B7UcIvBxDTlsJk%2B3ugyN%2FXheWRFSBHlIteUnQU8hH4Ufc9%2B9zZSB2NOTOvQkmxXu%2FPkFPASu9%2Fb86J%2BsjsnVEZ9Ar2b424YbDs6zprr%2BR9jpbqW8mSFRXGrNLngykKiI5pjA%3D%3D"}],"group":"cf-nel","max_age":604800}
nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
vary: Accept-Encoding
cf-cache-status: DYNAMIC
server: cloudflare
cf-ray: 850340455fee82b9-ARN
alt-svc: h3=":443"; ma=86400
```
 And apparently the preload links is more or less a hit or miss due to the fact that Cloudflare skips the 103 responses if they can deliver it directly from their cache fast enough.

> Early Hints may be emitted less frequently on requests where the content is cacheable. Cloudflare CDN is more likely to retrieve a response header before the asynchronous Early Hints lookup finishes if the response has been cached. Cloudflare will not send a 103 response if the main response header is already available.

So might not be any benefits there but I'll leave it up to you guys if we should merge or close this PR.


madsnedergaard:
Can you elaborate a bit more on this - what's the cons/pros of merging it? :)

VIKTORVAV99:
> Can you elaborate a bit more on this - what's the cons/pros of merging it? :)

If Cloudflare followed the spec to the dot (which they don't appear to do) it would tell browsers to download the preloaded assets even before the HTML is downloaded and parsed reducing the wall-time until all required assets are loaded.

There is no downsides to this except that we have to update the headers if we update the preloaded assets (only the fonts so far).

So it potentially has the possibility to speed up first load by telling the browser to download the fonts sooner with no downsides.

If you want do dig into the nitty gritty details this is a good starting point: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/103

madsnedergaard (approved):
Thanks for the detailed response, let's do it - I can enable it on the CloudFlare side :)

madsnedergaard:
Seems like it's working with the fonts as well now :)
Screenshot 2024-02-15 at 21 54 42

PR #6451, opened 2024-02-05
update(ENTSOe) - query_ENTSOE update to enable use for capacity parser

mathilde-daugy:
## Issue

image

## Description

set `span` as optional so that the `query_ENTSOE` function can keep working for the capacity update as the params already define the `periodStart` and `periodEnd` (annual frequency rather than hourly)

@VIKTORVAV99 I saw that this was updated last week, idk what the best way to go about this. Let me know if you have a better suggestion :)

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -475,7 +475,7 @@ def closest_in_time_key(x, target_datetime: datetime | None, datetime_key="datet
 def query_ENTSOE(
     session: Session,
     params: dict[str, str],
-    span: tuple | None =None,
+    span: tuple,
     target_datetime: datetime | None = None,
     function_name: str = "",
 ) -> str:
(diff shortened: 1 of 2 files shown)


parsers/ENTSOE.py
```diff
@@ -493,13 +493,13 @@ def query_ENTSOE(
...
             message="target_datetime has to be a datetime in query_entsoe",
         )
-
-    params["periodStart"] = (target_datetime + timedelta(hours=span[0])).strftime(
-        "%Y%m%d%H00"  # YYYYMMDDHH00
-    )
-    params["periodEnd"] = (target_datetime + timedelta(hours=span[1])).strftime(
-        "%Y%m%d%H00"  # YYYYMMDDHH00
-    )
+    if span is not None:
```
VIKTORVAV99, line 496: This is fine but then we should also make sure the `periodStart` and `period end` parameters are already in the parameters object and throw an exception if they are not.

Otherwise we might accidentally query entso-e without valid parameters in the future.

Also didn't this already overwrite the parameters that was sent from the query function?

Btw have you tried requesting all the capacity data from 2015 and forwards? Might be possible to get all the capacity data in one go for each region which would make backfills and retroactive updates trivial.

mathilde-daugy:
i reverted the changes and only modified the capacity parser
in the future, please add me to PRs with modifications to major parsers so that i know what is happening. :)

VIKTORVAV99:
> i reverted the changes and only modified the capacity parser in the future, please add me to PRs with modifications to major parsers so that i know what is happening. :)

Will do, I didn't realise the capacity parser used the same fetch function.
If I have the time I'll try and create a better test coverage for them the next slack week.

VIKTORVAV99 (approved):

torokati44:
This more or less reverted https://github.com/electricitymaps/electricitymaps-contrib/pull/6446. :/

PR #6452, opened 2024-02-05
Add nuclear parser to JP-SK production parser 

wobniarin:
## Issue
the parser for JP-SK only accounts for unknown and solar, and nuclear has to be fetched from a different website (nuclear power plant operator) and reading the information from a GIF. 

## Description
This PR creates a specific parser for JP-SK where nuclear power can be fetched and added to the data point for a given timestamp.

This is the [website](https://www.yonden.co.jp/energy/atom/ikata/ikt722.html) we want to extract nuclear power production from 

image

We also include tests to validate that we can parse the nuclear power production

### Preview

Result after implementing the parser 
image

### Double check

### Expected reviews 

@VIKTORVAV99  and @unitrium I added you two since I think you can provide valuable feedback on this parser and the tests I have implemented.


VIKTORVAV99 (commented):
Just some small suggestions but otherwise it looks good!

(We don't particularly like pytz either 😅)

VIKTORVAV99:
Just some notes that are not part of the review. I know this PR is not using the ParserEvent classes which I think is okay since the core JP parser don't either. When this is updated this parser should be updated in the same PR.

wobniarin:
> Just some notes that are not part of the review. I know this PR is not using the ParserEvent classes which I think is okay since the core JP parser don't either. When that is updated this parser should be updated in the same PR.

Makes sense, thanks for the heads-up! :)

parsers/JP_SK.py
```diff
@@ -0,0 +1,102 @@
...
+    filtered_img_tags = [tag for tag in images_links if "ikt721-1" in tag["src"]]
+    if len(filtered_img_tags) == 0:
+        raise Exception("No image found")
+    img_url = IMAGE_CORE_URL + filtered_img_tags[0]["src"]
+    return img_url
+
+
+def get_nuclear_power_value_and_timestamp_from_image_url(
+    img_url: str, session: Session
+) -> (float, datetime):
```
VIKTORVAV99, line 80:
```suggestion
def get_nuclear_power_value_and_timestamp_from_image_url(
    img_url: str, session: Session
) -> tuple[float, datetime]:
```

parsers/JP_SK.py
```diff
@@ -0,0 +1,102 @@
...
+    """This method reads the image from the image url and extracts nuclear power and timestamp from it."""
+    response_image = session.get(img_url)
+    image = Image.open(BytesIO(response_image.content))
+    width, height = image.size
+    img = image.crop((0, 0, width, height))
+    text = image_to_string(img)
```
VIKTORVAV99, line 86:
```suggestion
    text = str(image_to_string(img))
```

VIKTORVAV99 (approved):
Just two tiny comments to fix two type errors, they don't change any functionality though so approving.

Nice work!

PR #6455, opened 2024-02-06
Misc improvements to capacity update 

mathilde-daugy:
## Description

renamed poetry command for consistency + update prettier command in `capacity_update` function

### Double check


VIKTORVAV99 (approved):
Do we need to change the readme as well to account for the command change?

mathilde-daugy:
good catch, will do :)

PR #6462, opened 2024-02-07
update(FR parser) - review historical + add battery storage

mathilde-daugy:
## Issue
The source has two datasets: 1 consolidated and 1 real-time. Originally we thought that the real-time data was migrated to the consolidated set after 9 months but it turns out that it isn't the case and only the real-time dat gets populated
image

We have one month of missing data in the DB (Jun-2022) because of this so this PR fixes this issue

The parser did not use the parser validation scheme and battery storage was not captured

## Description

Update includes:
- consistency for parsing past dates
- parser validation
- battery storage

### Preview

image

### Double check


parsers/FR.py
```diff
@@ -27,31 +30,28 @@
...
+TZ = ZoneInfo("Europe/Paris")
+SOURCE = "opendata.reseaux-energies.fr"
 
 
 def get_dataset_from_datetime(target_datetime: datetime) -> str:
     """Returns the dataset to query based on the target_datetime. The real-time API returns no values for target datetimes older than 9 months and we need to query the consolidated dataset."""
-    if target_datetime < arrow.now(tz="Europe/Paris").shift(months=-9):
-        # API called is Données éCO2mix régionales consolidées et définitives for datetimes that are more than 9 months in the past
+    if target_datetime < datetime(2022, 5, 31, tzinfo=TZ):
+        # API called is Données éCO2mix régionales consolidées et définitives for datetimes before mai 2022
```
unitrium, line 54:
```suggestion
        # API called is Données éCO2mix régionales consolidées et définitives for datetimes before May 2022
```

parsers/FR.py
```diff
@@ -88,7 +88,11 @@ def get_data(
     params["apikey"] = get_token("RESEAUX_ENERGIES_TOKEN")
     # make request and create dataframe with response
     response = r.get(API_ENDPOINT, params=params)
-    data = json.loads(response.content)
+    if response.status_code != 200:
```
unitrium, line 91:
```suggestion
    if not response.ok:
```
Shortcut

parsers/FR.py
```diff
@@ -139,74 +142,33 @@ def fetch_production(
...
-            if -50 < row[1][key] < 0:
-                # set small negative values to 0
-                logger.warning(f"Setting small value of {key} ({value}) to 0.")
-                production[value] = 0
+    df_production_reindexed = df_production_reindexed.rename(columns=MAP_GENERATION)
+    df_production_reindexed = df_production_reindexed.groupby(
+        df_production_reindexed.columns, axis=1
+    ).sum()
+
+    datapoints = ProductionBreakdownList(logger)
```
unitrium, line 150:
```suggestion
    production_mixes = ProductionBreakdownList(logger)
```
For readability I recommend using clear variables

unitrium (approved):
Looks good to me!

VIKTORVAV99 (changes requested):
Can we update the snapshots to include the sourceType as well?

parsers/FR.py
```diff
@@ -216,24 +178,14 @@ def fetch_consumption(
...
-                .datetime,
-                "consumption": row.consommation,
-                "source": "opendata.reseaux-energies.fr",
-            }
+        consumption_list.append(
+            zoneKey=zone_key,
+            consumption=row.consommation,
+            datetime=row.Index.to_pydatetime(),
+            source=SOURCE,
+            sourceType=EventSourceType.measured,
```
VIKTORVAV99, line 188: Technically this don't need to be included as it's the default. But it don't hurt either.

VIKTORVAV99 (approved):

PR #6464, opened 2024-02-08
Add estimation feedback card

silkeholmebonnen:
## Issue

Getting user feedback on the estimation card [ELE-3734](https://linear.app/electricitymaps/issue/ELE-3734/feedback-card-for-estimation-card)

## Description

This PR adds a small feedback section below the estimation card. This will be shown the first time a user has expanded and afterwards collapsed the estimation card. Local storage is used to ensure that the user only sees the feedback card once (if they clear their localstorage, they will see it again). The feedback will be sent to [Sapier](https://zapier.com/app/dashboard). 
### Preview

Dark:
image
image
image
Light:
image

### Double check


web/public/locales/en.json
```diff
@@ -90,6 +90,24 @@
...
+  "estimation-feedback": {
+    "title": {
+      "state-1": "Help us improve!",
+      "state-3": "Thank you!"
+    },
+    "subtitle": {
+      "state-1": "Please rate the following statement.",
+      "state-3": "Your response helps us understand and improve the quality of our communication and methodology."
+    },
+    "rate-question": "The description above was easy to understand:",
```
tonypls, line 102: Users may think, what description above? If it's only going to be feedback of the estimation card we could say "The description of the data estimation was easy to understand"

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+          state={state}
+          setState={setState}
+          estimationMethod={estimationMethod}
+        />
+      </div>
+    </div>
+  );
+}
+
+function InputField({
```
tonypls, line 64: It would be useful if the input could grow with the users input, you could use a [textarea element](https://react.dev/reference/react-dom/components/textarea) here.

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+    >
+      {pillContent.map((content) => (
+        <ToggleGroupPrimitive.Item
+          data-test-id={`feedback-pill-${content}`}
+          key={content}
+          value={String(content)}
+          aria-label={String(content)}
+          onClick={() => handlePillClick(String(content))}
+          className={`
+          mr-2 inline-flex h-9 w-[60px] select-none items-center justify-center rounded-full border border-neutral-200 text-black  dark:border-gray-700 dark:text-white
```
tonypls, line 228: As the screen becomes larger these move quite far apart from each other, it could be nicer to fix the margin and allow the pill to grow to fill the new space. 
image

tonypls (commented):
Looks great!

Adding the new icons will add a tiny bit of size to the bundle, we already send an icon (electricitymaps-icon.svg) to the client and use the tailwind property `dark:invert` to change it to white for darkmode. Could be nice to remain as lightweight as possible and use this one again. 

I've added some general and user style feedback. Ping me when it's out of draft and I'll come back to give a more in depth code review.

web/src/features/panels/zone/FeedbackCard.tsx
VIKTORVAV99: I would also try and center the buttons, right now they are a little to the left (just a nitpick though).

web/public/locales/en.json
silkeholmebonnen: Good point. I talked to Alex and we agree so I've changed it to be "The description of the data estimation was easy to understand"

web/src/features/panels/zone/FeedbackCard.tsx
silkeholmebonnen: I changed it in the latest commit. Does it look like you envisioned?

web/src/features/panels/zone/FeedbackCard.tsx
silkeholmebonnen: @tonypls I've changed it so the pill since grows instead. @VIKTORVAV99 I also removed some padding so it should be centered now

VIKTORVAV99:
Looks good!
Just some small stuff:
the text area don't auto expand but I'm not sure if there is a easy fix, I remember having this problem years ago with no solution.

There is a small selected area in the bottom right corner that looks a little off.

image


Both of these are just small nitpicks tho and are not considered blockers by me.

silkeholmebonnen:
> Looks good! Just some small stuff: the text area don't auto expand but I'm not sure if there is a easy fix, I remember having this problem years ago with no (easy) solution.
> 
> There is a small selected area in the bottom right corner that looks a little off.
> 
> image
> Both of these are just small nitpicks tho and are not considered blockers by me.
> 
> Would also like to bump the text size a little here: image It's tiny and hard to read and I don't think @Alportan would mind :)

I've disabled the resize option so the thing in the bottom right corner should be gone, and the area auto expands now. 

Good catch with the text, because @Alportan  actually already updated the Figma to have bigger text in that section, I just missed it.

VIKTORVAV99 (approved):
Looks great now!

Nice work!

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+import { HiOutlineX } from 'react-icons/hi';
+import { useTranslation } from 'translation/translation';
+
+export default function FeedbackCard({
+  estimationMethod,
+}: {
+  estimationMethod?: string;
+}) {
+  const [isClosed, setIsClosed] = useState(false);
+  const [state, setState] = useState('1');
```
madsnedergaard, line 13: I think we should use some proper names here to make it easier to understand what each state means :) 

You could use an enum like we do for `TimeAverages` - this means we're using strings, but in a way where you can't misspell it and it's easier to use (your editor can show you what options are available, etc.)

```ts
enum FeedbackState {
  OPEN = 'open',
  SUCCESS = 'success',
  // etc.
}

// Usage in component
const [state, setState] = useState(FeedbackState.OPEN);

// Usage for type
function FeedbackActions({state}: { state: FeedbackState }) {}
```

web/src/features/panels/zone/EstimationCard.cy.tsx
```diff
@@ -0,0 +1,22 @@
+import EstimationCard from './EstimationCard';
```
madsnedergaard, line 1: Kudos for adding the Cypress tests here 🥳

web/public/locales/en.json
```diff
@@ -90,6 +90,23 @@
       "total-electricity-consumption": "Total electricity consumption by source"
     }
   },
+  "estimation-feedback": {
+    "title": {
+      "state-1": "Help us improve!",
```
madsnedergaard, line 95: See my comment further down about naming the states, which I think should also be applied here :)

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -15,6 +19,21 @@ export default function EstimationCard({
...
+  useEffect(() => {
+    setcurrentlyShowingFeedbackCard(
+      feedbackEnabled &&
+        showEstimationFeedbackCard(
+          collapsedNumber,
+          currentlyShowingFeedbackCard,
+          localStorage.getItem('feedbackCardStatus')
+        )
+    );
+  }, [feedbackEnabled, collapsedNumber]);
```
madsnedergaard, line 35: I added the feature-flag wrapping here, it was super easy thanks to the way you made it 🙌 

Here is an example :
https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/5f26d852-d075-4b25-a489-bccd278836a1

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+    </ToggleGroupPrimitive.Root>
+  );
+}
+
+function getQuestionTranslation(field: string, state?: string) {
+  const { __ } = useTranslation();
+  if (state !== undefined) {
+    if (state == '2') {
+      return __(`estimation-feedback.${field}.state-1`);
+    }
```
madsnedergaard, line 266: What is happening here? 😅

web/src/features/panels/zone/FeedbackCard.tsx
tonypls: Looks great!

web/src/features/panels/zone/util.ts
```diff
@@ -77,3 +77,20 @@ export function getDisclaimer(zoneId: string) {
   const zoneConfig = config.zones[zoneId];
   return zoneConfig?.disclaimer;
 }
+
+export function showEstimationFeedbackCard(
```
tonypls, line 81: It would be nice to use the same pattern we use for the onboarding modal here and use an [atom with storage](https://jotai.org/docs/utilities/storage) it's stored as onboardingSeen. It could make sense to call it estimationFeedbackSeen and default it to false. The atoms are nice because you can read and set them easily from any component

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -23,7 +42,17 @@ export default function EstimationCard({
...
     }
     case 'estimated': {
-      return <EstimatedCard estimationMethod={estimationMethod} />;
+      return (
+        <div>
+          <EstimatedCard
+            estimationMethod={estimationMethod}
+            setCollapsedNumber={setCollapsedNumber}
+          />
+          {currentlyShowingFeedbackCard && (
```
tonypls, line 51: Nitpick but it's quite nice to use the pattern isXVisibile for visibilty naming so isFeedbackCardVisibile in this case

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -63,11 +93,16 @@ function BaseCard({
...
   pillType?: string;
   textColorTitle: string;
+  setCollapsedNumber?: Dispatch<SetStateAction<number>>;
 }) {
   const [isCollapsed, setIsCollapsed] = useState(
     estimationMethod == 'outage' ? false : true
   );
+
   const handleToggleCollapse = () => {
+    if (setCollapsedNumber) {
```
tonypls, line 103: As mentioned in the utils, we should be able to simplify this by using an atom

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+    </div>
+  );
+}
+
+function PillContent({
+  pillContent,
+  handlePillClick,
+  CurrentPillNumber,
+}: {
+  pillContent: number[];
```
tonypls, line 221: If this is a string array we won't have to cast it to string below, but if possible it might be nicer to have them all as numbers and not cast them to strings if that works.

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+  handlePillClick: (identifier: string) => void;
+  CurrentPillNumber: string;
```
tonypls, line 223: camelCase here

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+}) {
+  return (
+    <ToggleGroupPrimitive.Root
+      className={'flex-start mb-2 flex flex-row items-center justify-between gap-3'}
+      type="multiple"
+      aria-label="Font settings"
```
tonypls, line 229: Not sure what Font settings is describing here

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+  CurrentPillNumber,
+}: {
+  pillContent: number[];
+  handlePillClick: (identifier: string) => void;
+  CurrentPillNumber: string;
+}) {
+  return (
+    <ToggleGroupPrimitive.Root
+      className={'flex-start mb-2 flex flex-row items-center justify-between gap-3'}
+      type="multiple"
```
tonypls, line 228:
```suggestion
      type="single"
```

web/src/features/panels/zone/FeedbackCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+            <SubmitButton handleSave={handleSave} />
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+function ActionPills({
+  setState,
```
tonypls, line 177: This could do with a more descriptive name, what state is being set?

tonypls (commented):
Looking good! 

A few code style comments and suggested improvements. I hope it's not too nitpicky, but I saw this as a nice opportunity to share some knowledge.

web/src/features/panels/zone/FeedbackCard.tsx
silkeholmebonnen: Good idea. I've made the changes to accommodate this in commit [change state to feedbackState and use enum](https://github.com/electricitymaps/electricitymaps-contrib/pull/6464/commits/ab89aaea7c667f9412162100bd811d9a96da05de)

web/src/features/panels/zone/EstimationCard.tsx
silkeholmebonnen: Cool! I didn't know that &&ff after remote=true would show that. What makes the &&ff work like that?

web/src/components/Pill.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const Element = isButton ? 'button' : 'div';
+
+  const handleClick = () => {
+    if (onClick) {
+      onClick(identifier);
+    }
+  };
+
+  return (
+    <Element
```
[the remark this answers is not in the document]
silkeholmebonnen: True. Changes made in commit [pill changes](https://github.com/electricitymaps/electricitymaps-contrib/pull/6464/commits/bfb8c5b2676bb9cfb866e212700add3d0ca212e8)

web/src/components/Pill.tsx
```diff
@@ -0,0 +1,35 @@
...
+  classes: string;
+  text: string;
+  isButton?: boolean;
+  isDisabled?: boolean;
+  onClick?: (identifier: string) => void;
+  identifier?: string;
+}): JSX.Element {
+  const Element = isButton ? 'button' : 'div';
+
+  const handleClick = () => {
```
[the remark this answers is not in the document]
silkeholmebonnen: True! Changed in commit [remove unnecessary code in Pill](https://github.com/electricitymaps/electricitymaps-contrib/pull/6464/commits/0927c6254596f7c52e112c85e97d5738f70ee500)

web/src/features/panels/zone/util.ts
silkeholmebonnen: Good point! I've changed it in the commit [change to use atoms](https://github.com/electricitymaps/electricitymaps-contrib/pull/6464/commits/fee2b04bb2a1c9970e4ae0ff40445c9296bb3724)

web/src/features/panels/zone/FeedbackCard.tsx
silkeholmebonnen: Do you think `setFeedbackState` is descriptive enough?

web/src/features/panels/zone/FeedbackCard.tsx
silkeholmebonnen: It wasn't possible to use numbers all the way as the value of the `ToggleGroupPrimitive.Item` expected a string, so I changed it to a string array

silkeholmebonnen:
Thank you for taking the time to make a thorough review @tonypls @madsnedergaard, I appreciate it!

web/src/features/panels/zone/EstimationCard.tsx
madsnedergaard: We have this component that only shows when the query parameter is used :)

Screenshot 2024-02-27 at 10 23 05

web/src/features/panels/zone/FeedbackCard.tsx
madsnedergaard: It makes good sense now, thanks for the change :)

web/src/features/panels/zone/EstimationCard.tsx
VIKTORVAV99: @madsnedergaard have you actually documented this anywhere? Might be good to do for our future selfs.

_I didn't know about this particular feature either..._

madsnedergaard (approved):
All good on my side, thanks for bearing with our pretty extensive code reviews 😄

tonypls (approved):
LGTM!! 🎉

PR #6467, opened 2024-02-09
Revert-revert - update get_capacity_data function again

mathilde-daugy:
### Double check


unitrium:
Looks alright to me, we should test it in the estimation pipeline locally first though.

unitrium (approved):

PR #6442, opened 2024-02-02
feat(productionBreakdownList): Add a function to update matching datetime with new values

VIKTORVAV99:
## Issue

ENTSO-E publishes generation forecasts for wind and solar that are updated at 3 different times, day-ahead, intraday and current. Intraday and current are more accurate than than day-ahead.

## Description

This PR is my hackday project.

It adds a way to create an updated production breakdown list from two existing lists of values if the zoneKey, source and sourceType matches.

This will allow us to update the ENTSO-E parser to parse not just day-ahead values but get the most updated and accurate values.

### Todo

- [x] Add tests for empty lists.
- [x] Add tests on the event level.
- [x] Test the inclusion of setters.

### Double check

- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

___

## **Type**
enhancement, tests

___

## **Description**
- Introduced a new method `update_production_breakdowns` in `ProductionBreakdownList` to update existing production breakdowns with new values, supporting both production and storage updates.
- Added `update` class method in `Mix`, `ProductionMix`, and `StorageMix` classes to facilitate the updating of mix values.
- Implemented `__setitem__` in `Mix` class to allow setting values using bracket notation.
- Extensive testing includes scenarios with empty lists, different zoneKeys, sources, and sourceTypes, ensuring robustness and reliability of the update functionality.
- Additional tests validate the correct behavior of the new `update` and `__setitem__` methods in mix classes.

___

## **Changes walkthrough**
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>event_lists.py</strong><dd><code>Add Method to Update Production Breakdowns</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

electricitymap/contrib/lib/models/event_lists.py
<li>Added a new method <code>update_production_breakdowns</code> to update existing <br>production breakdowns with new values.<br> <li> Ensured compatibility with empty lists and added logger support.

</details>
    

  </td>
[...]

PR #6473, opened 2024-02-12
feat(ENTSOE.py): Use update function + get all forecasts

VIKTORVAV99:
## Issue

We are not using the most up to date forecasts for ENTSO-E zones.

## Description

This PR takes advantage of the new update functions introduced in #6442 to resolve the above issue + adds a snapshot test to ensure we don't regress things in the future.

### Double check


parsers/ENTSOE.py
```diff
@@ -48,6 +52,13 @@
 DEFAULT_TARGET_HOURS_REALTIME = (-DEFAULT_LOOKBACK_HOURS_REALTIME, 0)
 DEFAULT_TARGET_HOURS_FORECAST = (-24, 48)
 
+
+class WindAndSolarProductionForecastTypes(Enum):
+    DAY_AHEAD = "A01"
+    INTRADAY = "A40"
+    CURRENT = "A18"
```
VIKTORVAV99, line 59: Do anyone know if the order is guaranteed in enums? I couldn't tell from the python documentation (or I missed something).

pierresegonne: The overlord says:

> Yes, the order you define your enum members in Python is guaranteed to be preserved. This means:

> When you iterate over the enum using a for loop or call Enum.values(), the members will come out in the order you declared them (DAY_AHEAD, INTRADAY, CURRENT).
When you compare enum members using operators like == or <, the comparison will be based on their defined order.
This behavior is consistent across different Python versions and implementations.
While the order is preserved by default, it's important to not rely on it for logic unless explicitly intended. If the order has significance, it's recommended to document it clearly and potentially consider refactoring to a namedtuple or a class with explicit ordering if needed.

> Remember, clarity and maintainability are important principles in coding!

pierresegonne: So I guess if you want to be certain about the order, you could create a constant:

```
ORDERED_FORECAST_TYPES = [WindAndSolarProductionForecastTypes.DAY_AHEAD, ...] # Descending priority
```

Which would make things extra clear

parsers/ENTSOE.py
```diff
@@ -59,6 +59,14 @@ class WindAndSolarProductionForecastTypes(Enum):
     CURRENT = "A18"
 
 
+# The order of the forecast types is important for the parser to use the most recent data
+# This ensures that the order is consistent across all runs even if the enum is changed
+ORDERED_FORCAST_TYPES = [
```
pierresegonne, line 64:
```suggestion
ORDERED_FORECAST_TYPES = [
```

parsers/ENTSOE.py
```diff
@@ -1391,7 +1399,7 @@ def fetch_wind_solar_forecasts(
         session = Session()
     raw_forecasts = {}
     domain = ENTSOE_DOMAIN_MAPPINGS[zone_key]
-    for data_type in WindAndSolarProductionForecastTypes:
+    for data_type in ORDERED_FORCAST_TYPES:
```
pierresegonne, line 1402:
```suggestion
    for data_type in ORDERED_FORECAST_TYPES:
```

pierresegonne:
I think it's looking good, let me know when you want a final review :) 

Annnndd HHaaaappppyyyy Biiirrrtthhhddaaaayyyyyy!

VIKTORVAV99:
@pierresegonne now with the other PR merged it should be good to go! 🙌🏻

pierresegonne (approved):

parsers/ENTSOE.py
```diff
@@ -10,10 +10,14 @@
 Day-ahead Price
 Generation Forecast
 Consumption Forecast
+
+Link to the API documentation:
+https://documenter.getpostman.com/view/7009892/2s93JtP3F6
```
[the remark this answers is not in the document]
VIKTORVAV99: Yeah I thought that would come in handy for the future 😅

PR #6476, opened 2024-02-13
chore(CI): Change the CI structure to allow for required checks

VIKTORVAV99:
## Issue

The way we were structuring our CI was interfering with the possibility to mark jobs as required in GitHub.

## Description

This PR moves all of the jobs that should be required into the main CI file so we can actually mark them as required even if they are skipped without blocking PRs.

### Double check


VIKTORVAV99:
The old prettier check is still expected because the name changed, I'll fix this before merging.

madsnedergaard (approved):
It's not pretty, but it works 😄

PR #6479, opened 2024-02-13
add(capacity-parser) - Indian zones

ghost:
## Issue

Indian capacity can be extracted from the [CEA API ](https://cea.nic.in/api/installed_capacity.php)
Renewable capacity still needs to be added manually from [MNRE](https://cdnbbsr.s3waas.gov.in/s3716e1b8c6cd17b771da77391355749f3/uploads/2024/02/202402122140462737.pdf) (it is only available in pdf)

## Description

Add parser for conventional capacity -> CEA
Add function for REN capacity: the pdf can be converted to excel and then the data is extracted from the excel using the `get_renewable_capacity` function

### Double check


VIKTORVAV99 (approved):
LGTM!

PR #6481, opened 2024-02-14
PE parser: add unit tests and remove usage of arrow module

gianantoniopini:
## Issue

#6135 

## Description

This PR adds unit tests for the PE parser and removes usage of the `arrow` module from the PE parser.

### Double check


parsers/test/test_PE.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+        self.assertMatchSnapshot(
+            [
+                {
+                    "datetime": element["datetime"].isoformat(),
+                    "production": element["production"],
+                    "source": element["source"],
+                    "zoneKey": element["zoneKey"],
+                    "sourceType": "measured",
```
VIKTORVAV99, line 41:
```suggestion
                    "sourceType": element["sourceType"].value,
```
This just brings it in line with our other tests and makes it easy to update if we ever change this.

VIKTORVAV99 (commented):
Looks good!

Just one tiny comment to take a look at then I'll merge this.

parsers/test/mocks/PE/response_20240205.json
VIKTORVAV99, line 1: One thing we can do if we want to reduce the size of these are minifying the files, it should have no impact on the tests.

VIKTORVAV99:
Merging this as is right now as we have a follow up PR that converts it to use parser event lists as well.

VIKTORVAV99 (approved):

gianantoniopini:
👍

PR #6484, opened 2024-02-14
format(config): use consistent dash for key-value pair to signify list

tonypls:
## Issue
Our yaml isn't consistent which means it's not loading in correctly to our ModeCapacity class

## Description
This PR formats the zone yaml files capacity list key-value pairs to start with a dash and indent the following list items appropriately

### Double check


VIKTORVAV99:
@tonypls mind taking a look at the stuff ruff flagged?

tonypls:
> @tonypls mind taking a look at the stuff ruff flagged?

I had left my linting script in there! trimmed it now ✂️

VIKTORVAV99 (approved):
Approving but we probably want to fix this in the capacity parser as well so we don't introduce more cases like this.

PR #6485, opened 2024-02-15
feat(web): Show capacity sources in the app

tonypls:
## Issue
We have fancy capacity sources but we aren't sharing them with the world

## Description
This PR enables the front-end to display capacity sources if they're available in the app-backed (PR open [over there,](https://github.com/electricitymaps/electricitymaps/pull/4591) branch preview won't show them till that's been deployed)

### Preview

image

### Double check


web/src/features/charts/tooltipCalculations.ts
```diff
@@ -46,6 +47,10 @@ export function getProductionTooltipData(
 
   const generationTypeCapacity = capacity ? capacity[selectedLayerKey] : undefined;
   const generationTypeProduction = production[generationType];
+  const capacitySource =
+    capacitySources && generationTypeCapacity
+      ? capacitySources[selectedLayerKey]
+      : undefined;
```
VIKTORVAV99, line 53: This can be simplified a bit by using optional chaining since we wouldn't need to check the capacity sources object.

web/src/types.ts
```diff
@@ -97,6 +97,7 @@ export interface ZoneDetail extends ZoneOverview {
   completenessPercentage?: number;
   // Capacity is only available on hourly details
   capacity?: { [key in ElectricityModeType]: number | null };
+  capacitySources?: { [key in ElectricityModeType]: string };
```
VIKTORVAV99, line 100: This could probably also potentially be null in some cases, better include that than exclude it.

VIKTORVAV99 (commented):
Just some small notes, will give a final review after the backend PR.

madsnedergaard (commented):
Code looks good, looking forward to testing it on staging when the backend part is done :)

tonypls:
Backend data and config updates are now deployed so you should be able to test all elements of this on the branch previews 👍

web/src/features/charts/tooltips/BreakdownChartTooltip.tsx
```diff
@@ -201,6 +203,12 @@ export function BreakdownChartTooltipContent({
               {__('tooltips.ofinstalled')}
               <br />
               <MetricRatio value={usage} total={(capacity ??= 0)} format={formatPower} />
+              {capacitySource && (
+                <small>
+                  {' '}
+                  ({__('country-panel.source')}: {capacitySource})
```
VIKTORVAV99, line 209: Do you know if we have more than one source for a single capacity mode for any zone?

VIKTORVAV99 (approved):
🥳

Left a tiny comment but I don't think it's really relevant (yet at least).

web/src/features/charts/tooltips/BreakdownChartTooltip.tsx
tonypls: I'm 99% sure we don't. In the future if we add sources for the annual view then there will probably be multiple sources for one generation type and hopefully this pattern will work

VIKTORVAV99: Yeah otherwise we just add a intl list format and it will look all nice with multiple entries.

PR #6488, opened 2024-02-15
update(capacity-config) - PA

mathilde-daugy:
## Issue

image

## Description

capacity outdated and unavailable from EMBER so switched to IRENA

### Double check

### DIFF
--- a/config/zones/PA.yaml
+++ b/config/zones/PA.yaml
@@ -4,16 +4,50 @@ bounding_box:
   - - -77.2425664944
     - 9.61161001224
 capacity:
-  biomass: 33
-  coal: 307.86
-  gas: 449.5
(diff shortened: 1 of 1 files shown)


config/zones/PA.yaml
```diff
@@ -4,16 +4,50 @@ bounding_box:
...
-  hydro storage: 0
-  nuclear: 0
-  oil: 243
-  solar: 522
-  unknown: 0
-  wind: 270
+  biomass:
+    datetime: '2022-01-01'
+    source: IRENA.org
+    value: 68.0
```
VIKTORVAV99, line 10: Tony added a PR that made sure all of these were list items instead of objects so things worked properly in other areas. Could we do the same here?

PR #6489, opened 2024-02-15
update return types for capacity_update_configuration

mathilde-daugy:
## Issue

see #6484 - Our yaml isn't consistent which means it's not loading in correctly to our ModeCapacity class

## Description

- update to `update_zone_capacity_config` and `generate_zone_capacity_config` so that the new capacity format for each mode is a list rather than a dict

### Preview

image

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing it at the source!

PR #6494, opened 2024-02-17
Update react-map-gl to version 7.1.7

VIKTORVAV99:
## Issue
We are using outdated dependencies that has us using workarounds that have since been fixed upstreams.

## Description

This PR is pretty straight forward actually but unfortunately it contains a lot of changes in code (but not functionality) to satisfy type checking and upstream changes.

> [!IMPORTANT]
> This bumps the minimum supported version of WebGL to 2, this might have an impact on the minimum iOS version we can support.

### Double check


VIKTORVAV99:
I thought this would be a easy PR but I for the life of me can't seem to figure out a way to satisfy the remaining type errors without significant changes.

If you want to take a look @madsnedergaard or @tonypls please feel free to do so otherwise I think I'll simplify this PR a bit and just update maplibre for now.

web/src/features/map/map-layers/StatesLayer.tsx
```diff
@@ -1,42 +1,32 @@
...
-  } as mapboxgl.LinePaint;
-
-  const stateLabelLayout = {
-    'symbol-placement': 'point',
-    'text-size': 12,
-    'text-letter-spacing': 0.12,
-    'text-transform': 'uppercase',
-    'text-font': ['poppins-semibold'],
-  } as mapboxgl.SymbolLayout;
-
```
VIKTORVAV99, line 24: Honestly I have no idea why I had to inline these for the typing to work but that is the only thing that seems to fix the issue.

If someone can explain it please do 😅

web/src/features/map/map-layers/ZonesLayer.tsx
```diff
@@ -9,69 +9,71 @@ export default function ZonesLayer() {
...
-          ['boolean', ['feature-state', 'selected'], false],
-          0.7,
-          0,
-        ],
-      } as mapboxgl.FillPaint,
-    }),
-
-    [theme]
-  );
-
```
VIKTORVAV99, line 60: Same here 🫠

VIKTORVAV99:
While the ESLint checks that are flagged are valid they were already existing in the codebase before so I won't fix them in this PR. It is safe to merge them anyway.

VIKTORVAV99:
Requested a review from both of you since this changes some core stuff. Better to get 4 eyes on it than 2 and I don't want to break stuff 😅.

web/src/features/weather-layers/wind-layer/WindLayer.tsx
```diff
@@ -16,7 +15,7 @@
 
 type WindyType = ReturnType<typeof Windy>;
 let windySingleton: Maybe<WindyType> = null;
-const createWindy = async (canvas: HTMLCanvasElement, data: any, map: MapboxMap) => {
+const createWindy = async (canvas: HTMLCanvasElement, data: any, map: maplibregl.Map) => {
```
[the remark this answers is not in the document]
VIKTORVAV99: Unrelated to my changes, can be ignored in this PR.

web/cypress/e2e/map.spec.ts
```diff
@@ -52,6 +52,6 @@
     // cy.get('[data-test-id=dark-mode-button]').click().click();
 
     // eslint-disable-next-line cypress/require-data-selectors
-    cy.get('.maplibregl-ctrl-zoom-in.mapboxgl-ctrl-zoom-in').click();
+    cy.get('.maplibregl-ctrl-zoom-in.maplibregl-ctrl-zoom-in').click();
```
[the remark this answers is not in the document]
VIKTORVAV99: Also unrelated to my changes, but we should look into it in the future during a slack week sometime.

tonypls:
Quick thing I've noticed is the solar layer won't work unless you refresh with it switched on then reload and turn it on and off

VIKTORVAV99:
> Quick thing I've noticed is the solar layer won't work unless you refresh with it switched on then reload and turn it on and off

Not exactly sure why it was broken but I combined the hooks in the solar layer as they where basically doing the same thing and now it works 🤷🏻‍♂️

The benefit of this is that we also have one less hook that will fire which should improve performance ever so slightly 🙂

web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -16,7 +15,7 @@ interface ExchangeArrowProps {
   data: ExchangeArrowData;
   viewportWidth: number;
   viewportHeight: number;
-  map: MapboxMap;
+  map: maplibregl.Map;
```
madsnedergaard, line 18: How does this actually work when the import is not defined and we don't set it as a global value anywhere? 🤔

web/src/features/map/map-layers/ZonesLayer.tsx
madsnedergaard: It's quite annoying, but I guess we can always try to improve it again later ¯\\\_(ツ)\_/¯

madsnedergaard (approved):
Looks good to me now :)

VIKTORVAV99:
> Looks good to me now :)

Now the only question is if we need to bump the min iOS version or not?

https://caniuse.com/?search=webgl2

Only 0.25% use pre iOS 15 versions globally and I suspect it's less in our case?

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: Maplibre is now exporting its own types that TS is picking up and since it's using a namespace it don't need importing. Importing Map directly also conflicts with the built-in Map type so I thought this was better.

madsnedergaard: Ah okay, fair enough - it's a bit too much "magic" in my opinion when you can use something that isn't declared anywhere, but if it works and it's only for types I guess it can't hurt :D

VIKTORVAV99: It's the same as the react and html or the builtin types that "just work".

VIKTORVAV99:
I can't seem to get the local IOS version to build, it fails at:
```
✖ Updating iOS native dependencies with pod install - failed!
```
Could one of you give it a go with the target set to iOS 15 @madsnedergaard or @tonypls?

madsnedergaard:
> I can't seem to get the local IOS version to build, it fails at:
> 
> ```
> ✖ Updating iOS native dependencies with pod install - failed!
> ```
> 
> Could one of you give it a go with the target set to iOS 15 @madsnedergaard or @tonypls?

Screenshot 2024-03-18 at 16 10 03


Seems all good even when I change "min target version" to iOS 15 :)

VIKTORVAV99:
I have updated the iOS targets to set iOS 15 as the min version, built it locally but wouldn't mind a second look at it.

madsnedergaard:
LGTM with new changes as well :)

VIKTORVAV99:
Will merge it then 🙂

PR #6495, opened 2024-02-17
feat(parsers): add BB parser

### LINKED ISSUE #6459: Data for Barbados: generation & demand (fossil, wind, solar, storage charge & discharge)
# Production
Hourly solar/wind/fossil/discharge graphs for ~1½ day ago (date at the bottom of each graph): https://dataservices.acelerex.com/barbados/grid-operation/generation/current
 - At the top of the page it says "Values: &lt;date> &lt;time>", but the values at the top of the page do not align with the values in the graphs at the same time stamp. I wondered if it could be (near) real-time generation, but it turned out that if I [converted that timestamp from *my* time zone to Barbados time](https://www.timeanddate.com/worldclock/converter.html?iso=20240206T100000&p1=46) I found a matching data point in the graph at that time.

# Consumption
Hourly  demand and storage charging graphs for ~1½ day ago: https://dataservices.acelerex.com/barbados/grid-operation/demand/current
 - At the top of the page it says "Values:", this time without a date or timestamp. Again there was a matching data point if I converted the last whole hour local time to Barbados time.

----
Note: I found the dashboard embedded at [this page](https://energy.gov.bb/dashboard/)

Ugzuzg:
## Issue

Closes #6459

## Description

A parser for BB with `fetch_production` and `fetch_consumption` implemented.

### Double check


parsers/BB.py
```diff
@@ -0,0 +1,181 @@
...
+    production_list = ProductionBreakdownList(logger=logger)
+
+    for _index, row in operational_data.iterrows():
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=datetime.fromisoformat(row["datetime"]),
+            production=ProductionMix(
+                solar=row["24 hr monitored solar production"],
+                wind=row["24 hr monitored wind production"],
+                unknown=row["24 hr fossil production"],
```
Ugzuzg, line 138: Not sure what to map fossil production to.

silimotion: Hi!
I've found this info in a PDF in the energy.gov.bb webpage:

Screenshot_20240217_174656

The data, albeit dated, shows thermal is oil. "Bagasse" is a bit confusing as it seems to not be reported on the API, or to be aggregated with oil.

q--: https://dataservices.acelerex.com/barbados/grid-operation/generation/long-term-forecast shows, "Long Term Forecast" tab shows they currently don't have, or at least didn't forecast having, biomass or landfill gas capacity.

Maybe it doesn't exist anymore?

silimotion: Barbados IRENA profile shows no bioenergy capacity. Maybe there is some plant that uses mixed fuel? Or maybe some factory burns bagasse for self consumption?

parsers/BB.py
```diff
@@ -0,0 +1,181 @@
...
+            self.token = self.get_token()
+        r.headers["Authorization"] = f"Bearer {self.token}"
+        return r
+
+    def get_token(self) -> str:
+        url = base_url + "/authentication"
+        response = requests.post(
+            url, json={"dashboardId": "barbados", "strategy": "jwt-public"}
+        )
+        return response.json()["accessToken"]
```
Ugzuzg, line 39: The token is issued for one hour. Will a single session ever run for longer than that? Is historical data fetched in a single session?

Wondering if implementing token renewal is needed.

parsers/BB.py
q--: I tried looking for context. This document which I found by googling for ["Fuel Inputs for Electricity Production 2015 Barbados"](https://faolex.fao.org/docs/pdf/bar199547.pdf) says 

>Over the last 20 years, the fuel feedstock used in Barbados for the production of electricity
consisted in large measure of fossil fuels. The main fuels included heavy fuel oil, kerosene,
diesel and natural gas. Within the sugar industry, electricity production was mainly fuelled
by bagasse.


Elsewhere it says:

>Also, in light of a declining sugar cane industry on the island, there is increasing confidence
in the agricultural sector that there can be greater use of biomass and biogas in the
energy sector. Bagasse has, at times, been used to generate energy in the sugar cane
industry and there is considerable interest in building on this experience.

It's quite possible that bagasse is no longer used, or that it's only used in captive power plants that aren't grid-connected.

VIKTORVAV99 (commented):
Left some notes, most are just some small suggestions but some need to be looked at more closely.

I would also love if this had some tests associated with it, a simple snapshot test is probably the easiest.

parsers/BB.py
VIKTORVAV99: Should not be needed as far as I know, but it also wouldn't hurt to have it in there if you want to do it.

parsers/BB.py
```diff
@@ -0,0 +1,181 @@
...
+    production_list = ProductionBreakdownList(logger=logger)
+
+    for _index, row in operational_data.iterrows():
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=datetime.fromisoformat(row["datetime"]),
+            production=ProductionMix(
+                solar=row["24 hr monitored solar production"],
+                wind=row["24 hr monitored wind production"],
+                oil=row["24 hr fossil production"],
```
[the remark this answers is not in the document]
Ugzuzg: There's a discussion above which suggests that it is.

parsers/BB.py
```diff
@@ -0,0 +1,181 @@
...
+        expected_date = target_datetime.astimezone(
+            ZoneInfo("America/Barbados")
+        ).strftime("%Y-%m-%d")
+
+        for i in range(10):
+            url_date = (
+                (target_datetime + timedelta(days=i))
+                .astimezone(ZoneInfo("America/Barbados"))
+                .strftime("%Y-%m-%d")
+            )
```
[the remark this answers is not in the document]
Ugzuzg: No, it's trying to find the requested date. The problem is that the api query only allows filtering by the date of document creation, not the date of actual data. On 2000-01-03 (document created_at) they post a dataset for 2000-01-01 or 2000-01-02. We can query by created_at, but it's not the actual date we want, so we need to look through a couple of days to find the one we want in the response itself.

parsers/BB.py
```diff
@@ -0,0 +1,181 @@
...
+                wind=row["24 hr monitored wind production"],
+                oil=row["24 hr fossil production"],
+            ),
+            storage=StorageMix(
+                battery=(
+                    row["24 hr storage discharging"] + row["24 hr storage charging"]
+                )
+                * -1
+            ),
+            source="dataservices.acelerex.com/barbados",
```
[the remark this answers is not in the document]
Ugzuzg: I'm not completely sure what the source should be, but the domain name on its own makes little sense, as you won't be able to get to the actual source by opening it.

But if it is always just a domain, I'll change it.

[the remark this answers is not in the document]
q--: Another option would be to change it to `energy.gov.bb`; after all, that's where the dashboard is [embedded](https://energy.gov.bb/dashboard/), and you wouldn't know the actual dashboard is hosted at another domain unless you did "Open Frame in new tab" (or the like) in your browser.

I don't speak for ElectricityMaps, but that would be my preference.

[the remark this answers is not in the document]
Ugzuzg: I like that, @q--. @VIKTORVAV99, thoughts?

parsers/BB.py
q--: I also noticed in the ["Barbados National Energy Policy (BNEP) 2019 - 2030"](https://energy.gov.bb/download/national-energy-policy-2019-2030/), probably where you found the diagram, that the values in the diagram are rounded. One page up (page 42), it has the percentages with one decimal point, showing, among others, that a bit of natural gas was also used for electricity generation in 2015:

>Heavy Fuel Oil – 74%
>• Kerosene – 17.4%
>• Bagasse – 5.6%
>• Diesel – 2.2%
>• Natural Gas – 0.2%
>• Solar – 0.6%
(It also notes that solar electricity increased to 2.4% in 2016, though we don't need that info as the data source returns solar data.)

The page also has the feedstock consumed in MMBTU (million British Thermal Units), which should allow for an even more exact calculation.

It also says "* Accounts for use of natural gas to produce electricity in the manufacturing sector.", which might imply [embedded generation](https://en.wikipedia.org/wiki/Captive_power_plant).
-----
Anyway, no more recent data than 2015

parsers/BB.py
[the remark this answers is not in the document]
VIKTORVAV99: Works for me 👍🏻

parsers/BB.py
q--: I found some graphs in the 2021 "[MM_IADB_Final IRRP Report_Activity B – Energy.gov.bb](https://energy.gov.bb/download/mm_iadb_final-irrp-report_activity-b/) that suggest that, in 2021, Barbados only had oil-based and solar-based electricity.

Figure 7.10 and 7.11 from the linked report showing two scenarios with identical installed capacity in 2021 as a starting point

VIKTORVAV99 (approved):
Code wise this looks good, so approving.

And if everyone involved in this PR agrees on the mode mapping we can merge it right away.

Thanks for doing this!

parsers/BB.py
Ugzuzg: These charts suggest that "fossil" is oil, right?

q--: Yes. My main remaining reservation is that it's still somewhat dated... I wish we had more recent data.

But I guess tracking it as oil is okay.

q--:
Thanks Viktor  and Ugzuzg!

PR #6496, opened 2024-02-18
Fix entsoe capacity parser

VIKTORVAV99:
## Issue

The parser was still operating under the assumption that we fetched 2 years of data when we do not. It also didn't use the new internal mapping for the fuel types aggregating hydro storage with hydro production.

## Description

Fixes the parser + updates the capacities for 2023 and 2024.

### Double check

### DIFF
--- a/capacity_update.py
+++ b/capacity_update.py
@@ -53,4 +53,4 @@ def capacity_update(
         update_zone(zone, parsed_target_datetime, session, update_aggregate)
 
     print("Running prettier...")
-    run_shell_command("pnpx prettier@2 --write .", cwd=ROOT_PATH)
+    run_shell_command("pnpx prettier@2 --write config/zones --cache", cwd=ROOT_PATH)
--- a/config/zones/PT.yaml
+++ b/config/zones/PT.yaml
@@ -7,10 +7,10 @@ capacity:
   biomass:
     - datetime: '2023-01-01'
       source: entsoe.eu
-      value: 681.0
+      value: 647.0
     - datetime: '2024-01-01'
       source: entsoe.eu
-      value: 647.0
+      value: 681.0
   coal:
     - datetime: '2023-01-01'
       source: entsoe.eu
@@ -18,10 +18,10 @@ capacity:
   gas:
     - datetime: '2023-01-01'
       source: entsoe.eu
-      value: 4520.0
+      value: 4585.0
     - datetime: '2024-01-01'
       source: entsoe.eu
-      value: 4585.0
+      value: 4433.0
   geothermal:
     - datetime: '2023-01-01'
       source: entsoe.eu
@@ -29,15 +29,12 @@ capacity:
   hydro:
     - datetime: '2023-01-01'
       source: entsoe.eu
-      value: 4371.0
+      value: 4489.0
     - datetime: '2024-01-01'
       source: entsoe.eu
-      value: 4489.0
+      value: 4485.0
   hydro storage:
     - datetime: '2023-01-01'
-      source: entsoe.eu
-      value: 2827.0
-    - datetime: '2024-01-01'
       source: entsoe.eu
       value: 3707.0
   nuclear:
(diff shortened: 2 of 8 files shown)


electricitymap/contrib/capacity_parsers/ENTSOE.py
```diff
@@ -69,21 +70,15 @@ def fetch_production_capacity(
...
-                    "datetime": end_date.strftime("%Y-01-01"),
-                    "source": SOURCE,
-                }
-            capacity_dict[ENTSOE_PARAMETER_BY_GROUP[fuel_code]]["value"] += value
+        point = timeseries.find_all("point")
+        value = float(point[0].find_all("quantity")[0].contents[0])
+        if ENTSOE_CODE_TO_EM_MAPPING[fuel_code] not in capacity_dict:
+            capacity_dict[ENTSOE_CODE_TO_EM_MAPPING[fuel_code]] = {
+                "value": 0,
+                "datetime": f"{target_datetime.year}-01-01",
```
VIKTORVAV99, line 78: Why are we hardcoding the month and day here instead of using the returned start data (with hours). They differ by a few hours for the zones which is what messed up PT capacities.

config/zones/PT.yaml
VIKTORVAV99, line 1: Basically all of the capacities here were wrong by 1 year as the start and end date was offset by 1 day (floored) compared to the other zones. I do not really know why.

electricitymap/contrib/capacity_parsers/ENTSOE.py
```diff
@@ -53,7 +54,7 @@ def query_capacity(
         session,
         params,
         target_datetime=target_datetime,
-        span=(0, 72),
+        span=(0, 72),  # DO NOT USE A NEGATIVE LOOKBACK
```
VIKTORVAV99, line 57: Since the lookup don't include a negative datetime we only get 1 year now.

electricitymap/contrib/capacity_parsers/ENTSOE.py
```diff
@@ -69,21 +70,15 @@ def fetch_production_capacity(
...
-            if ENTSOE_PARAMETER_BY_GROUP[fuel_code] not in capacity_dict:
-                capacity_dict[ENTSOE_PARAMETER_BY_GROUP[fuel_code]] = {
-                    "value": 0,
-                    "datetime": end_date.strftime("%Y-01-01"),
-                    "source": SOURCE,
-                }
-            capacity_dict[ENTSOE_PARAMETER_BY_GROUP[fuel_code]]["value"] += value
+        point = timeseries.find_all("point")
+        value = float(point[0].find_all("quantity")[0].contents[0])
+        if ENTSOE_CODE_TO_EM_MAPPING[fuel_code] not in capacity_dict:
```
VIKTORVAV99, line 75: This new EM mapping was not used meaning hydro production and hydro storage was combined when ran.

capacity_update.py
```diff
@@ -53,4 +53,4 @@ def capacity_update(
         update_zone(zone, parsed_target_datetime, session, update_aggregate)
 
     print("Running prettier...")
-    run_shell_command("pnpx prettier@2 --write .", cwd=ROOT_PATH)
+    run_shell_command("pnpx prettier@2 --write config/zones --cache", cwd=ROOT_PATH)
```
VIKTORVAV99, line 56: `--cache` just speed things up by basically skipping files that have not been changed and limiting the scope to `config/zones` ensures only the zone config files are checked instead of everything in contrib.

mathilde-daugy (approved):

config/zones/PT.yaml
q--: It's the only EU country other than Ireland that's in this time zone, in the rest of the EU it's at least an hour later. Could that be the cause?

VIKTORVAV99: Ah that could be it, I thought all time zones was in UTC but I haven't double checked that.

PR #6497, opened 2024-02-20
Improve Canary Islands oil emission factors

silimotion:
## Description

This PR adds better emission factors for the Canary Islands. In a similar fashion to my previous #6406, I've used the ETS emissions data. In this case, I've got the total electricity produced each year from the EM app. I have no reason to think this data isn't accurate, as the there is no estimation in the data, and the results are similar to other emission factors published by REE. I didn't use those because they are aggregated and less accurate.

In islands where there were several power plants, I've added the emissions values, and divided by the total production amount. The mappings are as follows:

- El Hierro: 
  - Llanos Blancos
- La Palma: 
  - Los Guinchos
- La Gomera: 
  - El Palmar
- Tenerife:
  - Granadilla
  - Candelaria
  - Guía de Isora
  - Arona
- Gran Canaria:
  - Barranco de Tirajana
  - Jinámar
- Fuerteventura-Lanzarote:
  - Las Salinas
  - Punta Grande

As a note:

This same process could be used for other Spanish regions, but the historical data available in the app is not valid. In some cases, because there is no data for the ETS years (Ceuta, Melilla). In others because the data was being recorded wrongfully (this was resolved in the parser rewrite) and has not been updated yet. In those cases the data could be updated in the backend, or I could build my own script to download the data.

In Mallorca there are plants with several fuels so the ETS path is useless.

### Double check


silimotion:
The ones fixed in the rewrite were Ibiza and Menorca. Formentera mapping was changed so I think it could be rerun or at least edited. I think that in Mallorca there were no changes.

If it is possible it would be nice to also update Cyprus as I've seen some years have big estimation proportions, creating spurious "unknown" production. Also, the oil factors were changed recently but the calculated factors for total production were not updated. I don't know if it is a bug or just that the backend needs more time. Cyprus historical parser did not record capacity so be careful not to overwrite anything.

VIKTORVAV99:
A lot of these will change by next month as I have already re-fetched and reprocessed a lot of the data (Cyprus included). But it won't be noticeable until the aggregates update.

VIKTORVAV99:
I'll have to take a look at the Spanish islands though, if the mapping was changed completely and the old modes are None now it might not get overwritten properly.

config/zones/ES-CN-FVLZ.yaml
```diff
@@ -196,6 +202,10 @@ parsers:
   consumption: ES.fetch_consumption
   production: ES.fetch_production
 sources:
+  EU-ETS, REE:
+    link:
+      https://drive.google.com/file/d/15UeUHyhcjVw8fGLc0Zx_HELSx20cw_-e/view?usp=share_link,
+      https://demanda.ree.es
```
VIKTORVAV99, line 208: This seems to be the same link as for ENTSO-E, this seems like a misstake? Did you forget to update it?

VIKTORVAV99 (changes requested):
As far as I can tell the emission factors themselves are reasonable. But the source link seems to be the same as the one we use for ENTSO-E which don't seem right.

config/zones/ES-CN-FVLZ.yaml
VIKTORVAV99: The source needs to be a link to the actual calculation so they can be verified by anyone at a later date.

It can be a spreadsheet or Jupyter notebook or something similar but it needs to have some sort of reproducibility.

silimotion: Ok... I can make a notebook with the calculations... Where can I upload it though? I assume you would want to have edit access, etc...?

VIKTORVAV99: https://colab.research.google.com would be a good place to upload it, we don't need write access just read access so we can copy it if need be.

silimotion:
Hi!
I was a bit busy with classes and it has been a long time, but I just finished the notebook!
It is available in [Github](https://github.com/silimotion/Emission_factors_research/blob/main/Spanish_non_peninsular_factors.ipynb) and in [Colab](https://colab.research.google.com/github/silimotion/Emission_factors_research/blob/main/Spanish_non_peninsular_factors.ipynb).

I appreciate any feedback or questions!

Some other notes:
- Since I opened this new ETS data has been published, so I added it.
- I added new factors for the Balearic Islands as well.
- I couldn't find ETS data for the biomass (incinerator plants) in Mallorca and Melilla.
- I didn't add factors for Ceuta or Melilla because they both have a power plant with gas and oil production.
  - In the case of Melilla, however, it seems "gas turbines" run on oil so the parser's mapping may need to be changed. If so, then the only fuel would be oil, and factors could be calculated.
  - In the case of Ceuta, they seem to actually run on gas but their production is insignificant, so maybe we could use some heuristic or approximation for this zone.

VIKTORVAV99:
Thanks for the update!

I'm pretty busy at the moment but I'll try and take a look at this the next few days! But it looks pretty good at first glance!

madsnedergaard (approved):
This is amazing, thanks for your work on this 😍 
I took the liberty of adding "IPCC 2014" to the lifecycle sources to indicate that this is where we have the LCA "addition" from the direct numbers, but otherwise I think this work is stellar and it's great to have the notebook!

Issue #5837, opened 2023-09-06
How should we name balancing authorities in USA? Feedback needed! 🙏 


madsnedergaard:
I'd like to get some feedback on how people would prefer the balancing authorities in US to be named.

Currently we're going with the full title on the app (e.g. California Independent System Operator) instead of the shorter California ISO (what they use on [their own website](http://www.caiso.com/about/Pages/News/default.aspx)) or even shorter CAISO.

Same with ERCOT vs Electric Reliability Council of Texas, and so on.

### Where we need your help

Since the internal team is either European or deeply involved in the energy sector, we're lacking an understanding of how everyday citizens in the US uses these names. How does the media use the names?

The current titles are often too long to be displayed in the app, so it would be nice if we could generally use the shorter names and maybe add the full title in a tooltip - but we don't want to make that move before hearing from you :)

madsnedergaard:
Thanks @brandongalbraith, that's very helpful 🙏

fhteagle:
Short abbreviation + tooltip for full BA name makes sense to me as well. In the specific case of the BA for my location, WAPA -> Western Area Power Administration is probably good enough. The geopgraphical and system boundary zones within WAPA are kind of a mess anyway. So trying to point at US-NW-WACM to match the EIA reporting pools is either necessary or overkill, depending on how you look at it.

madsnedergaard:
Thanks both! Just to make sure it's clear, it's column D that is the new, simplified name :)
So we are already planning on droppoing PUD.

> I would drop the hyphens

Do you mean the ones used in the new name, like "Duke Energy - Progress East"?

fhteagle:
What I get for looking at it on mobile, did not even see column D lol

Agreed that the hyphens are probably not necessary. "WAPA Rocky Mountain" works for me.

PR #6498, opened 2024-02-20
Use special "displayName" for zones and show full name in tooltip

madsnedergaard:
## Issue

Part 1 of #5837

## Description

This makes it possible show to a shorter name for a zone, while having the full name shown in a tooltip. This means we can use "ERCOT" for Texas in tooltips, while still having the actual name available to anyone curious :)

I will follow up with a second PR with the actual translation updates.

### Double check

### DIFF
--- a/web/src/features/panels/zone/ZoneHeaderTitle.tsx
+++ b/web/src/features/panels/zone/ZoneHeaderTitle.tsx
@@ -3,7 +3,12 @@ import { TimeDisplay } from 'components/TimeDisplay';
 import TooltipWrapper from 'components/tooltips/TooltipWrapper';
 import { HiArrowLeft } from 'react-icons/hi2';
 import { Link } from 'react-router-dom';
-import { getCountryName, getZoneName, useTranslation } from 'translation/translation';
+import {
+  getCountryName,
+  getFullZoneName,
+  getZoneName,
+  useTranslation,
+} from 'translation/translation';
 import { createToWithState } from 'utils/helpers';
 
 import { getDisclaimer } from './util';
@@ -14,13 +19,19 @@ interface ZoneHeaderTitleProps {
   isAggregated?: boolean;
 }
 
+const MAX_TITLE_LENGTH = 25;
+
 export default function ZoneHeaderTitle({ zoneId }: ZoneHeaderTitleProps) {
   const { __ } = useTranslation();
-  const title = getZoneName(zoneId);
+  const zoneNameShort = getZoneName(zoneId);
+  const zoneNameFull = getFullZoneName(zoneId);
+  const zoneName = zoneNameFull.length < MAX_TITLE_LENGTH ? zoneNameFull : zoneNameShort;
+  const showTooltip =
+    zoneName !== zoneNameFull || zoneNameShort.length >= MAX_TITLE_LENGTH;
   const returnToMapLink = createToWithState('/map');
   const countryName = getCountryName(zoneId);
   const disclaimer = getDisclaimer(zoneId);
-  const showCountryPill = zoneId.includes('-') && !title.includes(countryName);
+  const showCountryPill = zoneId.includes('-') && !zoneNameShort.includes(countryName);
 
   return (
     <div className="flex w-full grow flex-row overflow-hidden pb-2 pl-2">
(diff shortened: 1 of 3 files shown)


VIKTORVAV99:
@madsnedergaard I really like this idea but can we try and get some of the other open PRs that affects translations first? In theory this PR shouldn't cause merge conflicts if we just add names but you never know; and I don't want merge conflicts in all the PRs, including this one 😅

https://github.com/electricitymaps/electricitymaps-contrib/pull/6227 should be ready to go, and then once that is merged I should be able to finish up the translations in https://github.com/electricitymaps/electricitymaps-contrib/pull/6443 without merge conflicts (and doing the same work twice).

web/src/translation/translation.ts
```diff
@@ -32,29 +32,18 @@ export const translateIfExists = (key: string) => {
...
-export const getZoneName = (zoneCode: string) =>
-  translateIfExists(`zoneShortName.${zoneCode}.zoneName`);
+export const getFullZoneName = (zoneCode: string) => {
+  return translateIfExists(`zoneShortName.${zoneCode}.zoneName`);
+};
+export const getZoneName = (zoneCode: string) => {
+  const displayName = translateIfExists(`zoneShortName.${zoneCode}.displayName`);
+  const fullName = getFullZoneName(zoneCode);
+  return displayName || fullName;
+};
```
VIKTORVAV99, line 42: If you want to do a micro optimisation size wise you can extract the `zoneShortName.${zoneCode}.zoneName` to arrow function to avoid the extra string size from declaring it twice 😅

madsnedergaard:
Ready for review @VIKTORVAV99, I have also added a single zone for now so we can better test the functionality :)

web/src/translation/translation.ts
```diff
@@ -4,29 +4,18 @@ export const translateIfExists = (key: string) => {
   return i18next.exists(key) ? i18next.t(key) : '';
 };
 
-export const getZoneName = (zoneCode: string) =>
-  translateIfExists(`zoneShortName.${zoneCode}.zoneName`);
+export const getFullZoneName = (zoneCode: string) => {
+  return translateIfExists(`zoneShortName.${zoneCode}.zoneName`);
+};
+export const getZoneName = (zoneCode: string) => {
+  const displayName = translateIfExists(`zoneShortName.${zoneCode}.displayName`);
```
VIKTORVAV99, line 11: Have we tested that this translateIfExists works properly with the i18n system?

I can't remember if I did that or not when migrating.

web/src/features/panels/zone/ZoneHeaderTitle.tsx
```diff
@@ -14,12 +14,18 @@ interface ZoneHeaderTitleProps {
   isAggregated?: boolean;
 }
 
+const MAX_TITLE_LENGTH = 25;
+
 export default function ZoneHeaderTitle({ zoneId }: ZoneHeaderTitleProps) {
-  const title = getZoneName(zoneId);
+  const zoneNameShort = getZoneName(zoneId);
```
VIKTORVAV99, line 20: There is no guarantee this is actually shorter since it only would be if there is a display name property for the zone.

I think we should just rename it to zoneDisplayName (and the other to zoneFullName).

VIKTORVAV99 (commented):
The styling here don't match the details page, is this something we want to adress in here?
image
It's more noticeable now since the title doesn't get truncated.

Perhaps a little out of scope but could you increase the title length a tiny bit here as well if you change the tooltip?
image

web/src/translation/translation.ts
madsnedergaard: We are already using it, so I sure hope that you did test it 😅 
But should be fine, otherwise current production app would be broken

web/src/translation/translation.ts
```diff
@@ -4,29 +4,18 @@ export const translateIfExists = (key: string) => {
...
+  const fullName = getFullZoneName(zoneCode);
+  return displayName || fullName;
```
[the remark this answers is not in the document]
madsnedergaard: Yeeeah, I thought about that too... Done now!

madsnedergaard:
> The styling here don't match the details page, is this something we want to adress in here?

Did some changes now.

> Perhaps a little out of scope but could you increase the title length a tiny bit here as well if you change the tooltip?

Fixed (and added Storybook stories as it was a pain to develop on) :)

web/src/translation/translation.ts
VIKTORVAV99: So do I! 😅

But remember who did the thorough review 😉

VIKTORVAV99 (approved):
One tiny thing, otherwise this looks good to me!

web/src/features/map/MapTooltip.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+          zoneId={zoneKey}
+          date={'2022-01-01'}
+        />
+      </div>
+    </div>
+  );
+}
+
+const meta: Meta<typeof TestWrapper> = {
+  title: 'Modal/MapTooltip',
```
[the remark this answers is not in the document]
madsnedergaard: Good catch! :D

PR #6501, opened 2024-02-21
Avoid showing the "estimated" badge on graphs only based on outage message

madsnedergaard:
## Issue

Closes https://github.com/electricitymaps/electricitymaps-contrib/issues/6499

## Description

We should not assume data is estimated just because there is an outage message as it can be there for other reasons :)

### Preview

See staging branch.

### Double check

### DIFF
--- a/web/src/features/panels/zone/ZoneDetails.tsx
+++ b/web/src/features/panels/zone/ZoneDetails.tsx
@@ -63,7 +63,7 @@ export default function ZoneDetails(): JSX.Element {
   const zoneMessage = data?.zoneMessage;
   const cardType = getCardType({ estimationMethod, zoneMessage, timeAverage });
   const hasEstimationPill =
-    ['estimated', 'outage'].includes(cardType) || Boolean(estimatedPercentage);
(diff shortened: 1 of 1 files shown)


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -63,7 +63,7 @@ export default function ZoneDetails(): JSX.Element {
   const zoneMessage = data?.zoneMessage;
   const cardType = getCardType({ estimationMethod, zoneMessage, timeAverage });
   const hasEstimationPill =
-    ['estimated', 'outage'].includes(cardType) || Boolean(estimatedPercentage);
+    ['estimated'].includes(cardType) || Boolean(estimatedPercentage);
```
VIKTORVAV99, line 66: At this point I don't think we need to use includes anymore... 😅

madsnedergaard: lol, fair point :D

VIKTORVAV99 (approved):

Issue #6490, opened 2024-02-15
[Data Issue]: FI outdated wind and solar capacity


msaynevirta:
### When did this happen?

Since generation capacity parsers were implemented.

### What zones are affected?

FI

### What is the problem?

Solar and wind capacity data is out of date for Finland. TSO Fingrid's API provides more up to date data for these:
- https://data.fingrid.fi/en/datasets/267
- https://data.fingrid.fi/en/datasets/268

VIKTORVAV99:
@mathilde-daugy how do we want to handle cases like this? Can we update the capacities monthly or will that inflate the config files to unmanageable levels?

In this case I'd even want a live (separate) parser for it but we don't support that yet.

mathilde-daugy:
We can't update the capacities every month or it will create a monster. The idea is to update capacities at most every quarter because it also takes some time. 

You are welcome to build a specific parser that combines Fingrid data and ENTSOe data :)

mathilde-daugy:
> I checked this with Fingrid. According to the ENTSO-E Transparency Platform's guidelines, only individual power plant capacities over 1 MW is to be reported there. For solar power, this is industrial scale, which is still really small in Finland, but it is accelerating! Most solar power in Finland (currently ~991 MW of ~1018 MW) is ”roof top” <1MW.

This is very valuable, thank you :)

pnook:
What is nowadays the way to update the capacity values? Manually or via the procedure mentioned above? According to Fingrid data, e.g. wind power capacity is 1GW more than shown in electricitymaps data.

VIKTORVAV99:
> What is nowadays the way to update the capacity values? Manually or via the procedure mentioned above? According to Fingrid data, e.g. wind power capacity is 1GW more than shown in electricitymaps data.

Both work, the important part is to follow the existing structure in the code with the date versioned capacities.

PR #6505, opened 2024-02-22
update(capacity) - new source for FI solar capacity 

mathilde-daugy:
## Issue
see issue #6490 

## Description

- Create capacity for FI that merges ENTSOE capacity and FINGRID capacity (solar and wind)
- Updated config

### Preview

image

### Double check


electricitymap/contrib/capacity_parsers/FI.py
```diff
@@ -0,0 +1,62 @@
+from datetime import datetime, timedelta
+from typing import Any
+
+from requests import Response, Session
+
+from electricitymap.contrib.capacity_parsers.ENTSOE import (
+    fetch_production_capacity as fetch_entsoe_production_capacity,
+)
```
VIKTORVAV99, line 8: I'm a little concerned about importing other parsers without having some form of testing in place to ensure we don't break other parsers. 

For example I have https://github.com/electricitymaps/electricitymaps-contrib/pull/6496 open which could change how things are working in here.

mathilde-daugy:
will look at testing on monday

mathilde-daugy:
@VIKTORVAV99 I see that you resolved your own comment on this PR. Is there anything you need me to do in particular or can we merge this?

VIKTORVAV99:
Ah no, I'll review it right away, I had just kind of forgotten about this 😬

electricitymap/contrib/capacity_parsers/FI.py
```diff
@@ -0,0 +1,78 @@
...
+from electricitymap.contrib.config import ZoneKey
+
+API_KEY = "a411d83c96bb4abb8812dba80bdaed64"
```
VIKTORVAV99, line 11: We really should not store the API key here. It should be a ENV variable.

(PS: you should revoke this API key if possible as it's already in the git history).

VIKTORVAV99 (changes requested):

electricitymap/contrib/capacity_parsers/FI.py
mathilde-daugy: done!

VIKTORVAV99 (approved):
LGTM!

PR #6506, opened 2024-02-22
fix(UY): Sanitise UY parser

pierresegonne:
## Issue

We've been parsing production events where all modes are at 0 for UY. These events cause downstream issues as they are not flagged for estimation.

## Description

* Adopts parser validation for the fetch function in the `UY` parser. 
* Sets all production to null if all is 0

### Double check

### DIFF
--- a/parsers/UY.py
+++ b/parsers/UY.py
@@ -18,6 +18,7 @@ from electricitymap.contrib.lib.models.event_lists import (
     TotalConsumptionList,
 )
 from electricitymap.contrib.lib.models.events import ProductionMix
+from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
@@ -106,7 +107,7 @@ def fetch_data(
         )
 
 
-def fix_solar_production(dt: datetime, value: pd.Series) -> int:
+def fix_solar_production(dt: datetime, value: float) -> int:
     """sets solar production to 0 during the night as there is only solar PV in UY"""
     if (dt.hour <= 5 or dt.hour >= 20) and value != 0:
         return 0
@@ -150,7 +151,7 @@ def fetch_production(
                     correct_negative_with_zero=True,
                 )
         production_list.append(
-            zoneKey=zone_key,
+            zoneKey=ZoneKey(zone_key),
             datetime=dt.to_pydatetime().replace(tzinfo=UY_TZ),
             source=UY_SOURCE,
             production=production_mix,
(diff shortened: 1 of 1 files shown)


parsers/UY.py
```diff
@@ -99,12 +106,12 @@ def fetch_data(
...
 
-def fix_solar_production(dt: datetime, row: pd.Series) -> int:
+def fix_solar_production(dt: datetime, value: pd.Series) -> int:
     """sets solar production to 0 during the night as there is only solar PV in UY"""
-    if (dt.hour <= 5 or dt.hour >= 20) and row.get("value") != 0:
+    if (dt.hour <= 5 or dt.hour >= 20) and value != 0:
         return 0
     else:
-        return round(row.get("value"), 3)
+        return round(value, 3)
```
VIKTORVAV99, line 114: This has a type error as pd.Series can't be passed to round.

parsers/UY.py
```diff
@@ -99,12 +106,12 @@ def fetch_data(
...
 
 
 @refetch_frequency(timedelta(days=1))
```
VIKTORVAV99, line 117: Can't comment on the line below but the zoneKey should be changed to match the ZoneKey type not string.

FelixDQ (commented):
LGTM! Soon we will probably handle this in the outlier detection pipeline instead, but this makes sense for now!

parsers/UY.py
```diff
@@ -170,17 +181,17 @@ def fetch_consumption(
...
-            "consumption": round(consumption[dt]["consumption"], 3),
-            "source": "pronos.adme.com.uy",
-        }
-        all_data_points += [data_point]
-    return all_data_points
+    consumption = data[["consumption"]]
+
+    for dt, row in consumption.iterrows():
+        consumption_list.append(
+            zoneKey=zone_key,
```
VIKTORVAV99, line 188: Same here so we don't end up with type errors.

FelixDQ (approved):

VIKTORVAV99 (commented):
Some small comments but nothing major.

However I can't seem to run it and get this exception:
> parsers.lib.exceptions.ParserException: UY.py Parser: Impossible to fetch data url from ADME

Is it working for you?

pierresegonne:
> However I can't seem to run it and get this exception:

Same happened for me before / after the changes - I think the source is flaky right now. It works as expected if you provide a delayed timestamp (try yesterday)

parsers/UY.py
pierresegonne: Good catch

parsers/UY.py
pierresegonne: Good catch

parsers/UY.py
pierresegonne: Good catch

PR #6507, opened 2024-02-23
feat: refactor SG parser to use event classes

yujia21:
## Issue

## Description

This PR updates the SG parser to use event classes as per new parser guidelines. 

Note to reviewer: prior to the change, the SG parser returned a dictionary, but with the change it will return a list of one dictionary. Not sure if this is the ideal pattern for parsers returning just one Event, happy to do a change.

### Double check

### DIFF
--- a/parsers/SG.py
+++ b/parsers/SG.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import re
-from collections import defaultdict
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
 from typing import Any
@@ -68,7 +67,6 @@ def get_solar(session: Session, logger: Logger) -> float | None:
     Fetches a graphic showing estimated solar production data.
     Uses OCR (tesseract) to extract MW value.
     """
-
     solar_image = Image.open(session.get(SOLAR_URL, stream=True).raw)
 
     solar_mw = __detect_output_from_solar_image(solar_image, logger)
@@ -185,15 +183,14 @@ def fetch_production(
         gen_type["Label"]: parse_percent(gen_type["Value"]) for gen_type in mix_section
     }
 
-    generation_by_type = defaultdict(float)  # this dictionary will default keys to 0.0
     production_breakdowns = ProductionBreakdownList(logger)
-
+    mix = ProductionMix()
     for gen_type, gen_percent in gen_types.items():
         gen_mw = gen_percent * generation
         mapped_type = TYPE_MAPPINGS.get(gen_type)
 
         if mapped_type:
-            generation_by_type[TYPE_MAPPINGS[gen_type]] += gen_mw
+            mix.add_value(TYPE_MAPPINGS[gen_type], gen_mw)
 
         else:
             # unrecognized - log it, then add into unknown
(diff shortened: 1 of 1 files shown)


parsers/SG.py
```diff
@@ -197,22 +206,24 @@ def fetch_production(
             logger.warning(msg)
             generation_by_type["unknown"] += gen_mw
 
-    generation_by_type["solar"] = get_solar(requests_obj, logger)
+    generation_by_type["solar"] = 0#get_solar(requests_obj, logger)
```
VIKTORVAV99, line 209: What is going on here? We kind of need solar data...

parsers/SG.py
```diff
@@ -197,22 +206,24 @@ def fetch_production(
...
     generation_by_type.update({"nuclear": 0, "wind": 0, "hydro": 0})
-
-    return {
-        "datetime": sg_data_to_datetime(data),
-        "zoneKey": zone_key,
-        "production": generation_by_type,
-        "storage": {},  # there is no known electricity storage in Singapore
-        "source": "emcsg.com, ema.gov.sg",
-    }
+    mix = ProductionMix(**generation_by_type)
```
VIKTORVAV99, line 213: While functional the purpose of the Parser classes is to unify and remove unneeded code from the parsers. 

For example this piece of code on line 197:
```python
        if mapped_type:
            generation_by_type[TYPE_MAPPINGS[gen_type]] += gen_mw
```

Should be:
```python
        if mapped_type:
            mix.add_value([TYPE_MAPPINGS[gen_type]], gen_mw)
```
Where mix is the production mix.
This is to ensure we reduce the floating point errors we might introduce from doing math operations and to ensure that if one of the values are negative it's corrected before it's summed with the other values.

parsers/SG.py
```diff
@@ -180,6 +188,7 @@ def fetch_production(
     }
 
     generation_by_type = defaultdict(float)  # this dictionary will default keys to 0.0
```
VIKTORVAV99, line 190: We should get rid of this default dict.

VIKTORVAV99 (changes requested):
Looks like a good start!

Some things need to be changed though before we can merge this.

q--:
This pull request reminded me again that I needed to bring up that Singapore generation mapping for anything other than solar makes wrong assumptions, and that 'unknown' isn't actually renewable.

https://github.com/electricitymaps/electricitymaps-contrib/issues/6508

Apologies for not filing an issue sooner.

yujia21:
Updated with suggested fixes! 

Btw not sure why but the get_solar function is still throwing an error on my local:
```
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7f0d2db0ffb0>
```
but I see the data seems live in production.

VIKTORVAV99:
Code wise this looks good now, will take a look at the issue linked by q-- before approving thought to see if there is something we need to change in this PR.

VIKTORVAV99:
I'm going to merge this and then we can adress the issue in a follow up PR instead.

VIKTORVAV99 (approved):

PR #6510, opened 2024-02-25
feat: refactor TH parser to use event classes

yujia21:
## Issue

### Preview

This PR updates the TH parser to use event classes as per new parser guidelines.

Instead of calling fetch_production from fetch_consumption as in the original code, I called the underlying _fetch_data directly as I felt it was cleaner and more future proof.

### Double check

### DIFF
--- a/parsers/TH.py
+++ b/parsers/TH.py
@@ -2,11 +2,19 @@ import json
 import re
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
+from typing import Any
 from zoneinfo import ZoneInfo
 
 from bs4 import BeautifulSoup
 from requests import Session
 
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM!

Thanks for updating it!

VIKTORVAV99:
Just as a future reference, it looks like the price domain is returning 404 not found for me. But this PR does not change that so it should be addressed separately.

PPsyrius:
@VIKTORVAV99 It seems like MEA changed their website again - I'll fix its pricing calculation later over the weekends.

I don't know if it's just me but Thailand's data in general appears to be down altogether.

PR #6512, opened 2024-02-25
perf: Avoid unnecessary property access

VIKTORVAV99:
## Issue
Property access is somewhat slow in JS.

## Description

Since we are using the value `worldTopography.objects[k]` many times in this function it makes sense to assign it to a local variable so we don't have to keep doing unnecessarily peppery lookups in the full object tree.

It also helps with maintainability, readability, and a very small reduction in size for the generated code.

### Double check


VIKTORVAV99:
The main bottleneck in this function is still the merge and mesh functions of `topojson-client`, I'll see if I can provide upstream changes for that and if that fails look into alternative actions.

PR #6513, opened 2024-02-26
Update RS validation parameters

VIKTORVAV99:
## Issue

We are sometimes getting junk data from ENTSO-E, and the current validation is not sufficient enough.

Closes: https://github.com/electricitymaps/electricitymaps-contrib/issues/6468
Closes: [ELE-3864](https://linear.app/electricitymaps/issue/ELE-3864/update-validation-rules-in-the-entso-e-parser-for-rs)

## Description

This PR updated the validation to filter out more bad data.

I think the 800 MW is the lower limit we can use without discarding good data.

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -455,8 +455,12 @@ VALIDATIONS: dict[str, dict[str, Any]] = {
         "expected_range": (2000, 25000),
     },
     "RS": {
-        "required": ["coal"],
+        "required": ["biomass", "coal", "gas", "hydro", "unknown"],
         "expected_range": {
+            "coal": (
+                800,
+                7000,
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -455,8 +455,12 @@ class WindAndSolarProductionForecastTypes(Enum):
         "expected_range": (2000, 25000),
     },
     "RS": {
-        "required": ["coal"],
+        "required": ["biomass", "coal", "gas", "hydro", "unknown"],
         "expected_range": {
+            "coal": (
+                800,
```
VIKTORVAV99, line 461: I think 800 is a safe lower limit based on our historical data. Together with the other required modes it should cover most of the cases.

wobniarin (approved):
LGTM! thanks for that!

PR #6517, opened 2024-02-28
CL parser: add snapshot test and remove usage of arrow module

gianantoniopini:
## Issue

#6135 

## Description

This PR adds a snapshot test for the CL parser and removes usage of the `arrow` module from the CL parser.

### Double check


parsers/test/mocks/CL/response_historical_20240224.json
gianantoniopini, line 1: I have tried to minify this .json mock file: after minification the file size went from 58K to 37K. But I decided to revert that commit and keep the file unminified, primarily because of readability. The commit with the minified version is 20ba10. Happy to revert back to the minified version, if it is needed.

VIKTORVAV99: I think either is fine, the point of the snapshot tests is not really to see if they match the expected input data (I sure hope they do though) but rather to ensure no accidental regressions are made in future edits.

So readability is not really a concern IMO.

gianantoniopini: While working on the parser code, I have found it helpful to be able to see what the response from the external api actually looks like. So, at least for me, the .json mock file also served this purpose, as it contains a sample response from the external api.

parsers/CL.py
```diff
@@ -108,10 +108,10 @@ def production_processor_historical(raw_data):
         clean_datapoint = {}
         date, hour = datapoint["fecha"], datapoint["hora"]
         hour -= 1  # `hora` starts at 1
-        date = arrow.get(date, "YYYY-MM-DD", tzinfo="Chile/Continental").shift(
-            hours=hour
-        )
-        clean_datapoint["datetime"] = date.datetime
+        parsed_datetime = datetime.strptime(date, "%Y-%m-%d").replace(
+            tzinfo=TIMEZONE
+        ) + timedelta(hours=hour)
```
VIKTORVAV99, line 113:
```suggestion
        parsed_datetime = datetime.fromisoformat(date).replace(
            tzinfo=TIMEZONE
        ) + timedelta(hours=hour)
```

parsers/CL.py
```diff
@@ -174,9 +174,12 @@ def fetch_production(
...
 
-    arr_target_datetime = arrow.get(target_datetime)
-    start = arr_target_datetime.shift(days=-1).format("YYYY-MM-DD")
-    end = arr_target_datetime.format("YYYY-MM-DD")
+    if target_datetime:
+        target_datetime_aware = target_datetime.replace(tzinfo=TIMEZONE)
+    else:
+        target_datetime_aware = datetime.now(tz=TIMEZONE)
+    start = (target_datetime_aware + timedelta(days=-1)).strftime("%Y-%m-%d")
+    end = target_datetime_aware.strftime("%Y-%m-%d")
```
VIKTORVAV99, line 182:
```suggestion
    start = (target_datetime_aware + timedelta(days=-1)).date().isoformat()
    end = target_datetime_aware.date().isoformat()
```

VIKTORVAV99 (commented):
Nice work! Looks like it works as expected!

I'm not a huge fan of strptime and strftime so I left suggestions that use the iso builtin versions instead.

gianantoniopini:
Just applied your suggestions related to the use of `datetime` isoformat methods. Thank you!

VIKTORVAV99 (approved):
LGTM! Thanks for doing this and special thanks for adding tests!

PR #6518, opened 2024-02-28
Capacity update AE and OM

florianscheidl:
## Issue

Missing recent capacity data for AE and OM, which are required for new estimation model.

## Description

Using Climatescope data, I updated the respective capacity values.

### Preview

N/A

### Double check


VIKTORVAV99 (approved):
LGTM!

Will be nice with all these historical capacities!

PR #6521, opened 2024-02-29
Fix: make sure that text doesn't start with estimation-card.

silkeholmebonnen:
## Issue

[ELE-3897](https://linear.app/electricitymaps/issue/ELE-3897/qa-estimation-tag-on-app)

## Description

This PR makes sure that estimation card never shows any text that starts with the 'estimation-card.', and instead use the default text.

### Preview

Before:
image
After:
image

### Double check

### DIFF
--- a/web/src/features/panels/zone/EstimationCard.tsx
+++ b/web/src/features/panels/zone/EstimationCard.tsx
@@ -75,7 +75,9 @@ function getEstimationTranslation(
       : t(`estimation-card.${estimationMethod?.toLowerCase()}.${field}`);
 
   const genericTranslation = t(`estimation-card.estimated_generic_method.${field}`);
-  return exactTranslation.length > 0 ? exactTranslation : genericTranslation;
+  return exactTranslation.length > 0 && !exactTranslation.startsWith('estimation-card.')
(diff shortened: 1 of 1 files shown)


web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -75,7 +75,9 @@ function getEstimationTranslation(
       : t(`estimation-card.${estimationMethod?.toLowerCase()}.${field}`);
 
   const genericTranslation = t(`estimation-card.estimated_generic_method.${field}`);
-  return exactTranslation.length > 0 ? exactTranslation : genericTranslation;
+  return exactTranslation.length > 0 && !exactTranslation.startsWith('estimation-card.')
+    ? exactTranslation
+    : genericTranslation;
```
VIKTORVAV99, line 80: Do we even need the length call anymore? I assume it will always be above 0 now as it returns the identifier (for some reason).

silkeholmebonnen: True

VIKTORVAV99 (approved):
LGTM!

PR #6533, opened 2024-03-01
parser validation - add function to filter out events with total production of 0 MW

mathilde-daugy:
## Issue
The issue has been seen for PE and some US zones. The parser returns only 0 for all modes and these data points are not flagged by the outlier detection pipeline
image

## Description

- [ ] Update `ProductionBreakdownList` to add function `filter_only_zero_production` that filters out events with a total production of 0 MW. 
- [ ] Add filter to `PE.py`
- [ ] Add filter to `EAI.py`
- [ ] Add test 

### Preview

NA

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -279,6 +279,35 @@ def update_production_breakdowns(
...
+        breakdowns: "ProductionBreakdownList",
+    ) -> "ProductionBreakdownList":
+        """A method to filter out production breakdowns with a total production of 0 MW."""
+        events = ProductionBreakdownList(breakdowns.logger)
+        for event in breakdowns.events:
+            total_production = sum(
+                {
+                    k: v for k, v in event.production.dict().items() if v is not None
+                }.values()
+            )
```
VIKTORVAV99, line 293: Is it not possible to call `all()` directly and assert if they are `None or 0` instead?

I think that would be simpler (and faster as it can exit early) if possible but I'm not sure if we have set up the internal methods in the production mix for that.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -279,6 +279,35 @@ def update_production_breakdowns(
 
         return production_breakdowns
 
+    @staticmethod
+    def filter_only_zero_production(
```
VIKTORVAV99, line 283: Do we really want this as an extra function?
I would expect this to be the default behaviour of the ProductionBreakdownList with the possibility to override it for zones that can have 0 production like AX and US-FLA-HST.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -279,6 +279,35 @@ def update_production_breakdowns(
...
+        breakdowns: "ProductionBreakdownList",
+    ) -> "ProductionBreakdownList":
+        """A method to filter out production breakdowns with a total production of 0 MW."""
+        events = ProductionBreakdownList(breakdowns.logger)
+        for event in breakdowns.events:
```
VIKTORVAV99, line 288: Maybe change up the naming here, I would expect `event` to be a child of `events` not `breakdown.events` at first glance.

electricitymap/contrib/lib/models/event_lists.py
mathilde-daugy: This is meant as a temporary fix that would then be handled by the outlier detection pipeline

VIKTORVAV99: Would it really need to be handled by the outlier detection though?
We already filter out events that are all None so I would just extend on that with a boolean flag to turn it off.

mathilde-daugy: The whole point of the outlier detection is to ingest all data points and then flag them rather than filtering them before the database

VIKTORVAV99: Does that mean we want to start returning all the null values as well?

I was under the assumption that we were going to filter out all clear errors as close to the parsers as possible and use the outlier detection to handle suspicious events that might not be clear errors.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -279,6 +279,35 @@ def update_production_breakdowns(
...
+                )
+                continue
+            if valid:
+                events.append(
+                    zoneKey=event.zoneKey,
+                    datetime=event.datetime,
+                    production=event.production,
+                    storage=event.storage,
+                    source=event.source,
+                )
```
VIKTORVAV99, line 308:
```suggestion
            if event.production is not None and all(
                (v[1] == 0 or v[1] is None) for v in event.production
            ):
                events.logger.warning(
                    f"Discarded production event for {event.zoneKey} at {event.datetime} because all production values are 0."
                )
                continue
            events.append(
                zoneKey=event.zoneKey,
                datetime=event.datetime,
                production=event.production,
                storage=event.storage,
                source=event.source,
                )
```
This should be a bit simpler and easier to understand

electricitymap/contrib/lib/models/event_lists.py
unitrium: @VIKTORVAV99, the long term vision is that we do not do any kind of filtering at the parser level. We rather want to ingest everything we can because we want to be able to report on why we dropped certain parser events, also analyze the frequency at which they appear and other things which we are not able to do if it's filtered at parser level.

Right now we do not have a sufficiently mature infrastructure to handle all edge cases so we do this filtering at this level but all of this should be handled in the future outside of the parser hence why a central function which we can remove later on is probably the best place. Also better to not introduce too much complexity.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -279,6 +279,35 @@ def update_production_breakdowns(
 
         return production_breakdowns
 
+    @staticmethod
+    def filter_only_zero_production(
+        breakdowns: "ProductionBreakdownList",
+    ) -> "ProductionBreakdownList":
+        """A method to filter out production breakdowns with a total production of 0 MW."""
```
unitrium, line 286:
```suggestion
        """
        TODO: Remove once the internal outlier detection is able to handle this.
        A method to filter out production breakdowns with a total production of 0 MW."""
```

electricitymap/contrib/lib/models/event_lists.py
unitrium: I agree with Viktor's suggestion, I think it's more efficient.

unitrium (approved):
Please implement Viktor's suggestion before merging.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: > why a central function which we can remove later on is probably the best place.
This is what I mean, we already do this filtering in one place with the validator decorators. Would it not be better to extend that in some way so we can remove them in one place later instead of every parser we decide to use this function?

PR #6536, opened 2024-03-01
fix(estimation labels & cards): Small fixes for Null data due to filtering

VIKTORVAV99:
## Issue
When we are actively filtering data due to zones being estimated for prolonged durations we are sending Null data instead of undefined.

We also had a small bug where it the card type would determine if the estimated badge was shown instead of the estimation method.

## Description
This PR relaxes the check to just see if the Estimation method values to true instead of not undefined, and uses the estimation method to check if we should show the estimation badge for the bar breakdown chart instead.

### Preview
Before:
image
After:
image

### Double check


silkeholmebonnen (approved):
LGTM

PR #6537, opened 2024-03-02
feat(ExchangeList): Add Exchange._update() and ExchangeList.update_exchanges() methods

VIKTORVAV99:
## Issue

We want to be able to collect the most up to date exchange forecasts from ENTSO-E (an possibly others).

## Description
I went with a slightly different approach in this PR by adding dunder functions to the class. It results in slightly cleaner code (IMO) but could be harder to understand.

I'll let you guys decide if this is preferable over the method in https://github.com/electricitymaps/electricitymaps-contrib/pull/6442 otherwise I'll change it to use the same approach as there.

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -142,6 +142,15 @@ def _get_unique_source_type(cls, events: pd.DataFrame) -> EventSourceType:
...
     events: list[Exchange]
 
+    def __contains__(self, datetime) -> bool:
+        return any(event.datetime == datetime for event in self.events)
+
+    def __getitem__(self, datetime) -> Exchange:
+        return next(event for event in self.events if event.datetime == datetime)
+
+    def __setitem__(self, datetime, event: Exchange):
+        self.events[self.events.index(self[datetime])] = event
```
VIKTORVAV99, line 152: These can potentially be moved up to the parent classes and be re-used between all the parser class lists if we want to.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -189,6 +198,32 @@ def merge_exchanges(
...
+                updated_event = Exchange.update(existing_event, new_event)
+                exchanges[new_event.datetime] = updated_event
+            else:
+                exchanges.append(
+                    new_event.zoneKey,
+                    new_event.datetime,
+                    new_event.source,
+                    new_event.netFlow,
+                    new_event.sourceType,
+                )
```
VIKTORVAV99, line 223: With the changes in the follow up PR it will look like this instead:
```python
        for new_event in new_exchanges:
            if new_event in exchanges:
                existing_event = exchanges[new_event]
                updated_event = Exchange.update(existing_event, new_event)
                exchanges[new_event] = updated_event
            else:
                exchanges._append_event(new_event)
```
Which could be further reduced by removing the intermediate variables if we want.

pierresegonne (commented):
I think it can be a good idea to use the magic methods, but at the condition that we ensure that we are consistent with other Implementations. 

Should we do it in this PR? Maybe not, it's a bit frustrating, but if we truly want to separate concerns, it would be better to just add the functionality in this PR and follow up with another one to transition to magic methods throughout. 

In this case, I'd be ok going forward with magic methods in this PR, but we need to be 100% sure that we'll follow-up afterwards to ensure consistency

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: append is not a magic method in it self and to avoid too much logic in the magic methods this made the most sense imo, but I have made follow up changes locally.

VIKTORVAV99:
> I think it can be a good idea to use the magic methods, but at the condition that we ensure that we are consistent with other Implementations.
> 
> Should we do it in this PR? Maybe not, it's a bit frustrating, but if we truly want to separate concerns, it would be better to just add the functionality in this PR and follow up with another one to transition to magic methods throughout.
> 
> In this case, I'd be ok going forward with magic methods in this PR, but we need to be 100% sure that we'll follow-up afterwards to ensure consistency

I'll push my local branch as a draft PR so you can evaluate how I have done in there, that moves the magic methods to the base classes and adds appropriate overrides in all the child functions for proper typing.

unitrium:
I started reviewing this but I won't have enough time to finish it today, if you don't mind waiting for me, I will give you feedback tomorrow.

electricitymap/contrib/lib/models/event_lists.py
unitrium: Those changes make sense to me, but I think they should be in the parent class rather.

electricitymap/contrib/lib/models/events.py
```diff
@@ -430,6 +430,33 @@ def create(
                 },
             )
 
+    @staticmethod
+    def update(event: "Exchange", new_event: "Exchange") -> "Exchange":
```
unitrium, line 434: Not sure I fully understand why this would be usef by someone if they already created an Exchange.
Maybe this should be an internal method of the list rather?
I'm not sure someone would understand what this is for.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -189,6 +198,32 @@ def merge_exchanges(
...
+        exchanges: "ExchangeList", new_exchanges: "ExchangeList", logger: Logger
+    ) -> "ExchangeList":
+        """Given a new batch of exchanges, update the existing ones."""
+        if len(new_exchanges) == 0:
+            return exchanges
+        elif len(exchanges) == 0:
+            return new_exchanges
+
+        for new_event in new_exchanges.events:
+            if new_event.datetime in exchanges:
```
unitrium, line 212: Hum, I'm thinking that the datetime indexing is not super explicit, since this is initially a list of events.
This should probably be explicited in the class description.

unitrium (changes requested):
Overall I think the change is good, I think we should just abstract it more and do it in the parent class.
Otherwise we have a very different behaviour for Exchanges and other types.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: That is what I do in the follow up PR, but I can do it directly in here as well and then update the other PR to just contain the logic changes I want to make.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: That is what I do in the follow up PR, but I can do it directly in here as well and then update the other PR to just contain the logic changes I want to make.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: In my follow up PR I am changing this to be a dict internally to streamline a lot of the logic without changing how it works externally.

It drastically simplifies the logic of all the magic functions.

That PR is still a draft though as it hasn't beef finalized or fully tested yet.

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: The reason this is in here is because it vastly simplifies the logic in the lists and can be abstracted away (could even be possible to do in the top level event_list if all the events types have a update function).

But I'll prepend it weith `_`to mark it as private.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: Added a it to the top level EventList class as there is where the magic functions are implemented now. Let me know if you want to make it even more clear.

unitrium:
Will review a bit later during the week.

pierresegonne:
I'll let @unitrium do the final review

unitrium (approved):
Looks good to me! Thanks for adding this!

PR #6543, opened 2024-03-04
remove the use of badge

silkeholmebonnen:
## Issue

[ELE-3909](https://linear.app/electricitymaps/issue/ELE-3909/replace-the-use-of-badge-with-estimationbadge-and-outagebadge)

## Description

The goal of this PR is to reduce the amount of duplicate code by using the `OutageBadge` and `EstimationBadge` instead of `Badge` whenever possible.

I tried to do some rework in `EstimationCard` but decided that it was okay the way it is.

### Double check


VIKTORVAV99 (approved):
LGTM! 🥳

BTW this shrunk our bundle size by approximately 2kb 🙌🏻

tonypls (approved):
LGTM 🎉

PR #6544, opened 2024-03-04
(#6011) Refactor AX parser to use electricitymaps event classes

AlisCode:
## Issue

#6011

## Description

This PR upgrades the parser for Aaland (ZoneKey: AX) to use electricitymaps' event classes, as described in the issue.

### Double check

### DIFF
--- a/parsers/AX.py
+++ b/parsers/AX.py
@@ -97,7 +97,7 @@ def formatted_data(
     consumption = TotalConsumptionList(logger)
     for data in data_list:
         corrected_date = date - timedelta(minutes=15 * data_list.index(data))
-        if data_type == "production":
+        if data_type == "production" and zone_key is not None:
             production_mix = ProductionMix(wind=data["wind"], oil=data["fossil"])
             production_breakdowns.append(
                 datetime=corrected_date,
@@ -105,7 +105,7 @@ def formatted_data(
                 source=SOURCE,
                 zoneKey=zone_key,
             )
-        elif data_type == "consumption":
+        elif data_type == "consumption" and zone_key is not None:
             consumption.append(
                 datetime=corrected_date,
                 consumption=data["consumption"],
--- a/parsers/AX.py
+++ b/parsers/AX.py
@@ -6,6 +6,14 @@ from zoneinfo import ZoneInfo
 from bs4 import BeautifulSoup
 from requests import Response, Session
 
+from electricitymap.contrib.lib.models.event_lists import (
+    ExchangeList,
+    ProductionBreakdownList,
+    TotalConsumptionList,
+)
+from electricitymap.contrib.lib.models.events import ProductionMix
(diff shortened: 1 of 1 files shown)


parsers/AX.py
```diff
@@ -84,48 +92,40 @@ def formated_data(
     )
     if date > date_time:
         date = date - timedelta(days=1)
-    return_list = []
+    exchanges = ExchangeList(logger)
+    production_breakdowns = ProductionBreakdownList(logger)
+    consumption = TotalConsumptionList(logger)
     for data in data_list:
         corrected_date = date - timedelta(minutes=15 * data_list.index(data))
         if data_type == "production":
```
VIKTORVAV99, line 100:
```suggestion
        if data_type == "production" and zone_key is not None:
```

parsers/AX.py
```diff
@@ -84,48 +92,40 @@ def formated_data(
...
-                    "source": SOURCE,
-                }
+            production_mix = ProductionMix(wind=data["wind"], oil=data["fossil"])
+            production_breakdowns.append(
+                datetime=corrected_date,
+                production=production_mix,
+                source=SOURCE,
+                zoneKey=zone_key,
             )
         elif data_type == "consumption":
```
VIKTORVAV99, line 108:
```suggestion
        elif data_type == "consumption" and zone_key is not None:
```

VIKTORVAV99 (commented):
Thanks for doing this!
The changes look good functionality wise but I left suggestions to remove two type errors.

AlisCode:
Hi @VIKTORVAV99 , thank you for reviewing :)

I fixed the type errors as you suggest (even though I think it would have been best to just make the argument `zone_key` of `formatted_data` not `None` ? I looked at the call sites and `zone_key` is always defined). 

Out of curiosity for next time I want to contribute, what are you guys' tooling to typecheck Python stuff ?

VIKTORVAV99:
> Hi @VIKTORVAV99 , thank you for reviewing :)
> 
> 
> 
> I fixed the type errors as you suggested (even though I think it would have been best to just make the argument `zone_key` of `formatted_data` a `ZoneKey` as opposed to `ZoneKey | None` ? I looked at the call sites and `zone_key` is always defined). 
> 
> 
> 
> Out of curiosity for next time I want to contribute, what are you guys' tooling to typecheck Python stuff ?

It would not be defined for the exchange data where it's assigned a none value.

In the CI we use RUFF to run static analysis and locally I use the pylance plugin for VSCode that implements the pyright type checker. We plan on running this in the CI as well when we have removed the existing type errors.

VIKTORVAV99 (approved):
LGTM!

Thanks for updating it!

AlisCode:
> It would not be defined for the exchange data where it's assigned a none value.

Well that one slipped past, and is indeed quite obvious. That's why you need tooling :) 

Thank you for the merge and the tooling recommendation!

PR #6546, opened 2024-03-07
chore: fix non f formatted strings

pierresegonne:
## Issue

Update contrib is blocked by this

### Double check


VIKTORVAV99 (approved):
Do you know which rule it was?
We should probably add that to the contrib config to avoid it in the future as well.

pierresegonne:
@VIKTORVAV99 UP032

PR #6547, opened 2024-03-07
Silkebonnen/ele 3900 implement production source legend component

silkeholmebonnen:
## Issue

[ELE-3900](https://linear.app/electricitymaps/issue/ELE-3900/implement-production-source-legend-component)

## Description

The goal of this PR is to add the production source legends to the bar breakdown chart and it's tooltip.

### Preview

image

### Double check


web/src/features/charts/bar-breakdown/constants.ts
```diff
@@ -7,3 +9,18 @@ export const RECT_OPACITY = 0.8;
...
+  'hydro storage': '8',
+  'battery storage': '6',
+  biomass: '9',
+  geothermal: '8',
+  nuclear: '8',
+  gas: '9',
+  coal: '9',
+  oil: '9',
+  unknown: '8',
+};
```
madsnedergaard, line 26: Wouldn't it make more sense to get @Alportan to make SVGs in the same height so we don't need all the code around the different heights? :)

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <image
+        transform={`translate(${LABEL_MAX_WIDTH - 1.5 * PADDING_Y - 10}, ${getIconPadding(
+          electricityType
+        )})`}
+        width={14}
+        height={iconHeight[electricityType as ElectricityModeType]}
+        xlinkHref={`/images/production-source/${electricityType}.svg`}
+      />
+    </g>
+  );
```
madsnedergaard, line 30: This is mixing svg elements and the image, I think it would be better to just have a `div` instead of `rect` if possible :)

web/src/features/charts/bar-breakdown/constants.ts
VIKTORVAV99: I don't even think they need to be the same size as you can override that in the SVG directly with the height and width parameters.

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
VIKTORVAV99: Following up on this, I think it would make even more sense to import the svgs directly using a dynamic import and style it instead of an image. By importing the SVG vite should be able to do some bundle magic and include it in our bundle to avoid HTTP calls.

web/public/images/production-source/oil.svg
madsnedergaard, line 1: Not a huge fan of the oil icon @Alportan, it's not very eligible when so small. How about a drop and potentially also something else for gas so they don't seem too similar? Oil could also be a barrel :)

Screenshot 2024-03-07 at 19 49 35

madsnedergaard:
Would be cool to also use it in the tooltips if you have time 😄 

Screenshot 2024-03-07 at 19 47 29

web/public/images/production-source/oil.svg
Alportan: Makes sense, let's have a look at it tomorrow 👍

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
madsnedergaard: Aaah wait, now I understand why you're using the `rect` though, didn't realise it was all within a big SVG! Then it makes good sense...

Agree on Viktor's comment though :)

VIKTORVAV99: I guess it would also be possible to make them into components as svg code is also valid html if that is easier.

web/src/features/charts/bar-breakdown/constants.ts
Alportan: I'll create an SVG with the same proportions (8X8 px) and play with the negative space inside the frame itself.

VIKTORVAV99:
Looking good!

I'll do a full review soon but just wanted to float something first.
@Alportan have we considered moving the flags to the right of the exchanges here so it's all aligned?
image
Maybe even make them a square? I do think we have that option directly from the library.

VIKTORVAV99:
> @madsnedergaard Hovering on the icon still doesn't work, but idk how to fix it.

I think the issue is that we are using translate to move the icon to the right.
If we instead moved both the icons and the flags I think we could solve it (Alex said it was fine if we did).

madsnedergaard:
> > @madsnedergaard Hovering on the icon still doesn't work, but idk how to fix it.
> 
> I think the issue is that we are using translate to move the icon to the right. If we instead moved both the icons and the flags I think we could solve it (Alex said it was fine if we did).

I just pushed a fix for it @silkeholmebonnen - I don't know why it has to have a `pointer-event-none` class, but it works ¯\\\_(ツ)\_/¯

VIKTORVAV99:
> > > @madsnedergaard Hovering on the icon still doesn't work, but idk how to fix it.
> > 
> > 
> > I think the issue is that we are using translate to move the icon to the right. If we instead moved both the icons and the flags I think we could solve it (Alex said it was fine if we did).
> 
> I just pushed a commit that fixes it @silkeholmebonnen - I don't know why it has to have a `pointer-event-none` class, but it works ¯\_(ツ)_/¯

The whole SVG inside SVG is also messing with styles. I was thinking it would be nice to add the backgrounds directly to the svgs but then they disappear when placed inside the other SVG element...

VIKTORVAV99 (approved):
LGTM!

If you want to move the flags as well feel free to do so but as we mentioned in Figma it's up to you to decide 🙂.

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
```diff
@@ -0,0 +1,30 @@
...
+        rx={2}
+      />
+      <g
+        transform={`translate(${LABEL_MAX_WIDTH - 1.5 * PADDING_Y - 7}, 3)`}
+        width={14}
+        height={8}
+      >
+        <ProductionSourceIcon source={electricityType} />
+      </g>
+    </g>
```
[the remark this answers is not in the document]
Alportan: Maybe we can still wait with updating the flags after all, as I really can't anticipate the direction of the placement of the icons in general. I'll explore a bit more and come back to you, but ideally it would make more sense to maybe left align everything and having the icons leading instead of trailing (this will also make it more consistent with the tooltips and the general direction). Hope that makes sense! 🙌

PR #6550, opened 2024-03-10
refactor: use event classes in US-MISO parser

yujia21:
## Issue

## Description

Use event classes in US-MISO parser. I assumed the corresponding ZoneKey is US-MIDW-MISO but if that's not the right one I'll update it!

### Double check

### DIFF
--- a/parsers/US_MISO.py
+++ b/parsers/US_MISO.py
@@ -4,10 +4,19 @@
 
 from datetime import datetime
 from logging import Logger, getLogger
+from typing import Any
 
 from dateutil import parser, tz
 from requests import Session
 
+from electricitymap.contrib.config import ZoneKey
+from electricitymap.contrib.lib.models.event_lists import (
+    ProductionBreakdownList,
+)
+from electricitymap.contrib.lib.models.events import EventSourceType, ProductionMix
+
+SOURCE = "misoenergy.org"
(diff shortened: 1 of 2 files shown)


parsers/US_MISO.py
```diff
@@ -76,40 +84,39 @@ def data_processer(json_data, logger: Logger):
     tzinfos = {"EST": tz.gettz("America/New_York")}
     dt = parser.parse(time_data, tzinfos=tzinfos)
 
-    return dt, production
+    return dt, mix
 
 
 def fetch_production(
-    zone_key: str = "US-MISO",
+    zone_key: ZoneKey = ZoneKey(ZONE),
```
VIKTORVAV99, line 91: Just a small note, you don't have to add a default value here. We usually do it for single zone parsers because it's easy to understand but it's not required by any means.

VIKTORVAV99 (changes requested):
Looks like there are some failing tests that need changing but otherwise I don't have much to note, looks good code wise!

yujia21:
> Looks like there are some failing tests that need changing but otherwise I don't have much to note, looks good code wise!

Haha I noticed it at the same time as you were reviewing - added a commit for that!

VIKTORVAV99 (approved):
LGTM!

PR #6553, opened 2024-03-11
refactor: use event classes in US-SPP parser

yujia21:
## Issue

## Description

Use event classes in US-SPP parser

### Double check

### DIFF
--- a/parsers/US_SPP.py
+++ b/parsers/US_SPP.py
@@ -5,14 +5,26 @@
 from datetime import datetime, timedelta, timezone
 from io import StringIO
 from logging import Logger, getLogger
+from typing import Any
 
 import pandas as pd
 from dateutil import parser
 from requests import Session
 
+from electricitymap.contrib.lib.models.event_lists import (
+    ExchangeList,
+    ProductionBreakdownList,
+    TotalConsumptionList,
+)
+from electricitymap.contrib.lib.models.events import (
+    EventSourceType,
+    ProductionMix,
+)
+from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.validation import validate_exchange
 
+SOURCE = "spp.org"
 US_PROXY = "https://us-ca-proxy-jfnx5klx2a-uw.a.run.app"
 HOST_PARAMETER = "host=https://marketplace.spp.org"
 
@@ -34,6 +46,8 @@ MAPPING = {
     "Diesel Fuel Oil": "oil",
     "Waste Disposal Services": "biomass",
     "Coal": "coal",
+    "Other": "unknown",
+    "Waste Heat": "unknown",
 }
 
 TIE_MAPPING = {"US-MISO->US-SPP": ["AMRN", "DPC", "GRE", "MDU", "MEC", "NSP", "OTP"]}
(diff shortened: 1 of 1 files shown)


parsers/US_SPP.py
```diff
@@ -296,37 +301,32 @@ def fetch_wind_solar_forecasts(
...
             )
             continue
 
-        datapoint = {
-            "datetime": dt,
-            "production": production,
-            "zoneKey": zone_key,
-            "source": "spp.org",
-        }
+        mix = ProductionMix(solar=solar, wind=wind)
```
VIKTORVAV99, line 310: I noticed the source can return the following:
```python
{'correctedModes': [], 'datetime': datetime.datetime(2024, 3, 19, 7, 0, tzinfo=datetime.timezone.utc), 'production': {'solar': nan, 'wind': nan}, 'source': 'spp.org', 'sourceType': <EventSourceType.forecasted: 'forecasted'>, 'storage': {}, 'zoneKey': 'US-CENT-SWPP'}
```
I'll double check where we want to handle this with the rest of the team and see if we want to handle this in the parser or more likley in the parser classes directly. I'll put this on hold until then but so far it looks good. Should have an answer by tomorrow on how we want to handle it.

VIKTORVAV99: I have created https://github.com/electricitymaps/electricitymaps-contrib/pull/6558 as a fix, will do a final review of this after that is merged.

parsers/US_SPP.py
```diff
@@ -380,50 +380,56 @@ def format_exchange_data(
...
-        data_point = {
-            "sortedZoneKeys": sorted_zone_keys,
-            "netFlow": round(data_dt.values[0], 4),
-            "datetime": dt.to_pydatetime(),
-            "source": "spp.org",
-        }
-        all_data_points.append(data_point)
-    validated_data_points = [x for x in all_data_points if validate_exchange(x, logger)]
+        exchange_list.append(
+            zoneKey=sorted_zone_keys,
```
VIKTORVAV99, line 387: `sorted_zone_keys` need to be wrapped with a ZoneKey instance for the type to match.

parsers/US_SPP.py
```diff
@@ -341,12 +341,12 @@ def fetch_live_exchange(
...
-    zone_key2: str,
+    zone_key1: ZoneKey,
+    zone_key2: ZoneKey,
     session: Session | None = None,
     target_datetime: datetime | None = None,
     logger: Logger = getLogger(__name__),
-) -> list:
+) -> list[dict[str, Any]]:
     filename = target_datetime.strftime("TieFlows_%b%Y.csv")
     file_url = f"{US_PROXY}/file-browser-api/download/historical-tie-flow?{HOST_PARAMETER}&path={filename}"
```
VIKTORVAV99, line 351: Not really part of your changes but I donät think the target datetime should be None here, type wise. Or it should be checked that it's not none before the filename is generated as `.strftime` is not available on `None`.

yujia21: Fixed both! Sorry it's been a while!

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for doing this and sorry for the delay in the review!

PR #6555, opened 2024-03-11
Improvements to mobile app deployment workflow

madsnedergaard:
## Description

This makes it easier to release new app updates.
You can now run just one command for each platform and be sure that it has built and used the latest version of the app :)

I also bumped Fastlane version to latest.


mobileapp/fastlane/Fastfile
```diff
@@ -6,6 +6,14 @@
 # Uncomment the line if you want fastlane to automatically update itself
 # update_fastlane
 
+private_lane :build_web_app do
+  ensure_env_vars(
+  env_vars: ['SENTRY_AUTH_TOKEN', 'VITE_PUBLIC_ELECTRICITYMAP_PUBLIC_TOKEN']
+)
```
madsnedergaard, line 12: This requires having these variables in your `.env.default` file

VIKTORVAV99:
How much of a problem would it be to run this as a part of our CI/CD? Not for all deployments but maybe a scheduled release once per month?

tonypls:
> How much of a problem would it be to run this as a part of our CI/CD? Not for all deployments but maybe a scheduled release once per month?

I think it needs you to authenticate with the developer consoles with a developer account which would make it kinda hard

PR #6558, opened 2024-03-13
fix: treat nan values as none in parser events

VIKTORVAV99:
## Issue

NaN values where not filtered out properly.

## Description

This PR changes the validation to treat NaN values more or less the same as we treat None values.

### Double check

- [x] I have added tests for my changes.


electricitymap/contrib/lib/models/events.py
```diff
@@ -40,9 +41,11 @@ def add_value(
         """
         existing_value: float | None = getattr(self, mode)
         if existing_value is not None:
-            value = 0 if value is None else value
+            value = 0 if value is None or math.isnan(value) else value
             self.__setattr__(mode, existing_value + value)
         else:
+            if value is not None and math.isnan(value):
+                value = None
```
VIKTORVAV99, line 48: Bit unsure if we want to do this but was the simplest way to ensure it was functioning as expected. Would be happy to modify it though.

electricitymap/contrib/lib/models/events.py
```diff
@@ -40,9 +41,11 @@ def add_value(
         """
         existing_value: float | None = getattr(self, mode)
         if existing_value is not None:
-            value = 0 if value is None else value
+            value = 0 if value is None or math.isnan(value) else value
             self.__setattr__(mode, existing_value + value)
         else:
+            if value is not None and math.isnan(value):
```
unitrium, line 47: Does this work with numpy nans as well ?

unitrium (approved):
Looks pretty straightforward to me, please just implement my suggestion about readability before and make sure it works wit np.nan as well as it's a commonly used library but I would hope it does.

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: It does and I added tests for it just to make sure.

PR #6559, opened 2024-03-14
Issue 6135 Remove Arrow Usage from CA_AB

nboswell216:
## Description
Removes usage of arrow from CA_AB parser, all datetimes are still timezone aware

### Double check


parsers/CA_AB.py
```diff
@@ -81,7 +81,9 @@ def fetch_price(
         if row[1] != "-":
             prices.append(
                 zoneKey=zone_key,
-                datetime=arrow.get(row[0], "MM/DD/YYYY HH", tzinfo=TIMEZONE).datetime,
+                datetime=datetime.strptime(row[0], "%m/%d/%Y %H").replace(
+                    tzinfo=TIMEZONE
+                ),
```
VIKTORVAV99, line 86: This fails with:
```python
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: 4
```
It will needs to be addressed before we can merge it.

VIKTORVAV99 (changes requested):
Thanks for opening the PR, there is a small error that needs to be fixed but otherwise things looks good.

parsers/CA_AB.py
nboswell216: Fixed ValueError by converting %H to 00 index.

parsers/CA_AB.py
```diff
@@ -79,9 +79,12 @@ def fetch_price(
     prices = PriceList(logger)
     for row in csv.reader(response.text.split("\r\n\r\n")[2].splitlines()[1:]):
         if row[1] != "-":
+            date, hour = row[0].split()
             prices.append(
                 zoneKey=zone_key,
-                datetime=arrow.get(row[0], "MM/DD/YYYY HH", tzinfo=TIMEZONE).datetime,
+                datetime=datetime.strptime(
+                    f"{date} {int(hour) - 1}", "%m/%d/%Y %H"
+                ).replace(tzinfo=TIMEZONE),
```
VIKTORVAV99, line 87: The date no longer matches the old version and is off by 1 hour, I'll take a deeper look at it later today but I think you need to add back one hour using `timedelta()` after it has been converted to a datetime.

nboswell216: I see that now. My previous test must have been in a time window that didn't show the >2h difference. I also looked back at PR #6179 and it didn't seem clear if timedelta would be need or not - it looks like it is. I have corrected it with timedelta.

VIKTORVAV99 (approved):
LGTM!

Thanks for fixing it!

PR #6563, opened 2024-03-18
Implement button component

silkeholmebonnen:
## Issue

[ELE-3903](https://linear.app/electricitymaps/issue/ELE-3903/implement-button-component)

## Description

The goal of this PR is to implement the new Button component and use it in the places where we were previously using the old button component + the link "See our commercial API offerings" in the left panel (see preview). This is a part of the project 'Make it more clear what the data sources are', so later the button component will also be used for this.

### Preview

image

#### On mobile
image
image
image

### Double check


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -89,6 +91,14 @@ export default function ZoneDetails(): JSX.Element {
...
           <BarBreakdownChart hasEstimationPill={hasEstimationPill} />
           <Divider />
+          <Button
+            size="lg"
+            type="link"
+            icon={<MdOutlineCloudDownload size={20} />}
+            href="https://www.electricitymaps.com/get-our-data?"
+          >
+            {'See our commercial API offerings'}
+          </Button>
```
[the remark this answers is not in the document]
silkeholmebonnen: He changed it so the green button is only in the bottom of the page and this link is in the top. I'll change it to use a translation string later:)
image

web/src/features/map-controls/LanguageSelector.tsx
```diff
@@ -24,10 +25,14 @@ export function LanguageSelector({ isMobile }: { isMobile?: boolean }) {
...
-            <HiLanguage size={21} />
-            {t('tooltips.selectLanguage')}
-          </div>
+          <Button
+            size="lg"
+            type="secondary-elevated"
+            icon={<HiLanguage size={20} />}
+            backgroundClasses="min-w-[330px]"
+          >
+            {t('aria.label.selectLanguage')}
```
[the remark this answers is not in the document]
silkeholmebonnen: Nope, that's a mistake

web/src/components/Button.tsx
```diff
@@ -1,54 +1,150 @@
...
+        return 'bg-zinc-50 dark:bg-gray-800 border border-neutral-200 dark:border-gray-700';
+      }
+      return 'bg-brand-green';
+    }
+    case 'secondary': {
+      return 'border dark:border-gray-700 border-neutral-200 bg-inherit';
+    }
+    case 'secondary-elevated': {
+      return 'border dark:border-gray-700 border-neutral-200 bg-neutral-100 dark:bg-gray-800';
+    }
```
madsnedergaard, line 88: Will we actually be using the `secondary` type in the new design changes? Otherwise I'd prefer to rename `secondary-elevated` to just `secondary`, as I find the "elevated" name quite confusing. In my mind, elevation means it has shadows and is "lifted" above something else, but here it just has a background color 🤔 

On another note, I think it would be better to call it "[ghost](https://uxplanet.org/ghost-buttons-in-ux-design-4cf3717334f8)" or "[outline](https://m3.material.io/components/buttons/guidelines#3742b09f-c224-43e0-a83e-541bd29d0f05)" to follow common naming practices. This makes it a bit easier to know what it is, as "secondary" does not say anything about how it will look :)

cc @Alportan as this is also the case in Figma

web/src/features/map-controls/MapControls.tsx
```diff
@@ -36,13 +36,15 @@ function MobileMapControls() {
...
+        type="secondary-elevated"
         aria-label="open info modal"
         onClick={handleOpenInfoModal}
         icon={<HiOutlineInformationCircle size={21} />}
       />
       <Button
-        className="m-0 bg-white/80 p-3 backdrop-blur-sm dark:bg-gray-800/80"
+        size="lg"
+        type="secondary-elevated"
         aria-label="open settings modal"
```
madsnedergaard, line 48: Can we keep the reduced opacity and `backdrop-blur-sm` classes here for the nice glass-like visuals?

web/src/features/modals/InfoModal.tsx
```diff
@@ -35,38 +35,51 @@ export function InfoModalContent() {
       </div>
       <div>
         <Button
+          size="lg"
+          type="secondary-elevated"
+          backgroundClasses="my-2"
```
madsnedergaard, line 40: Instead of adding the margin on each button here, you could add a `space-y-2` class on the `<div>` above it :)

web/src/features/map-controls/MapControls.tsx
VIKTORVAV99: There should be a background-glass variant somewhere that is either done or a work in progress (in the Figma).

web/src/features/modals/InfoModal.tsx
```diff
@@ -35,38 +35,51 @@ export function InfoModalContent() {
...
           {t('info-modal.feedback-button')}
         </Button>
         <Button
-          background="linear-gradient(to right, #04275c 0%, #040e23 100%)"
-          textColor="#fff"
+          size="lg"
+          type="primary"
+          backgroundClasses="bg-gradient-to-r from-[#04275c] to-[#040e23] mb-2"
+          forgroundClasses="text-white dark:text-white"
+          focusOutlineColor="focus:outline-[#040e23]"
```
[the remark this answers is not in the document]
silkeholmebonnen: That's true. Actually I think it would be better to just add it as a part of the forgroundclasses

web/src/features/map-controls/MapControls.tsx
silkeholmebonnen: Yes:)

web/src/features/modals/SettingsModal.tsx
```diff
@@ -45,8 +45,10 @@ function WeatherToggleButton({
```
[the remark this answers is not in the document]
Alportan: Good points:

1. We had a chat with Silke about this. I would move away from using buttons to indicate toggle functionality. One hack mini project I had in mind is changing these buttons to toggle switches instead as they are more common UI components in this case.
2. Uhh, that's on me. I see that the grey somehow sneaked in while the original component uses a `background-base` token (#FFFFFF).
image

web/src/components/Button.tsx
Alportan: Hmmm, I went with `primary`, `secondary`, `tertiary`, `link` approach as they are more semantically showing hierarchy if that makes sense. 

I removed the `secondary-elevated` in the meantime for simplicity but maybe Silke implemented while I was also working on the components 🙈. That's on me, I owe you an apology @silkeholmebonnen .

Alportan: We should still do the naming convention survey and get a common understanding of common naming conventions. Ideally we should roll it out in our Contrib Slack as well to get a more complete overview. 🤔

madsnedergaard: Hmm yeah, I guess that makes sense - I'm fine with this for now, we can always bring the topic back up another time if it becomes too confusing :)

web/src/components/Button.tsx
```diff
@@ -1,54 +1,139 @@
...
-      href={href}
-      type={type}
-      onClick={onClick}
     >
-      {icon}
-      {children}
-    </As>
+      <As
+        className={twMerge(
+          `flex h-full w-full flex-row items-center justify-center rounded-full text-sm font-semibold focus:bg-inherit
```
silkeholmebonnen, line 41: I'm not sure if we want the `focus:bg-inherit`. I added it because I thought it looked better on mobile, when clicking the 'hide wind/solar layer' and 'color blind mode', so they do not look hovered when they have been pressed. But on web the hover disappear when the button has been clicked, and idk if we want that. What do you think is the best solution @madsnedergaard @VIKTORVAV99 ?

madsnedergaard: I think that's okay for now since we'll be moving to the switches later on anyway :)

madsnedergaard (approved):
Looks good to me now! 💪

web/src/components/Button.tsx
VIKTORVAV99: I agree with Mads, let's not let such small details hold us back when we know we will be changing it soon anyhow.

web/src/features/charts/ChartTitle.tsx
```diff
@@ -22,24 +21,9 @@ export function ChartTitle({
...
-      {hasLink && (
-        <div className="flex flex-row items-center pb-2 text-center text-[0.7rem]">
-          <HiOutlineArrowDownTray className="min-w-[12px]" size={12} />
-          <a
-            href="https://electricitymaps.com/?utm_source=app.electricitymaps.com&utm_medium=referral&utm_campaign=country_panel"
-            target="_blank"
-            rel="noreferrer"
-            className="pl-0.5 text-left text-[#4178ac] no-underline hover:underline dark:invert"
-          >
-            {t('country-history.Getdata')}
```
VIKTORVAV99, line 39: If this string is no longer used we should remove it from all the translation files so we don't forget about it and have a bloated translation file.

silkeholmebonnen:
@VIKTORVAV99 I requested your review again since I deleted the the `country-history.GetData` from the translation files and made some other small changes

VIKTORVAV99 (approved):
LGTM!

Thanks for removing the now bloated strings :)

PR #6570, opened 2024-03-21
CL parser: refactor to use event classes

gianantoniopini:
## Issue

#6011

## Description

This PR refactors the CL parser (CL-SEN zone) to use event classes, as described in the issue. The existing snapshot has been updated accordingly: numbers rounding, addition of `sourceType` and `correctedModes` fields.

Additionally, usage of the live API and live parser has been removed from the CL parser. This has been done in order to simplify the code and because the live parser was disabled anyway and is not needed any longer.

### Double check


parsers/CL.py
```diff
@@ -193,26 +209,39 @@ def fetch_production(
...
+
+
+@refetch_frequency(timedelta(days=1))
+def fetch_production(
+    zone_key: ZoneKey = ZoneKey("CL-SEN"),
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list[dict[str, Any]]:
+    return _fetch_production(zone_key, session, target_datetime, logger)
```
VIKTORVAV99, line 244: What is the point of doing this and not just use the old function like before?

parsers/CL.py
```diff
@@ -141,37 +145,49 @@ def production_processor_historical(raw_data):
...
-    session: Session | None = None,
-    target_datetime: datetime | None = None,
+def _fetch_production(
+    zone_key: ZoneKey,
+    session: Session | None,
+    target_datetime: datetime | None,
     logger: Logger = getLogger(__name__),
-):
-    if target_datetime is None and ENABLE_LIVE_PARSER:
+    enable_live_parser: bool = ENABLE_LIVE_PARSER,
```
VIKTORVAV99, line 153: And why is this passed as a argument if it's only assigned the constant?

It would be simpler to use the constant directly as before.

VIKTORVAV99 (changes requested):
I think there are some larger discussions that need to be made in this parser with the rest of the team as well, for example I don't think we need or want the live parser at all if the since the historical parser is more accurate and it has been turned off so far.

Removing it could greatly simplify the code. Either way there are some comments that needs to be addressed.

parsers/CL.py
gianantoniopini: The reason why I introduced the `_fetch_production` internal function is to be able to unit test the parser's code that processes live data. The newly added unit test `test_snapshot_live_data` calls the `_fetch_production` function with parameter `enable_live_parser` set to `True`.
So the code that was in the `fetch_production` function has been moved to the internal `_fetch_production` function, which also provides the `enable_live_parser` parameter.

I was myself not sure if to add or not the new test `test_snapshot_live_data` for the live parser, because as you mentioned it is currently disabled. However, as the parser code still supports it and since as part of this PR I had to also update that code, I decided that it would be better to have a unit test covering the live parser.

parsers/CL.py
gianantoniopini: Yes, parameter `enable_live_parser` in the `_fetch_production` function has a default value of `ENABLE_LIVE_PARSER`. This way the `fetch_production` function will produce the same results as before, as live parser is disabled.
However, it is now possible to call the `_fetch_production` function with parameter `enable_live_parser` set to `True` from a unit test, in order to cover the code processing live data. This is in fact done in the newly added test `test_snapshot_live_data`.

gianantoniopini:
Thanks for reviewing this. 👍 

I agree that removing the parser's code that deals with live data (in the case where `ENABLE_LIVE_PARSER` was set to `True`) will greatly simplify the code.
For this PR I worked under the assumption that live parser might be re-enabled in the future. See also comment on line 32 of CL.py . Therefore I also updated that code to use event classes.

parsers/CL.py
```diff
@@ -126,7 +130,7 @@ def production_processor_historical(raw_data):
 
     ordered_data = sorted(combined.values(), key=itemgetter("datetime"))
 
-    if ENABLE_LIVE_PARSER:
+    if enable_live_parser:
         # For consistency with live API, hydro and geothermal must be squeezed into unknown
         for datapoint in ordered_data:
             if "unknown" not in datapoint:
```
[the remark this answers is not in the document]
gianantoniopini: I believe this is already covered by the changes applied in the `_fetch_production` function, because that is the only place where the `production_processor_historical` function is called. Or perhaps I misunderstood?

gianantoniopini:
Thank you, I will remove the code related to live parser then.

gianantoniopini:
Live parser code has now been removed. The PR description has been updated accordingly.

VIKTORVAV99 (approved):
LGTM!

Thanks for doing this!

gianantoniopini:
Thank you!

PR #6571, opened 2024-03-22
refactor(JP-KY): remove arrow usage

shuuji3:
## Issue

#6135

## Description

- This PR removes `arrow` from `parsers/JP-KY.py`. (fa9f0b0f0731e12ab353fa16c863facb4ed7f69a)
- Also, I did some refactoring on the code to improve code readability. (72ca498896bbe114002a6699217267109a2a61be) (I run the test before/after the refactoring so the behavior should not be changed.)

### Preview

N/A

### Double check

### DIFF
--- a/parsers/JP-KY.py
+++ b/parsers/JP-KY.py
@@ -10,7 +10,7 @@ from requests import Session, get
 
 from parsers import occtonet
 
-TIMEZONE = ZoneInfo('Asia/Tokyo')
+TIMEZONE = ZoneInfo("Asia/Tokyo")
 
 
 def fetch_production(
@@ -22,6 +22,7 @@ def fetch_production(
     """Requests the last known production mix (in MW) of a given zone."""
     if target_datetime:
         raise NotImplementedError("This parser is not yet able to parse past dates")
+
     data = {
         "zoneKey": zone_key,
         "datetime": None,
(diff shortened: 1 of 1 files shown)


parsers/JP-KY.py
```diff
@@ -40,81 +41,85 @@ def fetch_production(
...
-    minutes = int(re.findall(r"(?<=時)[\d]+(?=分)", ts)[0])
+
     # get date
-    ds = soup.find("div", class_="puChangeGraph")
-    date = re.findall(r"(?<=chart/chart)[\d]+(?=.gif)", str(ds))[0]
+    date_div = soup.find("div", class_="puChangeGraph")
+    date_str = re.findall(r"(?<=chart/chart)[\d]+(?=.gif)", str(date_div))[0]
+    year_str = date_str[:4]
+    month_str = date_str[4:6]
+    day_str = date_str[6:]
```
VIKTORVAV99, line 57: Why do we need to split them up into all these strings?
Cant we just parse it as a date and then use that?

VIKTORVAV99 (approved):
LGTM!

Just one tiny comment I'd like a response to, but I think we can merge this regardless.


PS: Thanks for removing all the abbreviations.

parsers/JP-KY.py
shuuji3: Ah, that's correct. I was influenced by the original code a bit.

I'll adjust that date parsing soon.

VIKTORVAV99: Yeah I'm aware and I want to update but some of the libs we use in the backend + the version of pandas we use (I think) don't support Python 3.11 yet and we want to keep them in sync.

It's really annoying though 😅

shuuji3: That makes sense😅 I'm OK with 3.10 for now. Hopefully these dependencies will gradually adopt to newer versions.

PR #6573, opened 2024-03-23
refactor: remove arrow from CO parser

nboswell216:
## Description
Removes arrow dependency from CO parser.
Note: It seems the 'fecha' datapoint sometimes is missing the full millisecond precision, so there is a check to add a trailing zero if that is the case

### Double check


parsers/CO.py
```diff
@@ -74,9 +74,11 @@ def fetch_live_consumption(
     demand_list = TotalConsumptionList(logger)
 
     for datapoint in demand_data:
+        if ("." in datapoint["Fecha"]) and (len(datapoint["Fecha"].split(".")[1]) != 3):
+            datapoint["Fecha"] = datapoint["Fecha"].ljust(23, "0")
```
VIKTORVAV99, line 78: I think adding a comment to what this does and why will help with future maintainability.

parsers/CO.py
```diff
@@ -91,19 +93,22 @@ def fetch_historical_consumption(
...
+        "Sistema",
+        target_datetime_in_tz.date(),
+        target_datetime_in_tz.date(),
     )
 
     if not df_consumption.empty:
         hour_columns = [col for col in df_consumption.columns if "Hour" in col]
         for hour_col in hour_columns:
-            target_datetime_in_tz = target_arrow_in_tz.datetime.replace(
+            target_datetime_in_tz = target_datetime_in_tz.datetime.replace(
```
VIKTORVAV99, line 111: This should not call datetime on a datetime object.

parsers/CO.py
```diff
@@ -91,19 +93,22 @@ def fetch_historical_consumption(
 ) -> list[dict[str, Any]]:
     demand_list = TotalConsumptionList(logger)
     # Convert datetime to local time
-    target_arrow_in_tz = arrow.get(target_datetime).to(TZ)
+    target_datetime_in_tz = datetime.fromisoformat(target_datetime).replace(tzinfo=TZ)
```
VIKTORVAV99, line 96: You can not pass a datetime object to fromisoformat, you can manipulate it directly.

I also think this should use `.astimezone()` instead of replace, but double check this.

parsers/CO.py
```diff
@@ -133,41 +138,51 @@ def fetch_production(
...
     df_recursos = None
     df_generation = None
-    target_arrow_in_tz = arrow.now()
+    target_datetime_in_tz = datetime.now()
 
     if target_datetime is None:
-        target_arrow_in_tz = arrow.now().floor("day").to(TZ).shift(days=-XM_DELAY_MIN)
+        target_datetime_in_tz = datetime.now().replace(
+            hour=0, minute=0, second=0, microsecond=0
+        ).replace(tzinfo=TZ) + timedelta(days=-XM_DELAY_MIN)
```
VIKTORVAV99, line 146: Same here, the second replace should probably be `.astimezone()`.

The timedelta can also be writtern as `- timedelta(days=XM_DELAY_MIN)` instead to improve readability.

parsers/CO.py
```diff
@@ -133,41 +138,51 @@ def fetch_production(
...
+                "Recurso",
+                target_datetime_in_tz.date(),
+                target_datetime_in_tz.date(),
             )
 
             if not df_generation.empty and not df_recursos.empty:
                 break
     else:
-        target_arrow_in_tz = arrow.get(target_datetime).to(TZ)
+        target_datetime_in_tz = target_datetime.replace(tzinfo=TZ)
```
VIKTORVAV99, line 170: And here with the `.astimezone()`.

parsers/CO.py
```diff
@@ -221,7 +236,7 @@ def fetch_production(
     else:
         raise ParserException(
             parser="CO",
-            message=f"{target_arrow_in_tz.datetime}: no production data available",
+            message=f"{target_datetime_in_tz.datetime}: no production data available",
```
VIKTORVAV99, line 239: `.datetime` cannot be called on a datetime objec, only arrow objects.

parsers/CO.py
```diff
@@ -238,37 +253,41 @@ def fetch_price(
...
-    target_arrow_in_tz = arrow.now()
+    target_datetime_in_tz = datetime.now()
 
     if target_datetime is None:
         # Allow retries for most recent data
         for xm_delay in range(XM_DELAY_MIN, XM_DELAY_MAX + 1):
-            target_arrow_in_tz = arrow.now().floor("day").to(TZ).shift(days=-xm_delay)
+            target_datetime_in_tz = datetime.now().replace(
+                hour=0, minute=0, second=0, microsecond=0
+            ).replace(tzinfo=TZ) + timedelta(days=-xm_delay)
```
VIKTORVAV99, line 263: Same here as the other comment above for the code that does the same.

parsers/CO.py
```diff
@@ -238,37 +253,41 @@ def fetch_price(
...
+                target_datetime_in_tz.date(),
             )
 
             if not df_price.empty:
                 break
     else:
-        target_arrow_in_tz = arrow.get(target_datetime).to(TZ)
+        target_datetime_in_tz = datetime.fromisoformat(target_datetime).replace(
+            tzinfo=TZ
+        )
```
VIKTORVAV99, line 277: And here with both skipping fromisoformat and using astimezone again.

VIKTORVAV99 (changes requested):
Several small fixes needed but looks to be on the right track!

Thanks for upgrading this parser.

nboswell216:
@VIKTORVAV99 I researched the difference between .replace() and astimezone() for changing timezones and indeed it looks like astimezone() is the preferred method. I used replace() in my refactor for both CA_AB and CA_BC - do you want me to go back and update those to astimezone(), or are you ok with those as they are?

VIKTORVAV99:
> @VIKTORVAV99 I researched the difference between .replace() and astimezone() for changing timezones and indeed it looks like astimezone() is the preferred method. I used replace() in my refactor for both CA_AB and CA_BC - do you want me to go back and update those to astimezone(), or are you ok with those as they are?

In general I think where we used `.replace()` in arrow we should still use replace and where arrow used `.to()` we should use `.astimezone()` instead.

The absolut best way to ensure there is no missmatch between the two would be to add snapshot tests before modifying them and then making sure the date times don't change (more than the tz info).

parsers/CO.py
```diff
@@ -91,19 +97,22 @@ def fetch_historical_consumption(
 ) -> list[dict[str, Any]]:
     demand_list = TotalConsumptionList(logger)
     # Convert datetime to local time
-    target_arrow_in_tz = arrow.get(target_datetime).to(TZ)
+    target_datetime_in_tz = target_datetime.astimezone(tz=TZ)
```
VIKTORVAV99, line 100: Target datetime might be `None` here and we should therefor not call `.astimezone()` directly here.

We should do something like this instead:
```python
    now = datetime.now(tz=TIMEZONE)
    target_datetime_in_tz = (
         now if target_datetime is None else target_datetime.astimezone(TIMEZONE)
     )
```

parsers/CO.py
```diff
@@ -74,9 +74,15 @@ def fetch_live_consumption(
...
+        # Ensure consistent millisecond precision for ISO compliance:
+        # Checks if the 'Fecha' field contains a decimal point (indicating milliseconds)
+        # and if the milliseconds portion is less than three digits. If so, it pads the 'Fecha'
+        # value with zeros to achieve the required ISO standard precision.
+        if ("." in datapoint["Fecha"]) and (len(datapoint["Fecha"].split(".")[1]) != 3):
+            datapoint["Fecha"] = datapoint["Fecha"].ljust(23, "0")
         demand_list.append(
             zoneKey=zone_key,
-            datetime=arrow.get(datapoint["Fecha"], tzinfo=TZ).datetime,
+            datetime=datetime.fromisoformat(datapoint["Fecha"]).astimezone(tz=TZ),
```
VIKTORVAV99, line 85: I would probably use replace here but I think both should do the same.

As long as the output is what you expect this is fine.

VIKTORVAV99 (commented):
Two small comments, otherwise it looks great!

VIKTORVAV99:
Sorry for taking so long to review this but I have been unable to get the parser working locally, it seems to be unrelated to your changes but I don't really want to merge this without a side by side comparison of the data since we don't have any tests for this parser.

madsnedergaard:
Would love to get this PR approved and merged, but unfortunately we're getting geoblocked over here :(

If this still works for you, would you mind running `poetry run test_parser CO` on `master` and on your branch here, and share the results to verify they are the same? :)

nboswell216:
This looks like it was resolved by https://github.com/electricitymaps/electricitymaps-contrib/commit/1a03862d4e7e66e91355bf819a9d6fd3dcbb4dbb. I will close this PR

PR #6576, opened 2024-03-23
refactor: remove arrow from CA_BC parser

nboswell216:
## Description
Removes arrow dependency from CA_BC parser
renames datetime variable to parsed_datetime to avoid conflict

### Double check


VIKTORVAV99 (approved):
LGTM!

Thanks for fixing it!

PR #6577, opened 2024-03-23
feat(ENTSOE.py): Parse and combine both kinds of exchange forecasts

VIKTORVAV99:
## Issue

Currently we are only parsing the Total forecasts which are the most accurate but this means we are missing out on the day ahead forecasts which are available further into the future than the total forecasts are.

## Description

This PR duplicates some functionality for the sake of simple logic in order to parse the day_ahead forecasts and then update them with the total forecasts so we get benefits of the longer day_ahead forecasts while retaining the shorter term accuracy of the total forecasts.

TODO:

- [x] Add new snapshot features that matches the current lookup periods to ensure the new code work as expected.

#### Latest value on master
```python
 {'datetime': datetime.datetime(2024, 3, 23, 22, 0, tzinfo=datetime.timezone.utc),
  'netFlow': 0.0,
  'sortedZoneKeys': 'EE->FI',
  'source': 'entsoe.eu',
  'sourceType': <EventSourceType.forecasted: 'forecasted'>}
```
#### Latest value on this branch
```python
 {'datetime': datetime.datetime(2024, 3, 24, 22, 0, tzinfo=datetime.timezone.utc),
  'netFlow': -358.0,
  'sortedZoneKeys': 'EE->FI',
  'source': 'entsoe.eu',
  'sourceType': <EventSourceType.forecasted: 'forecasted'>}
```

### Double check


parsers/ENTSOE.py
```diff
@@ -926,13 +936,45 @@ def parse_exchange(
         datetime_start = datetime.fromisoformat(
             zulu_to_utc(timeseries.find_all("start")[0].contents[0])
         )
-        # Only use contract_marketagreement.type == A01 (Total to avoid double counting some columns)
```
pierresegonne, line 929: What about this? Are we not risking the double counting if we start ingesting the other forecasts?

parsers/ENTSOE.py
```diff
@@ -1161,26 +1205,95 @@ def _fetch_and_parse_exchange(
...
+            )
+
         # Then we try to get the export data
-        raw_exchange_lists.append(
-            _fetch_and_parse_exchange(domain_pair, is_import=False)
-        )
+        raw_exchange = _fetch_exchange(domain_pair, is_import=False)
 
-    return ExchangeList(logger).merge_exchanges(raw_exchange_lists, logger)
+        if raw_exchange is None:
```
pierresegonne, line 1252: Can't we separate more clearly the forecast and non forecasts cases?

parsers/ENTSOE.py
VIKTORVAV99: Both kinds are returned in the same API call so in order to parse it we need to make this dynamic.

Right now it calls the parser function twice, first for the day-ahead values and a second time for the total values to then combine them.

Functionally it should do the same but just use a dynamic config instead of a static value.

parsers/ENTSOE.py
VIKTORVAV99: We can but it would likely result in more code duplication, if that's okay I can do right ahead and split them apart further.

parsers/ENTSOE.py
```diff
@@ -926,13 +936,45 @@ def parse_exchange(
...
+    exchange_list = ExchangeList(logger)
+
+    soup = BeautifulSoup(xml_text, "html.parser")
+    # Get all points
+    for timeseries in soup.find_all("timeseries"):
+        resolution = str(timeseries.find_all("resolution")[0].contents[0])
+        marketAgreementType = timeseries.find("contract_marketagreement.type").contents[
+            0
+        ]
+        if marketAgreementType and marketAgreementType != market_type.value:
```
VIKTORVAV99, line 973: @pierresegonne here is the functional equivalent of the previous check.

VIKTORVAV99:
@pierresegonne is this something we can merge? Or are there anything I still need to adress/change?

pierresegonne (approved):

VIKTORVAV99:
As I discussed with @Garsdal we will leave this un-merged while he is looking into the exchange data we currently have, so nothing suddenly changes, and then he'll merge it at his discretion when he is ready.

PR #6583, opened 2024-03-27
Fix: move flag and fix text length in breakdownchart

silkeholmebonnen:
## Issue

The text length in the breakdownchart was not correct after the production source legend was implemented. See the "before" screenshots, this is what it looks like on the app at the moment.

## Description

This PR moves the flag so it is in the same place as the production source legend. I tried to fix the length without moving the flag, but I thought it looked much better when moving the flag. 

### Preview

Before:
Screenshot 2024-03-27 at 13 48 06
Screenshot 2024-03-27 at 13 49 42
Now:
Screenshot 2024-03-27 at 13 50 37
Screenshot 2024-03-27 at 13 49 30

### Double check


VIKTORVAV99 (approved):
LGTM!

PR #6584, opened 2024-03-27
feat(leftpanel): implement the accordion list component

silkeholmebonnen:
## Issue

[AVO-81](https://linear.app/electricitymaps/issue/AVO-81/implement-the-accordion-list-component)

## Description

The goal of this PR is to implement the accordion component and to use it to implement the data source list. 

### Preview

image

### Double check


silkeholmebonnen:
I have made this draft PR because there are a few things I am unsure about.

I have added the svg's for both light and dark because I couldn't figure out a better way to do it. Since @Alportan is using font awesome for icons in figma I cannot find a better way to use the icons than adding them as svg's. This is fine for now maybe but we probably want to think about how we do it in the future. I think Alex put it as a topic for the next sync.

I don't see a way to show the Exchange data sources, because we don't get that information (right?). What would we need to do in order to get this information?

Is the logic of what I have done so for right?

@madsnedergaard @VIKTORVAV99

VIKTORVAV99:
> I have made this draft PR because there are a few things I am unsure about.
> 
> I have added the svg's for both light and dark because I couldn't figure out a better way to do it. Since @Alportan is using font awesome for icons in figma I cannot find a better way to use the icons than adding them as svg's. This is fine for now maybe but we probably want to think about how we do it in the future. I think Alex put it as a topic for the next sync.
> 
> I don't see a way to show the Exchange data sources, because we don't get that information (right?). What would we need to do in order to get this information?
> 
> Is the logic of what I have done so for right?
> 
> @madsnedergaard @VIKTORVAV99

Will give some more detailed comments directly on the related code but looks pretty good so far. Regarding the Font Awesome icons there should probably be a NPM package we can use instead to import the icons, that way we would get the benefit of bundling them. This can be a separate PR though.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -84,6 +85,17 @@ function BarBreakdownChart({
...
+  const emissionData = [
+    ...new Set(
+      [
+        ...Object.values(currentZoneDetail?.dischargeCo2IntensitySources || {}),
+        ...Object.values(currentZoneDetail?.productionCo2IntensitySources || {}),
+      ].flatMap((item) => item.split('; '))
+    ),
+  ]
+    .filter((item) => !item.startsWith('assumes'))
+    .sort();
```
VIKTORVAV99, line 97: This is going to be a very costly operation performance wise but there is a few ways we can make it faster. Will try a few things locally over easter and give some more concrete suggestions.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -138,8 +150,93 @@ function BarBreakdownChart({
...
+        <div className="mr-1">{icon}</div>
+        <div className="text-md font-semibold">{title}</div>
+      </div>
+      <div className="flex flex-col gap-2">
+        {sources.map((source, index) => (
+          <div key={index} className="text-xm">
+            {source}
+          </div>
+        ))}
+      </div>
```
VIKTORVAV99, line 239: This should probably have some indentation to clearly mark it as a group of sources. @Alportan what do you think?

image

web/src/components/Accordion.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function Accordion({
+  isCollapsedDefault = true,
+  onClick,
+  badge,
+  className,
+  icon,
+  children,
+  title,
+}: {
```
VIKTORVAV99, line 13: This should probably also accept a atom so we can perserve the state in a way where the component is still reusable.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -138,8 +150,93 @@ function BarBreakdownChart({
...
     </div>
   );
 }
 
 export default BarBreakdownChart;
+
+function GetSourceArrayFromDictionary(sourceDict: {
+  [key in ElectricityModeType]: string[] | null;
+}): string[] {
+  const sourceArray: string[] = [];
```
VIKTORVAV99, line 201: If this is made into a set.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -138,8 +150,93 @@ function BarBreakdownChart({
...
+  if (sourceDict == null) {
+    return sourceArray;
+  }
+  for (const key of Object.keys(sourceDict)) {
+    const capacitySource = sourceDict?.[key as ElectricityModeType];
+    if (capacitySource != null) {
+      for (const source of capacitySource) {
+        if (!sourceArray.includes(source)) {
+          sourceArray.push(source);
```
VIKTORVAV99, line 210: You don't need to do this.

madsnedergaard:
> I have added the svg's for both light and dark because I couldn't figure out a better way to do it. Since @Alportan is using font awesome for icons in figma I cannot find a better way to use the icons than adding them as svg's. This is fine for now maybe but we probably want to think about how we do it in the future. I think Alex put it as a topic for the next sync.
 
 Let's take this discussion on the next team sync, there's many angles and details to it :)

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
silkeholmebonnen: I agree. Should I change it to have indentation @Alportan ?

web/src/components/Accordion.tsx
silkeholmebonnen: What state do we need to preserve? If the accordion is collapsed or expanded?

VIKTORVAV99: Yeah so if you switch between zones it stays open and don't auto close again. And by passing a atom we can do this for each graph later on.

silkeholmebonnen: I have made it accept an atom now. I also made an atom for datasources and I don't know if we actually want this. What do you think?

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
Alportan: Hmmm, I did played a bit with indentation when I was designing, but (at least on mobile 📱) it takes alot of screen estate to indent them especially for longer sources which can get into 2+ rows. 

I would still keep it left aligned and maybe increase the padding between the sections a bit. What do you think about that? @VIKTORVAV99 @silkeholmebonnen

Alportan: Let me play a bit and get back to you!

VIKTORVAV99: Increasing the padding comes with its own negative effects (also mostly for mobile) with showing less information on the vertical space instead.

Your concerns are very valid though so maybe we should leave this as is for now and test out both? Could even have something more conditional based on screen size where  small screens have no indentation and somewhat larger screen does (think iPhone vs Iphone Plus or along those lines).

Alportan: I'll add a small task to revise the source alignment to left aligned, with the icon leading, as we're heading towards that direction anyways with the new design. 🙌

silkeholmebonnen:
I added indentation to the sources, and moved the Source component to it's own file because I realised that this would be used to to implement the other data source lists below the other graphs.

web/public/images/industry_dark.svg
VIKTORVAV99, line 1: You created an icons folder in the other PR right? Do you want to move these there as well or how do you feel about these?

VIKTORVAV99 (approved):
LGTM!

Nice work!

web/public/images/industry_dark.svg
silkeholmebonnen: True, I'll change it. I also added tracking events for the data sources, and for the contrib button.

madsnedergaard (approved):
Looks good to me :)

PR #6587, opened 2024-03-27
refactor: upgrade BG parser with event classes

amv213:
## Issue

#6011

## Description

Hello! This PR upgrades the BG parser to use event classes.

This is my first contribution so do not hesitate to let me know if there is anything I did wrong / should be done differently 🙌🏽 

### Double check


parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
-    """Main method, never used by the Electricity Map backend, but handy for testing."""
-
-    print("fetch_production() ->")
-    print(fetch_production())
+def fetch_production(
+    zone_key: ZoneKey = ZoneKey("BG"),
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list[dict]:
```
amv213, line 91: question: should I have returned here just a single event like before?

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
+) -> list[ProductionBreakdown]:
+    """Fetches all production events from the backend source API."""
+    if target_datetime is not None:
+        raise ParserException(
+            PARSER, "This parser is not yet able to parse historical data", zone_key
+        )
+
+    time = datetime.now(timezone.utc)
+    response = session.get(SOURCE_API_URL)
+    if not response.status_code == 200:
```
VIKTORVAV99, line 45: Small tip:
This can be simplified with:
```suggestion
    if not response.ok:
```

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
+            hydro=production.get("hydro"),
+            nuclear=production.get("nuclear"),
+            oil=production.get("oil"),
+            solar=production.get("solar"),
+            unknown=production.get("unknown"),
+            wind=production.get("wind"),
+        ),
+        storage=None,
+    )
+    return [event]
```
VIKTORVAV99, line 83: While this works it would be simpler and better to do this directly instead of getting the values from the production object already created.

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
+        raise ParserException(
+            PARSER,
+            f"Exception when fetching production error code: {response.status_code}: {response.text}",
+            zone_key,
+        )
+
+    response_payload = response.json()
+    logger.debug(f"Raw generation breakdown: {response_payload}")
 
     production = {}
```
VIKTORVAV99, line 55: This should be a production mix object instead:
```suggestion
    production = ProductionMix()
```
But we also need a ProductionBreakdownList object here as well so we can append it to something.

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
-    }
-
-    return data
+    for header, value in response_payload:
+        for production_type_bg in PRODUCTION_TYPES_BG_TO_EN:
+            if header.startswith(production_type_bg):
+                production_type_en = PRODUCTION_TYPES_BG_TO_EN[production_type_bg]
+                production[production_type_en] = (
+                    production.get(production_type_en, 0.0) + value
+                )
```
VIKTORVAV99, line 62: If the production mix class is used above this can be simplified to:
```suggestion
                production.add_value(production_type_en, value)
```

parsers/BG.py
VIKTORVAV99: No returning a list is the correct way to go about it, even if it just has a single value.

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
+
+    event = ProductionBreakdown.create(
```
VIKTORVAV99, line 64: If the above suggestions are implemented you can do this instead:
```python
productionBreakdownList.append(
  zoneKey=zone_key,
  datetime=time,
  source=SOURCE,
  production=production,
)
```
Which is a lot simpler and has built in protections.

parsers/BG.py
```diff
@@ -1,72 +1,109 @@
...
+    production_list = ProductionBreakdownList(logger)
+    for event in all_production_events:
+        production_list.append(
+            zoneKey=event.zoneKey,
+            datetime=event.datetime,
+            source=event.source,
+            production=event.production,
+            storage=event.storage,
+            sourceType=event.sourceType,
+        )
```
VIKTORVAV99, line 108: then you can remove this and return inline the other function directly. Avoiding a lot of additional computations and assignments.

parsers/test/test_BG.py
```diff
@@ -0,0 +1,76 @@
...
+    fixture_requests_mock,
+):
+    """That a ParserException is raised if requesting historical data (not supported yet)."""
+    fixture_requests_mock.get(SOURCE_API_URL, json=[])
+
+    with pytest.raises(
+        ParserException, match="This parser is not yet able to parse historical data"
+    ):
+        historical_datetime = datetime.now(timezone.utc) - timedelta(days=1)
+        fetch_production(target_datetime=historical_datetime)
```
VIKTORVAV99, line 76: A simpler way to do this is to add a snapshot test which requires a lot less manual work and only need a fixture (what is returned from the api) and then the request to that is mocked.

VIKTORVAV99 (changes requested):
A good start! Thank you!

I have added a lot of comments on how to simplify the parser a bit. But if you have any questions feel free to ping me directly!

parsers/test/test_BG.py
amv213: Thanks @VIKTORVAV99 !

I initially tried using snapshots but the auto-generated snapshots were ending up with invalid syntax (e.g. `sourceType`): 

```
def test_snapshot(snapshot):
    response = fetch_production()
    snapshot.assert_match(response)
```

```
# -*- coding: utf-8 -*-
# snapshottest: v1 - https://goo.gl/zC4yUc
from __future__ import unicode_literals

from snapshottest import GenericRepr, Snapshot


snapshots = Snapshot()

snapshots['test_snapshot 1'] = [
    {
        'correctedModes': [
        ],
        'datetime': GenericRepr('datetime.datetime(2024, 3, 27, 19, 13, 2, 948224, tzinfo=datetime.timezone.utc)'),
        'production': {
            'biomass': 20.83,
            'coal': 760.02,
            'gas': 452.7,
            'hydro': 655.81,
            'nuclear': 2120.73,
            'solar': 0,
            'wind': 197.86
        },
        'source': 'eso.bg',
        'storage': {
        },
        'zoneKey': 'BG'
    }
]
```

Do you know if I am doing anything wrong? Snapshots for other parsers seem to serialise properly...

parsers/BG.py
amv213: Will do - thanks! 

thought: should I remove `"Товар РБ": "consumption"` from the mapping above then? If not feeding things directly into the `ProductionMix` here could potentially raise an error as `consumption` is not a valid mode

parsers/test/test_BG.py
VIKTORVAV99: I think you just need to call `.value` on the sourceType as it is a enum to get the actual value and not the enum object representation as shown here.

parsers/BG.py
VIKTORVAV99: Remove it from the production mapping but leave it as a comment so we know what it means in the future.

parsers/test/test_BG.py
VIKTORVAV99: Here is how it is done in another parser (but for exchange forecasts but the principle is the same):
```python
        self.assertMatchSnapshot(
            [
                {
                    "datetime": element["datetime"].isoformat(),
                    "netFlow": element["netFlow"],
                    "source": element["source"],
                    "sortedZoneKeys": element["sortedZoneKeys"],
                }
                for element in exchange_forecast
            ]
        )
```

amv213: I see - thanks! 

It looks like this bug is why we can't simply do `self.assertMatchSnapshot(response)` https://github.com/syrusakbary/snapshottest/issues/120

parsers/BG.py
```diff
@@ -1,72 +1,69 @@
...
+
+    response_payload = response.json()
+    logger.debug(f"Raw generation breakdown: {response_payload}")
+
+    production_mix = ProductionMix()
+    for header, value in response_payload:
+        production_type, _delimiter, _percentage = header.rpartition(" ")
+        production_mode = PRODUCTION_TYPE_TO_PRODUCTION_MODE.get(production_type)
+        if production_mode in PRODUCTION_MODES:
+            production_mix.add_value(production_mode, value)
```
VIKTORVAV99, line 60: We should skip this check, if there is a mode that is not a part of the production modes (which is checked internally) we want to raise logs for it so we can map it to something valid.

parsers/test/test_BG.py
```diff
@@ -0,0 +1,68 @@
+from datetime import datetime, timedelta, timezone
+
+import pytest
+import requests_mock
+from freezegun import freeze_time
+
+from electricitymap.contrib.parsers.lib.exceptions import ParserException
```
VIKTORVAV99, line 7: This is a symlink which might break on windows (we have had some trouble with that in the past) and it should be imported from just `parsers.lib.exceptions` instead. (Sadly we can't remove the symlink due to internal uses.)

parsers/BG.py
```diff
@@ -1,72 +1,67 @@
...
-    "Малки ВЕЦ": "hydro",  # 74
-    "ВяЕЦ": "wind",  # 488
-    "ФЕЦ": "solar",  # 0
-    "Био ТЕЦ": "biomass",  # 29
-    "Био ЕЦ": "biomass",  # 29
-    "Товар РБ": "consumption",  # 3175
+from electricitymap.contrib.lib.models.event_lists import ProductionBreakdownList
+from electricitymap.contrib.lib.models.events import ProductionMix
+from electricitymap.contrib.lib.types import ZoneKey
+from electricitymap.contrib.parsers.lib.exceptions import ParserException
```
VIKTORVAV99, line 9: Same here about the exception. the others are fine though.

VIKTORVAV99 (commented):
Some final comments, but functionality wise this works great.

parsers/test/test_BG.py
```diff
@@ -0,0 +1,68 @@
...
+from parsers.BG import SOURCE_API_URL, fetch_production
+
+
+@pytest.fixture()
+def fixture_requests_mock():
+    """A mocker for requests.
+
+    TODO(amv213): replace with builtin pytest 'requests_mock' fixture when upgrading to requests-mock>=1.5.0
```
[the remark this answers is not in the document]
amv213: Thanks for the link @VIKTORVAV99 - I will use that template / switch to `unittest`-style tests

thought: maybe it would be less awkward on `requests-mock>=1.5.0` where `requests_mock` is available as a builtin fixture, but this is what that entire `test_CEB.py` file would look like `pytest`-style: 
```python
import json
from importlib import resources

from requests_mock import ANY

from electricitymap.contrib.lib.types import ZoneKey
from parsers.CEB import fetch_production


def test_production(snapshot, requests_mock):
    requests_mock.get(
        ANY, 
        json=json.loads(
            resources.files("parsers.test.mocks.CEB")
            .joinpath("response.text")
            .read_text()
        ),
    )

    production = fetch_production(zone_key=ZoneKey("LK"))
    
    snapshot.assert_match(
        [
            {
                "datetime": element["datetime"].isoformat(),
                "production": element["production"],
                "storage": element["storage"],
                "source": element["source"],
                "zoneKey": element["zoneKey"],
            }
            for element in production
        ]
    )

```
which I find more readable / maintainable... but I'm probably just biased because I haven't written `unittest`-style tests in a very long time 😉 

Either way, I completely agree with your suggestion! It will also avoid having the test structure / framework diverge across parsers which is probably the most important thing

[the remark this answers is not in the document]
VIKTORVAV99: We might be able to upgrade the package and streamline the tests even further at one point but I don't think that should be a goal in this PR.

VIKTORVAV99 (approved):
LGTM!

Thanks for making all the changes!

PR #6588, opened 2024-03-27
refactor_remove arrow from CA_NB

nboswell216:
## Description
Removes "usage" of arrow in CA_NB parser
Note: This parser does not seem to have used arrow at least since the upgrade to Python 3.10 - the only thing keeping this on the list when running `grep -El '(import arrow|arrow)' parsers/*.py` is a comment. The comment was removed.

### Double check


VIKTORVAV99 (approved):
Thanks for cleaning this up!

PR #6590, opened 2024-03-28
fix(ERP_PGCB.py): fix BD unknown data and upgrade to event lists

### LINKED ISSUE #6542: [Data Issue]: BD imports are counted as generation in source
### When did this happen?

Ongoing since start?

### What zones are affected?

BD

### What is the problem?

It seems like the source is including imports as generation but we do not, so where are inferring an unknown production that is almost 1=1 with the imports. The parser needs to be updated to account for this and then all data needs to be re-fetched and reprocessed.

VIKTORVAV99:
## Issue
Fixes #6542

## Description

Updates the BD parser to parser classes and fixes #6542. The snapshots are seeing significant changes due to the fact that the sorting changed and I fixed an error in the datetime calculations.

### Double check


parsers/ERP_PGCB.py
```diff
@@ -73,17 +77,13 @@ def parse_table_body(table_body: Tag) -> list[dict]:
...
-        except ValueError:
-            # very old data points are sometimes in HH:mm format
-            parsed_time = datetime.strptime(row_items[1], "%H:%M").time()
+        assert isinstance(row_items[1], str)
+        hour = int(row_items[1][0:2])
+        # The hour is backwards looking so if it's 01:00, it's actually 00:00
+        hour = hour - 1
+        minute = int(row_items[1][3:5])
+        second = int(row_items[1][6:8])
+        parsed_time = time(hour, minute, second)
```
VIKTORVAV99, line 86: This is the only part I am unsure of, the values looks more correct when doing this and it's more logical but there are a few datapoints where both 00:00 and 24:00 exists. But then again I also found two 01:00 values for one day so hard to tell which is right.

VIKTORVAV99:
@q-- I would appreciate it if you took a look at this as well if you have the time. 🙂

q--:
> @q-- I would appreciate it if you took a look at this as well if you have the time. 🙂

q--:
> @q-- I would appreciate it if you took a look at this as well if you have the time. 🙂

Hi,

Unfortunately I don't have much time right now, so I don't think I'll be able to help out with this one. Thank you for fixing this, though!

VIKTORVAV99:
> > @q-- I would appreciate it if you took a look at this as well if you have the time. 🙂
> 
> Hi,
> 
> Unfortunately I don't have much time right now, so I don't think I'll be able to help out with this one. Thank you for fixing this, though!

No worries!
And thanks for flagging the issue!

unitrium (approved):

PR #6591, opened 2024-03-29
improve datetime naiveness check

amv213:
## Description

A very small PR to improve datetime validation in event classes. 

According to the official [datetime docs](https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive):
> - d is aware iff: `d.tzinfo is not None and d.tzinfo.utcoffset(d) is not None`
> - d is naive iff: `d.tzinfo is None or d.tzinfo.utcoffset(d) is None`

This more exhaustive check shouldn't change much but It might help avoid some sneaky edge cases getting through.

### Double check

### DIFF
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -1,5 +1,4 @@
 # pylint: disable=no-member
-import datetime as dt
 import math
 from abc import ABC, abstractmethod
 from collections.abc import Set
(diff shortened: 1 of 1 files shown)


electricitymap/contrib/lib/models/events.py
```diff
@@ -296,8 +300,8 @@ def _validate_zone_key(cls, v):
         return v
 
     @validator("datetime")
-    def _validate_datetime(cls, v: dt.datetime, values: dict[str, Any]):
-        if v.tzinfo is None:
+    def _validate_datetime(cls, v: datetime, values: dict[str, Any]):
```
VIKTORVAV99, line 303: The reason we imported datetime again as dt was because if we use `datetime` directly we get this issue:
```log
Variable not allowed in type expressionPylancereportInvalidTypeForm
(variable) datetime: Unknown
```
So if you revert that the other changes looks good and we can merge this.

VIKTORVAV99 (approved):
LGMT!

I made a small change to the return type but that was only for type hinting to work properly.

PR #6359, opened 2024-01-17
Fix 6203 tooltip overflows behind screen

PmplCode:
## Issue

Closes #6203 , although it is my first time taking part on an open source project and i would like guidance.
## Description

tooltip remains visible everytime even with scrolled screen.
### Preview

https://www.linkedin.com/posts/eloipampliegajose_opensource-frontenddev-electricitymaps-activity-7153079821013508096-yzqW
### Double check

- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [ ] I have run `pnpx prettier --write .` and `poetry run format` to format my changes.

PR #6594, opened 2024-03-29
perf(exchange layer): deduplicate exchange tooltips

VIKTORVAV99:
## Issue
We have a duplicate code that is functioning almost exactly the same and is creating a lot of unnecessary dom elements of which half are hidden at any given time.

## Description
This PR deduplicates the code and unifies the MobileExchangeTooltip and ExchangeTooltip as well as ExchangeTooltipWrapper and MobileExchangeTooltipWrapper into just one single component respectively that can be used by both mobile and none mobile.

Some highlights:
- Changes the click behaviour on "desktop" so the tooltip doesn't disappear when you click on the arrow.
- Deduplicates the code between the tooltips and wrappers into one.
- Adds a useDimension hook to get if the screen size is below or above the mobile breakpoint.
- Optimises the early returns of the wrapper to reduce the amount of function calls that are made if we return early.
- Removes unneeded dom elements in the tooltip.
- Reduces the dom count by 792 elements (dependant on how many arrows are rendered)

### Double check


web/src/features/exchanges/ExchangeTooltip.tsx
```diff
@@ -23,16 +24,20 @@ export default function ExchangeTooltip(
   const zoneTo = key.split('->')[isExporting ? 1 : 0];
   const [timeAverage] = useAtom(timeAverageAtom);
   const isHourly = timeAverage === TimeAverages.HOURLY;
+  const { isMobile } = useDimensions();
```
VIKTORVAV99, line 27: This could be passed as a prop instead, not sure which is preferable TBH.

web/src/hooks/dimensions.ts
VIKTORVAV99, line 1: I thought we had a hook that did something similar already (to `isMobile`) but I can't find it in the hooks folder?

VIKTORVAV99: @madsnedergaard do you have any insight here?

silkeholmebonnen:
The tooltip on mobile goes out of the screen sometimes, and it didn't before. 
Screenshot 2024-04-04 at 08 50 26
image


It seems like this was already a problem on desktop before, but we might want to change it in this PR as well?
Screenshot 2024-04-04 at 08 51 06

VIKTORVAV99:
> The tooltip on mobile goes out of the screen sometimes, and it didn't before. 
> 
> Screenshot 2024-04-04 at 08 50 26
> 
> image
> 
> 
> 
> 
> 
> It seems like this was already a problem on desktop before, but we might want to change it in this PR as well?
> 
> Screenshot 2024-04-04 at 08 51 06

Nice find!

There is another PR open at #6359 that focuses on that for other tooltips so I'll try and get that merged as soon as possible and see if it fixes this as well or can be reused in here.

web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -26,124 +23,82 @@ function ExchangeArrow({
...
-  // ... or if it would be rendered outside of viewport.
-  if (transform.x + 100 * transform.k < 0) {
+  if (
+    // or if the arrow would be very tiny
+    transform.k < 0.1 ||
+    // or if it would be rendered outside of viewport.
+    transform.x + 100 * transform.k < 0 ||
+    transform.x - 100 * transform.k > viewportWidth ||
+    transform.y - 100 * transform.k > viewportHeight
+  ) {
```
silkeholmebonnen, line 52: Why do we not check for `absFlow < 1` and `transform.y + 100 * transform.k < 0` anymore?

web/src/hooks/dimensions.ts
silkeholmebonnen: In Map.tsx isMobile is defined: `const isMobile = window.innerWidth < 768;`
Should we use the useDimensions instead here?

web/src/features/exchanges/ExchangeTooltip.tsx
```diff
@@ -23,16 +24,20 @@ export default function ExchangeTooltip(
...
-        <div className="flex items-center pb-2">
-          <ZoneName zone={zoneFrom} textStyle="max-w-[165px]" /> <p className="mx-2">→</p>{' '}
+        <div className={divClass}>
+          <ZoneName zone={zoneFrom} textStyle="max-w-[165px]" />
+          {isMobile ? <p className="ml-0.5">↓</p> : <p className="mx-2">→</p>}{' '}
           <ZoneName zone={zoneTo} textStyle="max-w-[165px]" />
           <b className="font-bold">
-            : {isHourly ? formatPower(roundedNetFlow) : formatEnergy(roundedNetFlow)}
+            {isMobile ? '' : ':'}{' '}
+            {isHourly ? formatPower(roundedNetFlow) : formatEnergy(roundedNetFlow)}
```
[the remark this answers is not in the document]
VIKTORVAV99: We don't show it for anything else right now but there is a long term plan to do so in the future.

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: Abs flow is checked further up so we exit sooner and avoid functions calls that we won't use anyway.

The other check is a miss on my side so I'll add that back 🙂

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -17,10 +18,27 @@ export default function TooltipWrapper(
...
   }
+  const [isOpen, setIsOpen] = useState(false);
+  const { isMobile } = useDimensions();
+
+  // Declare the event handlers outside of the JSX to avoid re-creating them on every render.
+  const handleMouseEnter = isMobile ? () => {} : () => setIsOpen(true);
+  const handleMouseLeave = isMobile ? () => {} : () => setIsOpen(false);
+  const handleClick = isMobile ? () => setIsOpen(!isOpen) : () => {};
+  const handleContentClick = isMobile ? () => setIsOpen(false) : () => {};
+  const handleContentPointerDownOutside = isMobile ? () => setIsOpen(false) : () => {};
```
madsnedergaard, line 29: How about something like this to make it a bit more readable? :)

```suggestion
  const noop = () => {};
	const openTooltip = () => setIsOpen(true);
	const closeTooltip = () => setIsOpen(false);
	const toggleTooltip = () => setIsOpen(!isOpen);
	
  const handleMouseEnter = isMobile ? noop :  openTooltip;
  const handleMouseLeave = isMobile ? noop : closeTooltip;
  const handleClick = isMobile ? toggleTooltip : noop;
  const handleContentClick = isMobile ? closeTooltip : noop;
  const handleContentPointerDownOutside = isMobile ? closeTooltip : noop;
```

web/src/hooks/dimensions.ts
madsnedergaard: Usage example: `  const isBiggerThanMobile = useBreakpoint('sm');`

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -17,10 +18,27 @@ export default function TooltipWrapper(
   if (!tooltipContent) {
     return children;
   }
+  const [isOpen, setIsOpen] = useState(false);
+  const { isMobile } = useDimensions();
```
madsnedergaard, line 22: Let's use our existing hook instead: `const isBiggerThanMobile = useBreakpoint('sm');`

web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -26,124 +23,82 @@ function ExchangeArrow({
...
 
-  if (transform.y + 100 * transform.k < 0) {
-    return null;
-  }
+  const setIsMoving = useSetAtom(mapMovingAtom);
 
-  if (transform.x - 100 * transform.k > viewportWidth) {
-    return null;
-  }
+  useEffect(() => {
```
madsnedergaard, line 58: Doesn't this break the [Rules of Hooks](https://legacy.reactjs.org/docs/hooks-rules.html), as the hook is only conditionally rendered? Guess this was also a problem before though, so we should probably use the linting plugin for hooks: https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks

web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -26,124 +23,82 @@ function ExchangeArrow({
...
-      </MobileTooltipWrapper>
-      <TooltipWrapper
-        tooltipClassName="max-h-[256px] max-w-[512px] top-[-76px] hidden md:flex"
-        tooltipContent={<ExchangeTooltip exchangeData={data} />}
-        side="right"
-        sideOffset={10}
+    <TooltipWrapper
+      tooltipClassName="max-h-[256px] max-w-[512px] top-[-76px]"
+      tooltipContent={<ExchangeTooltip exchangeData={data} />}
+      side="right"
```
madsnedergaard, line 81: Should the side be `top` for mobile and `side` on desktop like it was before?

madsnedergaard (commented):
Some comments, but nice :)

web/src/hooks/dimensions.ts
VIKTORVAV99: Maybe we can switch both to the useBreakpoint hook?
Will look into how the other PR I mentioned does it as well.

web/src/components/tooltips/TooltipWrapper.tsx
VIKTORVAV99: Will do!

web/src/components/tooltips/TooltipWrapper.tsx
VIKTORVAV99: the sm breakpoint is a lot smaller than what we use as isMobile on the map, should it not be md? Or should we change the map as well?

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: I do not know but since the renders should not change between the hours themselves and we are re rendering all hours I think we should be safe, but I'll look into it.

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: I'll double check!

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -17,10 +18,27 @@ export default function TooltipWrapper(
...
+  const handleMouseLeave = isMobile ? () => {} : () => setIsOpen(false);
+  const handleClick = isMobile ? () => setIsOpen(!isOpen) : () => {};
+  const handleContentClick = isMobile ? () => setIsOpen(false) : () => {};
+  const handleContentPointerDownOutside = isMobile ? () => setIsOpen(false) : () => {};
+
   return (
-    <Tooltip.Provider disableHoverableContent>
-      <Tooltip.Root delayDuration={0}>
-        <Tooltip.Trigger asChild>{children}</Tooltip.Trigger>
+    <Tooltip.Provider>
```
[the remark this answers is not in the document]
VIKTORVAV99: With the offset the tooltip have we never reach the inside of the tooltip anyway but I added it back in case this changes in the future.

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: I'll leave it as is now and then I'll open a new PR with the new eslint plugin and fix everything that needs to be fixed in there.

web/src/components/tooltips/TooltipWrapper.tsx
VIKTORVAV99: Ended up using isMobile and negated the value of the hook (was quicker and less error prone than changing reversing all the logic.

I also used `'md'` to match what it was before, but I'm open to changing this.

madsnedergaard (approved):

PR #6595, opened 2024-03-30
refactor(us-pjm): remove usage of arrow from US-PJM parser

amv213:
## Issue

## Description

This PR removes the dependency on `arrow` from the `US-PJM` parser.

### Double check

    - `US-PJM` gave a `KeyError`, but `US-MIDA-PJM` worked. I am not sure if they are the same thing.


parsers/US_PJM.py
```diff
@@ -167,7 +176,7 @@ def fetch_production(
                     production[mode] = row.get("mw")
             data_point = {
                 "zoneKey": zone_key,
-                "datetime": arrow.get(dt).replace(tzinfo="US/Eastern").datetime,
+                "datetime": dt.to_pydatetime().replace(tzinfo=TIMEZONE),
```
amv213, line 179: note: "US/Eastern" has been [renamed](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) to "America/New_York"

parsers/US_PJM.py
```diff
@@ -220,18 +220,18 @@ def get_miso_exchange(session: Session) -> tuple:
         raise ValueError("US-MISO->US-PJM flow direction cannot be determined.")
 
     find_timestamp = soup.find("div", {"id": "body_0_divTimeStamp"})
-    dt_naive = parser.parse(find_timestamp.text)
-    dt_aware = add_default_tz(dt_naive)
+    dt = parser.parse(find_timestamp.text, tzinfos={"EDT": -4 * 3600, "EST": -5 * 3600})
+    if _is_naive(dt):
+        dt = dt.replace(tzinfo=TIMEZONE)
```
amv213, line 225: note: `dt_naive` wasn't really naive as it includes EDT / EST information. 

I was getting
```
UnknownTimezoneWarning: tzname EDT identified but not understood. 
Pass `tzinfos` argument in order to correctly return a timezone-aware datetime.
In a future version, this raise an exception.
category=UnknownTimezoneWarning)
```

The new code should make sure that those are mapped to known utc offsets. The old code was remapping everything to "America/New_York" so I think it could have potentially been assigning the wrong UTC offset if e.g. the timestamp was marked as EDT and New York was observing EST (although I imagine that in practice - at least for live data - those would always be in sync)

Just in case there are really naive timestamps coming through I have kept the fallback to make sure those would get assigned the correct timezone.

parsers/US_PJM.py
```diff
@@ -261,13 +261,15 @@ def get_exchange_data(interface, session: Session) -> list:
...
+    date = datetime.combine(datetime.now(TIMEZONE), time())  # truncate to day
 
     converted_flows = []
     for flow in flows:
-        arr_time = arrow.get(flow[1], "h:mm A")
-        arr_dt = arr_date.replace(hour=arr_time.hour, minute=arr_time.minute).datetime
-        converted_flow = (flow[0], arr_dt)
+        time_of_the_day = datetime.strptime(flow[1], "%I:%M %p").replace(
+            tzinfo=TIMEZONE
+        )
```
amv213, line 270: note: make sure to use `%I` and not `%H` or `%p` wouldn't take effect

parsers/US_PJM.py
```diff
@@ -261,13 +261,15 @@ def get_exchange_data(interface, session: Session) -> list:
...
+        dt = date.replace(hour=time_of_the_day.hour, minute=time_of_the_day.minute)
+        converted_flow = (flow[0], dt)
```
amv213, line 272: note: I couldn't test this because this function is not working for me (I think the API is gone)

parsers/US_PJM.py
```diff
@@ -261,13 +261,15 @@ def get_exchange_data(interface, session: Session) -> list:
 
     flows = zip(actual_load, time_vals, strict=True)
 
-    arr_date = arrow.now("America/New_York").floor("day")
+    date = datetime.combine(datetime.now(TIMEZONE), time())  # truncate to day
```
VIKTORVAV99, line 264: This is a neat trick instead of using replace.

VIKTORVAV99 (commented):
Looks good! Just some small notes.

VIKTORVAV99 (approved):

PR #6596, opened 2024-03-30
refactor(us-prepa): remove usage of arrow from US-PREPA parser

amv213:
## Issue

## Description

This PR removes the dependency on arrow from the US-PREPA parser.

### Double check

- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"` (KeyError)

### DIFF
--- a/parsers/US_PREPA.py
+++ b/parsers/US_PREPA.py
@@ -12,12 +12,12 @@ import json
 import re
 from datetime import datetime
 from logging import Logger, getLogger
+from zoneinfo import ZoneInfo
 
-# The arrow library is used to handle datetimes
-import arrow
 from requests import Session
 
-timezone_name = "America/Puerto_Rico"
+TIMEZONE = ZoneInfo("America/Puerto_Rico")
+
 US_PROXY = "https://us-ca-proxy-jfnx5klx2a-uw.a.run.app"
 HOST_PARAMETER = "?host=https://aeepr.com"
 GENERATION_BREAKDOWN_URL = (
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Unfortunately this parser is no longer working for `PR`, there is however new parsers that need to be finished for one of them to be merged. I'll see if I can get that sorted during this weekend.

If you still want to merge this that is fine by me as it's not working right now anyway but once again unfortunately your changes won't make a difference. Sorry.

amv213:
Ah no worries - thanks for the heads up @VIKTORVAV99!

Happy to merge while we are at it -  tbh I am just randomly going through the parsers where I see `arrow` being used 😅 

Out of curiosity, what's the best way to figure out if a parser is being used? Should I move them somewhere if I stumble on them? Also, is there a way to figure out that e.g. here the zone key for the parser was `PR`? Thanks!

VIKTORVAV99:
> Ah no worries - thanks for the heads up @VIKTORVAV99!
> 
> Happy to merge while we are at it - tbh I am just randomly going through the parsers where I see `arrow` being used 😅
> 
> Out of curiosity, what's the best way to figure out if a parser is being used? Should I move them somewhere if I stumble on them? Also, is there a way to figure out that e.g. here the zone key for the parser was `PR`? Thanks!

The easiest way is to search for the filename (without .py) in the config files (a global search will make things way easier) and then see which exchanges and zones use the parser.

Unfortunately we don't have a better way to do it right now.

VIKTORVAV99:
As for moving the parsers if they are not used anywhere (no references at all) you can move them to `parsers/archived`.

VIKTORVAV99 (approved):

amv213:
Perfect! Thanks @VIKTORVAV99 🔥

PR #6610, opened 2024-04-01
refactor(pa): upgrade PA parser with event classes

amv213:
## Issue

## Description

This PR upgrades the PA parser to use event classes, and drops usage of arrow in favour of the datetime library.

### Double check


parsers/PA.py
```diff
@@ -97,33 +110,67 @@
...
+def _localise_spanish_date(date: str) -> str:
+    """Localises a date containing full name (lowercase) spanish months, replacing them with zero-padded decimal numbers.
+
+    This avoids having to mess up the global locale to be able to parse the date.
+    """
+    return re.sub(
+        "|".join(_SPANISH_CALENDAR.keys()),
+        lambda m: _SPANISH_CALENDAR[m.group(0)],
+        date,
+    )
```
amv213, line 146: note: the production API returns 'spanish_dates' like `'31-marzo-2024 5:26:36'`

`arrow` had a neat one-liner to parse them:
`date = arrow.get(spanish_date, "DD-MMMM-YYYY H:mm:ss", locale="es", tzinfo="America/Panama")`

but unfortunately `datetime.strptime()` doesn't have a similar way to provide a locale and we would be forced to set it globally for it to work - [which is problematic](https://stackoverflow.com/a/32785195)

This function "translates" the month to its corresponding number so that then it can be parsed in a locale-independent way

VIKTORVAV99 (approved):
LGTM!

PR #6612, opened 2024-04-01
refactor(cr): upgrade CR parser with event classes

amv213:
## Issue

## Description

This PR upgrades the CR parser to use event classes.

### Double check


parsers/test/test_CR.py
```diff
@@ -0,0 +1,122 @@
...
+                    "zoneKey": element["zoneKey"],
+                    "production": element["production"],
+                    "storage": element["storage"],
+                    "source": element["source"],
+                    "sourceType": element["sourceType"].value,
+                    "correctedModes": element["correctedModes"],
+                }
+                for element in production
+            ]
+        )
```
VIKTORVAV99, line 50: FYI I found a "better" snapshot library that both fixes the enum issue and uses pytest style assertions.

We won't be able to use it just yet but hopefully soon.

amv213: Nice find @VIKTORVAV99 - looks really good! 

I have actually never used snapshot testing libraries before, but from a first look I think it would definitely be worth (thinking of) switching to `syrupy`. If anything, just the fact that it is actively maintained would be a very big plus and I think would prevent bad surprises / major blockers down the line. In addition, as you said, it also seems to be able to do everything that `snapshottest` can do... but "better" 😅  

The docs do say though that the two packages cannot coexist in the same environment because of argument conflicts and overlapping fixtures. Did you already have in mind a possible migration path? 

If you are not too keen for a big-bang migration, one possibility could be to initially keep using `snapshottest` and slowly migrate the existing tests to use `pytest`-style tests and`snapshottest`'s `snapshot` fixture / `snapshot.assert_match()`. This would gradually drop all explicit references to `snapshottest`, which we currently have just to be able to use `snapshottest.TestCase`. Once it's all done, I think it should then be possible to simply swap the `snapshottest` package with `syrupy`, and all tests would still work out-of-the-box: the `snapshot` fixture would this time be the one provided by `syrupy`, and the docs claim that it would understand `snapshottest`'s `snapshot.assert_match(actual)` statement (although it would natively prefer `snapshot == actual`)

VIKTORVAV99: Honestly I was just going to do one big PR that does it all. All snapshots would change anyway due to a different serialization process for the date times anyway. Since we don’t need to explicitly loop over them and there would be no chance to call isoformat anymore. But perhaps it would just be better to do it incrementally as you mention.

I’m also going to see if we can change it one file at a time without conflicts to see what our options are in regards to that. If that’s an alternative migration would be easy.

I’m going to have a chat with the people in the engineering team tomorrow and see what they say.

parsers/CR.py
```diff
@@ -30,161 +42,133 @@
...
-        "capacity": {},
-        "production": {},
-        "storage": {},
-        "source": "grupoice.com",
-    }
-
-
-def extract_exchange(raw_data, exchange) -> float | None:
+def _parse_exchange_data(
+    exchange_data: list[dict], sorted_zone_keys: str
```
VIKTORVAV99, line 46:
```suggestion
    exchange_data: list[dict], sorted_zone_keys: ZoneKey
```

parsers/CR.py
```diff
@@ -30,161 +42,133 @@
...
-    if target_datetime:
-        raise NotImplementedError("This parser is not yet able to parse past dates")
-
-    sorted_zones = "->".join(sorted([zone_key1, zone_key2]))
-
-    if sorted_zones not in EXCHANGE_JSON_MAPPING:
-        raise NotImplementedError("This exchange is not implemented.")
 
-    s = session or Session()
+    sorted_zone_keys = "->".join(sorted([zone_key1, zone_key2]))
```
VIKTORVAV99, line 138:
```suggestion
    sorted_zone_keys = ZoneKey("->".join(sorted([zone_key1, zone_key2])))
```

parsers/CR.py
```diff
@@ -30,161 +42,133 @@
...
-        "datetime": dt.datetime,
-        "netFlow": flow,
-        "source": "enteoperador.org",
-    }
+    net_flow = _parse_exchange_data(response.json(), sorted_zone_keys=sorted_zone_keys)
 
-    return exchange
+    exchange_list = ExchangeList(logger)
+    exchange_list.append(
+        zoneKey=ZoneKey(sorted_zone_keys),
```
VIKTORVAV99, line 165:
```suggestion
        zoneKey=sorted_zone_keys,
```

VIKTORVAV99 (approved):
LGTM!

I added some small suggestions that should be looked at first before merging though.

PR #6614, opened 2024-04-02
refactor(opennem) upgrade OPENNEM parser with event classes and replace arrow

amv213:
## Issue

## Description

This PR upgrades the OPENNEM parser to use event classes, and drops usage of `arrow` in favour of the `datetime` library.

### Double check

### DIFF
--- a/parsers/OPENNEM.py
+++ b/parsers/OPENNEM.py
@@ -2,7 +2,6 @@ from collections.abc import Mapping
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
 
-import arrow
 import pandas as pd
 import requests
 from requests import Session
@@ -66,8 +65,8 @@ SOURCE = "opennem.org.au"
 def dataset_to_df(dataset):
     series = dataset["history"]
     interval = series["interval"]
-    dt_start = arrow.get(series["start"]).datetime
-    dt_end = arrow.get(series["last"]).datetime
+    dt_start = datetime.fromisoformat(series["start"])
+    dt_end = datetime.fromisoformat(series["last"])
     data_type = dataset["data_type"]
     _id = dataset.get("id")
 
@@ -151,7 +150,7 @@ def generate_url(
     if target_datetime:
         network = ZONE_KEY_TO_NETWORK[zone_key]
         # We will fetch since the beginning of the current month
-        month = arrow.get(target_datetime).floor("month").format("YYYY-MM-DD")
+        month = target_datetime.strftime("%Y-%m-%d")
         if is_flow:
             url = (
                 f"http://api.opennem.org.au/stats/flow/network/{network}?month={month}"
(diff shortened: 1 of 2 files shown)


parsers/OPENNEM.py
```diff
@@ -256,156 +245,147 @@ def filter_dataset(ds: dict) -> bool:
...
-                "hydro": sum_vector(capacities, OPENNEM_PRODUCTION_CATEGORIES["hydro"]),
-                "wind": sum_vector(capacities, OPENNEM_PRODUCTION_CATEGORIES["wind"]),
-                "biomass": sum_vector(
-                    capacities, OPENNEM_PRODUCTION_CATEGORIES["biomass"]
-                ),
-                "solar": sum_vector(capacities, OPENNEM_PRODUCTION_CATEGORIES["solar"]),
-                "hydro storage": capacities.get(OPENNEM_STORAGE_CATEGORIES["hydro"][0]),
-                "battery storage": capacities.get(
-                    OPENNEM_STORAGE_CATEGORIES["battery"][0]
-                ),
```
amv213, line 317: question: @VIKTORVAV99 what is the recommended way to deal with this capacity data given that it cannot be captured in a `ProductionMix`? 

I see references to `OPENNEM.fetch_production_capacity` in some configs but I can't seem to find it. Is that a function that I need to write or was it dynamically generated from something else?

VIKTORVAV99: So this is a small mess right now and I don't really know why but the capacity parsers live in an entirely different folder.

And yes as you mention there is no way to currently return the live capacity data. Honestly I'd leave it as a draft for now until I have discussed with the team on how to handle this.

But I'll probably end up creating a capacity mix class or something similar so we can keep parsing this data.

amv213: Perfect! Thanks

VIKTORVAV99:
Just adding this as a suggestion here but it's by no means required. If you want you can split this PR up by removing arrow in one and migrating it to parser classes in the other while we figure out how to handle live capacities with the new system.

PS: You can also add a PR that adds you as a contributor to all to zones you have helped upgrade. It's optional but you deserve the credit!
(It's done in the zone configs).

parsers/OPENNEM.py
VIKTORVAV99: So I have been trying to figure out to do this, what I thought was going to be pretty straight forward is not as the space in the "hydro storage" and "battery storage" modes. This means they can't be used as variables which is kind of needed for it to plug into the existing solution we have. All workarounds I have create more errors than I can count 😅

There is a TODO about removing that space and simplifying things a bit and I think I'll have to do that first before working on the "capacity mixes".

amv213: No worries!! Thanks for the update

VIKTORVAV99:
Closing this down as it's no longer relevant after some major refactoring of the OPENNEM parser.

PR #6620, opened 2024-04-03
refactor(gt): remove usage of arrow from GT parser

amv213:
## Issue

## Description

This PR removes the dependency on `arrow` from the GT parser.

### Double check

### DIFF
--- a/parsers/GT.py
+++ b/parsers/GT.py
@@ -4,21 +4,21 @@
 Mayorista (AMM) API.
 """
 
-# Standard library imports
 import collections
-from datetime import datetime
+from datetime import datetime, time, timedelta
 from logging import Logger, getLogger
+from typing import Literal
+from zoneinfo import ZoneInfo
 
-# Third-party library imports
-import arrow
 from requests import Session
 
-# Local library imports
 from parsers.lib import validation
 
 DEFAULT_ZONE_KEY = "GT"
+TIMEZONE = ZoneInfo("America/Guatemala")
+
 PRODUCTION_THRESHOLD = 10  # MW
-TIMEZONE = "America/Guatemala"
+
 DOMAIN = "wl12.amm.org.gt"
 URL = f"https://{DOMAIN}/GraficaPW/graficaCombustible"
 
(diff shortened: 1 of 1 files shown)


parsers/GT.py
```diff
@@ -95,36 +106,51 @@ def fetch_production(
     ]
 
 
-def index_api_data_by_hour(json):
-    """The JSON data returned by the API is a list of objects, each
-    representing one technology type. Collect this information into a list,
-    with the list index representing the hour of day.
-    """
+def _get_api_data(
+    session: Session,
```
VIKTORVAV99, line 110: Since you are changing this anyway `Session` here should be `Session | None`.

parsers/GT.py
```diff
@@ -32,21 +32,28 @@ def fetch_consumption(
...
+    api_data = _get_api_data(session, URL, target_datetime, target="consumption")
+
+    day_datetime = datetime.combine(
+        target_datetime, time(), tzinfo=TIMEZONE
+    )  # truncate to day
     results = [
         {
             "consumption": row["DEM SNI"],
-            "datetime": date_time.replace(hour=hour).datetime,
+            "datetime": day_datetime + timedelta(hours=hour),
```
VIKTORVAV99, line 50: Any reason you went with timedelta instead of replace?

parsers/GT.py
```diff
@@ -95,36 +106,51 @@ def fetch_production(
...
-def index_api_data_by_hour(json):
-    """The JSON data returned by the API is a list of objects, each
-    representing one technology type. Collect this information into a list,
-    with the list index representing the hour of day.
-    """
+def _get_api_data(
+    session: Session | None,
+    url: str,
+    date_time: datetime,
+    target: Literal["consumption", "production"],
```
VIKTORVAV99, line 113: I would probably use a string enum instead to reduce the risk of miss spellings in the future. It can also be used as a type and has type hinting so you don't need the literal.

VIKTORVAV99 (commented):
Just some notes before approving but seems to work great!

parsers/GT.py
amv213: I mostly just thought it expressed a bit more intuitively the logic.

The previous code was flooring the date_time to "hour" and then replacing that hour with those returned by the API, which confused me slightly. I thought it would be slightly more intuitive to start with today's day and iteratively step through hour by hour.

VIKTORVAV99 (approved):
LGTM!

Thanks for fixing it!

PR #6621, opened 2024-04-03
refactor(md): remove usage of arrow from MD parser

amv213:
## Issue

## Description

This PR removes the dependency on `arrow` from the MD parser.

### Double check


parsers/MD.py
```diff
@@ -147,21 +146,17 @@ def get_archive_data(session: Session | None = None, dates=None) -> list:
 
     archive_url = archive_base_url
     if date1 and date2:
```
VIKTORVAV99, line 148: Well this is weird.

The function expects 2 dates in a tuple or list but it's ever only called with 1...

Not only should this never happen IMO but it's just causing the one date to be passed to be duplicated...

However the date passed itself is a datetime object, not a string which is causing `.fromisoformat()` to fail.

You should be able to change it to just accept one date in the function and then call fromisoformat directly on it.

amv213: Oh wow... thanks for catching that! The function calls in `__main__` feeding in strings threw me off!

Should hopefully all be sorted now

parsers/MD.py
```diff
@@ -129,43 +122,26 @@ def template_exchange_response(
...
             ArchiveDatapoint(
-                arrow.get(
-                    entry[0], archive_datetime_format, tzinfo="Europe/Chisinau"
-                ).datetime,
+                datetime.strptime(entry[0], "%Y-%m-%d %H:%M").replace(tzinfo=TZ),
                 *map(float, entry[1:]),
             )
             for entry in data
         ]
+        return sorted(archive_datapoints, key=lambda x: x.datetime)
```
VIKTORVAV99, line 144: Using `itemgetter` instead of lambda is both simpler and faster (there is other benefits as well but are not relevant in our case). So I would recomend that.

But does this need to be sorted?

VIKTORVAV99 (commented):
I can't seem to test this at the moment as the source is done but left a comment, will revisit it later.

parsers/MD.py
amv213: TIL - thanks @VIKTORVAV99 !

The API is/was returning data in reverse chronological order, so production and consumption data are also returned unsorted (at least until we update to event classes). I wasn't too sure of how the backend handles these, but just in case I thought it would be good to sort things properly to limit surprises / keep things intuitive

amv213:
I can't seem to be getting any data anymore either (same with the code on `master` though) 👀

parsers/MD.py
VIKTORVAV99: The backend don't care at all which order things are returned in, this is for us humans.

VIKTORVAV99:
> I can't seem to be getting any data anymore either (same with the code on `master` though) 👀

Noticed the same thing, will try and dig into it when I have more time but things could have changed at the source.

amv213:
I have updated the code to get it working again but no idea why things changed so drastically API-side between now and when I first opened the PR 👀 Definitely worth having a good look over to make sure everything is still correct

parsers/MD.py
```diff
@@ -251,8 +228,7 @@ def fetch_production(
 ) -> list[dict] | dict:
     """Requests the production mix (in MW) of a given country."""
     if target_datetime:
-        archive_data = get_archive_data(session, target_datetime)
-
+        archive_data = get_archive_data(target_datetime, session=session)
         datapoints = []
```
VIKTORVAV99, line 232: Would it be possible for you to make it only return the archive data?

I noticed the live production data also just grabs the current timestamp which is something we want to avoid if possible.

Otherwise this seems to work as expected.

parsers/MD.py
```diff
@@ -280,30 +256,14 @@ def fetch_production(
         field_values = get_data(session)
 
         production = {
-            "solar": None,
-            "wind": None,
-            "biomass": 0.0,
+            "solar": field_values[FIELD_LABEL_to_INDEX["solar"]],
+            "wind": field_values[FIELD_LABEL_to_INDEX["wind"]],
+            "biomass": field_values[FIELD_LABEL_to_INDEX["biomass"]],
             "nuclear": 0.0,
```
VIKTORVAV99, line 262: Could possibly remove this nuclear 0 as well, we generally only want to return the data that are reported by the API, not assign static values.

parsers/MD.py
amv213: Sure! The only downside I can think of is that we would lose the more granular information about the solar / wind / biomass distribution, which seems to only be available on the live API. Happy to do that anyway?

amv213: Perfect- will do. Should I use only `get_archive_data` in all of the other functions too?

VIKTORVAV99: If possible then yes. We really want data that is mapped to an actual datetime and not just the time when the parser runs.

amv213:
Thank you for all of your comments @VIKTORVAV99! Ready for another review.

I wasn't sure if there are existing guidelines around how much data should be returned from these kind of APIs which can theoretically return data for arbitrary time windows. At the moment I've set it to return one day of data for the day of the requested target datetime. Do not hesitate to let me know if that needs changing. For example, maybe you prefer one day of data but centered around the target datetime? i.e. target_datetime +- 12h instead of 24h day(target_datetime)

VIKTORVAV99:
> Thank you for all of your comments @VIKTORVAV99! Ready for another review.
> 
> I wasn't sure if there are existing guidelines around how much data should be returned from these kind of APIs which can theoretically return data for arbitrary time windows. At the moment I've set it to return one day of data for the day of the requested target datetime. Do not hesitate to let me know if that needs changing. For example, maybe you prefer one day of data but centered around the target datetime? i.e. target_datetime +- 12h instead of 24h day(target_datetime)

There is no set guidelines as far as I know but we like overlapping data and the flowtracing pipeline always runs on the last 2 days so if we can return the last 2 days of data that would be great. Then we can also set the refetch frequency to 2 days so the internal tools knows how many requests to make when refetching data manually.

Thanks for doing this!

parsers/MD.py
```diff
@@ -210,107 +164,62 @@ def fetch_price(
...
+    datapoints = []
+    for entry in archive_data:
+        datapoint = template_consumption_response(
+            zone_key, entry.datetime, entry.consumption
+        )
+        datapoints.append(datapoint)
+    return datapoints
 
 
+@refetch_frequency(timedelta(days=2))
```
amv213, line 190: question: would this refetch once every two days or twice a day?

VIKTORVAV99: every 20 minutes ;)

This is just so that when we manually refetch large sections of data we create one request for every 2 days worth of data instead of 1 per hour or whatever the default is.

VIKTORVAV99 (approved):
LGTM! 

Thanks for fixing it and cleaning up some of the code debt!

PR #6623, opened 2024-04-03
feat(pa): use json APIs for PA parser

amv213:
## Description

This PR updates the `PA` parser to use directly the backend json APIs rather than having to scrape the frontend HTML pages for data. 

This should hopefully make the code more readable / easier to test.

### Double check


parsers/PA.py
```diff
@@ -313,44 +245,25 @@ def fetch_exchange(
...
-        "NI->PA": net_flow_ni,  # Nicaragua to Panama
-        "PA->SV": net_flow_sv,  # Panama to El Salvador
-    }
+        # insert net flow taking into account alphabetical sorting of zone keys
+        sorted_exchange_key = ZoneKey("->".join(sorted([country, "PA"])))
+        net_flows[sorted_exchange_key] = (
+            net_flow_country
+            if sorted_exchange_key.endswith("PA")
+            else -net_flow_country
+        )
```
amv213, line 266: issue(maybe): not an issue for this PR (also because I don't think that there is any exchange using this parser) but I don't think that this function is / was calculating the correct things (unless I am understanding things wrong 😅 )

Looking at https://sitr.cnd.com.pa/m/pub/int.html, the flows are between neighbouring countries, not all to `PA`. Also if we tried `fetch_exchange()` on any other combination than "CR->PA" it indeed fails saying that those are not valid exchanges.

Should I just remove this function? If not also happy to try fixing it in another PR

parsers/PA.py
```diff
@@ -211,69 +175,37 @@ def fetch_production(
...
+    production_units = response_payload["unit"]
+    for production_unit in production_units:
+        unit_name = production_unit["name"]
+        unit_generation = float(production_unit["value"])
+        if (
+            unit_name in MAP_THERMAL_GENERATION_UNIT_NAME_TO_FUEL_TYPE
+            and unit_generation >= 0
+        ):  # Ignore self-consumption
+            unit_fuel_type = MAP_THERMAL_GENERATION_UNIT_NAME_TO_FUEL_TYPE[unit_name]
+            production_mix.add_value(unit_fuel_type, unit_generation)
```
amv213, line 208: note: maybe the only downside is that now we cannot know dynamically which plants are "thermic" plants (because there is no way to know from the json which units are displayed on the webpage under the "Termicas" table). So if a new unknown fossil plant was to be built we wouldn't sum it up / get a warning. So far though we "know" of all existing plants 🙌🏽

parsers/test/mocks/PA/production_live_20211230.json
amv213, line 1: note: this is exactly the same data that was in `PA_nominal_generation.html`, which I have ported by hand here to replicate what would have been the corresponding json API response that backed it up. In this way we can ensure that the new code gives exactly the same results as those historical results.

parsers/test/test_PA.py
```diff
@@ -15,30 +17,94 @@ def setUp(self):
         self.session = requests.Session()
         self.session.mount("https://", self.adapter)
 
-    @freeze_time("2021-12-30 09:58:40", tz_offset=-5)
-    def test_fetch_production(self):
+    def test_fetch_production_live(self):
+        for frozen_time, mock_json_response in [
+            ("2021-12-30 09:58:40", "production_live_20211230.json"),
+            ("2024-04-03 02:43:25", "production_live.json"),
+        ]:
```
amv213, line 24: note: testing against a new recent snapshot but also the historical snapshot to make sure we didn't break anything / code is fully backwards compatible (e.g. can handle legacy fossil fuel plants that may no longer exist)

parsers/PA.py
VIKTORVAV99: Since this is unused it would be fine to remove it but if you want to fix it that would also be great 🙂

Totally up to you though IMO.

parsers/PA.py
amv213: Unfortunately the data is coming down like this https://sitr.cnd.com.pa/m/pub/data/gen.json , where `"units"` is just a long list of energy plant names and associated production - without any information about which plants are "Thermica", which are solar, which are hydroelectric, etc...

The grouping into production modes is introduced only at the frontend level https://sitr.cnd.com.pa/m/pub/gen.html (couldn't quite figure out if they had a dynamic way to render that or if essentially they are hard coding which plant goes into which table) so I fear that there is not really a way to be able to know which plants are thermic without either "knowing it" or scraping the html page...

Happy to revert if risking to miss out on new plants is a blocker!

parsers/test/test_PA.py
amv213: Amazing!! I'll definitely try that in a follow up PR (just to not stack too many changes here)

unitrium:
Closing this PR as it has been stale for 2years and now there are conflicts with the current implementation of the parser.

PR #6625, opened 2024-04-03
Update US balancing authority naming

### LINKED ISSUE #5837: How should we name balancing authorities in USA? Feedback needed! 🙏

madsnedergaard:
## Issue

Closes #5837, follow-up to #6498

## Description

This updates a lot of `zoneName` to remove "Inc" and similar, as well as adding a bunch of `displayName` properties with nicer names.

**This should only be merged after #6498 has been released and tested!**


madsnedergaard:
This is good to go now, ERCOT looks fine :)
https://app.electricitymaps.com/zone/US-TEX-ERCO

VIKTORVAV99 (approved):
Question here, do we want to remove the translations from the other languages or just leave them for now?

Otherwise it looks good to me!

madsnedergaard:
I'd say let's get this merged now, and then at some point in the future we should do a more thorough cleaning up of all translation files 😅

VIKTORVAV99:
> I'd say let's get this merged now, and then at some point in the future we should do a more thorough cleaning up of all translation files 😅 

I have a plan for that! Not enough to be its own project just yet but I'll restructure a few things at one point so we can use already available tooling.

PR #6626, opened 2024-04-03
EIA capacity update for 2024

VIKTORVAV99:
## Issue

We are missing EIA 2024 capacities.

## Description

Updates all EIA zones with 2024 capacities.

### Double check


config/zones/US-NW-BPAT.yaml
KabelWlan, line 1: If you're already making changes to the capacities. I would change the wind capacity as the generation from the AVRN zone is included in the BPAT zone. I would therefore change the wind capacity to 5314.8

config/zones/US-NW-NWMT.yaml
```diff
@@ -52,6 +55,9 @@ capacity:
     - datetime: '2023-10-01'
       source: EIA.gov
       value: 452.9
+    - datetime: '2024-01-01'
+      source: EIA.gov
+      value: 763.6
```
KabelWlan, line 60: I would use the aggregated number for NWMT, GWA and WWA as you aggregate generation for those zones. Furthermore, the Clearwater Wind 1 wind farm is in the EIA report wrongly accounted to PSEI. The generation is shown in the NWMT zone. The aggregated number would thus be 1528.1

KabelWlan (commented):
Two comments regarding zones that you aggregate generation. I think you should thus aggregate the capacities as well.

config/zones/US-NW-BPAT.yaml
VIKTORVAV99: I honestly thought it was already doing that...

Since we are now using a parser for updating these values we want to avoid manual changes as much as possible. However I think fixing the parser itself is out of scope for this PR but I'll open a issue about it so it can be adressed later.

pierresegonne:
Looks good, can you confirm that you've reallocated capacity for zones that are aggregated into others?

VIKTORVAV99:
> Looks good, can you confirm that you've reallocated capacity for zones that are aggregated into others?

I have not, that would require parser changes. I have only run the parser itself in this PR.

PR #6630, opened 2024-04-04
ENTE parser: remove usage of arrow module

gianantoniopini:
## Issue

#6135 

## Description

This PR removes usage of the `arrow` module from the ENTE parser.
To help prevent regressions, some unit tests have also been added for this parser.

### Double check


parsers/ENTE.py
```diff
@@ -25,6 +27,10 @@
 }
 
 
+def floor_to_minute(dt: datetime) -> datetime:
+    return dt.replace(second=0, microsecond=0)
+
+
 def fetch_production(
```
gianantoniopini, line 34: I've noticed that when running `poetry run test_parser HN production` from the command line, the `fetch_production` function of the **HN** parser is called, not the ENTE parser. So I'm wondering if the `fetch_production` funtion of the ENTE parser is actually used or needed at all?

VIKTORVAV99: Looks like ENTE is only used for exchanges at the moment so no it would not be needed. I would not recommend removing parser functions but since this only outputs unknown data I don't think we would use it anyhow as it is now. But if it's not causing any larger issues I'd say leave it.

gianantoniopini: Thank you for the clarification. I will leave it as it is then, as this is not a big issue.

parsers/ENTE.py
```diff
@@ -25,6 +27,10 @@
 }
 
 
+def floor_to_minute(dt: datetime) -> datetime:
+    return dt.replace(second=0, microsecond=0)
```
VIKTORVAV99, line 31: This is pretty nice, maybe we should create a common set of helper functions for this in the future.

VIKTORVAV99 (approved):
LGTM!

And thanks for adding the snapshots!

parsers/ENTE.py
gianantoniopini: Yes, if other parsers are (or will) perform this logic, then it would make sense to move this `floor_to_minute` function to a common shared module. There is already a `parsers/lib/utils.py` file, which might be a good candidate to move this function to.

PR #6634, opened 2024-04-04
feat(NL): Use NED.nl as source for solar data

VIKTORVAV99:
## Issue

ENTSO-E don't have any good solar data for NL.

## Description

This PR adds a parser that gets the solar values from NED.nl and the production data from ENTSO-E, it then updates the ENTSO-E data with the data from NED.nl, it does this for both production and forecasts.

>[!IMPORTANT]
>We need to change the estimation method for NL in conjunction with merging this and then refetch and reprocess all the historical data.

>[!WARNING]
>We need to add a API KEY for NED before this is merged!

### Double check

### DIFF
--- a/DATA_SOURCES.md
+++ b/DATA_SOURCES.md
@@ -141,10 +141,10 @@ Real-time electricity data is obtained using [parsers](https://github.com/electr
   - California (US-CAL-CISO): [CAISO](http://www.caiso.com/Pages/default.aspx)
   - Hawaii, Oahu Island (US-HI-OA): [Islandpulse](https://www.islandpulse.org/)
   - New England (US-NE-ISNE): [NEISO](https://www.iso-ne.com/isoexpress/)
-  - New York (US-NY-NYIS): [NYISO](http://www.nyiso.com/public/markets_operations/market_data/graphs/index.jsp)
+  - New York (US-NY-NYIS): [NYISO](https://www.nyiso.com/real-time-dashboard)
   - PJM (US-MIDA-PJM): [PJM](http://www.pjm.com/markets-and-operations.aspx)
-  - Puerto Rico (PR): [AEEPR](https://aeepr.com/en-us/Pages/Generaci%C3%B3n.aspx)
-  - Southwest Power Pool (US-CENT-SWPP): [SPP](https://marketplace.spp.org/pages/generation-mix)
+  - Puerto Rico (PR): [AEEPR](https://aeepr.com/#/operacion)
+  - Southwest Power Pool (US-CENT-SWPP): [SPP](https://portal.spp.org/integrated-marketplace)
 - Uruguay: [UTE](https://apps.ute.com.uy/SgePublico/ConsPotenciaGeneracionArbolXFuente.aspx)
 - Vietnam: [EVN](https://www.nldc.evn.vn/)
   &nbsp;</details>
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "electricitymaps-web",
-  "version": "1.128.0",
+  "version": "1.129.0",
   "description": "A real-time visualisation of the CO2 emissions of electricity consumption",
   "repository": {
     "type": "git",
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -122,7 +122,7 @@
   "estimation-card": {
     "link": "Read about our estimation models",
     "outage-details": "See details",
-    "estimated_forecast_hierarchy": {
+    "estimated_forecasts_hierarchy": {
       "title": "Data is estimated",
       "pill": "Estimated",
       "body": "The data for this zone is estimated using a combination of internal and external production forecasts and Time Slicer Averages. This aims to provide a more accurate representation than Time Slicer Average alone but the data may still be subject to inaccuracies. The data will be updated as soon as real-time data becomes available."
--- a/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
+++ b/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
@@ -114,6 +114,7 @@ function BarBreakdownChart({
       <BySource
         hasEstimationPill={hasEstimationPill}
         estimatedPercentage={currentZoneDetail.estimatedPercentage}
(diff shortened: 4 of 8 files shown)


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -302,20 +302,36 @@ def update_production_breakdowns(
...
-            else:
-                production_breakdowns.append(
+                updated_production_breakdowns.append(
+                    updated_event.zoneKey,
+                    updated_event.datetime,
+                    updated_event.source,
+                    updated_event.production,
+                    updated_event.storage,
+                    updated_event.sourceType,
+                )
```
VIKTORVAV99, line 332: This will be simplified in another PR and abstracted to a helper function.

electricitymap/contrib/lib/models/events.py
```diff
@@ -655,20 +655,19 @@ def _update(
...
-                f"Cannot update events from different sources: {event.source} and {new_event.source}"
-            )
         if event.sourceType != new_event.sourceType:
             raise ValueError(
                 f"Cannot update events from different source types: {event.sourceType} and {new_event.sourceType}"
             )
         production_mix = ProductionMix._update(event.production, new_event.production)
         storage_mix = StorageMix._update(event.storage, new_event.storage)
+        source = ", ".join(
+            set(event.source.split(", ")) | set(new_event.source.split(", "))
```
VIKTORVAV99, line 665: This ensures that if we where to update and already updated list that now contain two sources these are deduplicated.

parsers/NED.py
```diff
@@ -0,0 +1,224 @@
...
+        mix = ProductionMix()
+        for data in data_dict:
+            clean_type = int(data["type"].split("/")[-1])
+            if clean_type in TYPE_MAPPING:
+                mix.add_value(
+                    TYPE_MAPPING[clean_type],
+                    _kwh_to_mw(data["volume"]),
+                )
+            else:
+                logger.warning(f"Unknown type: {clean_type}")
```
VIKTORVAV99, line 133: This is a bit overkill for just solar but makes it easy to add other modes as well in the future, like wind for example.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -302,20 +302,36 @@ def update_production_breakdowns(
         production_breakdowns: "ProductionBreakdownList",
         new_production_breakdowns: "ProductionBreakdownList",
         logger: Logger,
+        matching_timestamps_only: bool = False,
```
pierresegonne, line 305: Maybe worth adding a comment at this level to explain what this parameter does?

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -324,7 +340,19 @@ def update_production_breakdowns(
                     new_event.sourceType,
                 )
 
-        return production_breakdowns
+        if matching_timestamps_only is False:
+            for existing_event in production_breakdowns.events:
```
pierresegonne, line 344: Same here, could be nice to have an helper function for that

parsers/NED.py
```diff
@@ -0,0 +1,224 @@
...
+        else NedClassification.MEASURED.value,
+        "activity": NedActivity.PRODUCTION.value,
+        "validfrom[before]": (target_datetime + timedelta(days=2 if forecast else 1))
+        .date()
+        .isoformat(),
+        "validfrom[after]": (target_datetime - timedelta(days=0 if forecast else 1))
+        .date()
+        .isoformat(),
+    }
+    headers = {"X-AUTH-TOKEN": get_token("NED_KEY"), "accept": "application/json"}
```
pierresegonne, line 90: We're using `${SOURCE}_TOKEN` for all (except the EIA...)

parsers/NED.py
```diff
@@ -0,0 +1,224 @@
...
+    response = requests.get(URL, params=params, headers=headers)
+    if not response.ok:
+        raise ParserException(
+            parser="NED.py",
+            message=f"Failed to fetch NED data: {response.status_code}, err: {response.text}",
+        )
+    return response.json()
+
+
+def format_data(json: Any, logger: Logger, forecast: bool = False):
```
pierresegonne, line 100: Can you add a type for the return value pls?

parsers/NED.py
```diff
@@ -0,0 +1,224 @@
...
+    ENTSOE_raw_data = ENTSOE_query_production(
+        ENTSOE_DOMAIN_MAPPINGS[zone_key], session, target_datetime=target_datetime
+    )
+    if ENTSOE_raw_data is None:
+        raise ParserException(
+            parser="NED.py",
+            message="Failed to fetch ENTSOE data",
+            zone_key=zone_key,
+        )
+    ENTSOE_parsed_data = ENTSOE_parse_production(
```
pierresegonne, line 165: Could be extracted to its own function?

pierresegonne (approved):
Mostly minor comments to make code more readable - nothing blocking, thanks for doing it!

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: Will add that in another PR so this don't require another review.

pierresegonne (approved):

mhooi:
First of all thank you for implementing this major improvement so fast. 

@VIKTORVAV99: you also wrote it in one of your comments but next to solar I think it is good to add also wind energy. Currently only Wind generation from Sea is reported correctly on ENTSO-E. Wind generation from land is not included but is luckily in cluded in the NED.NL API. 

Looking for example at Electricity maps to the wind data on the 28th of April 11h00 there is only a value of 2.86 GW. 
When we look at [https://co2monitor.nl/en/energysources/history](url) we see the same 2.8GW for wind at see but there is also roughly 5 GW reported under wind from land. SO Electricity maps is highly underestimating the wind generation. 

I hope after having implemented the code the addition of Wind energy is relatively easy?

mhooi:
Ofcourse, no hurry. I'm just a bit eager to have finally "correct" or alligned data for the Netherlands :p 2.8 vs 7.8 GW is for example a big difference.

VIKTORVAV99:
Agreed!

I will open a internal issue about it so we can keep track of it 🙂

Issue #6636, opened 2024-04-06
[frontend] Climate Impact by Area ranking: some zones without data are sorted below low-impact areas.


jayaddison:
**Describe the bug**
The homepage displays a ranking of areas by climate impact, and it appears that these should be sorted with the lowest-impact area at the top of the list.

When viewing the homepage at the moment, I find three regions where data is unavailable displayed at the top of the list; after those a region with data and less-than 20g CO2/kWh appears next.

That's the bugreport; the regions without data shouldn't be ranked at the top of the list I don't think?

**To Reproduce**
1. Go to https://app.electricitymaps.com/
2. Inspect the "Climate Impact by Area" widget near the top-left of the page.
3. Observe regions where data is currently unavailable at the top of the list.

**Expected behavior**
The top-rated area/zone should be one with the lowest climate impact based on currently available data.

**Screenshots**
![image](https://github.com/electricitymaps/electricitymaps-contrib/assets/55152140/0fdad69f-9807-46d6-8f9b-0b84d8c82730)

**Desktop:**
 - OS: Debian Linux
 - Browser: Firefox
 - Version: v155.9

**Additional context**
Data seems to have previously been available in Canada's Saskatchewan, another zone that currently appears ranked as lower-impact than others despite an absence of live data at the moment.

jarek:
More data points: both Hong Kong and Indonesia have only yearly data, and Saskatchewan only has daily (but not hourly) data. (You can see this with the "Display data from the past" switcher in lower left corner of the page.)

I see Hong Kong and Indonesia at the top of "hourly" climate impact by area list, and after viewing Saskatchewan, it is also added to that list. When viewing the "daily" climate impact by area list, Hong Kong and Indonesia are still at the top, but Saskatchewan is in the correct place.

So it seems related to not having hourly/daily data for an area.

andiz2:
Hi guys! I would like to dive deep into this 😄 
Is there any documentation about where the datas is fetched from and where is it saved into so the browser would took it from?

Regards,
andiz2

VIKTORVAV99:
We should be able to fix this in the front end by filtering out nullish data but I'll also check if there is a simple fix for this that can be made in the backend.

andiz2:
I've sorted it from front-end side.
Changed the filter from getRankingPanelData to:
`    .filter(
      (zone) =>
        zone.co2intensity !== undefined && zone.co2intensity !== null &&
[...]

PR #6637, opened 2024-04-08
Update getRankingPanelData.ts to omit null values for co2intensity

andiz2:
## Issue

Closes: #6636

## Description

Fixed bug in the map frontend which was about filtering null values for co2intensity.

### Double check
Wasn't able to run `poetry run test_parser "DE"` because of some dependency errors: 
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
clikit 0.6.2 requires crashtest<0.4.0,>=0.3.0, but you have crashtest 0.4.1 which is incompatible.

But since I've changed only 1 line of code I think shouldn't be any issues. 😄 


web/src/features/panels/ranking-panel/getRankingPanelData.ts
```diff
@@ -58,7 +58,7 @@ export const getRankedState = (
     })
     .filter(
       (zone) =>
-        zone.co2intensity !== undefined &&
+        zone.co2intensity !== undefined && zone.co2intensity !== null &&
```
VIKTORVAV99, line 61:
```suggestion
        Boolean(zone.co2intensity) &&
```
I would simplify this check to this, not only does it produce less code but it makes the code easier to understand IMO.

VIKTORVAV99 (commented):
The change itself looks good but I have a suggestion on how we can make it simpler.

I also have some general feedback on the PR itself. There is no need to run the parser testing if you are no modifying any parsers. But if you do you need to set up poetry and use that instead of pip.

What you need to ensure however is that all the code is formatted using prettier. Otherwise it won't pass the CI and we will not be able to merge it. My suggested render this redundant though so it's just for future PRs.

web/src/features/panels/ranking-panel/getRankingPanelData.ts
andiz2: Thanks for reply @VIKTORVAV99!
I like your solution, looks well too.
I have a question about this phrase: "But if you do you need to set up poetry and use that instead of pip." - can you be more explicit about it?
I've tryied to run poetry format but somehow I have this error..maybe someone can help about it..I would like to fix it and contribute to future issues since project is impressive!
"poetry run format
  File "/home/beast/.local/bin/poetry", line 5, in <module>
    from poetry.console.application import main
  File "/usr/lib/python3/dist-packages/poetry/console/__init__.py", line 1, in <module>
    from .application import Application
  File "/usr/lib/python3/dist-packages/poetry/console/application.py", line 3, in <module>
    from cleo import Application as BaseApplication
ImportError: cannot import name 'Application' from 'cleo' (/home/beast/.local/lib/python3.10/site-packages/cleo/__init__.py)"

VIKTORVAV99: Have you followed the steps we have described here https://github.com/electricitymaps/electricitymaps-contrib/wiki/Set-up-local-environment?

Those should contain everything you need and if they don't please let me know or open a issue so we can update it.

andiz2: I've followed the guide and have troubles installing poetry...:
"
python3 --version      
Python 3.10.12
beast@DESKTOP-TFSG0II:/mnt/c/Users/admin/Documents/code js/electricity_fork$ pipx install poetry
⚠️  File exists at /home/beast/.local/bin/poetry and points to /home/beast/.local/bin/poetry, not
    /home/beast/.local/pipx/venvs/poetry/bin/poetry. Not modifying.
  installed package poetry 1.8.2, installed using Python 3.10.12
    - poetry (symlink missing or pointing to unexpected location)
done! ✨ 🌟 ✨
beast@DESKTOP-TFSG0II:/mnt/c/Users/admin/Documents/code js/electricity_fork$ poetry --version
  File "/home/beast/.local/bin/poetry", line 5, in <module>
    from poetry.console.application import main
  File "/usr/lib/python3/dist-packages/poetry/console/__init__.py", line 1, in <module>
    from .application import Application
  File "/usr/lib/python3/dist-packages/poetry/console/application.py", line 3, in <module>
    from cleo import Application as BaseApplication
ImportError: cannot import name 'Application' from 'cleo' (/home/beast/.local/lib/python3.10/site-packages/cleo/__init__.py)
"
I don't understand where is the issue and where it comes from

andiz2:
I've sorted the poetry install, ready to make a new pull request with your idea as solution if it not bothers you.

VIKTORVAV99:
> I've sorted the poetry install, ready to make a new pull request with your idea as solution if it not bothers you.

You don't need to close this PR, you can simply push the changes to this branch and it will update.

VIKTORVAV99 (approved):
LGTM! Thanks for fixing it!

PR #6642, opened 2024-04-08
Silkebonnen/avo 83 move all graphs to cards

silkeholmebonnen:
## Issue

[AVO-83](https://linear.app/electricitymaps/issue/AVO-83/move-all-graphs-to-cards)

## Description

The goal of this PR is to move the graphs into cards.

### Double check


web/src/features/charts/CarbonChart.tsx
```diff
@@ -31,7 +32,7 @@ function CarbonChart({ datetimes, timeAverage }: CarbonChartProps) {
     return <NotEnoughDataMessage title="country-history.carbonintensity" />;
   }
   return (
-    <>
+    <GraphCard className="pb-2">
```
VIKTORVAV99, line 35: In the Figma the unit have moved places from the right side to under the title. Do we want to do that in this PR or do you want to make a follow up PR for that?

web/src/features/charts/PriceChart.tsx
```diff
@@ -51,9 +52,9 @@ function PriceChart({ datetimes, timeAverage }: PriceChartProps) {
...
 
   return (
-    <>
+    <GraphCard>
       <ChartTitle translationKey="country-history.electricityprices" />
-      <div className="relative overflow-hidden">
+      <div className="relative">
         {isPriceDisabled && (
           <div className="absolute top-0 -ml-3 h-full w-[115%]">
             <div className="h-full w-full rounded bg-white opacity-90 dark:bg-gray-900" />
```
VIKTORVAV99, line 60: This overlay is now overflowing and is causing the details panel to have a horizontal scroll.

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -102,7 +103,10 @@ function BarBreakdownChart({
     .sort();
 
   return (
-    <div className="text-sm" ref={ref}>
+    <div
+      className="mt-4 rounded-2xl border-[1px] border-neutral-200 px-4 pb-2 text-sm dark:border-gray-700"
```
madsnedergaard, line 107: Tiny nitpick, but instead of `border-[1px]` you can just use `border` :)

web/src/features/charts/CarbonChart.tsx
silkeholmebonnen: I'll do it in another PR because I'm not completely sure of what we want to write as the units

VIKTORVAV99: The vertical axis cuts into the graph now. Which I believe is due to the margin the unit is causing.
image

As for what to write for the units I think that should be the same as the unit's that are already there for now.

silkeholmebonnen: True, I hope it is fixed now:)

VIKTORVAV99 (approved):
LGTM!

PR #6643, opened 2024-04-08
refactor(ge): upgrade GE parser with event classes

amv213:
## Issue

## Description

This PR upgrades the `GE` parser to use event classes. I have also added a couple of tests so that we can catch regressions going forward (I took the occasion to try out `pytest` style tests using fixtures that would be compatible with us slowly migrating towards using `syrupy` for snapshot testing)

While adding the tests I spotted a small bug in the step that fetches historical production data: the data returned was not the data for the 24 hours in the requested UTC day, but for the 24 hours of the GE day in which the requested date fell into i.e. straddling across UTC day boundaries. Now the data returned covers the full 24 hours of the UTC day requested. Now though, given that the UTC day straddles GE day boundaries, the logic parsing the historical data API response had to be slightly generalised as in reality the API can and will return more than one table (one for each GE day).

Also I have added a small update to the `RU` exchange key, which is now called `RU-1` (the GE parser is not used for `GE->RU-1` so it wasn't that important)

### Double check


parsers/GE.py
```diff
@@ -3,167 +3,246 @@
...
+
+        filtered_breakdowns.append(
+            zoneKey=event.zoneKey,
+            datetime=event.datetime,
+            production=event.production,
+            storage=event.storage,
+            source=event.source,
+            sourceType=event.sourceType,
+        )
+    return filtered_breakdowns
```
VIKTORVAV99, line 59: This is not needed anymore and can be skipped. We have built and are building out a outlier detection layer in the backend instead.

parsers/GE.py
```diff
@@ -3,167 +3,246 @@
...
-    zone_key1: str = "GE",
-    zone_key2: str = "TR",
+    zone_key1: ZoneKey,
+    zone_key2: ZoneKey,
     session: Session | None = None,
     target_datetime: datetime | None = None,
-    logger=getLogger(__name__),
-) -> list[dict] | dict:
+    logger: Logger = getLogger(__name__),
+) -> list[dict]:
```
VIKTORVAV99, line 178: Just as a FYI, we will probably switch to using ENTSO-E directly instead of this source as it don't have any fixed timestamps to read and the fact that it's using ENTSO-E for historical data anyway which is polluting the database with both sources.

VIKTORVAV99 (commented):
I'll give this a full review later today but added two early comments.

parsers/GE.py
```diff
@@ -3,167 +3,212 @@
...
+        for timestamp, production in itemised_production_data.dropna(
+            axis="columns", how="all"
+        ).items():
+            dt = timestamp.to_pydatetime().replace(tzinfo=TIMEZONE)
+
+            production_mix = ProductionMix()
+            production_mix.add_value("gas", production["Thermal Power Plants"])
+            production_mix.add_value("hydro", production["Hydro Power Plants"])
+            production_mix.add_value("solar", production["Wind Power Plants"])
+            production_mix.add_value("wind", production["Solar Power Plants"])
```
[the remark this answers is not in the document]
amv213: whoops - will fix and double check

VIKTORVAV99 (approved):
LGTM!

Thanks for fixing it!

PR #6650, opened 2024-04-09
fix: Update GE exchange configurations to use ENTSOE.py instead of GE.py

VIKTORVAV99:
## Issue

We were polluting the database with a mix of ENTSOE and GE values.

Slack discussion here: https://electricitymaps.slack.com/archives/C01PWQYKX2R/p1712559608324009

## Description

Switches to use ENTSOE only to avoid the polluted data and ensure it gets properly overridden if it's updated later on.

### Double check

### DIFF
--- a/config/exchanges/AM_GE.yaml
+++ b/config/exchanges/AM_GE.yaml
@@ -2,6 +2,6 @@ lonlat:
   - 44.313
   - 41.205
 parsers:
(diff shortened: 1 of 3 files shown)


wobniarin (approved):
hey, I read the slack thread. Thanks for doing that!

PR #6651, opened 2024-04-09
Enable ESlint rules for React Hooks

VIKTORVAV99:
## Issue

Closes: AVO-172

### Double check


VIKTORVAV99:
Turns out we hade quite few hook violations...

I'll fix them after #6263 is merged as they have a lot of overlap in the files they change.

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -142,33 +142,33 @@ function AreaGraph({
...
-  const startTime = datetimes?.at(0);
-  const lastTime = datetimes?.at(-1);
-  const interval = datetimes?.at(-2);
+  const startTime = datetimes.at(0);
+  const lastTime = datetimes.at(-1);
+  const interval = datetimes.at(-2);
 
-  if (!startTime || !lastTime || !interval) {
-    return null;
-  }
```
VIKTORVAV99, line 151: There is no point in moving this one as all the values would have been used by the time it goes through all the hooks.

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -142,33 +142,33 @@ function AreaGraph({
...
   // The endTime needs to include the last interval so it can be shown
   const endTime = useMemo(
-    () => new Date(lastTime.getTime() + intervalMs),
+    () => (lastTime ? new Date(lastTime.getTime() + intervalMs) : null),
     [lastTime, intervalMs]
   );
-  const datetimesWithNext = useMemo(() => [...datetimes, endTime], [datetimes, endTime]);
-  if (!endTime) {
-    return null;
-  }
```
VIKTORVAV99, line 164: Same here

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -205,18 +205,12 @@ function AreaGraph({
...
     },
     [setGraphIndex, setSelectedLayerIndex]
   );
 
-  // Don't render the graph at all if no layers are present
-  if (isEmpty(layers)) {
-    console.error('No layers present in AreaGraph');
-    return null;
-  }
-
```
VIKTORVAV99, line 219: We should exit way before this is even run so this should never be a problem.

web/src/features/charts/graphUtils.ts
```diff
@@ -43,17 +43,20 @@ export const detectHoveredDatapointIndex = (
 export const getTooltipPosition = (isMobile: boolean, marker: { x: number; y: number }) =>
   isMobile ? { x: 0, y: 0 } : marker;
 
-// TODO: Deprecate this
-export const isEmpty = (object: any) =>
-  [Object, Array].includes((object || {}).constructor) &&
-  Object.entries(object || {}).length === 0;
-
```
VIKTORVAV99, line 50: Which meant we can remove this as well since it's no longer used.

web/src/features/charts/graphUtils.ts
```diff
@@ -43,17 +43,20 @@ export const detectHoveredDatapointIndex = (
...
-export const getTimeScale = (width: number, startTime: Date, endTime: Date) =>
-  scaleTime()
+export const getTimeScale = (
+  width: number,
+  startTime?: Date | null,
+  endTime?: Date | null
+) => {
+  if (!startTime || !endTime) {
+    return null;
+  }
```
VIKTORVAV99, line 55: As per react documentation conditionals should be handled inside the functions instead of the component for optimal hook functionality.

web/src/components/Accordion.tsx
```diff
@@ -20,27 +20,20 @@ export default function Accordion({
...
-  const [isCollapsed, setIsCollapsed] = useState(isCollapsedDefault);
-
+  const [isCollapsed, setIsCollapsed] = useAtom(isCollapsedAtom);
   useEffect(() => {
-    if (collapsedAtom !== null) {
-      setIsCollapsed(collapsedAtom);
+    if (isCollapsed === undefined) {
+      setIsCollapsed(isCollapsedDefault);
     }
-  }, [collapsedAtom]);
```
VIKTORVAV99, line 34: I'm a bit unsure about these changes but it was the best way I could think of to resolve the fact that the atoms can't be conditionals.

web/src/features/panels/zone/EstimationCard.cy.tsx
```diff
@@ -67,10 +79,10 @@ describe('EstimationCard with known estimation method', () => {
...
 
   it('Toggles when collapse button is clicked', () => {
-    cy.get('[data-test-id="collapse-up"]').should('not.exist');
-    cy.get('[data-test-id="collapse-down"]').should('exist');
-    cy.get('[data-test-id="body-text"]').should('not.exist');
-    cy.get('[data-test-id="methodology-link"]').should('not.exist');
+    cy.get('[data-test-id="collapse-up"]').should('exist');
+    cy.get('[data-test-id="collapse-down"]').should('not.exist');
+    cy.get('[data-test-id="body-text"]').should('exist');
+    cy.get('[data-test-id="methodology-link"]').should('exist');
```
VIKTORVAV99, line 85: Since the state is now persisted using an atom these become more or less inverted as it's clicked in the previous test.

I was hoping to find a easy solution to isolate them but I don't think my solution is fully isolating them.

silkeholmebonnen: The estimation card was closed by default before but in `ResetEstimationCard` you set it to be open by default so this gets inverted. I think we might want to set `setIsCollapsed(true);` in `ResetEstimationCard` to get the same results as before. But for the outage card it was open by default so we might need to make another one for that

web/src/features/time/TimeController.tsx
```diff
@@ -39,7 +39,7 @@ export default function TimeController({ className }: { className?: string }) {
         index: datetimes.length - 1,
       });
     }
-  }, [data]);
+  }, [data, datetimes, setSelectedDatetime]);
```
silkeholmebonnen, line 42: Why is this added?

web/src/utils/state/atoms.ts
```diff
@@ -46,8 +46,10 @@ export const feedbackCardCollapsedNumberAtom = atom(0);
...
-export const dataSourcesCollapsedBarBreakdown = atom(true);
+export const dataSourcesCollapsedBarBreakdown = atom<boolean | undefined>(true);
 
-export const dataSourcesCollapsedBreakdown = atom(true);
+export const dataSourcesCollapsedBreakdown = atom<boolean | undefined>(true);
 
-export const dataSourcesCollapsedEmission = atom(true);
+export const dataSourcesCollapsedEmission = atom<boolean | undefined>(true);
+
+export const estimationCardCollapsed = atom<boolean | undefined>(undefined);
```
silkeholmebonnen, line 55: Why is it necessary to have `atom<boolean | undefined>` and not just `atom<boolean>`?

silkeholmebonnen (commented):
I have a bunch of comments, but I just want to make sure I understand it. But it looks good overall:)

web/src/components/Accordion.tsx
VIKTORVAV99: Not just the rule per se but react internally requires all hooks to be called in the same order on every render to function properly internally.

This means the calling of the atom itself can't be a conditional but you can choose to not use the atom later on.

web/src/features/charts/elements/AreaGraph.tsx
VIKTORVAV99: They are used in other hook calls which means that by the point we would exit early if we moved it all the values would have already been used making the check if they are fall pointless.

web/src/features/charts/elements/AreaGraph.tsx
VIKTORVAV99: We should exit and not render the graph at all if there is no data for it. This is handled when the graphs are created, not in this file itself. Or at least should be.

web/src/features/map/Map.tsx
```diff
@@ -145,6 +145,10 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
     isSourceLoaded,
     spatialAggregate,
     isSuccess,
+    isLoading,
+    isError,
+    worldGeometries.features,
+    theme.clickableFill,
```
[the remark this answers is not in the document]
VIKTORVAV99: Because all of these are in one way or another used inside the hook, that means the outcome is dependant on their value and if they change the hook should rerun. They probably won't change though but to react they *can* change which is why it wants them here.

web/src/features/panels/zone/EstimationCard.cy.tsx
VIKTORVAV99: Ah, I'll double check.

web/src/features/time/TimeController.tsx
VIKTORVAV99: datetimes.length is called inside the hook which requires datetimes to be a part of the dependency array, not sure why setSelectedDatetime is required but it would not be happy without it.

web/src/utils/state/atoms.ts
VIKTORVAV99: I didn't want it to have a state from the begging so it can be used for both the estimation card and the outage card, but perhaps it's better to split them into two and just use a boolean value.

VIKTORVAV99:
> I have a bunch of comments, but I just want to make sure I understand it. But it looks good overall:)

No problem!

I tried limiting the functional changes as much as possible but the only thing I am a bit unsure of is the atoms and the accordion/estimation/outage cards.

VIKTORVAV99:
> Thanks for taking on this one 💚 
> 
> 
> 
> Looks really good, couple of small fixes:
> 
> 
> 
> 1. Solar layer can get stuck on
> 
> 2. The line on the charts that previous followed the selectedDatetime is now missing
> 
> 
> 
> This was quite a big PR 😅, could have been nice to isolate the hooks fixes from other fixes but overall very nice 🙌 

They are all hook fixes (if we include the dependency arrays) 😅

Fix one thing and it complains about the next, it's all pretty simple but deep issues though.

I'll take a look at the solar layer and selectedDatetime though.

web/src/features/panels/zone/EstimationCard.cy.tsx
VIKTORVAV99: Not sure where you mean I set it to open by default here?
I set it to undefined which means it should use the default values provided to the card type, either open or closed.

web/src/features/weather-layers/solar/SolarLayer.tsx
```diff
@@ -93,7 +93,7 @@ export default function SolarLayer({ map }: { map?: maplibregl.Map }) {
         map.removeSource('solar');
       }
     };
-  }, [map, node, isVisibleReference.current]);
```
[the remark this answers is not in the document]
VIKTORVAV99: I have added it back but ESlint throws a warning now:
```
React Hook useEffect has an unnecessary dependency: 'isVisibleReference.current'. Either exclude it or remove the dependency array. Mutable values like 'isVisibleReference.current' aren't valid dependencies because mutating them doesn't re-render the component.
```
However since it's just a warning we can adress it later.

web/src/features/weather-layers/solar/SolarLayer.tsx
```diff
@@ -93,7 +93,7 @@
...
+  }, [map, node, setIsLoadingSolarLayer, isVisibleReference.current]);
```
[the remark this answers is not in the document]
VIKTORVAV99: Solar layer don't appears to be working correctly without it. We should address that in another PR.

VIKTORVAV99:
I can't really figure out how to make the Estimation cards do what we want with these new restrictions without significant changes.

We could restore them and skip the changes in here, ESlint would error in this PR but let other PRs pass in the future. We would have to change how earthly works though as it also run ESLint for some reason?

web/src/features/panels/zone/EstimationCard.cy.tsx
silkeholmebonnen: Ahh okay I understand, but I don't think the solution to this is to change the test. I will try to figure out another way to do it:)

tonypls:
> I can't really figure out how to make the Estimation cards do what we want with these new restrictions without significant changes.
> 
> We could restore them and skip the changes in here, ESlint would error in this PR but let other PRs pass in the future. We would have to change how earthly works though as it also run ESLint for some reason?

Could we address the fixes that don't require significant changes then turn the ESLint rule off until we want to address them?

VIKTORVAV99:
> > I can't really figure out how to make the Estimation cards do what we want with these new restrictions without significant changes.
> > We could restore them and skip the changes in here, ESlint would error in this PR but let other PRs pass in the future. We would have to change how earthly works though as it also run ESLint for some reason?
> 
> Could we address the fixes that don't require significant changes then turn the ESLint rule off until we want to address them?

We could, but then we run the risk of reintroducing some of them, but might be worth it anyway?

I would keep all the changes except for the estimation card stuff if we were to do this.

tonypls:
> > > I can't really figure out how to make the Estimation cards do what we want with these new restrictions without significant changes.
> > > We could restore them and skip the changes in here, ESlint would error in this PR but let other PRs pass in the future. We would have to change how earthly works though as it also run ESLint for some reason?
> > 
> > 
> > Could we address the fixes that don't require significant changes then turn the ESLint rule off until we want to address them?
> 
> We could, but then we run the risk of reintroducing some of them, but might be worth it anyway?
> 
> I would keep all the changes except for the estimation card stuff if we were to do this.

It's a nice improvement even without the rule enabled. Yeah I think that's a logical split on what to include. Is it possible to exclude a line or file from the linting?

VIKTORVAV99:
It is, so we could also do that, which is probably preferable.

VIKTORVAV99:
Oh god now this PR have the same issue as #6699...

I'll try and fix it in that PR directly then we can hopefully merge this.

tonypls:
Let me know when this needs review again 🚀

tonypls (approved):
LGTM!

madsnedergaard:
Kudos for getting this one through and making the codebase a better place ❤️

PR #6654, opened 2024-04-10
refactor(gt): upgrade GT parser with event classes

amv213:
## Issue

## Description

This PR upgrades the `GT` parser to use event classes and adds tests to catch regressions.

While doing so I have also updated the data retrieval logic so that the parser returns correctly all data for the target day of interest. I have also and added an option to allow fetching more than one day of data if we want to.

I think that the last datapoint that was getting returned for live production data was also slightly-wrong as it contained live data that was still getting tallied up for that hour.

### Double check


parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
+            )
+
+        # The JSON data returned by the API is a list of objects, each representing one technology type.
+        daily_payloads[local_day] = response.json()
+
+    if not any(daily_payload for daily_payload in daily_payloads.values()):
+        raise ParserException(
+            PARSER,
+            f"Exception when fetching {target.value}: no data available for UTC day {target_day_utc.strftime('%Y-%m-%d %Z')}. "
+            f"Note that historical data is only available for the last 1 year.",
```
amv213, line 80: note: I randomly found this out, but it might be a problem for reproducibility of past results / workflows

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
+        consumption_list.append(
+            zoneKey=zone_key,
+            datetime=dt,
+            consumption=consumption,
+            source=SOURCE,
+        )
+    return consumption_list.to_list()
 
 
+@refetch_frequency(timedelta(days=1))
```
amv213, line 145: note: I have set it to `1` but happy to increase this if it's preferred

The data API only supports asking for one day at a time so asking for e.g. two days of data every two days in reality results in the same number of HTTP calls being made than asking for one day of data every day. Happy to go with whichever combination of refetch frequency / num_backlog_days you prefer

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
-            "zoneKey": zone_key,
-        }
-        for hour, row in enumerate(api_data)
-    ]
-    return [
-        validation.validate(
-            result, logger, floor=PRODUCTION_THRESHOLD, remove_negative=True
-        )
-        for result in results
-    ]
```
amv213, line 111: note: I have dropped this functionality

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
-
-    is_historical_data = date_time.date() < datetime.now(TIMEZONE).date()
-
-    # For live consumption data, an hour's data isn't updated until the hour has passed,
-    # so the current (and future) hour(s) should not be included in the results.
-    # For live production data, the API will return zero-filled future data until the end of the day,
-    # so future hours should not be included in the results.
-    cutoff_index = (
-        date_time.hour if target == ApiKind.CONSUMPTION else date_time.hour + 1
-    )
```
amv213, line 145: fix: I found out that this wasn't quite true and that an hour's data not being updated until the hour has passed applied to any kind of data - not only to consumption data. This resulted in getting production data whose last datapoint was still temporary data

For example, a production request made at 9:30 UTC had this (in-progress) datapoint for the 9->10 block: 
```
{'datetime': datetime.datetime(2024, 4, 10, 9, 0, tzinfo=datetime.timezone.utc), 'zoneKey': 'GT', 'production': {'biomass': 304.24, 'coal': 202.9, 'gas': 15.386, 'geothermal': 35.449, 'hydro': 95.92, 'oil': 173.165, 'solar': 0.0, 'unknown': 440.2, 'wind': 1.69}, 'storage': {}, 'source': 'wl12.amm.org.gt', 'sourceType': <EventSourceType.measured: 'measured'>, 'correctedModes': []}
```

while the same request made at 10:30 UTC had (completed) updated values:
```
{'datetime': datetime.datetime(2024, 4, 10, 9, 0, tzinfo=datetime.timezone.utc), 'zoneKey': 'GT', 'production': {'biomass': 305.42683, 'coal': 203.60906, 'gas': 11.92, 'geothermal': 33.34, 'hydro': 92.37983, 'oil': 188.9405, 'solar': 0.0, 'unknown': 440.72003, 'wind': 0.06}, 'storage': {}, 'source': 'wl12.amm.org.gt', 'sourceType': <EventSourceType.measured: 'measured'>, 'correctedModes': []}, 
```

The new code should cut things off more correctly (i.e. always cutoff the in-progress hour of live data)

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
+            dt_utc = dt_local.astimezone(timezone.utc)
+
+            # ignore data outside of range of interest
+            if dt_utc < target_day_utc_start or dt_utc > target_day_utc_end:
+                continue
+
+            # For live data, the current hour's data isn't finished updating until the hour has passed,
+            # so the current hour should not be included in the results.
+            if dt_utc >= hour_now_utc:
+                continue
```
amv213, line 101: thought: I though of potentially still including the "live" datapoint(`dt_utc == hour_now_utc`) and we could then make it count as e.g. of being of type "forecasted"... but not sure if it would really be valid to consider that as a real forecast so I went for just skipping it

parsers/test/snapshots/snap_test_GT.py
amv213, line 1: note: 

live data (with time frozen at `"2024-04-10 12:28:00 UTC"`) correctly returns datapoints from the start of `2024-04-10 UTC` up to `11:00` (i.e. the `11:00 -> 12:00` production data), avoiding the "in-progress" data (`12:00->1:00`)

historical data (for a `"2023-07-16T12:00:00+00:00"` target) correctly returns data for the entire `2023-07-16 UTC` day

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
-    # The JSON data returned by the API is a list of objects, each
-    # representing one technology type. Collect this information into a list,
-    # with the list index representing the hour of day.
-    results = [collections.defaultdict(float) for _ in range(24)]
-    for row in response_payload:
-        # The API returns hours in the range [1, 24], so each one refers to the
-        # past hour (e.g., 1 is the time period [00:00, 01:00)). Shift the hour
-        # so each index represents the hour ahead and is in the range [0, 24),
-        # e.g., hour 0 represents the period [00:00, 01:00).
-        results[int(row["hora"]) - 1][row["tipo"]] = row["potencia"]
```
amv213, line 135: thought: should we rather timestamp the data a bit like the API is doing e.g. timestamp the `[00:00, 01:00)` production values as being the production at `1:00` rather than the production at `00:00` ?

parsers/GT.py
VIKTORVAV99: Nice!
As long as the datetimes are the same it causes no long term issues but very nice to have it fixed regardless!

parsers/GT.py
VIKTORVAV99: Probably a good idea, forecasted values generally does not update and is often separate from the real data at the source as well.

We might support other types as well in the future and then a in_progress or partial might be a good thing to keep in mind.

parsers/GT.py
VIKTORVAV99: We use forward looking dates internally so in this case the production for the 00:00 to 01:00 period should be marked with time of 00:00. Was it not doing that before?

amv213: Perfect - thanks for the clarification! Yes, that is what it was doing (and still is)

parsers/GT.py
VIKTORVAV99: The most important thing is that the data is available for the last 6 months or so then our automatic refetch pipeline will be able to fill any gaps that might occur if the source is not available during the first parser runs.

parsers/test/conftest.py
VIKTORVAV99, line 1: I would prefer if this was added as an independent PR that also removed all the current usage of the same snippet of code in the tests.

Just so the PRs are focused on one thing so if anything where to go wrong and we had to revet it the other functional changes are still there.

parsers/GT.py
```diff
@@ -1,163 +1,206 @@
...
+        production_mix.add_value("geothermal", row["VAPOR"])
+        production_mix.add_value("hydro", row["AGUA"])
+        production_mix.add_value("oil", row["BUNKER"] + row["DIESEL"])
+        production_mix.add_value("solar", row["IRRADIACIÓN"])
+        production_mix.add_value(
+            "unknown",
+            # 'BIOMASA/CARBÓN not always present, e.g. 15/07/2023 TZ
+            row.get("BIOMASA/CARBÓN", 0.0) + row["CARBÓN/PETCOKE"] + row["SYNGAN"],
+        )
+        production_mix.add_value("wind", row["VIENTO"])
```
VIKTORVAV99, line 182: I would prefer if this was done as a loop instead with a mapping to the maps modes. That way we can avoid the get.("...", 0.0) as we really don't want default values to be set in the parsers.

By doing it in a loop we can also get more accurate floating point correction as it's done after each addition instead of on the total. This should probably not be a major issue but still.

(it's still there in some legacy parsers but should be removed whenever possible).

There looks to be some remapping that should be done here as well from the unknown to known modes but I'll address that later.

parsers/GT.py
```diff
@@ -1,163 +1,215 @@
...
+    "CARBÓN": "coal",
+    "GAS NATURAL": "gas",
+    "VAPOR": "geothermal",
+    "AGUA": "hydro",
+    "BUNKER": "oil",
+    "DIESEL": "oil",
+    "IRRADIACIÓN": "solar",
+    "BIOMASA/CARBÓN": "unknown",  # not always present, e.g. 15/07/2023 TZ
+    "CARBÓN/PETCOKE": "unknown",
+    "SYNGAN": "unknown",
```
VIKTORVAV99, line 39: And this to gas.

parsers/GT.py
```diff
@@ -1,163 +1,215 @@
...
+        num_backlog_days=0,
     )
 
-    return results if is_historical_data else results[:cutoff_index]
+    production_breakdown_list = ProductionBreakdownList(logger)
+    for dt, row in api_data.items():
+        production_mix = ProductionMix()
+        for key in row:
+            if key not in PRODUCTION_TYPE_TO_PRODUCTION_MODE:
+                continue
```
VIKTORVAV99, line 188: Let's also log a warning here so we know it's missing from the mapping.

VIKTORVAV99 (commented):
Just some small final notes but looks good!

VIKTORVAV99 (approved):
LGTM!

Thanks again for all your help with the parsers!

PR #6655, opened 2024-04-10
refactor(gb): upgrade GB parser to use event classes

amv213:
## Issue

## Description

This PR upgrades the `GB` parser to use event classes, drops `arrow` in favour of `datetime`, and adds tests to catch regressions.

### Double check


parsers/GB.py
```diff
@@ -1,71 +1,99 @@
...
+) -> list[dict]:
+    """Returns one-hourly prices for the requested day."""
 
-    r = session or Session()
-    formatted_from = now.shift(days=-1).format("DD/MM/YYYY")
-    formatted_to = now.format("DD/MM/YYYY")
+    now = datetime.now(timezone.utc)
+    target_datetime = (
+        now if target_datetime is None else target_datetime.astimezone(timezone.utc)
+    )
```
amv213, line 33: thought: The previous code was returning two days of data, which I thought was a bug so I changed it to just one. 

In light of the recent comments in the other PRs though it was probably a feature 😅 Should I revert? (and would I need to bump the refetch frequency to `2` then?)

parsers/GB.py
```diff
@@ -1,71 +1,99 @@
...
+            tzinfo=timezone.utc
         )
+        for daily_zone_data in daily_market_data:
+            zone_code = daily_zone_data.get("perimetre")
 
-        for item in donnesMarche:
-            if item.get("granularite") != "Global":
+            # Data for Germany / Luxembourg is not set / reported as aggregate region
+            if zone_code in {"DE", "DL"}:
                 continue
```
amv213, line 60: note: I don't know what the best way to deal with these really is / how to interpret the API data

"DE" (germany) data is always null, and "DL" data has values. "DL" might maybe represent Germany + Luxembourg? In fact https://www.rte-france.com/eco2mix/les-donnees-de-marche displays a single aggregate `Allemagne + Lux` curve 👀  either way, I don't think they would be usable for us

parsers/GB.py
VIKTORVAV99: It would be nice with 2 days of data. But it's not critical.

parsers/GB.py
amv213: I see - thanks! I am very much not familiar with the "real life" side of things 😅 

So happy to just skip over these as done here?

parsers/GB.py
amv213: Will do and update the snapshots - shouldn't be too much of a problem

parsers/GB.py
VIKTORVAV99: Yep I think so. But I'll double check when doing a full code review.

parsers/GB.py
```diff
@@ -1,71 +1,104 @@
...
+
+                price_list.append(
+                    zoneKey=zone_key,
+                    datetime=dt,
+                    source="rte-france.com",
+                    price=price,
+                    currency="EUR",
+                    sourceType=EventSourceType.forecasted
+                    if is_future
+                    else EventSourceType.measured,
```
VIKTORVAV99, line 91: If the price is set on a day ahead market it is fine to say it's only measured as this won't change.

We should probably add a `EventSourceType.day_ahead`.

parsers/GB.py
```diff
@@ -1,71 +1,104 @@
...
-    for donnesMarche in obj:
-        if donnesMarche.tag != "donneesMarche":
-            continue
+    xml_tree = ET.fromstring(response.content)
 
-        start_date = arrow.get(
-            arrow.get(donnesMarche.attrib["date"]).datetime, "Europe/Paris"
+    price_list = PriceList(logger=logger)
+    for daily_market_data in xml_tree.iterfind("donneesMarche"):
+        day = datetime.strptime(daily_market_data.get("date"), "%Y-%m-%d").replace(
```
VIKTORVAV99, line 57: Small nitpick here, calling get on daily_market_data can potentially return none if the object is missing which is casuing type errors since strptime only accepts strings.

VIKTORVAV99 (commented):
Just some small notes but working as expected!

parsers/GB.py
amv213: I see! Thanks @VIKTORVAV99 

I just realised now then that this API can return prices from the full day ahead, just by updating the code to ask for `dateFin` = day + 1 (and to double check data for day + 2 or more returns nulls instead of having values). Is this something that would be useful to add here? (and I imagine still marking them as 'measured' right?)

VIKTORVAV99: Yes that would be very nice! It would open up more possibilities in the future.

amv213: Perfect! I have updated to return day-ahead prices for live data, but let me know if you prefer to return "day ahead" prices also for historical data

parsers/GB.py
```diff
@@ -1,71 +1,115 @@
...
+
+            granularity = daily_zone_data.get("granularite")
+            if granularity != "Global":
                 continue
-            value = None
-            for value in item:
-                if value.text == "ND":
+
+            for value in daily_zone_data:
+                price = None if value.text == "ND" else float(value.text)
```
VIKTORVAV99, line 88:
```suggestion
                price = None if value.text == "ND" else float(value.text) if value.text else None
```

VIKTORVAV99 (commented):
Just one small thing to fix a type error, otherwise things looks good!

VIKTORVAV99 (approved):
LGTM!

Thanks for upgrading it!

PR #6659, opened 2024-04-12
feat(app): add cacheBuster and use URL internally

VIKTORVAV99:
## Issue

Closes: AVO-166

## Description

This PR does two things, restructures the query functions to use URL objects internally and adds a cache busting function to the state and zone queries.

### Double check


web/src/api/helpers.ts
```diff
@@ -68,6 +68,18 @@ export function getBasePath() {
...
+  const current_date = new Date();
+  const date_to_hour = new Date(
+    current_date.getFullYear(),
+    current_date.getMonth(),
+    current_date.getDate(),
+    current_date.getHours(),
+    current_date.getMinutes()
+  );
+  return date_to_hour.toISOString();
+}
```
madsnedergaard, line 81: Wouldn't it be simpler to do something like this instead?

```suggestion
export function cacheBuster(): string {
  const currentDate = new Date();
  currentDate.setSeconds(0);
  currentDate.setMilliseconds(0);
  
  return currentDate.toISOString();
}
```

madsnedergaard (commented):
One comment, otherwise it looks good to me

web/src/api/helpers.ts
VIKTORVAV99: It would!

madsnedergaard (approved):
LGTM! This will not actually do anything until we switch the domain to `app-backend.electricitymaps.com` though, but might be worth doing that in a separate PR :)

PR #6660, opened 2024-04-14
refactor(md): upgrade MD parser to use event classes

amv213:
## Issue

## Description

This PR upgrades the MD parser to use event classes and adds tests to catch regressions.

I have also fixed / updated the `fetch_price()` function, which was returning prices that were no longer correct as of `2023-12-31` (or before `2015-07-31`)

### Double check


parsers/MD.py
```diff
@@ -1,103 +1,96 @@
...
+    datetime(2015, 7, 31, tzinfo=timezone.utc): 145.0,
+    # https://www.legis.md/cautare/getResults?doc_id=134854&lang=ro
+    datetime(2023, 12, 31, tzinfo=timezone.utc): 201.0,
+    # https://www.legis.md/cautare/getResults?doc_id=142391&lang=ro
+    datetime(2024, 3, 21, tzinfo=timezone.utc): 185.0,
+    # TODO(amv213): update as new tariffs get rolled out...
+}
+
+
+class ArchiveDatapoint(NamedTuple):
```
VIKTORVAV99, line 71: Should this have a `@dataclass` decorator?

parsers/MD.py
```diff
@@ -1,103 +1,96 @@
...
+    # https://www.legis.md/cautare/getResults?doc_id=40589&lang=ro
+    datetime(2012, 5, 11, tzinfo=timezone.utc): 80.2,
+    # https://www.legis.md/cautare/getResults?doc_id=84436&lang=ro
+    datetime(2015, 7, 31, tzinfo=timezone.utc): 145.0,
+    # https://www.legis.md/cautare/getResults?doc_id=134854&lang=ro
+    datetime(2023, 12, 31, tzinfo=timezone.utc): 201.0,
+    # https://www.legis.md/cautare/getResults?doc_id=142391&lang=ro
+    datetime(2024, 3, 21, tzinfo=timezone.utc): 185.0,
+    # TODO(amv213): update as new tariffs get rolled out...
+}
```
VIKTORVAV99, line 68: Are these constant? As in they don't change hourly?

And do you know if they are just the electricity price or the total cost (including transmission fees and taxes)?

parsers/MD.py
```diff
@@ -147,125 +146,163 @@ def get_archive_data(
...
+    )  # assume base price for times before we could find references
+    for dt, new_price in prices:
+        if target_datetime < dt:
+            break
+        price = new_price
+
+    price_list = PriceList(logger=logger)
+    price_list.append(
+        zoneKey=zone_key,
+        datetime=target_datetime,
```
VIKTORVAV99, line 181: Could we turnacate this to the hour?

parsers/MD.py
```diff
@@ -1,103 +1,96 @@
...
+    datetime: datetime
+    consumption: float
+    planned_consumption: float
+    production: float
+    planned_production: float
+    tpp: float  # production from thermal power plants
+    hpp: float  # production from hydro power plants
+    res: float  # production from renewable energy sources
+    exchange_UA_to_MD: float
+    planned_exchange_UA_to_MD: float
```
VIKTORVAV99, line 83: Would it be possible to add exchange forecasts as well? I see the parser fon't contain it right now. Feel free to do that in a seperate PR though.

VIKTORVAV99 (commented):
Some small comments but it looks good!

parsers/MD.py
amv213: Yes, these are constants. It looks like each ['licence holder' has a regulated price](https://www.anre.md/energie-electrica-3-290), and every so often the National Agency for Energy Regulation of the Republic of Moldova (ANRE) emits new decrees to update them to new ones (e.g. like [here](https://www.legis.md/cautare/getResults?doc_id=142391&lang=ro) for the latest one). Each decree abrogates the previous one, so I tried to walk them all back one by one to try to cover all historical changes.

I think that the price includes transmission but not taxes: [here](https://www.anre.md/energie-electrica-3-290) they refer to it as "Tariff/Regulated price money/kWh (without VAT)", and both the ANRE decrees and [moldoelectrica](https://moldelectrica.md/ro/activity/tariff) themselves refer to it as "tariful pentru serviciul de transport al energiei electrice (TVA excluded)" which seems to be translating as "Tariff for the electricity transmission service"

parsers/MD.py
amv213: We can turn it into a dataclass instead of a named tuple if we prefer, but otherwise a `NamedTuple` doesn't need that decorator. 

I quickly updated the existing `collections.namedtuple` to a `typing.NamedTuple` - because it allows defining them in this more dataclass-y style that I find much more readable / intuitive

parsers/MD.py
amv213: Sure! I can try having a look into that 🙌🏽 I'll do that in a separate PR

parsers/MD.py
amv213: (The previous code was essentially always returning the `datetime(2015, 7, 31, tzinfo=timezone.utc): 145.0` price)

parsers/MD.py
VIKTORVAV99: It's fine, I don't really have a preference either way. :)

VIKTORVAV99 (approved):
LGTM!

Thanks for doing it and adding all the tests!

parsers/MD.py
VIKTORVAV99: Thanks!

Everything looks good then.

parsers/MD.py
amv213: @VIKTORVAV99 what would be the best way to implement this? Would it be just a matter of adding to the existing values returned by the `fetch_exchange()` function new datapoints with the planned values and "forecasted" source type like in the example [here](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser#fetch_exchange_forecast) or would it be a dedicated separate function just returning forecasts? I saw that some other parsers are using something called `fetch_exchange_forecast`

VIKTORVAV99: Oh yeah it should be a different function. I'll make sure to update the wiki.

PR #6661, opened 2024-04-14
refactor(ch): remove usage of arrow from CH parser

amv213:
## Issue

## Description

This PR removes the dependency on arrow from the `CH` parser.

P.S. Testing the parser locally with `poetry run test_parser CH`, it looks like the parser does not return recent enough data, but this seems to be unrelated to these changes as it is the case also with the code that is already on `master`
```
min returned datetime: 2024-04-11 10:00:00+00:00 UTC
max returned datetime: 2024-04-14 05:00:00+00:00 UTC  :( >2h from now !!! (now=2024-04-14T08:56:23+00:00 UTC)
```

### Double check

### DIFF
--- a/parsers/CH.py
+++ b/parsers/CH.py
@@ -1,17 +1,12 @@
-#!/usr/bin/env python3
-
-
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from logging import Logger, getLogger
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM!

Thanks for upgrading it!

PR #6663, opened 2024-04-15
Silkebonnen/avo 187 add data source accordions to other graphs

silkeholmebonnen:
## Issue

[AVO-187](https://linear.app/electricitymaps/issue/AVO-187/add-data-source-accordions-to-other-graphs)

## Description

The goal of this PR is to add the data source accordion to the emissionchart, carbonchart and breakdownchart.

### Double check

### DIFF
--- a/web/src/features/charts/CarbonChart.tsx
+++ b/web/src/features/charts/CarbonChart.tsx
@@ -1,6 +1,11 @@
+import Accordion from 'components/Accordion';
+import Divider from 'features/panels/zone/Divider';
+import { IndustryIcon } from 'icons/industryIcon';
 import { useTranslation } from 'react-i18next';
 import { TimeAverages } from 'utils/constants';
+import { dataSourcesCollapsedEmission } from 'utils/state/atoms';
 
+import { DataSources } from './bar-breakdown/DataSources';
 import { ChartTitle } from './ChartTitle';
 import AreaGraph from './elements/AreaGraph';
 import { getBadgeText, noop } from './graphUtils';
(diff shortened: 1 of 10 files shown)


web/src/features/charts/CarbonChart.tsx
```diff
@@ -50,6 +56,18 @@ function CarbonChart({ datetimes, timeAverage }: CarbonChartProps) {
...
+      <Divider />
+      <Accordion
+        title={t('data-sources.title')}
+        className="text-md"
+        isCollapsedAtom={dataSourcesCollapsedEmission}
+      >
+        <DataSources
+          title={t('data-sources.emission')}
+          icon={<IndustryIcon />}
+          sources={[...emissionSourceToProductionSource.keys()].sort()}
```
silkeholmebonnen, line 68: I made this a Map instead of just a list so it would be easier to implement [AVO-84](https://linear.app/electricitymaps/issue/AVO-84/add-the-production-source-legend-under-data-sources-for-emission). When I do this I will also change the way emissiondata is found in the barbreakdownchart (barbreakdownchart line 96-105).

VIKTORVAV99:
Let's merge #6642 first then I'll do a full in depth review but so far I don't see anything that needs changing.

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -192,3 +193,60 @@ export const getExchangesToDisplay = (
...
+    }
+  }
+}
+
+function getProductionSourcesToAdd(
+  entry: [string, string],
+  productionSourceArray: string[] | undefined,
+  storageType?: boolean
+): string[] {
+  const productionSource = storageType ? entry[0] + ' storage' : entry[0];
```
VIKTORVAV99, line 239: Would really like a
```js
`${}`
```
here instead. Actually somewhat surprised we didn't have a ESLint rule that covered using the addition operator on strings.

VIKTORVAV99 (approved):
Just one minor thing but don't think there is much else that need noting.

Nice work!

PR #6664, opened 2024-04-15
feat(md): add fetch_exchange_forecast for MD parser

amv213:
## Issue

As suggested in https://github.com/electricitymaps/electricitymaps-contrib/pull/6660

## Description

This PR adds the ability to fetch exchange forecasts for the MD parser.

### Double check


parsers/MD.py
```diff
@@ -72,17 +72,17 @@ class ArchiveDatapoint(NamedTuple):
...
+    planned_consumption: float | None
+    production: float | None
+    planned_production: float | None
+    tpp: float | None  # production from thermal power plants
+    hpp: float | None  # production from hydro power plants
+    res: float | None  # production from renewable energy sources
+    exchange_UA_to_MD: float | None
+    planned_exchange_UA_to_MD: float | None
+    exchange_RO_to_MD: float | None
+    planned_exchange_RO_to_MD: float | None
```
amv213, line 85: note: it looks like the API can return null values, so I have updated the code to make sure we can handle them without crashing

e.g. https://moldelectrica.md/utils/archive2.php?id=table&type=array&date1=14.04.2024&date2=16.04.2024

parsers/MD.py
```diff
@@ -263,16 +267,38 @@ def fetch_exchange(
...
+            netflow = (
+                entry.exchange_RO_to_MD
+                if use_actual
+                else entry.planned_exchange_RO_to_MD
+            )
+            if netflow is None:
+                continue
+
+            if netflow != 0:
+                netflow *= -1
```
VIKTORVAV99, line 301: There is a lot of duplicate code here, first of all we should instantiate the netflow variable outside the if statements. Then we can move out the 
```python
if netflow != 0:
    netflow *= -1
```
and only call it once.

This check should also be changed into:
```python
if netflow != 0 and not None:
    netflow *= -1
```

then we can remove the:
```python
if netflow is None:
    continue
```
checks as the parser class handles None values internally.

VIKTORVAV99 (commented):
Just a small note really.

parsers/MD.py
amv213: Thanks for the suggestion! Something like this? 

```python
for entry in archive_data:

    if sorted_zone_keys == ZoneKey("MD->UA"):
        netflow = (
            entry.exchange_UA_to_MD
            if use_actual
            else entry.planned_exchange_UA_to_MD
        )

    elif sorted_zone_keys == ZoneKey("MD->RO"):
        netflow = (
            entry.exchange_RO_to_MD
            if use_actual
            else entry.planned_exchange_RO_to_MD
        )
    else:
        raise NotImplementedError(f"{sorted_zone_keys} pair is not implemented")

    if netflow != 0 and netflow is not None:
        netflow *= -1

    exchange_list.append(
        zoneKey=sorted_zone_keys,
        datetime=entry.datetime,
        netFlow=netflow,
        source=SOURCE,
        sourceType=event_type,
    )
```

VIKTORVAV99: Exactly but you need to add `netflow = None` before the first if statement otherwise you can run into type errors with it not being defined.

amv213: Perfect, added! (I thought the `raise` in the final catch-all `else:` block would have forced `netflow` to always be defined)

VIKTORVAV99: It's quite possible it would, I didn't test it locally but I also know that some type checking systems have a hard time narrowing down types in some scenarios so better to just define it.

(I've been having related issues with that in TypeScript lately)

VIKTORVAV99 (approved):
LGTM!

Thanks for adding this!

PR #6668, opened 2024-04-17
feat(ch): update CH parser with 2023 and 2024 data

amv213:
## Description

I noticed that the hard-coded solar capacity data used by the CH parser might be a bit out of date. Similarly to what was done in https://github.com/electricitymaps/electricitymaps-contrib/pull/5529, this PR adds values for 2023 and 2024 and updates the existing values based on the latest data that I found available on the dashboard linked to in the code comments.

P.S. dashboards / data would also be available on the same website for hydro, wind, and biomass capacities. Is this something we would be interested in integrating too? 

- https://www.uvek-gis.admin.ch/BFE/storymaps/EE_Elektrizitaetsproduktionsanlagen/data/wind.csv 
- https://www.uvek-gis.admin.ch/BFE/storymaps/EE_Elektrizitaetsproduktionsanlagen/data/biomass.csv
- https://www.uvek-gis.admin.ch/BFE/storymaps/EE_Elektrizitaetsproduktionsanlagen/data/water.csv

And if it can be useful, the data that powers the pie-chart showing the current-year capacities is here:
- https://www.uvek-gis.admin.ch/BFE/storymaps/EE_Elektrizitaetsproduktionsanlagen/data/daten_elektrizitaetsproduktionsanlagen.csv

What's nice is that it includes capacities for sources (e.g. nuclear) that don't otherwise have their own dashboard, but we cannot get any historical data beyond the current year

### Double check


parsers/CH.py
```diff
@@ -9,32 +10,60 @@
...
+        16.994,2005,850
+        19.241,2006,1044
+        26.007,2007,1514
+        42.472,2008,2725
+        80.181,2009,4627
+        123.568,2010,6489
+        250.203,2011,11138
+        500.19,2012,18076
+        806.067,2013,25407
+        995.055,2014,32790
```
VIKTORVAV99, line 38: While great that you found all this data we don't need anything before 2015-01-01 as that is the absolute earliest we have some confidence in the data we are collecting and flowtracing. (2015 is when the ENTSO-E transparency platform launched).

VIKTORVAV99:
Sorry about taking a while to review this, we found a bug in the logic related to the capacities that #6703 should temporarily fix.

I will be taking some extra care when reviewing this so we don't reintroduce the error while we are looking for a more permanent fix but that means it can take a bit longer to be reviewed.

VIKTORVAV99 (commented):
Since this returns floats instead of int's it should work without issue with the backend, however there is a merge conflict which needs to be resolved.

Once that is sorted I'll I'd be happy to merge this :)

madsnedergaard (approved):
I fixed the merge conflict, so looks good to me now :)

PR #6673, opened 2024-04-17
chore: upgrade to pnpm 9

VIKTORVAV99:
## Description

Update our projects to use PNPM version 9

### Double check


web/package.json
```diff
@@ -190,6 +190,6 @@
   },
   "engines": {
```
madsnedergaard, line 191: I tried with the newly supported `packageManager` field, but it requires a specific version and doesn't allow pinning just the major or minor version :(
```
  "packageManager": "pnpm@9.0.2",
```

madsnedergaard (approved):

madsnedergaard:
Merging this now as deployments are blocked by it :)

PR #6674, opened 2024-04-18
feat: add productionSourceLegendList to BreakDownChart

silkeholmebonnen:
## Issue

[AVO-160](https://linear.app/electricitymaps/issue/AVO-160/add-the-production-source-legend-under-origin-of-electricity-graph)
## Description

The goal of this PR is to add a list of the production source legends below the breakdownchart.

### Preview

image

### Double check


silkeholmebonnen:
I was thinking about if we should add a tooltip to the legends below the graphs, but I'm not sure what the content of the tooltip should be. What do you think @madsnedergaard @VIKTORVAV99 ?

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
```diff
@@ -9,16 +9,18 @@ export default function ProductionSourceLegend({
...
-    <g className="pointer-events-none">
-      <rect
-        fill={modeColor[electricityType as ElectricityModeType]}
-        width={14}
-        height={14}
-        rx={2}
-      />
-      <g transform={`translate(3, 3)`} width={14} height={8}>
-        <ProductionSourceIcon source={electricityType} />
+    <svg width={14} height={14}>
```
madsnedergaard, line 12: Could we make them a bit bigger? On my non-retina screen they become really hard to discern :)
Screenshot 2024-04-18 at 11 19 07

I guess it's the same in the bar-breakdown-chart at the top, think @Alportan discussed it a some point, but can't remember what the conclusion was 🤔

Alportan: Good point! Yes, I do have another size for the icons. Maybe we can update it in the main component instead so it will reflect on all uses 🙌 

The new size is 16x16 instead, maybe we can try that and see if it helps in this case! 😊

silkeholmebonnen: I have changed them to be 16x16 now. What do you think?

image

Alportan: Looking good! I think it's already better 🙌 

@madsnedergaard what are your thoughts?

madsnedergaard (approved):
Looks great to me! 🥳

PR #6677, opened 2024-04-18
add unit below the graph titles

silkeholmebonnen:
## Issue

[AVO-186](https://linear.app/electricitymaps/issue/AVO-186/add-units-below-the-graph-titles)

## Description

The goal of this PR is to add units below graph titles.

### Double check


web/src/features/charts/graphUtils.ts
```diff
@@ -112,6 +113,31 @@ export const getNextDatetime = (datetimes: Date[], currentDate: Date) => {
...
+    ? getUnit(formatPower(getTotalElectricityAvailable(currentZoneDetail, mixMode)))
+    : getUnit(formatEnergy(getTotalElectricityAvailable(currentZoneDetail, mixMode)));
+}
+
+function getUnit(valueAndUnit: any) {
+  const regex = /\s+(.+)/;
+  const match = valueAndUnit.match(regex);
+  return match[1];
+}
+
```
silkeholmebonnen, line 140: I couldn't figure out a good way to determine the unit of the BarBreakDownChart, so it is a bit hacky... Suggestions are very welcome:)

VIKTORVAV99:
> Why is the e2e failing? 

Unknown but I'll look into it. 👍🏼

VIKTORVAV99:
> Why is the e2e failing?

The reason for the e2e tests failure was that we were entering the mouse on the graph and then reading the expected element. When you then removed the units to the right the graph expanded to fill that void shifting the value the mouse was hovering on ever so slightly so it went into the neighbouring value instead.

Since this was somewhat expected logically, we just didn't consider it, I have updated the value so the e2e tests should pass now.

web/public/locales/en.json
```diff
@@ -176,6 +176,10 @@
       "body": "The data consists of an aggregation of hourly values. {{percentage}}% of the production values are estimated."
     }
   },
+  "units": {
+    "g-kwh": "g / kWh",
```
VIKTORVAV99, line 180: I think this is where the unit is defined so commenting here.

It should be CO2eq / kWh to match the tooltip here:
image

web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
VIKTORVAV99, line 1: I can't comment directly on the affected lines in here but this still have the units in the axis. We should probably remove them to keep it consistent.
image

web/public/locales/en.json
VIKTORVAV99: I would also not add these as translations but rather as constants instead as these should not change between languages as they are base SI units, prefixes and chemical compositions.

web/src/features/charts/graphUtils.ts
```diff
@@ -112,6 +113,34 @@ export const getNextDatetime = (datetimes: Date[], currentDate: Date) => {
...
+  currentZoneDetail: ZoneDetail,
+  mixMode: Mode,
+  timeAverage: TimeAverages
+) {
+  if (displayByEmissions) {
+    return getUnit(
+      formatCo2(getTotalEmissionsAvailable(currentZoneDetail, mixMode)) +
+        ' ' +
+        t('ofCO2eq')
+    );
```
madsnedergaard, line 127: Won't the output of this one always be the unit you're giving it, in which case we could just return that?
```suggestion
    return  t('ofCO2eq')
```

web/src/features/charts/graphUtils.ts
madsnedergaard: I think it's fine to do it this way.

The alternative would be to refactor the logic in `formatPower` and `formatEnergy` functions so we have a separate function for getting the unit which we could then reuse here (see pseudo-code example below), but I don't think that's necessarily worth it for this :)


```js
function calculatePowerUnit(value) {
  // Move logic from formatPower here
}
function formatPower(value) {
  const formattedValue = ...
  const unit = calculatePowerUnit(formattedValue)
  return `${formattedValue} ${unit}`
}

File: web/src/features/charts/graphUtils.ts (hunk 3 above)
Comment (silkeholmebonnen) [reply web/src/features/charts/graphUtils.ts:127]: The translation here is `of CO₂eq` but the formatCo2 function adds the number of tons e.g. `6 kt` so it is `6 kt of CO₂eq`  and then the get unit removes the number so it ends up being `kt of CO₂eq` like the unit is on the tooltip.

File: web/public/locales/en.json (hunk 2 above)
Comment (silkeholmebonnen) [reply web/public/locales/en.json:180]: Okay, I just used `g / kWh` because that was the unit next to the graph before, but I agree, it makes more sense to use `gCO2eq / kWh`. I'll move it out of the translation files.

File: web/src/features/charts/graphUtils.ts (hunk 1 above)
Comment (silkeholmebonnen) [reply web/src/features/charts/graphUtils.ts:143]: This might be a better approach, but the `formatPower` uses `d3-format` which adds the M, G, depending on the number, before W so it kind of does both things at the same time. So I couldn't really figure out a way to separate it into two different functions.

File: web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
Comment (silkeholmebonnen) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: In the design @Alportan didn't remove these, but he did on the charts. Do you think we should remove it here as well? @madsnedergaard @VIKTORVAV99

Comment (VIKTORVAV99) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: Id want to keep it consistent between the graphs so my cote is to remove them.
I would not consider it a blocker for this PR though.

Comment (Alportan) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: I'm fine with removing them. I removed the ones on the other charts as they were taking horizontal space. I was in a very mobile focused state of mind so I mainly caught the ones that looked odd 😅

Comment (silkeholmebonnen) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: I don't think we should remove the units from the axis because of this example, where there are both used `t`, `kt` and `Mt`. As or what the unit below the title should say in this case I'm not sure. Maybe just `t of CO2eq` ?
image

Comment (Alportan) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: Good catch! Should this graph follow a similar logic as we currently do in the production graph? There we use 0.2GW 0.4 GW etc instead of 200MW. 400MW etc. (keeping 1 unit for the whole graph)? 🙋

Comment (silkeholmebonnen) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: I think I like `200 kt` ... `1 Mt` better than `0.2 Mt` ... `1 Mt`, but I'm not sure. What do you think @VIKTORVAV99 @tonypls?

Comment (VIKTORVAV99) [reply web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx:1]: I would prefer a single unit I think but let's table that for a follow up PR and get this merged first.

This is more of a visual preference than a usability and understanding problem I think.

File: web/src/features/charts/graphUtils.ts [hunk 4]
```diff
@@ -4,9 +4,10 @@ import { bisectLeft } from 'd3-array';
 // // https://observablehq.com/@d3/d3-selection-2-0
 import { scaleTime } from 'd3-scale';
 import { pointer } from 'd3-selection';
-import { TFunction } from 'i18next';
+import { t, TFunction } from 'i18next';
```
VIKTORVAV99, line 7: t here should be imported and created bu the useTranslation hook I believe. I'm not sure it would update properly otherwise when the user changes language.

web/src/features/charts/graphUtils.ts
VIKTORVAV99: Let's keep it like this for now and maybe add a task for it either the next cycle if you think it's important enough or just as a slack week item.

web/src/features/charts/graphUtils.ts
silkeholmebonnen: good catch!

VIKTORVAV99:
> Why is e2e failing again? Heeelp @VIKTORVAV99

I'm taking a look 🙂

silkeholmebonnen:
> > Why is e2e failing again? Heeelp @VIKTORVAV99
> 
> Not exactly sure where but it seems like the hooks you in the latest commit execute conditionally causing this:
> 
> ```
> (uncaught exception)Error: Rendered more hooks than during the previous render.
> (uncaught exception)Error: Rendered more hooks than during the previous render.
> ```
> 
> Kind of why we need that other PR 😅

So should we just try to get the other PR merged and then see if that helps?

tonypls:
> > > > Why is e2e failing again? Heeelp @VIKTORVAV99
> > 
> > 
> > > 
> > 
> > 
> > > Not exactly sure where but it seems like the hooks you in the latest commit execute conditionally causing this:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > (uncaught exception)Error: Rendered more hooks than during the previous render.
> > 
> > 
> > > (uncaught exception)Error: Rendered more hooks than during the previous render.
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > Kind of why we need that other PR 😅
> > 
> > 
> > So should we just try to get the other PR merged and then see if that helps?
> 
> It should tell us what the issue is at least, otherwise I have time to take a deeper look tomorrow.

I was about to say hooks (in this case the `useTranslation`) hook can only be called from components (the files should start with a capital letter if it's a component) but it looks like you found the problem yourself 🎉

silkeholmebonnen:
It seems be happy when the `t` is given as a property. I think it's also better practice to do it like that

silkeholmebonnen:
> > > > > Why is e2e failing again? Heeelp @VIKTORVAV99
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Not exactly sure where but it seems like the hooks you in the latest commit execute conditionally causing this:
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > (uncaught exception)Error: Rendered more hooks than during the previous render.
> > > 
> > > 
> > > > (uncaught exception)Error: Rendered more hooks than during the previous render.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Kind of why we need that other PR 😅
> > > 
> > > 
> > > So should we just try to get the other PR merged and then see if that helps?
> > 
> > 
> > It should tell us what the issue is at least, otherwise I have time to take a deeper look tomorrow.
> 
> I was about to say hooks (in this case the `useTranslation`) hook can only be called from components (the files should start with a capital letter if it's a component) but it looks like you found the problem yourself 🎉

Ahh okay I didn't know that hooks could only be called from components but that makes sense. I think I need to read a bit about hooks:)

tonypls (approved):
Nice work, I've done a user test and is working great 🎉

PR #6679, opened 2024-04-19
fix(ntesmo): fix data fetching for AU-NT parser

amv213:
## Issue

Might close https://github.com/electricitymaps/electricitymaps-contrib/issues/6024 

## Description

I was playing around with the `AU-NT` parser and I noticed I had some problems getting data for some dates. This PR should fix the issues:

- the 'current year' data at https://ntesmo.com.au/data/daily-trading is paginated. The existing code was parsing just the first page so requesting data for any day of the current year that was not within the last few days present on that first page would fail

- some consumption data can have some disclaimers at the bottom, which crashed some of the existing logic as those rows ended up into the parsed dataframe

image

P.S. coincidentally https://github.com/electricitymaps/electricitymaps-contrib/issues/6024 just popped up this morning, but not sure if it's for these same reasons or not 

### Double check


parsers/NTESMO.py
```diff
@@ -71,90 +69,91 @@ class Generator(TypedDict):
...
-    return index
-
+    this_year = target_datetime.year == today_date.year
+    if this_year:
+        # current year's report cards are paginated (9 per page)
+        num_pages = int(math.ceil(365 / 9))
+        urls = (
+            f"{API_URL}?result_70160_result_page={page_number}"
+            for page_number in range(1, num_pages + 1)
+        )
```
amv213, line 92: thought: we could maybe even use some logic to pre-compute which page_number the data for the requested day would fall in, and query that page directly

But maybe that would be too risky?

parsers/lib/config.py
```diff
@@ -16,6 +17,7 @@ def refetch_frequency(frequency: timedelta):
     assert isinstance(frequency, timedelta)
 
     def wrap(f):
+        @wraps(f)
```
amv213, line 20: note: [small best practice](https://realpython.com/primer-on-python-decorators/#a-few-real-world-examples) / allows to access the underlying unwrapped function with `f.__wrapped__`

VIKTORVAV99: Even if this is the best practice and probably something we should do it should not be done in this PR.

Please submit a separate PR for it as touching the core code that is shared between all there parsers need extra consideration and testing before merging.

VIKTORVAV99 (commented):
Looks good to me but to avoid this failing with historical data we should update it to handle plants that are missing from the mapping in a better way.

I think we should just add the non matching plants to unknown and log a warning.

For reference:
```
electricitymap-contrib-py3.10➜  web git:(pr/amv213/6679) poetry run test_parser AU-NT --target_datetime="2022-03-11"
WARNING:test_parser:Session already has adapters, they will be overriden.
  File "<string>", line 1, in <module>
  File "/Users/viktor/Repos/electricitymaps-contrib/.venv/lib/python3.10/site-packages/click/core.py", line 1134, in __call__
    return self.main(*args, **kwargs)
  File "/Users/viktor/Repos/electricitymaps-contrib/.venv/lib/python3.10/site-packages/click/core.py", line 1059, in main
    rv = self.invoke(ctx)
  File "/Users/viktor/Repos/electricitymaps-contrib/.venv/lib/python3.10/site-packages/click/core.py", line 1401, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/viktor/Repos/electricitymaps-contrib/.venv/lib/python3.10/site-packages/click/core.py", line 767, in invoke
    return __callback(*args, **kwargs)
  File "/Users/viktor/Repos/electricitymaps-contrib/test_parser.py", line 66, in test_parser
    res = parser(
  File "/Users/viktor/Repos/electricitymaps-contrib/parsers/lib/config.py", line 22, in wrapped_f
    result = f(*args, **kwargs)
  File "/Users/viktor/Repos/electricitymaps-contrib/parsers/lib/config.py", line 46, in wrapped_f
    result = f(*args, **kwargs)
  File "/Users/viktor/Repos/electricitymaps-contrib/parsers/NTESMO.py", line 304, in fetch_production
    return parse_production_mix(production_mix, logger)
  File "/Users/viktor/Repos/electricitymaps-contrib/parsers/NTESMO.py", line 186, in parse_production_mix
    raise ParserException(
parsers.lib.exceptions.ParserException: NTESMO.py Parser: New generator {'Unnamed: 26'} detected in AU-NT, please update the mapping of generators.
```

parsers/lib/config.py
amv213: Sorry about that - I have reverted this change 🙌🏽 

Should I remove it from the `retry_policy` decorator down below as well? I think that that one is being used only in `NTESMO`, but happy to get rid of it there too. 

The motivation was to have access to the raw `__wrapped__` function in the tests in order to not have the mocked sessions overwritten by the decorator.

amv213:
Thanks @VIKTORVAV99 for the reproducible example! I have updated the code to just warn on an unknown power plant, rather than erroring out. 

(Looking into it, in reality the error was caused by the opposite situation: the 2022 data has data for less power plants than those that we know of today (as they didn't exist). I fixed it by making sure we drop those "Unnamed" / empty columns before parsing.)

At the same time, I have also noticed that we were missing out on data from newer power plants, because the excel sheet was being read only up to column `AA`, but nowadays there is data up to `AE`. I have updated to read those columns too from the spreadsheet, and added the new plants to the mapping.

parsers/NTESMO.py
```diff
@@ -71,90 +75,91 @@ class Generator(TypedDict):
...
-            index[dt.date()] = a["href"]
-    return index
+    today_date = datetime.now(AUSTRALIA_TZ).date()
+    target_datetime = target_datetime.astimezone(AUSTRALIA_TZ)
 
+    # cap target datetime if more recent that what the API makes available:
+    # data is being published after 2 days at the moment
+    _DELAY_IN_DAYS = 2
+    if abs(today_date - target_datetime.date()).days < _DELAY_IN_DAYS:
+        target_datetime = target_datetime - timedelta(days=_DELAY_IN_DAYS)
```
amv213, line 89: thought: should we shift this by `_DELAY_IN_DAYS` + 1 ?

It looks like e.g. today (Sun 21 08:00 UTC / 17:30 ACST ) there is still no data for Fri 19  ACST. So maybe it means that data gets published on the 3rd ACST day?

parsers/lib/config.py
VIKTORVAV99: It makes sense but I'm not exactly sure how it will impact things on the backend so I'd rather have it in a separate PR I can test out locally with just these changes.

The best would be to move them both to a separate PR.

Thanks in advance!

parsers/NTESMO.py
```diff
@@ -177,93 +183,151 @@ def parse_consumption(
...
-                    ]
-        production_mix.append(data_point)
-    return production_mix
+        production_breakdown_list.append(production_breakdown)
+
+    return production_breakdown_list
 
 
 @refetch_frequency(timedelta(days=1))
-@retry_policy(retry_policy=retry_strategy)
```
VIKTORVAV99, line 223: Why where these moved/removed?

I'm not sure it will work as expected if they are as the backend only really look at the top function declarations (the ones that are imported for the parser to run).

amv213:
Ok - one last update (hopefully)! 

I have noticed some weird behaviour where they seem to have just bulk uploaded (somewhere around Monday 22 April ~23:00 ACST) data for the 19th, 20th, and 21st ... 

The https://ntesmo.com.au/data/daily-trading page is actually still showing me [including the python script] data only up to the 18th, and I can only see the new data (sometimes) in an incognito window or using a secret "page 0" https://ntesmo.com.au/data/daily-trading?result_70160_result_page=0 (not available through the normal carousel at the bottom)...

All of this to say that we may not be able to rely on a fixed delay of 2 days in data uploads, so I have modified the code so that we actually don't need to hardcode any of this logic: while looking through the data available, we just make a note of which day is the latest one available and we return that one if requesting data for a later target date.

Hopefully this should handle things more gracefully!

parsers/NTESMO.py
```diff
@@ -177,93 +204,154 @@ def parse_consumption(
...
+    # some older data may miss columns for newer plants
+    raw_production_mix_df = raw_production_mix_df.dropna(axis=1, how="all")
+
+    generation_units = set(raw_production_mix_df.columns)
+    generation_units.remove("Period Start")
+    generation_units.remove("Period End")
+    unknown_units = generation_units.difference(PLANT_MAPPING.keys())
+    logger.warning(
+        f"New generator(s) {unknown_units} detected in AU-NT, please update the mapping of generators."
+    )
```
VIKTORVAV99, line 222: This should probably be guarded behind an if statement to check if the set is empty or not first.

parsers/NTESMO.py
```diff
@@ -71,90 +76,112 @@ class Generator(TypedDict):
...
+        page = response.text
+        soup = BeautifulSoup(page, "html.parser")
+        for a in soup.find_all("a", href=True):
+            if a["href"].startswith("https://ntesmo.com.au/__data/assets/excel_doc/"):
+                timestamp = a.find("div", {"class": "smp-tiles-article__title"}).text
+                dt = (
+                    datetime.strptime(timestamp, "%d %B %Y")
+                    .replace(tzinfo=AUSTRALIA_TZ)
+                    .date()
+                )
```
VIKTORVAV99, line 102: When using `--target_datetime=2018-03-11` for example this returns multiple strings of dates which is causing the date parsing to fail.

VIKTORVAV99 (commented):
Some comments relating to historical data fetching, real time seems to be working as expected.

parsers/NTESMO.py
amv213: Looks like there are a bunch of non-standard cards on the website whose `article__title` has some text in addition to the data or e.g. have malformatted dates  👀 

```
# examples
"05 April 2018 - Publishing Recommencement"
"05 January 2018 - System Notice"
"18 June2018"
```

I'll try to go through all years and see if I can implement some parsing logic that can shield us from all of those

amv213: Ok - I have updated the code to be able to handle all (known) special cases!

parsers/NTESMO.py
```diff
@@ -95,11 +96,24 @@ def _scan_daily_report_urls(
...
+
+                try:
+                    date_time = datetime.strptime(timestamp, "%d %B %Y")
+                except ValueError:
+                    # some dates are not formatted properly or will contain extra information
+                    # e.g. "05 April 2018 - Publishing Recommencement", "05 January 2018 - System Notice", "18 June2018"
+                    # so extract individual %d %B %Y groups, potentially separated by 0 or more spaces
+                    match = re.search(r"(\d{2})\s*([a-zA-Z]+)\s*(\d{4})", timestamp)
+                    if not match:
+                        raise
```
amv213, line 108: I thought that here we might as well just re-raise the original `ValueError` from the failed `datetime.strptime(timestamp, "%d %B %Y")` attempt as it already contains a useful error message about the problem

VIKTORVAV99 (approved):
LGTM!

Thanks you for doing this! Hopefully it fixes the issue with this parser being very smappy with it's errors 😅

PR #6680, opened 2024-04-19
Switch app-backend domain

madsnedergaard:
## Description

This would make the app use the app-backend.electricitymaps.com domain that uses CloudFlare to give us better caching opportunities, lower latency and other improvements.

Fixes: AVO-144
Fixes: AVO-174


VIKTORVAV99 (approved):
LGTM!

VIKTORVAV99:
BTW with it being proxied through CloudFlare we can now almost get realtime information about the cache status (even if there is no caching yet.

image

PR #6690, opened 2024-04-22
fix tooltip positioning so it never goes out of the screen

silkeholmebonnen:
## Issue

The chart tooltip currently gets displayed outside of the screen at times, see image below. 

image

## Description

This PR fixes this issue, so it will be placed at the top if it otherwise would have been outside of the screen. 

### Preview

image

### Double check

### DIFF
--- a/web/src/components/tooltips/utilities.ts
+++ b/web/src/components/tooltips/utilities.ts
@@ -48,7 +48,7 @@ export const getOffsetTooltipPosition = ({
 
   const tooltipPosition = {
     x: mousePositionX + xOffset,
-    y: mousePositionY - yOffset,
+    y: Math.max(mousePositionY - yOffset, 0),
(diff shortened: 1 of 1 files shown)


silkeholmebonnen:
> There is already a PR open for this at #6359 which I believe is a little more comprehensive. Would love your review on that and then perhaps we can merge it instead?
> 
> Although I don't see a problem in merging this first and then the other either.

I think the only difference between the two is that his PR stops at the when it hits the header, where this PR stops at the screen. So if we prefer to get it to stop at the header we could merge his PR afterwards, but it might be nice to merge this until we merge his PR.

VIKTORVAV99 (approved):
LGTM!

I think we should merge the PR later on as well though so it stops at the header as that should be the end of the "app content".

PR #6692, opened 2024-04-23
refactor(parser validation): Remove required mode rule

unitrium:
## Issue

The required mode rule has been moved to the quality layer in the back-end therefore there's no need for it in the parser space anymore.

## Description

Removing the required mode validation rule from all parsers.

### Double check

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -363,25 +363,18 @@ VALIDATIONS: dict[str, dict[str, Any]] = {
     # It will still work if data is present but 0.
     # "expected_range" and "floor" only count production and storage
     # - not exchanges!
-    "AT": {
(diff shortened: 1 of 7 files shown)


parsers/ENTSOE.py
```diff
@@ -1001,14 +941,6 @@ def validate_production(
 
     if validation_criteria:
         return validate(datapoint, logger=logger, **validation_criteria)
-
-    # NOTE: Why are there sepcial checks for these zones?
-    if zone_key.startswith("DK-"):
```
pierresegonne, line 1006: Aren't we removing all validation for these zones?

VIKTORVAV99: They don't have any other validation going on so it shouldn't be a problem.

PR #6694, opened 2024-04-24
NZ parser: Remove usage of arrow (refactor) and add snapshot test

Jonmcd1:
## Issue

[#6135](https://github.com/electricitymaps/electricitymaps-contrib/issues/6135)

## Description

This PR removes usage of arrow in NZ.py and uses datetime instead. 
It also adds a snapshot test case (test_NZ.py), which tests NZ.py on two different snapshots of each of the two URLs it uses. This test case can be run by running `python -m unittest parsers/test/test_NZ.py` from the root folder.

### Double check


VIKTORVAV99 (approved):
LGTM!

Thanks for doing this and special thanks for adding the tests!

PR #6695, opened 2024-04-25
feat(web): show exchange arrows after closing side panel

tonypls:
## Issue
Mobile exchange arrows are not present when navigation back from the side panel

## Description
This PR updates the map moving atom when navigating back so the exchange arrows become visible again after navigating back from the full screen side panel


web/src/features/panels/LeftPanel.tsx
```diff
@@ -97,9 +97,10 @@ function MobileHeader() {
 
 function OuterPanel({ children }: { children: React.ReactNode }) {
   const [isOpen, setOpen] = useAtom(leftPanelOpenAtom);
-  const onCollapse = () => setOpen(!isOpen);
   const location = useLocation();
 
+  const onCollapse = () => setOpen(!isOpen);
+
```
VIKTORVAV99, line 103: Im curious, doe the location of this matter? or was it just a preference change?

VIKTORVAV99 (approved):
Looks good just one small question really.

web/src/features/panels/LeftPanel.tsx
tonypls: Doesn't matter 😆 I originally thought the fix might come from this file so I had changed it a bit. My ordering logic was bunch together variable definitions then have a line between functions defined within a component

silkeholmebonnen (approved):
LGTM!

PR #6696, opened 2024-04-26
remove arrow NG.py

lin0110:
## Description

remove arrow usage from NG.py

### Double check

- [O] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [O] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.


parsers/NG.py
```diff
@@ -42,19 +43,20 @@ def fetch_production(
...
-    timestamp = (
-        arrow.get(target_datetime)
-        .to("Africa/Lagos")
-        .replace(minute=0, second=0, microsecond=0)
+
+    if target_datetime is None:
+        target_datetime = datetime.now(timezone.utc).isoformat()
+    timestamp = datetime.fromisoformat(target_datetime).replace(
+        minute=0, second=0, microsecond=0, tzinfo=TIMEZONE
     )
```
VIKTORVAV99, line 51:
```suggestion
    if target_datetime is None:
        target_datetime = datetime.now(timezone.utc)
    timestamp = target_datetime.replace(minute=0, second=0, microsecond=0).astimezone(
        TIMEZONE
    )
```
We don't need to convert the object to a string and then back to a datetime. Also since we are using arrows `.to()` function we should use the equivelant datetime function called `.astimezone()`.

VIKTORVAV99 (commented):
Hi and welcome to the Electricity Contrib repository.

Thanks for creating a PR for this.
I have a small comment with a suggestion but besides that this looks good!

lin0110:
Is there anything else I need to do?

VIKTORVAV99:
> Is there anything else I need to do?

Nope, I just missed that this PR had been updated. In the future feel free to re-request a review then it will show up as an action needed from my side. 🙂

lin0110:
If possible, please review it quick. If this pull request is being accepted, I can get extra credits for my course, which is going to end. And thanks for the help you have provided so far!! I really appreciate it!

VIKTORVAV99 (approved):
LGTM!

Thanks for doing this!

PR #6697, opened 2024-04-27
fix(options requests): Fix API request in getFeatureFlags function

VIKTORVAV99:
## Issue
For every request we send with authentication headers we also create a OPTIONS request, this request can't be cached and is creating unneeded server load as this endpoint is not even authenticated on the backend.

## Description

Removes the extra headers causing the browser to trigger the OPTIONS requests for /feature-flags.

### Double check


VIKTORVAV99:
I have no idea what is going on with the component test as it is passing locally.

tonypls:
> I have no idea what is going on with the component test as it is passing locally.

I've tried re-adding the request method to see if that helps cypress 🤷

VIKTORVAV99:
I'm gonna open a PR to update cypress, maybe that will help.

tonypls:
Interesting that we only use auth for a our backend tests 😆 

I'm curious why auth was removed and if it's something we want again.

VIKTORVAV99:
Fixed some black magic level stuff happening, removed all errors, made it run locally in both electron and edge...


...and it still won't work in the CI!

VIKTORVAV99:
I just noticed we aren't intercepting this the same way as with the other endpoints. More specifically it doesn't have a registered alias. In theory it shouldn't matter but I'll try that and see if that makes any difference.

VIKTORVAV99:
How would you guys feel about temporarily disabling these tests until I have migrated them to E2E tests?

I would like to get this out there so we can reduce the amount of requests the server is doing as fast as possible.

The actual get requests are not impacted in any way by this PR.

web/src/features/feature-flags/api.ts
```diff
@@ -1,17 +1,13 @@
...
 
 export async function getFeatureFlags(): Promise<FeatureFlags> {
   const path: URL = new URL(`/${QUERY_KEYS.FEATURE_FLAGS}`, getBasePath());
-  const requestOptions: RequestInit = {
-    method: 'GET',
-    headers: await getHeaders(path),
-  };
 
   try {
-    const response = await fetch(path, requestOptions);
```
madsnedergaard, line 14: This is one option, but would it be better to enable auth in the backend instead of removing it? 🤔

VIKTORVAV99: No cause that would still trigger options requests, which I want to avoid if it's not needed.

Right now every request we make with custom headers triggers two requests from the browser. One OPTIONS request to see if the action is allowed and then the GET request.

These OPTIONS request are now the primary cause of the spikes in the app-backend cause they cannot be cached by cloudflare.

madsnedergaard: Hmm yeah okay, I see - guess we won't be hurt by DDOS if it's cached on Cloudflare 😅

madsnedergaard:
> How would you guys feel about temporarily disabling these tests until I have migrated them to E2E tests?
> 
> I would like to get this out there so we can reduce the amount of requests the server is doing as fast as possible.
> 
> The actual get requests are not impacted in any way by this PR.

If fine with skipping e2e here in favor of getting this released before tomorrow :)

madsnedergaard (approved):

PR #6699, opened 2024-04-27
fix(options requests): Fix OPTIONS request for state

VIKTORVAV99:
## Issue

Same as for feature flags, every time we are fetching a state we create 2 requests, one OPTIONS request that can't be cached and one GET request that we are able to cache. This increases server load more than needed as these are not authenticated requests.

## Description

Removes the authentication headers from state requests to prevent OPTIONS requests.

### Double check

### DIFF
--- a/web/src/api/getState.ts
+++ b/web/src/api/getState.ts
@@ -10,12 +10,8 @@ import { cacheBuster, getBasePath, getHeaders, QUERY_KEYS } from './helpers';
 const getState = async (timeAverage: string): Promise<GridState> => {
   const path: URL = new URL(`v8/state/${timeAverage}`, getBasePath());
   path.searchParams.append('cacheKey', cacheBuster());
-  const requestOptions: RequestInit = {
-    method: 'GET',
-    headers: await getHeaders(path),
-  };
 
-  const response = await fetch(path, requestOptions);
+  const response = await fetch(path);
 
   if (response.ok) {
     const result = (await response.json()) as GridState;


VIKTORVAV99:
@tonypls would you be comfortable merging this without us figuring out #6697?

tonypls:
> @tonypls would you be comfortable merging this without us figuring out #6697?

Yeah sound good! I'm a bit worried that it's going to be hard to quickly identify regressions after the spike in error rate we saw on deployment yesterday. [Issue here](https://linear.app/electricitymaps/issue/AVO-233/improve-how-we-handle-distributing-web-app-updates) 

Perhaps we can use this change to do a small deployment and see if we can gain any more insight into the cause of the increased error rate.

VIKTORVAV99:
> > @tonypls would you be comfortable merging this without us figuring out #6697?
> 
> 
> 
> Yeah sound good! I'm a bit worried that it's going to be hard to quickly identify regressions after the spike in error rate we saw on deployment yesterday. [Issue here](https://linear.app/electricitymaps/issue/AVO-233/improve-how-we-handle-distributing-web-app-updates) 
> 
> 
> 
> Perhaps we can use this change to do a small deployment and see if we can gain any more insight into the cause of the increased error rate.

Hmm yeah that sounds good, and I'd be happy to pick up the service worker parts again.

I know we were worried about the ability to roll back but I really don't think that is an issue or at least not big enough to be a blocker for it.

tonypls:
> > > @tonypls would you be comfortable merging this without us figuring out #6697?
> > 
> > 
> > Yeah sound good! I'm a bit worried that it's going to be hard to quickly identify regressions after the spike in error rate we saw on deployment yesterday. [Issue here](https://linear.app/electricitymaps/issue/AVO-233/improve-how-we-handle-distributing-web-app-updates)
> > Perhaps we can use this change to do a small deployment and see if we can gain any more insight into the cause of the increased error rate.
> 
> Hmm yeah that sounds good, and I'd be happy to pick up the service worker parts again.
> 
> I know we were worried about the ability to roll back but I really don't think that is an issue or at least not big enough to be a blocker for it.


That would be awesome if you could prioritize app delivery, we get lots of errors after trying to read the client_version. I'm not sure service worker is going to solve this so feel free to re-work this approach if there's a better way to handle it. 

Now the rewrite is complete I think there's less concern about the ability to rollback (we did think by pointing to the instance of the old app I believe).

tonypls (approved):
LGTM

VIKTORVAV99:
> > > > @tonypls would you be comfortable merging this without us figuring out #6697?
> > > 
> > > 
> > > Yeah sound good! I'm a bit worried that it's going to be hard to quickly identify regressions after the spike in error rate we saw on deployment yesterday. [Issue here](https://linear.app/electricitymaps/issue/AVO-233/improve-how-we-handle-distributing-web-app-updates)
> > > Perhaps we can use this change to do a small deployment and see if we can gain any more insight into the cause of the increased error rate.
> > 
> > 
> > Hmm yeah that sounds good, and I'd be happy to pick up the service worker parts again.
> > I know we were worried about the ability to roll back but I really don't think that is an issue or at least not big enough to be a blocker for it.
> 
> That would be awesome if you could prioritize app delivery, we get lots of errors after trying to read the client_version. I'm not sure service worker is going to solve this so feel free to re-work this approach if there's a better way to handle it.
> 
> Now the rewrite is complete I think there's less concern about the ability to rollback (we did think by pointing to the instance of the old app I believe).

When we are using a service worker I don't think we will need that part at all as it will have it's own list of assets with hashes and use that to check if there is a new version of the files in question using.

PR #6703, opened 2024-04-29
fix(CH.py): fix capacity type for downstream insert

pierresegonne:
## Issue

The current implementation casts the capacity value to `int64` -> that causes ingestion issues downstream

## Description

* Use instead floats, which seem to be supported

### Double check


pierresegonne:
We'll be fixing the issue downstream

PR #6704, opened 2024-04-29
chore(deps): bump ruff to 0.4

VIKTORVAV99:
## Description

Astral has released a new "major" version of ruff. We should upgrade to get the latest benefits which include a 20–40% speedup for something that is already crazy fast 😍.

The release blog post are here:
https://astral.sh/blog/ruff-v0.4.0

And the actual changelog are here:
https://github.com/astral-sh/ruff/releases/tag/v0.4.0

I will create a secondary PR that does this in the backend as well 🙂

### Double check


VIKTORVAV99:
I'll resolve the conflicts as soon as possible but they aren't really conflicts at all so it's fine to review as is.

tonypls (approved):
LGTM

PR #6707, opened 2024-05-01
feat(web): Record time slider interactions with plausible

tonypls:
## Issue
We don't know how users interact with out time slider

## Description
This PR sends a custom event to Plausible for interactions with the time slider component, I've alos created a goal and custom property to record the unique interactions with the time slider and property of selected index.

### DIFF
--- a/web/src/components/TimeSlider.tsx
+++ b/web/src/components/TimeSlider.tsx
@@ -74,7 +74,7 @@ export const getThumbIcon = (
 
 function trackTimeSliderEvent(selectedIndex: number, timeAverage: TimeAverages) {
   trackEvent('Time Slider Button Interaction', {
-    selectedIndex: `${timeAverage} :${selectedIndex}`,
+    selectedIndex: `${timeAverage}: ${selectedIndex}`,
   });
 }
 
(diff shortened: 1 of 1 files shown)


web/src/components/TimeSlider.tsx
```diff
@@ -88,7 +93,10 @@ export function TimeSliderBasic({
       max={numberOfEntries}
       step={1}
       value={selectedIndex && selectedIndex > 0 ? [selectedIndex] : [0]}
-      onValueChange={(value) => onChange(value[0])}
+      onValueChange={(value) => {
+        onChange(value[0]);
+        trackTimeSliderEvent(value[0]);
```
silkeholmebonnen, line 98: As far as I can tell `value[0]` is just a number, e.g. for hourly its 0-24 and monthly 0-12, but how do we know if that number corresponds to hourly, daily, monthly or yearly? Or are we not interested in that?

silkeholmebonnen (commented):
Looks good, just one comment:)

web/src/components/TimeSlider.tsx
tonypls: Good point, I think we should include the time aggregate too. I'll update the PR to include it

PR #6708, opened 2024-05-01
feat(web): Record language selection in Plausible

tonypls:
## Issue
We don't know much about our users preferred languages

## Description
This PR posts an event to Plausible when a language is selected.

I've also created the following events in Plausible:
- Custom goal `Language Selected`
- Custom property `language`

### DIFF
--- a/web/src/features/map-controls/LanguageSelector.tsx
+++ b/web/src/features/map-controls/LanguageSelector.tsx
@@ -3,6 +3,7 @@ import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { HiLanguage } from 'react-icons/hi2';
 import { languageNames } from 'translation/locales';
+import trackEvent from 'utils/analytics';
 
 import MapButton from './MapButton';
 import MapOptionSelector from './MapOptionSelector';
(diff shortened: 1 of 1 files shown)


silkeholmebonnen:
Just checked in plausible and I cannot see a goal called `Language Selected`. I should be able to see it under `site settings` -> `goals`, right? Or am I looking at the wrong place

tonypls:
> Just checked in plausible and I cannot see a goal called `Language Selected`. I should be able to see it under `site settings` -> `goals`, right? Or am I looking at the wrong place

It should be there, make sure you're looking at the app.electricitymaps.com page. Another way to test it is to visit this branch preview on a browser without ad blocker (I just use incognito mode) and change the language, then see if it's present as an event.

silkeholmebonnen:
Ahh I was looking at electricitymaps.com page. How can i add app.electricitymaps.com so I can see it in Plausible?

tonypls:
> Ahh I was looking at electricitymaps.com page. How can i add app.electricitymaps.com so I can see it in Plausible?

I've invited you

PR #6711, opened 2024-05-01
chore(web): remove mouseflow script

tonypls:
## Issue
We currently have disabled Mouseflow but we're still running it in the app scripts
image

## Description
This PR removes the mouseflow script as it currently fails so all our users are downloading mouseflows js without any use.

I think it's best to clean this up for now and add it or something else again in the future if we want it.


madsnedergaard (approved):
Makes sense :)

PR #6713, opened 2024-05-02
feat(web): add tracking events for data source accordions

silkeholmebonnen:
## Issue

[AVO-212](https://linear.app/electricitymaps/issue/AVO-212/add-tracking-events)
## Description

This PR posts an event to Plausible when a data source accordion is clicked, with a property `chart` to get information about which chart's accordion that has been clicked. 

### Double check


web/src/features/charts/BreakdownChart.tsx
```diff
@@ -110,6 +111,13 @@ function BreakdownChart({
       />
       <Divider />
       <Accordion
+        onClick={() => {
+          trackEvent('Data Sources Clicked', {
```
tonypls, line 115: Now we're getting quite a few of these I'm wondering if it would be nice to have an enum to centralized them all in our constant.ts file. What do you think?

web/src/features/charts/EmissionChart.tsx
```diff
@@ -59,6 +60,9 @@ function EmissionChart({ timeAverage, datetimes }: EmissionChartProps) {
       />
       <Divider />
       <Accordion
+        onClick={() => {
```
tonypls, line 63: These will all trigger on open and close, could be nice to just capture open.

web/src/features/charts/BreakdownChart.tsx
VIKTORVAV99: I love enums so I'm in favour of this. It usually also means the app bundles get smaller once it's used more than 2-3 times.

VIKTORVAV99: Agreed!

VIKTORVAV99 (approved):
Code wise I think it looks good!
I would just double check that these populate as expected (use the preview link and click on them a bit and see that the plausible stuff shows up where it should 😅) before merging this.

silkeholmebonnen:
> Code wise I think it looks good! I would just double check that these populate as expected (use the preview link and click on them a bit and see that the plausible stuff shows up where it should 😅) before merging this.

Works as expected!

PR #6715, opened 2024-05-02
chore(mockserver): Update mock data, cleanup old data, use gfs v8

VIKTORVAV99:
## Issue

We have migrated to V8 for all endpoints, and the mock server were missing the `/feature-flags` endpoint. The current V8 mock data was also missing a few keys.

## Description
This updates all the v8 data, removes the old data and adds a empty response to the `/feature-flags` endpoint so it don't give a 404 error.

This is needed to create more extensive and updated e2e tests.

### Double check


madsnedergaard (approved):
LGTM :)

PR #6716, opened 2024-05-03
Update the dependabot config to use multi directory support

VIKTORVAV99:
## Issue
We where duplicating some of the configs.

## Description

Update the dependant config to take advantage of the new multi directory support so we don't have to store multiple duplicate configs.

### Double check


silkeholmebonnen (commented):
Nice work, just a couple of questions:)
What configs where we duplicating?
Why can we delete the commented out code?

VIKTORVAV99:
> Nice work, just a couple of questions:) What configs where we duplicating? Why can we delete the commented out code?

The configs we where duplicating is the commented out code.

Before Dependabot didn't support multi directory configs so we needed one config entry per directory. Now they do so we can remove the old duplicate code.

I also re-enabled the mockserver because it has a low amount of dependencies (just 3 I think) and they are not business critical even if something where to go wrong.

silkeholmebonnen (approved):
Thank you for explaining @VIKTORVAV99, I think it looks good! :sun_with_face:

PR #6721, opened 2024-05-04
feat(app): Update capacitor to the latest version

VIKTORVAV99:
## Issue

Part of https://linear.app/electricitymaps/issue/AVO-236/add-privacy-manifest-to-ios-app

Closes: AVO-241

## Description
In order for us to be fully compliant we need to update the capacitor versions, I figured that might make it a good opportunity to bump the major versions as well.

### iOS
Note I could not build this locally for production for iOS, IDK if this is a recent change on Apple side or if there is an issue with my local install, but it needs to be tested.

This is probably just a signing issue and won't affect things later on.
image

Building for testing seems to work though so it should work just fine for production as well.
Test build:
image

### Android
The android app builds and runs just fine in the emulator.

### Double check


VIKTORVAV99:
Since this includes a shift to JAVA 17 it could potentially help speed things up on android devices as well but that's not really something I have been able to test.

madsnedergaard:
Will try to test this thoroughly later today :)

madsnedergaard (approved):
Looks good to me on simulators!
I think it's worth merging this and then doing a beta build on both platforms that we test thoroughly in order to better test in a production environment :)

VIKTORVAV99:
> Looks good to me on simulators!
> 
> I think it's worth merging this and then doing a beta build on both platforms that we test thoroughly in order to better test in a production environment :)

Agreed!

PR #6728, opened 2024-05-06
feat(web): add production source legends to data source accordion for carbon and emission chart

silkeholmebonnen:
## Issue

[AVO-84](https://linear.app/electricitymaps/issue/AVO-84/add-the-production-source-legend-under-data-sources-for-emission)

## Description

The goal of this PR is to add the relevant production source legends to the data sources for emission factors, so you are able to see which sources are used for which production sources.

### Preview
image

### Double check


tonypls:
Should this also show under the production breakdown in the sources there too?

silkeholmebonnen:
> Should this also show under the production breakdown in the sources there too?

Yes! I totally forgot about that. I'll take a look at it now

silkeholmebonnen:
It should be in the bar-breakdown chart as well now:)

VIKTORVAV99:
> Would it make sense that the power generation and emission factor data sources are including under all three charts?
> 
> Also is it correct that multiple emission factors have coal and gas for example in an hourly view? I would assume there is only one emission factor used for each production mode at hourly granularity

IMO Not really.
The carbon intensity graph for example is dictated by the exchanges to a large amount, especially in Europe. And it'd rather show no sources for that particular graph than incomplete sources if that makes sense?

silkeholmebonnen:
> > Would it make sense that the power generation and emission factor data sources are including under all three charts?
> > Also is it correct that multiple emission factors have coal and gas for example in an hourly view? I would assume there is only one emission factor used for each production mode at hourly granularity
> 
> In my opinion not really. The carbon intensity graph for example is dictated by the exchanges to a large amount, especially in Europe. And it'd rather show no sources for that particular graph than incomplete sources if that makes sense?

So which graphs would you show sources for? And which sources would you show?

VIKTORVAV99:
> > > Would it make sense that the power generation and emission factor data sources are including under all three charts?
> 
> > > Also is it correct that multiple emission factors have coal and gas for example in an hourly view? I would assume there is only one emission factor used for each production mode at hourly granularity
> 
> > 
> 
> > In my opinion not really. The carbon intensity graph for example is dictated by the exchanges to a large amount, especially in Europe. And it'd rather show no sources for that particular graph than incomplete sources if that makes sense?
> 
> 
> 
> So which graphs would you show sources for? And which sources would you show?

Preferably we should only show the sources that are relevant for each graph without causing additional cognitive load. So if a source is not directly relevant we should not show it. Therefore I don't think it makes sense to show say the emission factors on the emission chart as it only aggregated values. Where the data comes from makes sense but then it should not be listed as production sources in my opinion.

silkeholmebonnen:
> > > > Would it make sense that the power generation and emission factor data sources are including under all three charts?
> > 
> > 
> > > > Also is it correct that multiple emission factors have coal and gas for example in an hourly view? I would assume there is only one emission factor used for each production mode at hourly granularity
> > 
> > 
> > > 
> > 
> > 
> > > In my opinion not really. The carbon intensity graph for example is dictated by the exchanges to a large amount, especially in Europe. And it'd rather show no sources for that particular graph than incomplete sources if that makes sense?
> > 
> > 
> > So which graphs would you show sources for? And which sources would you show?
> 
> Preferably we should only show the sources that are relevant for each graph without causing additional cognitive load. So if a source is not directly relevant we should not show it. Therefore I don't think it makes sense to show say the emission factors on the emission chart as it only aggregated values. Where the data comes from makes sense but then it should not be listed as production sources in my opinion.



> > > > Would it make sense that the power generation and emission factor data sources are including under all three charts?
> > 
> > 
> > > > Also is it correct that multiple emission factors have coal and gas for example in an hourly view? I would assume there is only one emission factor used for each production mode at hourly granularity
> > 
> > 
> > > 
> > 
> > 
> > > In my opinion not really. The carbon intensity graph for example is dictated by the exchanges to a large amount, especially in Europe. And it'd rather show no sources for that particular graph than incomplete sources if that makes sense?
> > 
> > 
> > So which graphs would you show sources for? And which sources would you show?
> 
> Preferably we should only show the sources that are relevant for each graph without causing additional cognitive load. So if a source is not directly relevant we should not show it. Therefore I don't think it makes sense to show say the emission factors on the emission chart as it only aggregated values. Where the data comes from makes sense but then it should not be listed as production sources in my opinion.

What do you think @madsnedergaard @tonypls @Alportan ?

tonypls:
Make sense to not show in the cases where it's aggregated from imports. It's usually better to over communicate rather than under communicate, I would suggest showing both the emission factors and 

Would this make sense?

Sources: EF = Emission Factor, P = Power, C = Capacity

Electricity charts consumption toggle on:
- Consumption by source: EF, P, C
- Carbon intensity: P
- Origin of electricity: EF, P

Electricity charts production toggle on:
- Consumption by source: EF, P, C
- Carbon intensity: EF, P
- Origin of electricity: EF, P

Emissions charts consumption toggle on:
- Consumption by source: EF, P, C
- Carbon emissions: P
- Origin of electricity: EF, P

Emissions charts production toggle on:
- Consumption by source: EF, P, C
- Carbon emissions: EF, P
- Origin of electricity: EF, P

silkeholmebonnen:
We discussed it in the office and decided that we will get this out as a v1 and then optimize it later. We can discuss v2 it in the next team sync when you're there @VIKTORVAV99

silkeholmebonnen:
@tonypls I have removed the data sources when the graph is disabled. It only happens in the 'Daily carbon emission origin' graph (`BreakdownChart.tsx` with consumption and emissions toggle) right?

web/src/features/charts/bar-breakdown/DataSources.tsx
```diff
@@ -1,25 +1,65 @@
...
+            {sourceToProductionSources.get(source)?.map((productionSource, index) => (
+              <span key={index} className="self-center object-center text-xs">
+                <ProductionSourceLegend
+                  electricityType={productionSource as ElectricityModeType}
+                  transform="translate(0,3.5)"
+                />
+              </span>
+            ))}
+          </span>
+        </p>
```
VIKTORVAV99, line 61: The sources and text are not aligned here, the source icons are placed higher up than the text. 
image

I suspect we have to use something other than `items-baseline`.

VIKTORVAV99 (commented):
I'm still not sold on including it in the carbon emissions graph as none of the sources are actually represented in any of the data we show:
image

web/src/features/charts/bar-breakdown/DataSources.tsx
silkeholmebonnen: It doesn't look like that for me... weird. Does it still look like that?

madsnedergaard:
> I'm still not sold on including it in the carbon emissions graph as none of the sources are actually represented in any of the data we show

Hmm, in that case we also shouldn't show the Production source as well on the co2 intensity graph. 

I'm in favor of adding the whole thing and then reconsidering later if it becomes too much or too confusing :)

web/src/features/charts/bar-breakdown/DataSources.tsx
silkeholmebonnen: Okay in safari it looks weird but in chrome it looks good... I'll look into it

silkeholmebonnen: Should be fixed now!

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
```diff
@@ -5,11 +5,13 @@ import ProductionSourceIcon from './ProductionsSourceIcons';
...
 export default function ProductionSourceLegend({
   electricityType,
+  transform,
 }: {
   electricityType: ElectricityModeType;
+  transform?: string;
 }) {
   return (
-    <svg width={16} height={16}>
+    <svg width={16} height={16} transform={transform}>
```
tonypls, line 14: Is this transform ever passed in?

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -233,35 +247,30 @@ function updateMapWithSources(
...
 ): string[] {
-  const productionSource = storageType ? `${entry[0]} storage` : entry[0];
-  const sourceInfo = entry[1];
-
-  if (sourceInfo.startsWith('assumes')) {
+  if (emissionFactorSource.startsWith('assumes')) {
     return [];
-  }
-  if (productionSourceArray === undefined) {
+  } else if (productionSourceArray === undefined) {
```
tonypls, line 270: You could double equals here to be more flexible with the falseyness

web/src/features/charts/hooks/useBarElectricityBreakdownChartData.ts
```diff
@@ -56,12 +57,15 @@ export default function useBarBreakdownChartData() {
   );
   const height = isConsumption ? exchangeY + exchangeHeight : exchangeY;
 
+  const emissionSourceToProductionSource = getEmissionData(zoneData);
```
tonypls, line 60: I'm not sure on the naming here, is this a function that converts an emission source to a production source? If it's a variable then the word "to" implies it converts and could be rephrased

web/src/features/charts/bar-breakdown/ProductionSourceLegend.tsx
silkeholmebonnen: True. I have removed it now.

web/src/features/charts/hooks/useBarElectricityBreakdownChartData.ts
silkeholmebonnen: I have named it `sourceProductionSourceMapping` now. I searched a bit on naming conventions for Maps and a lot of suggestions have "to" in it. I think that `sourceToProductionSourceMapping` might be better, what do you think?

tonypls (commented):
Looks great, I think we're almost there! Thanks for iterating many times 🙌 

I think the sources here might look better left aligned:
image

VIKTORVAV99:
> Looks great, I think we're almost there! Thanks for iterating many times 🙌
> 
> I think the sources here might look better left aligned: image

What do you mean by left aligned? No indentation?
If so that was discussed here: https://github.com/electricitymaps/electricitymaps-contrib/pull/6584#discussion_r1541474399

tonypls:
> > Looks great, I think we're almost there! Thanks for iterating many times 🙌
> > I think the sources here might look better left aligned: image
> 
> What do you mean by left aligned? No indentation? If so that was discussed here: [#6584 (comment)](https://github.com/electricitymaps/electricitymaps-contrib/pull/6584#discussion_r1541474399)

Ah I didn't see that comment but the indentation here seems to have crept even more to the right, to the point where the final source takes 3 lines.

VIKTORVAV99:
> > > Looks great, I think we're almost there! Thanks for iterating many times 🙌
> > > I think the sources here might look better left aligned: image
> > 
> > 
> > What do you mean by left aligned? No indentation? If so that was discussed here: [#6584 (comment)](https://github.com/electricitymaps/electricitymaps-contrib/pull/6584#discussion_r1541474399)
> 
> Ah I didn't see that comment but the indentation here seems to have crept even more to the right, to the point where the final source takes 3 lines.

I see what you mean now, it looks like the indentation has almost doubled. That we should fix directly but we should hold of on the full left.

madsnedergaard (approved):
Looks good to me :)

PR #6733, opened 2024-05-10
tmp-fix(FR parser): switch FR production parser to ENTSO-E

unitrium:
## Issue

From @florianscheidl's investigations: the FR parser is currently not working returning only timestamps and no data. Because of the long week-end we cannot expect any fix before next week earliest. 

## Description

Switch temporarely FR to ENTSO-E until we have an update on the parser.

### Double check

### DIFF
--- a/config/zones/FR.yaml
+++ b/config/zones/FR.yaml
@@ -359,7 +359,7 @@ parsers:
   consumptionForecast: ENTSOE.fetch_consumption_forecast
   generationForecast: ENTSOE.fetch_generation_forecast
   price: ENTSOE.fetch_price
-  production: FR.fetch_production
+  production: ENTSOE.fetch_production
   productionCapacity: ENTSOE.fetch_production_capacity
   productionPerModeForecast: ENTSOE.fetch_wind_solar_forecasts
 price_displayed: false
--- a/poetry.lock
+++ b/poetry.lock
@@ -1386,4 +1386,4 @@ scripts = ["xmltodict"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">= 3.10, < 3.11"
-content-hash = "ad142d6791f2052f4bbb070fdfbe59d31bc441fa1391e4c0f767af55469b0f31"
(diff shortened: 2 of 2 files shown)


VIKTORVAV99:
I'll regenerate the lock file as that is failing for some reason.

VIKTORVAV99:
Just note that this will be blocked in the monorepo unless the current ruff stuff is merged first.

unitrium:
Which ruff stuff @VIKTORVAV99 ?

VIKTORVAV99:
> Which ruff stuff @VIKTORVAV99 ?

This https://github.com/electricitymaps/electricitymaps/pull/5047

Since the ruff PR has already been merged in contrib it needs the corresponding monorepo changes or the update contrib job won't work due to the lock files being out of sync.

unitrium:
Ok then let's get them merged asap if we can as FR is a priority zone, otherwise if it's breaking too many things in the mono repo we should revert the ruff PR.

VIKTORVAV99:
> Ok then let's get them merged asap if we can as FR is a priority zone, otherwise if it's breaking too many things in the mono repo we should revert the ruff PR.

As far as I can tell it's just small formatting and syntax changes. And should not be breaking but I didn't want to merge it without another set of eyes on it first.

florianscheidl (approved):
LGTM, I checked if it would make sense to change to the ENTSO-E consumption data as well but they also do not have consumption data atm, so only changing production seems good 👍.

florianscheidl:
I'd support fixing the parser on the monorepo for now and merging contrib early next week to avoid things breaking on long-weekend Friday?

florianscheidl:
> > Ok then let's get them merged asap if we can as FR is a priority zone, otherwise if it's breaking too many things in the mono repo we should revert the ruff PR.
> 
> As far as I can tell it's just small formatting and syntax changes. And should not be breaking but I didn't want to merge it without another set of eyes on it first.

Or we can go for that 👍

VIKTORVAV99 (approved):
Let's merge this either way, I don't expect the ruff PR to be a issue.

PR #6744, opened 2024-05-13
feat(parsers): Allow the usage of wind_onshore and wind_offshore internally

VIKTORVAV99:
## Description

This these changes need more testing to ensure it works for all current implementation and works well with the update methods.

It will serve as the base for supporting both offshore and onshore wind internally in the parsers so we can better handle when one of them is missing.

### Double check
- [ ] I added tests for the relevant changes.

### DIFF
--- a/electricitymap/contrib/lib/models/events.py
+++ b/electricitymap/contrib/lib/models/events.py
@@ -120,6 +120,7 @@ class ProductionMix(Mix):
         """
         if self.wind is not None:
             return self.wind
+
         if self.wind_onshore is not None and self.wind_offshore is not None:
             return _none_safe_round(self.wind_onshore + self.wind_offshore)
         if (
@@ -130,6 +131,16 @@ class ProductionMix(Mix):
             return _none_safe_round(
                 (self.wind_onshore or 0) + (self.wind_offshore or 0)
             )
+        if (
+            "wind_onshore" in self.__fields_set__
+            and "wind_offshore" not in self.__fields_set__
+        ):
+            return self.wind_onshore
+        if (
+            "wind_offshore" in self.__fields_set__
+            and "wind_onshore" not in self.__fields_set__
+        ):
+            return self.wind_offshore
 
     def dict(  # noqa: A003
         self,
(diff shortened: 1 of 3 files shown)


madsnedergaard:
What are your thoughts on doing it at this level, vs just adding something to `query_wind_solar_production_forecast`?
I am unsure if it becomes harder to understand what's going on if it's "hidden" in the model class - e.g. "the parser takes wind_offshore, but why is it not shown in the app?". Might be me overthinking things :)

VIKTORVAV99:
> What are your thoughts on doing it at this level, vs just adding something to `query_wind_solar_production_forecast`?
> 
> I am unsure if it becomes harder to understand what's going on if it's "hidden" in the model class - e.g. "the parser takes wind_offshore, but why is it not shown in the app?". Might be me overthinking things :)
> 
> But doing a quick, temporary "hack" would enable us to think more about the structure and ensuring we build something scalable

Doing it at this level means it will apply to everything use the ProductionMix class including the normal production breakdown as well as we don't want partial wind in there either.

It is also required if the update and merge methods are going to work properly as they are abstracted away from the actual parser (this is the issue we are facing today in part).

madsnedergaard:
Thanks for the clarifying comment, I see you've already thought well on this! ❤️
Let me know if I can help (later tonight), also happy to review and test this :)

VIKTORVAV99:
@madsnedergaard this is starting to get a bit complex and based on how the ENTSO-E parser actually and how it returns data I'm not sure it will actually work like I expected it to. The ENTSO-E API returns every individual mode as a time series with just missing values if they are well missing (compared to N/A and N/E on the web interface (not sure how they do that)). It means it's basically impossible to tell if a mode is temporarily missing (like in the DE situation) or permanently missing (say offshore wind in Luxembourg.

The good news is I found an easier way to do in just the ENTSO-E parser that does work so I think I'll close this in favour of that (I'll wait until the other is merged though).

The general idea of what I want to do here still works on a theoretical basis though but it probably need to run in the backend and I think we should directly ingest more source types like offshore and onshore wind instead of just wind then we could flag this in the outlier detection project with the existing rules we already have. I realise that is a bigger discussion we must have with proper shaping and planning but I think it is something we should prioritise as soon as we can. (It would have other benefits as well).

I'll work on writing a motivational letter/brief for this during slack week and why I think it should be prioritised.

PR #6745, opened 2024-05-13
feat(web): create applied methodologies accordion

silkeholmebonnen:
## Issue

[AVO-189](https://linear.app/electricitymaps/issue/AVO-189/create-the-applied-methodologies-accordion)

## Description

This PR creates the applied methodologies accordion which links to the methodology page of our website. 

### Preview

image
Screenshot 2024-05-13 at 14 44 25

### Double check


web/src/features/panels/zone/MethodologyCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+            className={`text-sm font-semibold `}
+          >
+            <span>{t('left-panel.applied-methodologies.flowtracing')}</span>
+          </a>
+          <a
+            href="https://www.electricitymaps.com/methodology#carbon-intensity-and-emission-factors"
+            target="_blank"
+            rel="noreferrer"
+            className={`text-sm font-semibold `}
+          >
```
VIKTORVAV99, line 38: Should we add some form of hover effect here to indicate these are external links like we do on the buttons in the navbar?
image

silkeholmebonnen: Personally I don't think so because it is a link and not a button. What do you think @Alportan ?

Alportan: Agreed with Silke, since they are simple links, it should be fine. We might thing of adding an icon in the future if needed to indicate that it opens an external website, but I don't think it's a blocker for this PR 🙌

web/src/features/charts/bar-breakdown/GraphCard.tsx
```diff
@@ -1,6 +1,6 @@
 import { twMerge } from 'tailwind-merge';
 
-export function GraphCard({
+export function RoundedCard({
```
tonypls, line 3: Be nice to keep the file name the same as the exported component when possible 🛠️

VIKTORVAV99 (approved):
LGTM!

I took care of the merge conflict (that I caused with another PR) so should be good to go now. Could you please wait with merging this until we have had a chance to deploy the current changes in the update contrib PR though?

PR #6754, opened 2024-05-15
fix(web): axis units overflowing graphcard

silkeholmebonnen:
## Issue

These two pictures show 2 issues related to axis units overflowing.

This issue is caused by the `formatCo2` function not handling negative values correctly, partly because the value to match set in `NetExchangeChart` was not taking into account that the number could be negative, and partly because the method itself didn't account for negative values. 
image
The second issue is that the gray overlay was not wide enough to cover the axis units.
image

## Description

This PR fixes these two issues so that the axis show negative values correctly and it widens the gray overlay.

### Preview

image
image

### Double check


web/src/features/charts/BreakdownChart.tsx
```diff
@@ -75,7 +75,7 @@ function BreakdownChart({
       <div className="relative">
         {isBreakdownGraphOverlayEnabled && (
           <div className="absolute top-0 h-full w-full">
-            <div className=" h-full w-full bg-white opacity-50 dark:bg-gray-800" />
+            <div className=" h-full w-full scale-x-150 bg-white opacity-50 dark:bg-gray-800" />
```
silkeholmebonnen, line 78: I think this is a bit of a hack but I couldn't really figure out a better way to make the child div be wider than its parent.

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -20,7 +20,7 @@ import GraphHoverLine from './GraphHoverline';
 import ValueAxis from './ValueAxis';
 
 const X_AXIS_HEIGHT = 20;
-const Y_AXIS_WIDTH = 20;
+const Y_AXIS_WIDTH = 26;
```
VIKTORVAV99, line 23: Just an idea here, but not sure if it should be in this PR or just a future improvement but it would be nice to calculate the actual with needed instead of having to use a constant.

VIKTORVAV99 (dismissed):
LGTM!

web/src/features/charts/elements/AreaGraph.tsx
silkeholmebonnen: Yes that would be nice! In the design review meeting we discussed units which will also affect how much space is needed, so I think it is a future thing.

silkeholmebonnen:
I just changed some things after we discussed it in the meeting. I should have let you know, sorry!

silkeholmebonnen:
It should be ready for review again now!

web/src/features/charts/DisabledMessage.tsx
```diff
@@ -0,0 +1,7 @@
+export function DisabledMessage({ message }: { message: JSX.Element | string }) {
```
tonypls, line 1:
```suggestion
export function DisabledMessage({ message }: { message: JSX.Element }) {
```

Should always return an element right?

web/src/features/charts/BreakdownChart.tsx
```diff
@@ -72,14 +73,15 @@ function BreakdownChart({
...
-          <div className="absolute top-0 h-full w-full">
-            <div className=" h-full w-full bg-white opacity-50 dark:bg-gray-800" />
-            <div className="absolute left-1/2 top-1/2 z-10 -translate-x-1/2 -translate-y-1/2 whitespace-nowrap rounded-sm bg-gray-200 p-2 text-center text-sm shadow-lg dark:bg-gray-900">
-              Temporarily disabled for consumption. <br /> Switch to production view
-            </div>
-          </div>
+          <DisabledMessage
+            message={
+              <>
+                Disabled for consumption. <br /> Switch to production view
```
tonypls, line 81:
```suggestion
                Disabled for consumption view. <br /> Please switch to production view to enable the origin chart
```
or similar?

VIKTORVAV99: Should this not be a translation string so it can well be translated if we are going to change it anyway?

madsnedergaard:
Wow, so cool how quickly and easily you just fixed the disabled message - you're awesome! 💪

web/src/features/charts/DisabledMessage.tsx
silkeholmebonnen: That is not the return type but the type of `message` . I could change it to only be of type string so we could also keep it as a translation string as @VIKTORVAV99 suggests.

silkeholmebonnen:
Ready for review again ⭐

VIKTORVAV99 (approved):
LGTM! 🎉

PR #6761, opened 2024-05-17
chore(app): Remove SentryRoutes as we don't use performance monitoring

VIKTORVAV99:
## Issue

We don't use performance monitoring with Sentry so we should not import code related to that.

## Description

Removes code and imports we don't use related to Sentry performance monitoring.

Sentry documentation: https://docs.sentry.io/platforms/javascript/guides/react/features/react-router/#react-router-v6

### Preview
The sentry bundle goes from `76.15 kB` to `75.82 kB` and the index bundle goes from `257.26 kB` to `257.23 kB` but there are runtime benefits as well (even though I did not measure this 😅).

### Double check


tonypls (approved):
LGTM, marginal gains but gains all the same

PR #6764, opened 2024-05-18
Add some french translation

phiphou:
- add some French translation (estimation feedback part)
- sort keys on French locale file to follow the default English locale file order in order to ease further translations.


VIKTORVAV99 (approved):
Thanks for adding these!

PR #6766, opened 2024-05-21
Fix issues with graphs on iOS 16 and below

### LINKED ISSUE #6723: The diagrams of 'carbon intensity', 'electricity origin' and 'electricity price' expand infinity to the right.
**To Reproduce**

1. open the website or app.
2. click on a country 
3. scroll a bit down and look at the diagrams. They start 'compressed' on the left and then expand infinitely to the right.

**Expected behavior**
The diagrams shouldn't expand at all but fill the whole width of the screen.

**Screenshots**
image

image

image

**Smartphone**
 - Device: iPhone 7XL
 - OS: iOS 15.8.2
 - Browser: Safari
 - App-Version: 1.129.0

madsnedergaard:
## Issue

Closes #6723

## Description

The native Resize Observer implementation does not work on SVG elements in Safari 16 and below.

### Preview

Screenshot 2024-05-21 at 10 19 25


madsnedergaard:
> I'll try and test this later, I need to download older iOS versions to emulate 😅

@tonypls do you have an older Simulator version installed and can test this?
Otherwise I'd mostly worry about testing that it works for current versions as I have tested on iOS 15 :)

tonypls (approved):
LGTM, tested on both iOS 15 and 17.

One question, we have the .at polyfil in both the index.html and in the util. Do we want this to init twice?

VIKTORVAV99:
> LGTM, tested on both iOS 15 and 17.
> 
> One question, we have the .at polyfil in both the index.html and in the util. Do we want this to init twice?

We do not...

madsnedergaard:
> LGTM, tested on both iOS 15 and 17.
> 
> One question, we have the .at polyfil in both the index.html and in the util. Do we want this to init twice?

Ah great catch, I couldn't understand why we had it in the file but never initiated it 😅

PR #6775, opened 2024-05-24
perf(ENTSOE.py): Adds small perf improvements to the ENTSO-E parser

VIKTORVAV99:
## Description

I needed a small break from reading articles and writing academic stuff so I fixed some performance stuff in the ENTSOE.py parser. Will add a test that double check that the LRU caching works as expected later today.

### Double check


madsnedergaard (approved):
Haven't tested it, but looks like fine from a quick glance :D

PR #6781, opened 2024-05-27
Get caller-location and feature-flags from new endpoint

madsnedergaard:
## Description

The caller-location (used to center the map on your area) is being cached and thus wrong for other users, so we need to move it out of the state files.

And to avoid adding an extra request just for this, we are merging the feature-flags and the caller-location into the same, new `/v8/meta` endpoint.

### Preview

https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/d2bfece6-9e87-4ec2-8b5e-3d098eac138a


web/src/api/getMeta.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+export function useMeta(): Meta {
+  return (
+    useQuery<Meta>([QUERY_KEYS.META], async () => getMeta(), {
+      suspense: true,
+    }).data ?? {}
+  );
+}
```
VIKTORVAV99, line 36: Not sure if it should be here or further downstream but should we guard against localhost/non valid IPs returning Nan as indicated by this sentry issue?

https://electricitymaps.sentry.io/issues/5092639410/?project=4504366922989568&query=is%3Aunresolved+nan&referrer=issue-stream&statsPeriod=90d&stream_index=0

Would be nice to eliminate those when we are tinkering with it.

madsnedergaard: Good idea, done :)

VIKTORVAV99: Thanks! Now let's just hope that is the actual cause 😅

madsnedergaard: Seems like it was mostly coming from a "Facebook Bot" browser, so I assume it's not really affecting any users :)

mockserver/server.js
```diff
@@ -45,8 +45,14 @@ app.get('/v8/gfs/solar', (req, res, next) => {
...
 
-app.get('/feature-flags', (req, res) => {
-  res.json({});
+app.get('/v8/meta', (req, res) => {
+  res.json({
+    features: {
+      'feature-flag-1': true,
+      'flag-2': false,
+    },
+    callerLocation: [0, 0],
```
tonypls, line 54:
```suggestion
    callerLocation: [138, 36],
```
This should help it pass the new cypress test

mockserver/server.js
```diff
@@ -45,16 +45,6 @@ app.get('/v8/gfs/solar', (req, res, next) => {
...
 
-app.get('/v8/meta', (req, res) => {
-  res.json({
-    features: {
-      'feature-flag-1': true,
-      'flag-2': false,
-    },
-    callerLocation: [0, 0],
-  });
-});
```
VIKTORVAV99, line 50: I would keep this in the mockserver anyway so it can be used for local testing by our contributors.

tonypls (approved):
LGTM

mockserver/server.js
tonypls: It's in there! Just not twice, any more 🙈

VIKTORVAV99: oh 😅

VIKTORVAV99: I don't see it though 👀

madsnedergaard: It's loading the JSON file directly if nothing is specifically defined - so a contributor can just change the JSON file and it will update :)

VIKTORVAV99: oh (again), I thought it needed a registered route?

PR #6784, opened 2024-05-28
2024-03 capacity updates for EIA zones

florianscheidl:
## Issue

Capacities for EIA zones were out of date.

## Description

This PR includes capacity updates from March 2024.

### Double check

### DIFF
--- a/config/zones/US-CAL-BANC.yaml
+++ b/config/zones/US-CAL-BANC.yaml
@@ -44,6 +44,9 @@ capacity:
     - datetime: '2023-10-01'
       source: EIA.gov
       value: 325.7
+    - datetime: '2024-03-01'
+      source: EIA.gov
+      value: 327.0
   unknown:
     - datetime: '2023-10-01'
       source: EIA.gov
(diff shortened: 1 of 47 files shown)


VIKTORVAV99:
I have a PR open at #6626 but there was some questions about the aggregated zones. Have you addressed these in this PR or are we in the same boat here?

florianscheidl:
Ah, didn't see the thread, we did not specifically address this issue here.

What is the status of Pierre's message?

> Can you check that it was the case before as well? If it's the case, then let's go ahead with it

VIKTORVAV99:
I have not have had the time to double check if these were manually adjusted before but as far as I could tell we don't aggregate things on a parser level.

Personally I think it makes sense to merge on for them (probably this now as the data should be more accurate) and address that later but I'm unsure how it will affect the new validation rules.

florianscheidl:
Sounds reasonable to me. Indeed it was the validation rules that triggered me to check for updates. 
I agree it would make sense to keep this in mind in the future and merge for now. If we are very sceptical of the way the validation is affected, we could temporarily disable the validation pipeline for a specific set of zones.

PR #6786, opened 2024-05-29
feat(web): add links to data sources

silkeholmebonnen:
## Issue

[AVO-232](https://linear.app/electricitymaps/issue/AVO-232/add-links-to-the-emission-factors-sources)
[AVO-278](https://linear.app/electricitymaps/issue/AVO-278/add-link-to-data-sourcesmd-for-capacity-data)

## Description

This PR adds links to the data sources. 

I made a `Link` component and changed the methodology accordion to use this component so ensure consistency.

### Preview

image

### Double check


silkeholmebonnen:
This PR is made from #6728, and should not be merged until that is merged.

web/src/utils/constants.ts
```diff
@@ -77,3 +77,106 @@ export const timeAxisMapping: Record<TimeAverages, keyof Duration> = {
   monthly: 'months',
   yearly: 'years',
 };
+
```
madsnedergaard, line 80:
```suggestion
/**
 * A mapping between the source name and a link to the source.
 * Sources are coming from zone YAML files (capacity and emission factors) or from parsers (production data).
 * The links have been collected from various sources and are not guaranteed to be up-to-date.
 * TODO: In future each link should ideally be stored in the zone configuration YAML files instead of here,
 * so that the link can be updated along with the source and more easily found and maintained.
 */
```

web/src/features/charts/bar-breakdown/DataSources.tsx
```diff
@@ -70,3 +73,29 @@ function SourcesWithLegends({
...
+  if (!source.includes('.')) {
+    return null;
+  }
+
+  if (source.includes('http')) {
+    return source;
+  }
+
+  return `http://${source}`;
+}
```
madsnedergaard, line 101: We should probably do a quick unit test on this one :)

madsnedergaard (commented):
I changed the base branch to make it easier to see the diff - this means there's some config changes right now, but they will disappear once the other branch is up to date with `master` :)

web/src/features/charts/bar-breakdown/DataSources.tsx
madsnedergaard: Nice, thanks! 🥳

madsnedergaard:
Moving [this discussion](https://github.com/electricitymaps/electricitymaps-contrib/pull/6789#discussion_r1621407656) here:

> Do we want to keep this as a JSON file instead? We keep other config mappings as JSON so it makes sense to do the same here?

I think it would make sense to somehow keep this information available in the zone config files in the future so that you can find the source and a link to it in the same place. This requires some additional work though, so we decided not to do that now. I'd suggest we just keep the map in frontend code like now, instead of doing an intermediate JSON step - and then make a proper YAML setup for it later on :)

web/src/utils/constants.ts
```diff
@@ -77,3 +77,112 @@ export const timeAxisMapping: Record<TimeAverages, keyof Duration> = {
...
   yearly: 'years',
 };
+/**
+ * A mapping between the source name and a link to the source.
+ * Sources are coming from zone YAML files (capacity and emission factors) or from parsers (production data).
+ * The links have been collected from various sources and are not guaranteed to be up-to-date.
+ * TODO: In future each link should ideally be stored in the zone configuration YAML files instead of here,
+ * so that the link can be updated along with the source and more easily found and maintained.
+ */
+export const sourceLinkMapping: { [key: string]: string } = {
```
madsnedergaard, line 87: Moving [this comment](https://github.com/electricitymaps/electricitymaps-contrib/pull/6789#discussion_r1621407656) by @VIKTORVAV99 :

> Do we want to keep this as a JSON file instead? We keep other config mappings as JSON so it makes sense to do the same here?

I think it would make sense to somehow keep this information available in the zone config files in the future so that you can find the source and a link to it in the same place. This requires some additional work though, so we decided not to do that now. I'd suggest we just keep the map in frontend code like now, instead of doing an intermediate JSON step - and then make a proper YAML setup for it later on :)

VIKTORVAV99: Once we have it in the YAML I can do a optimized mapping like I did for the contributors 🙌🏻

web/src/components/Link.tsx
```diff
@@ -0,0 +1,12 @@
+export function Link({ href, linkText }: { href: string; linkText: string }) {
+  return (
+    <a
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
```
madsnedergaard, line 6: Great with a Link component! We could use it in several other places, but that would probably also require more customization and stuff we don't really need now :)

On thing here though, I'm a bit unsure on the `noreferrer` attribute here, it basically means that the "receiver" will not see where the user was coming from. This also means that if we use this component for the links to our commercial website, all traffic will show as "direct" instead of "coming from the app".

web/src/components/Link.tsx
```diff
@@ -0,0 +1,12 @@
...
+      className="text-sm font-semibold text-emerald-800 underline underline-offset-2 dark:text-emerald-500"
+    >
+      {linkText}
```
madsnedergaard, line 9: I would prefer to use `children` here instead, so we're less restrictive on what can be entered and usage feels more natural IMO (e.g. `<Link href="#">My link text</Link>`)

web/src/features/charts/graphUtils.test.ts
```diff
@@ -140,3 +141,42 @@ describe('getTotalEmissionsAvailableOrElectricityAvailable', () => {
...
+    expect(extractLinkFromSource('source2', sourceLinkMapping)).to.equal(
+      'http://mappedlink2.com'
+    );
+  });
+
+  it('real example: should return mapped link if source is in sourceLinkMapping', () => {
+    expect(extractLinkFromSource('Climatescope', sourceLinkMapping)).to.equal(
+      'https://www.global-climatescope.org/'
+    );
+  });
```
madsnedergaard, line 165: Dunno how much it matters, but if we stick to a consistent naming it could be something like this:
```suggestion
  it('should work with a real link', () => {
    expect(extractLinkFromSource('Climatescope', sourceLinkMapping)).to.equal(
      'https://www.global-climatescope.org/'
    );
  });
```

web/src/features/charts/graphUtils.ts
```diff
@@ -210,3 +210,25 @@ export function getBadgeText(chartData: AreaGraphElement[], t: TFunction) {
...
+
+  if (!source.includes('.')) {
+    return null;
+  }
+
+  if (source.includes('http')) {
+    return source;
+  }
+
+  return `http://${source}`;
```
madsnedergaard, line 233:
```suggestion
  // We on purpose don't use https due to some sources not supporting it (and the majority that does will automatically redirect anyway)
  return `http://${source}`;
```

web/src/utils/constants.ts
madsnedergaard: That would be great!

madsnedergaard (commented):
A few small comments, but otherwise this is great 🥳

web/src/components/Link.tsx
silkeholmebonnen: I get this error when I give rel as a prop. So I think we need it to open the page in a new tab.

image

web/src/components/Link.tsx
silkeholmebonnen: Good point!

web/src/components/Link.tsx
madsnedergaard: Let's update the `.eslintrc.json` with this then: 
```json
    "react/jsx-no-target-blank": [
      "error",
      {
        "allowReferrer": true
      }
    ],
```

web/src/components/Link.tsx
```diff
@@ -0,0 +1,20 @@
...
+}: {
+  href: string;
+  rel?: string;
+  children: React.ReactNode;
+}) {
+  return (
+    <a
+      href={href}
+      target="_blank"
+      rel={`noopener ${rel}`}
```
madsnedergaard, line 14: It's currently returning `undefined` and I don't think we need to make this dynamic :)
```suggestion
      rel="noopener"
```

web/src/components/Link.tsx
```diff
@@ -0,0 +1,20 @@
+export function Link({
+  href,
+  rel,
```
madsnedergaard, line 3:
```suggestion
```

web/src/components/Link.tsx
```diff
@@ -0,0 +1,20 @@
...
+  children,
+}: {
+  href: string;
+  rel?: string;
```
madsnedergaard, line 7:
```suggestion
```

madsnedergaard (approved):

PR #6788, opened 2024-05-29
Clean up various links and names of sources

madsnedergaard:
## Issue

@silkeholmebonnen and I have been working on adding links for all data sources (#6786). While working on this, the sources changed here were quite messy and stood out from how we usually format and write sources.

## Description

Updates source URLs and names, to make it more clear what the source is as well as make it easier to link to the sources from the app (#6786).

### Double check


parsers/ajenti.py
```diff
@@ -18,29 +18,33 @@
...
     },
     # Flinders Island
     # https://github.com/electricitymaps/electricitymaps-contrib/issues/2533
     # https://en.wikipedia.org/wiki/Flinders_Island
+    # https://www.hydro.com.au/clean-energy/hybrid-energy-solutions/success-stories/flinders-island
     "AU-TAS-FI": {
         "hub": "flindershub",
         "method": "SendDashboard",
         "tz": "Australia/Hobart",
-        "source": "https://www.hydro.com.au/clean-energy/hybrid-energy-solutions/success-stories/flinders-island",
```
VIKTORVAV99, line 34: You don't want to change these in database too do you? 😅

These are live sources so we will end up with both on the temporal aggregates.

parsers/ajenti.py
```diff
@@ -18,29 +18,33 @@
 from signalr import Connection
 
 ZONE_PARAMS = {
+    # King's Island
```
q--, line 21: * King Island, not King's Island

parsers/ajenti.py
madsnedergaard: I can manually go through the DB entries afterwards to clean them up :)

parsers/ajenti.py
madsnedergaard, line 21: Fixes typo
```suggestion
    # King Island
```

madsnedergaard: Thanks, good catch!

Fixes typo
```suggestion
    # King Island
```

parsers/ajenti.py
VIKTORVAV99: If it's easier we could also do it in the app-backend when we are deduplicating the sources by stripping everything but the domain.

config/zones/QA.yaml
```diff
@@ -4,47 +4,48 @@ bounding_box:
   - - 52.11654707100007
     - 26.660101630000042
 capacity:
+  _comment: Emission factor calculations are available at https://docs.google.com/spreadsheets/d/1ownS3qse6zDG4gNnKEZDroIKLNiqZcULqp34LJUxrL0/edit?usp=sharing
```
VIKTORVAV99, line 7: This is not allowed in our current pydantic config, but I think we should change that rather than removing it.

madsnedergaard: Since we don't show this anywhere, should we just make it a code comment like this instead?
```suggestion
# Capacity calculations are available at https://docs.google.com/spreadsheets/d/1ownS3qse6zDG4gNnKEZDroIKLNiqZcULqp34LJUxrL0/edit?usp=sharing
```

VIKTORVAV99: > Since we don't show this anywhere, should we just make it a code comment like this instead?

Yes but there is a risk some other tool or something will remove it if it's run as a part of a script. Up to you IMO but would be really annoying if that happened and we would probably end up needing to fix it anyway.

madsnedergaard: I moved the comment to top-level, as I am unsure if having the key lower might break some existing functionality that is only expecting actual production modes.

VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99 (approved):
LGTM! 🎉

tonypls (approved):
LGTM 😝

PR #6789, opened 2024-05-30
feat(web): add a tooltip to clarify why production modes can have multiple emission factor sources

silkeholmebonnen:
## Issue

[AVO-256](https://linear.app/electricitymaps/issue/AVO-256/add-a-tooltip-to-clarify-why-production-modes-can-have-multiple)

## Description

The goal of this PR is to add a info icon with a tooltip to clarify why production modes can have multiple emission factor sources.

This PR is dependent on #6728, and should not be merged until that has been merged. 

### Preview

Web:
image
image

Mobile:
image
image

### Double check


web/src/utils/constants.ts
```diff
@@ -77,3 +77,106 @@ export const timeAxisMapping: Record<TimeAverages, keyof Duration> = {
   monthly: 'months',
   yearly: 'years',
 };
+
+export const sourceLinkMapping: { [key: string]: string } = {
```
VIKTORVAV99, line 81: Do we want to keep this as a JSON file instead?
We keep other config mappings as JSON makes sense to do the same here?

silkeholmebonnen: @madsnedergaard What do you say?

silkeholmebonnen: This discussion should probably take place in #6786

madsnedergaard: [Moved here](https://github.com/electricitymaps/electricitymaps-contrib/pull/6786#discussion_r1625698767)

web/src/features/charts/DataSources.tsx
```diff
@@ -53,6 +88,19 @@ export function DataSources({
...
 
+function EmissionFactorTooltip({ t }: { t: TFunction<'translation', undefined> }) {
+  return (
+    <Portal.Root className="pointer-events-none absolute left-0 top-0 z-50 flex h-full w-full flex-col content-center items-center justify-center bg-black/20 pb-40">
+      <div className="dark:border-1 relative mx-6 h-auto min-w-64 rounded-xl border bg-zinc-50 p-4 text-left text-sm opacity-80 shadow-md dark:border-gray-700 dark:bg-gray-900">
+        {t('country-panel.emissionFactorDataSourcesTooltip')}
+      </div>
+      <button className="p-auto pointer-events-auto mt-2 flex h-10 w-10 items-center justify-center self-center rounded-full border bg-zinc-50 text-black shadow-md sm:hidden">
+        <HiXMark size="24" />
+      </button>
```
madsnedergaard, line 99: This button is currently hidden on sizes where it should still be visible - perhaps we can remove the `sm:hidden` here as the whole thing is only visible on smaller screens? :) 

Screenshot 2024-06-18 at 21 34 43

web/src/features/charts/DataSources.tsx
```diff
@@ -53,6 +88,19 @@ export function DataSources({
...
 
+function EmissionFactorTooltip({ t }: { t: TFunction<'translation', undefined> }) {
+  return (
+    <Portal.Root className="pointer-events-none absolute left-0 top-0 z-50 flex h-full w-full flex-col content-center items-center justify-center bg-black/20 pb-40">
+      <div className="dark:border-1 relative mx-6 h-auto min-w-64 rounded-xl border bg-zinc-50 p-4 text-left text-sm opacity-80 shadow-md dark:border-gray-700 dark:bg-gray-900">
+        {t('country-panel.emissionFactorDataSourcesTooltip')}
+      </div>
+      <button className="p-auto pointer-events-auto mt-2 flex h-10 w-10 items-center justify-center self-center rounded-full border bg-zinc-50 text-black shadow-md">
+        <HiXMark size="24" />
+      </button>
```
VIKTORVAV99, line 99: Tiny thing but this button should probably change colour depending on the mode that is used so it matches the other backgrounds.

Right now it looks like this in dark mode:
image
And like this in light mode:
image

silkeholmebonnen: It is a part of the design, so it is actually on purpose. Maybe @Alportan can elaborate on why:)

Alportan: @VIKTORVAV99 after some testing on dark theme, I discovered that people didn't really see the button because of the lack of contrast. So I would keep it white, no matter the selected theme. It might look like a bug, but it's a feature 😄

VIKTORVAV99: Hmm...

I guess that's fine then but it looks weird 😅

madsnedergaard (approved):

PR #6791, opened 2024-05-30
fix(web): change order of production sources

silkeholmebonnen:
## Issue

[AVO-218](https://linear.app/electricitymaps/issue/AVO-218/electricity-production-by-source-update-order-of-electricity-sources)

## Description

The goal of this PR is to change the order of the production sources in the `BarBreakdownChart`.

The new order takes into account renewables first (sorted alphabetically), low carbon (nuclear), fossil fuels (sorted alphabetically) and storage (sorted alphabetically).

image

### Preview

image

### Double check


web/src/utils/constants.ts
```diff
@@ -56,15 +56,15 @@ export const modeColor: { [mode in ElectricityModeType]: string } = {
 };
 
 export const modeOrder = [
```
madsnedergaard, line 58: Adding a comment to explain the logic here so it's clear :)
```suggestion
/**
 * The order of modes use the following logic:
 * 1. renewables first (sorted alphabetically)
 * 2. low carbon (nuclear)
 * 3. fossil fuels (sorted alphabetically)
 * 4. storage (sorted alphabetically)
 */
```

madsnedergaard:
Hmm, I'm wondering if it would be better to combine this with the new colors or not - if we do, we could update the modal message to say that we have revamped colors and the order. On the other hand, it might be harder for people to transition to a completely new thing where all visual cues they were used to have changed. But unless we wait a fairly long time between merging the two PRs, this argument is not too valid 🤔 

What do you think @Alportan?

Alportan:
> Hmm, I'm wondering if it would be better to combine this with the new colors or not - if we do, we could update the modal message to say that we have revamped colors and the order. 

I like the idea of including this change in the modal as well in case we decide releasing them at once. 🌟

>On the other hand, it might be harder for people to transition to a completely new thing where all visual cues they were used to have changed. But unless we wait a fairly long time between merging the two PRs, this argument is not too valid 🤔

Good point. They are somewhat related, so it might make sense to change them both at the same time. It might take some more time for people to get used to the new everything, but it might be better than changing it over 2 separate releases, but I'm not entirely sure. I'm curious what are your thoughts as well @tonypls 🙌

tonypls:
> > Hmm, I'm wondering if it would be better to combine this with the new colors or not - if we do, we could update the modal message to say that we have revamped colors and the order.
> 
> I like the idea of including this change in the modal as well in case we decide releasing them at once. 🌟
> 
> > On the other hand, it might be harder for people to transition to a completely new thing where all visual cues they were used to have changed. But unless we wait a fairly long time between merging the two PRs, this argument is not too valid 🤔
> 
> Good point. They are somewhat related, so it might make sense to change them both at the same time. It might take some more time for people to get used to the new everything, but it might be better than changing it over 2 separate releases, but I'm not entirely sure. I'm curious what are your thoughts as well @tonypls 🙌

I think the idea of communicating this in a modal is a very good one, whether that's part of the colors modal or it's own.

Is this a change we might want to feature flag or A-B test?

VIKTORVAV99:
> > > Hmm, I'm wondering if it would be better to combine this with the new colors or not - if we do, we could update the modal message to say that we have revamped colors and the order.
> 
> > 
> 
> > I like the idea of including this change in the modal as well in case we decide releasing them at once. 🌟
> 
> > 
> 
> > > On the other hand, it might be harder for people to transition to a completely new thing where all visual cues they were used to have changed. But unless we wait a fairly long time between merging the two PRs, this argument is not too valid 🤔
> 
> > 
> 
> > Good point. They are somewhat related, so it might make sense to change them both at the same time. It might take some more time for people to get used to the new everything, but it might be better than changing it over 2 separate releases, but I'm not entirely sure. I'm curious what are your thoughts as well @tonypls 🙌
> 
> 
> 
> I think the idea of communicating this in a modal is a very good one, whether that's part of the colors modal or it's own.
> 
> 
> 
> Is this a change we might want to feature flag or A-B test?

Not exactly sure what we would measure with a A/B test if it's just the modal?

If we are going to A/B test things I think we need to do that on the order itself (and modal) and then use the survey component to see what they think about it.

madsnedergaard (approved):
LGTM, ready to merge with the 2 other PRs

tonypls (approved):
LGTM 👍

PR #6792, opened 2024-05-30
feat(web): add announcement modal for the production source color changes

silkeholmebonnen:
## Issue

[AVO-279](https://linear.app/electricitymaps/issue/AVO-279/implement-the-modal-and-make-it-expire-after-x-months)

## Description

This PR adds a new modal that shows the new color changes to the productions source legends. It will only show until June 30. I don't know if this is the date that we want?
It appears after the onboarding modal, if this is shown.

### Preview

image

### Double check


web/src/icons/announcementSourceColorImage.tsx
silkeholmebonnen, line 1: I added the image Alex made as an SVG. I don't know if we prefer this or an png?

web/src/App.tsx
```diff
@@ -88,6 +89,7 @@ export default function App(): ReactElement {
               </Suspense>
               <Suspense>
                 <OnboardingModal />
+                <AnnouncementModal />
```
silkeholmebonnen, line 92: I was a bit unsure about where to add the `AnnouncementModal` here

web/src/features/modals/AnnouncementModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const hasAnnouncementBeenSeenAtom = atomWithStorage(
+  'announcementSeen',
+  localStorage.getItem('announcementSeen') ?? false
+);
+
+const handleTierGaterClick = (event: React.MouseEvent<HTMLDivElement>) => {
+  event.stopPropagation();
+  event.preventDefault();
+};
```
silkeholmebonnen, line 17: This was the only way that I could get the modal to stay when clicking it, since its parent element has a onClick that makes it dissapper. Suggestions are welcome to do it another way

web/src/features/modals/AnnouncementModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+  // Check if onboarding has been seen in the current session
+  useEffect(() => {
+    const onboardingSeenInSession = sessionStorage.getItem('onboardingSeenInSession');
+    if (onboardingSeenInSession) {
+      setHasOnboardingBeenSeenInSession(true);
+    } else if (!hasOnboardingBeenSeen) {
+      sessionStorage.setItem('onboardingSeenInSession', 'true');
+      setHasOnboardingBeenSeenInSession(true);
+    }
+  }, [hasOnboardingBeenSeen]);
```
silkeholmebonnen, line 38: Me and @tonypls decided that the announcementModal should only be shown if the user has seen the onboarding modal before but not in the current session, as it we would assume that it would be a new user who would not care that we changed the color. 

I couldn't really figure out a better way of doing this, so suggestions to do it another way a very welcome ⭐

web/src/icons/announcementSourceColorImage.tsx
VIKTORVAV99: We prefer SVGs over PNGs in general as SVGs are usually smaller yet sill always crisp on all resolutions.

web/src/App.tsx
VIKTORVAV99: Looks like the correct location but it should probably be in it's own suspense boundary and possible dynamically imported, especially if the SVG is imported directly in the component.

silkeholmebonnen: I tried to put it in its own suspense but the e2e test failed. I think it is because the announcementmodal should only be shown if the onboardingmodal has not been shown in the current session, so it needs to wait fo the onboarding modal to load before it can be loaded correctly.

web/src/features/modals/AnnouncementModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export default function AnnouncementModal() {
+  const { t } = useTranslation();
+  const [hasAnnouncementBeenSeen, setHasAnnouncementBeenSeen] = useAtom(
+    hasAnnouncementBeenSeenAtom
+  );
+  const [hasOnboardingBeenSeen] = useAtom(hasOnboardingBeenSeenAtom);
+  const [hasOnboardingBeenSeenInSession, setHasOnboardingBeenSeenInSession] =
+    useState(false);
+  const isExpired = new Date() > new Date('2024-06-30');
```
madsnedergaard, line 27: Before we actually get this out and mobile builds reviewed and released it might be a bit short time-span, so I suggest we extend it like this
```suggestion
  const isExpired = new Date() > new Date('2024-07-15');
```

web/src/features/modals/AnnouncementModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+  event.stopPropagation();
+  event.preventDefault();
+};
+
+export default function AnnouncementModal() {
+  const { t } = useTranslation();
+  const [hasAnnouncementBeenSeen, setHasAnnouncementBeenSeen] = useAtom(
+    hasAnnouncementBeenSeenAtom
+  );
+  const [hasOnboardingBeenSeen] = useAtom(hasOnboardingBeenSeenAtom);
```
madsnedergaard, line 24: We should also support the `skip-onboarding` query parameter that some people use for embedding the map on dashboards. See past example here: https://github.com/electricitymaps/electricitymaps-contrib/pull/5840/files#diff-0f9eeb8600aad95a9ee44374749ce1d84315dd3e062985528e002e29a688fb21R62-R64

web/src/features/modals/AnnouncementModal.tsx
madsnedergaard: Not quite sure I understand why we need both the general check and the session check?

What am I missing with just the logic like the old modal did? :)
```js
const visible =
    Boolean(hasOnboardingBeenSeen) &&
    !hasTotalEnergyIntroBeenSeen &&
    !skipOnboarding &&
    !isExpired;
```

web/src/features/modals/AnnouncementModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    return null;
+  }
+
+  return (
+    <div
+      role="presentation"
+      onClick={handleDismiss}
+      onKeyDown={handleDismiss}
+      className="absolute left-0 top-0 z-50 flex h-full w-full flex-col items-center justify-center bg-black/20"
```
madsnedergaard, line 58: Why could we not reuse the Modal component, was the design too different?

web/src/locales/en.json
```diff
@@ -13,6 +13,10 @@
   "settings-modal": {
     "title": "Settings"
   },
+  "announcement-modal": {
+    "title": "Introducing a visual update to our app!",
+    "content": "We've revamped the colors representing electricity sources, bringing them closer to industry standards."
+  },
```
madsnedergaard, line 19: As a general FYI, I don't think it we need to do translations for short-lived messages like this one :)

web/src/features/modals/AnnouncementModal.tsx
madsnedergaard: I also think it's a bit of a shame to not use [Radix modal](https://www.radix-ui.com/primitives/docs/components/dialog) here which is much more accessible :)

madsnedergaard (commented):
Some comments :)

web/src/features/modals/AnnouncementModal.tsx
madsnedergaard: On second thought it's probably not too critical for this short-lived one, so I'm also okay with sticking to your current implementation for now if this becomes blocking :)

web/src/features/modals/AnnouncementModal.tsx
silkeholmebonnen: The old modal would show up after a user had just seen the onboarding modal, but we decided that a user that has just seen the onboarding modal probably wouldn't be interested in the announcement modal. 

So we only want to show the announcement modal if the user has not seen the onboarding modal in the current session, but it has been seen before in another session.

web/src/features/modals/AnnouncementModal.tsx
silkeholmebonnen: The design was too different to reuse it. I can try to use the Radix's, if you would like. What is the reason why it is smart to use?

web/src/features/modals/AnnouncementModal.tsx
madsnedergaard: Aaah I see, makes sense then :)

web/src/features/modals/AnnouncementModal.tsx
madsnedergaard: Okay, fair :) Here's why it's nice (in addition to being tested in a multitude of browsers and environments):

Screenshot 2024-06-04 at 23 37 46

With that said I'm also fine with not during it here, but next time it would be good to revamp a generic Modal component that we can more easily reuse :)

web/src/components/Link.tsx
```diff
@@ -0,0 +1,12 @@
+export function Link({ href, linkText }: { href: string; linkText: string }) {
```
madsnedergaard, line 1: I will fix the upcoming merge conflict here once https://github.com/electricitymaps/electricitymaps-contrib/pull/6786 is merged :)

madsnedergaard (approved):

PR #6800, opened 2024-06-03
feat(PWA + Service worker): Add PWA manifest and use service worker to handle updates

VIKTORVAV99:
## Issue

Closes: AVO-299
Closes: AVO-300
Closes: AVO-301
Closes: AVO-302
Closes: AVO-303
Closes: #5031 

## Description

Enables the build config for a Service Worker and PWA manifest to make the web app more app like with a controlled update behaviour that keeps all the files in sync.

It also helps support reoccurring users in low connectivity areas by caching assets such as code, fonts and images between visits. This goes beyond browser caching and are implemented in the following way:
- Precaching: All our code and fonts are preached which means they are always downloaded in full and available to the browser when needed, this ensures we should not see any more failed to import errors.
- Runtime caching: All images and icons are cached using runtime caching. This means that they are added to the cache the first time they are used instead of during the "install". This helps keep the assets and network connections to a minimum during the install as we support multiple versions of the images (gif vs webp or colourblind more for the arrows for example).

### Double check


VIKTORVAV99:
@madsnedergaard @tonypls I'm having a hard time figuring out why the cypress tests time out here, they seem to work just fine locally in both electron and edge. Any insights you might have would be very appreciated.

VIKTORVAV99:
@tonypls I applied your workaround, it's not the most elegant solution as you pointed out but it seems to do the trick. So this should be ready for a final review now 🙂

madsnedergaard:
Inspiration for applying on beforeEach
https://gist.github.com/jamesfulford/26181c332c25d6464c3524510d18e75e

web/src/components/Toast.tsx
```diff
@@ -7,15 +7,19 @@ type Props = {
...
 };
 
-function Toast(props: Props) {
+function Toast({ title, description, toastAction, toastActionText, toastClose }: Props) {
   const { t } = useTranslation();
-  const { title, description, toastAction, toastActionText } = props;
   const [open, setOpen] = useState(true);
   const handleToastAction = () => {
-    toastAction && toastAction();
+    toastAction?.();
```
madsnedergaard, line 18: Is this really valid? Looks so weird 😅

web/src/features/service-worker/UpdatePrompt.tsx
```diff
@@ -0,0 +1,44 @@
...
+    updateServiceWorker,
+  } = useRegisterSW({
+    onRegisteredSW: (_swURL, registration) => {
+      registration &&
+        setInterval(
+          () => {
+            console.info(`Checking for app update...`);
+            registration.update();
+          },
+          import.meta.env.PROD ? ONE_HOUR : 10 * 1000
```
madsnedergaard, line 19: Would it be a problem to check more often than once an hour? And what was the reason for doing the `prompt` approach vs `autoUpdate`?

web/vite.config.cts
```diff
@@ -41,6 +41,50 @@ const sentryPluginOptions: SentryVitePluginOptions = {
   },
 };
 
+const PWAManifest: Partial<ManifestOptions> = {
+  name: 'Electricity Maps',
+  short_name: 'Electricity Maps',
+  start_url: '/',
+  display: 'standalone', // For some reason, the type is not recognized without the cast
```
madsnedergaard, line 48: Is this comment still valid? I don't get the "cast" thing at least :)

web/src/components/Toast.tsx
VIKTORVAV99: haha yeah it's valid syntax 😅

We can reconsider it but then we would always need an action which we might not want if we are planning to use this toast for other things. Like notifying the users they are offline or something similar.

web/vite.config.cts
```diff
@@ -41,6 +41,50 @@ const sentryPluginOptions: SentryVitePluginOptions = {
   },
 };
 
+const PWAManifest: Partial<ManifestOptions> = {
```
madsnedergaard, line 44: Would be nice if we could add these to support better installs :)
Screenshot 2024-06-19 at 13 39 03

madsnedergaard (commented):
A few minor comments, but otherwise it looks good to me. I haven't actually tested the updating and how it works with a new version, let me know if you know a good way to do that?  😅

web/src/features/service-worker/UpdatePrompt.tsx
VIKTORVAV99: AutoUpdate is pretty ruthless and will update as soon as a new version is found and force reload the application, not the best UX but I suppose it's a "safer" option implementation wise.

web/vite.config.cts
VIKTORVAV99: Ah no, I fixed that in another way 😅

I'll remove it.

web/vite.config.cts
VIKTORVAV99: They are added as a separate milestone in linear, to keep the effort and scope of the initial PR as small as possible.

web/src/features/service-worker/UpdatePrompt.tsx
madsnedergaard: In the past we (somehow) had it fetch new assets in the background, but still prompt users to "reload" before it was used. But might be too much for now, we can always move that direction later on if we want :)

web/vite.config.cts
madsnedergaard: Great!

web/src/features/service-worker/UpdatePrompt.tsx
VIKTORVAV99: That might require us to build a more custom service worker, something we can do but I'm not sure it's worth it TBH. At least not yet.

madsnedergaard:
Not sure how to test this PRfurther, but it would be really nice to do a proper test with a new update on staging.electricitymaps.com - perhaps we can sync on that after this is merged?

VIKTORVAV99:
> Not sure how to test this PRfurther, but it would be really nice to do a proper test with a new update on staging.electricitymaps.com - perhaps we can sync on that after this is merged?

The easiest way to test it locally with the service worker in effect would be to do the following:

1. run `pnpm run build --mode=development --watch`
2. run `pnpm run preview` in another terminal
3. edit some code
4. wait 10-20 sec
5. see update prompt

Hopefully the above steps should trigger the service worker to update.

But I wouldn't mind deploying it to staging and testing it in a live environment either.

madsnedergaard:
Thanks, that worked :D 

I think we can make the prompt look nicer and make it more clear what the "primary" action is with some changes like this, I'll make a commit with the changes if that's okay with you :)

Screenshot 2024-06-20 at 11 02 23

VIKTORVAV99:
> Thanks, that worked :D
> 
> I think we can make the prompt look nicer and make it more clear what the "primary" action is with some changes like this, I'll make a commit with the changes if that's okay with you :)
> 
> Screenshot 2024-06-20 at 11 02 23

There is a separate PR for that https://github.com/electricitymaps/electricitymaps-contrib/pull/6822 😉

madsnedergaard:
> > Thanks, that worked :D
> > I think we can make the prompt look nicer and make it more clear what the "primary" action is with some changes like this, I'll make a commit with the changes if that's okay with you :)
> > Screenshot 2024-06-20 at 11 02 23
> 
> There is a separate PR for that #6822 😉

lol, it also seemed weird that you didn't do anything 🤦 
Let's get that PR merged first then?

VIKTORVAV99:
> > > Thanks, that worked :D
> > > I think we can make the prompt look nicer and make it more clear what the "primary" action is with some changes like this, I'll make a commit with the changes if that's okay with you :)
> > > Screenshot 2024-06-20 at 11 02 23
> > 
> > 
> > There is a separate PR for that #6822 😉
> 
> lol, it also seemed weird that you didn't do anything 🤦 Let's get that PR merged first then?

It's based of this branch to avoid conflicts but if you want to merge it into this PR first I'm fine with that 🙂

madsnedergaard (approved):
Looks good to me (once the updated design is merged in)

VIKTORVAV99:
Merging this now and creating a contrib PR and marking it as blocked so it's not accidentally released before we are comfortable with it in production/staging.

PR #6804, opened 2024-06-04
bug(web): update timeslider on hour change

tonypls:
## Issue
The timeslider was not updating with the changing hours

## Description
This PR adds the isLoading variable to the useMemo dependency array that sets the list of datetimes used in the timeslider.

P.S. removed random unused import


VIKTORVAV99 (approved):
Is there not a better or more robust way to do this? I'm just worried we might be introducing things that we later have to fix in order to be compliant with the rules of react?

However I think that might be out of scope for this pr in particular as we do this elsewhere as well so I'll approve this but I do think we should try and solve these issues once and for all!

PR #6807, opened 2024-06-05
bug(web): map overlays blocking touch events on mobile

tonypls:
## Issue

Empty map overlay component is blocking touch events on smaller devices

## Description

This PR hides the component on smaller devices

### DIFF
--- a/web/src/components/MapOverlays.tsx
+++ b/web/src/components/MapOverlays.tsx
@@ -12,7 +12,7 @@ export default function MapOverlays() {
   const surveyEnabled = useFeatureFlag('feedback-micro-survey') && !hasSeenSurveyCard;
 
   return (
-    <div className="fixed bottom-4 right-4 z-20 hidden flex-col items-end space-y-3 sm:flex">
+    <div className="pointer-events-none fixed bottom-4 right-4 z-20 hidden flex-col items-end space-y-3 sm:flex">
       <Suspense>
         <FeatureFlagsManager />
       </Suspense>
--- a/web/src/components/MapOverlays.tsx
+++ b/web/src/components/MapOverlays.tsx
@@ -12,7 +12,7 @@ export default function MapOverlays() {
   const surveyEnabled = useFeatureFlag('feedback-micro-survey') && !hasSeenSurveyCard;
 
   return (
-    <div className="fixed bottom-4 right-4 z-20 flex flex-col items-end space-y-3">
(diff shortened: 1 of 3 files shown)


VIKTORVAV99:
Not sure if this is just the micro surveys doing but it seems like this entire area is blocking all the map interaction (on both desktop and mobile) likely due to the fact that the micro survey is a lot wider than the others.

image

Also should the legend really move? Is that not confusing and causing undue layout shifts and renders?

web/src/components/MapOverlays.tsx
```diff
@@ -12,7 +12,7 @@ export default function MapOverlays() {
   const surveyEnabled = useFeatureFlag('feedback-micro-survey') && !hasSeenSurveyCard;
 
   return (
-    <div className="fixed bottom-4 right-4 z-20 flex flex-col items-end space-y-3">
+    <div className="fixed bottom-4 right-4 z-20 hidden flex-col items-end space-y-3 sm:flex">
       <Suspense>
         <FeatureFlagsManager />
       </Suspense>
```
VIKTORVAV99, line 18: I think we need to wrap each of these in their own div so they only take up the exact width and high each of the elements require.

tonypls:
> Not sure if this is just the micro surveys doing but it seems like this entire area is blocking all the map interaction (on both desktop and mobile) likely due to the fact that the micro survey is a lot wider than the others.
> 
> image
> Also should the legend really move? Is that not confusing and causing undue layout shifts and renders?

I updated the clicks to only propagate where required. I think it's a bit nicer for users that it renders below the legend as it's a common place for feedback cards. 

Let's focus on the bug here and we can think about UX improvements after if we like 👨‍🎨

web/src/components/MapOverlays.tsx
tonypls: I'm not sure this will help 🤔

VIKTORVAV99: Wouldn't wrapping each of them in their individual divs ensure they don't cover any more of the screen than they should? Right now it forms a big (transparent) box around the extreme points.

tonypls: I tried wrapping in divs when you suggested and it still grew to fill the larger container

VIKTORVAV99 (approved):
LGMT! 🎉

web/src/components/MapOverlays.tsx
VIKTORVAV99: Ah, that's unfortunate. Let's go with the current approach then.

PR #6808, opened 2024-06-06
Update Belgian Solar capacity

Fred-3000:
Same source as my previous update, also clarified the source (https://www.elia.be/en/grid-data/generation-data/solar-pv-power-generation-data).


config/zones/BE.yaml
```diff
@@ -45,9 +45,9 @@ capacity:
     - datetime: '2023-01-01'
       source: entsoe.eu
       value: 6475.0
-    - datetime: '2024-01-01'
-      source: entsoe.eu
-      value: 8789.0
+    - datetime: '2024-06-06'
+      source: elia.be
+      value: 9986.05
```
madsnedergaard, line 50: Keeping the old one:
```suggestion
    - datetime: '2024-01-01'
      source: entsoe.eu
      value: 8789.0
    - datetime: '2024-06-06'
      source: elia.be
      value: 9986.05
```

madsnedergaard (approved):
LGTM, would be nice to also get wind and other modes updated, but I couldn't find datasets on those (that somewhat matched ENTSO-E values) - so let's just merge this one as is :)

PR #6812, opened 2024-06-07
fix(parser-validation): Remove parser validation on US-SE-SEPA

unitrium:
## Issue

The backend validation is now active, so we should start relying less on parser validation which might discard a point. As a starter we disable it for US-SE-SEPA as right now we discard all datapoints because there's no solar but actually this zone should almost never have solar.


wobniarin (approved):
Merci!! :)

PR #6813, opened 2024-06-07
bug(web): close map control tooltip when menu button is  clicked

tonypls:
## Issue
Currently our map button tooltips persist after a selection menu is opened
## Description
This PR changes the pointer events for a map button depending on if it's a toggle button or a menu.

We want to tooltip to disappear when the menu is open


tonypls:
> This only partially fixes the problem by hiding the tooltip if the menu is open but the original tooltip is still open while the mouse hovers on the icon (but disappear when moved outside). image

I can't seem to replicate this 🤔

VIKTORVAV99:
> > This only partially fixes the problem by hiding the tooltip if the menu is open but the original tooltip is still open while the mouse hovers on the icon (but disappear when moved outside). image
> 
> I can't seem to replicate this 🤔

And neither can I anymore...
Very strange but I guess this works as expected then?
I'll approve and then if the other things shows back up we'll handle it then.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #6815, opened 2024-06-08
fix some misspells in french translation

phiphou:
Fix some misspells in french translation in the feedback part.


VIKTORVAV99 (approved):
Thanks!

PR #6818, opened 2024-06-10
Tvs/test cy2

tonypls:
### Double check

### DIFF
--- a/web/cypress/e2e/countrypanel.spec.ts
+++ b/web/cypress/e2e/countrypanel.spec.ts
@@ -1,9 +1,24 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 // TODO: Convert to component test
 describe('Country Panel', () => {
   beforeEach(() => {
     cy.interceptAPI('v8/state/hourly');
     cy.interceptAPI('v8/state/last_hour');
     cy.interceptAPI('v8/meta');
+    cy.visit('/map', {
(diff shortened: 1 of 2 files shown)


web/cypress/e2e/countrypanel.spec.ts
```diff
@@ -1,15 +1,45 @@
...
+  //   }
+  // });
 
   it('interacts with details', () => {
     cy.interceptAPI('v8/details/hourly/DK-DK2');
 
-    cy.visit('/zone/DK-DK2?lang=en-GB');
+    cy.visit('/zone/DK-DK2?lang=en-GB', {
+      onBeforeLoad(win) {
+        delete win.navigator.__proto__.serviceWorker;
```
tonypls, line 40: I think this is making it work here

VIKTORVAV99: I think this will prevent the service worker from registering, since we are running building the cypress build with `--mode=testing` can we leverage that to prevent it from building/mounting instead?

ESLint don't seem to be too happy about it but I guess that is fine in tests.

tonypls: Would that turn the service worker off for all tests?

As much as possible I think it would be good to test e2e with the service worker running and prevent it from registering only where it causes us issues. What do you think?

VIKTORVAV99: It would, and probably all prevent all the imported code from generating properly which is not optimal.

So let's try your fix and ignore it in the code scanning and hope that fixes things.

PR #6819, opened 2024-06-10
perf(loading spinner): Optimize loading spinner and inline code

VIKTORVAV99:
## Issue
We are doing a http call for a critical part of the core functionality, this is potentially causing performance problems.

## Description
Since we have moved many of the icons into JSX components I think it's time we did it for this one as well. I also took the opportunity to take advantage of JSX to minimise the size impact.

This increases the bundle size by ~1kb but removes 1-2 https calls depending on if the users switches the theme or not.

Closes: #6821 

### Double check


VIKTORVAV99:
oh god that grew more than it should have, which is very strange as it's working when I build it locally and run with preview...

Why can't things ever be simple...

tonypls:
Because it's quite a bit larger than it used to be you now see if under the onboarding modal when reloading:
image

VIKTORVAV99:
> Because it's quite a bit larger than it used to be you now see if under the onboarding modal when reloading: image

Yeah that was not intentional, for some reason it seems when I tested it locally I got a cached version so it all looked great...

Should be fixed now tho.

tonypls:
The previous one started animated as soon as it loads, this one takes some time to start the animation. Not blocking but could be a nice improvement

PR #6822, opened 2024-06-11
chore(update-prompt): Update the prompt to match our new visual design

VIKTORVAV99:
## Issue
The update prompt was a bit dated and didn't match our current visual design.
Closes: AVO-315

## Description

Reworks the modal to use less HTML elements and match our current style a bit better. It also have some new props to make it more extensible in the future.

### TODO:
- [x] Update button stories.
- [x] Create a story for the toast.
- [x] Update en text.
- [x] Add translations.

### Preview
image
image

image

### Double check


VIKTORVAV99:
Ah shit, I knew this would happen eventually, prettier wants one thing and ESLint another...

web/src/components/Toast.tsx
```diff
@@ -1,6 +1,9 @@
 import * as ToastPrimitive from '@radix-ui/react-toast';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import { FaXmark } from 'react-icons/fa6';
```
VIKTORVAV99, line 4: @tonypls I noticed you used `import { HiOutlineX } from 'react-icons/hi'`, was there some special reason for that?

@Alportan do you have the icon name which you prefer?
(I just want to prevent us from importing duplicate icons/code)

tonypls: Nope, we should try unify our x marks, I think this is our most commonly used one `import { HiXMark } from 'react-icons/hi2';`

VIKTORVAV99: Great!

I'll switch all the x icons to use that :)

web/src/components/Button.tsx
```diff
@@ -22,10 +25,11 @@ export function Button({
   foregroundClasses, // textColor, etc.
   size = 'lg',
   type = 'primary',
+  asDiv, // If true, renders a div instead of a button to avoid nested buttons in components like ToastPrimitive.Action
```
VIKTORVAV99, line 28: Not super happy this was needed but it seemed to be the only way to avoid nested buttons while still re-using our buttons. Happy to hear about other ideas to solve it if you guys have them though.

web/src/locales/en.json
```diff
@@ -324,10 +324,15 @@
...
+    "update": "Update",
+    "dismiss": "Dismiss"
+  },
   "misc": {
     "maintitle": "Live 24/7 CO₂ emissions of electricity consumption",
     "oops": "Oops! We are having trouble reaching the server. We will try again in a few seconds.",
-    "newversion": "A new app version is available!",
     "webgl-not-supported": "The map can't be rendered because this browser does not support WebGL.",
     "retrynow": "Retry now",
     "database-ad": "Looking for historical data? <a href=\"https://electricitymaps.com?utm_source=app.electricitymaps.com&utm_medium=referral\" target=\"_blank\">Check out our Database!</a>",
```
VIKTORVAV99, line 338: Just curious, is this even used anywhere...?

web/src/components/Button.tsx
madsnedergaard: An alternative would be to use the Radix Button component, which has a nice pattern for solving it:

Screenshot 2024-06-20 at 11 15 39

https://www.radix-ui.com/themes/docs/components/button

But that does require some more work though, dunno if we want to do that now :)

VIKTORVAV99: I tried that first and it messed up all our styling and there was seemingly no easy way to fix it unfortunately. Looked even worse than before at one point...

web/src/components/Toast.tsx
```diff
@@ -28,42 +41,37 @@ function Toast({ title, description, toastAction, toastActionText, toastClose }:
...
+          {toastAction && (
+            <ToastPrimitive.Action
+              altText="view now"
+              className="mx-2 flex items-center justify-center"
+              onClick={handleToastAction}
+            >
+              <Button size="md" asDiv>
+                {toastActionText}
+              </Button>
+            </ToastPrimitive.Action>
```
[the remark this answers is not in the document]
VIKTORVAV99: This does work but it strips most if not all styles, asDiv was the easiest way to preserve our own styling with minimal changes. And radix has no primitives for buttons so no way to integrate them nicely as far as I can tell?

web/src/components/Button.tsx
madsnedergaard: Oh okay, thanks for explaining. Let's roll with your solution then. In the future we could make it a bit more generic with a `as` prop that allows rendering it as whatever, but for now I think this is okay :)

web/src/components/Toast.tsx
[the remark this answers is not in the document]
madsnedergaard: Shame 🥲

web/src/components/Toast.tsx
```diff
@@ -28,42 +41,37 @@ function Toast({ title, description, toastAction, toastActionText, toastClose }:
       <ToastPrimitive.Root
         open={open}
         onOpenChange={setOpen}
-        className="radix-state-open:animate-toast-slide-in-right radix-swipe-end:animate-toast-swipe-out radix-state-closed:animate-toast-hide fixed left-1/2 top-16 z-50 w-1/4 self-center rounded-lg  bg-white shadow translate-x-radix-toast-swipe-move-x radix-swipe-cancel:translate-x-0 radix-swipe-cancel:duration-200 radix-swipe-cancel:ease-[ease] dark:bg-gray-900"
+        duration={duration}
+        type="background"
+        className="bg-gray/80 fixed top-16 z-50 m-2 inline-block max-w-max self-center rounded-lg shadow backdrop-blur-sm xs:left-1/2 xs:w-max xs:-translate-x-1/2 xs:transform dark:bg-gray-800/80"
```
madsnedergaard, line 46:
```suggestion
        className="bg-white/80 fixed top-16 z-50 m-2 inline-block max-w-max self-center rounded-lg shadow backdrop-blur-sm xs:left-1/2 xs:w-max xs:-translate-x-1/2 xs:transform dark:bg-gray-800/80"
```
Looks better white white background in my opinion and it matches our map buttons :)

**Current**
Screenshot 2024-06-20 at 11 40 43

**Proposed**
Screenshot 2024-06-20 at 11 42 41

VIKTORVAV99: I thought I had gotten it right the first time but thanks! And I agree, much better!

madsnedergaard (approved):
It looks great and it works 🥳

web/src/components/Toast.tsx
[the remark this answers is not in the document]
VIKTORVAV99: Looks like HeadlessUI (made by the same folk as TailwindCSS) has a "primitive" for a Button we could potentially use that supports rendering as multiple types.

https://headlessui.com/react/button

Overall headless also looks pretty sweet and easy to use.

PR #6824, opened 2024-06-12
fix(config): Update expected production parser delays for some zones

florianscheidl:
## Issue

Our production parsers have varying delays across zones but this is not fully captured in the config files at the moment.

## Description

Step-by-step (according to parser availability monitoring), we will update the expected delays on demand. @VIKTORVAV99, for context, the update is computed as outlined in https://github.com/electricitymaps/electricitymaps/pull/5200

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #6826, opened 2024-06-13
fix(GB->NO-NO2): change GB->NO-NO2 to use ENTSO-E

VIKTORVAV99:
## Issue

Seems like the flows are reversed right now.
## Description

Switches parsers to ENTSO-E to fix the above issue and get more data while we look into the core issue in the other parser.

### Double check


florianscheidl (approved):
LGTM 👍

PR #6827, opened 2024-06-13
fix(config): Update expected production parser delays for US zones

florianscheidl:
## Issue

Our production parsers have varying delays across zones but this is not fully captured in the config files at the moment.

## Description
This is the second step of updating expected delays. For context, the update is computed as outlined in https://github.com/electricitymaps/electricitymaps/pull/5200.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #6829, opened 2024-06-14
chore(translations): Clean up unused translations

VIKTORVAV99:
## Issue
We have some legacy translations that are not used anymore but still included in the language bundles which are bloating things.

Closes: AVO-322

## Description

This PR adds a script to remove any key that are not in the `en.json` file from any of the other language files.

It removes some translations from `en.json` that are no longer used.

Runs the script.

### Double check


web/src/locales/sv.json
```diff
@@ -134,7 +132,6 @@
     },
     "estimated_time_slicer_average": {
       "title": "Data är för närvarande uppskattad",
-      "pill": "Otillgänglig",
```
VIKTORVAV99, line 137: I wonder if we should just add this key to the en file instead?

web/src/locales/pt-PT.json
```diff
@@ -980,8 +964,7 @@
       "zoneName": "Índia Nordeste"
     },
     "IN-NO": {
-      "countryName": "Índia",
-      "+oneName": "Índia Norte"
```
VIKTORVAV99, line 984: I'll clean up this one.

web/src/locales/pt-PT.json
```diff
@@ -132,7 +132,6 @@
     },
     "estimated_time_slicer_average": {
       "title": "Os dados atuais são estimados",
-      "pill": "Indisponível",
```
VIKTORVAV99, line 135: Same here as for the Swedish one

web/src/locales/sv.json
tonypls: I think it's okay to remove for now. We should re-add this with translations from "Preliminary" in the future, English also needs this for the badges in AVO-282. I'll add some context there

VIKTORVAV99:
> LGTM, it's a little confusing now we have two en.json, nice work all the same!

Yeah and I think we should consider moving it directly to the API project but we can discuss that on a later point if we need to actually move it.

PR #6830, opened 2024-06-14
Turn exchange arrows gray for zones with an outage

madsnedergaard:
## Issue

It feels weird seeing colorful arrows with a carbon-intensity value, when the zone is grayed out due to an outage.

## Description

This turns the export arrows of a zone with an ongoing outage gray.

We do however still want to include the estimated exchange in the zone details, as the estimated values are used in the flowtracing algorithms and as such we want to make sure they are available.

### Preview
Screenshot 2024-06-14 at 15 34 07

Screenshot 2024-06-14 at 15 34 18

### DIFF
--- a/web/src/hooks/arrows.ts
+++ b/web/src/hooks/arrows.ts
@@ -11,6 +11,10 @@ import exchangesToExclude from '../../config/excluded_aggregated_exchanges.json'
 const exchangesConfig: Record<string, any> = exchangesConfigJSON;
 const { exchangesToExcludeZoneView, exchangesToExcludeCountryView } = exchangesToExclude;
 
+/**
+ * Determines if the carbon intensity of an exchange should be hidden due to a temporary zone outage.
+ * By not passing on carbon intensity, we are able to still show a grey arrow with the power value.
+ */
 function shouldHideExchangeIntensity(
   exchange: string,
   zonesWithOutages: string[],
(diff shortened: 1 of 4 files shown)


web/src/locales/en.json
```diff
@@ -291,6 +291,7 @@
     "changeTheme": "Change theme",
     "noParserInfo": "No data available",
     "temporaryDataOutage": "Live data temporarily unavailable",
+    "temporarilyUnavailable": "Temporarily unavailable",
```
VIKTORVAV99, line 294: I'm thinking that maybe we should say we are using fallbacks instead?
That way it makes sense that the carbon intensity is still visible in the details of the importing country?

(Just a thought on my side though, but it might be even more confusing).

web/src/hooks/arrows.ts
```diff
@@ -11,6 +11,19 @@ import exchangesToExclude from '../../config/excluded_aggregated_exchanges.json'
...
+  value: number
+) {
+  const [zone1, zone2] = exchange.split('->');
+
+  return (
+    (zonesWithOutages.includes(zone1) && value > 0) ||
+    (zonesWithOutages.includes(zone2) && value < 0)
+  );
+}
+
```
VIKTORVAV99, line 26: I had a little hard time understanding what this was supposed to do at first glance. Could you maybe add a comment and or a test for our future selfs?

madsnedergaard: Thanks for the feedback, added some comments and tests now :)

web/src/locales/en.json
madsnedergaard: I see your point, but not sure how we could phrase it in a very short, yet easily understandable way :sweat_smile:

VIKTORVAV99 (approved):
LGTM! 🎉

It will be nice to finally use those grey arrows! 😅

PR #6834, opened 2024-06-17
chore(deps): Update to Vite 5

VIKTORVAV99:
## Issue

Previously this was blocked by Cypress but since they added support we should upgrade!

## Description

Upgrades to Vite 5 and changes the file extension on the config file to just `.ts`

There are slight changes in bundle sizes, both increases and decreases so I don't think that should be a concern for us.

### TODO:
- [ ] Figure out why all the arrows in the cypress component test is throwing 404 statuses effectively making them hidden.

### Double check


VIKTORVAV99:
TODO: Figure out why all the arrows in the cypress component test is throwing 404 statuses effectively making them hidden.

PR #6835, opened 2024-06-17
feat(bun): Use bun for development

VIKTORVAV99:
### Double check


VIKTORVAV99:
@madsnedergaard Bun works fine for local development but it seems pnpm still requires node for some reason. Now we can fix this in one of two ways, having both bun and node installed in earthly, which seems a bit cumbersome. Or just use node + pnpm in earthly.

Another alternative would be to switch package manager too and dump pnpm entirely but dependabot and the github dependency graph don't support the bun lock file (yet) and we would probably have to switch to https://www.mend.io/renovate/ if we wanted automatic dependency updates? I would be okay with manual updates as well FYI as we still require that for the major versions.

PR #6856, opened 2024-06-17
feat(config): update zones .yaml + zone pydantic model to include region and country

Garsdal:
## Issue

There is currently no clear source-of-truth for regions and countries of zones. 

## Description

* Included a `region` and `country` attribute in the pydantic config model
* Include `region` and `country` keys across all .yaml zone configurations

## Notes on data sources
All regions and countries are following the ISO-3166 standard with values taken from here: https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.csv 

We want to use these new attributes for aggregation and grouping of our zones. It was also considered if this could be done using existing information, and I checked the subZoneNames. We do have the field `sub_zone_names: list[ZoneKey] | None = Field(None, alias="subZoneNames")` in our zone model, which indicates a `parent_zone` and `sub_zone` relation. However, this is not populated exhaustively across all configurations to be used instead of a `country` attribute.

### Preview
An example of a US zone including `country` and `region`
image

### Double check


config/zones/AD.yaml
```diff
@@ -1,14 +1,16 @@
...
   hydro: 46
   solar: 0.231
 contributors:
-  - Manu1400
-  - autipial
+- Manu1400
+- autipial
 timezone: Europe/Andorra
+region: Europe
+country: AD
```
VIKTORVAV99, line 16: Do we really need the country? It should always be the 2 first chars of the zoneKey.

I'm not against adding it I'm just wondering why you might need this?

VIKTORVAV99: I forgot about those! 😅 Let's add the key then. I'm unsure about the other naming though but that we can look at later if need be.

Garsdal:
> I ran prettier to ensure the formatting stays the same and it passes the CI and all the changes look good now.
> 
> We would like to avoid merging more PRs until the latest frontend changes has been deployed tomorrow so I'll approve it then

Garsdal:
> I ran prettier to ensure the formatting stays the same and it passes the CI and all the changes look good now.
> 
> We would like to avoid merging more PRs until the latest frontend changes has been deployed tomorrow so I'll approve it then 🙂

I will make a new commit today to try and fix the indentation :) I am not sure if it is breaking, but I still dislike that this PR changes 40.000 lines of code. And regarding the merge, I am in no hurry, but thanks for checking it so fast!

VIKTORVAV99:
> > I ran prettier to ensure the formatting stays the same and it passes the CI and all the changes look good now.
> 
> > 
> 
> > We would like to avoid merging more PRs until the latest frontend changes has been deployed tomorrow so I'll approve it then 🙂
> 
> 
> 
> I will make a new commit today to try and fix the indentation :) I am not sure if it is breaking, but I still dislike that this PR changes 40.000 lines of code. And regarding the merge, I am in no hurry, but thanks for checking it so fast!

Indentation should already be fixed 😉

config/zones/CA-BC.yaml
```diff
@@ -15,22 +15,22 @@ emissionFactors:
...
 isLowCarbon:
   unknown:
     - datetime: '2019-01-01'
-      source: 'assumes 87% hydro, 5% biomass, 4% gas, 2% wind and 2% from various sources'
+      source: assumes 87% hydro, 5% biomass, 4% gas, 2% wind and 2% from various sources
       value: 0.94
 isRenewable:
   unknown:
     - datetime: '2019-01-01'
-      source: 'assumes 87% hydro, 5% biomass, 4% gas, 2% wind and 2% from various sources'
```
Garsdal, line 33: Are these changes okay? Is it due to the formatting? @VIKTORVAV99

VIKTORVAV99: Perfectly fine, as long as it passes prettier it's good to go.

VIKTORVAV99 (approved):
LGMT! 🎉

config/zones/NA.yaml
```diff
@@ -77,3 +77,5 @@ fallbackZoneMixes:
         unknown: 0.0
         wind: 0.015505325244864603
 timezone: Africa/Windhoek
+region: Africa
+country: null
```
tonypls, line 81: Nulll?

Garsdal: I decided to use `null` instead of `Other` for zones with an ambiguous country. It might potentially be a sensitive topic, so I thought it was the most appropriate to leave the value as null in these cases. Please bring other suggestions if you have any!

tonypls (commented):
Just the one small comment but otherwise LGTM

config/zones/NA.yaml
tonypls: I think that's a nice approach but this is Namibia and as far as I know it's not contested?

config/zones/GL.yaml
```diff
@@ -1 +1,3 @@
 timezone: America/Nuuk
+region: Americas
+country: GL
```
madsnedergaard, line 3: As a fellow dane I am very disappointed in you on this one 😉 🇬🇱

Garsdal: I used the ISO-3166-1 (https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.csv) standard to generate the configs, but it hurts my danish heart 😭

PR #6857, opened 2024-06-17
perf(manual-chunking): Remove `country-flag-icons` from manual chunking

VIKTORVAV99:
## Description

Since this is now only imported into the Flag component we no longer need to add it as a manual chunk, this improve performance by budding it directly where it's used (together with the Flag component) and removes a lot of import links that was included in half the bundles for some reason.

Should yield a very minor performance boost when it comes to http downloads and first visits to the app.

### Double check

### DIFF
--- a/web/vite.config.cts
+++ b/web/vite.config.cts
@@ -11,7 +11,6 @@ import tsconfigPaths from 'vite-tsconfig-paths';
 const manualChunkMap = {
   '@sentry': 'sentry',
   '@radix-ui': 'radix',
-  'country-flag-icons': 'flags',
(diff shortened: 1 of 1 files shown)


silkeholmebonnen (commented):
Looks good. Can we somehow test if the performance actually is increased?

VIKTORVAV99:
> Looks good. Can we somehow test if the performance actually is increased?

The performance increase itself is so small it's not really measurable with all other variations that exists but I manually verified the size changes and the import links and they look much nicer now.

silkeholmebonnen (approved):
Looks good ⭐

PR #6862, opened 2024-06-18
chore(dependabot): Adjust dependabot config

VIKTORVAV99:
## Description

Let's ensure the patch versions are updated as well.

And let's ensure we keep react up to date.

### Double check


silkeholmebonnen (approved):
LGTM! Do you know why we ignored the patch versions before?

VIKTORVAV99:
> LGTM! Do you know why we ignored the patch versions before?

It was before they had grouping of dependencies and it became very spammy, but should be fine now with the groups as they should ensure there don't become too many PRs.

PR #6865, opened 2024-06-18
fix(exchange tooltip): Remove hidden class to fix exchange tooltip on mobile

VIKTORVAV99:
## Issue

Fixes: AVO-320

### Double check


madsnedergaard (approved):
Nice, I was also curious about this when working on gray-arrows, but then thought it was on purpose that it didn't appear on mobile 😅

VIKTORVAV99:
> Nice, I was also curious about this when working on gray-arrows, but then thought it was on purpose that it didn't appear on mobile 😅 

Nope, it just slipped by in a review 😬

Which probably means we should expand the cypress tests but they are already so flaky on the CI...

PR #6866, opened 2024-06-18
Hide zones with insufficient data from the map

madsnedergaard:
## Issue

We have some zones that are always showing the same carbon intensity for every single hour of the year. This means we are showing potentially quite wrong data (Thailand as example has an average of 550g co2/kWh for 2023 according to [Our World in Data](https://ourworldindata.org/grapher/carbon-intensity-electricity?tab=chart&time=2017..latest&country=~THA) vs our 409g CO2/kWh).

## Description

This PR hides certain zones from the map to avoid showing incorrect colors and numbers on the map.
I decided to reuse the `aggregates_displayed` property in zone configurations to avoid introducing more new ways of doing things. To make changes minimal I added a "none" entry for the zones, but an alternative implementation (which requires a bit more work) would be to return an empty array instead (`aggregates_displayed: []`).

**Zones hidden:**
- All subzones in Mexico
- Thailand
- Kuwait
- Bahrain

### Preview

Screenshot 2024-06-18 at 22 48 17

### Double check

### DIFF
--- a/web/src/features/panels/zone/util.ts
+++ b/web/src/features/panels/zone/util.ts
@@ -20,6 +20,7 @@ export const getHasSubZones = (zoneId?: string) => {
 
 export enum ZoneDataStatus {
   AGGREGATE_DISABLED = 'aggregate_disabled',
+  FULLY_DISABLED = 'fully_disabled',
   NO_INFORMATION = 'no_information',
   NO_REAL_TIME_DATA = 'dark',
   AVAILABLE = 'available',
@@ -44,8 +45,6 @@ export const getZoneDataStatus = (
   // If there is no config for the zone, we assume we do not have any data
   const zoneConfig = config.zones[zoneId];
   if (!zoneConfig) {
-    console.log(zoneConfig);
-
     return ZoneDataStatus.NO_INFORMATION;
   }
 
@@ -53,6 +52,9 @@ export const getZoneDataStatus = (
     config.zones[zoneId].aggregates_displayed &&
     !config.zones[zoneId].aggregates_displayed.includes(timeAverage)
   ) {
+    if (config.zones[zoneId].aggregates_displayed[0] === 'none') {
+      return ZoneDataStatus.FULLY_DISABLED;
+    }
     return ZoneDataStatus.AGGREGATE_DISABLED;
   }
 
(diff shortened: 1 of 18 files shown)


web/src/locales/en.json
```diff
@@ -88,6 +88,7 @@
     "helpUs": "Help us identify the problem by taking a look at the <a href=\"{{link}}\" target=\"_blank\">runtime logs</a> or contact the data provider.",
     "noParserInfo": "We do not yet have any information about this area.<br/> Want to help fix this? You can contribute by <a href=\"{{link}}\" target=\"_blank\">adding area data</a>.",
     "aggregationDisabled": "Data is not available for the selected time period due to missing granular breakdown. Try a broader time period.",
+    "fullyDisabled": "The data we have available for this zone is insufficient and therefore not shown. If you are aware of data sources for this area, please <a href=\"{{link}}\" target=\"_blank\">let us know</a>.",
```
VIKTORVAV99, line 91: Just a small nitpick here, it wound be nice to use single quotes in the links so we can avoid the escape chars. But I see it's the same in all the other ones I guess it's fine.

web/src/features/panels/zone/util.ts
```diff
@@ -44,15 +45,16 @@ export const getZoneDataStatus = (
   // If there is no config for the zone, we assume we do not have any data
   const zoneConfig = config.zones[zoneId];
   if (!zoneConfig) {
-    console.log(zoneConfig);
-
```
VIKTORVAV99, line 48: you can potentially hide this in a
`if (import.meta.env.DEV){}` clause instead if you want it for debug builds.

VIKTORVAV99 (commented):
Does this require the backend to be updated/use this data as well to function or should it work in just the app?

It doesn't seem to work in the preview links.

web/src/locales/en.json
madsnedergaard: Yeah, I just copy-pasted - but happy to make an issue for generally fixing this problem afterwards :)

web/src/features/panels/zone/util.ts
madsnedergaard: It was an old left-over console log that seemed like a mistake, so I just removed it :)

madsnedergaard:
> Does this require the backend to be updated/use this data as well to function or should it work in just the app?

Backend already works with no changes with this approach :)

web/src/locales/en.json
VIKTORVAV99: Sounds good, something to do next slack week!

VIKTORVAV99 (approved):
LGMT! 🎉

web/src/locales/en.json
madsnedergaard: Captured in Linear here: https://linear.app/electricitymaps/issue/AVO-336/remove-escaped-quotes-from-translation-files

PR #6885, opened 2024-06-21
hackday: Virtual lists for ranking panel

VIKTORVAV99:
## Issue

We are rendering a lot of DOM elements with this list when we don't really have to.
Closes: AVO-335

## Description

Adds `@tanstack/react-virtual` to use a virtual list instead of rendering everything, drastically cutting down the DOM elements.

The amount of elements on `/map` goes from `3 915 elements` to `692 elements` with this change (200 zones with data). 

### Double check


tonypls (approved):
LGTM 🎉

PR #6886, opened 2024-06-21
feat(hack): in app storage buster

tonypls:
## Issue
The dev process on feature flagged stuff can be quite slow to delete local storage keys

## Description
This PR adds a new button to the featureflags manager (find with the ff query param, `app.el...com/map?ff`)

The button clears all local storage and reloads the page.

### DIFF
--- a/web/src/features/feature-flags/FeatureFlagsManager.tsx
+++ b/web/src/features/feature-flags/FeatureFlagsManager.tsx
@@ -7,6 +7,11 @@ import { useSearchParams } from 'react-router-dom';
 import { useFeatureFlags } from './api';
 import { FeatureFlags } from './types';
 
+function handleClearLocalStorage() {
+  localStorage.clear();
+  location.reload();
+}
+
 function Content({ features }: { features: FeatureFlags }) {
   const queryClient = useQueryClient();
   const mutation = useMutation({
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
This is not related to your changes directly but how do we feel about limiting these dev features to only dev environments and stripping it from production builds?

tonypls:
> This is not related to your changes directly but how do we feel about limiting these dev features to only dev environments and stripping it from production builds?

It can be nice for us to be able to test this stuff on production if we are debugging. 

Are you thinking to limit users or for reducing bundle size?

VIKTORVAV99:
> > This is not related to your changes directly but how do we feel about limiting these dev features to only dev environments and stripping it from production builds?
> 
> 
> 
> It can be nice for us to be able to test this stuff on production if we are debugging. 
> 
> 
> 
> Are you thinking to limit users or for reducing bundle size?

Mostly reducing the bundle size, but it's only a 5 kB change so far so it's fine for now, but if we add more dev tools it might be nice to limit them.

tonypls:
> > > This is not related to your changes directly but how do we feel about limiting these dev features to only dev environments and stripping it from production builds?
> > 
> > 
> > It can be nice for us to be able to test this stuff on production if we are debugging.
> > Are you thinking to limit users or for reducing bundle size?
> 
> Mostly reducing the bundle size, but it's only a 5 kB change so far so it's fine for now, but if we add more dev tools it might be nice to limit them.

Is there a way we could only download them if the query param is active?

web/src/features/feature-flags/FeatureFlagsManager.tsx
```diff
@@ -34,23 +39,31 @@ function Content({ features }: { features: FeatureFlags }) {
...
+            </Switch.Root>
+          </div>
+        ))}
+      </div>
+      <button
+        className="mt-2 self-end rounded  bg-green-900 p-1 text-xs text-white"
+        onClick={handleClearLocalStorage}
+      >
+        Clear Local Storage
+      </button>
```
silkeholmebonnen, line 66: Can we maybe increase the text size a bit? It is very small

silkeholmebonnen (commented):
Nice work! I'm sure I will be using this feature 🙌

web/src/features/feature-flags/FeatureFlagsManager.tsx
tonypls: 🙌 bumped it up!

silkeholmebonnen (approved):
LGTM 🥳

PR #6887, opened 2024-06-21
Remove US-CENT-SPA from fossil production requirement

florianscheidl:
## Issue

US-CENT-SPA is a zone which only has hydro production. However, so far, we have required that the parser reports one of the fossil production technologies, leading to the parser to fail recently (previously they had reported unknown: 0).

## Description
In the PR, we add the zone to the list of fossil-free production.

### Double check


wobniarin (approved):
LGTM 🚀

PR #6888, opened 2024-06-22
chore(storybook): Use storybook features for button stories

VIKTORVAV99:
## Issue
The way we did things before don't really allow you to change any of the props and see any changes.

Closes: AVO-343
## Description
Turn examples into individual stories and override some of the default stuff to provide a better DX.

For example we are creating a radio button to toggle the icon on or off.

>[!NOTE]
>We can always add more examples if we need them later.

### Preview
image

### Double check


web/src/components/Button.tsx
VIKTORVAV99, line 1: These changes are just so we don't have to declare the custom class every time we use the button and want it to shrink and it allows storybook to easily change it as it's a boolean.

silkeholmebonnen (approved):
Nice work! Looks way better now 🤩

PR #6889, opened 2024-06-22
chore(types): more specific badge types

VIKTORVAV99:
## Issue

It seems like we where declaring a type we where not expected but this did not fail type checking because we where not specific enough.

## Description

Adds the more specific type and use that where applicable.

### Double check


web/src/components/Badge.tsx
```diff
@@ -1,6 +1,8 @@
+export type PillTypes = 'default' | 'warning';
```
silkeholmebonnen, line 1: Small thing but would it be more correct to call it `PillType`?

silkeholmebonnen (approved):
Looks good to me! Just a small nitpick

web/src/components/Badge.tsx
VIKTORVAV99: I'm actually not sure what the convention is but I was thinking that the property `pillType` should be a type of `PillTypes` if that makes sense?

I'd be happy to change it if you wish though.

silkeholmebonnen: I did a little google search and it seems that the convention is singular, so `PillType` ([Types in typescript](https://imgeorgemor.medium.com/you-dont-know-how-to-write-types-in-typescript-d8cfaf4857e5))

VIKTORVAV99: Great!

I'll change it then!

PR #6890, opened 2024-06-22
perf: Move CSS import to `main.tsx`

VIKTORVAV99:
## Issue

Since we didn't import the CSS in the same place as the other CSS files it created a separate bundle at build time. And since CSS is render blocking both where needed for the map to render.

## Description

Moves the CSS import to our `main.tsx` where our other CSS imports are so they end up in the same bundle.

### Preview
On master:
```
dist/assets/MapWrapper-05820bd7.css               63.75 kB │ gzip:   9.07 kB
dist/assets/index-cfaab156.css                    76.15 kB │ gzip:  13.26 kB
```
On this branch:
```
dist/assets/index-4f5ad7ed.css                   139.90 kB │ gzip:  22.25 kB
```

### Double check

### DIFF
--- a/web/src/features/map/Map.tsx
+++ b/web/src/features/map/Map.tsx
@@ -1,5 +1,3 @@
-import 'maplibre-gl/dist/maplibre-gl.css';
-
 import useGetState from 'api/getState';
 import ExchangeLayer from 'features/exchanges/ExchangeLayer';
 import ZoomControls from 'features/map-controls/ZoomControls';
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
If we want to save a few kB on the css bundle we can also switch to https://lightningcss.dev as the minifier (it's as simple as adding the dependency and changing a single config option).

tonypls (approved):
LGTM

PR #6892, opened 2024-06-22
chore: use modern tailwind syntax

VIKTORVAV99:
## Issue

These will be removed in a tailwindcss v4.

## Description

Since this will be removed we might as well start using them already.

### Double check


silkeholmebonnen (approved):
LGTM 🟢

PR #6893, opened 2024-06-23
chore: switch storybook to ts

VIKTORVAV99:
## Issue

The storybook config still uses js and old cjs syntax.

## Description

Switches the storybook config to use typescript.

### Double check


web/tsconfig.node.json
```diff
@@ -17,5 +17,5 @@
     },
     "transpileOnly": true
   },
-  "include": ["vite.config.cts", "cypress.config.ts", "scripts", "geo"]
+  "include": ["vite.config.cts", "cypress.config.ts", "scripts", "geo", ".storybook"]
```
silkeholmebonnen, line 20: What is the reason for adding `.storybook` here?

silkeholmebonnen (commented):
Looks good! I just have one question:)

web/tsconfig.node.json
VIKTORVAV99: So it's picked up by `tsc` (the typescript compiler) and actually gets type checked (with the expected rules). So far it looks like it's just config files so it should be in the "node" config instead of the general app config.

silkeholmebonnen: Why should it be inside the node config if it is only config files?

VIKTORVAV99: > Why should it be inside the node config if it is only config files?

We have a different set of rules for that compared to the web app and in the future once we have gotten rid of TSNode I think we want to change the module resolution of the web app.

@madsnedergaard has some more insight on why it was originally split.

silkeholmebonnen: Okay, I don't think I really understand what is supposed to be in what tsconfig file and why, so I don't really feel comfortable approving it. I don't know if it is even important but maybe someone else should review it?

VIKTORVAV99: That's perfectly fine!


I'll ask Mads to take a look as I think it was he that set it all up. 😅

madsnedergaard: That's very fair! The reason it's split is that they have different environments - the main app has [the DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction) available, while the other doesn't :)

madsnedergaard (approved):
LGTM

PR #6894, opened 2024-06-23
perf: use optional chaining everywhere

VIKTORVAV99:
## Issue

We are not using the most efficient syntax.

## Description

Use optional chaining everywhere I found that it could be used + simplify development check in analytics.

Needs https://github.com/electricitymaps/electricitymaps-contrib/pull/6891 to take full advantage of the perf changes but functionality wise works just fine without it.

### Double check


silkeholmebonnen (approved):
LGTM 🏆

PR #6897, opened 2024-06-23
chore(deps-dev): remove unused dependency

VIKTORVAV99:
## Issue

We have a dependency we don't use anywhere.

## Description

Removes the unused dependency.

### Double check


VIKTORVAV99:
Turns out it is used in earthly under another name...

PR #6899, opened 2024-06-24
feat(web): dynamic import of featureflag component

tonypls:
## Issue
Dev tools are served to all clients

## Description
This PR dynamically imports the FeatureFlagManager component depending on if the query param is present.


VIKTORVAV99 (approved):
Works great!

Thanks for doing this 💚

The service worker does appear to download it in the background but only after all the other code is donwloaded so we should still get a lot of benefits from when it comes to first visits and not bloating the runtime code if the parameter is not set.

PR #6903, opened 2024-06-26
Update production parser delay thresholds

florianscheidl:
## Issue

For our alerting, we use zone-wise delay thresholds for their respective production parsers. However, for several zones, we have not computed this and therefore it does not appear in the zone config yet.

## Description

In this PR, we add computed production parser delay thresholds for all zones.

### Double check

### DIFF
--- a/config/zones/AR.yaml
+++ b/config/zones/AR.yaml
@@ -44,6 +44,9 @@ contributors:
   - systemcatch
   - corradio
   - nessie2013
+country: AR
+delays:
+  production: 5
 disclaimer: Unknown production is the aggregation of thermal production, mainly gas.
 emissionFactors:
   direct:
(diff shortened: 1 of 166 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

Just curious how this is calculated though, like why is Norway at 7 hours delay when the expected delay from ENTSO-E is 2 hours?

florianscheidl:
It's something we are fine-tuning at the moment. I'm using a 0.75-quantile of measured delays and adding a 5-hour offset to ensure we don't get alerted right away for zones that once have a slightly greater-than usual delay.

VIKTORVAV99:
> It's something we are fine-tuning at the moment. I'm using a 0.75-quantile of measured delays and adding a 5-hour offset to ensure we don't get alerted right away for zones that have a slightly greater-than usual delay.

Sounds great!

However I'm wondering if it doesn't make more sense to just keep the offset in the function that alerts us rather than inside the config themselves? But that's just a minor implementation detail really.

florianscheidl:
Good point, we might want to move towards that in the future 👍

PR #6904, opened 2024-06-26
storybook: global language selector

VIKTORVAV99:
## Issue

Many of our components are directly using translations but we have no way to control the language in storybook right now.

Closes: AVO-347

## Description

Adds a decorator and global controls to the main storybook preview file as instructed here: https://storybook.js.org/recipes/react-i18next

>[!NOTE]
>There is an add-on that does this as well but we'll actually end up with *more* code if we go down that route? So I opted to just do the simple solution.

### Double check


madsnedergaard:
I like the simple approach!
Doesn't seem to work for me though 🤔 
https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/7ea6974f-dec7-4e16-add2-5d0bc967dc6b

VIKTORVAV99:
> I like the simple approach!
> 
> Doesn't seem to work for me though 🤔 
> 
> https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/7ea6974f-dec7-4e16-add2-5d0bc967dc6b
> 
> 
> 
> 

Did you make sure it's not a component that has its own i18next wrapper?

I'm going to remove those in a follow up PR.

I think you can test it best on the FAQ panel.

VIKTORVAV99:
> I like the simple approach! Doesn't seem to work for me though 🤔 https://github.com/electricitymaps/electricitymaps-contrib/assets/3296643/7ea6974f-dec7-4e16-add2-5d0bc967dc6b

Hehe, turns out these are hardcoded in storybook...

```js
export const Default: Story = {
  argTypes: {
    ratio: {
      control: { type: 'number', min: 0, max: 1, step: 0.1 },
    },
  },
  args: {
    ratio: 0.5,
    name: 'Renewable', // <- This one...
  },
};
```

madsnedergaard:
Oh, that explains it! 😄

madsnedergaard (approved):

PR #6905, opened 2024-06-26
feat(web): Display preliminary instead of estimated when using TSA

tonypls:
## Issue
We have changed the estimation card to read Preliminary for time slicer average data but not the rest of the badges

## Description
This PR updates the chart badges to display "Preliminary" instead of "Estimated" when the estimation method is Time Slicer Average. 

The map tooltip data does not have the estimation method data to conditionally display here as it only gets estimated as a boolean. Setting this as preliminary for hour would not be accurate for mode reconstruct breakdown zones.


web/src/features/charts/graphUtils.ts
```diff
@@ -198,6 +198,14 @@ export function getBadgeText(chartData: AreaGraphElement[], t: TFunction) {
     (day) => day.meta.estimationMethod || day.meta.estimatedPercentage === 100
   );
 
+  const allTimeSlicerAverageMethod = chartData.every(
+    (day) => day.meta.estimationMethod === EstimationMethods.TSA
+  );
```
VIKTORVAV99, line 203: I'm wondering if it doesn't start to make sense to create a custom function for these instead so we only loop over it once and set different Boolean flags instead?

Or at the very least re arrange these so the TSA is checked first and then the early TSA return then the rest of the function as it were.

web/src/hooks/getEstimationTranslation.ts
```diff
@@ -0,0 +1,21 @@
+import { useTranslation } from 'react-i18next';
+
+export function useGetEstimationTranslation(
+  field: 'title' | 'pill' | 'body',
+  estimationMethod?: string,
```
VIKTORVAV99, line 5: Since we are using Enums now this should accept the enum only.

web/src/hooks/getEstimationTranslation.ts
```diff
@@ -0,0 +1,21 @@
...
+  estimatedPercentage?: number
+) {
+  const { t } = useTranslation();
+
+  const exactTranslation =
+    (estimatedPercentage ?? 0) > 0 && estimationMethod === 'aggregated'
+      ? t(`estimation-card.aggregated_estimated.${field}`, {
+          percentage: estimatedPercentage,
+        })
+      : t(`estimation-card.${estimationMethod?.toLowerCase()}.${field}`);
```
VIKTORVAV99, line 15: Should we change the enums to be lower case by default so we don't need to call `toLowerCase()`?

web/src/utils/constants.ts
```diff
@@ -208,3 +208,12 @@ export const sourceLinkMapping: { [key: string]: string } = {
   '2021 annual mean carbon intensity by Electricity Maps':
     'https://github.com/electricitymaps/electricitymaps-contrib/wiki/Emission-factors#emission-factors-for-storage',
 };
+
+export enum EstimationMethods {
+  TSA = 'ESTIMATED_TIME_SLICER_AVERAGE',
+  CONSTRUCT_BREAKDOWN = 'ESTIMATED_CONSTRUCT_BREAKDOWN',
+  CONSTRUCT__ZERO_BREAKDOWN = 'ESTIMATED_CONSTRUCT_ZERO_BREAKDOWN',
```
VIKTORVAV99, line 215:
```suggestion
  CONSTRUCT_ZERO_BREAKDOWN = 'ESTIMATED_CONSTRUCT_ZERO_BREAKDOWN',
```

VIKTORVAV99 (commented):
I really like this direction with the use of Enums, added some more small initial feedback points.

web/src/hooks/getEstimationTranslation.ts
tonypls: Sometimes they are taken from the api object and are uppercase or lowercase, I'll make it so the translations and enums match the casing but I think in the future could improve this by matching this as one case between front and backend.

web/src/hooks/getEstimationTranslation.ts
tonypls: Done! ...and in many other places 😎

web/src/features/charts/graphUtils.ts
```diff
@@ -193,19 +193,38 @@ export function getElectricityProductionValue({
...
+      estimatedCount++;
+    }
+  }
+
+  return {
+    allTimeSlicerAverageMethod: tsaCount === chartData.length,
+    allEstimated: estimatedCount === chartData.length,
+    hasEstimation: estimatedCount > 0,
+  };
+}
```
VIKTORVAV99, line 214: Looks way better than all the redundat loops!

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for doing one step extra and ensuring we have strict types for the estimation methods!

web/src/features/charts/graphUtils.ts
tonypls: That functional evangelist in me really struggled typing `let` 😆

tonypls:
/review
--pr_reviewer.extra_instructions="
In the `possible issues` section, emphasize the following:
- Is the code logic efficient?
- 
"
--pr_reviewer.num_code_suggestions="4"

VIKTORVAV99:
Was this a accidental change we made here?
image
It should not say imprecise here but rather show the estimated percentage like the other badge.

tonypls:
> Was this a accidental change we made here? image It should not say imprecise here but rather show the estimated percentage like the other badge.

Good spotting, I'll look into it!

PR #6907, opened 2024-06-26
chore(deps-dev): vitest stable and fix code

VIKTORVAV99:
## Issue
We are on a very outdated vitest version and we should upgrade.

## Description

Updates vitest and the reporter to the latest version + some code changes to the scripts the new version complained about (which should be more performant anyhow).

This will bring several benefits but the biggest one might be the GitHub reporter that will automatically annotate any files that causes tests to fail.

### Double check


madsnedergaard (approved):
LGTM!

PR #6913, opened 2024-06-28
feat: include timezone in hourly displays

VIKTORVAV99:
## Description

Adds timezones to hourly time displays to reduce confusion in shared screenshots and the like.

##### SV:
image

##### EN:
image

##### SV:
image

##### EN:
image

### Double check


web/src/utils/formatting.test.ts
```diff
@@ -247,3 +256,110 @@ describe('formatCo2', () => {
...
+describe('formatDate', () => {
+  it('handles hourly data', () => {
+    const actual = formatDate(
+      new Date('2021-01-01T00:00:00Z'),
+      'en',
+      TimeAverages.HOURLY
+    );
+    const expected = 'January 1, 2021 at 1:00 AM GMT+1';
+    expect(actual).to.deep.eq(expected);
+  });
```
silkeholmebonnen, line 323: Can we create one more test on another language than en?

silkeholmebonnen (approved):
Nice work! One optional suggestion:)

web/src/utils/formatting.test.ts
VIKTORVAV99: Sure thing, I can change this to a snapshot test and then we can add all the languages we want :)

silkeholmebonnen:
I just noticed this! Can you create a follow up PR to fix it?

image

VIKTORVAV99:
> I just noticed this! Can you create a follow up PR to fix it? Or maybe it's a good time to move the dates on tooltips now? [AVO-95](https://linear.app/electricitymaps/issue/AVO-95/app-move-the-date-on-tooltips) image

Ah!
I'll pull the ripcord on AVO-95 and do that tomorrow.

madsnedergaard:
I think we should have been a bit more thorough on checking all cases where a date is present 😅

Screenshot 2024-07-16 at 20 36 13

VIKTORVAV99:
> I think we should have been a bit more thorough on checking all cases where a date is present 😅
> 
> 
> 
> Screenshot 2024-07-16 at 20 36 13
> 
> 

This tooltip will soon get a revamp as well...
But yes we probably should have been more careful about this change, visually.

madsnedergaard:
Did you try/consider using date-fns or [dayjs with AdvancedFormat](https://day.js.org/docs/en/plugin/advanced-format) to display dates in a shorter way?

VIKTORVAV99:
> Did you try/consider using date-fns or [dayjs with AdvancedFormat](https://day.js.org/docs/en/plugin/advanced-format) to display dates in a shorter way?

Since we don't use dayjs I'm not super keen on replacing every date-fns function with it. Cause having two date time libraries makes no sense. And with date-fns we need to import all the locales separately (it's built to be as tiny as possible) and I think they would format things roughly the same.

In theory we can build or own locale with Intl and control the format more tightly but I simply don't think it's worth it.

Especially since the tooltips themselves are the problem as they have not been migrated to the new design yet.

madsnedergaard:
> > Did you try/consider using date-fns or [dayjs with AdvancedFormat](https://day.js.org/docs/en/plugin/advanced-format) to display dates in a shorter way?
> 
> Since we don't use dayjs I'm not super keen on replacing every date-fns function with it. Cause having two date time libraries makes no sense. And with date-fns we need to import all the locales separately (it's built to be as tiny as possible) and I think they would format things roughly the same.
> 
> In theory we can build or own locale with Intl and control the format more tightly but I simply don't think it's worth it.
> 
> Especially since the tooltips themselves are the problem as they have not been migrated to the new design yet.

I think it's a larger problem than just tooltips, but happy to discuss it separately at another time :)

This is on MacOS 14" with Arc:
Screenshot 2024-07-16 at 20 59 14

And this is in full-width on an external display:
Screenshot 2024-07-16 at 20 59 41

PR #6916, opened 2024-06-30
chore(deps): migrate to @tanstack/react-query@v5

VIKTORVAV99:
## Description

We should upgrade to take better advantage of newer features and integrations later on, with this upgrade we can later use https://jotai.org/docs/extensions/query which seems like it could simplify our data handling, especially when combined with derived and composed atoms.

### Double check


web/src/main.tsx
```diff
@@ -45,7 +45,7 @@ const queryClient = new QueryClient({
       retry: MAX_RETRIES,
       refetchOnWindowFocus: false,
       // by default data is cached and valid forever, as we handle invalidation ourselves
-      cacheTime: Number.POSITIVE_INFINITY,
+      gcTime: Number.POSITIVE_INFINITY,
```
VIKTORVAV99, line 48: I'm actually not sure we should disable garbage collection here, as I think that means the cache (garbage) grows indefinitely?

What was the original reasoning for this?

web/src/api/getMeta.ts
```diff
@@ -29,8 +29,9 @@ export async function getMeta(): Promise<Meta> {
 
 export function useMeta(): Meta {
   return (
-    useQuery<Meta>([QUERY_KEYS.META], async () => getMeta(), {
-      suspense: true,
+    useSuspenseQuery<Meta>({
+      queryKey: [QUERY_KEYS.META],
+      queryFn: async () => getMeta(),
     }).data ?? {}
```
VIKTORVAV99, line 35: I think we can remove the `.data ?? {}` part here as data should always be defined when using suspense.

web/src/main.tsx
madsnedergaard: Hmm, as far as I remember it is by default 5 min and it would refetch data files every 5 min which seemed very wasteful given that nothing will change within the same hour. This is still the same situation with Cloudflare cache, as we don't want the client to redownload data every 5 min.

And since we automatically invalidate at the switch of the hour, it also isn't useful in any way :)

I can't remember how it actually works with the garbage collection part... But since we will overwrite/fetch fresh data on every new session, will that not "overwrite" the old one?

VIKTORVAV99: I have no idea if it will overwrite it or save a reference to the old data. I know there are some features around tanstack router that allows you to serve the "old" data in case the request fails or temporarily while it's fetching the new stuff.

I'll have to dig into the details but should be fine to leave as is for now.

web/src/api/getWeatherData.ts
```diff
@@ -105,25 +105,28 @@ async function getWeatherData(type: WeatherType) {
...
+  type: WeatherType;
+}
+
+const useGetWeather = (options: UseWeatherQueryOptions) => {
+  const { type, ...queryOptions } = options;
+  // Ensure getWeatherData returns Promise<Maybe<GfsForecastResponse>>
+  // and FIVE_MINUTES is a number representing milliseconds
+  return useQuery<Maybe<GfsForecastResponse>>({
+    queryKey: [type],
+    queryFn: () => getWeatherData(type), // Removed unnecessary async/await
```
madsnedergaard, line 119: No need to keep this comment
```suggestion
    queryFn: () => getWeatherData(type),
```

web/src/api/getWeatherData.ts
```diff
@@ -105,25 +105,28 @@ async function getWeatherData(type: WeatherType) {
...
-  );
+interface UseWeatherQueryOptions
+  extends Omit<UseQueryOptions<Maybe<GfsForecastResponse>>, 'queryKey'> {
+  type: WeatherType;
+}
+
+const useGetWeather = (options: UseWeatherQueryOptions) => {
+  const { type, ...queryOptions } = options;
+  // Ensure getWeatherData returns Promise<Maybe<GfsForecastResponse>>
+  // and FIVE_MINUTES is a number representing milliseconds
```
madsnedergaard, line 116: This comment suggest we should name the constant differently (like `FIVE_MINUTES_IN_MS`) 😅

madsnedergaard (approved):
Code looks good to me :)

web/src/api/getWeatherData.ts
VIKTORVAV99: I mean all time in JS is in milliseconds but I guess this is what I get for asking CoPilot to document my changes...

I'll remove the comments that are not needed 😅

PR #6935, opened 2024-07-01
chore(deps-dev): remove whatwg-fetch

VIKTORVAV99:
## Description

We don't need this Polyfill anymore as mentioned in, https://github.com/electricitymaps/electricitymaps-contrib/pull/6933, so we can remove it 🙂.

### Double check


silkeholmebonnen (approved):
LGTM!

PR #6941, opened 2024-07-02
feat(web): add link to com website from header logo

tonypls:
## Issue
Were a subdomain of electrictymaps.com but it's not entirely clear how to get back
## Description
Based on some feedback from GTM, it would be normal for the logo to link back to the homepage.

This PR adds the link to electricitymaps.com on the header icon that opens in a new tab


VIKTORVAV99 (commented):
Do we want to mark it as an external link?

tonypls:
> Do we want to mark it as an external link?

Hmm, I like the have consistency but also don't love messing with the logo. 

It seems like in general people have pointer finger on their logo and at most a small animation.

VIKTORVAV99:
> > Do we want to mark it as an external link?
> 
> 
> 
> Hmm, I like the have consistency but also don't love messing with the logo. 
> 
> 
> 
> It seems like in general people have pointer finger on their logo and at most a small animation.

I'll let you decide, I think it's fine either way.

VIKTORVAV99 (approved):

tonypls:
> > > Do we want to mark it as an external link?
> > 
> > 
> > Hmm, I like the have consistency but also don't love messing with the logo.
> > It seems like in general people have pointer finger on their logo and at most a small animation.
> 
> I'll let you decide, I think it's fine either way.

Let's put it out there and we can update it later if we think of a better implementation

PR #6946, opened 2024-07-03
feat: update emission factors for ME, BA, RS, MK, AL

FelixDQ:
fixes GMM-81

Updates hydro and nuclear emission factors to use UNECE as with other European zones.


VIKTORVAV99 (approved):
LGTM! 🎉

Any reason AL is not included?

FelixDQ:
We don't have data for that zone as far as I can tell, so I excluded it for now

FelixDQ:
I added AL as well

PR #6947, opened 2024-07-03
Update Solar and Wind capacity for Portugal

Musketeiro:
Solar and wind figures have been updated as per the most recent REN data.


VIKTORVAV99 (changes requested):
Hi and thanks for keeping these updated!

While the new values look good we don't want to remove the old ones, if you could make that quick change we can merge this 🙌🏻

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for changing it!

PR #6948, opened 2024-07-03
chore(web): typing and refactoring

tonypls:
## Issue
Found some motivation to tidy up some long standing any types

## Description
This PR refactors and adds types where I could figure them out. There's still a few more to go but this is a nice start. I've tried to minimize the changes of the actual functional code.


web/src/features/charts/elements/GraphBackground.tsx
```diff
@@ -1,11 +1,22 @@
...
+  valueScale: ScaleLinear<number, number>;
+  datetimes: Date[];
+  mouseMoveHandler?: (index: number | null) => void;
+  mouseOutHandler?: () => void;
+  isMobile: boolean;
+  svgNode: SVGElement;
+}
+
+const GraphBackground: React.FC<GraphBackgroundProps> = React.memo(
   ({
```
VIKTORVAV99, line 20:
```suggestion
const GraphBackground = React.memo(function graphBackground({
```
And then the props in the inner function as types.

VIKTORVAV99: or we can just turn of this annoying ESLint rule...?

web/src/features/charts/elements/GraphBackground.tsx
```diff
@@ -1,70 +1,67 @@
...
+  datetimes,
+  mouseMoveHandler,
+  mouseOutHandler,
+  isMobile,
+  svgNode,
+}: GraphBackgroundProps) {
+  const [x1, x2] = useMemo(() => timeScale.range(), [timeScale]);
+  const [y2, y1] = useMemo(() => valueScale.range(), [valueScale]);
+  const width = useMemo(() => x2 - x1, [x1, x2]);
+  const height = useMemo(() => y2 - y1, [y1, y2]);
```
silkeholmebonnen, line 29: I'm not sure if it is worth using the useMemo here, and I think this looks a bit simpler:

`const width = x2 - x1;
   const height = y2 - y1;`

web/src/features/charts/elements/GraphBackground.tsx
```diff
@@ -1,70 +1,67 @@
...
+  displayName?: string;
+}
 
-    // Mouse hover events
-    let mouseOutRectTimeout: string | number | NodeJS.Timeout | undefined;
-    const handleRectMouseMove = (event_: any) => {
-      if (mouseOutRectTimeout) {
-        clearTimeout(mouseOutRectTimeout);
-        mouseOutRectTimeout = undefined;
-      }
```
silkeholmebonnen, line 29: I'm not sure that I understand why we had this in the first place. Do you know what the purpose of it was?

silkeholmebonnen (approved):
LGTM ⭐ Just one question and an optional suggestion

web/src/features/charts/elements/GraphBackground.tsx
```diff
@@ -1,70 +1,67 @@
...
+  );
 
-    // Don't render if the dimensions are not positive
-    if (width <= 0 || height <= 0) {
-      return null;
+  const handleRectMouseOut = useCallback(() => {
+    if (mouseOutHandler) {
+      mouseOutHandler();
     }
+  }, [mouseOutHandler]);
```
VIKTORVAV99, line 45:
```suggestion
    mouseOutHandler?.()
```

web/src/translation/i18n.ts
```diff
@@ -25,7 +25,7 @@ i18n
 i18n.on('languageChanged', function (lng: keyof typeof localeToFacebookLocale) {
   document.documentElement.setAttribute('lang', lng);
   // TODO: Use react-helmet to manage meta tags
-  document.title = `Electricity Maps | ${i18n.t('misc.maintitle')}`;
+  document.title = `Electricity Maps | ${t('misc.maintitle')}`;
```
VIKTORVAV99, line 28: Will this update when the language changes? I think that was the original reason we keept the i18n import here.

VIKTORVAV99 (commented):
Just some small comments

web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -194,17 +193,3 @@ export const getExchangesToDisplay = (
...
-  useEffect(() => {
-    const headerElement = document.querySelector('header');
-    if (headerElement) {
-      const height = headerElement.offsetHeight;
-      setHeaderHeight(height * 1.1);
-    }
-  }, [window.innerWidth, window.innerHeight]);
-
-  return headerHeight;
-};
```
VIKTORVAV99, line 210: Can you remove the useEffect and useState imports in this file that are no longer needed?

web/src/translation/i18n.ts
tonypls: It did in my testing but it would be great validate further

web/src/features/charts/elements/GraphBackground.tsx
tonypls: No idea but I've learned that if something seems complicated there's probably a good reason why it's in the code, I would only remove this if I could explain why it was there!

web/src/features/charts/elements/GraphBackground.tsx
tonypls: Good point, I've updated them

web/src/features/charts/elements/GraphBackground.tsx
tonypls: I kinda like the explicit way an if statement reads but I've updated it anyway!

tonypls:
Add some updates from PR feedback 👍

VIKTORVAV99 (approved):
LGTM! 🎉

PR #6949, opened 2024-07-03
fix: Fix accordion violating rules of react

VIKTORVAV99:
## Issue
The accordion component is violating the rules of react and might cause issues and will not be optimised in React 19 when using the React Compiler. It also violates our lint rules about the rules of React.

## Description
I got inspired by @tonypls with the PR to fix our types in #6948 so I decided to fix some violations of my own.

This PR strips out the conditionals around the hooks and removes the use effect hook streamlining the entire component logic. It does this by requiring all accordions to have an atom that controls their state instead of marking it as an optional property.

There is no other (easy) way to do this without violating said rules I think we just have to save all accordion states going forward, I don't think that will be an issue personally but I'm interested in your opinion on this.

PS: This gives some insight as to why useEffect should be avoided if possible: https://react.dev/learn/you-might-not-need-an-effect

### Double check


web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -220,6 +223,8 @@ function BaseCard({
   );
 }
 
+const infoCardCollapsedAtom = atom<boolean | null>(true);
```
VIKTORVAV99, line 226: Having a shared state for these brings the benefit of reducing Content Layout Shifts when navigating between different zones or toggling the aggregations.

tonypls:
Would be nice to test on a branch preview!

VIKTORVAV99:
> Would be nice to test on a branch preview!

I have no idea why the CI pipeline didn't run but I took care of the merge conflict which triggered a CI run so you should have a preview link soon 🙂

VIKTORVAV99:
@tonypls I think I came up with a better solution while taking a swim.

What if we just moved out the state out of the Accordion component?

That way we could use either a local state with useState or a global state with useAtom that can be persisted. That should give us more options and be cleaner in general.

What do you think?

tonypls:
> @tonypls I think I came up with a better solution while taking a swim.
> 
> What if we just moved out the state out of the Accordion component?
> 
> That way we could use either a local state with useState or a global state with useAtom that can be persisted. That should give us more options and be cleaner in general.
> 
> What do you think?

image

tonypls:
> @tonypls I think I came up with a better solution while taking a swim.
> 
> What if we just moved out the state out of the Accordion component?
> 
> That way we could use either a local state with useState or a global state with useAtom that can be persisted. That should give us more options and be cleaner in general.
> 
> What do you think?

Swimming

PR #6950, opened 2024-07-03
feat: simplify dark mode class toggle logic

VIKTORVAV99:
## Description
Instead of using `.add()` and `.remove()` inside a conditional we simply use `.toggle()` with the boolean inside.

This is nicer and a tiny bit smaller, `113 bytes` vs `160 bytes`, which is an added bonus albeit very small.

### Double check


silkeholmebonnen (approved):
LGTM ⭐

PR #6952, opened 2024-07-04
feat(config): re-enable price for nordpool zones

tonypls:
## Issue
We had to remove price data for Nordpool zones!

## Description
This PR removes the filtering flags for price data that we previously didn't have the licensing to share. We've now purchased the license and can re-enable this feature 🎉


VIKTORVAV99 (commented):
Do you know if we also have to change the source values or is it enough that we have the license?

madsnedergaard:
Should we coordinate some SoMe content when releasing this? :)

tonypls:
> Do you know if we also have to change the source values or is it enough that we have the license?

Good question, I think the source is still ENTSOE but the source of the source is Nordpool. It's probably better to use Nordpool 🤔

tonypls:
> Should we coordinate some SoMe content when releasing this? :)

Good idea! We might be able to turn around some negative app reviews too. I've made issues here AVO-374 AVO-375

VIKTORVAV99:
> > Do you know if we also have to change the source values or is it enough that we have the license?
> 
> 
> 
> Good question, I think the source is still ENTSOE but the source of the source is Nordpool. It's probably better to use Nordpool 🤔 

We don't actually say the source for the price data in the app yet so I guess we could either skip it or hardcode it for now.

I think the ENTSO-E license also allows us to use it as long as we have the Nordpool license as well.

VIKTORVAV99 (approved):

PR #6954, opened 2024-07-04
feat: Update selectedDatetimeIndexAtom to use a Date object and use derived Atom for string format

VIKTORVAV99:
## Description

The selectedDatetimeIndexAtom in the atoms.ts file was updated to store the datetime as a Date object instead of a datetimeString. This change was made to improve consistency, simplify the code, reduce code duplication and increase the performance of the app.

To still be able to use the string value for the areas that require it I created a derived atom that takes the datetime from the index atom and returns it in a string format.

This change allows Vite to further optimise our bundles reducing the amount of preached bundles by 2 from 75 to 73 entries while also lowering the overall code size by 1kb.

### Double check

### DIFF
--- a/web/src/components/TimeDisplay.tsx
+++ b/web/src/components/TimeDisplay.tsx
@@ -1,4 +1,4 @@
-import { useAtomValue } from 'jotai/react';
+import { useAtomValue } from 'jotai';
 import { useTranslation } from 'react-i18next';
 import { formatDate } from 'utils/formatting';
 import { selectedDatetimeIndexAtom, timeAverageAtom } from 'utils/state/atoms';
--- a/web/src/components/TimeDisplay.tsx
+++ b/web/src/components/TimeDisplay.tsx
@@ -1,14 +1,16 @@
-import { useAtom } from 'jotai';
+import { useAtomValue } from 'jotai/react';
 import { useTranslation } from 'react-i18next';
 import { formatDate } from 'utils/formatting';
 import { selectedDatetimeIndexAtom, timeAverageAtom } from 'utils/state/atoms';
 
 export function TimeDisplay({ className }: { className?: string }) {
   const { i18n } = useTranslation();
-  const [timeAverage] = useAtom(timeAverageAtom);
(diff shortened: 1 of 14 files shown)


web/src/hooks/arrows.ts
```diff
@@ -55,22 +55,22 @@ export function filterExchanges(
...
   const zonesWithOutages = useMemo(() => {
-    const zoneData = data?.data?.datetimes?.[selectedDatetime?.datetimeString]?.z;
+    const zoneData = data?.data?.datetimes?.[selectedDatetimeString]?.z;
     return zoneData
       ? Object.entries(zoneData)
           .filter(([_, value]) => value.o)
           .map(([zone]) => zone)
       : [];
-  }, [data, selectedDatetime]);
+  }, [data?.data?.datetimes, selectedDatetimeString]);
```
silkeholmebonnen, line 70: Why can we change it to be `data?.data?.datetimes` instead of `data` as it was before?

web/src/components/TimeDisplay.tsx
```diff
@@ -1,14 +1,16 @@
-import { useAtom } from 'jotai';
+import { useAtomValue } from 'jotai/react';
```
silkeholmebonnen, line 1: Why do we import `useAtomValue` from `jotai` sometimes and `jotai/react` other times?

silkeholmebonnen (commented):
Nice work! Just a couple of questions:)

web/src/hooks/arrows.ts
VIKTORVAV99: Sure but this was a ESLint rule that made the suggestion/change. I'll revert it and then we can look into that separately.

web/src/components/TimeDisplay.tsx
VIKTORVAV99: Honestly I'm not sure, this is the import path that was suggested by VSCode. I can change it back and align them in another PR.

silkeholmebonnen (approved):
LGTM

PR #6955, opened 2024-07-04
feat(Nordpool parser): Nordpool price parser

VIKTORVAV99:
## Description
Adds a parser to parse price data directly from the Nordpool API instead of ENTSO-E for the supported regions.

Closes: AVO-379

### TODO:
- [x] Add snapshot tests to prevent regressions.

### Double check


parsers/NORDPOOL.py
```diff
@@ -0,0 +1,196 @@
...
+    "SE-SE2": "SE2",
+    "SE-SE3": "SE3",
+    "SE-SE4": "SE4",
+}
+
+INVERTED_ZONE_MAPPING = {value: key for key, value in ZONE_MAPPING.items()}
+
+
+def _generate_new_nordpool_token(session: Session | None = None) -> str:
+    global NORDPOOL_TOKEN_EXPIRATION
```
unitrium, line 68: In general, global variables are not the best because a variable might be modified unexpectedly in another context. 
If we keep it only to this file and it doesn't grow too much this should be ok. However, if we start fetching a lot of different types in parallel using code from this file this might cause problems in the long run. So overall, I'm ok for now but we should be open to other solutions.

parsers/NORDPOOL.py
```diff
@@ -0,0 +1,196 @@
...
+
+from electricitymap.contrib.lib.models.event_lists import PriceList
+from electricitymap.contrib.lib.types import ZoneKey
+
+from .lib.utils import get_token
+
+NORDPOOL_BASE_URL = "https://data-api.nordpoolgroup.com/api/v2/"
+
+# Set this as a global const outside of the functions to cache the token for the current parser instance
+NORDPOOL_TOKEN = None
```
unitrium, line 16: I think those two should go together so they are not out of sync. So I would recommend using something like:
```suggestion
@dataclass
class NordpoolToken:
   token: str | None = None
   expiration: timedelta | None = None
TOKEN = NordpoolToken()
```

unitrium (approved):
No major objections to this, in the long run maybe the global variable might cause troubles though but it's ok for now.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -48,6 +48,11 @@ def __contains__(self, datetime) -> bool:
     def __setitem__(self, datetime, event: Event):
         self.events[self.events.index(self[datetime])] = event
 
+    def __add__(self, other: "EventList") -> "EventList":
```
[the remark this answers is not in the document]
VIKTORVAV99: While true, this is what I would expect two lists being "added" to actually do.
I can define it further down in the class structure to only work with the same class though if you wish.

parsers/NORDPOOL.py
VIKTORVAV99: Would this also solve the global issue or would it cause the parsers to fetch a new token on every request?
I would like to avoid querying their auth server more than absolutely necessary.

parsers/NORDPOOL.py
```diff
@@ -0,0 +1,194 @@
...
+from electricitymap.contrib.lib.types import ZoneKey
+
+from .lib.utils import get_token
+
+NORDPOOL_BASE_URL = "https://data-api.nordpoolgroup.com/api/v2/"
+
+
+@dataclass
+class NordpoolToken:
+    token: str
```
unitrium, line 18: @VIKTORVAV99 I changed some behavior here.
I removed the possibility of having a None str or a None expiration date. If a token is created then it should always have both otherwise a token cannot be created.
This is a better practice than before by bundling the two and making sure they are always attached to each other.

parsers/NORDPOOL.py
```diff
@@ -0,0 +1,194 @@
...
+    return datetime_string.replace("Z", "+00:00")
+
+
+def _query_nordpool(
+    endpoint: NORDPOOL_API_ENDPOINT,
+    params: dict[str, str],
+    logger: Logger,
+    session: Session,
+):
+    global CURRENT_TOKEN
```
unitrium, line 135: @VIKTORVAV99 a tip here, if you need to update a global variable, you need to reference it with global otherwise you would only be updating the local reference.

unitrium: The previous code was not working because of that.

parsers/NORDPOOL.py
```diff
@@ -0,0 +1,194 @@
...
+def zulu_to_utc(datetime_string: str) -> str:
+    """Converts a zulu time string to a UTC time string."""
+    return datetime_string.replace("Z", "+00:00")
+
+
+def _query_nordpool(
+    endpoint: NORDPOOL_API_ENDPOINT,
+    params: dict[str, str],
+    logger: Logger,
+    session: Session,
```
unitrium, line 133: Regarding the session, it's better to have once a default parameter which can be None and then in all subsequent calls enforce it to be the expected type. Then you don't have to handle the None type anymore.
This was causing the issue you were experiencing with the tests as it was resetting the session due to a default argument which was None and therefore not using the mock.

parsers/test/test_NORDPOOL.py
```diff
@@ -0,0 +1,68 @@
...
+class TestNordpoolPrice(TestNordpool):
+    def test_price_parser_se(self):
+        mock_token = Path(base_path_to_mock, "token.json")
+        mock_data_current_day = Path(base_path_to_mock, "se_current_day_price.json")
+        mock_data_next_day = Path(base_path_to_mock, "se_next_day_price.json")
+
+        self.adapter.register_uri(
+            POST,
+            "https://sts.nordpoolgroup.com/connect/token",
+            json=loads(mock_token.read_text()),
```
unitrium, line 36: A note here, if you pass a string then the mock will return a string as it was before. But if you give it here a dict then it will return a dict when calling `.json`.

unitrium (commented):
I updated the code, now the tests are working and the global token seems cleaner to me :)

parsers/NORDPOOL.py
VIKTORVAV99: 🤦🏻‍♂️ 

Well that sure explains why nothing was working.

Thanks for fixing it!

parsers/NORDPOOL.py
VIKTORVAV99: Thanks!

I'm not really used to working with globals in python 😅.

VIKTORVAV99:
@unitrium would you feel comfortable with merging it like this? (I'll take care of the merge conflict.)

PR #6958, opened 2024-07-05
WIP: Tanstack router

VIKTORVAV99:
### Double check


VIKTORVAV99:
Closing this for now but the TL;DR is that it works pretty well, the migration is not super hard but there are some structural patterns we need to change and it's not entirely straight forward on how to do what we do on mobile (hide the ranking panel) but there are some solutions for that as well if we wanted to change.

PR #6960, opened 2024-07-05
fix: add missing exchange between DK and NL

VIKTORVAV99:
## Issue

We where missing a exchange confix for the aggregated exchange between DK and NL.

## Description

Adds the exchange config with the same location, rotation and capacity as the exchange between DK-DK1 and NL.

### Double check

- [ x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.


tonypls (approved):
LGTM

PR #6911, opened 2024-06-28
feat: Collapsable historical picker

VIKTORVAV99:
## Description
Adds a collapsible time controller/picker using an accordion instead of the bottom sheet (which is woefully unmaintained and outdated).

It also changes all necessary core components such the Badge, TimeAverageToggle, TimeAxis and TimeSlider to match with the new visual design.

Closes: AVO-357

### Preview
image
image

### Double check

- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

PR #6961, opened 2024-07-05
chore: Update the style of the legends + cleanup

VIKTORVAV99:
## Description

Updates the style of the legend container to match the new historical controller in #6911 simplify the DOM state by using conditional rendering instead of hidden classes and reuse the same wrapper component for all 3 legends.

### Preview
image

### Double check


silkeholmebonnen:
There are a merge conflict, so I will wait on approving until it is fixed but it looks good!

VIKTORVAV99:
I'll take care of the conflict right away :)

silkeholmebonnen:
You can only see half of the last 0. And idk if it is just because I'm used to how it looked before but wouldn't it look better if we made it a bit wider so that the text could fit on one line?
image

VIKTORVAV99:
@silkeholmebonnen should be ready to review/merge now 🙂

web/src/components/legend/LegendContainer.tsx
```diff
@@ -11,7 +11,7 @@ export default function LegendContainer(): ReactElement {
   const isWindLayerEnabled = useAtomValue(isWindLayerEnabledAtom);
 
   return (
-    <div className="invisible  flex w-[224px] flex-col rounded bg-white/90 px-1 py-2 shadow-xl backdrop-blur-sm sm:visible dark:bg-gray-800">
+    <div className="flex min-w-64 max-w-min flex-col gap-2 rounded-2xl bg-white/80 p-2 backdrop-blur dark:bg-gray-800/80">
```
silkeholmebonnen, line 14: I think it is a bit hard to see the difference between this element and the background in light mode now that there is no shadow.

Before:
image

After:
Screenshot 2024-07-26 at 09 05 10

silkeholmebonnen (commented):
Just a small comment, otherwise it looks good!

web/src/components/legend/LegendContainer.tsx
VIKTORVAV99: I'll double check how we do it for the other translucent elements. 👍🏼

VIKTORVAV99:
@Alportan turns out the toggle buttons (production/consumption for example) don't use a white background but rather `gray-200`, I think we should align these between all the "map elements" (time controller, toggles, legends, buttons) but idk if `gray-200` is optimal.

silkeholmebonnen:
@VIKTORVAV99 Let me know when this is ready for review:)

VIKTORVAV99:
> @VIKTORVAV99 Let me know when this is ready for review:)

From my side it's good to go but I think @Alportan might want to weigh in on the colors, but perhaps we will adjust that in it's own PR.

Alportan:
> @Alportan turns out the toggle buttons (production/consumption for example) don't use a white background but rather `gray-200`, I think we should align these between all the "map elements" (time controller, toggles, legends, buttons) but idk if `gray-200` is optimal.

Just summing up our back-and-forth on Slack @VIKTORVAV99 , we decided to still use a white background, with a soft grey border instead of the `grey-200` background, which draws unnecessary attention to the legend item.

In a future PR, we'll probably unify the other map elements to match this, and maybe even set up some global rules so we won't have to think about it in the future. 🙌

silkeholmebonnen (approved):
LGTM 🌟 
Just a small comment: It looks a bit weird that the feature flags container has a shadow when the carbon intensity legend doesn't, but since the feature flag is not something that a user would see it is not blocking. 
image

Issue #6592, opened 2024-03-29
Greece is partially coloured. Islands are left out


mpelas:
## Description
Aegean sea island are Greek but are left without colour. Crete is colorless too. From thasos to Kastelorizo one country as a whole. 
Avoid using incomplete vector layers. Greece is a archipelago country with so many islands worth visiting. 

Regards

VIKTORVAV99:
Hi!

Our zones are not meant to be matched 1:1 with country borders but rather the grid systems. When it was added, and the last time I checked with ENTSO-E late last year, the Greece islands were still not connected to the main grid. Which is why they have been kept separate as the grid conditions of Greece would not actually match what is happening on the island.

If you have information that they have since been combined and that the islands production and consumption is now included in the ENTSO-E data we have no issue merging the zones.

gerrykou:
Hey, yeah makes sense, there is an update on the network and you can find the update on the etsoe grid maps
https://www.entsoe.eu/data/map/
From the grid map, i can see those islands connected:
Crete, Andros, Tinos, Syros, Mykonos, Paros, Naxos.

VIKTORVAV99:
> Hey, yeah makes sense, there is an update on the network and you can find the update on the etsoe grid maps
> 
> https://www.entsoe.eu/data/map/
> 
> From the grid map, i can see those islands connected:
> 
> Crete, Andros, Tinos, Syros, Mykonos, Paros, Naxos.
> 
> 

I can see that they have indeed made some progress which is great!
I still see that some of the lines are dotted which means they are under construction but I'll send an email to ENTSO-E double checking the status and if the reported production now contains these islands.

PR #6962, opened 2024-07-06
chore: Merge GR zones

VIKTORVAV99:
## Issue

Closes: #6592

## Description

Previously these islands where not connected to the mainland grid but this has since changed with high voltage AC and DC power lines and the zones should now be merged.

ENTSO-E has also signaled that the production these islands will be included in the Greece production values.

### Preview
image

### Double check


silimotion:
Two things:

- As the islands have been recently connected, shouldn't they show divided in the historical view?
- Does the capacity data available in entsoe reflect the recent changes in topology?

Thanks!

VIKTORVAV99:
> Two things:
> 
> 
> 
> - As the islands have been recently connected, shouldn't they show divided in the historical view?
> 
> - Does the capacity data available in entsoe reflect the recent changes in topology?
> 
> 
> 
> Thanks!

For the first one:
There is a two part problem here, firstly we don't have the technical capacity to do this right now simply because the map is a static configuration. This might change in the future.

Secondly, which is a bigger problem imo as it's outside of our control is simply when they started reporting the combined values. As far as I know we where just previously operating on the assumption that the grids are separated and therefore exchange values can't affect the islands, but seemingly there is nothing blocking the TSO from reporting the production itself as part of Greece so it might be more accurate to include it in the historical view in any case.

Happy to discuss this further.

As for capacity, ENTSO-E only has yearly values so it might not reflect that until we have 2025 values for Greece, but once we do it should be available. But I can double check with ENTSO-E just to be sure.

As a note I did not see any sudden increase in production capacity between 2023 and 2024.

Depending on the generator sizes it might not be a legal requirement to include these either.

tonypls (approved):
LGTM, do we also want to remove all the translations for GR-IS? 

Would also be nice to link any relevant announcements / articles about the connections to the mainland in this PR as a future reference

VIKTORVAV99:
> LGTM, do we also want to remove all the translations for GR-IS?
> 
> Would also be nice to link any relevant announcements / articles about the connections to the mainland in this PR as a future reference

Will do a quick removal of the translations 👍🏻

And then update the PR with relevant links.

PR #6970, opened 2024-07-09
feat(web): Add tooltip to production source legends in BreakdownChart

silkeholmebonnen:
## Issue

[AVO-234](https://linear.app/electricitymaps/issue/AVO-234/origin-of-electricity-add-legend-tooltip-for-sources)

## Description

This PR adds a tooltip to the production source legend list in the BreakdownChart.

### Preview
image
image

Mobile:
image
image

### Double check

### DIFF
--- a/web/src/features/charts/ProductionSourceLegendList.tsx
+++ b/web/src/features/charts/ProductionSourceLegendList.tsx
@@ -61,9 +61,9 @@ function ProductionSourceTooltip({
 function ProductionSourceTooltipInner({ sources }: { sources: ElectricityModeType[] }) {
   return (
     <div className="flex flex-col gap-1.5">
-      {sources.map((source, index) => (
+      {sources.map((source) => (
         <div key={source} className="flex flex-row gap-2">
-          <ProductionSourceLegend key={index} electricityType={source} />
+          <ProductionSourceLegend key={source} electricityType={source} />
           <div className="text-xs font-medium">{source}</div>
         </div>
       ))}
--- a/web/src/features/charts/ProductionSourceLegendList.tsx
+++ b/web/src/features/charts/ProductionSourceLegendList.tsx
@@ -20,8 +20,8 @@ export default function ProductionSourceLegendList({
       tooltipContent={<ProductionSourceTooltip sources={sources} isMobile={isMobile} />}
       tooltipClassName={
         isMobile
-          ? 'bg-transparent shadow-none border-transparent dark:bg-transparent items-center flex flex-col p-0'
-          : 'rounded-2xl min-w-44 dark:bg-gray-900/80 dark:border-1 dark:border-gray-700 mx-5 backdrop-blur bg-white/80'
+          ? 'bg-transparent shadow-none border-transparent dark:bg-transparent items-center flex flex-col p-0 mb-2'
+          : 'rounded-2xl min-w-44 dark:bg-gray-900/80 dark:border-1 dark:border-gray-700 mx-5 backdrop-blur bg-white/80 mb-2'
       }
       side="bottom"
       isMobile={isMobile}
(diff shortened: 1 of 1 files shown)


silkeholmebonnen:
@Alportan Do you have any feedback?

Alportan:
> @Alportan Do you have any feedback?

Looking good! 😎 Maybe one small thing, could we use the frosted background blur on the modal? 🙌 
image

VIKTORVAV99:
> > @Alportan Do you have any feedback?
> 
> Looking good! 😎 Maybe one small thing, could we use the frosted background blur on the modal? 🙌 image
> 
> I believe we already use it on a few components/elements.

I have been using background-blur (not sm) to not get the edges to bleed tough, it would be nice if it was the same here :)

silkeholmebonnen:
I added the blur and I also changed the mobile version to not be in the middle of the screen, but right above or below the production source icons, because thats what it looked like in figma. Will you take a look at it again @Alportan, and let me know if you would rather have in be in the center on mobile.

Alportan:
> I added the blur and I also changed the mobile version to not be in the middle of the screen, but right above or below the production source icons, because thats what it looked like in figma. Will you take a look at it again @Alportan, and let me know if you would rather have in be in the center on mobile.

Looks good with the blur! 🫶

Regarding the center modal on mobile, that's a good question. I'm unsure about it for now because adding a tooltip on top of the graph might be disruptive, as the main goal is to help users identify sources on the graph.

It's important to keep the design consistent and avoid too many custom components or logic. However, we should review all current use-cases first. There’s a future task to [assess and unify tooltips](https://www.notion.so/electricitymaps/Revise-and-unify-all-tooltips-d4f78af5dcba4ab695b940f33f0ed4a0?pvs=4), so we can start with this implementation and refine it later as needed.

silkeholmebonnen:
> > I added the blur and I also changed the mobile version to not be in the middle of the screen, but right above or below the production source icons, because thats what it looked like in figma. Will you take a look at it again @Alportan, and let me know if you would rather have in be in the center on mobile.
> 
> Looks good with the blur! 🫶
> 
> Regarding the center modal on mobile, that's a good question. I'm unsure about it for now because adding a tooltip on top of the graph might be disruptive, as the main goal is to help users identify sources on the graph.
> 
> It's important to keep the design consistent and avoid too many custom components or logic. However, we should review all current use-cases first. There’s a future task to [assess and unify tooltips](https://www.notion.so/electricitymaps/Revise-and-unify-all-tooltips-d4f78af5dcba4ab695b940f33f0ed4a0?pvs=4), so we can start with this implementation and refine it later as needed.

Okay, I'll keep it as it is for now then:)

web/src/features/charts/ProductionSourceLegendList.tsx
```diff
@@ -10,10 +13,59 @@ export default function ProductionSourceLegendList({
...
+          ? 'bg-transparent shadow-none border-transparent dark:bg-transparent items-center flex flex-col p-0 mb-2'
+          : 'rounded-2xl min-w-44 dark:bg-gray-900/80 dark:border-1 dark:border-gray-700 mx-5 backdrop-blur bg-white/80 mb-2'
+      }
+      side="bottom"
+      isMobile={isMobile}
+    >
+      <div className={twMerge('flex w-fit flex-row gap-1.5 py-1', className)}>
+        {sources.map((source, index) => (
+          <ProductionSourceLegend key={index} electricityType={source} />
+        ))}
```
VIKTORVAV99, line 32: As the bot suggested you can use the source object directly as a key here as well just as you did later in the code.

VIKTORVAV99 (approved):
LGTM! 🎉

Just one minor thing but nothing blocking! Nice work!

PR #6974, opened 2024-07-11
feat: Custom zoom buttons

VIKTORVAV99:
## Issue
There was two todos and AVO-342.

## Description
Uses a `<MapProvider>` and later the `useMap()` hook to implement our own zoom buttons, this allows us to keep a unified design, remove the maplibre css and implement the aria labels that was mentioned in the todo comments.

>[!NOTE]
>These icons are a mess, being in different sizes and from different libraries with different styles. I and Alex agreed that there should be an operate PR to fix this later on.

Closes: AVO-342

>[!IMPORTANT]
> We can probably use the `useMap()` hook in more places to simplify our logic a bit but that's something for a later PR.

#### New:
image

#### Old:
image

### Double check


web/src/features/map-controls/MapButton.tsx
```diff
@@ -1,5 +1,4 @@
-import * as Toggle from '@radix-ui/react-toggle';
```
VIKTORVAV99, line 1: Should we move this toggle thing to the button component or is it fine to just get rid of it?

web/src/components/Button.tsx
```diff
@@ -9,12 +9,14 @@ interface ButtonProps {
   disabled?: boolean;
   size?: SizeOptions;
   shouldShrink?: boolean;
-  type?: 'primary' | 'secondary' | 'tertiary' | 'link';
+  type?: 'primary' | 'secondary' | 'tertiary' | 'link' | 'transparent';
```
VIKTORVAV99, line 12: Maybe this should be called something like glass instead?

madsnedergaard:
> These icons are a mess, being in different sizes and from different libraries with different styles. I and Alex agreed that there should be an operate PR to fix this later on.

Yes, we should align these when possible, but be aware that we cannot get them from the same package so we do need to maintain our own custom icons here. They could be aligned and stylised in one common way though :)

VIKTORVAV99:
> > > > These icons are a mess, being in different sizes and from different libraries with different styles. I and Alex agreed that there should be an operate PR to fix this later on.
> > > 
> > > 
> > > Yes, we should align these when possible, but be aware that we cannot get them from the same package so we do need to maintain our own custom icons here. They could be aligned and stylised in one common way though :)
> > 
> > 
> > Why can't we get them from the same package? I'm just talking about the button icons here and all of them appears to be available in FontAwesome6 (free).
> 
> Because they don't have the same icons necessarily - see the translation and solar icons that look quite different in FontAwesome. Not saying we shouldn't change the icons, but I wanted to clarify why we in the past decided not to use the same icon package :)

That's why it's a separate PR, we need to figure out exactly which icons to use. I use the Sun and Moon icon from FontAwesome6 in #6982 which are the ones that Alex have used in the Figma designs (but for the charts). We might want to align it here as well but we should probably talk it through first.

web/src/features/map-controls/MapButton.tsx
```diff
@@ -9,35 +8,29 @@ interface MapButtonProperties {
   tooltipText?: string;
   className?: string;
   dataTestId?: string;
-  asToggle?: boolean;
   ariaLabel?: string;
+  backgroundClasses?: string;
 }
 
 export default function MapButton({
```
madsnedergaard, line 15: We now have a button within a button, which messes up keyboard navigation:

Screenshot 2024-07-16 at 15 04 43

https://github.com/user-attachments/assets/53441915-adf5-4244-82bf-5e2e0fa8b4bb

madsnedergaard (commented):
Some comments, but so nice to get rid of the hacky implementation :)

web/src/components/Button.tsx
VIKTORVAV99: eeh, I guess that's fine with type checking but it is harder to spell for sure...

I'll change it though.

web/src/features/map-controls/MapButton.tsx
VIKTORVAV99: the asDiv prop I implemented for the update prompt should fix this. I'll double check that it doesn't cause any issues first though.

web/src/features/map-controls/ZoomControls.tsx
```diff
@@ -1,23 +1,29 @@
...
-      //TODO: Find a way to use a __('tooltips.zoomIn') as aria-label here
-      //TODO: Find a way to use a __('tooltips.zoomOut') as aria-label here
-    />
+    <div className="flex flex-col">
+      <MapButton
+        icon={<FaPlus size={20} />}
+        onClick={() => map?.zoomIn()}
+        ariaLabel={t('tooltips.zoomIn')}
+        dataTestId="zoom-in"
+        backgroundClasses="rounded-none rounded-t-full"
```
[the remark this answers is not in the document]
VIKTORVAV99: I'll change it but honestly this looks better than the alternative to me 😅

VIKTORVAV99:
It seems like the wind layer stopped working...?
Which I don't get cause the solar layer is still working as expected so it's not the buttons that broke it as the logic is shared between those.

madsnedergaard:
> It seems like the wind layer stopped working...? Which I don't get cause the solar layer is still working as expected so it's not the buttons that broke it as the logic is shared between those.

Strange indeed, it works on other preview branches so it's something specifically here that is causing it 🤔

VIKTORVAV99:
> Can't remember anything about it at all, but here be dragons and we have struggled with it - happy to help on it :)

I'll see if I can make them always on and then conditionally render them instead. Hopefully that solves the issue and makes it easier for us to unify the logic and get rid of a few use effect hooks.

madsnedergaard:
Hey, should we put this back in draft? Seems like there's still quite some work to be done to ensure the wind layers work 😬

VIKTORVAV99:
> Hey, should we put this back in draft? Seems like there's still quite some work to be done to ensure the wind layers work 😬 

Yeah we might as well. I still have no idea why they actually break 😅

They shouldn't IMO but I haven't had time to look into it.

madsnedergaard:
Let's close this for now, I think we might end up reworking the controls in the future and then we'll keep this in mind :)

PR #6975, opened 2024-07-11
Type hint Fix

ZhipengXue97:
## Description
Fix type check warnings reported by Pyre@Google, which were outdated after code modifications.

## Detail
1. update the return type of function `fetch_production` from `dict[str, Any]` to `Optional[dict[str, Any]]`, since it could raise an exception and return None after commit https://github.com/electricitymaps/electricitymaps-contrib/commit/9c8fed25a8be3d9ce91abd368015a10faadb12d0
2. update the return type of function `fetch_cea_production` from `dict[str, Any]` to `Optional[dict[str, Any]]`, since it could raise an exception and return None after commit https://github.com/electricitymaps/electricitymaps-contrib/commit/da640af193365fdaabcf41276bf2a27810a06334

### DIFF
--- a/parsers/CA_AB.py
+++ b/parsers/CA_AB.py
@@ -10,7 +10,7 @@ import re
 import urllib.parse
 from datetime import datetime, timedelta
 from logging import Logger, getLogger
-from typing import Any, Optional
+from typing import Any
 from zoneinfo import ZoneInfo
 
 # Third-party library imports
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
Hi!

We are using PyRight locally at the moment which is not showing these type issues.
Could you maybe explain a bit more why they are needed?

ZhipengXue97:
Done, update the type to fit ruff format

VIKTORVAV99 (approved):
Thanks!

LGTM now! 🎉

PR #6980, opened 2024-07-12
perf: Add useCallbacks and isHourly derived atom

VIKTORVAV99:
## Description
This PR does two things:

1. Wraps two component callbacks in useCallback so they are not re-created on every render.
2. Creates and uses a derived atom of `isHourly`.

### Double check


madsnedergaard (approved):
Looks good to me! :)

PR #6981, opened 2024-07-13
perf: use derived atoms for solar and wind state

VIKTORVAV99:
## Description

Uses derived atoms for the wind and solar enabled state to reduce code duplication and minimise unneeded imports.

### Double check

### DIFF
--- a/web/src/utils/state/atoms.ts
+++ b/web/src/utils/state/atoms.ts
@@ -29,8 +29,19 @@ export const spatialAggregateAtom = atomWithStorage(
 );
 export const productionConsumptionAtom = atomWithStorage('mode', Mode.CONSUMPTION);
 
-export const solarLayerEnabledAtom = atomWithStorage('solar', ToggleOptions.OFF);
+export const solarLayerAtom = atomWithStorage('solar', ToggleOptions.OFF);
+export const isSolarLayerEnabledAtom = atom(
+  (get) =>
+    get(solarLayerAtom) === ToggleOptions.ON &&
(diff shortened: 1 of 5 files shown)


web/src/utils/state/atoms.ts
```diff
@@ -29,8 +29,19 @@ export const spatialAggregateAtom = atomWithStorage(
 );
 export const productionConsumptionAtom = atomWithStorage('mode', Mode.CONSUMPTION);
 
-export const solarLayerEnabledAtom = atomWithStorage('solar', ToggleOptions.OFF);
+export const solarLayerAtom = atomWithStorage('solar', ToggleOptions.OFF);
+export const isSolarLayerEnabledAtom = atom(
+  (get) =>
+    get(solarLayerAtom) === ToggleOptions.ON &&
+    get(selectedDatetimeIndexAtom).index === 24
```
madsnedergaard, line 36: We need to add a check for `isHourly` (from #6980), as it is currently possible to make the solar layer show on daily view (on day 25) 😬 

Screenshot 2024-07-16 at 14 40 24

madsnedergaard (commented):
LGTM, but reading the existing code made me realise we have a silly bug in here :)

web/src/utils/state/atoms.ts
VIKTORVAV99: Should be fixed now

madsnedergaard (approved):
Thanks for fixing the bug :)

PR #6982, opened 2024-07-15
perf: use icons instead of SVGs in TimeSlider

VIKTORVAV99:
## Description

Splits out icon changes in TimeSlider from #6911  so it is easier to review that PR and can be merged faster.

The reasoning for doing this and not using the SVG files as backgrounds are two fold. One we always need them so it's smart to bundle them directly with the code and reduce the http calls we make. Two they no longer match our visual design of the other icons and changing them to Font Awesome versions fixes that as we use Font Awesome in other parts of the app.

### Preview

image
image
image
image
image
image

### Double check


web/src/components/TimeSlider.tsx
```diff
@@ -3,16 +3,14 @@ import { scaleLinear } from 'd3-scale';
 import { useNightTimes } from 'hooks/nightTimes';
 import { useDarkMode } from 'hooks/theme';
 import { useAtom } from 'jotai';
+import { ReactElement } from 'react';
+import { FaArrowsLeftRight, FaMoon, FaSun } from 'react-icons/fa6';
```
madsnedergaard, line 7: I think the sun icon looks pretty bad, especially on a non-retina screen:

Screenshot 2024-07-16 at 20 46 25

I'd much rather use the existing icons, but turn them into inline React components instead as a first step 🤷

web/src/components/TimeSlider.tsx
```diff
@@ -62,14 +60,19 @@ export const getTrackBackground = (
 export const getThumbIcon = (
   selectedIndex?: number,
   sets?: NightTimeSet[]
-): ThumbIconPath => {
+): ReactElement => {
+  const size = 14;
   if (selectedIndex === undefined || !sets || sets.length === 0) {
-    return 'slider-thumb.svg';
+    return <FaArrowsLeftRight size={size} pointerEvents="none" />;
```
madsnedergaard, line 66: I might be biased here, but I think the existing one is much cleaner and nicer than the new FA one.

Screenshot 2024-07-16 at 20 49 14

madsnedergaard (commented):
How about keeping existing icons (but moving them to React components) for now, and then we later on take a thorough revisit to all icon usage instead? I personally don't think it's worth using a consistent icon package if it means sacrificing eligibility and UX  
¯\\\_(ツ)\_/¯

VIKTORVAV99:
> How about keeping existing icons (but moving them to React components) for now, and then we later on take a thorough revisit to all icon usage instead? I personally don't think it's worth using a consistent icon package if it means sacrificing eligibility and UX  
> 
> ¯\\\_(ツ)\_/¯

The problem with the existing icons are that they are not uniform and scaling them is not the easiest without a lot of manual work.

There are outline versions of these icons as well but I would much prefer to use the icon library than have them as custom components since that means we will use a consistent icon style. I'd be happy to discuss it on the design review it's Alex though if we have the time (tomorrow).

madsnedergaard:
> > How about keeping existing icons (but moving them to React components) for now, and then we later on take a thorough revisit to all icon usage instead? I personally don't think it's worth using a consistent icon package if it means sacrificing eligibility and UX
> > ¯\_(ツ)_/¯
> 
> The problem with the existing icons are that they are not uniform and scaling them is not the easiest without a lot of manual work (or runtime calculations).
> 
> There are outline versions of these icons as well but I would much prefer to use the icon library than have them as custom components since that means we will use a consistent icon style. I'd be happy to discuss it on the design review it's Alex though if we have the time (tomorrow).

Did you have a chance to discuss this yesterday? :)

VIKTORVAV99:
> > > How about keeping existing icons (but moving them to React components) for now, and then we later on take a thorough revisit to all icon usage instead? I personally don't think it's worth using a consistent icon package if it means sacrificing eligibility and UX
> 
> > > ¯\_(ツ)_/¯
> 
> > 
> 
> > The problem with the existing icons are that they are not uniform and scaling them is not the easiest without a lot of manual work (or runtime calculations).
> 
> > 
> 
> > There are outline versions of these icons as well but I would much prefer to use the icon library than have them as custom components since that means we will use a consistent icon style. I'd be happy to discuss it on the design review it's Alex though if we have the time (tomorrow).
> 
> 
> 
> Did you have a chance to discuss this yesterday? :)

I did not, so let's pause this PR for a bit until next week.

I don't necessarily believe that this impacts the UX in a negative way though. Having a uniform icon style brings it's own UX benefits.

But I'll test it a bit on lower resolution devices as well and see how it looks.

web/src/components/TimeSlider.tsx
VIKTORVAV99: We decided to switch to Lucide icons as you suggested so hopefully it looks good now 😬.

tonypls:
Code looks good to me, but I think it's more a UX change that is suited to @Alportan to review

VIKTORVAV99:
The proposed updates should be live now, let me know what you think 🙂

tonypls (approved):
LGTM

VIKTORVAV99:
@Alportan you have any additional feedback or can I merge this?

Alportan:
> @Alportan do you have any additional feedback or can I merge this?

Looking good, green light from my side! 🎉  You can merge! 😎

PR #6985, opened 2024-07-15
feat(config): add tag in zones config for generation only zones

FelixDQ:
fixes GMM-94

Adds a tag to generation only zones to allow identification in the backend. These zones should be handled differently as they don't have any consumption.

source: https://www.eia.gov/electricity/gridmonitor/about

### DIFF
--- a/config/zones/US-CAR-YAD.yaml
+++ b/config/zones/US-CAR-YAD.yaml
@@ -205,3 +205,4 @@ sources:
   IPCC 2014:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
 timezone: US/Eastern
+generation_only: true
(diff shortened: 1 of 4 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

madsnedergaard (approved):
LGTM as well, I'll get to work on geometries! 🤓

madsnedergaard (changes requested):
Spoke too soon, we also need to update the TS interface `ZoneConfig` in here: `web/geo/types.ts`

VIKTORVAV99:
> Spoke too soon, we also need to update the TS interface `ZoneConfig` in here: `web/geo/types.ts`

Is there any way to catch that as a part of the CI run as well so we don't miss this in the future?

madsnedergaard (approved):
Back to LGTM, fixed it in #6988 instead 😄

PR #6987, opened 2024-07-15
perf: set experimental min chunk size to 3500 bytes

VIKTORVAV99:
## Description

Sets the experimentalMinChunkSize to 3500 bytes which removes 2 of the smallest chunks and bundles them with a common parrent of the module that use it instead.

### Double check


madsnedergaard:
Would you mind explaining why this is preferable to more smaller files? :)

VIKTORVAV99:
> Would you mind explaining why this is preferable to more smaller files? :)

Basically if the files are smaller than roughly 4kb with overhead it costs more to crate the http connection than you gain by splitting it. And since browsers impose a limit on the maximum of parallel connections (8 on chromium and 4 on safari I believe) waiting for the server takes longer than downloading the actual content which means they are blocking other requests without any real benefit.

madsnedergaard:
> > Would you mind explaining why this is preferable to more smaller files? :)
> 
> Basically if the files are smaller than roughly 4kb with overhead it costs more to create the http connection than you gain by splitting it up in different files. And since browsers impose a limit on the maximum of parallel connections (8 on chromium and 4 on safari I believe) waiting for the server takes longer than downloading the actual content which means they are blocking other requests without any real benefit.

I was under the impressions that with HTTP/2 multiple smaller files would always be beneficial, but that makes sense - thanks for explaining :)

madsnedergaard (approved):
LGTM

VIKTORVAV99:
> > > Would you mind explaining why this is preferable to more smaller files? :)
> 
> > 
> 
> > Basically if the files are smaller than roughly 4kb with overhead it costs more to create the http connection than you gain by splitting it up in different files. And since browsers impose a limit on the maximum of parallel connections (8 on chromium and 4 on safari I believe) waiting for the server takes longer than downloading the actual content which means they are blocking other requests without any real benefit.
> 
> 
> 
> I was under the impressions that with HTTP/2 multiple smaller files would always be beneficial, but that makes sense - thanks for explaining :)

Http/2 and later H3 does improve it a lot and it does make sense to split up the files to benefit from multiplexing but as it is right now er are already saturating the allowed connections.

Basically you want to split them up if your just loading 1 file at the time and combine them if your loading 8+ at the time.

PR #6963, opened 2024-07-06
chore: clean up world and round to 3 decimals

VIKTORVAV99:
## Issue

The world file contained a lot of issues around duplicate coordinate, gaps, miss alignments and more. This prevented us from using a resolution of 0.001 degrees instead of 0.0001 degrees. Making the generated world file larger than it should have been.

## Description

I've taken care of all the issues I could find, fixed the alignment of some US and Nordic zones and changed the resolution to 0.001 degrees as it was originally supposed to have.

#### From:
image
#### To:
image

### Double check

- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [ ] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

PR #6988, opened 2024-07-15
Adds new generation_only type and ensure it's available for scripts

madsnedergaard:
## Description

Removes 3 generation-only zones from the US geometries:
- `US-CAR-YAD`
- `US-NW-GRID`
- `US-SE-SEPA`

We still want to parse the data and make it accessible in the app via URL directly, but we do not want to show the zones on the map. Therefore we remove them from the geometries only.

Closes: AVO-398

### Preview
Screenshot 2024-07-15 at 23 13 03

### Double check


madsnedergaard:
There's something weird going on with GR-IS, dunno why it's now gone from the map: https://cf2f96ab.electricitymaps-staging.pages.dev/zone/GR-IS 🤔

VIKTORVAV99:
> There's something weird going on with GR-IS, dunno why it's now gone from the map: https://cf2f96ab.electricitymaps-staging.pages.dev/zone/GR-IS 🤔 
> 
> 
> 
> **EDIT:** Lol, I couldn't understand why my local build was different from prod-app, but now I see #6962 that hasn't been released yet 🤦 

There is also the massive PR at #6963 this is going conflict with 😬. So many changes to the world file in that since I cleaned up all the datapoints and corrected many of the us bidding zones as well as the Nordic ones to match SVK.se.

web/scripts/generateZonesConfig.ts
```diff
@@ -37,6 +37,7 @@ const getConfig = (): CombinedZonesConfig => {
     'parsers',
     'subZoneNames',
     'aggregates_displayed',
+    'generation_only',
```
VIKTORVAV99, line 40: Instead of including this would it not make more sense to exclude all the zone configs that contain this?

And do any of these zones have exchanges we somehow need to account for that we might not have considered yet?

web/geo/world.geojson
VIKTORVAV99, line 1: Nice! Looks like there wasn't a conflict at all!

web/geo/types.ts
```diff
@@ -11,6 +11,7 @@ export interface ZoneConfig {
   contributors?: string[];
   disclaimer?: string;
   estimation_method?: string;
+  generation_only?: boolean;
```
VIKTORVAV99, line 14: This relates to the other comment, if we exclude the zones all together we don't need this.

web/geo/generateAggregates.ts
```diff
@@ -85,8 +85,11 @@ const generateAggregates = (fc: WorldFeatureCollection, zones: OptimizedZonesCon
     })
     .filter((zone) => zone !== null);
 
-  if (skippedZones.length > 0) {
-    for (const zone of skippedZones) {
+  const skippedZonesWithConsumption = skippedZones.filter(
+    (z) => !zones[z].generation_only
+  );
```
VIKTORVAV99, line 90: And then we might not need this either.

VIKTORVAV99 (commented):
Just some small thoughts. But if you prefer it this way I'd be happy to approve it.

web/scripts/generateZonesConfig.ts
madsnedergaard: Hmm, since we still want to allow accessing via URL I don't think we can just do that? 🤔

VIKTORVAV99: Wouldn't that be very weird if we didn't display it on the map?

Like if someone shared a screenshot or mentioned it no other users would be able to find it unless they edited the URL (which is impossible if you use the App Store versions of the app).

madsnedergaard: We have the same with old India zones: https://app.electricitymaps.com/zone/IN-MH

I really don't think that will be a big problem and it's easier for debugging to see the zone on the app. Happy to reconsider if we ever reach a situation where this becomes a problem :)

VIKTORVAV99: Didn't we delete those? 😅 🙈

madsnedergaard: Nope, not yet - too much work since internal tests are relying on it 😬

VIKTORVAV99: Ah well, let's just use your current approach then and we can revisit when we delete the old India zones (and exchanges).

VIKTORVAV99 (approved):
LGMT! 🎉

PR #6989, opened 2024-07-17
chore(deps): update to react-i18next 15

VIKTORVAV99:
## Description

I have made som upstream changes in react-i18next to reduce the code size a bit and make it slightly more performant. Coupled with #6891 this should give us a nice size reduction (it's already cutting the size by 1kb without it).

### Double check


silkeholmebonnen (approved):
LGTM

PR #6991, opened 2024-07-18
add back minimal GR-IS config for internal requirements

VIKTORVAV99:
## Description

We still need a config for this zone for some internal reasons. 

### Double check


VIKTORVAV99:
@madsnedergaard leaving it up to you about what to do here but maybe a redirect to GR would be enough?

PR #6993, opened 2024-07-19
Revert "parser validation - add function to filter out events with total production of 0 MW"

FelixDQ:
Reverts electricitymaps/electricitymaps-contrib#6533

We have now added this rule to our backend QA pipeline, so we can remove this from the parser validation.


VIKTORVAV99 (approved):
LGMT! 🎉

Kind of surprised you didn't have to change any tests, apparently there is nothing testing if this thing was working 😬.

PR #6995, opened 2024-07-19
fix(web): replace renderToString

silkeholmebonnen:
## Issue

[AVO-262](https://linear.app/electricitymaps/issue/AVO-262/find-alternative-to-rendertostring-from-react-domserver)

## Description

This PR avoids using renderToString by using a new component `Headline` instead. It reduces the bundle size with ~69 kb. 

### Double check


VIKTORVAV99 (approved):
LGMT! 🎉

I took the liberty to remove the old translations as otherwise it would have looked like this in other languages than English:
image
Now it will fall back to English instead.

Just a small tip:
When we either need to delete the other translations or use new translation keys when doing major changes to the translations or it can't fall back to English. That can cause it to render improperly, like now, or to have the wrong meaning.

silkeholmebonnen:
It is a little bit weird that the production source is in the selected language, but the rest of the sentence is in english. Do you think that is fine for now?
image

VIKTORVAV99:
> It is a little bit weird that the production source is in the selected language, but the rest of the sentence is in english. Do you think that is fine for now? image

I think that is fine for now.
That problem exist elsewhere in the code as well.

PR #6999, opened 2024-07-22
fix: remove generation only zones from ranking

VIKTORVAV99:
## Issue

We are currently showing zones in the ranking that we don't have on the map in staging.

## Description
This checks if it's a generation only zone and hides it from the ranking if it is to match the map.

### Double check


tonypls (approved):
LGTM

PR #7000, opened 2024-07-22
chore: some micro optimizations in util

VIKTORVAV99:
## Description

Adds some micro optimizations to util functions.
There is a small size benefit from this but more inportantly the code is easier to navigate and more consistant.

### Double check

### DIFF
--- a/web/src/features/panels/zone/util.ts
+++ b/web/src/features/panels/zone/util.ts
@@ -4,14 +4,14 @@ import { TimeAverages } from 'utils/constants';
 import zonesConfigJSON from '../../../../config/zones.json'; // Todo: improve how to handle json configs
 import { CombinedZonesConfig } from '../../../../geo/types';
 
-const config = zonesConfigJSON as unknown as CombinedZonesConfig;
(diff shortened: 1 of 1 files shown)


web/src/features/panels/zone/util.ts
```diff
@@ -68,32 +68,27 @@ export const getZoneDataStatus = (
   return ZoneDataStatus.NO_REAL_TIME_DATA;
 };
 
-export function getContributors(zoneId: string) {
-  return {
-    zoneContributorsIndexArray: config.zones[zoneId]?.contributors as number[],
-    contributors: config.contributors,
-  };
-}
+export const getContributors = (zoneId: string) => ({
```
tonypls, line 71: Very nitpicky but we try use regular function definitions as much as possible

VIKTORVAV99: I know but arrow functions are smaller when bundled basically because of this:
with functions:
```js
function one(){return "1"};
function two(){return "2"};
function three(){return "3"};
```
with arrow functions:
```js
const one=()=>"1", two=()=>"2", three=()=>"3";
```

tonypls (approved):
LGTM

PR #7001, opened 2024-07-22
fix: Align no data tooltip with other map tooltips

VIKTORVAV99:
## Issue
Closes: AVO-430

## Description
Aligns the design of the no data tooltip with how the other zone hover tooltips look.

### Preview
image

### Double check


web/src/components/ZoneName.tsx
```diff
@@ -5,7 +5,7 @@ import { CountryFlag } from './Flag';
 
 export function ZoneName({ zone, textStyle }: { zone: string; textStyle?: string }) {
   return (
-    <div className="flex items-center">
+    <div className="flex min-w-0 items-center">
```
VIKTORVAV99, line 8: This is so the truncation works properly. Basically the min width needs to be anything other than auto.

silkeholmebonnen (approved):
Nice work! 

This did not come from your PR but I was wondering if we could change this, since it looks a bit weird in my opinion. But maybe we should do it in a follow up PR?
Screenshot 2024-07-24 at 09 39 41

VIKTORVAV99:
> Nice work!
> 
> This did not come from your PR but I was wondering if we could change this, since it looks a bit weird in my opinion. But maybe we should do it in a follow up PR? Screenshot 2024-07-24 at 09 39 41

That should be a follow up PR IMO. I'll add a linear task for slack week.

PR #7004, opened 2024-07-23
perf: use useCallbacks for handlers and deconstruct events

VIKTORVAV99:
## Issue
The handlers are not wrapped in useCallbacks and are re-created on each render.

## Description

Wraps the handlers in useCallbacks to avoid recreating them and deconstructs the event objects to optimize the size a bit.

### Double check


tonypls (approved):
LGTM

PR #7005, opened 2024-07-24
Design Hack: Remove icons from chart titles

VIKTORVAV99:
## Issue
The icons take up too much space on mobile.

## Description

Removes the icons from the chart titles but keep the icon files themselves if we want to use them in the future.

### Double check


web/src/features/charts/ChartTitle.tsx
```diff
@@ -8,11 +8,10 @@ type Props = {
   translationKey: string;
   hasLink?: boolean;
   badgeText?: string;
-  icon?: JSX.Element;
   unit?: string;
 };
 
-export function ChartTitle({ translationKey, badgeText = undefined, icon, unit }: Props) {
```
silkeholmebonnen, line 15: If we keep the icon files, should we keep this prop as well? Or do you mean that we might use the icons elsewhere in the future?

VIKTORVAV99: We might use them elsewhere in the future. I don't think we will be using them inside the titles though  so it's fine to remove the prop but @Alportan can confirm.

silkeholmebonnen (approved):
LGTM 🎉 

I believe Alex told me the same thing about the icons in the title

web/src/features/charts/ChartTitle.tsx
Alportan: We won't be using them inside titles so all good 👌

PR #7011, opened 2024-07-24
feat: add lucide-react and change X icon

VIKTORVAV99:
## Description
Since we agreed to change to Lucide icons I have added it to our dependencies and changed the X (close) icon to ensure it's working (it looks identical as far as I can tell).

### Double check

### DIFF
--- a/web/src/features/charts/DataSources.tsx
+++ b/web/src/features/charts/DataSources.tsx
@@ -2,8 +2,8 @@ import * as Portal from '@radix-ui/react-portal';
 import { Link } from 'components/Link';
 import TooltipWrapper from 'components/tooltips/TooltipWrapper';
 import { TFunction } from 'i18next';
+import { X } from 'lucide-react';
 import { useTranslation } from 'react-i18next';
-import { HiXMark } from 'react-icons/hi2';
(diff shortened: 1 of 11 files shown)


web/src/features/charts/DataSources.tsx
```diff
@@ -95,7 +95,7 @@ function EmissionFactorTooltip({ t }: { t: TFunction<'translation', undefined> }
         {t('country-panel.emissionFactorDataSourcesTooltip')}
       </div>
       <button className="p-auto pointer-events-auto mt-2 flex h-10 w-10 items-center justify-center self-center rounded-full border bg-zinc-50 text-black shadow-md">
-        <HiXMark size="24" />
```
VIKTORVAV99, line 98: 24 is the default size.

tonypls (approved):
LGTM, the X has a slightly thicker line thickness but I think that's okay.

VIKTORVAV99:
> LGTM, the X has a slightly thicker line thickness but I think that's okay.

That is adjustable but I think we should try and keep as close to the defaults as possible so we don't end up shooting ourselves in the foot.

PR #7019, opened 2024-07-28
chore: Migrate MapMovement component test to E2E

VIKTORVAV99:
## Issue

The current component test currently rely on behaviour that are from outside the component itself, this is bad.

It is also preventing us from upgrading to Vite 5 as the external behaviour have changed (this only affects the testing).

## Description

Migrates the component test to be part of the E2E map test and makes it more robust and tests more features and components.

### Double check


tonypls (approved):
LGTM, nice little tidy up and desctructuring

PR #7020, opened 2024-07-29
WIP: Side panel design hack

VIKTORVAV99:
### Double check


madsnedergaard:
Closing this, dunno what it is, but don't think it's worth having open PRs for internal hacking around like this :)

VIKTORVAV99:
> Closing this, dunno what it is, but don't think it's worth having open PRs for internal hacking around like this :)

Was supposed to be converted to a non hack version but we always got busy with other stuff. Alex got some designs if someone is picking this up in the future.

PR #7023, opened 2024-07-30
feat: convert EUR/MWh to EUR/KWh

VIKTORVAV99:
## Issue
Closes: AVO-443
## Description

Creates a helper function to convert prices (which currently only handles EUR) and uses that to convert `EUR/MWh` to `EUR/KWh`

- [x] Add tests for the new function to ensure it's working as expected.

### Double check


web/src/features/charts/bar-breakdown/utils.ts
```diff
@@ -192,3 +193,22 @@ export const getExchangesToDisplay = (
...
+ * If no currency is provided, the parameters are returned as is.
+ */
+export const convertPrice = (
+  value?: number,
+  currency?: string,
+  unit: EnergyUnits = EnergyUnits.MEGAWATT_HOURS
+): { value?: number; currency?: string; unit: EnergyUnits } => {
+  if (currency === 'EUR') {
+    if (value) {
+      value = round(value / 1000, 4);
```
tonypls, line 209: I think this should be 2dp as a cent is the most granular unit of the currencies we use

VIKTORVAV99: I did some research and testing and the problem with 2 decimal points is that we loose a lot of information.

For example if the price is 0,56 EUR per MWh it would be 0,0006 EUR per kWh if using 4 decimals but 0.01 EUR per kWh if we used 2 decimals.

The rounding to 4 decimals causes an error of 0.6 if converted back to per MWh but if the same is done with 2 decimals the error would be 9.44, which is very bad.

Other sites that show the price per kWh shows it with 4 decimals as well.

VIKTORVAV99: @Alportan with the above in mind you to think it makes sense to have a toggle in the settings to show prices as per MWh or per kWh?

I would assume some energy experts and the like want to have the most accurate value all the time, even if it's per MWh?

Alportan: Hmm, good point! I think a toggle in the settings does make a lot of sense. Should I sketch something? 🙌

VIKTORVAV99: > Hmm, good point! I think a toggle in the settings does make a lot of sense. Should I sketch something? 🙌

If you wish! 

I'm thinking that maybe we should try and move stuff into a settings modal next cycle. Even if it's just the current stuff and then we can expand from there.

tonypls: Happy for this to go through as it is if you're happy with the 4dp

tonypls (approved):
LTGM, will leave it up to you if you want to do further changes here then tag me other wise a V2 branch works 👍

VIKTORVAV99:
I think I'll merge this as is and then I'll sync a bit with Alex on the settings modal and how we can best incorporate a setting with that.

PR #7025, opened 2024-07-31
chore: removes english translations from other languages

VIKTORVAV99:
## Issue

There was English translations in other language files.

## Description

Removes the above mentioned translations as they will fall back to English anyway. This ensures that the translations are not duplicated and will help contributors spot the missing translations.

### Double check


tonypls (approved):
LGTM

PR #7026, opened 2024-07-31
fix(NL): add reallocation from unknown to solar

FelixDQ:
Entsoe data for NL differs largely between real-time and historical. Specifically it seems like the real-time production is drastically undereported. After a delay of ~1 month, the data is updated to include a large share of unknown. It seems like the majority of this unknown production can be attributed to solar.

image

Our current parser just replaces the solar production with the one from ned.nl. This doesn't work for historical data as the solar data now occurs as both part of the unknown mix and as solar production. This change subtracts the solar production from the unknown which makes real-time and historical data more consistent. See on the graph below how this changes the consolidated data from before July 3rd

image

This solution is still not perfect as there is still a discrepancy between historical and real-time data.
* Biomass doesn't seem to appear in real-time data
* There's still a share of unknown that is unaccounted for


VIKTORVAV99:
CC @tonypls, perhaps we should ask about biomass during our meeting with NED.nl next week?

@FelixDQ have you tried comparing the remaining unknown with the difference in offshore wind reported between ENTSO-E and NED.nl?

VIKTORVAV99:
> Nope, that would be a great thing to test as well

I have some stuff to wrap up but let me know if I can be of assistance!

FelixDQ:
We'll go with this fix for now to alleviate the most critical problems and then we can investigate these further improvements afterwards

parsers/NED.py
```diff
@@ -240,10 +272,6 @@ def fetch_production_forecast(
     NED_data = format_data(json_data, logger, forecast=True)
     ENTSOE_data = _get_entsoe_forecast_data(zone_key, session, target_datetime, logger)
 
-    combined_data = ProductionBreakdownList.update_production_breakdowns(
-        production_breakdowns=ENTSOE_data,
-        new_production_breakdowns=NED_data,
-        logger=logger,
-        matching_timestamps_only=True,
-    )
+    combined_data = combine_production_data(ENTSOE_data, NED_data, logger)
```
VIKTORVAV99, line 275: By adding this to the forecasts the forecasts are now always 0 unknown as we only get forecasted wind and solar values.

This could be a problem if we decide to rely more on forecast values in the future for our estimation methods. I'd rather not pollute the database with static 0 values that are almost guaranteed to be wrong.

This could work if NED.nl also forecasts unknown (other) data but I don't think they do.

FelixDQ: Sure, I reverted to the old method for forecasts. I have no idea if this will be accurate, but we don't really use these for anything atm.

VIKTORVAV99: It should be, the only thing these are used for is the new estimation model but that is only used for Germany at the moment.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7048, opened 2024-08-01
fix: persist ranking on search

VIKTORVAV99:
## Issue

The ranking panel used the index as the ranking but this always changed when searching in meaning if you searched a single zone it would always be ranked as 1.

## Description
This PR rewrites the whole filtering and sorting logic to account for this and persist the ranking even in searches, something that it looked like it was always intended to do as there was already a ranking property that was always set to undefined.

As a bonus it will now also match using the zone key.

This should partially help with the issue described in https://github.com/electricitymaps/electricitymaps-contrib/issues/6971

### TODO:

- [x] Add tests.

#### On master
image

#### On this branch
image

### Double check


tonypls (approved):
LGTM, as nice extra feature would be that the search value doesn't get lowercased from the user perspective.

VIKTORVAV99:
> LGTM, as nice extra feature would be that the search value doesn't get lowercased from the user perspective.

I have a few things to follow up on this (like searching in English should always work no matter the language) so I can add that as well.

strassburger:
looks like solved:
the issue described in https://github.com/electricitymaps/electricitymaps-contrib/issues/6971
Thanks for fixing.
Rainer

PR #7057, opened 2024-08-02
fix: revert bump to @tailwindcss/forms

VIKTORVAV99:
## Issue
It seems like a breaking visual change was introduced in 0.5.4 

## Description
Reverts the package to v 0.5.3 for now.

### Double check


VIKTORVAV99:
Self merging this so things look pretty before the demo 🙂

PR #7058, opened 2024-08-02
refactor(web): change styling of exchange tooltips to match the other map tooltips

silkeholmebonnen:
## Issue

[AVO-431](https://linear.app/electricitymaps/issue/AVO-431/unify-exchange-tooltip)

## Description

This PR changes the styling of the exchange tooltips to match the other map tooltips

### Preview

image
image

### Double check

### DIFF
--- a/web/src/features/exchanges/ExchangeArrow.tsx
+++ b/web/src/features/exchanges/ExchangeArrow.tsx
@@ -60,7 +60,8 @@ function ExchangeArrow({
   };
 
   // Setting the top position from the arrow tooltip preventing overflowing to top.
-  let tooltipClassName = 'max-h-[256px] max-w-[512px] md:flex';
+  let tooltipClassName =
+    'max-h-[256px] max-w-[512px] md:flex rounded-2xl border-neutral-200 bg-white dark:bg-gray-900 dark:border-gray-700 dark:border';
   if (!isMobile) {
     tooltipClassName += transform.y - 76 < headerHeight ? ' top-[76px]' : ' top-[-76px]';
   }


silkeholmebonnen:
The size of exchange tooltips are not the same as the map tooltips, but I don't think that is something that we necessarily want. Let me know if you think otherwise. 

Also I just noticed this:
image
I will create a slack friendly task on linear for it.

VIKTORVAV99:
> The size of exchange tooltips are not the same as the map tooltips, but I don't think that is something that we necessarily want. Let me know if you think otherwise.
> 
> Also I just noticed this: image I will create a slack friendly task on linear for it.

I would maybe try and move down the export amount to the lower right corner (so it's on the same line as the carbon intensity) and let the zone names take up more space. That way we eliminate some weird whitespace and allow more content to be shown. But as you said that could be a separate PR.

VIKTORVAV99 (commented):
While a very big improvement I think we should try and make them the same size (at least hight) and move some elements around to make them more visual similar to (moving the zone names to the top and adding the datetime) to minimise the CLS and jumping as much as possible here:

https://github.com/user-attachments/assets/6adf6e1c-123c-4ff4-b6e1-9a5ee721d639

@Alportan what do you think?

Alportan:
> While a very big improvement I think we should try and make them the same size (at least hight) and move some elements around to make them more visual similar to (moving the zone names to the top and adding the datetime) to minimise the CLS and jumping as much as possible here:
> 
>  Skarminspelning.2024-08-02.kl.11.37.47.mov 
> @Alportan what do you think?

Hmm, good idea! I would advise doing it in an upcoming PR as I believe this would also require some design changes 🙌  

We can start with this and follow-up with the height update later, what do you think @VIKTORVAV99 @silkeholmebonnen ?

VIKTORVAV99 (approved):
LGTM!

Let's do the other potential changes in a follow up PR as suggested by Alex.

PR #7061, opened 2024-08-03
fix(US-MIDA-PJM): use appropriate timezone

### LINKED ISSUE #7018: US-MIDA-PJM production parser down
## Description

This is an automatic error report generated for USA PJM Interconnection (US-MIDA-PJM).

Issues:

## Suggestions
- Try running the parser locally using the command `poetry run test_parser US-MIDA-PJM production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/US-MIDA-PJM.html">Explore the runtime logs</a>

You can see an overview of all parser issues [here](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Parser-issues).

elisezhg:
## Issue

Closes #7018

## Description

- Use the right timezone to filter the `datetime_beginning_ept` column. It was using the UTC timezone which is 4 hours ahead, causing no data to be returned when the query was run at times when UTC was already on the next day.
- (Clean up: remove `session` default value since it's mutable)

### Preview

```zsh
❯ poetry run test_parser US-MIDA-PJM production
Parser result:
[{'datetime': datetime.datetime(2024, 8, 2, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='America/New_York')),
  'production': {'coal': 26735.0,
                 'gas': 52148.0,
                 'hydro': 317.0,
                 'nuclear': 32124.0,
                 'oil': 1534.0,
                 'solar': 6.0,
                 'unknown': 641.0,
                 'wind': 1962.0},
  'source': 'pjm.com',
  'storage': {'battery': 0.0},
  'zoneKey': 'US-MIDA-PJM'},
 ...
---------------------
took 1.13s
min returned datetime: 2024-08-02 04:00:00+00:00 UTC
max returned datetime: 2024-08-03 00:00:00+00:00 UTC  -- OK, <2h from now :) (now=2024-08-03T01:17:32+00:00 UTC)
```
### Double check


parsers/US_PJM.py
```diff
@@ -106,8 +106,6 @@ def fetch_consumption_forecast_7_days(
 
     if target_datetime:
         raise NotImplementedError("This parser is not yet able to parse past dates")
-    if not session:
-        session = Session()
```
VIKTORVAV99, line 110: This should not be removed, if anythin it's the default value that should be changed to None as using mutable defaults in python can potentially cause all kinds of issues.

VIKTORVAV99 (commented):
LGTM except for one tiny thing, could you fix that before I merge this?

VIKTORVAV99:
BTW how did you test this actually works?

To me it seems like it's this that's the problem not the timezone itself:
```python
        "datetime_beginning_ept": target_datetime.strftime("%Y-%m-%dT%H:00:00.0000000"),
```
as it's not converting the incoming datetimes to match the expected timezone?

parsers/US_PJM.py
elisezhg: Thanks, fixed for the whole file!

elisezhg:
> BTW how did you test this actually works?
> 
> To me it seems like it's this that's the problem not the timezone itself:
> 
> ```python
>         "datetime_beginning_ept": target_datetime.strftime("%Y-%m-%dT%H:00:00.0000000"),
> ```
> 
> as it's not converting the incoming datetimes to match the expected timezone?

From what I understand, the incoming datetimes are already in EPT so no need to convert them, but we were using a UTC datetime to filter them. This is causing a mismatch when you run the parser late at night, since UTC is already on the next day and we end up trying to fetch data for the next day...
At least that's the problem I ran into when running `poetry run test_parser US-MIDA-PJM production` last night! After changing the timezone, I was able to get the data so I hope it closes the issue

VIKTORVAV99:
> > BTW how did you test this actually works?
> 
> > 
> 
> > To me it seems like it's this that's the problem not the timezone itself:
> 
> > 
> 
> > ```python
> 
> >         "datetime_beginning_ept": target_datetime.strftime("%Y-%m-%dT%H:00:00.0000000"),
> 
> > ```
> 
> > 
> 
> > as it's not converting the incoming datetimes to match the expected timezone?
> 
> 
> 
> From what I understand, the incoming datetimes are already in EPT so no need to convert them, but we were using a UTC datetime to filter them. This is causing a mismatch when you run the parser late at night, since UTC is already on the next day and we end up trying to fetch data for the next day...
> 
> At least that's the problem I ran into when running `poetry run test_parser US-MIDA-PJM production` last night! After changing the timezone, I was able to get the data so I hope it closes the issue

The problem is that the backend will call the parser with UTC datetimes when refetching data and your fix only applies when the datetime is none.

elisezhg:
> > > BTW how did you test this actually works?
> > 
> > 
> > > 
> > 
> > 
> > > To me it seems like it's this that's the problem not the timezone itself:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > ```
> > >     "datetime_beginning_ept": target_datetime.strftime("%Y-%m-%dT%H:00:00.0000000"),
> > > ```
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > as it's not converting the incoming datetimes to match the expected timezone?
> > 
> > 
> > From what I understand, the incoming datetimes are already in EPT so no need to convert them, but we were using a UTC datetime to filter them. This is causing a mismatch when you run the parser late at night, since UTC is already on the next day and we end up trying to fetch data for the next day...
> > At least that's the problem I ran into when running `poetry run test_parser US-MIDA-PJM production` last night! After changing the timezone, I was able to get the data so I hope it closes the issue
> 
> The problem is that the backend will call the parser with UTC datetimes when refetching data and your fix only applies when the datetime is none.

Thanks for the clarification, the fix is done!

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing this and sorry for the delay!

PR #7064, opened 2024-08-05
fix(SG): update OCR configuration

FelixDQ:
The image from the source was updated, so we have to change the configuration to make it work again. This fixes the preprocessing and cropping.


VIKTORVAV99 (approved):
LGTM! 🎉

Feel free to create a new release to get this out right away. Any changes in contrib are fine to release as is.

PR #7067, opened 2024-08-06
chore(web): fix feedback card wording and font weight

tonypls:
## Issue
Feedback card doesn't match the design
image

## Description
This PR updates the wording and styling of the in app satisfaction survey to better match the design.

We should try make this a bit more re-usable next time we add a new feature feedback survey.


Alportan (approved):
LGTM 🎉 Thanks for picking this up so fast Tony! 🫶

web/src/locales/en.json
```diff
@@ -124,14 +124,16 @@
       "primary-question": "The description of the data estimation was easy to understand:"
     },
 
-    "primary-question": "How satisfied are you with your experience using the Electricity Maps app?",
+    "primary-question": "How would you rate your overall satisfaction with the Electricity Maps app?",
```
VIKTORVAV99, line 127: Is this added to any other language that we need to adjust as well?

tonypls: Yes it's also in 3 other languages which translate to "How satisfied are you with your experience using the Electricity Maps app?" I think it's a finer detail though, here's the Swedish: "Hur nöjd är du med din upplevelse av att använda Electricity Maps-appen?" Does it still capture the general question?

VIKTORVAV99: Yeah I think that would still work. I can add the Swedish translations for the new strings if you wish?

tonypls: yeah sounds good!

PR #7069, opened 2024-08-09
chore: clean up unused prop in BarBreakdownChart

VIKTORVAV99:
## Issue

While looking into AVO-361 I found a prop that was not used but still drilled through multiple components.

## Description
Removes said prop from these components.

### Double check


tonypls (approved):
LGTM

PR #7071, opened 2024-08-12
feat: Add button to our subreddit

VIKTORVAV99:
## Description
Adds a button in the ranking panel accordion to our official subreddit at https://reddit.com/r/electricitymaps

### Preview

image

### Double check


Alportan (approved):
LGTM 🎉

It does stand out the most because of the colour, but I don't think it's an issue 👌

VIKTORVAV99:
> LGTM 🎉
> 
> 
> 
> It does stand out the most because of the colour, but I don't think it's an issue 👌 

Yeah I don't think there is much we can do about that, on light mode GitHub stands out the most for example.

PR #7072, opened 2024-08-12
chore: Improve test coverage

VIKTORVAV99:
## Issue

Our test coverage is pretty lacking.

## Description
This PR adds 101 new unit tests for utility functions and marks some tests as todos.

It also makes some of the functions a bit more robust based on the test outcomes to handle some edge cases.

This is mostly low hanging fruit that is somewhat easy to test (pure functions and no translation dependency).

### Double check


web/src/features/panels/zone/__snapshots__/util.test.ts.snap
VIKTORVAV99, line 1: This output does not seem optimal we should probably only return the contributors for the specific zone.

Will take a look at it in a separate PR though.

web/src/utils/helpers.test.ts
```diff
@@ -48,7 +52,7 @@ describe('getProductionCo2Intensity', () => {
 
   it('returns the correct value when the type is battery storage', () => {
     const actual = getProductionCo2Intensity('battery storage', zoneDetailMock);
-    expect(actual).to.eq(155.11);
+    expect(actual).to.eq(0);
```
VIKTORVAV99, line 55: This produced false values before due to how the mock data was structured.

tonypls (approved):
Also LGTM

VIKTORVAV99:
Woohoo! 🥳

PR #7073, opened 2024-08-13
feat(web): US Specific survey

tonypls:
## Issue
We want to understand our US user more so we can build great features 🦅

AVO-425

## Description

This PR adds a new in app survey that targets US specific users. The targeting is currently taken from the userLocation on map load and will not be displayed if users load directly to a zone. 

We could consider adding a trigger when any US zone is clicked that would widen the funnel of users we're surveying. 

## Testing instructions
Add the `ff` query param i.e. `app.electricitymaps.com/map?ff` to open the feature flags manager and toggle on the survey then you'll also need a VPN that locations you inside the US. The survey should appear on map load when not viewing a zone. If testing is painful add a comment and I'll make it easier 👍


web/src/components/MapOverlays.tsx
```diff
@@ -1,26 +1,35 @@
...
+} from 'utils/state/atoms';
 import { useIsBiggerThanMobile } from 'utils/styling';
 
 import SurveyCard from './app-survey/SurveyCard';
 import LegendContainer from './legend/LegendContainer';
 
 export default function MapOverlays() {
   const hasSeenSurveyCard = useAtomValue(hasSeenSurveyCardAtom);
+  const hasSeenUsSurveyCard = useAtomValue(hasSeenUsSurveyCardAtom);
+  const [userLocation] = useAtom(userLocationAtom);
```
VIKTORVAV99, line 18: Any reason why this is not using useAtomValue like the others?

web/src/components/app-survey/FeedbackCard.tsx
```diff
@@ -39,20 +51,26 @@ export default function FeedbackCard({
...
   const isMobile = useIsMobile();
 
   const handleClose = () => {
     setIsClosed(true);
     if (surveyReference === 'Map Survey') {
       setHasSeenSurveyCard(true);
     }
+    if (surveyReference === 'US Survey') {
+      setHasSeenUsSurveyCard(true);
+    }
```
VIKTORVAV99, line 73: These are mutually exclusive right?

Should we use if else statements instead?

web/src/utils/state/atoms.ts
```diff
@@ -46,10 +46,10 @@ export const isWindLayerEnabledAtom = atom(
     get(selectedDatetimeIndexAtom).index === 24
 );
 
-export const solarLayerLoadingAtom = atom(false);
-export const windLayerLoadingAtom = atom(false);
+export const solarLayerLoadingAtom = atom<boolean>(false);
+export const windLayerLoadingAtom = atom<boolean>(false);
```
VIKTORVAV99, line 50: Just curious why the type was added here?

web/src/utils/state/atoms.ts
```diff
@@ -67,16 +67,18 @@ export const feedbackCardCollapsedNumberAtom = atom(0);
...
 
-export const dataSourcesCollapsedEmission = atom(true);
+export const dataSourcesCollapsedEmission = atom<boolean>(true);
 
-export const userLocationAtom = atom(undefined);
+export const userLocationAtom = atom<string | undefined>(undefined);
 
 export const hasSeenSurveyCardAtom = atomWithStorage('hasSeenSurveyCard', false);
 
+export const hasSeenUsSurveyCardAtom = atomWithStorage('hasSeenUsSurveyCard', false);
```
VIKTORVAV99, line 80: But not here?

VIKTORVAV99 (commented):
Just some small questions but otherwise it looks good.

web/src/components/MapOverlays.tsx
tonypls: Prettier complained when I tried that originally but now it's typed I think it's okay, updated it now.

web/src/utils/state/atoms.ts
tonypls: I was adding a type below where it couldn't be inferred and thought it was also kinda nice to have them explicit where possible.

web/src/utils/state/atoms.ts
tonypls: The cognitive load required was higher 🤷

VIKTORVAV99 (approved):
LGTM! 🎉

(but please removed the unused import before merging)

web/src/components/app-survey/FeedbackCard.tsx
tonypls: Yeah they are mutually exclusive. I think it's a touch easier to read / extend without the else in there. In the case there is an overlap it could be nice that it closes them both

PR #7076, opened 2024-08-14
switch back to useQuery instead of useSuspenseQuery

VIKTORVAV99:
## Issue
The switch from useQuery to useSuspenseQuery was intended to alleviate some CLS and allow components to load more gracefully. This had the unintended consequence of making the bottom sheet disappear when a zone is loading as it would suspend the whole component without a fallback.

## Description
Reverts that particular change to bring back the old behaviour.

### Double check


VIKTORVAV99:
@Alportan can you double check that this fixes the issue as well?

tonypls:
Works nicely on my machine but It needs some more null checks to pass the linting

tonypls (approved):
LGTM! What do you think @VIKTORVAV99 ?

PR #7077, opened 2024-08-14
feat: Update ENTSO-E parser with Energy Storage

VIKTORVAV99:
## Description

Updates the ENTSO-E parser with the PsrType mapping for Energy Storage to our battery storage mode.

This is an addition ENTSO-E made yesterday (2024-08-13) so there are not any zones that are reporting values for it yet. Which means that so far it's untested and therefore this PR is in a draft until there is a zone we can verify it's working as expected on.

Closes: GMM-516

### Double check


VIKTORVAV99:
It works!

PR #7079, opened 2024-08-15
chore: use Lucide icons for Accordion and simplify DOM

VIKTORVAV99:
## Issue

We are still using old Font Awesome icons here and the DOM could be simplified a bit to reduce redundant div elements and logic checks.

Part of: AVO-439

## Description

Changes the icons to Lucide versions, they have a slightly higher line width but otherwise they are identical.

The other code changes have been tested to work the same as the original code and logic.

### Double check


silkeholmebonnen (approved):
LGTM 🪴

PR #7080, opened 2024-08-15
chore: use Lucide icons for ZoneHeaderTitle and simplify DOM

VIKTORVAV99:
## Issue

We where not using Lucide icons and the DOM was bloated.

Part of: AVO-439

## Description

Switches to use Lucide icons and simplifies the the DOM structure to reduce the amount of rendered elements and size of the code.

This reduces the total app size from `4071.75 KiB` to `4070.90 KiB` spread across the index file, style.css and the ZoneDetails bundles.

### Preview

> [!NOTE]
> Previously the `i` was not an icon and therefore had some unexpected behaviour on hover.
> It however means we can't style it exactly the same way.

#### On master:
image

#### On this branch:
image

### Double check

### DIFF
--- a/web/src/features/panels/zone/ZoneHeaderTitle.tsx
+++ b/web/src/features/panels/zone/ZoneHeaderTitle.tsx
@@ -3,7 +3,7 @@ import { TimeDisplay } from 'components/TimeDisplay';
 import TooltipWrapper from 'components/tooltips/TooltipWrapper';
 import { mapMovingAtom } from 'features/map/mapAtoms';
 import { useSetAtom } from 'jotai';
-import { HiArrowLeft } from 'react-icons/hi2';
+import { ArrowLeft, Info } from 'lucide-react';
 import { Link } from 'react-router-dom';
 import { getCountryName, getFullZoneName, getZoneName } from 'translation/translation';
 import { createToWithState } from 'utils/helpers';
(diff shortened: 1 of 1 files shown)


web/src/features/panels/zone/ZoneHeaderTitle.tsx
```diff
@@ -31,49 +31,41 @@ export default function ZoneHeaderTitle({ zoneId }: ZoneHeaderTitleProps) {
...
+            <h1 className="grow truncate" data-test-id="zone-name">
+              {zoneName}
+            </h1>
+          </TooltipWrapper>
+          {showCountryPill && (
+            <div className="flex w-auto items-center rounded-full bg-gray-200 px-2 py-0.5 text-sm dark:bg-gray-800/80">
+              <p className="w-full truncate">{countryName ?? zoneId}</p>
             </div>
-          </div>
+          )}
```
silkeholmebonnen, line 63: I think it makes more sense to keep the pill left aligned as it was before.

Before:
image

Now:
image

silkeholmebonnen (commented):
Just one comment:)

web/src/features/panels/zone/ZoneHeaderTitle.tsx
VIKTORVAV99: I can change it back but my thinking was that this reduces CLS when switching between US zones as it keeps the USA badge in the same place.

Do you think that it's important that it's left aligned?

silkeholmebonnen: I don't think it is that important but I think it looks a bit strange in a case like this:

image

I was also wondering if it was intentional to remove padding from the pill? Maybe it's just me being used to how it was before but I think the pill could use a bit more padding.

VIKTORVAV99: > I don't think it is that important but I think it looks a bit strange in a case like this:
> 
> image
> 
> I was also wondering if it was intentional to remove padding from the pill? Maybe it's just me being used to how it was before but I think the pill could use a bit more padding.

I didn't change the padding on the pill 👀 only replaced the left margin with a gap for the entire group.

But I can change back the alignment and then we can reevaluate that later, the custom pill should probably be changed to use the Badge component anyhow.

silkeholmebonnen (approved):
LGTM 😄

PR #7081, opened 2024-08-15
chore: use Lucide icons for MapControls and reuse logic

VIKTORVAV99:
## Issue

We still used old icons here, and we duplicated some logic.

Part of: AVO-439

## Description

Switches the icons to Lucide versions and refractors the atom logic a little so it can be reused in more places.

This reduces the bundle size from `4071.77 KiB` to `4070.87 KiB`.

### Double check

### DIFF
--- a/web/src/features/map-controls/LanguageSelector.tsx
+++ b/web/src/features/map-controls/LanguageSelector.tsx
@@ -1,7 +1,7 @@
 import { Button } from 'components/Button';
+import { Languages } from 'lucide-react';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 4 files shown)


silkeholmebonnen:
The icon disappears when I turn solar and wind on:
Screenshot 2024-08-19 at 15 15 27

VIKTORVAV99:
> The icon disappears when I turn solar and wind on:
> 
> Screenshot 2024-08-19 at 15 15 27
> 
> 

That's not good...

I'll look into it.

silkeholmebonnen (approved):
LGTM 🌷

PR #7085, opened 2024-08-16
chore: use Lucide icons for LeftPanel and FAQ

VIKTORVAV99:
## Issue

We were still using non-Lucide icons here.

Part of: AVO-439

## Description

Switches the icons to Lucide.

### Double check


silkeholmebonnen:
I would prefer if the crevron icon more in the center here
image

silkeholmebonnen (approved):
LGTM. One small thing is that the chevron was grey before on light mode but now it is black. You can decide yourself if this is something that you want to change or if you think it looks better in black 😎 

Now:
image

Before:
image

VIKTORVAV99:
> LGTM. One small thing is that the chevron was grey before on light mode but now it is black. You can decide yourself if this is something that you want to change or if you think it looks better in black 😎
> 
> Now: image
> 
> Before: image

I think we'll leave it as is now, I have been working on a redesign of the left panel and that button so this will probably change as soon as I have time to wrap that up.

PR #7086, opened 2024-08-16
chore: use Lucide icons in DataSources and simplify DOM

VIKTORVAV99:
## Issue

We were not using Lucide icons for all of the icons here.

Part of: AVO-439

## Description

Switches to use Lucide icons.

### Double check


web/src/features/charts/DataSources.tsx
```diff
@@ -90,13 +85,11 @@ export function DataSources({
...
   return (
-    <Portal.Root className="pointer-events-none absolute left-0 top-0 z-50 flex h-full w-full flex-col content-center items-center justify-center bg-black/20 pb-40">
+    <Portal.Root className="pointer-events-none absolute left-0 top-0 z-50 flex h-full w-full flex-col content-center items-center justify-center gap-2 bg-black/20 pb-40">
       <div className="dark:border-1 relative mx-6 h-auto min-w-64 rounded-xl border bg-zinc-50 p-4 text-left text-sm opacity-80 shadow-md dark:border-gray-700 dark:bg-gray-900">
         {t('country-panel.emissionFactorDataSourcesTooltip')}
       </div>
-      <button className="p-auto pointer-events-auto mt-2 flex h-10 w-10 items-center justify-center self-center rounded-full border bg-zinc-50 text-black shadow-md">
-        <X />
-      </button>
+      <Button icon={<X />} type="secondary" backgroundClasses="pointer-events-auto" />
```
VIKTORVAV99, line 92: @Alportan I know we have discussed this button before (the white button) but I decided to switch it to use the button component here and therefore it will follow the theme.

I think it should be fine as it has an outline and it will help with keeping things uniform as other tooltips don't have this white icon.

I will also switch other tooltips to use the button component so it's uniform in both style and functionality.

web/src/features/charts/DataSources.tsx
```diff
@@ -55,12 +55,7 @@ export function DataSources({
             side="bottom"
             isMobile={isMobile}
           >
-            <div>
-              <IoInformationCircleOutline
-                className="text-emerald-800 dark:text-emerald-500"
-                size={20}
-              />
-            </div>
+            <Info className="text-emerald-800 dark:text-emerald-500" size={20} />
```
silkeholmebonnen, line 58: I think the icon is a bit big here. I think either 16 or 18 is better

silkeholmebonnen (commented):
Small detail but looks good

web/src/features/charts/DataSources.tsx
VIKTORVAV99: I think 16 works but then we need to center align these otherwise it looks very strange, would that work?

silkeholmebonnen (approved):
LGTM 🐟

PR #7087, opened 2024-08-18
chore: use Lucide icons in Header and simplify DOM

VIKTORVAV99:
## Issue

We were not using Lucide icons here and the DOM was bloated.

Part of: AVO-439

## Description

Switches to Lucide icons and simplifies the DOM, the colour used is the same colour that will be used in other places (such as info icons in some places).

#### On master:
image
image

#### On this branch:
image
image

### Double check


silkeholmebonnen (approved):
LGTM 🌟

PR #7088, opened 2024-08-19
fix(parser): use proxy for US-MIDA-PJM

FelixDQ:
## Issue
fixes GMM-137

The parser was working locally but was broken in production. It seems like we have been blocked. This changes the parser to use the proxy. I needed to do a few more changes as it for some reason started returning the data in a slightly different format (and gzip'd).


VIKTORVAV99 (approved):
Tested locally and it LGTM! 🎉

PR #7089, opened 2024-08-19
fix(parsers):India parser updated for new data format

annaib00:
## Issue
No India renewable data due to new data format from 1/7/2024.

## Description
From 1/7/2024 a column was removed from IN data files. Updated code to handle this format difference. 

### Double check


FelixDQ (approved):
Great!

PR #7090, opened 2024-08-19
chore: use Lucide icons in ToggleButton and clean up styles

VIKTORVAV99:
## Issue
We were just using a `i` as an icon.

Closes: AVO-440

## Description

Switches to use an actual icon and cleans up some of the styles.

#### On master:
image
image

#### In this branch:
image
image

### Double check


web/src/components/ToggleButton.tsx
```diff
@@ -79,16 +80,16 @@ export default function ToggleButton({
                 role="button"
                 tabIndex={0}
                 className={twMerge(
-                  'ml-2 inline-flex h-[29px] w-[29px] select-none justify-center self-center rounded-full bg-white shadow dark:border dark:border-gray-500/80 dark:bg-gray-600/50',
+                  'inline-flex select-none items-center justify-center',
                   isToolTipOpen && 'pointer-events-none'
                 )}
               >
-                <div className="self-center text-xs font-bold">i</div>
+                <Info className="text-gray-500" />
```
silkeholmebonnen, line 87: I think the icon should be a bit bigger
```suggestion
                <Info className="text-gray-500" size={30} />
```

VIKTORVAV99: I considered it but then it felt too big compared to the other info icons (the one in the zone header in particular) but we can try bumping it a bit and see if you wish?

VIKTORVAV99: Now it looks like this:
image


Not a huge fan of it as now the icons look miss aligned and too big compared to the rest?

silkeholmebonnen: I prefer it as it is right now as it aligns with the outer circle. If you want to we can probably get alex take on it tomorrow ?

VIKTORVAV99: Sounds good, there is no hurry here so it's better we take our time with it.

VIKTORVAV99: I just saw Alex had made a prototype design for this so I'll go ahead and implement that: 
image

Should fix both our concerns I think.

VIKTORVAV99:
@Alportan could you double check this. It kind of pops a little too much on darkmode but the gray-500 don't have enough contrast.

Alportan:
> @Alportan could you double check this. It kind of pops a little too much on darkmode but the gray-500 don't have enough contrast.

Hmm, I agree. It does stand out too much. 
image

I would suggest the following icon colours:
🌝 Light mode: `gray-300`
🌚 Dark mode: `neutral-500`

Alportan (approved):
Looks good on my side! ✨

We might revise it later when we make a decision on the elevation colours for backgrounds (the blur effect) to keep the colour consistent with the info icon on the side panel. 🙌

VIKTORVAV99:
> Looks good on my side! ✨
> 
> We might revise it later when we make a decision on the elevation colours for backgrounds (the blur effect) to keep the colour consistent with the info icon on the side panel. 🙌

Sounds good, and revising later is not a problem, in fact I'd be happy to do it so we have a consistent and unified design!

VIKTORVAV99:
@silkeholmebonnen does this look good to you as well?

silkeholmebonnen (approved):
LGTM

Issue #7091, opened 2024-08-20
Better translations


phiphou:
Working on a french translation, I came to an issue.

The key "comes from" should give "vient de".

But the full sentence "... comes from solar" shoud give "vient DU solaire".

And if it comes from wind, it would give "vient DE L'élien".

And for geothermy, it would give "vient DE LA géothermie".

I finaly replaced the wording "comes from" by a translation for "is generated by".

It allows a better translation, "...comes from solar" becomes "...est générée par solaire".

But that's not perfect, the good translation should be "...est générée par DU solaire", or for wind "...est générée par DE L'éolien".

One solution would be to add a different translation keys for each energy sources, like this : 

```json
{
  "comesFrom_wind": "est générée par de l'",
  "comesFrom_solar": "est générée par du",
  "comesFrom_geothermy": "est générée par de la"
}
```

Any french speaker can give us his views on this point ?

It also may touch langages like german (der/die/das)...

VIKTORVAV99:
@tonypls what is your take on this?

It means we would have to revise the logic on the tooltips a bit but it's doable.

phiphou:
Please let me know when you dive in this ;)

phiphou:
Forgot to mention the problem is the same with keys "electricityStoredUsing" and "emissionsStoredUsing".

"...is stored using wind" should give "...est stockée en utilisant [de l' / l'] éolien", and not "...est stockée en utilisant éolien".

"...is stored using solar" should give "...est stockée en utilisant [le / du] solaire", and not "...est stockée en utilisant solaire".

Use the "via" word may help, we would have "...est stockée via éolien". That's not perfect but probably better.

It would be usefull to have feedback on this from french speaker...

VIKTORVAV99:
> Forgot to mention the problem is the same with keys "electricityStoredUsing" and "emissionsStoredUsing".
> 
> "...is stored using wind" should give "...est stockée en utilisant [de l' / l'] éolien", and not "...est stockée en utilisant éolien".
> 
> "...is stored using solar" should give "...est stockée en utilisant [le / du] solaire", and not "...est stockée en utilisant solaire".
> 
> Use the "via" word may help, we would have "...est stockée via éolien". That's not perfect but probably better.
> 
> It would be usefull to have feedback on this from french speakers...

I don't know French but how would this work with hydro and battery. Would it be the same for those? (those are the only storage modes we support at the moment).

phiphou:
[...]

PR #7092, opened 2024-08-20
Add some french translations

phiphou:
Add some french translations.

But this new wording made me think about a better way to translate things for language using words like "l'/le/la" or "der/die/das".

I've opened an issue for that. #7091


VIKTORVAV99 (approved):
LGMT! 🎉

Thanks for adding these and opening the issue for further discussion!

PR #7094, opened 2024-08-20
French translations

phiphou:
Add some french translations.

I discussed #7091 with french lovers of the project via Twitter DMs, and we came to a better solution.

Using the "via" word offers a better translation than the "par" word.

So this PR translate "comes from solar" by "est générée via solaire".

It don't totaly fix it, the "multiple keys for each energy source" is still the best in my opinion.


VIKTORVAV99 (approved):
LGTM!

Cool that you where able to iterate it that quickly! 🚀

phiphou:
In french, we say "quand c'est fait c'est plus à faire", wich means "when it's done, you don't have to do it anymore" ;)

PR #7095, opened 2024-08-20
chore: bump vite to v5 and vitest to v2

VIKTORVAV99:
## Issue

We are using a outdated version of Vite that don't benefit from the same optimizations as Vite 5 has.

## Description
Bumps Vite to version 5 and Vitest to version 2 along with updates to their respective extensions. This will allow us to produce more efficient bundles and should help with the performance of the app, especially on first visit as we will be making fewer http requests.

The bump to vitest should have no impact on our testing but ensures we get the latest bugfixes and performance enhancements.

### Double check


madsnedergaard (commented):
Sorry for the delay here! Did you verify that none of the breaking changes affects us? :) 
https://vitejs.dev/guide/migration

VIKTORVAV99:
> Sorry for the delay here! Did you verify that none of the breaking changes affects us? :) 
> 
> https://vitejs.dev/guide/migration

There was one breaking change to a test that prevented this in the past but I have taken care of that in another PR already. Tested building and running locally as well and I couldn't spot any issues.

madsnedergaard (approved):
Okay great, let's do it then :)

PR #7096, opened 2024-08-21
fix(parsers):Change-some-interconnector-direction

annaib00:
## Issue
Interconnector wrong direction GB-IE and GB-NL

## Description
Change interconnector direction for GB-IE and GB-NL. Issue due to fetching data from Elexon which is given wrt GB, however due to alphabetical order we have the exchange as "GB->IE/NL", which is wrt IE/NL. 

### Double check


FelixDQ (approved):
🙌

PR #7099, opened 2024-08-22
feat(web): Deep linking prototype

tonypls:
## Issue
We can't see historic hourly data in the app. I wanted to tweet hourly charts from a few days ago but this wasn't possible.

## Description
This draft PR allows the frontend to send state and zone requests with startDate and endDate query params from the url as follows:

`http://localhost:5173/zone/NZ?startDate=2024-08-07T10:00:00Z&endDate=2024-08-08T10:00:00Z`

This requires the backend PR to be running and will hopefully serve as some inspiration for us when we do more work to tackle deep linking in the future 🚀


tonypls:
Closing as we're implementing this in a better way. #7397

PR #7101, opened 2024-08-22
chore: use Lucide icons instead of SVG backgrounds and simplify DOM

VIKTORVAV99:
## Issue

We were using SVG background instead of Icon components.

## Description

Removes the SVG backgrounds and replaces them with Lucide icons. It also further simplifies the DOM and styling of the Accordion component to fix a title overflow issue in some languages.

### Double check


VIKTORVAV99:
Waiting for https://github.com/lucide-icons/lucide/pull/2372 until I can finalize this with the proper icons.

VIKTORVAV99:
@Alportan this is the PR that uses the new Lucide icon you added, however the Figma seems to miss the replacement icon for the aggregated card (the one on top if you switch to daily, monthly or yearly) so I just picked one of the graph icons without the axis.

Could you double check this one?

Alportan:
> @Alportan this is the PR that uses the new Lucide icon you added, however the Figma seems to miss the replacement icon for the aggregated card ... so I just picked one of the graph icons without the axis.

Good catch, I think the graph icon looks great! 🙌 


> 
> Could you double check this one?

I have a few comments, that now, with the new icons, the icon gets really close to the card title. What's the current padding between the icon and the title? Maybe we can increase it a tiny bit.

image

Additionally, can we use the same icon as we do on the card on the `Preliminary` badge?

image

image

VIKTORVAV99:
> > @Alportan this is the PR that uses the new Lucide icon you added, however the Figma seems to miss the replacement icon for the aggregated card ... so I just picked one of the graph icons without the axis.
> 
> Good catch, I think the graph icon looks great! 🙌
> 
> > Could you double check this one?
> 
> I have a few comments, that now, with the new icons, the icon gets really close to the card title. ~What's the current padding between the icon and the title?~ I realised I was the culprit adding it in Figma 🙈 Let's increase it midway to 6px.
> 
> image
> Additionally, can we use the same icon as we do on the card on the `Preliminary` badge?
> 
> image image

As discussed in Slack I tweaked the style changes.

For fixing the preliminary I'll do that in a follow up PR (It's already in the works) to keep the scope of this PR mostly to the icons themselves and not so much the logic around them.

Alportan (approved):
Looks good! 🙌

PR #7103, opened 2024-08-23
chore: remove AnnouncementModal

VIKTORVAV99:
## Issue

This modal was only meant to be temporary and since it's not longer shown we should remove it.

## Description

Removes the announcement modal from the app along with the assets used for it.

### Double check


tonypls (approved):
LGTM

PR #7104, opened 2024-08-23
fix: Fix rules of react violation in Accordion

VIKTORVAV99:
## Description

Fixes the rules of react by extracting the state to the parent component instead. This allows us to use both Atoms and `useState` which should allow for more flexibility than my other PR.

Best part of all is that there is no changes needed to the tests which (hopefully) means the behaviour is identical even though its highly simplified.

Closes #6949 

### Double check


web/src/utils/state/atoms.ts
```diff
@@ -65,11 +65,11 @@ export const feedbackCardCollapsedNumberAtom = atom(0);
...
 export const colorblindModeAtom = atomWithStorage('colorblindModeEnabled', false);
 
-export const dataSourcesCollapsedBarBreakdown = atom<boolean>(true);
+export const dataSourcesCollapsedBarBreakdownAtom = atom<boolean>(true);
 
-export const dataSourcesCollapsedBreakdown = atom<boolean>(true);
+export const dataSourcesCollapsedBreakdownAtom = atom<boolean>(true);
 
-export const dataSourcesCollapsedEmission = atom<boolean>(true);
+export const dataSourcesCollapsedEmissionAtom = atom<boolean>(true);
```
VIKTORVAV99, line 72: This is just so it aligns with the rest of our naming scheme.

web/src/features/charts/BreakdownChart.tsx
```diff
@@ -123,7 +126,8 @@ function BreakdownChart({
             }}
             title={t('data-sources.title')}
             className="text-md"
-            isCollapsedAtom={dataSourcesCollapsedBreakdown}
+            isCollapsed={dataSourcesCollapsedBreakdown}
+            setState={setDataSourcesCollapsedBreakdown}
```
VIKTORVAV99, line 130: The fact that these always end up the same length is very satisfying.

madsnedergaard: length

web/src/features/panels/zone/MethodologyCard.tsx
```diff
@@ -3,16 +3,20 @@ import { Link } from 'components/Link';
 import { RoundedCard } from 'features/charts/RoundedCard';
 import { t } from 'i18next';
 import { EmapsIcon } from 'icons/emapsIcon';
+import { useState } from 'react';
 import trackEvent from 'utils/analytics';
 
 export default function MethodologyCard() {
+  const [accordionState, setAccordionState] = useState(true);
```
madsnedergaard, line 10: The naming here was a bit confusing to me, I thought it would be open by default now - maybe worth renaming props to be consistent?

```suggestion
  const [isCollapsed, setIsCollapsed] = useState(true);
```

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -168,13 +168,14 @@ function BaseCard({
     >
       <Accordion
         onClick={() => handleToggleCollapse()}
```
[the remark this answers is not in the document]
VIKTORVAV99: It does a little more like tracking events and stuff but maybe it would be enough to remove the state management from the function?

madsnedergaard (approved):
LGTM! :)

PR #7107, opened 2024-08-26
feat(app): enrich sentry events

VIKTORVAV99:
## Issue

We are currently lacking some context on why some issues are happening and how many people they affect.

We also can't tell if the issues are coming from the web or capacitor versions of the app in an easy way.

## Description
This PR does a few small things to help with the above issues and just to make our sentry information a bit more useful.

- It creates and sets a UUID as the user, this ensures we get a user count for each app install (not the actual user count if they use more than one device).

- It add tags to check which language the app is in and which language is preferred by the browser to try and correlate it with the pesky insert before/after node issues we are seeing.

- It adds a DSN specific for the Capacitor app so we can collect these issues separately, this will also allow us to integrate it with the fast lane plugin later on.

- It changes the version name in Sentry's to match Sentry's idea of server, which means including a unique package identifier (in this case `app@`), this will allow for automatic regression flagging.

- It changes the second part of the server string to point to a git tag as per Sentry recommendations.

- It sets the commit options in the Vite plugin to auto so we associate commits with each release, so we can see which files were changed between releases.

> [!NOTE]
> Once this PR is merged I will do a few rapid releases under the week to ensure things are working as expected.

### Double check


web/src/translation/i18n.ts
```diff
@@ -22,14 +23,16 @@
...
+i18n.on('languageChanged', (lng: keyof typeof localeToFacebookLocale) => {
   document.documentElement.setAttribute('lang', lng);
   // TODO: Use react-helmet to manage meta tags
   document.title = `Electricity Maps | ${t('misc.maintitle')}`;
   // Optional chaining added to ensure jsdom works
   document
     .querySelector('meta[property="og:locale"]')
     ?.setAttribute('content', localeToFacebookLocale[lng || 'en']);
+
+  Sentry.setTag('app_locale', lng);
```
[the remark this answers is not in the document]
VIKTORVAV99: Hmm. This is how Sentry recommends doing it in their docs. https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/tags/

VIKTORVAV99:
Not sure what is going on with the Earthly ESLint test, running this locally show no errors?

Would you guys be able to double check?

tonypls:
> Not sure what is going on with the Earthly ESLint test, running this locally show no errors?
> 
> Would you guys be able to double check?

Tried pushing a fix, I'm not sure why it was failing but prettier complained locally 🤞

VIKTORVAV99:
> > Not sure what is going on with the Earthly ESLint test, running this locally show no errors?
> > Would you guys be able to double check?
> 
> Tried pushing a fix, I'm not sure why it was failing but prettier complained locally 🤞

Sounds good. Hopefully it works!

(I don't actually know why the docs always import the whole of Sentry).

tonypls (approved):
LGTM!

web/src/main.tsx
```diff
@@ -13,14 +14,22 @@ import { BrowserRouter } from 'react-router-dom';
...
 if (isProduction) {
   Sentry.init({
-    dsn: 'https://bbe4fb6e5b3c4b96a1df95145a91e744@o192958.ingest.sentry.io/4504366922989568', //We should create a capacitor project in Sentry for the mobile app
+    dsn: Capacitor.isNativePlatform()
+      ? 'https://dfa9d3f487a738bcc1abc9329a5877c6@o192958.ingest.us.sentry.io/4507825555767296' // Capacitor DSN
+      : 'https://bbe4fb6e5b3c4b96a1df95145a91e744@o192958.ingest.us.sentry.io/4504366922989568', // Web DSN
     tracesSampleRate: 0, // Disables tracing completely as we don't use it and sends a lot of data
+    initialScope: (scope) => {
+      scope.setUser({ app_uuid: getSentryUuid() }); // Set the user context with a random UUID for Sentry so we can correlate errors with users anonymously
+      scope.setTag('browser_locale', window.navigator.language); // Set the language tag for Sentry to correlate errors with the user's language
```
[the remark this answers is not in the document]
tonypls: Could be nice to keep it consistent 👍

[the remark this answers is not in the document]
VIKTORVAV99: I'll update them 👍🏻

VIKTORVAV99:
@madsnedergaard do you want to take a look as well or would it be fine if I merged this?

web/src/utils/getSentryUuid.ts
```diff
@@ -0,0 +1,12 @@
+export const getSentryUuid = () => {
```
madsnedergaard, line 1: Hmm, I think this is a borderline GDPR problem, as we now have an "identifier" for a specific user.. But as long as we're not collecting more data it's probably okay, just want to make sure we keep it in mind going forward :)

VIKTORVAV99: Is it really though, it's completely anonymous and random?

We need something unique otherwise we won't be able to get a user count to correlate errors with.

madsnedergaard (approved):
LGTM! 
It would be nice to not store the UUID if the user is blocking Sentry with their adblocker, but I couldn't figure out how to test that - so I'm fine leaving it like this :)

web/src/utils/getSentryUuid.ts
madsnedergaard: As far as I understand, it's not actually anonymous but pseudonymised data, since the identifier is for the same person and we keep adding data on this user over time (for each crash report).

> You UUID is an indirect identifier that allows you to link all the actions of a single user together. You don't know the "name" of you user anymore, but with all that info you can single out one specific user and still analyse their activity: you are in a case of identifiable user.

With that said I still don't think it's a biggie and I personally think the benefits win here, just want to make sure we're aware of the "consequences" :)

VIKTORVAV99: That makes sense. Thanks for the additional context.

PR #7108, opened 2024-08-26
fix(translation): Remove hardcoded string in favour of translation strings

VIKTORVAV99:
## Issue

We had a hardcoded sting that could not be translated.

## Description

Fixes the above issue.

### Double check


silkeholmebonnen (approved):
LGTM 🇺🇸 🇸🇪

PR #7105, opened 2024-08-25
Add somme french translations

phiphou:
Add somme french translations for this part :

FireShot Capture 127 - Electricity Maps - Émissions CO₂ de la consommation électrique en tem_ - app electricitymaps com

But : 

- found no translation key for "Share the app"
- for the "About the app" part, there is a translation in the info key object, but it seems it's not the used one.

PR #7109, opened 2024-08-26
Add some french translations

phiphou:
Add french translation for the new "share the app" text in left panel.

ps : I think you could add a little more space before the buttons that follw this text.

Complete #7105

### DIFF
--- a/web/src/locales/fr.json
+++ b/web/src/locales/fr.json
@@ -3,7 +3,8 @@
     "title": "À propos de l'application",
     "text": "L'application Electricity Maps fournit des informations en temps réel sur le web et sur les mobiles pour les citoyens, les experts en énergie, les ONG et les décideurs politiques, favorisant ainsi une meilleure compréhension des défis de la transition énergétique.",
     "open-source-text": "Cette application est <a href='{{link}}' target='_blank'>Open Source</a> et alimentée par la communauté.",
-    "version": "Version de l'application: {{version}}"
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGMT! 🎉

I'll also make a note about the more space, but it's a balance to allow for longer feedback button strings and smaller screens.

phiphou:
> LGMT! 🎉
> 
> I'll also make a note about the more space, but it's a balance to allow for longer feedback button strings and smaller screens.

That's a good point, my translation adds 9 characters to the english one, don't know if it will cause a UI issue...

VIKTORVAV99:
> > LGMT! 🎉
> > I'll also make a note about the more space, but it's a balance to allow for longer feedback button strings and smaller screens.
> 
> That's a good point, my translation adds 9 characters to the english one, don't know if it will cause a UI issue...

It shouldn't but some languages can be even longer (looking at you German 👀).

VIKTORVAV99:
Maybe I spoke too soon 😅

You can check how it looks here: https://staging.electricitymaps.com/

I'll see if we can make it a little better alignment wise but I think we might have to accept that it will be two rows?

phiphou:
For french, I can remove 6 characters : "application" can be shortened to "appli" (like app for application).

For buttons in the "about" part, maybe we can decrease font size a little bit, no ?

PR #7114, opened 2024-08-27
fix: use id prop for Sentry instead of custom prop

VIKTORVAV99:
## Issue

It seems like our custom prop is just added as additional info and is not used to detect unique users.

## Description

Uses the reserved id prop instead. This should hopefully fix the issue.

### Double check


madsnedergaard (approved):
Did you test this locally? Might be the fastest way to see if it works :)

VIKTORVAV99:
> Did you test this locally? Might be the fastest way to see if it works :)

You mean by creating a release and deliberately breaking things?

I did not but I can...

VIKTORVAV99:
Oh that's simpler. 

I'll be away from the computer for a while but I'll do that when I get back.

VIKTORVAV99:
So I tried testing it. And while I get the Sentry error prompt and can provide feedback it's not showing up in sentry...

Are we filtering events based on URL or similar?

VIKTORVAV99:
Got it to work in staging and it's working as expected!

PR #7115, opened 2024-08-27
fix: nordpool session

VIKTORVAV99:
## Issue

We didn't provide a default session.

### Double check


VIKTORVAV99:
Self merging this as the parser probably won't work without it.

PR #7118, opened 2024-08-28
feat(app): use new TimeAverageToggle

VIKTORVAV99:
## Issue

Part of: AVO-357

## Description

Uses the new TimeAverage toggle of #6911 to make the PR easier to review and reduce the scope of the other PR.

It also removes the last icon we used from FontAwesome as the icon is no longer in the design.

### Preview
image

### Double check


Alportan:
Looks good from a design perspective ✅  Hurray 🎉 

One question, when I switch between the toggle items, I see a "Data is temporary unavailable" message. Is that because of dev environment?

Screen Cast 2024-08-28 at 3 06 47 PM

VIKTORVAV99:
> Looks good from a design perspective ✅ Hurray 🎉
> 
> One question, when I switch between the toggle items, I see a "Data is temporary unavailable" message. Is that because of dev environment? I tried a gif, was hard to see due to it's frame rate, but I added a screenshot instead 😅
> 
> Screen Cast 2024-08-28 at 3 06 47 PM

Not because of the dev environment but we have a nasty bug somewhere in the time control logic where a useEffect hook is triggering rendering twice that is proving hard to fix without breaking other stuff.

This is why I added the suspense query before as it should in theory at least hide the issue but that made the whole time controller flash instead which was also sub optimal. What I think I'm gonna try next is creating a simple skeleton like we talked about or some form of hybrid so only parts of the controller are affected.

web/src/components/TimeAverageToggle.tsx
```diff
@@ -17,42 +20,41 @@ export interface TimeAverageToggleProps {
...
+    () =>
+      Object.keys(TimeAverages).map((time) =>
+        createOption(time.toLowerCase() as TimeAverages, t)
+      ),
+    [t]
   );
 
   return (
-    <ToggleGroupPrimitive.Root
+    <ToggleGroupRoot
```
madsnedergaard, line 32: Would be nice to add a keyboard focus state, seems like it worked better before (not sure why though) :)

web/src/features/time/TimeControllerWrapper.tsx
```diff
@@ -32,7 +32,7 @@ function BottomSheetWrappedTimeController() {
       blocking={false}
       header={<TimeHeader />}
     >
-      <TimeController className="p-2 pt-1 min-[370px]:px-4" />
+      <TimeController className="p-2 min-[370px]:px-4" />
```
madsnedergaard, line 35: On smaller screens (and languages with longer text) it would be nice to have some top padding again:

**Before:**


Screenshot 2024-08-30 at 14 55 23


**This PR:**

Screenshot 2024-08-30 at 14 55 17

Alportan: Good catch Mads! 🙌 Some extra padding around the header would be nice 🫶

VIKTORVAV99: The padding itself is actually doubled but I think it's the overflow that's the problem by setting a fixed hight of the "title" to avoid CLS caused by the badge.

But I'll try out a few solutions to see what is working best 🙂

web/src/components/TimeAverageToggle.tsx
Alportan: Good catch, we can go with a simple outline of the pill as an easy fix 🙌 

image

VIKTORVAV99: Are you adding the colors to the Figma or is it the brand color now (but with the dark variant as well)?

Alportan: Brand color, but[ added it to Figma](https://www.figma.com/design/O1DtS6UyZHSH4IymG8Ufsw/%F0%9F%8E%B9-Design-Hack?node-id=5257-5632&t=ntt7JkKDAaO1ylDA-1) as well! 🙌 My bad, I didn't link it before 🙈

madsnedergaard: `shift+tab` two times on a refreshed view takes you to it, then you can go sideways with arrows - so I think it would be nice to just quickly a `focus:outline-brand-green` kinda class as a first step :)

VIKTORVAV99: Yeah I think that would work!

I'll add it then!

madsnedergaard (approved):
LGTM 🥳

PR #7120, opened 2024-08-28
feat(Accordion): Animate the Accordion

VIKTORVAV99:
## Description

Animates the accordion with `@react-spring/web`, it animates both the height and the icon for a smooth rotation.

### Preview

https://github.com/user-attachments/assets/88dbb6ec-2097-4d79-91ff-6b419152c675

### Double check


VIKTORVAV99:
@madsnedergaard if you have any idea on how to fix the initial rotation for the icon I would appreciate it, otherwise I think I'll skip that for now.

In the meantime I'll try and figure out these tests...

web/src/components/Accordion.tsx
```diff
@@ -46,12 +76,18 @@ export default function Accordion({
...
+          style={{ rotate: spring.rotate.to((r) => `${r}deg`) }}
           data-test-id={isCollapsed ? 'collapse-down' : 'collapse-up'}
         />
       </button>
-      {!isCollapsed && <div>{children}</div>}
+      <animated.div style={{ height: spring.height }}>
+        {/* The div below is used to measure the height of the children
+         * DO NOT REMOVE IT
+         */}
+        {renderChildren && <div ref={ref}>{children}</div>}
```
VIKTORVAV99, line 89: An alternative to this would be to create a AccordionContent component that exposes a forward ref. That might be more future prof but will require us to change all the places where we use the accordion.

madsnedergaard (approved):
This looks nice - it still seems like a lot of complicated code to maintain when we could just use a component from the library we're already using, but since you already did this let's use it 😅

VIKTORVAV99:
> This looks nice - it still seems like a lot of complicated code to maintain when we could just use a component from the library we're already using, but since you already did this let's use it 😅 

The other component library, radix, don't have a spring effect. It just uses a boring old css transition 😉

Took a while to figure out how to do this but now it's pretty easy to wrap my head around. Basically you have a starting position and a target position and that's it. So it would have been a lot less complicated if it always had the same starting position for example.

Still need to figure out the tests. And if we want to remove the children from the dom once it's collapsed but other than that it seems to work pretty well now.

The gauges are next so I can finally remove recharts...

madsnedergaard:
> > This looks nice - it still seems like a lot of complicated code to maintain when we could just use a component from the library we're already using, but since you already did this let's use it 😅
> 
> The other component library, radix, don't have a spring effect. It just uses a boring old css transition 😉
> 
> Took a while to figure out how to do this but now it's pretty easy to wrap my head around. Basically you have a starting position and a target position and that's it. So it would have been a lot less complicated if it always had the same starting position for example.
> 
> Still need to figure out the tests. And if we want to remove the children from the dom once it's collapsed but other than that it seems to work pretty well now.
> 
> The gauges are next so I can finally remove recharts...

Haha fair :D

web/src/features/charts/FuturePrice.cy.tsx
```diff
@@ -20,24 +20,16 @@ describe('FuturePrice only positive values', () => {
     cy.get('[data-test-id="future-price"]').should('not.exist');
     cy.get('[data-test-id="collapse-button"]').click();
     cy.get('[data-test-id="future-price"]').should('be.visible');
-  });
```
VIKTORVAV99, line 23: For some unknown reason the tests fails when in their own it statements, but works when it's combined. 🤷🏻‍♂️

Alportan:
> @Alportan could you double check the visual and navigational stuff in here? and @cadeban could you give the code a quick double check since I had to change some stuff after the approving review?

Probably most thing I noticed several times after opening it for the first time, and I remember we talked about it, is that the accordion looks like it's extended by default (arrow down) and you first close it, then open it again. See the gif below 👇 
Screen Cast 2024-10-03 at 10 58 45 AM


Also a few knit picks related to spacing 😅 We might not need to address this in case we move towards the modular design, but I wanted to [document them somewhere](https://www.figma.com/design/O1DtS6UyZHSH4IymG8Ufsw/%F0%9F%8E%B9-Design-Hack?node-id=5427-8160&t=rEIpuXK6UXo0Q6MM-1).
image

Alportan:
I wanted to add to my previous comment that the animation is super smooth 🤩 and I accidentally closed the PR 🤦

web/src/features/charts/FuturePrice.cy.tsx
VIKTORVAV99: Both! But I can't really tell why. It's working as a combined test though.

web/src/components/Accordion.tsx
```diff
@@ -34,18 +39,45 @@ export default function Accordion({
...
+    }),
+    [isCollapsed, isTopExpanding, observerHeight]
+  );
+
   const handleToggleCollapse = () => {
     onClick?.();
 
     isCollapsed && onOpen?.();
 
+    setRenderChildren(true);
```
cadeban, line 59: I know this might be a bit old school, but consider refactoring `handleToggleCollapse` to use if/else instead (see snippet below).
1. I think it's a little easier to read the intent. One can see the open and the close cases clearly.
2. We don't have to rely on a closure, which can cause confusion. At first glance, `onRest: () => setRenderChildren(isCollapsed)` looks like we are intending to `setRenderChildren` to true when we're collapsed and false when we're open, which is the opposite behavior of what we would expect. It works because the onRest closure captured the `isCollapsed` at the time that the callback was created, which is the previous state so it will work as expected, but it's hard to understand immediately.
3. We don't need to toggle the eslint rule.

```javascript
  const handleToggleCollapse = () => {
    onClick?.();
    setRenderChildren(true);

    if (isCollapsed) {
      onOpen?.();

      api.start({
        to: {
          height: observerHeight,
          rotate: isTopExpanding ? 90 : -90,
        }
      });
    } else {
      api.start({
        to: {
          height: 0,
          rotate: isTopExpanding ? -90 : 90,
        },
        onRest: () => {
          setRenderChildren(false);
        },
      });
    }

    setState(!isCollapsed);
  };
  ```

File: web/src/components/Accordion.tsx [hunk 4]
```diff
@@ -34,18 +39,45 @@ export default function Accordion({
...
+      onRest: () => {
+        setRenderChildren(isCollapsed);
+      },
+    });
     setState(!isCollapsed);
   };
 
-  const Icon: LucideIcon = isTopExpanding === isCollapsed ? collapsedIcon : expandedIcon;
+  const Icon: LucideIcon =
+    (collapsedIcon && expandedIcon && isTopExpanding === isCollapsed
```
cadeban, line 75: This expression is challenging to parse. Consider adding some form of documentation, i.e. comment, saving subexpressions into a well named variable, extracted into a function, etc

web/src/components/Accordion.tsx
```diff
@@ -34,18 +39,45 @@ export default function Accordion({
...
+  );
+
   const handleToggleCollapse = () => {
     onClick?.();
 
     isCollapsed && onOpen?.();
 
+    setRenderChildren(true);
+
+    api.start({
```
cadeban, line 61: I'm not super familiar with react-spring. Does it handle `api.stop()` on unmount for us, or do we need to call it in a clean up function?

web/src/features/panels/zone/MethodologyCard.tsx
```diff
@@ -7,16 +7,16 @@ import { useState } from 'react';
 import trackEvent from 'utils/analytics';
 
 export default function MethodologyCard() {
-  const [isCollapsed, setIsCollapsed] = useState(true);
+  const [accordionState, setAccordionState] = useState(true);
```
cadeban, line 10: Consider using`isCollapsed/setIsCollapsed`.  `accordionState === true` is kind of confusing semantically

cadeban:
The animation:

web/src/features/charts/FuturePrice.cy.tsx
cadeban: 🤔 Curious. I don't think it'll be a blocker if it works as a combined test. I might play around with it locally though

cadeban: So I branched this branch, reintroduced the tests, and it [worked](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/11164629034/job/31034519914?pr=7275)! Seems like you fixed whatever the issue was 🎉

cadeban: Maybe it had something to do with the [initial state](https://github.com/electricitymaps/electricitymaps-contrib/pull/7273)?

VIKTORVAV99: Maybe?

But thanks for re-adding them!

web/src/components/Accordion.tsx
VIKTORVAV99: Tried it locally and it seems to work just as well and is a lot more readable!

Feel free to commit it directly or add it as a suggestion instad so you get the credit that you very much deserve!

PS: I like to take size and performance optimisations to the extreme, maybe a bit to far at time so don't hesitate to call me out on it if it goes a step to far 😅

web/src/components/Accordion.tsx
VIKTORVAV99: Really just a work around for the custom icons, provided to the component. I'll make sure to document it though and I've also talked with Alex about as well 🙂

web/src/components/Accordion.tsx
VIKTORVAV99: From what I've read in their documentation and some previous knowledge I think they handle that for us. But I'll try to double check this as well.

cadeban: Yeah I skimmed the docs briefly and didn't see anything, but thought I'd  double-check since you have more xp here 👍

web/src/components/Accordion.tsx
cadeban: Updated. And it's good to have many different perspectives!

VIKTORVAV99:
@cadeban do you feel this is in a merge ready state now?

cadeban (approved):

VIKTORVAV99:
> I still think we should consider updating [this](https://github.com/electricitymaps/electricitymaps-contrib/pull/7120#discussion_r1786362168), but I wouldn't consider it a blocker

Ah I missed that, I'll update it before merging.

PR #7151, opened 2024-09-04
fix: Fix unknown locale error

VIKTORVAV99:
## Issue

It seems like the users locales are not sanetized properly so if they pass gibbrish to the lang query then the whole thing crashes at different points.

Resolves: [APP-WEB-3NP](https://electricitymaps.sentry.io/issues/5660167599/?project=4504366922989568&query=is:unresolved+issue.priority:%5Bhigh,+medium%5D+locale&statsPeriod=90d&stream_index=0)

## Description

This PR adds a check to ensure the locale is valid before setting it as the global and if it's not falls back to english and logs a warning.

### Double check

### DIFF
--- a/web/src/translation/i18n.test.ts
+++ b/web/src/translation/i18n.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from 'vitest';
+
+import { sanitizeLocale } from './i18n';
+
+describe('sanitizeLocale', () => {
+  it.each([
+    ['en', 'en'],
+    ['en-US', 'en-US'],
+    ['fr', 'fr'],
+    ['fr-FR', 'fr-FR'],
+  ])('should return the locale if it is valid: %s', (input, expected) => {
+    expect(sanitizeLocale(input)).toBe(expected);
+  });
+
+  it.each([
+    ['de).', 'de'],
+    ['de-DE-1996', 'de-DE'],
+    ['de-DE-1996-1996', 'de-DE'],
+  ])('should sanitize partially valid locales: %s', (input, expected) => {
+    expect(sanitizeLocale(input)).toBe(expected);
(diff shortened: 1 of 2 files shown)


web/src/translation/i18n.ts
```diff
@@ -17,6 +17,14 @@ i18n
       order: ['querystring', 'localStorage', 'sessionStorage', 'navigator', 'htmlTag'],
       lookupQuerystring: 'lang',
       caches: ['localStorage'],
+      convertDetectedLanguage: (lng: string) => {
+        try {
+          return Intl.getCanonicalLocales(lng)[0];
```
madsnedergaard, line 22: Hmm, we could go further here and strip out non letter characters as well (e.g. `DE).` -> `DE`) as this would help fix invalid markdown links and fix the concrete Sentry error you linked to :)

VIKTORVAV99: We can try but we can't strip all non alphabetic letters, de-AT for example is a perfectly valid locale.

But I can whip up some simple validation.

web/src/translation/i18n.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  it.each([
+    ['e'],
+    ['deAT'],
+    ['deNOX%20wird'],
+    ['itcarbon_intensity'],
+    ['carbon_intensity'],
+  ])('should return "en" if the locale is invalid', (input) => {
+    expect(sanitizeLocale(input)).toBe('en');
+  });
```
VIKTORVAV99, line 38: I suspect some of these could be valid but there is no way to tell so I think this is the best we could do.

madsnedergaard (approved):
This is great, thanks for the update <3

PR #7152, opened 2024-09-04
fix(parsers):Update TR parser to handle TR API update

annaib00:
## Issue
TR down due to TR API update

## Description
Update TR parser to get TGT ticket in order to get data from TR API. 

### Double check


parsers/TR.py
```diff
@@ -56,8 +57,9 @@ def _str_to_datetime(date_string: str) -> datetime:
     Converts string received into datetime format.
     String received is almost in isoformat, just missing : in the timezone
```
VIKTORVAV99, line 58: If it's indeed a ISO format now you can remove this function.

FelixDQ (approved):
Yay, LGTM!

PR #7153, opened 2024-09-04
chore(RoundedCard): add forward ref to RoundedCard so it's reusable for BarBreakdownChart

VIKTORVAV99:
## Issue

We are duplicating styles when we don't need to.

## Description

Adds a forward ref to the RoundedCard so it can be used for the BarBreakdownChart as well.

### Double check


silkeholmebonnen (approved):
LGTM

PR #7154, opened 2024-09-04
fix(parsers): Fix to include new Adani import column in BD data

annaib00:
## Issue
New Adani import column in data for BD

## Description
Fixed parser to handle the new Adani import column and take this into account in exchange with IN-EA. The column is NoneType before the 28th of august 2024.

### Double check


parsers/ERP_PGCB.py
```diff
@@ -214,12 +215,24 @@ def fetch_production(
...
+                )
+
+            else:
+                unknown_source_mw = (
+                    row["total_generation"]
+                    - known_sources_sum_mw
+                    - row["bd_import_bheramara"]
+                    - row["bd_import_tripura"]
+                    - row["bd_import_adani"]
+                )
```
FelixDQ, line 234: Would something like this be simpler? That way, we don't need to write the entire "formula" twice.

```suggestion
              unknown_source_mw = (
                    row["total_generation"]
                    - known_sources_sum_mw
                    - row["bd_import_bheramara"]
                    - row["bd_import_tripura"]
                )

            # Adani import was added after this date
            if target_datetime is None or target_datetime > datetime(2024, 8, 28):
                unknown_source_mw -=  row["bd_import_adani"]
```

FelixDQ (commented):
One suggestion for slightly more readable code, but otherwise it LGTM

FelixDQ (approved):

PR #7160, opened 2024-09-05
chore: ensure better internal type checking in Button

VIKTORVAV99:
## Issue

The internal types where a bit to relaxed and could easily cause misstakes.
## Description

Replaces the internal string validation with explicit string literals.

### Double check


tonypls (approved):
LGTM!

Issue #5393, opened 2023-05-12
[Data Issue]: MX data link throws 403


celsomilne:
### When did this happen?

2023-05-12

### What zones are affected?

MX

### What is the problem?

# Description

It appears that the link for the production data for MX is broken. The [production source page](https://www.cenace.gob.mx/Paginas/SIM/Reportes/EnergiaGeneradaTipoTec.aspx) does not appear to have its data available for download. Running tests on it returns a 403 error, as does clicking on the files on the page itself (i.e., visit the URL and click on the CSV file). 

```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.cenace.gob.mx/SIM/VISTA/REPORTES/EnergiaGenLiqAgregada.aspx
```

# Other available data

I am not familiar with the operation of the SIM, but I believe there are alternate sources of data:

* Demand data for the last 24 hours: See the graph [here](https://www.cenace.gob.mx/Paginas/SIM/OperacionSEN.aspx)

## Market Day Ahead (MDA) data
* Electrical power by load zone: See published data [here](https://www.cenace.gob.mx/Paginas/SIM/Reportes/ResultadosMDA.aspx)
* Prices: Published data [here](https://www.cenace.gob.mx/Paginas/SIM/Reportes/PreEnerServConMDA.aspx)

## Real time prices
* Real time prices: [here](https://www.cenace.gob.mx/Paginas/SIM/Reportes/PreEnerServConMTR.aspx)

## Demand forecasts
* Generation forecasts: [here](https://www.cenace.gob.mx/Paginas/SIM/Reportes/PronosticoGeneracionInter.aspx)
* Demand forecasts: [here](https://www.cenace.gob.mx/Paginas/SIM/Reportes/PronosticosDemanda.aspx)

# Suggested fix

I am not sure which of these data sources are going to be the right ones, but I believe a solution will may include all / some of them! MDA data may be a good enough proxy for now, as might reconciling them with real time prices.

VIKTORVAV99:
These are some great sources!

I just took a quick look at them but I think there are definitely something here we can use!

And yeah it does look like they are forbidden (403) and I hit that fot the last source you provided as well. 

It seems like we can implement some of this data more or less right away (the generation forecasts for wind and solar) but others will need some investigation. The teams are quite busy at the time but I'll check in with them to see if this is enough to start estimating these zones once we have collected enough data.

shuuji3:
[...]

PR #7163, opened 2024-09-06
fix: Mexico CENACE parser

tuxity:
## Issue

Broken CENACE parser, not reported (it seems)

Fixes: #5393

## Description

- update URL
- dynamically load __VIEWSTATE
- fix a deprecation warning

### Preview

N/A

### Double check


FelixDQ:
Thank you @tuxity!

It seems like the data source geoblocks requests from Europe, so we need to use our proxy. I'll try to set it up next week.

tuxity:
> Thank you @tuxity!
> 
> 
> 
> It seems like the data source also started geoblocking requests from Europe, so we need to use our proxy. I'll try to set it up next week.

Sure, let me know if I can do something to speed it up 👍

FelixDQ:
I just added the URL to our proxy whitelist. Would you be able to change the parser so that it uses our US proxy? You can see how it was done for the California parser here: https://github.com/electricitymaps/electricitymaps-contrib/blob/master/parsers/US_CA.py

tuxity:
@FelixDQ Done, but it seems from my end the URL is not correctly whitelisted
https://us-ca-proxy-jfnx5klx2a-uw.a.run.app/Paginas/SIM/Reportes/EnergiaGeneradaTipoTec.aspx?host=https://www.cenace.gob.mx%22 returns `Host https://www.cenace.gob.mx" is not allowed`

FelixDQ:
I think that it works if you remove the %22:

https://us-ca-proxy-jfnx5klx2a-uw.a.run.app/Paginas/SIM/Reportes/EnergiaGeneradaTipoTec.aspx?host=https://www.cenace.gob.mx

FelixDQ:
It looks like the proxy somehow returns `text/plain` instead of `application/octet-stream`. I can run the parser with a breakpoint and see the response. 

i.e. if i outcomment the following, then the parser works as expected:
```py
    if (
        "Content-Type" not in response.headers
        or response.headers["Content-Type"] != "application/octet-stream"
    ):
        raise Exception(
            f"Error while fetching csv for date {datestr}: No CSV was returned by the API. Probably the data for this date has not yet been published."
        )
```

VIKTORVAV99:
@tuxity please do not use force-push it breaks any local copy of your branch anyone else would have.

That only makes it harder for us to review and makes it a lot harder to follow the changes in the PR.

tuxity:
@VIKTORVAV99 I'm doing a rebase on master to avoid doing merge commits. It's a known process when working on your own branch. Why would anyone have a local copy of my branch if I'm alone working on that?

Anyway, PR is ready for review I don't plan to push more code change in this PR.

VIKTORVAV99:
> @VIKTORVAV99 I'm doing a rebase on master to avoid doing merge commits and I'm not squashing commits so you can follow PRs changes. It's a known process when working on your own branch. Why would anyone have a local copy of my branch if I'm alone working on that?
> 
> Anyway, PR is ready for review I don't plan to push more code change in this PR.

How else do you expect us to test it if we don't have a local copy?

I for one have a broken local branch that can't be updated cleanly now.

Just do merge commits, they'll all be squashed when we merge to master.

tuxity:
Isn't what Github workflows supposed to test when updating a PR? If you have to manually test each PRs every time then the processes aren't optimized at all. I will be glad to take a look at that.

VIKTORVAV99:
> Isn't what Github workflows supposed to test when updating a PR? If you have to manually test each PRs every time then the processes aren't optimized at all. I will be glad to take a look at that.
> 
> If you want to update your local branch you can:
> ```
> git reset --hard HEAD~10
> git pull
> ```
> And I'm sure there is also a solution using rebase.
> 
> EDIT:
> Here a better way to reset your local branch:
> ```
> git fetch
> git reset --hard origin/fix_mexico_parser
> ```

Sure but the MX parser has no tests and that only test logic and don't actually make requests to the parser source.

It would be incredibly flaky if they suddenly updates the data blocking all other PR, we would have to add tokens to the workflows that would only be usable if push the branch to this repository which only members can do or use a very insecure alternative that would give anyone access to the tokens.

This is simply things we can not do as it would cause more problems than it solves.

tuxity:
From what I saw there is definitely some improvements to be made around that. I will take a look when I can.

FelixDQ (approved):
LGTM! Thanks a lot for your work on fixing the parser @tuxity 

Sorry for the slow review. We had a team hackathon last week

PR #7162, opened 2024-09-06
chore: remove duplicate function

VIKTORVAV99:
## Issue

We had a function that was nearly identical.

## Description

Removes the duplicate function and cleans up all usage areas of it.

### Double check

- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

PR #7164, opened 2024-09-06
perf: extract isConsumption checks to derived atom

VIKTORVAV99:
## Issue

We are doing a lot of equality checks when we really only need to do one.

## Description

Built on top of #7162 

Extracts the `mode === Mode.CONSUMPTION` into a derived atom and propagate it's value.

### Double check


tonypls (approved):
LGTM! A lesson in handling boolean truthiness in a simple way in the future, nicely done!

PR #7166, opened 2024-09-09
perf: memoize useZoneDataSources

VIKTORVAV99:
## Issue

This function is called 4 times for each timestamp (one time per graph). If you use the time slider this mean we have a lot of function calls and overhead since this is called a maximum of 120 times (30 days x 4) with the exact same data. We can avoid a lot of this overhead my memoizing the internal function calls since they are deterministic.

## Description

Memoizes the calls, moves the early returns to the individual functions so the rules of react are followed with the new memoization and uses arrow functions for some better minimization results.

I did some quick performance tracing and it looks like all the overhead is removed after doing this. (hard to do exact comparisons without profiling builds).

### Double check


silkeholmebonnen (approved):
LGTM! I'm just curious, how do you decide on when to use function vs const?

VIKTORVAV99:
> LGTM! I'm just curious, how do you decide on when to use function vs const?

Generally I always use arrow functions as they have multiple benefits over normal return statements such as multi const declarations `const a=1, b=2` and implicit returns if they are one liners.

The only real exception is the react components as we have a ESLint rule forcing the function use so far.

The other instance I'd use a function over a arrow function is if you need access to its arguments as those are by design not available in arrow functions.

PR #7167, opened 2024-09-09
Update mobile app release workflow

madsnedergaard:
## Description

This changes how we manage certificates, to ensure we use one shared one for the team. This should make onboarding new developers significantly easier :)


madsnedergaard:
Hey @tonypls, can you verify that this works on your end as well? :)

mobileapp/README.md
```diff
@@ -5,14 +5,13 @@ This is a capacitor project that builds the mobile apps from the web directory
...
 - Follow this guide: https://capacitorjs.com/docs/getting-started/environment-setup (but skip the Android SDK part)
-- install JDK v8 <-- to avoid having to create an Oracle account(!), you can find a `jdk-8u321-macosx-x64.dmg` in our internal Google Drive.
 - `brew install gradle`
 - install Android Studio - make sure you open it and go through the install wizard in the start
   - Also go to Tools > SDK Manager and install SDK v29
 - Follow the steps here: https://www.brainfever.co.uk/2022/02/04/build-tool-32-1-0-rc1-is-missing-dx-at/
   - instead of last step, add the following to your `.zshrc` file:
   ```bash
-  export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_321.jdk/Contents/Home
+  export JAVA_HOME='/Applications/Android Studio.app/Contents/jbr/Contents/Home'
```
tonypls, line 14: I haven't tested this part, I use this `export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-21.jdk/Contents/Home`

and on my machine the above is `'/Applications/Android Studio.app/Contents/jre/Contents/Home'` - The jbr -> jre

mobileapp/README.md
```diff
@@ -5,14 +5,13 @@ This is a capacitor project that builds the mobile apps from the web directory
 ## Prerequisites
 
 - Follow this guide: https://capacitorjs.com/docs/getting-started/environment-setup (but skip the Android SDK part)
```
tonypls, line 7: Could also be worth mentioning ruby versions and which ruby version manager to use?

mobileapp/README.md
madsnedergaard: That means you're on an older Android Studio version then :) 
I think this simplifies it a lot, so we don't need to manage and update the separate JDK - the downside is that we need to use the same-ish Android Studio version, but maybe that's an okay tradeoff?

mobileapp/.gitignore
```diff
@@ -9,6 +9,7 @@ fastlane/screenshots
 fastlane/metadata
 fastlane/test_output
 fastlane/fastlane-key.json
+gc_keys.json
```
madsnedergaard, line 12: Can't remember if this one is actually needed or not, let's make sure you try without it next time we make a mobile release. If it is, I can see if I can do something to make it shared/accessible :)

mobileapp/README.md
madsnedergaard: That makes sense, my setup is like this:

```suggestion
- Setup Ruby environment: Use [`rbenv`](https://github.com/rbenv/rbenv) to manage versions, and make sure you install and set `3.1.2` as the default
- Follow this guide: https://capacitorjs.com/docs/getting-started/environment-setup (but skip the Android SDK part)
```

madsnedergaard:
Merging this as it's the best way to figure out if it works ¯\\\_(ツ)\_/¯

PR #7190, opened 2024-09-10
chore(CI): add workflow to CI that ensures all translation are in sync

silkeholmebonnen:
## Description

WIP

### Double check


web/scripts/detectTranslationsOutOfSync.js
VIKTORVAV99, line 1: We do already have the syncTranslations script, would it not be better to modify that instead of duplicating it?

silkeholmebonnen: I thought I would leave it so someone could use the syncTranslation script to fix the translations out of sync

silkeholmebonnen:
Is there a better way to test CI changes than make a PR and see if it works as expected? @VIKTORVAV99

VIKTORVAV99:
> Is there a better way to test CI changes than make a PR and see if it works as expected? @VIKTORVAV99

Nope, which is a bit annoying...

silkeholmebonnen:
Why is it only the Labeler workflow that is running while the others just say expected?

web/scripts/detectTranslationsOutOfSync.js
VIKTORVAV99: What I was thinking was we could modify it so it looks for a env variable called say CI and if it does it collects a list of files with errors instead and prints those and exit with a error code (to fail the CI).

VIKTORVAV99:
> Why is it only the Labeler workflow that is running while the others just say expected? @VIKTORVAV99

That usually happens when GitHub is starting to have problems...

But try merging master, that works sometimes.

.github/workflows/CI.yml
```diff
@@ -112,6 +112,14 @@ jobs:
         run: |
           jsonlint -q web/src/locales/*.json
 
+  check-translation-files-out-of-sync:
+    name: Translations sync
+    needs: [setup]
+    if: >-
+      needs.setup.outputs.json != 'false' ||
+      needs.setup.outputs.workflows != 'false'
+    uses: ./.github/workflows/translations.yml
```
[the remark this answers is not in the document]
silkeholmebonnen: What do you mean by inline? To not have a seperate workflow file?

silkeholmebonnen:
It looks like it works now. Can you review it @VIKTORVAV99 🙏

.github/workflows/CI.yml
```diff
@@ -112,6 +112,23 @@ jobs:
         run: |
           jsonlint -q web/src/locales/*.json
 
+  check-translation-files-out-of-sync:
+    name: Translations sync
```
VIKTORVAV99, line 116: Could we call this something like `Validate Translation Files` or something else that is a little more descriptive?

VIKTORVAV99 (approved):
LGTM! 🎉

Just one small comment about the name but otherwise it seems to be working as intended.

Nice work!

PR #7191, opened 2024-09-11
perf: use state data for gauges

VIKTORVAV99:
## Issue

Currently when you press on a zone the gauges first go to `?` and then animate to their position when the data has loaded, this is not ideal as it causes a lot of visual changes for just a split second. 

## Description

We can fix the above issue by using the global state data instead, this is always loaded and available and is the same we use for the map tooltips ensuring these are always in sync.

Ideally we should also use this for the carbon intensity graph but that requires deeper changes that would be better suited to it's own PR if we decide to do that.

### Double check


cadeban (approved):
LGTM - looks much snappier! I also like that you de-duplicated the gauge components 👍

VIKTORVAV99:
> LGTM - looks much snappier! I also like that you de-duplicated the gauge components 👍

Yeah I did, since they use the same data structure now I figured it would be best to just do it or else we will forget about it and have duplicated code.

Thanks for the review!

Issue #7194, opened 2024-09-11
Timeline slider and scale buttons invisible in Firefox


torokati44:
**Describe the bug**

The usual controls to switch between Hourly/Daily/Monthly/Yearly aggregates, and to view historical data, are invisible.
They still work, just can't be seen.

**To Reproduce**
Open the app in Firefox.

**Expected behavior**

All the controls are visible.

**Screenshots**

![image](https://github.com/user-attachments/assets/60f2ca09-f49b-4613-a5e2-629325bb9f02)

**Desktop (if relevant):**
 - OS: Fedora Linux 40
 - Browser: Firefox
 - Version: 130.0

**Additional context**

It started happening just now, when I "updated" to the new version of the app (v1.164.0). It worked fine yesterday (on v1.163.0) - although there was https://github.com/radix-ui/primitives/issues/2777 already.

torokati44:
Disabling any of the following properties of the style applied to the
```css
*,
:before,
:after
```
selector:
```css
  --tw-blur: ;
  --tw-brightness: ;
  --tw-contrast: ;
  --tw-grayscale: ;
  --tw-hue-rotate: ;
  --tw-invert: ;
  --tw-saturate: ;
  --tw-sepia: ;
  --tw-drop-shadow: ;
  ```
... in the Developer Tools makes the controls appear as normal.

VIKTORVAV99:
Thanks for reporting it!

That sure is odd! 👀

I'll work on a fix right away.

torokati44:
On Firefox Android, it works in portrait mode (when the controls are in the pull-up drawer), but not in landscape mode, when they are in the floating panel:
![Screenshot_20240911-160357](https://github.com/user-attachments/assets/83344769-7507-458c-89ef-72b711168cd5)
![Screenshot_20240911-160405](https://github.com/user-attachments/assets/f28819f7-40df-47ed-bda0-9a7d0e030a5d)

VIKTORVAV99:
This is the offending PR: https://github.com/electricitymaps/electricitymaps-contrib/pull/7118

The problem is that I don't see anything obvious that would cause this. But it does seem to be related to the tailwind styles.

VIKTORVAV99:
According to https://caniuse.com/css-backdrop-filter it should be supported but something is not working as expected.

VIKTORVAV99:
[...]

PR #7195, opened 2024-09-11
fix: fix TimeAverageToggle not visible on Firefox

VIKTORVAV99:
## Issue

Fixes: #7194 

## Description

I don't know exactly why but apparently `backdrop-blur` breaks things when it's used here.
Let's just remove it for now since it don't affect the current styles and re-evaluate how this will be used in the future.

### Double check


VIKTORVAV99:
Self merging this to get the fix out quickly!

PR #7198, opened 2024-09-12
perf: streamline getContributors logic

VIKTORVAV99:
## Issue

The way this was set up and later modified was not exactly optimal.

## Description

Streamlines the logic here to avoid returning all contributors, something that was not logical and potentially uses more memory than it needs as it can't release the full array.

And reduces the need for multiple index lookups in the array, something that should help with performance even if it's small.

### Double check


madsnedergaard (approved):
Looks good to me! Also when testing the staging branch, I realised that I haven't been good at adding myself as contributor anywhere 🥲

VIKTORVAV99:
> Looks good to me! Also when testing the staging branch, I realised that I haven't been good at adding myself as contributor anywhere 🥲

Samesies

I did in the early day but then I forgot/didn't bother.

Maybe we should fix that...
We can add ourself to all the zones (or just hack the function to always add us) 😎

PR #7193, opened 2024-09-11
Changed CollapseButton and made a copy called ShareButton

seburu:
## Issue

Needed a share button

## Description

Creating a share button and edit the collapse left panel button

PR #7199, opened 2024-09-12
feat(web): Add copy image to clipboard functionality (#7193)

wobniarin:
## Description
We mock a screenshot image and we copy it to our clipboard so it can be shared :) The mocked image will be removed once we have the feature of generating a screenshot.

Note that Clipboard API does not support both Text and Image at the same time. We are just focusing on sharing an image for now. 

### Preview

image

We copied an image to the clipboard 🎉 

### Double check
N/A


web/package.json
```diff
@@ -75,6 +75,7 @@
     "@visx/group": "^3.3.0",
     "@visx/shape": "^3.5.0",
     "@visx/text": "^3.3.0",
+    "clipboard": "^2.0.11",
```
VIKTORVAV99, line 78: I would avoid adding a package for this if we can. Take a look at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard instead.

wobniarin: Thanks for the suggestion! Since we are hacking we are just trying things out as quickly as possible :) But if you guys implement this in production then it makes sense to build it nicer!

PR #7205, opened 2024-09-12
Add french translations for tooltips

phiphou:
Add french translations for the main indicators tooltips.

I followed the English wording, so "from generating electricity" gives "issues de la production d'électricité".

But it would be cool if this tooltip text was dynamic : the term "generating" should be "production" or "consumption" depending on the app current mode.

### DIFF
--- a/web/src/locales/fr.json
+++ b/web/src/locales/fr.json
@@ -290,7 +290,9 @@
   "tooltips": {
     "carbonintensity": "Intensité carbone",
     "zoneHeader": {
-      "lowcarbon": "Inclut les énergies renouvelables et le nucléaire"
+      "lowcarbon": "Inclut les énergies renouvelables et le nucléaire",
+      "renewable": "Inclut le solaire, l'éolien, l'hydroélectricité, la biomasse et la géothermie",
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Good call about production vs consumption!

I'll see what we can do about that, but it should be fairly easy to add.

VIKTORVAV99 (approved):
LGTM! 🎉

phiphou:
> Good call about production vs consumption!
> 
> I'll see what we can do about that, but it should be fairly easy to add.

I see two options : 

- you reuse the terms production / consumption and their translation
- we use two different keys in translation files

Not sure about what is the best choice.

VIKTORVAV99:
> > Good call about production vs consumption!
> > I'll see what we can do about that, but it should be fairly easy to add.
> 
> I see two options :
> 
> * you reuse the terms production / consumption and their translations
> * we use two different keys in translation files
> 
> Not sure about what is the best choice.

I think the easiest might be to just do two different keys since replacing parts of it could create unexpected behavior in some languages, especially non latin ones (I think).

PR #7207, opened 2024-09-13
chore: clean up generated config and config types

VIKTORVAV99:
## Issue

We had some Any types, included data we didn't use, and had sub-optimal config generation.

## Description

This PR does 3 small things:

- It cleans up 2 Any types
- It removes the contributor array from a zone config if it don't have any contributors.
- It removes the capacity data from exchanges as that is unused as we use the data sent through the backend.

### Double check


web/geo/types.ts
```diff
@@ -21,13 +21,23 @@ export interface ZoneConfig {
...
     productionPerUnit?: string;
   };
-  [key: string]: any;
+  [key: string]:
+    | undefined
+    | string[]
+    | number[][]
+    | string
+    | boolean
+    | { [key: string]: string };
```
tonypls, line 30: Haha, at this point I'm not sure it's better than any 🤣

VIKTORVAV99: Agreed but at least ESLint don't complain 😅

tonypls (approved):
LGTM

web/geo/types.ts
tonypls: We could add this somewhere 😆 
```
export interface Any {
  // Primitive types
  string: string;
  number: number;
  boolean: boolean;
  null: null;
  undefined: undefined;
  symbol: symbol;
  bigint: bigint;

  // Object types
  object: object;
  array: any[];
  tuple: [any, any];
  function: (...args: any[]) => any;

  // Union type
  union: string | number | boolean;

  // Intersection type
  intersection: { a: string } & { b: number };

  // Literal types
  stringLiteral: "literal";
  numberLiteral: 42;
  booleanLiteral: true;

  // Enum
  enum: "A" | "B" | "C";

  // Generic types
  generic: Array<any>;
  promise: Promise<any>;
  map: Map<any, any>;
  set: Set<any>;

  // Class
  class: { new (...args: any[]): any };

  // Recursive type
  recursive: Any;

  // Optional property
  optional?: any;

  // Index signature for additional properties
  [key: string]: unknown;
}
```

PR #7209, opened 2024-09-13
Tvs/bottom sheet

tonypls:
## Issue
We have a bulky bottom sheet

## Description
This PR replaces [React Spring Bottom Sheet](https://www.npmjs.com/package/react-spring-bottom-sheet) with our own lightweight bottom sheet, inspired by one I made [here](https://www.npmjs.com/package/bottom-sheet-dialog).

## Testing guidelines
On device iOS and android:
- Interactions
- Slider doesn't move sheet up and down
- Bottom sheet remains in place when time averages are toggled


web/src/features/time/BottomSheet.ts
```diff
@@ -0,0 +1,340 @@
...
+    });
+  }
+  if (hasTailwindBackground) {
+    Object.assign(backdrop.style, {
+      position: 'absolute',
+      left: '0',
+      right: '0',
+      bottom: '0',
+      transition: 'height 0.3s ease-out',
+    });
```
VIKTORVAV99, line 36: Do we need both of these?

And even if we do it would be nice to deduplicate most of the code by only adding/changing the background color if tailwind is not used.

web/src/features/time/BottomSheet.ts
```diff
@@ -0,0 +1,340 @@
...
+    setupElementStyles(element);
+    element.append(backdropElement);
+    updateChildrenHeight();
+    setSnap(0);
+    setupAccessibility();
+    return addEventListeners();
+  }
+
+  function setupAccessibility(): void {
+    element.setAttribute('role', 'dialog');
```
VIKTORVAV99, line 78: Instead of setting a ARIA attribute we should use the `<dialog>` element:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

web/src/features/time/BottomSheet.ts
```diff
@@ -0,0 +1,340 @@
...
+    return () => {
+      element.removeEventListener('touchstart', handlers.start);
+      element.removeEventListener('touchmove', handlers.move);
+      element.removeEventListener('touchend', handlers.end);
+      element.removeEventListener('mousedown', handlers.start);
+      window.removeEventListener('mousemove', handlers.move);
+      window.removeEventListener('mouseup', handlers.end);
+      window.removeEventListener('resize', resizeHandler);
+      resizeObserver.disconnect();
+    };
```
VIKTORVAV99, line 172: Would it be possible to add these directly to the elemetns in a more "react" syle code?

We should be gettng a lot of the logic in here by default then.

web/src/features/time/BottomSheet.ts
```diff
@@ -0,0 +1,340 @@
...
+
+    // If velocity is low, or at the edges, snap to closest
+    return closestSnapIndex;
+  }
+
+  function calculateNewHeight(height: number): number {
+    const [lowestSnapPoint, highestSnapPoint] = [
+      Math.min(...state.snapPointsWithChildHeight),
+      Math.max(...state.snapPointsWithChildHeight),
+    ];
```
VIKTORVAV99, line 263: Why use array assignemnt here instead of direct assignement to the two constants?

It will be compiled to `const lowestSnapPoint = Math.min(...state.snapPointsWithChildHeight), highestSnapPoint = Math.max(...state.snapPointsWithChildHeight)` at build anyhow (and then minified) which means using direct assignment is slightly more efficent.

web/src/features/time/BottomSheet.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    const rubberBandFactor = 1;
+    if (height > highestSnapPoint) {
+      const overscroll = height - highestSnapPoint;
+      return highestSnapPoint + (1 - Math.exp(-overscroll / 200)) * 50 * rubberBandFactor;
+    }
+
+    const underscroll = lowestSnapPoint - height;
+    return lowestSnapPoint - (1 - Math.exp(-underscroll / 200)) * 50 * rubberBandFactor;
```
VIKTORVAV99, line 276: Would this not be better suited to something like react-spring annimations?

web/src/features/time/ReactBottomSheet.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }, [snapPoints, backgroundColor, excludeElementRef]);
+
+  useEffect(() => {
+    if (bottomSheetInstanceReference.current) {
+      bottomSheetInstanceReference.current.snapTo(currentSnapIndex);
+    }
+  }, [snapPoints, currentSnapIndex]);
+
+  return <div ref={bottomSheetReference}>{children}</div>;
```
VIKTORVAV99, line 52: This would be the element that is turned into a dialog element instead.

web/src/features/time/TimeController.tsx
```diff
@@ -16,89 +16,94 @@ import { useIsBiggerThanMobile } from 'utils/styling';
...
+        </div>
       </div>
-    </div>
-  );
-}
+    );
+  }
+);
+
+TimeController.displayName = 'TimeController';
```
VIKTORVAV99, line 108: If you use a function declaration inside the forward ref this is not needed.

With the changes in react 19 with ref as a prop we won't need forward ref anymore so it will be a cleaner change then if we used the fucntion declaration syntax instead.

VIKTORVAV99 (commented):
Some early comments, but it would be really great if you could add some documentation to some of the functions. It's quite hard to understand how they interact sometimes.

Seems to be working great though!

tonypls:
@VIKTORVAV99 Thanks for the comments, very much appreciate the feedback, I'll try add changes when I have time but I'm not sure we'll be able to prioritize this during this cycle. Feel free add commits if you like and I'll take a look next slack week.

tonypls:
Closing this, will re-open if we have the appetite for it in the future.

PR #7210, opened 2024-09-15
feat(SEO): add react-helmet-async

VIKTORVAV99:
## Issue

Closes: AVO-483
## Description

Adds react-helmet-async and uses it to replace our current meta tag manipulation. It replaces our custom logic so it's possible to extend it's functionality in upcoming PRs.

>[!NOTE]
>I decided to use react-helmet-async over react-helmet as it's a more modern fork that is maintained as react-helmet was throwing errors since it used react internal functions that should not be used anymore. react-helmet-async does not have this problem.


tonypls (approved):
LGTM! Looks great and nice to split this into pieces, makes the reviewing easier

PR #7211, opened 2024-09-15
chore(SEO): use semantic HTML

VIKTORVAV99:
## Issue

Closes: AVO-518

## Description

Uses semantic HTML to improve content discovery by search engines.

### Double check


web/src/components/Accordion.tsx
```diff
@@ -35,7 +35,7 @@ export default function Accordion({
   const Icon: LucideIcon = isTopExpanding === isCollapsed ? ChevronUp : ChevronDown;
 
   return (
-    <div className="flex flex-col gap-1.5 py-1">
+    <section className="flex flex-col gap-1.5 py-1">
```
silkeholmebonnen, line 38: Is section the right one to use here? As far as I can tell `<details>` would be more appropriate to use

web/src/features/charts/RoundedCard.tsx
```diff
@@ -6,14 +6,14 @@ export const RoundedCard = forwardRef<
   { children: ReactNode; className?: string }
 >(function RoundedCard({ children, className }, reference) {
   return (
-    <div
+    <section
```
silkeholmebonnen, line 9: Would `<article>` be better to use here?

web/src/features/panels/ranking-panel/RankingPanel.tsx
```diff
@@ -52,8 +52,8 @@ export default function RankingPanel(): ReactElement {
   return (
     <div className="flex max-h-[calc(100vh-236px)] flex-col py-3 pl-4 pr-1 ">
       <div className="pb-5">
-        <div className="font-poppins text-lg font-medium">{t('ranking-panel.title')}</div>
-        <div className="text-sm">{t('ranking-panel.subtitle')}</div>
+        <h1>{t('ranking-panel.title')}</h1>
+        <h2 className="text-sm">{t('ranking-panel.subtitle')}</h2>
```
silkeholmebonnen, line 56: This also impacts the size. But that might be fine?
Before:
image
Now:
image

silkeholmebonnen (commented):
A few comments but looks good

web/src/components/Accordion.tsx
VIKTORVAV99: Details seem to have it's own accordion behaviour that we can't control in a easy way: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

Section works the same as a div but has more semantic meaning and should always include a heading or distinct set of elements (like buttons with names) which fits our needs here.

web/src/features/charts/RoundedCard.tsx
VIKTORVAV99: Don't think so, in my view article would be used mostly for text content and not visualisations like this that can change (using the time slider).

web/src/features/panels/ranking-panel/RankingPanel.tsx
VIKTORVAV99: @Alportan what do you think?

I think the H1 is fine but maybe the font should be changed in the h2, but it is a subtitle so maybe we should use the "title font"?

Alportan: So it's a custom, smaller H2? I think it works as it is. Don't have any strong preferences here 🙌

VIKTORVAV99: Yeah it's a h2 but a tad smaller.

Mostly it's just so we have a h1, h2 and then the zone titles are h3 so we should have them all.

Alportan: Sounds good! All good on my side 👏

web/src/features/charts/RoundedCard.tsx
silkeholmebonnen: I juat saw this and thought that product card would be similar to this. But it's not a dealbreaker for me so I'll let you decide
image

silkeholmebonnen (approved):
LGTM ⭐

web/src/features/charts/RoundedCard.tsx
VIKTORVAV99: I think it's safer to keep it as a section for now.

PR #7212, opened 2024-09-16
chore(SEO): Use the correct link for get our data

VIKTORVAV99:
## Issue

We are not actually linking to the get our data page. This can hurt the SEO for that page.

## Description

Updates the link to https://electricitymaps.com/get-our-data and adjusts the Swedish translation slightly.

### Double check

### DIFF
--- a/web/src/locales/sv.json
+++ b/web/src/locales/sv.json
@@ -221,7 +221,7 @@
     }
   },
   "header": {
-    "get-data": "Hämta våra data",
+    "get-data": "Använd vår data",
     "blog": "Blogg",
     "open-source": "Öppen källkod",
     "hiring": "Vi anställer!",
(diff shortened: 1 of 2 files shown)


web/src/locales/sv.json
```diff
@@ -221,7 +221,7 @@
     }
   },
   "header": {
-    "get-data": "Hämta våra data",
+    "get-data": "Använd vår data",
```
VIKTORVAV99, line 224: Technically this means "Use our data" but it's better than "Hämta" which means more like Download than Get in this context. "Skaffa" could also be used but it's less formal.

Alportan (approved):
Ohh my... you did it! 🥳 

Looking good, happy to see this done! ✅

PR #7213, opened 2024-09-16
feat(SEO): Add dynamic canonical links

VIKTORVAV99:
## Issue

All our current pages have the same canonical link.

Fixes: AVO-487

## Description

Uses react helmet async to control our canonical links, when the language is `en` there is no query parameters added as they don't control the visual content. If any other language is use the param `?lang=` is set.

### Double check


web/src/App.tsx
```diff
@@ -73,6 +75,7 @@ export default function App(): ReactElement {
       >
         <title>{`Electricity Maps | ${t('misc.maintitle')}`}</title>
         <meta property="og:locale" content={i18n.languages[0]} />
+        {canonicalLink}
```
VIKTORVAV99, line 78: @cadeban would you prefer if this was a component instead?
Might be better DX to so it better matches how it used?

web/src/hooks/getCanonicalLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const useGetCanonicalLink = () => {
+  const { i18n } = useTranslation();
+  const { pathname } = useLocation();
+  const currentLanguageKey = i18n.languages[0];
+  return (
+    <link
+      rel="canonical"
+      href={`${baseUrl}${pathname}${
+        currentLanguageKey == 'en' ? '' : `?lang=${currentLanguageKey}`
```
cadeban, line 17: nit: consider using strict equality here, since there isn't a need for type conversion

web/src/App.tsx
VIKTORVAV99: Yeah, that is a concern. Not sure what is the best way to do it as I don't want to duplicate the functionality across the pages/components.

VIKTORVAV99: Switched it up to use a combination of a hook + link elements which should adress both concerns.

cadeban: Yeah, I hear you re: duplication. It seems like the `react-helmet-async` [unions](https://github.com/nfl/react-helmet?tab=readme-ov-file#features) all of the Helmet definitions. Maybe we could do something like define a `CanonicalHelmet` component that could consolidate the hook, link, and helmet and then update `app.tsx` to something like below?

```jsx
// app.tsx
...
  <Helmet htmlAttributes={{...}} prioritizeSeoTags>
    <title>{`Electricity Maps | ${t('misc.maintitle')}`}</title>
    <meta property="og:locale" content={i18n.languages[0]} />
  </Helmet>
  <CanonicalHelmet />
...
```

cadeban: (Otherwise, lgtm!)

VIKTORVAV99: We will be putting more stuff in the other helmets like title and description manipulation so I'm not sure having multiple helmets on each page would be good either.

Perhaps we can go with this for now and then look for optimisations once we have all the tags in place?

cadeban (approved):
LGTM

web/src/App.tsx
cadeban: sgtm!

PR #7214, opened 2024-09-16
feat(parsers): add proxy to malaysia (MY-WM) parser

FelixDQ:
I added a new @use_proxy decorator that uses a new proxy with locations for Malaysia. This should only be used as a fallback proxy for parsers where the other proxies doesn't work.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7216, opened 2024-09-17
fix(MY): change order of decorators

FelixDQ:
I added the decorators in the wrong order in #7214 while testing the error message 🤦 

This fixes the parser


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7217, opened 2024-09-18
Tvs/add country name as hover

tonypls:
## Issue
Before:
image

## Description
After:
image

Note, only works when a country name exists in the translations / zone configs. Works for Corsica and Sicily etc where the country can need further infomation.


madsnedergaard:
I don't know, I quite like TODO 😂

cadeban (approved):
lgtm

cadeban:
TODO is the best country

PR #7221, opened 2024-09-20
fix: abbreviate month in hourly datetimes

VIKTORVAV99:
## Issue

On very small screens the date can overflow now that we added the timezone which is causing some layout issues.

## Description
Abbreviate the months in hourly view and update the tests to match the new expected format.

#### Master
image

#### This branch
image

### Double check


Alportan (approved):
Looks amazing and solving a bunch of small resolution issues! 😍

PR #7222, opened 2024-09-20
feat(SEO): Add dynamic meta titles

VIKTORVAV99:
## Issue

We are using the same meta title for all our pages which could be hurting our SEO and goes against best practices.

Fixes: AVO-484

## Description
Updates the links to use a new suffix, `'title' | App | Electricity Maps` so we can match this with other parts of our websites and products. `'title' | Blog | Electricity Maps` or `'title' | Docs | Electricity Maps`, etc.

The title itself will be either the existing one on the `/map` view or the translated zone name on zone pages.

### Double check

### DIFF
--- a/web/src/utils/constants.ts
+++ b/web/src/utils/constants.ts
@@ -1,6 +1,7 @@
 import type { Duration } from 'date-fns';
 import { ElectricityModeType } from 'types';
 
+export const metaTitleSuffix = ' | App | Electricity Maps';
 export const baseUrl = 'https://app.electricitymaps.com';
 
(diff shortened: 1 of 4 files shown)


cadeban (approved):
lgtm

web/src/utils/constants.ts
```diff
@@ -1,6 +1,7 @@
 import type { Duration } from 'date-fns';
 import { ElectricityModeType } from 'types';
 
+export const metaTitleSuffix = ' | App | Electricity Maps';
```
cadeban, line 4: 👍

PR #7223, opened 2024-09-20
feat(web): AGI solves the wind layer, maybe

tonypls:
## Issue
The wind layer trails remained on safari and were super resource intensive

## Description
This PR updates the wind layer to have better compatibility


VIKTORVAV99 (approved):
LGTM! 🎉

Issue #3404, opened 2021-09-23
Add a sitemap.xml file


madsnedergaard:
### Description

Improve SEO further!

Following up from the work in #3363 where the canonical links where optimized, it would be great to also create a `sitemap.xml` file to reflect this.

I'm not sure if we should include `?lang=XX` entries in the sitemap - anyone with SEO experience that know what's the best approach here? :)

PR #7224, opened 2024-09-21
feat(SEO): Generate static sitemap at build

VIKTORVAV99:
## Issue

We don't have a sitemap for the app, which could be causing discovery issues in google and other search engines.

Closes: AVO-494
Closes: #3404

## Description

Creates a script that generates a sitemap at build (mostly so it will always be up to date and so we don't forget to update it). Currently it only builds basic links without any language tags, but this should be sufficient for our purposes for now at least.

>[!IMPORTANT]
>This currently don't include alternative language links. I'll add this once I figure out how to best approach AVO-524.

### Double check


web/public/robots.txt
```diff
@@ -1 +1,3 @@
-User-agent: *
\ No newline at end of file
+User-agent: *
+
+Sitemap: http://app.electricitymaps.com/sitemap.xml
```
VIKTORVAV99, line 3: This is valid syntax and recommended by google: https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt

web/scripts/generateSitemap.js
```diff
@@ -0,0 +1,32 @@
+// Generates a static sitemap for the app based on the zones in the config file
+
+import fs from 'node:fs';
+import path from 'node:path';
+
+import zonesConfig from '../config/zones.json' assert { type: 'json' };
```
madsnedergaard, line 6: Do we want to actually add zones with no data to the sitemap? :)

madsnedergaard: Great, thanks for the reply 🙌

cadeban, line 6: Is our node version during the earthly build process pegged to a specific version or will it default to the most recent major LTS? I think Node updated [import asserts -> import attributes](https://nodejs.org/docs/latest-v20.x/api/esm.html#import-attributes) in >20.10.0 and not I'm not certain if we need to handle that.

cadeban: Otherwise, it looks good! Really love that we won't have to manually manage it :D

VIKTORVAV99: Ah good spotting, it should use the latest 20.x version so should be fine to switch. I'll go ahead and switch, if it passes CI it should be fine.

VIKTORVAV99: So slight problem, prettier screams at me for using `with`. We'll have to keep this for now and then switch later once we have upgraded prettier (which is not as straight forward as it sounds sadly).

cadeban: Prettier whyy 😭 - SGTM

cadeban (approved):

web/scripts/generateSitemap.js
VIKTORVAV99: Pretty sure this one is on us, we use Prettier 2 so we can do some hacky thing to avoid creating a top level package.json config. Sounds like we might have to bite the bullet and do it though. Probably as a workspace so we don't duplicate dependencies.

PR #7226, opened 2024-09-23
perf: optimize getHasSubZones

VIKTORVAV99:
## Description

Just a small size and performance optimization to getHasSubZones by taking advantage of a Boolean constructor and optional chaining, functionality wise it's identical to the previous method but it's more elegant IMO.

### Double check


madsnedergaard:
To me, "elegant" often means it's becoming [clever code](https://www.joshwcomeau.com/career/clever-code-considered-harmful/) 😉 

While your solution here might be slightly fewer lines of code, I personally find it harder to read and understand what is going on. In my opinion it's not worth sacrificing readability over such a small improvement (which I also couldn't reproduce with a basic benchmark on https://jsbench.me/, see screenshot below) especially given that our codebase is open-source and we want more people than ourselves to be able to read and understand what is going on :) 


Screenshot of benchmark, dunno if I'm reproducing it fairly though:
Screenshot 2024-09-23 at 16 45 06

VIKTORVAV99:
> https://jsbench.me/

This is what I get using the would be code output instead of the boolean constructor (vite will transform this to `!!`):
image

I'd say the new and hybrid are the same speed but the new one will be slightly smaller code wise.

So the new code is faster, but I do agree it's slightly harder to understand at first glance so I'd be happy to describe it step by step in a comment (or a few).

madsnedergaard:
We're talking 23 million operations per second vs 26 million operations, right? So calling it 300 times would make it 1.51 **micro**second slower for the user if I'm doing math right (or well, if ChatGPT is doing math correctly - I didn't even try 🙈).

VIKTORVAV99:
> We're talking 23 million operations per second vs 26 million operations, right? So calling it 300 times would make it 1.51 **micro**second slower for the user if I'm doing math right (or well, if ChatGPT is doing math correctly - I didn't even try 🙈).

I wouldn't trust GPT with math but it will be small 😅

I'll leave it up to you but I do prefer smaller and faster code in general. Especially since this probably won't be changed much.

So feel free to either close or merge it (or tell me to add a bunch of comments).

madsnedergaard (approved):
Approving to move on, but given that this codebase is public and we want people on various levels to contribute, I think you should generally be cautious with optimizations like this 😉

VIKTORVAV99:
> Approving to move on, but given that this codebase is public and we want people on various levels to contribute, I think you should generally be cautious with optimizations like this 😉

I added a jsdoc comment with the previous version of the code. Hopefully this should get improve things a little. But in the future I'll save the heavy optimisations for where it's making a bigger difference.

Issue #7219, opened 2024-09-18
Electricity price should be in cent/kWh


SHSGAGNON:
**Is your feature request related to a problem? Please describe.**
The feature improve readability

**Describe the solution you'd like**
The Hourly Electricity Price graph should be in cent/kWh for better visibility. This is the unit used commonly for electricity bill. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

VIKTORVAV99:
> **Is your feature request related to a problem? Please describe.**
> 
> The feature improve readability
> 
> 
> 
> **Describe the solution you'd like**
> 
> The Hourly Electricity Price graph should be in cent/kWh for better visibility. This is the unit used commonly for electricity bill. 
> 
> 
> 
> **Describe alternatives you've considered**
> 
> N/A
> 
> 
> 
> **Additional context**
> 
> N/A
> 
>

Hi!

I assume you mean EURO cent?

We tried to match it other websites that showed electricity prices for Europe in per kWh (most use per MWh as that's the trading unit).

SHSGAGNON:
I know. The MWh has factor 10 difference to the common daily unit, now you have a factor 1/100. I just want my eyes to quickly decide if it is expensive or cheap from customer point of view.

madsnedergaard:
Thanks for sharing your thoughts here, since we are showing the wholesale spot market price and not your actual residential price including all tariffs, we have decided to stick with €/MWh to make it clear it's not the same thing :)

PR #7227, opened 2024-09-23
feat(web): add future price 

silkeholmebonnen:
## Issue

[AVO-455](https://linear.app/electricitymaps/issue/AVO-455/add-upcoming-price-section-for-nordpool-areas)

## Description

This PR adds a section to the price card that shows future prices for the zones that we have future price for (nordpool zones). 

### Preview

image

### Double check


package.json
VIKTORVAV99, line 1: Did you mean to add this to the top level and not /web?

web/src/features/charts/PriceChart.tsx
```diff
@@ -42,7 +44,11 @@ function PriceChart({ datetimes, timeAverage }: PriceChartProps) {
     }));
   }
 
-  if (!chartData[0]?.layerData?.price) {
+  if (
+    chartData[0]?.layerData?.price === undefined ||
+    Number.isNaN(chartData[0]?.layerData?.price) ||
+    chartData[0]?.layerData?.price === null
+  ) {
```
silkeholmebonnen, line 51: This had to be changed since the previous implementation would result in not showing price data if the first price entry was 0.

package.json
silkeholmebonnen: No, I'll change it

silkeholmebonnen:
Feel free to take a look at the future price @tonypls @Alportan @cadeban 🌞

silkeholmebonnen:
I'll just do a few changes @VIKTORVAV99. I found a bug that I will fix and we decided in the office that we will try and switch to MWh.

VIKTORVAV99:
> I'll just do a few changes @VIKTORVAV99. I found a bug that I will fix and we decided in the office that we will try and switch to MWh.

Sounds good, I'll probably not have time to take a deep look at this until tomorrow or Wednesday. 🙂

VIKTORVAV99:
> I'll just do a few changes @VIKTORVAV99. I found a bug that I will fix and we decided in the office that we will try and switch to MWh.

@Alportan there is also a issue open here #7219 to show prices in Euro Cent per KWh, which might be an option. But I agree on doing MWh now and then we can look into cents later on.

Alportan:
> > I'll just do a few changes @VIKTORVAV99. I found a bug that I will fix and we decided in the office that we will try and switch to MWh.
> 
> @Alportan there is also a issue open here #7219 to show prices in Euro Cent per KWh, which might be an option. But I agree on doing MWh now and then we can look into cents later on.

Nice! 🙌 Let's start with switching back to MWh for now. 

It's sneaky, but this way we can monitor incoming feedback the other way (people being upset about switching back to MWh instead of kWh) then we have some more evidence for the need of a toggle between the 2 views. 🥷

silkeholmebonnen:
Should be ready for review now. The only thing missing is the highligt of the current hour (as seen here) but that could also be done in v2.

image

VIKTORVAV99 (commented):
Looks really good!

Some minor comments but overall it's a pretty clean implementation!

VIKTORVAV99:
I did notice a blocker for this but it's related to the backend, basically we need to double check that we have a license for the zone before we send the day ahead data. If you look at [Croatia for example](https://6d57fa4d.electricitymaps-staging.pages.dev/zone/HR) then we hide the data in the graph due to license issues but you can still see the day ahead prices.

silkeholmebonnen:
> I did notice a blocker for this but it's related to the backend, basically we need to double check that we have a license for the zone before we send the day ahead data. If you look at [Croatia for example](https://6d57fa4d.electricitymaps-staging.pages.dev/zone/HR) then we hide the data in the graph due to license issues but you can still see the day ahead prices.

Until we change it in the backend we would be able to do a simple check like this in the frontend to ensure that we only show future price if the price is not disabled:
`{!isPriceDisabled && <FuturePrice futurePrice={futurePrice} />}`

Do you think we should do that for now?

VIKTORVAV99:
> > I did notice a blocker for this but it's related to the backend, basically we need to double check that we have a license for the zone before we send the day ahead data. If you look at [Croatia for example](https://6d57fa4d.electricitymaps-staging.pages.dev/zone/HR) then we hide the data in the graph due to license issues but you can still see the day ahead prices.
> 
> 
> 
> Until we change it in the backend we would be able to do a simple check like this in the frontend to ensure that we only show future price if the price is not disabled:
> 
> `{!isPriceDisabled && <FuturePrice futurePrice={futurePrice} />}`
> 
> 
> 
> Do you think we should do that for now?

Not sure that's good enough, we would still send the data. But it shouldn't be too hard to do it in the backend as we are already doing it for the other price information.

silkeholmebonnen:
> > > I did notice a blocker for this but it's related to the backend, basically we need to double check that we have a license for the zone before we send the day ahead data. If you look at [Croatia for example](https://6d57fa4d.electricitymaps-staging.pages.dev/zone/HR) then we hide the data in the graph due to license issues but you can still see the day ahead prices.
> > 
> > 
> > Until we change it in the backend we would be able to do a simple check like this in the frontend to ensure that we only show future price if the price is not disabled:
> > `{!isPriceDisabled && <FuturePrice futurePrice={futurePrice} />}`
> > Do you think we should do that for now?
> 
> Not sure that's good enough, we would still send the data. But it shouldn't be too hard to do it in the backend as we are already doing it for the other price information.

Okay, I will change it in the backend ASAP:)

Alportan:
> Not sure that's good enough, we would still send the data. But it shouldn't be too hard to do it in the backend as we are already doing it for the other price information.

How long do you estimate it would take to implement the backend changes for checking the license before sending day-ahead data? 🙋 

Given that we're getting more user feedback requesting prices in MWh instead of kWh (and now in the form of appstore reviews), we could consider cherry-picking just that change and release it separately if the full set of changes wouldn't be as straightforward. 💡

Alternatively, if the backend changes are quick, we should proceed with the full implementation. 💪

What do you think the best approach would be? 🙌 @VIKTORVAV99 @silkeholmebonnen

VIKTORVAV99:
> > Not sure that's good enough, we would still send the data. But it shouldn't be too hard to do it in the backend as we are already doing it for the other price information.
> 
> How long do you estimate it would take to implement the backend changes for checking the license before sending day-ahead data? 🙋
> 
> Given that we're getting more user feedback requesting prices in MWh instead of kWh (and now in the form of appstore reviews), we could consider cherry-picking just that change and release it separately if the full set of changes wouldn't be as straightforward. 💡
> 
> Alternatively, if the backend changes are quick, we should proceed with the full implementation. 💪
> 
> What do you think the best approach would be? 🙌 @VIKTORVAV99 @silkeholmebonnen

They just got merged 😉

VIKTORVAV99 (approved):
LGTM! 🎉

I look forward to spending hours and hours looking at prices!

Nice work!

Issue #6345, opened 2024-01-14
Precision of emission factors is misleading


corradio:
## Description
The number of digits used for the emission factors are leading people to believe they are more precise than they are:
<img width="421" alt="image" src="https://github.com/electricitymaps/electricitymaps-contrib/assets/1655848/96357a54-d8e2-46ce-a035-8275ef93ecf2">

This tweet made me realise it:
https://twitter.com/robbie_andrew/status/1733081430672941174

I suggest rounding to the nearest 10g (or the nearest g) to avoid this.

VIKTORVAV99:
I think rounding to the whole number makes the most sense, at least for our calculated numbers, otherwise things might be even weirder since nuclear currently is below 10g for most of Europe.

But how should we handle it when some of the scientific sources we use have higher precision?

corradio:
> But how should we handle it when some of the scientific sources we use have higher precision?

Good point - I think this could be an exception.

w-flo:
I don't think these need to be rounded in the config file if they're just intermediate results from some calculation like the ETS/ENTSO-E mapping. Otherwise, you could argue that before starting the emission factor calculation, total emissions and total electricity generation of each plant should be rounded as well, and then round those intermediate emission factors of power plants before combining them for the country factor. So IMHO, intermediate values should be stored in high precision. (Of course we don't really need 7 decimal places, but I wouldn't say that's something that needs fixing either.)

When emission factors are displayed in the app / website for public consumption though, they should always be rounded in some way of course, to avoid the misleading impression. I believe that's already implemented everywhere?

VIKTORVAV99:
> I don't think these need to be rounded in the config file if they're just intermediate results from some calculation like the ETS/ENTSO-E mapping. Otherwise, you could argue that before starting the emission factor calculation, total emissions and total electricity generation of each plant should be rounded as well, and then round those intermediate emission factors of power plants before combining them for the country factor. So IMHO, intermediate values should be stored in high precision. (Of course we don't really need 7 decimal places, but I wouldn't say that's something that needs fixing either.)
> 
> When emission factors are displayed in the app / website for public consumption though, they should always be rounded in some way of course, to avoid the misleading impression. I believe that's already implemented everywhere?

Thanks for your view and insights, and yes they are already rounded in the frontend.

[...]

PR #7237, opened 2024-09-26
feat(zone-config): update discharge emission factors for all zones

silkeholmebonnen:
## Issue

[AVO-475](https://linear.app/electricitymaps/issue/AVO-475/update-discharge-emission-factors-for-all-zones)

## Description

This PR updates the discharge emission factors for all zones using the script `update_co2eq_parameters_overrides.py`. Unfortunately it also makes a bunch of formatting changes (I'm sorry reviewer).

### Double check

### DIFF
--- a/config/zones/AD.yaml
+++ b/config/zones/AD.yaml
@@ -11,6 +11,6 @@ capacity:
 contributors:
   - Manu1400
   - autipial
-timezone: Europe/Andorra
-region: Europe
 country: AD
+region: Europe
+timezone: Europe/Andorra
--- a/config/zones/AE.yaml
+++ b/config/zones/AE.yaml
@@ -29,6 +29,29 @@ capacity:
     - datetime: '2023-01-01'
       source: Climatescope
       value: 39451.8
+  nuclear:
+    - datetime: '2020-08-19'
+      source: IAEA PRIS
+      value: 1310
+    - datetime: '2021-09-14'
+      source: IAEA PRIS
+      value: 2620
+    - datetime: '2022-10-08'
+      source: IAEA PRIS
+      value: 3930
+    - datetime: '2024-03-24'
+      source: IAEA PRIS
+      value: 5240
+  oil:
+    - datetime: '2016-01-01'
+      source: Climatescope
+      value: 219.9
+    - datetime: '2017-01-01'
+      source: Climatescope
+      value: 218.6
+    - datetime: '2018-01-01'
+      source: Climatescope
+      value: 71.4
   solar:
     - datetime: '2016-01-01'
       source: Climatescope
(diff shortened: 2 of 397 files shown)


silkeholmebonnen:
Should I update the test to match the new emission factors?

VIKTORVAV99:
> Should I update the test to match the new emission factors?

Yes :)

VIKTORVAV99:
If possible could we round the output to 2 decimals?

FelixDQ:
It would be nice to have a pre-commit or script that ensures the keys are always in the same order. It's kind of impossible to review PRs like this. I guess that we'll have to trust the script for now

VIKTORVAV99:
> It would be nice to have a pre-commit or script that ensures the keys are always in the same order. It's kind of impossible to review PRs like this. I guess that we'll have to trust the script for now

Most of the changes should be easy to review, it's only a couple of manually edited files that are out of order (they where not sorted by key before).

I'll see about either adding a pre-commit hook for it, or a simple check in the CI, or both for future PRs.

silkeholmebonnen:
How can I test this locally? 

I have tried running `poetry run test` from contrib but it fails with the error message:

⚙︎ pytest 
/bin/sh: /Users/sshb/Desktop/emaps/electricitymaps/contrib/.venv/bin/pytest: /Users/sshb/Desktop/electricitymaps/contrib/.venv/bin/python: bad interpreter: No such file or directory
FAILED: pytest

silkeholmebonnen:
> If possible could we round the output to 2 decimals?

Yes, but it would require an update to the script. But do we want to lose the precision of the numbers?

FelixDQ (approved):
Let's go ahead with the current precision. We can adjust later

madsnedergaard:
> > If possible could we round the output to 2 decimals?
> 
> Yes, but it would require an update to the script. It's an easy fix and I already tried to run it but do we want to lose the precision of the numbers?

For future I would say yes, the precision doesn't help us (we only show 2 decimals in app anyway) and might make people believe it's extremely accurate :)

Here's a previous discussion on the topic: https://github.com/electricitymaps/electricitymaps-contrib/issues/6345

silkeholmebonnen:
> > > If possible could we round the output to 2 decimals?
> > 
> > 
> > Yes, but it would require an update to the script. It's an easy fix and I already tried to run it but do we want to lose the precision of the numbers?
> 
> For future I would say yes, the precision doesn't help us (we only show 2 decimals in app anyway) and might make people believe it's extremely accurate :)
> 
> Here's a previous discussion on the topic: #6345

I created a slack friendly task for this: [AVO-561](https://linear.app/electricitymaps/issue/AVO-561/round-discharge-emission-factors)

PR #7249, opened 2024-09-30
feat(config): add malaysia country zone

FelixDQ:
Adds the country-level zone for Malaysia


config/zones/MY.yaml
```diff
@@ -0,0 +1,10 @@
+contributors:
+  - kruschk
+  - systemcatch
+  - nessie2013
```
VIKTORVAV99, line 4: Technically not needed since the build script will aggregate the contributors from the subzones to the parent zone, but why don't you add yourself as well ;)

VIKTORVAV99 (approved):
LGTM! 🎉

@tonypls maybe you can double check just in case?
The country level data is not expected to work until after it's merged.

config/zones/MY.yaml
```diff
@@ -0,0 +1,11 @@
+country: MY
+disclaimer: 'This zone is partially using fully estimated data from MY-EM'
+estimation_method:
+  CONSTRUCT_BREAKDOWN
```
tonypls, line 4: Do you think it would make sense to make a follow up PR and change all of these for country level to `estimation_method: SPATIAL_AGGREGATE`?

tonypls (approved):
LGTM 🎉

config/zones/MY.yaml
VIKTORVAV99: > Do you think it would make sense to make a follow up PR and change all of these for country level to `estimation_method: SPATIAL_AGGREGATE`?

Either that or we just check that it has subzones, but maybe that's not strict enough?
If you want to change them now that is fine with me though. Then we can look into optimizing it later.

VIKTORVAV99:
I'll merge this and create a new release so the data is hidden on the map like it should be.

PR #7272, opened 2024-10-03
fix(web): day a head price styling features

silkeholmebonnen:
## Issue

[AVO-553](https://linear.app/electricitymaps/issue/AVO-553/v2-day-a-head-price-styling-features)

## Description

This PR:
-  adds highlights the current hour
- adds boldness of "Now" 
- disable future price for countries without day-a-head price (see [AU-SA](https://app.electricitymaps.com/zone/AU-SA) for reference of where it should be disabled but is currently not).
- Right aligns the price for €
- Puts the currency symbol in front for all other currencies than € (I'm not sure which one should be the fallback. I did a bit of googling and thought it was better this way around)

### Preview

image
image
image

### Double check


web/src/features/charts/FuturePrice.tsx
```diff
@@ -176,9 +179,17 @@ export function PriceBar({
...
+      <p className="min-w-[66px] text-nowrap text-end text-sm font-semibold tabular-nums">
+        {`${price} ${symbol}`}
+      </p>
+    );
+  }
   return (
     <p className="min-w-[66px] text-nowrap text-sm font-semibold tabular-nums">
-      {`${price} ${getSymbolFromCurrency(currency)}`}
+      {`${symbol}${price}`}
     </p>
```
VIKTORVAV99, line 193: @Alportan do you think we can put the currency symbol on the right side for all currencies? I think most would would have it on the right, especially if we start using more local currencies in the future?

We might even want to consider switching to named currencies like `USD`, `EUR`, `SEK`, `DKK` etc which should all be placed on the right.

silkeholmebonnen: > @Alportan do you think we can put the currency symbol on the right side for all currencies? I think most would would have it on the right, especially if we start using more local currencies in the future?

Tony commented on the fact that it was on the right side, and I think many english speakers would notice it. So I don't think that we should put it on the right side for all currencies.

> We might even want to consider switching to named currencies like `USD`, `EUR`, `SEK`, `DKK` etc which should all be placed on the right.

I think that depends on the language, in english you put it in front, that is what [this](https://www.scribens.com/grammar-rules/typography/currency-format.html) article says at least

silkeholmebonnen: I think you're actually supposed to write the currency code on the left always and where you put the symbol depends on the language: [currency symbols and codes](https://proofed.com/knowledge-hub/currency-symbols-and-codes/)

Alportan: Gotta love localisation 🙃

VIKTORVAV99: Here are the more in depth documentation around the various options: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#currency_2

silkeholmebonnen:
@cadeban do you have time to review this?

web/src/features/charts/FuturePrice.tsx
```diff
@@ -74,16 +73,27 @@ export function FuturePrice({ futurePrice }: { futurePrice: FuturePriceData | nu
             <ul>
               {Object.entries(filteredPriceData).map(
                 ([date, price]: [string, number]) => (
-                  <li key={date}>
+                  <li
+                    key={date}
+                    className={
+                      isNow(date, granularity)
+                        ? `rounded-md bg-[#18214F]/10 dark:bg-[#848EC0]/10`
```
cadeban, line 80: nit: If we're using these hex codes in multiple places, then we can consider giving them more human-readable names in the [tailwind config](https://tailwindcss.com/docs/customizing-colors#naming-your-colors)

web/src/features/charts/FuturePrice.tsx
```diff
@@ -263,16 +287,29 @@ const getColor = (
...
+    return 'bg-danger dark:bg-red-400';
+  } else if (price === minPrice) {
+    return 'bg-success dark:bg-emerald-500';
   } else {
     return 'bg-[#18214F] dark:bg-[#848EC0]';
   }
 };
+
+const isNow = (date: string, granularity: number): boolean =>
+  normalizeToGranularity(new Date(date), granularity).getTime() ==
```
cadeban, line 305: nit: Do we need type conversion here? Might be more type safe to use `===`

web/src/features/charts/FuturePrice.tsx
```diff
@@ -263,16 +287,29 @@ const getColor = (
...
-    return 'bg-success dark:bg-emerald-500';
-  } else if (
+  if (
     normalizeToGranularity(new Date(date), granularity) <
     normalizeToGranularity(new Date(), granularity)
   ) {
     return 'bg-[#18214F] dark:bg-[#848EC0] opacity-50';
+  } else if (price === maxPrice) {
+    return 'bg-danger dark:bg-red-400';
+  } else if (price === minPrice) {
```
cadeban, line 297: :thumbs

web/src/features/charts/FuturePrice.tsx
```diff
@@ -175,10 +181,28 @@ export function PriceBar({
...
+    currency: currency,
+    maximumSignificantDigits: 4,
+    currencyDisplay: 'narrowSymbol',
+  }).format(price);
   return (
-    <p className="min-w-[66px] text-nowrap text-sm font-semibold tabular-nums">
-      {`${price} ${getSymbolFromCurrency(currency)}`}
+    <p
+      className={`min-w-[66px] overflow-clip text-nowrap text-sm font-semibold tabular-nums ${
+        Number.isNaN(Number(priceString.at(-1))) ? 'text-end' : 'text-start'
```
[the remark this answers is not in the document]
cadeban: 💯

cadeban (approved):
LGTM! Sound logic & looks ✨ nice ✨  I left a couple of nits, but nothing blocking.

web/src/features/charts/FuturePrice.tsx
silkeholmebonnen: @Alportan do you think we will use this color again (the purple price bar color)? If we are not planning on using it again right now, I think it's fine to leave it as hex

Alportan: Are these the colours for the price data graph? If yes, then yes, it would be cool to color code certain data that today uses a neutral color even more, such as price and exchanges. I second adding it to the `config` 🙌

silkeholmebonnen: Yes! I'll add it to the config:)

PR #7274, opened 2024-10-03
fix(web): day ahead price styling features v3

silkeholmebonnen:
## Issue

[AVO-559](https://linear.app/electricitymaps/issue/AVO-559/v3-day-ahead-price)

## Description

This PR highlights the top/bottom 5 percentage as green / red instead of just one value. It also shows 0 as a circle (12x*12 px) instead of nothing.

This is a follow up PR to #7272, so lets get that merged first.

### Preview
Before:
Screenshot 2024-10-09 at 11 39 32

After:
Screenshot 2024-10-09 at 11 39 38

image
image

### Double check


Alportan:
Quick feedback that @VIKTORVAV99 brought up 💡 

Maybe we can ditch the `At` completely for a cleaner view. Looking at it again, it might not necessarily bring more value, so we can ditch it altogether and go for a cleaner (and more small device friendly) option. ✨ What do you think @silkeholmebonnen ? 🙌

silkeholmebonnen:
> Quick feedback that @VIKTORVAV99 brought up 💡
> 
> Maybe we can ditch the `At` completely for a cleaner view. Looking at it again, it might not necessarily bring more value, so we can ditch it altogether and go for a cleaner (and more small device friendly) option. ✨ What do you think @silkeholmebonnen ? 🙌

Sounds good! I tried to play around with how we can avoid the space being so big when it is in languages that doesn't have AM/PM or similar, but I didn't find a good solution. Do you have an idea @VIKTORVAV99?   
image

VIKTORVAV99:
> > Quick feedback that @VIKTORVAV99 brought up 💡
> 
> > 
> 
> > Maybe we can ditch the `At` completely for a cleaner view. Looking at it again, it might not necessarily bring more value, so we can ditch it altogether and go for a cleaner (and more small device friendly) option. ✨ What do you think @silkeholmebonnen ? 🙌
> 
> 
> 
> Sounds good! I tried to play around with how we can avoid the space being so big when it is in languages that doesn't have AM/PM or similar, but I didn't find a good solution. Do you have an idea @VIKTORVAV99?   
> 
> image
> 
> 

I don't think the space is an issue but if we think it is I'd either pad it to center the time a bit more or give more space to the actual graph.

But like I said I don't think it's an issue.

Alportan:
> > Quick feedback that @VIKTORVAV99 brought up 💡
> > Maybe we can ditch the `At` completely for a cleaner view. Looking at it again, it might not necessarily bring more value, so we can ditch it altogether and go for a cleaner (and more small device friendly) option. ✨ What do you think @silkeholmebonnen ? 🙌
> 
> Sounds good! I tried to play around with how we can avoid the space being so big when it is in languages that doesn't have AM/PM or similar, but I didn't find a good solution. Do you have an idea @VIKTORVAV99? 

Yeah, it does look a bit off when there's too much spacing 🥲 Maybe there is an elegant solution to this 🤞

Alportan:
> > > Quick feedback that @VIKTORVAV99 brought up 💡
> > 
> > 
> > > 
> > 
> > 
> > > Maybe we can ditch the `At` completely for a cleaner view. Looking at it again, it might not necessarily bring more value, so we can ditch it altogether and go for a cleaner (and more small device friendly) option. ✨ What do you think @silkeholmebonnen ? 🙌
> > 
> > 
> > Sounds good! I tried to play around with how we can avoid the space being so big when it is in languages that doesn't have AM/PM or similar, but I didn't find a good solution. Do you have an idea @VIKTORVAV99?
> > image
> 
> I don't think the space is an issue but if we think it is I'd either pad it to center the time a bit more or give more space to the actual graph.
> 
> But like I said I don't think it's an issue.

Definitely not a blocker, but would be nice to have a more fluid layout when it comes to localising.

web/src/features/charts/FuturePrice.tsx
```diff
@@ -46,13 +47,37 @@ export function FuturePrice({ futurePrice }: { futurePrice: FuturePriceData | nu
...
+    if (current) {
+      observer.observe(current);
+    }
+
+    return () => {
+      if (current) {
+        observer.unobserve(current);
+      }
+    };
+  }, [priceBarContainerReference, negativePercentage, isCollapsed]);
```
VIKTORVAV99, line 75: Do we really need a use effect?

It should only be used as a escape hatch if it can't fit in the normal rendering flow of react.

Wouldn't the useResizeObserver hook be a better option?

web/src/features/charts/FuturePrice.tsx
```diff
@@ -190,7 +252,7 @@ function TimeDisplay({ date, granularity }: { date: string; granularity: number
 
   if (isNow(date, granularity)) {
     return (
-      <p className="min-w-[82px] text-sm" data-test-id="now-label">
+      <p className={`min-w-[72px] pl-1 text-sm font-semibold`} data-test-id="now-label">
```
VIKTORVAV99, line 255: You can add a custom size 18 to the tailwind config to avoid using a custom spacing here instead.

72px = 4.5rem = tailwind 18 (don't exist in the default theme).

web/src/features/charts/FuturePrice.tsx
```diff
@@ -202,9 +264,7 @@ function TimeDisplay({ date, granularity }: { date: string; granularity: number
   }).format(datetime);
 
   return (
-    <p className="min-w-[82px] text-nowrap text-sm tabular-nums">
-      {`${t(`country-panel.price-chart.at`)} ${formattedDate}`}
-    </p>
+    <p className={`min-w-[72px] text-nowrap text-sm tabular-nums`}>{formattedDate}</p>
```
VIKTORVAV99, line 267: then it can be reused here as well.

web/src/features/charts/FuturePrice.tsx
```diff
@@ -263,16 +323,35 @@ const getColor = (
...
+  } else if (
+    priceIn5Percentile(price, maxPrice, minPrice, true) &&
+    maxPrice != minPrice
+  ) {
+    return 'bg-danger dark:bg-red-400';
+  } else if (
+    priceIn5Percentile(price, maxPrice, minPrice, false) &&
+    maxPrice != minPrice
+  ) {
+    return 'bg-success dark:bg-emerald-500';
```
VIKTORVAV99, line 340: Should the dark not be bg-success-dark?

@Alportan

web/src/features/charts/futurePriceUtils.ts
```diff
@@ -72,3 +72,30 @@ export const negativeToPostivePercentage = (
...
+  negativePercentage: number,
+  isNegative: boolean
+): number => {
+  if (!width) {
+    return 0;
+  }
+  if (isNegative) {
+    return ((width - 12) * negativePercentage) / 100 + 12;
+  }
+  return ((width - 12) * (100 - negativePercentage)) / 100 + 12;
```
VIKTORVAV99, line 87: Would be nice with some comments that describe what these magic numbers are for our future selfs, or extracting them to constants as they seem to be re-used a lot.

web/src/features/charts/futurePriceUtils.ts
```diff
@@ -72,3 +72,30 @@ export const negativeToPostivePercentage = (
...
+  price: number,
+  maxPrice: number,
+  minPrice: number,
+  inTop: boolean
+): boolean => {
+  const priceRange = maxPrice - minPrice;
+  if (inTop) {
+    return price >= maxPrice - priceRange * 0.05;
+  }
+  return price <= minPrice + priceRange * 0.05;
```
VIKTORVAV99, line 100: Same here.

VIKTORVAV99 (commented):
Some comments but looks like it's working!

web/src/features/charts/FuturePrice.tsx
silkeholmebonnen: bg-emerald-500 is the same as bg-success-dark, except that it doesn't work bg-success-dark doesn't behave as expected. That's why i used bg-emerald-500, but maybe i should just fix it in the tailwind config instead

VIKTORVAV99: > bg-emerald-500 is the same as bg-success-dark, except that it doesn't behave as expected. That's why i used bg-emerald-500, but maybe i should just fix it in the tailwind config instead

In what way does it not behave as expected? 👀

silkeholmebonnen: It uses the bg-succes color

Alportan: `bg-success-dark` is indeed mapped to `emerald-500` in Figma, but it wasn't showing correctly if I recall correctly (I think @silkeholmebonnen did mentioned this to me a while back). If there's a way to make it work, it would be nice, otherwise we can also revise it later, maybe together with most colors. 

We could do another pair-code/design session, similar to the one for typography 🙌

Alportan: > `bg-success-dark` is indeed mapped to `emerald-500` in Figma, but it wasn't showing correctly if I recall correctly (I think @silkeholmebonnen did mentioned this to me a while back). If there's a way to make it work, it would be nice, otherwise we can also revise it later, maybe together with all colors to make sure the repo and Figma match.
> 
> We could do another pair-code/design session, similar to the one for typography 🙌

Alportan: Yes please 🙏

Alportan: Slack week topic, but we're mid way haha 😅

silkeholmebonnen: Okay seems to work for me as well, idk what happened before. I'll change it to use `dark:bg-success-dark`

web/src/features/charts/FuturePrice.tsx
silkeholmebonnen: I tried just using the useResizeObserver without the useEffect but I could not get it to work. If you have time and think it is important then it would be nice if you could try it. But you decide

VIKTORVAV99: I can take a quick look but otherwise I don't consider this a blocking thing.

silkeholmebonnen:
@VIKTORVAV99, do you think I should try to refactor like GA-reviews says?

VIKTORVAV99:
> @VIKTORVAV99, do you think I should try to refactor like GA-reviews says?

No I don't think we need that, I would open a backlog issue about looking into using the resizeObserver hook though. But I don't think it's blocking.

I'll give this a final review tomorrow but I don't anticipate any issues🙂

web/src/features/charts/FuturePrice.tsx
```diff
@@ -81,31 +107,34 @@ export function FuturePrice({ futurePrice }: { futurePrice: FuturePriceData | nu
...
-                        {hasNegativePrice && (
-                          <div
-                            className="flex flex-row justify-end"
-                            style={{
-                              width: `${negativePercentage}%`,
-                            }}
-                            data-test-id="negative-price"
-                          >
-                            {price < 0 && (
+                    <div>
```
VIKTORVAV99, line 110: I don't think this div is needed?

VIKTORVAV99 (approved):
LGTM! 🎉

I have one minor comment though, and I think we should look into the resize observer again at some point but none of them are blocking!

silkeholmebonnen:
I just noticed this:
image

This was what happened when I tried using the useResizeObserver, and I thought it was only when I switched to useResizeOberser, but apparently it does it now as well... To recreate open the futureprice and adjust the width of the browser. I will look into fixing this before we merge this PR

VIKTORVAV99:
> I just noticed this: image
> 
> This was what happened when I tried using the useResizeObserver, and I thought it was only when I switched to useResizeOberser, but apparently it does it now as well... To recreate open the futureprice and adjust the width of the browser. I will look into fixing this before we merge this PR

I didn't notice that when using it normally, but if you want to take another look at it don't let me stop you!

silkeholmebonnen:
I think I found a much simpler way of calculating the width that fixes all the problems (not using a useEffect and the resizing problem) @VIKTORVAV99

web/src/features/charts/FuturePrice.tsx
```diff
@@ -81,31 +81,33 @@ export function FuturePrice({ futurePrice }: { futurePrice: FuturePriceData | nu
...
+                        />
+                        <div className="flex h-full w-full flex-row self-center">
+                          {hasNegativePrice && price < 0 && (
+                            <div
+                              className="flex flex-row justify-end"
+                              style={{
+                                width: `calc(calc(calc(100% - 12px) * ${
+                                  negativePercentage / 100
+                                }) + 12px)`,
+                              }}
```
VIKTORVAV99, line 106:
```suggestion
                                                            style={{
                                width: `calc(100% * ${negativePercentage / 100} + 12px - 12px * ${negativePercentage / 100})`,
                              }}
```
You should be able to use just one calc. Making this even simpler! But very nice approach!

web/src/features/charts/FuturePrice.tsx
```diff
@@ -115,27 +117,32 @@ export function FuturePrice({ futurePrice }: { futurePrice: FuturePriceData | nu
...
+                            </div>
+                          )}
+                          {price >= 0 && (
+                            <div
+                              data-test-id="positive-price"
+                              style={{
+                                width: `calc(calc(calc(100% - 12px) * ${
+                                  (100 - negativePercentage) / 100
+                                }) + 12px)`,
+                              }}
```
VIKTORVAV99, line 129:
```suggestion
                              style={{
                                width: `calc(100% * ${(100 - negativePercentage) / 100} + 12px - 12px * ${(100 - negativePercentage) / 100})`,
                              }}
```
same here. You'll probably have to run the formatting afterwards though 😅

silkeholmebonnen:
Do you want to give it a final review or do you think its good to go now?

VIKTORVAV99:
> Do you want to give it a final review or do you think its good to go now?
> 
> 

I took a quick look at the last changes and I don't think it needs a new review, so just go ahead and merge it!

PR #7281, opened 2024-10-04
Error fix for broken path generation in windows at generateSitemap.js and Error: EPERM fix 

shing1Sks:
## Issue

1.Due to an extra leading '/' or '\\' in the dirname variabe a broken path such as  path: 'C:\\C:\\Users\\SHREYASH%20KUMAR%20SINGH\\Desktop\\hackfest\\electricitymaps-contrib\\web\\public' like this is generated 
2.An EPERM error

## Description

Added an extra check to check if an extra  '/' or '\\' exists and removing it if it does for the issue 1

Wrapped the directory check code in try catch block to better handle the issues related and it handles the EPERM error

### Preview

1.
issue1
2.
issue2
-- final works --
after_changes

### Double check

- [*] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [*] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

### DIFF
--- a/web/src/components/Accordion.tsx
+++ b/web/src/components/Accordion.tsx
@@ -1,5 +1,10 @@
-import { ChevronDown, ChevronUp, LucideIcon } from 'lucide-react';
+import { animated, useSpring } from '@react-spring/web';
+import { ChevronRight, LucideIcon } from 'lucide-react';
+import { useState } from 'react';
 import { twMerge } from 'tailwind-merge';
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (changes requested):
As far as I can tell this should work just as well but please split out the un-related parser changes into their own PR.

PR #7282, opened 2024-10-05
fix: fix logic in ZoneHeaderTitle to match tooltip

VIKTORVAV99:
## Issue

The logic was different between the Tooltip and ZoneHeaderTitle and this works better in languages like Swedish for example.

## Description

Just convert the strings to lower case before converting them.

### Preview

Before:
image

After:
image

### Double check


VIKTORVAV99:
@Alportan this country pill thingy, should it use the badge component instead of this custom thing?

And should we use that on the tooltips as well as we talked about before?

Alportan:
> @Alportan this country pill thingy, should it use the badge component instead of this custom thing?
> 
> And should we use that on the tooltips as well as we talked about before?

Yes to badge component and yes to tooltips. Just to double check, the badge would only be shown in case the country name is not included in the zone name in any way. Did that logic stay the same?

VIKTORVAV99:
> > @Alportan this country pill thingy, should it use the badge component instead of this custom thing?
> 
> > 
> 
> > And should we use that on the tooltips as well as we talked about before?
> 
> 
> 
> Yes to badge component and yes to tooltips. Just to double check, the badge would only be shown in case the country name is not included in the zone name in any way. Did that logic stay the same?

Yes it would only show if the country is not part of the zone name. The only logic change in here is to make the comparison case insensitive so that for example in Swedish "Mellansverige" are recognized to contain "Sverige". The tooltips already do this but apparently not the title.

VIKTORVAV99:
I'll create a follow-up PR for that to keep the changes here to an absolute minimum.

VIKTORVAV99:
Poking @tonypls for a review 👀

PR #7284, opened 2024-10-06
Force added CAPBridgeViewController.swift

### LINKED ISSUE #6966: iOS header overlaps with UI
**Describe the bug**
The iOS header (time, date, battery etc..) overlaps with the UI (see top right of the screen).

**To Reproduce**
Open app

**Expected behavior**
iOS header should not overlap

**Screenshots**
Screenshot 2024-07-09 at 14 15 47 jpeg

**Additional context**
Using the app on iPad (classic)

Xeno-Tek:
## Issue

Closes #6966

## Description

This pull request fixes the issue where the iOS status bar (time, date, battery, etc.) overlaps with the app UI, as described in issue #6966. The problem was that the `webView` was not respecting the safe area insets, causing the UI content to be displayed underneath the status bar.

### Changes Made:

- **Modified `loadView()` Method:**
  - Wrapped the `webView` inside a container `UIView`.
  - Added the `webView` as a subview of the container.
  - Applied Auto Layout constraints to the `webView` to ensure it respects the safe area by anchoring it to the `containerView.safeAreaLayoutGuide`.
  - Set the container view as the main view of the view controller (`self.view = containerView`).
  
- **Added Documentation:**
  - Included detailed documentation comments for the `loadView()` method to explain the setup and purpose of each step.

### Before Fix:
The iOS status bar overlaps with the app UI elements.

### After Fix:
The app UI now respects the safe area, and content is displayed correctly below the status bar.

## Double Check

- [x] I have tested the app on various iOS devices, including those with and without notches, to ensure the UI displays correctly and the status bar does not overlap with the content.
- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top-level directory to format my changes.
- [x] I have verified that the changes do not introduce any new issues or affect existing functionality.
- [x] I have added appropriate documentation comments to the modified code for clarity and maintainability.

## Additional Context

- **Issue Reference:** This fix addresses the issue raised by @corradio where the iOS header overlaps with the UI on iPad devices.
- **Testing:** The fix has been tested on iPad 10th generation in both portrait and landscape orientations, ensuring that the UI elements are properly aligned and no longer overlap with the status bar.

### DIFF
--- a/mobileapp/node_modules/.pnpm/@capacitor+ios@6.0.0_@capacitor+core@6.0.0/node_modules/@capacitor/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/mobileapp/node_modules/.pnpm/@capacitor+ios@6.0.0_@capacitor+core@6.0.0/node_modules/@capacitor/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -1,0 +1,401 @@
+import UIKit
+import WebKit
+import Cordova
+
+@objc open class CAPBridgeViewController: UIViewController {
+    private var capacitorBridge: CapacitorBridge?
+    public final var bridge: CAPBridgeProtocol? {
+        return capacitorBridge
+    }
+
+    public fileprivate(set) var webView: WKWebView?
+
+    public var isStatusBarVisible = true
+    public var statusBarStyle: UIStatusBarStyle = .default
+    public var statusBarAnimation: UIStatusBarAnimation = .fade
+    @objc public var supportedOrientations: [Int] = []
+
+    public lazy final var isNewBinary: Bool = {
+        if let curVersionCode = Bundle.main.infoDictionary?["CFBundleVersion"] as? String,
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Thanks for taking a look at this but I'm not sure this is the way we want to go about it. I'll have to discuss this with the rest of the team before proceeding any further.

Since the current issue only affects Ipad due to the fact we didn't check the safe area values in the css for the "desktop version" of the app I think it would be simpler to modify this behaviour in CSS.

VIKTORVAV99:
Overwriting the node modules are also risky as it would most likely not work as the node modules are overwritten when a new version of capacitor is used. Even if it's a patch release so this way of doing it would not be a permanent fix.

Xeno-Tek:
I'm planning to fix this with CSS. Could you point me to the specific file/folder where the issue occurs?


Thanks

VIKTORVAV99:
The problem isn't in a specific file or folder, but a good place to start would be to see if you can change it in the App.tsx file But there might be changes required elsewhere as well.

PR #7290, opened 2024-10-07
ENTSOE: Add more production types and battery storage

torokati44:
Sources:

https://transparency.entsoe.eu/content/static_content/Static%20content/web%20api/Guide.html#_psrtype

image

https://documenter.getpostman.com/view/7009892/2s93JtP3F6#93160892-f305-43d8-80e7-545535250034

image

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -91,6 +91,11 @@ ENTSOE_PARAMETER_DESC = {
     "B18": "Wind Offshore",
     "B19": "Wind Onshore",
     "B20": "Other",
+    "B21": "AC Link",
+    "B22": "DC Link",
+    "B23": "Substation",
+    "B24": "Transformer",
+    "B25": "Energy storage",
 }
 ENTSOE_PARAMETER_BY_DESC = {v: k for k, v in ENTSOE_PARAMETER_DESC.items()}
 ENTSOE_PARAMETER_GROUPS = {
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -104,9 +109,12 @@ class WindAndSolarProductionForecastTypes(Enum):
         "oil": ["B06"],
         "solar": ["B16"],
         "wind": ["B18", "B19"],
-        "unknown": ["B20", "B13", "B15"],
+        "unknown": ["B20", "B13", "B15", "B21", "B22", "B23", "B24"],
```
VIKTORVAV99, line 112: This is a risky change, we currently don't know if AC link and DC link would duplicate the data we get from the exchange data.

We also don't know what Substation or Transformer mean in this context, if it just includes their consumption then unknown could become negative. Which would not be correct.

VIKTORVAV99 (commented):
Hi!

We already have a open draft PR that adds battery storage here: https://github.com/electricitymaps/electricitymaps-contrib/pull/7077

The reason we haven't merged it yet is because there is no reported data for any zone yet which means we can't test it or add a snapshot test for it yet. It should work but we want to be on the safe side.

As for the other data sources I don't think we want to add those, especially without more knowledge of what data is actually reported for each of them and if we get that data elsewhere already.

parsers/ENTSOE.py
torokati44: That's fair - I thought of these too, but hoped that either someone knows better than me, or that this change could be verified in a test deployment first (assuming there is any actual data coming in with these types from ENTSO-E).

torokati44:
> The reason we haven't merged it yet is because there is no reported data for any zone yet which means we can't test it or add a snapshot test for it yet. 

Ah, I see.

torokati44:
Closing in favor of https://github.com/electricitymaps/electricitymaps-contrib/pull/7077.

Sorry for the spam, I should have checked beforehand... :man_facepalming:

parsers/ENTSOE.py
VIKTORVAV99: The best way would be to test this locally, using the `poetry test_parser <zone>` command.

But if the data is not added to the zone in question then it becomes impossible to test and if it suddenly changes later a bunch of weird things can happen.

PR #7292, opened 2024-10-07
ESIOS.py shifted arrow to datetime only

shing1Sks:
## Issue

for - Remove usage of arrow #6135 for ESIOS.py 

## Description

removed arrow and used datetime only

### Preview

tested with test_ESIOS.py

parser_ok

code_ok

### Double check

- [*] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [*] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.


parsers/ESIOS.py
```diff
@@ -90,9 +88,13 @@ def fetch_exchange(
 
         net_flow *= EXCHANGE_MULTIPLICATION_FACTOR_MAP[zone_key]
 
+        exchange_datetime = datetime.strptime(
+            value["datetime_utc"], "%Y-%m-%dT%H:%M:%SZ"
+        ).replace(tzinfo=TIMEZONE)
```
VIKTORVAV99, line 93: The timezone here should be UTC as the value sent from the server is in Zulu (Z).

I would also look into using `datetime.fromisofomat()` instead. But since we don't we are still on Python 3.10 it can't parse zulu directly and you'll have to replace the Z with `+00:00` which means the same thing but uses the extended ISO format.

VIKTORVAV99 (changes requested):

parsers/ESIOS.py
shing1Sks: so okay for the current python verion in use i need to manually replace the z with +00:00
fix

this seems better ?

VIKTORVAV99: Yeah it's simpler, then we can just remove the replacement when we upgrade our python version.

shing1Sks:
i have commited the new changes !

VIKTORVAV99 (approved):
LGTM! 🎉

I added a snapshot test as well to ensure it don't regress in the future.

PR #7294, opened 2024-10-08
iOS header overlaps with UI #6966 (Fixed)

Xeno-Tek:
### **iOS header overlaps with UI (Fixed):**
fix: resolve iOS header overlap by adding safe area insets to CSS

---

### **Issue:**

Closes [#6966](https://github.com/electricitymaps/electricitymaps-contrib/issues/6966)

---

### **Description:**

This PR resolves the issue where the iOS header (time, date, battery, etc.) overlaps with the UI when the app is viewed on iPads. The problem stems from the layout not considering the iOS-specific safe area insets.

- Added CSS safe area insets (`env(safe-area-inset-*)`) to ensure proper padding for iOS devices.
- Adjusted `padding-top`, `padding-bottom`, and side insets for better alignment and to prevent header overlap.
- Maintained the integrity of existing styles while applying the necessary layout fixes.

---

### **Preview:**

Please find below a screenshot demonstrating the resolution of the issue:

373942909-946e6ccf-916e-4ad6-96c6-3c046f8de167

---

### **Double Check:**

- [x] I have tested my changes on an iOS device (iPad).
- [x] I have verified that the safe area insets do not affect non-iOS devices.

### DIFF
--- a/parsers/IN.py
+++ b/parsers/IN.py
@@ -7,7 +7,6 @@ from logging import Logger, getLogger
 from typing import Any
 from zoneinfo import ZoneInfo
 
-import arrow
 import numpy as np
 import pandas as pd
 from bs4 import BeautifulSoup
@@ -22,7 +21,7 @@ from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.exceptions import ParserException
 
 IN_TZ = ZoneInfo("Asia/Kolkata")
-START_DATE_RENEWABLE_DATA = arrow.get("2020-12-17", tzinfo=IN_TZ).datetime
+START_DATE_RENEWABLE_DATA = datetime(2020, 12, 17, tzinfo=IN_TZ)
 CONVERSION_GWH_MW = 0.024
 GENERATION_MAPPING = {
     "THERMAL GENERATION": "coal",
@@ -498,7 +497,7 @@ def daily_to_hourly_production_data(
 
 
 def get_start_of_day(dt: datetime) -> datetime:
-    dt_localised = arrow.get(dt).to(IN_TZ).datetime
+    dt_localised = dt.astimezone(IN_TZ)
     dt_start = dt_localised.replace(hour=0, minute=0, second=0, microsecond=0)
     return dt_start
 


VIKTORVAV99:
@Xeno-Tek how does this look in the various iPhone versions?

Right now the map covers the whole screen, this is something we want to keep but just move foreground elements so they don't overlap.

Xeno-Tek:
### iPhone SE

**With Insets:**
iPhone SE with Insets

**Without Insets:**
iPhone SE without Insets

### iPhone 16 Pro Max

**With Insets:**
iPhone 16 Pro Max with Insets

**Without Insets:**
iPhone 16 Pro Max without Insets

Xeno-Tek:
@VIKTORVAV99 Alternatively I’ve also implemented a fix using media queries that specifically targets iPads without affecting iPhones or other devices. Below are the test results showcasing the impact of this solution:

- **iPad with media tags (code in index.css):**  
  Screenshot 2024-10-09 at 10 05 07 pm

- **iPad with media tags (landscape):**  
  Ipad with media tags

- **iPad without media tags (portrait):**  
  Ipad without media tags

- **iPad in horizontal orientation (with media tags):**  
  Horizontal Ipad with media tags

- **iPhone 16 Pro Max (unchanged by media tags):**  
  Iphone 16 Pro Max unchanged with media tags

- **iPhone SE (3rd generation, unchanged by media tags):**  
  Iphone SE 3rd generation unchanged by media tags

Please let me know which solution works best for your needs, and I'll be happy to create a new PR accordingly.

VIKTORVAV99:
The problem is that the map background for the top is also removed, we do not want this. We only want to move the titles and buttons we so they don't overlap.

Just turning them gray wastes a lot of background space and do not fit into our current or future design plans.

Xeno-Tek:
@VIKTORVAV99 Would this solution work? I turned off the status bar, which allows you to see the background of the map at the top, without losing any important details. Here are the test results:

- **iPad (10th generation) without status bar (landscape):**  
  

- **iPad (10th generation) without status bar (portrait):**  
  

- **iPhone 16 Pro Max without status bar:**  
  

Let me know if this approach aligns with your design preferences, and I'll proceed accordingly!

VIKTORVAV99:
> @VIKTORVAV99 Would this solution work? I turned off the status bar, which allows you to see the background of the map at the top, without losing any important details. Here are the test results:
> 
> * **iPad (10th generation) without status bar (landscape):**
>   
> * **iPad (10th generation) without status bar (portrait):**
>   
> * **iPhone 16 Pro Max without status bar:**
>   
> 
> Let me know if this approach aligns with your design preferences, and I'll proceed accordingly!

@Alportan how do you feel about this?

Alportan:
> > @VIKTORVAV99 Would this solution work? I turned off the status bar, which allows you to see the background of the map at the top, without losing any important details. Here are the test results:
> > 
> > * **iPad (10th generation) without status bar (landscape):**
> >   
> > * **iPad (10th generation) without status bar (portrait):**
> >   
> > * **iPhone 16 Pro Max without status bar:**
> >   
> > 
> > Let me know if this approach aligns with your design preferences, and I'll proceed accordingly!
> 
> @Alportan how do you feel about this?

I wouldn't hide the status bar completely. My go-to would be the second suggestion on using media queries that specifically targets iPads without affecting iPhones or other devices. 

This would fix the issue that's only visible on iPads, and keep all the other use-cases the same as they are today! What do you think @VIKTORVAV99 @Xeno-Tek? 🙌

Xeno-Tek:
@Alportan @VIKTORVAV99, I agree that the second option is the way to go. It provides flexibility for future changes and ensures we’re targeting iPads specifically without affecting other devices. Also, using media queries will make it easier to adjust the design later if the aspect ratios change.

VIKTORVAV99:
I'd be fine with using media queries but would preferably want to keep them in TailwindCSS classes instead of the css file as we want to have a little as possible in that.

Issue #7270, opened 2024-10-01
Missing exchange configuration between South Africa (ZA) and Lesotho (LS)


VIKTORVAV99:
## Description

Lesotho have a interconnector with South Africa and rely on them for a large part of their electricity. It would be nice to get the config added as a first step in getting this exchange on the map.

u7262835:
Hi ElectricityMaps Team, 

I am interested in working on this issue. It's my first time contributing to this project, and I think it would be a great opportunity for me to learn how the data-parsing works (whilst helping the project in the process). 

I have looked at other successfully-completed issues and pull-requests (such as Issue #7123 and Pull Request #7238) in an attempt to understand what these types of issues involve, and how the connections and parsers are implemented. 

What would I need to do/who would I need to contact to be assigned to this issue, and is there any additional information about this issue (or general exchange parser implementations) that I should be aware of? 

Thanks in advance!

VIKTORVAV99:
Hi!

This initial issue was only really meant to add the exchange configuration to the app so we can inform people that there is a exchange but we are missing the data.

But if you want to try and build a parser for it you are very welcome to do so, I would like that to be a separate PR though.

> What would I need to do/who would I need to contact to be assigned to this issue, and is there any additional information about this issue (or general exchange parser implementations) that I should be aware of?

This is the hard part, finding the data, I don't have a data source I can link to but for ZA data we are using data from [eskom.co.za](http://eskom.co.za/) so that would be a good start.

As for the parser implementation itself that will be different depending on what data can be found. But generally it includes transforming the data to our format and using our parser classes to get some general validation of the data. Information around this can be found here: https://github.com/electricitymaps/electricitymaps-contrib/wiki/Building-a-new-parser

u7262835:
Perfect, thank you very much for your speedy and thorough reply Viktor!

I will start by attempting the exchange configuration in the coming days, and after that is completed (if it is approved as an appropriate solution) I will have a deeper look into data sources and parsing. 

Thanks once again!

VIKTORVAV99:
Sounds like a plan!

Thanks for taking on this issue!

u7262835:
Hi Viktor (or other maintainers),

What's the best way to locally verify that the config file that I've created is correct? Is there an easy way to locally view the exchange configuration on a non-live map, or another preferred method of verification (or for changes such as the config files, is it just confirmed by the reviewing developer in the pull request)?

[...]

PR #7298, opened 2024-10-08
Added exchange file for Lesotho-South Africa.

u7262835:
## Issue

This pull request resolves Issue #7270 (Missing exchange configuration between South Africa (ZA) and Lesotho (LS)). 

## Description

This pull request adds an exchange configuration file (LS_ZA.yaml) containing the longitude, latitude, and rotation of the South Africa -> Lesotho exchange. The location of this exchange is at the Maseru border in accordance with the Southern African Power Pool grid map (https://www.sapp.co.zw/) and the Lesotho Electrification Master Plan Report (https://nul-erc.s3.amazonaws.com/public/documents/reports/lesotho-electrification-master-plan-grid-1532184999.pdf, pg. 9). 

No capacity information (or live, parse-able data) was found for this exchange. I will continue to search for this, and update it in a new pull request if found (however this seems unlikely for the near future, as the only free information found to date was static and outdated). 

### Preview

There are no applicable preview screenshots. The LS_ZA.yaml file should be reviewed by a maintainer/experienced ElectricityMaps developer for correctness before accepting the merge to confirm that it will create an exchange at an appropriate location (@VIKTORVAV99 suggested that he can do this: https://github.com/electricitymaps/electricitymaps-contrib/issues/7270#issuecomment-2399018342)). 

### Double check

- [x] I have tested my parser changes locally with `poetry run test_parser "zone_key"` (N/A)
- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes. (N/A)


u7262835:
@VIKTORVAV99 requesting a review regarding Issue #7270.

config/exchanges/LS_ZA.yaml
```diff
@@ -0,0 +1,4 @@
+lonlat:
+  - 27.454639
+  - -29.297917
+rotation: 87
```
VIKTORVAV99, line 4:
```suggestion
rotation: -87
```
I think we just need to rotate this. Otherwise it looks good to me.

u7262835: The exchange is intended to go from South Africa (source) into Lesotho (destination) (on the map, that will appear as going left-to-right with a slight angle, hence the 87deg instead of 90deg). 

Based on the other exchange configurations (such as South Australia -> Victoria, which goes left-to-right), a positive rotation value would direct it this way. Similarly, a negative value (such as Great Britain -> Ireland) would go right-to-left. 

Have I misunderstood the intended exchange direction and/or how the rotation values work?

Thanks again for your help, and please let me know if you need any more info from me to help clarify this!

VIKTORVAV99: No you have not, but the exchange direction is defined alphabetically so currently the exchange is going from LS into ZA.

But the rotation currently shows the opposite.

u7262835: Ahhh ok, thanks for clarifying! 

Should I rename the file to ZA_LS.ymal, or should I keep this file (and correct the rotation) and add a new file for ZA_LS.yaml?

u7262835: As far as I know there are no exchanges from Lesotho into South Africa, so I don’t think a LS-ZA exchange config is necessary.

u7262835: Thanks for the clarification Viktor; that makes sense.

I’ve updated the rotation and aligned this branch with the main branch, so the pull request should be ready for review again.

Thanks once again for your help, I really appreciate it!

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for adding it!

PR #7299, opened 2024-10-08
Avoid sending token

corradio:
## Issue
The electricity map secret token should never be sent in the header, but should only be used to sign the request.
I recommend rotating the secret.

Tested locally against our backend using the weather endpoint, bypassing cloudflare cache.


tonypls (approved):
LGTM, I've created a task to rotate.  Not 100% sure why this was being sent plain text.

PR #7300, opened 2024-10-08
Fix wind scale

corradio:
## Issue
The wind scale shown was not the same as the one computed as one file had a max set to 15 and the other to 30. Now both are consistent.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7301, opened 2024-10-08
fix(SEO): re-add the zone title

VIKTORVAV99:
## Issue

Not sure what happened but it seems like https://github.com/electricitymaps/electricitymaps-contrib/pull/7240 removed the Title from the zone header.

## Description

Re adds the title.

### Double check


cadeban (approved):
LGTM!

cadeban:
>Not sure what happened but it seems like https://github.com/electricitymaps/electricitymaps-contrib/pull/7240 removed the Title from the zone header.

So sorry about that! I think it might have happened during a merge conflict 😬

PR #7302, opened 2024-10-09
Feat: New Feature Popover

cadeban:
## Description
[AVO-554](https://linear.app/electricitymaps/issue/AVO-554/whats-new-tooltip): We have a lot of planned changes for our app, and we need a way to let people know about all of our cool new features.

Changes:
- Adds New Feature Popover component
  - Only supports 1 popover at a time
  - This component supports a portal version and a non-portal version
- Adds popover content and atomsWithStorage for Date Picker feature (not finalized)

Todo:
- [x] Dark Mode 🕶️ 
- [x] more tests
- [ ] storybook

Open for discussion:
- How do we want to handle multiple? Changelog, walkthrough (with max # of popovers)? -> just 1
- Do we want modal background? -> probably not

### Preview
Date Picker example (portal):
|Desktop view|Mobile view|
|---|---|
|Screenshot 2024-10-09 at 11 04 10|Screenshot 2024-10-09 at 11 04 37|

Future Price example (non-portal):
|Desktop view|Mobile view*| 
|---|---|
|Screenshot 2024-10-09 at 11 41 08|Screenshot 2024-10-09 at 11 41 37|

*Note how the in the non-portal version the popover is behind the TimeController

Dark mode:
Screenshot 2024-10-11 at 16 26 33

### Double check


cadeban:
@tonypls This PR still needs some tests, but you can take a look at the code whenever you have time

web/src/locales/en.json
```diff
@@ -52,6 +52,10 @@
     "preliminary-data": "Preliminary data may change over time.",
     "summary": "Discover real-time electricity insights with the Electricity Maps app!"
   },
+  "new-feature-popover": {
+    "title": "Select a Date to Explore",
+    "content": "Use the date picker to view historical data on electricity sources, CO2 emissions, and prices."
+  },
```
silkeholmebonnen, line 58: Do you think we should add another layer in this if we will add more feature popovers in the future? 
Maybe something like this:

```suggestion
  "new-feature-popover": {
    "date-picker": {
      "title": "Select a Date to Explore,"
      "content": "Use the date picker to view historical data on electricity sources, CO2 emissions, and prices."
    },
  },
```

web/src/features/time/TimeHeader.tsx
```diff
@@ -1,32 +1,44 @@
...
-            />
-          }
-          type="success"
-        />
+      {allLoaded && (
+        <NewFeaturePopover
+          side="top"
+          content={<NewFeaturePopoverContent />}
+          portal={false}
+        >
```
silkeholmebonnen, line 30: This should not be showing until we actually implement it. Meanwhile, should we just add it in a FF or what do you think? 

Also, do you think we should add an end date for when it should not show up for any users anymore? e.g a month after the release of the new feature?

web/src/features/time/TimeHeader.tsx
```diff
@@ -1,32 +1,44 @@
...
 
 export default function TimeHeader() {
   const { t, i18n } = useTranslation();
   const timeAverage = useAtomValue(timeAverageAtom);
   const selectedDatetime = useAtomValue(selectedDatetimeIndexAtom);
-  const { isLoading } = useGetState();
+  const { isLoading: isLoadingData } = useGetState();
+  const [isLoadingMap] = useAtom(loadingMapAtom);
+
+  const allLoaded = !isLoadingMap && !isLoadingData;
```
silkeholmebonnen, line 18: Why is this needed when it wasn't before?

If it is needed, then we could use useAtomValue to match the other atom values in the component

web/src/components/NewFeaturePopover/NewFeaturePopoverContent.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function NewFeaturePopoverContent() {
+  const { t } = useTranslation();
+
+  return (
+    <div className="flex flex-col text-left">
+      <div className="flex flex-row items-center gap-2">
+        <CalendarSearch size={16} />
+        <h3>{t(`new-feature-popover.title`)}</h3>
+      </div>
+      <p>{t(`new-feature-popover.content`)}</p>
```
silkeholmebonnen, line 13: Depending on how we decide to do the translation (see en.json comment), then it might be necessary to give the translation strings as props

silkeholmebonnen (commented):
Looks really good! Just a couple of questions and comments

web/src/features/time/TimeHeader.tsx
cadeban: At the time I originally wrote this, the popover was displaying over the loading screen. However, that no longer seems to be an issue, so I'll remove the isLoadingMap check

cadeban:
Will make a follow up PR to automatically dismiss the popover if someone clicks elsewhere on the app

VIKTORVAV99:
@cadeban to fix the lockfile just merge master and run pnpm install before committing it to git. PNPM has built in merge resolution for the lockfiles :)

tonypls:
Looks great, can the little arrow point to the back chevron? Encourage users to the interaction is the first step in interacting with this feature
Screenshot 2024-12-02 at 11 53 27 AM

tonypls (approved):
LGTM!

Be nice to get this out and follow up with the small tweaks!

PR #7304, opened 2024-10-09
chore(translations): remove languages that have a low amount of translated keys

VIKTORVAV99:
## Issue

We have a few languages that don't have enough translations, since we don't want to have a mixed experience we need to remove them for now.

## Description

Removes the 3 languages with the lowest amount of keys translated that we we can't translate ourselves.

### Double check


silkeholmebonnen (approved):
LGTM

PR #7306, opened 2024-10-10
chore(translation-files): remove escaped quotes

silkeholmebonnen:
## Issue

[AVO-336](https://linear.app/electricitymaps/issue/AVO-336/remove-escaped-quotes-from-translation-files)

## Description

This PR adds a script to remove escaped quotes and replace them with single quotes. This script has been used to remove all escaped quotes from translation files.

### Double check


silkeholmebonnen:
Should I add this script to the CI?

VIKTORVAV99:
> Should I add this script to the CI?

I would try and add it as a part of the existing scripts we have, like the syncTranslations script and the associated check. 

So we don't introduce these in the future as well.

web/src/locales/cs.json
```diff
@@ -94,41 +94,41 @@
...
-    "divideExistingArea-answer": "Určitě, když existují separátní zdroje dat pro každou část dané oblasti. Můžete <a href=\"#contribute\" class=\"entry-link\">nám pomoci přidáním nových datových zdrojů</a>.",
+    "divideExistingArea-answer": "Určitě, když existují separátní zdroje dat pro každou část dané oblasti. Můžete <a href='#contribute' class='entry-link'>nám pomoci přidáním nových datových zdrojů</a>.",
     "seeHistoricalData-question": "Můžu zobrazit historii oblasti starší než 24 hodin?",
-    "seeHistoricalData-answer": "Můžete si zakoupíte přístup do <a href=\"https://electricitymaps.com?utm_source=app.electricitymaps.com&utm_medium=referral\" target=\"_blank\">databáze historických dat</a>."
+    "seeHistoricalData-answer": "Můžete si zakoupíte přístup do <a href='https://electricitymaps.com?utm_source=app.electricitymaps.com&utm_medium=referral' target='_blank'>databáze historických dat</a>."
   },
   "methodology": {
     "groupName": "Metogologie",
-    "carbonIntensity-question": "Co je \"uhlíková intenzita\"?",
+    "carbonIntensity-question": "Co je 'uhlíková intenzita'?",
```
VIKTORVAV99, line 120: IDK about the grammar in various languages but do you think this could be a problem?

It will be significantly harder to fix if it is but I also don't want to make something "wrong" in languages just because we didn't know about it.

VIKTORVAV99 (commented):
Looks good as far as I can tell.

I just think we need to figure out the implications of doing this inside the translated strings themselves or if we should only do it inside HTML elements. Maybe double check with Tony or Alex if they have any insight into the topic.

But personally it's not a blocker for me.

web/src/locales/cs.json
silkeholmebonnen: I don't know if it will be a problem. @tonypls or @Alportan do have any knowledge about this?

tonypls: No idea! I think it should be okay to proceed and see if we get any feedback on the languages we're not sure about

silkeholmebonnen:
It does not work as I hoped. Do you know a smart way of committing and pushing changes in the CI @VIKTORVAV99 ?

VIKTORVAV99:
> It does not work as I hoped. Do you know a smart way of committing and pushing changes in the CI @VIKTORVAV99 ?

You can't really do that. There is a "work around" by using pull request target instead (which would work here as it's not using any secrets) but then we have to split out this CI script from the rest of the CI jobs.

The answer here describes it better than I could: https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

It might be possible to just tweak the permissions of this job but as it notes pull_request_target runs from the base branch so bad actors can use a PR to modify the script and mine crypto instead (that has happened, but not to us). But that means this PR would have to be merged before it works so it would not be easy to test it before unfortunately.

silkeholmebonnen:
Okay do you think I should just modify the script so that it just checks if there is an escaped quote and if there is then you have to run the script yourself, like we do with the 'Validate Translation Files' step?

VIKTORVAV99:
> Okay do you think I should just modify the script so that it just checks if there is an escaped quote and if there is then you have to run the script yourself, like we do with the 'Validate Translation Files' step?

Yeah I think that would be a good. I don't think we need the complexity of fully automating it.

silkeholmebonnen:
@VIKTORVAV99 this is ready for review now:)

scripts/remove_escaped_quotes_in_locales.py
VIKTORVAV99, line 1: Only thing to consider is if we want to keep this as a python script in the future or simple js script. But I don't consider this blocking.

Hacktoberfest did make me consider how we can simplify our tooling a bit, that's all :)

VIKTORVAV99 (approved):
LGTM! 🎉

Nothing blocking and it seems to be working as expected!

VIKTORVAV99:
Nice work!

scripts/remove_escaped_quotes_in_locales.py
silkeholmebonnen: I don't think it's worth spending time on right now that this is working. But its a fair point:)

PR #7313, opened 2024-10-14
feat(web): show breakdown chart in historical consumption views

tonypls:
## Issue
Breakdown chart isn't visible in historical consumption mode
AVO-578

## Description
This PR updates the breakdown chart to show in all views with a disclaimer of the data that is not included.


cadeban:
Probably want to fix this typo:
Screenshot 2024-10-14 at 11 31 04

cadeban (approved):
Approved on condition that we fix the typo ;)

PR #7314, opened 2024-10-14
feat(web): hide consumption toggle behind ff

tonypls:
## Issue
Consumption / production toggle can be unnecessary complexity. 
AVO-590

## Description
This PR uses a new feature flag `consumption-only` to conditionally hide the consumption production toggle on mobile and desktop.

This PR #7313 should mean all information is present, event without the toggle.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -8,15 +8,18 @@ import { AppStoreBanner } from 'components/AppStoreBanner';
 import LoadingOverlay from 'components/LoadingOverlay';
 import { OnboardingModal } from 'components/modals/OnboardingModal';
 import ErrorComponent from 'features/error-boundary/ErrorBoundary';
+import { useFeatureFlag } from 'features/feature-flags/api';
 import Header from 'features/header/Header';
 import UpdatePrompt from 'features/service-worker/UpdatePrompt';
 import { useDarkMode } from 'hooks/theme';
 import { useGetCanonicalUrl } from 'hooks/useGetCanonicalUrl';
+import { useSetAtom } from 'jotai';
 import { lazy, ReactElement, Suspense, useEffect, useLayoutEffect } from 'react';
 import { Helmet } from 'react-helmet-async';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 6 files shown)


web/src/features/map-controls/ZoomControls.tsx
```diff
@@ -1,12 +1,15 @@
+import { useFeatureFlag } from 'features/feature-flags/api';
 import { ReactElement } from 'react';
 import { NavigationControl } from 'react-map-gl/maplibre';
 
 export default function ZoomControls(): ReactElement {
+  const isConsumptionOnlyMode = useFeatureFlag('consumption-only');
+  const marginTop = isConsumptionOnlyMode ? 54 : 98;
```
cadeban, line 7: QQ: Where do these numbers come from?

web/src/App.tsx
```diff
@@ -46,6 +49,14 @@ export default function App(): ReactElement {
   const shouldUseDarkMode = useDarkMode();
   const { t, i18n } = useTranslation();
   const canonicalUrl = useGetCanonicalUrl();
+  const setConsumptionAtom = useSetAtom(productionConsumptionAtom);
```
cadeban, line 52: What's the plan with this atom? It doesn't seem like we're using it anywhere

tonypls: It's used here `setConsumptionAtom(Mode.CONSUMPTION);` just below. 

It's here so if a user has production mode stored in local storage and we change the feature flag it will set this to production

cadeban: Ahh gotcha gotcha 👍

cadeban (approved):
🚀

web/src/features/map-controls/ZoomControls.tsx
VIKTORVAV99: I think it's the absolute positioning the zoom buttons use. These comes from the Maplibre css file and the maplibre built-ins. I tried to fix it in #6974 but then broke the wind layer instead (which should be unrelated but apparently is not).

PR #7315, opened 2024-10-14
AVO-588: Match share/copy icon to functionality

cadeban:
## Description
[AVO-588](https://linear.app/electricitymaps/issue/AVO-588/change-share-button-to-copy-button-when-on-desktop): Match share/copy icon to functionality

### Preview
Desktop:
Screenshot 2024-10-14 at 13 11 54

Android:
Screenshot 2024-10-14 at 13 12 23

iPhone:
Screenshot 2024-10-14 at 13 12 11

### Double check

### DIFF
--- a/web/src/features/panels/zone/ShareButton.cy.tsx
+++ b/web/src/features/panels/zone/ShareButton.cy.tsx
@@ -1,21 +1,20 @@
 import { Capacitor } from '@capacitor/core';
 import { ToastProvider } from '@radix-ui/react-toast';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { s } from 'vite/dist/node/types.d-aGj9QkWt';
 
 import { ShareButton } from './ShareButton';
 
 const queryClient = new QueryClient();
 
-describe.skip('Share Button', () => {
+describe('Share Button', () => {
   beforeEach(() => {
     cy.intercept('/feature-flags', {
       body: { 'share-button': true },
     });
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

Is the plan still to make our own share menu later on? Utilizing our X, Linkedin, Facebook and maybe Reddit buttons?

cadeban:
@VIKTORVAV99 Thanks for the speedy review!

I think we still plan on adding the share menu at some point, but right now it's not planned for this cycle since we have all of these other items that seem to have greater value add for storytelling in the app

VIKTORVAV99:
> @VIKTORVAV99 Thanks for the speedy review!
> 
> 
> 
> I think we still plan on adding the share menu at some point, but right now it's not planned for this cycle since we have all of these other items that seem to have greater value add for storytelling in the app

Great then we are on the same page! 🚀

tonypls (approved):
Nice! LGTM 🎉

cadeban:
@VIKTORVAV99 Quick update: after some discussion with Oli, this button in the zone header might be duplicative if use permalinks to the charts instead

VIKTORVAV99:
> @VIKTORVAV99 Quick update: after some discussion with Oli, this button in the zone header might be duplicative if use permalinks to the charts instead

Not sure I agree on that, there should probably be a way to share the entire zone as well, but perhaps it don't need to be the green pop version.

PR #7316, opened 2024-10-14
feat(web): zone specific datetimes

tonypls:
## Issue
AVO-586
It's not always valuable to have a zone date time formatted for a users location

## Description
This PR updates the date formatting to convert formatted dates to be zone specific, if no zone is selected it should revert to the users location. 

There were a few missing timezones in configs so I've also added those.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -539,7 +539,6 @@ jobs:
           restore-keys: ${{ runner.os }}-venv-poetry_${{steps.get-poetry-version.outputs.poetry-version}}
       - name: Install libxml2-dev and tesseract-ocr
         run: |
-          poetry env use 3.10
           sudo apt-get update
           sudo apt-get install libxml2-dev tesseract-ocr tesseract-ocr-eng
       - name: Install Poetry dependencies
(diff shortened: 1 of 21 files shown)


.github/workflows/CI.yml
```diff
@@ -539,6 +539,7 @@ jobs:
           restore-keys: ${{ runner.os }}-venv-poetry_${{steps.get-poetry-version.outputs.poetry-version}}
       - name: Install libxml2-dev and tesseract-ocr
         run: |
+          poetry env use 3.10
```
VIKTORVAV99, line 542: Is this really needed? Don't we specify the python version above?

tonypls: No but why does a python test workflow fail 😢

VIKTORVAV99: Hmm I can try removing the cache and see if that helps.

cadeban:
Do we provide any sort of disclaimer about the data delay or will users come to the app and expect to see current time?
Screenshot 2024-10-14 at 14 28 11

tonypls:
> Do we provide any sort of disclaimer about the data delay or will users come to the app and expect to see current time? Screenshot 2024-10-14 at 14 28 11

I think because the delay is unchanged we don't need to make a disclaimer right now. Hopefully we reduce the delay soon 🚀

cadeban:
And what was the verdict on `GMT-X`?

cadeban (approved):
LGTM!

cadeban:
Whoops, just seeing [Viktor's feedback](https://electricitymaps.slack.com/archives/C031YV8M940/p1728912933565289) in the channel now - maybe wait for his approval too

VIKTORVAV99 (approved):
LGTM! 🎉

Seems like my syncing issues are unrelated to this.

PR #7324, opened 2024-10-15
[WIP] AVO-592: Scroll to chart

cadeban:
## Description

[AVO-592](https://linear.app/electricitymaps/issue/AVO-592/scroll-to-chart-functionality)

### Double check

### DIFF
--- a/web/src/features/panels/LeftPanel.tsx
+++ b/web/src/features/panels/LeftPanel.tsx
@@ -120,16 +120,38 @@ export default function LeftPanel() {
     <OuterPanel>
       <Routes>
         <Route path="/" element={<HandleLegacyRoutes />} />
-        <Route
-          path="/zone/:zoneId"
-          element={
-            <ValidZoneIdGuardWrapper>
-              <Suspense fallback={<LoadingSpinner />}>
-                <ZoneDetails />
-              </Suspense>
-            </ValidZoneIdGuardWrapper>
-          }
-        />
+        <Route path="/zone/:zoneId">
+          <Route
+            index
+            element={
+              <ValidZoneIdGuardWrapper>
+                <Suspense fallback={<LoadingSpinner />}>
+                  <ZoneDetails />
+                </Suspense>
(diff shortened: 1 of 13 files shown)


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -99,33 +92,87 @@ export default function ZoneDetails(): JSX.Element {
...
 
+export function ZoneDetailRoutes() {
+  return (
+    <Routes>
+      <Route index element={<ZoneDetailsContent />} />
+      <Route path={LeftPanelToggleOptions.ELECTRICITY} element={<ZoneDetailsContent />} />
+      <Route path={LeftPanelToggleOptions.EMISSIONS} element={<ZoneDetailsContent />} />
+    </Routes>
+  );
+}
```
VIKTORVAV99, line 109: Not sure how I feel about this. We got to be real careful when we structure these URLs as they will basically live forever.

If we add electricity/emissions directly to the url at the end of the path I'm unsure how that plays into the deeper work we need to do around the URL structure for historical linking and consumption/production modes.

We would also need to add it to the sitemap and ensure it works with the canonical URL properly.

web/src/features/panels/LeftPanel.tsx
```diff
@@ -121,7 +121,7 @@ export default function LeftPanel() {
       <Routes>
         <Route path="/" element={<HandleLegacyRoutes />} />
         <Route
-          path="/zone/:zoneId"
+          path="/zone/:zoneId/*"
```
VIKTORVAV99, line 124: instead of a wildcard I think we can define `production` and `emissions` as valid paths instead. It should be safer going forward. (if we decide to include this now)

web/src/features/panels/zone/ZoneDetails.tsx
cadeban: 💯. This draft PR was just to test some things out for the chart linking work. If we choose to pursue this avenue, then we would probably want to write up a proper brief.

PR #7334, opened 2024-10-18
fix: errors in Mexico CENACE parser

tuxity:
## Description

- fix not correctly identifying API error
- if trying to fetch current month, return previous month data (up to 6 months) as current month if there is not yet available data

### Preview

image

### Double check


VIKTORVAV99 (approved):
This is a great improvement to the parser itself but I've been doing some digging and I'm not sure why we are not getting any data for MX at a country level so I can't promise this will fix the issue.

Unfortunately we have don't have much time to look into this on a deeper level before new years as we have a lot of other things already planned + the holidays are coming up as well.

In any case thanks for fixing some of the issues :)

VIKTORVAV99:
Before merging could you take care of the formatting issue?

tuxity:
@VIKTORVAV99 I will try to improve this during the upcoming weekend, it's not perfect. They take time to release the information of previous month, during the current month and this partial fix. Could we send `0` to `-1` instead of the previous month values?

VIKTORVAV99:
> @VIKTORVAV99 I will try to improve this during the upcoming weekend, it's not perfect. They take time to release the information of previous month, during the current month and this partial fix. Could we send `0` to `-1` instead of the previous month values?

No we can not manually set the values, if they are all zero it wouldn't work anyway and negative values aren't even accepted.

The problem isn't returning the past months data but rather it does not seem to be getting into our system at all even with manually re-fetching the data. In the past this has meant either geoblocking or that the IP has been blocked (googles data center IP) so we'd have to find a way around that.

parsers/CENACE.py
```diff
@@ -128,13 +128,10 @@ def fetch_csv_for_date(dt, session: Session | None = None):
...
-    if (
-        "Content-Type" not in response.headers
-        or response.headers["Content-Type"] == "text/html"
+    if "Content-Type" not in response.headers or "text/html" in response.headers.get(
+        "Content-Type"
     ):
-        raise Exception(
-            f"Error while fetching csv for date {datestr}: No CSV was returned by the API. Probably the data for this date has not yet been published."
-        )
+        return None
```
VIKTORVAV99, line 134:
```suggestion
    if (
        "Content-Type" not in response.headers
        or response.headers["Content-Type"] == "text/html"
    ):
        return None
```
Otherwise it trows a type errors as it's trying to compare the literal string with `str | None`. Changing the fallback would probably also work.

VIKTORVAV99 (commented):
Looks good, but one change introduced a new type error.

Could you please revert that?

parsers/CENACE.py
tuxity: You meant the `raise` `return` ? Also, If I remember correctly `response.headers["Content-Type"]` is an array and can have more values, that's why I used a `in`

VIKTORVAV99: Pylance is reporting the following:
```
Operator "in" not supported for types "Literal['text/html']" and "str | None"
Operator "in" not supported for types "Literal['text/html']" and "None"
```
This was not a problem before.

tuxity: Let me take a look at that 🤔

tuxity: Done!

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7336, opened 2024-10-18
feat(shareability): Add Reddit share button

VIKTORVAV99:
## Description

Adds a share feature to the reddit button.

### Double check


VIKTORVAV99:
Found out that YCombinator / Hacker News also have a share URL that works in a similar way. Do we want to add that as well?

tonypls (approved):
LGTM, let's keep it as this for now and talk about adding hacker news during slack week

PR #7342, opened 2024-10-21
chore(translation-files): Change Consumption to Electricity on the country-panel toggle

Alportan:
## Issue

[AVO-603](https://linear.app/electricitymaps/issue/AVO-603/update-consumption-to-electricity-in-country-panel-toggle)

## Description

Update Consumption label to Electricity, now that we soft-killed the Production/Consumption toggle.

### Double check


Alportan:
I thought that if we are to release a new version of the app, this would help a bit, now that the consumption/production toggle is not there anymore. For this fix, I only updated `electricityconsumption` translation-key and left `electricityproduction` as it is for now. 

We can clean that later, or I can also do it now. It's just that it might make sense to keep it in case we want to revert showing the toggle (hopefully we won't have to 🤞).

web/src/locales/da.json
```diff
@@ -86,7 +86,7 @@
     "lowcarbon": "Lav CO₂",
     "renewable": "Vedvarende",
     "electricityproduction": "Produktion",
-    "electricityconsumption": "Forbrug",
+    "electricityconsumption": "El",
```
VIKTORVAV99, line 89: I know my danish is not great but should it not be: "Elektricitet"

```suggestion
    "electricityconsumption": "Elektricitet",
```

VIKTORVAV99 (commented):
One minor question/suggestion otherwise it looks good I think.

Maybe nl.json is missing as well?

web/src/locales/da.json
Alportan: Just had a quick chat with Mathias, he recommends either "Elektricitet" or "Strøm". Let's go with the first and update it later in case the other is more appropriate. 🙌

VIKTORVAV99: I think elektricitet works best as that is the one we use in the related languages (mainly Swedish and German).

Alportan:
> One minor question/suggestion otherwise it looks good I think.
> 
> Maybe nl.json is missing as well?

Good point, I'll add that as well! 💪

VIKTORVAV99:
Looks like there is a test depending on this value in some language. I'll update it / them after lunch :)

VIKTORVAV99:
Oh I forgot to do it 😬

Alportan:
Ahh, damn. the test is still failing? 🙈

VIKTORVAV99:
> Ahh, damn. the test is still failing? 🙈

It's finding some other element than the one we want that is hidden.
I'm taking a look at it right now :)

VIKTORVAV99:
Might have done a tiny bit of cleanup as well 😅

VIKTORVAV99:
I'll let @cadeban or @tonypls do the final review since I changed the code a tiny bit.

web/src/features/panels/zone/DisplayByEmissionToggle.tsx
```diff
@@ -1,21 +1,18 @@
-import ToggleButton from 'components/ToggleButton';
-import { useAtom, useAtomValue } from 'jotai';
+import ToggleButton, { ToggleButtonOptions } from 'components/ToggleButton';
+import { useAtom } from 'jotai';
 import type { ReactElement } from 'react';
 import trackEvent from 'utils/analytics';
 import { LeftPanelToggleOptions, TrackEvent } from 'utils/constants';
-import { displayByEmissionsAtom, isConsumptionAtom } from 'utils/state/atoms';
+import { displayByEmissionsAtom } from 'utils/state/atoms';
```
cadeban, line 6: Ooh are we ready to say goodbye to production for real?

web/src/components/ToggleButton.tsx
```diff
@@ -14,8 +14,13 @@ import { ReactElement, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { twMerge } from 'tailwind-merge';
 
+export type ToggleButtonOptions = Array<{
```
cadeban, line 17: 👍

web/src/components/ToggleButton.tsx
```diff
@@ -57,7 +62,7 @@ export default function ToggleButton({
             key={`group-item-${key}`}
             value={value}
             onClick={() => onToggle(value)}
-            data-test-id={`toggle-button-${dataTestId}`}
+            data-test-id={`toggle-button-${dataTestId ?? value}`}
```
cadeban, line 65: I think we'll have to remember to use quotes when referencing the id if it defaults to value and value has spaces etc.

cadeban (approved):
lgtm! Thanks @Alportan and @VIKTORVAV99!

web/src/features/panels/zone/DisplayByEmissionToggle.tsx
VIKTORVAV99: Since we changed it to electricity it don't make much sense to switch between "Electricity" and "Production".

web/src/components/ToggleButton.tsx
VIKTORVAV99: Hmm good point, should we remove the fallback and just declare the IDs directly even if they are the same value?

PR #7343, opened 2024-10-21
chore: Update CODEOWNERS to include teams

VIKTORVAV99:
## Description

Now that we have teams in github we should take advantage of that 🙂


VIKTORVAV99:
> Hmm, will this not notify the whole team? 🤔

It should just automatically assign one person for review since we have auto assignment on for Avo. So I hope not...

It will however notify the self-service team 😉

madsnedergaard (approved):
Let's try, then we can readjust later if required :)
(On a side-note we could in the future create sub-teams like `self-service/engineering` or something like that, so it doesn't notify the full team)

VIKTORVAV99:
@tonypls what do you think about this?

tonypls (approved):
LGTM, let's try it out!

PR #7344, opened 2024-10-22
fix(parsers):Add new gas plant to PA parser

annaib00:
## Issue
Capacity outliers detected on a regular basis for unknown mode. 

## Description
Gas power plants, Proyecto Gatún 1-3, are added to the MAP_THERMAL_GENERATION_UNIT_NAME_TO_FUEL_TYPE mapper, that maps unknown production to a specific mode. By specifying these plants as gas, their production will not be categorised as unknown. 

Not completely sure the unknown outliers were only due to these specific plants missing in the mapper as the parser does not support historic data.

https://sitr.cnd.com.pa/m/pub/gen.html
https://generadoragatun.com/ 

### Preview
Screenshot 2024-10-22 at 10 47 27

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7348, opened 2024-10-22
Remove arrow from IL parser

tay:
## Issue

Related issue: https://github.com/electricitymaps/electricitymaps-contrib/issues/6135.

## Description

Removes arrow library from Israel IL parser.

I'm not sure if this is a dead code pathway, because I can't access any of the URL's and the [Parser issue page](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Parser-issues) says both the consumption and price parser have no recent data.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

They started geo-blocking the data a while ago but this should be a safe change as far as I can tell.

PR #7309, opened 2024-10-11
feat(web): start fetching data for zone at historical datetime

tonypls:
## Issue
Allow fetching of historical hourly data from the front end

## Description
This PR adds support for a user to visit the app via a historical link and navigate through the data for a given historical period.

Testing instructions: Run the backend: https://github.com/electricitymaps/electricitymaps/pull/5723
Visit: http://localhost:5173/zone/SE-SE2/hourly/2024-10-20T08:00:00Z

PR #7349, opened 2024-10-22
feat(web): lift the routes to app

tonypls:
## Issue
As an enabler for #7309 we need to have context of the routes in the getState API call which is triggered at the top of the tree. 

## Description
This PR lifts our routes to app.tsx for we have context of the params across the app.


web/src/features/panels/LeftPanel.tsx
```diff
@@ -117,37 +61,12 @@ function OuterPanel({ children }: { children: React.ReactNode }) {
       } ${isOpen ? '' : '-translate-x-full'}`}
     >
       {isMobile && <MobileHeader />}
-      <section className="h-full w-full">{children}</section>
+      <section className="h-full w-full">
+        <Suspense fallback={<LoadingSpinner />}>
+          {isZonePath ? <ZoneDetails /> : <RankingPanel />}
+        </Suspense>
```
VIKTORVAV99, line 67: While this works would it not be better to use a outlet and pass the component down from the router?

https://reactrouter.com/en/main/components/outlet#outlet

web/src/App.tsx
```diff
@@ -113,7 +157,20 @@ export default function App(): ReactElement {
...
+                      path="/zone/:zoneId/:urlTimeAverage?/:urlDatetime?"
+                      element={
+                        <ValidZoneIdGuardWrapper>
+                          <LeftPanel />
+                        </ValidZoneIdGuardWrapper>
+                      }
+                    />
+                    <Route path="*" element={<LeftPanel />} />
+                  </Routes>
+                </Suspense>
```
VIKTORVAV99, line 173: Then this would be more declarative and we could avoid using the left panel twice.

web/src/App.tsx
```diff
@@ -31,20 +33,57 @@ const TimeControllerWrapper = lazy(() => import('features/time/TimeControllerWra
...
+  if (zoneId !== upperCaseZoneId) {
+    return <Navigate to={`/zone/${upperCaseZoneId}?${searchParameters}`} replace />;
+  }
+
+  // Handle legacy Australia zone names
+  if (upperCaseZoneId.startsWith('AUS')) {
+    return (
+      <Navigate to={`/zone/${zoneId.replace('AUS', 'AU')}?${searchParameters}`} replace />
+    );
+  }
```
VIKTORVAV99, line 74: Just going to ask this while we are at it:

Should we move this to a redirect rule in Cloudflare instead?

VIKTORVAV99 (commented):
Works as I expect but I have some questions about the logic behind it.

Did you happen to look into https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis? I and Cadence want to switch to using data routing so we can use the fancy data APIs.

tonypls:
> Works as I expect but I have some questions about the logic behind it.
> 
> Did you happen to look into https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis? I and Cadence want to switch to using data routing so we can use the fancy data APIs.

Good point! and thanks for the feedback, I'll close this one and let's revisit with a better approach.

PR #7352, opened 2024-10-24
fix: Update AU-LH timezone

VIKTORVAV99:
Updates the timezone for AU-LH, maybe sentry will stop bothering us now 😅


tonypls (approved):
LGTM

PR #7354, opened 2024-10-25
Reverse chart layer opacity on hover

cadeban:
## Description

Previously, the hourly electricity origin chart reduced the opacity of the highlighted data series, making it challenging for a user to visually isolate the relevant data. This PR reverses the that behavior.

### Preview

|Before|After|
|---|---|
|Screenshot 2024-10-25 at 16 14 52|Screenshot 2024-10-28 at 15 44 24|
| Screenshot 2024-10-25 at 16 15 00|Screenshot 2024-10-28 at 15 44 30|

Dark Mode:
Screenshot 2024-10-28 at 15 44 07

Handling 0s in origin data series:
Screenshot 2024-10-28 at 19 30 53

### Double check

### DIFF
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -103,7 +103,8 @@ function AreaGraphLayers({
         );
 
         const isLayerSelected = selectedLayerIndex && selectedLayerIndex === ind;
-        const shouldLayerBeSaturated = isLayerSelected === null || isLayerSelected;
+        const shouldLayerBeSaturated =
+          isLayerSelected === null || layers.length === 1 || isLayerSelected;
 
         return (
           <React.Fragment key={layer.key}>
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable unicorn/no-null */
 /* eslint-disable react/jsx-handler-names */
 import { area, curveStepAfter } from 'd3-shape';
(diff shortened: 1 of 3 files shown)


web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -100,12 +110,26 @@ function AreaGraphLayers({
...
+        if (showHoverHighlight && emphasizeStroke) {
+          stroke = isDarkModeEnabled ? 'white' : 'black';
+        }
+
         return (
           <React.Fragment key={layer.key}>
             <path
-              className={layers.length > 1 ? 'sm:hover:opacity-75' : ''}
+              className={
+                shouldLayerBeSaturated ? 'sm:hover:opacity-100' : 'sm:opacity-30'
```
VIKTORVAV99, line 128:
```suggestion
                shouldLayerBeSaturated ? 'sm:hover:opacity-100' : 'sm:opacity-50'
```

Maybe personal preference but this feels a bit too much on dark more. I think 50 is enough now that we have the stroke outline 🙂

VIKTORVAV99 (approved):
LGTM! 🎉

But I have a tiny suggestion 🙂

PS: This is awesome!

web/src/features/charts/elements/AreaGraphLayers.tsx
cadeban: I think that's reasonable. I'll run it by @Alportan too so we're all on the same page

cadeban: I think I'll go ahead and merge this in as-is. It's pretty easy to update if we want to change it after

PR #7356, opened 2024-10-29
chore(map): Map cleanup

VIKTORVAV99:
## Issue

Closes: AVO-611

## Description

Updates the bidding zone borders between the SE subzones to be more accurate, adds the channel islands and removes the NKR zone.

### Double check


config/zones/GG.yaml
VIKTORVAV99, line 1: The translations for this zone already exists.

config/zones/JE.yaml
VIKTORVAV99, line 1: Same for this one.

VIKTORVAV99:
> Looks good! Do we want any additional messaging for the channel islands?
> 
> 
> 
> Screenshot 2024-11-06 at 12 05 07
> 
> 

Not just yet, the server error is due to the fact that the backend don't have access to the config files yet.

Once it does it should simply say we don't have any data for it. I'll see if I can make this more graceful in the backend at some point.

VIKTORVAV99:
@tonypls or @madsnedergaard do you have any objections? Otherwise I'll merge this later today.

VIKTORVAV99:
Holding off on merging this until I can do some manual testing with it against the API, shouldn't be any problems but let's just make sure.

madsnedergaard (approved):
Looks good to me on API side! 🙌 

It changes a 400 to 404 for NRK and adds more zones to /zones, but that's not really a problem - and in the future the /zones endpoint should probably only return zones we actually have data for!

VIKTORVAV99:
Merging this on Monday just to be extra careful.

PR #7360, opened 2024-10-30
Annaboye/gmm 172 ned data spiky

annaib00:
## Issue
Spiky NL Data

## Description
The API has an itemsPerPage input, one item is one type for one datetime. We in total want 12 types. If we ex query 10 items per page, then sometimes the API repeats one type on the next page and sometimes completely skips one type on the next page. ItemsPerPage is now set as a multiple of 12, such that we get all data for a number of datetimes per page. 

The API fetches data for entire days, so we add + 1 day to the target_datetime in order to get data from today. However, when fetching data from today the latest data is incomplete and the API also sometime gives future data from certain types. Therefore, data with datetime before (now - 0.5 hr) is removed.

OBS! One preview from current Intramaps (spiky), one from quick visualization of new parser (not spiky)

### Preview
Screenshot 2024-10-30 at 14 57 05
Screenshot 2024-10-30 at 14 57 35

### Double check


parsers/NED.py
```diff
@@ -199,10 +202,18 @@ def fetch_production(
...
 
-    return NED_data.to_list()
+    NED_data_list = NED_data.to_list()
+
+    # Add lag to avoid using data that is not yet complete and remove "future" data
+    NED_data_list = [
+        x
+        for x in NED_data_list
+        if x.get("datetime") < (datetime.now(timezone.utc) - timedelta(hours=0.5))
+    ]
```
VIKTORVAV99, line 214: Instead of doing this check here and reconstructing and looping over the entire list can we just check this before we append the event?

annaib00: Yes, I think we could do that. But in this case, should an if statement be added in the case of forecast = True? I'm not sure how the forecasts work, but I suppose we don't want to not append datetimes ahead of now for forecasts?

VIKTORVAV99: Good point!

I think combining the statements in a single if statement and just adding "`and not forecasts`" would do the trick?

annaib00: Yes, let's do that

VIKTORVAV99 (approved):
LGTM! 🎉

Nice work!

Just a heads up, even if we merge this today we probably won't deploy it until Monday :)

PR #7364, opened 2024-10-31
AVO-619: Remove partially estimated tag from charts

cadeban:
## Description
[AVO-619](https://linear.app/electricitymaps/issue/AVO-619/remove-partially-estimated-tag-from-charts-and-show-it-only-for): Remove partially estimated tag from charts
- Removes `preliminary` and `estimated` badges from charts except for bar breakdown
- Keeps zone disclaimer and tooltip badges

TODO: Discuss internally - does this conflict with our stance on scientific integrity?

### Preview
Hourly:
Screenshot 2024-10-31 at 12 28 53
Screenshot 2024-10-31 at 12 28 38
Screenshot 2024-10-31 at 12 29 15

Aggregation:
Screenshot 2024-10-31 at 12 43 08
Screenshot 2024-10-31 at 12 42 51

### Double check

### DIFF
--- a/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
+++ b/web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
@@ -1,14 +1,16 @@
 import * as Portal from '@radix-ui/react-portal';
+import EstimationBadge from 'components/EstimationBadge';
 import { getOffsetTooltipPosition } from 'components/tooltips/utilities';
+import { useGetEstimationTranslation } from 'hooks/getEstimationTranslation';
 import { useHeaderHeight } from 'hooks/headerHeight';
 import { TFunction } from 'i18next';
 import { useAtom, useAtomValue } from 'jotai';
-import { X } from 'lucide-react';
+import { CircleDashed, TrendingUpDown, X } from 'lucide-react';
 import React, { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { ElectricityModeType, ZoneKey } from 'types';
 import useResizeObserver from 'use-resize-observer';
-import { Charts, TimeAverages } from 'utils/constants';
+import { Charts, EstimationMethods, TimeAverages } from 'utils/constants';
 import {
   displayByEmissionsAtom,
   isConsumptionAtom,
@@ -68,6 +70,12 @@ function BarBreakdownChart({
   const headerHeight = useHeaderHeight();
 
   const titleText = useBarBreakdownChartTitle();
+  const estimationMethod = currentZoneDetail?.estimationMethod;
+  const pillText = useGetEstimationTranslation(
+    'pill',
+    estimationMethod,
+    currentZoneDetail?.estimatedPercentage
+  );
 
   if (isLoading) {
     return null;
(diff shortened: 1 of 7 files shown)


web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -115,7 +123,16 @@ function BarBreakdownChart({
       <ChartTitle
         titleText={titleText}
         unit={graphUnit}
-        isEstimated={hasEstimationPill}
+        badge={
+          hasEstimationPill ? (
+            <EstimationBadge
+              text={pillText}
+              Icon={
+                estimationMethod === EstimationMethods.TSA ? CircleDashed : TrendingUpDown
```
VIKTORVAV99, line 131: Could we add the `FORECASTS_HIERARCHY` method here as well (and other places we use preliminary) since it's also a TSA model enhanced with solar and wind forecasts?

Maybe it should be a follow up PR though.

cadeban: Good suggestion - if it's a quick change then I'll include it in this one!

cadeban: We use `EstimationMethods.TSA` in [other places in the code]( https://github.com/search?q=repo%3Aelectricitymaps%2Felectricitymaps-contrib+EstimationMethods.TSA&type=code) as well. I assume that we'll want to check for `FORECASTS_HIERARCHY` in these places too for consistency?

VIKTORVAV99: Probably yeah. Not sure if they are all identical but if you want to change them in here I'll help go through them on a 1:1 basis.

cadeban: Ok, I assumed that all of the instances in the search that stored the results in a variable named something like `isTSA` or `tsaCount` should be omitted. LMK if that instinct is wrong.
Omitted instances:
- https://github.com/electricitymaps/electricitymaps-contrib/blob/c1b42b0ef7e6d02b074cb5f0d4c8131bdc9661d7/web/src/features/charts/graphUtils.ts#L218
- https://github.com/electricitymaps/electricitymaps-contrib/blob/c1b42b0ef7e6d02b074cb5f0d4c8131bdc9661d7/web/src/features/panels/zone/EstimationCard.tsx#L56

VIKTORVAV99: Not sure it's wrong but I'm also not sure we are doing that correctly as is.

IMO Forecast Hierarchy should be considered TSAv2, under the hood it's still based on TSA with some specific forecast based estimation for solar and wind (for now).

So the question is for the estimation card and graph utils if we should consider TSA and Forecast Hierarchy to be the same thing?

I see that the estimation card asks for feedback if it's TSA based but IMO it should also do that for Forecast Hierarchy, maybe even display the TSA card itself or a variant of it as that uses the dashed circle Preliminary icon.

VIKTORVAV99: I would be fine with merging this as is now and then do some follow up cleanup, having a specific card for TSA don't even make much sense IMO, not anymore.

cadeban: > IMO Forecast Hierarchy should be considered TSAv2

Ahh, that's helpful. I wasn't sure what the relationship was between the two

cadeban:
closing in favor of https://github.com/electricitymaps/electricitymaps-contrib/pull/7508

PR #7388, opened 2024-11-05
Add Origin Chart Source Toggles

cadeban:
## Description
Allow the user to toggle sources on the origin chart.

TODO:
- [ ] Discuss with Alex how it should/behave

### Preview
No interaction:
Screenshot 2024-11-05 at 10 38 33

Selected:
Screenshot 2024-11-05 at 10 38 58

Selected buttons + Hover (not sure how we want to handle this):
Screenshot 2024-11-05 at 10 39 08

Dark Mode:
Screenshot 2024-11-05 at 10 40 18

### Double check


VIKTORVAV99:
Tiny question with a big impact here:

How does this work with exchanges? 👀
We don't even have labels for them...

web/src/features/charts/BreakdownChart.tsx
```diff
@@ -51,6 +52,18 @@ function BreakdownChart({
   const { t } = useTranslation();
   const isHourly = useAtomValue(isHourlyAtom);
 
+  const [selectedData, setSelectedData] = useState<Record<string, boolean>>({});
+
+  // TODO(cady): is there a better way to reset if we switch between emissions & electricity panels?
+  useEffect(() => setSelectedData({}), [displayByEmissions]);
```
VIKTORVAV99, line 58: Event listeners maybe?

Should be one for onButtonPress on the emissions and electricity buttons.

VIKTORVAV99: You might have to use an atom instead of a state though, but I still think it would be preferable if possible.

cadeban:
bump @tonypls @VIKTORVAV99!

web/src/features/charts/ProductionSourceLegendList.tsx
```diff
@@ -1,26 +1,42 @@
...
+            key={index}
+            type="tertiary"
+            size="sm"
+            foregroundClasses={'text-xs font-normal text-neutral-600 dark:text-gray-300'}
+            backgroundClasses={isSourceSelected ? 'bg-neutral-400/10' : ''}
+            onClick={onClick}
+            icon={<ProductionSourceLegend key={index} electricityType={source} />}
+          >
+            {capitalizedLabel}
+          </Button>
```
VIKTORVAV99, line 39: It's kind of hard to see which are selected in based on the buttons alone.

I would consider adding the success green as an outline to the button if it's selected.

Example:
image

Solar and hydro are selected here but the contrast/difference between them are too low. Especially for someone that might not have perfect eye sight.

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -110,7 +111,8 @@ interface AreagraphProps {
   tooltip: (props: InnerAreaGraphTooltipProps) => JSX.Element | null;
   tooltipSize?: 'small' | 'large';
   formatTick?: (t: number) => string | number;
-  showHoverHighlight?: boolean;
+  isDataInteractive?: boolean;
+  selectedData?: SelectedData;
```
VIKTORVAV99, line 115: I love it when the name + type end up the same length!

VIKTORVAV99 (commented):
Logic wise this looks sound to me, but I would try and revise the styling on the selected buttons a bit so it's more clear that they are in fact selected.

web/src/features/charts/ProductionSourceLegendList.tsx
cadeban: That's a fair point. I agree, that is difficult to see. However I think we use a green outline to indicate that an element is focused, so I think we should use a different outline color instead. Here's my update:

Screenshot 2024-11-14 at 15 39 11
Screenshot 2024-11-14 at 15 42 05

For dark mode, I used the bg and outline colors of the TimeController toggle and added a darker outline for the light theme.

VIKTORVAV99: Not using green is a fair point, but this already looks a lot better!

I'd keep @Alportan in the loop but I don't think he'll have any objections to this approach, but maybe a suggestion or two.

Otherwise this LGTM from my side of things and I'll approve it but maybe wait just a tiny bit to merge it if Alex has any feedback?

VIKTORVAV99 (approved):
LGTM! 🎉

Will be a useful feature for all those SoMe posts!

web/src/features/charts/ProductionSourceLegendList.tsx
Alportan: Looks good to me! 🙌 

Nitpick: We can give more contrast to the selected label text, but I think it's already clear enough which sources are selected now 💪

cadeban: @Alportan what did you have in mind? Bold text? Black text/white text depending on theme?

Alportan: More like black/white text, to make it pop 🎈

cadeban: Better?
Screenshot 2024-11-15 at 10 55 17
Screenshot 2024-11-15 at 10 56 37

Alportan: Yeees 🤩 Very small change, but it already stand out a bit more 🙌

PR #7393, opened 2024-11-06
Contribution lifecycle

VIKTORVAV99:
## Description

Let's make things a bit more clear by adding how our code actually get's deployed!

Closes: AVO-622


VIKTORVAV99:
This has already been approved by Julien so if there is notting obvious that are missing it just needs an approval.

silkeholmebonnen:
> It's using mermaidjs, you might need the extension for it so it renders locally. GitHub has this built it :)

Do we want to mention this to the people who might view it in their editor?

VIKTORVAV99:
> > It's using mermaidjs, you might need the extension for it so it renders locally. GitHub has this built it :)
> 
> Do we want to mention this to the people who might view it in their editor?

Maybe I'll just add it to the recommended extensions, but I don't think we need to mention it in the text. But it would be quite easy to add something small if we wanted to.

silkeholmebonnen (approved):
LGTM

Issue #7386, opened 2024-11-04
[Data Issue]: Netherlands: Unknown all mapped to geothermal, when it's actually a mix of various technologies?


q--:
### When did this happen?

Observed it today

### What zones are affected?

NL

### What is the problem?

So, current  NL Geothermal production is ~220MW [according to ElectricityMaps](https://github.com/user-attachments/assets/0b4cece5-8965-4127-a4b5-afffcc6dd754), which corresponds with the [production of 'other' at the data source NED.nl](https://github.com/user-attachments/assets/12ee3528-b061-4515-8746-6c10b2e65e69). However, NED.nl makes it clear that this actually a mixture of sources, [mentioning](https://ned.nl/en/node/78) hydropower, biogas and biomass* specifically).

\* Must be dedicated biomass, as biomass co-firing  with coal is already accounted for

FelixDQ:
Thank you for reporting @q--. I took a quick look, and the only thing that we are mapping to geothermal is "Type 9", which they classify as Geothermal here: https://ned.nl/nl/handleiding-api. I'm using the Google translated webpage, so I might be misunderstanding the docs. What is your analysis? Our mapping can be seen here:

https://github.com/electricitymaps/electricitymaps-contrib/blob/61d7447385a3ee3df4f2fa380098d312e4ba4f38/parsers/NED.py#L23-L36

q--:
I don't know about their API (I did notice there's also a `HeatPump` field though, so they're probably supporting or planning to support non-electricity energy).

However, [according to Geothermie Nederland](https://geothermie.nl/wp-content/uploads/GNL-Factsheet-Elektriciteit-uit-aardwarmte.pdf "2022 factsheet"), there is no geothermal electricity production in the Netherlands. Their [locations list](https://geothermie.nl/geothermie/locaties-in-nederland/) also doesn't list any geothermal electricity installations, only heat.

So my suspicion is that this category is actually biogas, dedicated biomass and hydropower (and perhaps other energy sources not mentioned) instead of geothermal. 

(I'm assuming that the category reported as 'geothermal' by ElectricityMaps is the same as the [production of 'other' at the data source NED.nl](https://ned.nl/en/node/78) because they produced the same amount of MW when I checked. Alternatively, it might've been a fluke that they were identical and `9` is actually geothermal heat or something. In any case, I don't think it's geothermal electricity because I don't know of any geothermal electricity plant(s) in the Netherlands.)

[...]

PR #7394, opened 2024-11-11
fix(parsers):Remove geothermal from NL parser

annaib00:
## Issue
Fix to issue #7386. Removing geothermal (type 9 in https://ned.nl/nl/handleiding-api) production from NL parser, as NL does not have any geothermal production. It is not added to unknown as we are already reporting more production than Eurostat and it could be geothermal heat production. 

### Double check


VIKTORVAV99 (approved):
Changes look good!

However do we have a plan for removing the geothermal data from the database?

It will not get overwritten as this will now return null values (not 0).

annaib00:
Ah, okay! No, we do not have a plan for that currently. Do you know if there are any nice way of doing this? Or should I just manually add a Geothermal value = 0 production i the parser?

VIKTORVAV99:
> Ah, okay! No, we do not have a plan for that currently. Do you know if there are any nice way of doing this? Or should I just manually add a Geothermal value = 0 production i the parser?

I'll leave this up to @FelixDQ to decide, we have in the past been hesitant to hardcode values as it increases maintenance and risks being wrong in the future. But on the other hand I don't think the Netherlands will add geothermal electricity generation.

Otherwise you'll have to write a database migration to remove them.

annaib00:
Okay, @FelixDQ is on vacation this week so I will look more into it when he is back :)

PR #7396, opened 2024-11-11
fix(parsers): Fix NED parser after geothermal removal

annaib00:
## Issue
Fix NED parser after geothermal removal so we don't get spiky data again. 
To ensure we get all data types for each hour the items per page has been updated so it's not a hardcoded value. 

If data from one datetime is split into several pages the API sometimes skip a type or repeats a type, so we want the itemPerPage to be a multiple of the types we are fetching. 

### Double check

### DIFF
--- a/parsers/NED.py
+++ b/parsers/NED.py
@@ -82,18 +82,23 @@ def _kwh_to_mw(kwh):
     return round((kwh / 1000) * 4, 3)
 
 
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

And nice removal of the hardcoded values!

PR #7397, opened 2024-11-12
feat(web): handle historcal linking with basic navigation

tonypls:
## Issue
View historical data with hourly granularity

## Description
Currently behind the feature flag `historical-linking`

This PR updates the state and zone requests to be able to query historical data at at hourly granularity and some basic navigation between time windows.

The navigation currently going back to the more recent utc 00:00 hour for the first click and subtracts 24 hours from there. 

In the spirit of deploying early and often I'd like to release this behind the feature flag and get it in the hands of our test users asap for feedback.

### DIFF
--- a/web/src/api/getState.ts
+++ b/web/src/api/getState.ts
@@ -1,6 +1,5 @@
 import type { UseQueryResult } from '@tanstack/react-query';
-import { useQuery, useQueryClient } from '@tanstack/react-query';
-import { useEffect } from 'react';
+import { useQuery } from '@tanstack/react-query';
 import { useParams } from 'react-router-dom';
 import type { GridState, RouteParameters } from 'types';
 import { TimeAverages } from 'utils/constants';
@@ -21,7 +20,9 @@ const getState = async (
     getBasePath()
   );
 
-  !targetDatetime && path.searchParams.append('cacheKey', cacheBuster());
+  if (!targetDatetime) {
+    path.searchParams.append('cacheKey', cacheBuster());
+  }
   const response = await fetch(path);
   if (response.ok) {
     const result = (await response.json()) as GridState;
@@ -32,24 +33,11 @@ const getState = async (
 };
 
 const useGetState = (): UseQueryResult<GridState> => {
-  const queryClient = useQueryClient();
   const { urlTimeAverage, urlDatetime } = useParams<RouteParameters>();
-  const isHourly = urlTimeAverage
-    ? URL_TO_TIME_AVERAGE[urlTimeAverage] === TimeAverages.HOURLY
-    : false;
-
-  const shouldUseLastHour = isHourly && !urlDatetime;
   const timeAverage = urlTimeAverage
     ? URL_TO_TIME_AVERAGE[urlTimeAverage]
     : TimeAverages.HOURLY;
-  console.log('shouldUseLastHour', shouldUseLastHour, isHourly, urlTimeAverage);
-  const lastHourQuery = useQuery<GridState>({
-    queryKey: [QUERY_KEYS.STATE, { aggregate: 'last_hour' }],
-    queryFn: () => getState('last_hour'),
-    enabled: shouldUseLastHour,
-  });
-
-  const fullDataQuery = useQuery<GridState>({
+  return useQuery<GridState>({
     queryKey: [
       QUERY_KEYS.STATE,
       {
@@ -58,26 +46,7 @@ const useGetState = (): UseQueryResult<GridState> => {
       },
     ],
     queryFn: () => getState(timeAverage, urlDatetime),
-    enabled: !shouldUseLastHour || (shouldUseLastHour && lastHourQuery.isSuccess),
   });
-
-  useEffect(() => {
-    if (shouldUseLastHour && fullDataQuery.data) {
-      queryClient.setQueryData(
-        [QUERY_KEYS.STATE, { aggregate: 'last_hour' }],
-        fullDataQuery.data
-      );
-    }
-  }, [shouldUseLastHour, fullDataQuery.data, queryClient]);
-
-  if (shouldUseLastHour) {
-    if (!fullDataQuery.data && lastHourQuery.data) {
-      return lastHourQuery;
-    }
-    return fullDataQuery;
-  }
-
-  return fullDataQuery;
 };
 
 export default useGetState;
--- a/web/src/api/getState.ts
+++ b/web/src/api/getState.ts
@@ -8,7 +8,7 @@ import { URL_TO_TIME_AVERAGE } from 'utils/state/atoms';
 import { cacheBuster, getBasePath, isValidDate, QUERY_KEYS } from './helpers';
 
 const getState = async (
-  timeAverage: TimeAverages | 'last_hour',
+  timeAverage: TimeAverages,
   targetDatetime?: string
 ): Promise<GridState> => {
   const shouldQueryHistorical =
--- a/web/src/api/getState.ts
+++ b/web/src/api/getState.ts
@@ -37,11 +37,15 @@ const useGetState = (): UseQueryResult<GridState> => {
     urlTimeAverage: UrlTimeAverages;
     urlDatetime?: string;
   }>();
-  const isHourly = urlTimeAverage === UrlTimeAverages['24h'];
+  const isHourly = urlTimeAverage
+    ? URL_TO_TIME_AVERAGE[urlTimeAverage] === TimeAverages.HOURLY
+    : false;
+
   const shouldUseLastHour = isHourly && !urlDatetime;
   const timeAverage = urlTimeAverage
     ? URL_TO_TIME_AVERAGE[urlTimeAverage]
     : TimeAverages.HOURLY;
+  console.log('shouldUseLastHour', shouldUseLastHour, isHourly, urlTimeAverage);
   const lastHourQuery = useQuery<GridState>({
     queryKey: [QUERY_KEYS.STATE, { aggregate: 'last_hour' }],
     queryFn: () => getState('last_hour'),
--- a/web/src/api/getState.ts
+++ b/web/src/api/getState.ts
@@ -1,17 +1,28 @@
 import type { UseQueryResult } from '@tanstack/react-query';
-import { useQuery } from '@tanstack/react-query';
-import { useAtom, useAtomValue } from 'jotai';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
 import type { GridState } from 'types';
-import { isHourlyAtom, timeAverageAtom } from 'utils/state/atoms';
+import { TimeAverages, UrlTimeAverages } from 'utils/constants';
+import { URL_TO_TIME_AVERAGE } from 'utils/state/atoms';
 
-import { cacheBuster, getBasePath, QUERY_KEYS } from './helpers';
+import { cacheBuster, getBasePath, isValidDate, QUERY_KEYS } from './helpers';
 
-const getState = async (timeAverage: string): Promise<GridState> => {
-  const path: URL = new URL(`v8/state/${timeAverage}`, getBasePath());
-  path.searchParams.append('cacheKey', cacheBuster());
+const getState = async (
+  timeAverage: TimeAverages | 'last_hour',
+  targetDatetime?: string
+): Promise<GridState> => {
+  const shouldQueryHistorical =
+    targetDatetime && isValidDate(targetDatetime) && timeAverage === TimeAverages.HOURLY;
+  const path: URL = new URL(
+    `v8/state/${timeAverage}${
+      shouldQueryHistorical ? `?targetDate=${targetDatetime}` : ''
+    }`,
+    getBasePath()
+  );
 
+  !targetDatetime && path.searchParams.append('cacheKey', cacheBuster());
   const response = await fetch(path);
-
   if (response.ok) {
     const result = (await response.json()) as GridState;
     return result;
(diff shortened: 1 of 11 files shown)


web/src/api/getState.ts
```diff
@@ -21,30 +32,55 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
-  const shouldFetchFullState =
-    !isHourly || hourZeroWasSuccessful || last_hour.isError === true;
+  useEffect(() => {
+    if (shouldUseLastHour && fullDataQuery.data) {
+      queryClient.setQueryData(
+        [QUERY_KEYS.STATE, { aggregate: 'last_hour' }],
+        fullDataQuery.data
+      );
+    }
+  }, [shouldUseLastHour, fullDataQuery.data, queryClient]);
```
VIKTORVAV99, line 74: Should we just remove the usage of last_hour?

It seems to add a lot of complexity for very small gains now that we have the cloudflare cache in place.

web/src/utils/formatting.ts
```diff
@@ -247,6 +247,33 @@ const formatDateTick = (
...
+    startDate.getFullYear() === endDate.getFullYear()
+  ) {
+    return `${new Intl.DateTimeFormat(locale, { month: 'short' }).format(
+      startDate
+    )} ${startDate.getDate()} - ${endDate.getDate()}, ${startDate.getFullYear()}`;
+  }
+
+  // If dates are in different months or years
+  return `${formatter.format(startDate)} - ${formatter.format(endDate)}`;
+}
```
VIKTORVAV99, line 275: 👀 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange

VIKTORVAV99 (changes requested):
Some early comments, but the only thing I would require a change of is the url handling so it supports upper case as well and to use the Intl formatRange function.

web/src/api/getState.ts
tonypls: I think it's working okay here. If we want to remove it let's take that in a separate PR

web/src/api/getState.ts
```diff
@@ -21,30 +32,55 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
 const useGetState = (): UseQueryResult<GridState> => {
-  const [timeAverage] = useAtom(timeAverageAtom);
-  const isHourly = useAtomValue(isHourlyAtom);
+  const queryClient = useQueryClient();
+  const { urlTimeAverage, urlDatetime } = useParams<{
+    urlTimeAverage: UrlTimeAverages;
+    urlDatetime?: string;
+  }>();
+  const isHourly = urlTimeAverage
+    ? URL_TO_TIME_AVERAGE[urlTimeAverage] === TimeAverages.HOURLY
```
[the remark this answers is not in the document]
tonypls: Should be fixed in the merge

web/src/utils/state/atoms.ts
```diff
@@ -18,7 +18,7 @@ import {
 
 export const timeAverageAtom = atom<TimeAverages>(TimeAverages.HOURLY);
 
-const URL_TO_TIME_AVERAGE: Record<string, TimeAverages> = {
+export const URL_TO_TIME_AVERAGE: Record<string, TimeAverages> = {
```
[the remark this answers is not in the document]
tonypls: I'd like it to be a stepping stone to unifying `hourly` -> `24h` everywhere. Maybe I'm optimistic thinking having it here makes it more likely to be a short term thing but I think it would be nice in a different PR and do a full switch over for consistency.

web/src/api/getState.ts
```diff
@@ -1,18 +1,28 @@
...
+  const shouldQueryHistorical =
+    targetDatetime && isValidDate(targetDatetime) && timeAverage === TimeAverages.HOURLY;
+  const path: URL = new URL(
+    `v8/state/${timeAverage}${
+      shouldQueryHistorical ? `?targetDate=${targetDatetime}` : ''
+    }`,
+    getBasePath()
+  );
 
+  !targetDatetime && path.searchParams.append('cacheKey', cacheBuster());
```
cadeban, line 24: Consider using an if statement instead of && - it's more readable imo

web/src/api/getState.ts
```diff
@@ -22,32 +32,52 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
+
+  const shouldUseLastHour = isHourly && !urlDatetime;
   const timeAverage = urlTimeAverage
     ? URL_TO_TIME_AVERAGE[urlTimeAverage]
     : TimeAverages.HOURLY;
-  const isHourly = urlTimeAverage ? timeAverage === TimeAverages.HOURLY : false;
-
-  // First fetch last hour only
-  const last_hour = useQuery<GridState>({
+  console.log('shouldUseLastHour', shouldUseLastHour, isHourly, urlTimeAverage);
```
cadeban, line 45: Remove console.log

web/src/api/getState.ts
```diff
@@ -22,32 +32,52 @@ const getState = async (timeAverage: string): Promise<GridState> => {
 };
 
 const useGetState = (): UseQueryResult<GridState> => {
-  const { urlTimeAverage } = useParams<RouteParameters>();
+  const queryClient = useQueryClient();
+  const { urlTimeAverage, urlDatetime } = useParams<RouteParameters>();
+  const isHourly = urlTimeAverage
+    ? URL_TO_TIME_AVERAGE[urlTimeAverage] === TimeAverages.HOURLY
```
cadeban, line 38: I think this can just be `const isHourly = URL_TO_TIME_AVERAGE[urlTimeAverage] === TimeAverages.HOURLY`

web/src/api/getState.ts
```diff
@@ -22,32 +32,52 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
+  const fullDataQuery = useQuery<GridState>({
+    queryKey: [
+      QUERY_KEYS.STATE,
+      {
+        aggregate: timeAverage,
+        targetDatetime: urlDatetime,
+      },
+    ],
+    queryFn: () => getState(timeAverage, urlDatetime),
+    enabled: !shouldUseLastHour || (shouldUseLastHour && lastHourQuery.isSuccess),
```
cadeban, line 61: I think that the second reference to `shouldUseLastHour` is redundant. It should be true if the first branch didn't short circuit

web/src/api/getZone.ts
```diff
@@ -35,17 +45,27 @@ const getZone = async (
...
+    queryKey: [
+      QUERY_KEYS.ZONE,
+      {
+        zone: zoneId,
+        aggregate: urlTimeAverage,
+        targetDatetime: urlDatetime,
+      },
+    ],
+    queryFn: async () => {
+      if (!zoneId) {
```
cadeban, line 64: Maybe we want to redirect the page instead? Or just use the presence of zoneId to enable/disable the query if we'd prefer to silently fail

web/src/api/getZone.ts
```diff
@@ -7,15 +7,25 @@ import { RouteParameters } from 'types';
...
+  const shouldQueryHistorical =
+    targetDatetime && isValidDate(targetDatetime) && timeAverage === TimeAverages.HOURLY;
+
+  const path: URL = new URL(
+    `v8/details/${timeAverage}/${zoneId}${
+      shouldQueryHistorical ? `?targetDate=${targetDatetime}` : ''
+    }`,
+    getBasePath()
+  );
+  !targetDatetime && path.searchParams.append('cacheKey', cacheBuster());
```
cadeban, line 28: nit: if statement?

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+    urlDatetime
+      ? newDate.setUTCDate(midnightDate.getUTCDate() - 1)
+      : newDate.setUTCDate(midnightDate.getUTCDate());
+
+    navigate({ datetime: formatDateToISO(newDate) });
+  }
+
+  function handleRightClick() {
+    if (!endDatetime) {
+      return;
```
cadeban, line 51: nit: Do we want to display some sort of message to the user if clicking the historical direction buttons fail?

web/src/api/getState.ts
```diff
@@ -22,32 +32,52 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
+  });
 
-  const shouldFetchFullState =
-    !isHourly || hourZeroWasSuccessful || last_hour.isError === true;
+  useEffect(() => {
+    if (shouldUseLastHour && fullDataQuery.data) {
+      queryClient.setQueryData(
```
cadeban, line 66: Question: Does `useQuery` not set query data?

web/src/api/getState.ts
```diff
@@ -22,32 +32,52 @@ const getState = async (timeAverage: string): Promise<GridState> => {
...
+    }
+  }, [shouldUseLastHour, fullDataQuery.data, queryClient]);
 
-  // Then fetch the rest of the data
-  const all_data = useQuery<GridState>({
-    queryKey: [QUERY_KEYS.STATE, { aggregate: timeAverage }],
-    queryFn: async () => getState(timeAverage),
-    enabled: shouldFetchFullState,
-  });
+  if (shouldUseLastHour) {
```
cadeban, line 73: It seems like we want to prefer `fullDataQuery` if present and use `lastHourQuery` otherwise or as a fallback? If so, I think we can just return: 
`return !fullDataQuery.data && lastHourQuery.data ? lastHourQuery : fullDataQuery;`

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+              language={i18n.languages[0]}
+              endDatetime={endDatetime}
+            />
+          }
+          type="success"
+        />
+      )}
+      <div className="flex flex-row items-center gap-2">
+        <ChevronLeft onClick={handleLeftClick} className="text-brand-green" />
+        <ChevronRight
```
cadeban, line 85: Consider using the Button component with these icons, since I think they already include a lot of a11y considerations

web/src/features/time/HistoricalTimeHeader.tsx
tonypls: I've improved the buttons so hopefully the isDisabled flag can communicate that there's no expected action

cadeban:
The code looks good and the functionality works as expected. 👌 

Last thing - the direction buttons are disabled in non-hourly views. Do we want to hide them? Or maybe show a tooltip on hover explaining that this functionality is not enabled yet for aggregate views?
Screenshot 2024-11-19 at 09 51 31

cadeban (approved):
lgtm!

tonypls:
@cadeban I noticed some inconsistency in the navigation logic so I simplified it! Could ya please review the last commit?

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -11,61 +11,39 @@ import { RouteParameters } from 'types';
...
   const navigate = useNavigateWithParameters();
-  const isHourly = useAtomValue(isHourlyAtom);
-  function handleLeftClick() {
-    if (!urlDatetime && !endDatetime) {
-      return;
-    }
 
-    const targetDate = urlDatetime ? new Date(urlDatetime) : endDatetime;
-    if (!targetDate) {
+  const handleTimeNavigation = (direction: 'left' | 'right') => {
```
cadeban, line 25: Subtracting/adding `TWENTY_FOUR_HOURS` is a lot easier to understand than previously. But I think I preferred the separate click handlers rather than passing in the direction (since we still need to create [distinct click handlers](https://github.com/electricitymaps/electricitymaps-contrib/pull/7397/commits/3fc3de50ee9d226127e1b7d9d26fae0e3bcb816e#diff-14a5cbe59e006d396425a12e6f93bc00e744e313563bb64bb23aecfb00165ab5R64) anyways). We could pull out the shared logic into functions. What do you think about something like this (I haven't actually tested this, so not sure if it's a direct translation, but something similar):
```javascript
const clampedDatetime = (date: Date): string => {
  // if date is greater than twentyFourHoursAgo, then go to latest
  if (date.getTime() >= Date.now() - TWENTY_FOUR_HOURS) {
    return '';
  }
  return date.toISOString();
};

const navigateDateOffset = (offset: number) => {
  const date = new Date(currentDate.getTime() + offset);
  navigate({ datetime: clampedDatetime(date) });
}

const handleLeftNavigation = () => navigateDateOffset(-TWENTY_FOUR_HOURS)

const handleRightNavigation = () => {
  if (!endDatetime || !urlDatetime) {
    return;
  }
  navigateDateOffset(TWENTY_FOUR_HOURS)
}
```

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -11,61 +11,39 @@ import { RouteParameters } from 'types';
...
 
-function setToMidnightUTC(date: Date): Date {
-  const newDate = new Date(date);
-  newDate.setUTCHours(0, 0, 0, 0);
-  return newDate;
-}
-function formatDateToISO(date: Date): string {
-  return date.toISOString().slice(0, -5) + 'Z';
-}
+const TWENTY_FOUR_HOURS = 24 * 60 * 60 * 1000;
```
cadeban, line 14: Can we add a comment that specifies that this is in milliseconds? Or rename it to `TWENTY_FOUR_HOURS_MS`?

web/src/features/time/HistoricalTimeHeader.tsx
cadeban: Also, are we still planning on anchoring midnight to midnight?

tonypls: I think for this v1 it's better to not anchor to midnight. From a users perspective this makes the most sense without limiting the current day to the elapsed hours.

web/src/features/time/HistoricalTimeHeader.tsx
tonypls: We can but this feels rather nitty and it currently follows our pattern in the api/helpers.ts

web/src/features/time/HistoricalTimeHeader.tsx
tonypls: Good point on the separate click handlers, I've pulled them apart to be completely isolated as we'll probably adjust their functionality to be even more unique in the future.

cadeban (approved):
lgtm!

tonypls:
> Some early comments, but the only thing I would require a change of is the url handling so it supports upper case as well and to use the Intl formatRange function.

Hey Viktor, have your requested changes been addressed?

PR #7398, opened 2024-11-12
fix(parsers): Update NL parser to include ENTSO-e data

annaib00:
## Issue
NED.nl missing historical data before 2021

## Description
Update parser to include historical ENTSO-e data. Before 2021 NED still has wind and solar, before 2017 NED only has wind, so ENTSO-e data is included to fill the missing types in these timeframes. 

### Double check


parsers/NED.py
```diff
@@ -145,6 +172,78 @@ def call_api(target_datetime: datetime, forecast: bool = False):
...
+        logger.info(
+            f"Failed to match {len(non_matching_indices)}  NED events with ENTSOE events"
+        )
+        ned_data.events = [
+            event
+            for idx, event in enumerate(ned_data.events)
+            if idx not in non_matching_indices
+        ]
+
+    return ned_data
```
VIKTORVAV99, line 244: We have this function on the ProductionBreakdownList class:
https://github.com/electricitymaps/electricitymaps-contrib/blob/64987b5bafdb2fff0be3fafa9e00b5219f000e01/electricitymap/contrib/lib/models/event_lists.py#L263-L304

Would that work instead of this custom one?
If so I think it would be better to use that so it is consistent with all other parsers.

annaib00: Ah okay, I will look into using that instead. I had worked on an outdated branch, so I closed this commit as I'm not very confident in working in git yet:)

VIKTORVAV99: That's okay! I don't mind at all 🙂

annaib00: @VIKTORVAV99 Hmm, I think it might be a bit challenging to use this function. 

Example if we are fetching data for a span 2016-12-31 to 2017-01-01, after 2017 we get wind and solar from NED and the rest from entsoe, but before 2017 we only get wind from NED and the rest from entsoe. So in order to use this function we would have to set the entsoe wind and solar values to zero after 2017, and only wind to zero before 2017, so we don't double up on wind or solar when merging the two frames. 

Perhaps this is still better in order to keep it consistent? What do you think? :)

VIKTORVAV99: I see!

We also have a similar static function that just updates values instead of merging them, perhaps that's better then?

If so we can fetch the ENTSOE data then the NED NL data and overwrite all ENTSOE data with the data that is available from NED NL.

That sounds like it should do what we want?

But if that don't work either I'm fine with keeping the current function if we add tests for it so we don't accidentally change something in the parser classes that break this function.

annaib00: I think that would work! 
I assume that it is the update_production_breakdowns function from the ProductionBreakdownList class? I will look into it, thanks!

VIKTORVAV99: Yeah that should be the one!

PR #7399, opened 2024-11-12
bug(web): multiple api calls when not hourly

tonypls:
## Issue
When not in the hourly view we are fetching data twice as the atoms initialize.

## Description
This PR updates our fetching to treat the url as the source of truth.

The bug is present on staging so it would be good to get this fix out before releasing contrib.

### DIFF
--- a/web/src/main.tsx
+++ b/web/src/main.tsx
@@ -23,6 +23,7 @@ import {
 } from 'react-router-dom';
 import i18n from 'translation/i18n';
 import { RouteParameters } from 'types';
+import { UrlTimeAverages } from 'utils/constants';
 import { createConsoleGreeting } from 'utils/createConsoleGreeting';
 import enableErrorsInOverlay from 'utils/errorOverlay';
 import { getSentryUuid } from 'utils/getSentryUuid';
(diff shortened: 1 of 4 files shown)


web/src/main.tsx
```diff
@@ -77,9 +78,20 @@ refetchDataOnHourChange(queryClient);
 export function ValidZoneIdGuardWrapper({ children }: { children: JSX.Element }) {
   const [searchParameters] = useSearchParams();
   const { zoneId, urlTimeAverage } = useParams<RouteParameters>();
+  const location = useLocation();
   if (!zoneId) {
     return <Navigate to="/map/24h" replace />;
   }
+  if (!urlTimeAverage) {
```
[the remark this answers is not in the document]
tonypls: I've added a check below 👍

web/src/main.tsx
```diff
@@ -171,23 +183,11 @@ const router = createBrowserRouter([
...
-        ),
+        element: <Navigate to="/map/24h" replace />,
       },
       {
         path: '/zone',
-        element: (
-          <Navigate
-            to="/map/24h"
-            replace
-            state={{ preserveSearch: true, preserveHash: true }}
```
VIKTORVAV99, line 157: I assume these are the defaults?

tonypls: It's to write to [the url state](https://developer.mozilla.org/en-US/docs/Web/API/History/state)

Which in this case we don't even use those keys. Cursor added it at some point and I thought it seemed logical but looking deeper it does nothing!

PR #7402, opened 2024-11-13
Update the app frontend to handle 72 hour window

cadeban:
## Description
[AVO-634](https://linear.app/electricitymaps/issue/AVO-634/update-the-app-frontend-to-handle-72-hour-windows): Update the app frontend to handle 72 hour windows

Changelog:
- Refetches on hourly change
- Adds ticks and formatting to x axis
- Updates chart title locales
- Allows pagination

**Change requires**: https://github.com/electricitymaps/electricitymaps/pull/6102

Related to:
- https://github.com/electricitymaps/electricitymaps/pull/5918
- https://github.com/electricitymaps/electricitymaps/pull/5990

### Preview
Screenshot 2024-11-21 at 17 51 59
Screenshot 2024-11-21 at 17 52 19

### Double check


VIKTORVAV99:
When I gave this a go with a quick and dirty prototype for 48 hours I ran into a issue with the latest label being hardcoded to index 24 (position 25), did you run into the same issue?

web/src/locales/en.json
```diff
@@ -262,48 +264,56 @@
   "country-history": {
     "carbonintensity": {
       "hourly": "Hourly carbon intensity",
+      "hourly_72": "Hourly carbon intensity",
```
VIKTORVAV99, line 267: Do you think we can do this in a way that don't duplicate the translations?

It's not a problem to do it like this but if there is a *easy* way for us to avoid it I would.

VIKTORVAV99:
> When I gave this a go with a quick and dirty prototype for 48 hours I ran into a issue with the latest label in the time controller being hardcoded to index 24 (position 25), did you run into the same issue?

VIKTORVAV99:
Some additional things we need to look at:

This date should probably be a range:
image

And the tooltips don't include the hours and timezone:
image
(we probably just need to adjust the isHourly atom to allow for this new aggregate as well)

cadeban:
@.tonypls's changes for historical frontend included the time range, so it should look ok now:
Screenshot 2024-11-21 at 15 30 17

Looking into the index labeling and tooltips!

VIKTORVAV99 (commented):
Don't we need to update this as well?
```ts
export const isHourlyAtom = atom((get) => get(timeAverageAtom) === TimeAverages.HOURLY);
```
This atom should probably return true for HORULY_72 as well, and it could maybe help you simplify some of the other logic.

tonypls:
Let me know if there's anyway I can help with this one!

cadeban:
FYI this PR should probably be ran against https://github.com/electricitymaps/electricitymaps/pull/6102 during review

cadeban:
Updated to work without https://github.com/electricitymaps/electricitymaps/pull/6102 per our discussion

tonypls:
Something curious happening with the time ticks:
image

cadeban:
😧 Good find, thanks @tonypls!

cadeban:
@tonypls @VIKTORVAV99 I fixed the time tick visual bug, so we should be all set for another round of reviews (when y'all have time, of course)

web/src/components/TimeAverageToggle.tsx
```diff
@@ -28,12 +28,18 @@ export interface TimeAverageToggleProps {
 function TimeAverageToggle({ timeAverage, onToggleGroupClick }: TimeAverageToggleProps) {
   const { t } = useTranslation();
   const historicalLinkingEnabled = useFeatureFlag('historical-linking');
+  const is72HourEnabled = useFeatureFlag('72-hours');
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { HOURLY_72, ...standardTimeAverages } = TimeAverages;
+  const timeOptions = is72HourEnabled ? TimeAverages : standardTimeAverages;
```
tonypls, line 35:
```suggestion
  const timeOptions = is72HourEnabled 
    ? Object.values(TimeAverages)
    : Object.values(TimeAverages).filter(option => option !== TimeAverages.HOURLY_72);
```

This could help not having an unused var

cadeban:
Screenshot 2024-12-09 at 13 08 52

web/src/features/time/TimeController.tsx
```diff
@@ -112,6 +112,7 @@ export default function TimeController({ className }: { className?: string }) {
     },
     [setSelectedDatetime, selectedDatetime.datetime, numberOfEntries, setTimeAverage]
   );
+
```
VIKTORVAV99, line 115: 👀

web/src/utils/formatting.ts
```diff
@@ -142,6 +142,7 @@ export const getDateTimeFormatOptions = (
   timezone?: string
 ): Intl.DateTimeFormatOptions => {
   switch (timeAverage) {
+    case TimeAverages.HOURLY_72:
     case TimeAverages.HOURLY: {
```
VIKTORVAV99, line 146: Now I'm glad I opted for the switch case instead of if statements 😅

VIKTORVAV99 (approved):

web/src/features/time/TimeController.tsx
cadeban: Ha, that's intentional. I thought more visual spacing could help differentiate between the logic and JSX of this component

web/src/utils/formatting.ts
cadeban: 💯 Made this super easy, so thanks!

PR #7403, opened 2024-11-13
fix(parsers): Make negative wind values zero GB-NIR

annaib00:
## Issue
Repeated ExpectedMode outliers detected for GB-NIR due to missing wind values

## Description
Sometimes EirGrid reports negative wind generation values (around -1 to -3 MW) these values are currently set to None in our parser which results in a lot of outliers being detected. It does not seem to be outliers but rather some consumption when the turbines stand still as it fits with EirGrids own low forecasted wind values. This PR makes the negative values zero instead of None. 

The issue is appearing repeatedly. 

### Preview
Screenshot 2024-11-13 at 13 06 44

Below EirGrids forecast and actual value for wind generation is seen, the negative values seem to be due to turbine consumption. 
Screenshot 2024-11-13 at 13 10 14

### Double check

### DIFF
--- a/parsers/SMARTGRIDDASHBOARD.py
+++ b/parsers/SMARTGRIDDASHBOARD.py
@@ -156,7 +156,7 @@ def fetch_production(
 
         productionMix = ProductionMix()
         if all([item["Value"], wind_prod]):
-            productionMix.add_value("wind", wind_prod)
+            productionMix.add_value("wind", wind_prod, correct_negative_with_zero=True)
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

I can't remember why we didn't make this the default...

annaib00:
Yes, you are right. I am doing a fix now. You can see if it looks like the good fix

PR #7405, opened 2024-11-13
Revert "chore(deps): bump jotai from 2.9.3 to 2.10.1 in /web"

tonypls:
Reverts electricitymaps/electricitymaps-contrib#7380

Fixes AVO-635


VIKTORVAV99 (approved):
Should we report the behavioral change to Jotai?

tonypls:
> > Should we report the behavioral change to Jotai?
> 
> I would if I could tell exactly why but I don't have the time to spend debugging right now

Hmm doesn't actually fix it 😢

VIKTORVAV99:
So are we keeping the old version?

I can try and take a look at it in a bit and see if there is anything else that might have affected this.

tonypls:
Doesn't solve the problem

PR #7406, opened 2024-11-13
fix(parsers): GB-NIR update to not add negative wind to unknown

annaib00:
## Issue
When receiving negative wind values they are effectively added to the unknown. Now we subtract the productionMix.wind which is zero for negative wind_prod values due to the "correct_negative_with_zero = True"

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Should work very well, and is nicer than what I had in mind!

PR #7407, opened 2024-11-13
chore(SEO): Sanitze zone names in URL

VIKTORVAV99:
## Issue

In some online platforms when you write a link like this https://app.electricitymaps.com/DK-DK1: Denmark, the colon is included which breaks the links. Similar problems can occur with (https://app.electricitymaps.com/DK-DK2) causing the trailing parentheses to be included in the link.

## Description

In order to avoid this I added a regex that only allows the zoneID to contain, letters, numbers and `-` which we use as a subzone delimiter. This should resolve the issue quite nicely.

I also took to opportunity to clean up the code a bit so it handles more things internally and not potentially causing multiple navigations requests. So instead of going `aus -> AUS -> AU` it simply goes `aus -> AU`.

### Double check

### DIFF
--- a/web/src/main.tsx
+++ b/web/src/main.tsx
@@ -123,21 +123,23 @@ export function ValidZoneIdGuardWrapper({ children }: { children: JSX.Element })
     return <Navigate to={`/map/24h?${searchParameters}`} replace />;
   }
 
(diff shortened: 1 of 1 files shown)


tonypls (approved):
LGTM

PR #7408, opened 2024-11-13
fruit(web): update US translations with SEO friendly zone names

tonypls:
## Issue
It's not always clear what zone is what in search results

## Description
This PR updates some US zone page meta titles to include full SEO zone names.

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -1756,7 +1756,8 @@
       "zoneName": "United States Minor Outlying Islands"
     },
     "US": {
-      "zoneName": "Contiguous United States"
+      "zoneName": "Contiguous United States",
+      "seoZoneName": "United States of America - USA"
     },
     "US-AK": {
       "countryName": "USA",
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -1757,7 +1757,7 @@
     },
     "US": {
       "zoneName": "Contiguous United States",
-      "seoZoneName": "United States of America - USA"
+      "seoZoneName": "United States of America - U.S."
     },
     "US-AK": {
       "countryName": "USA",
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -1920,7 +1920,8 @@
     },
     "US-NE-ISNE": {
       "countryName": "USA",
-      "zoneName": "New England ISO"
+      "zoneName": "ISO New England",
+      "seoZoneName": "ISO New England - ISO-NE"
     },
     "US-NW-AVA": {
       "countryName": "USA",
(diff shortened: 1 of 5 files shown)


web/src/locales/en.json
```diff
@@ -1801,16 +1801,19 @@
     "US-CAL-BANC": {
       "countryName": "USA",
       "displayName": "BANC",
-      "zoneName": "Balancing Authority of Northern California"
+      "zoneName": "Balancing Authority of Northern California",
+      "seoZoneName": "Balancing Authority of Northern California (BANC)"
```
VIKTORVAV99, line 1805:
```suggestion
      "seoZoneName": "Balancing Authority of Northern California - BANC"
```

web/src/translation/translation.ts
```diff
@@ -15,6 +15,14 @@ export const getZoneName = (zoneCode: string) => {
   return fullName;
 };
 
+export const getSEOZoneName = (zoneCode: string) => {
+  const seoZoneName = translateIfExists(`zoneShortName.${zoneCode}.seoZoneName`);
+  if (seoZoneName) {
+    return seoZoneName;
+  }
+  return getZoneName(zoneCode);
+};
```
VIKTORVAV99, line 24:
```suggestion
export const getSEOZoneName = (zoneCode: string) =>
  translateIfExists(`zoneShortName.${zoneCode}.seoZoneName`) || getZoneName(zoneCode);
```

VIKTORVAV99 (commented):
Two small things but otherwise it looks good!

Probably going to take a while for Google to catch on though.

web/src/translation/translation.ts
VIKTORVAV99: I'll allow it 😎

VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
Can we also change New England ISO to ISO New England https://staging.electricitymaps.com/zone/US-NE-ISNE/24h, this would then match their official naming:
image

The abbreviation should be `ISO-NE`

VIKTORVAV99:
@tonypls can you also update the zoneNames in the file the API uses?

Should be in /public/ instead.

web/src/utils/constants.ts
```diff
@@ -1,7 +1,7 @@
 import type { Duration } from 'date-fns';
 import { ElectricityModeType } from 'types';
 
-export const metaTitleSuffix = ' | App | Electricity Maps';
```
VIKTORVAV99, line 4: Do we really want to change this?

It currently follows the same style as our other websites and describes what they are rather than what they do.

Blog, Docs, etc

tonypls: Whoops, didn't mean to include this change in this PR.

PR #7411, opened 2024-11-13
chore(shareability): track successful shares (mobile only)

jbdietrich:
## Issue

We'd like to be able to track shares that are completed (In other words, the user completes the share action).

## Description

After this change, we record the result of the call to `CapShare.share` https://capacitorjs.com/docs/apis/share#shareresult. When it's defined (truthy), the share completed. When it's falsy, the share was not completed.

### Double check

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -8,7 +8,7 @@ _If you're looking for more info on the parsers, check out [how to setup Python 
 
 Prerequisites:
 
-- Ensure you have `NodeJS` and `pnpm` installed locally (`brew install pnpm`)
+- Ensure you have `NodeJS` (`v20.9.0` or above) and `pnpm` installed locally (`brew install pnpm`)
 - Run `pnpm install` in both the web and mockserver directories
 
 1. Start the mockserver: `pnpm run mockserver`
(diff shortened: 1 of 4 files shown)


web/README.md
```diff
@@ -8,7 +8,7 @@ _If you're looking for more info on the parsers, check out [how to setup Python
 
 Prerequisites:
 
-- Ensure you have `NodeJS` and `pnpm` installed locally (`brew install pnpm`)
+- Ensure you have `NodeJS` (`v20.9.0` or above) and `pnpm` installed locally (`brew install pnpm`)
```
VIKTORVAV99, line 11: I got to remember to update this in the future but it makes sense!

web/src/hooks/useShare.ts
```diff
@@ -21,12 +21,14 @@ export function useShare() {
...
       try {
-        await CapShare.share(shareData);
+        const result = await CapShare.share(shareData);
+        return result;
       } catch (error) {
         if (error instanceof Error && !/AbortError|canceled/.test(error.toString())) {
           console.error(error);
           callback?.(t('share-button.share-error'));
         }
+        return null;
```
VIKTORVAV99, line 31: Minor nit, I don't think you need to return null here, just return will do (null vs undefined) but it's not a problem either.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7412, opened 2024-11-13
chore(mobileapp): Fix iPad layout

### LINKED ISSUE #6966: iOS header overlaps with UI

VIKTORVAV99:
## Issue

Fixes #6966 

## Description

Ensures that the safe areas are respected in the relevant parts of the app.

### Preview

image

### Double check

### DIFF
--- a/web/src/features/map-controls/MapControls.tsx
+++ b/web/src/features/map-controls/MapControls.tsx
@@ -116,7 +116,7 @@ function DesktopMapControls() {
   const isConsumptionOnlyMode = useFeatureFlag('consumption-only');
 
   return (
-    <div className="pointer-events-none absolute right-3 top-2 z-20 hidden flex-col items-end md:flex">
+    <div className="pointer-events-none absolute right-3 top-2 z-20 mt-[env(safe-area-inset-top)] hidden flex-col items-end md:flex">
       <div className="pointer-events-auto mb-16 flex flex-col items-end space-y-2">
         {!isConsumptionOnlyMode && <ConsumptionProductionToggle />}
         <SpatialAggregatesToggle />
--- a/web/src/features/map-controls/ZoomControls.tsx
+++ b/web/src/features/map-controls/ZoomControls.tsx
@@ -9,7 +9,7 @@ export default function ZoomControls(): ReactElement {
     <NavigationControl
       style={{
         marginRight: 12,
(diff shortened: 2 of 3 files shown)


corradio:
Super cool! I'll remove myself from the review as I don't feel I have the background required to review this.
By the way, would we consider hiding the left panel in portrait mode? It would feel more intuitive to me, but maybe that's a different issue!

VIKTORVAV99:
> Super cool! I'll remove myself from the review as I don't feel I have the background required to review this. By the way, would we consider hiding the left panel in portrait mode? It would feel more intuitive to me, but maybe that's a different issue!

Different issue but yeah we can consider hiding it by default in portrait mode!

Perfectly fine with the review 👍🏻

cadeban:
Tested via xcode simulator and it looks good in both orientations:
Screenshot 2024-11-14 at 09 50 40
Screenshot 2024-11-14 at 09 50 23

VIKTORVAV99:
@cadeban mind pressing that green approval button is it looks good? 👀 💚

web/src/features/map-controls/MapControls.tsx
```diff
@@ -116,7 +116,7 @@ function DesktopMapControls() {
   const isConsumptionOnlyMode = useFeatureFlag('consumption-only');
 
   return (
-    <div className="pointer-events-none absolute right-3 top-2 z-20 hidden flex-col items-end md:flex">
+    <div className="pointer-events-none absolute right-3 top-2 z-20 mt-[env(safe-area-inset-top)] hidden flex-col items-end md:flex">
```
cadeban, line 119: nit: Should we consider adding a fallback value for all instances of `env(safe-area-inset-top)`?

VIKTORVAV99: Not sure we need to, I only added it where we do a calc to be on the safe side.

But I can do it if we want to, previously we didn't have any though (when it was just for mobile).

cadeban (approved):
lgtm!

PR #7423, opened 2024-11-17
Update FI.yaml

pnook:
Updated solar and wind capacities

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for updating this, that's a pretty significant change in the wind capacity!

PR #7424, opened 2024-11-18
fix(events): downgrade negative production values from warning to debug

FelixDQ:
The warning log is expected for many parsers and we are therefore polluting our log files with millions of these logs. I have therefore downgraded it to debug so that it only shows up when we use `test_parser.py` and stop storing it in google cloud logs.


annaib00 (approved):

VIKTORVAV99 (approved):
LGTM! 🎉

Makes sense!

PR #7428, opened 2024-11-18
fix(parsers): add hydro to bypassed modes for US-SE-SOCO

annaib00:
## Issue
Data currently removed for US-SE-SOCO due to missing hydro values. Hydro is added to bypassed modes, so we can start getting data for this zone again.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7432, opened 2024-11-19
use with instead of assert

KTibow:
## Description

`assert` is deprecated and in fact doesn't work in new versions of node
`with` is basically the same thing and has been supported since node 18
this pr changes `assert` to `with`

## Double check

### DIFF
--- a/web/scripts/generateSitemap.js
+++ b/web/scripts/generateSitemap.js
@@ -5,7 +5,7 @@ import path from 'node:path';
 
 import { fileURLToPath } from 'url';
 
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Thanks for the contribution but this is not going to work as Prettier fails on it, and to update prettier we need to change the entire repo setup to a pnpm workspace to have the same functionality we have today since they switched from commonjs to ESM.

This also means we need to change the Earthly config files which is used for building and deploying to production.

madsnedergaard:

PR #7434, opened 2024-11-20
Tonyvanswet/avo 647 match historical storytelling design for up to 24h

tonypls:
## Description
This PR updates the historical storytelling timeslider work on both mobile and web as well as better match with the design.

### DIFF
--- a/web/src/components/Button.tsx
+++ b/web/src/components/Button.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { twMerge } from 'tailwind-merge';
 
-type SizeOptions = 'xs' | 'sm' | 'md' | 'lg';
+type SizeOptions = 'sm' | 'md' | 'lg';
 
 type ButtonTypes = 'primary' | 'secondary' | 'tertiary' | 'link';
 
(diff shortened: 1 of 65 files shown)


cadeban (approved):
:)

web/src/components/Button.tsx
```diff
@@ -130,6 +130,9 @@ function getForeground(type: ButtonTypes) {
 function getSize(size: SizeOptions, type: ButtonTypes, isIconOnly: boolean) {
   if (isIconOnly) {
     switch (size) {
+      case 'xs': {
+        return 'min-w-5 min-h-5';
+      }
```
VIKTORVAV99, line 135: Let @Alportan know about this so he ads it to the design system as well :)

Alportan: Is it regarding the min width and height?

VIKTORVAV99: There is a new XS size!

So yes 😅

Alportan: Aaa, that's true, I completely missed that 🙈 I'll add it as well since I myself had a need for a smaller button sometimes, but never justified to add it due to accessibility. Although if the hitbox is large enough, we should be good 👍

cadeban:
@tonypls Mind if I merge this in?

tonypls:
> @tonypls Mind if I merge this in?

Go for it!

web/src/components/Button.tsx
tonypls: This was to stop the slider jumping around when the buttons were hidden but by fixing the height of the component it now works without the new button size. To keep things simpler in both design and implementation I removed the xs size :)

PR #7438, opened 2024-11-21
AVO-631: Integrate historical links and share dropdown

cadeban:
## Description
[AVO-631](https://linear.app/electricitymaps/issue/AVO-631/integrate-sharing-drop-down-with-historical-links)
Share urls now contain time aggregates and datetimes. Removed isHourly check now that time aggregate view state is stored in url.

To test, ensure that historical linking and more options dropdown feature flags are on.

Note that [live vs historical share](https://linear.app/electricitymaps/issue/AVO-632/share-live-vs-share-historical-data) is coming in a follow up PR. 

### Double check


web/src/hooks/useGetCurrentUrl.ts
```diff
@@ -0,0 +1,19 @@
...
+  const {
+    urlTimeAverage,
+    urlDatetime,
+    zoneId: zoneIdParameter,
+  } = useParams<RouteParameters>();
+  const zId = zoneId || zoneIdParameter;
+  const url =
+    baseUrl +
+    (zId
+      ? `/zone/${zId}/${urlTimeAverage}/${urlDatetime || new Date().toISOString()}`
```
tonypls, line 15: This isn't floored to 00:00Z, when the URL is not present. We could try it with something like this:

`  const selectedDatetime = useAtomValue(selectedDatetimeIndexAtom);
  const targetDate = urlDatetime ?? new Date(selectedDatetime.datetime).toISOString();`
  
  and would be nice to keep zoneId variable naming over zId

tonypls (commented):
Couple of comments!

web/src/hooks/useGetCurrentUrl.ts
cadeban: Updated, LMK if I need any additional changes!

cadeban:
@tonypls @VIKTORVAV99 would you please take another pass on this PR when y'all have time? Thanks!

Issue #7436, opened 2024-11-21
Using a lowercase zone in URL does not work


madsnedergaard:
**Describe the bug**

Going to https://app.electricitymaps.com/zone/DK-dk2 no longer works.

This seems to be related to the new `24h` parameter, as this works: https://app.electricitymaps.com/zone/DK-dk2/24h 🤔 

**To Reproduce**
1. Open the link above

silkeholmebonnen:
It works for me

madsnedergaard:
Really? I tried going to https://app.electricitymaps.com/zone/DK-dk2 in both Safari and Arc/Chrome, getting same result where it's going to `/map` instead of the zone 😅

VIKTORVAV99:
Even valid ones like https://app.electricitymaps.com/zone/AU don't seem to work for me.

madsnedergaard:
Here's a video showing what the experience is like for me:

https://github.com/user-attachments/assets/2a2d84a5-7844-41cc-9d52-08ff6d4bf990

silkeholmebonnen:
My mistake, doesn't work for me either

PR #7443, opened 2024-11-22
fix: fix redirect

VIKTORVAV99:
## Issue

Closes: #7436

## Description

The redirect with the timeAverage caused things to cascade and then fail, by removing it things are back to normal and the redirects are still working. And redirecting to /24h still works as well.

### Double check


tonypls:
What about `/zone/DK-dk2/30d?` can we fix the zone id and preserve the time average?

VIKTORVAV99:
> What about `/zone/DK-dk2/30d?` can we fix the zone id and preserve the time average?

I can look into it but I think it would be good to merge this in the meantime.

tonypls (approved):
LGTM!

Issue #5906, opened 2023-09-17
MN production parser down


electricitymapsbot:
## Description

This is an automatic error report generated for Mongolia (MN).

Issues:
- No recent data found for `consumption` parser
- No recent data found for `production` parser

## Suggestions
- Try running the parser locally using the command `poetry run test_parser MN production`
- <a href="https://storage.googleapis.com/electricitymap-parser-logs/MN.html">Explore the runtime logs</a>

You can see an overview of all parser issues [here](https://github.com/electricitymaps/electricitymaps-contrib/wiki/Parser-issues).

q--:
See https://github.com/electricitymaps/electricitymaps-contrib/issues/5905

electricitymapsbot:
Closing this as the electricity data is flowing again.

electricitymapsbot:
Reopening this as the electricity data is no longer flowing.

q--:
The [endpoint](https://disnews.energy.mn/convertt.php) used by the parser, but also the [website](//ndc.energy.mn), now redirects to a login page instead of returning the JSON.

This is probably a mistake, as the web page still prominently features the image in which the generation was displayed.

Their e-mail if someone wants to contact them: ndc@ndc.energy.mn

electricitymapsbot:
Closing this as the electricity data is flowing again.

electricitymapsbot:
Reopening this as the electricity data is no longer flowing.

q--:
Same issue as before
> The [endpoint](https://disnews.energy.mn/convertt.php) (used by the parser, but also the [website](//ndc.energy.mn)) now redirects to a login page instead of returning the JSON.
> 
> This is probably a mistake, as the web page still prominently features the image in which the generation was displayed.
> 
> Their e-mail if someone wants to contact them: [ndc@ndc.energy.mn](mailto:ndc@ndc.energy.mn)

tonypls:
Bump, still an issue. Would greatly appreciate any help on this one 🙏

q--:
> https://github.com/electricitymaps/electricitymaps-contrib/issues/5906#issuecomment-1985432604

I checked again, this time it isn't due to the endpoint being down as I can see the JSON response at https://disnews.energy.mn/convertt.php. Not sure what's causing it this time, then.

tonypls:
Ah thanks! I had a look and it appears there's a key error:

  File "/contrib/parsers/MN.py", line 96, in fetch_production
[...]

PR #7444, opened 2024-11-22
Fix Mongolia parser, add thermal + battery support [Mostly complete, needs some work so if someone else wants to complete this please do]

q--:
## Issue

See #5906

## Description

Fixes the Mongolia parser by correcting a renamed field in the API, determining that 'Songino' stands for battery (https://github.com/electricitymaps/electricitymaps-contrib/issues/5906#issuecomment-2495091821). It also maps `tpp` to thermal.

# Still to be done:
- [ ] From now on, I'm assuming that tpp = thermal and any remaining production = hydro, so 'unknown' needs to have an override for [oil + coal](https://www.iea.org/countries/mongolia/electricity) instead of oil + coal + hydro. Keep the old override for the old data though, that's not based on `tpp` production but on total unaccounted for production and therefore should include hydro).
- [ ] Test if it works
- [ ] Formatting

_(I wrote this in the GitHub web editor hoping to give someone else a starting point for fixing this, if someone else can do the last steps that would be appreciated. May even be tagged 'help wanted'.)_

### Double check

### DIFF
--- a/config/zones/MN.yaml
+++ b/config/zones/MN.yaml
@@ -32,6 +32,9 @@ emissionFactors:
         datetime: '2020-01-01'
         source: IEA 2020; assumes 92.85% coal, 5.78%% oil, 1.37% hydro in unknown
         value: 729.1
+      - datetime: '2024-11-25'
+        source: IEA 2020; assumes 94.14% coal, 5.86% oil in unknown
+        value: 739
   lifecycle:
     battery discharge:
       - datetime: '2023-01-01'
@@ -46,6 +49,9 @@ emissionFactors:
         datetime: '2020-01-01'
         source: IEA 2020; assumes 92.85% coal, 5.78%% oil, 1.37% hydro in unknown
         value: 799.2
+      - datetime: '2024-11-25'
+        source: IEA 2020; assumes 94.14% coal, 5.86% oil in unknown
+        value: 810
 fallbackZoneMixes:
   powerOriginRatios:
     - _source: Electricity Maps, 2023 average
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
Looks good from a first glance!

There is a risk that unknown can contain other types than just hydro as we don't know if all of their production modes are actually reported properly for all plants. But I think this is a reasonable assumption to make given the available data.

I can take a look at it later today and add the needed override values. Sadly other contributors can't edit existing PR, only EMaps members can.

q--:
Thank you!

A small thing I didn't think of when working on this: I could also imagine
that 'dts' is just coal and not coal + oil, it might've been established at
some point but I don't remember.

config/zones/MN.yaml
```diff
@@ -46,6 +49,9 @@ emissionFactors:
         datetime: '2020-01-01'
         source: IEA 2020; assumes 92.85% coal, 5.78%% oil, 1.37% hydro in unknown
         value: 799.2
+      - datetime: '2024-11-25'
+        source: IEA 2020; assumes 94.14% coal, 5.86% oil in unknown
```
q--, line 53: This `source` also says `IEA 2020` but has different values from two lines up, is that correct?

VIKTORVAV99: Yes, basically I used the same source but simply removed hydro since that will now be reported on it's own.

parsers/MN.py
```diff
@@ -99,6 +101,7 @@ def fetch_production(
     # 'unknown' consists of 92.8% coal, 5.8% oil and 1.4% hydro as per 2020; sources: IEA and IRENA statistics.
     query_data["leftoverMW"] = round(
         query_data["consumptionMW"]
+        - query_data["thermalMW"]
         - query_data["importMW"]
         - query_data["solarMW"]
         - query_data["windMW"],
```
annaib00, line 107: LGTM! 
Just a quick question, should any usage/withdrawings from battery storage be included here?

PaulRoms:

PR #7447, opened 2024-11-24
feat(parsers): add ERCOT parser

silkeholmebonnen:
## Issue

[AVO-582](https://linear.app/electricitymaps/issue/AVO-582/fix-ercot-parsers)

## Description

This PR adds a parser for ERCOT that gets real-time production (including battery storage), consumption, exchanges and historical production from ercot.com. 

This PR build on top of the work made in #5011 

### Preview

Output of running `poetry run test_parser US-TEX-ERCO`
```
[{'datetime': datetime.datetime(2024, 11, 23, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='US/Central')),
  'production': {'coal': 5067.958333333333,
                 'gas': 9792.533333333333,
                 'hydro': 5.799999999999998,
                 'nuclear': 5085.675,
                 'solar': 0.9583333333333334,
                 'unknown': 0.06666666666666667,
                 'wind': 21959.933333333334},
  'source': 'ercot.com',
  'storage': {'battery': -76.62675000000002},
  'zoneKey': 'US-TEX-ERCO'},
 ....
 {'datetime': datetime.datetime(2024, 11, 24, 7, 0, tzinfo=zoneinfo.ZoneInfo(key='US/Central')),
  'production': {'coal': 3603.0166666666664,
                 'gas': 5868.258333333334,
                 'hydro': 95.15833333333335,
                 'nuclear': 5086.958333333333,
                 'solar': 1652.9583333333333,
                 'unknown': 0.0,
                 'wind': 24439.916666666668},
  'source': 'ercot.com',
  'storage': {'battery': 387.5513333333333},
  'zoneKey': 'US-TEX-ERCO'}]
---------------------
took 1.08s
min returned datetime: 2024-11-23 06:00:00+00:00 UTC
max returned datetime: 2024-11-24 13:00:00+00:00 UTC  -- OK, <2h from now :) (now=2024-11-24T14:32:33+00:00 UTC)
```

Plots comparing EIA's and ERCOT's historical data.
September 2024:
comparison_historically_plot_september

Carbon intensity using EIA's and ERCOT's historical data for september 2024:
comparison_carbon_intensity_plot

RE % using EIA's and ERCOT's historical data for september 2024:
comparison_renewable_percentage_plot

ERCOT battery storage:
image
Gridstatus graph for the same period (5 minute granularity):
image

### Double check


parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+
+import gzip
+import json
+import zipfile
+from datetime import datetime, timedelta, timezone
+from io import BytesIO
+from logging import Logger, getLogger
+from zoneinfo import ZoneInfo
+
+import arrow
```
VIKTORVAV99, line 11: We are trying to remove arrow from the parsers so would be great if we could avoid adding it here.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+        }
+        all_data_points.append(data_point)
+
+    return all_data_points
+
+
+def fetch_live_exchange(
+    zone_key1: str,
+    zone_key2: str,
+    session: Session = Session(),
```
VIKTORVAV99, line 270: Generally we need to avoid setting mutable defaults, thous should default to None and then the session should be handled inside the function.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+        agg_data_point["sortedZoneKeys"] = sortedZoneKeys
+        agg_data_point["source"] = "ercot.com"
+        aggregated_data_points.append(agg_data_point)
+    validated_data_points = [x for x in all_data_points if validate_exchange(x, logger)]
+    return validated_data_points
+
+
+def fetch_production(
+    zone_key: ZoneKey = ZoneKey("US-TEX-ERCO"),
+    session: Session = Session(),
```
VIKTORVAV99, line 303: Same here.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+            session=session,
+            target_datetime=target_datetime,
+            logger=logger,
+        )
+    return production
+
+
+def fetch_consumption(
+    zone_key: ZoneKey = ZoneKey("US-TEX-ERCO"),
+    session: Session = Session(),
```
VIKTORVAV99, line 325: And here.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+            target_datetime=target_datetime,
+            logger=logger,
+        )
+    return consumption
+
+
+def fetch_exchange(
+    zone_key1: ZoneKey,
+    zone_key2: ZoneKey,
+    session: Session = Session(),
```
VIKTORVAV99, line 350: And here.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+    aggregated_data_points = []
+    for dt in sorted({item["datetime"] for item in all_data_points}):
+        agg_data_point = {}
+        agg_data_point["datetime"] = dt
+        values_dt = [
+            item["netFlow"] for item in all_data_points if item["datetime"] == dt
+        ]
+        agg_data_point["netFlow"] = sum(values_dt) / len(values_dt)
+        agg_data_point["sortedZoneKeys"] = sortedZoneKeys
+        agg_data_point["source"] = "ercot.com"
```
VIKTORVAV99, line 295: Usually we set the source as a constant in the top level.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,366 @@
...
+                continue
+
+            # We get the data in 15 minute intervals, so we need to sum the 4 columns that represent an hour
+            start_idx = hour * 4
+            end_idx = start_idx + 4
+            hour_cols = time_columns[start_idx:end_idx]
+            hour_value = row[hour_cols].sum()
+
+            if hour_dt not in datapoints_by_date:
+                datapoints_by_date[hour_dt] = {"storage": {}, "production": {}}
```
VIKTORVAV99, line 247: You can avoid a lot of these manual checks by using the parser classes that handles these checks internally. (Check the ENTSOE parser for an example).

I'd recommend using them as they also do some internal validation to ensure all datapoints are valid in a uniform way.

VIKTORVAV99 (commented):
I know this is still a draft but I couldn't help myself but to add some early feedback!

Overall it looks like it's working very well!

I'm available on slack if you got any specific questions :)

parsers/US_ERCOT.py
silkeholmebonnen: I don't think I can avoid all of them because the data is read for a whole day for one production mode and then afterwards a whole day for another production mode. So I need to temporarily store them before I can add the whole hour of production mix. If you can think of a better way to do it, let me know:)

But I will try to modify the code to use the parser classes as much as possible.

VIKTORVAV99: Ah I see!

Maybe you can read all of them and zip them? That would ensure they are all the same length as well. (just throwing out an idea, I don't know if it will work in this case).

But please do use the parser classes as they will track negative corrections and automatically round values to 1W incements.

VIKTORVAV99: Another alternative is to create 1 production / storage list for each mode and then merging all the production and storage. The parser classes have built in functions for that.

silkeholmebonnen: True but I don't think its worth spending time on now, so I'll leave it as it is, unless you have a strong opinion about it.

VIKTORVAV99: I'm not sure if the grid modeling team needs it otherwise I'm fine with it :)

parsers/test/mocks/ERCOT.json
```diff
@@ -0,0 +1,16231 @@
...
+          "Natural Gas": {
+            "gen": 9955.2,
+            "hsl": 13310.1,
+            "seasonalCapacity": 68234
+          }
+        },
+        "2024-11-23 00:24:57-0600": {
+          "Coal and Lignite": {
+            "gen": 5063.9,
+            "hsl": 9881.4,
```
VIKTORVAV99, line 186: Do you know what hsl stands for?

If it's the capacity that would be real interesting as technically we can parse live capacities. (follow up pr if that's the case)

parsers/test/test_ERCOT.py
```diff
@@ -0,0 +1,66 @@
...
+            self.assertIsNotNone(data)
+        with self.subTest():
+            self.assertEqual(data[0]["consumption"], 42015.44)
+        with self.subTest():
+            expected_dt = datetime(2024, 11, 24, 0, 0, tzinfo=ZoneInfo("US/Central"))
+            self.assertEqual(data[0]["datetime"], expected_dt)
+        with self.subTest():
+            self.assertEqual(data[0]["source"], "ercot.com")
+        with self.subTest():
+            self.assertEqual(data[0]["zoneKey"], "US-TEX-ERCO")
```
VIKTORVAV99, line 62: Do you think we can add a snapshot test as well, so if we change things in the future we can see exactly what has changed?

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,379 @@
...
+    consumption_list = TotalConsumptionList(logger)
+    for key in ["previousDay", "currentDay"]:
+        data_dict = data_json[key]
+        dt = datetime.strptime(data_dict["dayDate"], "%Y-%m-%d %H:%M:%S%z").replace(
+            tzinfo=TX_TZ
+        )
+        for item in data_dict["data"]:
+            if "systemLoad" in item:
+                consumption_list.append(
+                    zoneKey=zone_key,
```
VIKTORVAV99, line 115: Zone key here also need to have the type `ZoneKey` otherwise it causes type errors.

silkeholmebonnen:
> I just want to say that I'm super impressed by how atomic your commits are

That's not always the case haha 🙃

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,379 @@
...
+    return consumption_list
+
+
+def fetch_live_production(
+    zone_key: str,
+    session: Session | None = None,
+    logger: Logger = getLogger(__name__),
+) -> ProductionBreakdownList:
+    gen_data_json = get_data(url=RT_GENERATION_URL, session=session)["data"]
+    production_breakdowns = ProductionBreakdownList(logger)
```
VIKTORVAV99, line 129:
```suggestion
) -> ProductionBreakdownList:
    session = session or Session()
    gen_data_json = get_data(url=RT_GENERATION_URL, session=session)["data"]
    production_breakdowns = ProductionBreakdownList(logger)
```

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,379 @@
...
+            # Add storage data if available for this hour
+            if hour_dt in storage_by_hour:
+                storage_values = storage_by_hour[hour_dt]
+                if storage_values:
+                    storage.add_value(
+                        "battery", sum(storage_values) / len(storage_values)
+                    )
+
+            production_breakdowns.append(
+                zoneKey=zone_key,
```
VIKTORVAV99, line 175: Here as well for the zone key.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,379 @@
...
+    return validated_data_points
+
+
+def fetch_production(
+    zone_key: ZoneKey = ZoneKey("US-TEX-ERCO"),
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
+    if target_datetime is None or target_datetime > (
```
VIKTORVAV99, line 323:
```suggestion
) -> list:
    session = session or Session()
    if target_datetime is None or target_datetime > (
```

parsers/test/mocks/ERCOT.json
silkeholmebonnen: Maybe hourly system load (google search). But I cannot find the documentation for the API so I'm not sure.

VIKTORVAV99 (commented):
Some more small things, but it looks very good!

parsers/test/test_ERCOT.py
silkeholmebonnen: Felix said that he does this think testing in parsers is that important, so I'll let the two of you decide if it is worth the time or not.

FelixDQ (commented):
One comment, but otherwise LGTM

parsers/US_ERCOT.py
VIKTORVAV99: It throws type errors otherwise in these places.

parsers/test/test_ERCOT.py
VIKTORVAV99: Agreed, I'll see if I can add a snapshot test when I get back home. Should be fairly quick since most of the test infra needed is already there.

A snapshot just makes it easy to spot subtle changes in the values of we ever change something this depend on, like the EIA parser or the parser classes.

parsers/US_ERCOT.py
```diff
@@ -0,0 +1,379 @@
...
+            zone_key=zone_key, session=session, logger=logger
+        )
+    else:
+        production = fetch_historical_production(
+            zone_key=zone_key,
+            session=session,
+            target_datetime=target_datetime,
+            logger=logger,
+        )
+    return production.to_list()
```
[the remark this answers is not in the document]
silkeholmebonnen: Yes. Can I set it to 1 month even though the live production only returns data for today and yesterday? Or does both live and historical have to return the data in that frequency?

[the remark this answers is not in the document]
FelixDQ: Yes, that sounds good!
The refetch frequency is only for refetches. They don't have to return data with the same frequency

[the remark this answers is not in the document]
silkeholmebonnen: Now the historical production fetches the whole month, if target datatime is Feb then all of the data from Feb will be returned. I set the refetch frequency to 30 days, I hope that is fine even though it is not always exactly 30 days.

silkeholmebonnen:
I commented out everything that is storage related. Let's get this merged and figure out what to do about storage afterwards

parsers/US_ERCOT.py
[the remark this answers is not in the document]
VIKTORVAV99: > Now the historical production fetches the whole month, if target datatime is Feb then all of the data from Feb will be returned. I set the refetch frequency to 30 days, I hope that is fine even though it is not always exactly 30 days.

Maybe it would be safer to set it to 28 days? That way no matter what we should fetch every month, now technically it could skip feb if it's triggered on the "wrong" day.

FelixDQ (approved):

PR #7475, opened 2024-11-28
feat(web): add trackevents for historical time interactions

silkeholmebonnen:
## Issue

[AVO-657](https://linear.app/electricitymaps/issue/AVO-657/add-tracking-for-historical-story-telling)

## Description

This PR adds trackevents for the 3 buttons for historical time; Left button, Right button and Latest button.

### Double check


tonypls (commented):
Great! Maybe we could keep plausible cleaner with a goal of Historical Navigation that has properties of Left, Right and Latest? or Back, Forward and Latest?

silkeholmebonnen:
@tonypls can you add this goal to plausible? I don't think I have access

tonypls:
> @tonypls can you add this goal to plausible? I don't think I have access

Done! I've also changed you to an admin so you can add them in the future 👍

PR #7476, opened 2024-11-28
feat(web): add redirect toast for links older than 30 days

silkeholmebonnen:
## Issue

[AVO-654](https://linear.app/electricitymaps/issue/AVO-654/handle-links-older-than-30-days)

## Description

This PR adds a toast that pops up if a user has been redirected to the last date, because the url date was older than 30 days.

### Preview

image

### Double check


silkeholmebonnen:
Is this a replacement for [AVO-565](https://linear.app/electricitymaps/issue/AVO-535/historical-empty-state), or when do we want to show the empty state?

tonypls:
Great! @cadeban cc for vis

After chatting with Alex, we think it would make the most sense to send people to real time if it's expired. Maybe Cady can take over this branch?

VIKTORVAV99:
> Great! @cadeban cc for vis
> 
> 
> 
> After chatting with Alex, we think it would make the most sense to send people to real time if it's expired.

I also agree with this. It will also be more performant as it should be cached.

silkeholmebonnen:
> Great! @cadeban cc for vis
> 
> After chatting with Alex, we think it would make the most sense to send people to real time if it's expired. Maybe Cady can take over this branch?

Are you taking over from here @cadeban? Or should I continue

cadeban:
Looks like you've already made good progress on this, so I think you should keep going. I think we want this to be part of the launch tomorrow, so let's sync if we're not able to merge before EOD. How does that sound?

web/src/locales/en.json
```diff
@@ -247,7 +247,8 @@
       "daily": "30d",
       "monthly": "12mo",
       "yearly": "all"
-    }
+    },
+    "disclaimer": "This link is beyond the available dates. We've redirected you to the latest available date."
```
tonypls, line 251:
```suggestion
    "date-unavailable": "Date unavailable; redirecting to live data"
```

web/src/features/time/DateRedirectToast.tsx
```diff
@@ -0,0 +1,33 @@
...
+    if (isRedirectedToLatestDatetime) {
+      reference.current?.publish();
+    }
+    setIsRedirectedToLatestDatetime(false);
+  }, [isRedirectedToLatestDatetime, reference, setIsRedirectedToLatestDatetime]);
+
+  return (
+    <Toast
+      ref={reference}
+      description={t('time-controller.disclaimer')}
```
tonypls, line 25:
```suggestion
      description={t('time-controller.date-unavailable')}
```

web/src/locales/en.json
tonypls: Mostly to shorten the toast 👍

silkeholmebonnen:
My last commit is changing from this:
image
to this:
image

VIKTORVAV99:
> My last commit is changing from this: image to this: image

Love it! 💚

PR #7486, opened 2024-11-30
feat(parser): Add exchange parsing to the Nordpool parser

VIKTORVAV99:
## Issue

Sometimes ENTSO-E does maintenance and are down for hours which is causing us to get data lag.

## Description

This adds exchange parsing to the ENTSO-E parser, this works for all the nordic zones and the baltic zones as we have data access to those.

This will also on the amount of API calls we are doing to ENTSO-E which has been a problem in the past.

It's also a much simpler code which are easier to maintain.

### Double check


VIKTORVAV99:
We still need to determine which zones we want to enable this on, to begin with I suggest the internal NO or SE zones. (DK is using a 5 min parser so we should not replace that).

PR #7507, opened 2024-12-02
chore(deps): Update vite to v6

VIKTORVAV99:
## Description

Vite has released a new major version with some performance changes that benefit us around JSON parsing. This means that some of our files will be larger and the app will grow a bit but it should be significantly faster at parsing the JSON which benefits the map loading, config and languages a lot.

There is no migration needed as far as I can tell and all our existing tests pass without issues.

### Double check


silkeholmebonnen (approved):
LGTM 🎄

PR #7508, opened 2024-12-02
fix(web): change chart badges to show less when partially estimated

silkeholmebonnen:
## Issue

[AVO-620](https://linear.app/electricitymaps/issue/AVO-620/change-chart-badges-to-show-less-when-partially-estimated)

## Description

This PR:
1. Removes estimated badge when less than 1 %
2. Removes partially estimated badge in hourly view. Only show estimated badge if all hours are estimated
3. Show 'partially estimated' instead of 'estimated' on aggregated views if the graph contains some measured data.

### Preview

1 Before:
image
1 Now: 
image

2 Before:
image
2 After:
image

3 Before:
image
3 After:
image

### Double check

### DIFF
--- a/web/src/features/charts/graphUtils.ts
+++ b/web/src/features/charts/graphUtils.ts
@@ -274,3 +274,13 @@ export function extractLinkFromSource(
   // We on purpose don't use https due to some sources not supporting it (and the majority that does will automatically redirect anyway)
   return `http://${source}`;
 }
+
+export function hasSignificantEstimation(
+  estimationMethod: EstimationMethods | undefined,
+  estimatedPercentage: number | undefined
+) {
+  return (
+    Boolean(estimationMethod) ||
+    (estimatedPercentage != undefined && estimatedPercentage > 1)
+  );
+}
--- a/web/src/features/charts/graphUtils.ts
+++ b/web/src/features/charts/graphUtils.ts
@@ -5,7 +5,7 @@ import { TFunction } from 'i18next';
 import { CircleDashed, LucideIcon, TrendingUpDown } from 'lucide-react';
 import { MouseEvent } from 'react';
 import { ElectricityStorageType, GenerationType, Maybe, ZoneDetail } from 'types';
-import { EstimationMethods, modeOrder } from 'utils/constants';
+import { EstimationMethods, modeOrder, TimeAverages } from 'utils/constants';
 import { formatCo2, formatEnergy, formatPower } from 'utils/formatting';
 
 import { AreaGraphElement } from './types';
@@ -211,9 +211,11 @@ export function getElectricityProductionValue({
   return generationTypeStorage === 0 ? 0 : -generationTypeStorage;
 }
 
-function analyzeChartData(chartData: AreaGraphElement[]) {
+function analyzeChartData(chartData: AreaGraphElement[], timeAverage: TimeAverages) {
   let estimatedCount = 0;
   let tsaCount = 0;
+  let allEstimated = true;
+  const isHourly = timeAverage === TimeAverages.HOURLY;
   for (const chartElement of chartData) {
     if (chartElement.meta.estimationMethod === EstimationMethods.TSA) {
       tsaCount++;
@@ -221,20 +223,29 @@ function analyzeChartData(chartData: AreaGraphElement[]) {
     if (chartElement.meta.estimatedPercentage || chartElement.meta.estimationMethod) {
       estimatedCount++;
     }
+    if (
+      (!isHourly && chartElement.meta.estimatedPercentage !== 100) ||
+      (isHourly && !chartElement.meta.estimationMethod)
+    ) {
+      allEstimated = false;
+    }
   }
   return {
     allTimeSlicerAverageMethod: tsaCount === chartData.length,
-    allEstimated: estimatedCount === chartData.length,
-    hasEstimation: estimatedCount > 0,
+    allEstimated,
+    hasEstimation: isHourly ? allEstimated : estimatedCount > 0,
   };
 }
 
 export function getBadgeTextAndIcon(
   chartData: AreaGraphElement[],
-  t: TFunction
+  t: TFunction,
+  timeAverage: TimeAverages
 ): { text?: string; icon?: LucideIcon } {
-  const { allTimeSlicerAverageMethod, allEstimated, hasEstimation } =
-    analyzeChartData(chartData);
+  const { allTimeSlicerAverageMethod, allEstimated, hasEstimation } = analyzeChartData(
+    chartData,
+    timeAverage
+  );
   if (allTimeSlicerAverageMethod) {
     return {
       text: t(`estimation-card.${EstimationMethods.TSA}.pill`),
--- a/web/src/features/charts/CarbonChart.tsx
+++ b/web/src/features/charts/CarbonChart.tsx
@@ -30,7 +30,7 @@ function CarbonChart({ datetimes, timeAverage }: CarbonChartProps) {
 
   const hasEnoughDataToDisplay = datetimes?.length > 2;
 
-  const { text, icon } = getBadgeTextAndIcon(chartData, t);
+  const { text, icon } = getBadgeTextAndIcon(chartData, t, timeAverage);
   const badge = <EstimationBadge text={text} Icon={icon} />;
 
   if (!hasEnoughDataToDisplay) {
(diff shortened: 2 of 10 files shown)


silkeholmebonnen:
> I'm not sure you have seen the discussion in #7266 but there I suggested we use percentages everywhere instead of Estimated vs Partially Estimated. (this was and dirty PR before the linear issue was created).
> 
> But I think there is some value is showing the percentages all the time. Especially if we can increase the granularity of the estimation data (it exists in the database).

I see your point and it would be cool to do it later, but I think this is an improvement alone. Maybe we can leave this PR as it is but add it as a task for later?

web/src/features/charts/graphUtils.ts
```diff
@@ -211,30 +211,41 @@ export function getElectricityProductionValue({
   return generationTypeStorage === 0 ? 0 : -generationTypeStorage;
 }
 
-function analyzeChartData(chartData: AreaGraphElement[]) {
+function analyzeChartData(chartData: AreaGraphElement[], timeAverage: TimeAverages) {
   let estimatedCount = 0;
   let tsaCount = 0;
+  let allEstimated = true;
+  const isHourly = timeAverage === TimeAverages.HOURLY;
```
tonypls, line 218: Would be good to use the isHourly atom here

web/src/features/charts/graphUtils.ts
```diff
@@ -211,30 +211,41 @@ export function getElectricityProductionValue({
...
+    ) {
+      allEstimated = false;
+    }
   }
   return {
     allTimeSlicerAverageMethod: tsaCount === chartData.length,
-    allEstimated: estimatedCount === chartData.length,
-    hasEstimation: estimatedCount > 0,
+    allEstimated,
+    hasEstimation: isHourly ? allEstimated : estimatedCount > 0,
```
VIKTORVAV99, line 236: I would prefer if this was renamed, to represent the new state.

Now it rather means all estimated or has estimations.

Not sure what a good name is tho.

web/src/features/charts/graphUtils.ts
```diff
@@ -211,30 +211,41 @@ export function getElectricityProductionValue({
...
-function analyzeChartData(chartData: AreaGraphElement[]) {
+function analyzeChartData(chartData: AreaGraphElement[], timeAverage: TimeAverages) {
   let estimatedCount = 0;
   let tsaCount = 0;
+  let allEstimated = true;
+  const isHourly = timeAverage === TimeAverages.HOURLY;
   for (const chartElement of chartData) {
     if (chartElement.meta.estimationMethod === EstimationMethods.TSA) {
       tsaCount++;
     }
```
VIKTORVAV99, line 222: This would need to include the `FORECASTS_HIERARCHY` method as well if we want to ensure it's functioning the same way for Germany and other zones in the future.

VIKTORVAV99:
Maybe I am missing something but why are we doing this part:
> Removes partially estimated badge in hourly view. Only show estimated badge if all hours are estimated

As in why are we not showing partially estimated when only 1-2 hours are estimated? For some zones this is always the case, for example Greece and Shi Lanka.

It means you won't be able to tell this contains estimates if you don't use the tooltip (which is why we added the partially estimated on in the first place).
image
VS
image

tonypls:
> Maybe I am missing something but why are we doing this part:
> 
> > Removes partially estimated badge in hourly view. Only show estimated badge if all hours are estimated
> 
> As in why are we not showing partially estimated when only 1-2 hours are estimated? For some zones this is always the case, for example Greece and Shi Lanka.
> 
> It means you won't be able to tell this contains estimates if you don't use the tooltip (which is why we added the partially estimated on in the first place). image VS image

It's a trade-off! We're leaning more into having confidence in our estimations and relying on the tooltips to handle the cases like you've mentioned.

Here's a case where the data is 2% estimated but it's broadly badged with "Partially Estimated" from our time creating social media posts we found this was often a hurdle to us feeling like the data is accurate enough to share, which I think it more often than not is.

image

I believe we will soon know if this is leaning to far the other way quite quickly and will happily revert if that's the case. Either way we have high conviction rounding of <1% change is a positive step!

VIKTORVAV99:
> > Maybe I am missing something but why are we doing this part:
> 
> > 
> 
> > > Removes partially estimated badge in hourly view. Only show estimated badge if all hours are estimated
> 
> > 
> 
> > As in why are we not showing partially estimated when only 1-2 hours are estimated? For some zones this is always the case, for example Greece and Shi Lanka.
> 
> > 
> 
> > It means you won't be able to tell this contains estimates if you don't use the tooltip (which is why we added the partially estimated on in the first place). image VS image
> 
> 
> 
> It's a trade-off! We're leaning more into having confidence in our estimations and relying on the tooltips to handle the cases like you've mentioned.
> 
> 
> 
> Here's a case where the data is 2% estimated but it's broadly badged with "Partially Estimated" from our time creating social media posts we found this was often a hurdle to us feeling like the data is accurate enough to share, which I think it more often than not is.
> 
> 
> 
> image
> 
> 
> 
> I believe we will soon know if this is leaning to far the other way quite quickly and will happily revert if that's the case. Either way we have high conviction rounding of <1% change is a positive step!
> 
> 

Which is why I switched to only use percentages in #7266 as it gives a clear indication of exactly how much is estimated. (And hides it if it's less than 1%). (That PR predates the linear issue)

I just think we might be shooting ourselves in the foot by not being overly clear when the next time ENRSOE does maintenance and something weird happens.

tonypls:
> > > Maybe I am missing something but why are we doing this part:
> > 
> > 
> > > 
> > 
> > 
> > > > Removes partially estimated badge in hourly view. Only show estimated badge if all hours are estimated
> > 
> > 
> > > 
> > 
> > 
> > > As in why are we not showing partially estimated when only 1-2 hours are estimated? For some zones this is always the case, for example Greece and Shi Lanka.
> > 
> > 
> > > 
> > 
> > 
> > > It means you won't be able to tell this contains estimates if you don't use the tooltip (which is why we added the partially estimated on in the first place). image VS image
> > 
> > 
> > It's a trade-off! We're leaning more into having confidence in our estimations and relying on the tooltips to handle the cases like you've mentioned.
> > Here's a case where the data is 2% estimated but it's broadly badged with "Partially Estimated" from our time creating social media posts we found this was often a hurdle to us feeling like the data is accurate enough to share, which I think it more often than not is.
> > image
> > I believe we will soon know if this is leaning to far the other way quite quickly and will happily revert if that's the case. Either way we have high conviction rounding of <1% change is a positive step!
> 
> Which is why I switched to only use percentages in #7266 as it gives a clear indication of exactly how much is estimated. (And hides it if it's less than 1%). (That PR predates the linear issue)
> 
> I just think we might be shooting ourselves in the foot by not being overly clear when the next time ENRSOE does maintenance and something weird happens.

Sorry, yes there is some double work that has been done here. Regardless of what solution we use, let's get it shipped!

silkeholmebonnen:
@VIKTORVAV99 I see pros and cons of both solutions. But I think the ENTSOE maintenance is a good reason, also one of the main reasons why we got the labels in the first place. So I think we should get your PR out, but hiding the badge if less than 1%.

tonypls:
> @VIKTORVAV99 I see pros and cons of both solutions. But I think the ENTSOE maintenance is a good reason, also one of the main reasons why we got the labels in the first place. So I think we should get your PR out, but hiding the badge if less than 1%.

Thanks Silke! Let's merge @VIKTORVAV99 's PR. 

I think the part 3 or showing partially estimated is still super nice here! Could we trim down this PR to improve that when the other PR is merged in?

Would it be valuable to only show the hourly preliminary chart badge for certain zones? If parser is entsoe? The motivation here is hurdles from our SoMe posts sessions when trying to share US data.

silkeholmebonnen:
> I think the part 3 or showing partially estimated is still super nice here! Could we trim down this PR to improve that when the other PR is merged in?

Now it shows the estimated percentage, so that would kind of override the changes viktor made in his PR.

> Would it be valuable to only show the hourly preliminary chart badge for certain zones? If parser is entsoe? The motivation here is hurdles from our SoMe posts sessions when trying to share US data.

I feel like it would be hard to justify if we had to explain it to someone outside the organization. Or what would you say if you had to explain it to someone not from EMaps?

silkeholmebonnen:
The only thing I think we still need to do is "1. remove estimated badge when less than 1 %". Viktors PR only did that for the chart but not for maptooltip, barbreakdownchart and estimation accordion. But I think this might be easier to do in a new branch.

VIKTORVAV99:
> The only thing I think we still need to do is "1. remove estimated badge when less than 1 %". Viktors PR only did that for the chart but not for maptooltip, barbreakdownchart and estimation accordion. But I think this might be easier to do in a new branch.

Yeah originally it was only mentioned for the charts, but I think this is a good follow up.

Sorry for causing some friction with your work with the other PR so if you need any assistance don't hesitate to let me know.

PR #7509, opened 2024-12-02
chore(deps): Update i18next to v24

VIKTORVAV99:
## Description

i18next has released a new major version that drops support for some of their oldest backwards compatibility features, we don't use any of these so for us this will be pure gains in terms of performance and bundle size.

### Double check

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -88,7 +88,7 @@
     "d3-shape": "^3.2.0",
     "date-fns": "^3.6.0",
     "geojson": "^0.5.0",
-    "i18next": "^23.16.8",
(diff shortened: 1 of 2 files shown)


cadeban (approved):
I skimmed the changelog and didn't see anything that was pertinent to us. I tested the app on the branch too and it LGTM.

VIKTORVAV99:
> I skimmed the changelog and didn't see anything that was pertinent to us. I tested the app on the branch too and it LGTM.

Haha I might have already reviewed the upstream PR in the i18n repository to ensure everything was working for us, might have had a hand in creating it as well 😅

PR #7512, opened 2024-12-03
feat(web): service worker update button mobile

tonypls:
## Issue
When debugging on mobile it's not possible to test the service worker update path without waiting

## Description
This PR adds an update button to the info modal and adds some extra catching on the SW update function


web/src/features/modals/InfoModal.tsx
```diff
@@ -31,7 +40,18 @@ export function InfoModalContent() {
...
+        {needRefresh && (
+          <button
+            onClick={handleUpdate}
+            className="rounded bg-brand-green px-2 py-1 text-xs text-white hover:bg-brand-green/90"
+            title={t('updatePrompt.update')}
+          >
+            {t('updatePrompt.update')}
+          </button>
+        )}
+      </div>
```
VIKTORVAV99, line 54: While this might be a good addition I've been using the service worker controls in the developer console (in chrome) to do this.

web/src/features/service-worker/UpdatePrompt.tsx
```diff
@@ -12,17 +12,22 @@ function UpdatePrompt() {
...
-          import.meta.env.PROD ? ONE_HOUR : 10 * 1000
-        );
+      if (registration) {
+        const checkForUpdates = async () => {
+          console.info('Checking for app update...');
+          try {
+            await registration.update();
+          } catch (error) {
+            console.warn('Failed to check for updates:', error);
+          }
```
VIKTORVAV99, line 22: Nice addition!

web/src/features/service-worker/UpdatePrompt.tsx
```diff
@@ -12,17 +12,22 @@ function UpdatePrompt() {
...
+        };
+
+        setInterval(checkForUpdates, import.meta.env.PROD ? ONE_HOUR : 10 * 1000);
+      }
     },
     onRegisterError(error) {
-      console.error(`SW registration failed: ${error}`);
+      console.warn('SW registration failed:', error);
+      setNeedRefresh(false);
```
VIKTORVAV99, line 30: Not sure we should set this to false. Wouldn't that risk making people stuck on an old version?

web/src/features/modals/InfoModal.tsx
tonypls: The problem I find is when I'm trying to test on my mobile device

web/src/features/service-worker/UpdatePrompt.tsx
tonypls: Have removed it!

web/src/features/modals/InfoModal.tsx
VIKTORVAV99: Perhaps the settings modal would be more suited to that then?

I don't think we should be putting it in the info modal.

tonypls: I thought to put it here as it's right next to the version number and it only appears when an update is required, mostly as a dev feature.

web/src/features/service-worker/UpdatePrompt.tsx
tonypls: Hopefully this gives some more insight to our sentry errors.

web/src/features/modals/InfoModal.tsx
VIKTORVAV99: Makes sense but in the future I think we could have it as a non dev feature as well. To force updates when people are having issues.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7514, opened 2024-12-03
chore(web): use consistent testid naming

tonypls:
## Issue

userEvent.click(screen.getByTestId('dismiss-btn')) works only when the attribute is named data-testid and not data-test-id is due to how the testing library is configured to query elements.
Explanation:
Testing Library Convention:
The @testing-library/react library, which includes screen.getByTestId, is designed to work with the data-testid attribute by default. This is a convention used by the library to identify elements for testing purposes.

## Description
This PR renames all `data-test-id` attributes to `data-testid`


VIKTORVAV99 (approved):
Smaller bundles! Love it!


LGTM! 🎉

PR #7516, opened 2024-12-03
bug(web): wrong query key on zone request

tonypls:
## Issue
Out of sync state and zone data

## Description
I was still seeing the annoying out of sync data issue, after some debugging I found the query key was set to "24h" instead of "hourly" this fixes the issue. It would be nice to align on one standard to aggregations in the future.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7518, opened 2024-12-04
feat(web): include more options drop down in completed shares

tonypls:
## Issue
We're not tracking completed shares from the more options dropdown

## Description
This PR updates the useShare hook to include completed share tracking


web/src/features/panels/zone/ShareButton.tsx
```diff
@@ -44,23 +44,16 @@ export function ShareButton({
...
-      ).then((shareCompleted) => {
-        if (shareCompleted) {
-          trackShareCompletion();
-        }
-      });
-    } else {
-      copyToClipboard(shareUrl, toastMessageCallback);
-    }
-    trackShareClick();
+    await (hasMobileUserAgent && (await CapShare.canShare())
```
cadeban, line 47: In my opinion this is more readable as an if statement. Now a reader is required to know the precedence of `await` vs the ternary operator. It's also a little strange to have a single await applied to multiple branches.

Snce we only care about the side effects here, I think an imperative statement more accurately describes the intent of the code.

web/src/utils/analytics.ts
```diff
@@ -37,8 +37,10 @@ export enum ShareType {
   COMPLETED_SHARE = 'completed_share',
 }
 
-export const trackShare = (shareType: ShareType) => () =>
+export const trackShare = (shareType: ShareType) => () => {
+  console.log('share', shareType);
```
cadeban, line 41: nit: console.log here

web/src/hooks/useShare.ts
```diff
@@ -21,7 +26,11 @@ export function useShare() {
   const share = useCallback(
     async (shareData: ShareOptions, callback?: (argument: string) => void) => {
       try {
+        trackShareClick();
         const result = await CapShare.share(shareData);
+        if (result) {
+          trackShareCompletion();
```
cadeban, line 32: 🎉

web/src/utils/analytics.ts
tonypls: Good spotting!

web/src/features/panels/zone/ShareButton.tsx
tonypls: Same but our prettier rules force it to be ternary 🤦

cadeban: I should've known! 😧

cadeban: I'll approve this but I might look into changing Prettier when I have some down time...

cadeban (approved):

PR #7522, opened 2024-12-05
fix: Quick fix for the rounding logic in analyzeChartData

VIKTORVAV99:
## Description
Quick fix for the math in the function.

### Double check

### DIFF
--- a/web/src/features/charts/graphUtils.ts
+++ b/web/src/features/charts/graphUtils.ts
@@ -230,7 +230,7 @@ function analyzeChartData(chartData: AreaGraphElement[]) {
     estimatedTotal += chartElement.meta.estimatedPercentage ?? 0;
   }
   const calculatedTotal = round(
-    (estimatedTotal / total || (estimatedCount || tsaCount) / chartData.length) * 100,
+    estimatedTotal / total || ((estimatedCount || tsaCount) / total) * 100,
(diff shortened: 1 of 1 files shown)


silkeholmebonnen (commented):
What is FORECASTS_HIERARCHY and is it equivalent to TSA?

VIKTORVAV99:
> What is FORECASTS_HIERARCHY and is it equivalent to TSA?

Because it's the same as the TSA model but it replaces the solar and wind output with forecasts from ENTSO-E when they are available.

silkeholmebonnen:
It seems a bit weird that the aggregated percentage is larger than any of the years it is aggregating. Do you know why this happens?


https://github.com/user-attachments/assets/e2c7a763-589d-49ff-9eb8-87bcf2c55515

VIKTORVAV99:
> It seems a bit weird that the aggregated percentage is larger than any of the years it is aggregating. Do you know why this happens?
> 
>  calc.mov

The multiplication was in the wrong side of the parentheses 🤦🏻‍♂️

silkeholmebonnen (approved):
LGTM! 

I still think that we should think about if we want to hide the estimation percentage if e.g. only one datapoint is estimated. But this we can figure that out after, let's merge this!

PR #7528, opened 2024-12-06
Archive IQ (Iraq) parser

### LINKED ISSUE #7487: [Data Issue]: Data not available for IQ (Iraq) parser?
### When did this happen?

Observed locally attempting to run the IQ (Iraq) parser and access the dashboard URL referenced in the corresponding code.

### What zones are affected?

IQ, IQ-KUR

### What is the problem?

I've been examining the status of various parsers for #6135, and as part of that process, began reading the IQ (Iraq) parser code.

Although the relevant website is still online, the URL retrieved by the parser ( https://www.gdoco.org/vueips.php ) does not appear to return relevant generation/dispatch data at the moment.

I also checked a previous dashboard API IP address used previously ( until b54f446e2abb0d2391c67f65263723918a7bc15d ) and confirmed that that it remains non-functional.

jayaddison:
## Issue

Closes #7487.

## Description

Archive the parser and ~~remove the zone and exchange configurations~~ remove the `parser` entries in zone and exchange configurations for `IQ` (Iraq) - the URL referenced in the existing logic no longer contains data, and other parts of the website appear to be behind a non-public login / authentication step.  I'm not aware of other datasources currently (although if we find one perhaps we can repair the parser instead).

### Preview

N/A

### Double check

- [ ] ~~I have tested my parser changes locally with `poetry run test_parser "zone_key"`~~
- [ ] ~~I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.~~


VIKTORVAV99 (changes requested):
This is not just archiving the parser this is removing the zone config and exchange configs as well. This is not we can just do on a whim even if there is no working parser.

We have historical data for 2018 for Iraq so this will have a real impact on internal and external systems not to mention that the geographies are not updated so the app will have a new ocean in its place.


Archiving the parser is fine, removing the configs are not.

jayaddison:
Gotcha, OK - thanks @VIKTORVAV99.  I'll go back to inspect some previous archived parsers to check the process.  In particular I'm wondering what to do about the references to `IQ.fetch_consumption` that will remain after the parser is archived.

jayaddison:
I think / hope this is ready now - the changeset archives the parser code, and removes only the `parsers` references from the exchange/zone config, while leaving the geo polygons unmodified.

VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
I think it should be fine to merge but just in case I've asked for a second look before I merge this.

FelixDQ (approved):

jayaddison:
I'm not certain, but I'm wondering whether the Python-related CI failures here may be something to do with the in-progress transition to Ubuntu 24.04 for GitHub Actions runner images (ref: https://github.com/actions/runner-images/issues/10636).

A previous successful run used Ubuntu 22.04: https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12198657484/job/34030884220#step:1:8

And the same job definition failed from commit 2bc90bb2e1ab333824a033245931006d2ff371fb used Ubuntu 24.04: https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12256651021/job/34192388545#step:1:8

VIKTORVAV99:
Hmm it might be the cache, I'm going to try a few things and if that don't work delete the old cache entry.

jayaddison:
Thanks - I have been wondering about caching too, but haven't figured out how that could cause the [selection of Python3.12 as the interpreter for the failing build](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12256651021/job/34192388545#step:9:12) (that I think may be relevant).. puzzling.

VIKTORVAV99:
> Thanks - I have been wondering about caching too, but haven't figured out how that could cause the [selection of Python3.12 as the interpreter for the failing build](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12256651021/job/34192388545#step:9:12) (that I think may be relevant).. puzzling.

I think it restored a cache that was incompatible with the new version because either the binaries or the locations changed and therefore when it tried to run it nothing was available where it was expected to be.

And since the cache was restored the install step was skipped which probably would have solved the issue on it's own.

jayaddison:
Yep, OK, that makes sense.  A stale virtualenv referencing a no-longer-existent Python binary -- and the stale/incorrect cache that was loaded from should have been from within this branch itself (GitHub is fairly careful about cache re-use from unrelated branches).

I think it was probably from 97ab630f2da9e2611e4cc324e7dc1c28d72a5872 / https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12204617678/job/34050116470 (the cache [did not exist](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12204617678/job/34050116470#step:6:18) when that job began, and it was [created](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12204617678/job/34050116470#step:16:5) during its completion; as [apparent](https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12204617678/job/34050116054#step:16:3) in a parallel job from the same commit).

(apologies, I'm being a bit verbose / distracting, but I wanted to gain more understanding of what happened there)

jayaddison:
> I think this mostly relates to our own caching logic. But as you mention it could be a race condition as well but then one of the jobs should have worked I think.

Yep, provided that the cache created by each of them is idempotent-enough (it should be: caches can be created at fairly arbitrary times and then re-used later) and that the writing process is reasonably locked (it appears to be, since the latter job detected an in-process creation), I think the parallelism/race should be fine.

> I will look into the cache key logic but it seems like we have to add something ourselves to make it work the way I'd expect it to as the github contexts seem to be missing a runner version context/variable we can use.

Thanks - yep, that's the approach I'd follow too.  I'm going to try to determine whether it is `poetry`-specific (e.g. do other Python GitHub Actions workflow caches include entire virtualenv contents, including binary symlinks).

jayaddison:
This could be an instance of https://github.com/actions/setup-python/issues/374 ([installing `poetry`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L539-L541) _before_ [using `actions/setup-python`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L546-L549) means that `poetry` is bound to the system-installed Python environment, and in our case the cache retrieved was intended for an earlier version (the latter already identified as a factor))

jayaddison:
> This could be an instance of [actions/setup-python#374](https://github.com/actions/setup-python/issues/374) ([installing `poetry`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L539-L541) _before_ [using `actions/setup-python`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L546-L549) means that `poetry` is bound to the system-installed Python environment, and in our case the cache retrieved was intended for an earlier version (the latter already identified as a factor))

There is a potentially relevant note in the GitHub Actions `setup-python` action:

https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/docs/advanced-usage.md?plain=1#L323

> **Note:** If the `setup-python` version does not match the version specified in `pyproject.toml` and the python version in `pyproject.toml` is less than the runner's python version, `poetry install` will default to using the runner's Python version.

I discovered this thanks to a comment at https://github.com/actions/setup-python/issues/940#issuecomment-2348892922

In `electricitymaps-contrib`, the `pyproject.toml` file and also the `.python-version` file both lock to Python3.10 currently ([ref](https://github.com/electricitymaps/electricitymaps-contrib/blob/df31a2b3d41d906731f6feb3f36f10c241d7bae7/pyproject.toml#L11), [ref](https://github.com/electricitymaps/electricitymaps-contrib/blob/df31a2b3d41d906731f6feb3f36f10c241d7bae7/.python-version)) -- so I think the note is relevant (the runner Python version would be used).

:warning: I think this implies that either pinning the runner version (to `ubuntu-22.04`), OR updating the Python version in this repository, OR adjusting the GitHub Actions / cache workflow will be required at some point soon.

> I will look into the cache key logic but it seems like we have to add something ourselves to make it work the way I'd expect it to as the github contexts seem to be missing a runner version context/variable we can use.

One of the GitHub Actions `cache` action example for `pipenv` -- admittedly not `poetry`, but perhaps not too dissimilar -- mentions `${{ os.runner }}` that I think might achieve this?

https://github.com/actions/cache/blob/1bd1e32a3bdc45362d1e726936510720a7c30a57/examples.md?plain=1#L560-L563

If that _does_ work, then I'd suggest that could be a good option.  And if it works here, I'd  be happy to offer a pull request to `actions/cache` to suggest adding that as a `poetry` example for their `examples.md` file.

(apologies for the probably-too-verbose commentary.  there are likely quicker ways to keep things working that wouldn't require such pedantic detail)

VIKTORVAV99:
> > This could be an instance of [actions/setup-python#374](https://github.com/actions/setup-python/issues/374) ([installing `poetry`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L539-L541) _before_ [using `actions/setup-python`](https://github.com/electricitymaps/electricitymaps-contrib/blob/56d4c3d2fbbe75d617f9632fff852910ae33462a/.github/workflows/CI.yml#L546-L549) means that `poetry` is bound to the system-installed Python environment, and in our case the cache retrieved was intended for an earlier version (the latter already identified as a factor))
> 
> There is a potentially relevant note in the GitHub Actions `setup-python` action:
> 
> https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/docs/advanced-usage.md?plain=1#L323
> 
> > **Note:** If the `setup-python` version does not match the version specified in `pyproject.toml` and the python version in `pyproject.toml` is less than the runner's python version, `poetry install` will default to using the runner's Python version.
> 
> I discovered this thanks to a comment at [actions/setup-python#940 (comment)](https://github.com/actions/setup-python/issues/940#issuecomment-2348892922)
> 
> In `electricitymaps-contrib`, the `pyproject.toml` file and also the `.python-version` file both lock to Python3.10 currently ([ref](https://github.com/electricitymaps/electricitymaps-contrib/blob/df31a2b3d41d906731f6feb3f36f10c241d7bae7/pyproject.toml#L11), [ref](https://github.com/electricitymaps/electricitymaps-contrib/blob/df31a2b3d41d906731f6feb3f36f10c241d7bae7/.python-version)) -- so I think the note is relevant (the runner Python version would be used).
> 
> ⚠️ I think this implies that either pinning the runner version (to `ubuntu-22.04`), OR updating the Python version in this repository, OR adjusting the GitHub Actions / cache workflow will be required at some point soon.

We generally don't want to pin the version of the runner as that increases maintenance for us (more so than deleting the cache), increasing the Python version is something we will do as soon as all the dependencies support it (both in here and in our backend). But last I checked we where limited by pandas and similar libraries.


> > I will look into the cache key logic but it seems like we have to add something ourselves to make it work the way I'd expect it to as the github contexts seem to be missing a runner version context/variable we can use.
> 
> One of the GitHub Actions `cache` action example for `pipenv` -- admittedly not `poetry`, but perhaps not too dissimilar -- mentions `${{ os.runner }}` that I think might achieve this?
> 
> https://github.com/actions/cache/blob/1bd1e32a3bdc45362d1e726936510720a7c30a57/examples.md?plain=1#L560-L563
> 
> If that _does_ work, then I'd suggest that could be a good option. And if it works here, I'd be happy to offer a pull request to `actions/cache` to suggest adding that as a `poetry` example for their `examples.md` file.
> 
> (apologies for the probably-too-verbose commentary. there are likely quicker ways to keep things working that wouldn't require such pedantic detail)

We are already using this key `key: ${{ runner.os }}-venv-poetry_${{steps.get-poetry-version.outputs.poetry-version}}-${{ hashFiles('poetry.lock') }}` but it don't seem to over the  edge case we spotted here. So not sure how much else we can do with the default contexts so we might have to add another step that captures the OS versions as well.


Don't worry about providing too much context, that is always a good thing IMO!

jayaddison:
> We are already using this key key: ${{ runner.os }}-venv-poetry_${{steps.get-poetry-version.outputs.poetry-version}}-${{ hashFiles('poetry.lock') }} but it don't seem to over the edge case we spotted here. So not sure how much else we can do with the default contexts so we might have to add another step that captures the OS versions as well.

Thanks @VIKTORVAV99.  And ok, yep - I should have double-checked that detail before commenting! (I'd misinterpreted a previous comment of yours to mean that there wasn't any reference to the runner system/platform in the cache key).

I suppose a naive/simplistic approach to cachebreak could be to upgrade the runner image explicitly to use `ubuntu-24.04` when ready _and_ to change the cache key during the same changeset.

jayaddison:
> I suppose a naive/simplistic approach to cachebreak could be to upgrade the runner image explicitly to use `ubuntu-24.04` when ready _and_ to change the cache key during the same changeset.

...or perhaps configuring `with: cache: 'poetry'` on the `setup-python` action (mentioned slightly above the [note in a previous link](https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/docs/advanced-usage.md?plain=1#L323)) would handle the edge-case for us (?).

(I'm not sure yet, and doing some more reading about this)

jayaddison:
> > I suppose a naive/simplistic approach to cachebreak could be to upgrade the runner image explicitly to use `ubuntu-24.04` when ready _and_ to change the cache key during the same changeset.
> 
> ...or perhaps configuring `with: cache: 'poetry'` on the `setup-python` action (mentioned slightly above the [note in a previous link](https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/docs/advanced-usage.md?plain=1#L323)) would handle the edge-case for us (?).
> 
> (I'm not sure yet, and doing some more reading about this)

Using `with: cache: 'poetry'` might be an improvement for simplicity/readability purposes, but would it result in a cache miss during runner image change?

The [relevant `actions/setup-python` code](https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/src/cache-distributions/poetry-cache.ts#L50) shows that it does use `RUNNER_OS` -- which is the same as `runner.os` in the [GitHub Actions runner context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context) -- and the (host?) Python version.

The inclusion of the host Python version might be enough to cause a cache miss here.  I think it'd be possible to test that theory by opening a pull request explicitly using `ubuntu-22.04` for the Python Tests part of `CI.yml` plus `cache: 'poetry'`, and then subsequently commiting a single change to update to `ubuntu-24.04` for that same workflow job only.

I'm not sure whether I'll attempt that yet - but that is what I'm thinking about currently.

VIKTORVAV99:
> > > I suppose a naive/simplistic approach to cachebreak could be to upgrade the runner image explicitly to use `ubuntu-24.04` when ready _and_ to change the cache key during the same changeset.
> > 
> > 
> > ...or perhaps configuring `with: cache: 'poetry'` on the `setup-python` action (mentioned slightly above the [note in a previous link](https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/docs/advanced-usage.md?plain=1#L323)) would handle the edge-case for us (?).
> > (I'm not sure yet, and doing some more reading about this)
> 
> Using `with: cache: 'poetry'` might be an improvement for simplicity/readability purposes, but would it result in a cache miss during runner image change?
> 
> The [relevant `actions/setup-python` code](https://github.com/actions/setup-python/blob/0b93645e9fea7318ecaed2b359559ac225c90a2b/src/cache-distributions/poetry-cache.ts#L50) shows that it does use `RUNNER_OS` -- which is the same as `runner.os` in the [GitHub Actions runner context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context) -- and the (host?) Python version.
> 
> The inclusion of the host Python version might be enough to cause a cache miss here. I think it'd be possible to test that theory by opening a pull request explicitly using `ubuntu-22.04` for the Python Tests part of `CI.yml` plus `cache: 'poetry'`, and then subsequently commiting a single change to update to `ubuntu-24.04` for that same workflow job only.
> 
> I'm not sure whether I'll attempt that yet - but that is what I'm thinking about currently.

As far as I know (might be wrong here) poetry only caches the dependencies themselves not the venv which means we still would have a install step that recreates that. Might not be a huge deal though but would probably not be as performant as caching the whole venv and skipping the whole install step which we do now.

We will have to do some testing to see how it works (and how long it takes).

jayaddison:
I could be mistaken, but I _think_ that `actions/setup-python` with `cache: poetry` enabled will cache the entire relevant (multiple may exist) `virtualenv` created by `poetry`.  Very much +1 to testing required, though.

jayaddison:
When I tested the change I had in mind -- approximately: to use `actions/setup: cache: poetry` for the Python-related GitHub Actions workflow jobs -- in my own forked repository, tests passed across a migration of the base runner image from Ubuntu 22.04 to 24.04 -- but not for the reason I expected.  I had expected the cache to be invalidated by the upgrade (thanks to the host Python version number changing).  Instead it appears that the job on the 24.04 runner was able to use Python3.10 as intended and re-use the `hostedtoolcache` binaries to successfully run and complete the tests.

That puzzles me because I expected Python3.12 to be (unexpectedly) activated, as it was during the earlier GitHub Actions CI job failure: https://github.com/electricitymaps/electricitymaps-contrib/actions/runs/12256651021/job/34192388545#step:9:12

So: I don't think I'm going to suggest any changes to CI, because something about my reasoning is incomplete or flawed, but I think my testing illustrates that some alternative workflow job configuration options are available.

PR #7530, opened 2024-12-07
Pumping storage fix for CZ parser

### LINKED ISSUE #5426: [Data Issue]: new CZ (CEPS) parser is missing pumping demand for pumped hydro

dermotmolony:
## Issue

Closes #5426

## Description

Creates get_pumping_load function which fetches pumping load data from ceps.cz and computes the current load at each timestamp that is being used for pumping. 

Modifies fetch_production to call get_pumping_load, whose dictionary is then used to add the amount of energy being stored to hydro at each timestamp.

### Preview

The hydro storage values from fetch_production() showing the new positive values calculated from load_including_pumping-load
image

### Double check

- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"` - I got the error marking ":( >2h from now !!!" but the original CZ parser code produces the exact same error locally. I tried to see if it was a time conversion issue but didn't have any luck locating the source of the issue

### DIFF
--- a/parsers/CZ.py
+++ b/parsers/CZ.py
@@ -2,16 +2,11 @@ from datetime import datetime, timedelta
 from logging import Logger, getLogger
 
 from bs4 import BeautifulSoup
-
 # The request library is used to fetch content through HTTP
 from requests import Response, Session
 
 from electricitymap.contrib.lib.models.event_lists import (
-    ExchangeList,
-    ProductionBreakdownList,
-    ProductionMix,
-    StorageMix,
-)
+    ExchangeList, ProductionBreakdownList, ProductionMix, StorageMix)
 from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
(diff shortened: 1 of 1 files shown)


parsers/CZ.py
```diff
@@ -141,6 +141,51 @@ def __get_exchange_data(
     return exchanges.to_list()
 
 
+def get_pumping_load(
+    zone_key: ZoneKey = ZoneKey("CZ"),
+    session: Session = Session(),
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
```
VIKTORVAV99, line 149: This has the return type list.

parsers/CZ.py
```diff
@@ -141,6 +141,51 @@ def __get_exchange_data(
...
+            )
+
+    else:
+        ParserException(
+            "CZ.py",
+            f"There was no pumping load data returned for {zone_key} at {target_datetime}",
+            zone_key,
+        )
+
+    return pumping_load_at_time_dict
```
VIKTORVAV99, line 186: But this is returning a dict.

parsers/CZ.py
```diff
@@ -141,6 +141,51 @@ def __get_exchange_data(
...
+    content = make_request(session, payload, zone_key).text
+    xml = BeautifulSoup(content, "xml")
+    data_tag = xml.find("data")
+    pumping_load_at_time_dict = {}
+
+    if data_tag is not None:
+        for values in data_tag:
+            pumping_load_at_time_dict[values["date"]] = float(values["value1"]) - float(
+                values["value2"]
+            )
```
VIKTORVAV99, line 177: This also contains type errors. Such as `"__getitem__" method not defined on type "PageElement"`

parsers/CZ.py
```diff
@@ -174,17 +219,28 @@ def fetch_production(
...
-
             for k, v in mapper.items():
                 generator = translate_table_gen[k]
                 if k != "PsPP":
                     production.add_value(mode=generator, value=float(values[v]))
                 else:
-                    storage.add_value(mode=generator, value=float(values[v]) * -1)
+                    pumping_storage = float(values[v]) * -1
+                    if values["date"] in pumping_load_at_time:
+                        pumping_storage += pumping_load_at_time[values["date"]]
```
VIKTORVAV99, line 240: Why are you summing things up here?

If there are multiple storage entries they should be looped over and add_value should be called on each of them as it handles summing up values of the same type.

VIKTORVAV99 (changes requested):
There are some type issues to look at and how to better use the parser functions as intended but otherwise it seems to work.

Thanks for opening the PR.

madsnedergaard:

PR #7531, opened 2024-12-09
fix(parsers): Update NI parser as old urls stopped working

annaib00:
## Issue
Old NI parser urls stopped working. In this update data is now taken from https://www.cndc.org.ni/#seccionMapaInteractivo 

Note, the fetch_price does also not work, so this also needs to be updated, I just need to figure out where to find the price information. Also, in CR_NI.yaml ENTE.fetch_exchange is changed to NI.fetch_exchange as ENTE gives an SSL certificate error. 

### Double check


config/zones/NI.yaml
```diff
@@ -306,6 +306,7 @@ fallbackZoneMixes:
         wind: 0.13018731995607496
 parsers:
   price: NI.fetch_price
+  exchange: NI.fetch_exchange
```
annaib00, line 309: A bit unsure if this is necessary?

FelixDQ (approved):

VIKTORVAV99:
@annaib00 I think there is something wrong with the CI cache that is blocking merges right now. I'm looking into it.

PR #7539, opened 2024-12-11
feat(parsers): Update NI price parse

annaib00:
## Issue
NI price parser not working due to old link no longer working. Updated with new link and refetch frequency as this url only provides data for the current hour. 

### Double check

### DIFF
--- a/parsers/NI.py
+++ b/parsers/NI.py
@@ -1,10 +1,9 @@
-from datetime import datetime, time, timedelta
+from datetime import datetime, timedelta
 from logging import Logger, getLogger
 from typing import Any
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

parsers/NI.py
```diff
@@ -58,6 +58,7 @@ def extract_text(full_text: str, start_text: str, end_text: str | None = None):
         return full_text[start:end]
 
 
+@config.refetch_frequency(timedelta(days=1))
```
annaib00, line 61: Added a frequency of 1 day as the link only provide data for the current day

PR #7541, opened 2024-12-11
Exchange Data Disclaimer

cadeban:
## Description
Adding the 72 hour resolution makes the exchange data delay in the US more apparent and obvious (see below). This PR addresses that by adding a disclaimer if:
- The exchange data includes the latest 48 hours
- The exchange data is missing a significant amount of data points (threshold is set as >=20%, but is adjustable)

[Designs](https://www.figma.com/design/mfYDsUd0hRmhSapNgT9DvJ/Increase-the-share-ability-of-the-app?node-id=4402-93381&node-type=canvas&t=NgDPMiWPhfxPg9ni-0)

### Preview
Screenshot 2024-12-11 at 20 54 00

### Double check


cadeban:
@Alportan This screenshot shows CAISO at a 24h resolution and one or two exchanges are being reported, so the scale hides the fact that this is delayed. I don't necessarily think it requires a change to the logic, but I wanted to flag it in case you had concerns:
Screenshot 2024-12-11 at 20 55 37

cadeban:
Additionally, maybe we should consider running the copy by Jesper, since we're talking about how to address the real-time-ness in the app

Alportan:
> @Alportan This screenshot shows CAISO at a 24h resolution and one or two exchanges are being reported, so the scale hides the fact that this is delayed. I don't necessarily think it requires a change to the logic, but I wanted to flag it in case you had concerns: Screenshot 2024-12-11 at 20 55 37

Couldn't this be the issue we observed that the reported exchanges for the last 48h, specifically in CAISO, are way below the average? 

If you look at the magnitude, it usually goes up to 6-7GW, while the last 48h goes to only couple of hundreds of MW.
image

cadeban:
>Couldn't this be the issue we observed that the reported exchanges for the last 48h, specifically in CAISO, are way below the average? 
>
>If you look at the magnitude, it usually goes up to 6-7GW, while the last 48h goes to only couple of hundreds of MW.
> image

I think it's likely, but could you link me to the day that the screenshot is from?

cadeban:
Update: Jesper said this is fine for now, so this PR is ready for review

web/src/features/charts/MissingExchangeData.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  if (isLoading || isError || !data) {
+    return null;
+  }
+
+  const { chartData } = data;
+
+  const date = new Date();
+  date.setDate(date.getDate() - 1);
+  date.setHours(0, 0, 0, 0);
```
VIKTORVAV99, line 34: Ah JS you keep messing with me...

Why can I set seconds in the setHours function? You make no sense!

web/src/features/charts/MissingExchangeData.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const isWithinPrevious48Hours = date < new Date(chartData.at(-1)!.datetime);
+  const isMissingSignificantExchangeData = getIsMissingSignificantExchangeData(chartData);
+
+  if (isWithinPrevious48Hours && isMissingSignificantExchangeData) {
+    return (
+      <p className="prose my-1 rounded bg-gray-200 p-2 text-xs leading-snug dark:bg-gray-800 dark:text-white dark:prose-a:text-white">
+        {t('country-history.exchange-delay')}
+      </p>
+    );
+  }
```
VIKTORVAV99, line 45: It should probably return null here if this is false.

Undefined might work but I think the convention is that components should always return values (never undefined), but feel free to double check as im not 100% sure.

VIKTORVAV99 (commented):
Copy looks good as well but why are we not simply saying up to 48 hours?

tonypls:
Something about the padding doesn't quite look right. Maybe it's the blank space, if there's not a nice way to make it look good the current copy, maybe we could be slightly more verbose to fill out the empty space?

Something like: "Exchange data may be delayed by 24-48 hours" and make the background the same width as the chart content?

tonypls:
This is what I was thinking:
image

web/src/features/charts/MissingExchangeData.tsx
cadeban: 🤣 😭

cadeban:
Fair call! Will make the UI & copy updates

web/src/features/charts/MissingExchangeData.tsx
VIKTORVAV99: This is so dumb!
```
parseInt(0.5); // -> 0
parseInt(0.05); // -> 0
parseInt(0.005); // -> 0
parseInt(0.0005); // -> 0
parseInt(0.00005); // -> 0
parseInt(0.000005); // -> 0
parseInt(0.0000005); // -> 5
```

web/src/features/charts/MissingExchangeData.tsx
```diff
@@ -0,0 +1,46 @@
+import { useTranslation } from 'react-i18next';
+
+import { useNetExchangeChartData } from './hooks/useNetExchangeChartData';
+import { AreaGraphElement } from './types';
+
+const SIGNIFICANT_THRESHOLD = 0.2;
+
+const getTotal = (object: object) => Object.keys(object).length;
+const getNulls = (object: object) =>
+  Object.values(object).filter((v) => v === null).length;
```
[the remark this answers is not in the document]
cadeban: Hm, so my understanding is that null is the value that represents that we haven't received the data yet, unlike NaN and Infinity. I think we would probably want the application to error on Nan and Infinity, since that's probably an indication that we have a bug. What do you think?

tonypls (approved):
LGTM! Will merge to include in the next mobile release :shipit:

PR #7545, opened 2024-12-12
chore(web): tidy up timeAverage mapping and rename to timeRange

tonypls:
## Issue
We are mixing temporal granularities and time ranges. 

We currently have multiple variables for the same concept; timeAverages, urlTimeAverages and timeAggregates.

To deal with the urlTimeAverage being different from the timeAverage we use a mapping between the two (this was a temporary solution to get the new routing out quickly!).

## Description
This PR consolidates all time averages into one concept for a TimeRange of 24h, 72h, 30d, 12mo or all. This also renames the variable. 

This simplifies logic across the app, improves consistency by using enums with best practices. We still use the "hourly" naming on the backend so the mapping from timeRange to timeAverage happens at the API level, this would be an easy migration from this point. 

Also this refactor helped find a bug where the breakdown chart header wasn't correct for the 72h view, this is the second bug since I introduced this complexity 🤦 

I hope this reduces some tech debt and makes things easier to understand, let me know what you think!

This is quite a few lines to review! Let's use AI to help


web/src/features/charts/bar-breakdown/BarElectricityBreakdownChart.stories.ts
```diff
@@ -17,6 +17,7 @@ type Story = StoryObj<typeof BarElectricityBreakdownChart>;
 
 const data: ZoneDetails = {
   hasData: true,
+  //TODO Remove from backend, most likely unused
```
VIKTORVAV99, line 20: 👍🏻

web/src/features/charts/elements/AreaGraph.tsx
```diff
@@ -106,7 +106,7 @@ interface AreagraphProps {
   isDisabled?: boolean;
   height: string;
   datetimes: Date[];
-  selectedTimeAggregate: TimeAverages; // TODO: Graph does not need to know about this
+  selectedTimeRange: TimeRange; // TODO: Graph does not need to know about this
```
VIKTORVAV99, line 109: Is it not for the axis stuff?

web/src/features/charts/hooks/useOriginChartData.ts
```diff
@@ -180,14 +180,14 @@ interface ValuesInfo {
...
-  timeAggregate: string
+  timeRange: string
 ): ValuesInfo {
   const maxTotalValue = d3Max(historyData, (d: ZoneDetail) =>
     displayByEmissions
       ? getTotalEmissionsAvailable(d, true)
       : getTotalElectricityAvailable(d, true)
   );
-  const isHourly = timeAggregate === TimeAverages.HOURLY;
+  const isHourly = timeRange === TimeRange.H24;
```
VIKTORVAV99, line 190:
```suggestion
  // TODO: Make this use the atom as a single source of truth.
  const isHourly = timeRange === TimeRange.H24;
```

How does this work with the 72 hour view? That is also hourly?

web/src/features/charts/tooltips/PriceChartTooltip.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { timeAverageAtom } from 'utils/state/atoms';
+import { timeRangeAtom } from 'utils/state/atoms';
 
 import { convertPrice } from '../bar-breakdown/utils';
 import { InnerAreaGraphTooltipProps } from '../types';
 import AreaGraphToolTipHeader from './AreaGraphTooltipHeader';
 
 export default function PriceChartTooltip({ zoneDetail }: InnerAreaGraphTooltipProps) {
-  const [timeAverage] = useAtom(timeAverageAtom);
+  const [timeRange] = useAtom(timeRangeAtom);
```
VIKTORVAV99, line 11: I didn't know there was so many of these useAtom calls that should be useAtomValue instead...

I'll create a follow up PR for that.

VIKTORVAV99 (approved):
That was a lot of rows!

But as far as I can tell everything looks good!

web/src/features/charts/elements/AreaGraph.tsx
tonypls: Hmm, maybe not used once but used now?

VIKTORVAV99: Don't know. Let's leave it for now.

web/src/features/charts/hooks/useOriginChartData.ts
tonypls: Moving this up to a hook!

VIKTORVAV99 (approved):
Still looks good!

cadeban:
Woo thanks for fixing this!!! 🙌

PR #7546, opened 2024-12-12
Navigate back/forward through time by 24 hours in 72 hour view

cadeban:
## Description
Received feedback that it might be useful to navigate back/forward through time by 24 hours in 72 hour view.

Related BE PR: https://github.com/electricitymaps/electricitymaps/pull/6162

### Double check


cadeban:
c.c. @Alportan

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -23,7 +23,7 @@ import {
   timeAverageAtom,
 } from 'utils/state/atoms';
 
-const TIME_OFFSETS: Partial<Record<TimeAverages, number>> = {
+const WINDOW_OFFSETS: Partial<Record<TimeAverages, number>> = {
   [TimeAverages.HOURLY]: 24,
   [TimeAverages.HOURLY_72]: 72,
 };
```
VIKTORVAV99, line 29: Is this still needed?

VIKTORVAV99 (commented):
I don't mind the change but I think we need to update the cache function in the backend first or this will be awfully slow.

tonypls:
> I don't mind the change but I think we need to update the cache function in the backend first or this will be awfully slow.

I can do this tomorrow if you like, tag me on a linear ticket if required!

web/src/features/time/HistoricalTimeHeader.tsx
cadeban: We're still using it to calculate if the earliest datetime in a given historical window is within our historical limit

web/src/features/time/HistoricalTimeHeader.tsx
```diff
@@ -35,7 +35,12 @@ const clamp = (date: number, offset: number) => {
     return '';
   }
 
-  return new Date(newDate).toISOString();
+  // Offset to 3 hours ago (taking into consideration that we're over the end datetime so actually -2).This mimics the logic on the backend in get_latest_complete_flowtraced_dt
+  const d = new Date(newDate);
+  d.setUTCHours(new Date().getUTCHours() - 2);
+  d.setUTCMinutes(0, 0, 0);
+
+  return d.toISOString();
```
VIKTORVAV99, line 43: Do we have to hardcode this?

It will make it wrong in the future if we reduce the time delay on the app backend which is a goal.

cadeban: Of course - do you mind linking me to what you're referring to?

VIKTORVAV99: Just the static -2 hour offset really. If we manage to reduce the app backend delay to say 1 hour this will be "wrong" again.

So if possible it would be best to try and determine this from the state files first and last values instead.

If that's too complicated and takes too much time then maybe this is fine for now though.

tonypls: User's may visit from specific datetimes so we don't know the latest flowtraced from the backend. It should effect to many users and the outcome would be them being jumped to realtime ~3 hours which the data will still be visible for in the charts.

Hopefully when we do real-time in the future we can revisit this code and remove the delay!

tonypls:
@cadeban I hope you don't mind but in the spirit of getting this available for a mobile build I've refactored it further to have the same logic across 24 and 72h. My [grug brain](https://grugbrain.dev/) couldn't handle empty states and hooks for navigation so I attempted to make it more grug friendly.

web/src/features/time/HistoricalTimeHeader.tsx
cadeban: That's a fair point re: users visiting specific datetimes. I got tunnel-visioned trying to ensure that we're actually hitting the cache when users visit, since we have an hour discrepancy between the frontend and the cache

VIKTORVAV99 (approved):
It works as I expect.

Let's forget about specific implementation details for now so we can do a mobile release before Christmas and then revise and refractor later if needed.

web/src/features/time/HistoricalTimeHeader.tsx
VIKTORVAV99: We could send the latest flow traced data through the meta endpoint if it's needed later.

PR #7547, opened 2024-12-12
chore(config): update capacity in RO.yaml

mzaharie:
## Description

Add new link for capacities in Romania

### Double check

### DIFF
--- a/config/zones/RO.yaml
+++ b/config/zones/RO.yaml
@@ -1,4 +1,4 @@
-_comment: Biomass includes 19.227 MW of biogas and 6.03 MW of waste incineration
+_comment: Biomass includes 22.46 MW of biogas and 6.03 MW of waste incineration
 bounding_box:
   - - 20.2201924985
     - 43.6884447292
@@ -9,6 +9,9 @@ capacity:
     - datetime: '2023-01-01'
       source: entsoe.eu
       value: 118.0
+    - datetime: '2024-12-12'
+      source: anre.ro
+      value: 134.76
   coal:
     - datetime: '2023-01-01'
       source: entsoe.eu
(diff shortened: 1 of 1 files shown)


config/zones/RO.yaml
```diff
@@ -9,45 +9,70 @@ capacity:
     - datetime: '2023-01-01'
       source: entsoe.eu
       value: 118.0
+    - datetime: '2024-12-12'
```
VIKTORVAV99, line 12: Is this the date the data was published or updated?

We would like the use the date it was updated to prevent any regressions if it's updated in the future.

config/zones/RO.yaml
```diff
@@ -401,6 +426,8 @@ sources:
     link: https://docs.google.com/spreadsheets/d/1w5DJ7sPen6axIHU8TCVcuzNCjlct4I6JAbhUlw-ZXu8/edit?usp=sharing
   IPCC 2014:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
+  ANRE:
+    link: https://anre.ro/puteri-instalate/
```
VIKTORVAV99, line 430: This is not for capacities but rather carbon intensities.

Please remove it from here.

config/zones/RO.yaml
```diff
@@ -9,45 +9,70 @@ capacity:
     - datetime: '2023-01-01'
       source: entsoe.eu
       value: 118.0
+    - datetime: '2024-12-12'
+      source: ANRE
```
VIKTORVAV99, line 13: Could you turn these into the domain that was used?

VIKTORVAV99 (changes requested):
Thanks for creating this PR!

However there are some changes that has to be made to ensure it works with our existing systems.

config/zones/RO.yaml
mzaharie: The web page states these are the numbers today -- https://anre.ro/puteri-instalate/

PaulRoms:

PR #7548, opened 2024-12-13
feat(web): vertical line to show midnight

tonypls:
## Issue
Unable to see day widths

## Description
This PR updates the charts to display a dotted line on midnight, removes an extra tick on the TimeAxis for charts and aligns the ticks with the bars.


web/src/features/charts/OriginChart.tsx
```diff
@@ -24,7 +24,7 @@ interface OriginChartProps {
   timeRange: TimeRange;
 }
 
-// TODO(cady): fix types to use ElectricityModeType
+// TODO: fix types to use ElectricityModeType
```
cadeban, line 27: 😅

web/src/features/time/TimeAxis.tsx
```diff
@@ -41,13 +47,34 @@ const renderTick = (
...
+const isMidnight = (date: Date, timezone?: string) => {
+  if (!timezone) {
+    return date.getHours() === 0 && date.getMinutes() === 0;
+  }
+
+  return (
+    new Date(date.toLocaleString('en-US', { timeZone: timezone })).getHours() === 0 &&
+    new Date(date.toLocaleString('en-US', { timeZone: timezone })).getMinutes() === 0
+  );
+};
```
VIKTORVAV99, line 76: Question, why are these hardcoded to `en-US`?

cadeban (approved):
:shipit:

web/src/features/time/TimeAxis.tsx
tonypls: Good catch! I'll update it!

tonypls:
> These is a slight alignment issue here we should look at but it's not related to the changes in here. But rather an issue with the axis.
> 
> English looks fine with AM/PM: image
> 
> But not 24 hour clocks: image

If I can fix this in 10 minutes I'll do it! Otherwise I believe this is already an issue and may be suited to another PR 🙏

web/src/features/charts/OriginChart.tsx
cadeban: I'm just embarrassed that it slipped by me and made it into the codebase 🤦

VIKTORVAV99: Don't be, all my TODOs go into the codebase...

Hopefully I'll fix them all at some point 😅

PR #7551, opened 2024-12-13
fix(app): Prevent iframing

jbdietrich:
## Description

Disallows iframing the entire app.

### Double check


VIKTORVAV99 (approved):
Nice!

Do we want to declare all content security policies as real headers in this file?

We currently have some in the index.html file. I think it makes sense to have these in a unified place (and that the headers file is the way to go).

jbdietrich:
> We currently have some in the index.html file. I think it makes sense to have these in a unified place (and that the headers file is the way to go).

Ah! I didn't know about these. Yeah, I'd argue they should all be in the same place. Happy to make a follow-on PR next week to consolidate them.

PR #7552, opened 2024-12-13
chore(SEO): Use useMatches to get the correct pathname with urlTimeRange

VIKTORVAV99:
## Issue

We had a todo we could fix + SEO didn't pick up the `urlTimeRange`.

## Description

Uses the useMatches hook to get the pathname.

### Double check


tonypls (approved):
LGTM!

PR #7558, opened 2024-12-16
feat(web): pop-over for 72h view

tonypls:
## Issue
New feature! 

## Description
This PR removes the historical navigation pop-over and adds a new popover for the 72h view.


VIKTORVAV99 (approved):
LGTM! 🎉

I still think we should move away from the onboarding screen towards multiple popovers (showing one or 2 at the time) in the future so it would be nice if we could plan out some state management for that at some point.

VIKTORVAV99:
> > LGTM! 🎉
> 
> > 
> 
> > I still think we should move away from the onboarding screen towards multiple popovers (showing one or 2 at the time) in the future so it would be nice if we could plan out some state management for that at some point.
> 
> 
> 
> Live 247 CO₂ emissions of electricity consumption  App  Electricity Maps
> 
> 🙈 

Exactly! 🙈

PR #7559, opened 2024-12-16
fix(parsers): CAISO url fix

annaib00:
Small fix to CAISO parser. They seem to have changed their urls a bit causing no data. 

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7562, opened 2024-12-16
feat(zones):Update capacity value for CH and add unknown capacity value

annaib00:
## Issue
No unknown capacity value for CH. The unknown production value is calculated as the difference in total_production and the sum of consumption, exchange and storage, so we do not have a capacity value for this production mode. Unknown capacity production value is here added as the total production capacity. 

Other values are updated: 
https://www.uvek-gis.admin.ch/BFE/storymaps/EE_Elektrizitaetsproduktionsanlagen/?lang=en 
ENTSOE

### Double check

### DIFF
--- a/config/zones/CH.yaml
+++ b/config/zones/CH.yaml
@@ -5,18 +5,69 @@ bounding_box:
     - 47.808380127
 
 capacity:
-  biomass: 648.4
-  coal: 0
-  gas: 304.7
-  geothermal: 0
-  hydro: 10438.1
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
While the changes look good should we change this to the new capacity format that is versioned by date?

FelixDQ (approved):

PR #7567, opened 2024-12-16
bug(web): midnight line not visible on mobile devices

tonypls:
## Issue
Midnight line not showing on mobile devices

## Description
This PR updates the logic for finding midnight to use the JavaScript Intl api

### DIFF
--- a/web/src/features/time/TimeAxis.tsx
+++ b/web/src/features/time/TimeAxis.tsx
@@ -46,16 +46,18 @@ const renderTick = (
       opacity={1}
       transform={`translate(${scale(value)},0)`}
     >
-      {isMidnightTime && !isTimeController && (
-        <line
-          stroke="currentColor"
-          strokeDasharray="2,2"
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
Think you can throw in a fix for this in the same PR?
image

I think you just need to check that isHourly is true as well.

tonypls:
> Think you can throw in a fix for this in the same PR? image
> 
> I think you just need to check that isHourly is true as well.

How do I replicate this? Doesn't show on 30d for me

VIKTORVAV99:
> > Think you can throw in a fix for this in the same PR? image
> 
> > 
> 
> > I think you just need to check that isHourly is true as well.
> 
> 
> 
> How do I replicate this? Doesn't show on 30d for me

I just set it to English and it appeared, but it might be specific to US English.

I can try a few steps tomorrow if this don't work.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7571, opened 2024-12-17
Add exports+storage to origin charts

corradio:
## Issue

This PR adds storage and exports to the origin chart by showing them as negative.

## Description

@lav-julien and myself discussed the following graph and were unsure if the wind drop was curtailment.

image

We wanted to check if the load stayed constant (indicating no missing data), and realised the exports were not visible here.
It reminded me of [this debate](https://github.com/electricitymaps/electricitymaps-contrib/issues/1030) from 2018 where a similar issue was discussed.

Important point: by adding exports and storage to the origin graph, we actually bring back more consistency with the top bar chart (which is supposed to represent a slice of the origin graph at a particular point) by visualising the bars that are negative.

### Preview

image
image

Tooltips also work out of the box as they use the same system as the bar chart at the top:

image

### Double check

### DIFF
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -53,7 +53,7 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => (shouldHideEmptyData ? d[1] > 0 : Number.isFinite(d[1])));
+    .defined((d) => Number.isFinite(d[1]));
 
   // Mouse hover events
   let mouseOutTimeout: string | number | NodeJS.Timeout | undefined;
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -53,7 +53,16 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => (shouldHideEmptyData ? d[1] != 0 : Number.isFinite(d[1])));
+    .defined((d) => {
+      if (!Number.isFinite(d[0]) || !Number.isFinite(d[1])) {
+        return false;
+      }
+      if (shouldHideEmptyData && d[0] == d[1]) {
+        // zero area height
+        return false;
+      }
+      return true;
+    });
 
   // Mouse hover events
   let mouseOutTimeout: string | number | NodeJS.Timeout | undefined;
--- a/web/src/features/charts/elements/AreaGraphLayers.tsx
+++ b/web/src/features/charts/elements/AreaGraphLayers.tsx
@@ -53,7 +53,7 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => Number.isFinite(d[1]));
+    .defined((d) => (shouldHideEmptyData ? d[1] != 0 : Number.isFinite(d[1])));
 
   // Mouse hover events
   let mouseOutTimeout: string | number | NodeJS.Timeout | undefined;
--- a/web/src/features/charts/hooks/useOriginChartData.ts
+++ b/web/src/features/charts/hooks/useOriginChartData.ts
@@ -93,7 +93,7 @@ export default function useOriginChartData() {
       // Add exchanges
       for (const [key, exchangeValue] of Object.entries(value.exchange)) {
         // in GW or MW
-        entry.layerData[key] = Math.max(0, exchangeValue / valueFactor);
+        entry.layerData[key] = exchangeValue / valueFactor;
         if (displayByEmissions) {
           // in gCO₂eq/hour
           entry.layerData[key] =
@@ -143,7 +143,7 @@ function getStorageValue(
     return Number.NaN;
   }
 
-  const invertedValue = -1 * Math.min(0, storageValue);
+  const invertedValue = -1 * storageValue;
 
   return displayByEmissions
     ? invertedValue * value.dischargeCo2Intensities[storageKey] * valueFactor


web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -53,7 +53,7 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => (shouldHideEmptyData ? d[1] > 0 : Number.isFinite(d[1])));
+    .defined((d) => Number.isFinite(d[1]));
 
   // Mouse hover events
   let mouseOutTimeout: string | number | NodeJS.Timeout | undefined;
```
VIKTORVAV99, line 59: Unrelated but why does this have a NodeJS.Timeout type when it's only running in the browser?

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -53,7 +53,7 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => (shouldHideEmptyData ? d[1] > 0 : Number.isFinite(d[1])));
```
VIKTORVAV99, line 56: Mind deleting this part a little further up as it's no longer used?
```ts
  const shouldHideEmptyData = isDataInteractive && layers.length > 1;
```

VIKTORVAV99: Oh and the changes here causes these empty lines to re-appear:
image

I think we should just check that it's not 0 instead of removing this. If that makes sense.

The production app looks like this:
image

corradio: Excellent catch @VIKTORVAV99 !! Let me change this now

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -53,7 +53,7 @@ function AreaGraphLayers({
...
+    .defined((d) => Number.isFinite(d[1]));
```
corradio, line 56:
```suggestion
    .defined((d) => (shouldHideEmptyData ? d[1] != 0 : Number.isFinite(d[1])));
```

web/src/features/charts/elements/AreaGraphLayers.tsx
corradio: No idea :(

web/src/features/charts/elements/AreaGraphLayers.tsx
corradio: Actually, when I change it to `!= 0`, then the negative storage graph is gone:

image


If the condition is removed (i.e. only non-finite points are removed), then the (negative) storage graph is back:

image

Do you see this as well?

VIKTORVAV99: Yeah I noticed the same. I didn't expect it to remove the whole thing.

I can test a few things locally as well.

VIKTORVAV99: Does this work for you? It seems to work for the zones I have tested. 
```ts
    .defined((d) =>
      shouldHideEmptyData ? (d[1] > 0 ? d[1] != 0 : d[0] != 0) : Number.isFinite(d[1])
    );
```
Basically is just checks which end value it should use.

We can probably clean it up a bit though.

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -53,7 +53,7 @@ function AreaGraphLayers({
     .x((d: any) => timeScale(d.data.datetime))
     .y0((d) => valueScale(d[0]))
     .y1((d) => valueScale(d[1]))
-    .defined((d) => (shouldHideEmptyData ? d[1] > 0 : Number.isFinite(d[1])));
+    .defined((d) => (shouldHideEmptyData ? d[1] != 0 : Number.isFinite(d[1])));
```
VIKTORVAV99, line 56:
```suggestion
    .defined((d) =>
      shouldHideEmptyData ? (d[1] > 0 ? d[1] != 0 : d[0] != 0) : Number.isFinite(d[1])
    );
```

web/src/features/charts/elements/AreaGraphLayers.tsx
corradio: yes that works. I rewrote it like this to:
1. check NaNs
2. remove points that start and end at the same place (y-axis)

```js
    .defined((d) => {
      if (!Number.isFinite(d[0]) || !Number.isFinite(d[1])) {
        return false;
      }
      if (shouldHideEmptyData && d[0] == d[1]) {
        // zero area height
        return false;
      }
      return true;
    });
```

My tests show that the above both hides solar values that equal zero, and shows (negative) storage values. What do you think?

VIKTORVAV99: It looks like it works just as well and is a bit easier to understand, especially the equal part!

If we want to optimize it though we can do this though:
```ts
.defined((d) => 
  Number.isFinite(d[0]) && 
  Number.isFinite(d[1]) && 
  (!shouldHideEmptyData || d[0] !== d[1])
);
```

corradio: I'll let you decide which one you prefer!

VIKTORVAV99: Hehe maybe don't as I'll always opt for the most optimized version, it's a curse...

So maybe @tonypls can decide by the time we wake up tomorrow or we'll just merge one of them 😅

web/src/features/charts/elements/AreaGraphLayers.tsx
corradio:
```suggestion
    .defined((d) => {
      if (!Number.isFinite(d[0]) || !Number.isFinite(d[1])) {
        return false;
      }
      if (shouldHideEmptyData && d[0] == d[1]) {
        // zero area height
        return false;
      }
      return true;
    });
```

web/src/features/charts/elements/AreaGraphLayers.tsx
tonypls: As the charts are already quite complex, I'd lean towards the less optimized and easy to ready if statement implementation that mere mortals can read :)

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7572, opened 2024-12-17
feat(config): add bboxes for upcomming zones

FelixDQ:
## Issue

I added these to our internal weather system, so I thought that it would be good to keep it consistent with this config as well


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7574, opened 2024-12-17
Moves parser files to a separate Earthfile target

madsnedergaard:
## Description

Separates parsers into it's own Earthly target to simplify dependency management in our internal build system.


FelixDQ (commented):
If i'm understanding this correctly, this also removes `parsers` from `+prepare`, `+build` and `+test`. Is that a problem? Are the tests still being run?

madsnedergaard:
> If i'm understanding this correctly, this also removes `parsers` from `+prepare`, `+build` and `+test`. Is that a problem? Are the tests still being run?

Whoops good point, fixed now 😅

FelixDQ (approved):

VIKTORVAV99:
@madsnedergaard mind merging this into master after the next release has been made?

I think you will maybe have to do some changes in the monorepo as well so would be nice if it was stand alone.

PR #7575, opened 2024-12-18
feat(web): Default to 72h view

tonypls:
## Issue
24h is less insightful than 72h

## Description
This PR updates the app to default to the 72h view


VIKTORVAV99:
Should we remove the 72h feature flag logic as well if we are making it the default?

tonypls:
> Should we remove the 72h feature flag logic as well if we are making it the default?

I still wanted to have the ability to revert now you say it, we're too far down the 72h route to turn it off.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7576, opened 2024-12-18
feat(web): add estimation text for general purpose zone model

silkeholmebonnen:
## Issue

[GMM-203](https://linear.app/electricitymaps/issue/GMM-203/add-new-estimation-label-for-general-purpose-zone-model)

## Description

This PR adds a new description for general purpose zone model estimations. 

### Preview

image

### Double check


silkeholmebonnen:
You are very welcome to come with suggestions for what the description should be!

FelixDQ:
Amazing, thank you @silkeholmebonnen! I think that there should be a hyphen in "real-time", but otherwise LGTM!

There might be a future question of how the estimation labels are perceived. When reading the labels "imprecise", "not real-time", or "incomplete", it sounds like the showcased data has these problems, but the labels are actually an indication of the input data. Our estimation models do make the data "precise", "real-time", and "complete".

Anyway, I think that is a problem to be tackled later, but I'm curious to hear your thoughts @tonypls @Alportan

tonypls:
This is great! 

I think the pill could say estimated instead of real-time. 

Here's a potential description too: "Real-time source data for this zone is not published. Current values are modelled using a combination of live weather data and historical production data from each source."

VIKTORVAV99:
Do we want to mention that they are using monthly and yearly sources?

That's the main reason they are being estimated even for the historical date times.

jbdietrich:
Just FYI: I'd like us to simplify our taxonomy of estimations to classify all zones as using [one of three types](https://www.notion.so/electricitymaps/Zone-tiers-1582abac42b080d782add8b769e44b92?pvs=4).

I think for now what you all have suggested is a good solution, I just want to flag that I see this changing in the near future. I'd suggest taking @tonypls's wording and going for it.

madsnedergaard:
Let's get this one merged and out today! If @silkeholmebonnen isn't working then I can add the suggestions from @tonypls and merge it later today

madsnedergaard:
Pushed the changes suggested by @tonypls with some minor tweaks. It could definitely be better(!), but this is at least a starting point :)

madsnedergaard (approved):

madsnedergaard:
The "estimated" pill next to "estimated" in title is a bit weird imo, but we can adjust later if anyone has a better suggestion :)

Screenshot 2024-12-19 at 14 24 15

Alportan:
> The "estimated" pill next to "estimated" in title is a bit weird imo, but we can adjust later if anyone has a better suggestion :)
> 
> Screenshot 2024-12-19 at 14 24 15

I agree! Following James' comment, I believe we should revise and simpligy the terminology of all these labels and headlines. I'll keep it on the top of my head and maybe sync with James to understand the general model better 🤓

PR #7578, opened 2024-12-18
Fix formatting of unit for 72hrs

corradio:
## Issue

Raised on LinkedIn and by @VIKTORVAV99 

### Preview

image

### Double check


VIKTORVAV99 (approved):
I was refracting this so we can use the atom but let's do this real quick. 😅


LGTM! 🎉

PR #7580, opened 2024-12-18
Fix parsers Earthly target

madsnedergaard:
## Issue

Follow-up to #7574

## Description

Turns out it didn't quite work as intended, but this should work now (I tested locally).


madsnedergaard:
Got something working now!
Here's a build using this branch with everything succeeding: https://drone.electricitymaps.com/electricitymaps/electricitymaps/20052

tonypls (approved):
If it works, it works!

PR #3092, opened 2021-04-14
Split Newfoundland and Labrador into 2 zones

FelixDQ:
Prerequisite for https://github.com/tmrowco/electricitymap-contrib/pull/3077

PR #7582, opened 2024-12-19
chore(zones): Merge CA-NL-LB and CA-NL-NF into CA-NL

VIKTORVAV99:
## Issue
The data source have these merged so we need to do so as well.
Part of: GMM-197

## Description

Merges the zones so we can add a General Purpose Zone Estimation Model for it.

### Double check


jarek:
Out of curiosity which data source do you have for it merged?

for reference, the split was done in #3092 but then I think the parser was never actually written, I guess because the data source doesn't give a lot of data

VIKTORVAV99:
> Out of curiosity which data source do you have for it merged?
> 
> 
> 
> for reference, the split was done in #3092 but then I think the parser was never actually written, I guess because the data source doesn't give a lot of data

We will be using monthly data from https://www.cer-rec.gc.ca that will be processed into hourly values based on local weather patterns and sunset/sunrise. While not as accurate on a hourly level we've decided to do this for all the missing Canadian zones where we are missing data sources. If we can find better data sources for them in the future we will happily use those.

FelixDQ (approved):

PR #7584, opened 2024-12-19
chore(date formatting): clean up date formatting in the app

VIKTORVAV99:
## Issue

It's not clear if the data is for a single point in time or for a range of time.
## Description

Switches to use a date range for the date formatting and uses date fns functions instead of get/set UTC date all the time.

### Preview

image

image

### Double check

### DIFF
--- a/web/src/components/Time.tsx
+++ b/web/src/components/Time.tsx
@@ -1,37 +1,32 @@
+import { useAtomValue } from 'jotai';
 import { useParams } from 'react-router-dom';
 import { RouteParameters } from 'types';
-import { TimeRange } from 'utils/constants';
-import { formatDate, formatDateRange } from 'utils/formatting';
+import { formatDate } from 'utils/formatting';
 import { getZoneTimezone } from 'utils/helpers';
+import { timeRangeAtom } from 'utils/state/atoms';
 
 export function FormattedTime({
   datetime,
   language,
-  timeRange,
   className,
   zoneId,
+  isTimeHeader = false,
   endDatetime,
 }: {
   datetime: Date;
   language: string;
-  timeRange?: TimeRange;
   className?: string;
   zoneId?: string;
+  isTimeHeader?: boolean;
   endDatetime?: Date;
 }) {
+  const timeRange = useAtomValue(timeRangeAtom);
   const { zoneId: pathZoneId } = useParams<RouteParameters>();
   const timeZoneZoneId = zoneId || pathZoneId;
   const timezone = getZoneTimezone(timeZoneZoneId);
-  if (timeRange) {
-    return (
-      <time dateTime={datetime.toISOString()} className={className}>
-        {formatDate(datetime, language, timeRange, timezone)}
-      </time>
-    );
-  }
   return (
     <time dateTime={datetime.toISOString()} className={className}>
-      {endDatetime && formatDateRange(datetime, endDatetime, language, timezone)}
+      {formatDate(datetime, language, timeRange, timezone, isTimeHeader, endDatetime)}
     </time>
   );
 }
--- a/web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
+++ b/web/src/features/charts/tooltips/AreaGraphTooltipHeader.tsx
@@ -2,18 +2,15 @@ import { HorizontalDivider } from 'components/Divider';
 import EstimationBadge from 'components/EstimationBadge';
 import { FormattedTime } from 'components/Time';
 import { useGetEstimationTranslation } from 'hooks/getEstimationTranslation';
-import { useAtomValue } from 'jotai';
 import { CircleDashed, TrendingUpDown } from 'lucide-react';
 import { useTranslation } from 'react-i18next';
-import { EstimationMethods, TimeRange } from 'utils/constants';
-import { endDatetimeAtom } from 'utils/state/atoms';
+import { EstimationMethods } from 'utils/constants';
 
 import ProductionSourceIcon from '../ProductionsSourceIcons';
 
 interface AreaGraphToolTipHeaderProps {
   squareColor: string;
   datetime: Date;
-  timeRange: TimeRange;
   title: string;
   hasEstimationPill?: boolean;
   estimatedPercentage?: number;
@@ -24,7 +21,6 @@ interface AreaGraphToolTipHeaderProps {
 export default function AreaGraphToolTipHeader({
   squareColor,
   datetime,
-  timeRange,
   title,
   hasEstimationPill = false,
   estimatedPercentage,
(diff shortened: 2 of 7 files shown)


corradio:
While this is technically more correct, I wonder if this impacts the simplicity of the current user experience.
For example, when presenting Electricity Maps, I've enjoyed the simplicity of saying "at 1pm, the grid produced XYZ". Here we're presenting a slightly more complicated story (between 1pm and 2pm, the grid produced XYZ).
I'm just wondering if the need to be more correct is creating more information overload than necessary, and whether this is worth it.
For yearly values in particular: don't we want to refer to them as e.g. "2024" instead of "2024-2025"? There it may bear confusion: if both are present, which one do we mean? What we actually mean is "2024-01-01T00:00:00 -> 2024-12-31T23:59:59", and that may be more easily communicated by saying "2024"?

I personally need to think a bit more about it and probably show it around to a few people to get some feedback before I can give proper feedback.

PS: In case you're wondering if this is a departure from what we did with the x-ticks on the area charts: there I think this was necessary because for e.g. hourly values, it didn't make sense to have the midnight line start at the middle of the hour 00:00 (it should start at beginning of the value for the hour 00:00).

Will think about it more.

VIKTORVAV99:
> While this is technically more correct, I wonder if this impacts the simplicity of the current user experience.
> 
> For example, when presenting Electricity Maps, I've enjoyed the simplicity of saying "at 1pm, the grid produced XYZ". Here we're presenting a slightly more complicated story (between 1pm and 2pm, the grid produced XYZ).
> 
> I'm just wondering if the need to be more correct is creating more information overload than necessary, and whether this is worth it.
> 
> For yearly values in particular: don't we want to refer to them as e.g. "2024" instead of "2024-2025"? There it may bear confusion: if both are present, which one do we mean? What we actually mean is "2024-01-01T00:00:00 -> 2024-12-31T23:59:59", and that may be more easily communicated by saying "2024"?
> 
> 
> 
> I personally need to think a bit more about it and probably show it around to a few people to get some feedback before I can give proper feedback.
> 
> 
> 
> PS: In case you're wondering if this is a departure from what we did with the x-ticks on the area charts: there I think this was necessary because for e.g. hourly values, it didn't make sense to have the midnight line start at the middle of the hour 00:00 (it should start at beginning of the value for the hour 00:00).
> 
> 
> 
> Will think about it more. You just opened pandora's box! 😁 

Haha I'm fine with just opening the discussion and closing this PR if we decide against it.

But it was I did this after the feedback we did on LinkedIn in relation to not knowing if the data was for a single point in time or if it was for a period.

While it's implicit for us (and probably other people working with time series data often) I'm not sure everyone would find it clear.

But talk to a few people, see what they think and then we can take it from there!

VIKTORVAV99:
Fixed the type issues so you can do some side by side comparisons!

corradio:
I've done a bit more thinking here. From a user perspective, we're having two different situations:
1. we're showing an *instantaneous* value representing power (e.g. GW). We do that when we consider the time interval/frequency to be small enough that it can be considered instantaneous. Right now we only do it at hourly level.
2. we're showing a *volume* of electricity (e.g. GWh). We do that when we're considering interval/frequencies that are large enough that we have to talk about electricity used during an interval. We do that for days, months, years.

From a user perspective, I currently think an intuitive way to talk about 1. is to say "*At* XX:XX, X *GW* were produced using X".
For 2., I think an intuitive way is to say "*During* XX, X *GWh* were produced using X". So use "At" for an instant, and "During" for an interval. Watts for instants, Watt-hours for intervals. For us it would mean:
- At 05:00, X GW were produced using X
- During 15 Dec 2024, X GWh were produced using X
- During Nov 2024, X GWh were produced using X
- During 2024, X GWh were produced using X

In all cases, we never have to rely on showcasing two values (e.g. "01:00-02:00"), and I believe the language of "at" and "during" is both simple and precise.

An important note is that the formatting depends on the frequency of the data (hourly, daily, monthly..) and not the window selected (24h, 72h, ...).

An outstanding question is whether we deem hourly data to be an instant (category 1) or an interval (category 2). I am slightly tempted to categorise it into an interval (category 2), which would mean using "During 05:00", and the unit suffix "Wh". It would bring consistency to the whole app and simplify the code. On the flip side, it would make it harder for us when we add 5min or 15min data, as I assume these would classify as instants (category 1).

VIKTORVAV99:
> I've done a bit more thinking here. From a user perspective, we're having two different situations:
> 
> 1. we're showing an *instantaneous* value representing power (e.g. GW). We do that when we consider the time interval/frequency to be small enough that it can be considered instantaneous. Right now we only do it at hourly level.
> 
> 2. we're showing a *volume* of electricity (e.g. GWh). We do that when we're considering interval/frequencies that are large enough that we have to talk about electricity used during an interval. We do that for days, months, years.
> 
> 
> 
> From a user perspective, I currently think an intuitive way to talk about 1. is to say "*At* XX:XX, X *GW* were produced using X".
> 
> For 2., I think an intuitive way is to say "*During* XX, X *GWh* were produced using X". So use "At" for an instant, and "During" for an interval. Watts for instants, Watt-hours for intervals. For us it would mean:
> 
> - At 05:00, X GW were produced using X
> 
> - During 15 Dec 2024, X GWh were produced using X
> 
> - During Nov 2024, X GWh were produced using X
> 
> - During 2024, X GWh were produced using X
> 
> 
> 
> In all cases, we never have to rely on showcasing two values (e.g. "01:00-02:00"), and I believe the language of "at" and "during" is both simple and precise.
> 
> 
> 
> An important note is that the formatting depends on the frequency of the data (hourly, daily, monthly..) and not the window selected (24h, 72h, ...).
> 
> 
> 
> An outstanding question is whether we deem hourly data to be an instant (category 1) or an interval (category 2). I am slightly tempted to categorise it into an interval (category 2), which would mean using "During 05:00", and the unit suffix "Wh". It would bring consistency to the whole app and simplify the code. On the flip side, it would make it harder for us when we add 5min or 15min data, as I assume these would classify as instants (category 1).

I'll see what we can do about the formatting but we use the Intl standard and default JS options for this to ensure it's localized not only by language but by region as well.

And because dates are what they are it's vastly different in the US vs say Europe.

But I'll get back to you about what is possible!

tonypls:
I think the idea here is more correct but we should ask ourselves "are we spending our time solving a problem that brings significant value to our users?". I've already spent time digging into all the total energy change PR's so I can bring more informed input to this discussion and I'm sure you all have too. 

We would also like to simplify the "data overload" present in the tooltips and adding more characters to them should be considered as a trade-off here of simplicity vs correctness.

I think the "during" approach makes sense to but does bring some challenges if we move to 5 & 15 minute data. It could be best to park this discussion and tackle it in a holistic way (simple, consistent and correct) when we shift to finer granularity?

VIKTORVAV99:
It seems like the new ENTSO-E website is using ranges: 
image
(tooltip title to the far left in the middle)

madsnedergaard:
Moved to draft until we have made a decision on this :)

PR #7590, opened 2024-12-20
bug(web): try fix dynamically imported modules failing

tonypls:
## Issue
~0.1% of users experiencing crashes some of which are due to failed dynamic imports when rolling out updates

## Description
This PR checks for a vite preload error and reloads the window,  [docs here](https://vite.dev/guide/build#load-error-handling)

Also calls event.preventDefault() so the error isn't thrown.


VIKTORVAV99 (approved):
LGTM! 🎉

I wonder why it just doesn't do this on its own.

tonypls:
> LGTM! 🎉
> 
> I wonder why it just doesn't do this on its own.

I've updated it a bit further as I thought it might throw users into a loop without us knowing about it. Updated version clears the cache, which I think is the root of the problem (trying to import mapwrapper-xyz123.js when mapwrapper-abc123.js is loaded) and if this fails still sends the error to sentry

VIKTORVAV99:
> > LGTM! 🎉
> 
> > 
> 
> > I wonder why it just doesn't do this on its own.
> 
> 
> 
> I've updated it a bit further as I thought it might throw users into a loop without us knowing about it. Updated version clears the cache, which I think is the root of the problem (trying to import mapwrapper-xyz123.js when mapwrapper-abc123.js is loaded) and if this fails still sends the error to sentry

Taking a deeper look in a few min!

web/src/main.tsx
```diff
@@ -46,6 +43,41 @@ if (isProduction) {
...
+  try {
+    if ('serviceWorker' in navigator) {
+      const registrations = await navigator.serviceWorker.getRegistrations();
+      await Promise.all(registrations.map((r) => r.unregister()));
+    }
+
+    if ('caches' in window) {
+      const keys = await caches.keys();
+      await Promise.all(keys.map((key) => caches.delete(key)));
+    }
```
VIKTORVAV99, line 59: What is this supposed to do?

I looked at both staging and production and `caches.keys()` contains the following:
```
0 : "workbox-precache-v2-https://app.electricitymaps.com/"
1 : "images"
2 : "icons"
```
I don't think we need to delete the last two?

But I suppose it don't hurt too much either?

web/src/main.tsx
```diff
@@ -4,7 +4,7 @@ import 'react-spring-bottom-sheet/dist/style.css';
 import './index.css';
 
 import { Capacitor } from '@capacitor/core';
-import * as Sentry from '@sentry/react';
```
VIKTORVAV99, line 7: Are we sure there are no side effects to changing this?

We are no longer importing the whole of Sentry.

tonypls: Hmm, good point, I've included it again!

web/src/main.tsx
tonypls: I think it's safer to remove them all, I believe it should be a rare case.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7591, opened 2024-12-20
Update IM.yaml

CHirch:
Added capacity information Isle of Man and added interconnector FR - JE

## Description

Added capacity information from manxutilities homepage and Interconnector https://www.jec.co.uk/about-us/our-businesses/energy-division/

### Double check


config/zones/IM.yaml
```diff
@@ -10,5 +28,12 @@ fallbackZoneMixes:
         gas: 0.988
         hydro: 0.007
         oil: 0.005
+    - _source: https://www.manxutilities.im/media/3096/manx-utilities-annual-report-and-accounts-31-march-2024.pdf
+      datetime: '2024-12-20'
```
VIKTORVAV99, line 32: This datetime means it will only be used from this date and forward, I assume this is not what you want?

VIKTORVAV99 (commented):
Everything looks good but there is one tiny thing to look at!

Thanks for the PR!

config/zones/IM.yaml
CHirch: I was not sure about the fuction of the datetime attribute as I was not able to find it in the documentation. I tried to reverse engineer from other config files an thought it was the date of the last edit.

VIKTORVAV99: It's from the first date the data is valid. 

So if it is for example 2023-01-01 it will be used from 2023-01-01 and forward.

If there is one for 2023-01-01 and 2024-01-01 then the 2023-01-01 value will be used between 2023-01-01 and 2024-01-01 and then 2024-01-01 will be used for all values after 2024-01-01.

madsnedergaard (approved):
LGTM, sorry about the delay :)

PR #7592, opened 2024-12-20
fix(map): Fix Minnesota map

VIKTORVAV99:
## Issue
The map boundaries have become outdated. 

## Description

Updates the map.

### Double check


VIKTORVAV99:
@jbdietrich as we discussed earlier I think this solves the immediate issue but we will have to look into the core issue more in depth after the holidays.

jbdietrich (approved):

jbdietrich:
Let's try to get this out today!

VIKTORVAV99:
@jbdietrich and it's deployed!

PR #7595, opened 2024-12-20
fix(parsers): Add missing mapping for AM to the ENTSOE parser

VIKTORVAV99:
## Issue

The mapping was missing so the parser failed...

## Description

Pretty awkward but we where missing the mapping for the AM bidding zone causing the parser to error when trying to fetch the exchange between `AM` and `GE`.


VIKTORVAV99:
Self merging this one as it's a simple fix.

Issue #7598, opened 2024-12-21
Some french translations are not displayed


phiphou:
Some french translations are not displayed while there is some in translation files.

Some exemples: 

![Capture d'écran 2024-12-21 130033](https://github.com/user-attachments/assets/4ebfcf72-dd1c-4e6a-9404-83e966c8933d)

![Capture d'écran 2024-12-21 125930](https://github.com/user-attachments/assets/0bd65cbe-84ff-4dd2-adae-9ae0f0416916)

VIKTORVAV99:
Yeah it's a known issue, the current system looks at the resolution to get the title, so they exists for 24 hours but not 72 hours.

So we just need to copy them over basically (as we have some things planned 👀 for 24 hours), this is a bit unfortunate but just the reality of things.

I can copy them over (and for all the other languages) but we won't be making another release until after new years.

PR #7599, opened 2024-12-21
fix(translation): Add 72h translations

VIKTORVAV99:
## Issue

Fixes: #7598 

## Description

Copies over the 24h translations to 72h as well as they are the same at the moment.

### Double check


cadeban (approved):
lgtm

PR #7600, opened 2024-12-21
fix(translation): Remove hardcoded string in favor of translation strings

VIKTORVAV99:
## Issue

Related to: #7598

## Description

Removes the hardcoded translation string in favor of translation strings.

### Double check


cadeban (approved):
👍

PR #7601, opened 2024-12-21
perf: Optimize analyzeChartData

VIKTORVAV99:
## Description

While pretty optimized already when it comes to the logic we can optimize this a bit further when it comes to runtime performance by reducing property access as much as possible, this also have a small size improvement.

### Double check


cadeban (approved):
Easier to read too!

PR #7602, opened 2024-12-21
perf: Optimize callbacks in AreaGraph.tsx

VIKTORVAV99:
## Issue

These callbacks where not optimized and used `useMemo` instead of `useCallback`.

## Description

Switches to use `useCallback` and memoizes one more callback.

### Double check


silkeholmebonnen (approved):
I read a bit about it and understand that it is basically doing the same, it just looks nicer. LGTM

VIKTORVAV99:
> Why do we want to use useCallback instead of useMemo?
> 
> 

Basically useMemo is for objects and useCallback is for functions.

So instead of having to do `()=>()=>{}` we can do `()=>{}` with use callback. 

So instead of creating a intermediate object and having a extra anonymous function call (since useMemo works by saving the object value to a anonymous function) we can return the function directly while still memorizing it creating a stable function without the additional assignment.

Small but simple performance optimization to follow best practices.

PR #7605, opened 2024-12-22
perf: Extract regex patterns + more arrow functions

VIKTORVAV99:
## Issue

If we declare the regex inside the functions they will be recompiled every time the function is called, but if we extract them to their own constants outside the functions then they are only compiled once.

## Description

Extract the regex patterns + convert some functions to arrow functions for consistency and size.

### Double check

### DIFF
--- a/web/src/components/ShareIcon.tsx
+++ b/web/src/components/ShareIcon.tsx
@@ -3,6 +3,8 @@ import { Share, Share2 } from 'lucide-react';
 import { memo } from 'react';
 import { DEFAULT_ICON_SIZE } from 'utils/constants';
 
+const IOS_DEVICE_PATTERN: RegExp = /Mac|iPad|iPhone|iPod/;
+
 export function ShareIcon({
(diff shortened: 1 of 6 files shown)


cadeban:
There is probably an opportunity for consolidating the regexes (for example, `MOBILE_USER_AGENT_PATTERN` in utils/helpers is the same as `MOBILE_DEVICE_PATTERN` in wind-layer/utils), but that's outside of the scope of this PR.

It's much nicer to have the regexes extracted!

cadeban (approved):

VIKTORVAV99:
> There is probably an opportunity for consolidating the regexes (for example, `MOBILE_USER_AGENT_PATTERN` in utils/helpers is the same as `MOBILE_DEVICE_PATTERN` in wind-layer/utils), but that's outside of the scope of this PR.
> 
> It's much nicer to have the regexes extracted!

Yeah I thought the same but that is a later PR!

PR #7609, opened 2024-12-23
perf: memoize hot-paths

VIKTORVAV99:
## Issue
We have some functions that are called hundred if not even thousands of times while a normal user is using the app.

On of these functions, `getLocalTime` is called 72 times per timeAxis, of which there is normally at least 5 on every zone view. This means it's called 360 times at just startup when visiting a zone.

It is also called 360 times more every time the date changes using the slider or 288 times every time the hover state changes on a AreaGraph.

## Description

Clearly there is some logic optimizations that can be done here to prevent it from being called as many times (up to 9720 times if you just use the slider to go back 72 hours) but in the meantime we can memoize the function at least so it returning cached results. This means that only the first 72 calls (for the 72 hour view) will actually go though to the internal logic and after that they will all be cached.

This PR enables such optimizations for 3 identified hot-paths that are called often with deterministic results.

> [!WARNING]
> Technically there is a risk of a memory leak here if the app never restarts as the caches will grow to infinity with new dates. But I think this risk is pretty low in reality.

> [!NOTE]
> This is a hard PR to test as nothing should have changed visually if it was done right. But I suggest checking out the branch locally and adding a `console.log()` statement for either one of the source functions and then checking how many times it was called. It should be at most 72 hours if you use the `getLocalTime` one and only use the 72 hour view.

### Double check

### DIFF
--- a/web/src/utils/helpers.ts
+++ b/web/src/utils/helpers.ts
@@ -1,4 +1,5 @@
 import { callerLocation, useMeta } from 'api/getMeta';
+import { memoize } from 'es-toolkit';
 import {
   useLocation,
   useMatch,
@@ -43,9 +44,10 @@ export function useUserLocation(): callerLocation {
 /**
  * Converts date to format returned by API
  */
-export function dateToDatetimeString(date: Date) {
-  return date.toISOString().split('.')[0] + 'Z';
-}
+export const dateToDatetimeString = (date: Date) =>
+  date.toISOString().split('.')[0] + 'Z';
+export const memoizedDateToDatetimeString = memoize(dateToDatetimeString);
+
 export function getProductionCo2Intensity(
   mode: ElectricityModeType,
   zoneData: ZoneDetail
@@ -240,6 +242,7 @@ export const hasMobileUserAgent = () =>
 
 export const isValidHistoricalTimeRange = (timeRange: TimeRange) =>
   historicalTimeRange.includes(timeRange);
+export const memoizedIsValidHistoricalTimeRange = memoize(isValidHistoricalTimeRange);
 
 export const getLocalTime = (date: Date, timezone?: string) => {
   if (!timezone) {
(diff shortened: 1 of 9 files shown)


VIKTORVAV99:
@madsnedergaard this don't fix the re-render issues (which is a much larger task) but it does reduce the impact a bit!

VIKTORVAV99:
Note that #7610 reduces the need for this as it eliminates a lot of the re-renders that called these functions. That's this is still useful but maybe not as much as before.

web/src/utils/helpers.ts
```diff
@@ -1,4 +1,5 @@
 import { callerLocation, useMeta } from 'api/getMeta';
+import { memoize } from 'es-toolkit';
```
madsnedergaard, line 2: Lol, didn't even know that there's a new `lodash` in town 😄 
Seems pretty overkill for just this memo function IMO:

Screenshot 2025-02-18 at 21 37 27

That said I'm not against having it per se, but I do feel like it might be more trouble than what it's worth at this point :)

web/src/utils/helpers.ts
```diff
@@ -240,6 +242,7 @@ export const hasMobileUserAgent = () =>
 
 export const isValidHistoricalTimeRange = (timeRange: TimeRange) =>
   historicalTimeRange.includes(timeRange);
+export const memoizedIsValidHistoricalTimeRange = memoize(isValidHistoricalTimeRange);
```
madsnedergaard, line 245: nit: It's already done so it's fine, but another time I'd probably not rename the exported function but keep that the same - to me, `isValidHistoricalTimeRange` is what the user should care about and the fact that it is memoized is an internal implementation detail that is not relevant to expose for something like this :)

madsnedergaard (approved):
LGTM, only a few minor comments that aren't blocking :)

web/src/utils/helpers.ts
VIKTORVAV99: That's the benefit of the ES-toolkit, fully modular and uses es modules so if we are only using the memorize function that's the only code we are importing!

So size wise the addition is pretty small. About as much code as the basic boilerplate for one component.

madsnedergaard: Nice, there's probably also other custom code we could get rid of then 🙂

PR #7610, opened 2024-12-23
perf: memoize TimeAxis

VIKTORVAV99:
## Issue

This component where not memoized and are re-rendered very often!

## Description

Memoizes the whole component, switches to use Group from `@visx/group` for some nicer syntax, creates, intermediate variable to avoid calling the same function twice and removes some unused code.

### Double check


tonypls (approved):
LGTM!

PR #7611, opened 2024-12-23
perf: memoize MapTooltip internals

VIKTORVAV99:
## Issue

The internals of the tooltips re-render when you move the mouse even if it's on the same zone.

## Description

Memoizes the internal components of the map tooltip so they are stable, also memoizes the gauges, co2 square and estimated badge to be stable even between zones. This helps especially on zones which are fully estimated or have no data at all.

### Double check

### DIFF
--- a/web/src/components/ZoneGauges.tsx
+++ b/web/src/components/ZoneGauges.tsx
@@ -1,18 +1,19 @@
 import { useAtomValue } from 'jotai';
+import { memo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { StateZoneData } from 'types';
 import { getCarbonIntensity, getFossilFuelRatio, getRenewableRatio } from 'utils/helpers';
 import { isConsumptionAtom } from 'utils/state/atoms';
 
 import CarbonIntensitySquare from './CarbonIntensitySquare';
-import { CircularGauge } from './CircularGauge';
+import CircularGauge from './CircularGauge';
 
 interface ZoneGaugesWithCO2SquareProps {
   zoneData: StateZoneData;
   withTooltips?: boolean;
 }
 
-export default function ZoneGaugesWithCO2Square({
+function ZoneGaugesWithCO2Square({
   zoneData,
   withTooltips = false,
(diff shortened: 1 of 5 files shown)


web/src/components/ZoneGauges.tsx
```diff
@@ -49,3 +50,10 @@ export default function ZoneGaugesWithCO2Square({
     </div>
   );
 }
+
+export default memo(
+  ZoneGaugesWithCO2Square,
+  (prevProps, nextProps) =>
+    JSON.stringify(prevProps.zoneData) === JSON.stringify(nextProps.zoneData) &&
+    prevProps.withTooltips === nextProps.withTooltips
+);
```
VIKTORVAV99, line 59: The question here is if this costs more than it's worth. I'd be happy to delete it if that is the case.

VIKTORVAV99: Decided to remove it and move the empty object creation outside the function so it's stable.

web/src/features/map/MapTooltip.tsx
```diff
@@ -9,33 +9,28 @@ import ZoneGaugesWithCO2Square from 'components/ZoneGauges';
...
 import { StateZoneData } from 'types';
 import { round } from 'utils/helpers';
 import { selectedDatetimeStringAtom } from 'utils/state/atoms';
 
 import { hoveredZoneAtom, mapMovingAtom, mousePositionAtom } from './mapAtoms';
 
-export function TooltipInner({
+const emptyZoneData: StateZoneData = {
+  p: {},
+  c: {},
```
madsnedergaard, line 22: Is ci/fr/rr not needed on the empty data any longer?

madsnedergaard (approved):
LGTM, one minor question

web/src/features/map/MapTooltip.tsx
VIKTORVAV99: They where always optional as the internal function uses optional chaining, I don't really know why I added them last time.

PR #7613, opened 2024-12-24
fix(parsers): make validate_reasonable_time relate to utc dates

consideRatio:
## Issue

Actual issue fixed: A log message was provided with two dates logged but without timezone details, making the comparison hard when it could have been made easy by using UTC systematically.

A bundled edge case change:
Via https://github.com/electricitymaps/electricitymaps-contrib/issues/1168#issuecomment-369597354 we got a course check to ensure data imported was in the right milennia. This course check will now check the year using UTC datetime which probably won't matter. Maybe it shouldn't been changed at all as it wasn't what I looked to fix and could muddy this PR - happy to revert and not change it.

### Double check

  _Comment_: This could impact all parsers, and I have not run them all.

### DIFF
--- a/parsers/lib/quality.py
+++ b/parsers/lib/quality.py
@@ -40,14 +40,14 @@ def validate_datapoint_format(datapoint: dict[str, Any], kind: str, zone_key: Zo
 
 
 def validate_reasonable_time(item, k):
-    data_time = arrow.get(item["datetime"])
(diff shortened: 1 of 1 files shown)


consideRatio:
Nevermind, I didn't see the +0800 etc in the end of the string. I figure its more helpful to compare with utc/utc, but i think this is a small incremental improvement at best and its worth doing right from the start if its to merit review capacity.

PR #7624, opened 2024-12-27
perf: optimize object creation in ExchangeArrow.tsx

VIKTORVAV99:
## Issue

We are creating a object for no reason which is causing both size and performance downsides.

## Description

Removes the object creation and are using pure const declarations, object deconstruction for the parent values, and compute the comparison factor once.

### Double check


web/src/features/exchanges/ExchangeArrow.tsx
```diff
@@ -108,29 +108,26 @@ function ExchangeArrow({
...
-  const transform = {
-    x: projection.x,
-    y: projection.y,
-    k: 0.04 + (mapZoom - 1.5) * 0.1,
-    r: rotation + (netFlow > 0 ? 180 : 0),
-  };
+  const { x, y } = map.project(lonlat);
+  const k = 0.04 + (mapZoom - 1.5) * 0.1;
+  const r = rotation + (netFlow > 0 ? 180 : 0);
+  const factor = 100 * k;
```
tonypls, line 114: Could do some nice naming here:

```suggestion
  const baseZoom = 0.04 + (mapZoom - 1.5) * 0.1;
  const directionalRotation = rotation + (netFlow > 0 ? 180 : 0);
  const scaledZoom = 100 * k;
```

Would also need to update below

tonypls (approved):
Looks like it should work! Small naming improvment

web/src/features/exchanges/ExchangeArrow.tsx
VIKTORVAV99: I originally kept the names the same so the diff would be as small as possible but I'd be happy to rename them before merging!

tonypls: Sounds good!

PR #7627, opened 2024-12-27
refactor/fix(parser): swap arrow in favour of datetime in IN_WE, and etc.

consideRatio:
## Issue

## Description

I ended up doing a lot of other refactoring in this PR as well =/ I aim not to do that to reduce the complexity of code review and keep PRs minimalistic, but I went for it anyhow this time as I struggled a while to get it to run correctly and had to understand the code better.

My gained code comprehension ended up making me fix a bug, where fetching could lead getting old data even though there was new data available.

I also fixed what I think is a bug that could have led to a bit incorrect data by mixing a few data points from an AM hour (like 04) into an PM hour (like 16:00) or vice versa.

## `is_expected_downtime` removed

I asked about this in https://github.com/electricitymaps/electricitymaps-contrib/pull/6184#issuecomment-2564564977, but since I've got fresh data from the API during this expected downtime, I went with removing it for now. I'll look to see if the API is actually providing data regularly going forward, and will be ready to add back exception handling if needed.

## Double check

  ```
  poetry run test_parser "IN-WE" consumption
  poetry run test_parser "IN-WE" consumption --target_datetime="2024-12-27 00:00 +05:30"
  poetry run test_parser "IN-SO->IN-WE"
  poetry run test_parser "IN-SO->IN-WE" --target_datetime="2024-12-27 00:00 +05:30"
  poetry run test_parser "IN-EA->IN-WE"
  poetry run test_parser "IN-NO->IN-WE"
  ```

### DIFF
--- a/parsers/IN_WE.py
+++ b/parsers/IN_WE.py
@@ -75,19 +75,24 @@ def fetch_data(
             message=f"{target_datetime}: {kind} data is not available",
         ) from e
 
+    # The source data is a 12 hour format without mentioning if its AM/PM, so
+    # 12:15 (AM or PM) could mean 00:15 or 12:15. This is addressed by relying
+    # on the full date worth of ordered data, where we assume the second time
+    # 12:xx shows up it must have transitioned to PM time.
+    am_time = True
+    checkpoint = False
+    dt_format = "%Y-%d-%m %I:%M:%S %p"
     for item in data:
-        # replace datetime string with datetime object
-        dt = datetime.strptime(item[datetime_col], "%Y-%d-%m %H:%M:%S").replace(
-            tzinfo=ZONE_INFO
-        )
+        dt_string = item[datetime_col] + (" AM" if am_time else " PM")
+        dt = datetime.strptime(dt_string, dt_format).replace(tzinfo=ZONE_INFO)
+        if am_time:
+            if not checkpoint and dt.hour != 0:
+                checkpoint = True
+            elif checkpoint and dt.hour == 0:
+                am_time = False
+                dt_string = item[datetime_col] + (" AM" if am_time else " PM")
+                dt = datetime.strptime(dt_string, dt_format).replace(tzinfo=ZONE_INFO)
         item[datetime_col] = dt
-
-        # round the seconds to 0 as a non-robust strategy to later distinguish
-        # AM from PM
-        if dt.second >= 30:
-            item[datetime_col] = dt + timedelta(seconds=60 - dt.second)
-        else:
-            item[datetime_col] = dt - timedelta(seconds=dt.second)
     return data
 
 
@@ -98,27 +103,12 @@ def filter_raw_data(
 ) -> pd.DataFrame:
     """
     From 24 hours of data, filter out a specific hour of interest.
-
-    The source data is a 12 hour format without mentioning if its AM/PM, so
-    11:15 could mean 11:15 or 23:15.
     """
     datetime_col = KIND_MAPPING[kind]["datetime_col"]
-    region_col = KIND_MAPPING[kind]["region_col"]
 
-    # 00:mm is never reported, only 12:mm, so we need to filter based on an hour
-    # value from one to twelve instead of zero to eleven
-    one_to_twelve = (target_datetime.hour % 12) or 12
-    two_hours_df = pd.DataFrame(
-        [item for item in data if item[datetime_col].hour == one_to_twelve]
+    return pd.DataFrame(
+        [item for item in data if item[datetime_col].hour == target_datetime.hour]
     )
-
-    # We rely on the data being ordered, and drop first/last duplicates
-    # depending on the hour of interest.
-    one_hour_df = two_hours_df.drop_duplicates(
-        subset=[region_col, datetime_col],
-        keep="first" if target_datetime.hour < 12 else "last",
-    )
-    return one_hour_df
 
 
 def format_exchanges_data(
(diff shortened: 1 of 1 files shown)


parsers/IN_WE.py
```diff
@@ -29,48 +30,60 @@
...
 def is_expected_downtime() -> bool:
-    current_day = datetime.now().weekday()
+    """
+    Details about this expected downtime is currently being clarified via
+    https://github.com/electricitymaps/electricitymaps-contrib/pull/6184#issuecomment-2564302911.
+    """
     expected_outage_days = [5, 6, 0]  # Saturday, Sunday and Monday
-
-    if current_day in expected_outage_days:
+    if datetime.now(ZONE_INFO).weekday() in expected_outage_days:
```
consideRatio, line 52: I think this is a small bugfix, but this is not critical in any way I think.

parsers/IN_WE.py
```diff
@@ -29,48 +30,60 @@
...
         return True
-
     return False
 
 
 def get_date_range(dt: datetime):
+    """Returns 24 datetime objects for a given datetime's date, one for each hour."""
+    now_date_as_dt = datetime.combine(
+        datetime.now(ZONE_INFO).date(), datetime.min.time()
+    ).replace(tzinfo=ZONE_INFO)
```
consideRatio, line 61: Thisis a floor("day") equiavlent operation.

parsers/IN_WE.py
```diff
@@ -29,48 +30,60 @@
...
+    Details about this expected downtime is currently being clarified via
+    https://github.com/electricitymaps/electricitymaps-contrib/pull/6184#issuecomment-2564302911.
+    """
     expected_outage_days = [5, 6, 0]  # Saturday, Sunday and Monday
-
-    if current_day in expected_outage_days:
+    if datetime.now(ZONE_INFO).weekday() in expected_outage_days:
         return True
-
     return False
```
VIKTORVAV99, line 54: Nitpick but this could simply do:
```python
return datetime.now(ZONE_INFO).weekday() in expected_outage_days
```

parsers/IN_WE.py
```diff
@@ -29,48 +30,60 @@
...
     payload = {"date": target_datetime.strftime("%Y-%m-%d")}
-
-    resp: Response = r.post(url=KIND_MAPPING[kind]["url"], json=payload)
+    resp: Response = session.post(url, json=payload)
 
     try:
         data = json.loads(resp.json().get("d", {}))
     except Exception as e:
-        if is_expected_downtime():
+        if is_expected_downtime(target_datetime):
```
VIKTORVAV99, line 86: This should not have a datetime passed as it cause type issues.

parsers/IN_WE.py
```diff
@@ -147,37 +171,28 @@ def format_exchanges_data(
 
 
 def format_consumption_data(
-    data: dict, zone_key: str, target_datetime: datetime
+    data: dict,
+    zone_key: str,
```
VIKTORVAV99, line 175: Not sure why this was included before but zone_key is not used anywhere in this function.

VIKTORVAV99 (commented):
There is just one tiny type issue that needs fixing before it can be merged otherwise things looks good to me.

Added two tiny nitpicks as well but those are not required.

Thanks for creating the PR!

parsers/IN_WE.py
consideRatio: Ah, of course, I'll fix it!

parsers/IN_WE.py
consideRatio: Nice catch! I removed the `is_expected_downtime`, it seems to work but its currently such downtime, so I think they have made it work reliably.

parsers/IN_WE.py
consideRatio: Resolved!

consideRatio:
Still doing some work, making it draft

parsers/IN_WE.py
```diff
@@ -29,65 +30,66 @@
...
-        return True
-
-    return False
-
-
 def get_date_range(dt: datetime):
+    """Returns 24 datetime objects for a given datetime's date, one for each hour."""
+    now_date_as_dt = datetime.combine(
+        datetime.now(ZONE_INFO).date(), datetime.min.time()
+    ).replace(tzinfo=ZONE_INFO)
```
consideRatio, line 50: `datetime.now` shouldn't be used, but `dt`. Will fix this regression I've introduced.

consideRatio:
@VIKTORVAV99 I ran into various bugs with the code as it was, such as getting old data even when there was new data. I ended up refactoring things to understand things well enough to resolve them.

This makes this PR quite messy because it includes bugfixes, significant amounts of refactoring, and then as the smaller part - the arrow removal.

VIKTORVAV99:
> @VIKTORVAV99 I ran into various bugs with the code as it was, such as getting old data even when there was new data. I ended up refactoring things to understand things well enough to resolve them.
> 
> 
> 
> This makes this PR quite messy because it includes bugfixes, significant amounts of refactoring, and then as the smaller part - the arrow removal.

That's fine and I appreciate the heads up but it will probably take me a bit longer to review so might not get that done before new years as I have some stuff to do tomorrow.

consideRatio:
Not sure this fixes https://github.com/electricitymaps/electricitymaps-contrib/issues/6047 reliably, but quite confident it greatly improves the situation

VIKTORVAV99 (approved):
LGTM! 🎉

There is no type errors and the output is the same so I'd say that is more than good enough reason to merge this as is!

Thanks a lot!

PR #7628, opened 2024-12-28
perf: Memoize Link.tsx

VIKTORVAV99:
## Issue

Every time a zone was clicked the whole header re-rendered.

## Description

Memoizes the Link component and uses callbacks for the onClick functions in the header so they are not recreated every time (even though there is a lot less re-renders now), following best practices.

### Double check


web/src/features/header/Header.tsx
```diff
@@ -57,10 +59,16 @@ function MenuLink({
...
-  };
   const { t } = useTranslation();
+
+  const onFAQClick = useCallback(() => {
+    setIsFAQModalOpen(true);
+  }, [setIsFAQModalOpen]);
+
+  const handleClick = useCallback(() => {
+    trackEvent(TrackEvent.HEADER_LINK_CLICKED, { linkId: 'get-data' });
+  }, []);
```
VIKTORVAV99, line 70: I'll move this outside the function instead as it don't need to be inside the component due to static data.

PR #7629, opened 2024-12-28
refactor(parser): swap arrow in favour of datetime in IN_UT

consideRatio:
## Issue

Note though that this parser is in a very bad state for other reasons:
- seems to be unused currently, as its not tied to India's subzones.
- seems to be broken, because:
  - the parser works against a website that isn't serving data at the URL any more
  - the new website URL isn't providing data, not even outdated data

### Double check

  ```
  poetry run test_parser "IN-UT" production
  poetry run test_parser "IN-UP->IN-UT"
  poetry run test_parser "IN-HP->IN-UT"
  ```

### DIFF
--- a/parsers/IN_UT.py
+++ b/parsers/IN_UT.py
@@ -2,13 +2,27 @@
 
 from datetime import datetime
 from logging import Logger, getLogger
+from zoneinfo import ZoneInfo
 
-import arrow
 from requests import Session
 
 from .lib import web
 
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

If these parsers really are not working anymore I think we could remove them but I need to talk with some people about that first.

Thanks!

PR #7630, opened 2024-12-29
refactor/fix(parser): swap arrow in favour of datetime in CO, and etc.

consideRatio:
## Issue

---

This commit replaces arrow, does some additional refactoring, and fixes a small bug.

The bug was that the requested data was from a date determined by a parser executing local datetime instead of the API local datetime.

The additional refactoring includes for example:
- De-nesting code under conditionals by raising an error early if needed under a conditional, but then de-nesting the other code block.
- Renaming local functions by prefixing them with an underscore
- Coalescing code that sometimes is run once and sometimes looped over by always looping (but sometimes only one iteration).

This commit was made without the ability to test the parser against the API, as the API was down or wasn't accessible from Sweden.

### Double check

  No! The API isn't accessible for me, so I haven't been able to test this!


parsers/CO.py
```diff
@@ -17,11 +17,19 @@
 from parsers.lib.config import refetch_frequency
 from parsers.lib.exceptions import ParserException
 
+# FIXME: Issues as of 2024-12-29
+#
+# 1. The API isn't accessible from anywhere, for example Sweden can't access it.
+#
+# 2. There is a use of `replace(hour=int(col[-2:]) - 1)`, but should it really
+#    be -1 there?
```
VIKTORVAV99, line 25: Will do a proper review later but the only reason I see for this is if they use a 1 to 24 hour accounting method instead of 0 to 23.

But will try and figure it out later.

VIKTORVAV99 (approved):
LGTM! 🎉

Haven't managed to test it, even with a VPN to Bogata, so not sure what else we can do at the moment.

I'll approve and merge it so we can get rid of arrow but It's likley we need to revisit this parser later.

Thanks for the PR!

PR #7631, opened 2024-12-29
perf: Refractor and memoize ZoneHeaderTitle.tsx

VIKTORVAV99:
## Issue

We are calling a lot of internal functions to compute the correct name, flag, disclaimer and county pill and the component is re-rendered often.

## Description

Splits out the actual title stuff into it's own component which is memoized to avoid calling all the internal on every datetime change.

> [!NOTE]
> The contents and logic is the same except for the memo so there is no changes in how things work on a logic level or visual level.

### Double check


tonypls (approved):
LGTM!

PR #7633, opened 2024-12-29
perf: Memoize ToggleButton.tsx and specific toggles

VIKTORVAV99:
## Issue

These toggles are re-rendered often even if their states or props don't change.

## Description

Memoizes the toggles, moves out the object declaration to outside the function so it's stable and uses useCallbacks to ensure the internal functions are stable between renders.

### Double check


tonypls (approved):
LGTM!

PR #7638, opened 2024-12-30
perf: Memoize Axis.tsx

VIKTORVAV99:
## Issue

The axis where not memoized and re-rendered very often!

## Description

Memoizes the component and makes the props stable to ensure it's only re-rendered when needed.

### Double check


cadeban (approved):
LGTM, though I like your idea about splitting the axisLegendText. 

Were you able to get some numbers comparing before and after?

VIKTORVAV99:
> LGTM, though I like your idea about splitting the axisLegendText.
> 
> Were you able to get some numbers comparing before and after?

Not performance wise but it re-rendered like 100 times per second if not more when you hovered on anything in the barBreakdownChart...

VIKTORVAV99:
For some reason the performance is extra bad right now but here is the problem:

https://github.com/user-attachments/assets/b26e76e9-56e0-4baa-a84e-51a442c88640

This is running in the dev server though and things are a bit better in production builds.

PR #7640, opened 2024-12-30
chore(config): Remove other forecast keys

florianscheidl:
## Description

Deletes the unused constant OTHER_FORECAST_KEYS

### Double check

### DIFF
--- a/electricitymap/contrib/config/constants.py
+++ b/electricitymap/contrib/config/constants.py
@@ -13,11 +13,6 @@ WEATHER_DATABASE_KEYS = [
     "dewpoint",
     "precipitation",
 ]
-OTHER_FORECAST_KEYS = [
-    "price",
-    "production",
-    "consumption",
-]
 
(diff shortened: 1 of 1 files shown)


electricitymap/contrib/config/constants.py
```diff
@@ -13,11 +13,6 @@
     "dewpoint",
     "precipitation",
 ]
-OTHER_FORECAST_KEYS = [
-    "price",
-    "production",
-    "consumption",
-]
```
VIKTORVAV99, line 20: Why are these in contrib to start with? 👀

VIKTORVAV99 (approved):
LGTM! 🎉

electricitymap/contrib/config/constants.py
florianscheidl: I think it's legacy that we copied to the monorepo later but never deleted this 🙈

PR #7646, opened 2024-12-31
perf: Memoize FuturePrice.tsx

VIKTORVAV99:
## Issue

Same as with the methodology card this is re-rendered when the selected datetime changes even though it's not affected by it.

## Description
Memoizes the FuturePrice component.

### Double check


silkeholmebonnen (approved):
LGTM 😄

PR #7568, opened 2024-12-17
feat(web): Update information copy and CTAs

tonypls:
## Issue
Users perceive us as an NGO 

## Description
This PR updates the app description and about sections on the ranking panel and info modal to better explain the value provided by electricity maps. Removes clutter from the ranking panel and makes "Access the API" CTA's more discoverable

PR #7647, opened 2024-12-31
perf/fix: Remove trans component from RankingPanel.tsx

VIKTORVAV99:
## Issue

A innocent little change introduced in #7568 caused our bundle size to increase by 7.3 kb and the subtitle to constantly re-render when the datetime changes.
## Description
Switches back to using the `t()` function instead to fix this regression

### Double check


tonypls:
This was to create add the correct breakpoint, see image:
image

Is there another way we can break the line after "from" and remove the <br/>?

tonypls:
> This was to create add the correct breakpoint, see image: image
> 
> Is there another way we can break the line after "from" and remove the <br forwardslash> ?

VIKTORVAV99:
> This was to create add the correct breakpoint, see image:
> 
> image
> 
> 
> 
> Is there another way we can break the line after "from" and remove the br?

Hmm I see, by looking at the documentation it don't look like it was intended to do that.

i18next does keep some html tags like <b> and <i>, I'll see if I can create a PR for <br> as well. Let's leave this open until then.

PR #7650, opened 2025-01-01
fix(parser, SV): adjust to data source changes

consideRatio:
## Issue

Expected to make #5158 self-close, as the parser now works.

I went ahead and added myself as a contributor in the zone config.

Noteworthy parser changes:
- updated previously hardcoded POST request data to match what was seen now at https://estadistico.ut.com.sv/OperacionDiaria.aspx
- updated parsing of POST request response to adjust to small changes
- removed a lot of no longer needed logic thanks to a cleaner data response

### Double check

- [x] I have tested my parser changes locally with `poetry run test_parser "zone_key"`
- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.

PR #7651, opened 2025-01-01
chore: update poetry.lock

consideRatio:
I'm proposing a new dependency in #7650, but updating the poetry.lock file there led to a lot of other things, so, I'm updating it in isolation without any dependency change here for a clean review.

All I've done is run `poetry lock` in the master branch.


consideRatio:
This wasn't as trivial as I hoped, something breaks doing this it seems - closing for now as I lack experience with poetry

PR #7669, opened 2025-01-01
chore: Update ENTSO-E capacities with values from 2017 to 2025

VIKTORVAV99:
## Issue
For the outlier detection to work properly we need good capacity data.

## Description
Updates the ENTSO-E zones with capacity data from 2017 to 2025.

> [!NOTE]
> Several zones still show N/A for 2025 data so we'll likely need to update these again soon.

### Double check


FelixDQ (approved):
Cool! I can't really verify these numbers, but would you be able to monitor https://intramaps.electricitymaps.com/overview/ after we've merged this to see if we start flagging a lot of events for some zones?

VIKTORVAV99:
> Cool! I can't really verify these numbers, but would you be able to monitor https://intramaps.electricitymaps.com/overview/ after we've merged this to see if we start flagging a lot of events for some zones?

Will do!
Nothing should change for the historical data until we re-run the pipeline. I'll do so gradually when I have some extra time and monitor the results!

consideRatio:
@VIKTORVAV99 test failure here is fixed by #7693 - impacts the Python test in master branch

config/zones/AL.yaml
```diff
@@ -5,50 +5,56 @@ bounding_box:
...
       source: entsoe.eu
       value: 0.0
   nuclear:
-    - datetime: '2023-01-01'
+    - datetime: '2021-01-01'
       source: entsoe.eu
       value: 0.0
   oil:
-    - datetime: '2023-01-01'
+    - datetime: '2021-01-01'
```
tonypls, line 39: @VIKTORVAV99 is this correct? Seems weird to change the date back 2 years?

VIKTORVAV99: Yes, this capacity format basically only record the date of the first time that value has been seen and then that's valid until the next value.

So if the capacity for a mode is:
0 for year 1
2 for year 2
2 for year 3
4 for year 4

Then the you can skip the year 3 in the config as the year 2 value will be valid until year 4.

config/zones/HU.yaml
```diff
@@ -55,20 +139,47 @@ capacity:
     - datetime: '2024-10-01'
       source: mavir.hu
       value: 3964.8
+    - datetime: '2025-01-01'
+      source: entsoe.eu
+      value: 7473.0
```
torokati44, line 144: A change to this value this large is _sus_ to me, even though ENTSO-E also shows the same. Maybe it includes household-size (<=50kW) installations too now?

Maybe there has been a change in how metrics are summarized, but unless the live production data will also include these, things could get confusing.

torokati44: What do you think, @VIKTORVAV99?

VIKTORVAV99: There really is no way to tell from the data as far as I can see so our bet bet would be to reach out to ENTSOE about it, it's not a priority for me at the moment though as there are bigger data issues to deal with at the moment.

PR #7672, opened 2025-01-02
chore(parsers): test with syrupy and pytest over snapshottest and unittest

### LINKED ISSUE #7667: Developer documentation about re-generating test snapshots
In #7650 I learned about tests against snapshot data, and ended up manually updating it. However, based on the pytest failure message ``1 snapshots failed in 1 test suites. Inspect your code or run with `pytest --snapshot-update` to update them``, I figure its possible to do automatically.

However, I failed to update the snapshot data using either `poetry run test --snapshot-update` or `pytest --snapshot-update`, and instead it was as if I just ran the test without the passed flag.

I expect that it wasn't meant to be updated manually, but then how should it be done?

consideRatio:
## Issue

Fixes #7667 by making `--snapshot-update` work reliably for all parsers.

## Description

I've worked all tests except those in the root folder -- they weren't similar to the others, and didn't make use of snapshottest either, so I figured I better leave them.

This is a massive PR and tough to review, but I hope that the fact that it mostly works on the tests makes it a bit easier as its not likely to cause failures in production.

### Removal of `unittest.TestCase` classes

Change motivation:
- [pytests' documents that use of fixtures with unittest TestCase class](https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses) is limited, individual functions of a class can't get a fixture set and `syrupy` is meant to use the `snapshot` fixture in test functions. Maybe there is a way to work unittest.TestCase with `syrupy`, but I failed to find it and figured it was better to go with a more pytest native test suite.
- I've come to perceive that code reads a bit cleaner as `assert "something" in var` over for example `self.assertIn(something, var)` as is typical in unittest classes.

### Introduction of `conftest.py`

`pytest` respects `conftest.py` as a way to initialize things before running tests, where for example one can define fixtures for use by the tests. I used this to define an `adapter` and `session` fixture, to replace the behavior of unittest's `setUp` function that otherwise often initialized `self.adapter` and `self.session`.

### Double check


pyproject.toml
```diff
@@ -76,8 +76,8 @@ parsers = [
 scripts = ["xmltodict"]
 
 [tool.poetry.group.dev.dependencies]
-snapshottest = "^0.6.0"
 pytest = "^7.4.0"
+syrupy = "^4.8.0"
```
consideRatio, line 80: `snapshottest` removed, and `syrupy` added

parsers/test/test_US_SPP.py
```diff
@@ -1,9 +1,4 @@
-#!/usr/bin/env python3
```
consideRatio, line 1: I removed shebangs like these, as they aren't meant to be executable files as pytest test files.

parsers/test/test_US_SPP.py
```diff
@@ -1,9 +1,4 @@
...
-
-"""Tests for US_SPP.py"""
```
consideRatio, line 3: I removed obvious docstrings like these as well.

electricitymap/contrib/capacity_parsers/tests/test_EMBER.py
```diff
@@ -14,13 +12,6 @@
...
-class testEmber(unittest.TestCase):
-    def test_map_variable_to_mode(self):
-        test_df["mode"] = test_df.apply(map_variable_to_mode, axis=1)
-        self.assertEqual(
-            test_df["mode"].tolist(), ["oil", "geothermal", "oil", "solar"]
-        )
-
-
-if __name__ == "__main__":
-    unittest.main()
```
consideRatio, line 26: Removed these as well

parsers/test/test_CY.py
```diff
@@ -1,69 +1,62 @@
...
 
 from electricitymap.contrib.lib.types import ZoneKey
 from parsers.CY import HISTORICAL_SOURCE, REALTIME_SOURCE, fetch_production
 
 
-class TestFetchProduction(TestCase):
-    def setUp(self):
-        self.adapter = Adapter()
-        self.session = Session()
-        self.session.mount("https://", self.adapter)
```
consideRatio, line 15: All of these could be removed in favor of the `adapter` and `session` fixtures in `conftest.py`.

parsers/test/test_CY.py
```diff
@@ -1,69 +1,62 @@
...
-                }
-                for element in production
-            ]
+@pytest.fixture(autouse=True)
+def mock_response(adapter):
+    with open(
+        "parsers/test/mocks/CY/response_historical_20240318.html", "rb"
+    ) as response:
+        adapter.register_uri(
+            GET, HISTORICAL_SOURCE.format("18-03-2024"), content=response.read()
```
consideRatio, line 21: When a `setUp` included a mock for multiple tests, or if the same mocking was initialized multiple times in all tests, I created a file-local fixture like this to make responses get mocked.

parsers/test/test_AW.py
```diff
@@ -1,45 +1,35 @@
...
+        {
+            "datetime": element["datetime"].isoformat(),
+            "production": element["production"],
+            "storage": element["storage"],
+            "source": element["source"],
+            "zoneKey": element["zoneKey"],
+            "sourceType": element["sourceType"].value,
+        }
+        for element in production
+    ]
```
VIKTORVAV99, line 35: I even tested changing this to just:
```python
    assert snapshot == production
```
which now works perfectly!

I don't think you should go and change every test like this in this PR but it will be real nice that we can update these in the future and get even more accurate snapshots instead of serializing the data like we had to do with the old library!

VIKTORVAV99 (commented):
WOO HOO!

Very nice change, I have been looking forward to this but never had time to prioritize it myself so I really appreciate this a lot!

With that said there is a huge amount of changes in here and while they look straight forward I'll have take the weekend to give this a proper review.

But hopefully we can merge this on Monday!

parsers/test/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+def adapter():
+    adapter = Adapter()
+    yield adapter
+
+
+@pytest.fixture
+def session(adapter):
+    session = Session()
+    session.mount("https://", adapter)
+    yield session
```
VIKTORVAV99, line 16: Could we do this for HTTP too?

I'm not sure we can guarantee that everything will be HTTPS all the time.

parsers/test/test_ES.py
```diff
@@ -1,216 +1,188 @@
...
-        )
-
-
-if __name__ == "__main__":
-    main()
+### El Hierro
+# Consumption
+@patch("requests.Response")
+@patch("parsers.ES.Session.get")
+def test_fetch_consumption(mocked_session_get, mocked_response, session):
```
VIKTORVAV99, line 12: This seems to still work but do you think using unittest.mock for this will cause issues in the future?

VIKTORVAV99 (approved):
LGTM! 🎉

Two tiny comments all thing considering but you can consider those optional, will merge this either when those are resolved (either code wise or just marked as such) or at the end of the weekend.

parsers/test/test_ES.py
consideRatio: No i think it will be fine, its just a way to manipulate a functions response and such I think, independent on how a test is executed etc, but I think it can be done also by using the request adapter as i saw in most other tests.

But, its a difference to how its done elsewhere, and being systematic is often positive from a maintenance perspective long term. I didnt feel very confident to make a call or dig into this so it seemed reasonable to not change as part of this

parsers/test/conftest.py
consideRatio: True true, i didn't feel confident about the impact of mount, so i went for https nonstop instead and transktioned two tests to work against an https endpoint as it also worked in those cases.

Do you think it is fine to do a http:// alongside here without a negative consequence do you think?

I'm leaning towards adding it if its needed in the future, but adding a comment about it for now saying that it hasnt been needed so far, but that it probably can be added whenever it is

consideRatio: I'll read up and trial a bit!

parsers/test/test_ES.py
consideRatio: I'm happy to try migrate this to look like most other test using adapter fixture, but i figure its better to avoid bundling it in this oversized PR

consideRatio:
Thank you soo much @VIKTORVAV99 for your quick review efforts, its such a great experience to contribute!

parsers/test/test_ES.py
VIKTORVAV99: Let's keep it as is for now then!

parsers/test/conftest.py
VIKTORVAV99: We have other adapters for non test purposes that does it and they seem to work. You can take a look here if you wish:
https://github.com/electricitymaps/electricitymaps-contrib/blob/9e8130fe2207d85453421babad8a5d5dac2aa55c/parsers/lib/session.py#L16-L22

So I think is should be fine!

VIKTORVAV99:
> Thank you soo much @VIKTORVAV99 for your quick review efforts, its such a great experience to contribute!

Thanks! It really means a lot! 💚

VIKTORVAV99:
You PRs are always clean and straight to the point (even if they refractor code) so it's my pleasure to review them!

consideRatio:
Rebased without other changes on master branch and pushed the changes in a dedicated commit in cf62a74, and marked the adapter / `http://` comment as resolved.

PR #7674, opened 2025-01-03
perf: Memoize BarBreakdownChart.tsx

VIKTORVAV99:
## Issue

The BarBreakdownChart is not memoized and re-renders often!

## Description

Memoizes the BarBreakdownChart so we don't spend resources re-rendering and re-computing the same values all the time!

#### Before:

https://github.com/user-attachments/assets/c9e43b99-a65b-4057-b7e9-f0f46990a7ba

#### After:

https://github.com/user-attachments/assets/221e21ec-8234-408a-beba-0ca77a5208a7

### Double check


web/src/features/charts/bar-breakdown/EmptyBarBreakdownChart.tsx
```diff
@@ -12,45 +13,34 @@ import { getDataBlockPositions } from './utils';
...
       scaleLinear()
-        .domain([
-          -maxCO2eqExport || 0,
-          Math.max(maxCO2eqProduction || 0, maxCO2eqImport || 0),
-        ])
+        .domain([MIN_CO2EQ, MAX_CO2EQ])
         .range([0, width - LABEL_MAX_WIDTH - PADDING_X]),
-    [maxCO2eqExport, maxCO2eqProduction, maxCO2eqImport, width]
+    [width]
   );
```
madsnedergaard, line 43: Oh, interesting with the fancy dynamic calculations that are using fixed values 😂

madsnedergaard (approved):
Looks good to me, thanks 🙌 

On a side-note, I wonder how well React Compiler would have handled the memo changes you've been doing - but on the other hand I also don't want to switch to React 19 and a beta Compiler at this point, so I still think it makes sense that you've done these great optimisations ❤️

web/src/features/charts/bar-breakdown/EmptyBarBreakdownChart.tsx
VIKTORVAV99: I think we just copied it over and never bothered to change it...

VIKTORVAV99: Heck I'm not even sure this is used anymore...

VIKTORVAV99:
> Looks good to me, thanks 🙌
> 
> On a side-note, I wonder how well React Compiler would have handled the memo changes you've been doing - but on the other hand I also don't want to switch to React 19 and a beta Compiler at this point, so I still think it makes sense that you've done these great optimisations ❤️

It would likely have memoized the components themselves but I don't think it could have done it all as some functions are declared inside the JSX as well (bad) and other structural changes that where needed like moving objects outside the function scopes. At least not the way I understood it. All code need to be reactive and follow the rules of react for it to work.

React 19 alone would probably not be a problem, most of our dependencies have already added peer support for it but I agree on the compiler part. Probably not worth it to run considering how complex the app actually is. 

But if I can remove all of this memo code later on and it just worked I'd be overjoyed!

PR #7676, opened 2025-01-04
fix: Prevent mockdata being counted towards language stats

VIKTORVAV99:
## Issue

Currently GitHub says HTML is the most popular language we use even though it's not...
image

## Description
Adds a line in our `.gitattributes` file that tells https://github.com/github-linguist/linguist to ignore all the files in `parsers/test/mocks` or any of it's subdirectories.

That should make Python number 1 again...


madsnedergaard (approved):
LGTM :)

PR #7677, opened 2025-01-04
chore: replace deprecated pkg_resources, cleanup __init__.py files

consideRatio:
## Description

pkg_resources is deprecated and its recommended to make use of `importlib.resources` instead, which is what this PR does.

From https://setuptools.pypa.io/en/latest/pkg_resources.html

image

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for the cleanup!

PR #7678, opened 2025-01-04
chore(tests): prefer use of requests_mock over unittest.mock

consideRatio:
## Description

Changes are triggered by the observation and discussion in https://github.com/electricitymaps/electricitymaps-contrib/pull/7672#discussion_r1902172671, where test_ES.py was the sole test file using unittest.mock's patch decorator, where all others used requests_mock.

I figure requests_mock is cleaner for us to use as we specifically mock requests, but its also nice to be consistent so that understanding one test leads to understanding other tests.

### Double check

  The parser wasn't changed, only tests.

### DIFF
--- a/parsers/test/__snapshots__/test_ES.ambr
+++ b/parsers/test/__snapshots__/test_ES.ambr
@@ -1,0 +1,144 @@
+# serializer version: 1
+# name: test_fetch_consumption
+  list([
+    dict({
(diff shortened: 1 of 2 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for even more cleanup!

PR #7679, opened 2025-01-04
chore(tests): compare snapshots directly with parsers output

consideRatio:
## Description

In https://github.com/electricitymaps/electricitymaps-contrib/pull/7672#discussion_r1901408703 @VIKTORVAV99 realized that with syrup's serialization, snapshots could be compared directly against the parsers output objects without first being processed. This PR makes use of that, reducing a lot of lines in test files, but increasing a bit of lines in snapshot files.

### Double check

  Not relevant because no parsers were touched, only their tests.


parsers/test/test_NORDPOOL.py
```diff
@@ -39,24 +39,12 @@ def test_price_parser_se(adapter, session, snapshot):
...
-            "datetime": element["datetime"].isoformat(),
-            "price": element["price"],
-            "currency": element["currency"],
-            "source": element["source"],
-            "zoneKey": element["zoneKey"],
-            "sourceType": element["sourceType"].value,
-        }
-        for element in price
-    ]
-
```
consideRatio, line 59: This PR is mostly changes like these, but included a few other detail changes in files already using snapshots where I transitioned to file-consistent use of snapshots when it made sense.

parsers/test/__snapshots__/test_BG.ambr
```diff
@@ -4,7 +4,7 @@
     dict({
       'correctedModes': list([
       ]),
-      'datetime': '2024-01-01T12:00:00+00:00',
+      'datetime': FakeDatetime(2024, 1, 1, 12, 0, tzinfo=datetime.timezone.utc),
```
VIKTORVAV99, line 7: interesting...

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for doing this!

The tests look a lot cleaner now!

As for the changes to the ENTSO-E parser, those make sense and I'm pretty sure they are covered by the unit tests for the shared parser functions now but let's keep the snapshot tests as that is still parser specific.

PR #7683, opened 2025-01-05
perf: Memoize to optimise sliding

corradio:
## Issue

I've always been puzzled why our slider lags whereas the one of the [debug ENTSOE app](https://debug-entsoe-tp-app-frontend-jfnx5klx2a-ew.a.run.app/), using the old Electricity Maps frontend, is very snappy.
I've also wanted to understand a bit better the React profiler and brush up my skills, and was further inspired by the work @VIKTORVAV99 has recently done, so here's a contribution!
(fyi @jbdietrich, @madsnedergaard and @cadeban)

## The Chrome profiler
One can use the built-in Chrome profiler. I recorded some samples while dragging the slider back and forth (after having selected a country so the side panel is visible). It's quite clear that the AreaGraph being re-rendered takes too much CPU (see bottom right).
Screenshot 2025-01-05 at 10 13 39

## React DevTools
It's hard debug further, so I installed the [React DevTools](https://react.dev/learn/react-developer-tools). Then I fired up the new Profiler tab.

Screenshot 2025-01-05 at 10 19 15
Screenshot 2025-01-05 at 10 19 33

We see that the component is memoized but re-rendered as its props change. We even can see which props cause the re-render.

## Conclusion
All the AreaGraphs are completely re-computed and re-rendered each time we change the datetime value (i.e. slide). Same goes for the TimeController.

## After the changes
The ZoneDetails now only causes an update of the part of the AreaGraph that depends on the current time being changed (that's the long vertical green lines we see on the flame graph for each of the AreaGraphs). The rest remains memo'ized.

image

The TimeController's header is not re-rendered (only the slider is):
image

The ExchangeLayer (the arrows changing) now takes most of the render time

image

## Performance
Performance before/after is best tested in a production build, which can be achieved using `pnpm build && pnpm preview`. Compared to the one running in prod, I see noticeable differences!

### Double check


web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -58,6 +64,52 @@ export default function ZoneDetails(): JSX.Element {
 
   useScrollHashIntoView(isLoading);
 
+  const datetimes = useMemo(
+    () => Object.keys(data?.zoneStates || {})?.map((key) => new Date(key)),
+    [data]
+  );
```
corradio, line 70: This was recomputed each time, causing re-render of child components depending on it

web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -105,29 +149,7 @@ export default function ZoneDetails(): JSX.Element {
...
-
-          <MethodologyCard />
-          <HorizontalDivider />
-          <div className="flex items-center justify-between gap-2">
-            <div className="text-sm font-semibold">{t('country-panel.forecastCta')}</div>
-            <ApiButton size="sm" />
-          </div>
-          <Attribution zoneId={zoneId} />
-        </ZoneDetailsContent>
+        {zoneDetailsContent}
```
corradio, line 152: Moving this component into a `useMemo` requires moving it before any `return` statement, which required me to move some declarations further up the file.

web/src/features/time/TimeController.tsx
```diff
@@ -104,25 +104,31 @@ export default function TimeController({ className }: { className?: string }) {
     (timeRange: TimeRange) => {
       // Set time slider to latest value before switching aggregate to avoid flickering
       setSelectedDatetime({
-        datetime: selectedDatetime.datetime,
+        datetime: datetimes.at(-1),
```
corradio, line 107: The `onToggleGroupClick` was changed each time we changed the current time, causing a full re-render of the TimeController child elements. Given the comment above, it seems like the this change should give us the independence from currently selected time, and not hinder the UX.
In the end I'm not sure why changing time range should depend on the currently selected datetime.

web/src/utils/helpers.ts
```diff
@@ -81,35 +82,47 @@ export function useNavigateWithParameters() {
...
+      basePath,
+      isZoneRoute,
+      location.hash,
+      location.search,
+      navigator,
+      previousDatetime,
+      previousTimeRange,
+      previousZoneId,
+    ]
+  );
```
corradio, line 125: Wrapping this into a `useCallback` is a best practice:
According to [this](https://react.dev/reference/react/useCallback#optimizing-a-custom-hook):
> If you’re writing a [custom Hook,](https://react.dev/learn/reusing-logic-with-custom-hooks) it’s recommended to wrap any functions that it returns into useCallback

web/src/utils/state/atoms.ts
```diff
@@ -30,10 +30,13 @@ export function useTimeRangeSync() {
...
-    setTimeRange(newTimeRange);
-    navigateWithParameters({ timeRange: newTimeRange });
-  };
+  const setTimeRangeAndNavigate = useCallback(
+    (newTimeRange: TimeRange) => {
+      setTimeRange(newTimeRange);
+      navigateWithParameters({ timeRange: newTimeRange });
+    },
+    [setTimeRange, navigateWithParameters]
+  );
```
corradio, line 39: Same as above.

web/src/features/time/TimeController.tsx
```diff
@@ -104,25 +104,31 @@ export default function TimeController({ className }: { className?: string }) {
...
         index: numberOfEntries,
       });
       setTimeRange(timeRange);
       trackEvent(TrackEvent.TIME_AGGREGATE_BUTTON_CLICKED, { timeRange });
     },
-    [setSelectedDatetime, selectedDatetime.datetime, numberOfEntries, setTimeRange]
+    [setSelectedDatetime, datetimes, numberOfEntries, setTimeRange]
```
corradio, line 113: Note that `setTimeRange` was changing each render, see changes in its method body further below.

VIKTORVAV99:
WoHoo!

Personally I've been using https://github.com/aidenybai/react-scan to get profiling and change detection. Which is a bit simpler.

As you notice there is a lot of cascading changes required for all of this memoization which is probably why it's taken so long.

I found some eslint rules we can enable to prevent regressions in the future though!

Looks like there are some type errors to sort out before it's ready to be merged but let me know if you need any assistance with those!

corradio:
Thanks! Seems like we're good to go on the review. I urge you to look at the staging deployment and compare with the prod version. I see quite the difference!

VIKTORVAV99:
> Thanks! Seems like we're good to go on the review. I urge you to look at the staging deployment and compare with the prod version. I see quite the difference!

Will do! And give it a review as well! In general I think there is a lot things we can improve in the code structure wise like splitting it up in more files so it's easier to understand what does what.

VIKTORVAV99:
@corradio since this is relevant to the slider do you know if anyone has ever used these analytics?

image

We send one each time the slider changes and well that can't help with performance and I noticed it often gets blocked due to rate limits anyhow.

Do you think I can remove this?

web/src/features/time/TimeController.tsx
```diff
@@ -102,27 +102,38 @@ export default function TimeController({ className }: { className?: string }) {
...
+
+  // This is required to avoid rerending the toggle needlessly
+  const timeRangeToggle = useMemo(
+    () => (
+      <TimeRangeToggle
+        timeRange={selectedTimeRange || TimeRange.H24}
+        onToggleGroupClick={onToggleGroupClick}
+      />
+    ),
+    [selectedTimeRange, onToggleGroupClick]
```
VIKTORVAV99, line 129: Same here as with the other comment, I think it would be best to memo the component itself and not use a useMemo hook with it.

VIKTORVAV99 (commented):
Two comments on implementation but it looks like it's working as expected functionality wise!

corradio:
> @corradio since this is relevant to the slider do you know if anyone has ever used these analytics?
> 
> image (it's the index position of the slider on each time range)
> We send one each time the slider changes and well that can't help with performance and I noticed it often gets blocked due to rate limits anyhow.
> 
> Do you think I can remove this?

oh wow, I would definitely vote to remove this. I don't really see how the index is useful information.

web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -58,6 +64,54 @@ export default function ZoneDetails(): JSX.Element {
...
+      isError,
+      zoneDataStatus,
+      hasEstimationPill,
+      datetimes,
+      timeRange,
+      displayByEmissions,
+      zoneId,
+      t,
+    ]
+  );
```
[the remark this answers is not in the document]
corradio: I just tried but unfortunately it doesn't work because the component is caused to re-render because its children have changed (not its props!).

image

Therefore, we need to memoize the computation of the children as well, i.e. this part:

image

web/src/features/time/TimeController.tsx
corradio: Indeed this works. Will change

VIKTORVAV99 (approved):
Let's go! 🎉 🚀

I'll drop the analytics when I change the horizons!

tonypls:
Nice! Much snappier!

PR #7685, opened 2025-01-05
fix(web): don't allow persistant production mode

tonypls:
## Issue
We sometime receive negative feedback that the exchange arrows have been removed.

## Description

This PR removes the storage of the consumption / production mode state, when users now reload this will be reset to consumption.


VIKTORVAV99 (approved):
I suppose this makes sense but I liked that it was persistent, but I guess an extra button press would be fine.


LGTM! 🎉

PR #7686, opened 2025-01-05
fix(parsers): use_proxy decorator to not override default args

consideRatio:
## Description

When `WEBPROXY_USERNAME` / `WEBPROXY_PASSWORD` is configured and a `use_proxy` decorated parser function was called without args/kwargs, like done when using `python -m parsers.KPX` instead of `poetry run test-parsers KR`, then the following error showed up:

```
Error(s) creating production breakdown Event 2025-01-06 07:30:00+09:00: 1 validation error for ProductionBreakdown
zoneKey
  none is not an allowed value (type=type_error.none.not_allowed)
```

It was because the `use_proxy` logic always passed a `zone_key` args explicitly, even if it wasn't received, which made a `None` value get passed explicitly and short-circuiting the `fetch_...` functions default value.

## Changes

- Makes calling `fetch_production()` without args or kwargs not run into issues stemming from `zone_key` being explicitly passed as None.
- Adds a comment on what webshare.io subscription is useful for use by this project.
- Re-configures session.proxies back to what it was in a way that I think now will work, while previously I think we just stored a reference to the proxies object that we then also modified.

  This likely won't impact anyone.
- Adjusted logging level from Error to Warning. As it kept running without further complaints it felt more like a warning, and it could reasonably succeed without a proxy depending on where the code is running.
- Adjusted the code for readability, which is an opinionated matter of course.

### Double check

  Things still work, with and without having `WEBSHARE_...` environment configured.

### DIFF
--- a/parsers/lib/config.py
+++ b/parsers/lib/config.py
@@ -62,6 +62,11 @@ def use_proxy(country_code: str):
     and other proxy methods (like cloud run datacenter proxies) do not work. Note that this
     proxy service is not free and is charged per GB.
 
+    If you create an account with webshare.io to develop parsers for this
+    project, you should subscribe to the "residential" proxy type (not "static
+    residential"), as this choice includes most countries. As a practical
+    example, South Korea proxies are only available using this subscription.
+
     Args:
         country_code (str): The ISO 3166-1 alpha-2 code of the country for which the proxy should be used.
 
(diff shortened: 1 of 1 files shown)


consideRatio:
I'm not sure what the test failure is about, I saw it also in #7687.

consideRatio:
Force pushed a rebase on master, hoping to see tests pass now.

consideRatio:
Merged as part of #7741 that was based on top of this!

PR #7688, opened 2025-01-06
chore(ci system): colorize pytest output and summarize slowest tests

consideRatio:
## Description

- configures `pytest` to colorize output, even in when run via GitHub Actions as it otherwise wouldn't.
- configures `pytest` to provide a summary of the ten slowest tests

### Colorized output

Colorized output makes the text quite a bit more readable. Here is a failure without color:

Before|After
-|-
image|image

### Slowest tests summarized

image


consideRatio:
Test failures seems unrelated, I opened #7689

VIKTORVAV99 (approved):
LGTM! 🎉

I do like the color!
Weird that it's not on by default TBH!

PR #7692, opened 2025-01-06
fix(ci system): install poetry after picking python version

### LINKED ISSUE #7689: Tests failing in master via CI system
image

I'm not sure, but I saw this in three separate recent PRs doing different things, so I figure its unrelated to the PRs and something that has showed up recently due to a change in external dependencies or similar.

consideRatio:
## Issue

- fixes #7689 

## Description

- Poetry should probably be installed after picking the Python version
- If poetry.lock has been changed, we shouldn't re-use a cached .venv folder, so the restore-keys are removed.


consideRatio:
Closed in favor of #7693 for now.

PR #7693, opened 2025-01-06
fix: pin poetry to v1 pending migration to v2

### LINKED ISSUE #7689: Tests failing in master via CI system

consideRatio:
## Issue

- Fixes #7689 

## Description

Poetry v2 was released yesterday, and it broke the CI system among other things. To use poetry v2, there is quite a bit of changes needed. Due to that, I think the best path forward short term is to pin poetry to v1 and then work the migration to v2 without time pressure.

### Double check

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -497,7 +497,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Install poetry
         run: |
-          pipx install poetry
+          pipx install "poetry==1.*"
       - name: Get poetry version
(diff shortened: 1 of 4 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

I'll see what we can do about upgrading poetry internally as well, but that might take some time have already scoped up the next weeks priorities.

PR #7696, opened 2025-01-06
fix(parsers, MX): drop seldom occuring datapoints with an hour set to 25

consideRatio:
### Double check

### DIFF
--- a/parsers/CENACE.py
+++ b/parsers/CENACE.py
@@ -142,9 +142,18 @@ def fetch_csv_for_date(dt, session: Session | None = None):
 
     # cleanup and parse the data
     df.columns = df.columns.str.strip()
+
+    # transform 01-24 entries where 24 means 00 the next day
     df["Hora"] = df["Hora"].apply(lambda x: "00" if int(x) == 24 else f"{int(x):02d}")
     df["Dia"] = pd.to_datetime(df["Dia"], format="%d/%m/%Y")
     df.loc[df["Hora"] == "00", "Dia"] = df["Dia"] + pd.Timedelta(days=1)
+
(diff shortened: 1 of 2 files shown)


parsers/CENACE.py
```diff
@@ -142,9 +142,18 @@ def fetch_csv_for_date(dt, session: Session | None = None):
...
+
+    # transform 01-24 entries where 24 means 00 the next day
     df["Hora"] = df["Hora"].apply(lambda x: "00" if int(x) == 24 else f"{int(x):02d}")
     df["Dia"] = pd.to_datetime(df["Dia"], format="%d/%m/%Y")
     df.loc[df["Hora"] == "00", "Dia"] = df["Dia"] + pd.Timedelta(days=1)
+
+    # The hour column has been seen at least once (3rd Nov 2024) to include 1-25
+    # hours rather than the expected 1-24, due to this, we are for now dropping
+    # such entries if they show up
+    df = df.drop(df[df["Hora"] == "25"].index)
```
VIKTORVAV99, line 154: How does this even happen...

Is it possible this is related to daylight savings?

consideRatio: I considered that as well, but this was 3rd Nov, and I don't think it would change in November - also, and it seems MX hasn't had DST since 2022 or something like that.

VIKTORVAV99: Then IDK what it can be but since it's not every day we should probably just get rid of that datapoint.

I don't even know where it would end up as our database is strictly 24 hours and not 25 😅

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7697, opened 2025-01-06
Vik/horizons update

VIKTORVAV99:
## Issue

Closes: AVO-696

### Double check


web/src/components/TimeSlider.tsx
```diff
@@ -74,12 +72,6 @@ export const getThumbIcon = (
   );
 };
 
-function trackTimeSliderEvent(selectedIndex: number, timeRange: TimeRange) {
-  trackEvent(TrackEvent.TIME_SLIDER_BUTTON, {
-    selectedIndex: `${timeRange}: ${selectedIndex}`,
-  });
-}
```
VIKTORVAV99, line 81: We don't really use this and it's sending way to many requests when you move the slider to the point where it even gets rate limited!

VIKTORVAV99:
@cadeban @tonypls @madsnedergaard would love a review here!

Logic wise it shouldn't be very big changes but I did rename a few things and removed the 24 hour view so there is a lot of changed files, sorry about that!

Would really appreciate if we can fast track this one!

web/src/features/map/MapTooltip.tsx
```diff
@@ -112,7 +112,7 @@ export default function MapTooltip() {
   const { zoneId } = hoveredZone;
 
   const { x, y } = mousePosition;
-  const zoneData = data?.data?.datetimes[selectedDatetimeString]?.z[zoneId];
+  const zoneData = data?.datetimes[selectedDatetimeString]?.z[zoneId];
```
madsnedergaard, line 115: Nice to get rid of the data-data part :D

web/src/main.tsx
```diff
@@ -213,29 +216,29 @@ const router = createBrowserRouter([
       },
       {
         path: '/map',
-        element: <Navigate to="/map/72h" replace />,
+        element: <Navigate to="/map/hourly/72h" replace />,
```
madsnedergaard, line 219: Why do we need the `72h` in the URL if it's always that? 🤔

web/src/components/TimeRangeToggle.tsx
```diff
@@ -49,51 +45,27 @@ function TimeRangeToggle({ timeRange, onToggleGroupClick }: TimeRangeToggleProps
...
     >
-      {options.map(({ value, label, dataTestId }) =>
-        value === TimeRange.H72 ? (
-          <NewFeaturePopover
-            side="top"
-            content={<NewFeaturePopoverContent />}
-            isOpenByDefault={isNewFeaturePopoverEnabled}
-            key="popover"
-          >
+      {options.map(
```
madsnedergaard, line 48: Why do we still label the buttons `72h` and `3mo`? Sounds very "internal" and like we forgot to add a translation, especially now that there's space for longer labels 🤔

madsnedergaard (commented):
Two comments, haven't tested or reviewed the code very thoroughly, but otherwise it looks good. Would like to make sure we reconsider the labels before approving though :)

web/src/api/helpers.ts
```diff
@@ -98,6 +98,7 @@ export function isValidDate(dateString: string) {
 export const TIME_RANGE_TO_TIME_AVERAGE: Record<TimeRange, string> = {
   [TimeRange.H72]: 'hourly',
   [TimeRange.M3]: 'daily',
+  [TimeRange.M12]: 'monthly',
```
madsnedergaard, line 101: 😅 
Screenshot 2025-01-09 at 16 42 21

web/src/main.tsx
VIKTORVAV99: Oli wants both in there, so both are in there...

web/src/api/helpers.ts
VIKTORVAV99: Kind of by design as I need to re-adjust the app backend again, PR incoming!

web/src/components/TimeRangeToggle.tsx
madsnedergaard: Is there a discussion somewhere I can read? :)

madsnedergaard: Is there a discussion somewhere I can read? :)

madsnedergaard: Cc @corradio, would love to understand the reasoning behind this :)

corradio: We're talking about a few different things here, so let me attempt to summarise.
I agree the labels can be optimised. Nonetheless, this was out of scope for this work (which has already been more delayed than we'd want). The intent here is to avoid scope creep, so finding better labels is out of scope.
Second, @VIKTORVAV99 suggested renaming to daily/weekly/yearly etc.., which we won't do at this stage because it changes the intent of the button from a range to a resolution. It also won't work as we have two views that are monthly (one that spans for a year, and one that spans all years).

Nonetheless, nothing should prevent anyone from making suggestions on better labels, as long as they don't change the concept of the buttons away from ranges (as as long as they happen *after* this is merged!). Anticipating a further question: the idea of changing the concept of buttons from ranges to resolutions is one that requires more thought because we used to have it before, so I'd like to avoid the back and forth without a deeper investigation and possibly with user-testing.

Hope this makes sense, else, ping me on slack.

web/src/main.tsx
corradio: I think @madsnedergaard is asking about the `72h`, not the fact that `hourly` is present?

VIKTORVAV99: I think he means why do we need both, but the reason is that if we need to supports but monthly all and yearly all without saying monthly all and yearly all in the url we need another limiter, this really increases the complexity for the URL syncing though so I’m honestly not sure it’s worth it for just clean URLs, but it’s done now.

before we could just read the url parameter and check if it’s valid, now we need to read both of them and do a bunch of conditional logic as well.

corradio: If that's the question then indeed we need both resolution and range in the URL because now the same range can have multiple resolutions (and vice-versa). The URL encoding could have been made by concatenating both (e.g. `monthly_all`), but I believe we benefit in the long-run from a cleaner encoding that clearly separates the two parameters, now that one can't be inferred from the other.

web/src/api/helpers.ts
VIKTORVAV99: Backend PR is deploying now so this should be fixed momentarily.

VIKTORVAV99:
Added redirects for 24h to 72h and 30d to 3mo in the last commit but I have not found a easy way to support it with all so I think that is best to handle externally in Cloudflare for now.

web/src/main.tsx
madsnedergaard: Makes sense, it just seemed silly with `/hourly/72h` if hourly is only 72h, but I talked with Viktor and see now the future path and why it makes sense :)

web/src/components/TimeRangeToggle.tsx
madsnedergaard: Thanks for the detailed reply, appreciate it :)

madsnedergaard (approved):
LGTM ❤️

VIKTORVAV99:
> LGTM ❤️

Let's hope this don't break everything and I'll follow up with the cloudflare redirects!

madsnedergaard:
We scoped out getting the yearly averages back on the app, but I just want to highlight here that we need to prioritise that at some point soon as we are already getting questions about it :)

VIKTORVAV99:
> We scoped out getting the yearly averages back on the app, but is there any plan on how/when to add it back (via a new toggle?) I just want to highlight here that we need to prioritise that at some point soon as we are already getting questions about it :)

They are technically still there, just hidden so as soon as there is a new way of selecting the aggregate AND resolution they can be added back.

VIKTORVAV99:
> We scoped out getting the yearly averages back on the app, but is there any plan on how/when to add it back (via a new toggle?) I just want to highlight here that we need to prioritise that at some point soon as we are already getting questions about it :)

Could you open a linear issue for it if there isn't one already?

PR #7699, opened 2025-01-07
fix(web): don't allow persistant country mode

tonypls:
## Issue

Users may get stuck in the country mode without know how to change

## Description

This PR removes the local storage for the country / zone selection.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7700, opened 2025-01-07
add command for generating bounding boxes

FelixDQ:
This command wasn't defined in the package.json. This PR adds the command to generate the zone bounding box:

Example
```
pnpm generate-bbox CF
```
Adds the bounding box to the `CF.yaml`

FYI @PaulRoms


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7703, opened 2025-01-08
feat(web): link data portal from app

tonypls:
## Issue
We don't have enough visibility of the data portal in the app

## Description
This PR adds download button to the zone header which links to the appropriate page of the data portal

## Preview
image

## TODO
Only show button when zone has data in the data portal
Finalize design and button placement


web/src/components/buttons/DownloadCsv.tsx
```diff
@@ -0,0 +1,36 @@
...
+function getCountryToDownload(zoneId?: string) {
+  if (!zoneId) {
+    return '';
+  }
+
+  const zoneName = getZoneName(zoneId);
+  const zoneCountryName =
+    zoneId && zoneId.includes('-') && getCountryName(zoneId)?.toLowerCase();
+  return zoneCountryName || zoneName;
+}
```
VIKTORVAV99, line 17: This don't really work as it gets the translated name of the country. So it breaks for everything except English.

web/src/components/buttons/DownloadCsv.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const url = `https://www.electricitymaps.com/data-portal/${countryToDownload.toLowerCase()}?utm_source=app&utm_medium=download_button&utm_campaign=csv_download`;
+
+  return (
+    <Button
+      href={url}
+      icon={<Download size={DEFAULT_ICON_SIZE} />}
+      type="tertiary"
+      size="md"
+      aria-label="Download CSV Data"
+    />
```
VIKTORVAV99, line 32: Can we mark that this is an external link just like we do in the header the desktop view?

web/src/components/buttons/DownloadCsv.tsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+}
+
+export default DownloadCsv;
```
VIKTORVAV99, line 36: Let's Memo it just in case.

VIKTORVAV99 (changes requested):
We need the fix the button in other languages before it can be merged.

web/src/components/buttons/DownloadCsv.tsx
tonypls: In line with trying to unify the journey I think this should feel as "one platform" as possible, I feel the external link will feel disconnected from the unified experience

web/src/components/buttons/DownloadCsv.tsx
tonypls: Resolved by changing to zoneIds

web/src/components/buttons/DownloadCsv.tsx
VIKTORVAV99: I get that but this will also be available on mobile which could lead to unexpected navigations.

I don't think that should be a blocker though.

VIKTORVAV99: Looks like the new portal does the same thing: image

VIKTORVAV99 (approved):
LGTM! 🎉

But please do consider marking it as external in some way. I don't think that is clear in the moment and might confuse some users.

VIKTORVAV99:
@tonypls why are we moving it into the dropdown?

Then the link won't be very visible and tink most people will miss that we added it?

It also won't be discoverable by search engines so there will be no bi-directional links created for SEO purposes.

tonypls:
> @tonypls why are we moving it into the dropdown?
> 
> Then the link won't be very visible and tink most people will miss that we added it?
> 
> It also won't be discoverable by search engines so there will be no bi-directional links created for SEO purposes.

We're trying to repurpose the more options drop down to handle this, it might be less discoverable but we can improve this in the future if we think it's important.

VIKTORVAV99 (approved):
LGTM! 🎉

tonypls:
Hopefully one last time 😄 

Should look like this on mobile:

image

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7709, opened 2025-01-08
refactor(parsers, ENTSOE): comment document/process types systematically

consideRatio:
Adds comments systematically of `documentType` and `processType` values sent to the ENTSO-E API, based on "Definition" and "Description", as looked up in 
https://eepublicdownloads.entsoe.eu/clean-documents/EDI/Library/Core/entso-e-code-list-v36r0.pdf.

Comments like these became useful while debugging #5318 when I had to match whats fetched via the API and what was looked up manually using the ENTSO-E website.

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -503,7 +503,12 @@ def query_consumption(
     domain: str, session: Session, target_datetime: datetime | None = None
 ) -> str | None:
     params = {
+        # System total load - Total load', including losses without power used
+        # for energy storage, is equal to generation deducted with exports,
+        # added with imports and deducted with power used for energy storage.
         "documentType": "A65",
+        # Realised - The process for the treatment of realised data as opposed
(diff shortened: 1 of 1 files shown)


parsers/ENTSOE.py
```diff
@@ -623,8 +642,11 @@ def query_generation_forecast(
 
     # Note: this does not give a breakdown of the production
     params = {
-        "documentType": "A71",  # Generation Forecast
-        "processType": "A01",  # Realised
```
consideRatio, line 627: Incorrect comment here about Realised btw.

VIKTORVAV99 (approved):
LGTM! 🎉

parsers/ENTSOE.py
VIKTORVAV99: I think this was just copied from another process and then modified and the comment slipped by, maybe by me 😅

PR #7710, opened 2025-01-09
Fix mobile paddings

corradio:
## Issue

I got a bit annoyed seeing these paddings on my iPhone SE:
* top padding is too big and screen estate is lost
* no bottom padding on bottom drawer

image

## Description

8px at the top of the country header needs to be removed (but only on mobile)
image

..and 8px additional needs to be removed from the padding of the mobile buttons
image

The result is much more balanced and we won 16px of vertical space:
image

Note the logo has some padding inconsistencies, but that's because the padding is built into the svg :/

The bottom drawer padding is fixed by simply changing the snap points.

image

### Double check


tonypls (approved):
LGTM

PR #7731, opened 2025-01-13
Adding Bounding box for weather-pre-processing

PaulRoms:
## Issue

List of countries here - https://docs.google.com/spreadsheets/d/1tIZ8kB9MhTmsUukCtisjE4wRJjVYJK8J4SdB-DIP4Mc/edit?gid=142795917#gid=142795917

If cell in red -> No bounding box possible because not defined. I will ask Felix how to deal with it. 

### Double check


config/zones/FJ.yaml
```diff
@@ -1,3 +1,8 @@
+bounding_box:
+  - - -180.5
+    - -19.62909541607167
+  - - 180.5
+    - -15.649580957852429
```
PaulRoms, line 5: -180.5 and 180.5. 
This looks off no? @FelixDQ ?

VIKTORVAV99: Ugh this island is right on top of the 180 line so it might need manual adjustment.

FelixDQ: Yes that will cause problems. Let's change it to go from `177.133330` and `179.9`

FelixDQ (approved):

PR #7732, opened 2025-01-13
fix: Fix capacity update script

VIKTORVAV99:
## Issue

This started to fail for some reason and the eval is throwing a type issue.

## Description

Removes the eval as it should not be needed.

### Double check


PaulRoms (approved):
LGTM

PR #7735, opened 2025-01-13
fix/feat(parsers, MN): fix json parsing error and add coal data

consideRatio:
## Issue

For #5906 that should resolve after a while if it works by its own.

## Description

The JSON API used had a new format which broke the parser.

- I made it so that wind data is now collected from another key clearly for the wind data based on seeing a wind power icon used to represent it.
  image
- I ignored the JSON API key `Songino` as it wasn't obvious enough what to do about it, but it could be battery storage perhaps based on the icon, see discussion following https://github.com/electricitymaps/electricitymaps-contrib/issues/5906#issuecomment-2161197001.
- See also highlighted decision below

## Highlighted decision

- I mapped `dts` to `coal`, even without understanding what `dts` refers to.
  - It seems that the previous parser's `unknown` calculation was calculating the same value `dts` is set to
  - It seems that it's a coal, oil, or gas plant `dts` represents based on the icon for it in https://ndc.energy.mn/
  - It seems that Mongolia's electricity production is dominated by coal (https://en.wikipedia.org/wiki/Energy_in_Mongolia).

### Double check

### DIFF
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -15,77 +15,85 @@ from electricitymap.contrib.lib.models.event_lists import (
 from electricitymap.contrib.lib.models.events import ProductionMix
 from parsers.lib.exceptions import ParserException
 
-NDC_GENERATION = "https://disnews.energy.mn/convertt.php"
+NDC_API = "https://disnews.energy.mn/convertt.php"
 TZ = ZoneInfo("Asia/Ulaanbaatar")  # UTC+8
 
-# Query fields to web API fields
-JSON_QUERY_TO_SRC = {
-    "time": "date",
-    "consumptionMW": "syssum",
-    "solarMW": "sumnar",
-    "windMW": "sums",
-    "importMW": "energyimport",  # positive = import
-    "temperatureC": "t",  # current temperature
+# https://ndc.energy.mn/ provides a list with icons and numbers next to them,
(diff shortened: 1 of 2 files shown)


parsers/MN.py
```diff
@@ -15,104 +15,102 @@
...
-
-    # Calculated 'unknown' production from available data (consumption, import, solar, wind, tpp).
-    # 'unknown' consists of 92.8% coal, 5.8% oil and 1.4% hydro as per 2020; sources: IEA and IRENA statistics.
-    query_data["leftoverMW"] = round(
-        query_data["consumptionMW"]
-        - query_data["importMW"]
-        - query_data["solarMW"]
-        - query_data["windMW"],
-        13,
-    )
```
consideRatio, line 106: This calculation maps exactly to the `dts` value, which is presented in the website https://ndc.energy.mn/ as:
image

Due to this, I figure it should be used directly. If it should be coal or unknown is not in any way obvious in my mind - whats clear is that Mongolia is mostly using coal though, and they didn't mind reporting this as some kind of fossil power plant with the icon.

consideRatio:
Force pushed a rebase on master.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7737, opened 2025-01-14
fix(parsers, KW): update regexp and workaround legacy connect issue

consideRatio:
## Issue

The KW consumption parser is down, but this fixes it.

There were two issues:
- A HTML response regexp no longer managed to identify the system load and was updated
- `requests.Session().get()` no longer worked due to an issue, which could be worked around in a way that I assume is fine because I assume no sensitive traffic is being sent when making a request to fetch this data.

### Double check


consideRatio:
(Rebased)

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7740, opened 2025-01-14
Update geo and configs

VIKTORVAV99:
## Issue

Closes: GMM-271

## Description
Updates the geofile with new and merged zones, updated or removes configs, adjusts exchanges and mexico parser.

In order to still collect consumption data for the zones they are archived instead of removed completely, hopefully this means we'll still collect consumption data for mexico in case we find more granular data in the future and have some use for it.

TODO:

- [x] Geo updates
- [x] Add new configs
- [x] Adjust existing configs
- [x] Update exchanges
- [x] Update CENACE exchange parser to sum values for ERCOT

### Double check


VIKTORVAV99:
@PaulRoms this contains the map updates only so far, but if you want to add some zone new zone zone configs feel free to do so. Otherwise I'll do it before the end of the week + adjust the existing ones.

web/geo/world.geojson
VIKTORVAV99, line 1: QGIS is not made for git so don't even try and look at the diff at this file or things might crash.

VIKTORVAV99:
It seems we have a small issue with the maldives where the islands are too small to even render on the map properly. The bigger ones work but how do we want to handle this?

VIKTORVAV99:
This should be ready to merge now after a few more changes than I'd liked. If you want I can split out the mexico parts in it's own PR as that is what is causing most of the trouble.

config/zones/US-HI.yaml
seljaks, line 1: Do we need a BB here?

VIKTORVAV99: Maybe but I was not sure if you wanted to include all the islands in the BB since not all of them have grid infrastructure. But I can easily add it if you think that is fine.

parsers/CENACE.py
```diff
@@ -294,8 +302,9 @@ def fetch_consumption(
     if demand_td is None:
         raise ParserException("CENACE.py", "Could not find demand cell", zone_key)
     demand = float(demand_td.text.replace(",", ""))
```
seljaks, line 304: This is a potential source of future confusion because we're overwriting the timezone value in the yaml file. I feel that checking timezone in ZONES_CONFIG and RETIRED_ZONES_CONFIG and setting after is worth it, even if it is more verbose.

parsers/CENACE.py
```diff
@@ -308,28 +317,32 @@ def fetch_consumption(
 
 
 if __name__ == "__main__":
```
seljaks, line 319: not sure what this long list of prints is doing, but I'm guessing it's irrelevant whether we add the updated exchange fetches e.g. MX->US-CAL-CISO

seljaks (commented):
Looks awesome. I've left some comments that we can resolve individually.
Broader note, I checked the replica and we're still getting mexico exchanges, so let's hope that stays the same 🤞

config/zones/US-HI.yaml
seljaks: Unsure myself. My guess is that something that includes the area between Hawai'i and Kauai covers the vast majority of the grid. Not sure if/how this bb definition impacts other parts of the app.

parsers/CENACE.py
VIKTORVAV99: These are just for de-bugging when running the parser file directly, you'll find them in a lot of the parsers but they are not used by the backend at all.

I just got annoyed with all the type issues 😅

config/zones/US-HI.yaml
VIKTORVAV99: It does not, we don't use the bb at all in the frontend and actually strip it out at build.

seljaks: Then I'm completely fine with leaving it blank, since the BB that defines which weather data to pull is also separate from this.

VIKTORVAV99: > Then I'm completely fine with leaving it blank, since the BB that defines which weather data to pull is also separate from this. Side note: what is the bb key then supposed to do? 😅

Some of the old forecast models use the BoundingBox defined in these configs instead of the internal ones we now have. There is a migration plan in place for those I think and once that is done we can remove them all from contrib.

I really just added the other ones because it was really simple to do and we can copy paste them into the internal one.

parsers/CENACE.py
VIKTORVAV99: Would you consider this is something we need to fix before merging this or is this more of follow up task?

seljaks: I would do it before merging. If you're swamped I can also make the change.

seljaks (approved):

parsers/CENACE.py
VIKTORVAV99: I'll take a look at it later today but either way I'll probably merge this tomorrow morning so it's not blocking any further tasks longer than that.

If that's the case I'll follow up and do it in a new PR.

PaulRoms (approved):
lgtm

PR #7741, opened 2025-01-14
fix/tests(parsers, DO): adopt use_proxy decorator that makes things work again, and add snapshot test

consideRatio:
## Issue

- for #6992 

## Description

- Adopts `use_proxy` as that makes the request to the data source succeed instead of time out
- Adds a snapshot test, which is very relevant now that `use_proxy` is required and people typically won't have a webshare.io account, and it then helps anyone to test code changes against the mocked response
- Based on top of the commit in #7686 to not workaround how the tests were written where the default zone_key parameter value is used instead of being passed explicitly -- which should be fine but isn't until #7686 or equivalent lands.

### Double check


VIKTORVAV99 (approved):
Sorry for the delayed review, had lots to do last week!

But this looks good to me and seems to work as expected!

PR #7744, opened 2025-01-15
show CA-SK data on the map

FelixDQ:
We now have a estimation model running for CA-SK, so we can show the hourly data


PaulRoms (approved):
lgtm

PR #7746, opened 2025-01-15
fix/feat(parsers, BD, BD_NP): fix to handle HTML changes and add BD_NP exchange

consideRatio:
## Issue

- There isn't an github issue open, but it should be one open I think. See #7747 about this.

## Description

- Fixes an error raised by the parser seeing a new data column (Nepal import!)
- Fixes a bug where datetime comparison fails when passing a datetime with a timezone, like done running `poetry run test_parser "BD->NP" --target_datetime="2024-11-14 00:00 +06:00"` for example
- Adds historical test for the BD parser.
- Adds the exchange config and data for BD_NP, where NP is expected to export 40MW regularly between June 15 and November 15. This is based on finding the new "Nepal" column and reading finding [an article](https://kathmandupost.com/money/2024/11/16/nepal-begins-historic-power-export-to-bangladesh-via-india).

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7749, opened 2025-01-16
removing AL parser since it is not working

PaulRoms:
## Issue
Removing AL parser which is not working currently. AL will be a fully estimated country until we get a good parser.


VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
For future reference we opted to use a estimation model for now so we can also flowtrace historical values and for the fact that despite us reaching out the ENTSO-E several times the TSO data has not been added yet.

The TSOs own website also don't offer a API and their CSV downloads are not easily downloaded programatically.

If this changes in the future we will re-evaluate the use of the estimation model.

PR #7754, opened 2025-01-20
fix(VN): remove subZoneNames field from yaml file

seljaks:
fixes test failure in flowtracing


VIKTORVAV99 (approved):
LGTM! 🎉

We should probably make flowtracing more robust so it can handle empty arrays and objects as well though.

PR #7759, opened 2025-01-22
feat(parsers): remove CH production parser

silkeholmebonnen:
## Issue

[GMM-251](https://linear.app/electricitymaps/issue/GMM-251/ch-data-no-longer-supported-by-entsoe)

## Description

ENTSOE has stopped reporting for Switzerland. I cannot find hourly data for CH so I have added it to the general purpose zone development (see [PR](https://github.com/electricitymaps/electricitymaps/pull/6473)). Therefore we need to remove the CH production parser, which is not getting any production data anyways.


PaulRoms (approved):
lgtm

PR #7766, opened 2025-01-25
tidy(web): Move test assertions over to jest style

### LINKED ISSUE #7761: Migrate frontend tests to use Jest test style
## Description
We would like to align our tests and make this repository more ready for bun as it looks very promising.
The first step to do so is to align our current tests to use Jest style syntax only instead of both Chai and Jest syntax.

Basically all instances of:
```ts
expect(result).to.deep.eq()
```
Should be switched to:
```ts
expect(result).toEqual()
```

Both are supported by vitest so there should be no conflicts in doing this step by step.

wegry:
## Issue

Closes #7761.

## Description

This PR should remove the older Chai(?) style assertions in the flavor of `expect(...).to.be.null` in favor of `expect(...).toBeNull()`.

### Preview

There should be no changes to non-test files.

### Double check


VIKTORVAV99:
Thanks for this PR @wegry!

You don't need to worry about the ESLint failure in the code scanning, that can be silenced but the type checking need to pass before we can merge this!

Otherwise this looks really good!

web/src/features/map-controls/MapControls.cy.tsx
VIKTORVAV99, line 1: This is not a vitest test but a cypress component test which is why it's failing.

Could you revert the changes to this file?

web/src/features/modals/SettingsModal.cy.tsx
VIKTORVAV99, line 1: Same here.

web/src/features/map-controls/MapControls.cy.tsx
wegry: Seems reasonable. See f2dd413.

web/src/features/modals/SettingsModal.cy.tsx
wegry: https://github.com/electricitymaps/electricitymaps-contrib/pull/7766#discussion_r1931212190

wegry:
@VIKTORVAV99 Is there anything I can try to get CI to run through? I can't see why Git checkout is failing for the `CI / Setup (pull_request)` job.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7771, opened 2025-01-27
fix: revert MX configs

VIKTORVAV99:
## Description

Reverts the changes to Mexico so we can create a new deployment without breaking other internal tools.

### Double check


seljaks (approved):
Didn't double check the lonlat numbers or if all the exchanges are the same as before. Overall lgtm.

VIKTORVAV99:
> Didn't double check the lonlat numbers or if all the exchanges are the same as before. Overall lgtm.

Should be!

Used a revert of the other PR as a base and cherry picked the files. Only thing that's edited is the world file.

PR #7776, opened 2025-01-29
Uzbekistan update of installed capacities

### LINKED ISSUE #7775: [Data Issue]: Uzbekistan installed capacities and production
### When did this happen?

Current time

### What zones are affected?

UZ

### What is the problem?

On the map it's showing how much electricity is coming from different sources in Uzbekistan. 
The installed capacities of Uzbekistan is empty on [config/zones/UZ.yaml](https://github.com/electricitymaps/electricitymaps-contrib/blob/master/config/zones/UZ.yaml). From where map is getting this information?
[I've added capacities information](https://github.com/electricitymaps/electricitymaps-contrib/commit/5badb77c2d7c530a44991af5f7cd36d7b85e0bce) with official sources. I think someone should approve it to be accepted? After then will it be visible on map?

jafarsaidov:
## Issue
Uzbekistan update of installed capacities

Closes #7775

## Description
Most sources used for Central Asia, including Uzbekistan, are quite out-of-date. I updated capacities based on latest official reports by governmental bodies of Uzbekistan.

### Double check


config/zones/UZ.yaml
```diff
@@ -3,6 +3,23 @@ bounding_box:
     - 36.68830730933749
   - - 73.61853118531187
     - 46.06232808111943
+capacity:
+  solar:
+  - datetime: '2024-08-30'
+    source: 'The Ministry of Energy of the Republic of Uzbekistan'
```
VIKTORVAV99, line 9: We would prefer the domains of these instead so we can create clickable links.

Other than that it looks good as far as I can tell!

jafarsaidov: I have added link for each source as a comment. is it okay now?

VIKTORVAV99: Not exactly, they should be formatted the way they are in FR.yaml for example. The current source is too long to be displayed in the app without overlapping. This is why just the domain is preferable. Or at least a shortening, the way "European Network of Transmission System Operators for Electricity" becomes "ENTSO-E"

It will look very very strange in the app otherwise.

madsnedergaard:
Hey @jafarsaidov, the checks are failing here because the content isn't formatted quite right :)
You can fix that by running this command in your terminal:

```bash
npx prettier@2 --write config/zones
```

VIKTORVAV99 (approved):

PR #7799, opened 2025-02-03
chore(zones): Split up ES-CN-FVLZ

VIKTORVAV99:
## Issue
Closes: GMM-402

### TODO

- [x] Add zone name and translations.
- [x] Add emission factor data.

### Double check


VIKTORVAV99:
@silimotion mind just double checking the emission factors I added? They are based on your notebook but a second set of eyes wouldn't hurt 🙂

silimotion:
Factors seem to be ok.

As you can see Fuerteventura 2022 factor is a bit high but I would assume it can be explained similarly as with Menorca. Or maybe because they bought less-energy-per-CO2 fuel, as that year fuel prices were a bit wacky.

VIKTORVAV99:
> Factors seem to be ok.
> 
> 
> 
> As you can see Fuerteventura 2022 factor is a bit high but I would assume it can be explained similarly as with Menorca. Or maybe because they bought less-energy-per-CO2 fuel, as that year fuel prices were a bit wacky.

Thanks!
And yeah that's what I figured as well so I think it's fine.

PaulRoms:
Quick one @VIKTORVAV99: Does this affect the forecating to split a zone?

seljaks (approved):
LGTM. Didn't go through the entire yaml, bounding boxes look good.

PaulRoms:
@Garsdal @stavnahumr FYI

VIKTORVAV99:
> Quick one @VIKTORVAV99: Does this affect the forecating to split a zone? 

It really shouldn't unless they are forecasting a non priority zone without estimates 😅

I really hope they don't do that.

VIKTORVAV99:
The old zone will also still be available in the API since it's archived so that's not a concern either.

VIKTORVAV99:
This will remove the ES-CN-FVLZ zone from everything but the API and data collection, it does not impact any of the data for the ES zone as they are not aggregated. There is also no neighboring zones topology wise as it's an island with no exchanges.<br><br>It's also not estimated so there is no completeness guarantees either currently.

PaulRoms (approved):
LGTM

PR #7802, opened 2025-02-04
Portugal Solar Capacity Update

Musketeiro:
Solar capacity update according to the most recent REN data.


VIKTORVAV99 (commented):
These numbers look very different from the ones from ENTSO-E, do you have a link to the capacity data you used so I can take a deeper look at it?

Musketeiro:
> These numbers look very different from the ones from ENTSO-E, do you have a link to the capacity data you used so I can take a deeper look at it?

The source is REN, the national grid operator. Scroll down to the table labelled 'Installed Capacity'.
https://datahub.ren.pt/en/electricity/monthly-balance/

ENTSOE is surely using old data. If you check the older values in the PT.yaml file, you'll see that according to REN the installed capacity in July 2024 was already over 3.1 GW.

And if one includes rooftop/self-supply, the installed capacity is already higher than 5.6 GW according to a very recent news report. https://x.com/mpradoexpresso/status/1886574047192080850

PR #7809, opened 2025-02-05
feat(contrib): add config/ folder to module build + change import

Garsdal:
## Issue

The python module contrib did not correctly contain the config/ folder which is needed to load zone .yaml files. 

## Description

* We change the includes to `"config/**/*.yaml"` to have all nested .yaml files needed for config
* We add `{ include = "config", to = "electricitymap/contrib/config" }` to build the config/ folder inside of the required module
* We add a conditional `CONFIG_DIR` based on if `os.environ.get("RUNNING_IN_RAY") == "true"` which means that the python module is running as a self-contained module.
    - This means that we can now leave the original config/ folder location and CONFIG_DIR, to not break any of our existing deployments, while still making sure that our code can be packaged as a module and run on Ray.

### Preview

We can successfully import contrib and print `ZONES_CONFIG` on Ray with these changes:
image

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7816, opened 2025-02-09
LT zone: Update hydro and hydro storage (2025)

Manu1400:
## Description

In LT zone 🇱🇹 :
* Update hydro (2025)
* Update hydro storage (2025)

### check


VIKTORVAV99:
I'm going to close this PR since we usually don't accept wikipedia as a source and it looks like the same plant was counted twice here.

Sorry it took so long for someone to review it the PR!

PR #7825, opened 2025-02-11
chore: Unblock zones

VIKTORVAV99:
## Issue

GMM-386

## Description

Enables a few zones we have disabled historically because of data quality issues.

### Double check


PaulRoms (approved):
LGTM

PR #7832, opened 2025-02-14
Update US_CA.py

alejandranavcas:
## Description

include fetch_wind_solar_forecasts() method

### Double check

### DIFF
--- a/parsers/US_CA.py
+++ b/parsers/US_CA.py
@@ -3,16 +3,21 @@
 from datetime import datetime, timedelta, timezone
 from logging import Logger, getLogger
 from zoneinfo import ZoneInfo
+from typing import Any
 
 import numpy as np
 import pandas
 from requests import Session
+import requests
+import zipfile
+import io
 
 from electricitymap.contrib.lib.models.event_lists import (
     ProductionBreakdownList,
     TotalConsumptionList,
 )
 from electricitymap.contrib.lib.models.events import ProductionMix, StorageMix
+from electricitymap.contrib.lib.models.events import EventSourceType
 from electricitymap.contrib.lib.types import ZoneKey
 from parsers.lib.config import refetch_frequency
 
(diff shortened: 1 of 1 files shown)


parsers/US_CA.py
```diff
@@ -208,16 +216,112 @@ def fetch_exchange(
...
+                "startdatetime": startdatetime,
+                "enddatetime": enddatetime,
+            },
+        },
+    }
+
+    target_url = _generate_oasis_url(oasis_config)
+
+    # Make a request to download the ZIP file and open the ZIP file in memory
+    response = requests.get(target_url)
```
VIKTORVAV99, line 284:
```suggestion
    response = session.get(target_url)
```

parsers/US_CA.py
```diff
@@ -208,16 +216,112 @@ def fetch_exchange(
...
+    all_production_events = (
+        df.copy()
+    )  # all events with a datetime and a production breakdown
+    production_list = ProductionBreakdownList(logger)
+    for _index, event in all_production_events.iterrows():
+        event_datetime = _index
+        production_mix = ProductionMix(solar=event["Solar"], wind=event["Wind"])
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=event_datetime,
```
VIKTORVAV99, line 304: This event_datetime does not have the type `datetime` causing type issues.

parsers/US_CA.py
```diff
@@ -208,16 +216,112 @@ def fetch_exchange(
...
+    df = df.groupby([COL_DATETIME, COL_DATATYPE], as_index=False).sum()
+    df = df.pivot(index=COL_DATETIME, columns=COL_DATATYPE, values="MW")
+
+    all_production_events = (
+        df.copy()
+    )  # all events with a datetime and a production breakdown
+    production_list = ProductionBreakdownList(logger)
+    for _index, event in all_production_events.iterrows():
+        event_datetime = _index
+        production_mix = ProductionMix(solar=event["Solar"], wind=event["Wind"])
```
VIKTORVAV99, line 301: While this works it would be best to use:
`production_mix.add_value("solar", event["Solar], correct_negative_with_zero=True)` to prevent any edge cases where solar would be forecasted as negative, if it was and it used direct assignment it would be set to None instead of 0.

(we should probably change this in the future but yeah)

parsers/US_CA.py
```diff
@@ -208,16 +216,112 @@ def fetch_exchange(
...
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=event_datetime,
+            production=production_mix,
+            source="oasis.caiso.com",  # or BASE_OASIS_URL
```
VIKTORVAV99, line 306: It's the domain that should be used so this is correct.

VIKTORVAV99 (commented):
Works great but got some minor comments!

Would be great if a snapshot test was added here using the zip file as a mock.

alejandranavcas:
Thanks Viktor! I will make those changes to the script and set up a snapshot test :)

parsers/test/test_US_CA.py
```diff
@@ -0,0 +1,34 @@
...
+            ]  # Get the first file in the zip (there is only one file)
+            with z.open(csv_filename) as f:  # Read the CSV file into a pandas DataFrame
+                df = pd.read_csv(f)
+        return df
+
+    # Mock the call to _get_oasis_data with above (with patch)
+    with patch("parsers.US_CA._get_oasis_data", _mock_get_oasis_data):
+        parsed_wind_solar_forecasts = US_CA.fetch_wind_solar_forecasts(
+            zone_key=ZoneKey("US-CAL-CISO"), logger=getLogger("test")
+        )
```
VIKTORVAV99, line 32: Usually we just patch the session adapter to mock the return of the network call but this is fine as well.

VIKTORVAV99 (commented):
Only one tiny thing left to change, you just need to add this:
```yaml
  productionPerModeForecast: US_CA.fetch_wind_solar_forecasts
```
To the parsers object in: `config/zones/US-CAL-CISO.yaml`

And go ahead and add yourself to the contributors array in the same file as well 🙂

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7835, opened 2025-02-16
chore(ENTSOE parser): Temporary re-map DE->SE-SE4 to fetch historical data

VIKTORVAV99:
## Issue

Part of: GMM-415

## Description

Temporary changes the mapping for these zones so we can get the historical data for this exchange.

### Double check


PaulRoms (approved):
LGTM

PR #7840, opened 2025-02-19
Update US_NY.py

alejandranavcas:
## Description

include the fetch_generation_forecast() method

### Double check


VIKTORVAV99:
> Hiii :) Some questions/comments: Should I add a test to this function? And if so, what kind of test? And also, what should I add to the parsers object in config/zones/US-NY-NYIS ?
> 
> Thanks :)

Snapshot tests are enough since you are using the TotalProductionList parser class and it's events that already have full test coverage.

And for the parsers object it should be the parser (name of the file) + . + method. 

So in this case it would be `generationForecast: US_NY.fetch_generation_forecast`

Then after that is added you can test that it works with `poetry run test_parser "US-NY-NYIS" generationForecast`

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7842, opened 2025-02-20
feat(web): update estimation text

silkeholmebonnen:
## Issue

[GMM-529](https://linear.app/electricitymaps/issue/GMM-529/update-the-estimation-description-for-the-3-models)

## Description

This PR updates the text for the estimation models and deletes the ones that we don't use anymore.

### Double check

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -182,39 +182,24 @@
     "link": "Read about our estimation models",
     "outage-details": "See details",
     "ESTIMATED_FORECASTS_HIERARCHY": {
-      "title": "Data is estimated",
-      "pill": "Estimated",
-      "body": "The data for this zone is estimated using a combination of internal and external production forecasts and Time Slicer Averages. This aims to provide a more accurate representation than Time Slicer Average alone but the data may still be subject to inaccuracies. The data will be updated as soon as real-time data becomes available."
+      "title": "Data is preliminary",
+      "pill": "Preliminary",
+      "body": "The data for this hour has not been reported yet and is based on preliminary data."
     },
     "ESTIMATED_TIME_SLICER_AVERAGE": {
       "title": "Data is preliminary",
       "pill": "Preliminary",
-      "body": "The data provider is not reporting data for this hour. The displayed values are estimated based on historical measurements and will be updated once measured data is available. Note that estimated values may not account for real-time weather events."
+      "body": "The data for this hour has not been reported yet and is based on preliminary data."
     },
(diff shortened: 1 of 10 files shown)


web/src/locales/en.json
```diff
@@ -182,40 +182,25 @@
...
-      "body": "Electricity generation data for this zone is incomplete; missing values have been calculated using an estimation model."
-    },
-    "ESTIMATED_RECONSTRUCT_BREAKDOWN": {
-      "title": "Data is always estimated",
-      "pill": "Imprecise",
-      "body": "The published data for this zone only provides the total production. Renewable sources are estimated using live weather data, while the fossil sources are deducted based on historical data for each production source."
+      "pill": "Partially estimated",
+      "body": "The reported data for this zone is incomplete; estimation models are used to estimate missing production modes."
     },
     "estimated_generic_method": {
```
VIKTORVAV99, line 199: What are we using this one for?
Can we delete it now that we are slimming these down?

web/src/locales/en.json
```diff
@@ -182,40 +182,25 @@
     "link": "Read about our estimation models",
     "outage-details": "See details",
     "ESTIMATED_FORECASTS_HIERARCHY": {
-      "title": "Data is estimated",
-      "pill": "Estimated",
-      "body": "The data for this zone is estimated using a combination of internal and external production forecasts and Time Slicer Averages. This aims to provide a more accurate representation than Time Slicer Average alone but the data may still be subject to inaccuracies. The data will be updated as soon as real-time data becomes available."
+      "title": "Data is preliminary",
+      "pill": "Preliminary",
+      "body": "The data for this hour has not been reported yet and is based on preliminary data."
```
VIKTORVAV99, line 187: Since we are changing this one (and TSA) do we need to go over the other translations as well?

VIKTORVAV99 (commented):
Changes look good but just two questions before approving.

web/src/locales/en.json
silkeholmebonnen: I think it's good to have a fallback, in case there is something that we missed

web/src/locales/en.json
silkeholmebonnen: Yes thats probably a good idea. Should we delete them for now and then update them along the way or what do you think?

VIKTORVAV99: Maybe delete them, it will allow you to move faster.

VIKTORVAV99 (approved):
LGTM! 🎉

Issue #7847, opened 2025-02-20
feat: Add support for 5 min Solar and Wind forecast to the Energinet (DK.py) parser


VIKTORVAV99:
Energinet have 5 minute solar and wind forecasts available for DK-DK1 and DK-DK2.<br>In order to increase our data quality we would like to use these instead of the hourly forecasts we get from ENTSO-E currently.<br><br>These changes should be made to the `DK.py` parser and then the config files for DK-DK1 and DK-DK2 at `DK-DK1.yaml` and `DK-DK2.yaml` should be modified to use the new parser.<br>Documentation for the forecasts can be found here: [https://www.energidataservice.dk/tso-electricity/Forecasts_5Min](https://www.energidataservice.dk/tso-electricity/Forecasts_5Min) and it should use the **Forecast Current** value.<br><br>For any solution to be approved and merged it needs to use the parser classes defined in the repository for built in validation and error correction. There should also be a snapshot test added to prevent any future regressions of the code.

PR #7851, opened 2025-02-21
update DK parser to fetch wind and solar forecasts

t-couery:
## Issue

Implement feat: Add support for 5 min Solar and Wind forecast to the Energinet (DK.py) parser #7847

Closes: #7847
Closes: CON-229
## Description

In order to increase the data quality this PR implement the Energinet 5mn forecast instead of the hourly forecasts from ENTSO-E.

### Preview

`poetry run test_parser DK-DK1 productionPerModeForecast`

```
 {'correctedModes': [],
  'datetime': datetime.datetime(2025, 2, 22, 12, 0, tzinfo=datetime.timezone.utc),
  'production': {'solar': 554.799988, 'wind': 1889.0},
  'source': 'energidataservice.dk',
  'sourceType': <EventSourceType.forecasted: 'forecasted'>,
  'storage': {},
  'zoneKey': 'DK-DK1'}
```

`poetry run test_parser DK-DK2 productionPerModeForecast`
```
{'correctedModes': [],
  'datetime': datetime.datetime(2025, 2, 22, 12, 0, tzinfo=datetime.timezone.utc),
  'production': {'solar': 408.320007, 'wind': 751.0},
  'source': 'energidataservice.dk',
  'sourceType': <EventSourceType.forecasted: 'forecasted'>,
  'storage': {},
  'zoneKey': 'DK-DK2'}

```
### Double check

- [X] I have tested my parser changes locally with `poetry run test_parser DK-DK1 productionPerModeForecast"`


parsers/DK.py
```diff
@@ -122,11 +147,50 @@ def fetch_exchange(
...
+
+    forecast = ProductionBreakdownList(logger=logger)
+    for date_time, productionMix in productionMixs.items():
+        forecast.append(
+            zoneKey=zone_key,
+            production=productionMix,
+            datetime=date_time,
+            source=SOURCE,
+            sourceType=EventSourceType.forecasted,
+        )
```
VIKTORVAV99, line 186: I think this whole block can be simplified a little by using https://docs.python.org/3/library/itertools.html#itertools.groupby and group everything by the datetime and then loop over it, that way we don't have to do these dict lookups and such and can do everything in one loop.

Otherwise this looks good and I'll do a full review tomorrow!

t-couery: Yes, of course, I modified the code with itertools.groupby

parsers/DK.py
```diff
@@ -122,11 +148,47 @@ def fetch_exchange(
...
+    zone_key: ZoneKey,
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list:
+    data = fetch_data(zone_key, session, target_datetime, logger, True)
+
+    # Group data by datetime Minutes5UTC
+    data["records"].sort(key=lambda x: x["Minutes5UTC"])
+    grouped_data = groupby(data["records"], key=lambda x: x["Minutes5UTC"])
```
VIKTORVAV99, line 167:
```suggestion
    grouped_data = groupby(data["records"], key=itemgetter("Minutes5UTC"))
```
There should be no need to pre-sort the data as it's sorted by the internal functions.
And using a itemgetter instead of a lambda for simple property selection slightly more efficient as it uses C code internally and don't need to create a function on each iteration (the lambda).

parsers/DK.py
```diff
@@ -122,11 +148,47 @@ def fetch_exchange(
                     tzinfo=timezone.utc
                 ),
                 netFlow=flow(sorted_keys, datapoint),
-                source="energidataservice.dk",
+                source=SOURCE,
```
VIKTORVAV99, line 151: Nice with a constant here instead!

VIKTORVAV99 (approved):
LGTM! 🎉

A tiny suggestion but otherwise I'll merge this either later today or Monday morning.

parsers/DK.py
t-couery: Updated

parsers/DK.py
t-couery: I don't understand this point SOURCE is not the good option in this case ?

VIKTORVAV99: It is! And we prefer the constant over strings scattered all over the code. So this is very good and I just wanted to highlight that!

PR #7853, opened 2025-02-24
Update LT.yaml

pnook:
updated solar and wind capacities

### Double check


config/zones/LT.yaml
```diff
@@ -483,6 +490,8 @@ sources:
     link: https://docs.google.com/spreadsheets/d/1w5DJ7sPen6axIHU8TCVcuzNCjlct4I6JAbhUlw-ZXu8/edit?usp=sharing
   IPCC 2014:
     link: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
+  litgrid:
```
VIKTORVAV99, line 493: These sources are only for the carbon intensity so this is not needed here and it's enough with the domain in the capacity entry.

VIKTORVAV99 (commented):
Thanks for creating the PR, just a tiny thing then we can merge it!

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7854, opened 2025-02-24
fix(web): change map tooltip color for preliminary

silkeholmebonnen:
## Issue

The map tooltip should also be grey for prelimiary
image

## Description

This makes the tooltip grey for preliminary

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7861, opened 2025-02-26
Change name in method in US_NY

alejandranavcas:
## Description

The data fetched is load forecast so the method name should be consumption (not generation or production).

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7862, opened 2025-02-27
wind solar forecast US_NEISO

alejandranavcas:
create wind solar forecast parser
create test for parser
create snapshot (.ambr) from test
upload mocks for test
update config zone US-NE-ISNE

### Double check


parsers/US_NEISO.py
```diff
@@ -225,11 +231,110 @@ def fetch_exchange(
...
+        merged_data["Date"]
+        + " "
+        + (merged_data["Hour Ending"] - 1).astype(str)
+        + ":00",
+        format="%m/%d/%Y %H:%M",
+    )
+    merged_data["datetime_iso"] = merged_data["datetime"].dt.strftime(
+        "%Y-%m-%dT%H:%M:%S"
+    )
+    data_final = merged_data[["datetime_iso", "solar", "wind"]]
```
VIKTORVAV99, line 308: Everything looks good functionality wise but this part has a lot of type issues. If it's possible to fix that would be great otherwise we can merge it without the fixes.

VIKTORVAV99: I mostly meant this:
image

But we can merge it without fixing that.

VIKTORVAV99:
It seems the test is a bit unstable, it passed when I ran it locally but not when run on the CI. I suspect it's something to do with the timezone handling somewhere so that's probably something we should try and fix.

alejandranavcas:
I think now the timezone is fixed and seems all tests passed!
About the type issue, if we could merge like this it would be nice :)

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7867, opened 2025-02-28
Ss/add new fuel types to eia parser

seljaks:
@VIKTORVAV99 think this is the best I can do with fuel-type mapping, will add tests for the storage handling later since that function is a bit confusing.


parsers/EIA.py
```diff
@@ -501,9 +522,9 @@ def fetch_production_mix(
         # As null values can cause problems in the estimation models if there's
```
VIKTORVAV99, line 522: Oh man I thought this was fixed and removed already 👀

Looks like we have some cleanup to do (In another PR)

parsers/EIA.py
```diff
@@ -338,16 +338,32 @@
     "US-MIDW-MISO->US-SE-SOCO": {"US-MIDW-MISO->US-SE-AEC"},
 }
 
+# # IMPORTANT: the order of the fields must match the order of the fields in the database
```
VIKTORVAV99, line 341: Why is this important?

VIKTORVAV99 (commented):
Do you know if there is any zone that is reporting geothermal values?

Would love to test and compare using live data.

parsers/EIA.py
seljaks: it's not, copied from the power breakdown struct. removed

parsers/EIA.py
seljaks: yeah feels like parsers could use a bit of a make over - I'll write up a pitch for it

seljaks:
US-CAL-IID is an example where they used to report other (i.e. unknown for us) and are (as of February) reporting same amount as geothermal, which is throwing off the production amounts. It's what sparked this investigation in the first place - see [GMM-555](https://linear.app/electricitymaps/issue/GMM-555/us-cal-iid-has-100percent-data-validated-with-0-confidence) for details

parsers/EIA.py
VIKTORVAV99: This check is already included in the parser classes as far as I know, but we should have a snapshot test for all production parsers then we can make these changes with more confidence.

seljaks:
@VIKTORVAV99 I'm getting a weird test failure in something completely unrelated -  do you have any insight?

VIKTORVAV99:
Hmm, it don't know why that is happening. It should not have passed CI.<br><br>I can merge it without that test though if we want/need to.

seljaks:
I'm getting the same test failure on master as well. Going to mark it for xfail to get this through

parsers/test/test_EIA.py
```diff
@@ -778,3 +781,36 @@ def test_texas_loses_one_mode(adapter, session):
...
+                "coal": 1,
+                "wind": 1,
+                "hydro": 1,
+                "nuclear": 1,
+                "oil": 1,
+                "unknown": 1,
+                "solar": 1,
+                "geothermal": 1,
+            },
+            "storage": {"battery": 1, "hydro": 1},
```
VIKTORVAV99, line 813: FYI this means it's charging, negative is dischargin. It's backwards, I know. But to change it we need some deeper re-structuring.

VIKTORVAV99 (approved):
LGTM! 🎉

parsers/test/test_EIA.py
seljaks: makes sense, the parser is producing negative numbers. Should I change the test? From what I can tell what will happen if I set the fixture to have a negative value, then all the other production values will become negative aswell.

VIKTORVAV99: No I think it's fine for now, but I think we will have to take a deeper look at this parser, possibly re-build it from the core and then we can add proper tests along the way.<br><br>I not confident enough in our test coverage here to just go about and change things in the future.

seljaks: Agree 100%, I'm writing a pitch for the next week that addresses this. If nothing else there's way too much processing logic happening at the request level here.

VIKTORVAV99: Most of it should already be covered by our own parser classes as well or should be covered by them (like the all None values). So I think we can slim this down a lot!

PR #7887, opened 2025-03-02
fix(locales): correct country and zone names for Saint Martin en.json

t-couery:
## Description

Correct country and zone names for Saint-Martin in en.json
countryName and zoneName were switched and the country was the wrong one

### Preview

```
    "MF": {
      "countryName": "France",
      "zoneName": "Saint-Martin"
    },
```

### Double check


web/src/locales/en.json
```diff
@@ -1356,8 +1356,8 @@
       "zoneName": "Montenegro"
     },
     "MF": {
-      "countryName": "Saint Martin",
-      "zoneName": "French"
+      "countryName": "France",
+      "zoneName": "Saint-Martin"
```
VIKTORVAV99, line 1360: Since this has it's own ISO code it should just be:
```suggestion
      "zoneName": "Saint-Martin"
```

Note however that this zone is currently not added to Electricity Maps, due to it's size and lack of data.

VIKTORVAV99 (changes requested):

VIKTORVAV99:
The CI is failing because the files are not in sync. I'll take care of it but it means we can't merge it just yet.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7889, opened 2025-03-05
Capacity update for EMBER

PaulRoms:
## Issue

Adding capacity for a couple of General Purpose Zone Development and using EMBER API instead of downloading the CSV each time

### Double check


PaulRoms:
Error in the function - 
ex: 'AG' 

If you fill forward 
Datetime | Config_capacity_fforward | EMBER
 '2017-01-01' | 80.0 | 80.0
 '2018-01-01' | 70.0 | 70.0
 '2019-01-01' | 70.0 | 80.0
 '2020-01-01' | 70.0 | 80.0
 '2021-01-01' | 60.0 | 60.0
 '2022-01-01' | 60.0 | 60.0
 '2023-01-01' | 60.0 | 60.0

The error is because in 2019, even if the true value is 80, since it has been seen in the past, it is not longer consider. It is incorrect.

VIKTORVAV99 (approved):
LGTM! 🎉

Didn't double check all the capacities but those that I did looked good!

BastienBigue (approved):

PR #7894, opened 2025-03-06
chore: Clean up configs and language files

VIKTORVAV99:
## Issue

We have some left over stuff that is not used anymore.

## Description

- Moves the IQ-KUR zone to retired zones.
- Removes the IQ->IQ-KUR exchange.
- Removes the now unused Hawaii island zone names. 

### Double check


silkeholmebonnen:
Why are we removing  IQ-KUR?

VIKTORVAV99:
It has already been merged in the app, and it has no data. The reason it's moved to retired zones is to ensure there are no breaking changes in the API but eventually it should be deleted completely.

silkeholmebonnen (approved):
LGTM

PR #7895, opened 2025-03-07
fetch wind and solar forecasts in CA_ON

alejandranavcas:
includes snapshot test
includes mock files for test
includes snapshot from test

### Double check


parsers/CA_ON.py
```diff
@@ -257,9 +258,196 @@ def fetch_exchange(
...
+) -> list[dict[str, Any]]:
+    """Requests the forecasts for wind and solar (in MW) of Canada Ontario zone for 33 days ahead."""
+    session = session or Session()
+
+    if target_datetime is None:
+        target_datetime = datetime.now(TIMEZONE)
+
+    # Extract the file in adequacy folder until last date (33 days later)
+    # In reality they have forecast until 34 days, but depends at what time of the day the url is extracted, the 34th day might not be published yet
+    end_date = target_datetime + timedelta(days=33)
```
VIKTORVAV99, line 411: Do we really need 33 days of hourly forecasts to be fetched every 20 minutes?

It seems a bit excessive when we don't forecast that far ahead.

And it means we need to do 33 http requests which in total takes almost a minute to run.

It also don't seem to work as the solar output is 0 for the later dates and the wind output is static.

parsers/CA_ON.py
```diff
@@ -257,9 +258,196 @@ def fetch_exchange(
...
+    return mixes
+
+
+def read_VGForecastSummary_report(root):
+    """Helper function for fetch_wind_solar_forecasts() that reads VGForecastsSummary report for a given xml root"""
+
+    ns = {"ns0": "http://www.ieso.ca/schema"}
+
+    # Dictionary to store extracted data
+    mixes: defaultdict[datetime, ProductionMix] = defaultdict(ProductionMix)
```
VIKTORVAV99, line 352: Why mixes instead of returning ProductionLists?

It's easier to combine the production lists later on than looping over all the mixes again and the datetimes are already added.

ProductionLists can be combined with a simple `+` operator.

parsers/CA_ON.py
```diff
@@ -257,9 +258,196 @@ def fetch_exchange(
...
+    return consumption_list.to_list()
+
+
+def read_adequacy_report(root):
+    """Helper function for fetch_wind_solar_forecasts() that reads Adequacy report for a given xml root"""
+    NAMESPACE = "{http://www.ieso.ca/schema}"
+    ns = {"ns0": "http://www.ieso.ca/schema"}
+
+    # Dictionary to store extracted data
+    mixes: defaultdict[datetime, ProductionMix] = defaultdict(ProductionMix)
```
VIKTORVAV99, line 322: Same here.

parsers/CA_ON.py
```diff
@@ -257,9 +258,196 @@ def fetch_exchange(
...
+
+        for forecast_ontario_demand in xml.findall(".//ns0:ForecastOntDemand", ns):
+            for demand in forecast_ontario_demand.findall(".//ns0:Demand", ns):
+                delivery_hour = demand.findtext(NAMESPACE + "DeliveryHour")
+                date_object = datetime.strptime(forecast_date.text, "%Y-%m-%d").replace(
+                    tzinfo=TIMEZONE
+                )
+                date_object = date_object.replace(hour=int(delivery_hour) - 1)
+                energy_mw = demand.findtext(NAMESPACE + "EnergyMW")
+                all_consumption_events[date_object] = energy_mw
```
VIKTORVAV99, line 300: This creates a type error as energy_mw can be none.

It should be checked if it's a real value first or the types should be updated.

parsers/CA_ON.py
```diff
@@ -257,9 +258,196 @@ def fetch_exchange(
...
+    logger: Logger = getLogger(__name__),
+) -> list[dict[str, Any]]:
+    """Requests the demand forecast (in MW) of Canada Ontario zone for 33 days ahead hourly."""
+    session = session or Session()
+
+    if target_datetime is None:
+        target_datetime = datetime.now(TIMEZONE)
+
+    # Dictionary to store extracted data
+    all_consumption_events: defaultdict[datetime, float] = defaultdict(float)
```
VIKTORVAV99, line 274: Why is this not creating consumption lists directly and combining them?

parsers/CA_ON.py
alejandranavcas: How should these cases be handled?
So, if energy_mw is None, negative, etc, I just don't include it as a consumption event? (I only include it if it is a positive float value?)

parsers/CA_ON.py
alejandranavcas: I've changed it now to fetch 7 days ahead. 
The solar output is sometimes still 0 and wind is static, but those are the number they report from the reports in IESO

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7899, opened 2025-03-10
price compliance

VIKTORVAV99:
## Issue
Our license does not include these zones.

## Description

Hides the data as we are not allowed to re-distribute it.

### Double check


seljaks (approved):
lgtm, assuming this is just a flag about whether or not we display price in FE. Also assuming the default setting is `true`

VIKTORVAV99:
> lgtm, assuming this is just a flag about whether or not we display price in FE. Also assuming the default setting is `true`

Yep 👍🏼

PR #7900, opened 2025-03-10
chore: Update capacities

VIKTORVAV99:
## Issue

New capacities are available.

## Description

This PR does two things.
1st it updates all the historical capacities to capture retroactive changes.
2nd it includes new capacities that has been released since last it was run.

### Double check

### DIFF
--- a/config/zones/AT.yaml
+++ b/config/zones/AT.yaml
@@ -71,6 +71,12 @@ capacity:
     - datetime: '2020-01-01'
       source: entsoe.eu
       value: 0.0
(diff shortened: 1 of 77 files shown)


VIKTORVAV99:
@PaulRoms if we are adding capacity to the GPZD model how far back should we add it for the measured zones?

PaulRoms:
> @PaulRoms if we are adding capacity to the GPZD model how far back should we add it for the measured zones?

If we can do until 2021 it is best

PR #7901, opened 2025-03-10
Adds new cross-platform sidebar 🌟 

madsnedergaard:
## Description

We are working on bringing the map and our other offerings at Electricity Maps like the API and Datasets closer together. With this PR we are adding a unified sidebar across the platforms that will make it easier to discover what we offer :)

> [!NOTE]
> The dark-mode color scale is not ideal and it differs from https://portal.electricitymaps.com, but I decided to opt for the least "intrusive" design here for now, and then we can revisit the dark-mode greyscales entirely later on

### Preview

412_1x_shots_so
604_1x_shots_so

Screenshot 2025-03-11 at 00 23 54


tonypls (approved):
LGTM

VIKTORVAV99:
Maybe it's out of scope but can we get a bit of rounding here?
image

That corner looks so sharp!

VIKTORVAV99:
On another note will this be added to the portal as well?

There is currently no way to navigate back to the map if you press one of the other buttons...

madsnedergaard:
Yes, it's behind a feature-flag, but will be rolled out before this is merged in.<br>Regarding the sharp corner(s), you're in for a big treat tomorrow 🙈

PR #7903, opened 2025-03-11
feat(metrics/parser-data): Increase production parser delay for AU-NT, AU-WA and ZA

BastienBigue:
## Description
We have frequent alert on those zones because the data source is reporting data with some delay.

We want to avoid noisy alerts of those parsers when it is just the datasource that is delayed and we cannot take action on those alerts. To avoid noisy alerts, we increase the delay threshold with which we trigger alerts for those zones. 

### Double check

### DIFF
--- a/config/zones/AU-NT.yaml
+++ b/config/zones/AU-NT.yaml
@@ -14,7 +14,7 @@ country: AU
 delays:
   consumption: 30
   price: 30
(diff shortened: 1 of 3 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7904, opened 2025-03-11
fix(EIA): hardcode fix for us-cal-iid

seljaks:
@VIKTORVAV99 I tried real hard to make this not feel so ugly, but gave up. Also tried to find a way to add this correction to the correctedModes list so that we'd have some metadata about it, but I had to stop because the code was giving me a headache 😅


VIKTORVAV99 (approved):
If we are rebuilding the parser inte future we can make it nicer then.


LGTM! 🎉

PR #7908, opened 2025-03-13
Design revamp part II: Remove the header and rework visual style

madsnedergaard:
## Description

Part of a bigger 2025 refresh of the app! 🚀 

This PR focuses on:
- removing the header 
- changing the grayscale
- updating visual style with floating translucent containers everywhere


tonypls:
The space under the description here feels a bit wasted, could reduce the lower padding:
image

tonypls (commented):
The contrast of this toggle and the toggle background needs improving both light and dark mode are quite hard to see it's a toggle. 
image

web/src/hooks/colors.ts
```diff
@@ -28,7 +28,7 @@ export const colors: Colors = {
   },
   dark: {
     co2Scale: defaultCo2Scale,
-    oceanColor: '#343D4C',
+    oceanColor: '#0A0A0A',
```
tonypls, line 31: Is this intentional? Quite the change 🌃

tonypls:
> The space under the description here feels a bit wasted, could reduce the lower padding: image

Also the same on the country panel, I think could have a bit less padding at the bottom here:

image

tonypls (approved):
Some small suggestions but LGTM after those!

VIKTORVAV99:
This button no longer works:
image

web/src/hooks/colors.ts
VIKTORVAV99: It's very dark 😅

madsnedergaard: Wdyt @Alportan, is it too far? It's now using `neutral-950` now like on the Portal :)

VIKTORVAV99 (approved):
It's just the very dark map but feel free to merge after talking with Alex about it.

madsnedergaard:
> Spoke to soon, I can't scroll on mobile 😅

Haha, I was already on it! ❤️

madsnedergaard:
> > Spoke to soon, I can't scroll on mobile 😅
> 
> Haha, I was already on it! ❤️ Fixed now

Oh wait, now I lost the timecontroller thingy 😓

madsnedergaard:
Couldn't reproduce the problem after all, but will test again before merging in the full thing :)

web/src/hooks/colors.ts
Alportan: I do prefer the darker side 🌚

PR #7909, opened 2025-03-13
fetch_consumption_forecast and fetch_wind_solar_forecasts

alejandranavcas:
create these parsers
include snapshot test for both
upload mocks
created snapshot
update config file for US-TEX-ERCO

### Double check


parsers/US_ERCOT.py
```diff
@@ -65,6 +72,17 @@
...
+# Report type IDs
+# Wind production forecast: https://www.ercot.com/mp/data-products/data-product-details?id=NP4-732-CD
+WIND_POWER_PRODUCTION_HOURLY_AVERAGED_ACTUAL_AND_FORECASTED_VALUES_RTID = 13028
+
+# Solar production forecast: https://www.ercot.com/mp/data-products/data-product-details?id=np4-737-cd
+SOLAR_POWER_PRODUCTION_HOURLY_AVERAGED_ACTUAL_AND_FORECASTED_VALUES_RTID = 13483
+
+# Load forecast: https://www.ercot.com/mp/data-products/data-product-details?id=np3-560-cd
+LOAD_FORECAST_BY_FORECAST_ZONE = 12311
+
```
VIKTORVAV99, line 84: These should probably be Enums.

And while I like clear names could we make them a tiny bit shorter?

parsers/US_ERCOT.py
```diff
@@ -374,3 +392,140 @@ def fetch_exchange(
...
+    return consumption_list.to_list()
+
+
+def fetch_wind_solar_forecasts(
+    zone_key: ZoneKey = ZoneKey("US-TEX-ERCO"),
+    session: Session | None = None,
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list[dict[str, Any]]:
+    """Requests wind and solar power data in MW. target_datetime not implemented (takes the latest report)"""
```
VIKTORVAV99, line 469: Does it support it at all?

It would be good to have so we can backfill all the data to use more advanced estimation methods.

alejandranavcas:
Thanks for your comments! 
I've now changed the Enums and also have modified the parsers to include the target_datetime

VIKTORVAV99 (approved):
LGTM! 🎉

PR #7916, opened 2025-03-14
chore(parsers): Use enums for Emaps production and storage modes

VIKTORVAV99:
## Issue

We currently have a lot of these strings in the codebase.

## Description

Replaces mode strings with enums to ensure they are consistent between parsers. Also uses the new enums internally instead of a list of the modes to ensure that if we add new modes it's supported everywhere at once.

### Double check


parsers/test/__snapshots__/test_ENTSOE.ambr
```diff
@@ -8373,7 +8373,7 @@
   list([
     dict({
       'correctedModes': list([
-        'gas',
+        <ProductionModes.GAS: 'gas'>,
```
VIKTORVAV99, line 8376: Do you think this will be a issue? if so I can make it so we call the `str()` method on them first to convert them to plain strings

seljaks: I would call string for backwards compatibility reasons, but I'm not sure if/where we use the corrected modes list.
The point about sets is good, but you probably have a better idea how often merging ProductionBreakdowns lists is. If it's not common, then I'd keep it as is

parsers/lib/config.py
```diff
@@ -10,6 +11,53 @@
...
+class BaseModeEnum(str, Enum):
+    """Base class for Mode enums."""
+
+    def __str__(self) -> str:
+        return self.value
+
+    @classmethod
+    def values(cls) -> list[str]:
+        """Return a list of all enum string values."""
+        return [mode.value for mode in cls]
```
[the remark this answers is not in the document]
seljaks: Not sure if relevant, but you can get to the underlying values with Enum.__members__ and then do .keys()/.values() on that.
While dict_keys and dict_items behave like lists, they have optimized implementations in python

electricitymap/contrib/lib/models/events.py
```diff
@@ -202,11 +202,12 @@ def merge(cls, production_mixes: list["ProductionMix"]) -> "ProductionMix":
...
         for production_mix in production_mixes:
-            for mode in set(PRODUCTION_MODES).intersection(
-                production_mix.__fields_set__
-            ):
-                value = getattr(production_mix, mode)
-                merged_production_mix.add_value(mode, value)
+            # Process all set production modes
+            for mode in ProductionModes.values():
+                if mode in production_mix.__fields_set__:
+                    merged_production_mix.add_value(mode, getattr(production_mix, mode))
```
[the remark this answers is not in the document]
seljaks: it's also more readable

electricitymap/contrib/lib/models/events.py
```diff
@@ -153,7 +153,7 @@ def __setattr__(
         and to check for negative values and set them to None.
         This method also keeps track of the modes that have been corrected.
         """
-        if name not in PRODUCTION_MODES:
+        if name not in ProductionModes.values():
```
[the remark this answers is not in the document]
seljaks: I think this is fine from a performance pov, the call to the api is always going to be the slowest thing in a parser.

seljaks (commented):
Looks really nice, left some comments that we can tackle one by one.

parsers/lib/config.py
```diff
@@ -1,15 +1,19 @@
...
 from enum import Enum
 from inspect import signature
 from logging import getLogger
+from typing import TypeVar
 
 import requests
 from requests import Session
 from requests.adapters import HTTPAdapter, Retry
 
+ModeEnumType = TypeVar("ModeEnumType", bound="BaseModeEnum")
```
VIKTORVAV99, line 15: This combined with the type changes below ensure that the right type is returned even when we use derived classes/enum.

parsers/lib/config.py
```diff
@@ -19,19 +23,19 @@ def __str__(self) -> str:
         return self.value
 
     @classmethod
-    def values(cls) -> list[str]:
-        """Return a list of all enum string values."""
-        return [mode.value for mode in cls]
+    def values(cls: type[ModeEnumType]) -> ValuesView[ModeEnumType]:
```
VIKTORVAV99, line 26: Should this be renamed to members?

parsers/lib/tests/__snapshots__/test_config.ambr
VIKTORVAV99, line 1: Adding a snapshot test so it's easy to see the output and so it will catch any future changes.

parsers/test/__snapshots__/test_ENTSOE.ambr
VIKTORVAV99: It's pretty common actually as for example the ENTSOE parser creates one list per mode and then merges the mode specific lists into one final list.

It's very much dependent on the data structure of the parser.

I'll call str() on this to maintain backwards compatibility.

parsers/lib/config.py
seljaks: Tough call, but I think values is better, people are more used to dict and it's methods, than enum functionality. So hiding this and keeping a more dict-like interface is nice

parsers/lib/config.py
seljaks: awesome, practicing type driven development :D

seljaks (approved):
🙌

parsers/test/__snapshots__/test_ENTSOE.ambr
VIKTORVAV99: Turns out the internal stuff already uses set for this so it's only the dict representation that uses a list, IDK why it does that but I don't think it's worth changing now.

Added a call to map to cast these to strings though.

Issue #5089, opened 2023-02-14
Close mobile exchange tooltip when opening another


tonypls:
## Description
As mentioned in #5070

As a mobile user I want the previous exchange tooltip to close when I select a different exchange so that I can clearly see that map

Current state:
![image](https://user-images.githubusercontent.com/18622768/218732130-550a52a5-9ae7-424d-9c45-e6890413c62b.png)

VIKTORVAV99:
@tonypls didn't one of your latest PRs fix this?

tonypls:
Yeah #5195

PR #7919, opened 2025-03-17
Design revamp part III: Aligns tooltips

madsnedergaard:
## Issue

_Part of a bigger 2025 refresh of the app! 🚀_

## Description

This aligns the various tooltips in look and feel.

Also fixes the regression of #5089 that happened at some point.

### Double check


madsnedergaard:
Oy, let me just get #7908 merged first before reviewing this PR!

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -1,15 +1,15 @@
 import * as Tooltip from '@radix-ui/react-tooltip';
```
madsnedergaard, line 1: This component now supports both:
- Hover tooltips like the ones on the sidebar
- Clickable tooltips (exchange arrows) and making sure only one of them is open at a time

VIKTORVAV99:
@madsnedergaard are the graph tooltips supposed to be glassy as well? I don't think they are at the moment.

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -19,18 +19,41 @@ export default function TooltipWrapper({
...
   const isMobile = useIsMobile();
   if (!tooltipContent) {
     return children;
   }
 
   // Helpers
-  const openTooltip = () => setIsOpen(true);
-  const closeTooltip = () => setIsOpen(false);
-  const toggleTooltip = () => setIsOpen(!isOpen);
+  const openTooltip = () => {
```
VIKTORVAV99, line 32: All of these need to be in useCallback hooks to prevent re-renders. Since you are changing it anyway do you mind adding that?

madsnedergaard: Gotcha, will do

web/src/components/tooltips/TooltipWrapper.tsx
```diff
@@ -1,43 +1,64 @@
...
+      setLocalIsOpen((previous) => !previous);
+    }
+  }, [tooltipId, openTooltipId, setOpenTooltipId]);
+
+  // Simple event handlers using memoized functions
+  const handleMouseEnter = () => !isMobile && openTooltip();
+  const handleMouseLeave = () => !isMobile && closeTooltip();
+  const handleClick = () => isMobile && toggleTooltip();
+  const handleContentClick = () => isMobile && closeTooltip();
+  const handleContentPointerDownOutside = () => isMobile && closeTooltip();
```
VIKTORVAV99, line 55: The functions themselves might be memorized but these declarations are not as they are assigned at component creation and don't use useCallback.

They either need to be moved outside the component so they have stable references or use useCallback as well.

madsnedergaard: Ugh, I had them memo'd but then Claude told it wasn't necessary given that the open/close/toggle functions and given that isMobile is not going to change 🥲

cadeban (approved):
+1 to Viktor's review :)

PR #7921, opened 2025-03-17
fix(parsers): fix JP-KN parser

silkeholmebonnen:
## Issue

JP-KN has not received parser data since 12/03. 

## Description

This PR crops the image differently so the OCR can read the text on the image. 

### Double check


seljaks (approved):
lgtm

PR #7923, opened 2025-03-17
Design revamp Part IV: New Search

madsnedergaard:
## Issue

_Last part of a bigger 2025 refresh of the app! 🚀_

## Description

Almost last PR!

- Removes the ranking panel completely
- Replaces it with a new floating search bar that expands into a list of results when in use
- Shows the logo on mobile map
- Removes zones from translation files that are not available on the map (I manually checked each of these 😅)


web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -0,0 +1,42 @@
+import { t } from 'i18next';
+import { GridState } from 'types';
+
+import { ZoneRowType } from './ZoneList';
+
+export const getAllZones = () => {
+  // Get all zone data directly from translations
+  // TODO: Get rid of zones not available on the map!
```
madsnedergaard, line 8: Addressing this in a follow-up PR

cadeban:
Very nice! Couple of visual nits:
1. The top padding of the zone info view looks a little thick for me on desktop: 
Screenshot 2025-03-17 at 18 56 25

2. The bottom padding looks a little tight (though maybe it only seems that way because of the top padding)
Screenshot 2025-03-17 at 18 56 43

cadeban:
Have we considered using the NewFeaturePopover to highlight the new search functionality?

cadeban:
😌 
> Screenshot 2025-03-17 at 19 09 34

web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -0,0 +1,41 @@
...
+  const filtered = Object.entries(zoneData).filter(([zoneKey, zone]) => {
+    const searchLower = searchTerm.toLowerCase();
+    return (
+      zoneKey.toString().toLowerCase().includes(searchLower) ||
+      zone.zoneName?.toLowerCase().includes(searchLower) ||
+      zone.countryName?.toLowerCase().includes(searchLower) ||
+      zone.seoZoneName?.toLowerCase().includes(searchLower) ||
+      zone.displayName?.toLowerCase().includes(searchLower)
+    );
+  });
```
VIKTORVAV99, line 34: Not needed for this PR but would be nice if we could search across all languages somehow.

web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -0,0 +1,41 @@
...
+  >;
+
+  return zoneData;
+};
+
+export const getFilteredList = (
+  searchTerm: string,
+  zoneData: Record<string, ZoneRowType>
+): ZoneRowType[] => {
+  if (!searchTerm) {
```
VIKTORVAV99, line 20: We should probably require the length of the term to be above 1 so we don't match every singe zone that has an a in it for example.

It should not be the same as NoResults though.

web/src/features/panels/search-panel/NoResultsIllustration.tsx
VIKTORVAV99, line 1: Is there no Lucide Icon we can use for this?

Seems like a lot of SVG code for a rarely used icon?

(or at least run it through SVGO to optimize it)

web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -0,0 +1,41 @@
...
+    return [];
+  }
+
+  // Filter zones based on search term
+  const filtered = Object.entries(zoneData).filter(([zoneKey, zone]) => {
+    const searchLower = searchTerm.toLowerCase();
+    return (
+      zoneKey.toString().toLowerCase().includes(searchLower) ||
```
VIKTORVAV99, line 28: why is this not using zoneID (which is the zoneKey) as before?

Would be easier to deal with one object if we want to add fuzzy search using https://github.com/farzher/fuzzysort for example.

madsnedergaard:
> 1. The top padding of the zone info view looks a little thick for me on desktop:

Oh thanks, forgot to push that change! 😅 

> 2. The bottom padding looks a little tight (though maybe it only seems that way because of the top padding)

Good point, fixed.

> Have we considered using the NewFeaturePopover to highlight the new search functionality?

I did not, but I'm also unsure what it should say since functionality is more or less the same :)

> I had such great plans for the ranking panel, it will be missed! 🪦💐

I have wanted to kill it for YEARS, so I'm very stoked about this! :D

web/src/features/panels/search-panel/NoResultsIllustration.tsx
madsnedergaard: Nope... I did optimise it a bit, but don't want to spend more time on saving 1 kb more - I can spend that time adding support instead, which I think is going to be way more valuable for users 😉

web/src/features/panels/search-panel/getSearchData.ts
VIKTORVAV99: FYI that library only seems to support English so that might be a dealbreaker.

madsnedergaard: Requires more work and reprocessing the translations object to add it in. Added a comment though that we can change this if we need it in the future :)

web/src/features/panels/search-panel/ZoneList.tsx
```diff
@@ -1,7 +1,6 @@
 import { useVirtualizer } from '@tanstack/react-virtual';
```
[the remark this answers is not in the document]
madsnedergaard: We do potentially have that (typing just one or two letters can give a lot of results), and it also works so I don't feel a big desire to rework this to save 4kb 😉

web/src/features/panels/search-panel/getSearchData.ts
madsnedergaard: Added ability to search for English names if user has selected a different language:
Screenshot 2025-03-17 at 23 42 02

web/src/features/panels/search-panel/getSearchData.ts
madsnedergaard: It's a fair suggestion, but given that it would increase complexity here (another state the search and results list can be in and that requires matching UI) and doesn't really seem to be a problem as long as we still have virtualised list, I'll postpone this for now :)

web/src/features/panels/search-panel/ZoneList.tsx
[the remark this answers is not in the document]
VIKTORVAV99: Was mostly concerned about the runtime overhead of creating and maintaining the virtual list as it will keep changing every time the search term changes. It's only really useful to keep the dom count low which was important when all zones where in the list by default, now it's kind of the opposite 😅

cadeban:
RE: NewFeaturePopover - True, we probably don't need a new feature popover for the search. But some sort of acknowledgement about the new design could be helpful for visitors who might be expecting/used to the older design

cadeban:
@madsnedergaard I'm going to just quickly test mobile stuff and then approve if it looks good

cadeban:
Hm, I'm not able to scroll the left panel in the browser's mobile view anymore

madsnedergaard:
> Hm, I'm not able to scroll the left panel in the browser's mobile view anymore

Oh, I must have messed that up in a merge - will fix it again, thanks for testing! 🙌

cadeban (approved):
🚀

PR #7926, opened 2025-03-17
feat(zone-configs): add capacity for USA

silkeholmebonnen:
## Description

Adds capacity data for the US

### Double check

### DIFF
--- a/config/zones/AE.yaml
+++ b/config/zones/AE.yaml
@@ -4,79 +4,113 @@ bounding_box:
   - - 56.88363691500007
     - 26.574791972000085
 capacity:
-  gas:
-    - datetime: '2016-01-01'
-      source: Climatescope
-      value: 26394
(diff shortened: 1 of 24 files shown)


PaulRoms (approved):
LGTM

VIKTORVAV99:
Just curious but why do we need this?

We don't run any estimation models on aggregated data and for the app we use the aggregated capacity of all subzones regardless of what's listed here.

silkeholmebonnen:
> Just curious but why do we need this?
> 
> We don't run any estimation models on aggregated data and for the app we use the aggregated capacity of all subzones regardless of what's listed here.

I'm not sure we're using it but I added it because we wanted capacities for US-AK and US-HI, which I could not find on IRENA since they are subzones of the US, so I decided to just add capacities for the US, since they also seemed to be outdated

PR #7927, opened 2025-03-17
Increase delay for CO and MX

BastienBigue:
## Issue
MX and CO report data with delay (resp. unto 1.5 month and 3.5 days). 
Specify those delays in the configuration.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7930, opened 2025-03-17
fix(security): replace usage of tj-actions

jbdietrich:
## Description

https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

### Double check


VIKTORVAV99 (approved):
Let's do this for now and then maybe look into some alternatives that support immutable actions once that's fully rolled out.

PR #7932, opened 2025-03-18
Design makeover 2025 🚀 

madsnedergaard:
## Description

We have given the app a small visual overhaul and removed elements (the "ranking panel" and header) to give more space for exploring what really matters - the map and the data itself. 

At the same time, we have added a new sidebar navigation to help people explore what else we offer at Electricity Maps (like the free datasets or the API).

Lastly, we are moving the FAQ and feedback to https://help.electricitymaps.com/ to unify all this information in a single place. The Help Center will receive more content as we learn what questions users actually have :)


VIKTORVAV99:
If we are removing/moving the FAQ can we also remove all the translations for them?

VIKTORVAV99:
> > If we are removing/moving the FAQ can we also remove all the translations for them?
> 
> It's still available on mobile for now (we will probably remove it once we have a new mobile release going out), so let's keep it around for a bit longer :)
> 
> Also, I tagged you for review, but there should be no unreviewed changes to this PR

I'll take a final look if I have the time but feel free to merge without it if you have other approving reviews. If I have some suggestions/feedback I might add them later and we can follow up in a new PR if needed.

PR #7936, opened 2025-03-19
feat(parsers): add price parser for spp north hub

silkeholmebonnen:
## Issue

[GMM-598](https://linear.app/electricitymaps/issue/GMM-598/add-price-parser-for-spp-north-hub)

## Description

This PR adds a realtime price parser and a dayahead price parser to SPP north hub.

### Preview

To get the latest realtime price
`poetry run test_parser US-CENT-SWPP realtimeLocationalMarginalPrice`

To get the realtime price at a specific datetime
`poetry run test_parser US-CENT-SWPP realtimeLocationalMarginalPrice --target_datetime "2025-03-19 12:30"`

To get the latest day ahead price
`poetry run test_parser US-CENT-SWPP dayaheadLocationalMarginalPrice`

To get the day ahead price for a specific datetime
`poetry run test_parser US-CENT-SWPP dayaheadLocationalMarginalPrice --target_datetime "2025-03-17 05:00"`

You can compare with [Grid Status](https://www.gridstatus.io/graph/lmp?iso=spp&date=2025-03-17), but be aware of 2 things:
- When SPP reports data for e.g. *:30, then we store this as *:30, but Grid Status shows it as *:25
- Grid Status use local time zone and we store in UTC

### Double check


silkeholmebonnen:
I'll make another PR for the database migrations to create tables to store this data in

parsers/US_SPP.py
```diff
@@ -420,6 +430,77 @@ def fetch_exchange(
...
+    target_datetime: datetime | None = None,
+    logger: Logger = getLogger(__name__),
+) -> list[dict[str, Any]]:
+    if target_datetime is None:
+        target_datetime = datetime.now(tz=timezone.utc)
+    if target_datetime.tzinfo is None:
+        target_datetime = target_datetime.replace(tzinfo=timezone.utc)
+
+    closest_5_minutes_datetime = get_closest_5_minutes_datetime(target_datetime)
+    url = f"{REALTIME_PRICE_URL}&path=/{closest_5_minutes_datetime.strftime('%Y')}/{closest_5_minutes_datetime.strftime('%m')}/By_Interval/{closest_5_minutes_datetime.strftime('%d')}/RTBM-LMP-SL-{closest_5_minutes_datetime.strftime('%Y%m%d%H%M')}.csv"
```
jbdietrich, line 446: This is a gnarly format string -- maybe extract to a helper function or join them into a path beforehand for readability.

parsers/lib/parsers.py
```diff
@@ -21,6 +23,8 @@
     "productionPerModeForecast": PRODUCTION_PER_MODE_FORECAST_PARSERS,
     "exchange": EXCHANGE_PARSERS,
     "price": PRICE_PARSERS,
+    "dayaheadPrice": DAYAHEAD_PRICE_PARSERS,
```
jbdietrich, line 26: I think we might need to distinguish LMP from other price "day ahead" is partially covered by our existing "price" parser key.

jbdietrich (approved):

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
...
+            price=row["LMP"],
+            currency="USD",
+            source=SOURCE,
+        )
+
+    price_list = prices.to_list()
+    return price_list
+
+
+def get_closest_5_minutes_datetime(datetime: datetime) -> datetime:
```
VIKTORVAV99, line 499: The variable here is shadowing the datetime built in module, this makes it so you can't use the built in module and might have unexpected results.

I would advice against this and simply rename it to target_datetime or something similar.

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
...
+    cdt_datetime = datetime.astimezone(tz=ZoneInfo("America/Chicago"))
+    rounded_cdt = cdt_datetime.replace(
+        minute=round(cdt_datetime.minute / 5) * 5, second=0, microsecond=0
+    )
+    return rounded_cdt
+
+
+def get_realtime_url(datetime: datetime) -> str:
```
VIKTORVAV99, line 507: Same here.

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
...
+    closest_5_minutes_datetime = get_closest_5_minutes_datetime(datetime)
+    year = closest_5_minutes_datetime.strftime("%Y")
+    month = closest_5_minutes_datetime.strftime("%m")
+    day = closest_5_minutes_datetime.strftime("%d")
+    hour = closest_5_minutes_datetime.strftime("%H")
+    minute = closest_5_minutes_datetime.strftime("%M")
+    return f"{REALTIME_PRICE_URL}&path=/{year}/{month}/By_Interval/{day}/RTBM-LMP-SL-{year}{month}{day}{hour}{minute}.csv"
+
+
+def get_dayahead_url(datetime: datetime) -> str:
```
VIKTORVAV99, line 517: And here.

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
...
+    year = datetime.strftime("%Y")
+    month = datetime.strftime("%m")
+    day = datetime.strftime("%d")
```
VIKTORVAV99, line 520: Would this not work here?

```suggestion
    year = datetime.year
    month = datetime.month
    day = datetime.day
```

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
...
+    return rounded_cdt
+
+
+def get_realtime_url(datetime: datetime) -> str:
+    closest_5_minutes_datetime = get_closest_5_minutes_datetime(datetime)
+    year = closest_5_minutes_datetime.strftime("%Y")
+    month = closest_5_minutes_datetime.strftime("%m")
+    day = closest_5_minutes_datetime.strftime("%d")
+    hour = closest_5_minutes_datetime.strftime("%H")
+    minute = closest_5_minutes_datetime.strftime("%M")
```
VIKTORVAV99, line 513: Same here as the other suggestion.

VIKTORVAV99:
> I'll make another PR for the database migrations to create tables to store this data in

Don't forget to also update feeder electriicty to fetch these new datatypes on a regular basis. I don't think they will be included by default.

parsers/US_SPP.py
```diff
@@ -420,6 +430,97 @@ def fetch_exchange(
     return exchanges
 
 
+@refetch_frequency(timedelta(minutes=30))
```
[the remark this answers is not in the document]
silkeholmebonnen: Yes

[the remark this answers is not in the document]
silkeholmebonnen: Every datapoint is retrieved from a different csv file, so since we are fetching every 20 minutes I thought 30 min was ok

parsers/US_SPP.py
silkeholmebonnen: no because then the month would be formatted "3" instead of "03"

parsers/US_SPP.py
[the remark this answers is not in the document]
VIKTORVAV99: 30 min refetch frequency here though only applies to when we are refetching the data not the normal fetching.

So it don't matter with the 20 min fetching pattern, but if you refetch 1 day of data now it will schedule 48 tasks to refetch 24 hours of data.

In contrast the ENTSOE parser has a refetch frequency of 72 hours meaning it can return 3 days if data with 1 task. That makes it much more efficient.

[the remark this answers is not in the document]
silkeholmebonnen: I know but since this is also the function that will be used when we fetch then it will be a lot slower if we fetch a whole day of data when we in reality only want 20 minutes.

VIKTORVAV99:
Will give this a in depth review later but would love to see some more tests added for this and to ensure that the node values are actual strings and such.

Take inspiration from the other validator functions!

silkeholmebonnen:
@jbdietrich and @VIKTORVAV99 this is ready for review again

electricitymap/contrib/lib/models/events.py
```diff
@@ -837,3 +837,56 @@ def to_dict(self) -> dict[str, Any]:
             "source": self.source,
             "sourceType": self.sourceType,
         }
+
+
+class LMP(Price):
```
VIKTORVAV99, line 842: Can we make the class name more descriptive? Contributors should be able to tell what this means without digging though our documentation.

I suggest NodalPrice or NodePrice to make this a bit more generic, we don't really care that it's the local or marginal price at this level, just that it's for a specific node in a zone.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -492,3 +493,26 @@ def append(
         )
         if event:
             self.events.append(event)
+
+
+class LMPList(EventList):
```
VIKTORVAV99, line 498: Same here.

parsers/lib/parsers.py
```diff
@@ -13,6 +13,8 @@
 GENERATION_FORECAST_PARSERS = {}
 EXCHANGE_FORECAST_PARSERS = {}
 PRODUCTION_CAPACITY_PARSERS = {}
+DAYAHEAD_PRICE_PARSERS = {}
+REALTIME_PRICE_PARSERS = {}
```
VIKTORVAV99, line 17: These should be renamed to be LMP specific since we might add other price parsers in the future.

electricitymap/contrib/lib/models/events.py
```diff
@@ -837,3 +837,56 @@ def to_dict(self) -> dict[str, Any]:
...
+                datetime=datetime,
+                source=source,
+                price=price,
+                currency=currency,
+                node=node,
+                sourceType=sourceType,
+            )
+        except ValidationError as e:
+            logger.error(
+                f"Error(s) creating LMP Event {datetime}: {e}",
```
VIKTORVAV99, line 877: If we change the event name this should also change.

electricitymap/contrib/lib/models/events.py
```diff
@@ -837,3 +837,56 @@ def to_dict(self) -> dict[str, Any]:
...
+                extra={
+                    "zoneKey": zoneKey,
+                    "datetime": datetime.strftime("%Y-%m-%dT%H:%M:%SZ"),
+                    "kind": "LMP",
```
VIKTORVAV99, line 881: And this.

Kind should be as generic as possible IMO.

electricitymap/contrib/lib/models/events.py
```diff
@@ -837,3 +837,56 @@ def to_dict(self) -> dict[str, Any]:
...
+    def create(
+        logger: Logger,
+        zoneKey: ZoneKey,
+        datetime: datetime,
+        source: str,
+        price: float | None,
+        currency: str,
+        node: str,
+        sourceType: EventSourceType = EventSourceType.measured,
+    ) -> "LMP | None":
```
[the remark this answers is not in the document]
VIKTORVAV99: This is so we are properly overwriting the correct values in the Price class, you can technically do positional value attribution so it's important we do not disturb the original order of the attributes for the type system to work properly.

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: @jbdietrich do you agree with this? Specifically making this a generic NodalPrice type?

VIKTORVAV99:
Pushed some small changes to the test and validator, I hope you don't mind.

electricitymap/contrib/lib/models/events.py
jbdietrich: I think we need a dedicated subclass to represent LMP. `LocationalMarginalPrice` is probably the most faithful to what we're really trying to capture here. It's a little early for me to say whether `NodalPrice` is generalizable enough to make sense long term.

It's also worth keeping in mind this is a PoC and we might have to throw the entire implementation away if it doesn't work, so I am not too bothered about the naming. I wouldn't expect contributors to use this class for a little while yet. We should optimize for speed for an end-to-end working rather than the perfect naming.

VIKTORVAV99: If we add it someone will come and add more parsers for it 😅<br>Might be tomorrow or 6 months from now but we need to be prepared for that regardless.<br><br>But `LocationalMarginalPrice`works if we do the full name and not the abbreviation IMO.

VIKTORVAV99:
I created this PR, https://github.com/electricitymaps/electricitymaps-contrib/pull/7947/, to clean up our internal typings a bit. There is a risk this will cause some conflicts so if you wish I can hold off on merging it until this is merged.

electricitymap/contrib/lib/models/events.py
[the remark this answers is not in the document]
silkeholmebonnen: `sourceType` I get this error when I try your suggestion: Non-default argument follows default argument. Since it is working as expected I dont think it is a problem

electricitymap/contrib/lib/models/events.py
silkeholmebonnen: Changed it to `LocationalMarginalPrice`

silkeholmebonnen:
@VIKTORVAV99 and @jbdietrich can we try to get this merged today so I can continue with the rest and MOVE FASSTTT 🏎️ ⏩ 🐎

jbdietrich (approved):

PR #7938, opened 2025-03-19
Tonyvanswet/tig 450 apply vocab to the map

tonypls:
## Issue
We want to use consistent vocabulary across out platform

## Description
This PR updates the wording of:

- Consumption -> Flow-traced
- Production -> Not flowtraced (or still production, depends on the context)
- Exchanges -> Electricity Flows
- Live -> Now

This also removes the exchange disclaimer from many US zones as it's a top level disclaimer and we have contextual disclaimers on the origin and net flow chart. 

Adds a new flow-traced badge and toggle

## Preview
New toggle and prod / consumption ff turned off:
image

New badge:
image


tonypls:
> I think these buttons are really bad for a "toggle" state like this. First up, when I click the button I have no idea what "enabled" means (is that flow-traced on?) and it's really hard to tell if it's on or off.
> 
> What if the button was instead "disable exchanges"? That means it's flow-traced by default, but one can still see non-flow-traced as an opt-in functionality. I would label it as "disable exchanges" or similar instead of "non-flow-traced" as we otherwise need to explain what flow-traced even is, which I don't think makes sense here. The icon could then be an arrow with a line across it, which I think better signifies what clicking this button actually does.
> 
> cc @Alportan :)
> 
> Edit: Oh, and what are your thoughts on the badge - why is it needed? And could we instead say we only show a badge if flow-traced is disabled? And I would maybe even call it "production only" to make it easier for people to understand (imagine seeing a screenshot on social media - a crossed-out flow-traced requires a lot of context and knowledge, whereas production only kinda implies the same thing but is easier to understand IMO.)

The button isn't perfect, we hope to move toward a settings based approach without button. 

I see what you mean, "Disable exchanges" it's worth experimenting with!

VIKTORVAV99 (dismissed):
Not really a big fan of removing the old toggle and replacing it with this button, it does not feel intuitive, the icon is not clear and there is no explainer on what it does if you don't hover on it. Something that is not possible on touch devices (like iPads).

What is the underlying reason to not just change wording in the existing toggle until we have a settings "panel" in place?

And unless I missed it there is no way to even change this setting on mobile now?

Also this flow-traced badge takes up way to much space in a location where space is already slim, we simply can't add this here and sacrifice the usability and visual clarity for something 99% of the users will care about once.

image


Will mark this as requested changes until we are at least on parity with functionality compared to today and you can toggle it on mobile as well.

VIKTORVAV99:
BTW can we do the disclaimer removal in a separate PR, those are unrelated to the discussions here and can be merged in right away if we want.

tonypls:
I've trimmed this PR down to focus on the more minor vocab and style changes. Will take the flow-traced piece in a separate PR with a settings modal.

web/src/locales/de.json
```diff
@@ -47,11 +47,7 @@
     "emissions": "Emissionen",
     "by-source": {
       "emissions": "CO₂-Emissionen",
-      "electricity-production": "Stromerzeugung",
-      "electricity-consumption": "Stromverbrauch",
-      "total-emissions": "Gesamte CO₂-Emissionen",
-      "total-electricity-production": "Gesamte Stromerzeugung",
-      "total-electricity-consumption": "Gesamter Stromverbrauch"
+      "total-emissions": "Gesamte CO₂-Emissionen"
```
madsnedergaard, line 50:
```suggestion
      "electricity-mix": "Strommix",
      "total-emissions": "Gesamte CO₂-Emissionen"
```

Screenshot 2025-03-24 at 12 27 43

web/src/locales/fr.json
```diff
@@ -121,11 +120,7 @@
     "now": "Maintenant",
     "by-source": {
       "emissions": "Émissions carbone",
-      "electricity-production": "Production d'électricité",
-      "electricity-consumption": "Consommation d'électricité",
-      "total-emissions": "Émissions totales de carbone",
-      "total-electricity-production": "Production totale d'électricité",
-      "total-electricity-consumption": "Consommation totale d'électricité"
+      "total-emissions": "Émissions totales de carbone"
```
madsnedergaard, line 123:
```suggestion
      "electricity-mix": "mix électrique",
      "total-emissions": "Émissions totales de carbone"
```

Seems like the most used phrasing for "electricity mix" as far as I can tell

madsnedergaard (approved):
LGTM (but let's apply the FR and DE translations as they should be somewhat uncontroversial 😚)

PR #7939, opened 2025-03-19
feat(app): allow keyboard shortcuts for pro users

tonypls:
## Disclaimer
I got claude to code this while I made a cup of tea.

## Issue
Shortcuts not working

## Description
This PR adds keyboard shortcuts following this pattern:

1. Press "/": this should focus the search field no matter where you are in the interface
2. Type to filter the list of areas
3. optional: use up/down arrows to select an area
4. Press enter to navigate to it


VIKTORVAV99 (commented):
When I press Shift this blue line shows up:
image
And if I try to search it in mobile "mode" it's orange...
image

The first result is always highlighted:
image


Btw why was `/` selected over say `Command + F` which is the default search shortcut in most applications?

tonypls:
> When I press Shift this blue line shows up: image And if I try to search it in mobile "mode" it's orange... image
> 
> The first result is always highlighted: image
> 
> Btw why was `/` selected over say `Command + F` which is the default search shortcut in most applications?

Will take a look at disabling it on mobile. I don't know why / was selected but it's a common shortcut for the search field used in Github, Slack and others. I don't love it when apps override default browser behaviour such as cmd + f (Github does this in the merge editor and it's quite annoying) but if you think it's a must have I can add it too.

VIKTORVAV99:
I don't think it's a must have by any means but I also don't really know what else people would search for in this view.

Preferably this shortcut should be configurable by the user and use command f by default IMO. But it's not required by any means.

Perhaps we can look into that when we create a settings panel?

tonypls:
> I don't think it's a must have by any means but I also don't really know what else people would search for in this view.
> 
> Preferably this shortcut should be configurable by the user and use command f by default IMO. But it's not required by any means.
> 
> Perhaps we can look into that when we create a settings panel?

I like your thinking, hopefully we have a settings panel in the not to distant future! I see this feature mostly being used by emaps people when demo'ing the app to allow some slick navigation.

madsnedergaard (approved):
LGTM

PR #7942, opened 2025-03-20
feat(app): reduce spammy errors from blocked plausible

tonypls:
## Issue
We're often exceeding our warning for error events in one hour on sentry

## Description
We believe the large number of errors from plausible is due to updated adblocking or similar causing the script to fail. 

This PR updates sentry to ignore errors that originate from the plausible script


VIKTORVAV99 (approved):
LGTM! 🎉

cadeban (approved):

PR #7943, opened 2025-03-20
Tvs/estimated chart layer

tonypls:
## Issue
Estimated % isn't clear what is estimated

## Description
This PR adds an overlay to the carbon chart and origin chart to communicate what values in the chart is estimated.

## Preview
image


VIKTORVAV99:
Some early feedback:
This is not clear enough in dark mode, on a low contrast screen this would not even be visible
image

I would also prefer if legend used the icon we use elsewhere for preliminary/estimated data.

And if we change this we should also change the preliminary/estimated badge that are used elsewhere to keep the same visual style. Otherwise it will be hard for users to know we mean the same thing.

image


image

And is there a particular reason we are moving away from the orange?
image

I'd love to comment directly on a figma file or linear issue if there is one. Could you link it if it exists?

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -2,14 +2,18 @@
...
 import { ElectricityModeType } from 'types';
 import { modeColor } from 'utils/constants';
 
 import { detectHoveredDatapointIndex, getNextDatetime, noop } from '../graphUtils';
 import { SelectedData } from '../OriginChart';
 import { AreaGraphElement } from '../types';
 
+// Function to check if a data point has estimation method
+const hasEstimationMethod = (d: { data: AreaGraphElement }) =>
+  d.data.meta && d.data.meta.estimationMethod;
```
VIKTORVAV99, line 15:
```suggestion
  d.data.meta?.estimationMethod;
```

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -36,13 +41,50 @@ function AreaGraphLayers({
...
+            uniqueTimestamps.add(timestamp);
+            const next = getNextDatetime(datetimes, d.data.datetime);
+            // Only add if we have a valid next datetime
+            if (next) {
+              overlayData.push({
+                datetime: d.data.datetime,
+                nextDatetime: next,
+              });
+            }
+          }
```
VIKTORVAV99, line 76:
```suggestion
          const datetime = d.data.datetime;
          const datetimeStr = datetime.toISOString();
          if (!uniqueTimestamps.has(datetimeStr)) {
            uniqueTimestamps.add(datetimeStr);
            const next = getNextDatetime(datetimes, datetime);
            // Only add if we have a valid next datetime
            if (next) {
              overlayData.push({
                datetime: datetime,
                nextDatetime: next,
              });
            }
          }
```

web/src/features/charts/elements/AreaGraphLayers.tsx
```diff
@@ -2,14 +2,18 @@
...
+import React, { useMemo } from 'react';
 import { ElectricityModeType } from 'types';
 import { modeColor } from 'utils/constants';
 
 import { detectHoveredDatapointIndex, getNextDatetime, noop } from '../graphUtils';
 import { SelectedData } from '../OriginChart';
 import { AreaGraphElement } from '../types';
 
+// Function to check if a data point has estimation method
+const hasEstimationMethod = (d: { data: AreaGraphElement }) =>
```
VIKTORVAV99, line 14: This need to check for estimated percentage as well.

And if it just checks for the estimation method it should return a boolean not the estimation method itself.

web/src/features/charts/CarbonChart.tsx
```diff
@@ -22,6 +23,11 @@ function CarbonChart({ datetimes, timeRange }: CarbonChartProps) {
   const { t } = useTranslation();
   const co2ColorScale = useCo2ColorScale();
 
+  // Call hooks before any conditional returns
+  const estimationMethod = data?.chartData?.find((d) => d.meta?.estimationMethod)?.meta
```
VIKTORVAV99, line 27: Why not a common helper for these?

web/src/features/charts/OriginChart.tsx
```diff
@@ -72,6 +73,11 @@ function OriginChart({ displayByEmissions, datetimes, timeRange }: OriginChartPr
   const isHourly = useAtomValue(isHourlyAtom);
   const selectedData = useSelectedData(displayByEmissions);
 
+  // Call hooks before any conditional returns
+  const estimationMethod = data?.chartData?.find((d) => d.meta?.estimationMethod)?.meta
+    ?.estimationMethod;
```
VIKTORVAV99, line 78: Duplicate code should be avoided otherwise our maintenance cost will go up long term.

PR #7944, opened 2025-03-21
Dark Mode, Lighter Tides: A Sea of Subtle Shades 🌚 

### LINKED ISSUE #7937: New dark mode ocean color too black
I really love the new design. 

But I think the new ocean color is too black in dark mode.

What about a slightly less dark black, like the one used in left panel's background ?

Image

madsnedergaard:
## Issue

Closes #7937

## Description

During the redesign, we did perhaps go a bit too dark on the ocean - so this PR now reverts it back to a nice dark grey instead. Thanks for the feedback from users on this :)

While at it, I also fixed the icon so we use the same one across platforms.


tonypls (approved):
I got used to the very dark sea but this is also nice 🌚

PR #7960, opened 2025-03-25
Adding source for Madeira

PaulRoms:
## Issue
Adding new source for Madeira

### DIFF
--- a/config/zones/PT-MA.yaml
+++ b/config/zones/PT-MA.yaml
@@ -78,6 +78,8 @@ emissionFactors:
       value: 12.62
 region: Europe
 sources:
+  Production:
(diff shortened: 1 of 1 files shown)


config/zones/PT-MA.yaml
```diff
@@ -78,6 +78,8 @@ emissionFactors:
       value: 12.62
 region: Europe
 sources:
+  Production:
+    link: https://www.eem.pt/pt/conteudo/sustentabilidade/rotulagem-de-energia-eletrica/evolucao-mensal-do-mix-da-oferta-e-das-emissoes-de-co2/
```
VIKTORVAV99, line 82: This sources section is just for emission factors, the sources for production is gotten directly from flowtracing and the source data tables we have.

normally this would be from the parsers, but I think silke was working on adding it to GPZD as well.

VIKTORVAV99 (changes requested):

PR #7961, opened 2025-03-25
chore: update ruff

pierresegonne:
## Issue

Our ruff version is quite old, I want the new shiny one! (Related to https://docs.astral.sh/ruff/editors/migration/ -> the language server protocol used to enabled auto linting within code editors needs to be migrated to the native version)

## Description

* Updates the ruff version
* Updates the pre-commit ruff version

### Double check


seljaks:
Trying to fix the CI errors.

Does this impact the mono repo? Or should it leave the submodule alone?

pierresegonne:
> Does this impact the mono repo?

Not yet, only after we merge that one and update contrib on the monorepo

VIKTORVAV99 (approved):
LGTM! 🎉

We should look into if there are more autofix rules we can apply now but we can do that later in a new PR.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ openpyxl = { version = "^3.1.2", optional = true }
 pydataxm = { version = "^0.3.2", optional = true }
 ruamel-yaml = "^0.17.24"
 pycountry = { version = "^22.3.5", optional = true }
-ruff = "0.4.3"
+ruff = "~0.11"
```
VIKTORVAV99, line 30: This is a bit risky though, it might change randomly and that will trigger all lockfiles to be updated in the monorepo which creates huge builds that takes forever to merge as they are updated every time someone merges to master.

VIKTORVAV99: I think so. Considering it's locked in the precommit hook either way.

seljaks:
Also not sure if we have to update the `check` script to be `ruff format .` instead of `ruff format --check .` The second option only tells you what's wrong, but doesn't reformat anything.

VIKTORVAV99:
> Also not sure if we have to update the `check` script to be `ruff format .` instead of `ruff format --check .` The second option only tells you what's wrong, but doesn't reformat anything.

Ah yeah this also has implication on the exit code used by the CI, so we should not change that.

PR #7962, opened 2025-03-25
adding new pictures of the map

PaulRoms:
As title says


VIKTORVAV99:
But are these two images not just used for the readme?
Don't think we should spend a ton of time on that.

PaulRoms:
> But are these two images not just used for the readme? Don't think we should spend a ton of time on that.

I think so.

PaulRoms:
@VIKTORVAV99  looks good in the README.md

image

VIKTORVAV99 (approved):

PR #7965, opened 2025-03-26
fix(parser): add error handling to US SPP

silkeholmebonnen:
## Issue

If there is no csv for the requested time the parser crashes

## Description

This PR adds some error handling so it prints "Empty response for {datetime}" instead of crashing

### Double check


parsers/US_SPP.py
```diff
@@ -450,6 +454,9 @@ def fetch_realtime_locational_marginal_price(
         check_datetime = target_datetime - timedelta(minutes=minutes)
         url = get_realtime_url(check_datetime)
         raw_data = get_data(url, session)
+        if raw_data.empty:
+            print(f"Empty response for {check_datetime}")
```
jbdietrich, line 458: Why use print instead of the logger (honest question!)?

silkeholmebonnen: My mistake, I'll change it

jbdietrich (approved):

PR #7968, opened 2025-03-26
Price license pt es

julievanoverloop25:
## Description

ES and PT licensing issues solved, so price can be visible again in app


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7969, opened 2025-03-26
chore(parsers): Add snapshot tests to US_NEISO fetch_exchange + switch parser for CA-NB->US-NE-ISNE

VIKTORVAV99:
## Issue

Missing snapshot tests + not being able to refetch data + potential bug in the parser.

## Description

This pull request includes changes to the `config/exchanges/CA-NB_US-NE-ISNE.yaml` file and the `parsers/US_NEISO.py` file to update the exchange data fetching process and correct configurations.

Configuration updates:

* [`config/exchanges/CA-NB_US-NE-ISNE.yaml`](diffhunk://#diff-8ed5f71973eb2c0b661a6f665699c5fce754ff7e24421e992fa1ae7d8cb84cbaL5-R5): Updated the `exchange` parser from `CA_NB.fetch_exchange` to `US_NEISO.fetch_exchange` to correct the data source.

Exchange data fetching improvements:

* [`parsers/US_NEISO.py`](diffhunk://#diff-df2a26bf60da2306bd9af495626401f86072c04e9b59c4262ead8d9b87b988abL209-R213): Modified the `fetch_exchange` function to create a list of exchange lists (`unmerged_exchanges`) and merge them at the end, instead of merging a single exchange list. This ensures all exchanges are processed and merged correctly. [[1]](diffhunk://#diff-df2a26bf60da2306bd9af495626401f86072c04e9b59c4262ead8d9b87b988abL209-R213) [[2]](diffhunk://#diff-df2a26bf60da2306bd9af495626401f86072c04e9b59c4262ead8d9b87b988abR230-R234)

### Double check


parsers/US_NEISO.py
```diff
@@ -225,11 +227,11 @@ def fetch_exchange(
                 netFlow=datapoint["Actual"] * multiplier,
                 source=SOURCE,
             )
+        # Append the exchange list to the unmerged list
+        unmerged_exchanges.append(exchanges)
 
-    # This will merge exchanges with the same datetime
-    exchanges = ExchangeList.merge_exchanges([exchanges], logger)
```
VIKTORVAV99, line 230: While this worked this is very error prone and was the reason I decided to add snapshots here in the first place before fixing it.

The ExchangeList (and all EventLists) assumes there are one event per datetime but it's not enforced, if we changed that in the future this would have created a different output.

By creating intermediate lists and then merging that we solve the issue by having unique datetimes in each list and processing everything correctly.

config/exchanges/CA-NB_US-NE-ISNE.yaml
```diff
@@ -2,5 +2,5 @@ lonlat:
   - -67.7771
   - 46.139
 parsers:
-  exchange: CA_NB.fetch_exchange
```
VIKTORVAV99, line 5: This parser don't support fetching historical data but US_NEISO does, so let's use that instead and re-fetch as much data we can when it's merged.

parsers/US_NEISO.py
seljaks: agreed, I've found the ExchangeList confusing before

seljaks (approved):
lgtm

parsers/US_NEISO.py
VIKTORVAV99: I've wanted to change the internal representation of the lists to dicts where the key is the datetime, it would significantly simplify some of the code but it would have completely broken this as it where before...

PR #7970, opened 2025-03-27
feat(app): ignore any plausible errors

tonypls:
## Issue

Improvement on #7942 as we're still getting plausible errors

The original regex pattern /plausible\.local\.js/ was too specific. It would only match errors where that exact string appears in the error message.

## Description

Sentry's ignoreErrors is primarily designed to filter errors based on their message or type, not the entire context of where the error occurred.

This PR uses the beforesend approach which is more comprehensive because it serializes the entire error object to JSON and checks if "plausible" appears anywhere in that string, catching references to plausible in URLs, stack traces, and any other properties of the error object.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7976, opened 2025-03-29
Change imperative form into infinitive for some wordings in french translations.

phiphou:
When a french locutor read "share on Facebook", he can interpret it with "please do it" or "you can do it here".

I asked on X some french followers of mine what's their point on this and the majority ends up with the infinitive form "you can do it". They showed me that many major websites are doing this in their french translations.

So I change imperative form into infinitive for some wordings in french translations.

I kept imperative form for some other wordings, those that invite the user to do something ("please do it" instead of "you can do it").


VIKTORVAV99 (approved):
LGTM! 🎉

PR #7979, opened 2025-03-31
feat(geo): add GeoJSON examples

jbdietrich:
## Description

This PR contains example GeoJSON objects to represent individual wind and solar assets. We'll ask contributors to add to these in a forthcoming issue.

### Double check

### DIFF
--- a/web/geo/solar_assets.geojson
+++ b/web/geo/solar_assets.geojson
@@ -1,0 +1,28 @@
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "properties": {
+        "name": "Example Solar Installation",
+        "capacity_mw": 0.8,
+        "comission_start_date": "2016-04-23T00:00:00Z",
+        "comission_end_date": "2016-04-23T00:00:00Z",
+        "capacity_update_date": "2025-03-01T00:00:00Z",
+        "source": "Electricity Maps"
+      },
+      "geometry": {
+        "coordinates": [
+          [
+            [12.096283407855566, 55.68669003772624],
+            [12.096283407855566, 55.683746218327684],
+            [12.097683268758374, 55.683746218327684],
+            [12.097683268758374, 55.68669003772624],
+            [12.096283407855566, 55.68669003772624]
+          ]
+        ],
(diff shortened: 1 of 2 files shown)


web/geo/solar_assets.geojson
```diff
@@ -0,0 +1,28 @@
...
+          [
+            [12.096283407855566, 55.68669003772624],
+            [12.096283407855566, 55.683746218327684],
+            [12.097683268758374, 55.683746218327684],
+            [12.097683268758374, 55.68669003772624],
+            [12.096283407855566, 55.68669003772624]
+          ]
+        ],
+        "type": "Polygon"
+      }
```
VIKTORVAV99, line 25: Do we need this to be polygons or would points be enough?

jbdietrich: I think it's easiest to consume as a polygon. Are you worried about the barrier to entry?

jbdietrich: To add a bit of context: points are way better than nothing, so if you think it encourages contribution I am happy to scale back the minimum requirement to points. However, it is most valuable for us to have polygons -- particularly if we want to get more sophisticated with how we visualize the data + combine it with weather data.

VIKTORVAV99: The barrier of entry of creating polygons are for sure higher, you need a geojson tool to do it properly especially for more advanced shapes this open up for.

A point is just a single coordinate pair to that would be a lot easier to add and would not require any special tools.

If we are just showing a dot on the map that's all we need as well, polygons have their benefit if there is a large offshore windfarm or something like that though.

VIKTORVAV99: Maybe we can accept both?

jbdietrich: Sounds good, I'll add examples for both.

VIKTORVAV99: Should we have a minimum requirement of installed capacity? Like 1 or 10mw?

jbdietrich: I have that in the issue I'm writing up. Updated with point examples.

PR #7983, opened 2025-03-31
Add french translations for the new settings panel

phiphou:
add some french translations for the new settings panel.


BastienBigue (approved):
LGTM! 
Thanks for adding these!

PR #8008, opened 2025-04-01
Tonyvanswet/PLG 584 add gtm or hubspot direct intgration to the map

tonypls:
## Issue
Analytics in the app

## Description
This PR adds the Google Tag Manager (GTM) to the app and a cookie banner.

I've chosen to embed the cookie script in the app instead of wrapping it in GTM because While technically possible, loading the consent banner through GTM is generally not recommended. It creates a chicken-and-egg problem: GTM needs consent to fire tracking tags, but it can't get consent until it loads the banner, which it can't do without consent. Loading CookieYes directly in your index.html <head> is the correct approach, as it ensures the consent mechanism loads independently and early.

I've created a data layer to filter out any events triggered on native app builds. 

In CookieYes I've enabled Google Consent Mode which means Google Tag Manager will validate the consent before sending any events to Google Analytics.


VIKTORVAV99:
Do we need to do something about this?
image

This is with a adblocker on, I'm thinking that maybe we should do what we did for plausible and ignore it?

VIKTORVAV99:
uBlock Origin, don't remember all the blocking lists though 😅

tonypls:
> uBlock Origin, don't remember all the blocking lists though 😅

Hmm, okay. I think we can release this and retro-actively add the error to Sentry when we see it. 

I've also merged the onboarding removal and it's deploying now!

VIKTORVAV99 (approved):
LGTM! 🎉

PR #8012, opened 2025-04-03
Increase delay on IN zones and BA

BastienBigue:
## Issue

IN zones reports are regularly delayed. 
BA too. 

## Description
The maximum observed delay for 2025 for IN zones is around 5 days, so set this threshold in the zone configurations. 
Ba is set to 3 days.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8013, opened 2025-04-03
chore(cleanup): Remove strings for which the translations should be changed

VIKTORVAV99:
## Issue

After we added back the all_months option a lot of the translations for other languages still use monthly for both 12 months and all months.

## Description

Just removes them for now so they either fall back to english or can be translated again properly. There are some strings we don't even use anymore as well, those have been removed along with the TimeHeader component that is no longer used.

### Double check


cadeban (approved):
🚀 Thanks for cleaning that up!

PR #8014, opened 2025-04-03
increase width of timeslider's dropdown

phiphou:
increase width of timeslider's dropdown to allow bigger text to fit in it.


VIKTORVAV99:
Is this to allow this text for example?
image

VIKTORVAV99:
> > Is this to allow this text for example? image
> 
> Yes, exactly.
> 
> I also think it would be great to have a class to capitalize only the first word...

The classes we use come from tailwind so it's not something we have defined ourselves, but I can take a look at it and see if I find something that does that.

phiphou:
I know how to do it with tailwind, I can do another PR for that if you want.

VIKTORVAV99:
Sure, but it's not just up to me if it's merged, but if there is a open PR it's a lot easier to review.

phiphou:
I'll close this one and make another one with the two changes, as they are related.

PR #8015, opened 2025-04-03
Bigger timeslider dropdown & capitalize first letter

phiphou:
Increase time slider dropdown width to allow bigger text to fit in it.

Add a capitalize-first-letter class to avoid all words to be capitalized. (https://play.tailwindcss.com/J8Rp7HA2WF?file=css)


VIKTORVAV99:
@cadeban this looks good to me but I'm unsure how this works with twmerge and if there are any other ramifications of this. What are your thoughts?

Otherwise we'll have to either do this in JS/TS (we have a utility function) or ensure it's done in the translations.

cadeban:
I'll take a look at this later today!

PR #8023, opened 2025-04-04
Update zone name for the United States in locale files

### LINKED ISSUE #8020: Update US name in translation files.
## Description
We should update the zone name for the US in the app.

It should be changed from `"Contiguous United States"` to just `"United States"` for all languages, in the native language for the translation.

Brijeshthummar02:
## Issue

Fixes #8020 

## Description

Update zone name for the United States in locale files

### Double check

### DIFF
--- a/web/public/locales/en.json
+++ b/web/public/locales/en.json
@@ -1083,7 +1083,7 @@
       "zoneName": "Uganda"
     },
     "US": {
-      "zoneName": "Contiguous United States"
+      "zoneName": "United States"
     },
     "US-AK": {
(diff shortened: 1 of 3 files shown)


VIKTORVAV99:
This only updates two languages?
What about all the rest like Swedish, French, German, etc?

VIKTORVAV99:
Well that's what the issue is about, we can't just change the English name and not the others as then it don't mean the same thing.

I don't know what it's translated to in all languages as we have a lot of them. You'll have to search for the key and adjust each translation accordingly.

PR #8028, opened 2025-04-08
feat(app): Use icons instead of question marks

VIKTORVAV99:
## Issue

Fixes: PLA-1269

### Double check


VIKTORVAV99:
I realized I didn't mark this as ready for review...

tonypls:
image
The tooltips no longer move flip up and down on desktop and instead are cut off by the overflow.

Also the mobile tooltip doesn't have the x icon anymore

VIKTORVAV99:
> image
> 
> The tooltips no longer move flip up and down on desktop and instead are cut off by the overflow.
> 
> 
> 
> Also the mobile tooltip doesn't have the x icon anymore 

I'll take care of that 🫡

PR #8032, opened 2025-04-10
feat(app): change service worker to use autoUpdate option

tonypls:
## Issue
Frequent prompting to update and a soft refresh can reload a previous version

## Description
This PR updates the service worker to use the [autoUpdate strategy](https://vite-pwa-org.netlify.app/guide/auto-update.html)

There's potential that there are some errors / weird behavior as this rolls out so I've added the error boundary reload check.

The plan is:

The browser automatically checks for a service worker update under specific conditions, primarily:
Navigation: When you navigate to a page within the scope of the service worker (essentially, any page in your app).
Service Worker Startup: If the service worker hasn't run for 24 hours, the browser might check upon its startup.
The reload happens after the new service worker successfully activates. Because we set skipWaiting: true and clientsClaim: true, the activation should happen relatively quickly after the download finishes.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8039, opened 2025-04-12
Update capacities for Reunion (RE) #8024

### LINKED ISSUE #8024: Update capacities for Reunion (RE)
## Description
It seems like the capacities for Reunion (RE) has not been updated in a while and now the production is exceeding the capacity in some cases. This should be fixed by updated the capacities to the latest values that are available for the zone to help with internal validation of the data.

t-couery:
## Issue

Closes #8024

## Description

Update capacities for La Réunion.

I based my work on OpenData data from Réunion Island's capacities :
https://colab.research.google.com/drive/1HfQB7GCvo1gDp9jo2hi2fBslf9VR9HnG#scrollTo=qf_3DjRy0gkm

It should be noted that a new gas plant (Saint-Pierre) has been built since the last figures with a capacity of 51MW in addition to 40+40 MW = 121 MW

In 'Thermique non renouvelable' we have both biomas and gas (fuel engine are biomas now) the real value is 498 - 121(gas) = 377

### Preview

image

### Double check


config/zones/RE.yaml
```diff
@@ -4,13 +4,55 @@ bounding_box:
...
+    - datetime: '2025-04-12'
+      source: reunion.edf.fr
+      value: 210.0
+  wind:
+    - datetime: '2022-10-10'
+      source: reunion.edf.fr
+      value: 15.0
+    - datetime: '2025-04-12'
+      source: reunion.edf.fr
+      value: 15.0
```
VIKTORVAV99, line 55:
```suggestion
  wind:
    - datetime: '2022-10-10'
      source: reunion.edf.fr
      value: 15.0
```
Since the value is the same we only need the first entry.

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for updating these!

PR #8042, opened 2025-04-14
remove link to app in safari

tonypls:
## Issue
F07880DF-C5E0-49A7-A40A-00DC12306968

## Description

> <meta name="apple-itunes-app" content="app-id=1224594248" />: This tells Safari on iOS devices that there's an associated native app available on the App Store (with ID 1224594248). This often results in a banner suggesting the user install or open the app.


cadeban (approved):
Nice!

PR #8045, opened 2025-04-15
feat(parser): Use 5 min data for AU

VIKTORVAV99:
## Issue

Previously all data was resampled to 30 min as that was the resolution of solar rooftop data. This is no longer the case and we can ingest the 5 minute data directly.

## Description

Removes the resampling step.

### Double check


BastienBigue (approved):
🇦🇺

BastienBigue:
Is this the root cause of the AU parsers being regularly down ?

VIKTORVAV99:
> Is this the root cause of the AU parsers being regularly down ? I doubt it but... 🤔 

Nope, they are changing and building a new API.

I'm trying to get in touch with them about it because the rate limits in the new API is not enough for us and see if we can get a higher plan with more API calls and full historical data.

PR #8048, opened 2025-04-16
feat: EMBER capacity parser handles zone overrides for known modes

BastienBigue:
## Issue
BO unknown capacity is invalid : gas capacity is separated while CNDC (the BO data source) reports gas as termico that we map to unknown because it contains gas but also other modes. 

## Description
This PR : 
- Fix the EMBER capacity mapper so that zone specific overrides can be defined even for modes that are known.
- Define a gas capacity mapping override for BO
- Update the BO capacities with the capacity_update script


VIKTORVAV99:
Do we want to do this or simply re assign it in the quality validator?

It has the capabilities for it already.

I don't mind doing it here but if feels like we have overrides everywhere.

BastienBigue:
Actually, why do quality validation have overrides ? 
It looks like it is a way of solving a problem without solving the cause of it, no ?

VIKTORVAV99:
> @VIKTORVAV99 any more thoughts on this ? 
> 
> I still think that the best fix (because if fixes the root cause) is to have a capacity value that matches what the data providers actually reports, and thus make the fix in the capacity parser, but let me know if you prefer to fix this in the quality validation 🙂  

I'm not against it but I worry that it might be hard(er) to maintain in the long run as external contributors might add new capacity and if the reviewer don't know about this or have forgotten about it, it will break the validation. There is less risk of that if we go with the re-allocation method but on the other hand the capacities won't match.

Maybe that should be a part of a larger discussion in how we ingest capacities at one point but for now I'll just approve it and leave it up to you since it's for a single zone.

VIKTORVAV99 (approved):

PR #8049, opened 2025-04-16
Parser generation forecast data JP

alejandranavcas:
## Description

This PR is to include parser for generation forecast data in all zones in Japan (JP).

### Double check


parsers/JP.py
```diff
@@ -352,9 +361,178 @@ def parse_dt(row):
...
+    # Parse the different formats
+    if zone_key == "JP-CG":
+        with ZipFile(BytesIO(response.content)) as z:
+            target_file = f"{datestamp}_yosoku_chugoku.csv"
+            if target_file in z.namelist():
+                with z.open(target_file) as f:
+                    df = pd.read_csv(f, encoding="shift_jis", skiprows=startrow)
+            else:
+                print(f"File '{target_file}' not found in the zip archive.")
+                print(f"Available files: {z.namelist()}")
```
VIKTORVAV99, line 470: This should be a logger if you intend to use this for debugging in production.

parsers/JP.py
```diff
@@ -352,9 +361,178 @@ def parse_dt(row):
...
+    else:
+        content = response.content.decode("shift_jis")
+        df = pd.read_csv(StringIO(content), skiprows=startrow)
+
+    # Extract the correspondant information from the files for each zone
+    if zone_key == "JP-HKD" or zone_key == "JP-HR":
+        df = df[["日付", "時間帯_自", "エリア総発電量(kWh)"]]
```
VIKTORVAV99, line 477: Since the df is created in a conditional above it's potentially unbound here and downstream.

parsers/JP.py
```diff
@@ -352,9 +361,178 @@ def parse_dt(row):
...
+            event["datetime"] = datetime.strptime(
+                datetime_string, "%Y/%m/%d%H:%M"
+            ).replace(tzinfo=ZONE_INFO)
+        elif zone_key == "JP-TK":
+            time_int = int(event["Time"])
+            time_str_padded = (
+                f"{time_int:04d}" if time_int >= 100 else f"00{time_int:02d}"
+            )
+            time_obj = datetime.strptime(time_str_padded, "%H%M").time()
+            date_obj = datetime.strptime(str(int(event["Date"])), "%Y%m%d").date()
```
VIKTORVAV99, line 513: Why is this converted first to a int and then a string?

VIKTORVAV99 (commented):
Looks like it works but I'm not a huge fan of casting converting these objects to various types instead of parsing them directly, what is the reason for this?

parsers/JP.py
VIKTORVAV99: But the CSVs don't have any decimals and are stored as INTs directly, so the casting INTs should not be needed, so if there are decimals added that signals a bigger issue somewhere. Probably because we are using Pandas to read the CSV.

alejandranavcas: Yes, that's true. I'll look into that then :)

alejandranavcas: I found the issue: when using iterrows(), pandas converts the values to float because of the way it handles data internally. Now, I'll use itertuples() instead

alejandranavcas:
Thanks for all the feedback! 
I've now implemented the changes you suggested and also added `generationForecast: JP.fetch_generation_forecast` in the parsers object in the config files. Let me know if more changes should be done!

parsers/JP.py
```diff
@@ -352,9 +361,195 @@ def parse_dt(row):
...
+            )
+        else:
+            date_int = int(row.Date)
+            year = date_int // 10000
+            month = (date_int % 10000) // 100
+            day = date_int % 100
+            hour, minute = map(int, row.Time.split(":"))
+            event_datetime = datetime(year, month, day, hour, minute).replace(
+                tzinfo=ZONE_INFO
+            )
```
VIKTORVAV99, line 537:
```suggestion
        if (
            zone_key == "JP-HR"
            or zone_key == "JP-ON"
            or zone_key == "JP-TH"
            or zone_key == "JP-CB"
        ):
            event_datetime = datetime.strptime(
                f"{row.Date} {row.Time}", "%Y/%m/%d %H:%M"
            ).replace(tzinfo=ZONE_INFO)

        elif zone_key == "JP-TK":
            # Format as string in the format "YYYYMMDD HHMM" and parse
            event_datetime = datetime.strptime(
                f"{row.Date:08d} {row.Time:04d}", "%Y%m%d %H%M"
            ).replace(tzinfo=ZONE_INFO)
        else:
            # Use strptime to directly parse the date integer and time string
            event_datetime = datetime.strptime(
                f"{row.Date:08d} {row.Time}", "%Y%m%d %H:%M"
            ).replace(tzinfo=ZONE_INFO)
```
How about we do this instead? It gets rid of a lot of type issues and are less code overall.

alejandranavcas: Yes, thank you 👍  that looks so much better! But there was a zone that was not working correctly (JP-KY), so I just included it in another elif condition separately. I will push the commit now

VIKTORVAV99 (approved):
LGTM! 🎉

PR #8051, opened 2025-04-18
refactor: simplify option checks in ToggleButton toggles

bcmyguest:
## Issue

Relates to https://github.com/electricitymaps/electricitymaps-contrib/issues/8035

## Description

Goal here is to address the zone/country _and_ electricity flows in settings not being saved on refresh. To accomplish this, I initially intended to use `atomWithStorage` and let it be. However, I noticed that the `onToggle` function wasn't being triggered correctly when the value is changed on the toggle buttons (this may have been my error, but I was having a hard time running the app locally so it's difficult to tell for sure). To make the minor change I had to change the usages of this function which include the zone/country and the emission/electricity buttons

Edit:

Removed the `atomWithStorage` change but kept the rest

### Preview

Sorry it's lagging out my PC, I can try on another machine later. Should be easy enough to validate

1. change the value from the default in your settings
2. refresh the page

Expect

1. the value is what you set it to
2. this is accomplished by saving the value in localstorage

### Double check

- [ ] I have tested my parser changes locally with `poetry run test_parser "zone_key"` -> not applicable
- [x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes. -> NO -> need to do this, eslint was lagging out my PC as well :'(

### DIFF
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -61,12 +61,12 @@ function ToggleButton({
         type="single"
         aria-label="Toggle between modes"
         value={selectedOption}
+        onValueChange={onToggle}
       >
         {options.map(({ value, translationKey, dataTestId }, key) => (
           <ToggleGroupItem
             key={`group-item-${key}`}
             value={value}
-            onClick={() => onToggle(value)}
             data-testid={`toggle-button-${dataTestId ?? value}`}
             className={twMerge(
               'inline-flex h-7 w-full items-center whitespace-nowrap rounded-full bg-neutral-100/0 px-3 text-xs dark:border dark:border-neutral-400/0 dark:bg-transparent',
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -24,7 +24,8 @@ export type ToggleButtonOptions<T extends string> = Array<{
 interface ToggleButtonProperties<T extends string> {
   options: ToggleButtonOptions<T>;
   selectedOption: T;
-  onToggle: (option: T) => void;
+  // radix gives back an empty string if a new value is not selected
+  onToggle: (option: T | "") => void;
   tooltipKey?: string;
   transparentBackground?: boolean;
 }
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -16,26 +16,26 @@ import { twMerge } from 'tailwind-merge';
 
 import { GlassBackdrop } from './GlassContainer';
 
-export type ToggleButtonOptions = Array<{
-  value: string;
+export type ToggleButtonOptions<T extends string> = Array<{
+  value: T;
   translationKey: string;
   dataTestId?: string;
 }>;
-interface ToggleButtonProperties {
-  options: ToggleButtonOptions;
-  selectedOption: string;
-  onToggle: (option: string) => void;
+interface ToggleButtonProperties<T extends string> {
+  options: ToggleButtonOptions<T>;
+  selectedOption: T;
+  onToggle: (option: T) => void;
   tooltipKey?: string;
   transparentBackground?: boolean;
 }
 
-function ToggleButton({
+function ToggleButton<T extends string>({
   options,
   selectedOption,
   tooltipKey,
   onToggle,
   transparentBackground = true,
-}: ToggleButtonProperties): ReactElement {
+}: ToggleButtonProperties<T>): ReactElement {
   const { t } = useTranslation();
   const [isToolTipOpen, setIsToolTipOpen] = useState(false);
   const onToolTipClick = () => {
@@ -116,4 +116,6 @@ function ToggleButton({
   );
 }
 
-export default memo(ToggleButton);
+// react and typescript doesn't pass through generics so we need to cast
+// https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37087#issuecomment-1765701020
+export default memo(ToggleButton) as typeof ToggleButton;
--- a/web/src/components/ToggleButton.tsx
+++ b/web/src/components/ToggleButton.tsx
@@ -25,7 +25,7 @@ interface ToggleButtonProperties<T extends string> {
   options: ToggleButtonOptions<T>;
   selectedOption: T;
   // radix gives back an empty string if a new value is not selected
-  onToggle: (option: T | "") => void;
+  onToggle: (option: T | '') => void;
   tooltipKey?: string;
   transparentBackground?: boolean;
 }
--- a/web/src/features/map-controls/ConsumptionProductionToggle.tsx
+++ b/web/src/features/map-controls/ConsumptionProductionToggle.tsx
@@ -32,9 +32,7 @@ function ConsumptionProductionToggle({
       trackEvent(TrackEvent.PRODUCTION_CONSUMPTION_CLICKED, {
         productionConsumption: option,
       });
-      setCurrentMode(
-        currentMode === Mode.PRODUCTION ? Mode.CONSUMPTION : Mode.PRODUCTION
-      );
+      setCurrentMode(option as Mode);
     },
     [currentMode, setCurrentMode]
   );
--- a/web/src/features/map-controls/ConsumptionProductionToggle.tsx
+++ b/web/src/features/map-controls/ConsumptionProductionToggle.tsx
@@ -25,8 +25,8 @@ function ConsumptionProductionToggle({
 }): ReactElement {
   const [currentMode, setCurrentMode] = useAtom(productionConsumptionAtom);
   const onSetCurrentMode = useCallback(
-    (option: Mode | "") => {
-      if (option === "") {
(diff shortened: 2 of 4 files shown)


web/src/features/map-controls/ConsumptionProductionToggle.tsx
```diff
@@ -32,9 +32,7 @@ function ConsumptionProductionToggle({
       trackEvent(TrackEvent.PRODUCTION_CONSUMPTION_CLICKED, {
         productionConsumption: option,
       });
-      setCurrentMode(
-        currentMode === Mode.PRODUCTION ? Mode.CONSUMPTION : Mode.PRODUCTION
-      );
+      setCurrentMode(option as Mode);
     },
```
bcmyguest, line 36: We can use the option that the user selected here

web/src/features/map-controls/SpatialAggregatesToggle.tsx
```diff
@@ -27,27 +27,20 @@ function SpatialAggregatesToggle({
   const onSetCurrentMode = useCallback(
     (option: string) => {
       if (
-        (option === SpatialAggregate.ZONE && currentMode === SpatialAggregate.ZONE) ||
-        (option === SpatialAggregate.COUNTRY && currentMode === SpatialAggregate.COUNTRY)
+        (option === currentMode)
```
bcmyguest, line 30: The check already boiled down to this.

web/src/features/map-controls/SpatialAggregatesToggle.tsx
```diff
@@ -27,27 +27,20 @@ function SpatialAggregatesToggle({
...
         return;
       }
       trackEvent(TrackEvent.SPATIAL_AGGREGATE_CLICKED, { spatialAggregate: option });
-      setCurrentMode(
-        currentMode === SpatialAggregate.COUNTRY
-          ? SpatialAggregate.ZONE
-          : SpatialAggregate.COUNTRY
-      );
+      setCurrentMode(option as SpatialAggregate);
     },
```
bcmyguest, line 36: Same as above 

> We can use the option that the user selected here

web/src/features/map-controls/SpatialAggregatesToggle.tsx
```diff
@@ -27,27 +27,20 @@ function SpatialAggregatesToggle({
...
   );
 
   return (
     <ToggleButton
       options={options}
-      selectedOption={
-        currentMode === SpatialAggregate.ZONE ? options[1].value : options[0].value
-      }
+      selectedOption={currentMode}
       onToggle={onSetCurrentMode}
```
bcmyguest, line 44: Not clear why we weren't doing this already

web/src/utils/state/atoms.ts
```diff
@@ -59,8 +59,8 @@ export const selectedDatetimeStringAtom = atom<string>((get) => {
   return dateToDatetimeString(datetime);
 });
 
-export const spatialAggregateAtom = atom(SpatialAggregate.ZONE);
-export const productionConsumptionAtom = atom(Mode.CONSUMPTION);
+export const spatialAggregateAtom = atomWithStorage('spatialAggregate', SpatialAggregate.ZONE);
+export const productionConsumptionAtom = atomWithStorage('productionConsumption', Mode.CONSUMPTION);
```
VIKTORVAV99, line 63: This was removed by design before as people complained about zones being missing or exchanges not working (productionConsumption) while the toggle was simply turned off and they forgot.

While not optimal as a stopgap we decided to not save the setting across sessions anymore. We'll add a more robust solution in the future that will allow power users to toggle this while it will remain unsaved for regular users in the future. We just need to shape this a bit first.

VIKTORVAV99 (commented):
Some very nice changes in here in regards to simplifying the code, those we should definitely keep but we cant change the atoms just yet without me speaking with the team and coming up with a good solution that addresses both the concerns of the issue and the internal concerns that prompted this change.

So if you could just revert the atom changes themselves and do the type change I suggested we can merge this right away.

web/src/utils/state/atoms.ts
bcmyguest: Makes complete sense, I honestly was just looking to learn a bit about the codebase anyways. Thanks for the review! I'll make those changes (revert atoms, make the other minor fixes, change commit message).

web/src/features/map-controls/SpatialAggregatesToggle.tsx
```diff
@@ -27,27 +27,20 @@ function SpatialAggregatesToggle({
...
       ) {
         return;
       }
       trackEvent(TrackEvent.SPATIAL_AGGREGATE_CLICKED, { spatialAggregate: option });
-      setCurrentMode(
-        currentMode === SpatialAggregate.COUNTRY
-          ? SpatialAggregate.ZONE
-          : SpatialAggregate.COUNTRY
-      );
+      setCurrentMode(option as SpatialAggregate);
```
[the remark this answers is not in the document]
bcmyguest: I think changing line 28 will cause the type hint on lint 42 to complain - happy to resolve this but honestly not sure what the suggestion is in the case?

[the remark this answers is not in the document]
VIKTORVAV99: While they are named the same they are not in the same closure so they are independent of each other, the type that is set on line 28 is only used inside the useCallback closure.

And the type on options on 42 is already of the type SpatialAggregate. If anything it's the function signature for line 45 that need to change. But it should be fine to update the type for the toggle button to accept Mode and SpatialAggregate as well so we don't need to cast the types.

VIKTORVAV99:
Please don't use force push, it makes it more or less impossible to keep track of changes between commits and requires people that have checked out the branch locally make a hard reset against origin when there are changes made as the git tree no longer matches.

bcmyguest:
> Please don't use force push, it makes it more or less impossible to keep track of changes between commits and requires people that have checked out the branch locally make a hard reset against origin when there are changes made as the git tree no longer matches.

Sorry about this!

web/src/utils/state/atoms.ts
VIKTORVAV99: No problem at all, feel free to ping me if you have any questions about the code!

web/src/features/map-controls/SpatialAggregatesToggle.tsx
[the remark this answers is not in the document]
bcmyguest: I _think_ I get what you're saying but not 100% sure. Updated the ToggleButton component to accept anything that extends a string...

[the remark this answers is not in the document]
VIKTORVAV99: It's pretty late here and I've put away my laptop for the day so I'll review it properly tomorrow morning.

But it looks great and exactly what I had in mind! Nice work!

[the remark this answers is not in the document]
bcmyguest: Thanks! Appreciate the review, have a good night :) I'll be away till Monday anyways so no rush at all

VIKTORVAV99 (commented):
Looks great!

This toggle is used in one more place where the types need to be updated to pass the CI here: web/src/features/panels/zone/DisplayByEmissionToggle.tsx

When that's updated we can merge this!

web/src/components/ToggleButton.tsx
```diff
@@ -61,12 +61,12 @@ function ToggleButton({
         type="single"
         aria-label="Toggle between modes"
         value={selectedOption}
+        onValueChange={onToggle}
```
VIKTORVAV99, line 64: With this change it seems the value that is emitted when the values is the same as the current value is a empty string `""`. This messes with: 
```
      if (option === currentMode) {
        return;
      }
```
Which has caused some unexpected behavior when zone is pressed multiple times for example.

I think we need to update it to check against empty strings instead.
To me this seems like a bug on radix though so I'll also open a issue there, as to me this function should only run on value changes but currently that's not how it's working.

bcmyguest: Will fix with a comment

bcmyguest: Based on the above, I typed it as 'expected behaviour' but still added the comment to explain (cause I admit it looks strange)

VIKTORVAV99: Which is very weird as anonymous functions passed as props will cause the component to be unstable and re-render even if it where memod so that goes against best practices.

So I'm glad we are getting rid of it here.

bcmyguest: I think this is done, could you please approve the workflow when you get the chance?

VIKTORVAV99 (approved):
LGTM! 🎉

Once again, thanks a lot!

VIKTORVAV99:
> @VIKTORVAV99 thank you for your help, it was fun doing some learning here. I'll keep an eye out for new tickets :)

It was my sincere pleasure!

You are very welcome to contribute more and if you need anything or have any questions just ping me!

PR #8052, opened 2025-04-18
chore(parsers): Remove capacity parsing from OPENNEM parser as the source no longer reports capacity

VIKTORVAV99:
## Issue

The source stopped reporting capacity so we are constantly trying to parse None values all the time. This is also preventing us from using the new parser classes as they don't support capacity parsing just yet.

## Description

Removes code related to capacity parsing and updates snapshot.

### Double check


silkeholmebonnen (approved):
LGTM. Maybe we should also delete `productionCapacity: OPENNEM.fetch_production_capacity` from the AU zones, since there is no fetch_production_capacity function

VIKTORVAV99:
> LGTM. Maybe we should also delete `productionCapacity: OPENNEM.fetch_production_capacity` from the AU zones, since there is no fetch_production_capacity function

I think that's from another parser entirely with the same name that's located under capacity parsers. So should be okay for now.

PR #8053, opened 2025-04-18
feat(parser): Build OPENELECTRICITY parser

VIKTORVAV99:
## Description

This is very much a work in progress PR but it adds the basics of a new parser for open electricity.

Its still missing exchange flows and it's missing solar rooftop power. But the core is there and it's working as intended.

There is a python package we could use to try and simplify this: https://github.com/opennem/openelectricity-python but I's still in beta so IDK if we should do that yet.

### Double check


seljaks:
checked up on this while doing #8289 
AFAIK:
* python client is not yet stable
* `https://api.openelectricity.org.au/v4/data/network/` would work for production data but not for imports/exports
* they do accept "imports", "exports", "interconnectors" as valid "fueltech" but the api calls don't return any data
* we can filter for zones in the api to get a cleaner response

PR #8057, opened 2025-04-23
Added Turkish language support

altayaydin:
### Double check


web/src/locales/tr.json
VIKTORVAV99, line 1: Overall this looks good but there should be no english language strings in this file. Please remove them.

altayaydin: Let me translate all the strings in this file and I'll hit you up again.

altayaydin:
@VIKTORVAV99 It has been a long time but now it should be ready to merge. I updated the description as everything is translated.

madsnedergaard:
Thanks for contributing! 🙌
While we really appreciate the work you've put in here, we have made a decision to limit the amount of languages that we support in the app. Over the last few years, we have struggled with keeping all of the existing 28(!) translations up to date when adding new features and due to our limited resources most of the locales are only partially translated now leading to a suboptimal experience. Therefore we'd rather have fewer, higher quality translations than many half-done ones.

But we hope that we in the future can revisit this topic and bring your translations back to live, so this work is not lost! :)

PR #8058, opened 2025-04-23
fix(parsers/IN): fix fossil production parser for IN after spreadsheet structure changed

BastienBigue:
## Issue
Since 15/04/2025, a footer was added to the spreadsheet reported by NPP for IN production. 
It changes the structure of the spreadsheet and thus we are unable to parse it anymore. 

## Description
This PR removes the footer and the empty columns it generates once read by pandas to end up with the same dataframe structure as we had before. 

[Example of old spreadsheet](https://npp.gov.in/public-reports/cea/daily/dgr/31-01-2025/dgr2-2025-01-31.xls) (without footer)
[Example of new spreadsheet](https://npp.gov.in/public-reports/cea/daily/dgr/20-04-2025/dgr2-2025-04-20.xls) (with footer and column changes)
Notice that production mode (e.g. THER(GT)) moved from columns CDE to EFG, which, once read by pandas, ends up in different columns. 

Tested locally on historical and live spreadsheets. 

### Double check

Screenshot 2025-04-23 at 11 33 40


VIKTORVAV99 (approved):
LGTM! 🥳

seljaks (approved):

PR #8059, opened 2025-04-23
bug(web): purge old service worker registrations

tonypls:
## Issue
We still have users stuck on old app versions

## Description
This PR intends to get and unregister all service worker registrations then reload the app with a new registration.


VIKTORVAV99 (commented):
This is only temporary right?

Otherwise it will remove the service worker every time the page is reloaded which kind of removes the point of having it.

VIKTORVAV99:
Not sure this will even work though as this will only remove it after the app has been updated so if users are stuck updating this will do nothing and it will only remove it for those that are already on auto update.

tonypls:
Closing this one in favour of #8069

PR #8060, opened 2025-04-23
rebrand(web) update favicon

tonypls:
## Issue

New favicon 🚀


web/public/icons/favicon.svg
```diff
@@ -1,24 +1,20 @@
...
-    circle {
+    path {
       fill: black;
     }
     @media (prefers-color-scheme: dark) {
-      path,
-      circle {
+      path {
         fill: white;
       }
```
VIKTORVAV99, line 9: Don't this need one for the rect to fill as black when dark mode is preferred?

tonypls: I thought so too, but as it's just a clipping path it doesn't matter.

VIKTORVAV99 (approved):
Let's try it!

PR #8064, opened 2025-04-24
Ss/fix india renewable parser

seljaks:
## Issue
The code was reading the excel file incorrectly.
Improved the mapping of regions to zone_keys (the reason this was broken was because of an extra white space) 😠 
Removed the summing logic because the source already contained the sums we're interested in.


parsers/IN.py
```diff
@@ -7,7 +7,6 @@
 from typing import Any
 from zoneinfo import ZoneInfo
 
-import numpy as np
```
VIKTORVAV99, line 10: Nice!

VIKTORVAV99 (approved):
LGTM! 🎉

BastienBigue (approved):
Cool ! 
Let's merge this, zone is grey 😢

parsers/IN.py
seljaks: got caught by ruff :D

PR #8066, opened 2025-04-24
fix(parser): fix JP-KY parser

silkeholmebonnen:
## Issue

The JP-KY parser is down since the website has changed. 

## Description

This PR reads the consumption values from a csv file instead of from the website

### Preview

```
Parser result:
{'datetime': datetime.datetime(2025, 4, 24, 23, 50, tzinfo=zoneinfo.ZoneInfo(key='Asia/Tokyo')),
 'production': {'biomass': 0,
                'coal': 0,
                'gas': 0,
                'geothermal': None,
                'hydro': None,
                'nuclear': 3114.0,
                'oil': 0,
                'solar': 0.0,
                'unknown': 4796.0,
                'wind': None},
 'source': 'www.kyuden.co.jp',
 'storage': {},
 'zoneKey': 'JP-KY'}
---------------------
took 18.92s
min returned datetime: 2025-04-24 14:50:00+00:00 UTC
max returned datetime: 2025-04-24 14:50:00+00:00 UTC  -- OK, <2h from now :) (now=2025-04-24T14:58:01+00:00 UTC)
```

### Double check


silkeholmebonnen:
This parser was quite confusing to me because we get the values from different places. If the reviewer is also a bit confused here is some context:
- Get the realtime value of nuclear. We get multiple values which is values from multiple generators, but only at the current timestamp.
- Get the solar and consumption values from a csv file in 5 minute granularity. Only use the latest one since we only have realtime nuclear values.
- Get exchanges and calculate generation based on consumption and exchanges
- Calculate unknown from the generation, solar and nuclear

VIKTORVAV99 (approved):
LGTM! 🎉

silkeholmebonnen:
I'm thinking it might be a good idea to add a test for this parser, but I might do that in another PR to get this fix out

PR #8068, opened 2025-04-25
test(parsers): add tests for JP-KY

silkeholmebonnen:
## Issue

[GMM-762](https://linear.app/electricitymaps/issue/GMM-762/create-test-for-jp-ky)
## Description

This PR adds some tests for JP-KY since it is a bit of an unusual parser

### Double check


VIKTORVAV99 (commented):
Since you are already adding all the mocks and the other tests can we get a snapshot test or two to get the full output?

Seems like the current tests only test one event?

VIKTORVAV99 (approved):
LGTM! 🎉

PR #8069, opened 2025-04-25
no op service worker

tonypls:
## Issue
Users don't receive the latest version of the app

## Description
This is an intermediate step to clean up old service worker configurations that might be stored in users' browsers.

This change should purge outdated configurations and reset them to default settings. We can then determine whether to achieve our desired outcome with a properly configured service worker or remove it entirely, allowing Cloudflare to handle all caching.

Without the requirement for offline functionality, combined with Cloudflare's built-in cache invalidation and the laborious workflow of making and validating service worker changes, I'm leaning toward eliminating the service worker completely.

### DIFF
--- a/web/src/main.tsx
+++ b/web/src/main.tsx
@@ -1,3 +1,5 @@
+/// <reference types="vite-plugin-pwa/client" />
+
 // Init CSS
 import 'maplibre-gl/dist/maplibre-gl.css';
 import 'react-spring-bottom-sheet/dist/style.css';
@@ -31,6 +33,7 @@ import { createConsoleGreeting } from 'utils/createConsoleGreeting';
 import enableErrorsInOverlay from 'utils/errorOverlay';
 import { getSentryUuid } from 'utils/getSentryUuid';
 import { refetchDataOnHourChange } from 'utils/refetching';
+import { registerSW } from 'virtual:pwa-register';
 
 const isProduction = import.meta.env.PROD;
 if (isProduction) {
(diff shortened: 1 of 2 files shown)


VIKTORVAV99:
I'd be willing to try this but I don't think it will do what we want in it's current state.

I think we might have to build our own no-op service worker as currently it output this:
```js
if (!self.define) {
  let e,
    i = {};
  const n = (n, o) => (
    (n = new URL(n + ".js", o).href),
    i[n] ||
      new Promise((i) => {
        if ("document" in self) {
          const e = document.createElement("script");
          (e.src = n), (e.onload = i), document.head.appendChild(e);
        } else (e = n), importScripts(n), i();
      }).then(() => {
        let e = i[n];
        if (!e) throw new Error(`Module ${n} didn’t register its module`);
        return e;
      })
  );
  self.define = (o, s) => {
    const t =
      e ||
      ("document" in self ? document.currentScript.src : "") ||
      location.href;
    if (i[t]) return;
    let r = {};
    const c = (e) => n(e, t),
      f = { module: { uri: t }, exports: r, require: c };
    i[t] = Promise.all(o.map((e) => f[e] || c(e))).then((e) => (s(...e), r));
  };
}
define(["./workbox-84c915bb"], function (e) {
  "use strict";
  self.skipWaiting(),
    e.clientsClaim(),
    e.precacheAndRoute(
      [
        {
          url: "icons/icon-maskable.svg",
          revision: "6327b7a739ac643e9c3fd32311ec63c8",
        },
        { url: "icons/icon.svg", revision: "ec4a8d6a10320fe605581e0d331fc13e" },
        {
          url: "manifest.webmanifest",
          revision: "976908437319d02a02ae34fa7dd6caaf",
        },
      ],
      {},
    ),
    e.cleanupOutdatedCaches();
});
```
This don't look like it is no-op to me but it might work anyway as it don't include the code bundles anymore from what I can tell.

VIKTORVAV99:
@tonypls what do you want to do with this PR?

tonypls:
Close it!

PR #8071, opened 2025-04-25
update consumption forecast parser

alejandranavcas:
## Description

This PR is to change the fetch_consumption_forecast in JP.py function so that it fetches better data.
Before: hourly data only until the end of the day.
Now: the consumption forecast data is fetched from the same file as generation forecast data --> half hour data also until next day.

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8072, opened 2025-04-25
fix padding for safe area top

tonypls:
## Issue
image

## Description
This PR adds some padding for devices where a safe area top is required.


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8074, opened 2025-04-25
Fix "1 hour app delay"

ziqiyang1:
This PR removes the obsolete `1-hour-app-delay` feature flag and all related parameters throughout the frontend data‐fetching logic:

- **helpers.ts**:  
  - Dropped the `isAppDelay` parameter from `getParameters` signature and implementation.
- **getState.ts**:  
  - Removed `useFeatureFlag('1-hour-app-delay')` import and hook.  
  - Updated `getState` function to accept only timeRange and targetDatetime
  - Updated React Query hook to call `getState(timeRange, urlDatetime)` and removed `is1HourAppDelay` from its `queryKey`.
- **getZone.ts**:  
  - Removed `useFeatureFlag('1-hour-app-delay')` import and hook.  
  - Updated `getZone` function signature to timeRange and targetDatetime
  - Updated React Query hook and `queryKey` to match the new signature.

### DIFF
--- a/web/src/api/getZone.ts
+++ b/web/src/api/getZone.ts
@@ -1,11 +1,8 @@
 import type { UseQueryResult } from '@tanstack/react-query';
 import { useQuery } from '@tanstack/react-query';
-import { useFeatureFlag } from 'features/feature-flags/api';
 import { useAtomValue } from 'jotai';
 import { useParams } from 'react-router-dom';
-import invariant from 'tiny-invariant';
-import type { ZoneDetails } from 'types';
-import { RouteParameters } from 'types';
+import type { GridState, RouteParameters } from 'types';
 import { TimeRange } from 'utils/constants';
 import { isValidHistoricalTimeRange } from 'utils/helpers';
 import { getStaleTime } from 'utils/refetching';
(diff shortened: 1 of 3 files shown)


web/src/api/getZone.ts
```diff
@@ -21,73 +18,56 @@ import {
...
-
+): Promise<GridState> => {
   const shouldQueryHistorical =
     targetDatetime &&
     isValidDate(targetDatetime) &&
     isValidHistoricalTimeRange(timeRange);
 
   const path: URL = new URL(
-    `v10/details/${TIME_RANGE_TO_BACKEND_PATH[timeRange]}/${zoneId}${getParameters(
+    `v10/state/${TIME_RANGE_TO_BACKEND_PATH[timeRange]}${getParameters(
```
VIKTORVAV99, line 31: Why is the URL updated here? It will completely break the app as it's not the same data that is being fetched.

web/src/api/getZone.ts
```diff
@@ -21,73 +18,56 @@ import {
   TIME_RANGE_TO_BACKEND_PATH,
 } from './helpers';
 
-const getZone = async (
+const getState = async (
```
VIKTORVAV99, line 21: And why is this renamed?

web/src/api/getZone.ts
VIKTORVAV99, line 1: Did you just copy the getState file to the getZone file? You can't do that, they do two different things?

VIKTORVAV99 (changes requested):

PR #8077, opened 2025-04-28
Fix CY.py parser 403 error by adding User-Agent header

PaulRoms:
## Summary
This PR fixes the 403 Forbidden error in the Cyprus (CY) parser by adding a User-Agent header to the HTTP request.

## Changes
- Added a User-Agent header to the session.get() request in the fetch_production method
- The User-Agent mimics a standard web browser to bypass anti-scraping measures

## Context
The Cyprus electricity data website (tsoc.org.cy) was blocking our requests with a 403 Forbidden error. This is a common anti-scraping measure used by websites to prevent automated access. Adding a User-Agent header that mimics a legitimate web browser helps bypass these restrictions.

## Testing
- Tested the parser locally to verify it can now successfully fetch data without receiving 403 errors
- Confirmed the parser correctly extracts and processes electricity production data from Cyprus


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8080, opened 2025-04-28
fix(ES): decrease parser validation for spain

BastienBigue:
## Issue
Reduce threshold for Spain minimal production in ENTSOE parser


PaulRoms (approved):
LGTM

PR #8086, opened 2025-04-29
Show estimation markers

corradio:
## Issue

It's not clear when data is synthetic or measured (see [PLG-210](https://linear.app/electricitymaps/issue/PLG-210/its-not-clear-what-the-estimation-percent-on-graphs-refers-to)).

## Description

This PR is an attempt to visualise the dates where data is synthetic.

Temporary notes:
- [x] Remove "Partially estimated" as there are cases where data is partially estimated (e.g. TSA) but we don't show it as such)
- [x] For Tier A zones: show "Preliminary". For Tier B+C: show "Synthetic".

### Preview

image

### Double check


web/src/features/charts/bar-breakdown/BarBreakdownChart.tsx
```diff
@@ -1,17 +1,22 @@
 import * as Portal from '@radix-ui/react-portal';
+import useGetZone from 'api/getZone';
```
corradio, line 2: I moved the EstimationCard inside the BarBreakdownChart (which forces me to move some imports and calculations inside the BarBreakdownChart) as the message depends on the datetime selected, and thus should be confined to the area of the UI which is datetime specific.

Before:
image

After:
image

web/src/features/charts/graphUtils.ts
```diff
@@ -236,38 +237,35 @@ const analyzeChartData = (chartData: AreaGraphElement[]) => {
...
 
   if (estimatedTotal === 0) {
     return {};
   }
 
   if (estimatedTotal) {
     return {
-      text: t('estimation-card.aggregated_estimated.pill', {
-        percentage: estimatedTotal,
-      }),
```
corradio, line 252: this had custom logic which was conflicting with `getEstimationOrAggregationTranslation`. I therefore ensured it uses `getEstimationOrAggregationTranslation` instead of the custom logic.

web/src/features/charts/graphUtils.ts
```diff
@@ -236,38 +237,35 @@ const analyzeChartData = (chartData: AreaGraphElement[]) => {
...
     };
   }
-
-  if (allEstimated) {
-    return { text: t('estimation-badge.fully-estimated'), icon: TrendingUpDown };
-  }
-
-  if (hasEstimation) {
-    return { text: t('estimation-badge.partially-estimated'), icon: TrendingUpDown };
-  }
```
corradio, line 270: These should be handled by the `getEstimationOrAggregationTranslation` logic so the logic can be shared across the app.

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -105,12 +105,14 @@ export default function EstimationCard({
   }
   const selectedData = data?.datetimes[selectedDatetimeString]?.z[zoneKey];
 
-  const estimationMethod = selectedData?.em;
+  const estimationMethod = selectedData?.em || undefined;
 
-  if (!estimationMethod) {
+  const isAggregated = !isHourly;
+
+  if (!estimationMethod && !isAggregated) {
```
corradio, line 112: This now shows the estimation card on aggregated data. It wasn't showing before.

now:
image

web/src/features/panels/zone/EstimationCard.tsx
```diff
@@ -266,7 +277,6 @@ export function AggregatedCard({
 }) {
   return (
     <BaseCard
-      estimationMethod={EstimationMethods.AGGREGATED}
```
corradio, line 269: Aggregated is not an estimationMethod.

web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -84,13 +82,6 @@ export default function ZoneDetails(): JSX.Element {
           zoneDataStatus={zoneDataStatus}
         >
           <BarBreakdownChart hasEstimationPill={hasEstimationPill} />
-          {zoneDataStatus !== ZoneDataStatus.NO_INFORMATION && (
-            <EstimationCard
-              zoneKey={zoneId}
-              zoneMessage={zoneMessage}
-              estimatedPercentage={roundedEstimatedPercentage}
-            />
-          )}
```
corradio, line 93: Moved EstimationCard inside BarBreakdownChart

web/src/utils/constants.ts
```diff
@@ -245,7 +245,6 @@ export enum EstimationMethods {
   MODE_BREAKDOWN = 'ESTIMATED_MODE_BREAKDOWN',
   RECONSTRUCT_BREAKDOWN = 'ESTIMATED_RECONSTRUCT_BREAKDOWN',
   RECONSTRUCT_PRODUCTION_FROM_CONSUMPTION = 'ESTIMATED_RECONSTRUCT_PRODUCTION_FROM_CONSUMPTION',
-  AGGREGATED = 'aggregated',
```
corradio, line 248: Not an estimation model

web/src/utils/getEstimationTranslation.ts
```diff
@@ -0,0 +1,67 @@
...
+import { round } from 'utils/helpers';
+
+export default function getEstimationOrAggregationTranslation(
+  t: TFunction,
+  field: 'title' | 'pill' | 'body' | 'legend',
+  isAggregated: boolean,
+  estimationMethod?: EstimationMethods,
+  estimatedPercentage: number = 0
+) {
+  // TODO: This method mixes too many concerns and too many cases. Refactor required.
```
corradio, line 12: More context:
The logic of what estimation translation is used many places (various tooltips, pills etc..), and its logic is spread across this file and the translation file (which maps some estimationModels to specific labels).
I've tried to simplify things a bit more by ensuring all logic goes through this file. I also simplified some of the logic that selects what label to show.
However, there's a bigger refactor needed as the translation file still has many estimationModels (and ideally it shouldn't have any logic, but rather only contain the final labels we want to show). 
Finally, it's worth noting this file is also used to showcase aggregations (that are not estimated). We should ideally separate concerns.

I leave that exercise for a future refactor.

corradio:
@tonypls would you have time to make a first review pass here? I can fine-tune the UI with Alex as soon as this is done.

tonypls:
> @tonypls would you have time to make a first review pass here? I can fine-tune the UI with Alex as soon as this is done.

Yeah for sure! Will do it now

web/src/features/charts/EmissionChart.tsx
```diff
@@ -1,11 +1,14 @@
-import EstimationBadge from 'components/EstimationBadge';
+import { round } from '@turf/helpers';
```
tonypls: Could be nice to consistently use the round from utils/helpers

web/src/features/charts/OriginChart.tsx
```diff
@@ -103,17 +100,36 @@ function OriginChart({ displayByEmissions, datetimes, timeRange }: OriginChartPr
     );
   }
 
+  const estimated = chartData.map((d) => {
```
tonypls: Could be nice to have a hook that all the charts can use for this:

i.e.
```
  const useEstimationData = (chartData) => {
     return useMemo(() => {
       const estimated = chartData.map((d) => {
         const { estimationMethod, estimatedPercentage } = d.meta;
         const roundedEstimatedPercentage = round(estimatedPercentage ?? 0, 0);
         return estimationMethod != undefined || Boolean(roundedEstimatedPercentage);
       });
       const estimationMethod = chartData.find((d) => d.meta.estimationMethod)?.meta.estimationMethod;
       const someEstimated = estimated.some(Boolean);
       return { estimated, estimationMethod, someEstimated };
     }, [chartData]);
   };
```

tonypls (commented):
A very quick initial review.

Nice feature!

There's an issue with the translations for aggregated synthetic data that doesn't have a value for the % estimated, see image:
image

corradio:
> A very quick initial review.
> 
> Nice feature!
> 
> There's an issue with the translations for aggregated synthetic data that doesn't have a value for the % estimated, see image: image

This seems to be because the estimationMode is missing for aggregated data (e.g. Peru), cf https://linear.app/electricitymaps/issue/GMM-821/aggregations-are-missing-the-origin-field#comment-582ea897

web/src/features/charts/OriginChart.tsx
corradio: Good idea.

web/src/features/charts/EmissionChart.tsx
corradio: done!

web/src/features/charts/OriginChart.tsx
corradio: done

corradio:
> > A very quick initial review.
> > Nice feature!
> > There's an issue with the translations for aggregated synthetic data that doesn't have a value for the % estimated, see image: image
> 
> This seems to be because the estimationMode is missing for aggregated data (e.g. Peru), cf https://linear.app/electricitymaps/issue/GMM-821/aggregations-are-missing-the-origin-field#comment-582ea897

This was fixed by https://github.com/electricitymaps/electricitymaps-contrib/pull/8086/commits/2e341677ff5b79f3d5a2d0ee092ad5abfb661298

It turns out this is the reason why currently, aggregated zones with 0% estimations show as "imprecise":
image

I had removed the "imprecise" fallback to ensure all cases are accounted for.

tonypls:
The positioning of this box, touching the border doesn't look right and the jumpiness when scrolling between synthetic and non-synthetic data points isn't ideal:
image

In aggregated views the whole chart for some signals are labelled as synthetic (This is France 90D):
image

I would support the idea that aggregated measured data is measured enough to not be labelled as synthetic, regardless this should be consistent between price and other signals.

We could also consider the shading is only for granular data, in some cases 3 days in 1 year would make a whole year as preliminary even if this is only 1% of data:
image


Also worth considering in this case why is data from 2017 "preliminary" when data in 2024 isn't? Feels like synthetic would be better labelling for aggregated cases like this.
image

corradio:
Thanks for the extensive review.
> The positioning of this box, touching the border doesn't look right and the jumpiness when scrolling between synthetic and non-synthetic data points isn't ideal:

Done in https://github.com/electricitymaps/electricitymaps-contrib/pull/8086/commits/b6e7848f58af48c8470c47ceb7d8a23e9a83a042

> In aggregated views the whole chart for some signals are labelled as synthetic (This is France 90D):

This bug was recently introduced. I was able to pinpoint it to the details endpoint which now returns a `"MEASURED"` string (note the inside quotes which shouldn't be there) as `estimationMethod` instead of null/undefined (which we use for the non-aggregated endpoints). I think this may be related to some of the recent changes in the app-backend as a consequence of now having the `origin` field populated in the table of aggregates?
image

> We could also consider the shading is only for granular data, in some cases 3 days in 1 year would make a whole year as preliminary even if this is only 1% of data:

It would require changes across the frontend as the checks are scattered across tooltips etc.. and are not centralised. I suggest we do in a separate PR although I support and agree on the need to change this.

> Also worth considering in this case why is data from 2017 "preliminary" when data in 2024 isn't? Feels like synthetic would be better labelling for aggregated cases like this.

Which zone is this? This would happen if the aggregates predominantly contains e.g. TSA data. In this case, data will be marked as Preliminary (after all, we're still waiting for measured data for these zones - else, they would have been filled with another estimation model such as GPZD, and marked as Synthetic). This highlights to me that we haven't activated the right estimation models, or haven't refetched data. My conclusion is that the depiction is accurate.

tonypls:
> This bug was recently introduced. I was able to pinpoint it to the details endpoint which now returns a "MEASURED" string (note the inside quotes which shouldn't be there) as estimationMethod instead of null/undefined (which we use for the non-aggregated endpoints). I think this may be related to some of the recent changes in the app-backend as a consequence of now having the origin field populated in the table of aggregates?

Yeah, I believe we didn't have an origin field for aggregated views and DBT wraps strings in double quotes which makes out equality check in the background return false. I fixed this for the state response here: https://github.com/electricitymaps/electricitymaps/pull/8109/files

> It would require changes across the frontend as the checks are scattered across tooltips etc.. and are not centralised. I suggest we do in a separate PR although I support and agree on the need to change this.

Sounds good!

> Which zone is this? This would happen if the aggregates predominantly contains e.g. TSA data. In this case, data will be marked as Preliminary (after all, we're still waiting for measured data for these zones - else, they would have been filled with another estimation model such as GPZD, and marked as Synthetic). This highlights to me that we haven't activated the right estimation models, or haven't refetched data. My conclusion is that the depiction is accurate.

This one: [https://697b7cd9.electricitymaps-staging.pages.dev/zone/US-NW-TPWR/all/yearly](https://697b7cd9.electricitymaps-staging.pages.dev/zone/US-NW-TPWR/all/yearly)

Yeah it could be the correct labelling but from a users perspective it does raise some concerns as to why data is still preliminary after 8 years. A refetch or re-run of the correct estimation models for older data we share on the app would be a good thing!

corradio:
> > > We could also consider the shading is only for granular data, in some cases 3 days in 1 year would make a whole year as preliminary even if this is only 1% of data:
> >
> > It would require changes across the frontend as the checks are scattered across tooltips etc.. and are not centralised. I suggest we do in a separate PR although I support and agree on the need to change this.
> 
> Sounds good!

It turns out this is an issue. I realised that because the DBT tables now issues a `origin` field, it means that this DB field is actually the source of truth for whether a zone is estimated or not. Overriding this based on a estimation % will create inconsistencies with the API. Therefore, I've had to refactor the frontend code in order to never use the % to decide if a zone is estimated or not (this information should be decided by GMM in DBT instead). With the current implementation it means that an aggregated zone is estimated if it has > 50% estimated data.

Furthermore, I renamed the variable `hasEstimation` into `hasEstimationOrAggregationPill` in order to clarify that the pill should be shown if there's an estimation OR an aggregation. This will make it more easy to update the code in the future. 

Bonus: I also noticed the LoadChart was missing a pill.

The associated commit for these changes is https://github.com/electricitymaps/electricitymaps-contrib/pull/8086/commits/ba687baed9fe5917671a5eca7cc0a7167ee33de1.

tonypls:
Looks good! Just the naggy bot suggestions to tidy up the any types and I think we're good to go.

corradio:
Should be done now. Can you help me fix the Cypress test though?

tonypls:
> Should be done now. Can you help me fix the Cypress test though?

Done, please double check it's correct. I also removed an unused cardType prop and defaulted the estimation to GPZD, feel free to revert if this isn't ideal. You'll have better context of the intended changes here and I don't have time right now to work through it in detail.

corradio:
> > Should be done now. Can you help me fix the Cypress test though?
> 
> Done, please double check it's correct. I also removed an unused cardType prop and defaulted the estimation to GPZD, feel free to revert if this isn't ideal. You'll have better context of the intended changes here and I don't have time right now to work through it in detail.

Hi @tonypls,

First, let me say I really appreciate the support in getting this through.
I've looked at the EstimationCard, and it actually will not get rendered if the `estimationMethod` is undefined (unless the zone is aggregated, in which case it will show the aggregated card). I therefore have removed the test, which doesn't make sense IMO, as we should never show an estimation card without an `estimationMethod` (unless aggregated).
I've also reverted the change defaulting to GPZD, as I don't believe the frontend should fallback to anything (it should just follow what the backend tells it).

tonypls (approved):
LGTM! 

Some nice updates!

PR #8093, opened 2025-04-30
fix: set PT lower bound to 0

VIKTORVAV99:
## Issue
We still have some estimated datapoints for PT.

## Description
This pull request includes a small change to the `parsers/ENTSOE.py` file. The change updates the `expected_range` for "PT" to allow a lower bound of 0 instead of 600.


corradio:
Good catch!

corradio (approved):

PR #8094, opened 2025-04-30
Put solar before wind on graphs

corradio:
## Issue

I realised that some graphs missed the opportunity to showcase the daily solar patterns properly in situations when there wind, because wind is put under solar in the graphs.
I think this makes sense because solar has a more recognisable and regular pattern than wind.

See e.g.

image
image
image
image

## Description

This PR puts solar underneath wind.

### Preview

image
image
image
image

### Double check


VIKTORVAV99 (approved):

VIKTORVAV99:
Let's try it!

PR #8113, opened 2025-05-01
feat: add enum for parser data types

pierresegonne:
## Issue

* The parser data types were not defined properly, meaning that:
  * We wouldn't 100% certainly know if a zone config exposes an incorrect parser type
  * We have a lot of manual string writing to do "production", "production", ... when dealing with the handling of parsers
  * Is making it harder to update the set of supported data types as we would most likely forget to update some of the references that need to be updated

## Description

* Creates an enum for the supported parser data types
* Tests that the `Parsers` model contains 1:1 all the parser data types
* Effectively creates a runtime check for the validity of the types loaded under `PARSER_DATA_TYPE_TO_DICT`

### Double check


pierresegonne:
As discussed https://electricitymaps.slack.com/archives/C08LH4L9PFB/p1746519910179509 I'll remove all references to `productionPerUnit` in a follow-up PR and then bump both to the monorepo

tests/config/test_config_model.py
```diff
@@ -51,6 +57,24 @@ def test_zone_sources(self):
                             assert zone_sources is not None  # pyright type-narrowing
                             self.assertIn(source, zone_sources, f"zone: {zone_key}")
 
+    def test_parser_model_contains_all_parser_data_types(self):
```
pierresegonne, line 60: Should we also test that for Delays?

electricitymap/contrib/config/constants.py
```diff
@@ -2,7 +2,6 @@
 
 EXCHANGE_FILENAME_ZONE_SEPARATOR = "_"
 
-EXCHANGE_DATA_TYPES = ["exchange", "exchangeForecast"]
```
VIKTORVAV99, line 5: Just to double check, is nothing using this internally?

tests/config/test_config_model.py
VIKTORVAV99: I think the current changes are enough but if you want to add a test feel free to do so.

VIKTORVAV99 (approved):
LGTM! 🎉

electricitymap/contrib/config/constants.py
pierresegonne: don't think so no!

BastienBigue:
@pierresegonne 
We will revert your PR
It makes the feeder-electricity fail and we do not fetch any data at all : https://console.cloud.google.com/run/jobs/details/europe-west1/feeder-electricity-job/logs?inv=1&invt=Abwv4Q&project=tmrow-152415

PR #8114, opened 2025-05-01
fix: remove duplicates for JP-KY emission factors for battery and hydro discharge

BastienBigue:
There are duplicates in the emission factors config for JP-KY. 
It makes it unpractical to use : prevent from creating a pivot table.
This PR only keeps the latest emission factor computed for each year.


VIKTORVAV99 (approved):
LGTM! 🎉

Nice catch!

PR #8115, opened 2025-05-01
WIP: Nested nodes list

VIKTORVAV99:
## Issue
The current event lists where only designed for a zoneKey datetime primary key reference and to better enforce that on the parser level we need to make some changes in how we handle nodal events.

## Description
This pull request introduces a new `Node` event type and refactors the `LocationalMarginalPrice` model to use a `NodeList` for handling multiple nodes instead of a single node. The changes improve the flexibility and structure of the data model for handling node-related information.

### New `Node` event type:
* Introduced a new `Node` class in `electricitymap/contrib/lib/models/events.py`, which represents a node with attributes such as `node`, `price`, and `currency`. Added a validator for the `currency` field and a `to_dict` method for serialization.

### Refactoring `LocationalMarginalPrice`:
* Updated the `LocationalMarginalPrice` class to use `NodeList` instead of individual `node`, `price`, and `currency` fields. This includes changes to its `create` method and `to_dict` method for handling the new structure. [[1]](diffhunk://#diff-8249df35f59b9a94b35ab9717193815e6273ffac2a2578df5f6c9369640b1a2eL842-R883) [[2]](diffhunk://#diff-8249df35f59b9a94b35ab9717193815e6273ffac2a2578df5f6c9369640b1a2eL889-R900)
* Modified the `append` method in `LocationalMarginalPriceList` to accept a `NodeList` instead of individual node attributes.

### Supporting changes:
* Added a `NodeList` class in `electricitymap/contrib/lib/models/event_lists.py` for managing a collection of `Node` events. This includes methods for appending nodes and converting them to a sorted list of dictionaries.
* Updated imports in `electricitymap/contrib/lib/models/events.py` to include `NodeList`.
* Updated imports in `electricitymap/contrib/lib/models/event_lists.py` to include `Node`.

### DIFF
--- a/parsers/US_SPP.py
+++ b/parsers/US_SPP.py
@@ -457,22 +457,24 @@ def fetch_realtime_locational_marginal_price(
             logger.warning(f"Empty response for {check_datetime}")
             continue
 
-        grouped = raw_data.groupby(["Pnode", "GMTIntervalEnd"]).first()
+        # Group by timestamp (GMTIntervalEnd)
+        for interval_end, group in raw_data.groupby("GMTIntervalEnd"):
+            # Create a NodeList for this timestamp
+            nodes = NodeList(logger)
 
-        for (node, interval_end), row in grouped.iterrows():
+            # Add each node in this group to the NodeList
+            for _, row in group.iterrows():
+                nodes.append(node=row["Pnode"], price=row["LMP"], currency="USD")
+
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
@silkeholmebonnen and @jbdietrich this introduces the core basics of what I want to change. Still have some tests to add to ensure it's working as expected and then to modify the parsers.

PR #8116, opened 2025-05-02
bug(app): fix presentation error on missing price + currency

ronan-mch:
When price and currency are missing, the tooltip displays ? ? / MWh. Instead, we should display ? / MWh in this scenario.

### Double check


tonypls (approved):
LGTM

PR #8118, opened 2025-05-05
feat(zones): Add subzone names to Russia

florianscheidl:
## Issue
Russia is made up of multiple zones. At the same time, we have a country-level zone (called RU). The latter does not contain subZoneNames in its config.
EDIT: We've added the subzones as bypassedSubZones to bypass aggregation (it currently does not make sense for Russia due to missing data in the subzones).

## Description
Adds subZoneNames and bypassedSubZones for the country zone Russia (RU).


VIKTORVAV99:
This is intentional and has been discussed between GMM and Forecasts before, has something changed?

Previously it was decided we should not aggregate it unless there is data for all subzones (the aggregation will fail anyway making all of Russia gray in country mode).

florianscheidl:
> LGTM ! Have you tried to run one of the flowtracing pipeline with this config locally ?

I haven't but can still do.

VIKTORVAV99 (approved):
LGTM! 🎉

But please do run the test as suggested, not sure it can handle that all subzones are bypassed as is right now.
If not it's something we should add support for though to make the pipeline more robust.

florianscheidl:
> LGTM! 🎉
> 
> But please do run the test as suggested, not sure it can handle that all subzones are bypassed as is right now. If not it's something we should add support for though to make the pipeline more robust.

Have run flowtracing locally and it ran without any issues.

VIKTORVAV99:
Then let this rip and merge it!

PR #8126, opened 2025-05-08
chore(graphs): Move battery storage next to solar in origin graph

VIKTORVAV99:
## Issue

After we moved solar I think it makes sense to move battery storage as well as they are almost ramping up and down at the same time or vice versa.
## Description
This pull request makes a minor update to the `modeOrder` array in `web/src/utils/constants.ts` by reordering the position of `'battery storage'` to prioritize it above `'wind'` but below `'solar'`.

The aim of this is to try and keep the wind layer of the graph smoother by reducing the drastic changes in the solar output by complementing it with the battery storage values as they have a high correlation.

### Preview
With this change:
image

On master:
image


corradio:
Interesting! I was thinking about how to evaluate this, as the rationale presented here could also apply for wind (if there's a zone where wind & storage ramps up / down together).

I think it would make sense to follow the typical merit-order here, as this would represent how each technology is used to fill the gap once all previous technologies have been dispatched.

The question I have is whether battery storage is activated before or after wind & hydro storage?
My understanding was that typically battery storage is going to be more expensive than wind & hydro storage, and therefore, it will only be called upon to fill the remaining once solar/wind/hydro/hydro storage have been used.
If this understanding is correct, then I think the current order on master makes more sense, as we should see e.g. hydro patterns fill the gaps before battery storage kicks in.
Another thought is that battery storage probably attempts to fill the residual load (i.e. the gap once both solar *and* wind have been dispatched), so I would err on the side of not doing this change.

VIKTORVAV99:
I'll admit I didn't consider merit order at all but rather correlating signals to keep the above signals as smooth as possible.

Interestingly enough I can't really find where battery discharges usually fit into the merit order. Since most BESS use daily cycles the merit order might be lower than we think as otherwise they won't discharge enough to charge during the next surplus's cycle (in the case of California) which means curtailment instead.

I'll try and dig into it a bit and see where battery discharges usually fit in in grid scale capacities.

VIKTORVAV99:
This article seems to put batteries before wind in the merit order based on the skip reasons: https://modoenergy.com/research/balancing-mechanism-dispatch-skip-battery-energy-storage

Will see if I can find some corroborating data from another source as well.

VIKTORVAV99:
Since @madsnedergaard has done this I'll just close this PR.

PR #8127, opened 2025-05-09
chore(capacity): Update EIA capacities

VIKTORVAV99:
## Issue

For some zones there have been some significant changes that might affect outlier detection.

## Description

Instead of updating just the individual zones with flagged event's I just went wide and updated all EIA data.

### Double check


silkeholmebonnen (approved):
LGTM ⭐

PR #8128, opened 2025-05-09
chore(capacity): Update EMBER capacity

VIKTORVAV99:
## Issue

EMBER capacities are out of date.
## Description

Updates the EMBER capacities up to 2024-01-01 (latest available data).

### Double check

### DIFF
--- a/config/zones/CO.yaml
+++ b/config/zones/CO.yaml
@@ -17,6 +17,9 @@ capacity:
     - datetime: '2021-01-01'
       source: Ember, Yearly electricity data
       value: 350.0
+    - datetime: '2023-01-01'
+      source: Ember, Yearly electricity data
+      value: 390.0
   coal:
     - datetime: '2017-01-01'
       source: Ember, Yearly electricity data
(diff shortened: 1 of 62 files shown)


config/zones/MX.yaml
```diff
@@ -120,7 +129,7 @@ contributors:
   - consideRatio
 country: MX
 delays:
-  production: 1080 # = 45 * 24 Mexico is known for having a maximum delay of 45 days
```
silkeholmebonnen, line 123: Do we want to delete the comments?

config/zones/CO.yaml
```diff
@@ -102,18 +105,24 @@ capacity:
...
+    - datetime: '2022-01-01'
+      source: Ember, Yearly electricity data
+      value: 30.0
 contributors:
   - lav-julien
   - alixunderplatz
   - consideRatio
 country: CO
 delays:
-  production: 84 # As of 2025-03-17, Colombia's recent maximum delay is 3.5 days
```
silkeholmebonnen, line 116: Same here

silkeholmebonnen (approved):
Two small comments but not blocking 😄

config/zones/MX.yaml
VIKTORVAV99: No we do not...

I guess the YAML library don't keep comments, I'll add them back manually.

PR #8129, opened 2025-05-09
fix(parsers): Fix TAIPOWER parsing of battery and hydro storage

VIKTORVAV99:
## Issue

Fixes: GMM-771 (hopefully)

## Description

Fixes the parsing of the hydro and battery storage.

### Double check


silkeholmebonnen:
Should we add a test for this parser?

VIKTORVAV99:
> Should we add a test for this parser?
> 
> 

There is a snapshot test already with the old structure that's currently failing. I'll update it later today.

parsers/test/__snapshots__/test_TAIPOWER.ambr
```diff
@@ -3,36 +3,37 @@
...
+        'unknown': 1791.3,
+        'wind': 181.9,
       }),
       'source': 'taipower.com.tw',
       'sourceType': <EventSourceType.measured: 'measured'>,
       'storage': dict({
-        'hydro': -935.7,
+        'battery': 5.2,
+        'hydro': 474.0,
       }),
```
VIKTORVAV99, line 37: I think this is right but feel free to double check that I'm doning these conversions correctly.

BastienBigue: Conversions look good to me !

BastienBigue (approved):
LGTM

PR #8133, opened 2025-05-12
Python upgrade 3.11

ronan-mch:
## Description

Upgrades to Python 3.11

### Double check


VIKTORVAV99:
@ronan-mch mind if I push a small CI change so the proper tests run?

ronan-mch:
@VIKTORVAV99 - yes please.

VIKTORVAV99:
FYI there is a target python version for ruff, this don't need to change in this PR and just controls how the formatting rules are applied. There will be a lot of small changes when we change that so we can do it in a separate PR as it won't affect any functionality.

VIKTORVAV99 (approved):
LGTM! 🎉

Let's go!

PR #8137, opened 2025-05-13
feat(web): script to generate arrows

tonypls:
## Issue
Arrow script doesn't create webp files
## Description
This PR gets the old script working with webp files too.

Use with caution! I haven't validated the size of the files generated.


corradio:
Hi @tonypls 
I took the liberty to push three commits. The first one ensures the script can be run from the `script` folder (like before).
The second one changes the quality from 75 to 50, which ensures the files are ~ 10kb like before.
The third one regenerates the arrows.

corradio:
This probably still requires some work as the arrows changed a bit:
image

corradio (approved):

PR #8139, opened 2025-05-14
Hindi_translated

Deriverx2:
## Issue #4782

Translations for the new zones in India

## Description

Created Hindi translation 

### Double check

-  I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.


.vscode/settings.json
```diff
@@ -9,5 +9,8 @@
   },
   "python.testing.pytestArgs": [".", "--ignore", "electricitymap/contrib/parsers"],
   "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "[json]": {
+    "editor.defaultFormatter": "vscode.json-language-features"
+  }
```
VIKTORVAV99: This should not be in this PR.

web/src/locales/hi.json
```diff
@@ -0,0 +1,1894 @@
...
+        },
+        "ZA": {
+            "zoneName": "South Africa"
+        },
+        "ZM": {
+            "zoneName": "Zambia"
+        },
+        "ZW": {
+            "zoneName": "Zimbabwe"
+        }
```
VIKTORVAV99, line 1037: I assume these have translated versions of at least the countries, why are they not added?

If temporary these keys should not be added at all so it can fall back to the english translation.

VIKTORVAV99 (commented):
Thanks for the PR and sorry for the late review, there has been a lot of vacations and such!

.vscode/settings.json
Deriverx2: Removed it

web/src/locales/hi.json
Deriverx2: Translated the countries

madsnedergaard:
Thanks for contributing! 🙌
While we really appreciate the work you've put in here, we have made a decision to limit the amount of languages that we support in the app. Over the last few years, we have struggled with keeping all of the existing 28(!) translations up to date when adding new features and due to our limited resources most of the locales are only partially translated now leading to a suboptimal experience. Therefore we'd rather have fewer, higher quality translations than many half-done ones.

But if this changes in the future, we'll revisit this PR so thanks again! :)

PR #8144, opened 2025-05-19
fix(parsers): fix return type for some parsers

BastienBigue:
We enforced type checking of the returned value from the parsers. 
We expects the parsers to return a list. 

The following were returning a dict and thus were failing. 
The list of affected parsers is :
- production: JP-KN, JP-SK, JP-KY, GB-ORK, CA-AB, AU-WA-RI, AU-TAS-FI, AU-TAS-KI
- consumption: KW
- price: NZ
- exchange: GB->GB-ORK

### Double check

### DIFF
--- a/parsers/CA_AB.py
+++ b/parsers/CA_AB.py
@@ -121,40 +121,42 @@ def fetch_production(
         }
         for row in csv.reader(response.text.split("\r\n\r\n")[3].splitlines())
     }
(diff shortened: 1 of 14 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

seljaks:
love the simplicity :D

PR #8145, opened 2025-05-19
fix(parsers): NZ production parser slipped through the net

BastienBigue:
- return a list instead of a dict after type checking was enforced

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

We already merged the 3.11 python PR though so might not be able to deploy this right away.

PR #8147, opened 2025-05-19
feat(parsers): Add grid alert infrastructure

silkeholmebonnen:
## Issue

[GMM-790](https://linear.app/electricitymaps/issue/GMM-790/create-pjm-grid-alert-parser)

## Description

This PR adds the infrastructure, on the contrib side, to handle grid alerts, according to [iteration 1](https://docs.google.com/spreadsheets/d/1Td9mK2bLYqpHT82HO1b9JgyQ8D1LzgcarZTLHGFSpRE/edit?gid=1716536059#gid=1716536059) of the discussed schema 
```
{
   "source": free text, # url
   "zoneKey": EMaps zone,
   "locationRegion": free text, # if given, sub region reported by ISO
   "alertType": ..., # one of ["Informational", "Action", "Undefined"]
   "message": free text, # Main description (+ messageSubject)
   "issuedTime": timestamp, # timestamp when notification is issued
   "startTime": timestamp, # if provided, timestamp when event starts. default = issuedTime
   "endTime": timestamp, # if provided, timestamp when event ends
}
```

### Double check


alejandranavcas:
I implemented the proposed grid alerts schema here :) I commented out `sourceType`, because I don't know if it makes sense having it for grid alerts. Should I include it as `EventSourceType.measured` ? or just remove it from the schema?

silkeholmebonnen:
> I implemented the proposed grid alerts schema here :) I commented out `sourceType`, because I don't know if it makes sense having it for grid alerts. Should I include it as `EventSourceType.measured` ? or just remove it from the schema?

I think you should just remove it

electricitymap/contrib/lib/models/events.py
```diff
@@ -892,3 +892,112 @@ def to_dict(self) -> dict[str, Any]:
...
+                zoneKey=zoneKey,
+                locationRegion=locationRegion,
+                source=source,
+                # sourceType=sourceType,
+                alertType=alertType,
+                messageBody=messageBody,
+                issuedTime=issuedTime,
+                startTime=startTime,
+                endTime=endTime,
+                datetime=issuedTime,  # Event requires a datetime field
```
silkeholmebonnen, line 977: Do we want both datetime and issuedTime? Or should we just use datetime=issuedTime?

electricitymap/contrib/lib/models/events.py
```diff
@@ -892,3 +892,112 @@ def to_dict(self) -> dict[str, Any]:
...
+class GridAlertType(str, Enum):
+    informational = "informational"
+    action = "action"
+    undefined = "undefined"
+
+
+class GridAlert(Event):
+    locationRegion: str | None = None
+    alertType: GridAlertType
+    messageBody: str
```
silkeholmebonnen: Should we just call it message instead of messageBody?

electricitymap/contrib/lib/models/events.py
alejandranavcas: We have to use `datetime` because it's required by the event type. But when we defined the schema, the name of that variable is `issuedTime` (so that it can be easily identifiable and distinguishable).

electricitymap/contrib/lib/models/events.py
alejandranavcas: we can do that yes, since we are also including the message subject now

jbdietrich (approved):

PR #8150, opened 2025-05-20
fix(parsers): fix IN parser when there is no renewables

BastienBigue:
## Issue
Indian data is parsed from two sources, one for renewable modes, one for conventional. 
Currently, and since 14/05, the renewable modes are not available in the source. 
This is not handled by the parser : the case where renewable modes are missing raise a type error because None cannot be expanded into a dict when mixing both renewable and conventional data. 
This leads to no data being returned even though conventional data is available, and zones becoming grey on the map after the estimates horizon ends. 

## Description
This PR handles the error when the renewable modes are missing : it fetches the conventional ones and returns them even if the low carbon are missing. 
Since IN zones are mainly using conventional modes (at least 90%), I think this is acceptable. 
If conventional data is missing, no data is returned. 

### Double check

Tested locally : datapoint from 19/05 were returned with conventional only vs no data returned before the change


PaulRoms (approved):
LGTM

VIKTORVAV99 (approved):
LGTM! 🎉

PR #8158, opened 2025-05-22
Ss/solar gem

seljaks:


electricitymap/contrib/config/model.py
```diff
@@ -340,3 +342,54 @@ def _load_config_model() -> ConfigModel:
...
 )
+
+
+class DataCenter(StrictBaseModel):
+    displayName: str
+    lonlat: tuple[float, float] | None
+    provider: str
+    region: str
+    status: str
+    zoneKey: ZoneKey
```
VIKTORVAV99, line 353: Id include an array of aliases here, GCP data centers for example are split up in A, B, C sections which could be nice to capture if we are going to make a lookup API for them.

web/scripts/uploadGeoJSON.js
```diff
@@ -0,0 +1,65 @@
...
+      metadata: {
+        cacheControl: 'public, max-age=60',
+        contentType: 'application/json',
+        contentEncoding: 'gzip',
+      },
+    });
+
+    console.log(`${inputPath} uploaded to ${BUCKET_NAME} as ${destinationFile}`);
+  } catch (error) {
+    console.error('Upload failed:', error);
```
VIKTORVAV99, line 59: Don't you want to exit here as well?

web/src/api/getSolarAssets.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { getHeaders } from './helpers';
+
+const getSolarAssets = async (): Promise<any> => {
+  const path: URL = new URL(
+    'https://storage.googleapis.com/testing-gzipped-geojson/solar_power_plants.geojson'
+  );
+
+  const requestOptions: RequestInit = {
+    method: 'GET',
```
VIKTORVAV99, line 12: Get is the default.

web/src/features/map/Map.tsx
```diff
@@ -67,6 +72,9 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
   const setMousePosition = useSetAtom(mousePositionAtom);
   const [isLoadingMap, setIsLoadingMap] = useAtom(loadingMapAtom);
   const [hoveredZone, setHoveredZone] = useAtom(hoveredZoneAtom);
+  const [selectedSolarAsset, setSelectedSolarAsset] = useAtom(selectedSolarAssetAtom);
+  const setHoveredSolarAssetInfo = useSetAtom(hoveredSolarAssetInfoAtom);
+  const hoveredSolarAssetInfoValue = useAtomValue(hoveredSolarAssetInfoAtom);
```
VIKTORVAV99, line 77: Should use `useAtom` instead to get both values at once.

VIKTORVAV99 (commented):
I realized halfway through that maybe the request for review was not intentional 😅

Let me know if it is...

PR #8160, opened 2025-05-22
feat(data-centers): cleanup gcp regions + add azure regions 

wobniarin:
## Issue

There were too many data centers that we found for Google that did not correspond to any of the available gcp regions. We also did not have any regions for Azure.

## Description
We clean up gcp regions to only those available from gcp data source. 
We include azure regions that are operating. 

### Double check

### DIFF
--- a/config/data_centers/data_centers.json
+++ b/config/data_centers/data_centers.json
@@ -110,5 +110,461 @@
     "region": "europe-west3",
     "zoneKey": "DE",
     "status": "operational"
+  },
+  "azure-westus3": {
+    "provider": "azure",
+    "lonlat": [-111.763275, 34.395342],
+    "displayName": "West US 3",
+    "region": "West US 3",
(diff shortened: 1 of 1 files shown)


config/data_centers/data_centers.json
```diff
@@ -1,202 +1,570 @@
 {
   "gcp-us-central1": {
     "provider": "gcp",
-    "lonlat": [41.1755951, -95.8028031],
-    "displayName": "Southlands Campus (us-central1)",
+    "lonlat": [-95.8028031, 41.1755951],
+    "displayName": "Southlands Campus  (us-central1)",
```
corradio, line 5: Maybe only have one space here? (and same in many other places)

wobniarin: done

corradio (approved):

alejandranavcas (approved):
:)

PR #8161, opened 2025-05-22
[hackaton] solar assets layer

seljaks:


web/src/features/panels/asset/SolarAssetPanel.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  // Hardcoded fallback URL for now - remove this when real URLs are working
+  if (!isValidUrl) {
+    sourceUrl =
+      'https://www.globalenergymonitor.org/projects/global-solar-power-tracker/';
+    isValidUrl = true;
+  }
+
+  const commissionYear = properties.commission_year
+    ? String(Math.floor(Number(properties.commission_year)))
```
VIKTORVAV99, line 108: Same here about the conversion?

Why do we need to floor a year, it should already be an int and can either be turned into a string directly?

web/src/features/panels/asset/SolarAssetPanel.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const possibleUrlFields = Object.entries(properties).filter(
+    ([_, value]) =>
+      typeof value === 'string' &&
+      (value.startsWith('http://') ||
+        value.startsWith('https://') ||
+        value.startsWith('www.'))
+  );
+  console.log('[SolarAssetPanel] Possible URL fields:', possibleUrlFields);
+
+  const handleClose = () => {
```
VIKTORVAV99, line 51: This needs to be wrapped in a useCallback since it's later passed to a component and declared inside a component. Otherwise it will cause performance issues as it gets recreated on every render which cascades down if the reference is not stable.

web/src/features/panels/asset/SolarAssetPanel.tsx
```diff
@@ -0,0 +1,207 @@
...
+      title={name}
+      iconSrc="/images/solar_asset.png"
+      customHeaderStartContent={backButton}
+      contentClassName="p-3 md:p-4"
+    >
+      {/* Content for the solar asset details */}
+      <div className="space-y-3 text-sm text-gray-700 dark:text-gray-300">
+        {!Number.isNaN(capacityMw) && (
+          <div className="flex justify-between font-medium">
+            <span>Capacity:</span>
```
VIKTORVAV99, line 147: Should use a translation key and inline number formatting.

web/src/features/panels/asset/SolarAssetPanel.tsx
```diff
@@ -0,0 +1,207 @@
...
+            <span className="text-gray-900 dark:text-gray-100">
+              {`${capacityMw.toFixed(1)} MW`}
+            </span>
+          </div>
+        )}
+        <div className="flex items-center justify-between">
+          <span>Status:</span>
```
VIKTORVAV99, line 154: Same here with the translation key.

web/src/features/panels/asset/SolarAssetPanel.tsx
```diff
@@ -0,0 +1,207 @@
...
+              {isCommissionedInPast ? 'Operational since:' : 'Commission Year:'}
+            </span>
+            <span className="font-medium text-gray-900 dark:text-gray-100">
+              {commissionYear}
+            </span>
+          </div>
+        )}
+        {capacityUpdateDate && capacityUpdateDate !== 'N/A' && (
+          <div className="flex justify-between">
+            <span>Capacity Data Updated:</span>
```
VIKTORVAV99, line 175: And here.

web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -29,6 +46,50 @@ export const getAllZones = (language: string) => {
...
+            country: feature.properties.country,
+            capacity: feature.properties.capacity_mw
+              ? `${Number.parseFloat(String(feature.properties.capacity_mw)).toFixed(
+                  1
+                )} MW`
+              : undefined,
+            coordinates: feature.geometry.coordinates.slice(0, 2) as [number, number],
+            status: feature.properties.status,
+          };
+        });
```
VIKTORVAV99, line 84: These functions are going to be major performance bottle necks as it's more or less just two loops after each other and creating a massive object.

This should probably be done at build.

web/src/features/panels/search-panel/getSearchData.ts
```diff
@@ -57,3 +118,40 @@ export const getFilteredList = (
...
+  if (!searchTerm) {
+    return [];
+  }
+
+  const filteredZones = getFilteredList(searchTerm, zoneData);
+  const filteredSolarAssets = getFilteredSolarAssets(searchTerm, solarAssets);
+
+  // Combine results - zones first, then solar assets
+  return [...filteredZones, ...filteredSolarAssets];
+};
```
VIKTORVAV99, line 157: The above function should just be updated to include these directly as we are duplicating logic and doing work twice now, as well as unpacking and re-packing two potentially large arrays. If we add more assets like wind turbines in the future this is going to be a real problem.

web/src/features/panels/zone/ZoneDetails.tsx
```diff
@@ -138,72 +125,43 @@ export default function ZoneDetails(): JSX.Element {
...
-      </div>
-    );
-  }
-
   if (zoneDataStatus === ZoneDataStatus.NO_INFORMATION) {
     return <NoInformationMessage />;
   }
+  if (zoneDataStatus === false) {
+    return <NoInformationMessage />;
+  }
```
VIKTORVAV99, line 164: Should be combined with the above check.

web/src/features/panels/InterfacePanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+        'pointer-events-auto z-[21] flex h-full flex-col border-0 transition-all duration-500 sm:inset-3 sm:bottom-[8.5rem] sm:h-auto sm:border sm:pt-0',
+        'pt-[max(2.5rem,env(safe-area-inset-top))] sm:pt-0',
+        className
+      )}
+    >
+      {renderFullHeader
+        ? renderFullHeader()
+        : // Render built-in header only if title is provided (as a proxy for wanting the built-in header)
+          // Or if custom start/end content is provided, as they need a header bar.
+          (title || customHeaderStartContent || customHeaderEndContent) && (
```
VIKTORVAV99, line 56:
```suggestion
      {renderFullHeader?.() ??
          // Render built-in header only if title is provided (as a proxy for wanting the built-in header)
          // Or if custom start/end content is provided, as they need a header bar.
          (title || customHeaderStartContent || customHeaderEndContent) && (
```

web/src/features/panels/InterfacePanel.tsx
```diff
@@ -0,0 +1,118 @@
...
+      >
+        {(() => {
+          if (isLoading) {
+            return <LoadingSpinner />;
+          }
+          if (error) {
+            return <div className="p-3 text-center text-red-500 md:p-4">{error}</div>;
+          }
+          return children;
+        })()}
```
VIKTORVAV99, line 112: Same here about the anonymous/lambda function.

VIKTORVAV99 (commented):
Long review and a lot of comments, sorry about that but it's a big PR...

seljaks:
I've tackled the easier parts of this, the still unresolved stuff seems trickier.

web/src/features/map/Map.tsx
```diff
@@ -467,6 +837,7 @@ export default function MapPage({ onMapLoad }: MapPageProps): ReactElement {
         <SolarLayer />
       </CustomLayer>
       <ZoomControls />
+      <SolarAssetNameTooltip />
```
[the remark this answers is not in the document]
seljaks: I think it already is, conditional logic is inside the element

madsnedergaard:
Closing this as there's no reason to keep hackathon PRs in drafts, we can always revisit history if we decide to bring this back :)

PR #8165, opened 2025-05-22
Fix aws data center locations

cadeban:
## Description
Fix aws data center locations

### Double check


wobniarin (approved):
Good catch!! Thanks for the fix 🥰🥰

PR #8173, opened 2025-05-28
Create MISO grid alert parser

alejandranavcas:
## Description

Creates parser for the grid alerts issued by MISO through [html here](https://www.misoenergy.org/markets-and-operations/notifications/)

### Double check


parsers/US_MISO.py
```diff
@@ -217,11 +222,70 @@ def fetch_wind_solar_forecasts(
...
     # print("fetch_production() ->")
     # print(fetch_production())
 
-    print(fetch_consumption_forecast())
+    # print(fetch_consumption_forecast())
 
     # print("fetch_wind_solar_forecasts() ->")
     # print(fetch_wind_solar_forecasts())
+
+    pprint(fetch_grid_alerts())
```
jbdietrich, line 291: Why are all of the other print statements commented out?

alejandranavcas: Just because I didn't want to output the result of the rest of the parsers, but I can just uncomment them

jbdietrich (approved):

PR #8197, opened 2025-06-02
fix(CY): fix CY wind and conventional column name when parsing html

BastienBigue:
## Issue

CY production data is dropped by feeder-electricity because oil is not parsed correctly anymore. It is also the case of wind. 
This is due to a change in the name of the mode values in the parsed html.

## Description

This PR fixes the mode names. 
I checked that it worked for both real-time and historical url. 

### Preview

Before fix : 
Screenshot 2025-06-02 at 10 39 08

After fix :  
Screenshot 2025-06-02 at 10 38 16

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8199, opened 2025-06-02
fix: move all basicConfig calls into scopes to prevent import sideeff…

seljaks:
One of the calls to .basicConfig was causing duplicated logs in feeder-electricity

This was logging is configured inside each separate endpoint, or if code is used as a library then whoever is using the code is responsible for configuring logging. This makes these files safe to import with a different logging setup


BastienBigue (approved):
LGTM !

PR #8202, opened 2025-06-03
create NYISO grid alerts parser

alejandranavcas:
## Description

Creates parser for NYISO grid alerts

### Preview
Output from running `poetry run test_parser "US-NY-NYIS" gridAlerts`:

```
Parser result:
[{'alertType': <GridAlertType.undefined: 'undefined'>,
  'datetime': datetime.datetime(2025, 6, 4, 4, 0, tzinfo=zoneinfo.ZoneInfo(key='UTC')),
  'endTime': None,
  'issuedTime': datetime.datetime(2025, 6, 4, 4, 0, tzinfo=zoneinfo.ZoneInfo(key='UTC')),
  'locationRegion': None,
  'message': 'Start of day system state is NORMAL',
  'source': 'nyiso.com',
  'startTime': datetime.datetime(2025, 6, 4, 4, 0, tzinfo=zoneinfo.ZoneInfo(key='UTC')),
  'zoneKey': 'US-NY-NYIS'}]
---------------------
took 0.34s
min returned datetime: 2025-06-04 04:00:00+00:00 UTC
max returned datetime: 2025-06-04 04:00:00+00:00 UTC  :( >2h from now !!! (now=2025-06-04T09:36:32+00:00 UTC)
```

### Double check


silkeholmebonnen:
Do we want to fetch the grid alert if the state is normal? To me that does not seem like a "grid alert"

alejandranavcas:
> Do we want to fetch the grid alert if the state is normal? To me that does not seem like a "grid alert"

Fair point. I assume not, I'll change the code to only fetch when the status is different than "Normal"

parsers/US_NY.py
```diff
@@ -308,6 +313,73 @@ def fetch_consumption_forecast(
...
+    for _, notification in csv.iterrows():
+        # Parse and assign EDT timezone
+        dt_edt = datetime.strptime(
+            notification["Timestamp"], "%m/%d/%Y %H:%M:%S"
+        ).replace(tzinfo=TIMEZONE)
+        # Convert to UTC
+        dt_utc = dt_edt.astimezone(ZoneInfo("UTC"))
+
+        # Parse message
+        message = notification["Message"]
```
silkeholmebonnen, line 366: Do we want to trim the message so it does not contain the ** in the beginning and the end? 
If you run `poetry run test_parser US-NY-NYIS gridAlerts --target_datetime "2025-06-06"` you will see the ** in the message

silkeholmebonnen (approved):
LGTM!

PR #8207, opened 2025-06-04
Ireland solar

### LINKED ISSUE #8206: [Data Issue]: Ireland solar data missing
### When did this happen?

-

### What zones are affected?

IE

### What is the problem?

The dashboard that electricity maps pulls its data from for Ireland has recently started reporting solar as a separate metric similar to how it reports wind generation.  
https://www.smartgriddashboard.com/roi/solar/

Would be nice to have it for the odd sunny day we have here.

If it matters, this is for large scale solar farms only, but I suspect this is true for the demand side too.

johnhoran:
## Issue
Closes #8206

## Description
Solar data has been added for eirgrid to https://www.smartgriddashboard.com.  

This PR also switches the URL used as the old URL is no longer used by the dashboard and it doesn't report solar data.  

Additionally I'm aggregating the exchange data for the EWIC and Greenlink interconnectors.  I'm not sure if I'm doing the right thing here, running the original code has it returning 3 lines per time slot, (EWIC, Greenlink and an agregate), but the results in the UI are correct.  Looking at the code I suspect this isn't actually used and the data is pulled from elexon instead. 

### Double check


parsers/SMARTGRIDDASHBOARD.py
```diff
@@ -203,18 +237,30 @@ def fetch_exchange(
...
-        )
+    for time in effective_times:
+        exchanges = [
+            event
+            for event in exchange_data
+            if event["EffectiveTime"] == time
+            and event["FieldName"] in exchange_mapping["exchange"]
+        ]
+
+        net_flow = 0
```
VIKTORVAV99: This is a big no, it will mean that the default value is 0 even if the data is missing which means our backend won't be able to differentiate between missing data and actual 0 exchange data.

So please do not use any default values.

What you should do is create one exchange list for each exchange and then call the ExchangeList.merge_exchanges function with takes a list of ExchangeLists (with matching datetimes only option) which handles things in a uniform way and supports null values gracefully.

I'll be on vacation for the rest of the month but someone from the team should give a final review before I'm back, but might be a bit longer than usual due to vacations and such.

johnhoran: Thanks for the review, I tried to do this using net_flow or None on line 262, but merge_exchanges sounds like a better approach so I'll reimplement using that when I get a chance.

jbdietrich: Thanks @johnhoran! Feel free to ping me when you've had a chance to update and I'll give it another review.

johnhoran: Thanks @jbdietrich, I pushed the requested change so whenever you get a chance.

jbdietrich (approved):

johnhoran:
CI failures look like they are unrelated to this PR.

jbdietrich:
> CI failures look like they are unrelated to this PR.

Interesting, I'll have to take a look.

VIKTORVAV99:
> > CI failures look like they are unrelated to this PR.
> 
> 
> 
> Interesting, I'll have to take a look.

Small tip: Delete all the action caches 😉

PR #8211, opened 2025-06-10
fix(parser): drop data before appending

seljaks:
image
they report short term (stlf) and medium term (mtlf) load on a single time scale. When STFL isn't available we take MTFL. Points between full-hours get flagged as validation errors even though we don't expect the data to be there.


BastienBigue (approved):
LGTM

PR #8212, opened 2025-06-11
fix(parser): GB->GB-NIR filter for datetimes in the future

seljaks:
Parser returns data in full days at 15 minute intervals, so datetimes in the future are always None and spam our log.
Doesn't remove all the errors since sometimes their data is delayed, but in those cases I think we actually should be alerted


BastienBigue (approved):
LGTM

PR #8213, opened 2025-06-11
fix(parser): GB-NIR future datetimes filter

seljaks:
same as https://github.com/electricitymaps/electricitymaps-contrib/pull/8212


BastienBigue (approved):
LGTM

PR #8214, opened 2025-06-11
fix(exchanges): delete FR-COR -> IT exchange

silkeholmebonnen:
## Issue

[FR-COR -> IT exchange does not exist](https://linear.app/electricitymaps/issue/GMM-777/parser-bug-for-fr-cor)

## Description

This PR deletes the exchange FR-COR -> IT, as it does not exist in [entsoe](https://transparency.entsoe.eu/transmission-domain/physicalFlow/show?name=&defaultValue=false&viewType=TABLE&areaType=BORDER_BZN&atch=false&dateTime.dateTime=11.06.2025+00:00%7CCET%7CDAY&border.values=CTY%7C10YIT-GRTN-----B!BZN_BZN%7C10Y1001A1001A885_BZN_BZN%7C10Y1001A1001A74G&dateTime.timezone=CET_CEST&dateTime.timezone_input=CET+(UTC+1)+/+CEST+(UTC+2)) so it fails when we try to fetch it.

### Double check


config/exchanges/FR-COR_IT.yaml
```diff
@@ -1,10 +0,0 @@
-capacity:
-  - -700
-  - 625
-lonlat:
-  - 10.091148
-  - 42.686804
-parsers:
-  exchange: ENTSOE.fetch_exchange
-  exchangeForecast: ENTSOE.fetch_exchange_forecast
```
tonypls, line 9: This can still exist in the aggregated view, let's look into how we can remove the parser but keep the aggregated arrow.

silkeholmebonnen: Now it matches the exchange DE-DK and DK-SE

PR #8215, opened 2025-06-11
fix(AU-TAS): filter data with only 0 solar value

BastienBigue:
## Issue
AU-TAS sometimes reports data with only solar production value. At night, this value is 0. 
For now, our quality pipeline is unable to catch it before it gets picked up by the downstream pipelines. 
This leads to AU-TAS, and AU, to be grey on the map during Australian nights, while we could be using the estimates.

## Description
This PR add filtering of the erroneous data in the parser. 
The current design of the quality pipeline does not allow us to catch this outlier in due time. 

### Double check


tonypls (approved):
LGTM!

PR #8218, opened 2025-06-13
fix(parsers): fix DK-DK1 -> DK-DK2 parser errors

silkeholmebonnen:
## Issue

[Linear issue pointing to the log errors](https://linear.app/electricitymaps/issue/GMM-813/dk1-dk2-exchange-parser)

## Description

This PR makes the if check more explicit so that it does not catch a 0 value.

### Preview

Before:
Screenshot 2025-06-13 at 14 10 05

After:

image

### Double check


silkeholmebonnen:
It looks a bit weird with the "-0" but to keep the code simple I want to change as little as possible

seljaks (approved):
lgtm

PR #8223, opened 2025-06-16
Create email parser

alejandranavcas:
## Issue
[GMM-824](https://linear.app/electricitymaps/issue/GMM-824/create-email-parser)

## Description

Creation of the parser for emails receiving grid alerts. 
Email subscription to ERCOT, CAISO and SPP.

### Double check

### DIFF
--- a/parsers/grid_alerts.py
+++ b/parsers/grid_alerts.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 
 import requests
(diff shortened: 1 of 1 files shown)


parsers/grid_alerts.py
```diff
@@ -0,0 +1,148 @@
...
+
+from requests import Session
+from requests.auth import HTTPBasicAuth
+
+from electricitymap.contrib.config import ZoneKey
+from electricitymap.contrib.lib.models.event_lists import GridAlertList
+from electricitymap.contrib.lib.models.events import GridAlertType
+
+# CONFIGURATION
+MAILGUN_API_KEY = ""  # Replace with your actual Mailgun private API key
```
jbdietrich: How will this get injected at runtime when these things actually run?

alejandranavcas: This still needs to be changed. Probably by getting the token from the environment variables

jbdietrich:
Nice! Should we do some kind of structuring for the file so it's clear it's an email parser? Maybe a folder or something?

alejandranavcas:
> Nice! Should we do some kind of structuring for the file so it's clear it's an email parser? Maybe a folder or something?

What do you suggest @jbdietrich ? A folder like `parsers/emails/mailgun_grid_alerts.py` ? 
If we do not have any other scripts for now, maybe it is enough by giving a more explanatory name to the script like `emails_mailgun_grid_alerts.py` ?

jbdietrich:
Yeah, maybe like `parsers/email/grid_alerts.py`

jbdietrich (approved):

PR #8224, opened 2025-06-16
fix(parsers): replace none check with nan check

seljaks:
previous check wasn't actually doing anything because `nan != None`


BastienBigue (approved):
LGTM

PR #8229, opened 2025-06-18
Patch: data center region inverted coordinates

ryansingman:
## Issue

Fixes issue where some lonlat coordinates are reversed.

### Double check


pierresegonne:
@VIKTORVAV99 Don't understand why this does not get merged?

PR #8230, opened 2025-06-19
feat(config): geojson config

florianscheidl:
## Issue
We want to have access to the country name for a given zone.

## Description
We use the geojson file to derive this.

### DIFF
--- a/electricitymap/contrib/config/__init__.py
+++ b/electricitymap/contrib/config/__init__.py
@@ -6,6 +6,7 @@ from typing import Any
 
 from electricitymap.contrib.config.co2eq_parameters import generate_co2eq_parameters
 from electricitymap.contrib.config.reading import (
+    read_country_given_zone_from_geojson,
     read_data_centers_config,
     read_defaults,
     read_exchanges_config,
@@ -26,6 +27,7 @@ ZONES_CONFIG = read_zones_config(CONFIG_DIR)
 RETIRED_ZONES_CONFIG = read_zones_config(CONFIG_DIR, retired=True)
 EXCHANGES_CONFIG = read_exchanges_config(CONFIG_DIR)
 DATA_CENTERS_CONFIG = read_data_centers_config(CONFIG_DIR)
+ZONE_TO_COUNTRY_LOOKUP = read_country_given_zone_from_geojson()
 
 EU_ZONES = [
     "AT",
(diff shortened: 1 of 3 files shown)


electricitymap/contrib/config/__init__.py
```diff
@@ -26,6 +27,7 @@
 RETIRED_ZONES_CONFIG = read_zones_config(CONFIG_DIR, retired=True)
 EXCHANGES_CONFIG = read_exchanges_config(CONFIG_DIR)
 DATA_CENTERS_CONFIG = read_data_centers_config(CONFIG_DIR)
+ZONE_TO_COUNTRY_LOOKUP = read_country_given_zone_from_geojson()
```
Garsdal: I think to match the other convention this should have a CONFIG_DIR as input parameter here

electricitymap/contrib/config/reading.py
```diff
@@ -51,3 +52,17 @@ def read_data_centers_config(config_dir) -> dict[str, Any]:
     for data_centers in data_centers_config.values():
         all_data_centers.update(data_centers)
     return all_data_centers
+
+
+def read_country_given_zone_from_geojson() -> dict[ZoneKey, str]:
```
Garsdal: Seeing this code made me think, doesn't there exist a python utility function somewhere that reads the geojson?

Garsdal (commented):
I realized that we actually only need a python utility functions which reads the geojson, are we sure that does not exist already?

electricitymap/contrib/config/__init__.py
florianscheidl: It does not read from config though

Garsdal: See my other comment, I just wanted to make sure that we are not adding something that does not match the expected structure. See all of these reading functions have the suffix _CONFIG and depend on a config dir.

The geojson might be different since its a single file - but did we check if there exists a function in python which reads this?

electricitymap/contrib/config/reading.py
florianscheidl: I think it's only read in json file patcher functions, not python.

Garsdal: Couldn't we just import those existing functions from contrib and use them to load geojson object? Then we can filter the object as we want in the lib/config

electricitymap/contrib/config/__init__.py
florianscheidl: As we discussed, it's a bit of a mismatch with the top docstring, however, for simplicity, maybe we still best keep it here for now.

Garsdal (approved):
Nice! Thanks for bearing with my change requests

PR #8237, opened 2025-06-21
Use separate scale for exchange bar chart

Cubikon:
## Description

In countries like Germany with high renewable production (up to 100 GW solar), using the same scale for both the production and exchange graphs makes the exchange chart nearly unreadable, since exchange values rarely exceed 3 GW.

This PR introduces a separate scale for the exchange bar chart to improve visibility and ensure better comparability between zones.

### Preview

Before: 
grafik
After:
grafik

### Double check

### DIFF
--- a/web/src/features/charts/bar-breakdown/BarElectricityBreakdownChart.tsx
+++ b/web/src/features/charts/bar-breakdown/BarElectricityBreakdownChart.tsx
@@ -56,15 +56,14 @@ function BarElectricityBreakdownChart({
 
   // Use the whole history to determine the min/max values in order to avoid
   // graph jumping while sliding through the time range.
-  const [minPower, maxPower] = useMemo(
+  const [minProductionPower, maxProductionPower] = useMemo(
     () => [
       d3Min(
         Object.values(data.zoneStates).map((zoneData) =>
           Math.min(
             -zoneData.maxStorageCapacity || 0,
             -zoneData.maxStorage || 0,
-            -zoneData.maxExport || 0,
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
I won't be reviewing this right now at least and will leave that up to the rest of the team (or until I'm back from my vacation) but I'd like to add that it's generally bad practice to have two different scales in one graph, which despite being separate in code, this is visually.

This also don't really fit with our long term design visions. It really would have been great if a issue was opened first where this could have been discussed as it's not something that can just be approved and merged without consideration.

Cubikon:
Ok, sorry for not opening an issue first. I thought it wasn’t necessary for such a small change.

About the two different scales in one graph – I think it's actually the other way around. There are now two graphs with different data and separate scales, but the same maximum values for those scales. Since the data differs so much, using the same scale isn't practical for both.

I believe a graph should primarily serve the purpose of providing a clear visualization of the data – in this case, the exchange rates.

But im totally fine if this doesn’t fit with the long-term vision – I'm happy to leave it for now and let the team decide.

Thanks for your time, and enjoy your vacation.

madsnedergaard:
Hey @Cubikon, thanks for sharing the idea here :)
I can see pros and cons to both ways, but for now I think it's going to cause more confusion (it's easy to miss the different scale), and secondly I think there's a great benefit to seeing the relevant "size" of exports/imports compared to a countries own production.

We'll keep this idea and PR in mind when we at some point revisit this area - until then I'll close it, but we really appreciate getting feedback and ideas like this one 🙏

PR #8245, opened 2025-06-23
Feat/Co2 intensity filtering

kumarabinash:
## Issue
Not an issue. A feature suggestion.

## Description
This PR aims to add a feature which allows showing Co2 Intensity on the map for a selected range. Further documentation and details can be found below:
https://kumarabinash.atlassian.net/wiki/external/YmFlZmRmZWMyNmZhNDU2YWEyNThkMDkwODU5ZmE4OTU

This PR also fixes the following:
- Fixed svg alignment issue on ColorBar.tsx

### Feature Flag Backed

Feature Flag Name: `legend-co2-intensity-filtering`

To test it out without the feature flag, add the following temporary code after line 11 on `web/src/features/feature-flags/api.ts` & make sure to remove after test

```ts
if (name === 'legend-co2-intensity-filtering') {
  return true; // todo: Remove before commit
}
```

#### New Component (RangeSlider)
Screenshot 2025-06-23 at 12 34 10 PM

#### Usage of Range Slider & Preview of the feature
Feat 2

### Double check

### DIFF
--- a/web/src/components/RangeSlider.tsx
+++ b/web/src/components/RangeSlider.tsx
@@ -1,6 +1,6 @@
 import * as SliderPrimitive from '@radix-ui/react-slider';
 import * as Tooltip from '@radix-ui/react-tooltip';
-import { Menu } from 'lucide-react';
+import { ChevronFirst, ChevronLast } from 'lucide-react';
 import { memo, type ReactElement } from 'react';
 
 import LabelTooltip from './tooltips/LabelTooltip';
@@ -51,7 +51,7 @@ function RangeSlider({
             className="-ml-2.5 -mt-1.5 flex h-5 w-5 items-center justify-center rounded-full bg-white p-1 outline
            outline-1 outline-neutral-200 hover:outline-2 focus-visible:outline-2 focus-visible:outline-brand-green dark:bg-neutral-900 dark:outline-neutral-700 dark:focus-visible:outline-brand-green"
           >
-            <Menu className="rotate-90" size={12} />
+            <ChevronFirst strokeWidth={3} size={12} pointerEvents="none" />
           </SliderPrimitive.Thumb>
         </TooltipWrapper>
 
@@ -65,7 +65,7 @@ function RangeSlider({
             className="-mr-2.5 -mt-1.5 flex h-5 w-5 items-center justify-center rounded-full bg-white p-1 outline
            outline-1 outline-neutral-200 hover:outline-2 focus-visible:outline-2 focus-visible:outline-brand-green dark:bg-neutral-900 dark:outline-neutral-700 dark:focus-visible:outline-brand-green"
           >
-            <Menu className="rotate-90" size={12} />
+            <ChevronLast strokeWidth={3} size={12} pointerEvents="none" />
           </SliderPrimitive.Thumb>
         </TooltipWrapper>
       </SliderPrimitive.Root>
--- a/web/src/components/legend/Co2Legend.tsx
+++ b/web/src/components/legend/Co2Legend.tsx
@@ -2,7 +2,7 @@ import { extent } from 'd3-array';
 import { scaleLinear } from 'd3-scale';
 import { useCo2ColorScale } from 'hooks/theme';
 import { useAtom } from 'jotai/index';
-import { memo, type ReactElement } from 'react';
+import { memo, type ReactElement, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { CarbonUnits } from 'utils/units';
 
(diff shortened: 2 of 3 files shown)


VIKTORVAV99:
cc @jbdietrich

web/src/utils/helpers.ts
```diff
@@ -187,6 +187,30 @@ export const getCarbonIntensity = (
...
+  isFeatureEnabled: boolean,
+  co2intensity: number | undefined,
+  min: number,
+  max: number
+): number | undefined => {
+  if (!isFeatureEnabled) {
+    return co2intensity;
+  }
+
+  if (co2intensity && (co2intensity < min || co2intensity > max)) {
```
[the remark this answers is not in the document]
VIKTORVAV99: Since we don't have any negative values and the lowest value for carbon intensity is 4.1 this should be fine. So no need to change this.

web/src/utils/state/atoms.ts
```diff
@@ -20,6 +20,9 @@ import {
 
 export const timeRangeAtom = atom<TimeRange>(TimeRange.H72);
 
+// TODO: Maintain constants for lower and upper limits for co2 intensity and use from there
+export const co2IntensityRangeAtom = atom([0, 1500]);
```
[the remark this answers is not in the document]
VIKTORVAV99: The only other place this is used is in the color scale so this should be fine.

web/src/components/legend/Co2Legend.tsx
```diff
@@ -1,20 +1,61 @@
...
   return (
     <LegendItem
       label={t('legends.carbonintensity')}
       unit={CarbonUnits.GRAMS_CO2EQ_PER_KILOWATT_HOUR}
     >
-      <HorizontalColorbar colorScale={co2ColorScale} ticksCount={6} id={'co2'} />
+      {isCo2IntensityFilteringFeatureEnabled ? (
+        <RangeSlider
+          value={co2IntensityRange}
+          defaultValue={[rangeValues.at(0) as number, rangeValues.at(-1) as number]}
```
VIKTORVAV99: Small question? Since this is wrapped in a array won't it cause the component to re-render since the object reference keeps changing?
I.e should we extract it to a const and memoize it?

web/src/components/RangeSlider.tsx
```diff
@@ -0,0 +1,76 @@
...
+          tooltipContent={<LabelTooltip className={'text-xs'}>{value[0]}</LabelTooltip>}
+          side="top"
+          sideOffset={4}
+        >
+          <SliderPrimitive.Thumb
+            data-testid="range-slider-input-1"
+            className="-ml-2.5 -mt-1.5 flex h-5 w-5 items-center justify-center rounded-full bg-white p-1 outline
+           outline-1 outline-neutral-200 hover:outline-2 focus-visible:outline-2 focus-visible:outline-brand-green dark:bg-neutral-900 dark:outline-neutral-700 dark:focus-visible:outline-brand-green"
+          >
+            <Menu className="rotate-90" size={12} />
```
VIKTORVAV99: We would like to change this to use `chevron-first` and the other one to use `chevron-last` just to indicate these have hard stops at the end of the ranges.

web/src/components/legend/Co2Legend.tsx
kumarabinash: You're right 💪 This would create a new obj reference on every render, it should be a memoized const.
I'll add a fix for that before resolving the comment.

web/src/components/RangeSlider.tsx
kumarabinash: Got it! I'll update the component with chevron icons and add some screenshots to the PR for further review.

kumarabinash:
### About the latest commit:

#### Fixes 2 review comment

- Updated Icon
    Screenshot 2025-07-01 at 7 45 00 PM

    Screenshot 2025-07-01 at 7 51 39 PM

- Moved suggested value to a const & memoized it along with another value.

- Fixed one alignment issue (Verified this across all its usages)

     Screenshot 2025-07-01 at 8 08 08 PM

web/src/components/RangeSlider.tsx
kumarabinash: `chevron-first` and `chevron-last` look much better 🚀 

Please note the stokeWidth added to the code is intentional. This makes it look consistent with the timeline chevrons.

Without stroke
Screenshot 2025-07-01 at 7 45 27 PM

With Stroke
Screenshot 2025-07-01 at 7 45 00 PM

VIKTORVAV99 (approved):
LGTM! 🎉

Thanks for fixing it up!

kumarabinash:
Checklist:

- [x] Create a new feature flag if not already created `legend-co2-intensity-filtering` and enable it
- [ ] Optional draft release notes for internal / external communication

>You can now filter electricity data on the map by CO2 intensity ranges directly from the legend. This new filtering capability allows you to focus on specific consumption / production levels and visualize only the data that matters most to your analyses.

cc: @VIKTORVAV99

VIKTORVAV99:
> Checklist:
> 
> * [x]  Create a new feature flag if not already created `legend-co2-intensity-filtering` and enable it
> * [ ]  Optional draft release notes for internal / external communication
> 
> > You can now filter electricity data on the map by CO2 intensity ranges directly from the legend. This new filtering capability allows you to focus on specific consumption / production levels and visualize only the data that matters most to your analyses.
> 
> cc: @VIKTORVAV99

Just created the feature flag!
You can test it here if you want: https://staging.electricitymaps.com/zone/SE-SE3/72h/hourly?ff=

(just don't look to closely at our hacky feature flag panel, it needs some TLC...)

Release notes we usually send out internally on Fridays so I'd be happy to write up something for that. I'll also add some for the next release we create on GitHub that includes this feature.

PR #8246, opened 2025-06-24
feat(app-frontend): Visualize alerts on map

alejandranavcas:
## Issue

[GMM-838](https://linear.app/electricitymaps/issue/GMM-838/visualize-alerts-on-map)

## Description

Visualize the grid alert on the left panel of the map, implementing [this design](https://www.figma.com/design/O1DtS6UyZHSH4IymG8Ufsw/🎹-Design-Hack?node-id=4213-13685&p=f&m=dev). Not ALL details of the design are implemented yet. But the core functionalities yes: the icon on the zones where there is an "action" type alert and the content of the latest alert for a zone in a card on the left panel.

#### Alerts card on left panel
Screenshot 2025-06-30 at 16 11 12

#### Screenshot of the alert icon on the map for zones with an "action" alert type.
Screenshot 2025-06-30 at 15 30 39

### Double check


jbdietrich:
A few things I noticed while testing that might block a merge:

1. It's not clear how the grid alert relates to the time slider. This wasn't an explicit requirement of this version but it's a strange experience to slide back in time and have a grid alert for the current day show. It might be cool to be able to slide back in time and see the alerts that existed at the time you select (or maybe just remove them entirely if the time window is outside of the past)?

2. Timezones are tricky. Grid alerts are ingested as if they are UTC but each ISO issues them in a local timezone. We probably want to handle this in the data model.

3. We may need to pre-process the raw data from e.g. emails to format them nicely. We are basically dumping the raw text including footers and signatures into the Grid Alert message.

PaulRoms:
@VIKTORVAV99 -> Is it a good candidate for the cooldown week?

VIKTORVAV99:
@PaulRoms yes and no, @silke.bonnen is currently already working on this in the [Grid Alerts](https://linear.app/electricitymaps/project/grid-alerts-9bfc9e840141) project so I think we are already have that covered for now. We can of course assist if needed or if we can help speed things up!

VIKTORVAV99:
Closed in favor of #8281 that builds on top of this.

PR #8247, opened 2025-06-25
feat(parsers): change source for some IN parsers

BastienBigue:
## Issue
The following parsers are querying (https://www.wrldc.in/ and https://app.erldc.in/) that have been replaced by https://grid-india.in (to be accessed with a proxy) : 
IN-SO->IN-WE
IN-NO->IN-WE
IN-EA->IN-WE
BT->IN-EA

## Description

This PR uses the new source.
WIP : 
parsing of the PDF containing interregional exchanges is working

### Double check


annaib00:
The proxy currently does not work. We can fetch the data using a VPN, but when using either datacenter or residential proxies we cannot reach the data.

VIKTORVAV99 (commented):
There is a problem with the old IN_EA test file which no longer work after rewriting the parser. We should probably delete this and create a new test.

This is why the CI is not passing.

PR #8272, opened 2025-07-03
EIA capacity update

VIKTORVAV99:
## Issue

There have been some changes.

## Description

Updates the capacity for the EIA zones.

### DIFF
--- a/capacity_update.py
+++ b/capacity_update.py
@@ -37,7 +37,7 @@ def capacity_update(
     >>> poetry run capacity_update --source ENTSOE --target_datetime "2022-01-01"
     """
     logging.basicConfig(
-        level=logging.debug,
+        level=logging.DEBUG,
         format="%(asctime)s %(levelname)-8s %(name)-30s %(message)s",
     )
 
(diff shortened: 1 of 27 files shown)


capacity_update.py
```diff
@@ -37,7 +37,7 @@ def capacity_update(
     >>> poetry run capacity_update --source ENTSOE --target_datetime "2022-01-01"
     """
     logging.basicConfig(
-        level=logging.debug,
+        level=logging.DEBUG,
```
VIKTORVAV99, line 40: Not sure why this needed changing but things failed otherwise.

config/zones/US-CAR-DUK.yaml
```diff
@@ -40,6 +43,9 @@ capacity:
     - datetime: '2024-01-01'
       source: EIA.gov
       value: 1059.5
+    - datetime: '2025-03-01'
+      source: EIA.gov
+      value: 2590.0
```
VIKTORVAV99, line 48: Lots more capacity here 👀

capacity_update.py
seljaks: all caps is the level enum, lower case is a function like logger.debug that goes directly to the root logger.
don't dig too deep into the logging module, it's all pain down there

seljaks (approved):
lgtm

PR #8275, opened 2025-07-08
[Capacity]: Adjust BD wind capacity

VIKTORVAV99:
## Issue

It's a bit hard to get a firm date on when they actually started producing energy at the first wind farm in BD. But this source has when it connected to the grid which I think we can use. That should help get rid of some outliers.

But we still have some production from before this date, roughly for 4 months. Not really sure what we should do about that. This production currently get flagged since the reported capacity from Ember is 0, this obviously is not correct as we have production so I'm kind of leaning towards removing that capacity entry and allowing them through?

### DIFF
--- a/config/zones/BD.yaml
+++ b/config/zones/BD.yaml
@@ -104,9 +104,6 @@ capacity:
       source: Ember, Yearly electricity data
       value: 850.0
   wind:
-    - datetime: '2017-01-01'
-      source: Ember, Yearly electricity data
-      value: 0.0
     - _comment: https://en.powerchina.cn/2023-10/13/c_828590.htm
       datetime: '2023-10-12'
       source: POWERCHINA


config/zones/BD.yaml
```diff
@@ -107,10 +107,10 @@ capacity:
     - datetime: '2017-01-01'
       source: Ember, Yearly electricity data
       value: 0.0
```
VIKTORVAV99: Should we remove this?

Or move up the other start date?

We currently have production before 2023-10-12 as well that get's flagged because this is 0.

Alternatively we could adjust the rule so we can still add a deviation when the capacity is 0 but then the logic gets a little twisted.

seljaks:
I'm plus one on removing the capacity entry and allowing the 4 months of production data to go through

VIKTORVAV99:
> I'm plus one on removing the capacity entry and allowing the 4 months of production data to go through

I'll do that then and merge this afterwards.

PR #8277, opened 2025-07-09
Add updated YUKONENERGY for CA-YT

annaib00:
## Issue 
Update / add parser for CA-YT. 

## Description
There was a parser before but it was not in use and outdated. It has now been updated. The data is fetched directly from the html response. We get data for the past 24 hours (hourly) and for the past 5 minutes. 

### Double check


VIKTORVAV99 (commented):
Code wise I don't see any issues, but this is a GPZD zone and I don't know what the protocol is for how and what to do when we switch to live data again.

Would be good to get @PaulRoms, @seljaks and @BastienBigue input on this as I'm afraid we might overwrite all the GPZD data we have with TSA if we just switch it outright.

parsers/YUKONENERGY.py
```diff
@@ -40,80 +34,85 @@ def fetch_production(
...
-    See schema of the grid at http://www.atcoelectricyukon.com/About-Us/
+    # Extract JavaScript data
+    match_past_day = re.search(
+        r"var rows_chart_past_day = parseDates\((\[\[.*?\]\])\);", html, re.DOTALL
+    )
+    match_current = re.search(
+        r"var rows_chart_current = parseDates\((\[\[.*?\]\])\);", html, re.DOTALL
+    )
+    if not match_past_day or not match_current:
+        raise ParserException("CA_YT.py", "Cannot find data", zone_key)
```
jarek: Probably should be the actual filename? 

```suggestion
        raise ParserException("YUKONENERGY.py", "Cannot find data", zone_key)
```

Also a couple other spots in this file. I assume this was missed when the file was renamed.

jarek (commented):
Sad to see the documenting comments go :thinking:

parsers/YUKONENERGY.py
```diff
@@ -17,20 +19,12 @@
...
-    try:
-        return float(text[: text.index(" MW")])
-    except ValueError as e:
-        raise ParserException(
-            "YUKONENERGY.py",
-            f"Unable to parse power value from '{text}'",
-            ZONE_KEY,
-        ) from e
+FUEL_MAPPING = {
+    "thermal": "unknown",
```
PaulRoms, line 23: According to CER, Thermal in CA-YT is oil. 
you can check it via the table parser_data_production_low_frequency or the website: https://apps2.cer-rec.gc.ca/energy-future/?page=electricity&mainSelection=electricityGeneration&yearId=2023&sector=&unit=megawatts&view=source&baseYear=2036&compareYear=2050&noCompare=&priceSource=&scenarios=Global%20Net-zero&provinces=YT&provinceOrder=YT,SK,QC,PE,ON,NU,NT,NS,NL,NB,MB,BC,AB&sources=ALL&sourceOrder=OIL,BIO,BIOCCUS,COAL,COALCCUS,GAS,GASCCUS,WIND,WINDOFF,SOLAR,SOLARUSC,NUCLEAR,NUCLEARSMR,HYDRO,HYDROGEN,BATTERY

PaulRoms:
Is there any historical data available for this parser? 

Regarding GPZD and TSA, you are right @VIKTORVAV99 that is might not work as expected. The idea solution, if we don't have historical data would be to enable GPZD until a certain point and then the TSA takes from here

VIKTORVAV99:
> Is there any historical data available for this parser? 
> 
> 
> 
> Regarding GPZD and TSA, you are right @VIKTORVAV99 that is might not work as expected. The idea solution, if we don't have historical data would be to enable GPZD until a certain point and then the TSA takes from here

Only the last 24 hours, so we can't backfill any data.

Currently I don't think the estimator has a way of configuring GPZD only for some duration so it's a all or nothing approach right now.

We'd have to change how the estimator is configured if we wanted to change that. It would probably be a good change as there are other zones we could enable it for ass well that only have partial measured data.

PaulRoms:
> > Is there any historical data available for this parser?
> > Regarding GPZD and TSA, you are right @VIKTORVAV99 that is might not work as expected. The idea solution, if we don't have historical data would be to enable GPZD until a certain point and then the TSA takes from here
> 
> Only the last 24 hours, so we can't backfill any data.
> 
> Currently I don't think the estimator has a way of configuring GPZD only for some duration so it's a all or nothing approach right now.
> 
> We'd have to change how the estimator is configured if we wanted to change that. It would probably be a good change as there are other zones we could enable it for ass well that only have partial measured data.

Agree with you! Let me ask the team :)

parsers/YUKONENERGY.py
annaib00: Yes, the filenames should be changed! Some of the documenting comments were outdated, but we can keep some of the not outdated ones

parsers/YUKONENERGY.py
jarek: > According to CER, Thermal in CA-YT is oil.

This is wrong, there has also been LNG generation in Whitehorse since 2015. https://www.cer-rec.gc.ca/en/data-analysis/energy-markets/provincial-territorial-energy-profiles/provincial-territorial-energy-profiles-yukon.html

annaib00: Yes, we keeping it as unknown.

annaib00:
If we add TSA to the yaml under estimation_models I guess we can start getting parser data from this zone instead of GPZD ?

VIKTORVAV99:
Yeah. I would merge this first though so we have some measured data to base the initial TSA estimations on.

VIKTORVAV99 (approved):
LGTM! 🎉

PR #8278, opened 2025-07-10
Switch MD to use the ENTSOE parser

VIKTORVAV99:
## Issue

The current parser only get gas, hydro and renewable and assigns all renewable to biomass. This is a problem since most of the renewable is actually solar. This causes the quality validator to flag almost all of the events during the day.

## Description

This PR switches it to use the ENTSO-E parser which has the following production modes:
- Biomass
- Coal
- Gas
- Hydro
- Oil
- Solar
- Wind

So this is a clear improvement in the quality of the data we get.
However the catch is that this data is hourly while the current Moldova (MD) data is 15 minute. So this means we'll have to run a database migration after we have merged this and refetched all the data to remove the data from the old Moldova source. 

### Double check


BastienBigue:
LGTM ! Did you check that we had the same historical depth with ENSTOE than we have with the MD source ? 
We should also update the changelog to reflect this change to our users https://docs.google.com/spreadsheets/d/15ELdAhEQDQxuwlefZE6X7blrrRNwTUuDm-B1GMIxSvg/edit?gid=279914678#gid=279914678

VIKTORVAV99:
> LGTM ! Did you check that we had the same historical depth with ENSTOE than we have with the MD source ? 
> 
> We should also update the changelog to reflect this change to our users https://docs.google.com/spreadsheets/d/15ELdAhEQDQxuwlefZE6X7blrrRNwTUuDm-B1GMIxSvg/edit?gid=279914678#gid=279914678

I checked back to 2020, and we have data at least until then. Ill do a more exact one when I write the database migration so we don't lose any data.

VIKTORVAV99:
2020-01-15 seems to be the earliest ENTSO-E has any data for Moldova. Our current source go further back but I don't see the same issue with the earlier data as we do now, likely because they didn't have any wind or solar installed at the time. So this shouldn't cause any issues.

seljaks (approved):
great find

VIKTORVAV99:
I'll merge this when I have the migration ready. But to be on the safe side I might not do anything until Monday.

PR #8279, opened 2025-07-10
fix(parser): set timezone in grid alert parsers

silkeholmebonnen:
## Issue

[GMM-871](https://linear.app/electricitymaps/issue/GMM-871/modify-parsers-to-take-local-time-zone-into-account)

## Description

This PR sets the timezone in the datetime fields to be the local timezone. This seems to be the only grid alert parser that didn't take the local timezone into account.

### Double check

### DIFF
--- a/parsers/US_MISO.py
+++ b/parsers/US_MISO.py
@@ -264,7 +264,7 @@ def fetch_grid_alerts(
     for notification in notifications:
         publish_datetime = datetime.fromisoformat(
             notification["publishDateUnformatted"]
-        )  # in UTC
+        ).replace(tzinfo=TIMEZONE)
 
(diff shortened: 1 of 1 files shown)


jbdietrich (approved):

jbdietrich:
There may still be an issue with the time in the body of the message. In other words: as a viewer of the grid alert on the app, if the body of the message contains a time (e.g. "This grid alert is valid from 12:00", it's not clear which 12:00 that refers to). Here again the solution may have to be some kind of summary layer that can make structured data out of the unstructured text.

VIKTORVAV99 (approved):
Agree with James but that's for another PR.

silkeholmebonnen:
Also agree that this would be a good next step for the grid alerts project

PR #8281, opened 2025-07-11
feat(web): add grid alerts to app

silkeholmebonnen:
## Issue

[GMM-870](https://linear.app/electricitymaps/issue/GMM-870/only-show-grid-alert-on-the-map-if-it-is-currently-active)

## Description

This PR builds on top of [Alejandra's PR](https://github.com/electricitymaps/electricitymaps-contrib/pull/8246/files#diff-90f5cac94f449081ecfebea81e03ccab5a7ffed7850333bf9c5579f919474dc1). It shows grid alerts on the map but only for on hourly or 5 minute view, and only the hours that it is active.

### Preview

image

image

### Double check


web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <FiAlertTriangle
+                size={16}
+                className={`-translate-y-px text-${iconColor}`}
+              />
+            </div>
+          </Marker>
+        );
+      })}
+    </>
+  );
```
silkeholmebonnen: Alejandra implemented this and I can see that it is different from the other map layers but it seems to work. From a little chat with AI it also seemed to be easier to use Marker when rendering Icons

web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    x += (xi + xi1) * a;
+    y += (yi + yi1) * a;
+  }
+
+  area *= 0.5;
+  x /= 6 * area;
+  y /= 6 * area;
+
+  return [x, y];
+};
```
VIKTORVAV99, line 37: There are helpers from turf that can do this if we wish (same library we use to compile the map) but this works!

web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+function getWarningIconData(worldGeometries: MapGeometries, data: GridState | undefined) {
+  if (!worldGeometries?.features) {
+    return null;
+  }
+
+  if (!data || !data.alerts) {
+    return null;
+  }
```
VIKTORVAV99: Since both return null this can be one statement.

web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+      } else if (zoneFeature.geometry.type === 'MultiPolygon') {
+        // Use the centroid of the largest polygon
+        let largest = zoneFeature.geometry.coordinates[0];
+        for (let index = 1; index < zoneFeature.geometry.coordinates.length; index++) {
+          if (zoneFeature.geometry.coordinates[index][0].length > largest[0].length) {
+            largest = zoneFeature.geometry.coordinates[index];
+          }
+        }
+        centroid = getPolygonCentroid(largest);
+      }
```
VIKTORVAV99, line 66: This is potentially a very costly function to run performance wise so it should probably be turned into a hook and memoized so it only needs to run when the data changes.

EDIT: I see this is memoized later using useMemo :)

web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+export default function GridAlertsLayer() {
+  const { worldGeometries } = useGetGeometries();
+  const { data } = useGetState();
+
+  const co2ColorScale = useCo2ColorScale();
+
+  const warningIconData = useMemo(
+    () => getWarningIconData(worldGeometries, data),
+    [worldGeometries, data]
+  );
```
VIKTORVAV99, line 98: This works too :)

web/src/features/map/map-layers/GridAlertsLayer.tsx
VIKTORVAV99: Yeah I think markers are the way to go here.

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+import HorizontalDivider from 'components/HorizontalDivider';
+import { FormattedTime } from 'components/Time';
+import LabelTooltip from 'components/tooltips/LabelTooltip';
+import TooltipWrapper from 'components/tooltips/TooltipWrapper';
+import { RoundedCard } from 'features/charts/RoundedCard';
+import { useAtomValue } from 'jotai';
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { FiAlertTriangle, FiInfo } from 'react-icons/fi';
+import { IoFlaskOutline } from 'react-icons/io5';
```
VIKTORVAV99: Should these not use icons from Lucide so they match the rest of the icons we use?

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const startTime = new Date(zoneMessage.start_time ?? '');
+  if (selectedTimeRange === TimeRange.H72) {
+    startTime.setMinutes(0, 0, 0);
+  } else {
+    // set minutes to the closest 5 minutes in the past
+    const minutes = startTime.getMinutes();
+    const closest5Minutes = Math.round(minutes / 5) * 5;
+    startTime.setMinutes(closest5Minutes, 0, 0);
+  }
```
VIKTORVAV99, line 60: There are helpers from date-fns that can help floor, set and get the minutes in a easy way if you wish to use that.

This is pretty simple operations though so this is probably fine.

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <div className="px-4 py-3 pb-2">
+        <div className="flex flex-row items-center gap-1">
+          {icon}{' '}
+          <span className={`line-clamp-1 font-semibold ${colorClass}`} title={title}>
+            {title}
+          </span>
+        </div>
+        <div className="flex flex-row items-center gap-1 text-xs">
+          <FormattedTime
+            datetime={new Date(zoneMessage?.start_time ?? '')}
```
VIKTORVAV99: You can reuse the start time variable from before here.

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+          />
+          {zoneMessage?.end_time && (
+            <div className="flex flex-row items-center gap-1">
+              <div>-</div>
+              <FormattedTime
+                datetime={new Date(zoneMessage.end_time ?? '')}
+                language={i18n.languages[0]}
+                timeRange={TimeRange.H72}
+              />
+            </div>
```
VIKTORVAV99: This should probably be a time range as different regions and languages uses different separators. I think we have a helper for that.

This is probably fine as a first iteration if you are pressed for time though.

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <HorizontalDivider />
+        {!isLongMessage && <div className="font-normal">{parseTextWithLinks(text)}</div>}
+        {isLongMessage && (
+          <div className="flex flex-col gap-1">
+            <div className={`text-wrap font-normal ${isCollapsed ? 'line-clamp-2' : ''}`}>
+              {parseTextWithLinks(text)}
+            </div>
+            <div className="items-left flex flex-row gap-1">
+              <button
+                onClick={() => setIsCollapsed(!isCollapsed)}
```
VIKTORVAV99: There is a big no no to have anonymous functions inside the components as they will be re-created every time and make the component unstable.

This should be declared further up before the return statement and wrapped in either a useCallback or useMemo hook.

VIKTORVAV99 (commented):
Some comments, most are just suggestions.

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
silkeholmebonnen: Okay I will just leave it as is then

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
silkeholmebonnen: No because I modify the minutes so it is rounded to nearest hour or 5 minutes

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
silkeholmebonnen: I can do it but the helper function (`formatDateRange`) only formats to day and not hour, so I cannot use it

web/src/features/map/map-layers/GridAlertsLayer.tsx
silkeholmebonnen: I will just leave it as it since it works

silkeholmebonnen:
I noticed that it keeps rerendering when hovering over the map alert icon. I will try to figure out why and solve it


https://github.com/user-attachments/assets/ca55fd3e-f521-4f4f-9948-4b29f1cc57e3

silkeholmebonnen:
> I noticed that it keeps rerendering when hovering over the map alert icon. I will try to figure out why and solve it

With help from AI, I found a way to avoid flickering. I'm not sure it is the best way, so please let me know if you know a better way to do it

web/src/features/panels/zone/CurrentGridAlertsCard.tsx
VIKTORVAV99: Ah right, JS/TS constants that are not always constants...

But should not the start date that is used and the one that is shown be in sync?

silkeholmebonnen: It is just to include it at the current hour. Lets say the grid alert is at 4:15, then I want it to be shown when timer slider is at 4 but if I do not floor the start time when comparing then this doesn't work. But at the same time I still want to show 4:15 as the time stamp on the alert

`
if (selectedDatetime.datetime < startTime) {
    return null;
  }`

web/src/features/map/map-layers/GridAlertsLayer.tsx
```diff
@@ -113,7 +113,10 @@ export default function GridAlertsLayer() {
 
         return (
           <Marker key={zoneId} longitude={longitude} latitude={latitude} anchor="center">
-            <div className="flex items-center justify-center rounded-full bg-white/10 p-2">
+            <div
+              className="flex items-center justify-center rounded-full bg-white/10 p-2"
+              style={{ pointerEvents: 'none' }}
```
VIKTORVAV99: I think you might need the pointer events none on the marker instead.
I should accept custom classes so just add the `pointer-events-none` tailwind class.

web/src/features/map/MapTooltip.tsx
```diff
@@ -109,14 +109,40 @@ export default function MapTooltip() {
   const isMapMoving = useAtomValue(mapMovingAtom);
   const { data } = useGetState();
 
-  if (!hoveredZone || isMapMoving) {
+  // Debounce logic to prevent flicker
+  const [visibleZoneId, setVisibleZoneId] = useState<string | null>(null);
+  const hideTimeout = useRef<NodeJS.Timeout | null>(null);
```
VIKTORVAV99: We cant use NodeJS.Timeout type as it's not going to run in node.

VIKTORVAV99 (commented):
Seems like a lot of logic for something that should not be a concern and should just work with CSS as the markers should not be interactive.

`pointer-events-none` should do the trick but it might need to be moved further up in the chain, potentially to the entire layer.

web/src/features/map/map-layers/GridAlertsLayer.tsx
silkeholmebonnen: You're right:)

silkeholmebonnen:
@VIKTORVAV99 There is an active action alert right now so it is really easy to test:)

VIKTORVAV99:
> @VIKTORVAV99 There is an active action alert right now so it is really easy to test:)

Nice! Looks like it's working!

web/src/locales/en.json
```diff
@@ -157,6 +157,12 @@
       "now": "Now"
     }
   },
+  "grid-alerts-card": {
+    "show-more": "Show more",
+    "show-less": "Show less",
+    "experimental-mode": "This is free while in experimental mode",
```
VIKTORVAV99:
```suggestion
    "experimental-mode": "This is free while in the experimental phase",
```

web/src/locales/en.json
```diff
@@ -157,6 +157,12 @@
...
+    "tooltip": "This feature is currently in experimental phase and available at no additional cost while we gather feedback and refine the functionality."
```
VIKTORVAV99:
```suggestion
    "tooltip": "This feature is currently in an experimental phase and available at no additional cost while we gather feedback and refine the functionality."
```

PR #8287, opened 2025-07-15
5 min disclaimer

VIKTORVAV99:
## Description
Adds a card for the 5 min feature that describes that it is experimental.

Also threw in some re-render fixes 🙂

Figma reference: https://www.figma.com/design/O1DtS6UyZHSH4IymG8Ufsw/%F0%9F%8E%B9-Design-Hack?node-id=7639-618&t=Bm945NmSvH7GcG7D-0

### Preview

image
image

### Double check

### DIFF
--- a/web/src/features/panels/zone/ExperimentalCard.tsx
+++ b/web/src/features/panels/zone/ExperimentalCard.tsx
@@ -1,6 +1,7 @@
 import { RoundedCard } from 'features/charts/RoundedCard';
 import { FlaskConicalIcon } from 'lucide-react';
-import { memo } from 'react';
+import { memo, useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
 
 export interface ExperimentalCardProps {
   title: string;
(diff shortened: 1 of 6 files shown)


web/src/features/panels/zone/ExperimentalCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+  title: string;
+  description?: string;
+}
+
+function ExperimentalCard({ title, description }: ExperimentalCardProps) {
+  const { t } = useTranslation();
+  const supportsIntercom = typeof window !== 'undefined' && Boolean(window.Intercom);
+  const handleChat = useCallback(() => {
+    if (window.Intercom) {
+      window.Intercom('show');
```
madsnedergaard, line 16: I cannot actually write a message 🤔 
Dunno if we have disabled messages by accident, any idea @cadeban?

Screenshot 2025-07-16 at 15 53 47

madsnedergaard (approved):
LGTM! Works great :)

web/src/features/panels/zone/ExperimentalCard.tsx
VIKTORVAV99: According to @Alportan you need to be signed in. Idk how that whole part works but it looks like it might be an oversight? 😅

madsnedergaard: It wasn't like that previously so I think someone made a change in Intercom settings that accidentally disables it?

VIKTORVAV99: I don't know about that, that's outside my domain now. But it should be fine to merge this right?

madsnedergaard: Yes, given it's still feature-flagged we can merge this. But I don't think we should roll out before we have it fixed, otherwise we ask users to "leave feedback" that they can't 😅

VIKTORVAV99: We could just skip the feedback button if we want, or feature flag it as well (yes I know).

GMM is quite keen on pushing this out as it's the last milestone of the 5 minute project (even though its technically done).

madsnedergaard: Cady fixed it, go merge :)

PR #8288, opened 2025-07-17
feat(capacity_parsers):Change to openelectricity 

annaib00:
## Issue
Changed data source and null updates

## Description
The source has changed from opennem to openelectricity, this change is now implemented here. The file name has also been changed and the dependencies in the dependent AU yaml files. 

Then there has been an issue of updating the capacities with null values, this is fixed by implementing a dropna statement. I would like a second opinion on if this is sufficient. 

### Preview
image

### Double check


electricitymap/contrib/capacity_parsers/OPENELECTRICITY.py
```diff
@@ -97,6 +112,8 @@ def fetch_production_capacity_for_all_zones(
         capacity_df.groupby(["zone_key", "mode"])[["value"]].sum().reset_index()
     )
 
+    capacity_df = capacity_df.dropna(subset=["value"])
```
annaib00, line 112: I would like a second opinion on whether this is the good way to not get null capacity updates?

electricitymap/contrib/capacity_parsers/tests/test_OPENELECTRICITY.py
annaib00, line 1: I have also updated the capacity test to fit with the new file name

electricitymap/contrib/capacity_parsers/OPENELECTRICITY.py
VIKTORVAV99: Should be but maybe we can add a snapshot test with some deliberate null values in the mock data to ensure it works as expected?

annaib00: That is a good idea, however I am not entirely sure how to implement this?

VIKTORVAV99: I'd intercept the http call with a mock function and then just check what the parser returns.

Looks like this:
```
def test_fetch_consumption(adapter, session, snapshot):
    data = Path(base_path_to_mock, "DK-DK1_consumption.xml")
    adapter.register_uri(
        GET,
        ANY,
        content=data.read_bytes(),
    )

    assert snapshot == ENTSOE.fetch_consumption(ZoneKey("DK-DK1"), session)
```

VIKTORVAV99: Might have to use read_text() instead tho as this returns JSON

VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99 (approved):
LGTM! 🎉

VIKTORVAV99:
Let's hope it works now!

PR #8289, opened 2025-07-17
feat(openem): hackily add exchanges

seljaks:

> This is a hacky and temporary solution because the source is migrating to a new API. That API does not provide exchange or export/import data at the moment.

We parse data from openelectricity that powers their dashboard. The data includes total exports and imports for each of the connected zones in eastern Australia.

Most of the zones only have one exchange, so it's easy calculate the netflow to it's only neighbor. The only exchange that requires some math is AU-NSW->AU-VIC which can be calculated as described below

```
    Calculate AU-NSW->AU-VIC netflow from exports and imports of AU-NSW and AU-QLD.

    NSW has two exchanges one to QLD, one to VIC. QLD only has one exchange to NSW.
    We want to know
        AU-NSW->AU-VIC = NSW_exports_to_VIC - NSW_imports_from_VIC

    NSW has two exchanges one to QLD, one to VIC:
        NSW_exports = NSW_exports_to_QLD + NSW_exports_to_VIC
        NSW_imports = NSW_imports_from_QLD + NSW_imports_from_VIC

    and because QLD only has one exchange to NSW:
        NSW_exports_to_QLD = QLD_imports_from_NSW = QLD_imports
        NSW_imports_from_QLD = QLD_exports_to_NSW = QLD_exports

    thus
        NSW_exports_to_VIC = NSW_exports - QLD_imports
        NSW_imports_from_VIC = NSW_imports - QLD_exports

    and
        AU-NSW->AU-VIC = NSW_exports - QLD_imports - NSW_imports + QLD_exports
        = NSW_exports - NSW_imports + QLD_exports - QLD_imports
```

Side note: also cleaned up some of the parser code because it made no sense


VIKTORVAV99 (commented):
Looks good but I'd really like to add a snapshot test at the very least to ensure this is working correctly and that we don't get regressions in the future.

I can assist if needed 🙂

seljaks:
tests added!

parsers/OPENNEM.py
```diff
@@ -341,31 +329,210 @@ def fetch_exchange(
...
-        {
-            "datetime": dt.to_pydatetime(),
-            "netFlow": value * direction,
-            "source": SOURCE,
-            "sortedZoneKeys": key,
-        }
-        for dt, value in series.iteritems()
+        Exchange(datetime=dt, netFlow=netflow, zoneKey=exchange_key, source=SOURCE)
+        for dt, netflow in datetimes_and_netflows
     ]
```
VIKTORVAV99: Don't this need to use the ExchangeList model if we are to use exchanges internally as the Exchange class itself don't serialize to JSON automatically?

VIKTORVAV99: I'd be happy to make the changes if you wish?

PR #8293, opened 2025-07-21
Cleanup: Dead code removal

VIKTORVAV99:
## Description

Found some code that is not used at all and since we are moving all our validation to the quality_validator backend it seems prudent to just remove it.


annaib00 (approved):
LTGM:)

PR #8297, opened 2025-07-22
[Parsers]: Clean up parsers

VIKTORVAV99:
## Issue
Fixes: GMM-934
Fixes: GMM-935
Fixes: GMM-936
Fixes: GMM-938

## Description

Cleans up a few parsers so they don't report 0 by default if the data is missing so we can run expected modes rules on them.

### Double check


electricitymap/contrib/lib/models/events.py
```diff
@@ -710,7 +710,7 @@ def to_dict(self) -> dict[str, Any]:
             "sourceType": self.sourceType,
             "correctedModes": []
             if self.production is None
-            else list(map(str, self.production._corrected_negative_values)),
+            else sorted(map(str, self.production._corrected_negative_values)),
```
VIKTORVAV99, line 713: This made the snapshot output fail randomly as it was non deterministic since it uses a map internally which don't keep track of the insertion order.

parsers/HN.py
```diff
@@ -36,110 +42,203 @@
...
+def get_datetime(date: str | None, hour: int) -> datetime | None:
+    """
+    Returns a datetime object with the given date and hour, or None if date is invalid
+    """
+    if date is None or date == "" or date.strip() == "":
+        return None
+    try:
+        return datetime.strptime(date, "%m/%d/%Y").replace(
+            tzinfo=ZoneInfo("America/Tegucigalpa")
+        ) + timedelta(hours=hour)
```
VIKTORVAV99, line 239: Big changes in this parser but this is the most important one. Previously there was a bug where it added one more hour than it should have shifting all data one hour + making one hour each day estimated. This can be observed on the app.

annaib00 (approved):
LGTM!

PR #8299, opened 2025-07-24
Add Turks and Caicos to config and map

### LINKED ISSUE #8136: Add a zone for Turks and Caicos
There is no zone for the Caribbean state of Turks and Caicos. Data about energy production might take some work to put together to create estimates, but I think there should be a zone for Turks and Caicos as other islands in the region seem to be represented.

miqs1992:
## Issue

Closes: electricitymaps/electricitymaps-contrib#8136

## Description

Based on the request in the issue, I added

* TC.yaml to config/zones
  * Basic geography data
  * Capacity based on Ember API (only data up to 2023)
  * Mixin based on US Dep of Energy doc (I wanted to show some amount of solar energy which is not visible in Ember data)
  * Connection to Ember parser
* Added shape to world.geojson and zone names to en and pl locales
* Added mapping to [EMBER.py](<http://EMBER.py>)

### Preview

Screenshot 2025-07-24 at 10 27 30

### Double check


config/zones/TC.yaml
```diff
@@ -0,0 +1,45 @@
...
+  oil:
+    - datetime: '2017-01-01'
+      source: Ember, Yearly electricity data
+      value: 80.0
+    - datetime: '2018-01-01'
+      source: Ember, Yearly electricity data
+      value: 90.0
+    - datetime: '2023-01-01'
+      source: Ember, Yearly electricity data
+      value: 90.0
```
VIKTORVAV99, line 16: Since this value is the same as the previous one it can be removed.

VIKTORVAV99 (commented):
Thanks for the PR!

We are making some changes into what information is available in the zones config so I'll hold off on merging this for a little while longer but everything looks good.

I have one small suggestion around the capacity though.

VIKTORVAV99:
@miqs1992 could you also add the translations to `web/public/locales/en.json`, this is a legacy file still used by the API.

PaulRoms:

PR #8305, opened 2025-07-28
[Parser]: Clean up RU parser

VIKTORVAV99:
## Issue
Fixes GMM-937

## Description

Cleans up the RU parser and simplify the logic a lot, also adds new values for wind in `RU-1`
Also adds a snapshot test suite to ensure no unwanted changes occurred between the old and the new parser.

### Double check


parsers/RU.py
```diff
@@ -50,196 +66,94 @@
...
 
 def fetch_production(
-    zone_key: str = "RU",
+    zone_key: ZoneKey,
     session: Session | None = None,
     target_datetime: datetime | None = None,
     logger: Logger = getLogger(__name__),
 ) -> list[dict] | dict:
-    """Requests the last known production mix (in MW) of a given country."""
-    if zone_key == "RU":
```
VIKTORVAV99, line 66: There is no point in this as we will use the subzone data and aggregate it in the backend instead.

parsers/RU.py
```diff
@@ -50,196 +66,94 @@
...
         "RU-2": 2,
     }
-    if zone_key not in zone_key_price_zone_mapper:
-        raise NotImplementedError("This parser is not able to parse given zone")
-
-    if target_datetime:
-        target_datetime_tz = target_datetime.astimezone(tz=TIMEZONE)
-    else:
-        target_datetime_tz = datetime.now(TIMEZONE)
-    # Query at t gives production from t to t+1
```
BastienBigue, line 121: No need to keep this logic ?

BastienBigue (approved):
LGTM

parsers/RU.py
VIKTORVAV99: Honestly I couldn't figure out why it was in there in the first place. When we query data for 12:00 we expect data for 12:00-13:00 not 11:00-12:00.

I think this is just some legacy code that should never have been there and since the parser have not been updated in 6 years it's just been left there.

But I can add it back if you think otherwise.

BastienBigue: yeah, I think it make sense to remove it then. I wasn't sure of it by looking at the explanation only.

parsers/RU.py
```diff
@@ -50,196 +66,94 @@
...
-                )
-
-        # Date
-        hour = f"{int(datapoint['INTERVAL']):02d}"
-        row["datetime"] = datetime.strptime(f"{date} {hour}", "%Y.%m.%d %H").replace(
-            tzinfo=TIMEZONE
-        )
-        last_dt = datetime.now(TIMEZONE) - timedelta(hours=1)
-
-        # Drop datapoints in the future
```
BastienBigue, line 163: cf https://github.com/electricitymaps/electricitymaps-contrib/pull/8311#discussion_r2244714876

PR #8306, opened 2025-07-28
feat(parsers):Clean up CA-AB parser

annaib00:
## Issue
For [GMM-931](https://linear.app/electricitymaps/issue/GMM-931/migrate-ca-abpy)

> [!IMPORTANT]
> Should be merged after the backend PR, which will be made soon :) 

### Double check

### DIFF
--- a/parsers/CA_AB.py
+++ b/parsers/CA_AB.py
@@ -24,17 +24,31 @@ from electricitymap.contrib.lib.models.event_lists import (
     PriceList,
     ProductionBreakdownList,
 )
-from electricitymap.contrib.lib.models.events import EventSourceType, ProductionMix
+from electricitymap.contrib.lib.models.events import (
+    EventSourceType,
+    ProductionMix,
+)
 from electricitymap.contrib.lib.types import ZoneKey
-
-# Local library imports
-from parsers.lib import validation
 
 DEFAULT_ZONE_KEY = ZoneKey("CA-AB")
 MINIMUM_PRODUCTION_THRESHOLD = 10  # MW
 TIMEZONE = ZoneInfo("Canada/Mountain")
 URL = urllib.parse.urlsplit("http://ets.aeso.ca/ets_web/ip/Market/Reports")
 URL_STRING = urllib.parse.urlunsplit(URL)
+SOURCE = URL.netloc
+
+PRODUCTION_MAPPING = {
+    "COGENERATION": "gas",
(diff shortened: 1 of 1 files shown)


parsers/CA_AB.py
```diff
@@ -115,48 +129,26 @@ def fetch_production(
...
-            floor=MINIMUM_PRODUCTION_THRESHOLD,
-            remove_negative=True,
-        )
-    ]
+
+    production_breakdowns = ProductionBreakdownList(logger)
+    mix = ProductionMix()
+    for key in generation:
+        if key in PRODUCTION_MAPPING:
+            mix.add_value(PRODUCTION_MAPPING[key], generation.get(key))
```
VIKTORVAV99: Maybe add some waning logs here if the key is not in the production mapping (or a key that should be skipped) so we know if they add any new keys in the future.

parsers/CA_AB.py
```diff
@@ -115,48 +129,26 @@ def fetch_production(
...
+
+    date = get_csd_report_timestamp(response.text)
+
+    production_breakdowns.append(
+        zoneKey=zone_key,
+        datetime=date,
+        production=mix,
+        source=SOURCE,
+    )
```
VIKTORVAV99: The previous version also parsed storage, we should do the same here.

parsers/CA_AB.py
VIKTORVAV99:
```suggestion
            mix.add_value(PRODUCTION_MAPPING[key], generation.get(key), correct_negative_with_zero=True)
```

parsers/CA_AB.py
```diff
@@ -115,48 +129,26 @@ def fetch_production(
...
-                "zoneKey": zone_key,
-            },
-            logger,
-            floor=MINIMUM_PRODUCTION_THRESHOLD,
-            remove_negative=True,
-        )
-    ]
+
+    production_breakdowns = ProductionBreakdownList(logger)
+    mix = ProductionMix()
```
VIKTORVAV99:
```suggestion
    production = ProductionMix()
    storage = StorageMix()
```

VIKTORVAV99 (commented):
Looks good but I think we are missing storage?

Does the source still have this?

VIKTORVAV99:
I'll let you handle the merging once the other PR is ready and approved.

PR #8310, opened 2025-07-29
[Parser]: Clean up CA_QC

VIKTORVAV99:
## Issue
Fixes: GMM-932

## Description

Removes the 0 filling if data is missing and reworks the logic a bit to be simpler and more type safe.

> [!IMPORTANT]
> There is a matching backend PR that enables expected modes validation for gas and the other modes so there will be no loss in data quality.

### Double check


BastienBigue (approved):
LGTM !

PR #8311, opened 2025-07-29
[Parser]: Migrate OPENNEM

VIKTORVAV99:
## Issue

Fixes: GMM-929

### Double check


parsers/OPENNEM.py
```diff
@@ -74,80 +78,48 @@
...
-) -> list[dict]:
-    def filter_solar_production(obj: dict) -> bool:
-        return bool(
-            "solar" in obj.get("production", {})
-            and obj["production"]["solar"] is not None
-        )
-
-    # The latest data points for AU-TAS are regularly containing only values for solar
-    # and no other production values. This is likely an invalid report, so we filter it out.
-    def only_solar_production(obj: dict) -> bool:
```
BastienBigue, line 128: As explained in slack, with the current validation set up, removing will lead to invalid but not yet validated data to be processed by the flowtracing pipeline. 
Can we live with it for some time and bet that we will revamp the validation pipeline or should we keep this validation during this time ?

parsers/OPENNEM.py
```diff
@@ -164,173 +136,184 @@ def generate_url(zone_key: str, target_datetime: datetime | None) -> str:
...
     region = ZONE_KEY_TO_REGION.get(zone_key)
-    url = generate_url(
-        zone_key=zone_key,
-        target_datetime=target_datetime,
-    )
-
-    response = (session or requests).get(url)
-    response.raise_for_status()
-    datasets = response.json()["data"]
+    if not region:
```
BastienBigue: Do we need this check again ? It is already in fetch_datasets

parsers/OPENNEM.py
```diff
@@ -164,173 +136,184 @@ def generate_url(zone_key: str, target_datetime: datetime | None) -> str:
...
+            continue
+        production_breakdown_list = ProductionBreakdownList(logger=logger)
+        history = dataset["history"]
+        start = datetime.fromisoformat(history["start"])
+        interval_min = int(history["interval"][:-1])  # remove 'm' at the end
+        delta = timedelta(minutes=interval_min)
+        data = history["data"]
+        for i, value in enumerate(data):
+            dt = start + i * delta
+            if dt > now:
```
BastienBigue, line 177: Do we want to keep this ? 
I saw in your RU PR that this logic was removed. 
I am not sure that it is a problem if it is removed, unless the source provides crap data when in the future 🤔

BastienBigue (approved):

parsers/OPENNEM.py
VIKTORVAV99: Here the data is all 0 in the future so don't make much sense to include it. In the RU parser it returns actual data.

Not sure what we prefer TBH, but to some degree it makes sense to remove future data but on the other hand it's probably more exact than our estimations.

parsers/OPENNEM.py
VIKTORVAV99: I would prefer to remove all parser side validation but on the other hand I don't think we should risk data quality to do so. With the race condition edge case we observed it might make sense to include it (but only for AU-TAS) for now and ensure it's removed when we revamp the pipeline.

What do you think?

parsers/OPENNEM.py
BastienBigue: Good for me !

parsers/OPENNEM.py
BastienBigue: Agree 👍

parsers/OPENNEM.py
```diff
@@ -164,173 +136,201 @@ def generate_url(zone_key: str, target_datetime: datetime | None) -> str:
...
+                    zoneKey = event.zoneKey
+                    corrected_breakdown.append(
+                        zoneKey=zoneKey,
+                        datetime=dt,
+                        production=production,
+                        storage=storage,
+                        source=source,
+                    )
+                    break
+        merged_production = corrected_breakdown
```
VIKTORVAV99, line 240: It's working now!

@BastienBigue could you double check and approve again if it still looks good?

BastienBigue: Lovely !

BastienBigue (approved):
LGTM

PR #8313, opened 2025-07-30
Update capacities for AU

VIKTORVAV99:
## Issue

Outlier detection is flagging values because the capacities are out of date.


PaulRoms (approved):
lgtm

VIKTORVAV99:
Crap, shouldn't have had auto merged this. Just noticed the solar capacities are super low for some reason.

I'll create a PR to revert just the solar and then a task to look into why that might have happened for the cooldown week as this is parsed data.

PR #8315, opened 2025-07-31
Removes errornous solar capacities for AU

VIKTORVAV99:
## Issue
For some reason the capacity parser for AU was pulling way lower solar values than expected, the other modes look fine.

So let's remove these for now and look into the parser issue separately. 

### Double check


annaib00 (approved):
lgtm:)

PR #8317, opened 2025-07-31
Set refetch frequency for RU production

VIKTORVAV99:
## Issue
We'd like to refetch all this data but producing one task per hour in unnecessary, setting the refetch frequency ensure we get one task per 24 hours.


BastienBigue (approved):
LGTM

PR #8335, opened 2025-08-01
Remove validation code only used by the test_parser utility

VIKTORVAV99:
## Description

This code was only used by the test_parser utility and it's functionality has since been moved to the individual parser classes and the backend quality validator so this code is no longer needed.

### Double check


BastienBigue (approved):
LGTM 🧹

Issue #8318, opened 2025-07-31
[Data Issue]: CY generation data available from ENTSO-E TP


luisagrether:
### When did this happen?

"2025-07-01"

### What zones are affected?

"CY"

### What is the problem?

"I'd just quickly like to tell you that hourly generation and load data for Cyprus are now available from the ENTSO-E Transparency Platform after I personally contacted them about it. That means, you could theoretically stop scraping the data from the Cyprian TSOs website and instead get it directly from ENTSO-E."

VIKTORVAV99:
Thanks a lot!

We'll double check the data to ensure it matches what we expect but after an initial glance it looks very promising!

PR #8336, opened 2025-08-01
[Parser]: Add ENTSO-E parsers to CY

VIKTORVAV99:
## Issue

Related to #8318

## Description

While we still need to double check and decide what to do with the production data let's add the other signals we didn't get before.

### Double check


BastienBigue (approved):
LGTM 🇨🇾

PR #8337, opened 2025-08-02
[Capacity]: Backfill US capacities

VIKTORVAV99:
### Double check

### DIFF
--- a/config/zones/US-CAL-LDWP.yaml
+++ b/config/zones/US-CAL-LDWP.yaml
@@ -11,7 +11,16 @@ capacity:
     - datetime: '2017-10-01'
       source: EIA.gov
       value: 2.0
-    - datetime: '2021-01-01'
+    - datetime: '2018-10-01'
+      source: EIA.gov
+      value: 22.0
+    - datetime: '2019-01-01'
+      source: EIA.gov
+      value: 2.0
+    - datetime: '2019-07-01'
+      source: EIA.gov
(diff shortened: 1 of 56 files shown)


config/zones/US-CAL-LDWP.yaml
```diff
@@ -11,7 +11,16 @@ capacity:
...
+    - datetime: '2018-10-01'
+      source: EIA.gov
+      value: 22.0
+    - datetime: '2019-01-01'
+      source: EIA.gov
+      value: 2.0
+    - datetime: '2019-07-01'
+      source: EIA.gov
+      value: 3.5
+    - datetime: '2020-01-01'
```
BastienBigue, line 23: This seems odd 🤔

VIKTORVAV99: Yeah a bit but it's the data EIA is returning. Could be a test battery that was installed and then removed or something?

BastienBigue (approved):
I checked a few of the zones. Some seems fine, some seem odd. If we end up flagging a lot of US data because of capacity, we should remember this PR 😅

config/zones/US-CAL-LDWP.yaml
BastienBigue: Yeah maybe 🤷 
I guess we should trust EIA and hope the generation data is in sync with the capacities

PR #8338, opened 2025-08-04
feat(parsers):Add updated solar capacity GT

annaib00:
## Issue
False capacity outliers GT

## Description
Add updated capacity to Guatemala due to solar panel increase in beginning of 2025 (https://www.bnamericas.com/en/news/guatemala-solar-capacity-up-60-with-new-plant). 

We have a capacity parser (ember) for this zone but ember does not have this increase in their data yet. Not sure if we want to mix this up or just stick with the ember capacities even if they are outdated? 

### Preview
Screenshot 2025-08-04 at 10 09 26

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8342, opened 2025-08-05
Add BE,CH,DE / FR capacity

t-couery:
## Issue

The interconnection capacity between Switzerland, Belgium and Germany is missing with France.

## Description

This information comes from the CRE document. The capacity for Germany and Belgium are combined, but it is indicated that they are equal.
The document is not recent: 2023, but it does not appear that there have been any new interconnections until 2025, and the numbers are consistent with what has been observed on Electricity Map

Source CRE / RTE ( French electricity transport network) 2023
https://www.cre.fr/fileadmin/Documents/Rapports_et_etudes/2024/Rapport_Interconnexions_CRE.pdf
image


config/exchanges/CH_FR.yaml
```diff
@@ -1,3 +1,6 @@
+capacity:
+  - -2700
+  - 2700
```
VIKTORVAV99:
```suggestion
capacity:
  - -2700
  - 1100
```
Should it not be the above? From the image you shared it looks like they have 1.1 GW import capacity and 2.7 export capacity?

(reversed here since we do CH to FR and not FR to CH)

PaulRoms:
Interconnection capacity are a bit tricky: It is possible that FR Electricity transites via CH to go to DE and therefore I'm not sure we want to have AC interconnectors capacity in our map.
We are in the process of removing it

PR #8349, opened 2025-08-12
Paulromieu/cleaning parser data no used

PaulRoms:
## Issue
Some zones are using GPZD but we are still parsing production data for it. A solution can be to comment the line that for the parser (to know that it exists) while not fetching the data if not used. 

- TH: All mapped to 'unknown' because there is no available breakdown.
- BB: no data since 2025-01-24 03:00:00+00 
- CA-NS: No data since 2024-01-23 19:00:00+00 
- CA-PE: No data since 2024-07-19 13:30:00+00 
- IL: No data since: 2023-10-06 
- PR: No data since 2022-03-24


config/zones/TH.yaml
```diff
@@ -324,7 +324,7 @@ isRenewable:
 parsers:
   consumption: TH.fetch_consumption
   price: TH.fetch_price
-  production: TH.fetch_production
+  # production: TH.fetch_production
```
VIKTORVAV99: Can we get historical data for this one later on if we wanted to?

Otherwise I think we should keep collecting the data or change the parser to be a totalProduction parser and not production per mode all mapped to unknown.

That way we can use the data in future versions of GPZD models that can take totalProduction into account.

PaulRoms: Modified it to keep production: TH.fetch_production

VIKTORVAV99:
BTW for TH we should probably do this: GMM-982

VIKTORVAV99 (approved):
We can also archive the parsers if we want but let's just do this first.

PR #8352, opened 2025-08-13
fix(parsers):Change TR parser to replace with 0 for negative production

### LINKED ISSUE #7807: [Data Issue]: no vs 0 for solar in TR
### When did this happen?

"2025-01-27 23:00" and other hours in 2025.

### What zones are affected?

TR

### What is the problem?

It seems like we use no instead of 0 when there is no solar. As of right now we are removing the entire check for solar as an expected mode so if we are are actually missing solar we won't flag that for estimations and won't trigger any alerts. See this [PR](https://github.com/electricitymaps/electricitymaps/pull/6642) for reference.

annaib00:
## Issue

Replace negative solar production with 0 instead of None.

## Description
Closes #7807 

When there is some negative production (some self consumption perhaps) as for ex solar, instead of replacing with None we replace with 0. This allow us to add solar for TR to expected modes. 

**A PR for adding solar to expected modes for TR will also be made.** 

### Double check

### DIFF
--- a/parsers/TR.py
+++ b/parsers/TR.py
@@ -144,7 +144,11 @@ def fetch_production(
         mix = ProductionMix()
         for key, value in item.items():
             if key in INVERT_PRODUCTION_MAPPPING:
-                mix.add_value(INVERT_PRODUCTION_MAPPPING[key], value)
+                mix.add_value(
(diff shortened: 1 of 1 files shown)


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8353, opened 2025-08-14
use ENSTO-e Parser

PaulRoms:
## Issue
Our current CY parser is not working properly.
At the same time, the ENSTO-e parser for CY seems to work well. 

## Analysis: 
From the flowtracing pipeline, it seems that the CY parser is okay until 2025-March, after it is no longer correct.
image

From ENSTO-e, the data seems complete from 1st of March:
[Cyprus analysis ENSTO-e](https://docs.google.com/spreadsheets/d/19PQcZ75TL_0KsyDI8r3l03DwSaUNyD3zXkbKcT6ovDU/edit?gid=596322936#gid=596322936)
It is more or less in line with [EMBER monthly:](https://ember-energy.org/data/electricity-data-explorer/?entity=Cyprus) . 

## Actions:
- In lib -> Change the estimation time box
- Refetch Cyprus from 2025-03-01
- Re-rerun flowtracing.

### DIFF
--- a/config/zones/CY.yaml
+++ b/config/zones/CY.yaml
@@ -357,7 +357,8 @@ parsers:
   consumption: ENTSOE.fetch_consumption
   consumptionForecast: ENTSOE.fetch_consumption_forecast
   generationForecast: ENTSOE.fetch_generation_forecast
-  production: CY.fetch_production
+  # production: CY.fetch_production # Valid until end of February 2025
+  production: ENTSOE.fetch_production # Valid from 1st of March 2025
   productionCapacity: EMBER.fetch_production_capacity
   productionPerModeForecast: ENTSOE.fetch_wind_solar_forecasts
 region: Asia


VIKTORVAV99 (commented):
Are these fetching data at the same resolution?
Otherwise we need to run a migration script to remove data from the old source from the first time ENTSO-E has data. As there is no way to guarantee it won't fetch data earlier than March 1.

Oh and why do we need to change the estimation timebox? Both should be using Forecast Hierarchy

PaulRoms:
> Are these fetching data at the same resolution? Otherwise we need to run a migration script to remove data from the old source from the first time ENTSO-E has data. As there is no way to guarantee it won't fetch data earlier than March 1.
> 
> Oh and why do we need to change the estimation timebox? Both should be using TSA?

In current parser (CY): Data is every 15min
In ENTSOE-E: Data is every hour.
-> I propose to create a migration that will delete all parser data for CY from 2025-03-01 until now. 

"As there is no way to guarantee it won't fetch data earlier than March 1" -> This might be tackled later with a similar approach as the estimation time box.

For the sake of simplicity, for the estimation model, I would use TSA instead of Forecast Hierarchy for CY for the ENTSOE-E parser

VIKTORVAV99:
Why would you use TSA over Forecast Hierarchy? It's only pulling the solar and wind values from ENTSO-E and using those instead of TSA based solar and wind interpolation. If they are not available it falls back to using TSA.

> "As there is no way to guarantee it won't fetch data earlier than March 1" -> This might be tackled later with a similar approach as the estimation time box.

But then we need to have that in place before we merge this as we can't guarantee someone don't trigger a refetch of the data in the meantime or one of our refetch jobs does it?

PaulRoms:
I'm keen to keep Forecast Hierarchy! 

In theory yes. But I prefer to have CY working and take care of the parser time box at another time. I'm fearing the PR might never get merge if we want.

VIKTORVAV99 (approved):
LGTM! 🎉

PaulRoms:
> I'm keen to keep Forecast Hierarchy!
> 
> In theory yes. But I prefer to have CY working and take care of the parser time box at another time. I'm fearing the PR might never get merge if we want.

After second thought, let's wait for the Parser time box to happen and keep the PR open.

PR #8354, opened 2025-08-14
Paulromieu/adding battery to mongolia

PaulRoms:
## Issue
Adding Battery to Mongolian parser. 
website: https://ndc.energy.mn/

Currently: it is the afternoon in Mongolia, so I assume it is charging. 
image

So the value battery_storage is 12.7MW

### DIFF
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -12,7 +12,7 @@ from electricitymap.contrib.lib.models.event_lists import (
     ProductionBreakdownList,
     TotalConsumptionList,
 )
-from electricitymap.contrib.lib.models.events import ProductionMix
+from electricitymap.contrib.lib.models.events import ProductionMix, StorageMix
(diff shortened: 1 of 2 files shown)


parsers/MN.py
```diff
@@ -107,13 +108,16 @@ def fetch_production(
     prod_mix.add_value("coal", data["coal"])
     prod_mix.add_value("solar", data["solar"])
     prod_mix.add_value("wind", data["wind"])
+    storage_mix = StorageMix()
+    storage_mix.add_value("battery", data["battery_storage"])
```
VIKTORVAV99, line 112: I think we need to invert the value here as we treat negative as discharge and positive as charging.

VIKTORVAV99 (commented):
Looks like this branch contains the CY changes as well. We should separate the them.

PR #8355, opened 2025-08-14
Songino -> Battery storage

PaulRoms:
## Issue
Adding Battery to Mongolian parser. 
website: https://ndc.energy.mn/

Currently: it is the afternoon in Mongolia, so I assume it is charging. 
image

So the value battery_storage is 12.7MW

### DIFF
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -47,7 +47,7 @@ JSON_API_MAPPING = {
     "dts": "coal",
     "sumnar": "solar",
     "sumsalhit": "wind",
-    "Songino": "battery_storage"
+    "Songino": "battery_storage",
 }
 
 
--- a/parsers/MN.py
+++ b/parsers/MN.py
@@ -12,7 +12,7 @@ from electricitymap.contrib.lib.models.event_lists import (
     ProductionBreakdownList,
     TotalConsumptionList,
 )
-from electricitymap.contrib.lib.models.events import ProductionMix
+from electricitymap.contrib.lib.models.events import ProductionMix, StorageMix
 from parsers.lib.exceptions import ParserException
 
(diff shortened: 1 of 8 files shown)


parsers/MN.py
```diff
@@ -107,13 +108,16 @@ def fetch_production(
     prod_mix.add_value("coal", data["coal"])
     prod_mix.add_value("solar", data["solar"])
     prod_mix.add_value("wind", data["wind"])
+    storage_mix = StorageMix()
+    storage_mix.add_value("battery", data["battery_storage"])
```
VIKTORVAV99: Mentioned this on the other PR as well but I think we need to invert the value here for it to be correct.

Negative is discharge and positive charge.

annaib00: But if it is charging now and the value is positive, and charging is positive, then we should not invert, right?

VIKTORVAV99: Well that's the question. Is it charging or discharging? Unless the dashboard has both positive and negative values I would assume they always show discharge.

VIKTORVAV99: Usually the charging pattern is something like charging at night, discharging in the morning, charge mid day, discharge afternoon to meet normal consumption patterns.

PaulRoms: Let's wait a couple of hours and see if the value turns negative

PaulRoms: Let's see later how it changes!

PaulRoms:
18:00 
image

PaulRoms:
image

parsers/MN.py
PaulRoms: I have inverted it, it makes more sense.

PaulRoms:

PR #8357, opened 2025-08-15
temp fix for AU-* data

VIKTORVAV99:
This pull request makes a small but important change to the `process_production_datasets` function in `parsers/OPENNEM.py`. The logic for handling datapoints that only contain solar production has been generalized: it now applies to all regions, not just the AU-TAS zone.

- Generalized the removal of datapoints that only have solar production from being specific to the `AU-TAS` zone to all regions in the `process_production_datasets` function.

### DIFF
--- a/parsers/OPENNEM.py
+++ b/parsers/OPENNEM.py
@@ -217,27 +217,27 @@ def process_production_datasets(
         logger=logger,
     )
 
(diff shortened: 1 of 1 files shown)


BastienBigue (approved):
LGTM

PR #8358, opened 2025-08-15
remove estimation_method from config.

PaulRoms:
## Issue
Remove estimation_method from config zones and web.


VIKTORVAV99 (approved):
LGTM! 🎉

web/src/features/panels/zone/util.ts
```diff
@@ -54,12 +54,6 @@ export const getZoneDataStatus = (
     return ZoneDataStatus.NO_INFORMATION;
   }
 
-  // If there are no production parsers or no defined estimation method in the config,
-  // we assume we do not have data for the zone
-  if (zoneConfig.parsers === false && zoneConfig.estimation_method === undefined) {
-    return ZoneDataStatus.NO_INFORMATION;
-  }
-
```
BastienBigue, line 62: the estimation_method here is taken from a zones.json file that contains info for the app frontend. I suppose the zones.json is built from the zones config. 
I think that removing this logic will make zones such as Andorra appear as Ukraine :
I suppose this is NO_REAL_TIME_DATA : 
Screenshot 2025-08-15 at 13 53 22
while this is how it appears now (I suppose It is NO_INFORMATION): 
Screenshot 2025-08-15 at 13 54 07

If we are ok with this, then let's move on, otherwise, can we find an alternative ? Like using the zone_config table ? But could be out of scope...

VIKTORVAV99: I think that is fine.
Otherwise we would have to modify the app backend as well and I don't think PLG likes us making too many changes while they are trying to migrate the app.

madsnedergaard: I'll handle it on the new app :)

madsnedergaard:
Hey @PaulRoms, this wasn't merged - we have handled it in the app, so I'll merge this now :)

PR #8362, opened 2025-08-20
feat(entsoe-outages): add outage events + code to get xml tree

unitrium:
## Issue

Add OutageEvents and code to get xml from entso-e

### DIFF
--- a/parsers/ENTSOE.py
+++ b/parsers/ENTSOE.py
@@ -1306,6 +1306,4 @@ def _query_entsoe_zip_endpoint(
 
 
 if __name__ == "__main__":
-    _query_zip_endpoint(
-        ZoneKey("FR"), session=Session(), target_datetime=datetime.now(timezone.utc)
-    )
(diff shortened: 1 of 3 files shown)


parsers/ENTSOE.py
```diff
@@ -1258,5 +1261,51 @@ def fetch_wind_solar_forecasts(
...
+            ]
+            for xml_file in xml_files:
+                tree = ET.parse(f"{tmpdir}/extracted/{xml_file}")
+                trees.append(tree)
+    return trees
+
+
 if __name__ == "__main__":
-    fetch_price(ZoneKey("FR"))
+    _query_zip_endpoint(
```
florianscheidl: Shall we remove this again?

florianscheidl: Also, looks like typo (_query_entsoe_zip_endpoint)?

unitrium: Yes let's remove

PR #8364, opened 2025-08-20
feat(parsers): parse entsoe generation outage

florianscheidl:
## Description

We add a parser for ENTSOE generation unit outage files provided as XML.
An outage event is stored for each hour betwen the beginning and the end of the outage.

We add a fixture and a snapshot test.

### Preview

One event the following form (here the one stored as snapshot):
{
    'capacity_reduction': 1030.0,
    'datetime': datetime.datetime(2025, 8, 31, 22, 0, tzinfo=datetime.timezone.utc),
    'fuel_type': 'nuclear',
    'generator_id': '22WTIHANG000242R',
    'outage_type': <OutageType.PLANNED: 'planned'>,
    'reason': 'Shutdown date as provided by the law',
    'source': 'entsoe.eu',
    'zoneKey': 'BE',
  }

### Double check


electricitymap/contrib/lib/models/events.py
```diff
@@ -843,8 +843,16 @@ class OutageType(Enum):
     PLANNED = "planned"
     UNPLANNED = "unplanned"
 
+    @staticmethod
+    def mapping_code_to_type(code: str) -> "OutageType":
+        return {
+            "A53": OutageType.PLANNED,
```
unitrium, line 849: This is only for ENTSO-E, since this part is supposed to be more generic I would recommend to keep this mapping inside the ENTSO-E parser.

electricitymap/contrib/lib/models/events.py
```diff
@@ -843,8 +843,16 @@ class OutageType(Enum):
...
+            "A54": OutageType.UNPLANNED,
+        }[code]
+
 
 class Outage(Event):
+    sourceType: EventSourceType = EventSourceType.forecasted
```
unitrium, line 855: Actually now I remembered this field is not super clear so I think this would rather be measured especially if it's for a current event.
```suggestion
    sourceType: EventSourceType = EventSourceType.measured
```

parsers/ENTSOE.py
```diff
@@ -817,6 +824,66 @@ def parse_prices(
     return prices
 
 
+def parse_outages(
+    xml_text: str,
```
unitrium, line 828: Could this be rather an xml tree object ? Because the return of the other function is a list of xml trees ? Otherwise we need to change the other one.

unitrium (approved):
Looks good a few comments :)

PR #8365, opened 2025-08-21
feat(parsers): query, parse and fetch outages

florianscheidl:
## Description
Implements query, parse and fetch functions for outages.
Includes snapshot tests and fixtures to test parsing and fetching end-to-end.

Outages are parsed per datetime and asset. We track zoneKey, fuel_type, outage_type (planned vs unplanned) and capacity_reduction.

For downstream application, we aggregate across assets, giving one event per datetime, zoneKey and fuel_type. 
This is what we want to track in the database.

### Preview

The events have this form after aggregation:
Screenshot 2025-08-21 at 09 52 08

### Next steps

Have the data parsed into a database.

### Double check


electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -420,7 +420,8 @@ def append(
         source: str,
         capacity_reduction: float,
         fuel_type: str,
-        outage_type: OutageType,
+        outage_type: OutageType | None = None,
+        generator_id: str | None = None,
```
VIKTORVAV99, line 424: Do we need the generator ID?

Is this something that would work on other regions/sources or is it ENTSOE specific?

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -431,11 +432,38 @@ def append(
...
             fuel_type,
             outage_type,
+            generator_id,
             reason,
         )
         if event:
             self.events.append(event)
 
+    @staticmethod
+    def aggregate_across_generation_units(
```
VIKTORVAV99: aggregate_per_mode?

Right now it's not clear what exactly it's aggregating across the units.

electricitymap/contrib/lib/models/events.py
```diff
@@ -843,11 +843,19 @@ class OutageType(Enum):
...
+    @staticmethod
+    def mapping_code_to_type(code: str) -> "OutageType":
+        return {
+            "A53": OutageType.PLANNED,
+            "A54": OutageType.UNPLANNED,
+        }[code]
+
 
 class Outage(Event):
+    sourceType: EventSourceType = EventSourceType.forecasted
```
florianscheidl, line 848: Keeping this as forecasted because we get a validation error when we use measured (because we have future values).
I would argue that future outages are kind of forecasted. We could also remove the validation or introduce a new type.

electricitymap/contrib/lib/models/event_lists.py
```diff
@@ -431,11 +432,38 @@ def append(
...
+        )
+        aggregated_outages = OutageList(logger)
+        for _, row in aggregated_outages_df.iterrows():
+            aggregated_outages.append(
+                zoneKey=row["zoneKey"],
+                datetime=row["datetime"],
+                source=row["source"],
+                capacity_reduction=row["capacity_reduction"],
+                fuel_type=row["fuel_type"],
+            )
```
VIKTORVAV99: We have in the past tried to avoid converting to dataframes in the parser models as usually the performance overhead and maintenance cost is higher than the benefit. (Especially we try to update dependencies).

Is there a way to do this without pandas in a easy way?

electricitymap/contrib/lib/models/events.py
```diff
@@ -843,11 +843,19 @@ class OutageType(Enum):
     PLANNED = "planned"
     UNPLANNED = "unplanned"
 
+    @staticmethod
+    def mapping_code_to_type(code: str) -> "OutageType":
+        return {
+            "A53": OutageType.PLANNED,
+            "A54": OutageType.UNPLANNED,
+        }[code]
```
VIKTORVAV99: This should live in the parser as it's specific to ENTSOE

parsers/ENTSOE.py
```diff
@@ -56,6 +59,7 @@
 DEFAULT_LOOKBACK_HOURS_REALTIME = 72
 DEFAULT_TARGET_HOURS_REALTIME = (-DEFAULT_LOOKBACK_HOURS_REALTIME, 0)
 DEFAULT_TARGET_HOURS_FORECAST = (-24, 48)
+DEFAULT_HOURS_OUTAGES = (-120, 120)
```
florianscheidl: I chose this window somewhat randomly. Thought we should cover at least beyond our current forecast horizon for the "lookforward".

electricitymap/contrib/lib/models/events.py
```diff
@@ -843,11 +843,19 @@ class OutageType(Enum):
...
+
 
 class Outage(Event):
+    sourceType: EventSourceType = EventSourceType.forecasted
     capacity_reduction: float
     fuel_type: str
-    outage_type: OutageType
+    outage_type: str | None = None
```
VIKTORVAV99: Should this not use the outage type?

parsers/ENTSOE.py
```diff
@@ -1261,32 +1350,68 @@ def fetch_wind_solar_forecasts(
...
     logger: Logger = getLogger(__name__),
 ) -> list[ET.ElementTree]:
     """Some endpoints are returning zip objects in which the xml files are stored."""
     # TODO: Currently this cannot use the proxy because it returns an attachment.
+
     URL = "https://web-api.tp.entsoe.eu/api"
     if target_datetime is None:
         target_datetime = datetime.now(timezone.utc)
 
-    if not isinstance(target_datetime, datetime):
```
florianscheidl: Removed because should always be datetime (either by input or if None we set to datetime).

parsers/ENTSOE.py
VIKTORVAV99: That's a shitload of datapoints to fetch. How often is this intended to run if we move it into production?

parsers/ENTSOE.py
```diff
@@ -1261,32 +1350,68 @@ def fetch_wind_solar_forecasts(
...
-        raise ParserException(
-            parser="ENTSOE.py",
-            message="target_datetime has to be a datetime in query_entsoe",
-        )
-    params = {}
-
-    params["periodStart"] = (target_datetime + timedelta(hours=0)).strftime(
-        "%Y%m%d%H00"  # YYYYMMDDHH00
+    params["periodStart"] = (target_datetime + timedelta(hours=span[0])).strftime(
```
florianscheidl, line 1533: Implementing the span as in other query functions.

parsers/ENTSOE.py
```diff
@@ -817,6 +836,76 @@ def parse_prices(
...
+                    zulu_to_utc(f"{end_time}")
+                ).replace(minute=0, second=0, microsecond=0) + timedelta(
+                    hours=1
+                )  # round to the next hour\
+
+                # HACK: creating one datetime per hour but should rather have one event per outage and handle this downstream.
+                for dt in [
+                    datetime_start,
+                    *list(
+                        pd.date_range(datetime_start_rounded, datetime_end, freq="H")
```
VIKTORVAV99: What's the benefit of this vs the existing datetime parsing in the ENTSOE parser?

This is also not guaranteed to be hourly and could be 15 minute intervals I think.

parsers/ENTSOE.py
```diff
@@ -1261,32 +1350,68 @@ def fetch_wind_solar_forecasts(
     return forcast_breakdown_list.to_list()
 
 
+@refetch_frequency(timedelta(hours=DEFAULT_LOOKBACK_HOURS_REALTIME))
```
VIKTORVAV99: If it's fetching 120 hours in the past it should not use the default one. It will create a lot of unnecessary requests if we refetch the data.

parsers/ENTSOE.py
```diff
@@ -1261,32 +1350,68 @@ def fetch_wind_solar_forecasts(
...
+                parser="ENTSOE.py",
+                message=f"No generation outages data found for {_zone_key}",
+                zone_key=zone_key,
+            )
+        # Aggregated data are regrouped unde the same zone key.
+
+        non_aggregated_data.append(parse_outages(raw_outage_data, zone_key, logger))
+
+    return OutageList.aggregate_across_generation_units(
+        [outage for outage_list in non_aggregated_data for outage in outage_list],
```
VIKTORVAV99: This seems overly complex for something that should just be passed as a plain list and handled in the aggregation?

parsers/test/__snapshots__/test_ENTSOE.ambr
```diff
@@ -5873,6 +5873,30 @@
...
 # ---
+# name: test_fetch_outages
+  list([
+    dict({
+      'capacity_reduction': 170.0,
+      'datetime': Timestamp('2024-04-17 22:00:00+0000', tz='UTC'),
+      'fuel_type': 'gas',
+      'generator_id': None,
+      'outage_type': None,
+      'reason': None,
```
VIKTORVAV99: Why are these not captured in the test as well?

electricitymap/contrib/lib/models/events.py
VIKTORVAV99: We should not remove the validation but we could introduce a new type.

We have a special type for price for example since it's day ahead but neither measured nor forecasted.

parsers/ENTSOE.py
VIKTORVAV99: Yes but the way we are calling the parsers we can't guarantee that.

electricitymap/contrib/lib/models/event_lists.py
unitrium: I think it might work in some other regions as well, we saw some regions in the US that would also report this.

electricitymap/contrib/lib/models/events.py
unitrium: Ah that's why you used predicted!

parsers/ENTSOE.py
unitrium: I think let's rather keep this window small because the API has a limitation of 200 documents per call. So let's only get an hour max 2 otherwise we will encounter errors.

electricitymap/contrib/lib/models/event_lists.py
unitrium: I'm fine with that for now as part of the hack project. But I think overall I would prefer if we could rather do the aggregation step later in the pipeline so the flowtracing pipeline would aggregate those across generation units. Especially because then after we can map it to our different resolutions.

florianscheidl:
Thanks for the detailed review @VIKTORVAV99! 
I'll address the most pressing ones - in the interest of hack we'd like to skip some of the details (we're keeping it in this extra branch for now) and fix these when we merge the changes to master.

parsers/ENTSOE.py
unitrium: I don't think you need to round the start, the time alignment in flowtracing could take an unrounded datetime as a start and then align it properly especially with our different resolutions :)

VIKTORVAV99: Yep it will sort the events, then window them, and then break them down to 1 minute intervals internally.

unitrium (approved):
Looks good to me, feel free to resolve the comments or not so we can go ahead with the hack. For now it's fine I think later we should rather yield hourly or sub hourly outage events and let the flowtracing/time alignment pipeline do the aggregation :)

electricitymap/contrib/lib/models/event_lists.py
florianscheidl: Let's keep them for now. For the ENTSOE parsers we would be aggregating across units, so it would not be used for now.

electricitymap/contrib/lib/models/event_lists.py
florianscheidl: Let's keep it for now then for simplicity.

florianscheidl: We are keeping modes, zones and outage types, and aggregating across the generation units. Do you think another naming would be more readible?

electricitymap/contrib/lib/models/event_lists.py
florianscheidl: I think that's too much of an overhead for the hack but I think we should keep it in mind at a later point.

electricitymap/contrib/lib/models/event_lists.py
VIKTORVAV99: But then this don't really fit the existing model as you are not aggregating all the values and deduplicating them. Like all the other aggregation functions in the models?<br><br>It means that the generation IDs will never be passed to the database right?

electricitymap/contrib/lib/models/events.py
florianscheidl: In the end I'd like it to be a string, so the member goes into the dataframe and then the database. In all steps before storing it as event, I am therefore thinking we treat it as OutageType and then in the event as str.
We can handle this differently if you have any ideas?

VIKTORVAV99: String enums!

electricitymap/contrib/lib/models/event_lists.py
florianscheidl: Yes, that was the plan for now to reduce the complexity.

parsers/test/__snapshots__/test_ENTSOE.ambr
florianscheidl: As of now we are aggregating across generators, so the data structure allows for them but we are not using the fields.

parsers/ENTSOE.py
florianscheidl: Still TBD. I think we'd want to report outages ASAP, so at least hourly? Reduced the window now.

parsers/test/__snapshots__/test_ENTSOE.ambr
VIKTORVAV99: But then we are going to fill the database with just None for these? That don't really make sense.<br><br>So either this should be removed from the model or aggregated as well.

VIKTORVAV99: Not saying this needs to be done for the hack but yeah this is a data modeling issue that needs to be fixed before we put it in production.

parsers/ENTSOE.py
florianscheidl: Changing to one hour.

parsers/ENTSOE.py
florianscheidl: Fair, let's keep it then.

electricitymap/contrib/lib/models/event_lists.py
florianscheidl: We've decided to avoid the aggregation here, doing it downstream.

florianscheidl: Change of plans! We're keeping per generation unit and not aggregating here.

parsers/ENTSOE.py
florianscheidl: The issue is how long the event would be considered valid. We could introduce an end-time and then figure out later how to reconcile that to combine it with production or forecasted data. The reason we choose one event per hour was to make it simpler and to consider an outage to be valid for one hour. That way we can follow the same schema as we do for production data.

Issue #8134, opened 2025-05-12
[Data Issue]: Installed nuclear capacity in Ontario too high


mackrauss:
### When did this happen?

"2024-12-31 24:00"

### What zones are affected?

CA-ON

### What is the problem?

At the end of 2024 the remaining two reactors in Pickering (A1 and A4) where taken off-line and are being decommissioned, thus removing about 1000 MW of capacity.

Your map lists Ontario's nuclear capacity at 13.1 GW but IESO shows now 12,184 MW (12.2 GW) of installed nuclear capacity, reflecting the removal of Pickering A. See https://www.ieso.ca/Power-Data/Supply-Overview/Transmission-Connected-Generation

julian-st:
as of today it still shows 13.1 GW with ieso.ca as source

julian-st:
i got the CA_ON.py working locally but when I edit the date to 2025 in the main I get the following 
`python electricitymap/contrib/capacity_parsers/CA_ON.py 
https://www.ieso.ca/-/media/Files/IESO/Document-Library/planning-forecasts/reliability-outlook/ReliabilityOutlookTables_2025Mar.ashx
Traceback (most recent call last):
  File "/media/style/1d77bc8f-01cc-4482-8062-215ca9c3404c/project/electricitymaps-contrib/electricitymap/contrib/capacity_parsers/CA_ON.py", line 66, in <module>
    fetch_production_capacity("CA_ON", datetime(2025, 3, 1), Session())
  File "/media/style/1d77bc8f-01cc-4482-8062-215ca9c3404c/project/electricitymaps-contrib/electricitymap/contrib/capacity_parsers/CA_ON.py", line 50, in fetch_production_capacity
    df["mode"] = df["mode"].apply(lambda x: MODE_MAPPING[x])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/media/style/1d77bc8f-01cc-4482-8062-215ca9c3404c/project/electricitymaps-contrib/path/to/venv/lib/python3.12/site-packages/pandas/core/series.py", line 4935, in apply
    ).apply()
      ^^^^^^^
  File "/media/style/1d77bc8f-01cc-4482-8062-215ca9c3404c/project/electricitymaps-contrib/path/to/venv/lib/python3.12/site-packages/pandas/core/apply.py", line 1422, in apply
    return self.apply_standard()
[...]

PR #8368, opened 2025-08-24
Updating of Ontario data to new mapping for 2025 data

julian-st:
## Issue
see #8134 

## Description
this updates the Ontario data to 2025 for nuclear energy capacity

## Double check
[x ] I have run pnpx prettier@2 --write . and poetry run format in the top level directory to format my changes.


julian-st:
thanks. Yes this should fix the capacity issues

Issue #8369, opened 2025-08-24
[Data Issue]: Installed nuclear capacity in Ontario too high


julian-st:
### When did this happen?
"2024-12-31 24:00"

see #8134 
unfortunately my fix did not help

Further information: The files from the ieso.ca site are not available by month but by the month of the last full quarter. So I try to do a fix to get the right files from the ieso.ca site

### What zones are affected?

CA-ON

### What is the problem?

At the end of 2024 the remaining two reactors in Pickering (A1 and A4) where taken off-line and are being decommissioned, thus removing about 1000 MW of capacity.

Your map lists Ontario's nuclear capacity at 13.1 GW but IESO shows now 12,184 MW (12.2 GW) of installed nuclear capacity, reflecting the removal of Pickering A. See https://www.ieso.ca/Power-Data/Supply-Overview/Transmission-Connected-Generation

VIKTORVAV99:
> unfortunately my fix did not help

What do you mean?
It has not been deployed to our production environment yet so there is no changes expected until tomorrow at the earliest.

You can see the changes that have been deployed under this page: https://github.com/electricitymaps/electricitymaps-contrib/releases

julian-st:
ok thanks, then lets wait until tomorrow. I am hopeful that the changes I made to the [config/zones/CA-ON.yaml](https://github.com/electricitymaps/electricitymaps-contrib/pull/8368/files#diff-2d8295a51a9e8698bcd6381a0c42973eda0fce0c81901bbf555c693476bd7481) file are going to resolve this

VIKTORVAV99:
> ok thanks, then lets wait until tomorrow. I am hopeful that the changes I made to the [config/zones/CA-ON.yaml](https://github.com/electricitymaps/electricitymaps-contrib/pull/8368/files#diff-2d8295a51a9e8698bcd6381a0c42973eda0fce0c81901bbf555c693476bd7481) file are going to resolve this

The parsers are currently only used to change the information in the config files. So unless the values are updated there the code changes won't really make a difference to the current changes.

If you want to rerun it you should use: `poetry run capacity_update --zone="CA-ON" --target_datetime="<what date you want to update the capacity for>"`

That will create a updated capacity configuration. Note that if the value is the same as the previous entry it will be skipped.

julian-st:
The new version of the site now shows the updated values for the ontario capacity :)

PR #8370, opened 2025-08-24
Get the outlook file with the proper month

julian-st:
## Issue

see #8369 
The Canada Ontario data is outdated

## Description

The files from the ieso.ca site are not available by month but by the month of the last full quarter. So I do a fix to get the right files from the ieso.ca site

### Double check

- [ x] I have run `pnpx prettier@2 --write .` and `poetry run format` in the top level directory to format my changes.


annaib00 (commented):
LGTM! Thanks for adding this update! 

My only though is perhaps the associated datetime for the capacity in 'fetch_production_capacity' should then be the last quarter end date and not target_datetime, so we don't end up with unnecessary duplicated data within the same quarter?

annaib00 (approved):
LGTM - Thanks for the changes!

PR #8373, opened 2025-08-28
fix(TR): fetch the last 2 days of data to avoid missing the last hours of the day

BastienBigue:
## Issue
We miss the last hours of the day every day because we fetch the current day only. 

## Description
Fetch the last 2 days instead. 

### Double check


VIKTORVAV99 (approved):
LGTM! 🎉

PR #8387, opened 2025-09-23
Move static files

ronan-mch:
Static files need to live in the `electricitymap/contrib` folder to be included in the installed Python package. Updating the `includes` directive in `pyproject.toml` adds them to the source distribution, but not the installed package, thus the library breaks on import.

Unfortunately, I also needed to update the Javascript import statement. I don't think this is super clean so I'm open to alternate suggestions. 

🙏🏻 🙏🏻 🙏🏻 🙏🏻 🙏🏻


VIKTORVAV99:
I'm pretty sure you also need to update several scripts in both python and maybe JS but more importantly this is going to be a breaking change when merged to the backend. So you might have to update a lot of projects if we merge this.

ronan-mch:
> I'm pretty sure you also need to update several scripts in both python and maybe JS but more importantly this is going to be a breaking change when merged to the backend. So you might have to update a lot of projects if we merge this.

That doesn't sound like fun 😬 

Can you suggest an alternative?

VIKTORVAV99:
In the past we have used symlinks to get around this. That way you can both move them and not move them. Only downside is they don't work on windows.

ronan-mch:
> In the past we have used symlinks to get around this. That way you can both move them and not move them. Only downside is they don't work on windows.

I'm not sure if we can use symlinks here. If they pointed from the package to outside the package, they would not resolve properly.

VIKTORVAV99:
> > In the past we have used symlinks to get around this. That way you can both move them and not move them. Only downside is they don't work on windows.
> 
> I'm not sure if we can use symlinks here. If they pointed from the package to outside the package, they would not resolve properly.

What if you do the other way around?

Move the files and then create a symlink in the "old" location so we don't need to update all the monorepo projects?

VIKTORVAV99:
Have you tried running this version of the submodule with the monorepo? Maybe it's not that many projects that need to be updated they can use the new package directly?

BastienBigue:
What about creating a new package ? 
That would contain only the static yaml files. 
And then the contrib package would depend on this new package. 
And the project or libs that import contrib only imports the latter ?

VIKTORVAV99:
> What about creating a new package ? That would contain only the static yaml files. And then the contrib package would depend on this new package. And the project or libs that import contrib only imports the latter ?

I like this idea, but it would have to be in the open source repo if we still want contributors to be able to use utilities like `test_parser.py`

BastienBigue:
Of course, I think both have to still be in the open source repo. 
I think the question is can we make contrib be 2 packages instead of only one ? I don't know all the capabilities of poetry to say if it is possible or not...

VIKTORVAV99:
I think UV can, and if we are making this a package we can use that as the package manager instead of poetry.

ronan-mch:
> Of course, I think both have to still be in the open source repo. I think the question is can we make contrib be 2 packages instead of only one ? I don't know all the capabilities of poetry to say if it is possible or not...

I think it should be possible given a different pyproject definition. I'll take a look at this approach, but in that case, we would need to figure out how to reference them within contrib. If we use a separate package, that package will need to be hosted somewhere publicly and then imported into the contrib Python package. 

I suspect that the current approach might be cleaner overall, so it would be good to clarify what things will break. My assumption was that most consumers of these files will import contrib and then use the constants to look up the files. If that's not the case, please let me know. In the meantime, I'll fix the failing tests.

VIKTORVAV99:
Yeah I think only feeder electricity reads the parsers and the rest are just either using constants or looking up config files directly.

VIKTORVAV99:
I suspect this will be the best long term option if we split it up: [https://docs.astral.sh/uv/concepts/projects/workspaces/#getting-started](https://docs.astral.sh/uv/concepts/projects/workspaces/#getting-started)

ronan-mch:
> I think UV can, and if we are making this a package we can use that as the package manager instead of poetry.

If we can avoid introducing a new package manager as part of this change, I'd greatly appreciate that. The scope has crept too far already! 😆

VIKTORVAV99:
Haha yeah, let's just try and get this packaged for now 🙂

ronan-mch:
Packaging separately worked ok. See #8388.

PR #8392, opened 2025-09-29
Move parsers into package

ronan-mch:
The parsers need to be in the electricitymap/contrib directory if they are going to be used by upstream projects. The current setup uses separate directory AND a symlink with imports using both paths. This breaks when using the parsers in a package, so we need to use a single approach.


VIKTORVAV99 (commented):
Current changes looks good but it seems we also need to update the Earthfile and some markdown links for the CI to pass.

VIKTORVAV99 (approved):
LGTM! 🎉

unitrium:
Were you able to integrate contrib into the monorepo since this was merged ? It seems like builds are broken because of `contrib/parsers/lib/digits_comma.traineddata` in the parser folder

VIKTORVAV99:
> Were you able to integrate contrib into the monorepo since this was merged ? It seems like builds are broken because of `contrib/parsers/lib/digits_comma.traineddata` in the parser folder

FYI I have no idea what that file is and as far as I know we don't use it for anything in GMM.

unitrium:
> > Were you able to integrate contrib into the monorepo since this was merged ? It seems like builds are broken because of `contrib/parsers/lib/digits_comma.traineddata` in the parser folder
> 
> FYI I have no idea what that file is and as far as I know we don't use it for anything in GMM.

As far as I can tell it's something for tesseract which we use in some parsers to read images. I don't know if it's still usefull maybe the parsers that used it are no longer in use.

PR #8393, opened 2025-09-29
feat(zone config): add currency information to the Zone model

unitrium:
## Issue

We need to know the currency of a zone.

## Description

This PR adds a currency field on the zone configuration making it easy for us to propagate this information through out the stack.


config/zones/BB.yaml
```diff
@@ -56,9 +56,9 @@ fallbackZoneMixes:
         wind: 0.0
 parsers:
   consumption: BB.fetch_consumption
-  # production: BB.fetch_production
```
unitrium, line 59: That was probably removed during formating, I will revert it.

config/zones/AX.yaml
```diff
@@ -311,3 +311,4 @@ sources:
 timezone: Europe/Mariehamn
 zone_name: Åland Islands
 country_name: Åland Islands
+currency: EUR
```
unitrium, line 314: I have a doubt about this one @VIKTORVAV99 do you know more ?

unitrium: Indeed EUR

unitrium:
> This don't match what we are fetching.
> 
> DK, SE, NO are all EUR as well.
> 
> I don't have time to double check all of these manually so either you need a different generation script I can review or we need another solution.

@VIKTORVAV99 you mean they are traded in EUR ?

unitrium:
Will update DK, NO and SE.

unitrium:
@VIKTORVAV99 thanks for flagging I updated DK, NO, SE.
 I won't commit the script as it's just something we want to use once to update those files fast.

config/zones/BG.yaml
```diff
@@ -524,3 +524,4 @@ sources:
 timezone: Europe/Sofia
 zone_name: Bulgaria
 country_name: Bulgaria
+currency: BGN
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/CH.yaml
```diff
@@ -509,3 +509,4 @@ sources:
 timezone: Europe/Zurich
 zone_name: Switzerland
 country_name: Switzerland
+currency: CHF
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/CZ.yaml
```diff
@@ -518,3 +518,4 @@ sources:
 timezone: Europe/Prague
 zone_name: Czechia
 country_name: Czechia
+currency: CZK
```
BastienBigue:
```suggestion
currency: EUR
```

unitrium: Hum is CZ part of Nordpool and therefore traded in EUR ?

config/zones/GB-NIR.yaml
```diff
@@ -334,3 +334,4 @@ sources:
 timezone: Europe/London
 zone_name: Northern Ireland
 country_name: Northern Ireland
+currency: GBP
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/HR.yaml
```diff
@@ -508,3 +508,4 @@ sources:
 timezone: Europe/Belgrade
 zone_name: Croatia
 country_name: Croatia
+currency: HRK
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/HU.yaml
```diff
@@ -571,3 +571,4 @@ sources:
 timezone: Europe/Budapest
 zone_name: Hungary
 country_name: Hungary
+currency: HUF
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/CZ.yaml
unitrium: Just checked it is indeed

VIKTORVAV99:
> > This don't match what we are fetching.
> > DK, SE, NO are all EUR as well.
> > I don't have time to double check all of these manually so either you need a different generation script I can review or we need another solution.
> 
> @VIKTORVAV99 you mean they are traded in EUR ?

Yes NORDPOOL is dual pricing them in bot EUR and local currencies and we are parsing the EUR version to keep it consistent.

unitrium:
Actually it seems like we also need to have GB in EUR, as it's in EUR on Nordpool

BastienBigue:
Currently manually going through all those for which we have data

BastienBigue:
latest GB price is GBP (they have a separate nordpool market)

unitrium:
Thanks @BastienBigue I'm also going through them manually now

BastienBigue:
ME has not been updated, we have EUR data

config/zones/MK.yaml
```diff
@@ -306,3 +306,4 @@ sources:
 timezone: Europe/Skopje
 zone_name: North Macedonia
 country_name: North Macedonia
+currency: MKD
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/PL.yaml
```diff
@@ -601,3 +601,4 @@ sources:
 timezone: Europe/Warsaw
 zone_name: Poland
 country_name: Poland
+currency: PLN
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/RO.yaml
```diff
@@ -551,3 +551,4 @@ sources:
 timezone: Europe/Bucharest
 zone_name: Romania
 country_name: Romania
+currency: RON
```
BastienBigue:
```suggestion
currency: EUR
```

config/zones/RS.yaml
```diff
@@ -398,3 +398,4 @@ sources:
 timezone: Europe/Belgrade
 zone_name: Serbia
 country_name: Serbia
+currency: RSD
```
BastienBigue:
```suggestion
currency: EUR
```

unitrium:
Thanks @BastienBigue for the help!

unitrium:
@VIKTORVAV99 we went together with @BastienBigue  through the currencies. Let us know if you have further objections otherwise I will go ahead soon.

config/zones/SJ.yaml
```diff
@@ -3,3 +3,4 @@ region: Europe
 timezone: Arctic/Longyearbyen
 zone_name: Svalbard and Jan Mayen
 country_name: Svalbard and Jan Mayen
+currency: NOK
```
VIKTORVAV99, line 6: Don't know if these even have a electricity market. Should we still add it?

unitrium: I would say let's keep it for consistency, we also added things for Greenland, we can always revert back those.

VIKTORVAV99:
@BastienBigue does this mean we also need to update the `zone_config` DB tables?

BastienBigue:
> @BastienBigue does this mean we also need to update the `zone_config` DB tables?

@unitrium what will retrieve this currency info ? 
If app or API, then it might be a good idea to add it to the zone_config table, yes. If it is a forecasting pipeline that already consumes the static configs, then I think it is not needed to add it to the `zone_config` table.

unitrium:
It will be a pipeline that already reads this static config.

PR #8396, opened 2025-09-30
chore(parsers): refactor JP_KY parser to use event classes

gianantoniopini:
## Issue

#6011

## Description

This PR refactors the JP_KY parser to use event classes, as described in the issue.

Additionally, a mock patch target has been corrected in the parser tests. I discovered this small issue while trying to run the parser tests without network connectivity.

### Double check


electricitymap/contrib/parsers/JP_KY.py
```diff
@@ -111,13 +97,22 @@ def fetch_production(
...
-        data["production"]["solar"] = solar
-        data["production"]["nuclear"] = nuclear
-        data["production"]["unknown"] = unknown
+        production_list.append(
+            zoneKey=zone_key,
+            datetime=dt,
+            production=ProductionMix(
+                biomass=0,
+                coal=0,
+                gas=0,
```
VIKTORVAV99, line 106: Let's keep these for now as I don't really know what the impact of removing them would be but generally we want to avoid hardcoded data.

VIKTORVAV99:
Thanks for the PR!

gianantoniopini:
Thank you!

PR #8413, opened 2025-10-01
fix(parsers): ENTSO-E price parsing

VIKTORVAV99:
## Issue
Fixes: GMM-1075

## Description
Ensures we are filtering for the expected market type.

It also sets up the core parts of the parser and infra for parsing the intraday prices independently when ready.

### Tests
Adds a new snapshot test to test for both day_ahead and intraday, now I realise this don't actually rest anything as we are intercepting all requests and mocking them. But it gives a good visible representation of the issue with ES intraday as it has duplicate datetimes with different values that comes from the different sequences of the data. This is something we need to be able to model and capture internally somehow.


electricitymap/contrib/parsers/tests/__snapshots__/test_ENTSOE.ambr
```diff
@@ -5873,779 +5873,9255 @@
...
+      'zoneKey': 'ES',
+    }),
+    dict({
+      'currency': 'EUR',
+      'datetime': datetime.datetime(2025, 9, 28, 22, 0, tzinfo=datetime.timezone.utc),
+      'price': 77.77,
+      'source': 'entsoe.eu',
+      'sourceType': <EventSourceType.measured: 'measured'>,
+      'zoneKey': 'ES',
+    }),
```
VIKTORVAV99, line 8125: Here are the tricky datapoints...

We need to figure out what to do with these and what the forecast team actually needs.

electricitymap/contrib/parsers/ENTSOE.py
```diff
@@ -62,6 +62,7 @@
 # TODO: Switch this to a string enum when we migrate to Python 3.11
 class EntsoeTypeEnum(str, Enum):
     DAY_AHEAD = "A01"
+    INTRADAY_PRICE = "A07"
     TOTAL = "A05"
     INTRADAY = "A40"
```
PaulRoms, line 67: Why is there 2 intraDay?

electricitymap/contrib/parsers/ENTSOE.py
```diff
@@ -417,7 +418,10 @@ def query_exchange_forecast(
 
 
 def query_price(
-    domain: str, session: Session, target_datetime: datetime | None = None
+    domain: str,
+    session: Session,
+    target_datetime: datetime | None = None,
+    marketType: EntsoeTypeEnum = EntsoeTypeEnum.DAY_AHEAD,
```
PaulRoms, line 424: Any impact expected on other zone using query_price from ENTSOE?

PaulRoms (commented):
Some questions otherwise LGTM

electricitymap/contrib/parsers/ENTSOE.py
VIKTORVAV99: Because ENTSO-E seems to have different kinds based on the process type as well (not just market type). So there are multiple kind of intraday markets.

For now it's fine but if it goes out of control we'll have to clean it up.

electricitymap/contrib/parsers/ENTSOE.py
VIKTORVAV99: Not as far as I can tell, tested with a few zones at random and they still work as expected.

PaulRoms: LGTM then

electricitymap/contrib/parsers/ENTSOE.py
PaulRoms: sounds good to me

BastienBigue:
We'll clean the existing data, right ?

electricitymap/contrib/parsers/ENTSOE.py
```diff
@@ -1083,6 +1091,40 @@ def fetch_price(
     return parse_prices(raw_price_data, zone_key, logger).to_list()
 
 
+@refetch_frequency(timedelta(hours=DEFAULT_LOOKBACK_HOURS_REALTIME))
+def fetch_price_intraday(
```
BastienBigue, line 1096: Could we add a comment suggesting to not use this, or even comment this code so that we don't end up running it and fetch intraday price that would end up in the day ahead price table ?

VIKTORVAV99:
> We'll clean the existing data, right ?

Yes once this is merged and deployed I'll refetch all historical data which will overwrite the current faulty data if they differ.

After that I hope we can use Samos new project to reprocess it all!

electricitymap/contrib/parsers/ENTSOE.py
VIKTORVAV99: Will do :)

BastienBigue:
I'm just afraid that if we simply refetch, it'll be hard to know whether a data point has been refetched or not, which interval is actual day ahead, what interval is intraday

VIKTORVAV99:
> > We'll clean the existing data, right ?
> 
> Yes once this is merged and deployed I'll refetch all historical data which will overwrite the current faulty data if they differ.



> I'm just afraid that if we simply refetch, it'll be hard to know whether a data point has been refetched or not, which interval is actual day ahead, what interval is intraday

All of them should just say updated at (current date).

The only case where this would not happen is if the price is the same between the day_ahead and intraday market.

Otherwise we can add a new `EventSourceType` called `DAY_AHEAD` and use that to filter the data and check what was updated. But this should be used in all day ahead parsers, not just ENTSO-E price.

EDIT:
Ah wait, we don't pass this along to the DB...

BastienBigue:
Yeah, all of those that are updated will have the new updated_at, those that are not will stay here we'll mix intraday and day ahead

VIKTORVAV99:
> Yeah, all of those that are updated will have the new updated_at, those that are not will stay here we'll mix intraday and day ahead

But does that matter if the value is the same?

That should be the only reason it's not updated at right? Or I suppose the parser can fail as well.

BastienBigue:
> Yeah, all of those that are updated will have the new updated_at, those that are not will stay here we'll mix intraday and day ahead

> But does that matter if the value is the same?

> That should be the only reason it's not updated at right? Or I suppose the parser can fail as well.

Or if some datetimes are not available in the source.

VIKTORVAV99:
Unless we want to delete all data first I think we will have to live with that?

BastienBigue:
I would argue for backuping all the data we have for Spain. 
Also, if granularity of intraday is higher than granularity of day ahead, it will mess it up.

BastienBigue:
Or we can refetch, and see what the situation is after refetching

VIKTORVAV99:
> I would argue for backuping all the data we have for Spain. Also, if granularity of intraday is higher than granularity of day ahead, it will mess it up.

The granularity appears to be the same and switched to 15 min at the same time so I think that is okay.

But let's try a refetch and see what changes. I'm not even sure we have been parsing Intraday for all this time, I think it was added to the response after we started fetching.

PR #8414, opened 2025-10-02
update linguist-vendored path

VIKTORVAV99:
## Description

Since we changed the location of the mocks we should update this path as well.


ronan-mch (approved):
❤️

PR #8420, opened 2025-10-08
chore(infra): Add new target with only data-centers file

madsnedergaard:
## Description

We're pulling in too much in the API, when we only really need this for the data center locations.

### DIFF
--- a/config/Earthfile
+++ b/config/Earthfile
@@ -7,5 +7,9 @@ src-files:
   COPY . .
   SAVE ARTIFACT .
 
+data-centers-files:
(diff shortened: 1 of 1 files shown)


config/Earthfile
```diff
@@ -7,5 +7,9 @@ src-files:
   COPY . .
   SAVE ARTIFACT .
 
+data-centers-files:
+  COPY data_centers/data_centers.json ./data_centers/data_centers.json
+  SAVE ARTIFACT .
```
ronan-mch, line 12: Should this be addressed somewhere?

madsnedergaard: Yes, will be picked up in the API :)

PR #8422, opened 2025-10-09
Paulromieu/pe parser

PaulRoms:
## Issue
Fixing " TODO: remove this in the future, when this is automatically detected by QA layer"

## Post analysis: 
Running for July 2025: 
[Peru check](https://docs.google.com/spreadsheets/d/1DuQ_PYolBn9RSH2i_fqe5MRXtZge6PRkY7-HgG5K-gI/edit?gid=864986181#gid=864986181
)
Comparison with EMBER: 
- All modes are spot on expect Gas -> I suspect that the reporting source is not including all gas.

## Note:
Website: https://www.coes.org.pe/Portal/portalinformacion/generacion
- The bottom graph we are parsing is in MW and not MWh as display on the website. 
- There is a gas discrepancy between the top graph and the bottom graph.


electricitymap/contrib/parsers/PE.py
```diff
@@ -3,9 +3,9 @@
 from logging import Logger, getLogger
 from zoneinfo import ZoneInfo
 
+import pandas as pd
 from requests import Response, Session
 
-from electricitymap.contrib.lib.models.event_lists import ProductionBreakdownList
```
VIKTORVAV99, line 8: Why did you remove this?

It has a lot of built in sanity checks and such and should be used.

PaulRoms: Then I should put it back! I removed it for the simplicity as I'm not really familiar with it.

VIKTORVAV99: We have a long list where we track parser migration to them here #6011. Basically when all of them are migrated we can remove the .to_list() and use the internal data structure directly in the feeder electricity project!

electricitymap/contrib/parsers/PE.py
```diff
@@ -57,47 +57,63 @@ def fetch_production(
...
-        )
-        production_events = production_events[:-1]
+
+    # Create DataFrame with datetime as index and energy sources as columns
+    df_data = {"datetime": datetime_values}
+
+    for item in production_data:
+        source_name = item["Name"]
+        values = [data["Valor"] for data in item["Data"]]
+        df_data[source_name] = values
```
BastienBigue, line 123: If one datetime is missing for one mode, isn't there a risk of shifting this mode in time ?

electricitymap/contrib/parsers/PE.py
```diff
@@ -52,8 +53,8 @@ def fetch_production(
     response_url: Response = r.post(
         API_ENDPOINT,
         data={
-            "fechaInicial": (target_datetime - timedelta(days=1)).strftime("%d/%m/%Y"),
-            "fechaFinal": target_datetime.strftime("%d/%m/%Y"),
+            "fechaInicial": (target_datetime).strftime("%d/%m/%Y"),
+            "fechaFinal": (target_datetime + timedelta(days=1)).strftime("%d/%m/%Y"),
```
VIKTORVAV99, line 95: Why where these inverted?

Wont this mean we are fetching + day in the future now instead of 1 day in the past for continneous runs?

electricitymap/contrib/parsers/PE.py
PaulRoms: Added to the list!

electricitymap/contrib/parsers/PE.py
PaulRoms: Yes, it is how the website is working.
To get current day 9 of October you need to do: Between 9 and 10.

You can try here https://www.coes.org.pe/Portal/portalinformacion/generacion

You can also test with and without it and see the latest timestamp (I have done it myself)

VIKTORVAV99: Ah!

VIKTORVAV99 (approved):
LGTM! 🎉

BastienBigue (approved):
LGTM ! 🙌

electricitymap/contrib/parsers/PE.py
PaulRoms: There is check above to Validate that all energy sources have the same datetime values

BastienBigue: yes oups sorry, forgot to remove the comment after seeing it

PR #8423, opened 2025-10-09
Add storage to ERCOT 

PaulRoms:
Adding storage to ERCOT. This can not be fetch historically though :/ 

### DIFF
--- a/electricitymap/contrib/parsers/US_ERCOT.py
+++ b/electricitymap/contrib/parsers/US_ERCOT.py
@@ -4,7 +4,7 @@ import gzip
 import json
 import time
 import zipfile
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 from enum import Enum
 from io import BytesIO
 from logging import Logger, getLogger
@@ -28,6 +28,7 @@ from electricitymap.contrib.lib.models.events import (
     StorageMix,
 )
 from electricitymap.contrib.lib.types import ZoneKey
+from electricitymap.contrib.parsers.lib.config import refetch_frequency
 from electricitymap.contrib.parsers.lib.utils import get_token
 from electricitymap.contrib.parsers.lib.validation import validate_exchange
 
@@ -52,10 +53,10 @@ DAYAHEAD_LMP_URL = f"{US_PROXY}/api/public-reports/np4-190-cd/dam_stlmnt_pnt_pri
 REALTIME_LMP_URL = f"{US_PROXY}/api/public-reports/np6-788-cd/lmp_node_zone_hub"
 
 # These links are found at https://www.ercot.com/gridinfo/generation, and should be updated as new data is released
-HISTORICAL_GENERATION_URL = {
-    "2025": f"{US_PROXY}/files/docs/2025/02/08/IntGenbyFuel2025.xlsx?{HOST_PARAMETER}",
-    "all_previous": f"{US_PROXY}/files/docs/2021/03/10/FuelMixReport_PreviousYears.zip?{HOST_PARAMETER}",
-}
+# HISTORICAL_GENERATION_URL = {
+#     "2025": f"{US_PROXY}/files/docs/2025/02/08/IntGenbyFuel2025.xlsx?{HOST_PARAMETER}",
+#     "all_previous": f"{US_PROXY}/files/docs/2021/03/10/FuelMixReport_PreviousYears.zip?{HOST_PARAMETER}",
+# }
 GENERATION_MAPPING = {
     "Coal and Lignite": "coal",
     "Hydro": "hydro",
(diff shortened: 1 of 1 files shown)


VIKTORVAV99:
The parsers location changed recently, they all now live under `electricitymap/contrib/parsers`, not sure all tests and tooling will run otherwise.

electricitymap/contrib/parsers/US_ERCOT.py
```diff
@@ -107,26 +103,25 @@ def get_data(
     return data_json
 
 
-def parse_storage_data(session: Session):
+def parse_storage_data_live(session: Session) -> pd.DataFrame:
+    """
+    Parse the storage data from the ERCOT API. The data is returned in 5 minute intervals, so we need to group by 15 minute intervals and take the mean.
```
VIKTORVAV99: Maybe add why we need to do this?

I suppose it's because the historical data is 15 min?

PaulRoms: It is because the historical is 15min :/. 

But since there is no battery in the historical end point, maybe worth to switch to 5min in live for all. (and then we don't reprocess)

VIKTORVAV99: If historical is 15 min we and the realtime is 5 min we have a problem.<br>Because then the historical data would only overwrite every 3rd datapoint and we can't guarantee we won't reprocess it in the future.

VIKTORVAV99: Otherwise you can forward fill the historical data to be at 5 min resolutions but not sure which is better in this case.

PaulRoms: It will overwrite the battery also :/. 

I'm favor of not allowing reprocessing for the zone in favor of having more granualar data (temporally and per mode). What do you think?

VIKTORVAV99: It should not overwrite battery unless you set battery to 0 no?

It should not overwrite values with none values as far as I know but maybe I'm wrong?

VIKTORVAV99: FYI my new data structure would fix this by coincidence.

electricitymap/contrib/parsers/US_ERCOT.py
```diff
@@ -340,27 +333,19 @@ def fetch_live_exchange(
     return validated_data_points
 
 
-@refetch_frequency(timedelta(days=28))  # A month
```
VIKTORVAV99, line 343: This should be added back, otherwise it will create 1 refetch job for every hour instead of 1 per 28 days like now.

This don't actually stop the refetching.

electricitymap/contrib/parsers/US_ERCOT.py
```diff
@@ -340,27 +333,19 @@ def fetch_live_exchange(
...
-        )
-    else:
-        production = fetch_historical_production(
-            zone_key=zone_key,
-            session=session,
-            target_datetime=target_datetime,
-            logger=logger,
-        )
+    if target_datetime:
+        raise NotImplementedError("This parser is not yet able to parse past dates.")
```
VIKTORVAV99, line 346: I don't like this. If we have a single outage we would loose measured data.

electricitymap/contrib/parsers/US_ERCOT.py
```diff
@@ -54,13 +53,9 @@
 
 # These links are found at https://www.ercot.com/gridinfo/generation, and should be updated as new data is released
 HISTORICAL_GENERATION_URL = {
-    "2024": f"{US_PROXY}/files/docs/2024/02/08/IntGenbyFuel2024.xlsx?{HOST_PARAMETER}",
-    "2023": f"{US_PROXY}/files/docs/2023/02/07/IntGenbyFuel2023.xlsx?{HOST_PARAMETER}",
-    "2022": f"{US_PROXY}/files/docs/2022/02/08/IntGenbyFuel2022.xlsx?{HOST_PARAMETER}",
-    "2021": f"{US_PROXY}/files/docs/2021/02/08/IntGenbyFuel2021.xlsx?{HOST_PARAMETER}",
+    "2025": f"{US_PROXY}/files/docs/2025/02/08/IntGenbyFuel2025.xlsx?{HOST_PARAMETER}",
```
VIKTORVAV99: And if you are removing historical parsing this don't make any sense anymore.

electricitymap/contrib/parsers/US_ERCOT.py
PaulRoms: But since feeder is running every 20 min, and the parser is fetching the last 2 days of data, we should be good no?

VIKTORVAV99: That assumes the source is always available. We can't guarantee that.

electricitymap/contrib/parsers/US_ERCOT.py
PaulRoms: Maybe keep it commented in case?

VIKTORVAV99: That works.

electricitymap/contrib/parsers/US_ERCOT.py
PaulRoms: As far as I know, the ProductionBreakdown object is link to a datetime.
If we query historically where this is no battery, the ProductionBreakdown will have NULL for battery which will go to the DB like that. 
So from understanding, the data will be overwritted and delete the battery

VIKTORVAV99: So we really need to save these individually in the future if we don't want this to happen?

electricitymap/contrib/parsers/US_ERCOT.py
PaulRoms: Added back

electricitymap/contrib/parsers/US_ERCOT.py
PaulRoms: 100% with you on that!

